PR #279, opened 2025-05-07
Refactor: Enhance @-command, Autocomplete, and Input Stability

allenhutchison:
This PR introduces several improvements and bug fixes to the CLI's input handling, particularly around "@" commands and autocompletion:

Enhancements & Fixes:

   * @-Command Trigger:
       * Adjusted the trigger logic for "@" commands to activate if an "@" is preceded by a space, regardless of the character following it (previously required a non-space character after "@").
       * A lone "@" character in the input (e.g., explain @) is now passed directly to the language model as part of the query, instead of attempting to read files or erroring. Previously, it would either error.

   * Autocomplete & Input:
       * Tab Key for Selection: The Tab key now correctly selects the highlighted suggestion in the autocomplete list. If no suggestion is highlighted by arrow keys, Tab will select the first available suggestion.
       * ESC Key and Enter Focus: Resolved a bug where pressing ESC to dismiss the autocomplete suggestion list would prevent the Enter key from submitting commands afterwards. Input submission is now robust after dismissing
         suggestions.
       * Up-Arrow Navigation: Fixed an issue where pressing the up-arrow key when the autocomplete list first appeared would navigate the command history instead of the suggestion list. Suggestion list navigation now correctly
         takes precedence.
       * Double Submission/UI: Corrected a bug where using an "@" command could sometimes lead to the "read many files" UI or action being triggered twice.

  Overall Impact:

These changes result in a more intuitive, robust, and user-friendly experience when using "@" commands and interacting with the autocomplete suggestions. The input handling is now more predictable, especially in edge cases involving focus and event propagation.

### DIFF
--- a/packages/cli/src/ui/App.tsx
+++ b/packages/cli/src/ui/App.tsx
@@ -96,17 +96,8 @@ export const App = ({ config, settings, cliVersion }: AppProps) => {
 
   const isInputActive = streamingState === StreamingState.Idle && !initError;
 
-  const {
-    query,
-    setQuery,
-    handleSubmit: handleHistorySubmit,
-    inputKey,
-    setInputKey,
-  } = useInputHistory({
-    userMessages,
-    onSubmit: handleFinalSubmit,
-    isActive: isInputActive,
-  });
+  // query and setQuery are now managed by useState here
+  const [query, setQuery] = useState('');
 
   const completion = useCompletion(
     query,
@@ -114,6 +105,22 @@ export const App = ({ config, settings, cliVersion }: AppProps) => {
     isInputActive && (isAtCommand(query) || isSlashCommand(query)),
     slashCommands,
   );
+
+  const {
+    handleSubmit: handleHistorySubmit,
+    inputKey,
+    setInputKey,
+  } = useInputHistory({
+    userMessages,
+    onSubmit: (value) => {
+      // Adapt onSubmit to use the lifted setQuery
+      handleFinalSubmit(value);
+      setQuery(''); // Clear query from the App's state
+    },
+    isActive: isInputActive && !completion.showSuggestions,
+    query,
+    setQuery,
+  });
 
   // --- Render Logic ---
 
--- a/packages/cli/src/ui/components/InputPrompt.tsx
+++ b/packages/cli/src/ui/components/InputPrompt.tsx
@@ -131,8 +131,9 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
             // This onSubmit is for the TextInput component itself.
             // It should only fire if suggestions are NOT showing,
             // as useInput handles Enter when suggestions are visible.
-            if (!showSuggestions && query.trim()) {
-              onSubmit(query.trim());
+            const trimmedQuery = query.trim();
+            if (!showSuggestions && trimmedQuery) {
+              onSubmit(trimmedQuery);
             }
             // If suggestions ARE showing, useInput's Enter handler
             // would have already dealt with it (either completing or submitting).
--- a/packages/cli/src/ui/components/InputPrompt.tsx
+++ b/packages/cli/src/ui/components/InputPrompt.tsx
@@ -39,93 +39,83 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
 }) => {
   const { isFocused } = useFocus({ autoFocus: true });
 
-  const handleAutocomplete = useCallback(() => {
-    if (
-      activeSuggestionIndex < 0 ||
-      activeSuggestionIndex >= suggestions.length
-    ) {
-      return;
-    }
-    const selectedSuggestion = suggestions[activeSuggestionIndex];
-    const trimmedQuery = query.trimStart();
+  const handleAutocomplete = useCallback(
+    (indexToUse: number) => {
+      if (indexToUse < 0 || indexToUse >= suggestions.length) {
+        return;
+      }
+      const selectedSuggestion = suggestions[indexToUse];
+      const trimmedQuery = query.trimStart();
 
-    if (trimmedQuery.startsWith('/')) {
-      // Handle / command completion
-      const slashIndex = query.indexOf('/');
-      const base = query.substring(0, slashIndex + 1);
-      const newValue = base + selectedSuggestion.value;
-      setQuery(newValue);
-    } else {
-      // Handle @ command completion
-      const atIndex = query.lastIndexOf('@');
-      if (atIndex === -1) return;
-
-      // Find the part of the query after the '@'
-      const pathPart = query.substring(atIndex + 1);
-      // Find the last slash within that part
-      const lastSlashIndexInPath = pathPart.lastIndexOf('/');
-
-      let base = '';
-      if (lastSlashIndexInPath === -1) {
-        // No slash after '@', replace everything after '@'
-        base = query.substring(0, atIndex + 1);
+      if (trimmedQuery.startsWith('/')) {
+        // Handle / command completion
+        const slashIndex = query.indexOf('/');
+        const base = query.substring(0, slashIndex + 1);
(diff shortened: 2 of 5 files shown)


packages/cli/src/ui/components/InputPrompt.tsx
```diff
@@ -39,93 +39,83 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
...
-        handled &&
-        showSuggestions &&
-        (key.upArrow || key.downArrow || key.tab || key.escape || key.return)
-      ) {
-        // No explicit preventDefault needed, handled flag stops further processing
-      }
+      // Enter key when suggestions are NOT showing is handled by TextInput's onSubmit prop below
     },
-    { isActive: isFocused },
+    { isActive: true },
```
NTaylorMullen, line 118: Why always true vs. isFocused?

packages/cli/src/ui/components/InputPrompt.tsx
```diff
@@ -138,7 +128,14 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
           onChange={setQuery}
           placeholder="Enter your message or use tools (e.g., @src/file.txt)..."
           onSubmit={() => {
-            /* onSubmit is handled by useInput hook above */
+            // This onSubmit is for the TextInput component itself.
+            // It should only fire if suggestions are NOT showing,
+            // as useInput handles Enter when suggestions are visible.
+            if (!showSuggestions && query.trim()) {
```
NTaylorMullen: nit: should trim once and re-use

packages/cli/src/ui/hooks/useInputHistory.ts
```diff
@@ -4,13 +4,15 @@
...
 
-import { useState, useCallback } from 'react';
+import { useCallback, useState } from 'react';
 import { useInput } from 'ink';
 
 interface UseInputHistoryProps {
   userMessages: readonly string[];
   onSubmit: (value: string) => void;
   isActive: boolean;
+  query: string; // Added
```
NTaylorMullen:
```suggestion
  query: string;
```

packages/cli/src/ui/hooks/useInputHistory.ts
```diff
@@ -41,9 +44,8 @@ export function useInputHistory({
     (value: string) => {
       const trimmedValue = value.trim();
       if (trimmedValue) {
-        onSubmit(trimmedValue);
+        onSubmit(trimmedValue); // This will call handleFinalSubmit, which then calls setQuery('') from App.tsx
```
NTaylorMullen, line 47:
```suggestion
        onSubmit(trimmedValue);
```

packages/cli/src/ui/App.tsx
```diff
@@ -115,6 +106,22 @@ export const App = ({ config, settings, cliVersion }: AppProps) => {
...
   );
 
+  const {
+    handleSubmit: handleHistorySubmit,
+    inputKey,
+    setInputKey,
+  } = useInputHistory({
+    userMessages,
+    onSubmit: (value) => {
+      // Adapt onSubmit to use the lifted setQuery
```
NTaylorMullen, line 116:
```suggestion
```

packages/cli/src/ui/App.tsx
```diff
@@ -115,6 +106,22 @@ export const App = ({ config, settings, cliVersion }: AppProps) => {
...
+      handleFinalSubmit(value);
+      setQuery(''); // Clear query from the App's state
```
NTaylorMullen, line 118:
```suggestion
      setQuery('');
```

packages/cli/src/ui/App.tsx
```diff
@@ -96,17 +96,8 @@ export const App = ({ config, settings, cliVersion }: AppProps) => {
...
-    setQuery,
-    handleSubmit: handleHistorySubmit,
-    inputKey,
-    setInputKey,
-  } = useInputHistory({
-    userMessages,
-    onSubmit: handleFinalSubmit,
-    isActive: isInputActive,
-  });
+  // query and setQuery are now managed by useState here
```
NTaylorMullen, line 99:
```suggestion
```

packages/cli/src/ui/components/InputPrompt.tsx
```diff
@@ -39,93 +39,83 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
...
         } else if (key.downArrow) {
           navigateDown();
-          handled = true;
-        } else if ((key.tab || key.return) && activeSuggestionIndex >= 0) {
-          handleAutocomplete();
-          handled = true;
+        } else if (key.tab) {
+          if (suggestions.length > 0) {
+            const targetIndex =
+              activeSuggestionIndex === -1 ? 0 : activeSuggestionIndex;
```
NTaylorMullen, line 99: Is there a way to make the UI also reflect that the first entry in the list is always auto-selected?

NTaylorMullen (approved):
!! Played around with it locally and it's wayyyyy better. So many QoL fixes.

Left some comments then also filed one more improvement that came to mind: https://b.corp.google.com/issues/416292478

packages/cli/src/ui/components/InputPrompt.tsx
allenhutchison: Previously, the useInput hook in InputPrompt.tsx was conditionally activated using { isActive: isFocused }. While isFocused (from useFocus) was generally reliable, we encountered a scenario where, after suggestions were displayed and then hidden by pressing ESC, a re-render occurred. It appeared that for the subsequent Enter key press, the useInput hook might not have been considered "active" by Ink's internal mechanisms at the exact moment the Enter event was processed, even if our isFocused check would have been true.

By changing to { isActive: true }, we ensure Ink always calls our useInput callback function when an input event occurs for the component. Then, as the very first step inside our callback, we explicitly check if (!isFocused) return;. This shifts the responsibility of gating the input handling from Ink's isActive prop (which seemed to have a subtle timing or state issue in this specific re-render scenario) to our own explicit check of the isFocused state provided by useFocus. This made the Enter key handling robust even after the ESC-hide-suggestion sequence.

packages/cli/src/ui/hooks/useInputHistory.ts
allenhutchison: I left that comment in intentionally because of the changes to the way enter is being handled. Can remove if you feel like it's not helpful.

packages/cli/src/ui/components/InputPrompt.tsx
allenhutchison: Sure, can do it that way. How about I follow up in a different PR for that because it's going to affect files that aren't in this one.

packages/cli/src/ui/components/InputPrompt.tsx
allenhutchison: fixed.

packages/cli/src/ui/hooks/useInputHistory.ts
allenhutchison: fixed.

packages/cli/src/ui/components/InputPrompt.tsx
NTaylorMullen: Love it!

packages/cli/src/ui/hooks/useInputHistory.ts
NTaylorMullen: Don't feel strongly, feel free to keep!

packages/cli/src/ui/components/InputPrompt.tsx
NTaylorMullen: I see interesting. We'll have to see how this plays out for moments when the UI isn't visible etc. That being said I don't have a better solution off-hand.

PR #280, opened 2025-05-07
fix: use react-jsx for typecheck

KeijiBranshi:
this was causing a bunch of noisy error messages on people's PRs


package.json
```diff
@@ -16,7 +16,7 @@
     "debug": "NODE_ENV=development DEBUG=1 scripts/start.sh",
     "lint:fix": "eslint . --fix",
     "lint": "eslint . --ext .ts,.tsx",
-    "typecheck": "tsc --noEmit --jsx react",
+    "typecheck": "tsc --noEmit --jsx react-jsx",
```
NTaylorMullen, line 19: oh wow lol, good catch

PR #281, opened 2025-05-07
Don't crash on 429/500s.

NTaylorMullen:
- The root of this issue was actually a genai SDK bug that was fixed here: https://critique.corp.google.com/cl/753255997
- Upgrade to latest genai SDK for latest bug fixes (including the above)
- Removed specific 429 handling for uncaught rejections.

Fixes https://b.corp.google.com/issues/413760164

### DIFF
--- a/packages/server/src/core/client.ts
+++ b/packages/server/src/core/client.ts
@@ -133,39 +133,28 @@ export class GeminiClient {
   ): AsyncGenerator<ServerGeminiStreamEvent> {
     let turns = 0;
     const availableTools = this.config.getToolRegistry().getAllTools();
-    try {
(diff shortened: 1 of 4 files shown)


packages/server/src/core/client.ts
```diff
@@ -133,39 +133,30 @@ export class GeminiClient {
   ): AsyncGenerator<ServerGeminiStreamEvent> {
     let turns = 0;
     const availableTools = this.config.getToolRegistry().getAllTools();
-    try {
```
NTaylorMullen, line 136: Dropping this try-catch because we were already handling this sort of exception management centrally on the consumign side. Aka it resulted in double logging

PR #284, opened 2025-05-08
Fix bugs from useGeminiStream refactor

thkim1011:
1. Split at the last safe point to maximize output rendered directly to `<Static />`.
2. The initial state of the pendingHistoryItem.text should be empty string.
3. Accidentally used `gemini_content` instead of `gemini` as item type.

image

### DIFF
--- a/packages/cli/src/ui/hooks/useGeminiStream.ts
+++ b/packages/cli/src/ui/hooks/useGeminiStream.ts
@@ -31,7 +31,7 @@ import { isAtCommand } from '../utils/commandUtils.js';
 import { useSlashCommandProcessor } from './slashCommandProcessor.js';
 import { useShellCommandProcessor } from './shellCommandProcessor.js';
 import { handleAtCommand } from './atCommandProcessor.js';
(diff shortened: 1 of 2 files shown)


packages/cli/src/ui/hooks/useGeminiStream.ts
```diff
@@ -211,21 +212,19 @@ export const useGeminiStream = (
...
+              const afterText = geminiMessageBuffer.substring(splitPoint);
+              geminiMessageBuffer = afterText; // Continue accumulating from split point
               addItem(
-                { type: 'gemini_content', text: beforeText },
+                { type: 'gemini', text: beforeText },
                 userMessageTimestamp,
               );
               setPendingHistoryItem({
-                type: 'gemini_content',
+                type: 'gemini',
```
NTaylorMullen, line 223: Ah does this need to be `gemini`? The `gemini_content` bit specifically was intended to not show the little star icon after each split message

thkim1011: ohh that makes more sense. i'll make a quick follow up pr; i think i know how to address this properly.

PR #285, opened 2025-05-08
rename SANDBOX_EXEC_PROFILE as SEATBELT_PROFILE, and fix another accidental rephrasing

olcan:


packages/cli/src/utils/sandbox.ts
```diff
@@ -142,15 +142,15 @@ function entrypoint(workdir: string): string[] {
 
 export async function start_sandbox(sandbox: string) {
   if (sandbox === 'sandbox-exec') {
-    process.env.SANDBOX_EXEC_PROFILE ??= 'minimal';
+    process.env.SEATBELT_PROFILE ??= 'minimal';
```
NTaylorMullen, line 145: Nice change

PR #288, opened 2025-05-08
Make ascii logo simpler

miguelsolorio:
This helps reduce vertical space so that it fits nicer in smaller default terminals and also simplifies the logo name

Before
CleanShot 2025-05-08 at 10 11 00@2x

After
CleanShot 2025-05-08 at 10 09 22@2x


NTaylorMullen:
Love it! Merge away!

NTaylorMullen (approved):

PR #295, opened 2025-05-08
fix(sandbox): set --inspect-brk in production sandbox when env DEBUG is truthy

KeijiBranshi:
CC=@olcan

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,18 +22,20 @@ This repository contains the Gemini Code CLI tool.
 
 ## Building
 
+As with most Node projects, major development scripts can be found in the `package.json`. See that for the full list of commands.
+
 ### Prerequisites:
 
 The build toolchain requires `npm` and `jq` to be installed. You can use the `scripts/setup-dev.sh` script to install these prerequisites.
 
-To build the entire project, including the CLI package, run the following command from the root directory:
+To build the entire project, CLI and Sandbox Container Image (if applicable), run the following command from the root directory:
 
 ```bash
 npm install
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -53,11 +55,35 @@ To debug the CLI application using VS Code:
     ```bash
     npm run debug
     ```
-    This command runs `node --inspect-brk dist/gemini.js` within the `packages/cli` directory, pausing execution until a debugger attaches.
+    This command runs `node --inspect-brk dist/gemini.js` within the `packages/cli` directory, pausing execution until a debugger attaches. You can then open `chrome://inspect` in your Chrome browser to connect to the debugger. Alternatively, you can achieve the same effect by running `DEBUG=1 npm run start`.
 2.  In VS Code, use the "Attach" launch configuration (found in `.vscode/launch.json`). This configuration is set up to attach to the Node.js process listening on port 9229, which is the default port used by `--inspect-brk`.
 
 Alternatively, you can use the "Launch Program" configuration in VS Code if you prefer to launch the currently open file directly, but the "Attach" method is generally recommended for debugging the main CLI entry point.
 
+## Testing with npm link
```
NTaylorMullen:
```suggestion
## Using Gemini Code in other directories
```

NTaylorMullen (approved):

README.md
KeijiBranshi: done (with a slight adjustment). added more clarification that this points to the gemini source code locally on the user's machine.

PR #297, opened 2025-05-08
Add bundling support.

NTaylorMullen:
- This can now be invoked with `npm run bundle`, it creates a `bundle/` folder that has:
  - gemini.js
  - sandbox-macos-minimal.sb
  - sandbox-macos-strict.sb
  - shell.json /cc @olcan this is probably a reason why we may want to move away from extra files for our tools
  - shell.md

- This doesn't include any sort of automation for auto bundling pieces. It's just the root capability which we can weave into other locations.
  - /cc @KeijiBranshi I didn't make the decision to ship this as part of our normal release process for now because I felt there was value in the debugability / visibility of the normal source files.

/cc @teeler Running `npm run bundle` produces a singular folder that could now be checked into third_party; however, we probably want to have more automation around automatically updating and publishing said bundle.

Fixes https://b.corp.google.com/issues/411432723

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,5 @@ dist
 
 # Docker folder to help skip auth refreshes
 .docker
+
+bundle
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,6 +11,7 @@
         "packages/*"
       ],
       "devDependencies": {
+        "esbuild": "^0.25.4",
         "eslint": "^9.24.0",
         "eslint-config-prettier": "^10.1.2",
         "eslint-plugin-import": "^2.31.0",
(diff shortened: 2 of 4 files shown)


KeijiBranshi:
> * /cc @KeijiBranshi I didn't make the decision to ship this as part of our normal release process for now because I felt there was value in the debugability / visibility of the normal source files.

all good. just some notes for when we plan to release it as a bundled app:

- we'll need to update the `main` field in `packages/cli/package.json`. 
- the bundle itself should probably be moved inside the `packages/cli/dist/` or `packages/cli/bundle` folder (npm publishes the package folder, not the entire repo)
- we'll probably want to generate a typings file `.d.ts` along side the bundled `.js`
- if we want to develop/debug on the bundle, then we'll need to output sourcemaps too into the bundle folder

NTaylorMullen:
> > * /cc @KeijiBranshi I didn't make the decision to ship this as part of our normal release process for now because I felt there was value in the debugability / visibility of the normal source files.
> 
> all good. just some notes for when we plan to release it as a bundled app:
> 
> * we'll need to update the `main` field in `packages/cli/package.json`.
> * the bundle itself should probably be moved inside the `packages/cli/dist/` or `packages/cli/bundle` folder (npm publishes the package folder, not the entire repo)
> * we'll probably want to generate a typings file `.d.ts` along side the bundled `.js`
> * if we want to develop/debug on the bundle, then we'll need to output sourcemaps too into the bundle folder

Great callout, going to re-open the bundling support bug and copy & paste your bits

PR #298, opened 2025-05-09
Fix: Prevent CLI from crashing when a configured theme is not found

amirh:
Previously, if a theme specified in the user's settings was not found, the CLI would crash during startup. This was particularly affecting users upgrading from older versions as the "ANSI colors only" theme was renamed to "ANSI".

This commit adds error handling to catch the theme not found error during initial loading and when setting themes later. Instead of crashing, the application now logs a warning, displays an error message in the UI, and opens the theme selection dialog to allow the user to choose a valid theme.


NTaylorMullen (approved):
Wow super thorough. :shipit: !!

PR #302, opened 2025-05-09
Multiline editor

jacob314:
Switch to a multiline editor. First commit is the baseline implementation (thanks @NTaylorMullen).
Second commit is wiring it up to history and autocomplete. Thanks @KeijiBranshi for some pointers on how to try to simplify this logic and avoid cases where this logic is really tangled in some other CLI tools. In particular, I tried to avoid using refs to components.

For discussion:
Having both InputPrompt.tsx and MultilineTextEditor might not be worth the complexity given how coupled they are. This code would be a bit simpler if they are merged together. I think it could depend on whether we think we want to use the MultilineTextEditor for more than just the input prompt.

https://screencast.googleplex.com/cast/NjY3MjMxODAwMzY3NTEzNnw2NDAwMmU3NC1lZA


NTaylorMullen:
For other onlookers it's worth calling out that this is a port from Codex too. @jacob314 would you mind marking the areas / files where review is most needed vs. what's a direct copy n paste

packages/cli/src/ui/components/shared/multiline-editor.tsx
```diff
@@ -0,0 +1,499 @@
...
+ * mock receives `write()` calls.  Ink, however, listens for `readable` and
+ * uses the `read()` method to fetch the buffered chunk.  Bridge the gap by
+ * hooking into `EventEmitter.emit` so that every `data` emission also:
+ *   1.  Buffers the chunk for a subsequent `read()` call, and
+ *   2.  Triggers a `readable` event, matching the contract expected by Ink.
+ */
+
+// Preserve original emit to avoid infinite recursion.
+// eslint‑disable‑next‑line @typescript-eslint/no‑unsafe‑assignment
+const originalEmit = proto['emit'] as (...args: Array<any>) => boolean;
```
KeijiBranshi: nitpick(non-blocking):

```suggestion
const originalEmit = proto['emit'] as EventEmitter['emit'];
```

packages/cli/src/ui/components/shared/multiline-editor.tsx
```diff
@@ -0,0 +1,499 @@
...
+ * failures during unit tests.
+ * ----------------------------------------------------------------------- */
+
+// Cast through `unknown` ➜ `any` to avoid the `TS2352`/`TS4111` complaints
+// when augmenting the prototype with the stubbed `ref`/`unref` methods in the
+// test environment.  Using `any` here is acceptable because we purposefully
+// monkey‑patch internals of Node's `EventEmitter` solely for the benefit of
+// Ink's stdin stub – type‑safety is not a primary concern at this boundary.
+//
+const proto: any = EventEmitter.prototype;
```
KeijiBranshi: question: I'm not sure I understand the reason why this monkey patch had to go all the way down to Node's EventEmitter. Is there a particular reason this could not have been done by wrapping (or even patching) ink's exports?

EventEmitter is foundational to _all_ events in Node, not just stdin/stdout related ones. So bugs here have a _HUGE_ blast radius.

packages/cli/src/ui/components/shared/text-buffer.ts
```diff
@@ -0,0 +1,1008 @@
...
+
+  private clipboard: string | null = null;
+
+  constructor(text = '', initialCursorOffset = 0) {
+    this.lines = text.split('\n');
+    if (this.lines.length === 0) {
+      this.lines = [''];
+    }
+
+    this.setCursorOffset(initialCursorOffset);
```
jacob314, line 117: only change to this file is pulling in this method which was recently added to codex and is needed to keep the cursor offset updated after applying an autocomplete suggestion.

packages/cli/src/ui/App.tsx
```diff
@@ -223,14 +224,16 @@ export const App = ({ config, settings, cliVersion }: AppProps) => {
...
-                onSubmit={handleHistorySubmit}
+                editorState={editorState}
+                setEditorState={setEditorState}
+                onSubmit={inputHistory.handleSubmit}
                 showSuggestions={completion.showSuggestions}
                 suggestions={completion.suggestions}
                 activeSuggestionIndex={completion.activeSuggestionIndex}
-                navigateUp={completion.navigateUp}
-                navigateDown={completion.navigateDown}
+                navigateHistoryUp={inputHistory.navigateUp}
```
jacob314, line 243: open to alternate names here. Just trying to make this less confusing.
navigateHistoryUp and navigateHistoryDown could also be
navigateOutUp and navigateOutDown as conceptually they are when you use the up and down arrows to navigate out of the top or bottom of the multiline edit area.

packages/cli/src/ui/components/InputPrompt.tsx
```diff
@@ -73,32 +89,30 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
...
 
       resetCompletion(); // Hide suggestions after selection
-      setInputKey((k) => k + 1); // Increment key to force re-render and cursor reset
     },
-    [query, setQuery, suggestions, resetCompletion, setInputKey],
+    [query, setQuery, suggestions, resetCompletion, setEditorState],
   );
 
-  useInput(
+  const inputPreprocessor = useCallback(
```
jacob314, line 91: here I diverged from what Codex does as what they do here is too hacky. They pass around a ton of component refs and even require a timeout of 1ms (not 0) to get the timing to work out right.
This is a case where this method would be even simpler if merged directly into multiline-editor. Open to squashing these two components to reduce the complexity. All depends on whether we really want multiline-editor as a reusable piece or a powerful SuperInputPrompt component.

packages/cli/src/ui/components/shared/multiline-editor.tsx
```diff
@@ -0,0 +1,499 @@
...
+  // Keyboard handling.
+  // ---------------------------------------------------------------------------
+
+  useInput(
+    (input, key) => {
+      if (!focus) {
+        return;
+      }
+
+      if (inputPreprocessor?.(input, key) === true) {
```
jacob314, line 120: this line is a delta from the Codex version. Adding the preprocessor avoids a lot of hackery where Codex uses components refs and timeouts to make things work. They even have a buggy state where the cursor positions are wrong and they keep cached cursor positions with a 1ms timeout to make things run ok.

packages/cli/src/ui/components/shared/multiline-editor.tsx
```diff
@@ -0,0 +1,499 @@
...
+      if (
+        process.env['TEXTBUFFER_DEBUG'] === '1' ||
+        process.env['TEXTBUFFER_DEBUG'] === 'true'
+      ) {
+        console.log('[MultilineTextEditor] key event', { input, key });
+      }
+
+      // Up arrow - check for navigation override.
+      if (key.upArrow) {
+        if (buffer.current.getCursor()[0] === 0 && navigateUp) {
```
jacob314: these 3 lines for navigateUp and navigateDown are the other delta between this code and Codex. Doing this logic directly in this component does make this component aware of navigating up out of the component and navigating down past it but at least means we avoid the cursor bugs that Codex appears to have hacked around with timeouts.

packages/cli/src/ui/components/shared/multiline-editor.tsx
```diff
@@ -0,0 +1,499 @@
+/* eslint-disable license-header/header */
```
jacob314: All code except where commented is unchanged from Taylor's original port. If you view by commit in the history you can see the delta that I added to make history navigation robust enough.

packages/cli/src/ui/components/InputPrompt.tsx
```diff
@@ -5,36 +5,45 @@
...
+import { Text, Box, useFocus, Key } from 'ink';
 import { Colors } from '../colors.js';
 import { Suggestion } from './SuggestionsDisplay.js';
+import { MultilineTextEditor } from './shared/multiline-editor.js';
 
 interface InputPromptProps {
   query: string;
   setQuery: React.Dispatch<React.SetStateAction<string>>;
-  inputKey: number;
-  setInputKey: React.Dispatch<React.SetStateAction<number>>;
```
jacob314, line 17: I've combined the key and the cursor position into a state object aligning with cursor. open to better solutions.

packages/cli/src/ui/hooks/useInputHistory.ts
```diff
@@ -51,64 +52,75 @@ export function useInputHistory({
...
-          setQuery(originalQueryBeforeNav);
-        } else {
-          const newValue = userMessages[userMessages.length - 1 - nextIndex];
-          setQuery(newValue);
-        }
-        setInputKey((k) => k + 1);
-        didNavigate = true;
-      } else {
-        if (historyIndex !== -1 && !didNavigate) {
-          if (input || key.backspace || key.delete) {
```
jacob314, line 98: I've removed the checks from line 97 as they don't really make sense in the new scheme. If someone understands what they were trying to do happy to add them back. For now, history navigation seems ok without them.

packages/cli/src/ui/components/shared/text-buffer.ts
```diff
@@ -0,0 +1,1008 @@
+/**
```
jacob314, line 1: this file does not need a careful review. It is the same as Taylor's version modulo maybe a helper method.

jacob314 (commented):
Added comments to clarify what portions of the code need a careful review and what portions do not.

packages/cli/src/ui/components/shared/multiline-editor.tsx
```diff
@@ -0,0 +1,550 @@
...
+    inputKey,
+    setInputKey,
+    showSuggestions,
+    suggestions,
+    activeSuggestionIndex,
+    navigateUp,
+    navigateDown,
+    resetCompletion,
+  }: MultilineTextEditorProps,
+  ref: React.Ref<MultilineTextEditorHandle | null>,
```
KeijiBranshi: suggestion: like with other reactive programming frameworks, imperative getters are generally discouraged. I think more "react"-y approach here would be to add a prop: `onHandle: (handle: MultilineTextEditorHandle) => void` (maybe rename to something more representative like `onCursorState(s: CursorState) => void`). The parent component should be reacting to changes emitted by the child component, not providing it with a mutable ref to modify on the parent's behalf. 

iiuc, the cursor position at some point has to reach a rendering statement. does that mean this pitfall is being violated? https://screenshot.googleplex.com/AtN5xj6R3vVweR2

packages/cli/src/ui/components/shared/multiline-editor.tsx
```diff
@@ -0,0 +1,499 @@
...
+    onSubmit,
+    focus = true,
+    onChange,
+    initialCursorOffset,
+    widthUsedByParent,
+    navigateUp,
+    navigateDown,
+    inputPreprocessor,
+  }: MultilineTextEditorProps,
+  ref: React.Ref<MultilineTextEditorHandle | null>,
```
KeijiBranshi: iiuc, the cursor position at some point has to reach a rendering statement. does that mean this pitfall is being violated? https://screenshot.googleplex.com/AtN5xj6R3vVweR2

suggestion: like with other reactive programming frameworks, imperative getters are generally discouraged and refs like this have a way of coupling components. I think more "react"-y approach here would be to add a prop: `onHandle: (handle: MultilineTextEditorHandle) => void` (maybe rename to something more representative like `onCursorChange(s: Cursor) => void`). In other words, the parent component should be react to changes emitted by the child component:

```
const Parent = () => {
  const [row, setRow] = useState(0);
  
  // Note, this callback will be reinstantiated every time `row` changes
  const someCallback = useCallback(() => console.log(row), [row]);
  
  // Note, by having all state managed by `useState`, React can then
  // handle propagating those changes to other places that need to
  // stay up to date with the latest value (in this case, `someCallback`)
  return <MultilineTextEditorInner onCursor={(c) => setRow(c.row)} />
}
```

packages/cli/src/ui/components/shared/multiline-editor.tsx
```diff
@@ -0,0 +1,499 @@
...
+      getText: () => buffer.current.getText(),
+      moveCursorToEnd: () => {
+        buffer.current.move('home');
+        const lines = buffer.current.getText().split('\n');
+        for (let i = 0; i < lines.length - 1; i++) {
+          buffer.current.move('down');
+        }
+        buffer.current.move('end');
+        // Force a re-render
+        setVersion((v) => v + 1);
```
KeijiBranshi: suggestion: it's a code smell to force re-renders. at first glance, I don't think `buffer` needs to be a ref. It can just be a `useState`.


unsolicited details that may just confuse us more:

If we're forcing renders, then it means we're trying to do React's job for it. React manages the component rendering lifecycle by observing changes to props and state, specifically state managed by the `useState` hook. This is why it's important to manage most if not _all_ state via `useState` hook. React automatically propagates those changes out to other components and derived values.

packages/cli/src/ui/components/shared/multiline-editor.tsx
```diff
@@ -0,0 +1,499 @@
...
+    height,
+    width: effectiveWidth,
+  });
+  const [cursorRow, cursorCol] = buffer.current.getCursor();
+  const scrollRow = (buffer.current as any).scrollRow as number;
+  const scrollCol = (buffer.current as any).scrollCol as number;
+
+  return (
+    <Box key={version} flexDirection="column">
+      {buffer.current.getText().length === 0 && placeholder ? (
```
KeijiBranshi: suggestion: another reason buffer should be made into a `useState`: https://screenshot.googleplex.com/AtN5xj6R3vVweR2

packages/cli/src/ui/hooks/useInputHistory.ts
```diff
@@ -4,23 +4,24 @@
...
   setQuery: React.Dispatch<React.SetStateAction<string>>;
+  setEditorState: React.Dispatch<React.SetStateAction<EditorState>>;
 }
 
 interface UseInputHistoryReturn {
   query: string;
   setQuery: React.Dispatch<React.SetStateAction<string>>;
   handleSubmit: (value: string) => void;
-  inputKey: number;
-  setInputKey: React.Dispatch<React.SetStateAction<number>>;
```
KeijiBranshi, line 23: thought: nice, tbh this felt weird to me - having the current `inputKey` be passed as a prop. key presses are usually just short lived events that result in updated state elsewhere. so it's odd that we were maintaining that and passing it around to other components/hooks.

packages/cli/src/ui/hooks/useInputHistory.ts
```diff
@@ -51,64 +52,75 @@ export function useInputHistory({
...
+  const setQueryAndMoveCursor = useCallback(
+    (value: string) => {
+      setQuery(value);
+      setEditorState((s) => ({
+        key: s.key + 1,
+        initialCursorOffset: value.length,
+      }));
+    },
+    [setQuery, setEditorState],
+  );
```
KeijiBranshi: nitpick: I think if we just made `setQueryAndMoveCursor` a prop passed by the parent, then it would remove the need for both `setQuery` and `setEditorState` in this hook.

packages/cli/src/ui/hooks/useInputHistory.ts
```diff
@@ -29,11 +30,11 @@ export function useInputHistory({
   isActive,
   query,
   setQuery,
+  setEditorState,
```
KeijiBranshi: supernitpick: I _think_ React devs typically name callbacks passed as props 
`on{EventName}`, reserving `set{Thing}` for just `useState` return value:

```
const [state, setState] = useState()
```

The reason being that `set` implies some sort of state change will happen. But child components should _not_ be aware of their parent's state. The only thing communicated between Parent and Child components are unidirectional "props" (passed from parent to child). it's a pedantic different, but might give a little insight into the react philosophy.

packages/cli/src/ui/components/InputPrompt.tsx
```diff
@@ -5,36 +5,45 @@
...
+  editorState: EditorState;
+  setEditorState: React.Dispatch<React.SetStateAction<EditorState>>;
```
KeijiBranshi: nitpick-suggestion: as mentioned in the previous comment, react child components generally shouldn't be aware of their parents _state_. But making the type here an explicit `React.Dispatch<React.SetStateAction<T>>` breaks that contract a little (albeit superficially since types aren't real). 

```suggestion
  onEditorChange: (editorState: EditorState) => void;
```

KeijiBranshi (approved):
first pass. looks solid, but I think there's still a few ref use-cases that could/should be refactored to react-managed state. 

Might take some time to read through text-buffer.ts, so please don't block on me for that.

packages/cli/src/ui/components/shared/multiline-editor.tsx
jacob314: removed this code for now as I agree it is risky and seems like a hack to just make tests the way Codex was writing them work.

packages/cli/src/ui/components/shared/multiline-editor.tsx
jacob314: removed all this crap as it is for how Codex was doing tests. We can add something back later if we hit issues when we test this.

packages/cli/src/ui/components/shared/multiline-editor.tsx
jacob314: OMG. your comments were enough to get Gemini to do the right thing to completely refactor this code to use what I think are the correct functional reactive pattern.

packages/cli/src/ui/components/InputPrompt.tsx
NTaylorMullen: I'm ok having it as a super input prompt if there's significant clarity gains we can get from that. We didn't end up using the InputPrompt in multiple places fwiw

packages/cli/src/ui/hooks/useTerminalSize.ts
```diff
@@ -0,0 +1,32 @@
...
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { useEffect, useState } from 'react';
+
+const TERMINAL_PADDING_X = 8;
+
+export function useTerminalSize(): { columns: number; rows: number } {
```
NTaylorMullen, line 11: We'll probably end up using this more too.

packages/cli/src/ui/App.tsx
```diff
@@ -99,6 +99,10 @@ export const App = ({ config, settings, cliVersion }: AppProps) => {
 
   // query and setQuery are now managed by useState here
   const [query, setQuery] = useState('');
+  const [editorState, setEditorState] = useState<EditorState>({
```
NTaylorMullen, line 101: Ya the way we pass editor state around makes me sad.  If we had a super input prompt then history / submission could be handled by multi line text editor itself and it could track "last submissions" instead of plumbing all this through the whole UI

NTaylorMullen (approved):
Your screencast especially made me excited for this. My biggest comment around cleanliness is around us potentially wanting to refactor logic so that history management is more isolated to the text editor itself BUT that's not blocking.

packages/cli/src/ui/components/shared/multiline-editor.tsx
jacob314: removed and switched to useState on the TextBuffer. Not perfect but a step in the right direction.

packages/cli/src/ui/hooks/useInputHistory.ts
jacob314: this is the setter from  a
const [editorState, setEditorState] = useState call. Calling this does change the state so I'd worrying calling in onEditorState would be confusing. Apologies this is confusing as the current editorState wasn't needed in this context.

packages/cli/src/ui/App.tsx
jacob314: Sounds good. I'm submitting as is then refactoring as super input.

packages/cli/src/ui/components/InputPrompt.tsx
jacob314: done.

PR #303, opened 2025-05-09
feat: Improve theme not found handling

NTaylorMullen:
Modify  to return a boolean instead of throwing an error when a theme is not found. Update CLI startup and  hook to handle the boolean return value for more graceful error handling.


jacob314:
lgtm

PR #304, opened 2025-05-09
sandbox/seatbelt-aware system prompt, support for custom seatbelt profiles under project settings

olcan:


packages/server/src/core/prompts.ts
```diff
@@ -85,6 +85,25 @@ Rigorously adhere to existing project conventions when reading or modifying code
 -   **Help Command:** The user can use '/help' to display help information.
 -   **Feedback:** Direct feedback to ${contactEmail}.
 
+${(function () {
```
NTaylorMullen, line 88: I really like that we're conditioning what we display based off of the active experience

packages/server/src/core/prompts.ts
```diff
@@ -85,6 +85,25 @@ Rigorously adhere to existing project conventions when reading or modifying code
...
+  if (process.env.SANDBOX === 'sandbox-exec') {
```
NTaylorMullen, line 89: Should we have a config way to represent the type of sandboxign we're using?

packages/server/src/core/prompts.ts
```diff
@@ -85,6 +85,25 @@ Rigorously adhere to existing project conventions when reading or modifying code
...
+`;
+  } else if (process.env.SANDBOX) {
+    return `
+# Sandbox
+You are running in a sandbox container with limited access to files outside the project directory or system temp directory, and with limited access to host system resources such as ports. If you encounter failures that could be due to sandboxing (e.g. if a command fails with 'Operation not permitted' or similar error), when you report the error to the user, also explain why you think it could be due to sandboxing, and how the user may need to adjust their sandbox configuration.
+`;
+  } else {
+    return `
+# Outside of Sandbox
+You are running outside of a sandbox container, directly on the user's system. For critical commands that are particularly likely to modify the user's system outside of the project directory or system temp directory, as you explain the command to the user (per the Explain Critical Commands rule above), also remind the user to consider enabling sandboxing.
```
NTaylorMullen, line 102: We'll want a temp link in here eventually on guidance on how a user can enable sandboxign

olcan:
Maybe... depends on what you mean. SANDBOX is a pretty special env var (not
to be confused e.g. by the sandbox setting that we already have), but open
to ideas.

NTaylorMullen:
When we start supporting command line passing I imagine the config will be the source of truth for if / how sandboxing is enabled. That will also unify the interpretation of hte various settings passed through the sandboxing

olcan:
Yeah, SANDBOX specifically is a bit complicated. It represents a
combination of system environment and user settings, and it is naturally
passed through sandboxing commands and baked into the sandbox image. We
could copy it into the Config but calling it a "source of truth" may be a
stretch. Anyway I'm happy to help eliminate this or other env vars if/when
we want to do that. I'd like to understand better how we want/need to use a
central config.

PR #309, opened 2025-05-10
Adding open file as a tool

cperry-goog:
Opens file with "open" (mac) or "start" (windows) in the default application for that file type. This lets you edit files in text editors, or it just opened up this webpage for me which was really nice.

### DIFF
--- a/packages/server/src/tools/open-file.ts
+++ b/packages/server/src/tools/open-file.ts
@@ -1,0 +1,172 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import fs from 'fs';
+import path from 'path';
+import { Config } from '../config/config.js';
+import {
+  BaseTool,
+  ToolResult,
+  ToolCallConfirmationDetails,
+  ToolExecuteConfirmationDetails,
+  ToolConfirmationOutcome,
+} from './tools.js';
+import { SchemaValidator } from '../utils/schemaValidator.js';
+import { spawn } from 'child_process';
+
+export interface OpenFileToolParams {
+  filePath: string;
+}
+
+export class OpenFileTool extends BaseTool<OpenFileToolParams, ToolResult> {
+  static Name: string = 'open_file';
+
+  constructor(private readonly config: Config) {
+    const toolDisplayName = 'Open File';
+    const toolDescription =
+      'Opens the specified file in the default local text editor.';
+    const toolParameterSchema = {
+      type: 'object',
+      properties: {
+        filePath: {
+          type: 'string',
+          description: 'The absolute path to the file to open.',
+        },
+      },
+      required: ['filePath'],
+    };
+    super(
+      OpenFileTool.Name,
+      toolDisplayName,
+      toolDescription,
+      toolParameterSchema,
+    );
+  }
+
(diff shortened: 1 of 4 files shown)


packages/server/src/tools/open-file.ts
```diff
@@ -0,0 +1,172 @@
...
+        },
+      },
+      required: ['filePath'],
+    };
+    super(
+      OpenFileTool.Name,
+      toolDisplayName,
+      toolDescription,
+      toolParameterSchema,
+    );
```
NTaylorMullen, line 46: Ask GC to ensure that filepaths are within the root directory like in https://github.com/google-gemini/gemini-code/blob/main/packages/server/src/tools/read-file.ts#L68

packages/server/src/tools/open-file.ts
```diff
@@ -0,0 +1,172 @@
...
+  filePath: string;
+}
+
+export class OpenFileTool extends BaseTool<OpenFileToolParams, ToolResult> {
+  static Name: string = 'open_file';
+
+  constructor(private readonly config: Config) {
+    const toolDisplayName = 'Open File';
+    const toolDescription =
+      'Opens the specified file in the default local text editor.';
```
NTaylorMullen, line 30: Lets update this to have something along the lines of "Only when the user asks you..." otherwise the model might overly use this.

packages/server/src/tools/open-file.ts
```diff
@@ -0,0 +1,172 @@
...
+} from './tools.js';
+import { SchemaValidator } from '../utils/schemaValidator.js';
+import { spawn } from 'child_process';
+
+export interface OpenFileToolParams {
+  filePath: string;
+}
+
+export class OpenFileTool extends BaseTool<OpenFileToolParams, ToolResult> {
+  static Name: string = 'open_file';
```
NTaylorMullen, line 25:
```suggestion
  static Name: string = 'open_file_for_user';
```

Intentionally trying to make it a bit more vague so it's more of a user requested action.

packages/server/src/tools/open-file.ts
```diff
@@ -0,0 +1,172 @@
...
+
+  async execute(params: OpenFileToolParams): Promise<ToolResult> {
+    const validationError = this.validateToolParams(params);
+    if (validationError) {
+      return {
+        llmContent: [
+          `Command rejected: open "${params.filePath}"`, // Use quotes for consistency
+          `Reason: ${validationError}`,
+        ].join('\n'),
+        returnDisplay: `Error: ${validationError}`,
```
NTaylorMullen, line 109:
```suggestion
        returnDisplay: validationError,
```

packages/server/src/tools/open-file.ts
```diff
@@ -0,0 +1,172 @@
...
+      };
+    }
+
+    return new Promise((resolve) => {
+      let command: string;
+      let args: string[];
+
+      if (process.platform === 'win32') {
+        command = 'cmd'; // Using 'cmd' to ensure 'start' is available
```
NTaylorMullen, line 118: I'd cleanpu some of these informative comments

NTaylorMullen (commented):
Another route to this change is you may be able to get away with this tech by adjusting the system prompt OR shell tools description as food for thought.

PR #310, opened 2025-05-10
Allow users to cancel out of theme selector

miguelsolorio:
This allows users to hit `Escape` while in the theme selector and goes back to the main page with a cleared input

https://github.com/user-attachments/assets/9c75fa54-cf61-4e13-9e17-9dcabfa5d746


NTaylorMullen:
So simple, love it. :shipit:

NTaylorMullen (approved):

PR #312, opened 2025-05-10
Enable tools to cancel active execution.

NTaylorMullen:
- Plumbed abort signals through to tools
- Updated the shell tool to properly cancel active requests by killing the entire child process tree of the underlying shell process and then report that the shell itself was canceled.

Fixes https://b.corp.google.com/issues/416829935

### DIFF
--- a/packages/cli/src/ui/hooks/useGeminiStream.ts
+++ b/packages/cli/src/ui/hooks/useGeminiStream.ts
@@ -89,7 +89,7 @@ export const useGeminiStream = (
   }, [config, addItem]);
 
   useInput((_input, key) => {
-    if (streamingState === StreamingState.Responding && key.escape) {
+    if (streamingState !== StreamingState.Idle && key.escape) {
       abortControllerRef.current?.abort();
     }
   });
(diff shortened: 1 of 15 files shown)


packages/cli/src/ui/hooks/useGeminiStream.ts
```diff
@@ -89,7 +89,7 @@ export const useGeminiStream = (
   }, [config, addItem]);
 
   useInput((_input, key) => {
-    if (streamingState === StreamingState.Responding && key.escape) {
+    if (streamingState !== StreamingState.Idle && key.escape) {
```
NTaylorMullen, line 92: Tool execution can happen during confirmation states

packages/server/src/tools/shell.ts
```diff
@@ -174,18 +177,38 @@ export class ShellTool extends BaseTool<ShellToolParams, ToolResult> {
...
     let code: number | null = null;
-    let signal: NodeJS.Signals | null = null;
-    shell.on(
-      'close',
-      (_code: number | null, _signal: NodeJS.Signals | null) => {
-        code = _code;
-        signal = _signal;
-      },
-    );
+    let processSignal: NodeJS.Signals | null = null;
```
NTaylorMullen, line 180: /cc @olcan I added some kill logic

olcan:
nice, exciting to have cancellation. btw just noticed this change. lmk if there are issues, esp. w/ shell tool, and i can help investigate

PR #317, opened 2025-05-11
Workaround model bug where it returns invalid history items.

NTaylorMullen:
- Currently there's a bug in the API (or SDK?) where the SDK endpoint will commonly fail with:

```
Error: Failed to generate JSON content: got status: 400 Bad Request. {"error":{"code":400,"message":"* GenerateContentRequest.contents[5].parts: contents.parts must not be empty.\n","status":"INVALID_ARGUMENT"}}
```

- At times the model will respond with an empty parts list where if we send that back up to the API endpoint it explodes with the above. Using a curated history seems like a total hack around this prolbem, and even in the SDK (i'm following up on this), BUT helps mitigate this issue.

### DIFF
--- a/packages/server/src/utils/nextSpeakerChecker.ts
+++ b/packages/server/src/utils/nextSpeakerChecker.ts
@@ -60,7 +60,12 @@ export async function checkNextSpeaker(
   chat: Chat,
   geminiClient: GeminiClient,
 ): Promise<NextSpeakerResponse | null> {
-  const history = await chat.getHistory();
+  // We need to capture the curated history because there are many moments when the model will return invalid turns
+  // that when passed back up to the endpoint will break subsequent calls. An example of this is when the model decides
+  // to respond with an empty part collection if you were to send that message back to the server it will respond with
+  // a 400 indicating that model part collections MUST have content.
+  const history = await chat.getHistory(/* curated */ true);
+
   // Ensure there's a model response to analyze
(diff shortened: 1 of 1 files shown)


olcan:
"At times the model will respond with an empty parts list"... do we know when this happens? We should probably consider these errors and take them into the retry loop just like other errors. I've never seen this happen consistently except when resuming model messages that were already complete.

NTaylorMullen:
> olcan

That's a good idea; however, I also don't necessarily think an empty parts list is an issue either (not sure why the API doesn't like it). For instance, if the model think it doesn't need to respond it shouldn't be required to actually say something. Emitting the stop token should be sufficient.

Therefore, how do we differentiate between "the model is done" and empty parts list? There's a chance that we'd loop forever

olcan:
I've always considered an empty response an error in my uses of the API and
have never seen it happen naturally so I've been assuming models were not
allowed to stop on the first token after a user message. I don't know that
for a fact, but also feel like I would have seen it if it did happen
naturally. In which cases do you think the model doesn't need to respond to
a user message?

NTaylorMullen:
> In which cases do you think the model doesn't need to respond to
a user message?

An example that I've seen is when the model says "last thing I need to do is XYZ" -> "tool call XYZ" -> "no response"

olcan:
hmm ok admittedly i haven't done a lot of tool calling with latest models
so that's totally plausible

PR #321, opened 2025-05-12
system prompt for working with git

olcan:
- drop the odd "NEVER Commit" instruction that can confuse the model
- add a whole section on git if the project is managed by a git repository
- add an instruction about interactive shell commands (like git rebase -i but also more generally)

### DIFF
--- a/packages/server/src/core/prompts.ts
+++ b/packages/server/src/core/prompts.ts
@@ -12,6 +12,7 @@ import { ReadFileTool } from '../tools/read-file.js';
 import { ReadManyFilesTool } from '../tools/read-many-files.js';
 import { ShellTool } from '../tools/shell.js';
 import { WriteFileTool } from '../tools/write-file.js';
+import { execSync } from 'node:child_process';
 
 const contactEmail = 'gemini-code-dev@google.com';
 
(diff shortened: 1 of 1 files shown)


packages/server/src/core/prompts.ts
```diff
@@ -80,6 +80,7 @@ Rigorously adhere to existing project conventions when reading or modifying code
 -   **Parallelism:** Execute multiple independent tool calls in parallel when feasible (i.e. searching the codebase).
 -   **Command Execution:** Use the '${ShellTool.Name}' tool for running shell commands, remembering the safety rule to explain modifying commands first.
 -   **Background Processes:** Use background processes (via \`&\`) for commands that are unlikely to stop on their own, e.g. \`node server.js &\`. If unsure, ask the user.
+-   **Interactive Commands:** Try to avoid shell commands that are likely to require user interaction (e.g. \`git rebase -i\`). Use non-interactive versions of commands (e.g. \`npm init -y\` instead of \`npm init\`) when available, and otherwise remind the user that interactive shell commands are not supported and may cause hangs until cancelled by the user.
```
NTaylorMullen, line 83: Should double check this doesn't break background shell invocations for things like `node abc.js`

olcan: just checked and it worked fine; interactivity is a pretty distinct shell concept, but i guess you never know so lmk if you notice odd behavior

PR #323, opened 2025-05-13
feat: Update system prompt to include ! execution

cperry-goog:


packages/server/src/core/prompts.ts
```diff
@@ -84,6 +84,7 @@ Rigorously adhere to existing project conventions when reading or modifying code
 
 ## Interaction Details
 -   **Help Command:** The user can use '/help' to display help information.
+-   **Direct Command Execution:** Users can execute shell commands directly by prefixing their input with an exclamation mark (!). For example: !ls -la
```
olcan, line 87: i'm somewhat concerned this could confuse the model since shell command execution is a powerful thing that can change state of model's environment arbitrarily. it could be doubly confusing as these "passthrough" shell commands do not go into the context, so e.g. you can imagine the model suggesting the user do something repeatedly, or assume they did not do something when they did, etc

cperry-goog: I keep getting annoyed when it tells me to exit and go to the terminal to execute a command. That said, you are right. Any way to balance both or give up?

olcan:
Ah I see, I was wondering what the goal was. We can try letting the model
know that these commands will not be visible to it. May help to give them a
distinguishing name like “passthrough” commands to drive home the point
that these are sort of special, as if the user is running them in a
separate terminal (and you could use that wording in telling model it won’t
see these)

olcan:
By the way we have two characters that do this, $ and !, unless we removed
one. Also this may still confuse the model so I wouldn’t count on this
working super well generally and could have side effects unfortunately at
the size of system prompt we already have. :(

One thing I would love us to get to is behaviroal auto-evals, where you can
flag this behavior and we can automatically measure its occurrance rate and
then make these sorts of changes more confidently.

NTaylorMullen:
IMO we should remove `$` and align with Claude Code here under `!`

olcan:
Yeah agree

cperry-goog:
I'm going to abandon this change and instead put it into my own Gemini.md to test the behvior.

PR #327, opened 2025-05-13
This commit introduces the hierarchical memory feature, allowing GEMI…

allenhutchison:


packages/cli/src/config/config.ts
```diff
@@ -61,7 +77,237 @@ async function parseArguments(): Promise<CliArgs> {
...
+
+  return finalArgv;
+}
+
+/**
+ * Finds the root of the project containing the given directory.
+ * Project root is defined as the closest ancestor directory containing a '.git' folder.
+ * Returns null if no '.git' directory is found up to the filesystem root.
+ */
+async function findProjectRoot(startDir: string): Promise<string | null> {
```
NTaylorMullen, line 100: Ah this is an interesting shift. Why not treat the target dir (i.e. the one GC was opened in) as root?

packages/cli/src/config/config.ts
```diff
@@ -61,7 +77,237 @@ async function parseArguments(): Promise<CliArgs> {
...
+}
+
+/**
+ * Reads the content of multiple GEMINI.md files.
+ */
+async function readGeminiMdFiles(
+  filePaths: string[],
+  debugMode: boolean,
+): Promise<Array<string | null>> {
+  const contents: Array<string | null> = [];
```
NTaylorMullen: We'll probably want to mark the file paths in the content collection. For instance, if you have a sub-folder with its own GEMINI.md you probably don't want those instructions to impact interactions outside of that directory

packages/server/src/config/config.ts
```diff
@@ -86,6 +87,11 @@ export class Config {
   getUserAgent(): string {
     return this.userAgent;
   }
+
+  // Added getter for userMemory
+  getUserMemory(): string {
+    return this.userMemory;
```
NTaylorMullen, line 94: We'll probably want to conditionally refresh this in case users change/update GEMINI.md's

packages/server/src/core/prompts.ts
```diff
@@ -158,4 +168,6 @@ assistant: I can run \`rm -rf ./temp\`. This will permanently delete the directo
 # Final Reminder
 Your core function is efficient and safe assistance. Balance extreme conciseness with the crucial need for clarity, especially regarding safety and potential system modifications. Always prioritize user control and project conventions. Never make assumptions on the contents of files; instead use '${ReadFileTool.Name}' or '${ReadManyFilesTool.Name}' to ensure you aren't making broad assumptions. Finally, you are an agent - please keep going until the user's query is completely resolved.
 `;
+
+  return `${memoryPrefix}${basePrompt}`;
```
NTaylorMullen: Oh interesting, having the memory prefix the base prompt? What's the motivation there?

NTaylorMullen (commented):
Few other follow ups:
- Will probably want to replace one of the <Tips> with a mention of using GEMINI.md

packages/server/src/core/prompts.ts
allenhutchison: I'm going back and forth on this, I think in the end I'd rather put it at the end, but wanted to have a discussion on it here.

packages/cli/src/config/config.ts
allenhutchison: I want the GEMINI.md to be heirarchacal. So any GEMINI.md in the project root applies to any invocation underneath that project root.

packages/server/src/core/prompts.ts
NTaylorMullen: ya I think the core system instructions are more important so I lean towards having memory as a suffix rather than prefix

allenhutchison: Fixed.

packages/server/src/config/config.ts
allenhutchison: How about I add a new slash command /refreshmemory to accomplish this. This will allow the user to stay in control and won't update the way the app works unexpectedly (i.e. like on a git pull). We can come back and do something fancier like watching the files for changes and alerting the user later?

NTaylorMullen: That sounds like a reasonable middleground!

allenhutchison: ok we now have /refreshmemor and /showmemory I will probably want to refactor those into /memory refresh and /memory show later on to make room for /memory add or something similar down the line.

packages/cli/src/config/config.ts
allenhutchison: This is done now.

packages/cli/src/config/config.ts
allenhutchison: Discussed offline. I think we are on the same page now.

packages/cli/src/config/config.test.ts
```diff
@@ -0,0 +1,76 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+// packages/cli/src/config/config.test.ts
+
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+// import * as fsPromises from 'fs/promises';
```
NTaylorMullen, line 10: I assume these are commented out because of the root note below

packages/cli/src/config/config.ts
```diff
@@ -15,9 +19,32 @@ import {
...
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  warn: (...args: any[]) => console.warn('[WARN]', ...args),
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  error: (...args: any[]) => console.error('[ERROR]', ...args),
+};
+
 const DEFAULT_GEMINI_MODEL = 'gemini-2.5-pro-preview-05-06';
+const GEMINI_MD_FILENAME = 'GEMINI.md';
+const GEMINI_CONFIG_DIR = '.gemini';
+// TODO(adh): Refactor to use a shared ignore list with other tools like glob and read-many-files.
```
NTaylorMullen, line 35: 💯, even cooler we may want to eventually make this list configurable by the user and/or understand .gitignore

packages/cli/src/config/config.ts
```diff
@@ -61,46 +88,318 @@ async function parseArguments(): Promise<CliArgs> {
...
+    if (parentDir === currentDir) {
+      return null;
+    }
+    currentDir = parentDir;
+  }
+}
+
+async function collectDownwardGeminiFiles(
+  directory: string,
+  collectedPaths: string[],
```
NTaylorMullen: Possible to return the collected paths instead?

packages/cli/src/ui/hooks/useGeminiStream.ts
```diff
@@ -197,27 +235,16 @@ export const useGeminiStream = (
...
-            // we should maximize the amount of output sent to <Static />.
             const splitPoint = findLastSafeSplitPoint(geminiMessageBuffer);
             if (splitPoint === geminiMessageBuffer.length) {
-              // Update the existing message with accumulated content
               setPendingHistoryItem((item) => ({
                 type: item?.type as 'gemini' | 'gemini_content',
                 text: geminiMessageBuffer,
               }));
             } else {
-              // This indicates that we need to split up this Gemini Message.
```
NTaylorMullen: Should bring this one back

packages/cli/src/ui/hooks/useGeminiStream.ts
```diff
@@ -197,27 +235,16 @@ export const useGeminiStream = (
 
             geminiMessageBuffer += event.value;
 
-            // Split large messages for better rendering performance. Ideally,
```
NTaylorMullen: ditto

packages/server/src/tools/read-many-files.ts
```diff
@@ -100,6 +100,7 @@ const DEFAULT_EXCLUDES: string[] = [
   '**/*.odp',
   '**/*.DS_Store',
   '**/.env',
+  '**/GEMINI.md',
```
NTaylorMullen, line 103: ah interesting, good catch

NTaylorMullen (approved):
There's a number of stale / descriptive comments you could probably do a pass. I've been doing:

```
In my last commit I introduced a number of comments that weren't necessarily high-value. Could you cleanup my last commit for me and remove any non-high value comment that were added?
```

packages/cli/src/config/config.test.ts
allenhutchison: Yes, and cleaned up.

packages/cli/src/config/config.ts
allenhutchison: agree

packages/cli/src/ui/hooks/useGeminiStream.ts
allenhutchison: done

packages/cli/src/ui/hooks/useGeminiStream.ts
allenhutchison: done

packages/server/src/tools/read-many-files.ts
allenhutchison: Thanks.

PR #328, opened 2025-05-13
Add a new UI element to indicate use of memory files.

cornmander:


packages/cli/src/ui/App.tsx
```diff
@@ -213,11 +213,20 @@ export const App = ({ config, settings, cliVersion }: AppProps) => {
...
+                <Box>
+                  <Text color={Colors.SubtleComment}>cwd: </Text>
+                  <Text color={Colors.LightBlue}>
+                    {shortenPath(config.getTargetDir(), 70)}
+                  </Text>
+                </Box>
+                {config.getGeminiMdFileCount() > 0 && (
+                  <Box>
+                    <Text color={Colors.SubtleComment}>
+                      Using {config.getGeminiMdFileCount()} gemini.md files
```
NTaylorMullen, line 226: gemini.md -> the constant for the memory file name.

As a follow up we'll probably also want to ensure this updates IF a user happens to delete or create GEMINI.md files 😄

PR #331, opened 2025-05-13
Update package-lock.json

jacob314:
Looks like this file wasn't updated when other changes were made or there is something wrong with my local config.


KeijiBranshi (approved):
woops. that was me. 

I also fix this in #330

NTaylorMullen:
Still need this or fixed in Brandon's?

PR #332, opened 2025-05-13
Using `Enter` on slash commands will execute it

miguelsolorio:
This improves how the slash commands work, instead of requiring multiple `Enter` presses we can execute on the first press.

## Before

https://github.com/user-attachments/assets/6dfa31b8-24b4-4de8-9cd0-17fb525b7c87

## After

https://github.com/user-attachments/assets/5814e83e-15bb-4fc6-8c39-5671621f3c51


jacob314:
You will need to rebase as this needs to be tweaked slightly to work with the multiline editor. you might want to just run your prompt again to fix it as that could be faster than merging.

miguelsolorio:
Alright, let me create a new PR

PR #334, opened 2025-05-13
Use `Enter` on slash commands to execute

miguelsolorio:
This improves how the slash commands work, instead of requiring multiple `Enter` presses we can execute on the first press.

## Before

https://github.com/user-attachments/assets/6dfa31b8-24b4-4de8-9cd0-17fb525b7c87

## After

https://github.com/user-attachments/assets/5814e83e-15bb-4fc6-8c39-5671621f3c51


NTaylorMullen (approved):
Absolutely love this. Totally agree we should have this!

Also @miguelsolorio I'd love your take on the UI / appearance of the completion list as well

jacob314 (approved):
lgtm

PR #335, opened 2025-05-13
First step refactoring InputPrompt

jacob314:
Trying a chain of small pull requests to move logic out of the app and into InputPrompt


NTaylorMullen (approved):
Champ status with this @jacob314 !

PR #340, opened 2025-05-13
Refactor TextBuffer to be a React hook

jacob314:
Amazingly Gemini did the majority of this refactor in a single shot thanks to @KeijiBranshi's advice on what prompt to use to explain the refactor.

### DIFF
--- a/packages/cli/src/ui/components/shared/multiline-editor.tsx
+++ b/packages/cli/src/ui/components/shared/multiline-editor.tsx
@@ -4,10 +4,10 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { TextBuffer } from './text-buffer.js';
+import { useTextBuffer, type TextBufferHook } from './text-buffer.js';
 import chalk from 'chalk';
 import { Box, Text, useInput, useStdin, Key } from 'ink';
-import React, { useState, useCallback } from 'react';
+import React, { useEffect } from 'react';
 import { useTerminalSize } from '../../hooks/useTerminalSize.js';
 import { Colors } from '../../colors.js';
 
@@ -68,10 +68,6 @@ export const MultilineTextEditor = ({
   navigateDown,
   inputPreprocessor,
 }: MultilineTextEditorProps): React.ReactElement => {
-  const [buffer, setBuffer] = useState(
-    () => new TextBuffer(initialText, initialCursorOffset),
-  );
-
   const terminalSize = useTerminalSize();
   const effectiveWidth = Math.max(
     20,
@@ -81,35 +77,19 @@ export const MultilineTextEditor = ({
 
   const { stdin, setRawMode } = useStdin();
 
-  // TODO(jacobr): make TextBuffer immutable rather than this hack to act
-  // like it is immutable.
-  const updateBufferState = useCallback(
-    (mutator: (currentBuffer: TextBuffer) => void) => {
-      setBuffer((currentBuffer) => {
-        mutator(currentBuffer);
-        // Create a new instance from the mutated buffer to trigger re-render
-        return TextBuffer.fromBuffer(currentBuffer);
-      });
-    },
-    [],
-  );
+  const buffer: TextBufferHook = useTextBuffer({
+    initialText,
+    initialCursorOffset,
+    viewport: { height, width: effectiveWidth },
+    stdin,
+    setRawMode,
+  });
 
-  const openExternalEditor = useCallback(async () => {
-    const wasRaw = stdin?.isRaw ?? false;
-    try {
-      setRawMode?.(false);
-      // openInExternalEditor mutates the buffer instance
-      await buffer.openInExternalEditor();
-    } catch (err) {
-      console.error('[MultilineTextEditor] external editor error', err);
-    } finally {
-      if (wasRaw) {
-        setRawMode?.(true);
-      }
-      // Update state with the mutated buffer to trigger re-render
-      setBuffer(TextBuffer.fromBuffer(buffer));
+  useEffect(() => {
+    if (onChange) {
+      onChange(buffer.text);
     }
-  }, [buffer, stdin, setRawMode, setBuffer]);
+  }, [buffer.text, onChange]);
 
   useInput(
     (input, key) => {
@@ -131,7 +111,7 @@ export const MultilineTextEditor = ({
           input.length === 1 &&
           input.charCodeAt(0) === 5);
       if (isCtrlX || isCtrlE) {
-        openExternalEditor();
+        buffer.openInExternalEditor();
         return;
       }
 
@@ -142,8 +122,6 @@ export const MultilineTextEditor = ({
         console.log('[MultilineTextEditor] event', { input, key });
       }
 
-      let bufferMutated = false;
-
       if (input.startsWith('[') && input.endsWith('u')) {
         const m = input.match(/^\[([0-9]+);([0-9]+)u$/);
         if (m && m[1] === '13') {
@@ -151,14 +129,10 @@ export const MultilineTextEditor = ({
           const hasCtrl = Math.floor(mod / 4) % 2 === 1;
           if (hasCtrl) {
             if (onSubmit) {
-              onSubmit(buffer.getText());
+              onSubmit(buffer.text);
             }
           } else {
             buffer.newline();
-            bufferMutated = true;
-          }
-          if (bufferMutated) {
-            updateBufferState((_) => {}); // Trigger re-render if mutated
           }
           return;
         }
@@ -171,14 +145,10 @@ export const MultilineTextEditor = ({
           const hasCtrl = Math.floor(mod / 4) % 2 === 1;
           if (hasCtrl) {
             if (onSubmit) {
-              onSubmit(buffer.getText());
+              onSubmit(buffer.text);
             }
           } else {
             buffer.newline();
-            bufferMutated = true;
-          }
-          if (bufferMutated) {
-            updateBufferState((_) => {}); // Trigger re-render if mutated
           }
           return;
         }
@@ -186,63 +156,42 @@ export const MultilineTextEditor = ({
 
       if (input === '\n') {
         buffer.newline();
-        updateBufferState((_) => {});
         return;
       }
 
       if (input === '\r') {
         if (onSubmit) {
-          onSubmit(buffer.getText());
+          onSubmit(buffer.text);
         }
         return;
       }
 
       if (key.upArrow) {
-        if (buffer.getCursor()[0] === 0 && navigateUp) {
+        if (buffer.cursor[0] === 0 && navigateUp) {
           navigateUp();
           return;
         }
       }
 
       if (key.downArrow) {
-        if (
-          buffer.getCursor()[0] === buffer.getText().split('\n').length - 1 &&
-          navigateDown
-        ) {
+        if (buffer.cursor[0] === buffer.lines.length - 1 && navigateDown) {
           navigateDown();
           return;
         }
       }
 
-      const modifiedByHandleInput = buffer.handleInput(
-        input,
-        key as Record<string, boolean>,
-        { height, width: effectiveWidth },
-      );
-
-      if (modifiedByHandleInput) {
-        updateBufferState((_) => {});
-      }
-
-      const newText = buffer.getText();
-      if (onChange) {
-        onChange(newText);
-      }
+      buffer.handleInput(input, key as Record<string, boolean>);
     },
     { isActive: focus },
   );
 
-  const visibleLines = buffer.getVisibleLines({
-    height,
-    width: effectiveWidth,
-  });
-  const [cursorRow, cursorCol] = buffer.getCursor();
-  const scrollRow = buffer.getScrollRow();
-  const scrollCol = buffer.getScrollCol();
+  const visibleLines = buffer.visibleLines;
+  const [cursorRow, cursorCol] = buffer.cursor;
+  const [scrollRow, scrollCol] = buffer.scroll;
 
   return (
     <Box flexDirection="column">
-      {buffer.getText().length === 0 && placeholder ? (
+      {buffer.text.length === 0 && placeholder ? (
         <Text color={Colors.SubtleComment}>{placeholder}</Text>
       ) : (
         visibleLines.map((lineText, idx) => {
--- a/packages/cli/src/ui/components/shared/multiline-editor.tsx
+++ b/packages/cli/src/ui/components/shared/multiline-editor.tsx
@@ -7,7 +7,7 @@
 import { useTextBuffer } from './text-buffer.js';
 import chalk from 'chalk';
 import { Box, Text, useInput, useStdin, Key } from 'ink';
-import React, { useEffect } from 'react';
+import React from 'react';
 import { useTerminalSize } from '../../hooks/useTerminalSize.js';
 import { Colors } from '../../colors.js';
 
@@ -83,13 +83,8 @@ export const MultilineTextEditor = ({
     viewport: { height, width: effectiveWidth },
     stdin,
     setRawMode,
+    onChange, // Pass onChange to the hook
   });
-
-  useEffect(() => {
-    if (onChange) {
-      onChange(buffer.text);
-    }
-  }, [buffer.text, onChange]);
 
   useInput(
     (input, key) => {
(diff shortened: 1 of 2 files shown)


packages/cli/src/ui/components/shared/text-buffer.ts
```diff
@@ -75,975 +76,830 @@ function dbg(...args: unknown[]): void {
...
-        text: this.getText(),
-      });
-    }
-    return this.version !== beforeVer || cursorMoved;
-  }
+  // For rendering
+  visibleLines: string[];
 }
+
+export type TextBufferHook = ReturnType<typeof useTextBuffer>;
```
KeijiBranshi: suggestion: I don't think this is needed. the `TextBuffer` interface (ie `ReturnType<typeof useTextBuffer>`) is already exported. Consumers should just use that.

packages/cli/src/ui/components/shared/multiline-editor.tsx
```diff
@@ -4,10 +4,10 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { TextBuffer } from './text-buffer.js';
+import { useTextBuffer, type TextBufferHook } from './text-buffer.js';
```
KeijiBranshi: suggestion: see previous comment about `TextBuffer` vs `TextBufferHook`

packages/cli/src/ui/components/shared/multiline-editor.tsx
```diff
@@ -81,35 +77,19 @@ export const MultilineTextEditor = ({
...
-      await buffer.openInExternalEditor();
-    } catch (err) {
-      console.error('[MultilineTextEditor] external editor error', err);
-    } finally {
-      if (wasRaw) {
-        setRawMode?.(true);
-      }
-      // Update state with the mutated buffer to trigger re-render
-      setBuffer(TextBuffer.fromBuffer(buffer));
+  useEffect(() => {
```
KeijiBranshi: thought: feels a bit like code-smell. this could indicate that `buffer.text` should be a `useState` managed outside of `useTextBuffer`.

packages/cli/src/ui/components/shared/text-buffer.ts
```diff
@@ -75,975 +76,830 @@ function dbg(...args: unknown[]): void {
...
+  setRawMode,
+}: UseTextBufferProps): TextBuffer {
+  const [lines, setLines] = useState<string[]>(() => {
+    const l = initialText.split('\n');
+    return l.length === 0 ? [''] : l;
+  });
+
+  const [[initialCursorRow, initialCursorCol]] = useState(() =>
+    calculateInitialCursorPosition(lines, initialCursorOffset),
+  );
```
KeijiBranshi, line 136: suggestion: this ignores the `setState` return value from `useState`. That suggests this should just be a `useMemo` w/ nothing in its deps array. 

```suggestion
  const [initialCursorRow, initialCursorCol] = useMemo(() =>
    calculateCursorPosition(lines, initialCursorOffset),
  );
```

KeijiBranshi: ...or (maybe even better), you can nix the `useMemo` entirely. It looks like these are only used in the cursor state managers below:

```
  const [cursorRow, setCursorRow] = useState<number>(initialCursorRow);
  const [cursorCol, setCursorCol] = useState<number>(initialCursorCol);
```

I _think_ React will ignore subsequent updates to the initial state value provided to `useState`. The fact that it's "initial state" implies you only care about the first value to be provided there in the lifetime of the hook/component.

packages/cli/src/ui/components/shared/text-buffer.ts
```diff
@@ -75,975 +76,830 @@ function dbg(...args: unknown[]): void {
...
+
+  const [clipboard, setClipboard] = useState<string | null>(null);
+  const [selectionAnchor, setSelectionAnchor] = useState<
+    [number, number] | null
+  >(null);
+
+  const currentLine = useCallback(
+    (r: number): string => lines[r] ?? '',
+    [lines],
+  );
```
KeijiBranshi, line 156: nitpick(non-blocking): this probably doesn't need to be a callback. I see its mostly just called with `cursorRow` anyway. It _might_ be cleaner to just make it a value, then React will always refresh `currentLine` based on `lines`'s current state.

```suggestion
  const currentLine = lines[cursorRow] ?? '';
  const previousLine = lines[cursorRow-1] ?? '';
```

packages/cli/src/ui/components/shared/text-buffer.ts
```diff
@@ -75,975 +76,830 @@ function dbg(...args: unknown[]): void {
...
+  const currentLineLen = useCallback(
+    (r: number): number => cpLen(currentLine(r)),
+    [currentLine],
+  );
```
KeijiBranshi, line 160: suggestion: same as previous comment

```suggestion
  const currentLineLen = cpLen(currentLine);
```

packages/cli/src/ui/components/shared/text-buffer.ts
```diff
@@ -75,975 +76,830 @@ function dbg(...args: unknown[]): void {
...
+
+    if (newScrollRow !== scrollRow) {
+      setScrollRow(newScrollRow);
+    }
+    if (newScrollCol !== scrollCol) {
+      setScrollCol(newScrollCol);
+    }
+  }, [cursorRow, cursorCol, scrollRow, scrollCol, viewport]);
+
+  const _pushUndo = useCallback(() => {
```
KeijiBranshi: nitpick: `_` is usually an indication that something is unused. while this is internal, it's still used and so the `_` is unnecessary

```suggestion
  const pushUndo = useCallback(() => {
```

packages/cli/src/ui/components/shared/text-buffer.ts
```diff
@@ -75,975 +76,830 @@ function dbg(...args: unknown[]): void {
...
+
+  const undo = useCallback((): boolean => {
+    const state = undoStack[undoStack.length - 1];
+    if (!state) return false;
+
+    setUndoStack((prev) => prev.slice(0, -1));
+    const currentSnapshot = { lines: [...lines], cursorRow, cursorCol };
+    setRedoStack((prev) => [...prev, currentSnapshot]);
+    return _restoreState(state);
+  }, [undoStack, lines, cursorRow, cursorCol, _restoreState]);
```
KeijiBranshi, line 229: thought/suggestion: the logic around the undo/redo stack seems abstract enough that it could just be its own hook - eg `useUndoRedoStack` or `useHistoryStack`. I think if you pulled that out then just had `useTextBuffer` call it, it could help clean up the main logic for `useTextBuffer` quite a bit. 

here's what gemini spit out for me when I asked it to write a undo/redo stack hook:

```
import { useState, useCallback } from 'react';

/**
 * A React hook to manage an undo/redo history stack for any state,
 * using useState for history management.
 *
 * @param {any} initialState The initial state value.
 * @returns {Object} An object containing the current state, a function to set the state
 * (which also manages history), undo/redo functions, and boolean flags
 * indicating if undo/redo are possible.
 */
const useUndoRedoStack = (initialState) => {
  const [history, setHistory] = useState([initialState]);
  const [historyIndex, setHistoryIndex] = useState(0);
  const [currentState, setCurrentState] = useState(initialState);

  const canUndo = historyIndex > 0;
  const canRedo = historyIndex < history.length - 1;

  /**
   * Sets the current state and manages the history.
   * When a new state is set, it truncates any "future" (redoable) history.
   *
   * @param {any} newState The new state value.
   */
  const setState = useCallback((newState) => {
    setCurrentState(newState); // Immediately update the visible state

    // Only add to history if the new state is different from the current history entry
    if (newState !== history[historyIndex]) {
      setHistory(prevHistory => {
        let newHistory = prevHistory;
        // If we've undone some changes, clear the redo history
        if (historyIndex < prevHistory.length - 1) {
          newHistory = prevHistory.slice(0, historyIndex + 1);
        }
        const updatedHistory = [...newHistory, newState];
        setHistoryIndex(updatedHistory.length - 1); // Update index based on new history length
        return updatedHistory;
      });
    }
  }, [history, historyIndex]); // Dependencies are important for useCallback

  /**
   * Undoes the last state change.
   */
  const undo = useCallback(() => {
    if (canUndo) {
      setHistoryIndex(prevIndex => {
        const newIndex = prevIndex - 1;
        setCurrentState(history[newIndex]); // Set current state from history
        return newIndex;
      });
    }
  }, [canUndo, history]); // `history` is a dependency here because `history[newIndex]` depends on it

  /**
   * Redoes the last undone state change.
   */
  const redo = useCallback(() => {
    if (canRedo) {
      setHistoryIndex(prevIndex => {
        const newIndex = prevIndex + 1;
        setCurrentState(history[newIndex]); // Set current state from history
        return newIndex;
      });
    }
  }, [canRedo, history]); // `history` is a dependency here because `history[newIndex]` depends on it

  return {
    currentState,
    setState,
    undo,
    redo,
    canUndo,
    canRedo,
  };
};

export default useUndoRedoStack;
```

packages/cli/src/ui/components/shared/text-buffer.ts
```diff
@@ -75,975 +76,830 @@ function dbg(...args: unknown[]): void {
...
+      if (!state) return false;
+      setLines(state.lines);
+      setCursorRow(state.cursorRow);
+      setCursorCol(state.cursorCol);
+      return true;
+    },
+    [],
+  );
+
+  const getText = useCallback((): string => lines.join('\n'), [lines]);
```
KeijiBranshi: suggestion: this doesn't need to be a callback. React will always make sure it stays up to date

```suggestion
  const getText = lines.join('\n');
```

KeijiBranshi (approved):
🚀

packages/cli/src/ui/components/shared/text-buffer.ts
jacob314: done

packages/cli/src/ui/components/shared/multiline-editor.tsx
jacob314: done

packages/cli/src/ui/components/shared/text-buffer.ts
jacob314: done. Applied your final suggestion.

packages/cli/src/ui/components/shared/text-buffer.ts
jacob314: done. switched to a simple function as I agree callback caching doesn't matter here.

packages/cli/src/ui/components/shared/text-buffer.ts
jacob314: sadly need this for a bunch of different lines not just the current and next one. left as while it is a bit silly it works.

packages/cli/src/ui/components/shared/text-buffer.ts
jacob314: done

packages/cli/src/ui/components/shared/text-buffer.ts
jacob314: ack. I would prefer to put this off for now as I'm not quite following the benefit. for now we aren't calling undo/redo outside this class and we might want to just remove the undo/redo logic.

packages/cli/src/ui/components/shared/text-buffer.ts
jacob314: done

packages/cli/src/ui/components/shared/text-buffer.ts
jacob314: not done as this triggers lint failures that I don't know how to resolve unless we make a bunch of calls not use useCallback that should.
```
151:9  warning  The 'currentLine' function makes the dependencies of useCallback Hook (at line 258) change on every render. To fix this, wrap the definition of 'currentLine' in its own useCallback() Hook        react-hooks/exhaustive-deps
  151:9  warning  The 'currentLine' function makes the dependencies of useCallback Hook (at line 281) change on every render. To fix this, wrap the definition of 'currentLine' in its own useCallback() Hook        react-hooks/exhaustive-deps
  151:9  warning  The 'currentLine' function makes the dependencies of useCallback Hook (at line 299) change on every render. To fix this, wrap the definition of 'currentLine' in its own useCallback() Hook        react-hooks/exhaustive-deps
  151:9  warning  The 'currentLine' function makes the dependencies of useCallback Hook (at line 330) change on every render. To fix this, wrap the definition of 'currentLine' in its own useCallback() Hook        react-hooks/exhaustive-deps
  151:9  warning  The 'currentLine' function makes the dependencies of useCallback Hook (at line 356) change on every render. To fix this, wrap the definition of 'currentLine' in its own useCallback() Hook        react-hooks/exhaustive-deps
  151:9  warning  The 'currentLine' function makes the dependencies of useCallback Hook (at line 447) change on every render. To fix this, wrap the definition of 'currentLine' in its own useCallback() Hook        react-hooks/exhaustive-deps
  151:9  warning  The 'currentLine' function makes the dependencies of useCallback Hook (at line 482) change on every render. To fix this, wrap the definition of 'currentLine' in its own useCallback() Hook        react-hooks/exhaustive-deps
  151:9  warning  The 'currentLine' function makes the dependencies of useCallback Hook (at line 505) change on every render. To fix this, wrap the definition of 'currentLine' in its own useCallback() Hook        react-hooks/exhaustive-deps
  151:9  warning  The 'currentLine' function makes the dependencies of useCallback Hook (at line 608) change on every render. To fix this, wrap the definition of 'currentLine' in its own useCallback() Hook        react-hooks/exhaustive-deps
  151:9  warning  The 'currentLine' function makes the dependencies of useCallback Hook (at line 783) change on every render. To fix this, wrap the definition of 'currentLine' in its own useCallback() Hook        react-hooks/exhaustive-deps
  153:9  warning  The 'currentLineLen' function makes the dependencies of useCallback Hook (at line 356) change on every render. To fix this, wrap the definition of 'currentLineLen' in its own useCallback() Hook  react-hooks/exhaustive-deps
  153:9  warning  The 'currentLineLen' function makes the dependencies of useCallback Hook (at line 447) change on every render. To fix this, wrap the definition of 'currentLineLen' in its own useCallback() Hook  react-hooks/exhaustive-deps
  153:9  warning  The 'currentLineLen' function makes the dependencies of useCallback Hook (at line 608) change on every render. To fix this, wrap the definition of 'currentLineLen' in its own useCallback() Hook  r
  ```

File: packages/cli/src/ui/components/shared/text-buffer.ts (hunk 8 above)
Comment (jacob314) [reply packages/cli/src/ui/components/shared/text-buffer.ts:229]: if you want to refactor this the undo/redo code after my pull request lands, go for it! happy to review and approve.

File: packages/cli/src/ui/components/shared/multiline-editor.tsx (hunk 3 above)
Comment (jacob314) [reply packages/cli/src/ui/components/shared/multiline-editor.tsx]: agreed. moved it into text-buffer.tsx for now to make it a little less ugly. Added a TODO there to refactor. I think there is a thread that can be pulled on in app.tsx that will simplify this but would rather do it in a separate CL.
I niavely tried adding this code to text-buffer but that causes execution order issues
```
 const setLines = useCallback((newLinesOrUpdater:  React.SetStateAction<string[]>): void => {
    if (newLinesOrUpdater instanceof Function) {
      const wrappedCallback = (lines: string[]): string[] => {
        const newLines = newLinesOrUpdater(lines);
        onChange?.(newLines.join('\n'));
        return newLines
      };
      setLinesInternal(wrappedCallback);
    } else {
      const newLines = newLinesOrUpdater;
      onChange?.(newLines.join('\n'));
      setLinesInternal(newLines);
    }
  }, [setLinesInternal]);
  ```

packages/cli/src/ui/components/shared/text-buffer.ts
KeijiBranshi: bsaed on that warning, I'm guessing the code looked like this?

```
const currentLine = (r: number) => lines[r]
```

let me know if I'm mistaken. but if true, then that means `currentLine` gets reassigned/redefined on every re-render. This forces the `useCallback`s below that take `currentLine` as a dependency to reassign themselves as well, making the `useCallback` hook inconsequential.

The solutions for this would be to:

a. change `currentLine` to a constant string assignment

```
const currentLine = lines[row]
```

the `useCallbacks` below will _not_ be redefined so long as this line stays the same string. or...

b. remove the `useCallback` wrappers from the other functions below. there's some debate on whether `useCallback` is a necessary hook. Some find function definitions light weight enough that they question whether useCallback is an over-optimization.

PR #341, opened 2025-05-13
refactor: move nested debugmessage and slashcommand hooks outside of useGeminiStream

KeijiBranshi:
BUG=b/411426253

This refactors how slash commands are processed and how debug messages are managed within the app. The main changes include:

- Decoupling slash command processing from the `useGeminiStream` hook. The useSlashCommandProcessor is now utilized in `App.tsx`, and its `handleSlashCommand` function is passed down. Note, `useGeminiStream` never actually needed the other return value from that hook, `slashCommands`.
- Standardizing debug message handling by replacing direct setDebugMessage calls with an onDebugMessage callback. This callback originates in App.tsx and is passed through useGeminiStream to the various command handlers. Note, `useGeminiStream` never actually needed the `debugMessage` state inside the hook.

This refactoring is a preparatory step to address a larger issue concerning parallel tool confirmation dialogs. By iteratively reducing the amount of state managed by `useGeminiStream`, I'm hoping to make it easier to implement the fix for the aforementioned bug.

### DIFF
--- a/packages/cli/src/ui/App.tsx
+++ b/packages/cli/src/ui/App.tsx
@@ -38,6 +38,7 @@ interface AppProps {
 export const App = ({ config, settings, cliVersion }: AppProps) => {
   const { history, addItem, clearItems } = useHistory();
   const [startupWarnings, setStartupWarnings] = useState<string[]>([]);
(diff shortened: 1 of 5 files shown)


NTaylorMullen:
Appreciate the splitting out!

PR #342, opened 2025-05-14
fix: pass startup warnings to app as prop

KeijiBranshi:
this performs an unnecessary re-render on startup. `startupWarnings` is first initialized to `[]`, then `useEffect` kicks off and triggers a re-render with the new value of `startupWarnings`. 

this change pre-computes the startup warnings and passes it to `<App />` as a prop


packages/cli/src/ui/hooks/useAppEffects.ts
```diff
@@ -1,42 +0,0 @@
-/**
```
NTaylorMullen, line 1: I also found that GC would regularly misunderstand the purpose of this file based off of file name alone (`useAppEffects`). Happy to see this go!

PR #343, opened 2025-05-14
Support images with @ syntax.

jacob314:
Future work: support drag and drop automatically adding @ imports. For discussion: do we need special syntax for image imports e.g. @image:somePath or is that overkill as we can determine the type from files.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,6 +11,7 @@
         "packages/*"
       ],
       "devDependencies": {
+        "@types/mime-types": "^2.1.4",
         "esbuild": "^0.25.4",
         "eslint": "^9.24.0",
         "eslint-config-prettier": "^10.1.2",
@@ -28,8 +29,6 @@
     },
     "node_modules/@alcalzone/ansi-tokenize": {
       "version": "0.1.3",
-      "resolved": "https://registry.npmjs.org/@alcalzone/ansi-tokenize/-/ansi-tokenize-0.1.3.tgz",
-      "integrity": "sha512-3yWxPTq3UQ/FY9p1ErPxIyfT64elWaMvM9lIHnaqpyft63tkxodF5aUElYHrdisWve5cETkh1+KBw1yJuW0aRw==",
       "license": "MIT",
       "dependencies": {
         "ansi-styles": "^6.2.1",
@@ -41,8 +40,6 @@
     },
     "node_modules/@alcalzone/ansi-tokenize/node_modules/ansi-styles": {
       "version": "6.2.1",
-      "resolved": "https://registry.npmjs.org/ansi-styles/-/ansi-styles-6.2.1.tgz",
-      "integrity": "sha512-bN798gFfQX+viw3R7yrGWRqnrN2oRkEkUjjl4JNn4E8GxxbjtG3FbrEIIY3l8/hrwUwIeCZvi4QuOTP4MErVug==",
       "license": "MIT",
       "engines": {
         "node": ">=12"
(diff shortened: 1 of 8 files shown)


packages/cli/src/ui/hooks/useGeminiStream.ts
```diff
@@ -34,6 +34,56 @@ import { findLastSafeSplitPoint } from '../utils/markdownUtilities.js';
 import { useStateAndRef } from './useStateAndRef.js';
 import { UseHistoryManagerReturn } from './useHistoryManager.js';
 
+function partListUnionToString(value: PartListUnion): string {
```
jacob314, line 37: @NTaylorMullen is there a standard place I should move this to? I think you mentioned there was about to be a  good location for this.

packages/server/src/tools/read-many-files.ts
```diff
@@ -81,14 +82,6 @@ const DEFAULT_EXCLUDES: string[] = [
   '**/*.bz2',
   '**/*.rar',
   '**/*.7z',
-  '**/*.png',
```
jacob314, line 84: Open to a better solution. One idea is to only include images if explicitly referenced in a file path rather than the directory. Also open to going with a `@files` or `@images` sort of scheme and adding a separate read-many-media tool that supports only images and pdf.

package-lock.json
```diff
@@ -28,8 +29,6 @@
     },
     "node_modules/@alcalzone/ansi-tokenize": {
       "version": "0.1.3",
-      "resolved": "https://registry.npmjs.org/@alcalzone/ansi-tokenize/-/ansi-tokenize-0.1.3.tgz",
```
jacob314, line 31: not sure why my client is deleting all of these. Anyone know what I need to add so they appear?

packages/cli/src/ui/hooks/useGeminiStream.ts
NTaylorMullen: In the other PR I made [this suggestion](https://github.com/google-gemini/gemini-code/pull/183/files/7ee08b898943a3eef9feacefff661287a078e037#r2061102714). Basically just saying we could make a nice abstraction that could do this. Nothing exists currently to manage this BUT is a key part of our system.

packages/server/src/tools/read-many-files.ts
NTaylorMullen: I kind of like your suggestion of just inferring from file path for the time being

NTaylorMullen (approved):
> For discussion: do we need special syntax for image imports e.g. @image:somePath or is that overkill as we can determine the type from files.

I agree with you that it might be overkill to do `@image`. IMO I think we should stick with just normal file paths for now unless we see a need not to.

packages/server/src/tools/read-many-files.ts
```diff
@@ -374,14 +384,17 @@ Default excludes apply to common non-text files and large dependency directories
     }
     if (
       concatenatedContent.length === 0 &&
-      processedFilesRelativePaths.length === 0
+      processedFilesRelativePaths.length === 0 &&
+      images.length === 0
```
sethtroisi, line 388: Maybe add images.length to the output on line 360

packages/cli/src/ui/hooks/useGeminiStream.ts
jacob314: ok started adding a minimal GeminiCodeRequest abstraction in this pull request as a place to put this method. In a follow up pull request we can make it something more useful.

packages/server/src/tools/read-many-files.ts
jacob314: ok tweaked this so images are only added if they are specifically mentioned in the pattern to avoid degradations in performance if users accidentally include some large images or pdfs they don't care about just by adding a directory.
@*.png and cases like that work. also added tests.

packages/server/src/tools/read-many-files.ts
jacob314: cleaned this up so images are included in order. FYI one downside is the PartListUnion will now be
string[] instead of a single string if the case of multiple text files. This shouldn't matter though as I assume the Gemini API compacts the two cases down the same things. we could always cleanup ourself and concat together multiple adjacent strings in a PartListUnion that is an string[].

jacob314:
Sorry botched my github workflow. Please review the updated pull request in 
https://github.com/google-gemini/gemini-cli/pull/447
All comments should be addressed.

PR #344, opened 2025-05-14
Improvements to suggestions & slash commands

miguelsolorio:
Improvements to suggestions & slash commands
- Simplified suggestions to use color highlighting instead of bg color
- Added descriptions to `/` commands
- Added 2-column to `/` commands for better readability
- When typing `Enter` on a `/` command, it'll immediately execute

https://github.com/user-attachments/assets/b3332182-7265-4f5a-8c5d-592153b1fd39


packages/cli/src/ui/App.tsx
```diff
@@ -247,7 +247,7 @@ export const App = ({ config, settings, cliVersion }: AppProps) => {
                 resetCompletion={completion.resetCompletionState}
               />
               {completion.showSuggestions && (
-                <Box marginTop={1}>
+                <Box marginTop={0}>
```
```suggestion
                <Box>
```

NTaylorMullen (approved):
:shipit: Looks much cleaner!

miguelsolorio:
I've gotta update the PR to support long descriptions, the latest changes from `main` introduced longer descriptions

NTaylorMullen:
> I've gotta update the PR to support long descriptions, the latest changes from `main` introduced longer descriptions

If it helps the `<Text>` element has a `wrap="truncate-end"` attribute you can provide to auto truncate things. Or you can `wrap` if that works as well

miguelsolorio:
Yea I was trying to wrap the text and there were some alignment issues that were happening. Think I've got it working now:

CleanShot 2025-05-14 at 14 21 10@2x

PR #346, opened 2025-05-14
improve sandboxing status message, update README, remove dead code from a previous change to build_sandbox.sh

olcan:


packages/cli/src/ui/components/Footer.tsx
```diff
@@ -46,13 +46,15 @@ export const Footer: React.FC<FooterProps> = ({
...
+          {process.env.SANDBOX.replace(/^gemini-(?:code-)?/, '')}
+        </Text>
       ) : process.env.SANDBOX === 'sandbox-exec' ? (
         <Text color={Colors.AccentYellow}>
-          using macos seatbelt ({process.env.SEATBELT_PROFILE})
+          sandbox-exec ({process.env.SEATBELT_PROFILE})
         </Text>
       ) : (
-        <Text color={Colors.AccentRed}> WARNING: OUTSIDE SANDBOX </Text>
+        <Text color={Colors.AccentRed}>no sandbox (see README)</Text>
```
NTaylorMullen, line 57: External users wont have access to this README.md. What'd you have in mind for them?

olcan:
they would be exposed to README via github page no? is that not going to be
the primary distribution like codex?

NTaylorMullen:
> they would be exposed to README via github page no? is that not going to be the primary distribution like codex?

I wouldn't assume that to be true. Many folks will simply follow an installation mechanism via npm and wont be aware of where the code is (or if) hosted

olcan:
we can do basic things like dump the readme on /help
but i think there should be a project page that we can point to
ideally this is our github page, like codex, as i would expect for an open
source project
otherwise we would need something else, and the bar would be a lot higher,
like claude-code docs

NTaylorMullen:
@cperry-goog we were planning to have dedicated docs for GC I assume weren't we?

cperry-goog:
docs sg, my plan was no new sites, just use GH for the launch to keep it simple, for the exact reason @olcan mentions; I don't want a high bar.

olcan:
+1 also for keeping the "community project" vibes

PR #348, opened 2025-05-14
Add UI memory indicator. 

allenhutchison:


packages/cli/src/ui/App.tsx
```diff
@@ -62,6 +67,26 @@ export const App = ({
     handleThemeHighlight,
   } = useThemeCommand(settings, setThemeError);
 
+  const fetchGeminiMdFileCount = useCallback(async () => {
+    if (!config) return;
+    try {
+      const paths = await getGeminiMdFilePaths(
```
NTaylorMullen: nit: instead of calling the method directly can we call a method on our config object to make it extra clear that it manages all interactions.

packages/cli/src/ui/App.tsx
```diff
@@ -62,6 +67,26 @@ export const App = ({
...
+        process.cwd(),
```
NTaylorMullen: For instance I believe this should be `config.getTargetPath()` BUT would be easier to be a detail of the config object itself

NTaylorMullen (approved):

packages/cli/src/ui/App.tsx
allenhutchison: fair enough. I was trying to keep the change to one file. I've refactored it to be on the config module.

PR #349, opened 2025-05-14
Update Gemini Code verbiage -> Gemini CLI

NTaylorMullen:
- Did not update details that impact GC execution. Meaning packages are still named gemini-code (for now) and things that import them still import them as gemini-code.


KeijiBranshi (approved):
LGTM

PR #350, opened 2025-05-14
refactor: break submitQuery into smaller functions

KeijiBranshi:


packages/cli/src/ui/hooks/useGeminiStream.ts
```diff
@@ -82,248 +100,492 @@ export const useGeminiStream = (
...
         }
 
         if (pendingHistoryItemRef.current) {
           addItem(pendingHistoryItemRef.current, userMessageTimestamp);
           setPendingHistoryItem(null);
         }
 
-        setStreamingState(StreamingState.Idle);
+        if (
+          processingStatus === 'completed' ||
```
NTaylorMullen: nit: an enum for processing status instead of strings

packages/cli/src/ui/hooks/useGeminiStream.ts
```diff
@@ -82,248 +100,492 @@ export const useGeminiStream = (
...
-            // we should maximize the amount of output sent to <Static />.
-            const splitPoint = findLastSafeSplitPoint(geminiMessageBuffer);
-            if (splitPoint === geminiMessageBuffer.length) {
-              // Update the existing message with accumulated content
-              setPendingHistoryItem((item) => ({
-                type: item?.type as 'gemini' | 'gemini_content',
-                text: geminiMessageBuffer,
-              }));
-            } else {
-              // This indicates that we need to split up this Gemini Message.
```
NTaylorMullen, line 195: Lets bring back some of these bigger comments since they detail key oddities we need to account for

NTaylorMullen (approved):

packages/cli/src/ui/hooks/useGeminiStream.ts
KeijiBranshi: done. but just wanted to note for future reference that TypeScript string literal unions are effectively enums themselves: https://www.typescriptlang.org/docs/handbook/literal-types.html#string-literal-types

packages/cli/src/ui/hooks/useGeminiStream.ts
KeijiBranshi: done

PR #354, opened 2025-05-14
Add readline-like keybindings to the input prompts.

dewitt:
New keybindings in the main input prompt (when auto-suggestions are not active):
  - `Ctrl+L`: Clears the entire screen.
  - `Ctrl+A`: Moves the cursor to the beginning of the current input line.
  - `Ctrl+E`: Moves the cursor to the end of the current input line.
  - `Ctrl+P`: Navigates to the previous command in the input history.
  - `Ctrl+N`: Navigates to the next command in the input history.

In the multiline text editor (e.g., when editing a previous message):
   - `Ctrl+K`: Deletes text from the current cursor position to the end of the line ("kill line right").

### DIFF
--- a/packages/cli/src/ui/App.tsx
+++ b/packages/cli/src/ui/App.tsx
@@ -170,6 +170,11 @@ export const App = ({
     [setQuery, setEditorState],
   );
 
+  const handleClearScreen = useCallback(() => {
+    clearItems();
+    refreshStatic();
+  }, [clearItems, refreshStatic]);
+
   const completion = useCompletion(
     query,
     config.getTargetDir(),
@@ -288,6 +293,8 @@ export const App = ({
                 navigateSuggestionUp={completion.navigateUp}
                 navigateSuggestionDown={completion.navigateDown}
                 resetCompletion={completion.resetCompletionState}
+                setEditorState={setEditorState} // Added setEditorState prop
+                onClearScreen={handleClearScreen} // Added onClearScreen prop
               />
               {completion.showSuggestions && (
                 <Box marginTop={1}>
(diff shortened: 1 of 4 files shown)


allenhutchison:
Can you add Ctrl+U that one is fairly straightforward.

sethtroisi:
I'm so excited for this I try to get ctrl+a at least once a session.

packages/cli/src/ui/App.tsx
```diff
@@ -312,6 +317,8 @@ export const App = ({
                 navigateSuggestionUp={completion.navigateUp}
                 navigateSuggestionDown={completion.navigateDown}
                 resetCompletion={completion.resetCompletionState}
+                setEditorState={setEditorState} // Added setEditorState prop
```
```suggestion
                setEditorState={setEditorState}
```

packages/cli/src/ui/App.tsx
```diff
@@ -312,6 +317,8 @@ export const App = ({
...
+                onClearScreen={handleClearScreen} // Added onClearScreen prop
```
NTaylorMullen, line 321:
```suggestion
                onClearScreen={handleClearScreen}
```

packages/cli/src/ui/App.tsx
```diff
@@ -180,6 +180,11 @@ export const App = ({
     [setQuery, setEditorState],
   );
 
+  const handleClearScreen = useCallback(() => {
+    clearItems();
+    refreshStatic();
```
NTaylorMullen, line 185: This will probably have undesireable consequences BUT is probably better than nothing

NTaylorMullen (approved):

packages/cli/src/ui/App.tsx
dewitt: ACKed. Per the hallway conversation, properly handling a screen update sounds non-trivial. It might even be better to remove Ctrl-L since it's not going to work great this way. But it does *something* when people hit it out of muscle memory, and that might be sufficient reason to leave it in.

dewitt:
> Can you add Ctrl+U that one is fairly straightforward.

Done in [23ea7cd](https://github.com/google-gemini/gemini-cli/pull/354/commits/23ea7cdfd28c4e2b16523a5aebfe6bd9de61587d).

PR #355, opened 2025-05-15
/clear clears the console

sethtroisi:
This prevents lingering console artifacts from when the app was taller


dewitt (commented):
Nice. I copied this for how we handle Ctrl-L as well. We should probably refactor this out somewhere.

NTaylorMullen (approved):

PR #356, opened 2025-05-15
Run console.clear() in handleClearScreen when invoked by Ctrl-L.

dewitt:
Copied from sethtroisi@'s identical improvement to /clear in change #355.


NTaylorMullen (approved):
Nice!

PR #357, opened 2025-05-15
refactor: remove unused props clearItems, openThemeDialog, onSubmit

KeijiBranshi:
just some cleanup


NTaylorMullen (approved):
Oh ya we don't use those anymore. Nice cleanup

PR #358, opened 2025-05-15
Fix: Prevent UI tearing and improve display of long content

NTaylorMullen:
**Note: Review this with whitespace diffing off››**

This commit introduces several changes to better manage terminal height and prevent UI tearing, especially when displaying long tool outputs or when the pending history item exceeds the available terminal height.

- Calculate and utilize available terminal height in `App.tsx`, `HistoryItemDisplay.tsx`, `ToolGroupMessage.tsx`, and `ToolMessage.tsx`.
- Refresh the static display area in `App.tsx` when a pending history item is too large, working around an Ink bug (see https://github.com/vadimdemedes/ink/pull/717).
- Truncate long tool output in `ToolMessage.tsx` and indicate the number of hidden lines.
image

- Refactor `App.tsx` to correctly measure and account for footer height.

Fixes https://b.corp.google.com/issues/414196943

### DIFF
--- a/packages/cli/src/ui/App.tsx
+++ b/packages/cli/src/ui/App.tsx
@@ -5,8 +5,9 @@
  */
 
 import { useCallback, useEffect, useMemo, useState, useRef } from 'react';
-import { Box, DOMElement, measureElement, Static, Text, useStdout } from 'ink';
+import { Box, DOMElement, measureElement, Static, Text } from 'ink';
 import { StreamingState, type HistoryItem } from './types.js';
+import { useTerminalSize } from './hooks/useTerminalSize.js';
 import { useGeminiStream } from './hooks/useGeminiStream.js';
 import { useLoadingIndicator } from './hooks/useLoadingIndicator.js';
 import { useThemeCommand } from './hooks/useThemeCommand.js';
@@ -56,8 +57,7 @@ export const App = ({
   const [debugMessage, setDebugMessage] = useState<string>('');
   const [showHelp, setShowHelp] = useState<boolean>(false);
   const [themeError, setThemeError] = useState<string | null>(null);
-  const [availableTerminalHeight, setAvailableTerminalHeight] =
-    useState<number>(0);
+  const [footerHeight, setFooterHeight] = useState<number>(0);
   const {
     isThemeDialogOpen,
     openThemeDialog,
@@ -195,26 +195,24 @@ export const App = ({
 
   // --- Render Logic ---
 
-  // Get terminal dimensions
-
-  const { stdout } = useStdout();
-  const terminalWidth = stdout?.columns ?? 80;
-  const terminalHeight = stdout?.rows ?? 24;
-  const footerRef = useRef<DOMElement>(null);
+  const { rows: terminalHeight, columns: terminalWidth } = useTerminalSize();
+  const mainControlsRef = useRef<DOMElement>(null);
   const pendingHistoryItemRef = useRef<DOMElement>(null);
 
   // Calculate width for suggestions, leave some padding
   const suggestionsWidth = Math.max(60, Math.floor(terminalWidth * 0.8));
 
   useEffect(() => {
-    const staticExtraHeight = /* margins and padding */ 3;
-    const fullFooterMeasurement = measureElement(footerRef.current!);
-    const fullFooterHeight = fullFooterMeasurement.height;
+    if (mainControlsRef.current) {
+      const fullFooterMeasurement = measureElement(mainControlsRef.current);
+      setFooterHeight(fullFooterMeasurement.height);
+    }
+  }, [terminalHeight]); // Re-calculate if terminalHeight changes, as it might affect footer's rendered height.
 
-    setAvailableTerminalHeight(
-      terminalHeight - fullFooterHeight - staticExtraHeight,
-    );
-  }, [terminalHeight]);
+  const availableTerminalHeight = useMemo(() => {
+    const staticExtraHeight = /* margins and padding */ 3;
+    return terminalHeight - footerHeight - staticExtraHeight;
+  }, [terminalHeight, footerHeight]);
 
   useEffect(() => {
     if (!pendingHistoryItem) {
@@ -260,7 +258,14 @@ export const App = ({
             <Header />
             <Tips />
           </Box>,
-          ...history.map((h) => <HistoryItemDisplay availableTerminalHeight={availableTerminalHeight} key={h.id} item={h} />),
+          ...history.map((h) => (
+            <HistoryItemDisplay
+              availableTerminalHeight={availableTerminalHeight}
+              key={h.id}
+              item={h}
+              isPending={false}
+            />
+          )),
         ]}
       >
         {(item) => item}
@@ -272,12 +277,13 @@ export const App = ({
             // TODO(taehykim): It seems like references to ids aren't necessary in
             // HistoryItemDisplay. Refactor later. Use a fake id for now.
             item={{ ...pendingHistoryItem, id: 0 }}
+            isPending={true}
           />
         </Box>
       )}
       {showHelp && <Help commands={slashCommands} />}
 
-      <Box flexDirection="column" ref={footerRef}>
+      <Box flexDirection="column" ref={mainControlsRef}>
         {startupWarnings.length > 0 && (
           <Box
             borderStyle="round"
--- a/packages/cli/src/ui/App.tsx
+++ b/packages/cli/src/ui/App.tsx
@@ -4,8 +4,8 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { useCallback, useEffect, useMemo, useState } from 'react';
-import { Box, Static, Text, useStdout } from 'ink';
+import { useCallback, useEffect, useMemo, useState, useRef } from 'react';
+import { Box, DOMElement, measureElement, Static, Text, useStdout } from 'ink';
 import { StreamingState, type HistoryItem } from './types.js';
 import { useGeminiStream } from './hooks/useGeminiStream.js';
 import { useLoadingIndicator } from './hooks/useLoadingIndicator.js';
@@ -46,6 +46,7 @@ export const App = ({
   startupWarnings = [],
 }: AppProps) => {
   const { history, addItem, clearItems } = useHistory();
+  const [staticNeedsRefresh, setStaticNeedsRefresh] = useState(false);
   const [staticKey, setStaticKey] = useState(0);
   const refreshStatic = useCallback(() => {
     setStaticKey((prev) => prev + 1);
@@ -55,7 +56,8 @@ export const App = ({
   const [debugMessage, setDebugMessage] = useState<string>('');
   const [showHelp, setShowHelp] = useState<boolean>(false);
   const [themeError, setThemeError] = useState<string | null>(null);
-
+  const [availableTerminalHeight, setAvailableTerminalHeight] =
+    useState<number>(0);
   const {
     isThemeDialogOpen,
     openThemeDialog,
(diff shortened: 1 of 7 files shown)


packages/cli/src/ui/App.tsx
```diff
@@ -194,12 +196,51 @@ export const App = ({
...
   const { stdout } = useStdout();
   const terminalWidth = stdout?.columns ?? 80;
+  const terminalHeight = stdout?.rows ?? 24;
+  const footerRef = useRef<DOMElement>(null);
+  const pendingHistoryItemRef = useRef<DOMElement>(null);
+
   // Calculate width for suggestions, leave some padding
   const suggestionsWidth = Math.max(60, Math.floor(terminalWidth * 0.8));
 
+  useEffect(() => {
```
NTaylorMullen, line 205: @KeijiBranshi @thkim1011 So I couldn't find a better way to measure terminal height, propagate it effectively and react. to pending items that exceed maximum height. If there's a better way to do this I'm alllll ears

packages/cli/src/ui/App.tsx
```diff
@@ -194,12 +196,51 @@ export const App = ({
...
+    // in Ink which we have a PR out to fix: https://github.com/vadimdemedes/ink/pull/717
+    if (pendingItemDimensions.height > availableTerminalHeight) {
+      setStaticNeedsRefresh(true);
+    }
+  }, [pendingHistoryItem, availableTerminalHeight, streamingState]);
+
+  useEffect(() => {
+    if (streamingState === StreamingState.Idle && staticNeedsRefresh) {
+      setStaticNeedsRefresh(false);
+      refreshStatic();
```
NTaylorMullen, line 237: So the gist is, in order to work around the bug that Tae has out for fixes in Ink (things duplicating) we need to occasionally fully re-render the static list. I decided that a proper way to accomplish this is to refresh that static list a **single** time IF and only IF a pending item at some point exceeds the available terminal height. It's not a perfect solution and may even flash more than we like (which we can adjust in the future) but removes most duplication.

I have a hunch that we may end up needing to special case pending items to only account for non-confirmation based ones and potentially let those dupliate. Or maybe not, we'll see.

packages/cli/src/ui/App.tsx
```diff
@@ -230,145 +271,151 @@ export const App = ({
               key={'history-' + historyItem.id}
               item={historyItem}
               onSubmit={submitQuery}
+              availableTerminalHeight={availableTerminalHeight}
             />
           );
         }}
       </Static>
       {pendingHistoryItem && (
-        <HistoryItemDisplay
```
NTaylorMullen, line 228: All this looks more intimidating than it is. Ensure you're reviewing with [whitespace disabled](https://github.com/google-gemini/gemini-cli/pull/358/files?diff=unified&w=1). Basically I just wrapped everything in a `<Box>` so I could capture its item ref so I could then measure it to determine available terminal height. That resulted in me indenting everything 😄

packages/cli/src/ui/App.tsx
thkim1011: So while I was digging around the Ink codebase last week, I noticed that it was able to detect terminal resize with the following:

https://github.com/vadimdemedes/ink/blob/991aaa54eae39e1ad6d1964e747ea936d5a4ad12/src/ink.tsx#L110

Found docs for the 'resize' event here: https://nodejs.org/api/tty.html#event-resize. One thing we could do is attach a resize callback in the `useEffect` hook which will call `setAvailableTerminalHeight`. There's a corresponding method for removing event listeners that we can use for the useEffect cleanup: https://nodejs.org/api/events.html#emitterremovelistenereventname-listener.

packages/cli/src/ui/App.tsx
```diff
@@ -194,12 +196,51 @@ export const App = ({
 
   // --- Render Logic ---
 
-  // Get terminal width
+  // Get terminal dimensions
+
   const { stdout } = useStdout();
   const terminalWidth = stdout?.columns ?? 80;
+  const terminalHeight = stdout?.rows ?? 24;
```
KeijiBranshi: question: we have a dedicated hook for terminal size, one that updates stateful variables for React to react to: https://github.com/google-gemini/gemini-cli/blob/46e955897ecc2e772bf9f5be2c2766cfe2c0213a/packages/cli/src/ui/hooks/useTerminalSize.ts#L11-L32

Any chance this could be managed in that hook?

packages/cli/src/ui/App.tsx
```diff
@@ -194,12 +196,51 @@ export const App = ({
...
   const suggestionsWidth = Math.max(60, Math.floor(terminalWidth * 0.8));
 
+  useEffect(() => {
+    const staticExtraHeight = /* margins and padding */ 3;
+    const fullFooterMeasurement = measureElement(footerRef.current!);
+    const fullFooterHeight = fullFooterMeasurement.height;
+
+    setAvailableTerminalHeight(
+      terminalHeight - fullFooterHeight - staticExtraHeight,
+    );
```
KeijiBranshi: suggestion: `availableTerminalHeight` doesn't need to be stateful. it looks like it's always derived from `terminalHeight`, meaning it should just be a `useMemo`:

```typescript
const availableTerminalHeight = useMemo(() => 
  terminalHeight - measureElement(footerRef.current!) - 3
}, [terminalHeight])
```

packages/cli/src/ui/App.tsx
```diff
@@ -194,12 +196,51 @@ export const App = ({
...
+
+    // If our pending history item happens to exceed the terminal height we will most likely need to refresh
+    // our static collection to ensure no duplication or tearing. This is currently working around a core bug
+    // in Ink which we have a PR out to fix: https://github.com/vadimdemedes/ink/pull/717
+    if (pendingItemDimensions.height > availableTerminalHeight) {
+      setStaticNeedsRefresh(true);
+    }
+  }, [pendingHistoryItem, availableTerminalHeight, streamingState]);
+
+  useEffect(() => {
```
KeijiBranshi, line 234: suggestion: I think this could be merged with the useEffect above. No need for the intermediate `staticNeedsRefresh` state:

```
  useEffect(() => {
    if (!pendingHistoryItem) {
      return;
    }

    const pendingItemDimensions = measureElement(
      pendingHistoryItemRef.current!,
    );

    const outOfBounds = pendingItemDimensions.height > availableTerminalHeight
    const idle = streamingState === StreamingState.Idle
    if (outOfBounds & idle) {
      refreshStatic()
    }
  }, [pendingHistoryItem, availableTerminalHeight, streamingState]);
```

packages/cli/src/ui/App.tsx
```diff
@@ -230,145 +271,151 @@ export const App = ({
...
-          flexDirection="column"
-        >
-          {startupWarnings.map((warning, index) => (
-            <Text key={index} color={Colors.AccentYellow}>
-              {warning}
-            </Text>
-          ))}
-        </Box>
-      )}
+      <Box flexDirection="column" ref={footerRef}>
```
KeijiBranshi: thought(non-blocking): it's a little misleading to provide this the `footerRef` since there's clearly more elements (including a literal `Footer`) below. At the moment though, I can't think of anything better 🤷

packages/cli/src/ui/components/messages/ToolGroupMessage.tsx
```diff
@@ -16,19 +16,24 @@ interface ToolGroupMessageProps {
...
   onSubmit,
+  availableTerminalHeight,
 }) => {
   const hasPending = !toolCalls.every(
     (t) => t.status === ToolCallStatus.Success,
   );
   const borderColor = hasPending ? Colors.AccentYellow : Colors.AccentPurple;
 
+  const staticHeight = /* border */ 2 + /* marginBottom */ 1;
+  availableTerminalHeight -= staticHeight;
```
KeijiBranshi: suggestion: assign this to a new const or just pass `availableTerminalHeight - staticHeight` directly into the `ToolMessage` below.

I'm a little surprised React didn't complain about this: https://screenshot.googleplex.com/B8cjpAF4ScE3Dp3

General rule of thumb in React is to maintain immutability of variables. The only way to invoke change is via `setState` somewhere.

packages/cli/src/ui/components/messages/ToolMessage.tsx
```diff
@@ -12,14 +12,38 @@ import { DiffRenderer } from './DiffRenderer.js';
...
   name,
   description,
   resultDisplay,
   status,
+  availableTerminalHeight,
 }) => {
   const statusIndicatorWidth = 3;
   const hasResult = resultDisplay && resultDisplay.toString().trim().length > 0;
+  const staticHeight = /* Header */ 1;
+  availableTerminalHeight -= staticHeight;
```
KeijiBranshi: suggestion: same as previous comment.

packages/cli/src/ui/components/messages/ToolMessage.tsx
```diff
@@ -56,15 +80,22 @@ export const ToolMessage: React.FC<IndividualToolCallDisplay> = ({
...
       {hasResult && (
         <Box paddingLeft={statusIndicatorWidth} width="100%">
-          <Box flexDirection="row">
-            {/* Use default text color (white) or gray instead of dimColor */}
-            {typeof resultDisplay === 'string' && (
+          <Box flexDirection="column">
+            {typeof displayableResult === 'string' && (
               <Box flexDirection="column">
-                {MarkdownRenderer.render(resultDisplay)}
+                {MarkdownRenderer.render(displayableResult)}
```
KeijiBranshi: question: i recall we discussed making this it's own components `<MarkdownRenderer content={displayableResult} />`. Did that not work here?

Somewhat subjective, but I think there's merit to avoiding these sort of nested render functions: https://alexkondov.com/tao-of-react/#avoid-nested-render-functions

It helps react do the work of avoiding re-renders when the props don't change. In this case, markdown is _always_ re-rendering regardless of `displayableResult`

packages/cli/src/ui/components/messages/ToolMessage.tsx
```diff
@@ -56,15 +80,22 @@ export const ToolMessage: React.FC<IndividualToolCallDisplay> = ({
...
               </Box>
             )}
-            {typeof resultDisplay === 'object' && (
-              <DiffRenderer diffContent={resultDisplay.fileDiff} />
+            {typeof displayableResult === 'object' && (
+              <DiffRenderer diffContent={displayableResult.fileDiff} />
+            )}
+            {hiddenLines > 0 && (
```
KeijiBranshi, line 95: nitpick: 

```suggestion
            {hiddenLines && (
```

KeijiBranshi (approved):
approving to unblock. but left a few comments.

packages/cli/src/ui/App.tsx
NTaylorMullen: Ah so if we resize to be able to account for it. If we assume that the terminal stays a static height for the entire duration of the session would that be usable in these effects? Asking because I'm trying to see if it makes sense to group that with: https://b.corp.google.com/issues/417440848

packages/cli/src/ui/App.tsx
NTaylorMullen: GC says yes 😄

packages/cli/src/ui/App.tsx
NTaylorMullen: Unless I'm missing something wouldn't that be racey depending on when the backend system decided to move the pending item from pending -> static?

packages/cli/src/ui/App.tsx
NTaylorMullen: That's a great point. I'll rename the variable to something more specific like `mainControlsRef`

packages/cli/src/ui/App.tsx
KeijiBranshi: Sorry, I don't see the race that you're alluding to. 

This effect will (_should_?) get triggered when `pendingHistoryItem` changes (ie when its moved from pending -> static). In fact, a change in any of the 3 deps `[pendingHistoryItem, availableTerminalHeight, streamingState]` will do this. So you can assume this will _always_ run with the latest values

packages/cli/src/ui/components/messages/ToolGroupMessage.tsx
NTaylorMullen: Great callout, can do!

packages/cli/src/ui/components/messages/ToolMessage.tsx
NTaylorMullen: Have a [different PR](https://github.com/google-gemini/gemini-cli/pull/359) for that 😄 . Didn't end up being required for this.

packages/cli/src/ui/App.tsx
NTaylorMullen: The race I'm worried about is the case when the pending item is moved from pending -> static before streaming state is idle. Meaning it will trigger the effect, the refreshStatic check wont fire (because things aren't idle yet) and then the next time the effect triggers because of a streaming state change the pending history item will be `null` OR potentially wont be large enough to be considered out of bounds.

packages/cli/src/ui/components/messages/ToolMessage.tsx
NTaylorMullen: Actually fun note, we can't do this 😄 . It got treated as rendered content (thanks JavaScript coercing to string) and then broke Ink:
image

packages/cli/src/ui/App.tsx
NTaylorMullen: Merging this in for now but will react to any feedback!

PR #359, opened 2025-05-15
Refactor: Replace MarkdownRenderer with MarkdownDisplay component

NTaylorMullen:
- This commit refactors the Markdown rendering logic within the CLI UI. The existing `MarkdownRenderer.tsx` class-based component has been replaced with a new functional component `MarkdownDisplay.tsx`.
- The `MarkdownDisplay` component is a React.memoized component for improved performance and maintains the same core Markdown parsing and rendering capabilities.


packages/cli/src/ui/utils/MarkdownDisplay.tsx
```diff
@@ -0,0 +1,301 @@
...
+}
+
+function MarkdownDisplayComponent({
+  text,
+}: MarkdownDisplayProps): React.ReactElement {
+  if (!text) return <></>;
+
+  const lines = text.split('\n');
+  const headerRegex = /^ *(#{1,4}) +(.*)/;
+  const codeFenceRegex = /^ *(`{3,}|~{3,}) *(\S*?) *$/;
```
sethtroisi: the end of this feels silly " *(\S*?) *" is any number of spaces, any number of non whitespace, any length of spaces" the only strings this doesn't match are those containing tabs.

If you pulled this from somewhere else keep it, if you don't want to mess with regex keep it, if you want consider changing it for just `/^ *(`{3,}|~{3,}) *(\w*?) *$/` which allows a single word like bash or shell.

sethtroisi (approved):

packages/cli/src/ui/utils/MarkdownDisplay.tsx
```diff
@@ -0,0 +1,301 @@
...
+import { Colors } from '../colors.js';
+import { colorizeCode } from './CodeColorizer.js';
+
+interface MarkdownDisplayProps {
+  text: string;
+}
+
+function MarkdownDisplayComponent({
+  text,
+}: MarkdownDisplayProps): React.ReactElement {
```
KeijiBranshi: nitpick(non-blocking): prefer `React.FC` to declare component types rather than referencing `React.ReactElement` directly. Also, no need to say `*Component`, since it's implied (the `C` in `React.FC` stands for 'component').

```suggestion
const MarkdownDisplay: React.FC<MarkdownDisplayProps> = ({ text }) => {
```

packages/cli/src/ui/utils/MarkdownDisplay.tsx
```diff
@@ -0,0 +1,301 @@
...
+      _renderCodeBlock(`line-eof`, codeBlockContent, codeBlockLang),
+    );
+  }
+
+  return <>{contentBlocks}</>;
+}
+
+// Helper functions (adapted from static methods of MarkdownRenderer)
+
+function _renderInline(text: string): React.ReactNode[] {
```
KeijiBranshi: suggestion: for the same/similar reason that `MarkdownDisplay` was refactored to be a component, this should also probably be a component:

```suggestion
interface Inline {
  text: string
}
const Inline: React.FC<Inline> = ({ text }) => {
```

packages/cli/src/ui/utils/MarkdownDisplay.tsx
```diff
@@ -0,0 +1,301 @@
...
+  }
+
+  return nodes.filter((node) => node !== null);
+}
+
+function _renderCodeBlock(
+  key: string,
+  content: string[],
+  lang: string | null,
+): React.ReactNode {
```
KeijiBranshi: suggestion: same as previous comment

```suggestion
interface CodeBlock {
  key: string;
  content: string[];
  lang?: string
}
const CodeBlock: React.FC<CodeBlock> = ({
  key,
  content,
  lang,
}) => {
```

packages/cli/src/ui/utils/MarkdownDisplay.tsx
```diff
@@ -0,0 +1,301 @@
...
+  );
+}
+
+function _renderListItem(
+  key: string,
+  text: string,
+  type: 'ul' | 'ol',
+  marker: string,
+  leadingWhitespace: string = '',
+): React.ReactNode {
```
KeijiBranshi: suggestion: same as previous

```suggestion
interface ListItem {
  key: string;
  text: string;
  type: 'ul' | 'ol';
  marker: string;
  leadingWhitespace?: string;
}
const ListItem: React.FC<ListItem> = ({
  key,
  text,
  type,
  marker,
  leadingWhitespace = '',
}) => {
```

packages/cli/src/ui/utils/MarkdownDisplay.tsx
```diff
@@ -0,0 +1,301 @@
...
+        }
+      }
+    }
+  });
+
+  if (inCodeBlock) {
+    contentBlocks.push(
+      _renderCodeBlock(`line-eof`, codeBlockContent, codeBlockLang),
+    );
+  }
```
KeijiBranshi: suggestion(non-blocking): might be nice to offload this to a library. i found this one that comes with a markdown lexer:

https://marked.js.org/using_pro#lexer

packages/cli/src/ui/utils/MarkdownDisplay.tsx
```diff
@@ -0,0 +1,301 @@
...
+            <Text key={key}>
+              {linkText}
+              <Text color={Colors.AccentBlue}> ({url})</Text>
+            </Text>
+          );
+        }
+      } else if (
+        fullMatch.startsWith('<u>') &&
+        fullMatch.endsWith('</u>') &&
+        fullMatch.length > 6
```
KeijiBranshi: thought(non-blocking): there's an eslint rule https://eslint.org/docs/latest/rules/no-magic-numbers

we have a lot of magic numbers sprinkled throughout, so I'm wondering if it'd be good for us to go through, do a sweep, then enable this rule.

KeijiBranshi (approved):

packages/cli/src/ui/utils/MarkdownDisplay.tsx
NTaylorMullen: Can totally do that, for my own education what's the driver for that preference?

packages/cli/src/ui/utils/MarkdownDisplay.tsx
NTaylorMullen: This is 100% Vibe coded 😄 . Doesn't surprise me it's a bit nonsensical lol. Love the catch, will update!

packages/cli/src/ui/utils/MarkdownDisplay.tsx
NTaylorMullen: Totally agree. Fun history: I actually did this initially but ended up hitting some hiccups that I didn't want to spend time fixing SO vibe coded this renderer from 0 quickly. Once we have more time to invest in this area and cross-check all the implications I think leaning on a proper library is the right choice.

Filed for later: https://b.corp.google.com/issues/418110177

packages/cli/src/ui/utils/MarkdownDisplay.tsx
NTaylorMullen: Ya good callout. GC handled this like a champ 😄

packages/cli/src/ui/utils/MarkdownDisplay.tsx
NTaylorMullen: I personally love that; however, the original eslint rules were from g3. We should probably think of our stance on how aligned we want to be there. I'll start a thread in the dev crew chat.

packages/cli/src/ui/utils/MarkdownDisplay.tsx
NTaylorMullen: Fun note, GC found this and the list and suggested the same 😄

packages/cli/src/ui/utils/MarkdownDisplay.tsx
KeijiBranshi: it's just a nicely packaged `type` provided by React to ensure your Functional Component type signature is correct. it handles the `Prop` type usage _and_ the return type for you. 

imo, there's more room for subtle mistakes when piecing it together manually. nothing crazy, but just saves a few minutes of confusion (especially for newer React devs). For example, given the signature:

```typescript
interface Props {
  foo: string;
  bar: number;
}
function Component(props: Props): React.ReactElement {}
```

1. props _must_ be a single object, but there's no guard rail telling you that when you defined the function like the above example. See http://screen/55gJXRsypsuP5fc

2. `React.ReactElement` is a source of confusion for some (mostly its conflation with `React.ReactNode` and `JSX.Element`. See [here](https://stackoverflow.com/questions/58123398/when-to-use-jsx-element-vs-reactnode-vs-reactelement). But using `React.FC<Props>` handles that distinction for us.

NTaylorMullen: Ahhh I see, that makes it a lot more clear. Thank you for the background!

PR #370, opened 2025-05-15
refactor: shorten static history section code

KeijiBranshi:
just felt a little superfluous to check for 'header'. also the keys don't need to be strings/human readable. they're there for React to track.


packages/cli/src/ui/App.tsx
```diff
@@ -212,25 +212,17 @@ export const App = ({
...
-
-          const historyItem = item as HistoryItem;
-          return (
-            <HistoryItemDisplay
-              key={'history-' + historyItem.id}
-              item={historyItem}
-            />
-          );
-        }}
+      <Static
```
NTaylorMullen, line 215: Clever 😄

PR #260, opened 2025-05-06
Add sqlite database for saving commands

sethtroisi:
This is going to be used to log user input, model requests, and model response

PR #371, opened 2025-05-15
feat: Implement CLI and model memory management

allenhutchison:
This commit introduces a comprehensive memory management feature, allowing both users via CLI commands and the AI model itself to store and retrieve information persistently.

Key changes include:

CLI Enhancements:
- Refactored /showmemory and /refreshmemory into a unified /memory command structure.
- Added new CLI commands:
  - /memory add <text>: Saves a new piece of information.
  - /memory delete_last: Removes the most recently added memory.
  - /memory delete_all_added: Clears all memories added via this system.
- Memory is stored in ~/.gemini/GEMINI.md under a dedicated "## Gemini Added Memories" Markdown section, with each memory as a bullet point.
- Updated slashCommandProcessor.ts to handle new command syntax and dispatch to memory utility functions.
- Created memoryUtils.ts to encapsulate file I/O logic for the Markdown-based memory store.

Server-Side Tool:
- Introduced a new saveMemory tool (memoryTool.ts) enabling the AI model to save facts.
- The tool appends facts to the same ~/.gemini/GEMINI.md file under the "## Gemini Added Memories" section.
- Added tool schema (memoryTool.json) and description for the model (memoryTool.md).
- Registered the MemoryTool in the tool-registry.ts.
- Updated system prompts in prompts.ts to instruct the model on how and when to use the saveMemory tool.

Testing:
- Added comprehensive unit tests for the new CLI slash command logic in slashCommandProcessor.test.ts.
- Added unit tests for the server-side MemoryTool in memoryTool.test.ts, mocking file system operations.
- Updated snapshot tests for prompts.test.ts.snap due to changes in the system prompt.

All changes have been verified with the project's build, test, type-checking, and linting procedures.


packages/cli/src/config/memoryUtils.ts
```diff
@@ -0,0 +1,210 @@
...
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import * as fs from 'fs/promises';
+import * as path from 'path';
+import { homedir } from 'os';
+
+const GEMINI_CONFIG_DIR = '.gemini';
```
NTaylorMullen: This should be (or re-used) somewhere more centralized IMO

packages/cli/src/config/memoryUtils.ts
```diff
@@ -0,0 +1,210 @@
...
+export function getGlobalMemoryFilePath(): string {
+  return path.join(homedir(), GEMINI_CONFIG_DIR, GEMINI_MD_FILENAME);
+}
+
+/**
+ * Ensures proper newline separation before appending content.
+ */
+function ensureNewlineSeparation(currentContent: string): string {
+  if (currentContent.length === 0) return '';
+  if (currentContent.endsWith('\n\n') || currentContent.endsWith('\r\n\r\n'))
```
NTaylorMullen: Looking out for windows 👏

packages/cli/src/config/memoryUtils.ts
```diff
@@ -0,0 +1,210 @@
...
+    return '\n';
+  return '\n\n';
+}
+
+/**
+ * Adds a new memory entry to the global GEMINI.md file under the specified header.
+ */
+export async function addMemoryEntry(text: string): Promise<void> {
+  const filePath = getGlobalMemoryFilePath();
+  const newMemoryItem = `- ${text.trim()}`;
```
NTaylorMullen: Should trim `-` from the beginning of text just in case

packages/cli/src/config/memoryUtils.ts
```diff
@@ -0,0 +1,210 @@
...
+
+  try {
+    await fs.mkdir(path.dirname(filePath), { recursive: true });
```
NTaylorMullen: IIRC fs.mkdir throws if the filePath already exists. I.e. if you write two memory items is everything ok?

packages/cli/src/config/memoryUtils.ts
```diff
@@ -0,0 +1,210 @@
...
+  }
+}
+
+/**
+ * Deletes the last added memory entry from the "Gemini Added Memories" section.
+ */
+export async function deleteLastMemoryEntry(): Promise<boolean> {
+  const filePath = getGlobalMemoryFilePath();
+  try {
+    let content = await fs.readFile(filePath, 'utf-8');
```
NTaylorMullen, line 45: Should we be gracefully returning if the file doesn't exist? There's a few file path checks we may want to be more graceful about in this file. I wont leave more comments though 😄

packages/cli/src/config/memoryUtils.ts
```diff
@@ -0,0 +1,210 @@
...
+
+    await fs.writeFile(filePath, content, 'utf-8');
+    return true;
+  } catch (error) {
+    if (error instanceof Error && 'code' in error && error.code === 'ENOENT') {
+      return false;
+    }
+    console.error(`Error deleting last memory entry from ${filePath}:`, error);
+    throw new Error(
+      `Failed to delete last memory entry: ${error instanceof Error ? error.message : String(error)}`,
```
NTaylorMullen: We have a [getErrorMessage](https://github.com/google-gemini/gemini-cli/blob/8d9e1118c6035185c54e4cd1f3162d3c6bb6e254/packages/server/src/utils/errors.ts#L11) helper which could be re-used here and elsewhere you try and extract the error message

packages/cli/src/config/memoryUtils.ts
```diff
@@ -0,0 +1,210 @@
...
+    );
+  }
+}
+
+/**
+ * Deletes all added memory entries (the entire "Gemini Added Memories" section).
+ */
+export async function deleteAllAddedMemoryEntries(): Promise<number> {
```
NTaylorMullen, line 112: When creating the memory section should we be having a comment indicating that users shouldn't modify the section or mention that it's managed by GC?

packages/cli/src/ui/hooks/slashCommandProcessor.ts
```diff
@@ -6,22 +6,27 @@
...
 import { type PartListUnion } from '@google/genai';
-import { getCommandFromQuery } from '../utils/commandUtils.js';
+// import { getCommandFromQuery } from '../utils/commandUtils.js';
 import { UseHistoryManagerReturn } from './useHistoryManager.js';
 import { Config } from '@gemini-code/server'; // Import Config
 import { Message, MessageType, HistoryItemWithoutId } from '../types.js'; // Import Message types
 import {
   createShowMemoryAction,
-  SHOW_MEMORY_COMMAND_NAME,
+  // SHOW_MEMORY_COMMAND_NAME, // Removed as unused due to backward compatibility removal
```
```suggestion
```

packages/cli/src/ui/hooks/slashCommandProcessor.ts
```diff
@@ -6,22 +6,27 @@
...
 } from './useShowMemoryCommand.js';
-import { REFRESH_MEMORY_COMMAND_NAME } from './useRefreshMemoryCommand.js'; // Only import name now
-import process from 'node:process'; // For process.exit
+// import { REFRESH_MEMORY_COMMAND_NAME } from './useRefreshMemoryCommand.js'; // Removed as unused
```
```suggestion
```

packages/cli/src/ui/hooks/slashCommandProcessor.ts
```diff
@@ -78,25 +168,45 @@ export const useSlashCommandProcessor = (
...
-        action: performMemoryRefresh, // Use the passed in function
-      },
-      {
-        name: SHOW_MEMORY_COMMAND_NAME.substring(1), // Remove leading '/'
-        description: 'Displays the current hierarchical memory content.',
-        action: showMemoryAction,
+        name: 'memory',
+        description:
+          'Manage memory. Usage: /memory <show|refresh|add|delete_last|delete_all_added> [text for add]',
+        action: (mainCommand, subCommand, args) => {
```
NTaylorMullen, line 114: Love the separation you've built in here of subcommands and args ❤️

packages/cli/src/ui/hooks/slashCommandProcessor.ts
```diff
@@ -108,8 +218,12 @@ export const useSlashCommandProcessor = (
       refreshStatic,
       openThemeDialog,
       clearItems,
-      performMemoryRefresh, // Add to dependencies
+      performMemoryRefresh,
       showMemoryAction,
+      addMemoryAction,
+      deleteLastMemoryAction,
+      deleteAllAddedMemoryAction,
+      addMessage, // addMessage is used by memory actions
```
```suggestion
      addMessage,
```

packages/cli/src/ui/hooks/slashCommandProcessor.ts
```diff
@@ -120,36 +234,57 @@ export const useSlashCommandProcessor = (
...
-      // Add user message to history only if it's not a silent command or handled internally
-      // For now, adding all slash commands to history for transparency.
+
       addItem({ type: MessageType.USER, text: trimmed }, userMessageTimestamp);
 
+      let subCommand: string | undefined;
+      let args: string | undefined;
+
+      const commandToMatch = (() => {
+        if (trimmed.startsWith('?')) {
```
NTaylorMullen, line 177: I'm ok dropping `?` entirely if you want

packages/cli/src/ui/hooks/slashCommandProcessor.ts
```diff
@@ -120,36 +234,57 @@ export const useSlashCommandProcessor = (
...
       const userMessageTimestamp = Date.now();
-      // Add user message to history only if it's not a silent command or handled internally
-      // For now, adding all slash commands to history for transparency.
+
       addItem({ type: MessageType.USER, text: trimmed }, userMessageTimestamp);
 
+      let subCommand: string | undefined;
+      let args: string | undefined;
+
+      const commandToMatch = (() => {
```
NTaylorMullen, line 176: Maybe I'm missing it, why wrap in a lambda and insta invoke?

packages/cli/src/ui/hooks/slashCommandProcessor.ts
```diff
@@ -120,36 +234,57 @@ export const useSlashCommandProcessor = (
...
+      });
 
       return true;
     },
-    [addItem, slashCommands],
+    [
+      addItem,
+      slashCommands,
+      addMessage,
+      // performMemoryRefresh, // Removed based on lint suggestion and backward compat removal
```
```suggestion
```

packages/cli/src/ui/hooks/slashCommandProcessor.ts
```diff
@@ -120,36 +234,57 @@ export const useSlashCommandProcessor = (
...
+      // showMemoryAction,     // Removed based on lint suggestion and backward compat removal
```
```suggestion
```

packages/server/src/core/prompts.ts
```diff
@@ -82,6 +82,7 @@ Rigorously adhere to existing project conventions when reading or modifying code
 -   **Command Execution:** Use the '${ShellTool.Name}' tool for running shell commands, remembering the safety rule to explain modifying commands first.
 -   **Background Processes:** Use background processes (via \`&\`) for commands that are unlikely to stop on their own, e.g. \`node server.js &\`. If unsure, ask the user.
 -   **Interactive Commands:** Try to avoid shell commands that are likely to require user interaction (e.g. \`git rebase -i\`). Use non-interactive versions of commands (e.g. \`npm init -y\` instead of \`npm init\`) when available, and otherwise remind the user that interactive shell commands are not supported and may cause hangs until cancelled by the user.
+-   **Remembering Facts:** Use the 'saveMemory' tool to remember specific facts or pieces of information when the user explicitly asks, or when they state a clear, concise piece of information that seems important for future interactions. If unsure whether to save something, you can ask the user, "Should I remember that for you?"
```
NTaylorMullen: Will be curious to see how the "seems important for future interactions" will come into play. It might be too broad. ChatGPT has more specific verbiage around improving the tools understanding of future conversations etc. (feel free to ask ChatGPT for its `bio` tool schema if you're curious).

All that said it might be worth indicating that it should only be used to remember user specific information so the model doesn't start adding project memories to your global file.

packages/server/src/tools/memoryTool.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "saveMemory",
```
NTaylorMullen: Becaues of bundling I'm suggesting we go away from represent the schema in json (otherwise we need to maintain a lot of static assets like this along side the bundle). Would it be possible to inline?

packages/server/src/tools/memoryTool.md
```diff
@@ -0,0 +1,20 @@
+# Tool: saveMemory
```
NTaylorMullen: ditto

packages/server/src/tools/memoryTool.ts
```diff
@@ -0,0 +1,147 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import * as fs from 'fs/promises';
+import * as path from 'path';
+import { homedir } from 'os';
+import { BaseTool, ToolResult } from './tools.js';
+import memoryToolSchemaData from './memoryTool.json' with { type: 'json' };
+
+const GEMINI_CONFIG_DIR = '.gemini';
```
NTaylorMullen: Should have a central location for all these

packages/server/src/tools/memoryTool.ts
```diff
@@ -0,0 +1,147 @@
...
+const GEMINI_MD_FILENAME = 'GEMINI.md';
+const MEMORY_SECTION_HEADER = '## Gemini Added Memories';
+
+interface SaveMemoryParams {
+  fact: string;
+}
+
+function getGlobalMemoryFilePath(): string {
```
NTaylorMullen, line 56: Hmmm, seems like a lot of the memoryUtil.ts is duplicated here. Is there a chance for unification?

packages/server/src/tools/memoryTool.ts
```diff
@@ -0,0 +1,147 @@
...
+  //     return 'Parameter "fact" must be a non-empty string.';
+  //   }
+  //   return null; // All good
+  // }
+
+  // Optional: Override getDescription for a custom pre-execution message
+  // getDescription(params: SaveMemoryParams): string {
+  //   return `Will attempt to save the fact: "${params.fact}" to long-term memory.`;
+  // }
+
```
NTaylorMullen, line 152: Note we could also consider treating this as an "edit a GEMINI.md" file operation and present a confirmation. ChatGPT will auto remember things which is super convenient but I wonder how often folks will want to do that with a file that they check in regularly / maintain.

packages/server/src/tools/tool-registry.ts
```diff
@@ -127,7 +128,9 @@ export class ToolRegistry {
   private tools: Map<string, Tool> = new Map();
   private mcpClient: Client | null = null;
 
-  constructor(private readonly config: Config) {}
+  constructor(private readonly config: Config) {
+    this.registerTool(new MemoryTool());
```
NTaylorMullen: Any reason to register this in the ctor vs. with the other tools registrations?

packages/cli/src/config/memoryUtils.ts
olcan: this is also defined in settings.ts and i’ve imported it from there in a few places. i’m  not sure if calling it settings or config dir is future looking enough if we put state there like histories etc

packages/cli/src/ui/hooks/slashCommandProcessor.test.ts
```diff
@@ -0,0 +1,317 @@
...
+
+  beforeEach(() => {
+    mockAddItem = vi.fn();
+    mockClearItems = vi.fn();
+    mockRefreshStatic = vi.fn();
+    mockSetShowHelp = vi.fn();
+    mockOnDebugMessage = vi.fn();
+    mockOpenThemeDialog = vi.fn();
+    mockPerformMemoryRefresh = vi.fn().mockResolvedValue(undefined);
+    mockConfig = { getDebugMode: vi.fn(() => false) };
```
jacob314: you should be able to cast this to the type required by useSlashCommandProcessor rather than using any.

jacob314: change this to `{ getDebugMode: vi.fn(() => false) } as unknown as Config;`
is an acceptable solution here.

packages/cli/src/ui/hooks/slashCommandProcessor.test.ts
```diff
@@ -0,0 +1,317 @@
...
+describe('useSlashCommandProcessor', () => {
+  let mockAddItem: ReturnType<typeof vi.fn>;
+  let mockClearItems: ReturnType<typeof vi.fn>;
+  let mockRefreshStatic: ReturnType<typeof vi.fn>;
+  let mockSetShowHelp: ReturnType<typeof vi.fn>;
+  let mockOnDebugMessage: ReturnType<typeof vi.fn>;
+  let mockOpenThemeDialog: ReturnType<typeof vi.fn>;
+  let mockPerformMemoryRefresh: ReturnType<typeof vi.fn>;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  let mockConfig: any;
```
jacob314: change this to 
let mockConfig : Config;

packages/server/src/tools/memoryTool.test.ts
```diff
@@ -0,0 +1,151 @@
...
+// Mock dependencies
+vi.mock('fs/promises');
+vi.mock('os');
+
+const MEMORY_SECTION_HEADER = '## Gemini Added Memories'; // Match the constant in memoryTool.ts
+
+describe('MemoryTool', () => {
+  let memoryTool: MemoryTool;
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars, @typescript-eslint/no-explicit-any
+  let mockMkdir: any;
```
jacob314: these `any` can actually be removed and the type system seems happy and autocomplete within the file is better as there are more inferred types.

dewitt:
High-level consideration: I can see the value of storing this in a text-based, human-readable .md file for consistency with project-wide instructions, but as soon as we start treating those files for machine-read/writable structured data, as this PR does, then I expect we're eventually going to have a bad time. E.g., even this CL has to take into account OS-specific line-endings, whitespace trimming, section headers, and character encodings into account. This seems brittle enough and will surely get even worse over time as the things we want to persist get more complex and more numerous.

Would we consider a sqlite-backed structured file format instead? I see that PR #260 already opens the idea of using sqlite. Using it for memories would likely greatly simplify this code, make it less brittle, more portable, with the minor tradeoff of being slightly more opaque (but still fairly standard and introspectable and extensible). 

Food for thought.

packages/cli/src/ui/hooks/slashCommandProcessor.test.ts
jacob314: The reason this solution is slightly better is now you will get type check errors if you try to pass mockConfig somewhere that expected a type other than Config. `as unknown` isn't great though. The other option would be to more fully mocks out a Config object perhaps adding a helper method to a shared testing library that creates a suitable Config object for testing. That would avoid strange runtime errors if a test exercises code that is missing from mockConfig.

jacob314 (commented):
Added a couple comments about  how to resolve uses of `any` in the tests. Let me know if there are any more `any` I've missed.  I verified locally that the type system is satisfied with these changes.

allenhutchison:
We had the debate on Sqlite already and decided to use the md file. We'll use sqlite for conversation history. There is a lot of context in the bug. https://b.corp.google.com/issues/413419168

packages/cli/src/config/memoryUtils.ts
allenhutchison: fixed.

packages/cli/src/config/memoryUtils.ts
allenhutchison: Fixed.

packages/cli/src/config/memoryUtils.ts
allenhutchison: The fs.mkdir command with the { recursive: true } option is designed to behave like mkdir -p. This means it will create the directory if it doesn't exist, and it will not throw an error if the directory already exists. So, writing multiple memory items should be okay as the directory will be created on the first call and subsequent calls won't cause an error due to the directory already being there.

I've tested with making multiple entries.

packages/cli/src/config/memoryUtils.ts
allenhutchison: This works as expected when no file exists. For add operations a file is created for delete it just comes back and says there are no memories to delete.

packages/server/src/tools/memoryTool.ts
allenhutchison: Sorry I'm a little confused by this comment. Do you mean the memoryTool.json and memoryTool.md files? I know most of the tools define those inline. I felt like this pattern was a little easier to read and edit, but I'm happy to combine them like the other tools if you prefer. 

Or are you suggesting that .json and .md tool associated files go elsewhere in the project structure? I think that would be messy and I'd rather combine all in one file if that were the case. It would be too hard to remember to update the files in other locations IMHO.

packages/cli/src/config/memoryUtils.ts
allenhutchison: Done. We should probably move any utils that we want to be common across the server and cli packages to an independent location to make the dependencies and testing mocks more straightforward.

packages/cli/src/config/memoryUtils.ts
allenhutchison: The heading is on line 13: 
const MEMORY_SECTION_HEADER = '## Gemini Added Memories';

We could make that stronger language or we could avoid the trouble overall by removing the delete paths in this utility and only allow GC to add things.

packages/cli/src/ui/hooks/slashCommandProcessor.test.ts
allenhutchison: Thanks for this. Really helpful. I agree we need to make a pass through the codebase and create common testing mocks.

packages/cli/src/ui/hooks/slashCommandProcessor.ts
allenhutchison: Thanks. We'll need to do a clean up afterward to make suggestions work well here.

packages/cli/src/ui/hooks/slashCommandProcessor.ts
allenhutchison: Mostly just allows us to keep the additional variables scoped to the block and commandToMatch can be const. I could go either way, but GC really liked the pattern and I thought it was really clean.

packages/server/src/core/prompts.ts
allenhutchison: I've updated this prompt to make it more user specific and with guidance away from project stuff that would go in a project GEMINI.md

packages/server/src/tools/memoryTool.json
allenhutchison: Yes will do.

packages/server/src/tools/memoryTool.md
allenhutchison: ack.

packages/server/src/tools/memoryTool.ts
allenhutchison: I saw your other comments in context above. I'll inline the schema and markdown.

packages/server/src/tools/memoryTool.ts
allenhutchison: Yes, it's just a big refactor because I need to put memoryUtils in a common packages/utils directory to avoid circular dependencies. This would be a good place to move things like getError that you pointed me to earlier.

allenhutchison: ok that refactor is going to be way too big to do here. It modifies the build system and a bunch of other things. I'll look at it as a follow up.

packages/server/src/tools/tool-registry.ts
allenhutchison: No reason at all other than I was doing too many things at once yesterday.

packages/server/src/tools/memoryTool.test.ts
allenhutchison: got it. Thank you.

packages/server/src/tools/memoryTool.ts
allenhutchison: Went for a walk and decided for now to pull the addMemory function from memoryUtils into the tool. That allows for sharing from the server back up to the CLI and removes the duplicated code. The bigger refactor can come later.

packages/cli/src/config/memoryUtils.ts
NTaylorMullen: I kind of like the path of removing delete entirely. In most cases users will want to mutate some bullet points (probably not all).

packages/server/src/core/prompts.ts
```diff
@@ -82,6 +82,7 @@ Rigorously adhere to existing project conventions when reading or modifying code
...
 -   **Background Processes:** Use background processes (via \`&\`) for commands that are unlikely to stop on their own, e.g. \`node server.js &\`. If unsure, ask the user.
 -   **Interactive Commands:** Try to avoid shell commands that are likely to require user interaction (e.g. \`git rebase -i\`). Use non-interactive versions of commands (e.g. \`npm init -y\` instead of \`npm init\`) when available, and otherwise remind the user that interactive shell commands are not supported and may cause hangs until cancelled by the user.
+-   **Remembering Facts:** Use the 'saveMemory' tool to remember specific, *user-related* facts or preferences when the user explicitly asks, or when they state a clear, concise piece of information that would help personalize or streamline *your future interactions with them* (e.g., preferred coding style, common project paths they use, personal tool aliases). This tool is for user-specific information that should persist across sessions. Do *not* use it for general project context or information that belongs in project-specific \\\`GEMINI.md\\\` files. If unsure whether to save something, you can ask the user, "Should I remember that for you?"
```
NTaylorMullen:
```suggestion
-   **Remembering Facts:** Use the '${memoryTool.Name}' tool to remember specific, *user-related* facts or preferences when the user explicitly asks, or when they state a clear, concise piece of information that would help personalize or streamline *your future interactions with them* (e.g., preferred coding style, common project paths they use, personal tool aliases). This tool is for user-specific information that should persist across sessions. Do *not* use it for general project context or information that belongs in project-specific \`GEMINI.md\` files. If unsure whether to save something, you can ask the user, "Should I remember that for you?"
```

packages/server/src/tools/memoryTool.ts
```diff
@@ -0,0 +1,196 @@
...
+        description:
+          'The specific fact or piece of information to remember. Should be a clear, self-contained statement.',
+      },
+    },
+    required: ['fact'],
+  },
+};
+
+const memoryToolDescription = `
+# Tool: saveMemory
```
NTaylorMullen: Since this is the schema description you should drop this line and the `## Description` header below

packages/server/src/tools/memoryTool.ts
```diff
@@ -0,0 +1,196 @@
...
+- To remember conversational context that is only relevant for the current session.
+- To save long, complex, or rambling pieces of text. The fact should be relatively short and to the point.
+- If you are unsure whether the information is a fact worth remembering long-term. If in doubt, you can ask the user, "Should I remember that for you?"
+
+## Parameters
+
+- \`fact\` (string, required): The specific fact or piece of information to remember. This should be a clear, self-contained statement. For example, if the user says "My favorite color is blue", the fact would be "My favorite color is blue".
+`;
+
+export const GEMINI_CONFIG_DIR = '.gemini';
```
NTaylorMullen, line 48: This config dir and the gemini md filename below are probably better in the server's `config.ts` since other components load and use them (i.e. system prompt constructions)

NTaylorMullen (approved):

packages/server/src/tools/memoryTool.ts
allenhutchison: Config dir is an issue because it's also defined in the client config but we can't pull that here because client depends on server. So I'd rather keep that here until we can refactor a common const file into something like package/common/... 

For gemini md filename I considerd the server config.ts but figured this was a more likely spot for it to be found by someone looking for memory definitions.

packages/cli/src/config/memoryUtils.ts
allenhutchison: ok, I'm going to remove both the delete options.

packages/server/src/tools/memoryTool.ts
allenhutchison: done

packages/cli/src/config/memoryUtils.ts
allenhutchison: Done.

packages/server/src/tools/memoryTool.ts
NTaylorMullen: Fair enough!

PR #372, opened 2025-05-15
feat: add react best practices to gemini.md

KeijiBranshi:
Taken from https://github.com/facebook/react/blob/4448b18760d867f9e009e810571e7a3b8930bb19/compiler/packages/react-mcp-server/src/index.ts#L376C1-L441C94


allenhutchison:
What do you think of adding an attribution comment in the doc itself with the URL?

KeijiBranshi:
> What do you think of adding an attribution comment in the doc itself with the URL?

Great idea! Done in #382

PR #373, opened 2025-05-15
UI improvements for suggestions & status

miguelsolorio:
Fixes an issue where context strings were being wrapped:

**Before**
CleanShot 2025-05-15 at 16 16 49@2x

**After**
CleanShot 2025-05-15 at 16 17 44@2x

Update previous status to reduce visual noise

**Before**
CleanShot 2025-05-15 at 16 19 04@2x

**After**
CleanShot 2025-05-15 at 16 18 56@2x

Also added a bit of margin to footer for some breathing room


NTaylorMullen (approved):

NTaylorMullen:
I never considered the extra emphasis on tool calls, that's a good point though

PR #376, opened 2025-05-16
refactor: derive streaming state from tool calls and isresponding state

KeijiBranshi:
preliminary refactor to fix b/411426253

the main idea here is to derive streaming state from other factors rather than have it set by the numerous gemini event handlers. This minimizes the sources of truth, giving the several gemini event handlers 1 less state to fight over (they were all calling `setStreamingState`).

now, StreamingState.WaitingForConfirmation is derived from the active `pendingHistoryItemRef` There's probably an argument for decoupling tool confirmation state from that pending history item as well, but at least now there's a single source of truth for the state of all pending confirmations.

### DIFF
--- a/packages/cli/src/ui/hooks/shellCommandProcessor.ts
+++ b/packages/cli/src/ui/hooks/shellCommandProcessor.ts
@@ -8,7 +8,6 @@ import { exec as _exec } from 'child_process';
 import { useCallback } from 'react';
 import { Config } from '@gemini-code/server';
 import { type PartListUnion } from '@google/genai';
-import { StreamingState } from '../types.js';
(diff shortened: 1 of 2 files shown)


NTaylorMullen (approved):
I see what you mean about deriving it from pending history item being a little bit of code smell. That being said, I agree limiting sources of truth here makes a big difference. I know you marked this PR as draft but I'd say :shipit:

KeijiBranshi:
> I see what you mean about deriving it from pending history item being a little bit of code smell. That being said, I agree limiting sources of truth here makes a big difference. I know you marked this PR as draft but I'd say :shipit:

SGTM! thanks!

PR #378, opened 2025-05-16
fix: Ensure filename is available for diff rendering in write-file

NTaylorMullen:
This commit resolves a bug where the `write-file` operation could fail to render content due to a missing filename.

The fix involves:
- Ensuring `fileName` is consistently passed to `DiffRenderer.tsx` through `ToolConfirmationMessage.tsx`, `ToolMessage.tsx`, and `useGeminiStream.ts`.
- Modifying `edit.ts` and `write-file.ts` to include `fileName` in the `FileDiff` object.
- Expanding the `FileDiff` interface in `tools.ts` to include `fileName`.

Additionally, this commit enhances the diff rendering by:
- Adding syntax highlighting based on file extension in `DiffRenderer.tsx`.
- Adding more language mappings to `getLanguageFromExtension` in `DiffRenderer.tsx`.

# Before
image

# After
image

Fixes https://b.corp.google.com/issues/418125982

### DIFF
--- a/packages/cli/src/ui/components/messages/DiffRenderer.tsx
+++ b/packages/cli/src/ui/components/messages/DiffRenderer.tsx
@@ -105,6 +105,14 @@ export const DiffRenderer: React.FC<DiffRendererProps> = ({
 
   const parsedLines = parseDiffWithLineNumbers(diffContent);
 
+  if (parsedLines.length === 0) {
+    return (
+      <Box borderStyle="round" borderColor={Colors.SubtleComment} padding={1}>
+        <Text dimColor>No changes detected.</Text>
+      </Box>
+    );
+  }
+
   // Check if the diff represents a new file (only additions and header lines)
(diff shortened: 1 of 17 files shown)


packages/cli/src/ui/components/messages/DiffRenderer.tsx
```diff
@@ -233,16 +233,21 @@ const renderDiffContent = (
...
-    '.ts': 'typescript',
-    '.py': 'python',
-    '.json': 'json',
-    '.css': 'css',
-    '.html': 'html',
-    '.sh': 'bash',
-    '.md': 'markdown',
-    '.yaml': 'yaml',
-    '.yml': 'yaml',
+    js: 'javascript',
```
NTaylorMullen, line 244: This was actually just a straight up bug. The extension passed in wouldn't have `.`

PR #380, opened 2025-05-16
feat: Add support for Vertex AI and Vertex express mode

sasha-gitg:
This PR respects all the same environment variables as the Gen AI SDK.

I think there are a few ways we can do this so I want to get feedback before adding tests.

1. This approach where we set API_KEY to an empty string if Vertex AI is enabled. Gen AI client infers from the environment variables.

2. Add Vertex AI fields to ServerConfig

3. Similar to (1) but remove api_key from ServerConfig and rely on solely Gen AI SDK environment variable resolution instead of explicitly passing api_key to the client. Right now, js-genai doesn't support GEMINI_API_KEY and uses GOOGLE_API_KEY for the Gemini Developer API.


README.md
```diff
@@ -8,19 +8,42 @@ For more comprehensive documentation, please see the [full documentation here](.
...
+The Gemini CLI supports several ways to authenticate with Google's AI services. You'll need to configure **one** of the following methods:
+
+1.  **Gemini API Key:**
+
+    - Obtain your API key from Google AI Studio: [https://aistudio.google.com/app/apikey](https://aistudio.google.com/app/apikey)
+    - Set the `GEMINI_API_KEY` environment variable. You can do this temporarily in your current shell session:
+      ```bash
+      export GEMINI_API_KEY="YOUR_GEMINI_API_KEY"
+      ```
+      Or add it to your shell's configuration file (like `~/.bashrc`, `~/.zshrc`, or `~/.profile`) for persistence:
```
olcan, line 20: (sorry this is unrelated to your change and we can fix it in a separate change)
i thought we used to mention .env files here, cc @NTaylorMullen @cperry-goog @KeijiBranshi ?

olcan (approved):
LGTM

README.md
NTaylorMullen: We may have accidentally dropped that unintentionally? 😄

PR #381, opened 2025-05-16
[DRAFT - DO NOT SUBMIT] Add support for a distributable single executable application binary.

dewitt:
DO NOT SUBMIT - PR FOR DISCUSSION ONLY

This PR introduces the preliminary capability to build a single executable application (SEA) that bundles the node runtime and dependencies along with the packaged gemini cli code into a single redistributable binary for systems that do not have node installed or have a different version of node installed.

This approach is using the native node SEA capabilities under active development here: https://nodejs.org/api/single-executable-applications.html

Before finishing this work-in-progress PR, we will also likely want to make the builder itself hermetic too (e.g., dockerized) and also clean this up to make the SEA an optional part of the build flow, rather than performing the expensive and generally unnecessary `postject` step on all builds.

Node SEAs are new to me, so please offer thoughts and feedback about the viability of this approach.

CREATING PR FOR DISCUSSION ONLY


mattKorwel:

PR #383, opened 2025-05-16
Support auto wrapping of in the multiline editor.

jacob314:
Previously the editor was multiline but with annoying automatic horizontal scroll instead of auto wrapping. Switched that to support wrapping. Arrow based navigation operates on the wrapped lines so it is intuitive to users,.
Add tests and fix bugs caught by tests.
Correctly support characters such as 你好 and 🐄 that are more than one character wide. Without this support, characters like cause lines that should be wrapped not to be wrapped.

Added an example vscode launch config to debug a test.
video showing new wrapping behavior.
https://screencast.googleplex.com/cast/NDg4MDIwNjUyNjQxNDg0OHw2YWFiZTI4NC0wNA
Screenshot 2025-05-16 at 10 37 06 AM

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -22,6 +22,26 @@
       "skipFiles": ["<node_internals>/**"],
       "program": "${file}",
       "outFiles": ["${workspaceFolder}/**/*.js"]
+    },
+    {
+      "type": "node",
+      "request": "launch",
+      "name": "Debug CLI Test: text-buffer",
+      "runtimeExecutable": "npm",
+      "runtimeArgs": [
+        "run",
+        "test",
+        "-w",
+        "packages/cli",
+        "--",
+        "--inspect-brk=9229",
+        "--no-file-parallelism",
+        "${workspaceFolder}/packages/cli/src/ui/components/shared/text-buffer.test.ts"
+      ],
+      "cwd": "${workspaceFolder}",
+      "console": "integratedTerminal",
+      "internalConsoleOptions": "neverOpen",
+      "skipFiles": ["<node_internals>/**"]
     }
   ]
 }
--- a/packages/cli/src/ui/components/shared/multiline-editor.tsx
+++ b/packages/cli/src/ui/components/shared/multiline-editor.tsx
@@ -4,12 +4,13 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { useTextBuffer } from './text-buffer.js';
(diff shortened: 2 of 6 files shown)


.vscode/launch.json
```diff
@@ -22,6 +22,26 @@
       "skipFiles": ["<node_internals>/**"],
       "program": "${file}",
       "outFiles": ["${workspaceFolder}/**/*.js"]
+    },
+    {
+      "type": "node",
+      "request": "launch",
+      "name": "Debug CLI Test: text-buffer",
```
NTaylorMullen, line 29: nice! I actually just found myself wanting something like this last night

packages/cli/src/ui/components/shared/multiline-editor.tsx
```diff
@@ -202,39 +210,58 @@ export const MultilineTextEditor = ({
...
   );
 
-  const visibleLines = buffer.visibleLines;
-  const [cursorRow, cursorCol] = buffer.cursor;
-  const [scrollRow, scrollCol] = buffer.scroll;
+  const linesToRender = buffer.viewportVisualLines; // This is the subset of visual lines for display
+  const [cursorVisualRowRelative, cursorVisualColRelative] =
+    buffer.visualCursor; // This is relative to *all* visual lines
+  const scrollVisualRow = buffer.visualScrollRow;
+  const scrollHorizontalCol = buffer.scroll[1]; // Horizontal scroll is still from logical scroll
```
NTaylorMullen: Not sure I follow. Why is it `buffer.scroll[1]`?

NTaylorMullen (approved):
Didn't dig heavily into the math behind the wrapping (trusting in GC + tests)

packages/cli/src/ui/components/shared/multiline-editor.tsx
jacob314: not quite ready for review. Doing a bit more massaging to remove the horizontal scroll logic that is now obsolete.

jacob314: cleaned up. there is no more scrollHorizontalCol as it wasn't needed now that there is wrap.

.vscode/launch.json
jacob314: Would be nice if you could just run test cases and files directly from the source code. not sure why that isn't working in our project. Barring that we might just add launch configs to debug all our tests.

PR #384, opened 2025-05-16
refactor: only replace abortcontroller on user cancel

KeijiBranshi:
BUG=b/411426253

another preliminary refactor towards parallel tool confirmations. previously each tool call would create its own abort controller, but we only track one at a time. 

now, we only replace the abortcontroller when the user invokes the 'cancel' action. this way _all_ running tool calls can be cancelled by the same user input.

### DIFF
--- a/packages/cli/src/ui/hooks/useGeminiStream.ts
+++ b/packages/cli/src/ui/hooks/useGeminiStream.ts
@@ -63,7 +63,9 @@ export const useGeminiStream = (
 ) => {
   const toolRegistry = config.getToolRegistry();
   const [initError, setInitError] = useState<string | null>(null);
-  const abortController = useRef<AbortController>(new AbortController());
+  const [abortController, setAbortController] = useState<AbortController>(
+    new AbortController(),
+  );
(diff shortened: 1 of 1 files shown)


NTaylorMullen (dismissed):
If you could double check that canceling a long running tool still works with this that'd make me feel better 😄 

Prompt I use is `Sleep for 5` which will run the shell command to sleep for 5 at which point it's easy to cancel and verify.

KeijiBranshi:
> If you could double check that canceling a long running tool still works with this that'd make me feel better 😄
> 
> Prompt I use is `Sleep for 5` which will run the shell command to sleep for 5 at which point it's easy to cancel and verify.

thanks for the tip. the cancel works, but im seeing the pending history item is not cleared out until I type into the input prompt: http://go/scrcast/NTQ4NDE3NTQzMDc3ODg4MHwxMTA5ODRiZC1hNQ

ill fix that before merging

NTaylorMullen:
@KeijiBranshi this still applicable or candidate to close?

scottdensmore:
@NTaylorMullen / @KeijiBranshi can we close this?

PR #386, opened 2025-05-16
strip escape characters when pasting.

jacob314:
Escape characters completely break rendering of the query.
Screenshot 2025-05-16 at 1 12 25 PM


NTaylorMullen (approved):
Nice! and love the screenshot. Great usecase 😄

PR #394, opened 2025-05-16
[DRAFT] fix: manage items and pendingItem state in the same object

KeijiBranshi:
DIFFBASE=#384

see diffbase bug (http://go/scrcast/NTQ4NDE3NTQzMDc3ODg4MHwxMTA5ODRiZC1hNQ) where the pending item is shown twice. I think this was because the operation to move a pending item to the static history list is not atomic. It requires two separate `setState` calls, one of which is a `setRef` call under the hood. Since ref changes don't trigger re-renders, I think the UI hit the inbetween state where the pending item was added to the static history, but never refreshed the pending item to null. 

this fix/refactor moves the pendingItem state into the same stateful object as the rest of the history items. this allows us to make state changes between the two atomically.

### DIFF
--- a/packages/cli/src/ui/components/HistoryItemDisplay.tsx
+++ b/packages/cli/src/ui/components/HistoryItemDisplay.tsx
@@ -25,7 +25,7 @@ export const HistoryItemDisplay: React.FC<HistoryItemDisplayProps> = ({
   availableTerminalHeight,
   isPending,
 }) => (
-  <Box flexDirection="column" key={item.id}>
+  <Box flexDirection="column">
     {/* Render standard message types */}
     {item.type === 'user' && <UserMessage text={item.text} />}
     {item.type === 'gemini' && (
--- a/packages/cli/src/ui/App.tsx
+++ b/packages/cli/src/ui/App.tsx
@@ -46,7 +46,8 @@ export const App = ({
   cliVersion,
   startupWarnings = [],
 }: AppProps) => {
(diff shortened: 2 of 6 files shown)


packages/cli/src/ui/components/HistoryItemDisplay.tsx
```diff
@@ -25,7 +25,7 @@ export const HistoryItemDisplay: React.FC<HistoryItemDisplayProps> = ({
   availableTerminalHeight,
   isPending,
 }) => (
-  <Box flexDirection="column" key={item.id}>
+  <Box flexDirection="column">
```
NTaylorMullen, line 28: why drop the key?

NTaylorMullen (approved):
This is a beassst but I do like the separation you've made. I'd say feel free to get this in to unblock yourself. It'll also ensure you don't have to work through merge conflicts galore. Your current issue IMO is the hardest one left on the early access list so anything we can do to unblock your progress is P0 imo

KeijiBranshi:
Closing in favor of #448

packages/cli/src/ui/components/HistoryItemDisplay.tsx
KeijiBranshi: I don't think this key is necessary. I _think_ keys are only required for child components in an Array. While the calling parent component might put this in an Array 1 level up the UI tree, <Box /> (in the context of _this_ component) is not in an Array

PR #412, opened 2025-05-18
Introduce a small easter egg. Woof.

dewitt:
Also changes auto-completion and /help to skip over slash commands that don't contain a description to avoid spoiling the surprise.


olcan:
woof

NTaylorMullen (approved):
lol simple enough and even has some nice filtering bits to go along with it. Down for our first easter egg 😄

PR #413, opened 2025-05-18
ability to override core system prompt (via .gemini/system.md) and specify core tools via coreTools setting (e.g. coreTools:["ls", "GrepTool", ...]) ; added tests, but did not update docs for now

olcan:


packages/server/src/tools/memoryTool.ts
```diff
@@ -10,7 +10,7 @@ import * as path from 'path';
 import { homedir } from 'os';
 
 const memoryToolSchemaData = {
-  name: 'saveMemory',
+  name: 'save_memory',
```
NTaylorMullen, line 13: ya this was a good change

NTaylorMullen:
I don't think I'm following the `coreTools` inclusion. Isn't it just providing names of tools? How does it then invoke those toolss?

olcan:
coreTools setting just picks out a subset of core tools to be enabled (by
names), or can disable all core tools if given as []

NTaylorMullen:
> coreTools setting just picks out a subset of core tools to be enabled (by names), or can disable all core tools if given as []

Oooooh like a tools config where right now we only turning them on/off. That makes sense! I imagine we'll expand this a lot. Nice add!

PR #414, opened 2025-05-18
ability to write system prompt to file

olcan:


NTaylorMullen:
What's the usecase for this one?

olcan:
when you want to try changing system prompt, just use this to write it to
disk and edit there. it will be picked up on next request in convo, and if
we can implement backtracking it will be extremely easy to iterate on
system prompt in context, e.g. encourage use of git and retry last message

NTaylorMullen:
Ah, I think we can do this with npm watchers. I.e. hot reloading real changes from the codebase into the running terminal application. Honestly we may want to consider doing that no matter what for the dev flow.

olcan:
yeah i was hoping someone more experienced with npm watchers might enable
those in the project but i can imagine annoying complications. btw there
was a question in the claude podcast about how writing files can be unsafe
with source control. a good answer is that sometimes file changes trigger
code execution

NTaylorMullen:
Filed: https://b.corp.google.com/issues/418496392

PR #415, opened 2025-05-18
feat: Add auto-accept indicator and toggle

NTaylorMullen:
- This commit introduces a visual indicator in the CLI to show when auto-accept for tool confirmations is enabled. Users can now also toggle this setting on/off using Shift + Tab.
- This addresses user feedback for better visibility and control over the auto-accept feature, improving the overall user experience.
- This behavior is similar to Claude Code, providing a familiar experience for users transitioning from that environment.
- Added tests for the new auto indicator hook.

Demo Video: http://screencast/cast/NDc1MzI4NzA4ODcwMTQ0MHw2ZmM4MWVlNi0wNw
image

Fixes https://b.corp.google.com/issues/413740468

PR #416, opened 2025-05-18
fix: Ensure CWD and auto-accept indicator are always visible (even through confirmations)

NTaylorMullen:
- This commit addresses an issue where the Current Working Directory (CWD) and the auto-accept indicator were not consistently visible, especially when tool confirmations were displayed.
- Previously, the CWD could be hidden during tool confirmation prompts, potentially leading to confusion about the context in which Gemini CLI was operating.

## When confirmations are shown:
image

## When loading:
image

Fixes https://b.corp.google.com/issues/414289185


NTaylorMullen:

PR #418, opened 2025-05-18
fix(glob): Improve glob tool accuracy and output

NTaylorMullen:
This commit enhances the glob tool by:

- Ensuring that glob patterns are used effectively. Previously, simple file names without glob syntax (e.g., "file.ts") would only search the root directory. This change encourages more precise glob patterns (e.g., "**\/file.ts") for broader searches.
- Returning absolute file paths instead of relative paths. This provides clearer, less ambiguous output and avoids encouraging the use of relative paths in subsequent operations.
- Adding comprehensive tests for various globbing scenarios, including case sensitivity and path specifications.

These changes address an issue where the glob tool could not find an expected item when a specific path was provided without appropriate glob syntax, and improve the overall reliability and usability of the tool.

Fixes https://b.corp.google.com/issues/418486553

### DIFF
--- a/packages/server/src/tools/glob.test.ts
+++ b/packages/server/src/tools/glob.test.ts
@@ -1,0 +1,241 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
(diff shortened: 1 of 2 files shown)


NTaylorMullen:

PR #419, opened 2025-05-18
fix(shell): Improve error reporting for shell command failures

NTaylorMullen:
This commit enhances the  tool to provide more informative feedback to the user when a shell command fails, especially in non-debug mode.

Previously, if a command terminated due to a signal (e.g., SIGPIPE during a  with no upstream) or failed without producing stdout/stderr, the user would see no output, making it difficult to diagnose the issue.

Changes:
- Modified  to update the  logic.
- If a command produces no direct output but results in an error, signal, non-zero exit code, or user cancellation, a concise message indicating this outcome is now shown in .
- Utilized the existing  utility from  for consistent error message formatting, which also resolved previous TypeScript type inference issues.

This ensures users receive clearer feedback on command execution status, improving the tool's usability and aiding in troubleshooting.

Fixes https://b.corp.google.com/issues/417998119

### DIFF
--- a/packages/server/src/tools/shell.ts
+++ b/packages/server/src/tools/shell.ts
@@ -17,6 +17,7 @@ import {
   ToolConfirmationOutcome,
 } from './tools.js';
 import { SchemaValidator } from '../utils/schemaValidator.js';
+import { getErrorMessage } from '../utils/errors.js';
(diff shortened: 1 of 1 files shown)


NTaylorMullen:

PR #420, opened 2025-05-18
fix(cli): Remove duplicate auto-accept indicator

NTaylorMullen:
- The auto-accept edits indicator was appearing in two places: once next to the loading indicator and again in the CWD bar.
- This was introduced when the CWD bar was made always visible.
- This commit removes the duplicate indicator, leaving only the one in the CWD bar.

Fixes https://b.corp.google.com/issues/413740468


NTaylorMullen:

PR #423, opened 2025-05-19
feat(cli): Add ShellModeIndicator component

NTaylorMullen:
This commit introduces a new ShellModeIndicator component to visually signify when shell mode is active.

- Displays "shell mode enabled (! to toggle)" in the UI.
- The AutoAcceptIndicator is now hidden when shell mode is active to prevent UI clutter.
image

### DIFF
--- a/packages/cli/src/ui/App.tsx
+++ b/packages/cli/src/ui/App.tsx
@@ -16,6 +16,7 @@ import { useAutoAcceptIndicator } from './hooks/useAutoAcceptIndicator.js';
 import { Header } from './components/Header.js';
 import { LoadingIndicator } from './components/LoadingIndicator.js';
 import { AutoAcceptIndicator } from './components/AutoAcceptIndicator.js';
+import { ShellModeIndicator } from './components/ShellModeIndicator.js';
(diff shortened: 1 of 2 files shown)


olcan:
FWIW I think the lower left is a better place for this indicator, so I would swap this with the memory indicator.

NTaylorMullen:
> FWIW I think the lower left is a better place for this indicator, so I would swap this with the memory indicator.

I chose top right above the prompt box for now because that's also where the auto accept indicator is. That being said, absolutely open to different locations. /cc @miguelsolorio

miguelsolorio:
tbh, all of these indicators are a bit distracting. from a ux perspective, i'd push to minimize the noise even further. maybe these can be enabled via a setting.

NTaylorMullen:
> tbh, all of these indicators are a bit distracting. from a ux perspective, i'd push to minimize the noise even further. maybe these can be enabled via a setting.

ya we have definitely been getting into "Starship Enterprise" level cockpit for GC 😄

olcan:
Just wanted to share my opinions.
I think a mental model helps reduce cognitive load.
Context-related things above the box. That's where the context already is.
Interaction modes and tips below the box. We already have completions there.

PR #427, opened 2025-05-19
indicate system prompt override on bottom right; require GEMINI_SYSTEM_MD to make it more explicit; allow custom paths for read/write

olcan:


packages/cli/src/ui/components/Footer.tsx
```diff
@@ -74,6 +74,9 @@ export const Footer: React.FC<FooterProps> = ({
           <Text color={Colors.AccentRed}>▼ </Text>
         </Text>
       )}
+      {process.env.GEMINI_SYSTEM_MD && (
+        <Text color={Colors.AccentRed}>|⌐■_■|</Text>
+      )}
```
allenhutchison, line 79: This is a really minor thing, but we should probably put this on the left hand side with the GEMINI.md file notifications. That way all the notifications about files modifying the behaviour of the model are in the same place.

PR #428, opened 2025-05-19
feat(cli): Implement /bug command and add open dependency

allenhutchison:
- Implements the /bug command in the CLI:
    - Opens the GitHub issue page with a pre-filled title if a bug description is provided.
    - Opens the generic GitHub issue page if no description is provided.
- Adds open as a dependency to packages/cli/package.json.
    - The open library is used to open URLs in the default browser in a robust and cross-platform manner.
- Updates tests for the /bug command.

Fixes b/413139077


NTaylorMullen (approved):
How do you feel about also including information like machine, CLI version, history (if user will allow us) etc.?

allenhutchison:
I've added CLI Version, OS, Model, and Sandbox ENV. Let's start there and see what else we want to gather for bug reports.

PR #435, opened 2025-05-19
Add Logger for command history

sethtroisi:
Same as #260 but in a local branch not a forked repo

### DIFF
--- a/packages/cli/src/ui/App.tsx
+++ b/packages/cli/src/ui/App.tsx
@@ -28,15 +28,10 @@ import { Tips } from './components/Tips.js';
 import { ConsoleOutput } from './components/ConsolePatcher.js';
 import { HistoryItemDisplay } from './components/HistoryItemDisplay.js';
 import { useHistory } from './hooks/useHistoryManager.js';
+import { useLogger } from './hooks/useLogger.js';
 import process from 'node:process';
 import { MessageType } from './types.js';
-import {
-  getErrorMessage,
-  shortenPath,
-  type Config,
-  Logger,
-} from '@gemini-code/server';
-import { useLogger } from './hooks/useLogger.js';
+import { getErrorMessage, shortenPath, type Config } from '@gemini-code/server';
 
 interface AppProps {
   config: Config;
@@ -163,23 +158,28 @@ export const App = ({
   );
 
   const logger = useLogger();
-  const userMessages = useMemo(
-    () => {
-      const pastMessages = logger.current?.getPreviousUserMessages() || [];
-      if (pastMessages) {
-        return pastMessages as string[];
+  const [userMessages, setUserMessages] = useState<string[]>([]);
+
+  useEffect(() => {
+    const fetchUserMessages = async () => {
+      const pastMessages = (await logger?.getPreviousUserMessages()) || [];
+      if (pastMessages.length > 0) {
+        setUserMessages(pastMessages.reverse());
+      } else {
+        setUserMessages(
+          history
+            .filter(
+              (item): item is HistoryItem & { type: 'user'; text: string } =>
+                item.type === 'user' &&
+                typeof item.text === 'string' &&
+                item.text.trim() !== '',
+            )
+            .map((item) => item.text),
+        );
       }
-      return history
-        .filter(
-          (item): item is HistoryItem & { type: 'user'; text: string } =>
-            item.type === 'user' &&
-            typeof item.text === 'string' &&
-            item.text.trim() !== '',
-        )
-        .map((item) => item.text);
-    },
-    [history],
-  );
+    };
+    fetchUserMessages();
+  }, [history, logger]);
 
   const isInputActive = streamingState === StreamingState.Idle && !initError;
 
--- a/packages/cli/src/ui/App.tsx
+++ b/packages/cli/src/ui/App.tsx
@@ -31,7 +31,7 @@ import { useHistory } from './hooks/useHistoryManager.js';
 import { useLogger } from './hooks/useLogger.js';
 import process from 'node:process';
 import { MessageType } from './types.js';
-import { getErrorMessage, shortenPath, type Config } from '@gemini-code/server';
+import { getErrorMessage, type Config } from '@gemini-code/server';
 
 interface AppProps {
   config: Config;
--- a/packages/cli/src/ui/App.tsx
+++ b/packages/cli/src/ui/App.tsx
@@ -178,11 +178,10 @@ export const App = ({
               typeof item.text === 'string' &&
               item.text.trim() !== '',
           )
-          .map((item) => item.text)
+          .map((item) => item.text),
       ),
     [history, pastMessages],
   );
-
 
   const isInputActive = streamingState === StreamingState.Idle && !initError;
 
--- a/packages/cli/src/ui/App.tsx
+++ b/packages/cli/src/ui/App.tsx
@@ -29,9 +29,10 @@ import { Tips } from './components/Tips.js';
 import { ConsoleOutput } from './components/ConsolePatcher.js';
 import { HistoryItemDisplay } from './components/HistoryItemDisplay.js';
 import { useHistory } from './hooks/useHistoryManager.js';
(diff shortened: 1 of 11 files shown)


packages/cli/src/ui/App.tsx
```diff
@@ -166,17 +170,29 @@ export const App = ({
     [submitQuery, shellModeActive],
   );
 
+  const [pastMessages, setPastMessages] = useState<string[]>([]);
```
NTaylorMullen: Why have a separate past messages collection vs. reusing the history/pending ones?

packages/cli/src/ui/hooks/useGeminiStream.ts
```diff
@@ -115,6 +117,7 @@ export const useGeminiStream = (
     if (typeof query === 'string') {
       const trimmedQuery = query.trim();
       onDebugMessage(`User query: '${trimmedQuery}'`);
+      Logger.getInstance().logMessage(RoleType.USER, trimmedQuery);
```
NTaylorMullen: This should be in the history, also we should probably get rid of all of the `onDebugMessage` pieces in favor of the logger.

packages/server/src/core/client.ts
```diff
@@ -44,6 +46,7 @@ export class GeminiClient {
         },
       },
     });
+    this.logger = Logger.getInstance();
```
NTaylorMullen: For testing I imagine we'll need or want to be passing in a logger across the board, is it worth having a singleton? Typically rich loggers will have scopes / sub-logging abilities which also yields towards non-singleton patterns.

packages/server/src/core/logger.ts
```diff
@@ -0,0 +1,141 @@
...
+const CREATE_TABLE_SQL = `
+CREATE TABLE IF NOT EXISTS messages (
+    session_id TEXT,
+    message_id TEXT,
+    timestamp TEXT,
+    type TEXT,
+    message TEXT
+);`;
+
+export enum RoleType {
```
NTaylorMullen: We should definitely have categorical message types BUT having a role type feels a little different from that. This represents what the user has sent which will also exist in history no?

packages/server/src/core/logger.ts
```diff
@@ -0,0 +1,141 @@
...
+  USER = 'user',
+}
+
+export class Logger {
+  private static instance: Logger;
+  private db: sqlite3.Database | null = null;
```
NTaylorMullen: Instead of setting all of these to `null` you could:

```suggestion
  private db: sqlite3.Database | undefined;
```

packages/server/src/core/logger.ts
```diff
@@ -0,0 +1,141 @@
...
+  private sessionId: number | null = null;
+  private messageId: number | null = null;
+
+  private constructor() {}
+
+  static getInstance(): Logger {
```
NTaylorMullen: See earlier comment about the singleton pattern here

packages/server/src/core/logger.ts
```diff
@@ -0,0 +1,141 @@
...
+      const DB_DIR = path.resolve(process.cwd(), GEMINI_DIR);
+      const DB_PATH = path.join(DB_DIR, DB_NAME);
+      await fs.mkdir(DB_DIR, { recursive: true });
+
+      this.sessionId = Math.floor(Date.now() / 1000);
+      this.messageId = 0;
+
+      this.db = new sqlite3.Database(DB_PATH, (err: Error | null) => {
+        if (err) {
+          throw err;
```
NTaylorMullen: Is this method run synchronously? Where does this bubble up?

packages/server/src/core/logger.ts
```diff
@@ -0,0 +1,141 @@
...
+          reject(err);
+        } else {
+          resolve(rows.map((row) => row.message));
+        }
+      });
+    });
+  }
+
+  async logMessage(type: RoleType, message: string): Promise<void> {
+    if (!this.db) {
```
NTaylorMullen, line 99: How async is initialization? Asking because I'm worried this actually wont be guaranteed to work:

```
await logger.initialize()
await logger.logMessage(...)
```

packages/cli/src/ui/App.tsx
sethtroisi: things added to history get displayed on the screen and sent to the model.
I tried both ways by having history add a "display: bool and chat:bool" but it gets gnarly fast.

packages/cli/src/ui/hooks/useGeminiStream.ts
sethtroisi: +1 I'll add it to the follow up list (tracked in buganizer)

packages/server/src/core/logger.ts
sethtroisi: This represents who sent the message (only user for now but gemini, tool, system, and error later)

MessageType is a better name but it conflicts with the MessageType used by history, gemini-cli suggests `MessageSender` which I like better. I'll update

packages/server/src/core/logger.ts
sethtroisi: Done

packages/server/src/core/client.ts
sethtroisi: In the first pass I was having a lot of difficulty passing the instance everywhere it was needed. I'll see if gemini-code can refactor it now that the design is finalized.

packages/server/src/core/logger.ts
sethtroisi: Ack

packages/server/src/core/logger.ts
sethtroisi: the whole method is async, errors here bubble up into the try catch which then rejects the promise

packages/server/src/core/logger.ts
sethtroisi: if you await `logger.initialize` then logger.db will work. The second await is needed if you care about message ordering.

What's could lead to race conditions is the real code in useGeminiStream which doesn't have either await, but I think it's better to get to the prompt faster and possible lose some logging than wait on the db to connect.

packages/server/src/core/client.ts
sethtroisi: gemini-code did a great job refactoring and cleaning is all up so done!

packages/cli/src/ui/App.tsx
NTaylorMullen: Sorry I think I'm missing it, where is `pastMessages` used in a specialized way?

packages/server/src/core/logger.ts
NTaylorMullen: This logging tech is meant to be used for more than just history messages though isn't it?

packages/cli/src/ui/App.tsx
sethtroisi: on line 171 we create `userMessages` from the concat of `pastMessages + history` which gets used by `<inputPrompt>` (for autocomplete, up arrow and other things). `history` gets used a number of other places mostly directly `<Box>` on line 300 in rendering the UI

NTaylorMullen: Thank you for hanging in there with me. This piece is effectively restoring the history of our user messages.

So one concern I have with this approach is that it will restore all of the user interactions BUT wont restore model responses and wont pre-fill the client.ts with that history so that further interactions are possible.

packages/server/src/core/logger.ts
NTaylorMullen: While the message is async this call isn't awaited which means any errors thrown here will bubble up to the app itself (would probably crash the application).

packages/server/src/core/logger.ts
NTaylorMullen: One of the challenges is that if this fails and does `console.error` it will actually be surfaced to the user in the top-level UI

sethtroisi:
We discussed offline to resolve all the remaining comments. I'm pushing then OOO so feel free to revert if this breaks things.

PR #438, opened 2025-05-19
warn on cd in shell mode. done robustly based on lessons from shell tool. logs to console.warn for now, and does not restore (but see comment on how to restore)

olcan:


packages/cli/src/ui/hooks/shellCommandProcessor.ts
```diff
@@ -76,6 +86,15 @@ export const useShellCommandProcessor = (
               userMessageTimestamp,
             );
           }
+          if (fs.existsSync(pwdFilePath)) {
+            const pwd = fs.readFileSync(pwdFilePath, 'utf8').trim();
+            if (pwd !== targetDir) {
+              console.warn(
```
NTaylorMullen, line 92: Instead of a console warning, why not add an "error" or another "info" history item? Console.warning will stick something permanent at the bottom of the UI

olcan:
yeah sorry i was hoping you could decide the best approach, and even
potentially save/restore if you would like to have persistence in the new
shell mode; it is not as hard to do robustly as i thought, was able to use
(painful) learnings from shell tool

PR #443, opened 2025-05-20
Implement additional readline-like keybindings, including alt-left arrow and alt-right arrow.

dewitt:
This change adds keybinding support for:

  - `Ctrl+B`: Moves the cursor backward one character.
  - `Ctrl+F`: Moves the cursor forward one character.
  - `Alt+Left Arrow`: Moves the cursor backward one word.
  - `Alt+Right Arrow`: Moves the cursor forward one word.

Note, the keybindings are consistent with terminal handling on macOS, and represented by the `key.meta` binding in `ink`.

Closes b/411469305.

### DIFF
--- a/packages/cli/src/ui/components/InputPrompt.tsx
+++ b/packages/cli/src/ui/components/InputPrompt.tsx
@@ -114,12 +114,7 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
   );
 
   const inputPreprocessor = useCallback(
-    (
-      input: string,
-      key: Key,
-      _currentText?: string,
-      _cursorOffset?: number,
-    ) => {
+    (input: string, key: Key, currentText?: string, cursorOffset?: number) => {
       if (input === '!' && query === '' && !showSuggestions) {
         setShellModeActive(!shellModeActive);
         onChangeAndMoveCursor(''); // Clear the '!' from input
@@ -177,6 +172,21 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
         }
         if (key.ctrl && input === 'n') {
           inputHistory.navigateDown();
+          return true;
+        }
+        if (key.ctrl && input === 'b' && typeof cursorOffset !== 'undefined') {
+          setEditorState((s) => ({
+            key: s.key + 1,
+            initialCursorOffset: Math.max(0, cursorOffset - 1),
+          }));
+          return true;
+        }
+        if (key.ctrl && input === 'f' && typeof cursorOffset !== 'undefined') {
+          const currentTextLength = currentText?.length ?? query.length;
+          setEditorState((s) => ({
+            key: s.key + 1,
+            initialCursorOffset: Math.min(currentTextLength, cursorOffset + 1),
+          }));
           return true;
         }
       }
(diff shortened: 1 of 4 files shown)


packages/cli/src/ui/components/InputPrompt.tsx
```diff
@@ -156,29 +151,91 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
...
+        // Keybinding: Alt/Meta/Option + Left Arrow to move backward one word
+        if (key.meta && input === 'b' && typeof cursorOffset !== 'undefined') {
+          const text = currentText ?? query;
+          let i = cursorOffset - 1;
+          // Skip whitespace to the left of the cursor
+          while (i >= 0 && text[i] === ' ') {
+            i--;
+          }
+          // Skip non-whitespace characters (the word itself)
+          while (i >= 0 && text[i] !== ' ') {
```
NTaylorMullen: How do you feel about expanding this to be more than `!= ' '`. For instance if you have a `this.foo(hello)` the cursor will jump around the punctuation differently than the characters themselves.

packages/cli/src/ui/components/InputPrompt.tsx
```diff
@@ -156,29 +151,91 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
...
+        // Keybinding: Alt/Meta/Option + Right Arrow to move forward one word
+        if (key.meta && input === 'f' && typeof cursorOffset !== 'undefined') {
+          const text = currentText ?? query;
+          let i = cursorOffset;
+          // Skip whitespace to the right of the cursor
+          while (i < text.length && text[i] === ' ') {
+            i++;
+          }
+          // Skip non-whitespace characters (the word itself)
+          while (i < text.length && text[i] !== ' ') {
```
NTaylorMullen: ditto

NTaylorMullen (approved):
Love how elegant and simple it was to add this 😄

packages/cli/src/ui/components/InputPrompt.tsx
dewitt: That's a great catch.

I tested a few IDEs that I have installed and all do seem to stop on punctuation too, but we should at least be consistent with people's muscle memory, so I'll update it to use whatever VSCode uses as `editor.wordSeparators`.

Which a quick Google suggests is whitespace characters plus ```\~!@#%^&*()-=+[{]}|;:`'",.<>/?``` by default (it's over-rideable per-language), but I'll check the code and confirm. 

Interestingly, one IDE I checked had different logic moving forward vs backward. Whatever I do, I'll change it to work like VSCode, if not too hard.

packages/cli/src/ui/components/InputPrompt.tsx
dewitt: As above.

packages/cli/src/ui/components/InputPrompt.tsx
NTaylorMullen: Love that choice!

dewitt:
Update for @NTaylorMullen regarding non-word characters (https://github.com/google-gemini/gemini-cli/pull/443#discussion_r2096896303): I refactored pretty much this whole PR in subsequent commits to do all the movement processing in once place now, text-buffer.ts, and there we already had ```/[\s,.;!?]+$/```.

For now I'm going to keep this logic, so we'll be consistent with past behavior of other ink apps that supported word movement. The choice of characters can be updated later if we want.

PTAL.

NTaylorMullen (approved):

PR #447, opened 2025-05-20
Support Images and PDFs

jacob314:
Support images and pdfs if and only if the file name or extension of the resource is explicitly listed in the `@`
Added tests for read_many_files

### DIFF
--- a/packages/cli/src/ui/hooks/atCommandProcessor.ts
+++ b/packages/cli/src/ui/hooks/atCommandProcessor.ts
@@ -6,7 +6,7 @@
 
 import * as fs from 'fs/promises';
 import * as path from 'path';
-import { PartListUnion } from '@google/genai';
+import { PartListUnion, PartUnion } from '@google/genai';
 import {
   Config,
   getErrorMessage,
@@ -126,6 +126,7 @@ export async function handleAtCommand({
   }
 
   const contentLabel = pathPart;
+
   const toolRegistry = config.getToolRegistry();
   const readManyFilesTool = toolRegistry.getTool('read_many_files');
 
@@ -168,7 +169,6 @@ export async function handleAtCommand({
 
   try {
     const result = await readManyFilesTool.execute(toolArgs, signal);
-    const fileContent = result.llmContent || '';
 
     toolCallDisplay = {
       callId: `client-read-${userMessageTimestamp}`,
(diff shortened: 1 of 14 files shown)


packages/cli/src/ui/hooks/atCommandProcessor.ts
```diff
@@ -180,13 +180,29 @@ export async function handleAtCommand({
...
+
+    function addFileContentPart(fileContent: string) {
+      processedQueryParts.push({
+        text: `\n--- Content from: ${contentLabel} ---\n${fileContent}\n--- End Content ---`,
+      });
+    }
+    // Process the result from the tool
+    if (typeof result.llmContent === 'string') {
+      addFileContentPart(result.llmContent);
+    } else if (Array.isArray(result.llmContent)) {
```
jacob314: flagging that this method is wrong. Need to think about what the most idiomatic way to fix it is. I suspect I should wrap the entire llmContent returned with a single
content from with a single
```
---Content from: 
--- End Content -- 
``
block. 
Right now we are adding too many such blocks.

Comment (jacob314) [reply packages/cli/src/ui/hooks/atCommandProcessor.ts]: fixed. also added some tests for this. For discussion: we could merge adjacent text blocks if we care. Also unclear whether it matters that we are sometimes using {text: textAfter} and sometimes just using String for the parts. Not sure which is the most idiomatic.

File: packages/server/src/core/geminiRequest.ts [hunk 2]
```diff
@@ -0,0 +1,71 @@
...
+import { type PartListUnion, type Part } from '@google/genai';
+
+/**
+ * Represents a request to be sent to the Gemini API.
+ * For now, it's an alias to PartListUnion as the primary content.
+ * This can be expanded later to include other request parameters.
+ */
+export type GeminiCodeRequest = PartListUnion;
+
+export function partListUnionToString(value: PartListUnion): string {
```
NTaylorMullen, line 16: How do you feel about putting this on `GeminiCodeRequest` itself as a to string override? I.e. `request.toString()`

NTaylorMullen: Ah nevermind, I see you're also using this to interpret responses.

packages/server/src/tools/read-many-files.ts
```diff
@@ -319,28 +320,63 @@ Default excludes apply to common non-text files and large dependency directories
...
-          Math.min(contentBuffer.length, 1024),
-        );
-        if (sample.includes(0)) {
-          skippedFiles.push({
-            path: relativePathForDisplay,
-            reason: 'Skipped (appears to be binary)',
+        const mimeType = mime.lookup(filePath);
+        if (
+          mimeType &&
+          (mimeType.startsWith('image/') || mimeType === 'application/pdf')
```
NTaylorMullen, line 326: Interesting using base64 for pdf's also. Hmmm, is that used elsewhere? Asking because in Copilot we extarcted the raw text. If this works then that's fine 😄

NTaylorMullen (approved):

packages/server/src/tools/read-many-files.ts
jacob314: This does work. Verified using a pdf research paper. I think this aligns with the Gemini instructions for PDFs.
https://ai.google.dev/gemini-api/docs/document-processing?lang=python#inline_data

allenhutchison:
For PDFs it might be better to just use the file api: https://ai.google.dev/gemini-api/docs/document-processing?lang=node that would allow us to keep the conversation history / context cleaner and reduce the number of tokens going back and forth.

PR #448, opened 2025-05-20
feat: useToolScheduler hook to manage parallel tool calls

KeijiBranshi:
BUG=b/411426253
DIFFBASE: #468
DEMO: https://screencast.googleplex.com/cast/NjQwODk5NDU2NTM5MDMzNnwwM2Y2MDQ2MS1kOA

Incorporates the changes from the diffbase into the `useGeminiStream` hook.

# Bug Context

- parallel tool calls in general were just broken. The cli would confirm the first, then ignore the rest entirely. 
  - this was because we [stop iterating](https://github.com/google-gemini/gemini-cli/blob/2970f0a06c234d1cb2fafb2e02754973bd1fa26a/packages/cli/src/ui/hooks/useGeminiStream.ts#L294) through Gemini's response (an async iterable) after seeing a tool call confirmation request
  - additionally, moving the async iterator to the next item would continue [here](https://github.com/google-gemini/gemini-cli/blob/2970f0a06c234d1cb2fafb2e02754973bd1fa26a/packages/server/src/core/turn.ts#L205-L233), where the turn.ts would just immediately try to parse the response for a function that was never executed.
- Once that was addressed, there was an issue where tool call responses and incoming gemini messages (ie text responses to the function responses) were clashing with each other. 
  - This is because we only had [1 pending history item](https://github.com/google-gemini/gemini-cli/blob/2970f0a06c234d1cb2fafb2e02754973bd1fa26a/packages/cli/src/ui/hooks/useGeminiStream.ts#L59) component for both incoming Gemini messages _and_ tool call confirmation dialogs. 
  - Once one tool call finished, we'd kick off a [nested/recursive query](https://github.com/google-gemini/gemini-cli/blob/2970f0a06c234d1cb2fafb2e02754973bd1fa26a/packages/cli/src/ui/hooks/useGeminiStream.ts#L489) to gemini. 
  - Gemini would then respond while the next tool call was still running. 
- Lastly, there was a bit of duplicate state (tbh still is) between the cli and the server package for tool confirmations. both packages had logic to detect tool call confirmation requirements, and execute followup tool calls. 
  - to minimize the back and forth between cli + server in these confirmations, I opted to just move the tool call confirmation/scheduling state management entirely into the cli.
  - there's probably an argument for moving this entirely in the server, but I just stuck with the cli-side since React hooks inherently offer async state management utilities. coordinating that with the async-iterator between cli and server got a little messy. 
  - side note, we also may want to consider abstracting more of our state behind something like useReducer as it's pretty effective at more compelx state management.

### DIFF
--- a/packages/cli/src/ui/hooks/useGeminiStream.ts
+++ b/packages/cli/src/ui/hooks/useGeminiStream.ts
@@ -4,34 +4,28 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { useState, useRef, useCallback, useEffect } from 'react';
+import { useState, useRef, useCallback, useEffect, useMemo } from 'react';
 import { useInput } from 'ink';
 import {
   GeminiClient,
   GeminiEventType as ServerGeminiEventType,
   ServerGeminiStreamEvent as GeminiEvent,
   ServerGeminiContentEvent as ContentEvent,
-  ServerGeminiToolCallRequestEvent as ToolCallRequestEvent,
-  ServerGeminiToolCallResponseEvent as ToolCallResponseEvent,
-  ServerGeminiToolCallConfirmationEvent as ToolCallConfirmationEvent,
   ServerGeminiErrorEvent as ErrorEvent,
   getErrorMessage,
   isNodeError,
   Config,
   MessageSenderType,
   ServerToolCallConfirmationDetails,
-  ToolCallConfirmationDetails,
   ToolCallResponseInfo,
-  ToolConfirmationOutcome,
   ToolEditConfirmationDetails,
   ToolExecuteConfirmationDetails,
   ToolResultDisplay,
-  partListUnionToString,
+  ToolCallRequestInfo,
 } from '@gemini-code/server';
 import { type Chat, type PartListUnion, type Part } from '@google/genai';
 import {
   StreamingState,
-  IndividualToolCallDisplay,
   ToolCallStatus,
   HistoryItemWithoutId,
   HistoryItemToolGroup,
@@ -44,6 +38,7 @@ import { findLastSafeSplitPoint } from '../utils/markdownUtilities.js';
 import { useStateAndRef } from './useStateAndRef.js';
 import { UseHistoryManagerReturn } from './useHistoryManager.js';
 import { useLogger } from './useLogger.js';
+import { useToolScheduler, mapToDisplay } from './useToolScheduler.js';
 
 enum StreamProcessingStatus {
   Completed,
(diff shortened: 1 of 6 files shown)


packages/cli/src/ui/hooks/useToolScheduler.ts
```diff
@@ -0,0 +1,303 @@
...
+  useEffect(() => {
+    // effect for executing scheduled tool calls
+    const scheduledCalls = toolCalls.filter((t) => t.status === 'scheduled');
+    const awaitingConfirmation = toolCalls.some(
+      (t) => t.status === 'awaiting_approval',
+    );
+    if (!awaitingConfirmation && scheduledCalls.length) {
+      scheduledCalls.forEach(async (c) => {
+        try {
+          setToolCalls(setStatus(c.request.callId, 'executing'));
```
NTaylorMullen: I could see it being valuable having a different indicator for tool calls that are waiting to be confirmed (or not), i.e. the `?` indicator

packages/cli/src/ui/hooks/useToolScheduler.ts
```diff
@@ -0,0 +1,303 @@
...
+    functionResponse: {
+      id: request.callId,
+      name: request.name,
+      response: { error: error.message },
+    },
+  },
+  resultDisplay: error.message,
+});
+
+function mapStatus(status: Status): ToolCallStatus {
```
NTaylorMullen: Ah mapping the status to better enforce the boundary between cli and server righT?

packages/cli/src/ui/hooks/useToolScheduler.ts
KeijiBranshi: it does use the `?` for tools awaiting confirmation! I didn't explicitly address that, so it looks like it was already the case.

packages/cli/src/ui/hooks/useToolScheduler.ts
KeijiBranshi: awkwardly, I was trying to set a boundary/scope between this hook and the rest of the CLI package. Just so I could freely make edits to those in this file without breaking other parts of the CLI package.

packages/cli/src/ui/hooks/useGeminiStream.ts
```diff
@@ -209,156 +226,48 @@ export const useGeminiStream = (
...
-      originalDetails: ServerToolCallConfirmationDetails['details'],
-    ) {
-      let resultDisplay: ToolResultDisplay | undefined;
-      if ('fileDiff' in originalDetails) {
-        resultDisplay = {
-          fileDiff: (originalDetails as ToolEditConfirmationDetails).fileDiff,
-          fileName: (originalDetails as ToolEditConfirmationDetails).fileName,
-        };
-      } else {
-        resultDisplay = `~~${(originalDetails as ToolExecuteConfirmationDetails).command}~~`;
```
KeijiBranshi, line 340: TODO: this needs to be ported over to the tool scheduler too

PR #451, opened 2025-05-20
a2a server implementation

cornmander:
Adds a new standalone A2A server that runs the core agent logic.


NTaylorMullen:
Let me know once you feel like this is good for review!

PR #453, opened 2025-05-20
Merge InputPrompt and multiline-editor and move autocomplete logic directly into InputPrompt

jacob314:
multiline-editor was only used for the InputPrompt so having it separated resulted in extra complexity.
Similarly, keeping the autocomplete functionality at arms length from InputPrompt resulted in a more complex InputPrompt API.

This cleans up some ways we were using keys and ending up with multiple inconsistent ways to set text content.


packages/cli/src/ui/App.tsx
```diff
@@ -361,37 +327,16 @@ export const App = ({
...
-                      activeIndex={completion.activeSuggestionIndex}
-                      isLoading={completion.isLoadingSuggestions}
-                      width={suggestionsWidth}
-                      scrollOffset={completion.visibleStartIndex}
-                      userInput={query}
-                    />
-                  </Box>
-                )}
-              </>
+              <InputPrompt
```
NTaylorMullen, line 330: so clean

NTaylorMullen (approved):

dewitt:
This is much cleaner, and much improved. Not a proper review, but a quick thank you for cleaning this up.

jacob314:
Landing for now. to unblock others from continuing to polish InputPrompt. @KeijiBranshi if you spot some new React code smells that I've introduced let me know and I'll fix them in follow ups. Directionally I believe this pull request resolves a lot of the existing code smells passing passing tons of state around and forcing re-renders using keys.

PR #456, opened 2025-05-20
Update docs and tool description for read-many-files.

jacob314:
Unsurprisingly this pull request was purely generated by GC after fixing a case where it messed up string escaping.
Feel free to suggest removing cases where this is overly verbose.

### DIFF
--- a/packages/server/src/tools/read-many-files.ts
+++ b/packages/server/src/tools/read-many-files.ts
@@ -161,7 +161,14 @@ export class ReadManyFilesTool extends BaseTool<
     super(
       ReadManyFilesTool.Name,
       'ReadManyFiles',
-      `Reads content from multiple text files specified by paths or glob patterns within a configured target directory and concatenates them into a single string.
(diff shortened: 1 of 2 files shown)


allenhutchison (approved):

packages/server/src/tools/read-many-files.ts
```diff
@@ -161,20 +161,30 @@ export class ReadManyFilesTool extends BaseTool<
     super(
       ReadManyFilesTool.Name,
       'ReadManyFiles',
-      `Reads content from multiple text files specified by paths or glob patterns within a configured target directory and concatenates them into a single string.
+      `Reads content from multiple files specified by paths or glob patterns
+within a configured target directory. For text files, it concatenates their content
```
NTaylorMullen, line 165: This goes over to the LLM we wont want to do line breaks here and below

jacob314: Done. Removed line breaks within paragraphs. https://github.com/google-gemini/gemini-cli/pull/476

PR #457, opened 2025-05-21
feat: Allow Esc to exit shell mode

NTaylorMullen:
- Update InputPrompt.tsx to handle Esc key for exiting shell mode.
- Modify ShellModeIndicator.tsx to reflect the new keybinding. Fixes https://buganizer.corp.google.com/issues/419087952


sethtroisi:
Great change!

PR #462, opened 2025-05-21
Add globbing support to @-command file suggestions and resolution.

dewitt:
Implements recursive glob-based file search for both suggestions and execution of the `@` command.

- When typing `@filename`, suggestions will now include files matching `filename` in nested directories.
- Suggestions are sorted by path depth (shallowest first), then directories before files, then alphabetically.
- The maximum recursion depth for suggestions is set to 10.
- When executing an `@filename` command, if the file is not found directly, a recursive search (using the glob tool) is performed to locate the file.

This addresses the first request in issue #461 by allowing users to quickly reference deeply nested files without typing the full path. Also addresses b/416292478.

### DIFF
--- a/packages/cli/src/ui/hooks/useCompletion.ts
+++ b/packages/cli/src/ui/hooks/useCompletion.ts
@@ -180,44 +180,114 @@ export function useCompletion(
     const baseDirAbsolute = path.resolve(cwd, baseDirRelative);
 
     let isMounted = true;
+
+    const findFilesRecursively = async (
+      startDir: string,
+      searchPrefix: string,
+      currentRelativePath = '',
+      depth = 0,
+      maxDepth = 10, // Limit recursion depth
+      maxResults = 50, // Limit number of results
+    ): Promise<Suggestion[]> => {
+      if (depth > maxDepth) {
+        return [];
+      }
+
+      let foundSuggestions: Suggestion[] = [];
+      try {
+        const entries = await fs.readdir(startDir, { withFileTypes: true });
+        for (const entry of entries) {
+          if (foundSuggestions.length >= maxResults) break;
+
+          const entryPathRelative = path.join(currentRelativePath, entry.name);
+          if (entry.name.startsWith(searchPrefix)) {
+            foundSuggestions.push({
+              label: entryPathRelative + (entry.isDirectory() ? '/' : ''),
+              value: escapePath(
+                entryPathRelative + (entry.isDirectory() ? '/' : ''),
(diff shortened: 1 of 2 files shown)


allenhutchison:
Also addresses b/416292478 which I've reassigned to you.

packages/cli/src/ui/hooks/useCompletion.ts
```diff
@@ -180,44 +180,114 @@ export function useCompletion(
     const baseDirAbsolute = path.resolve(cwd, baseDirRelative);
 
     let isMounted = true;
+
+    const findFilesRecursively = async (
+      startDir: string,
+      searchPrefix: string,
+      currentRelativePath = '',
+      depth = 0,
+      maxDepth = 10, // Limit recursion depth
```
NTaylorMullen, line 189: We definitely need a max depth here to avoid blowing out the world; food for thought / another time would be awesome to also capture items / folders that are represented by files in latest commits or staging areas (assuming git repo).

Of course we'll eventually want a tri index for these sorts of things BUT that's its own challenge 😄

packages/cli/src/ui/hooks/useCompletion.ts
```diff
@@ -180,44 +180,114 @@ export function useCompletion(
...
+      maxResults = 50, // Limit number of results
```
NTaylorMullen, line 190: To make sure I'm understanding: for every key stroke we compute 50 items recursively that potentially match the query string right?

NTaylorMullen (approved):

packages/cli/src/ui/hooks/useCompletion.ts
dewitt: Yeah, we just break after 50. ("50 results ought to be enough for anybody.")

Then we sort then, first by folder depth (so `README.md` is ordered before `docs/README.md`), and then lexicographically (`foo/bar` before `foo/baz`) when showing the results.

packages/cli/src/ui/hooks/useCompletion.ts
dewitt: Yes. I thought about adding a cache for this that we can invalidate on git fetch or our own filechanges, or just use some sort of file-watcher. But amazingly the brute force even at maxDepth = 10 seems to introduce no appreciable latency, so maybe the easy way will work.

dewitt:
Accidentally requested a re-review. No blocking comments from @NTaylorMullen. Proceeding to squash and merge.

PR #465, opened 2025-05-21
fix exit code for shell tool

olcan:


NTaylorMullen:
Was it not showing the exit code before?

olcan:
yeah it would swallow the exit code due to wrapper before; same with shell
tool so fixed both

NTaylorMullen:
Ooooo this makes so much sense. I was having some funky behavior here and there about it not realizing the type of interactions / responses shell commands were having

olcan:
Yeah would be especially bad if there is no stderr to indicate the error.

PR #470, opened 2025-05-21
Fix: Prevent hang in large directories by using BFS for getFolderStru…

allenhutchison:
…cture

Refactors `getFolderStructure` to use a Breadth-First Search (BFS) when reading directory contents. This prevents the application from hanging in directories with a very large number of files or deep nesting when the `maxItems` limit is approached.

The previous Depth-First Search (DFS) could exhaustively explore a single deep branch before hitting `maxItems`, leading to long delays. The new BFS approach ensures a more balanced traversal, providing a wider view of the directory structure within the item limit.

Changes include:
- Replaced recursive DFS `readFullStructure` with an iterative BFS implementation.
- Removed the now redundant `reduceStructure` function.
- Updated `formatStructure` to work with the new BFS output and handle truncation indicators (`hasMoreFiles`, `hasMoreSubfolders`).
- Adjusted the summary message in `getFolderStructure`.

### DIFF
--- a/packages/cli/src/config/config.ts
+++ b/packages/cli/src/config/config.ts
@@ -144,7 +144,7 @@ async function collectDownwardGeminiFiles(
 
   if (debugMode)
     logger.debug(
-      `Scanning downward for GEMINI.md files in: ${directory} (scanned: ${scannedDirCount.count}/${maxScanDirs})`,
+      `Scanning downward for ${GEMINI_MD_FILENAME} files in: ${directory} (scanned: ${scannedDirCount.count}/${maxScanDirs})`,
     );
   const collectedPaths: string[] = [];
   try {
@@ -170,11 +170,13 @@ async function collectDownwardGeminiFiles(
           await fs.access(fullPath, fsSync.constants.R_OK);
           collectedPaths.push(fullPath);
           if (debugMode)
-            logger.debug(`Found readable downward GEMINI.md: ${fullPath}`);
+            logger.debug(
+              `Found readable downward ${GEMINI_MD_FILENAME}: ${fullPath}`,
+            );
         } catch {
           if (debugMode)
             logger.debug(
-              `Downward GEMINI.md not readable, skipping: ${fullPath}`,
+              `Downward ${GEMINI_MD_FILENAME} not readable, skipping: ${fullPath}`,
             );
         }
       }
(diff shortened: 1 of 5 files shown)


allenhutchison:
This fixes b/417749515

packages/cli/src/config/config.ts
```diff
@@ -130,8 +130,22 @@ async function collectDownwardGeminiFiles(
...
+      logger.debug(
+        `Max directory scan limit (${maxScanDirs}) reached. Stopping downward scan at: ${directory}`,
+      );
+    return [];
+  }
+  scannedDirCount.count++;
+
+  if (debugMode)
+    logger.debug(
+      `Scanning downward for GEMINI.md files in: ${directory} (scanned: ${scannedDirCount.count}/${maxScanDirs})`,
```
NTaylorMullen: Replace GEMINI.md with the constant

packages/server/src/utils/getFolderStructure.ts
```diff
@@ -37,286 +37,206 @@ interface FullFolderInfo {
...
 // --- Helper Functions ---
 
-/**
- * Recursively reads the full directory structure without truncation.
- * Ignored folders are included but not recursed into.
- * @param folderPath The absolute path to the folder.
- * @param options Configuration options.
- * @returns A promise resolving to the FullFolderInfo or null if access denied/not found.
- */
 async function readFullStructure(
```
NTaylorMullen, line 52: Could you have GC add tests for bits in this file

packages/server/src/utils/getFolderStructure.ts
```diff
@@ -37,286 +37,206 @@ interface FullFolderInfo {
...
   };
 
-  let totalChildrenCount = 0;
-  let totalFileCount = 0;
+  const queue: Array<{ folderInfo: FullFolderInfo; currentPath: string }> = [
+    { folderInfo: rootNode, currentPath: rootPath },
+  ];
+  let currentItemCount = 0;
+  // Count the root node itself as one item if we are not just listing its content
+  // currentItemCount++;
```
NTaylorMullen:
```suggestion
```

packages/server/src/utils/getFolderStructure.ts
```diff
@@ -37,286 +37,206 @@ interface FullFolderInfo {
...
+        // Add to queue for processing its children later
+        queue.push({ folderInfo: subFolderNode, currentPath: subFolderPath });
       }
     }
+    folderInfo.subFolders = subFoldersInCurrentDir;
   }
 
-  return rootReducedNode;
-}
+  // Recalculate totalChildren and totalFiles for rootNode based on its actual children
```
NTaylorMullen: Should this big comment be here? Doesn't look like there's much after it

packages/server/src/utils/getFolderStructure.ts
```diff
@@ -37,286 +37,206 @@ interface FullFolderInfo {
...
  * @param isLast Sibling indicator.
  * @param builder Array to build the string lines.
  */
-function formatReducedStructure(
-  node: ReducedFolderNode,
+function formatStructure(
+  node: FullFolderInfo,
   indent: string,
   isLast: boolean,
+  isRoot: boolean, // Added to handle root node name printing
```
NTaylorMullen:
```suggestion
  isRoot: boolean,
```

NTaylorMullen (approved):

packages/cli/src/config/config.ts
allenhutchison: Fixed throughout.

packages/server/src/utils/getFolderStructure.ts
allenhutchison: Added.

packages/server/src/utils/getFolderStructure.ts
allenhutchison: fixed

PR #471, opened 2025-05-21
use pending history item for shell mode, update as output is received

olcan:


NTaylorMullen:
Out of curiosity, what usecase did you find exploded for you without this?

olcan:
e.g. ! for i in $(seq 1 10); do echo "hello"; sleep 1; done
will now print hello every 1 second, instead of 10 hellos after 10 seconds
so if you have a long-running command, you will get streaming output

PR #474, opened 2025-05-21
Fix: Align WebSearchTool API key handling with GeminiClient

allenhutchison:
This change modifies the WebSearchTool to initialize the GoogleGenAI client in the same way as GeminiClient. It now allows the API key to be sourced from environment variables if not explicitly provided in the Config object.

This resolves an issue where WebSearchTool would fail to initialize if the API key was only present in the environment, unlike GeminiClient which could handle this scenario.

Fixes #473

### DIFF
--- a/packages/server/src/tools/web-search.ts
+++ b/packages/server/src/tools/web-search.ts
@@ -83,13 +83,11 @@ export class WebSearchTool extends BaseTool<
       },
     );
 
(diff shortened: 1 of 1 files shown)


NTaylorMullen:
It might be useful to also have a general client we construct at the tool registry layer and pass that down into various tools. Otherwise we'll find a lot of tools newing up their own clients

PR #478, opened 2025-05-22
Refactor: Streamline memoryUtils and update slash commands

allenhutchison:
Removes unused `deleteLastMemoryEntry` and `deleteAllAddedMemoryEntries` functions from `memoryUtils.ts` to simplify the API.

Updates the `/memory` slash command to reflect these changes, removing the `delete_last` and `delete_all_added` subcommands.

Includes new tests for `memoryUtils.ts` to improve coverage.

### DIFF
--- a/packages/cli/src/config/memoryUtils.ts
+++ b/packages/cli/src/config/memoryUtils.ts
@@ -9,10 +9,10 @@ import * as path from 'path';
 import { homedir } from 'os';
 import { SETTINGS_DIRECTORY_NAME } from './settings.js';
 import {
-  getErrorMessage,
+  // getErrorMessage, // Removed as it's not used
   MemoryTool,
   GEMINI_MD_FILENAME,
-  MEMORY_SECTION_HEADER,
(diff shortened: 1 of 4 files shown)


packages/cli/src/config/memoryUtils.ts
```diff
@@ -9,10 +9,10 @@ import * as path from 'path';
 import { homedir } from 'os';
 import { SETTINGS_DIRECTORY_NAME } from './settings.js';
 import {
-  getErrorMessage,
+  // getErrorMessage, // Removed as it's not used
```
dewitt, line 12: How do we get Gemini to actually stop with the unnecessary inline comments that do nothing but document what changes it just made? 

Currently GEMINI.md dictates:

> Only write high-value comments if at all. Avoid talking to the user through comments.

But that's clearly not working.

NTaylorMullen: Ya this might be an example of us wanting to test `GEMINI.md` as a user message instead of in system prompt /cc @olcan

PR #479, opened 2025-05-22
Chore: Integrate coverage reporting into CI

allenhutchison:
Updates the GitHub Actions workflow (`ci.yml`) to:
- Add a step to collect test coverage using `npm run coverage`.
- Add a step to upload coverage reports as build artifacts.
- Renames the existing test step to clarify it generates JUnit XML.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -87,8 +87,11 @@ jobs:
       - name: Install dependencies for testing
         run: npm ci # Install fresh dependencies using the downloaded package-lock.json
 
-      - name: Run tests
+      - name: Run tests (generate JUnit XML)
         run: npm run test:ci --workspaces --if-present
+
+      - name: Collect coverage
+        run: npm run coverage # This will run tests again, but also generate coverage
 
       - name: Publish Test Report
         uses: dorny/test-reporter@v1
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -87,9 +87,12 @@ jobs:
       - name: Install dependencies for testing
         run: npm ci # Install fresh dependencies using the downloaded package-lock.json
 
-      - name: Run tests
+      - name: Run tests (generate JUnit XML)
```
NTaylorMullen, line 90: Why is the clarification needed?

.github/workflows/ci.yml
```diff
@@ -87,9 +87,12 @@ jobs:
...
         run: npm run test:ci --workspaces --if-present
 
+      - name: Collect coverage
+        run: npm run coverage # This will run tests again, but also generate coverage
```
NTaylorMullen, line 94: I see, this means when things fail we'll have 2x failures though I imagine. Is there harm in making the "coverage" aspect of this a detail and making the Run tests step just do this?

NTaylorMullen (approved):
Sorry for the delay, I actually had pending comments yesterday that I somehow didn't publish?

.github/workflows/ci.yml
allenhutchison: I just wanted to be a little conservative here. I thought I'd get this into the CI and then look at merging the two. So this is a temp state.

PR #480, opened 2025-05-22
Refactor read-file and support images.

jacob314:
Modify turn.ts to support responses that are arbitrary `Part` objects rather than only strings.
Without this change, asking about a single image blows through Gemini's token limit.
Please review the changes to turn.ts most carefully in this pull request. The way I worked around the limitations on what structured data a tool response contains could be a bit of a hack. Would be great if there was a better way to do this.

Other things in this pull request:
Refactor duplicate code in read-file and read-many-files.
Made read-file also support images.
Added tests for impacted files.
Demo showing that you can now just ask about an image and Gemini CLI will find it and call the read_file tool to load it and then describe the result.
Screenshot 2025-05-21 at 5 20 50 PM

### DIFF
--- a/packages/server/src/utils/fileUtils.ts
+++ b/packages/server/src/utils/fileUtils.ts
@@ -1,0 +1,280 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import fs from 'fs';
+import path from 'path';
+import { PartUnion } from '@google/genai';
+import mime from 'mime-types';
+
+// Constants for text file processing
+const DEFAULT_MAX_LINES_TEXT_FILE = 2000;
+const MAX_LINE_LENGTH_TEXT_FILE = 2000;
+
+// Default values for encoding and separator format
+export const DEFAULT_ENCODING: BufferEncoding = 'utf-8';
+
+/**
+ * Checks if a path is within a given root directory.
+ * @param pathToCheck The absolute path to check.
+ * @param rootDirectory The absolute root directory.
+ * @returns True if the path is within the root directory, false otherwise.
+ */
+export function isWithinRoot(
+  pathToCheck: string,
+  rootDirectory: string,
+): boolean {
+  const normalizedPathToCheck = path.normalize(pathToCheck);
+  const normalizedRootDirectory = path.normalize(rootDirectory);
+
+  // Ensure the rootDirectory path ends with a separator for correct startsWith comparison,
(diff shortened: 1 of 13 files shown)


packages/server/src/utils/fileUtils.ts
```diff
@@ -0,0 +1,280 @@
...
+  // and it's not a known binary extension.
+  if (isBinaryFile(filePath)) {
+    return 'binary';
+  }
+
+  return 'text';
+}
+
+export interface ProcessedFileReadResult {
+  llmContent: string | Part; // string for text, Part for image/pdf/unreadable binary
```
NTaylorMullen: If you want to make this extra flexible you could just do `PartListUnion` OR if you want to acknowledge that this is going to be consumed by a tool you could surface it and returnDisplay as `toolResult: ToolResult`

packages/server/src/utils/fileUtils.ts
```diff
@@ -0,0 +1,280 @@
...
+          linesShown: [actualStartLine + 1, endLine],
+        };
+      }
+      case 'image':
+      case 'pdf': {
+        const contentBuffer = await fs.promises.readFile(filePath);
+        const base64Data = contentBuffer.toString('base64');
+        return {
+          llmContent: {
+            inlineData: {
```
NTaylorMullen, line 251: I actually didn't realize this would work... This is awesome!

packages/server/src/utils/fileUtils.ts
```diff
@@ -0,0 +1,280 @@
+/**
```
NTaylorMullen, line 1: Could you have GC add some tests for this file

NTaylorMullen (approved):

packages/server/src/utils/fileUtils.ts
jacob314: Switched to PartUnion which is the existing name for `string|Part`. PartListUnion would be ok but would make it slightly more clumsy to use from read-many-files where having it clear it is only a String | Part makes the case handling a tiny bit easier.

packages/server/src/utils/fileUtils.ts
jacob314: done

NTaylorMullen (approved):

jacob314:
Please take another look. All the content since when you reviewed last is in 
https://github.com/google-gemini/gemini-cli/pull/480/commits/667dc5b75779dde30237aa3d63b1057887cf6d32

NTaylorMullen (approved):

PR #481, opened 2025-05-22
fix: forward entire tool call confirmation object through useToolScheduler

KeijiBranshi:
file diffs were not getting displayed in the tool call confirmation boxes

### DIFF
--- a/packages/server/src/utils/nextSpeakerChecker.ts
+++ b/packages/server/src/utils/nextSpeakerChecker.ts
@@ -94,7 +94,7 @@ export async function checkNextSpeaker(
     return null;
   } catch (error) {
     console.warn(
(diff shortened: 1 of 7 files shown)


packages/server/src/utils/nextSpeakerChecker.ts
```diff
@@ -94,7 +94,7 @@ export async function checkNextSpeaker(
     return null;
   } catch (error) {
     console.warn(
-      'Failed to talk to Gemiin endpoint when seeing if conversation should continue.',
+      'Failed to talk to Gemini endpoint when seeing if conversation should continue.',
```
NTaylorMullen, line 97: lol nice

NTaylorMullen:
With this change because we were deconstructing it before we were losing the file diff info?

PR #487, opened 2025-05-22
interactive shell mode (for review only)

olcan:
(for review only)

enables interactive commands, i.e. any command that reads stdin for user input

requires a new dependency node-pty, which also requires 'pip install setuptools' or 'brew install python-setuptools' on mac; after extensive investigation, this appears to be the only reliable way to achieve interactivity for shell commands in node.js, at least on the mac

as a side benefit, creates a pseudo-terminal for the subprocess, which in principle would allow arbitrary interactive utilties (e.g. nano, vim, etc), although in quick testing there were limitations due to size and colors (mostly about ink theme carrying over to subprocess and not working well)


allenhutchison:
I was considering bringing in https://github.com/sindresorhus/open-editor for a subset of these use cases. It allows us to open a default editor on the users machine in a cross platform way. I realize this doesn't address all the cases you are trying to address in this PR, but maybe it gets us partway there without requiring the python bits.

olcan:
actually enabling cli editors was definitely not a goal here, and in fact i
would strongly prefer improving overall ux around the user's
preferred editors (e.g. vscode and obsidian for me). this whole thing
started with folks reporting issues with model trying to run interactive
commands like `npm init`, and then taylor getting excited about model
driving interactive tools and trying to get me fired up about it. :) so i
would say main goal here was to take some steps toward more graceful
handling of interactive commands; currently (with recent changes) they will
simply error out, which is also a fine place to be when focusing on
non-interactive commands

allenhutchison:
I see. I was jumping straight to the commands that open editors. We can proceed down that path in another way. One of my goals is to open an editor for memory files.

olcan:
ah cool, yeah preferred opening editors for memory, settings, etc, would be
awesome

NTaylorMullen:
Olcan and I chatted offline around this. I think the dependency is a bit too heavy to include at this stage in the product BUT knowing we can explore this is invaluable

olcan:
yeah closing and deleting remote branch; i'll keep around locally for
future reference

PR #493, opened 2025-05-22
Refactor(cli): Move memory add logic to server tool call

allenhutchison:
- Deletes packages/cli/src/config/memoryUtils.ts and its tests.
- Modifies slashCommandProcessor to return a directive for /memory add.
- Updates useGeminiStream to handle this directive and schedule the servers save_memory tool.
- Adjusts related types and tests in App.tsx and affected hooks.

First step for #472


NTaylorMullen (approved):
Nice unification. Might be worth digging through the diff real quick and deleting any extra non-high value comments

PR #502, opened 2025-05-23
feat: add emphasis to tool confirmations

KeijiBranshi:
DEMO=http://screen/BwDbY28GCYq2gDA

- use a green circle to indicate the status of 'scheduled' (but not yet 'executing') tool calls
- de-emphasize the text for non-focused tool calls during confirmation/cancellation
- added a trailing left-arrow indicator to the focused item during confirmation/cancellation
- drive by change: fix useToolScheduler to account for some but not all tool calls getting cancelled


NTaylorMullen:
Wow solid freaking UI choices here

PR #504, opened 2025-05-23
fix: do not retry cancelled tool calls

KeijiBranshi:
DEMO=http://screen/3iRCUm89PhQpEcz

we were hitting situations where the user would cancel a tool call, then the agent would immediately attempt to retry the cancelled tool call. 

I had a fair amount of luck with this prompt.


NTaylorMullen:
Ahh is this the case when we have multi-tool call confirmations and the user cancels one?

PR #506, opened 2025-05-23
feat: Modify loading indicator to support a paused state

jacob314:
The loading indicator (spinner, witty phrases, and elapsed time) will now pause and display "Waiting for user confirmation..." when the application is awaiting user input for a tool confirmation.

Changes:

-   `packages/cli/src/ui/hooks/useLoadingIndicator.ts`:
    -   Added an `isPaused: boolean` parameter to the `useLoadingIndicator` hook.
    -   The hook now returns an additional boolean `shouldShowSpinner`.
    -   When `streamingState` is `Responding` and `isPaused` is `true`:
        -   The `elapsedTime` counter is paused.
        -   `currentLoadingPhrase` is set to "Waiting for user confirmation...".
        -   `shouldShowSpinner` is set to `false`.
    -   Otherwise, the loading indicator behaves as before.
    -  Show a paused view of spinners within ToolMessage if the app is responding
       to ensure we don't flicker while responding if there are a mix of
       in progress and complete tasks.

This eliminates continual flicker while waiting for confirmation.

### DIFF
--- a/packages/cli/src/ui/App.tsx
+++ b/packages/cli/src/ui/App.tsx
@@ -182,8 +182,17 @@ export const App = ({
       handleSlashCommand,
       shellModeActive,
     );
-  const { elapsedTime, currentLoadingPhrase } =
-    useLoadingIndicator(streamingState);
+  const isPausedForConfirmation = useMemo(
+    () =>
+      pendingHistoryItems.some(
+        (item) =>
+          item?.type === 'tool_group' &&
+          item.tools.some((tool) => tool.status === 'Confirming'),
+      ),
+    [pendingHistoryItems],
+  );
+  const { elapsedTime, currentLoadingPhrase, shouldShowSpinner } =
+    useLoadingIndicator(streamingState, isPausedForConfirmation);
   const showAutoAcceptIndicator = useAutoAcceptIndicator({ config });
 
   const handleFinalSubmit = useCallback(
@@ -302,6 +311,7 @@ export const App = ({
               key={h.id}
               item={h}
               isPending={false}
+              streamingState={streamingState}
             />
           )),
         ]}
@@ -317,6 +327,7 @@ export const App = ({
             // HistoryItemDisplay. Refactor later. Use a fake id for now.
             item={{ ...item, id: 0 }}
             isPending={true}
+            streamingState={streamingState}
           />
         ))}
       </Box>
@@ -356,6 +367,7 @@ export const App = ({
           <>
             <LoadingIndicator
               isLoading={streamingState === StreamingState.Responding}
+              showSpinner={shouldShowSpinner}
               currentLoadingPhrase={currentLoadingPhrase}
               elapsedTime={elapsedTime}
             />
--- a/packages/cli/src/ui/components/messages/ToolMessage.tsx
+++ b/packages/cli/src/ui/components/messages/ToolMessage.tsx
@@ -7,7 +7,11 @@
 import React from 'react';
 import { Box, Text } from 'ink';
 import Spinner from 'ink-spinner';
-import { IndividualToolCallDisplay, ToolCallStatus } from '../../types.js';
+import {
+  IndividualToolCallDisplay,
+  StreamingState,
+  ToolCallStatus,
+} from '../../types.js';
 import { DiffRenderer } from './DiffRenderer.js';
 import { Colors } from '../../colors.js';
 import { MarkdownDisplay } from '../../utils/MarkdownDisplay.js';
@@ -21,6 +25,7 @@ export type TextEmphasis = 'high' | 'medium' | 'low';
 export interface ToolMessageProps extends IndividualToolCallDisplay {
   availableTerminalHeight: number;
   emphasis?: TextEmphasis;
+  streamingState?: StreamingState;
 }
 
 export const ToolMessage: React.FC<ToolMessageProps> = ({
@@ -30,6 +35,7 @@ export const ToolMessage: React.FC<ToolMessageProps> = ({
   status,
   availableTerminalHeight,
   emphasis = 'medium',
+  streamingState,
 }) => {
   const contentHeightEstimate =
     availableTerminalHeight - STATIC_HEIGHT - RESERVED_LINE_COUNT;
@@ -57,7 +63,7 @@ export const ToolMessage: React.FC<ToolMessageProps> = ({
     <Box paddingX={1} paddingY={0} flexDirection="column">
       <Box minHeight={1}>
         {/* Status Indicator */}
-        <ToolStatusIndicator status={status} />
+        <ToolStatusIndicator status={status} streamingState={streamingState} />
         <ToolInfo
           name={name}
           status={status}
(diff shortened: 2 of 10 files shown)


packages/cli/src/ui/components/messages/ToolMessage.tsx
```diff
@@ -99,15 +105,32 @@ export const ToolMessage: React.FC<ToolMessageProps> = ({
...
+        // executing until all confirmations are done. Showing a spinner would
+        // be misleading as the task is not actually executing at the moment
+        // and also has flickering issues due to Ink rendering limitations.
+        // If this hack becomes a problem, we can always add an additional prop
+        // indicating that the tool was indeed confirmed. If the tool was not
+        // confirmed we could show a paused version of the spinner.
+        <Text color={Colors.Gray}>✔</Text>
+      ) : (
+        <Spinner type="dots" />
+      ))}
```
KeijiBranshi, line 133: question: is this still needed?

I thought I updated this so that `ToolCallStatus.Executing` _does_ mean the tool is currently running. When it's been confirmed, but not yet executing, it's in the `ToolCallStatus.Pending` state. I had it show a green 'o' instead of the check though to avoid confusion with the `ToolCallStatus.Success` state.

packages/cli/src/ui/components/messages/ToolMessage.tsx
```diff
@@ -99,15 +105,32 @@ export const ToolMessage: React.FC<ToolMessageProps> = ({
   );
 };
 
-type ToolStatusIndicator = {
+type ToolStatusIndicatorProps = {
   status: ToolCallStatus;
+  streamingState?: StreamingState;
```
KeijiBranshi, line 110: thought(no-action): this is an example of "prop drilling" in React, where we pass the same prop down through multiple layers of React components. 

sometimes it's harmless and sometimes it's a code smell indicating we either need to:

- flatten our component tree, and/or
- break down object props (not relevant here), and/or
- use the Context API (more suitable when the same prop is drilled _many_ times over)

imo we don't need to address it here, but just something to keep in the back of our minds as our components grow in size/scope.

NTaylorMullen (approved):
This looks fine to me BUT if we want to have a paused state you'll probably want to undo the changes here: https://github.com/google-gemini/gemini-cli/pull/505

packages/cli/src/ui/hooks/useLoadingIndicator.ts
```diff
@@ -20,48 +23,79 @@ export const useLoadingIndicator = (streamingState: StreamingState) => {
...
-        setElapsedTime((prevTime) => prevTime + 1);
-      }, 1000);
-    } else if (timerRef.current) {
-      clearInterval(timerRef.current);
-      timerRef.current = null;
+      if (!isPaused) {
+        if (!timerRef.current) {
+          // No specific action needed here if timer wasn't running and we are not paused.
+          // Elapsed time continues from where it left off or starts from 0 if it's a fresh start.
+        }
```
KeijiBranshi, line 34: suggestion: remove since it does nothing

packages/cli/src/ui/hooks/useLoadingIndicator.ts
```diff
@@ -20,48 +23,79 @@ export const useLoadingIndicator = (streamingState: StreamingState) => {
   const phraseIntervalRef = useRef<NodeJS.Timeout | null>(null);
   const currentPhraseIndexRef = useRef<number>(0);
 
-  // Timer effect for elapsed time during loading
+  const [shouldShowSpinner, setShouldShowSpinner] = useState(true);
+
   useEffect(() => {
     if (streamingState === StreamingState.Responding) {
```
KeijiBranshi, line 29: thought: there is a `StreamingState.WaitingForConfirmation` state. I initially removed it since I thought `StreamingState.Responding` was sufficient (also i recall seeing the text input prompt appear in that state, even though it should be hidden). We could consider using that to dodge maintaining the separate isPaused bool.

packages/cli/src/ui/App.tsx
```diff
@@ -182,8 +182,17 @@ export const App = ({
...
+    () =>
+      pendingHistoryItems.some(
+        (item) =>
+          item?.type === 'tool_group' &&
+          item.tools.some((tool) => tool.status === 'Confirming'),
+      ),
+    [pendingHistoryItems],
+  );
+  const { elapsedTime, currentLoadingPhrase, shouldShowSpinner } =
+    useLoadingIndicator(streamingState, isPausedForConfirmation);
```
KeijiBranshi, line 195: suggestion(non-blocking): There is already a `StreamingState.WaitingForConfirmation` state. https://github.com/google-gemini/gemini-cli/blob/e9931816284aa7a2dbb8e97fa392a9c563425e2c/packages/cli/src/ui/hooks/useGeminiStream.ts#L417-L425

It's making the same check here. I also tried to set up `useToolScheduler` to hold off on tool execution until all tools have been confirmed anyway: https://github.com/google-gemini/gemini-cli/blob/e9931816284aa7a2dbb8e97fa392a9c563425e2c/packages/cli/src/ui/hooks/useToolScheduler.ts#L187-L190

We could consider using that to dodge maintaining the separate isPaused bool.

jacob314:
> This looks fine to me BUT if we want to have a paused state you'll probably want to undo the changes here: #505
Thanks. I've reverted #505 as part of this pull request. No strong opinion one way or the other.

KeijiBranshi (commented):
first pass. submitting now since I see I need to refresh the diff.

packages/cli/src/ui/hooks/useLoadingIndicator.ts
```diff
@@ -20,48 +23,79 @@ export const useLoadingIndicator = (streamingState: StreamingState) => {
...
+            currentPhraseIndexRef.current =
+              (currentPhraseIndexRef.current + 1) %
+              WITTY_LOADING_PHRASES.length;
+            setCurrentLoadingPhrase(
+              WITTY_LOADING_PHRASES[currentPhraseIndexRef.current],
+            );
+          }, PHRASE_CHANGE_INTERVAL_MS);
+        }
+      } else {
+        setShouldShowSpinner(false);
```
KeijiBranshi, line 78: thought: I'd need to fiddle with it a little to confirm, but can `shouldShowSpinner` just be a derived value? it seems like it's equivalent to:

```
const shouldShowSpinner = streamingState === StreamingState.Responding && !isPaused
```

doing it this way could avoid the extra cycle of running the effect, setting the state, and triggering the re-render

KeijiBranshi (approved):
approving to unblock. but left some comments for consideration

packages/cli/src/ui/components/messages/ToolMessage.tsx
jacob314: I think a ContextAPI would be appropriate for this. I'll send up a followup with that cleanup.

packages/cli/src/ui/hooks/useLoadingIndicator.ts
jacob314: done in follow up pull request.

packages/cli/src/ui/components/messages/ToolMessage.tsx
jacob314: Done. Switched this to a paused spinner and added a comment that this case should not be hit.
Open to also just removing this as this is somewhat paranoid defensive coding given your changes.

packages/cli/src/ui/hooks/useLoadingIndicator.ts
jacob314: tried to add this back in a followup pull request

packages/cli/src/ui/App.tsx
jacob314: done I think in followup pull request.

packages/cli/src/ui/hooks/useLoadingIndicator.ts
NTaylorMullen: Ya I think I ended up re-introducing WaitingForConfirmation in my one PR which I said you could revert. Might be a mix of bringing that change back to life and this

PR #507, opened 2025-05-23
Fix bug updating the cursor after navigating history.

jacob314:
The problem was calling 'home' immediately after setting the text resulted in inconsistent state as there wasn't the required async gapp to allow the state of textbuffer to update after calling setText. @KeijiBranshi curious if there is a clean way to avoid this issue.

The fix was as to just remove the call to move('home') as the call to  setText already moved the cursor to the end of the line as desired.
Added some tests although I left adding tests for InputPrompt itself for another pull request as that requires more thought to do reasonably.

### DIFF
--- a/packages/cli/src/ui/hooks/useInputHistory.ts
+++ b/packages/cli/src/ui/hooks/useInputHistory.ts
@@ -25,7 +25,7 @@ export function useInputHistory({
   onSubmit,
   isActive,
   currentQuery,
-  onChangeAndMoveCursor: setQueryAndMoveCursor,
+  onChangeAndMoveCursor,
 }: UseInputHistoryProps): UseInputHistoryReturn {
   const [historyIndex, setHistoryIndex] = useState<number>(-1);
   const [originalQueryBeforeNav, setOriginalQueryBeforeNav] =
@@ -65,14 +65,14 @@ export function useInputHistory({
     if (nextIndex !== historyIndex) {
       setHistoryIndex(nextIndex);
       const newValue = userMessages[userMessages.length - 1 - nextIndex];
-      setQueryAndMoveCursor(newValue); // Call the prop passed from parent
+      onChangeAndMoveCursor(newValue);
       return true;
     }
     return false;
   }, [
     historyIndex,
     setHistoryIndex,
-    setQueryAndMoveCursor,
+    onChangeAndMoveCursor,
     userMessages,
     isActive,
     currentQuery, // Use currentQuery from props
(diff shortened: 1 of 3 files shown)


packages/cli/src/ui/hooks/useInputHistory.ts
```diff
@@ -25,7 +25,7 @@ export function useInputHistory({
   onSubmit,
   isActive,
   currentQuery,
-  onChangeAndMoveCursor: setQueryAndMoveCursor,
+  onChangeAndMoveCursor,
 }: UseInputHistoryProps): UseInputHistoryReturn {
```
KeijiBranshi, line 29: nitpick: should we just make this `onChange` now?

Somewhat tangentially, I feel like there's a tott out there about conjunctions (ie. 'And' in this case) in function names

KeijiBranshi (approved):
> The problem was calling 'home' immediately after setting the text resulted in inconsistent state as there wasn't the required async gapp to allow the state of textbuffer to update after calling setText. @KeijiBranshi curious if there is a clean way to avoid this issue.

This fix looks super clean to me 🙂 . 

But if you mean more generally dealing with sequential updates on related state, then I might take that as a sign that state should be managed with a single composite useState under the hood. That way these separate states can be updated with setState in one atomic operation. Simpler wrapper 'setState' callbacks can be exported when consumers shouldn't need to know about those details. That seems to be happening with `setText` in this case?

Even more generally regarding sequential state updates, here's an official article on that: https://react.dev/learn/queueing-a-series-of-state-updates

Lastly, if youre state is just so complex that neither of the above will work, I think useReducer is the final power tool to reach for: https://react.dev/reference/react/useReducer

packages/cli/src/ui/hooks/useInputHistory.ts
jacob314: agree. tossed that cleanup here. I'm working on a personal presubmit code review hook that flags sloppy cases like this for cleanup.

jacob314:
> > The problem was calling 'home' immediately after setting the text resulted in inconsistent state as there wasn't the required async gapp to allow the state of textbuffer to update after calling setText. @KeijiBranshi curious if there is a clean way to avoid this issue.
> 
> This fix looks super clean to me 🙂 .
> 
> But if you mean more generally dealing with sequential updates on related state, then I might take that as a sign that state should be managed with a single composite useState under the hood. That way these separate states can be updated with setState in one atomic operation. Simpler wrapper 'setState' callbacks can be exported when consumers shouldn't need to know about those details. That seems to be happening with `setText` in this case?
> 
> Even more generally regarding sequential state updates, here's an official article on that: https://react.dev/learn/queueing-a-series-of-state-updates
> 
> Lastly, if youre state is just so complex that neither of the above will work, I think useReducer is the final power tool to reach for: https://react.dev/reference/react/useReducer

Thank you! Super useful resources that will help me think about this for future cases. That is how I understand the current setText behavior as well.

packages/cli/src/ui/hooks/useInputHistory.ts
jacob314: Agreed. Done. Added a explicit check for bad names like this to my personal prompt I am now running before sending out code for review.

PR #511, opened 2025-05-23
MCP SSE support

cornmander:
Add SSE transport for MCP.

Matches the config format used by other MCP clients.

### DIFF
--- a/packages/server/src/config/config.ts
+++ b/packages/server/src/config/config.ts
@@ -24,11 +24,16 @@ import { WebSearchTool } from '../tools/web-search.js';
 
 export class MCPServerConfig {
   constructor(
-    readonly command: string,
+    // For stdio transport
+    readonly command?: string,
     readonly args?: string[],
     readonly env?: Record<string, string>,
     readonly cwd?: string,
+    // For sse transport
+    readonly url?: string,
+    // Common
     readonly timeout?: number,
+    readonly transport: 'stdio' | 'sse' = 'stdio',
   ) {}
 }
 
--- a/packages/server/src/config/config.ts
+++ b/packages/server/src/config/config.ts
@@ -33,7 +33,6 @@ export class MCPServerConfig {
     readonly url?: string,
     // Common
     readonly timeout?: number,
-    readonly transport: 'stdio' | 'sse' = 'stdio',
   ) {}
 }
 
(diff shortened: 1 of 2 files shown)


packages/server/src/config/config.ts
```diff
@@ -24,11 +24,16 @@ import { WebSearchTool } from '../tools/web-search.js';
...
+    // For stdio transport
+    readonly command?: string,
     readonly args?: string[],
     readonly env?: Record<string, string>,
     readonly cwd?: string,
+    // For sse transport
+    readonly url?: string,
+    // Common
     readonly timeout?: number,
+    readonly transport: 'stdio' | 'sse' = 'stdio',
```
olcan: i was going to comment on this being redundant, but actually much more important is that our config/settings format was fully compatible (AFAIK) with other clients, at least Cursor and Claude Desktop, so we should try to maintain that; they don't use this extra field so we should not either, see e.g. https://docs.cursor.com/context/model-context-protocol. i'm not sure if the config format is published somewhere

olcan (commented):
looks good other than the non-standard config issue

NTaylorMullen (approved):

packages/server/src/config/config.ts
cornmander: Done.

olcan:
> looks good other than the non-standard config issue

I wanted to follow up on this based on a user on Gemini CLI Users reporting using transportType field in their config. I did some quick research and it seems Cline added this (initially as a required field but then made optional as it broke people's configs), presumably to distinguish between older "SSE" and the newer "Streamable HTTP" transport which was meant to replace SSE according to https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#streamable-http

Here is a relevant Client PR: https://github.com/cline/cline/pull/3714/files

I could not find any other meaningful reference to this setting. It is sort of annoying that there appears to be no published standard MCP client-server configurations though I may have missed something. That said, it may be safest to follow Cline and add this as an optional field if/when we want to support both SSE and StreamableHTTP.

cornmander:
VSC uses "type" as the field name: https://code.visualstudio.com/docs/copilot/chat/mcp-servers#_configuration-format

olcan:
🤦‍♂️

PR #512, opened 2025-05-23
Add an `/about` command

miguelsolorio:
Adds an `/about` command and moves CLI info into there to reduce the amount of info displayed in the footer

CleanShot 2025-05-23 at 10 35 39@2x

CleanShot 2025-05-23 at 10 35 47@2x


packages/cli/src/ui/hooks/slashCommandProcessor.ts
```diff
@@ -9,7 +9,7 @@ import { type PartListUnion } from '@google/genai';
 import open from 'open';
 import { UseHistoryManagerReturn } from './useHistoryManager.js';
 import { Config } from '@gemini-code/server';
-import { Message, MessageType, HistoryItemWithoutId } from '../types.js';
+import { Message, MessageType, HistoryItemWithoutId } from '../types.js'; // Removed HistoryItem
```
```suggestion
import { Message, MessageType, HistoryItemWithoutId } from '../types.js';
```

packages/cli/src/ui/types.ts
```diff
@@ -110,15 +119,26 @@ export enum MessageType {
   INFO = 'info',
   ERROR = 'error',
   USER = 'user',
+  ABOUT = 'about', // Added ABOUT type
```
NTaylorMullen:
```suggestion
  ABOUT = 'about',
```

NTaylorMullen (approved):
I love the about box but I'm not sure I'm a fan of the way GC added it as its own message type. Your changes seem pretty targeted though so maybe this is indeed the best route. We can go with this for now and course-correct if the message type expansion ends up being painful

miguelsolorio:
Closing in favor of #519

PR #513, opened 2025-05-23
feat: Publish test coverage summaries to PRs

allenhutchison:
This change introduces the ability to publish test coverage summaries directly to GitHub Pull Requests.

Key changes:
- Configured Vitest in both `cli` and `server` packages to generate Cobertura coverage reports (`cobertura-coverage.xml`) in addition to existing formats.
- Consolidated test execution and report generation in CI into a single `test:ci` script per workspace. This script now handles both JUnit XML and all coverage report generation.
- Updated the GitHub Actions workflow (`ci.yml`):
    - Replaced separate test and coverage collection steps with a single call to `npm run test:ci`.
    - Added a new step using `dorny/test-reporter` to parse the Cobertura reports and publish a coverage summary to PRs.
- Refined workspace `package.json` scripts:
    - The `test:ci` script in each workspace now explicitly defines JUnit and coverage reporters. This ensures `npm test` (local manual runs) does not generate JUnit XML, while `npm run test:ci` does.
    - Removed redundant `coverage` scripts from workspaces and the root `package.json`.


NTaylorMullen:
> Coverage report

sick

NTaylorMullen:
> ## Code Coverage Summary
> CLI Package Coverage Details
> Server Package Coverage Details
> _For detailed HTML reports, please see the 'coverage-reports-20.x' artifact from this CI run._

Wow if this is real that's higher % than I thought

allenhutchison:
The coverage results here are real, we're doing pretty good.

.github/workflows/ci.yml
```diff
@@ -102,6 +99,87 @@ jobs:
           reporter: java-junit
           fail-on-error: 'false'
 
+      - name: Prepare Coverage Comment
```
NTaylorMullen: Wow this is way more than I would have imagined. Epic in its own way. Is it possible to break out this part of the workflow into its own .yml file to keep the top level file simpler?

.github/workflows/ci.yml
```diff
@@ -102,6 +99,87 @@ jobs:
...
+        id: prep_coverage_comment
+        shell: bash
+        run: |
```
NTaylorMullen: Do we want to "continue on error" / make it so if this step fails it doesn't block the build?

.github/workflows/ci.yml
```diff
@@ -102,6 +99,87 @@ jobs:
...
+
+          echo "_For detailed HTML reports, please see the 'coverage-reports-${{ matrix.node-version }}' artifact from this CI run._" >> "$comment_file"
+
+          echo "comment_body<<EOF" >> "$GITHUB_OUTPUT"
+          cat "$comment_file" >> "$GITHUB_OUTPUT"
+          echo "EOF" >> "$GITHUB_OUTPUT"
+
+      - name: Post Coverage Comment
+        uses: thollander/actions-comment-pull-request@v3
+        if: always()
```
NTaylorMullen: ditto on when to run/continue

NTaylorMullen (approved):

NTaylorMullen:
> ## Code Coverage Summary
> Package	Lines	Statements	Functions	Branches
> CLI	41.62%	41.62%	47.22%	73%
> Server	64.82%	64.82%	67.03%	80.16%
> CLI Package - Full Text Report
> Server Package - Full Text Report
> _For detailed HTML reports, please see the 'coverage-reports-20.x' artifact from this CI run._

This is going to be fun to improve 😄

.github/workflows/ci.yml
allenhutchison: Yes, I was focused on just getting it to work, but now that we have something that works and looks ok I'll split it out into a function that we can call from the CI.

.github/workflows/ci.yml
allenhutchison: Agree, added.

.github/workflows/ci.yml
allenhutchison: done

NTaylorMullen:
Let the games begin 😄

PR #514, opened 2025-05-23
feat: Enable npx execution directly from GitHub URL

NTaylorMullen:
This commit modifies the packaging setup to allow the CLI to be executed directly from its GitHub URL using `npx`, for example: `npx https://github.com/google-gemini/gemini-cli` (once merged to main). This is achieved without requiring the bundle to be checked into the repository.

Key changes and motivations:

- Modify `scripts.prepare` to run `npm run bundle`: Ensures the CLI bundle is generated automatically when `npx` installs the package from a git URL. This replaces previous approaches (e.g., using `prepack`) which were not consistently triggered in the `npx` environment.

- Update `scripts.bundle` to use a direct path for `esbuild` and externalize `sqlite3`: Using `node_modules/.bin/esbuild` provides a more reliable way to invoke the bundler. Externalizing `sqlite3` is crucial for correctly handling its native addon, preventing runtime errors.

- Add `bin`, `files`, and root `sqlite3` dependency:
  - The `bin` field defines the `gemini` command.
  - The `files` array ensures the generated `bundle/` directory is recognized by npm.
  - `sqlite3` is added as a root dependency to ensure it's installed by `npx` when `gemini-code` is fetched, allowing the externalized module to be resolved.

These changes collectively ensure that the necessary build artifacts are created on-the-fly during `npx` installation, providing a seamless execution experience directly from the GitHub repository URL.

image

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -33,6 +33,7 @@ export default tseslint.config(
       'packages/cli/dist/**',
       'packages/server/dist/**',
       'eslint-rules/*',
+      'bundle/**',
     ],
   },
   eslint.configs.recommended,
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,6 +10,12 @@
       "workspaces": [
         "packages/*"
       ],
+      "dependencies": {
+        "sqlite3": "^5.1.7"
+      },
+      "bin": {
+        "gemini": "bundle/gemini.js"
+      },
       "devDependencies": {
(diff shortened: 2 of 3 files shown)


NTaylorMullen:
Getting this in to test against `main` (so I don't need to use branch syntax). Feel free to leave comments and i'll react after-the-fact

NTaylorMullen:
Works: 
image

PR #520, opened 2025-05-23
feat: Enable npx execution on tagged versions

NTaylorMullen:
- Modify root package.json to allow publishing
- Update "files" to include only essential bundled assets

This change makes it possible to use npx with tagged versions of the gemini-cli package (e.g., npx https://github.com/google-gemini/gemini-cli@early-access). It removes "private: true" and refines the "files" array to ensure that only the necessary bundled assets are included when the package is fetched by npx, optimizing download size and ensuring correct execution.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,6 @@
 {
   "name": "gemini-code",
   "version": "0.1.0",
-  "private": true,
   "type": "module",
   "workspaces": [
     "packages/*"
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -1,7 +1,6 @@
 {
   "name": "gemini-code",
   "version": "0.1.0",
-  "private": true,
```
NTaylorMullen, line 4: This is what's required to `npx` a tag in GitHub. Sadly this also means that if someone accidentally runs `publish` then it will go to npmjs early. It's unlikely any of us will accidentally publish so that's not super concerning BUT giving access to others makes this a bit more challenging because they will then have access

PR #521, opened 2025-05-23
Make console message support more robust to logging in the middle of rendering.

jacob314:
Use setTimeout to avoid crashing if messages are logged in the middle of rendering. Merge repeated log messages so logs are more likely to fit on the screen.

Demo showing repeated log messages.
Screenshot 2025-05-23 at 3 40 39 PM

### DIFF
--- a/packages/cli/src/ui/types.ts
+++ b/packages/cli/src/ui/types.ts
@@ -123,4 +123,5 @@ export interface Message {
 export interface ConsoleMessageItem {
   type: 'log' | 'warn' | 'error' | 'debug';
   content: string;
+  count?: number;
 }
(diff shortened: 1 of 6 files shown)


packages/cli/src/ui/types.ts
```diff
@@ -123,4 +123,5 @@ export interface Message {
 export interface ConsoleMessageItem {
   type: 'log' | 'warn' | 'error' | 'debug';
   content: string;
+  count?: number;
```
NTaylorMullen: Should make this required. Will always be at least 1 😄

NTaylorMullen (approved):
Absolutely love it. When I was hitting those 100x errors the other day this was exactly what I was hoping for

packages/cli/src/ui/types.ts
jacob314: absolutely. done

PR #486, opened 2025-05-22
Refactor: Improve console error/log display in CLI

jacob314:
This change refactors how console messages (errors, warnings, logs, debug output) are displayed in the CLI UI.

There is now a warning in the bottom right corner of the CLI with the number of errors if there are errors.
CTRL-D toggles showing and hiding a debug console showing the errors, warnings, and info.
@miguelsolorio please provide suggestions or submit follow up pull requests to make this more beautiful.
If this direction looks right I'll add some tests.
Screenshot 2025-05-22 at 12 42 34 AM
Screenshot 2025-05-22 at 12 42 23 AM

Key changes:

- Introduces a summary display (`ConsoleSummaryDisplay.tsx`) that shows a count of errors and warnings when `showErrorDetails` is false (e.g., `(!) N errors occurred (F12 for info)`).
- Introduces a detailed display (`DetailedMessagesDisplay.tsx`) that lists all console messages with appropriate coloring when `showErrorDetails` is true.
- CTRL-D now toggles the `showErrorDetails` state in `App.tsx` to switch between the summary and detailed views.
- `ConsolePatcher.tsx` (formerly `ConsoleOutput.tsx`) has been refactored to no longer render messages directly. Instead, it calls an `onNewMessage` prop passed from `App.tsx`.
- `App.tsx` now manages the state for console messages and controls the visibility of the summary and detailed views.
- The old console output has been removed from `Footer.tsx`.
- Resolve
d various TypeScript and ESLint issues that arose during the refactoring, particularly around Ink key handling and type definitions

PR #524, opened 2025-05-24
Add additional readline-like keybindings.

dewitt:
Adds the following conventional readline-like keybindings:

  - `Ctrl+H`: Delete the previous character.
  - `Ctrl+D`: Delete the next character.

Additionally, remaps the Debug Console command from Ctrl+D to Ctrl+O, which had been first introduced in PR #486.

### DIFF
--- a/packages/cli/src/ui/App.tsx
+++ b/packages/cli/src/ui/App.tsx
@@ -84,8 +84,7 @@ export const App = ({
     [consoleMessages],
   );
   useInput((input: string, key: InkKeyType) => {
-    // Check for Ctrl+D key press
-    if (key.ctrl && (input === 'd' || input === 'D')) {
+    if (key.ctrl && input === 'o') {
       setShowErrorDetails((prev) => !prev);
       refreshStatic();
     }
--- a/packages/cli/src/ui/components/ConsoleSummaryDisplay.tsx
+++ b/packages/cli/src/ui/components/ConsoleSummaryDisplay.tsx
@@ -27,7 +27,7 @@ export const ConsoleSummaryDisplay: React.FC<ConsoleSummaryDisplayProps> = ({
       {errorCount > 0 && (
         <Text color={Colors.AccentRed}>
           {errorIcon} {errorCount} error{errorCount > 1 ? 's' : ''}{' '}
-          <Text color={Colors.SubtleComment}>(CTRL-D for details)</Text>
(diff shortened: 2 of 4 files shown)


dewitt:
I just noticed that PR #486 used a common readline keybinding (ctrl+d) for the debugger window. Makes sense, but I wanted to see here if we could claim it back for deleting the next character. 

I didn't really have a great idea for what to change the debugger to, though, so I put ctrl+o for now as a placeholder.

In the process, I found ink's support for keybinding to be frustrating and unpredictable in general. I wanted to do something like ctrl+\ or ctrl+shift+D but none of the obvious things worked. We should revisit down the road.

CC @jacob314.

NTaylorMullen:
> I just noticed that PR https://github.com/google-gemini/gemini-cli/pull/486 used a common readline keybinding (ctrl+d) for the debugger window. Makes sense, but I wanted to see here if we could claim it back for deleting the next character.

100% claim it back, that just went in so no harm no foul. Are we able to do F12 (thinking developer console)?

NTaylorMullen:
> In the process, I found ink's support for keybinding to be frustrating and unpredictable in general. 

Agree, I had similar difficulty the other night trying to enable the `delete` key and ran into this bug: https://github.com/vadimdemedes/ink/pull/639 . We could implement our own `OnInput` and take full control of this if we want to (we probably will want to)

NTaylorMullen (approved):

dewitt:
>  Are we able to do F12 (thinking developer console)?

F12 was the first thing I tried.  : )  I can't even see the function keys in ink. It might be doable, but it's going to take some surgery. I just wanted to get ctrl+d reclaimed, then decide if we can get something better long-term than whatever placeholder we choose for the debugger for now.

dewitt:
And I do want to just say on the record I have much respect to ink for getting as much done here as it has. Just look at [src/parse-keypress.ts](https://github.com/vadimdemedes/ink/blob/991aaa54eae39e1ad6d1964e747ea936d5a4ad12/src/parse-keypress.ts).

This looks hard because it _is_ hard. We shouldn't be naive about how much work it would be to do this any better that it already is.

dewitt:
@jacob314 -- heads up that I'm going to merge this now, just to get the conventional keys working again. Happy to work with you to sort out a plan for the debugger, and please shout if you think this should be rolled back.

jacob314 (commented):
lgtm
Thanks for cleaning this up!

PR #525, opened 2025-05-24
Code review comment fixes and some refactors.

jacob314:
No intentional different behavior aside for tweaks suggested from the code review of https://github.com/google-gemini/gemini-cli/pull/506 Refactor: Extract console message logic to custom hook

This commit refactors the console message handling from `App.tsx` into a new custom hook `useConsoleMessages`.

This change improves the testability of the console message logic and declutters the main `App` component.

- Created `useConsoleMessages.ts` to encapsulate console message state and update logic.
- Updated `App.tsx` to utilize the new `useConsoleMessages` hook.
- Added unit tests for `useConsoleMessages.ts` to ensure its functionality.

I deleted and started over on LoadingIndicator.test.tsx as I spent way too much time trying to fix it before just regenerating the tests as the code was easier to write tests for from scratch and the existing tests were not that good (I added them in the previous pull request).

### DIFF
--- a/packages/cli/src/ui/App.tsx
+++ b/packages/cli/src/ui/App.tsx
@@ -41,6 +41,10 @@ import { useHistory } from './hooks/useHistoryManager.js';
 import process from 'node:process';
 import { getErrorMessage, type Config } from '@gemini-code/server';
 import { useLogger } from './hooks/useLogger.js';
+import {
+  StreamingContext,
+  StreamingContextType,
+} from './contexts/StreamingContext.js';
 
 interface AppProps {
   config: Config;
@@ -174,18 +178,14 @@ export const App = ({
       handleSlashCommand,
       shellModeActive,
     );
-  const isPausedForConfirmation = useMemo(
-    () =>
-      pendingHistoryItems.some(
-        (item) =>
-          item?.type === 'tool_group' &&
-          item.tools.some((tool) => tool.status === 'Confirming'),
-      ),
-    [pendingHistoryItems],
-  );
-  const { elapsedTime, currentLoadingPhrase, shouldShowSpinner } =
-    useLoadingIndicator(streamingState, isPausedForConfirmation);
+  const { elapsedTime, currentLoadingPhrase } =
+    useLoadingIndicator(streamingState);
   const showAutoAcceptIndicator = useAutoAcceptIndicator({ config });
+
+  const streamingContextValue: StreamingContextType = useMemo(
+    () => ({ streamingState }),
+    [streamingState],
+  );
 
   const handleFinalSubmit = useCallback(
     (submittedValue: string) => {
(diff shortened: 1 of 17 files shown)


packages/cli/src/ui/App.tsx
```diff
@@ -278,177 +278,176 @@ export const App = ({
...
-       */}
-      <Static
-        key={staticKey}
-        items={[
-          <Box flexDirection="column" key="header">
-            <Header />
-            <Tips />
-          </Box>,
-          ...history.map((h) => (
+    <StreamingContext.Provider value={streamingContextValue}>
```
jacob314, line 281: make sure to hide widespace when viewing this diff. There is no change to this block other than wrapping with StreamingContext.Provider.

packages/cli/src/ui/components/messages/ToolMessage.tsx
```diff
@@ -107,48 +106,43 @@ export const ToolMessage: React.FC<ToolMessageProps> = ({
...
+  const { streamingState } = useStreamingContext();
+  return (
+    <Box minWidth={STATUS_INDICATOR_WIDTH}>
+      {status === ToolCallStatus.Pending && (
+        <Text color={Colors.AccentGreen}>o</Text>
+      )}
+      {status === ToolCallStatus.Executing &&
+        (streamingState === StreamingState.Responding ? (
+          <Spinner type="dots" />
+        ) : (
```
jacob314, line 122: change of behavior here. I think I had this backwards here but failed to notice it last night.

packages/cli/src/ui/components/LoadingIndicator.tsx
```diff
@@ -8,36 +8,36 @@ import React from 'react';
...
 
 export const LoadingIndicator: React.FC<LoadingIndicatorProps> = ({
-  isLoading,
-  showSpinner,
   currentLoadingPhrase,
   elapsedTime,
   rightContent,
 }) => {
-  if (!isLoading) {
+  const { streamingState } = useStreamingContext();
```
NTaylorMullen, line 25: Ya way cleaner, love this choice

packages/cli/src/ui/components/LoadingIndicator.tsx
```diff
@@ -8,36 +8,36 @@ import React from 'react';
...
-      {showSpinner && (
+      {streamingState === StreamingState.Responding && (
         <Box marginRight={1}>
           <Spinner type="dots" />
         </Box>
       )}
       <Text color={Colors.AccentPurple}>
         {currentLoadingPhrase}
-        {isLoading && ` (esc to cancel, ${elapsedTime}s)`}
+        {` (esc to cancel${streamingState === StreamingState.WaitingForConfirmation ? '' : `, ${elapsedTime}s`})`}
```
NTaylorMullen: When confirmations are showing it already has a `No (esc)` next to it. This might be duplicative, We oculd probably just hide the `(esc to cancel)` entirely if we're waiting for a confirmation

packages/cli/src/ui/components/messages/ToolMessage.test.tsx
```diff
@@ -8,6 +8,7 @@ import { render } from 'ink-testing-library';
 import { ToolMessage, ToolMessageProps } from './ToolMessage.js';
 import { StreamingState, ToolCallStatus } from '../../types.js';
 import { Text } from 'ink';
+import { StreamingContext } from '../../contexts/StreamingContext.js'; // Import the context
```
NTaylorMullen: Man I was hoping me elevating the system prompt core mandates would fix these comments. Maybe we need to add that bit also in the final reminder or something. Gemini just loves talking in comments

```suggestion
import { StreamingContext } from '../../contexts/StreamingContext.js';
```

packages/cli/src/ui/components/messages/ToolMessage.tsx
NTaylorMullen: Ah I thought this was intentional originally but admittedly did second-look at it. I like your new path.

packages/cli/src/ui/components/messages/ToolMessage.tsx
```diff
@@ -107,48 +106,43 @@ export const ToolMessage: React.FC<ToolMessageProps> = ({
...
+          // Paused spinner to avoid flicker. This case should not currently
```
NTaylorMullen: If the streaming state is waiting for confirmation with a multi-confirmation I thinkkkk this might get hit? Maybe I'm missing something though

packages/cli/src/ui/hooks/useGeminiStream.ts
```diff
@@ -42,7 +42,7 @@ import { useToolScheduler, mapToDisplay } from './useToolScheduler.js';
 
 enum StreamProcessingStatus {
   Completed,
-  PausedForConfirmation,
+  // PausedForConfirmation, // Removed as streamingState now covers this
```
NTaylorMullen:
```suggestion
```

packages/cli/src/ui/hooks/useGeminiStream.ts
```diff
@@ -486,7 +496,7 @@ export const useGeminiStream = (
     },
     // eslint-disable-next-line react-hooks/exhaustive-deps
     [
-      isResponding,
+      streamingState, // Added streamingState
```
NTaylorMullen: Few extra comments to remove in this block

packages/cli/src/ui/hooks/useLoadingIndicator.ts
```diff
@@ -4,98 +4,54 @@
...
+      setTimerResetKey((prevKey) => prevKey + 1);
+      setRetainedElapsedTime(0); // Clear retained time when going back to responding
+    } else if (
+      streamingState === StreamingState.Idle &&
+      prevStreamingStateRef.current === StreamingState.Responding
+    ) {
+      setTimerResetKey((prevKey) => prevKey + 1); // Reset timer when becoming idle from responding
+      setRetainedElapsedTime(0);
+    } else if (streamingState === StreamingState.WaitingForConfirmation) {
+      // Capture the time when entering WaitingForConfirmation
```
NTaylorMullen, line 42: Oh super interesting! Great change

packages/cli/src/ui/hooks/usePhraseCycler.ts
```diff
@@ -0,0 +1,68 @@
...
+        clearInterval(phraseIntervalRef.current);
+        phraseIntervalRef.current = null;
+      }
+    } else if (isActive) {
+      if (phraseIntervalRef.current) {
+        clearInterval(phraseIntervalRef.current);
+      }
+      // Reset to the first witty phrase when starting to respond
+      currentPhraseIndexRef.current = 0;
+      setCurrentLoadingPhrase(WITTY_LOADING_PHRASES[0]);
```
NTaylorMullen, line 55: We could probably lift some of these constants into this file. At this point there's not much in the constants file 😄

packages/cli/src/ui/hooks/useTimer.ts
```diff
@@ -0,0 +1,65 @@
...
+
+import { useState, useEffect, useRef } from 'react';
+
+/**
+ * Custom hook to manage a timer that increments every second.
+ * @param isActive Whether the timer should be running.
+ * @param resetKey A key that, when changed, will reset the timer to 0 and restart the interval.
+ * @returns The elapsed time in seconds.
+ */
+export const useTimer = (isActive: boolean, resetKey: unknown) => {
```
NTaylorMullen, line 15: Nice extraction here

NTaylorMullen (approved):
These are some epic refactors in a really solid way. Hats off man.

packages/cli/src/ui/hooks/useGeminiStream.ts
```diff
@@ -123,85 +138,100 @@ export const useGeminiStream = (
...
 
-  const prepareQueryForGemini = async (
-    query: PartListUnion,
-    userMessageTimestamp: number,
-    signal: AbortSignal,
-  ): Promise<{ queryToSend: PartListUnion | null; shouldProceed: boolean }> => {
-    if (typeof query === 'string' && query.trim().length === 0) {
-      return { queryToSend: null, shouldProceed: false };
-    }
+  const prepareQueryForGemini = useCallback(
```
jacob314, line 140: There were a bunch of members here that needed to be useCallback to fix the ignored lint removed bellow. These changes are mechanical as the lint robustly identifies the dependencies.

packages/cli/src/ui/hooks/useGeminiStream.ts
```diff
@@ -484,19 +525,16 @@ export const useGeminiStream = (
         setIsResponding(false);
       }
     },
```
jacob314, line 526: the exhaustive-deps lint is a scary one to ignore as ignoring it will result in hard to track down bugs. Given this I opted to fix this lint and add useCallback to multiple other members in this method as the risk of hard to detect stale callback bugs is otherwise quite high.

packages/cli/src/ui/components/LoadingIndicator.tsx
jacob314: Done.

packages/cli/src/ui/components/messages/ToolMessage.test.tsx
jacob314: > Man I was hoping me elevating the system prompt core mandates would fix these comments. Maybe we need to add that bit also in the final reminder or something. Gemini just loves talking in comments
Done. Yeah I'm still seeing a few but not as many as before so not on as high an alert to remove.

packages/cli/src/ui/components/messages/ToolMessage.tsx
jacob314: ok removing the "Ttis case should not currently be hit" part. I was worried I might just be defensively coding.

packages/cli/src/ui/hooks/useGeminiStream.ts
jacob314: Done.

packages/cli/src/ui/hooks/useGeminiStream.ts
jacob314: Fixed this as well as the worrying issue on line 497 that we were ignoring the great exhaustive-deps lint so this callback was likely to be stale unless we were really lucky.

packages/cli/src/ui/hooks/usePhraseCycler.ts
jacob314: agreed these were a little silly to have in constants.ts. moved here.

packages/cli/src/ui/contexts/StreamingContext.tsx
```diff
@@ -0,0 +1,26 @@
...
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import React, { createContext } from 'react';
+import { StreamingState } from '../types.js';
+
+export interface StreamingContextType {
+  streamingState: StreamingState;
```
KeijiBranshi, line 11: nitpick: I think `StreamingContextType` could just be a `string`. No need to wrap in an object. This would clean up some of the context provider code too

packages/cli/src/ui/contexts/StreamingContext.tsx
```diff
@@ -0,0 +1,26 @@
...
+}
+
+export const StreamingContext = createContext<StreamingContextType | undefined>(
+  undefined,
+);
+
+export const useStreamingContext = (): StreamingContextType => {
+  const context = React.useContext(StreamingContext);
+  if (context === undefined) {
+    throw new Error(
```
KeijiBranshi, line 21: suggestion: I would discourage throwing errors in React components and hooks as it creates the entire UI tree. I typically see React code assign errors to a `useState`. Then consuming components render conditionally based on if that state is set. 

Also, I thought useContext already handles this check for a Provider

packages/cli/src/ui/contexts/StreamingContext.tsx
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+import React, { createContext } from 'react';
+import { StreamingState } from '../types.js';
+
+export interface StreamingContextType {
+  streamingState: StreamingState;
+}
+
+export const StreamingContext = createContext<StreamingContextType | undefined>(
```
KeijiBranshi, line 14: thought,suggestion(non-blocking): I realize I was the one to bring up context as an option, but I probably should've been more explicit about discouraging it. Context really should only be used for data that nearly _every_ component needs (theme/colors are the go to example). StreamingContext only seems to be needed by the input prompt area.

The more encouraged approach to fix deep prop-drilling is to flatten (normalize?) the UI tree - often by:

a. Removing single component wrappers (eg avoiding components like `const Component = () => <Subcomponent />`), or
b. Hoisting subcomponents up as props or children:

```
// instead of forwarding props to children...
const App = () => {
  const foo = useItem()
  const bar = useItem()
  return <List foo={foo} bar={bar} />
}
const List = ({ foo, bar }) => 
  <Flex direction='column'>
    <ListItem item={foo} />
    <ListItem item={bar} />
  </Flex>

// Pass the children as props themselves
const App = () => {
  const foo = useItem()
  const bar = useItem()
  return <List>
    <ListItem item={foo} />
    <ListItem item={bar} />
  </List>
}
const List = ({ children }) => 
  <Flex direction='column'>
    {children}
  </Flex>
```

packages/cli/src/ui/components/messages/ToolMessage.tsx
NTaylorMullen: Actually I kind of like the simplicity of the gray checkmark after playing with this today. Otherwise you get 2 identical spinners. One for the loading indicator and one for each tool that's running.

Alternatively we could change this spinner to be something from here: https://github.com/sindresorhus/cli-spinners

I'll play with it.

packages/cli/src/ui/contexts/StreamingContext.tsx
jacob314: Interesting. Why should context only be used for data that nearly every component needs? My understanding is that React's createContext primarily impacts the performance of components that consume the context, not every component in your application. Does that match your expectations? When the value of a Context Provider changes, only the components that are subscribed to that specific context will re-render. Given that I would think that this is an appropriate use for context. Components that need to be StreamingContext aware can be made so without flowing a bunch of logic through intermediate components that do not care about StreamingContext. Are there some other downsides of using context that you are concerned about?
If there are other downsides happy to try the "hoisting subcomponets up as props or children." approach.

packages/cli/src/ui/contexts/StreamingContext.tsx
jacob314: Good point. Cleaned up with https://github.com/google-gemini/gemini-cli/pull/585

packages/cli/src/ui/contexts/StreamingContext.tsx
jacob314: Agree we should generally avoid throwing errors in components and hooks but I think this case should be an exception. I'd rather have the whole UI fail to render if you have forgotten to provide a StreamingContext than to have it quietly use a default value which as far as I can tell is the only alternative with React.useContext.

PR #526, opened 2025-05-24
feat: Add GEMINI.md tip to UI

NTaylorMullen:
- Display a tip to create a GEMINI.md file if one doesn't exist.
- Pass config to Tips component so it can inspect the initial GEMINI.md count.
image

### DIFF
--- a/packages/cli/src/ui/components/Tips.tsx
+++ b/packages/cli/src/ui/components/Tips.tsx
@@ -7,20 +7,39 @@
 import React from 'react';
 import { Box, Text } from 'ink';
 import { Colors } from '../colors.js';
+import { type Config } from '@gemini-code/server';
 
-export const Tips: React.FC = () => (
-  <Box flexDirection="column" marginBottom={1}>
-    <Text color={Colors.Foreground}>Tips for getting started:</Text>
-    <Text color={Colors.Foreground}>
-      1.{' '}
-      <Text bold color={Colors.AccentPurple}>
-        /help
-      </Text>{' '}
-      for more information.
-    </Text>
-    <Text color={Colors.Foreground}>
-      2. Ask coding questions, edit code or run commands.
-    </Text>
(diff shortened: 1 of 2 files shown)


packages/cli/src/ui/components/Tips.tsx
```diff
@@ -7,20 +7,39 @@
...
+        for more information.
+      </Text>
+      <Text color={Colors.Foreground}>
+        2. Ask coding questions, edit code or run commands.
+      </Text>
+      <Text color={Colors.Foreground}>
+        3. Be specific for the best results.
+      </Text>
+      {geminiMdFileCount === 0 && (
+        <Text color={Colors.Foreground}>
```
NTaylorMullen, line 35: Would be super curious about folks take on the verbiage here. Especially @miguelsolorio

jacob314: Like that this only shows if there isn't a GEMINI.md file!
One idea. Tell users to run /init to create a GEMINI.md file as. part of the mesage. The init command could then have a lot of verbiage on how to create a file or create one for them after going through a canned prompt. 
I could also see this as going under a future /doctor flow rather than creating a separate /init flow just for this. The gemini doctor command could always prompt you to create a GEMINI.md file if you don't have one as well as visualizing whether you have one for your project and your current directory. It would also be a natural place to suggest copying over existing AGENT.md, AGENTS.md and similar files as a basis for your GEMINI.md.
Over on the Flutter CLI tool, we got a lot of mileage out of users getting used to running the doctor command to get their environment in a good state.

jacob314 (approved):
lgtm

packages/cli/src/ui/components/Tips.tsx
NTaylorMullen: Ya totalllly agree. I think having a more full-purpose GEMINI.md building experience would be really really nice. 

I think `/init` may be a bit bigger of a task though so I'm going to defer to a later date for now.

PR #540, opened 2025-05-25
refactor: Decouple new_string correction from old_string

NTaylorMullen:
- Previously, `new_string` was assumed to be over-escaped if `old_string` was.
- This change introduces an explicit check (`newStringPotentiallyEscaped`) to determine if `new_string` itself needs correction.
- If `new_string` is potentially escaped, its corrected using an LLM call; otherwise, the original `new_string` is used.
- This avoids unnecessary corrections to `new_string` when only `old_string` was problematic.


NTaylorMullen:

PR #541, opened 2025-05-25
Upon finding > 1 occurrences do not attempt auto-correction.

NTaylorMullen:
- When correcting edits prior to this if we found more than one occurrence we would try to auto-correct the old/new strings. There's no need in this situation because the tool has already provided too vague of an old_string to act upon. Instantly return.


NTaylorMullen:

PR #543, opened 2025-05-25
Correct edits even when auto-accept is enabled.

NTaylorMullen:
- Prior to this when a user would turn on auto-accept for edits we'd stop ensuring correct edits. This would result in a lot of back and forth by the model. This change also incoporates ensure correct edit into the normal execution flow.
- Added edit tests for this.


NTaylorMullen:

PR #544, opened 2025-05-25
Ensure edit correction isn't re-done after confirm.

NTaylorMullen:
- Edit corretion leans on LLM-isms to ensure we properly fix poorly escaped content. Beacues of this we need to ensure that we don't re-run edit correction in many cases.
  - To ensure this an `LruCache` has been added to capture intermediate steps of edit correction to avoid re-computations.
  - Max cache size is 50 currently. This means a user can have a muti-confirmation flow of 25 items without recomputing anything (assuming they all break edit correction).
- Laid some groundwork for future testing.

### DIFF
--- a/packages/server/src/utils/LruCache.ts
+++ b/packages/server/src/utils/LruCache.ts
@@ -1,0 +1,41 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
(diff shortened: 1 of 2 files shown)


NTaylorMullen:

PR #546, opened 2025-05-25
Fix(test): Improve write-file and editCorrector test suites

NTaylorMullen:
- Enhanced  by:
  - Mocking  and  utilities (, ) to allow for more focused unit testing of .
  - Adding comprehensive tests for the private  method, covering new and existing file scenarios, as well as error handling.
  - Expanding tests for  and  to verify behavior with the new content correction logic, including diff generation and directory creation.
- Refined  by:
  - Introducing robust mocking for  and its methods (, , ) to simulate LLM interactions accurately.
  - Adding extensive test scenarios for , categorized by how  matches and how  is processed, including direct matches, unescaping, and LLM-based corrections.
  - Including tests for edge cases like no matches or multiple matches.
  - Adding a  utility for better test isolation.

Final fix for https://github.com/google-gemini/gemini-cli/issues/484

### DIFF
--- a/packages/server/src/tools/write-file.test.ts
+++ b/packages/server/src/tools/write-file.test.ts
@@ -4,16 +4,49 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
+import {
+  describe,
+  it,
+  expect,
+  beforeEach,
+  afterEach,
+  vi,
+  type Mocked,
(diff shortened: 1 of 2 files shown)


NTaylorMullen:
Alright this is the final fix (famous last words) for #484. In all seriousness though I imagine things like 484 will continue to happen until the Gemini model course corrects and becomes better with escaping. We'll need to balance new changes in this area to ensure things continue to work well. These test suites should be a good starting point.

PR #554, opened 2025-05-26
Fix(chat): Prevent empty model response after function call

NTaylorMullen:
- Addresses a Gemini model bug where it may return an empty content object after a function response.
- Previously, the SDK attempted to inject an empty model message, which could disrupt curated history.
- This change modifies our custom  class to detect this scenario using an  utility and avoid pushing an unnecessary empty model message, thus preserving history integrity.

Workaround for https://b.corp.google.com/issues/420354090 Part of https://github.com/google-gemini/gemini-cli/issues/551

### DIFF
--- a/packages/server/src/core/geminiChat.ts
+++ b/packages/server/src/core/geminiChat.ts
@@ -16,6 +16,7 @@ import {
   GoogleGenAI,
   createUserContent,
 } from '@google/genai';
+import { isFunctionResponse } from '../utils/messageInspectors.js';
 
 /**
  * Returns true if the response is valid, false otherwise.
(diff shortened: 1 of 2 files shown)


packages/server/src/core/geminiChat.ts
```diff
@@ -292,12 +293,15 @@ export class GeminiChat {
...
     } else {
-      // Appends an empty content when model returns empty response, so that the
+      // When not a function response appends an empty content when model returns empty response, so that the
       // history is always alternating between user and model.
-      outputContents.push({
-        role: 'model',
-        parts: [],
-      } as Content);
+      // Workaround for: https://b.corp.google.com/issues/420354090
+      if (!isFunctionResponse(userInput)) {
```
olcan, line 299: should't this be a check for whether it is a model-role response instead, since otherwise there could still be multiple model responses back to back?

olcan: another thing is that i think it would safer if we can not only ensure alternation but also ensure that the messages are always non-empty, because i'd be concerned about empty messages getting treated differently (e.g. dropped or merged) in the backend

NTaylorMullen: AFAIK function responses are always user role. Maybe I'm misunderstanding your question.

PR #565, opened 2025-05-26
feat: Replace SQLite with JSON logging for macOS sandbox compatibility

NTaylorMullen:
- Removes the sqlite3 dependency and refactors the logging mechanism to use a JSON file (logs.json) instead of a database.
- This change is a temporary workaround to address issues with macOS sandboxing that were caused by the SQLite native module.
- Storing all logs in a single JSON file may introduce scalability concerns in the future.

image

Fixes https://github.com/google-gemini/gemini-cli/issues/522

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,9 +10,6 @@
       "workspaces": [
         "packages/*"
       ],
-      "dependencies": {
-        "sqlite3": "^5.1.7"
-      },
       "bin": {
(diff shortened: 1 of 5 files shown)


NTaylorMullen:
@sethtroisi feel free to revert this when you get back if needed. GC nearly one-shotted this replacement to unblock MacOS sandboxing for the time being.

PR #567, opened 2025-05-27
Display git commit info in the /about section.

dewitt:
This change detects the most recent git commit short hash and writes it to the `GIT_COMMIT_INFO` constant in `packages/cli/src/generated/git-commit.sh`, optionally appending the string "(local modifications)" if additional local changes after that commit are detected.

If set, this string is displayed in the `/about` dialog as well as passed into the `/bug` template.

Example:

```
> /about

╭───────────────────────────────────────────────────────────────────────────╮
│                                                                           │
│ About Gemini CLI                                                          │
│                                                                           │
│ CLI Version               development                                     │
│ Git Commit                43370ab (local modifications)                   │
│ Model                     gemini-2.5-pro-preview-05-06                    │
│ Sandbox                   sandbox-exec (minimal)                          │
│ OS                        darwin v23.11.0                                 │
│                                                                           │
╰───────────────────────────────────────────────────────────────────────────╯
```

Additionally, this change updates `.gitignore` to ignore the generated files, `scripts/clean.sh` to remove them, and adds a `npm run generate` stage for this and any other generators we need to write.


dewitt:
For discussion, two changes I might make before sending this out for review:

1. Persist the git version info to a generated file, rather than using the environment. This will make it easier and more consistent to find the app metadata on startup, since it can be included as data in the build itself, rather than relying on the environment to propagate build-time metadata.

2. Simply read the metadata directly when rendering the AboutBox page, rather than passing it along as optional state throughout the entire app. If we're going to use it elsewhere, we should plumb it through, but I don't think we will.

NTaylorMullen:
> For discussion, two changes I might make before sending this out for review:
> 
> 1. Persist the git version info to a generated file, rather than using the environment. This will make it easier and more consistent to find the app metadata on startup, since it can be included as data in the build itself, rather than relying on the environment to propagate build-time metadata.
> 2. Simply read the metadata directly when rendering the AboutBox page, rather than passing it along as optional state throughout the entire app. If we're going to use it elsewhere, we should plumb it through, but I don't think we will.

I like both of these suggestions.

packages/cli/src/ui/components/AboutBox.tsx
```diff
@@ -44,6 +46,18 @@ export const AboutBox: React.FC<AboutBoxProps> = ({
         <Text>{cliVersion}</Text>
       </Box>
     </Box>
+    {gitCommitInfo && gitCommitInfo !== 'N/A' && (
```
NTaylorMullen: Why have the `N/A` bit at all vs. leave undefined?

packages/cli/src/ui/components/HistoryItemDisplay.tsx
```diff
@@ -53,6 +53,7 @@ export const HistoryItemDisplay: React.FC<HistoryItemDisplayProps> = ({
         osVersion={item.osVersion}
         sandboxEnv={item.sandboxEnv}
         modelVersion={item.modelVersion}
+        gitCommitInfo={item.gitCommitInfo} // Added this line
```
NTaylorMullen:
```suggestion
        gitCommitInfo={item.gitCommitInfo}
```

packages/cli/src/ui/hooks/slashCommandProcessor.ts
```diff
@@ -56,6 +56,7 @@ export const useSlashCommandProcessor = (
           osVersion: message.osVersion,
           sandboxEnv: message.sandboxEnv,
           modelVersion: message.modelVersion,
+          gitCommitInfo: message.gitCommitInfo, // Added this line
```
```suggestion
          gitCommitInfo: message.gitCommitInfo,
```

allenhutchison:
Can you make sure this get's plumbed through the /bug handler as well? edit; Nevermind I see you did that already 👍

dewitt:
> Can you make sure this get's plumbed through the /bug handler as well? edit; Nevermind I see you did that already 👍

Good catch. I broke that with my last refactor. Put it back in.

dewitt:
> I like both of these suggestions.

Implemented. This will also provide a template if we need additional source generation down the road.

.gitignore
```diff
@@ -31,3 +31,6 @@ bundle
 # Test report files
 junit.xml
 packages/*/coverage/
+
+# Generated files
+packages/cli/src/generated/
```
NTaylorMullen, line 36: Why output it in the cli/src/generated folder vs. something more at the root?

dewitt: > Why output it in the cli/src/generated folder vs. something more at the root?

Indeed, I originally parented the output under a top-level `generated` directory, but ran into no end to places in our build process and codebase that expects all source to live under `packages/`, and a many of those places would need to be substantially updated to take multiple source directories.

But the fact that you also caught this makes me pause. So before I submit this, let me give another pass at refactoring this out into a single top-level directory and report back.

Thanks for the review!

dewitt: > Before I submit this, let me give another pass at refactoring this out into a single top-level directory and report back.

I spent a while moving things around to arrive at a single top-level `generated/` directory, and unfortunately I still don't like the tradeoff. It's doable, but it's _significantly_ more invasive, changing a dozen additional files in non-intuitive ways.

To sanity check, I investigated what other people do with tools like `protoc` for GRPC that generate nodejs src, and every example I found generates it in-tree as well.

And finally, even Gemini itself recommended the in-tree approach over a root-level generated/ directory, independently reaching the same conclusion I did in my experiments.

CC @NTaylorMullen for a final 👍 or 👎 on using an in-tree `packages/cli/src/generated/` directory for now, with a pledge to revisit if grows out of control.

NTaylorMullen: > To sanity check, I investigated what other people do with tools like protoc for GRPC that generate nodejs src, and every example I found generates it in-tree as well.
> 
> And finally, even Gemini itself recommended the in-tree approach over a root-level generated/ directory, independently reaching the same conclusion I did in my experiments.

These are good enough reasons for me. Only question I have is, is `generated` a common folder name used in TypeScript land for this sort of content? I would have figured it'd exist in `dist` or `out` or something.

Assuming it aligns, consider me signed off 😄

dewitt: > Only question I have is, is generated a common folder name used in TypeScript land for this sort of content?

Looks like graphql uses in-tree (i.e., `src/.../generated/`, or alternately `src/.../gql` named for the type) folders or directly into the source tree (i.e. `src/...`) alongside other code. GRPC examples I've seen generate the code in-tree right next the corresponding `.proto` files.

I'm going to go with the in-tree approach I have here. I can back this change out easier than I could back out the extensive changes required to hack in a top-level `generated/` dir, or the changes needed to create a separate `packages/generated/` project just for this one file.

PR #569, opened 2025-05-27
refactor: use React strict mode

KeijiBranshi:


demo.md
```diff
@@ -0,0 +1,79 @@
+---
```
NTaylorMullen: For some reason GH wont show the preview of this file. Might be because of this?

demo.md
```diff
@@ -0,0 +1,79 @@
...
+
+## Gemini CLI Demo
+
+**Objective:** Showcase the basic functionality and interaction flow of the Gemini CLI.
+
+**1. Setup (Prerequisite):**
+
+   Before starting, ensure your Gemini API key is configured. The CLI looks for the `GEMINI_API_KEY` environment variable.
```
NTaylorMullen: We also support vertex keys if folks want. You could link to the Setup documentation on the repo

demo.md
```diff
@@ -0,0 +1,79 @@
...
+
+   The CLI has several built-in "slash" commands. To see a list of available commands and their descriptions:
+   ```
+   > /help
+   ```
+   This will display commands like `/clear` (to clear the screen), `/theme` (to change UI themes), `/memory` (to interact with persistent context from `GEMINI.md` files), `/about`, and `/bug`.
+
+**6. Executing Shell Commands (Use with Caution):**
+
+   You can execute shell commands directly from the CLI by prefixing them with an exclamation mark (`!`).
```
NTaylorMullen: This now swaps into shell mode so you only need to type `!` once

packages/cli/src/gemini.tsx
```diff
@@ -70,12 +70,14 @@ async function main() {
       process.env.CLI_VERSION || readUpResult?.packageJson.version || 'unknown';
 
     render(
-      React.createElement(App, {
-        config,
-        settings,
-        cliVersion,
-        startupWarnings,
-      }),
+      <React.StrictMode>
```
NTaylorMullen, line 73: oh wow, it's that easy? I assume we'll probably need to react to a bajillion errors 😄

NTaylorMullen (approved):

KeijiBranshi:
@NTaylorMullen sorry, that demo.md file was just a temporary commit to try and show the Kaggle team Gemini-CLI's capabilities. Removed from this PR

packages/cli/src/gemini.tsx
KeijiBranshi: fingers crossed. if people see issues though, we should feel empowered to just revert this

PR #574, opened 2025-05-27
fix: change entrypoint from 'gemini-code' to 'gemini' in published sandbox

KeijiBranshi:


packages/cli/Dockerfile.sandbox
```diff
@@ -19,9 +19,10 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
   procps \
   psmisc \
   lsof \
+  socat \
```
olcan, line 22: hmm we've had the socat calls for a while... how did that work before?

KeijiBranshi: It might _not_ have worked 😅.

this Dockerfile is what's published to our image registry, but it's not what we're typically using locally in development. So there's not too many eyes catching errors here. I'm working on removing this one entirely though and publishing the top level Dockerfile's image

olcan: i'm pretty sure i tested (actually used) ports w/ socat in a sandbox when I added that feature because I needed it (on my mac but shouldn't matter since this is inside container), so i would have thought it was already installed via one of the other packages

PR #582, opened 2025-05-28
Add a keybinding for ctrl+w to delete the previous word.

dewitt:
Adds the following new keybindings to the cli text input buffer:

- `Ctrl+W` : Delete previous word

### DIFF
--- a/packages/cli/src/ui/components/shared/text-buffer.ts
+++ b/packages/cli/src/ui/components/shared/text-buffer.ts
@@ -1176,6 +1176,7 @@ export function useTextBuffer({
       else if (key['ctrl'] && input === 'a') move('home');
       else if (key['end']) move('end');
       else if (key['ctrl'] && input === 'e') move('end');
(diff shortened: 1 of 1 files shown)


NTaylorMullen (approved):
Easy!

That being said, at one point we should add tests for these.

jacob314 (commented):
lgtm

jacob314 (commented):
There are existing tests for handleInput in text-buffer.test.ts
Would be nice if Gemini CLI defaulted to attempting to add tests for new functionality without having to be nudged.

dewitt:
> Would be nice if Gemini CLI defaulted to attempting to add tests for new functionality without having to be nudged.

Blame me. This one I did by hand, since I knew where it needed to go, and Gemini wanted to put the code in (a) the wrong file, (b) write new code it didn't even need, and (c) did it's usual superfluous comment thing. 

Maybe it was bad prompting on my part, but for a one-liner I was surprised by how complex it tried to make things. Yet it didn't even offer to write a test.

PR #583, opened 2025-05-28
Refactor: Add GeminiRespondingSpinner to make use of streamingState idiomatic

jacob314:
This Spinner should be used everywhere in the UI where we want to use a spinner to avoid flicker caused by showing spinners when we are not responding. Using context in this way makes the way we are using Context more idiomatic as this context is now basically the state to theme how responding spinners are rendered.

- If `streamingState` is `Responding`, the component renders a spinner.
- If `nonRespondingDisplay` is  set it days the fallback content.
- Otherwise (not `Responding` and no `nonRespondingDisplay`), it renders `null`.

Usage in `LoadingIndicator.tsx` and `ToolMessage.tsx` has been updated to pass the appropriate `nonRespondingDisplay` string to maintain previous visual behavior. Tests for `LoadingIndicator.test.tsx` and `ToolMessage.test.tsx` have been updated accordingly. Fix bug where the default display for the Spinner in ToolMessage wasn't aligned with the Spinner type.

### DIFF
--- a/packages/cli/src/ui/components/LoadingIndicator.tsx
+++ b/packages/cli/src/ui/components/LoadingIndicator.tsx
@@ -32,7 +32,6 @@ export const LoadingIndicator: React.FC<LoadingIndicatorProps> = ({
     <Box marginTop={1} paddingLeft={0}>
       <Box marginRight={1}>
         <GeminiRespondingSpinner
-          spinnerType="dots"
           nonRespondingDisplay={
             streamingState === StreamingState.WaitingForConfirmation ? '⠏' : ''
           }
--- a/packages/cli/src/ui/components/LoadingIndicator.tsx
+++ b/packages/cli/src/ui/components/LoadingIndicator.tsx
@@ -6,10 +6,10 @@
 
 import React from 'react';
 import { Box, Text } from 'ink';
-import Spinner from 'ink-spinner';
 import { Colors } from '../colors.js';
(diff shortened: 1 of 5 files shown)


packages/cli/src/ui/components/LoadingIndicator.tsx
```diff
@@ -30,11 +30,14 @@ export const LoadingIndicator: React.FC<LoadingIndicatorProps> = ({
...
   return (
     <Box marginTop={1} paddingLeft={0}>
-      {streamingState === StreamingState.Responding && (
-        <Box marginRight={1}>
-          <Spinner type="dots" />
-        </Box>
-      )}
+      <Box marginRight={1}>
+        <GeminiRespondingSpinner
+          spinnerType="dots"
```
KeijiBranshi: nitpick: "dots" is the default, so no need to include it here

KeijiBranshi (approved):
LGTM!

packages/cli/src/ui/components/LoadingIndicator.tsx
jacob314: done. in a followup I'll try adding the 
react/jsx-no-redundant-props
lint rule to make cathing spurious props automatic.  This prop was already spurious with the existing code but it wasn't obvious it was.

jacob314: never mind... Gemini hallucinated that lint rule.

PR #586, opened 2025-05-28
much improved support for background processes, avoiding termination (via SIGPIPE) or eventual blocking (e.g. due to filled OS buffers)

olcan:


packages/cli/src/ui/hooks/shellCommandProcessor.ts
```diff
@@ -120,13 +120,19 @@ export const useShellCommandProcessor = (
...
           const handleOutput = (data: string) => {
-            output += data;
-            setPendingHistoryItem({
-              type: 'info',
-              text: output,
-            });
+            // continue to consume post-exit for background processes
+            // removing listeners can overflow OS buffer and block subprocesses
+            // destroying (e.g. child.stdout.destroy()) can terminate subprocesses via SIGPIPE
+            if (!exited) {
```
NTaylorMullen, line 129: For a properly running background process only way to exit is via `esc` right? If so we may want to await the `exit` on the abort signal to prevent flashing the UI

olcan:
Actually escape only applies to the foreground shell process. There is no
way to exit background processes. So if you do 'npm start' you will see
streaming output and you can stop it with Escape. But if you do 'npm start
&' that process will get detached and would have to be killed via its
process id, which the model knows.

PR #587, opened 2025-05-28
feat: Enhance web_fetch tool to process prompts with URLs using Gemini API

allenhutchison:
This commit refactors the  tool to leverage the Gemini API for processing natural language prompts that include URLs. Instead of simply fetching the content of a single URL, the tool can now accept a prompt with instructions and multiple URLs (up to 20).

The Gemini API will be used to understand the instructions and process the content from the provided URLs. This allows for more complex operations like summarization, data extraction, and question answering based on web content.

The tool now expects a  parameter instead of  and includes logic to handle grounding metadata (sources and citations) from the Gemini API response, appending them to the processed content. Error handling has also been updated to reflect the new API interaction.

The  is updated to pass the necessary  object to the
 constructor.

### DIFF
--- a/packages/server/src/tools/web-fetch.ts
+++ b/packages/server/src/tools/web-fetch.ts
@@ -4,18 +4,42 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+import { GoogleGenAI, GroundingMetadata } from '@google/genai';
 import { SchemaValidator } from '../utils/schemaValidator.js';
 import { BaseTool, ToolResult } from './tools.js';
 import { getErrorMessage } from '../utils/errors.js';
+import { Config } from '../config/config.js';
+import { getResponseText } from '../utils/generateContentResponseUtilities.js';
+
+// Interfaces for grounding metadata (similar to web-search.ts)
+interface GroundingChunkWeb {
+  uri?: string;
+  title?: string;
+}
+
+interface GroundingChunkItem {
+  web?: GroundingChunkWeb;
+}
+
+interface GroundingSupportSegment {
+  startIndex: number;
+  endIndex: number;
+  text?: string;
+}
+
+interface GroundingSupportItem {
+  segment?: GroundingSupportSegment;
+  groundingChunkIndices?: number[];
+}
 
 /**
  * Parameters for the WebFetch tool
  */
 export interface WebFetchToolParams {
   /**
-   * The URL to fetch content from.
+   * The prompt containing URL(s) (up to 20) and instructions for processing their content.
    */
-  url: string;
+  prompt: string;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


packages/server/src/tools/web-fetch.ts
```diff
@@ -24,23 +48,32 @@ export interface WebFetchToolParams {
...
           },
         },
-        required: ['url'],
+        required: ['prompt'],
         type: 'object',
       },
     );
+
+    const apiKeyFromConfig = this.config.getApiKey();
+    this.ai = new GoogleGenAI({
```
NTaylorMullen, line 73: We should centralize our `GoogleGenAI` creation so it only happens once. That way once we start counting tokens etc. we only have one place to track

packages/server/src/tools/web-fetch.ts
```diff
@@ -53,21 +86,24 @@ export class WebFetchTool extends BaseTool<WebFetchToolParams, ToolResult> {
...
-    } catch {
-      return `Invalid URL format: "${params.url}". Please provide a valid absolute URL (e.g., 'https://example.com').`;
+    if (!params.prompt || params.prompt.trim() === '') {
+      return "The 'prompt' parameter cannot be empty and must contain URL(s) and instructions.";
+    }
+    if (
+      !params.prompt.includes('http://') &&
+      !params.prompt.includes('https://')
+    ) {
+      return "The 'prompt' must contain at least one valid URL (starting with http:// or https://).";
```
NTaylorMullen, line 96: How often did you see this in your testing?

packages/server/src/tools/web-fetch.ts
```diff
@@ -53,21 +86,24 @@ export class WebFetchTool extends BaseTool<WebFetchToolParams, ToolResult> {
...
     return null;
   }
 
   getDescription(params: WebFetchToolParams): string {
-    const displayUrl =
-      params.url.length > 80 ? params.url.substring(0, 77) + '...' : params.url;
-    return `Fetching content from ${displayUrl}`;
+    const displayPrompt =
+      params.prompt.length > 100
+        ? params.prompt.substring(0, 97) + '...'
```
NTaylorMullen, line 104: Hmmm, what if we extracted the URLs instead of surfacing the prompt?

packages/server/src/tools/web-fetch.ts
```diff
@@ -53,21 +86,24 @@ export class WebFetchTool extends BaseTool<WebFetchToolParams, ToolResult> {
...
+        : params.prompt;
+    return `Processing URLs and instructions from prompt: "${displayPrompt}"`;
```
NTaylorMullen, line 106: I'd also shorten this. The description doesn't get a lot of real-estate in most default terminal sizes

NTaylorMullen (approved):

packages/server/src/tools/web-fetch.ts
allenhutchison: Yeah, makes sense. I'll file an issue for that refactor. Could go along with the factory work I was mentioning in the RFC you read earlier.

packages/server/src/tools/web-fetch.ts
allenhutchison: I saw a few cases where the model dropped the URL. I originally had a URL parameter and the model would drop the URL on a regular basis. I was just combining the URL and the prompt at that stage, and decided to drop the required URL parameter and let the model create a prompt. This functionality allows the model to ask for specific information out of the page rather than asking for the raw page.

packages/server/src/tools/web-fetch.ts
allenhutchison: I think it's important for debugging that people are able to see that this is a prompt to a model and not just a url fetch anymore. How big can the displayPrompt be?

packages/server/src/tools/web-fetch.ts
allenhutchison: It's only 100 characters, are you saying it should be shorter than that?

PR #589, opened 2025-05-28
feat: add git branch name to footer

KeijiBranshi:
DEMO=http://screen/ne3zVUSJyrT5N7T

### DIFF
--- a/packages/cli/src/ui/hooks/useGitBranchName.ts
+++ b/packages/cli/src/ui/hooks/useGitBranchName.ts
@@ -1,0 +1,66 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { useState, useEffect, useCallback } from 'react';
+import { exec } from 'node:child_process';
+import fs from 'node:fs';
+import fsPromises from 'node:fs/promises';
+import path from 'path';
+
+export function useGitBranchName(cwd: string): string | undefined {
+  const [branchName, setBranchName] = useState<string | undefined>(undefined);
+
+  const fetchBranchName = useCallback(
+    () =>
+      exec(
+        'git rev-parse --abbrev-ref HEAD',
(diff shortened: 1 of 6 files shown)


packages/cli/src/ui/hooks/useGitBranchName.ts
```diff
@@ -0,0 +1,66 @@
...
+import fsPromises from 'node:fs/promises';
+import path from 'path';
+
+export function useGitBranchName(cwd: string): string | undefined {
+  const [branchName, setBranchName] = useState<string | undefined>(undefined);
+
+  const fetchBranchName = useCallback(
+    () =>
+      exec(
+        'git rev-parse --abbrev-ref HEAD',
```
NTaylorMullen, line 19: What if git isn't installed?

packages/cli/src/ui/hooks/useGitBranchName.ts
```diff
@@ -0,0 +1,66 @@
...
+          }
+        },
+      ),
+    [cwd, setBranchName],
+  );
+
+  useEffect(() => {
+    fetchBranchName(); // Initial fetch
+
+    const gitHeadPath = path.join(cwd, '.git', 'HEAD');
```
NTaylorMullen, line 40: You may want to look up the dir or see if `git` is installed to do the checking here.

packages/cli/src/ui/hooks/useGitBranchName.ts
KeijiBranshi: the error is caught in the callback passed to `exec`. then we just return `undefined` for the `branchName`

packages/cli/src/ui/hooks/useGitBranchName.ts
KeijiBranshi: yeah, gemini asked about this at the time of vibe-coding. I just told it to keep it simple and only check current dir for now. but i can add this in a followup

PR #595, opened 2025-05-29
fix name in package.json

sethtroisi:


sethtroisi:
broken because of remote branch.

PR #600, opened 2025-05-29
allow writing to user cache directory on macos (fixes use of lyria mcp server [to generate songs] under seatbelt)

olcan:


acoliver:
@coderabbit please re-review after changes

PR #603, opened 2025-05-29
allow comments in settings when parsed in scripts via jq

olcan:


scripts/sandbox_command.sh
```diff
@@ -35,7 +35,7 @@ shift $((OPTIND - 1))
 # note it can be string or boolean, and if missing jq will return null
 USER_SETTINGS_FILE="$HOME/.gemini/settings.json"
 if [ -z "${GEMINI_SANDBOX:-}" ] && [ -f "$USER_SETTINGS_FILE" ]; then
-    USER_SANDBOX_SETTING=$(jq -r '.sandbox' "$USER_SETTINGS_FILE")
+    USER_SANDBOX_SETTING=$(sed -e 's/\/\/.*//' -e 's/\/\*.*\*\///g' -e '/^[[:space:]]*\/\//d' "$USER_SETTINGS_FILE" | jq -r '.sandbox')
```
NTaylorMullen, line 38: Is there a way we can avoid requiring `jq`? Feels like it's a core pre-req to build the repo which is challenging because that's one more barrier to entry for consumers

olcan:
let me see

PR #606, opened 2025-05-30
feat: add --show_memory_usage flag to display memory usage in status bar

jacob314:
This commit introduces a new `--show_memory_usage` CLI flag. When this flag is set, the application will display the current memory usage (RSS) in the status bar, updating periodically.

The memory usage is displayed in a human-readable format (MB or GB). If usage exceeds 2GB, the text is shown in red to draw users attention.

This helps users monitor the application's memory footprint during runtime without using Chrome DevTools.

`npm start -- --show_memory_usage`

Screenshot 2025-05-29 at 6 19 04 PM

Example when memory usage is high
Screenshot 2025-05-29 at 6 18 09 PM

### DIFF
--- a/packages/cli/src/config/config.test.ts
+++ b/packages/cli/src/config/config.test.ts
@@ -7,16 +7,10 @@
 // packages/cli/src/config/config.test.ts
 
 import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
-// import * as fsPromises from 'fs/promises';
-// import * as fsSync from 'fs';
 import * as os from 'os';
-// import * as path from 'path'; // Unused, so removing
-// import { readPackageUp } from 'read-package-up';
-// import {
-//   loadHierarchicalGeminiMemory,
-// } from './config';
-// import { Settings } from './settings';
-// import * as ServerConfig from '@gemini-code/server';
+import { loadCliConfig } from './config.js';
+import { Settings } from './settings.js';
+import * as ServerConfig from '@gemini-code/server';
 
 const MOCK_HOME_DIR = '/mock/home/user';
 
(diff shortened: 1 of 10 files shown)


NTaylorMullen:
What do you think about integrating this with the `-d` (debug) flag? Also I could imagine wanting a `/diagnostics` or `/debug` slash command to understand details mid-run

jacob314:
> What do you think about integrating this with the `-d` (debug) flag? Also I could imagine wanting a `/diagnostics` or `/debug` slash command to understand details mid-run
Interesting idea. I've integrated it so it is on by default with the debug flag as well as adding it to the existing `/bug` command. If we want we can later add a /debug or /diagnostics command as well. The other low cost integration point would be the /about command but I didn't do that as I don't want to distract users with the memory usage unless there is a problem.

I did keep the separate config flag for now  as this is something I want to be able run with 100% confidence that the memory usage I am seeing is not due to other debug mode functionality. I'm paranoid about this because in Flutter, memory usage in debug mode ended up being much higher than regular mode over time as debug mode ended up retaining a lot of references to objects omitted in release mode.

jacob314:
Example bug report with the new functionality:
```
## Describe the bug
A clear and concise description of what the bug is.

## Additional context
Add any other context about the problem here.

## Diagnostic Information
*   **CLI Version:** development
*   **Git Commit:** 1b06d92 (local modifications)
*   **Operating System:** darwin v23.11.0
*   **Sandbox Environment:** sandbox-exec (minimal)
*   **Model Version:** gemini-2.5-pro-preview-05-06
*   **Memory Usage:** 159.6 MB
```

PR #608, opened 2025-05-30
feat(cli): add pro model availability check and fallback to flash

allenhutchison:
Implements a check on CLI startup to verify the availability of the default pro model (gemini-2.5-pro-preview-05-06). If a 429 (rate limit) error is detected immediately, the CLI will silently switch to using the flash model (gemini-2.5-flash-preview-05-20) for the current session.

This change aims to improve the experience for users who may be on free-tier API keys that do not have access to the pro model, by automatically falling back to a usable alternative.

The check is performed with a short timeout to minimize startup delay. User notification of the switch is handled in the main CLI entry point to avoid duplicate messages due to potential re-executions of the config loading logic (e.g., during sandboxing).

Includes unit tests for the new model checking utility.

Addresses #529

### DIFF
--- a/packages/cli/src/utils/modelCheck.ts
+++ b/packages/cli/src/utils/modelCheck.ts
@@ -38,7 +38,7 @@ export async function getEffectiveModel(
   const body = JSON.stringify({ contents: [{ parts: [{ text: 'test' }] }] });
 
   const controller = new AbortController();
-  const timeoutId = setTimeout(() => controller.abort(), 500); // 100ms timeout for the request
+  const timeoutId = setTimeout(() => controller.abort(), 500); // 500ms timeout for the request
 
   try {
     const response = await fetch(endpoint, {
--- a/packages/cli/src/utils/modelCheck.ts
+++ b/packages/cli/src/utils/modelCheck.ts
@@ -35,10 +35,18 @@ export async function getEffectiveModel(
   const modelToTest = DEFAULT_GEMINI_MODEL;
   const fallbackModel = DEFAULT_GEMINI_FLASH_MODEL;
   const endpoint = `https://generativelanguage.googleapis.com/v1beta/models/${modelToTest}:generateContent?key=${apiKey}`;
-  const body = JSON.stringify({ contents: [{ parts: [{ text: 'test' }] }] });
+  const body = JSON.stringify({
+    contents: [{ parts: [{ text: 'test' }] }],
+    generationConfig: {
+      maxOutputTokens: 1,
+      temperature: 0,
+      topK: 1,
+      thinkingConfig: { thinkingBudget: 0, includeThoughts: false },
+    },
+  });
 
   const controller = new AbortController();
-  const timeoutId = setTimeout(() => controller.abort(), 500); // 500ms timeout for the request
+  const timeoutId = setTimeout(() => controller.abort(), 2000); // 500ms timeout for the request
 
   try {
     const response = await fetch(endpoint, {
--- a/packages/cli/src/utils/modelCheck.ts
+++ b/packages/cli/src/utils/modelCheck.ts
@@ -1,0 +1,67 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import {
+  DEFAULT_GEMINI_MODEL,
+  DEFAULT_GEMINI_FLASH_MODEL,
+} from '../config/config.js';
+
+export interface EffectiveModelCheckResult {
+  effectiveModel: string;
+  switched: boolean;
+  originalModelIfSwitched?: string;
+}
+
+/**
+ * Checks if the default "pro" model is rate-limited and returns a fallback "flash"
+ * model if necessary. This function is designed to be silent.
+ * @param apiKey The API key to use for the check.
+ * @param currentConfiguredModel The model currently configured in settings.
+ * @returns An object indicating the model to use, whether a switch occurred,
+ *          and the original model if a switch happened.
+ */
+export async function getEffectiveModel(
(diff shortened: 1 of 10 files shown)


allenhutchison:
There are some reasons not to like this:

1) In the case where the API key has access to pro we will wait for either the entire timeout (500ms) or until we get a 200 back. This adds latency to startup for the default case. I can't make it any faster than 500ms or we start to lose protection from the test.

2) This is even more of a problem because we run the config code twice (once when the cli starts, and again inside the sandbox) so you actually have to wait twice for the check.

packages/cli/src/utils/modelCheck.ts
```diff
@@ -0,0 +1,67 @@
...
+): Promise<EffectiveModelCheckResult> {
+  if (currentConfiguredModel !== DEFAULT_GEMINI_MODEL) {
+    // Only check if the user is trying to use the specific pro model we want to fallback from.
+    return { effectiveModel: currentConfiguredModel, switched: false };
+  }
+
+  const modelToTest = DEFAULT_GEMINI_MODEL;
+  const fallbackModel = DEFAULT_GEMINI_FLASH_MODEL;
+  const endpoint = `https://generativelanguage.googleapis.com/v1beta/models/${modelToTest}:generateContent?key=${apiKey}`;
+  const body = JSON.stringify({ contents: [{ parts: [{ text: 'test' }] }] });
```
NTaylorMullen: We actually might want to specify hyper parameters here to indicate how many tokens (like 1), how much thinking (zero) etc. it should use so that this initial request comes back as fast as humanly possible.

packages/cli/src/utils/modelCheck.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  const controller = new AbortController();
+  const timeoutId = setTimeout(() => controller.abort(), 500); // 500ms timeout for the request
```
NTaylorMullen: Why hold up the UI on this? Instead can't we show `...` as the model until it's decided?

packages/cli/tsconfig.json
```diff
@@ -7,6 +7,6 @@
     "types": ["node", "vitest/globals"]
   },
   "include": ["index.ts", "src/**/*.ts", "src/**/*.tsx", "src/**/*.json"],
-  "exclude": ["node_modules", "dist"],
+  "exclude": ["node_modules", "dist", "src/**/*.test.ts", "src/**/*.spec.ts"],
```
NTaylorMullen: intentional?

packages/cli/src/utils/modelCheck.ts
```diff
@@ -0,0 +1,67 @@
...
+  currentConfiguredModel: string,
+): Promise<EffectiveModelCheckResult> {
+  if (currentConfiguredModel !== DEFAULT_GEMINI_MODEL) {
+    // Only check if the user is trying to use the specific pro model we want to fallback from.
+    return { effectiveModel: currentConfiguredModel, switched: false };
+  }
+
+  const modelToTest = DEFAULT_GEMINI_MODEL;
+  const fallbackModel = DEFAULT_GEMINI_FLASH_MODEL;
+  const endpoint = `https://generativelanguage.googleapis.com/v1beta/models/${modelToTest}:generateContent?key=${apiKey}`;
```
NTaylorMullen, line 37: Any reason to not use the official sdk client for this?

NTaylorMullen (commented):
Honestly not so bad, left some comments.

One other thing we may want to consider is to make this "fallback" a little more visible to the user and internally represent this as 'paid' vs. 'unpaid' and surface the fact that we can't use pro as a more explicit mention.

Also one thing that I noticed today is that technnnnically there's a preview/experimental version of pro that's still on the free tier. We could potentially bake a "free approach" into our core Client and fall back to the experimental pro -> flash as an alternate

packages/cli/src/utils/modelCheck.ts
allenhutchison: Excellent suggestion. Fixed.

packages/cli/src/utils/modelCheck.ts
allenhutchison: We don't, currently, have a good way to set the model to something different later in the startup sequence, and to communicate it back to the CLI.

packages/cli/tsconfig.json
allenhutchison: src/**/*.test.ts was intentional. The build would fail on the vi namespace without this after I added the unit test. The spec is unneeded and I've removed it.

packages/cli/src/utils/modelCheck.ts
allenhutchison: My thinking was to keep this as fast as possible and as low overhead as possible. So I thought a simple HTTP request would be faster than initializing the sdk at this stage of the startup sequence. Especially since all I want is the response code.

NTaylorMullen: Fair. One thing that this will miss (could be ok) is the agent header for gemini-cli

packages/cli/src/utils/modelCheck.ts
NTaylorMullen: Why only a 500ms response timeout in that case then?

allenhutchison: I changed it to 2 seconds in the update. With the changes to the hyperparameters the model usually comes back much faster than that. 

My initial thinking for the 500ms was that I only care about the 429 response case, I don't actually care about the 200 response case. The model fails for paid vs free much faster (i.e. 429) than it succeeds in generating tokens (200) so we could have a really short wait, if we don't get a 429 immediately then we are probably fine to continue.

NTaylorMullen (approved):

PR #610, opened 2025-05-30
Change ReadFile to not have a result display.

NTaylorMullen:
- It's verbose on its own; however, if file content is truncated we'll indicate that in the result display.

@jacob314 this is one "fix" I did tonight. I felt like the extra result display didn't add enough value on its own.


jacob314 (approved):
lgtm

PR #613, opened 2025-05-30
feat: Implement retry with backoff for API calls

NTaylorMullen:
- Added a  utility function to handle transient errors (429, 5xx) when making API calls.
- Integrated this utility into , , , and  to improve resilience against intermittent network issues and rate limiting.
- Includes unit tests for the retry logic.

Fixes https://github.com/google-gemini/gemini-cli/issues/528

### DIFF
--- a/packages/server/src/utils/retry.test.ts
+++ b/packages/server/src/utils/retry.test.ts
@@ -1,0 +1,238 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+import { retryWithBackoff } from './retry.js';
+
+// Define an interface for the error with a status property
+interface HttpError extends Error {
+  status?: number;
+}
+
+// Helper to create a mock function that fails a certain number of times
+const createFailingFunction = (
+  failures: number,
+  successValue: string = 'success',
+) => {
+  let attempts = 0;
+  return vi.fn(async () => {
+    attempts++;
+    if (attempts <= failures) {
+      // Simulate a retryable error
+      const error: HttpError = new Error(`Simulated error attempt ${attempts}`);
+      error.status = 500; // Simulate a server error
+      throw error;
+    }
+    return successValue;
+  });
+};
+
+// Custom error for testing non-retryable conditions
+class NonRetryableError extends Error {
+  constructor(message: string) {
+    super(message);
+    this.name = 'NonRetryableError';
+  }
+}
+
+describe('retryWithBackoff', () => {
+  beforeEach(() => {
+    vi.useFakeTimers();
+  });
+
+  afterEach(() => {
+    vi.restoreAllMocks();
+  });
+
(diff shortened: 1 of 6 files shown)


packages/server/src/utils/retry.test.ts
```diff
@@ -0,0 +1,238 @@
...
+  failures: number,
+  successValue: string = 'success',
+) => {
+  let attempts = 0;
+  return vi.fn(async () => {
+    attempts++;
+    if (attempts <= failures) {
+      // Simulate a retryable error
+      const error = new Error(`Simulated error attempt ${attempts}`) as any;
+      error.status = 500; // Simulate a server error
```
jacob314, line 27: optional nit: can you instead create the actual error type used by Gemini or the web fetch tool? That way if those types change this test will fail.
it would also avoid using any.

packages/server/src/utils/retry.test.ts
```diff
@@ -0,0 +1,238 @@
...
+    });
+    await expect(promise).rejects.toThrow('Bad Request');
+    expect(mockFn).toHaveBeenCalledTimes(1);
+  });
+
+  it('should respect maxDelayMs', async () => {
+    const mockFn = createFailingFunction(3);
+    const delays: number[] = [];
+    const originalSetTimeout = global.setTimeout;
+    global.setTimeout = ((cb: any, ms: number) => {
```
jacob314: use 
vi.useFakeTimers();
rather than mocking setTimeout ourselves.

packages/server/src/utils/retry.ts
```diff
@@ -0,0 +1,230 @@
...
+ * @param error The error object.
+ * @returns The HTTP status code, or undefined if not found.
+ */
+function getErrorStatus(error: unknown): number | undefined {
+  if (typeof error === 'object' && error !== null) {
+    if (
+      'status' in error &&
+      typeof (error as { status?: unknown }).status === 'number'
+    ) {
+      return (error as { status: number }).status;
```
jacob314: more nits about as checks.  these as checks shouldn't be needed as Typescript should infer the type from the previous line

packages/server/src/utils/retry.test.ts
```diff
@@ -0,0 +1,238 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
jacob314, line 7: optional nit: if possible remove displaying no-explicit-any in this test.

jacob314 (approved):
lgtm

packages/server/src/utils/retry.test.ts
NTaylorMullen: Done!

packages/server/src/utils/retry.ts
NTaylorMullen: Done!

packages/server/src/utils/retry.test.ts
NTaylorMullen: I feel like this is an ok middlegorund primarily because we don't want the model churning on as many "any"s. If we get to the point of writing tests ourselves (instead of relying on GC) then I could see the intention to change this.

FWIW I added this bit manually 😄

PR #614, opened 2025-05-30
update shell output at an interval to reduce flicker

olcan:


packages/cli/src/ui/hooks/shellCommandProcessor.ts
```diff
@@ -16,6 +16,8 @@ import path from 'path';
 import os from 'os';
 import fs from 'fs';
 
+const OUTPUT_UPDATE_INTERVAL_MS = 1000;
```
jacob314, line 19: nit: this constant and the associated logic is duplicated in two places. could be nice to move it to a shared helper along with adding a test.

jacob314 (commented):
lgtm

PR #616, opened 2025-05-30
allow aborting of shell mode (!) commands, similar to shell tool commands. fix bug that prevented aborts after first abort. more robust killing logic

olcan:


packages/cli/src/ui/hooks/shellCommandProcessor.ts
```diff
@@ -37,7 +37,7 @@ export const useShellCommandProcessor = (
    * @returns True if the query was handled as a shell command, false otherwise.
    */
   const handleShellCommand = useCallback(
-    (rawQuery: PartListUnion): boolean => {
+    (rawQuery: PartListUnion, signal?: AbortSignal): boolean => {
```
NTaylorMullen, line 40: lets make this required

packages/server/src/tools/shell.ts
```diff
@@ -210,16 +201,28 @@ export class ShellTool extends BaseTool<ShellToolParams, ToolResult> {
...
 
     const abortHandler = () => {
-      if (shell.pid) {
+      if (shell.pid && !exited) {
         try {
-          // Kill the entire process group
+          // attempt to SIGTERM process group (negative PID)
+          // if SIGTERM fails after 200ms, attempt SIGKILL
           process.kill(-shell.pid, 'SIGTERM');
+          setTimeout(() => {
```
NTaylorMullen, line 209: Should we be weaving this into a promise that's awaitable to ensure things don't linger?

olcan:
yes good point

olcan:
actually, the tool only returns once the process exits, so there can be
no lingering of process, and the tool returns as soon as the process exits
and no sooner, so it seems fine

PR #624, opened 2025-05-30
feat(sandbox): use package config to dictate sandbox image name

KeijiBranshi:


packages/cli/package.json
```diff
@@ -25,6 +25,9 @@
   "files": [
     "dist"
   ],
+  "config": {
+    "sandboxImageUri": "gemini-cli-sandbox"
```
NTaylorMullen, line 29: ❤️

PR #632, opened 2025-05-30
Fix bug pasting multiline strings

jacob314:
This fixes the bug pasting in long multiline strings.
Sadly my test case passes without the fix so isn't quite right even though the fix robustly handles the real app.

@KeijiBranshi do you have an idea why my test isn't fixing the issue? More nuanced timing of multiple useState objects that I haven't quite wrapped my head around.

### DIFF
--- a/packages/cli/src/ui/components/shared/text-buffer.ts
+++ b/packages/cli/src/ui/components/shared/text-buffer.ts
@@ -524,35 +524,33 @@ export function useTextBuffer({
       const normalised = str.replace(/\r\n/g, '\n').replace(/\r/g, '\n');
       const parts = normalised.split('\n');
 
-      setLines((prevLines) => {
-        const newLines = [...prevLines];
-        const lineContent = currentLine(cursorRow);
-        const before = cpSlice(lineContent, 0, cursorCol);
-        const after = cpSlice(lineContent, cursorCol);
+      const newLines = [...lines];
+      const lineContent = currentLine(cursorRow);
+      const before = cpSlice(lineContent, 0, cursorCol);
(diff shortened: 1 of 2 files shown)


packages/cli/src/ui/components/shared/text-buffer.ts
```diff
@@ -524,35 +524,33 @@ export function useTextBuffer({
...
+        // Adjusted condition for inserting multiple lines
+        const remainingParts = parts.slice(1);
+        const lastPartOriginal = remainingParts.pop() ?? '';
+        newLines.splice(cursorRow + 1, 0, ...remainingParts);
+        newLines.splice(
+          cursorRow + parts.length - 1,
+          0,
+          lastPartOriginal + after,
+        );
+        setCursorRow(cursorRow + parts.length - 1);
```
jacob314, line 544: the problem line causing the crash was setCursorRow which previously took in its previous value even though the rest of the logic was tightly coupled to cursorRow being the previous value. For consistency I also changed
setLines to not take the previous value and instead required this use callback to list lines as a dependency.

Now at least you have a bunch of updates to the state that all occur together and don't depend on previous values of the state so the modification of the state is at least more atomic.

KeijiBranshi (approved):
It's kind of buried in the docs, but `setState` updater functions should be pure: https://screenshot.googleplex.com/Bc9vMbr3yVmULrR

In other words, we shouldn't be calling `setState` inside other `setState`s. So this change is a solid improvement in the long term. I have a _hunch_ that may have contributed to the discrepancy between the test and app behavior.

PR #636, opened 2025-05-30
Dont match slash commands in history

sethtroisi:
pressing up to a slash commands triggers the autocomplete box which interferes with pressing up (until you backspace the entire command). This is majorly frustrating. This fixes that by just ignoring slash commands in history. Maybe we'll want to do the same with at commands later but that's not bitting me in the same way yet.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,1 @@
+packages/server/dist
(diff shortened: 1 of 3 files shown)


NTaylorMullen (approved):

sethtroisi:
Fixed another way in https://github.com/google-gemini/gemini-cli/pull/734

PR #637, opened 2025-05-30
Update default & ANSI theme names

miguelsolorio:
Updated the default theme names (`VS2015` => `Default`, `VS` => `Default Light`) and added a new `ANSI Light`. There was no clean way to add light themes for Default + ANSI so had to add "light" in the name twice (would be nice to not have that duplication for light themes).

Also added logic so that users of the old name are routed to the new names when going to the theme selector. 

90% of this was done by hand 😅

CleanShot 2025-05-30 at 16 20 12@2x

CleanShot 2025-05-30 at 16 22 54@2x


NTaylorMullen (approved):
> 90% of this was done by hand 😅

The worst! I'm sorry 😢 . That being said, I love what you did!

PR #639, opened 2025-05-31
Add default build task for VS Code

miguelsolorio:
Simple addition that allows devs to hit `Cmd+Shift+B` when building the repo. Super handy shortcut.


NTaylorMullen (approved):
oooo I like!

PR #640, opened 2025-05-31
feat: Implement delayed Ctrl+C exit prompt

taeold:
This change introduces a small delay after the first Ctrl+C press, prompting the user to press Ctrl+C again to exit. This helps prevent accidental termination of the application.

- Added `exitOnCtrlC={false}` to the Ink render options in `gemini.tsx` to enable custom Ctrl+C handling.
- Implemented logic in `App.tsx` to:
    - Display "Press Ctrl+C again to exit." for 1 seconds after the first Ctrl+C.
    - Exit the application if Ctrl+C is pressed again during this period.
    - Revert to normal operation if the second Ctrl+C is not pressed within the timeout.
- Defined a constant `CTRL_C_PROMPT_DURATION_MS` for the timeout duration.

Fixes https://github.com/google-gemini/gemini-cli/issues/621

image (3)

### DIFF
--- a/packages/cli/src/gemini.tsx
+++ b/packages/cli/src/gemini.tsx
@@ -78,6 +78,7 @@ async function main() {
           startupWarnings={startupWarnings}
         />
       </React.StrictMode>,
+      { exitOnCtrlC: false },
     );
     return;
   }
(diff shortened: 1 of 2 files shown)


packages/cli/src/gemini.tsx
```diff
@@ -78,6 +78,7 @@ async function main() {
           startupWarnings={startupWarnings}
         />
       </React.StrictMode>,
+      { exitOnCtrlC: false },
```
NTaylorMullen, line 81: Ahhh this was the bit I was missing, I was like how is it working.

NTaylorMullen (approved):
@taeold this is awesome, I love the route you took with this!

Issue #607, opened 2025-05-30
"gemini --version" returns unknown


zsais:
## Describe the bug
"gemini --version" returns unknown. When running "/about", we are able to see the CLI version.

```
$  gemini --version
unknown
```

```
> /about


╭───────────────────────────────────────────────────────────────────────────────────────────────────╮
│                                                                                                   │
│ About Gemini CLI                                                                                  │
│                                                                                                   │
│ CLI Version                       0.1.0-0d99398.0                                                 │
│ Git Commit                        0d99398                                                         │
│ Model                             gemini-2.5-pro-preview-05-06                                    │
│ Sandbox                           sandbox-exec (minimal)                                          │
│ OS                                darwin v22.16.0
```

## Diagnostic Information
*   **CLI Version:** 0.1.0-0d99398.0
*   **Git Commit:** 0d99398
*   **Operating System:** darwin v22.16.0
*   **Sandbox Environment:** sandbox-exec (minimal)
*   **Model Version:** gemini-2.5-pro-preview-05-06

NTaylorMullen:
Ah great callout, for us missing this!

scottdensmore:
This should be fixed now.

zsais:
I just tested this with version 0.1.0-34b81ab.0 and am still seeing the "unknown" issue.

<img width="899" alt="Image" src="https://github.com/user-attachments/assets/283a2068-9fed-44b2-bdac-16d7b3aded36" />

NTaylorMullen:
@scottdensmore here's a fun one for your follow up

zsais:
I just updated to `@head` and see the same issue.

![Image](https://github.com/user-attachments/assets/9f681a50-73af-445d-88ec-ea925e0a2af6)

cperry-goog:
closing this preemptively as the PR is in (can you merge it?)

PR #642, opened 2025-05-31
Fix/cli version unknown

scottdensmore:
Fix for https://github.com/google-gemini/gemini-cli/issues/607


packages/cli/src/config/config.ts
```diff
@@ -73,6 +73,7 @@ async function parseArguments(): Promise<CliArgs> {
       description: 'Show memory usage in status bar',
       default: false,
     })
+    .version() // This will enable the --version flag based on package.json
```
NTaylorMullen, line 76: oooo so easy, had no idea this was a thing!

PR #643, opened 2025-05-31
Fix: Update git branch watcher to use .git/logs/HEAD

scottdensmore:
The previous implementation watched .git/HEAD for changes, which may not always update reliably on all systems or git versions when the branch changes.

This change updates the watcher to monitor .git/logs/HEAD, which is more consistently updated when the HEAD reflog changes, providing a more robust way to detect branch switches.

Fixes https://github.com/google-gemini/gemini-cli/issues/634

### DIFF
--- a/packages/cli/src/ui/hooks/useGitBranchName.ts
+++ b/packages/cli/src/ui/hooks/useGitBranchName.ts
@@ -37,14 +37,17 @@ export function useGitBranchName(cwd: string): string | undefined {
   useEffect(() => {
     fetchBranchName(); // Initial fetch
 
-    const gitHeadPath = path.join(cwd, '.git', 'HEAD');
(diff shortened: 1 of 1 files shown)


packages/cli/src/ui/hooks/useGitBranchName.ts
```diff
@@ -37,14 +37,17 @@ export function useGitBranchName(cwd: string): string | undefined {
   useEffect(() => {
     fetchBranchName(); // Initial fetch
 
-    const gitHeadPath = path.join(cwd, '.git', 'HEAD');
+    const gitLogsHeadPath = path.join(cwd, '.git', 'logs', 'HEAD');
```
NTaylorMullen, line 40: Nice!

PR #648, opened 2025-05-31
Add batch editing capabilities to Edit Tool

keithballinger:
Support for multiple edits in a single operation with comprehensive error handling and improved mode selection. Includes best practices documentation in prompts.

- Add edits array parameter for batch operations
- Add mode parameter supporting 'edit', 'create', 'overwrite'
- Enhance error handling with detailed failed edit tracking
- Update return type to include edit metrics and failure details
- Maintain backward compatibility with single edit operations
- Update prompts with Edit Tool best practices section


keithballinger:
test failure appears unrelated

packages/core/src/tools/edit.ts
```diff
@@ -84,18 +110,49 @@ Expectation for parameters:
...
           },
           new_string: {
             description:
-              'The exact literal text to replace `old_string` with, preferably unescaped. Provide the EXACT text. Ensure the resulting code is correct and idiomatic.',
+              'The exact literal text to replace `old_string` with (for single edit - backward compatibility). Provide the EXACT text. Ensure the resulting code is correct and idiomatic.',
+            type: 'string',
+          },
+          mode: {
+            description:
+              'Edit mode: "edit" (modify existing file), "create" (create new file), or "overwrite" (replace entire file content). Default is "edit".',
```
NTaylorMullen: Technically we have a separate write-file tool that takes care of creating and overwriting content. In a world where we have this, we'd want to drop the write-file. Or alternatively we'd want to drop this.

packages/core/src/core/prompts.ts
```diff
@@ -50,13 +50,29 @@ You are an interactive CLI agent specializing in software engineering tasks. You
 - **Confirm Ambiguity/Expansion:** Do not take significant actions beyond the clear scope of the request without confirming with the user. If asked *how* to do something, explain first, don't just do it.
 - **Explaining Changes:** After completing a code modification or file operation *do not* provide summaries unless asked.
 
+# Edit Tool Best Practices
+
+## Batch Editing
```
NTaylorMullen, line 55: Batch edits are super interesting. My initial gut here is why not parallel tool calls; BUT that then means you have a whole bunch of parallel edits changing the same document over and over again vs. one group of edits resulting in one diff.

packages/core/src/core/prompts.ts
```diff
@@ -50,13 +50,29 @@ You are an interactive CLI agent specializing in software engineering tasks. You
...
+- **Group Related Changes:** Use the edit tool's batch editing capability (via 'edits' array) when making multiple related changes to the same file. This is more efficient than individual edit calls.
+- **Context for Each Edit:** Even in batch operations, each edit still requires significant context (3+ lines before/after) to ensure unique identification.
+- **Partial Success Handling:** Batch edits may partially succeed. Review the 'editsApplied', 'editsFailed', and 'failedEdits' in results to understand what worked and retry failed edits with better context if needed.
+
+## Edit Modes
```
NTaylorMullen: Lets remove edit modes given we have a dedicated write-file for now

packages/core/src/core/prompts.ts
```diff
@@ -50,13 +50,29 @@ You are an interactive CLI agent specializing in software engineering tasks. You
...
 2. **Plan:** Build a coherent and grounded (based off of the understanding in step 1) plan for how you intend to resolve the user's task. Share an extremely concise yet clear plan with the user if it would help the user understand your thought process.
-3. **Implement:** Use the available tools (e.g., '${EditTool.Name}', '${WriteFileTool.Name}' '${ShellTool.Name}' ...) to act on the plan, strictly adhering to the project's established conventions (detailed under 'Core Mandates').
+3. **Implement:** Use the available tools (e.g., '${EditTool.Name}', '${WriteFileTool.Name}' '${ShellTool.Name}' ...)  to act on the plan, strictly adhering to the project's established conventions (detailed under 'Core Mandates'). For file modifications, prefer batch edits when making multiple related changes to the same file. Use appropriate edit modes: 'create' for new files, 'edit' for modifications, 'overwrite' for complete replacements.
```
NTaylorMullen: With edit mode removal you'll want to update this too

packages/core/src/core/prompts.ts
```diff
@@ -75,7 +91,7 @@ When requested to perform tasks like fixing bugs, adding features, refactoring,
...
+4. **Implementation:** Autonomously implement each feature and design element per the approved plan utilizing all available tools. When starting ensure you scaffold the application using '${ShellTool.Name}' for commands like 'npm init', 'npx create-react-app'. Use 'create' mode when generating new files and batch editing for related changes within files. Aim for full scope completion. Proactively create or source necessary placeholder assets (e.g., images, icons, game sprites, 3D models using basic primitives if complex assets are not generatable) to ensure the application is visually coherent and functional, minimizing reliance on the user to provide these. If the model can generate simple assets (e.g., a uniformly colored square sprite, a simple 3D cube), it should do so. Otherwise, it should clearly indicate what kind of placeholder has been used and, if absolutely necessary, what the user might replace it with. Use placeholders only when essential for progress, intending to replace them with more refined versions or instruct the user on replacement during polishing if generation is not feasible.
```
NTaylorMullen, line 242: ditto

packages/core/src/core/prompts.ts
```diff
@@ -99,8 +115,10 @@ When requested to perform tasks like fixing bugs, adding features, refactoring,
...
 - **Background Processes:** Use background processes (via \`&\`) for commands that are unlikely to stop on their own, e.g. \`node server.js &\`. If unsure, ask the user.
 - **Interactive Commands:** Try to avoid shell commands that are likely to require user interaction (e.g. \`git rebase -i\`). Use non-interactive versions of commands (e.g. \`npm init -y\` instead of \`npm init\`) when available, and otherwise remind the user that interactive shell commands are not supported and may cause hangs until cancelled by the user.
+- **Enhanced Editing:** The '${EditTool.Name}' tool supports both single edits and batch operations. Use batch editing (\`edits\` array) for multiple related changes to the same file. Each edit in a batch still requires significant context for precise targeting. Choose appropriate modes: \`create\` for new files, \`edit\` for modifications, \`overwrite\` for complete file replacement. Handle partial success - review \`editsApplied\`, \`editsFailed\`, and \`failedEdits\` results and retry failed edits with improved context when needed.
```
NTaylorMullen:
```suggestion
- **Enhanced Editing:** The '${EditTool.Name}' tool supports both single edits and batch operations. Use batch editing ('edits' array) for multiple related changes to the same file. Each edit in a batch still requires significant context for precise targeting. Choose appropriate modes: 'create' for new files, 'edit' for modifications, 'overwrite' for complete file replacement. Handle partial success - review 'editsApplied', 'editsFailed', and 'failedEdits' results and retry failed edits with improved context when needed.
```

packages/core/src/core/prompts.ts
```diff
@@ -99,8 +115,10 @@ When requested to perform tasks like fixing bugs, adding features, refactoring,
...
 - **Respect User Confirmations:** Most tool calls (also denoted as 'function calls') will first require confirmation from the user, where they will either approve or cancel the function call. If a user cancels a function call, respect their choice and do _not_ try to make the function call again. It is okay to request the tool call again _only_ if the user requests that same tool call on a subsequent prompt. When a user cancels a function call, assume best intentions from the user and consider inquiring if they prefer any alternative paths forward.
+</edits>
```
NTaylorMullen:
```suggestion
```

packages/core/src/core/prompts.ts
```diff
@@ -184,12 +202,30 @@ Okay, I see src/auth.py currently uses 'urllib'. Before changing it, I need to c
 (After confirming dependency)
 Okay, 'requests' is available. Let me double check how it's used across the codebase:
 [tool_call: ${GrepTool.Name} 'requests\\.(get|post)']
-I will now refactor src/auth.py.
-[tool_call: Uses ${EditTool.Name} or ${WriteFileTool.Name} edit tools following conventions]
-(After editing)
+I will now refactor src/auth.py using batch edits for the multiple related changes needed.
```
NTaylorMullen: Should put this little bit in the tool call aspect of this example, otherwise the model will talk outloud to the user about the tools its using and their impl details

packages/core/src/core/prompts.ts
```diff
@@ -184,12 +202,30 @@ Okay, I see src/auth.py currently uses 'urllib'. Before changing it, I need to c
...
+[tool_call: Uses ${EditTool.Name} with 'edits' array for multiple related changes, or ${WriteFileTool.Name} if completely rewriting]
+(After editing, reviewing results for any failed edits)
 [tool_call: Runs project-specific lint/typecheck commands found previously, e.g., ${ShellTool.Name} for 'npm run build', 'ruff', 'check', 'src/auth.py']
 </example>
 
+<example>
```
NTaylorMullen, line 350: If we get rid of edit mode we'd want to get rid of this example as well

packages/core/src/tools/edit.ts
```diff
@@ -278,89 +403,103 @@ Expectation for parameters:
...
-          this.config.setAlwaysSkipModificationConfirmation(true);
-        }
-      },
-    };
-    return confirmationDetails;
+      const editsCount =
+        (params.edits?.length || 0) + (params.old_string !== undefined ? 1 : 0);
+      const title =
+        editsCount > 1
+          ? `Confirm ${editsCount} Edits: ${shortenPath(makeRelative(params.file_path, this.rootDirectory))}`
```
NTaylorMullen, line 385: I imagine the model may make multiple edits in a single line (or array entry). This might be misleading

packages/core/src/tools/edit.ts
```diff
@@ -278,89 +403,103 @@ Expectation for parameters:
...
     }
-    const oldStringSnippet =
-      params.old_string.split('\n')[0].substring(0, 30) +
-      (params.old_string.length > 30 ? '...' : '');
-    const newStringSnippet =
-      params.new_string.split('\n')[0].substring(0, 30) +
-      (params.new_string.length > 30 ? '...' : '');
-    return `${shortenPath(relativePath)}: ${oldStringSnippet} => ${newStringSnippet}`;
+
+    // Handle single edit (backward compatibility)
```
NTaylorMullen: If the intention is that the model would never utilize a single edit we could consider removing this path and its schema entirely

packages/core/src/tools/edit.ts
```diff
@@ -84,18 +110,49 @@ Expectation for parameters:
...
+                  description:
+                    'The exact literal text to replace, preferably unescaped. CRITICAL: Must uniquely identify the single instance to change. Include at least 3 lines of context BEFORE and AFTER the target text, matching whitespace and indentation precisely.',
+                  type: 'string',
+                },
+                new_string: {
+                  description:
+                    'The exact literal text to replace `old_string` with, preferably unescaped. Provide the EXACT text. Ensure the resulting code is correct and idiomatic.',
+                  type: 'string',
+                },
+                description: {
```
NTaylorMullen: Today we can only show one description per tool call so I think having multiple here may result in unnecessary tokens.

NTaylorMullen (changes requested):
On paper I really like the direction of this change. That being said I am concerned about the risk level of it.

There's a few key callouts:
- You've done a really cool thing of enabling an edit 'mode' that the model can utilize to effectively replace the "write-file" tool; however, a lot of the existing workflows were testing against the `replace` / `write-file` combo (for better or worse).
- I really like the idea of taking multiple edits that you can then conditionally apply (the ones that work) + if you do need to make several edits to a file doing that in parallel is a disaster waiting to happen. You've fixed that.
- The system prompt changes are probably the most risky to me given its ambiguous and hard to measure impact (at least currently) BUT is something we can anecdotally measure with enough time.

To get this in we should pick a direction for write-file or edit-mode (one or the other, if edit mode we need to change the tool name too) and be ready to potentially revert if needed.

packages/core/src/tools/edit.ts
keithballinger: Agreed

packages/core/src/core/prompts.ts
keithballinger: Correct. This allows the model to express in one turn the edits to a file it wants to make vs  multiple calls .

packages/core/src/core/prompts.ts
NTaylorMullen: Decided not to do this

packages/core/src/core/prompts.ts
```diff
@@ -160,75 +177,29 @@ user: 1 + 2
...
-user: is 13 a prime number?
-model: true
-</example>
-
 <example>
 user: list files here.
 model: [tool_call: ${LSTool.Name} for path '.']
 </example>
 
-<example>
```
NTaylorMullen, line 173: Bring this one back

packages/core/src/core/prompts.ts
```diff
@@ -160,75 +177,29 @@ user: 1 + 2
...
-user: start the server implemented in server.js
-model: [tool_call: ${ShellTool.Name} for 'node server.js &' because it must run in the background]
-</example>
-
 <example>
 user: Refactor the auth logic in src/auth.py to use the 'requests' library.
 model:
 [tool_call: ${ReadFileTool.Name} to read src/auth.py]
-Okay, I see src/auth.py currently uses 'urllib'. Before changing it, I need to check if 'requests' is already a project dependency.
```
NTaylorMullen: You'll need to bring more of this example back. Some of the verbiage is valuable because it lets the user know what's happening and why without acknowledging the impl details themselves

packages/core/src/core/prompts.ts
```diff
@@ -160,75 +177,29 @@ user: 1 + 2
...
-[tool_call: ${ReadManyFilesTool.Name} for paths ['**/*.test.ts', 'src/**/*.spec.ts'] assuming someFile.ts is in the src directory]
-(After reviewing existing tests and the file content)
-[tool_call: ${WriteFileTool.Name} to create /path/to/someFile.test.ts with the test code]
-I've written the tests. Now I'll run the project's test command to verify them.
-[tool_call: ${ShellTool.Name} for 'npm run test']
+[tool_call: ${EditTool.Name} with mode 'create' to create config/database.yml]
 </example>
 
 <example>
-user: How do I update the user's profile information in this system?
```
NTaylorMullen: Bring this and the next example back

packages/core/src/tools/edit.ts
```diff
@@ -84,18 +114,50 @@ Expectation for parameters:
...
+          edits: {
+            description:
+              'Array of edit operations to apply. Each edit should have old_string and new_string properties.',
+            type: 'array',
+            items: {
+              type: 'object',
+              properties: {
+                old_string: {
+                  description:
+                    'The exact literal text to replace, preferably unescaped. CRITICAL: Must uniquely identify the single instance to change. Include at least 3 lines of context BEFORE and AFTER the target text, matching whitespace and indentation precisely.',
```
NTaylorMullen, line 112: Bring back this line

packages/core/src/tools/edit.ts
```diff
@@ -84,18 +114,50 @@ Expectation for parameters:
...
+              'The exact literal text to replace (for single edit - backward compatibility). CRITICAL: Must uniquely identify the single instance to change. Include at least 3 lines of context BEFORE and AFTER the target text, matching whitespace and indentation precisely. If this string is not the exact literal text (i.e. you escaped it), matches multiple locations, or does not match exactly, the tool will fail.',
             type: 'string',
           },
           new_string: {
             description:
-              'The exact literal text to replace `old_string` with, preferably unescaped. Provide the EXACT text. Ensure the resulting code is correct and idiomatic.',
+              'The exact literal text to replace `old_string` with (for single edit - backward compatibility). Provide the EXACT text. Ensure the resulting code is correct and idiomatic.',
+            type: 'string',
+          },
+          content: {
```
NTaylorMullen: Maybe I missed this in the last pass. Isn't this effectively `old_string: ''` and `new_string: '... content...'`?

NTaylorMullen (changes requested):
I think we may have gone too far in the opposite direction with the removal of the examples. In the original changeset you had added a number of examples so our goal here should be to have a set of examples that accurately represents how our tools should be used while encouraging different trains of thought from the model (i.e. exploration)

Also there were some bits added to edit.ts for creating a file (content) that I'm not sure are necessary

packages/core/src/core/prompts.ts
```diff
@@ -185,7 +254,7 @@ Okay, I see src/auth.py currently uses 'urllib'. Before changing it, I need to c
 Okay, 'requests' is available. Let me double check how it's used across the codebase:
 [tool_call: ${GrepTool.Name} 'requests\\.(get|post)']
 I will now refactor src/auth.py.
-[tool_call: Uses ${EditTool.Name} or ${WriteFileTool.Name} edit tools following conventions]
+[tool_call: Uses ${EditTool.Name} edit tool following conventions]
```
NTaylorMullen:
```suggestion
[tool_call: Uses ${EditTool.Name}]
```

packages/core/src/core/prompts.ts
```diff
@@ -231,6 +300,17 @@ I found the following 'app.config' files:
 To help you check their settings, I can read their contents. Which one would you like to start with, or should I read all of them?
 </example>
 
+<example>
+user: Create a new config file for the database settings.
+model:
+[tool_call: ${EditTool.Name} with mode 'create' to create config/database.yml]
+</example>
+
+<example>
```
NTaylorMullen, line 397: This already exists above it looks like (see line 264)

packages/core/src/tools/edit.ts
```diff
@@ -84,18 +114,50 @@ Expectation for parameters:
...
+                new_string: {
+                  description:
+                    'The exact literal text to replace old_string with, preferably unescaped. Provide the EXACT text. Ensure the resulting code is correct and idiomatic.',
+                  type: 'string',
+                },
+              },
+              required: ['old_string', 'new_string'],
+            },
+          },
           old_string: {
```
NTaylorMullen: Why have the batch and single versions vs. edits with 1 element

NTaylorMullen (changes requested):
I went through the previous comments and tried to resolve ones that were covered! There were a few big ones left open though, hopefully that helps!

file_learnings.md
NTaylorMullen, line 1: Nice 😄 . Should drop though

packages/core/src/config/config.ts
```diff
@@ -16,7 +16,7 @@ import { GrepTool } from '../tools/grep.js';
 import { GlobTool } from '../tools/glob.js';
 import { EditTool } from '../tools/edit.js';
 import { ShellTool } from '../tools/shell.js';
-import { WriteFileTool } from '../tools/write-file.js';
+
```
NTaylorMullen, line 19:
```suggestion
```

packages/core/src/core/geminiChat.ts
```diff
@@ -54,6 +54,7 @@ function isValidContent(content: Content): boolean {
  * @throws Error if the history does not start with a user turn.
  * @throws Error if the history contains an invalid role.
  */
+// hello keith
```
NTaylorMullen, line 57: lolololololol, should drop these

packages/core/src/core/prompts.ts
```diff
@@ -50,19 +50,183 @@ You are an interactive CLI agent specializing in software engineering tasks. You
...
+- **Group Related Changes:** Use the edit tool's batch editing capability (via 'edits' array) when making multiple related changes to the same file. This is more efficient than individual edit calls.
+- **Context for Each Edit:** Even in batch operations, each edit still requires significant context (3+ lines before/after) to ensure unique identification.
+- **Partial Success Handling:** Batch edits may partially succeed. Review the 'editsApplied', 'editsFailed', and 'failedEdits' in results to understand what worked and retry failed edits with better context if needed.
+
+
+## Examples
+
+### Batch Editing for Multiple Related Changes
+When making several related changes to the same file, use the 'edits' array:
+{
```
NTaylorMullen, line 65: We should codify these examples and JSON.stringify them to ensure that they evolve with the code and don't fall out of date

packages/core/src/core/prompts.ts
```diff
@@ -50,19 +50,183 @@ You are an interactive CLI agent specializing in software engineering tasks. You
...
+  "file_path": "/absolute/path/to/app.js",
+  "edits": [{
+    "old_string": "  // Initialize app\n  const app = express();\n  app.use(middleware);\n  \n  // Start server\n  app.listen(3000);",
+    "new_string": "  // Initialize app\n  const app = express();\n  app.use(middleware);\n  app.use(newMiddleware);\n  \n  // Start server\n  app.listen(3000);"
+  }]
+}
+
+
+### API v2 Style Edit Examples
```
NTaylorMullen, line 91: What's API v2?

packages/core/src/core/prompts.ts
```diff
@@ -50,19 +50,183 @@ You are an interactive CLI agent specializing in software engineering tasks. You
...
+      "new_string": "function newFunction() {\n  // updated logic...\n}"
+    }
+  ]
+}
+
+## JSON Escaping for Edit Operations
+
+**CRITICAL:** When using the edit_file tool, strings in old_string and new_string must use standard JSON escaping - one level only. The content must match exactly what appears in the actual file.
+
+### Escaping Rules
```
NTaylorMullen, line 148: How effective have you found this?

packages/core/src/tools/edit.ts
```diff
@@ -50,18 +48,29 @@ export interface EditToolParams {
...
+  newString: string;
+  error: string;
 }
 
 /**
  * Implementation of the Edit tool logic
  */
-export class EditTool extends BaseTool<EditToolParams, ToolResult> {
+export class EditTool extends BaseTool<EditToolParams, EditResult> {
   static readonly Name = 'replace';
```
NTaylorMullen, line 74: Should this still be `replace`? You have `edit_file` listed elsewhere

packages/core/src/tools/edit.ts
```diff
@@ -74,8 +83,8 @@ export class EditTool extends BaseTool<EditToolParams, ToolResult> {
   constructor(config: Config) {
     super(
       EditTool.Name,
-      'Edit',
-      `Replaces text within a file. By default, replaces a single occurrence, but can replace multiple occurrences when \`expected_replacements\` is specified. This tool requires providing significant context around the change to ensure precise targeting. Always use the ${ReadFileTool} tool to examine the file's current content before attempting a text replacement.
+      'EditFile',
+      `Replaces text within a file. By default, replaces a single occurrence, but can replace multiple occurrences when \`expected_replacements\` is specified. This tool also supports batch editing with multiple edits in a single operation. Requires providing significant context around the change to ensure precise targeting. Always use the ${ReadFileTool.Name} tool to examine the file's current content before attempting a text replacement.
```
NTaylorMullen, line 87: Since you removed the new file edit mode we'll probably want to call out creating new files here. Also I know I mentioned it elsewhere but we should probably pick a certain type of quotations consistently

packages/core/src/tools/edit.ts
NTaylorMullen: This is still true

packages/core/src/utils/editCorrector.test.ts.backup
NTaylorMullen, line 1: Intended?

packages/core/src/utils/editCorrector.ts
```diff
@@ -61,10 +61,16 @@ export interface CorrectedEditResult {
  */
 export async function ensureCorrectEdit(
   currentContent: string,
-  originalParams: EditToolParams, // This is the EditToolParams from edit.ts, without \'corrected\'
+  originalParams: EditToolParams & { old_string: string; new_string: string },
   client: GeminiClient,
   abortSignal: AbortSignal,
 ): Promise<CorrectedEditResult> {
+  // Ensure we have required old_string and new_string
+  if (!originalParams.old_string || !originalParams.new_string) {
```
NTaylorMullen, line 69: Given you mention the param settings here as being non-nullable you could probably drop this

packages/core/src/utils/positionBasedEditProcessor.ts
NTaylorMullen, line 1: What's this bit?

NTaylorMullen (approved):
Definitely a good amount of comments to address but after you've done that feel free to merge away!

PR #651, opened 2025-05-31
Ignore folders files

keithballinger:
### Summary

This PR introduces .gitignore file filtering across the CLI, new `/sandbox` commands for managing .gitignore files, and configuration options for both features. It also adds corresponding tests and documentation.

### Key Changes

- **File Filtering:**
  - The CLI now respects `.gitignore` patterns when using `@` (at) commands, file completions, and several internal tools.
  - New configuration section (`fileFiltering`) allows customizing ignore behavior and additional patterns.
  - Updates to tests and implementation ensure git-ignored files are excluded from operations and suggestions by default.

- **Sandbox Cleanup Commands:**
  - Adds `/sandbox` slash command with:
    - `status`: Shows a summary of git-ignored files that could be cleaned.
    - `clean`: Removes git-ignored files with options for dry-run, aggressive mode, and preservation patterns.
    - `list-ignored`: Lists up to 20 git-ignored files that can be cleaned.
  - Always preserves critical files like `.env.example`, `README*`, and `LICENSE*` unless aggressive mode is enabled.
  - Customizable via new `sandboxCleanup` config section.

- **Configuration Updates:**
  - Adds new options to `settings.json`: `fileFiltering` and `sandboxCleanup`.
  - Updates documentation to describe new settings and command usage.

- **Tests:**
  - Adds integration and unit tests for file filtering and sandbox cleanup logic.
  - Ensures correct merging and fallback for settings.

- **Other Improvements:**
  - Underlying core tools (`ls`, `glob`, `read-many-files`) now support git-aware filtering.
  - Refactors how file discovery and filtering is handled internally.


packages/cli/src/config/settings.ts
```diff
@@ -28,6 +28,22 @@ export interface Settings {
...
+  
+  // Git-aware file filtering settings
+  fileFiltering?: {
+    respectGitIgnore?: boolean;
+    customIgnorePatterns?: string[];
+    allowBuildArtifacts?: boolean;
+  };
+  
+  // Sandbox cleanup settings
+  sandboxCleanup?: {
```
NTaylorMullen: /cc @olcan for the sandboxing aspects of the PR (I'll tag you on the sandboxing bits so you don't have to go digging)

packages/cli/src/ui/hooks/atCommandProcessor.ts
```diff
@@ -134,9 +135,19 @@ export async function handleAtCommand({
 
   addItem({ type: 'user', text: query }, userMessageTimestamp);
 
+  // Initialize git-aware file discovery
+  const fileDiscovery = new FileDiscoveryService(config.getTargetDir());
+  const respectGitIgnore = config.getFileFilteringRespectGitIgnore();
+  const customIgnorePatterns = config.getFileFilteringCustomIgnorePatterns();
+  await fileDiscovery.initialize({
```
NTaylorMullen: Lets move this to the `gemini.tsx` so that we initialize once async at startup and then we can have a `config.getFileService()` here.

packages/cli/src/ui/hooks/slashCommandProcessor.ts
NTaylorMullen, line 1: @olcan

packages/cli/src/ui/hooks/useCompletion.ts
```diff
@@ -236,24 +247,55 @@ export function useCompletion(
...
       let fetchedSuggestions: Suggestion[] = [];
+      
+      // Initialize git-aware file discovery if config is available
+      let fileDiscovery: FileDiscoveryService | null = null;
+      if (config) {
+        try {
+          fileDiscovery = new FileDiscoveryService(cwd);
+          const respectGitIgnore = config.getFileFilteringRespectGitIgnore?.() ?? true;
+          const customIgnorePatterns = config.getFileFilteringCustomIgnorePatterns?.() ?? [];
+          await fileDiscovery.initialize({
```
NTaylorMullen: Aligned with my other comment, I think having a centralized version of this will clean this bit up a ton! 😄

packages/cli/src/ui/hooks/useCompletion.ts
```diff
@@ -236,24 +247,55 @@ export function useCompletion(
...
-              return {
-                label,
-                value: escapePath(label), // Value for completion should be just the name part
-              };
-            });
+          
+          // Filter entries using git-aware filtering
+          const filteredEntries = [];
+          for (const entry of entries) {
+            if (!entry.name.startsWith(prefix)) continue;
```
NTaylorMullen: Definitely double check me but I think entry here an absolute path which may not do what you think here

packages/core/src/services/fileDiscoveryService.ts
```diff
@@ -0,0 +1,82 @@
...
+
+export class FileDiscoveryService {
+  private gitIgnoreFilter: GitIgnoreFilter | null = null;
+  private projectRoot: string;
+
+  constructor(projectRoot: string) {
+    this.projectRoot = path.resolve(projectRoot);
+  }
+
+  async initialize(options: FileDiscoveryOptions = {}): Promise<void> {
```
NTaylorMullen, line 25: This might need to be "git aware" i.e. what if the repo isn't a git repo

packages/core/src/services/fileDiscoveryService.ts
```diff
@@ -0,0 +1,82 @@
...
+      // Always respect git ignore unless explicitly disabled
+      if (options.respectGitIgnore !== false && this.gitIgnoreFilter) {
+        if (this.gitIgnoreFilter.isIgnored(filePath)) {
+          return false;
+        }
+      }
+
+      // Apply custom ignore patterns
+      if (options.customIgnorePatterns) {
+        const relativePath = path.relative(this.projectRoot, path.resolve(this.projectRoot, filePath));
```
NTaylorMullen: Double check me but I think you can do:

```suggestion
        const relativePath = path.relative(this.projectRoot, filePath);
```

packages/core/src/services/sandboxCleanupService.ts
NTaylorMullen, line 1: @olcan

packages/core/src/tools/glob.ts
```diff
@@ -40,7 +46,7 @@ export class GlobTool extends BaseTool<GlobToolParams, ToolResult> {
    * Creates a new instance of the GlobLogic
    * @param rootDirectory Root directory to ground this tool in.
    */
-  constructor(private rootDirectory: string) {
+  constructor(private rootDirectory: string, private config?: any) {
```
NTaylorMullen: Lets make the config required + you'll also want to avoid `any` (will break linting)

```suggestion
  constructor(private rootDirectory: string, private config: Config) {
```

packages/core/src/tools/glob.ts
```diff
@@ -62,6 +68,11 @@ export class GlobTool extends BaseTool<GlobToolParams, ToolResult> {
               'Optional: Whether the search should be case-sensitive. Defaults to false.',
             type: 'boolean',
           },
+          respect_git_ignore: {
```
NTaylorMullen, line 74: Lets make this schema addition conditional based on if the repo is a git repo

packages/core/src/tools/ls.ts
```diff
@@ -65,7 +71,7 @@ export class LSTool extends BaseTool<LSToolParams, ToolResult> {
    * Creates a new instance of the LSLogic
    * @param rootDirectory Root directory to ground this tool in. All operations will be restricted to this directory.
    */
-  constructor(private rootDirectory: string) {
+  constructor(private rootDirectory: string, private config?: any) {
```
NTaylorMullen: Ditto comments about requiring this + making the git ignore bits conditional on the repo being a git repo

packages/core/src/tools/read-many-files.ts
```diff
@@ -119,7 +125,10 @@ export class ReadManyFilesTool extends BaseTool<
    * @param targetDir The absolute root directory within which this tool is allowed to operate.
    * All paths provided in `params` will be resolved relative to this directory.
    */
-  constructor(readonly targetDir: string) {
+  constructor(
+    readonly targetDir: string,
+    private config?: any,
```
NTaylorMullen: ditto here BUT lets also make the schema `respectGitIgnore` naming consistent with the other tools `respect_git_ignore`

packages/core/src/utils/gitIgnoreParser.ts
```diff
@@ -0,0 +1,110 @@
...
+export class GitIgnoreParser implements GitIgnoreFilter {
+  private ignorePatterns: string[] = [];
+  private projectRoot: string;
+
+  constructor(projectRoot: string) {
+    this.projectRoot = path.resolve(projectRoot);
+  }
+
+  async initialize(): Promise<void> {
+    await this.loadGitIgnorePatterns();
```
NTaylorMullen: Should just inline the load git ignore patterns here

NTaylorMullen (approved):
Super solid, consider me signed off after changes. Should wait for @olcan's sign-off too.

An epic PR!

NTaylorMullen:
> Thanks for your pull request! It looks like this may be your first contribution to a Google open source project. Before we can look at your pull request, you'll need to sign a Contributor License Agreement (CLA).
> 
> View this [failed invocation](https://github.com/google-gemini/gemini-cli/pull/651/checks?check_run_id=43232354152) of the CLA check for more information.
> 
> For the most up to date status, view the checks section at the bottom of the pull request.

@teeler do you happen to know why Keith might be running into this? He said his https://cla.developers.google.com/clas is all good

packages/cli/src/ui/hooks/atCommandProcessor.ts
keithballinger: Done!

packages/cli/src/ui/hooks/useCompletion.ts
keithballinger: Done!

packages/cli/src/ui/hooks/useCompletion.ts
keithballinger: Good call

packages/core/src/services/fileDiscoveryService.ts
keithballinger: added logic to detect if we are in a repository

packages/core/src/services/fileDiscoveryService.ts
keithballinger: yep

olcan:
> Super solid, consider me signed off after changes. Should wait for @olcan's sign-off too.
> 
> An epic PR!

Yeah this change touches on a much-needed and critical idea of centralizing the "sandbox": which files/folders are meant to be available to CLI and tools for reading or read-writing. Currently this has been implicit and scattered in various sandbox configuration options for system-level sandboxing, and on top of that we have had some tools (like read file tool) doing its own "sandboxing" (i.e. rejecting access) inconsistently with the system-level sandboxing, causing confusion and weakening safety. 

The central "file discovery service" introduced here could be the perfect place to solve the above issue, as well as introduce new features such as those in this PR. However this PR as it is now seems to be sandbox-unaware (w.r.t. system-level sandboxing) and also seems to assume that the repo is git-managed. So for example `/sandbox status` talks about summarizing git-ignored files when I would expect that to summarize system-level (or "unified") sandboxing status.

One way to make progress would be to not use the word "sandbox" (maybe /ignore or /gitignore if assuming git) for now to avoid any confusion, and then rename and unify (with system-level sandboxing, solving the above issue) in a subsequent change, which i'd be happy to also advise on. I would still make sure that this change makes sense for non-git repos, and also in general i'm assuming all the features added here have use cases already, which I can defer to @NTaylorMullen on.

packages/core/src/services/fileDiscoveryService.ts
```diff
@@ -74,9 +78,16 @@ export class FileDiscoveryService {
...
+    if (options.respectGitIgnore !== false && this.isGitRepo && this.gitIgnoreFilter) {
       return this.gitIgnoreFilter.isIgnored(filePath);
     }
     return false;
   }
+
+  /**
+   * Returns whether the project is a git repository
+   */
+  isGitRepository(): boolean {
```
NTaylorMullen: This should probably be on config itself.

keithballinger:
Thanks for takning a look!

> also seems to assume that the repo is git-managed. 
This was fixed by introducing  gitUtils.js which includes a isGitRepository function.

>  However this PR as it is now seems to be sandbox-unaware (w.r.t. system-level sandboxing) and So for example `/sandbox status` talks about summarizing git-ignored files when I would expect that to summarize system-level (or "unified") sandboxing status.

yep, removing those pieces - they just muddy the PR with extra stuff.

packages/cli/src/ui/hooks/useCompletion.ts
keithballinger: done

teeler:
> > Thanks for your pull request! It looks like this may be your first contribution to a Google open source project. Before we can look at your pull request, you'll need to sign a Contributor License Agreement (CLA).
> > View this [failed invocation](https://github.com/google-gemini/gemini-cli/pull/651/checks?check_run_id=43232354152) of the CLA check for more information.
> > For the most up to date status, view the checks section at the bottom of the pull request.
> 
> @teeler do you happen to know why Keith might be running into this? He said his https://cla.developers.google.com/clas is all good

Yea, hey @keithballinger - you seem to have signed the commit twice? 

```
✅ 29041be6d2b44750e493c2653f5678170827dff7 PR Opener: @keithballinger <ke****a​@gmail.com>
❌ 29041be6d2b44750e493c2653f5678170827dff7 Author: <kei*******ger​@google.com>
```

And with two different email addresses.

your @gmail address has signed the CLA, but your google.com address hasn't BUT more importantly as you are a Google employee you don't need to sign the CLA. ;) 

My suggestion is to fix your commit and either keep one or the other but not both. You should also register your github username internally at go/github. 

Ping me if you need help!

packages/cli/src/ui/hooks/useCompletion.ts
```diff
@@ -37,6 +37,7 @@ export function useCompletion(
   cwd: string,
   isActive: boolean,
   slashCommands: SlashCommand[],
+  config?: any,
```
NTaylorMullen: Should be concrete config type

```suggestion
  config: Config,
```

packages/cli/src/ui/hooks/useCompletion.ts
```diff
@@ -236,24 +252,59 @@ export function useCompletion(
     const fetchSuggestions = async () => {
       setIsLoadingSuggestions(true);
       let fetchedSuggestions: Suggestion[] = [];
+
+      // Get centralized file discovery service if config is available
+      let fileDiscovery = null;
+      if (config) {
+        try {
```
NTaylorMullen: Try catch around this is probably overkill. Can drop. Also the if (config) could be dropped too

packages/core/src/services/fileDiscoveryService.ts
```diff
@@ -0,0 +1,105 @@
...
+  private gitIgnoreFilter: GitIgnoreFilter | null = null;
+  private projectRoot: string;
+  private isGitRepo: boolean = false;
+
+  constructor(projectRoot: string) {
+    this.projectRoot = path.resolve(projectRoot);
+  }
+
+  async initialize(options: FileDiscoveryOptions = {}): Promise<void> {
+    this.isGitRepo = isGitRepository(this.projectRoot);
```
NTaylorMullen: Lets bake `isGitRepository` into the config object and pass if we're in a git repo in the constructor

packages/core/src/tools/glob.ts
NTaylorMullen: Just pinging this bit

packages/core/src/tools/ls.ts
```diff
@@ -84,6 +94,11 @@ export class LSTool extends BaseTool<LSToolParams, ToolResult> {
             },
             type: 'array',
           },
+          respect_git_ignore: {
```
NTaylorMullen, line 96: ditto

packages/core/src/tools/ls.ts
```diff
@@ -229,6 +253,18 @@ export class LSTool extends BaseTool<LSToolParams, ToolResult> {
         }
 
         const fullPath = path.join(params.path, file);
+        const relativePath = path.relative(this.rootDirectory, fullPath);
+
+        // Check if this file should be git-ignored (only in git repositories)
+        if (
+          respectGitIgnore &&
+          fileDiscovery.isGitRepository() &&
+          fileDiscovery.shouldIgnoreFile(relativePath)
```
NTaylorMullen, line 261: This flow will potentially ignore non-git ignore custom file paths if the repo is not a git repo. Feel free to remove the non-git ignore feature if you'd like (in a way it kind of overlaps with .aiexclude fwiw). I imagine it'd require a bit more surgery otherwise.

Seems like this pattern / concern applies to a few other places in this PR as well as a heads up

packages/core/src/tools/read-many-files.ts
```diff
@@ -155,6 +165,12 @@ export class ReadManyFilesTool extends BaseTool<
             'Optional. Whether to apply a list of default exclusion patterns (e.g., node_modules, .git, binary files). Defaults to true.',
           default: true,
         },
+        respect_git_ignore: {
```
NTaylorMullen, line 167: ditto

packages/core/src/utils/gitIgnoreParser.ts
```diff
@@ -0,0 +1,120 @@
...
+  private projectRoot: string;
+  private isGitRepo: boolean = false;
+
+  constructor(projectRoot: string) {
+    this.projectRoot = path.resolve(projectRoot);
+  }
+
+  async initialize(): Promise<void> {
+    this.isGitRepo = isGitRepository(this.projectRoot);
+    if (this.isGitRepo) {
```
NTaylorMullen, line 28: If initialize gets called here we should assume it's a git repo imo. i.e. the caller should do the repo check

packages/core/src/utils/gitUtils.ts
```diff
@@ -0,0 +1,73 @@
...
+
+import * as fs from 'fs';
+import * as path from 'path';
+
+/**
+ * Checks if a directory is within a git repository
+ * @param directory The directory to check
+ * @returns true if the directory is in a git repository, false otherwise
+ */
+export function isGitRepository(directory: string): boolean {
```
NTaylorMullen, line 15: See other comments about moving this bit into the config

packages/core/src/utils/gitUtils.ts
```diff
@@ -0,0 +1,73 @@
...
+    return false;
+  }
+}
+
+/**
+ * Finds the root directory of a git repository
+ * @param directory Starting directory to search from
+ * @returns The git repository root path, or null if not in a git repository
+ */
+export function findGitRoot(directory: string): string | null {
```
NTaylorMullen, line 49: Do we need to export this?

NTaylorMullen (approved):

packages/core/src/tools/glob.ts
keithballinger: this is done

packages/core/src/tools/ls.ts
keithballinger: done

packages/core/src/utils/gitIgnoreParser.ts
keithballinger: done

packages/core/src/services/fileDiscoveryService.ts
keithballinger: done

packages/cli/src/ui/hooks/useCompletion.ts
keithballinger: done

packages/core/src/services/fileDiscoveryService.ts
keithballinger: done-ish

packages/core/src/tools/ls.ts
keithballinger: done

scidomino (approved):

Issue #406, opened 2025-05-17
Use a more general `.md` file to guide agents


steren:
### Description of the feature request:

We [support `GEMINI.md`](https://github.com/google-gemini/gemini-cli/blob/main/docs/cli/configuration.md#4-geminimd-files-hierarchical-instructional-context), but the guidance in this file can be used by any agent, not only Gemini. 

So I suggest using a more generic `AGENTS.md` filename instead.

Potentially, both could be supported, and we should then decide if it's an override or a complement.

If an override, the order of selection could be:

1. `GEMINI.md`
2.  `AGENTS.md`

I.e. `GEMINI.md` could take precedence if present

If it's a complement, then `GEMINI.md` can just be appended to `AGENTS.md` when using Gemini CLI

steren:
CC @ptone

NTaylorMullen:
I do like the override aspect you detail here

olcan:
i'd probably start with override-only, because complement brings up questions around ordering, and another more general way to achieve that is to introduce some basic include/import syntax so one could include AGENTS.md in GEMINI.md wherever they like, or have both include a common base, etc

allenhutchison:
We're going to introduce a setting to allow the user to name the file they want to use. We'll still use GEMINI.md by default, but if the user specifies AGENTS.md in their settings we'll use that instead. 

I'm still not convinced that all models need the same instructions. We have some cases where we've added instructions in our own project files that are really unique to the way Gemini works. I don't know what quirks people have tried to work around in their other config files, nor do I know how those workarounds will be interpreted by Gemini.

Because of this, and because we are entering a market where there are other tools out there already I assume that we will be operating in environments that already have 1 or 2 instruction files available. I'd like our users to get the best impression of the tool and Gemini on our first run experience and I think slurping in a bunch of instructions for other models can detract from that experience. 

In addition to this, we auto-write to the GEMINI.md file when the model wants to remember something specific about you. I'm less comfortable auto-writing to another file that is used by other models.

So rather than auto-detect and merge, or strictly prioritize we are going to use GEMINI.md by default. If the user wishes to use a different file they can set that in .gemini/settings.json and we'll use that filename instead. That new name will be used throughout the entire scope of the settings.json.

steren:
`AGENTS.md` is seeing some [standardization / community alignment](https://agents.md/).

I suggest we re-open this issue.

Gemini CLI should look *by default* for any of `AGENTS.md` or `GEMINI.md`

[...]

PR #654, opened 2025-05-31
feat: allow custom filename for context files

allenhutchison:
This change introduces a new setting `contextFileName` that allows users to specify a custom filename for their hierarchical context files, which defaults to `GEMINI.md`.

This addresses issue #406 by providing flexibility for projects or users who prefer a different naming convention for their AI context.

The changes include:
- Updating CLI and server configurations to read and use the `contextFileName` setting.
- Modifying the memory discovery logic to search for the configured filename.
- Updating documentation to reflect the new setting and refer to "context files" more generally.
- Adjusting UI elements to display the correct filename.
- Adding new tests and updating existing ones to cover the new functionality.


packages/cli/src/config/config.ts
```diff
@@ -132,6 +133,17 @@ export async function loadCliConfig(settings: Settings): Promise<Config> {
   const argv = await parseArguments();
   const debugMode = argv.debug || false;
 
+  // Set the context filename in the server's memoryTool module BEFORE loading memory
+  // TODO(b/343434939): This is a bit of a hack. The contextFileName should ideally be passed
```
NTaylorMullen, line 138: Agree, thanks for the callout

packages/cli/src/config/config.ts
```diff
@@ -132,6 +133,17 @@ export async function loadCliConfig(settings: Settings): Promise<Config> {
...
+  // directly to the Config constructor in core, and have core handle setGeminiMdFilename.
+  // However, loadHierarchicalGeminiMemory is called *before* createServerConfig.
+  if (settings.contextFileName) {
+    setServerGeminiMdFilename(settings.contextFileName);
+  } else {
+    // Reset to default if not provided in settings.
+    setServerGeminiMdFilename('GEMINI.md');
```
NTaylorMullen: Should use our constant value

packages/cli/src/config/settings.ts
```diff
@@ -30,6 +30,7 @@ export interface Settings {
   mcpServerCommand?: string;
   mcpServers?: Record<string, MCPServerConfig>;
   showMemoryUsage?: boolean;
+  contextFileName?: string; // Keep this for now for setServerGeminiMdFilename
```
NTaylorMullen:
```suggestion
  contextFileName?: string;
```

packages/cli/src/ui/App.tsx
```diff
@@ -398,7 +398,8 @@ export const App = ({
                     </Text>
                   ) : geminiMdFileCount > 0 ? (
                     <Text color={Colors.SubtleComment}>
-                      Using {geminiMdFileCount} GEMINI.md file
+                      Using {geminiMdFileCount}{' '}
+                      {settings.merged.contextFileName || 'GEMINI.md'} file
```
NTaylorMullen: ditto

packages/core/src/config/config.ts
```diff
@@ -56,6 +56,7 @@ export interface ConfigParameters {
   alwaysSkipModificationConfirmation?: boolean;
   vertexai?: boolean;
   showMemoryUsage?: boolean;
+  contextFileName?: string; // This is used by the constructor to call setGeminiMdFilename
```
NTaylorMullen:
```suggestion
  contextFileName?: string;
```

packages/core/src/tools/memoryTool.ts
```diff
@@ -46,9 +46,19 @@ Do NOT use this tool:
 `;
 
 export const GEMINI_CONFIG_DIR = '.gemini';
-export const GEMINI_MD_FILENAME = 'GEMINI.md';
+export let GEMINI_MD_FILENAME = 'GEMINI.md';
```
NTaylorMullen: Hmmm, what if we called this:
```suggestion
export const DEFAULT_CONTEXT_FILENAME = 'GEMINI.md';
```

And then had all call sites that were looking to have the "set" version call `getCurrent..`

NTaylorMullen (approved):

packages/cli/src/config/config.ts
allenhutchison: done

packages/cli/src/ui/App.tsx
allenhutchison: done

packages/core/src/tools/memoryTool.ts
allenhutchison: Done.

PR #656, opened 2025-05-31
cleanup: removed duplicate check from Config.registerCoreTool()

bdmorgan:


NTaylorMullen (approved):
Hah! Nice catch

packages/core/src/config/config.ts
```diff
@@ -238,7 +238,7 @@ export function createToolRegistry(config: Config): ToolRegistry {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   const registerCoreTool = (ToolClass: any, ...args: unknown[]) => {
     // check both the tool name (.Name) and the class name (.name)
-    if (!tools || tools.has(ToolClass.Name) || tools.has(ToolClass.name)) {
+    if (!tools || tools.has(ToolClass.Name)) {
```
olcan, line 241: unfortunately this is not duplicate or redundant (but yeah confusing sadly)

NTaylorMullen: Oooooo the casing!

NTaylorMullen (commented):
@bdmorgan actually looks like there's a casing difference

packages/core/src/config/config.ts
bdmorgan: oh no sorry about that!

PR #658, opened 2025-05-31
feat: Refactor CLI header for customizable logo text

NTaylorMullen:
- Replaces the hardcoded ASCII art logo with the  component. This change makes it easier to customize the displayed text in the header.
- Adds  `ink-big-text` dependency.
- This change is mostly a no-op visually; however, there is a single extra line of text above the logo than there was before due to how the cfonts lib under the covers renders the logo. I think that's fine but feel free to disapprove.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2612,6 +2612,37 @@
         "node": ">=6"
       }
     },
+    "node_modules/cfonts": {
+      "version": "3.3.0",
+      "resolved": "https://registry.npmjs.org/cfonts/-/cfonts-3.3.0.tgz",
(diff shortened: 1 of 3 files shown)


scottdensmore:
@NTaylorMullen where is that component now? Is InkBigText now custom somewhere?

NTaylorMullen:
> @NTaylorMullen where is that component now? Is InkBigText now custom somewhere?

It's the official ink-big-text library 😄 . See: https://github.com/sindresorhus/ink-big-text

To be fair we could inline it, the logic is quite simple.

PR #659, opened 2025-05-31
Fix: Make file path case-insensitive in @-command

scottdensmore:
This commit addresses issue #561 by modifying the @-command functionality to handle file paths in a case-insensitive manner.

Changes include:
- Updated atCommandProcessor.test.ts to include tests for case-insensitive file paths.
- Modified useCompletion.ts to perform case-insensitive matching when suggesting file paths.


NTaylorMullen:
nice!

PR #660, opened 2025-05-31
Fix for validating getDescription in read_file tool call

anj-s:
Fixes https://github.com/google-gemini/gemini-cli/issues/533

* Adding an error check for the path params.

### DIFF
--- a/packages/core/src/tools/read-file.ts
+++ b/packages/core/src/tools/read-file.ts
@@ -93,6 +93,9 @@ export class ReadFileTool extends BaseTool<ReadFileToolParams, ToolResult> {
   }
 
   getDescription(params: ReadFileToolParams): string {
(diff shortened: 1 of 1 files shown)


packages/core/src/tools/read-file.ts
```diff
@@ -93,6 +93,13 @@ export class ReadFileTool extends BaseTool<ReadFileToolParams, ToolResult> {
   }
 
   getDescription(params: ReadFileToolParams): string {
+    if (
+      !params ||
+      typeof params.path !== 'string' ||
+      params.path.trim() === ''
+    ) {
+      return `[${ReadFileTool.Name}: Path information unavailable]`;
```
NTaylorMullen: This bit is shown to the user so I'd suggest:

```suggestion
      return `Path unavailable`;
```

NTaylorMullen (approved):

PR #663, opened 2025-06-01
Modify shortenPath and add param validation

### LINKED ISSUE #612: Fatal error of Edit tool being called with undefined string
**Describe the bug**
Image

anj-s:
Fixes #612 #646

* Added param validation for getDescription since we end up calling it with request args. Will follow up and see if we can call validateParams before the mapAndDisplay.
* Added logic to add the last segment to the file path we display for EditTool. Also increase max character length for old/new snippet to 80 character with us displaying at least 3 lines. Let me know if we should increase or keep this the same.


packages/core/src/tools/edit.ts
```diff
@@ -350,16 +350,35 @@ Expectation for parameters:
   }
 
   getDescription(params: EditToolParams): string {
+    if (!params.file_path || !params.old_string || !params.new_string) {
+      return `EditTool params cannot be empty {file_path: ${params.file_path}, old_string: ${params.old_string}, new_string: ${params.new_string}}`;
```
NTaylorMullen: This is surfaced to the user so we may want to phrase this more in a generic sense. i.e. `Model did not provide X`

packages/core/src/tools/edit.ts
```diff
@@ -350,16 +350,35 @@ Expectation for parameters:
...
     if (params.old_string === '') {
       return `Create ${shortenPath(relativePath)}`;
     }
-    const oldStringSnippet =
-      params.old_string.split('\n')[0].substring(0, 30) +
-      (params.old_string.length > 30 ? '...' : '');
-    const newStringSnippet =
-      params.new_string.split('\n')[0].substring(0, 30) +
-      (params.new_string.length > 30 ? '...' : '');
+    const MAX_SNIPPET_LINES = 3;
```
NTaylorMullen: AFAIK we have descriptions as one line and short because they're displayed to user. Do we happen to support 3x lines well in the UI?

packages/core/src/utils/paths.ts
```diff
@@ -68,6 +69,10 @@ export function shortenPath(filePath: string, maxLen: number = 35): string {
       currentLength = lengthWithSegment;
     } else {
       // Adding this segment would exceed maxLen
+      if (endPartSegments.length === 0) {
+        // Lets try to keep the last directory so that we have "firstDir/.../lastDir"
+        endPartSegments.unshift(lastDir);
```
NTaylorMullen: This on its own will potentially expand us past the max length. What if we changed the boundaries on the for loop to always have first and last segments and then we can check length etc. before and after?

packages/core/src/tools/edit.ts
anj-s: Done.

packages/core/src/tools/edit.ts
anj-s: Well if they are only one line, do you think we should not support displaying it as part of getDescription? I see a truncated line which does not give me info about the edit proposed.  I think we see the full diff before we ask for user confirmation. Would it be sufficient to show it at that point alone?

packages/core/src/utils/paths.ts
anj-s: Since we have this final check of max length on result on line 88 I figured we could set the last directory as the default value. Is that final check not sufficient? I can refactor the loop here if that makes more sense.

packages/core/src/tools/edit.ts
NTaylorMullen: I'm not sure what a 3x line description might look like to be honest. I could imagine it being great or being repetitive. Right now it truncates because it expects more content to land below it (the result display) to drive most of the understanding.

Just for clarity though the full diff will be showed as part of the result display (not the description).

packages/core/src/utils/paths.ts
NTaylorMullen: ooooo I think that max length check might be responsible for losing the final file path though isn't it? I.e. the root cause of https://github.com/google-gemini/gemini-cli/issues/646

NTaylorMullen: As for refactoring the loop, I think that's a path. I think ultimately for #646 I think the root prioritization should be:

file > base directory > inbetween directories

And trimming should be aware in that direction too imo

packages/core/src/tools/edit.ts
anj-s: Discussed. Will revert to previous change.

packages/core/src/utils/paths.ts
```diff
@@ -67,27 +68,25 @@ export function shortenPath(filePath: string, maxLen: number = 35): string {
...
     }
   }
 
-  // Construct the final path
-  let result = startComponent + separator + '...';
-  if (endPartSegments.length > 0) {
-    result += separator + endPartSegments.join(separator);
+  let result = endPartSegments.join(separator) + separator + lastDir;
+  // if we have reached the limit return the path as is
+
```
scidomino, line 76: I think you want the line return before this comment instead of after it.

packages/core/src/utils/paths.ts
```diff
@@ -67,27 +68,25 @@ export function shortenPath(filePath: string, maxLen: number = 35): string {
...
+  // truncate the result string from the beginning, prefixing with "...".
   if (result.length > maxLen) {
-    const keepLen = Math.floor((maxLen - 3) / 2);
-    if (keepLen <= 0) {
-      return filePath.substring(0, maxLen - 3) + '...';
-    }
-    const start = filePath.substring(0, keepLen);
-    const end = filePath.substring(filePath.length - keepLen);
-    return `${start}...${end}`;
+    const numCharsFromResult = maxLen - 3;
```
scidomino: extreme nit: inline this.

scidomino (approved):

packages/core/src/utils/paths.ts
```diff
@@ -51,14 +51,15 @@ export function shortenPath(filePath: string, maxLen: number = 35): string {
   }
 
   const firstDir = segments[0];
+  const lastDir = segments[segments.length - 1];
```
NTaylorMullen:
```suggestion
  const lastSegment = segments[segments.length - 1];
```

NTaylorMullen (approved):

packages/core/src/utils/paths.ts
anj-s: Done

packages/core/src/utils/paths.ts
anj-s: Removed comment since it is confusing.

packages/core/src/utils/paths.ts
anj-s: Done

PR #666, opened 2025-06-01
A2A Server Implementation

cornmander:
Basic A2A server implementation that uses the core package as the agent kernel.

Tool calls don't work yet and just get echoed back to the CLI client.

Run by starting the server and the CLI tester:

(in server package) npm start
(in a2alib package) npm run a2a:cli -- http://localhost:41242

Vendors code from https://github.com/google-a2a/a2a-samples/tree/main/samples/js/src to an a2alib package.


cornmander:
@NTaylorMullen I'm not a typescript or NPM packaging expert, so I wasn't exactly sure how to make the server package depend on the a2alib package, so there might be some stuff in metadata files like tsconfig.json and package.json that should be revised. Please LMK

packages/a2alib/src/client/README.md
```diff
@@ -0,0 +1,119 @@
...
+      message: {
+        role: 'user',
+        parts: [{ text: 'Hello, agent!', type: 'text' }],
+      },
+    };
+    // Method now returns Task | null directly
+    const taskResult: Task | null = await client.sendTask(sendParams);
+    console.log('Send Task Result:', taskResult);
+
+    // Get task status (pass only params)
```
scidomino: here and elsewhere, "pass only params" doesn't seem like a useful comment. What else are you going to pass?

packages/a2alib/src/client/README.md
```diff
@@ -0,0 +1,119 @@
...
+        parts: [{ text: 'Stream me some updates!', type: 'text' }],
+      },
+    };
+    // Pass only params to the client method
+    const stream = client.sendTaskSubscribe(streamParams);
+
+    // Stream now yields the event payloads directly
+    for await (const event of stream) {
+      // Type guard to differentiate events based on structure
+      if ('status' in event) {
```
scidomino: Here and elsewhere: Instead of looking at fields that are unique to that event type, look at the value in the "kind" field instead.

packages/a2alib/src/client/client.ts
```diff
@@ -0,0 +1,591 @@
...
+  TaskIdParams,
+  CancelTaskResponse,
+  TaskPushNotificationConfig, // Renamed from PushNotificationConfigParams for direct schema alignment
+  SetTaskPushNotificationConfigResponse,
+  GetTaskPushNotificationConfigResponse,
+  A2AErrorData, // Renamed from A2AError to A2AErrorData
+} from '../schema.js'; // Assuming schema.ts is in the same directory or appropriately pathed
+
+// Helper type for the data yielded by streaming methods
+type A2AStreamEventData =
```
scidomino, line 32: name this SendStreamingMessageSuccessResponse so it matches the name in the specification.

packages/a2alib/src/client/client.ts
```diff
@@ -0,0 +1,591 @@
...
+  | Message
+  | Task
+  | TaskStatusUpdateEvent
+  | TaskArtifactUpdateEvent;
+
+/**
+ * A2AClient is a TypeScript HTTP client for interacting with A2A-compliant agents.
```
scidomino: nit: A2AClient is a TypeScript HTTP client for interacting with A2A server.

packages/a2alib/src/client/client.ts
```diff
@@ -0,0 +1,591 @@
...
+    this.agentBaseUrl = agentBaseUrl.replace(/\/$/, ''); // Remove trailing slash if any
+    this.agentCardPromise = this._fetchAndCacheAgentCard();
+  }
+
+  /**
+   * Fetches the Agent Card from the agent's well-known URI and caches its service endpoint URL.
+   * This method is called by the constructor.
+   * @returns A Promise that resolves to the AgentCard.
+   */
+  private async _fetchAndCacheAgentCard(): Promise<AgentCard> {
```
scidomino: This seems to duplicate a lot of the functionality of the getAgentCard(). The name is misleading because it implies that it caches the agent card which it doesn't do. Combine them.

packages/a2alib/src/client/client.ts
```diff
@@ -0,0 +1,591 @@
...
+        throw new Error(
+          `Failed to fetch Agent Card from ${agentCardUrl}: ${response.status} ${response.statusText}`,
+        );
+      }
+      const agentCard = (await response.json()) as AgentCard;
+      if (!agentCard.url) {
+        throw new Error(
+          "Fetched Agent Card does not contain a valid 'url' for the service endpoint.",
+        );
+      }
```
scidomino, line 139: Caching this seems unnecessary since you can just get that from the agent card.

packages/a2alib/src/client/client.ts
```diff
@@ -0,0 +1,591 @@
...
+    // If serviceEndpointUrl is not set, it means the agent card fetch is pending or failed.
+    // Awaiting agentCardPromise will either resolve it or throw if fetching failed.
+    await this.agentCardPromise;
+    if (!this.serviceEndpointUrl) {
+      // This case should ideally be covered by the error handling in _fetchAndCacheAgentCard
+      throw new Error(
+        'Agent Card URL for RPC endpoint is not available. Fetching might have failed.',
+      );
+    }
+    return this.serviceEndpointUrl;
```
scidomino, line 203: We should just have the constructor create and cache the agentcard promise once and then call that promise here to fetch the url. This method should be a one liner.

cornmander:
@NTaylorMullen another thing I need your input on is the CLI package useToolScheduler code. Would like to understand more about the move to put this in the CLI code -- from an agent-to-agent interaction perspective you probably don't want agent A to be able to invoke tools internal to agent B to maintain some encapsulation and abstraction. But I can see a surface<->agent interaction needing this, where the agent code wants to trigger surface-side changes. So we should think about ways to support both.

packages/a2alib/src/client/client.ts
```diff
@@ -0,0 +1,591 @@
...
+          // Check if it's a JSON-RPC error structure
+          throw new Error(
+            `RPC error for ${method}: ${errorJson.error.message} (Code: ${errorJson.error.code}, HTTP Status: ${httpResponse.status}) Data: ${JSON.stringify(errorJson.error.data)}`,
+          );
+        } else if (!errorJson.jsonrpc) {
+          throw new Error(
+            `HTTP error for ${method}! Status: ${httpResponse.status} ${httpResponse.statusText}. Response: ${errorBodyText}`,
+          );
+        }
+      } catch (e) {
```
scidomino: This is doing some gross checking for errors it threw. 

Instead, have "await httpResponse.text()" in it's own try catch that throws a simple error that doesn't log the errorBodyText (since we don't have one). And then, once we know that succeeded, try parsing the errors.

packages/a2alib/src/client/client.ts
```diff
@@ -0,0 +1,591 @@
...
+          `HTTP error for ${method}! Status: ${httpResponse.status} ${httpResponse.statusText}. Response: ${errorBodyText}`,
+        );
+      }
+    }
+
+    const rpcResponse = (await httpResponse.json()) as
+      | JSONRPCResult<unknown>
+      | JSONRPCErrorResponse;
+
+    if (rpcResponse.id !== requestId) {
```
scidomino, line 240: lets just remove this.

packages/a2alib/src/client/client.ts
```diff
@@ -0,0 +1,591 @@
...
+        )
+          throw e;
+        throw new Error(
+          `HTTP error for ${method}! Status: ${httpResponse.status} ${httpResponse.statusText}. Response: ${errorBodyText}`,
+        );
+      }
+    }
+
+    const rpcResponse = (await httpResponse.json()) as
+      | JSONRPCResult<unknown>
```
scidomino, line 237: Make this a named type like A2AStreamEventData so this can just return it as the type directly.

packages/a2alib/src/client/client.ts
```diff
@@ -0,0 +1,591 @@
...
+  /**
+   * Sends a message to the agent and streams back responses using Server-Sent Events (SSE).
+   * Push notification configuration can be specified in `params.configuration`.
+   * Optionally, `params.message.contextId` or `params.message.taskId` can be provided.
+   * Requires the agent to support streaming (`capabilities.streaming: true` in AgentCard).
+   * @param params The parameters for sending the message.
+   * @returns An AsyncGenerator yielding A2AStreamEventData (Message, Task, TaskStatusUpdateEvent, or TaskArtifactUpdateEvent).
+   * The generator throws an error if streaming is not supported or if an HTTP/SSE error occurs.
+   */
+  async *sendMessageStream(
```
scidomino, line 276: There is a lot of common code between this and resubscribeTask(). combine them where possible.

packages/a2alib/src/client/client.ts
```diff
@@ -0,0 +1,591 @@
...
+        if (errorMessage.startsWith('HTTP error establishing stream')) throw e;
+        // Fallback if body is not JSON or parsing fails
+        throw new Error(
+          `HTTP error establishing stream for message/stream: ${response.status} ${response.statusText}. Response: ${errorBody || '(empty)'}`,
+        );
+      }
+      throw new Error(
+        `HTTP error establishing stream for message/stream: ${response.status} ${response.statusText}`,
+      );
+    }
```
scidomino: this duplicates a bunch of the error checking code found in the _postRpcRequest() you shoudl extract a helper function

scidomino (changes requested):

packages/a2alib/src/server/a2a_express_app.ts
```diff
@@ -0,0 +1,151 @@
...
+ */
+
+import express, { Request, Response, Express } from 'express';
+
+import { A2AError } from './error.js';
+import { schema } from './index.js';
+import { A2ARequestHandler } from './request_handler/a2a_request_handler.js';
+import { JsonRpcTransportHandler } from './transports/jsonrpc_transport_handler.js';
+
+export class A2AExpressApp {
```
scidomino, line 18: document this class

packages/a2alib/src/server/events/execution_event_queue.ts
```diff
@@ -0,0 +1,76 @@
...
+      this.resolvePromise();
+      this.resolvePromise = undefined;
+    }
+  }
+
+  /**
+   * Provides an async generator that yields events from the event bus.
+   * Stops when a Message event is received or a TaskStatusUpdateEvent with final=true is received.
+   */
+  async *events(): AsyncGenerator<AgentExecutionEvent, void, undefined> {
```
scidomino, line 43: Here and elsewhere: It's ok to use AsyncGenerator<AgentExecutionEvent>. Don't specify the other two unless you use them.

packages/a2alib/src/server/request_handler/a2a_request_handler.ts
```diff
@@ -0,0 +1,50 @@
...
+
+export interface A2ARequestHandler {
+  getAgentCard(): Promise<AgentCard>;
+
+  sendMessage(params: MessageSendParams): Promise<Message | Task>;
+
+  sendMessageStream(
+    params: MessageSendParams,
+  ): AsyncGenerator<
+    Message | Task | TaskStatusUpdateEvent | TaskArtifactUpdateEvent,
```
scidomino, line 27: Does the typescript binding not define this union type (SendStreamingMessageResponse ) already? We keep redefining it it lots of places.

packages/a2alib/src/server/request_handler/a2a_request_handler.ts
```diff
@@ -0,0 +1,50 @@
...
+  ): Promise<TaskPushNotificationConfig>;
+
+  getTaskPushNotificationConfig(
+    params: TaskIdParams,
+  ): Promise<TaskPushNotificationConfig>;
+
+  resubscribe(
+    params: TaskIdParams,
+  ): AsyncGenerator<
+    Task | TaskStatusUpdateEvent | TaskArtifactUpdateEvent,
```
scidomino, line 46: The specification says this returns SendStreamingMessageResponse which includes Message subtype (it's unlikely to send one though). For simplicity, have this return SendStreamingMessageResponse as well.

packages/a2alib/src/server/transports/jsonrpc_transport_handler.ts
```diff
@@ -0,0 +1,200 @@
...
+
+  /**
+   * Handles an incoming JSON-RPC request.
+   * For streaming methods, it returns an AsyncGenerator of JSONRPCResult.
+   * For non-streaming methods, it returns a Promise of a single JSONRPCMessage (Result or ErrorResponse).
+   */
+  async handle(
+    requestBody: unknown,
+  ): Promise<A2AResponse | AsyncGenerator<A2AResponse, void, undefined>> {
+    let rpcRequest: A2ARequest | undefined = undefined;
```
scidomino, line 39: use null instead of undefined.

packages/a2alib/src/server/transports/jsonrpc_transport_handler.ts
```diff
@@ -0,0 +1,200 @@
...
+      } else if (typeof requestBody === 'object' && requestBody !== null) {
+        // Cast. Further validation will confirm.
+        rpcRequest = requestBody as A2ARequest;
+      } else {
+        throw A2AError.parseError('Invalid request body type.');
+      }
+
+      if (
+        rpcRequest.jsonrpc !== '2.0' ||
+        !rpcRequest.method || // method is part of A2ARequest, so this check is fine
```
scidomino: remove this comment

packages/a2alib/src/server/transports/jsonrpc_transport_handler.ts
```diff
@@ -0,0 +1,200 @@
...
+    this.requestHandler = requestHandler;
+  }
+
+  /**
+   * Handles an incoming JSON-RPC request.
+   * For streaming methods, it returns an AsyncGenerator of JSONRPCResult.
+   * For non-streaming methods, it returns a Promise of a single JSONRPCMessage (Result or ErrorResponse).
+   */
+  async handle(
+    requestBody: unknown,
```
scidomino, line 37: Instead of passing the body which is of unknown type pass the express Request type.

packages/a2alib/src/server/transports/jsonrpc_transport_handler.ts
```diff
@@ -0,0 +1,200 @@
...
+    try {
+      if (typeof requestBody === 'string') {
+        // Attempt to parse, then cast. Further validation will confirm.
+        rpcRequest = JSON.parse(requestBody) as A2ARequest;
+      } else if (typeof requestBody === 'object' && requestBody !== null) {
+        // Cast. Further validation will confirm.
+        rpcRequest = requestBody as A2ARequest;
+      } else {
+        throw A2AError.parseError('Invalid request body type.');
+      }
```
scidomino, line 50: The Express documentation make it look like this is not the right way to parse a request body. 

I'm pretty sure that only one of these branches are necessary since this method only has one caller and presumably it does it the same way every time.

packages/a2alib/src/server/transports/jsonrpc_transport_handler.ts
```diff
@@ -0,0 +1,200 @@
...
+  constructor(requestHandler: A2ARequestHandler) {
+    this.requestHandler = requestHandler;
+  }
+
+  /**
+   * Handles an incoming JSON-RPC request.
+   * For streaming methods, it returns an AsyncGenerator of JSONRPCResult.
+   * For non-streaming methods, it returns a Promise of a single JSONRPCMessage (Result or ErrorResponse).
+   */
+  async handle(
```
scidomino, line 36: This method is gigantic and does several different things. Please break it up.

scidomino (changes requested):

packages/a2alib/src/client/README.md
cornmander: I'm going to remove all the code from the READMEs, as it's just going to become immediately stale.

packages/a2alib/src/server/a2a_express_app.ts
cornmander: done

packages/a2alib/src/server/request_handler/a2a_request_handler.ts
cornmander: SendStreamingMessageResponse is a union of SendStreamingMessageSuccessResponse and JSONRPCErrorResponse and SendStreamingMessageSuccessResponse is JSONRPCResult<
    Message | Task | TaskStatusUpdateEvent | TaskArtifactUpdateEvent>.

Is JSONRPCResult<T> type-equivalent to T? My understanding is that in JSONRPCResult, T is used for the "result" field. So maybe we should add the  Message | Task | TaskStatusUpdateEvent | TaskArtifactUpdateEvent as a type union called SendStreamingMessageSuccessResponseResult or something in the schema code. LMK.

packages/a2alib/src/server/request_handler/a2a_request_handler.ts
cornmander: Same as above, I think we can't use SendStreamingMessageResponse because it's JSONRPCResult<T> and not T.

packages/a2alib/src/server/transports/jsonrpc_transport_handler.ts
cornmander: Done.

packages/a2alib/src/server/transports/jsonrpc_transport_handler.ts
cornmander: done

packages/a2alib/src/client/README.md
cornmander: done

packages/a2alib/src/server/transports/jsonrpc_transport_handler.ts
cornmander: done

packages/a2alib/src/client/client.ts
cornmander: Done.

eslint.config.js
```diff
@@ -32,6 +32,8 @@ export default tseslint.config(
       'eslint.config.js',
       'packages/cli/dist/**',
       'packages/core/dist/**',
+      'packages/server/dist/**',
+      'packages/a2alib/dist/**',
```
NTaylorMullen, line 36: From a naming perspective do you view this as an a2a client, an a2a server or truly a lib? I.e. if it can be executed I don't know if I'd call it a lib 😄

packages/a2alib/package.json
```diff
@@ -0,0 +1,38 @@
+{
+  "name": "@gemini-code/a2alib",
+  "version": "0.1.0",
+  "main": "dist/index.js",
+  "types": "dist/index.d.ts",
+  "description": "Vendored A2A Client and Server",
```
NTaylorMullen, line 6: Vendored? 😄

packages/a2alib/src/schema.ts
NTaylorMullen, line 1: Are these auto-generated or provided somewhere else? Assuming you didn't write each of them yourself we should add a note at the top indicating their origin

packages/a2alib/src/server/README.md
```diff
@@ -0,0 +1,3 @@
+# A2A Server (JS)
+
+This directory contains a TypeScript server implementation for the Agent-to-Agent (A2A) communication protocol, built using Express.js.
```
NTaylorMullen, line 3: Could we call this project `a2a-server` or something to that extent?

packages/server/README.md
```diff
@@ -0,0 +1,3 @@
+# @gemini-code/server
```
NTaylorMullen, line 1: ditto

packages/server/src/agent.ts
NTaylorMullen, line 1: Part of me feels like the bits that make it into agent.ts could be their own project a2a-agent

NTaylorMullen (commented):
You've built so much! I think there's a few things we should be thinking about with this PR and I'll brain dump them down here:

1. What do we want the state of all this to be once we OSS / what story do we want to tell?
2. How much boiler plate (like clients) do we want to write?
3. What sort of reusability structures do we want to implement to enable others to use this.

With those in mind here's some of my take:
1. We shouldn't be re-implementing an a2a client. You've done a super impressive job of basically making a generic TypeScript a2a client. Ideally we'd use something already done so as a2a evolves we don't need to spend a ton of cycles updating our own implementation.
2. We should be representing Gemini CLI's agent as its own package that potentially could be hoisted by others. I.e. truly a library of some sort that's light weight.
3. We should have an a2a server that hosts our agent (from 2) in order to let other A2A clients talk to it.
4. We should have an example that utilizes an a2a client already in the ecosystem that talks to our Gemini CLI a2a server.
5. **Stretch goal:** Have the GC cli act as an A2A client that can talk to any server

All that said, I think there's a number of different PRs / commits that could make up the above BUT splitting out what you have would be a monumentous amount of work on its own most likely. Therefore, I think my high level ask for this PR is to have a vision associated with it that's aligned with public OSS release timeframe so that if we get it in as-is we can be confident that it will be in a good state for the general public come release.

packages/a2alib/package.json
cornmander: They're based on the implementations in https://github.com/google-a2a/a2a-samples/tree/main/samples/js/src, but those are just starting points and needed some moderate modification. Agree that in the medium term we should not be rolling our own here

packages/a2alib/src/schema.ts
cornmander: I believe they're autogenerated from https://github.com/google-a2a/A2A/tree/main/types

packages/server/package.json
```diff
@@ -0,0 +1,36 @@
...
+    "typecheck": "tsc --noEmit",
+    "prerelease:version": "node ../../scripts/bind_package_version.js",
+    "prerelease:deps": "node ../../scripts/bind_package_dependencies.js",
+    "prepack": "npm run build"
+  },
+  "files": [
+    "dist"
+  ],
+  "dependencies": {
+    "@gemini-code/a2alib": "file:../a2alib",
```
KeijiBranshi, line 24: suggestion: I think setting this to `*` should be fine. npm will automatically link to the local workspace

```suggestion
    "@gemini-code/a2alib": "*",
```

NTaylorMullen:
@cornmander should we close this one out?

mattKorwel:

PR #667, opened 2025-06-01
Update edit tool validation function to override validateToolParams

ngleo:
This fixes #557.

`shouldConfirmExecute` is already calling `validateToolParams` to skip confirmation if params are invalid but `validateToolParams` is not overridden in the edit tool.

This PR simply renames existing validation function to `validateToolParams` to override it.

### DIFF
--- a/packages/core/src/tools/edit.ts
+++ b/packages/core/src/tools/edit.ts
@@ -126,7 +126,7 @@ Expectation for parameters:
    * @param params Parameters to validate
    * @returns Error message string or null if valid
    */
-  validateParams(params: EditToolParams): string | null {
+  validateToolParams(params: EditToolParams): string | null {
     if (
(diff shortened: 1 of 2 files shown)


ngleo:
The edit tool call presents an error without confirmation.
ok

packages/core/src/tools/edit.ts
```diff
@@ -126,7 +126,7 @@ Expectation for parameters:
    * @param params Parameters to validate
    * @returns Error message string or null if valid
    */
-  validateParams(params: EditToolParams): string | null {
+  validateToolParams(params: EditToolParams): string | null {
```
NTaylorMullen, line 129: nice, good catch!

NTaylorMullen (approved):

PR #669, opened 2025-06-01
Added replace tool ability to replace more than 1 occurrence

bdmorgan:
- **cleanup: removed duplicate check from Config.registerCoreTool()**
- **Fixing issue 530: Replace should enable replacing more than 1 occurrence**

### DIFF
--- a/packages/core/src/config/config.ts
+++ b/packages/core/src/config/config.ts
@@ -238,7 +238,7 @@ export function createToolRegistry(config: Config): ToolRegistry {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   const registerCoreTool = (ToolClass: any, ...args: unknown[]) => {
     // check both the tool name (.Name) and the class name (.name)
-    if (!tools || tools.has(ToolClass.Name) || tools.has(ToolClass.name)) {
+    if (!tools || tools.has(ToolClass.Name)) {
       registry.registerTool(new ToolClass(...args));
     }
   };
--- a/packages/core/src/config/config.ts
+++ b/packages/core/src/config/config.ts
@@ -238,7 +238,7 @@ export function createToolRegistry(config: Config): ToolRegistry {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   const registerCoreTool = (ToolClass: any, ...args: unknown[]) => {
     // check both the tool name (.Name) and the class name (.name)
-    if (!tools || tools.has(ToolClass.Name)) {
+    if (!tools || tools.has(ToolClass.Name) || tools.has(ToolClass.name)) {
       registry.registerTool(new ToolClass(...args));
     }
   };
--- a/packages/core/src/tools/edit.ts
+++ b/packages/core/src/tools/edit.ts
@@ -75,7 +75,7 @@ export class EditTool extends BaseTool<EditToolParams, ToolResult> {
     super(
       EditTool.Name,
       'Edit',
-      `Replaces text within a file. By default, replaces a single occurrence, but can replace multiple occurrences when expected_replacements is specified. This tool requires providing significant context around the change to ensure precise targeting. Always use the ${ReadFileTool} tool to examine the file's current content before attempting a text replacement.
+      `Replaces text within a file. By default, replaces a single occurrence, but can replace multiple occurrences when \`expected_replacements\` is specified. This tool requires providing significant context around the change to ensure precise targeting. Always use the ${ReadFileTool} tool to examine the file's current content before attempting a text replacement.
 
 Expectation for parameters:
 1. \`file_path\` MUST be an absolute path; otherwise an error will be thrown.
--- a/packages/core/src/tools/edit.ts
+++ b/packages/core/src/tools/edit.ts
@@ -83,7 +83,8 @@ Expectation for required parameters:
 3. \`new_string\` MUST be the exact literal text to replace \`old_string\` with (also including all whitespace, indentation, newlines, and surrounding code etc.). Ensure the resulting code is correct and idiomatic.
 4. NEVER escape \`old_string\` or \`new_string\`, that would break the exact literal text requirement.
 **Important:** If ANY of the above are not satisfied, the tool will fail. CRITICAL for \`old_string\`: Must uniquely identify the single instance to change. Include at least 3 lines of context BEFORE and AFTER the target text, matching whitespace and indentation precisely. If this string matches multiple locations, or does not match exactly, the tool will fail.,
-**Multiple replacements:** Set \`expected_replacements\` to the number of occurrences you want to replace. The tool will replace ALL occurrences that match \`old_string\` exactly. Ensure the number of replacements matches your expectation.`,      {
+**Multiple replacements:** Set \`expected_replacements\` to the number of occurrences you want to replace. The tool will replace ALL occurrences that match \`old_string\` exactly. Ensure the number of replacements matches your expectation.`,
+      {
         properties: {
           file_path: {
             description:
@@ -102,7 +103,8 @@ Expectation for required parameters:
           },
           expected_replacements: {
             type: 'number',
-            description: 'Number of replacements expected. Defaults to 1 if not specified. Use when you want to replace multiple occurrences.',
+            description:
+              'Number of replacements expected. Defaults to 1 if not specified. Use when you want to replace multiple occurrences.',
             minimum: 1,
           },
         },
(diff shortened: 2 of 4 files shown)


packages/core/src/config/config.ts
```diff
@@ -238,7 +238,7 @@ export function createToolRegistry(config: Config): ToolRegistry {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   const registerCoreTool = (ToolClass: any, ...args: unknown[]) => {
     // check both the tool name (.Name) and the class name (.name)
-    if (!tools || tools.has(ToolClass.Name) || tools.has(ToolClass.name)) {
```
NTaylorMullen: This needs to come back. The casing is different 😄 . That being said this is the second time this change was made like this SO you're not alone. Did the model do this or was it by eye? Trying to see if we need to change things here

packages/core/src/tools/edit.ts
```diff
@@ -69,14 +75,20 @@ export class EditTool extends BaseTool<EditToolParams, ToolResult> {
     super(
       EditTool.Name,
       'Edit',
-      `Replaces a single, unique occurrence of text within a file. This tool requires providing significant context around the change to ensure uniqueness and precise targeting. Always use the ${ReadFileTool} tool to examine the file's current content before attempting a text replacement.
+      `Replaces text within a file. By default, replaces a single occurrence, but can replace multiple occurrences when expected_replacements is specified. This tool requires providing significant context around the change to ensure precise targeting. Always use the ${ReadFileTool} tool to examine the file's current content before attempting a text replacement.
```
NTaylorMullen:
```suggestion
      `Replaces text within a file. By default, replaces a single occurrence, but can replace multiple occurrences when \`expected_replacements\` is specified. This tool requires providing significant context around the change to ensure precise targeting. Always use the ${ReadFileTool} tool to examine the file's current content before attempting a text replacement.
```

packages/core/src/tools/edit.ts
```diff
@@ -69,14 +75,20 @@ export class EditTool extends BaseTool<EditToolParams, ToolResult> {
...
 
 Expectation for parameters:
 1. \`file_path\` MUST be an absolute path; otherwise an error will be thrown.
 2. \`old_string\` MUST be the exact literal text to replace (including all whitespace, indentation, newlines, and surrounding code etc.).
 3. \`new_string\` MUST be the exact literal text to replace \`old_string\` with (also including all whitespace, indentation, newlines, and surrounding code etc.). Ensure the resulting code is correct and idiomatic.
-4. NEVER escape \`old_string\` or \`new_string\`, that would break the exact literal text requirement.
-**Important:** If ANY of the above are not satisfied, the tool will fail. CRITICAL for \`old_string\`: Must uniquely identify the single instance to change. Include at least 3 lines of context BEFORE and AFTER the target text, matching whitespace and indentation precisely. If this string matches multiple locations, or does not match exactly, the tool will fail.`,
+4. \`expected_replacements\` (optional) specifies the number of replacements expected. Defaults to 1. Use this when you want to replace multiple occurrences of the same text.
```
NTaylorMullen: Should remove this bullet, this bit is trying to encourage the model not to escape things. Instead I'd bring back the old bullet 4 + change the title here to be "Expectation for required parameters"

packages/core/src/tools/edit.ts
```diff
@@ -69,14 +75,20 @@ export class EditTool extends BaseTool<EditToolParams, ToolResult> {
...
+5. NEVER escape \`old_string\` or \`new_string\`, that would break the exact literal text requirement.
+
```
NTaylorMullen:
```suggestion
```

packages/core/src/tools/edit.ts
```diff
@@ -69,14 +75,20 @@ export class EditTool extends BaseTool<EditToolParams, ToolResult> {
...
+**Important:** If ANY of the above are not satisfied, the tool will fail. 
+
+**Single replacement (default):** Include at least 3 lines of context BEFORE and AFTER the target text, matching whitespace and indentation precisely. If this string matches multiple locations, the tool will fail unless expected_replacements is specified.
```
NTaylorMullen: What's the motivation behind this one? I think we could probably get away without it

packages/core/src/tools/edit.ts
```diff
@@ -69,14 +75,20 @@ export class EditTool extends BaseTool<EditToolParams, ToolResult> {
...
+
+**Multiple replacements:** Set expected_replacements to the number of occurrences you want to replace. The tool will replace ALL occurrences that match old_string exactly. Ensure the replacement count matches your expectation.`,
```
NTaylorMullen:
```suggestion
**Multiple replacements:** Set \`expected_replacements\` to the number of occurrences you want to replace. The tool will replace ALL occurrences that match \`old_string\` exactly. Ensure the number of replacements matches your expectation.`,
```

packages/core/src/tools/edit.ts
```diff
@@ -86,14 +98,19 @@ Expectation for parameters:
...
           },
           new_string: {
             description:
               'The exact literal text to replace `old_string` with, preferably unescaped. Provide the EXACT text. Ensure the resulting code is correct and idiomatic.',
             type: 'string',
           },
+          expected_replacements: {
+            type: 'number',
+            description: 'Number of replacements expected. Defaults to 1 if not specified. Use when you want to replace multiple occurrences.',
+            minimum: 1,
```
NTaylorMullen, line 108: I didn't realize that the Gemini API supported this. Would have figured it'd explode

packages/core/src/utils/editCorrector.ts
NTaylorMullen, line 1: There's more in this file that you'll need to work through. Basically everywhere which used to have occurrences === 1 is now occurrencxes === expected replacements

packages/core/src/config/config.ts
bdmorgan: okay fixed - please re-check (I think I had something in context from yesterday due to the _other_ bug I tried to fix...)

packages/core/src/tools/edit.ts
bdmorgan: done

packages/core/src/tools/edit.ts
bdmorgan: done

packages/core/src/tools/edit.ts
bdmorgan: removed it

packages/core/src/tools/edit.ts
bdmorgan: done

packages/core/src/utils/editCorrector.ts
bdmorgan: fixed - please check

packages/core/src/utils/editCorrector.ts
```diff
@@ -89,6 +89,29 @@ export async function ensureCorrectEdit(
       );
     }
   } else if (occurrences > 1) {
```
NTaylorMullen, line 91: @bdmorgan this should be `occurrences > expectedReplacements`. Gist is, if we're in a state where the model wasn't explicit enough then we can early exit.

packages/core/src/utils/editCorrector.ts
```diff
@@ -89,6 +89,29 @@ export async function ensureCorrectEdit(
...
+    const expectedReplacements = originalParams.expected_replacements ?? 1;
+
+    // If user expects multiple replacements, return as-is
+    if (occurrences === expectedReplacements) {
```
NTaylorMullen, line 95: This wont ever hit

packages/core/src/utils/editCorrector.ts
```diff
@@ -89,6 +89,29 @@ export async function ensureCorrectEdit(
...
+      const result: CorrectedEditResult = {
+        params: { ...originalParams },
+        occurrences,
+      };
+      editCorrectionCache.set(cacheKey, result);
+      return result;
+    }
+
+    // If user expects 1 but found multiple, try to correct (existing behavior)
+    if (expectedReplacements === 1) {
```
NTaylorMullen, line 105: No need for this, it's same as below

packages/core/src/utils/editCorrector.ts
NTaylorMullen, line 1: Instead of doing `?? 1` throught I'd suggest having a single variable that you normalize off of in the beginning:

```
const expectedReplacements = originalParams.expected_replacements ?? 1
```

packages/core/src/utils/editCorrector.ts
bdmorgan: fixed

packages/core/src/utils/editCorrector.ts
bdmorgan: fixed

packages/core/src/utils/editCorrector.ts
bdmorgan: fixed

packages/core/src/utils/editCorrector.ts
bdmorgan: fixed

PR #676, opened 2025-06-01
[HACK - do not merge] Enable yolo mode by default for non-interactive mode to unblock evals.

NTaylorMullen:
- Do not merge proper yolo coming from @tolikmalibroda

image


NTaylorMullen:
No longer needed.

PR #680, opened 2025-06-02
Color enhancements

miguelsolorio:
- Consolidates messaging colors (parenthesis always use `SubtleComment`)
- Adds colors for line numbers to reduce visual noise
- Make shortcuts consistent (lowercase)

CleanShot 2025-05-31 at 12 48 11@2x

CleanShot 2025-06-01 at 21 25 14@2x

### DIFF
--- a/packages/cli/src/ui/utils/CodeColorizer.tsx
+++ b/packages/cli/src/ui/utils/CodeColorizer.tsx
@@ -97,6 +97,11 @@ export function colorizeCode(
     // Render the HAST tree using the adapted theme
     // Apply the theme's default foreground color to the top-level Text element
     const lines = codeToHighlight.split('\n');
+    const getHighlightedLines = (line: string) =>
+      !language || !lowlight.registered(language)
+        ? lowlight.highlightAuto(line)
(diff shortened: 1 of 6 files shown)


NTaylorMullen:
Looks SO good @miguelsolorio!! Honestly it's kind of crazy how much those subtle comment additions really amplify readability.

Btw, I noticed in your image that you're still using gemini-2.5-pro-03-25, any reason in particular vs the default 05-06?

NTaylorMullen (approved):

packages/cli/src/ui/utils/CodeColorizer.tsx
```diff
@@ -94,16 +94,28 @@ export function colorizeCode(
   const activeTheme = themeManager.getActiveTheme();
 
   try {
-    const hastTree =
```
jacob314, line 97: keep hasTree as a const rather than moving it into the body. Wonder why GC did this. It regresses performance slightly and makes the code a bit harder to read.

jacob314 (approved):
Looks good!
lgtm

PR #682, opened 2025-06-02
refactor: Update MCP tool discovery to use @google/genai - Also fixes JSON schema issues.

NTaylorMullen:
- Integrates `mcpToTool` from `@google/genai` for tool discovery, replacing the previous manual client interaction. This enables robust JSON Schema handling, allowing the `@google/genai` SDK to manage schema compatibility for MCP tools.
- Modifies `DiscoveredMCPTool` to accept a `CallableTool` instance instead of a raw MCP client.
- Adjusts `ToolRegistry` and related tests to align with the new discovery mechanism.
- Ensures MCP client connections are properly closed when no tools are registered from a server or if errors occur during discovery.
- Simplifies `DiscoveredMCPTool`s `execute` method to leverage `mcpTool.callTool` and handles the response `Part[]` for display.
- Enhances test mocks for `StdioClientTransport` and `SSEClientTransport` to include a `close` method.
- Improves error handling and logging during MCP tool discovery.

Some other key bits here is that the tool names are now shown nicely to the user (and to the model) as long as there's not conflicts. Also I built some content extraction logic to make MCP tool call outputs a little more human readable
image

Fixes https://b.corp.google.com/issues/420718585

### DIFF
--- a/packages/core/src/tools/mcp-client.ts
+++ b/packages/core/src/tools/mcp-client.ts
@@ -84,16 +84,13 @@ async function connectAndDiscover(
     transport.stderr.on('data', (data) => {
       const stderrStr = data.toString();
       // Filter out verbose INFO logs from some MCP servers
-      if (
-        !stderrStr.includes('] INFO') &&
-        !stderrStr.includes('DEBUG rustls')
-      ) {
+      if (!stderrStr.includes('] INFO')) {
         console.debug(`MCP STDERR (${mcpServerName}):`, stderrStr);
       }
     });
   }
 
-  const toolRegistry = config.getToolRegistry();
+  const toolRegistry = await config.getToolRegistry();
   try {
     const mcpCallableTool: CallableTool = mcpToTool(mcpClient);
     const discoveredToolFunctions = await mcpCallableTool.tool();
@@ -122,13 +119,14 @@ async function connectAndDiscover(
       }
 
       let toolNameForModel = funcDecl.name;
+
+      // Replace invalid characters (based on 400 error message from Gemini API) with underscores
+      toolNameForModel = toolNameForModel.replace(/[^a-zA-Z0-9_.-]/g, '_');
+
       const existingTool = toolRegistry.getTool(toolNameForModel);
       if (existingTool) {
         toolNameForModel = mcpServerName + '__' + toolNameForModel;
       }
-
-      // Replace invalid characters (based on 400 error message from Gemini API) with underscores
-      toolNameForModel = toolNameForModel.replace(/[^a-zA-Z0-9_.-]/g, '_');
 
       // If longer than 63 characters, replace middle with '___'
       // (Gemini API says max length 64, but actual limit seems to be 63)
@@ -161,25 +159,27 @@ async function connectAndDiscover(
       `Failed to list or register tools for MCP server '${mcpServerName}': ${error}`,
     );
     // Ensure transport is cleaned up on error too
-    if (transport instanceof StdioClientTransport) {
-      await transport.close();
-    } else if (transport instanceof SSEClientTransport) {
+    if (
+      transport instanceof StdioClientTransport ||
+      transport instanceof SSEClientTransport
+    ) {
       await transport.close();
     }
   }
 
-  // Note: We might want to keep the client connected if tools are registered successfully,
-  // or dispose/close it if no tools were found or an error occurred.
-  // For now, let's assume the client connection is managed by the lifecycle of the DiscoveredMCPTool instances
-  // or should be explicitly closed if no tools are usable from this server.
-  // If no tools were registered from this server, we can probably close the connection.
+  // If no tools were registered from this MCP server, the following 'if' block
+  // will close the connection. This is done to conserve resources and prevent
+  // an orphaned connection to a server that isn't providing any usable
+  // functionality. Connections to servers that did provide tools are kept
+  // open, as those tools will require the connection to function.
   if (toolRegistry.getToolsByServer(mcpServerName).length === 0) {
     console.log(
       `No tools registered from MCP server '${mcpServerName}'. Closing connection.`,
     );
-    if (transport instanceof StdioClientTransport) {
-      await transport.close();
-    } else if (transport instanceof SSEClientTransport) {
+    if (
+      transport instanceof StdioClientTransport ||
+      transport instanceof SSEClientTransport
+    ) {
       await transport.close();
     }
   }
--- a/packages/core/src/tools/mcp-client.ts
+++ b/packages/core/src/tools/mcp-client.ts
@@ -10,12 +10,9 @@ import { SSEClientTransport } from '@modelcontextprotocol/sdk/client/sse.js';
 import { parse } from 'shell-quote';
 import { Config, MCPServerConfig } from '../config/config.js';
 import { DiscoveredMCPTool } from './mcp-tool.js';
-import { ToolRegistry } from './tool-registry.js';
+import { CallableTool, FunctionDeclaration, mcpToTool } from '@google/genai';
 
-export async function discoverMcpTools(
-  config: Config,
-  toolRegistry: ToolRegistry,
-): Promise<void> {
+export async function discoverMcpTools(config: Config): Promise<void> {
   const mcpServers = config.getMcpServers() || {};
 
   if (config.getMcpServerCommand()) {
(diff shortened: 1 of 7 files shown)


packages/core/src/tools/mcp-client.ts
```diff
@@ -82,63 +73,84 @@ async function connectAndDiscover(
...
-    console.error('MCP ERROR', error.toString());
+    console.error(`MCP ERROR (${mcpServerName}):`, error.toString());
   };
 
   if (transport instanceof StdioClientTransport && transport.stderr) {
     transport.stderr.on('data', (data) => {
-      if (!data.toString().includes('] INFO')) {
-        console.debug('MCP STDERR', data.toString());
+      const stderrStr = data.toString();
+      // Filter out verbose INFO logs from some MCP servers
```
olcan, line 86: looks like not just INFO logs now

packages/core/src/tools/mcp-client.ts
```diff
@@ -82,63 +73,84 @@ async function connectAndDiscover(
...
+    for (const funcDecl of discoveredToolFunctions.functionDeclarations) {
+      if (!funcDecl.name) {
+        console.warn(
+          `Discovered a function declaration without a name from MCP server '${mcpServerName}'. Skipping.`,
+        );
+        continue;
+      }
+
+      let toolNameForModel = funcDecl.name;
+      const existingTool = toolRegistry.getTool(toolNameForModel);
```
olcan, line 126: this existence check should be done after toolNameForModel is finalized below

packages/core/src/tools/mcp-client.ts
```diff
@@ -148,6 +160,27 @@ async function connectAndDiscover(
     console.error(
       `Failed to list or register tools for MCP server '${mcpServerName}': ${error}`,
     );
-    // Do not re-throw, allow other servers to proceed.
+    // Ensure transport is cleaned up on error too
+    if (transport instanceof StdioClientTransport) {
```
olcan: these if statements should be combined

packages/core/src/tools/mcp-client.ts
```diff
@@ -148,6 +160,27 @@ async function connectAndDiscover(
...
+  // Note: We might want to keep the client connected if tools are registered successfully,
+  // or dispose/close it if no tools were found or an error occurred.
+  // For now, let's assume the client connection is managed by the lifecycle of the DiscoveredMCPTool instances
+  // or should be explicitly closed if no tools are usable from this server.
+  // If no tools were registered from this server, we can probably close the connection.
+  if (toolRegistry.getToolsByServer(mcpServerName).length === 0) {
+    console.log(
+      `No tools registered from MCP server '${mcpServerName}'. Closing connection.`,
+    );
+    if (transport instanceof StdioClientTransport) {
```
olcan: same as above, i.e. if statements can/should be combined

packages/core/src/tools/mcp-client.ts
```diff
@@ -148,6 +160,27 @@ async function connectAndDiscover(
...
-    // Do not re-throw, allow other servers to proceed.
+    // Ensure transport is cleaned up on error too
+    if (transport instanceof StdioClientTransport) {
+      await transport.close();
+    } else if (transport instanceof SSEClientTransport) {
+      await transport.close();
+    }
+  }
+
+  // Note: We might want to keep the client connected if tools are registered successfully,
```
olcan: i find this comment block confusing, would be good to rewrite this to just say what is being done here and why, with the latter being more important

packages/core/src/tools/mcp-tool.ts
```diff
@@ -62,15 +69,15 @@ Returns the MCP server response as a json string.
       DiscoveredMCPTool.whitelist.has(serverWhitelistKey) ||
       DiscoveredMCPTool.whitelist.has(toolWhitelistKey)
     ) {
-      return false; // server and/or tool already whitelisted
+      return false; // server and/or tool already allow listed
```
olcan, line 69: the variables are still called whitelist so this should be consistent

packages/core/src/tools/mcp-tool.ts
```diff
@@ -32,11 +32,18 @@ export class DiscoveredMCPTool extends BaseTool<ToolParams, ToolResult> {
   ) {
     description += `
 
-This MCP tool was discovered from a local MCP server using JSON RPC 2.0 over stdio transport protocol.
-When called, this tool will invoke the \`tools/call\` method for tool name \`${name}\`.
+This MCP tool was discovered from a local MCP server.`;
```
olcan: looks like the description was outdated since we added SSE, but also, the server is no longer "local" necessarily, so we should drop that word too

packages/core/src/tools/mcp-tool.ts
```diff
@@ -32,11 +32,18 @@ export class DiscoveredMCPTool extends BaseTool<ToolParams, ToolResult> {
...
+    if (serverName !== 'mcp') {
```
olcan, line 33: appending these things feels really hacky. could we clean this up as something like this:
if (serverName !== 'mcp') {
  This MCP tool named '<tool>' was discovered from MCP server '<server>'.
} else {
  This MCP tool named '<tool>' was discovered from a local MCP server.
}

packages/core/src/tools/mcp-tool.ts
```diff
@@ -62,15 +69,15 @@ Returns the MCP server response as a json string.
...
+      return false; // server and/or tool already allow listed
     }
 
     const confirmationDetails: ToolMcpConfirmationDetails = {
       type: 'mcp',
       title: 'Confirm MCP Tool Execution',
       serverName: this.serverName,
-      toolName: this.serverToolName,
-      toolDisplayName: this.name,
+      toolName: this.serverToolName, // Display original tool name in confirmation
```
olcan, line 76: i would say "Display original (server-assigned) tool name alongside server name"

packages/core/src/tools/mcp-tool.ts
```diff
@@ -62,15 +69,15 @@ Returns the MCP server response as a json string.
...
+      toolDisplayName: this.name, // Display mangled name as it's what user sees
```
olcan: I would say "Display global registry name exposed to model and user"

packages/core/src/tools/mcp-tool.ts
```diff
@@ -83,20 +90,58 @@ Returns the MCP server response as a json string.
...
+    return part; // Fallback for unexpected structure or non-FunctionResponsePart
+  };
+
+  if (result.length === 1) {
+    const processedResponse = processFunctionResponse(result[0]);
+    if (typeof processedResponse === 'string') {
+      // If it's a simple string, no need for JSON stringification or markdown
+      return processedResponse;
+    }
+    return '```json\n' + JSON.stringify(processedResponse, null, 2) + '\n```';
```
olcan: redundant code with immediately below

packages/core/src/tools/mcp-tool.ts
```diff
@@ -83,20 +90,58 @@ Returns the MCP server response as a json string.
...
+    const output = getStringifiedResultForDisplay(responseParts);
     return {
-      llmContent: output,
+      llmContent: responseParts,
       returnDisplay: output,
     };
   }
 }
+
+function getStringifiedResultForDisplay(result: Part[]) {
```
olcan, line 127: found this hard to follow, would benefit from a high level description of what kind of parts can exist and how we deal with them

packages/core/src/tools/tool-registry.ts
```diff
@@ -178,6 +178,20 @@ export class ToolRegistry {
     return Array.from(this.tools.values());
   }
 
+  /**
+   * Returns an array of tools registered from a specific MCP server.
+   */
+  getToolsByServer(serverName: string): Tool[] {
+    const serverTools: Tool[] = [];
+    for (const tool of this.tools.values()) {
+      // Check if the tool instance has a serverName property and if it matches
```
olcan: comment not useful

packages/core/src/tools/mcp-client.ts
NTaylorMullen: Bringing back to just be info logs

packages/core/src/tools/mcp-client.ts
NTaylorMullen: Ahhh really good catch!

packages/core/src/tools/mcp-tool.ts
NTaylorMullen: I'm not sure how important "local" is here. Going to simplify further.

PR #684, opened 2025-06-02
remove redundant call to discoverTools (already called during registry init)

olcan:


NTaylorMullen (approved):

olcan:
there is a complication w/ discoverTools now returning a promise.
i'll have to revisit this tomorrow

PR #685, opened 2025-06-02
fix: Refine model message consolidation for improved model interaction

NTaylorMullen:
- Adjusts chat history consolidation to allow separate model messages for text content and function calls.
- Previously, text and function call parts from the model were always merged into a single model message. This change allows them to exist as distinct consecutive model messages if they are of different types (e.g., a text response followed by a function call response).
- This is a partial reversion of earlier consolidation logic, aiming to address potential model quality issues observed. The hypothesis is that cleaner separation of response types, even if it means less strict user/model turn rotation, might lead to better model behavior.
- Consolidates adjacent text parts within the *same* model message if they are both text.
- Adds `isTextContent` helper and updates tests.

Part of https://b.corp.google.com/issues/421010429

### DIFF
--- a/packages/core/src/core/geminiChat.test.ts
+++ b/packages/core/src/core/geminiChat.test.ts
@@ -69,10 +69,7 @@ describe('GeminiChat', () => {
       expect(history.length).toBe(2);
       expect(history[0]).toEqual(userInput);
       expect(history[1].role).toBe('model');
-      expect(history[1].parts).toEqual([
-        { text: 'Model part 1' },
-        { text: 'Model part 2' },
-      ]);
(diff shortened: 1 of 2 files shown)


NTaylorMullen:
Getting this in for now. @olcan you'll be especially surprised by this PR. Gist is that I observed a lot of model quality issues over the weekend. Now while that could have been impacted by actual behind the scenes work I have my doubts. The only real impacting change I observed prior to this weekend was the message consolidation change. Therefore, to limit the risk in model quality regressions I'm bringing back a little bit of the non-consolidated world that we had before. Take a look at the tests to get a good idea of what this does, the commit message also does a decent job explaining.

PR #692, opened 2025-06-02
Fix theme selector bug

miguelsolorio:
This attempts to fix the theme selectors bug where if you select a theme the dialog still appears. @jacob314 since you wrote most of the code in this file, assigning to you for review. Not sure if this is a great implementation.

### DIFF
--- a/packages/cli/src/ui/hooks/useThemeCommand.ts
+++ b/packages/cli/src/ui/hooks/useThemeCommand.ts
@@ -4,7 +4,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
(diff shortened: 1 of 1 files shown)


scidomino (approved):
Does this conflict with https://github.com/google-gemini/gemini-cli/pull/703?

If not: aproved.

jacob314 (commented):
Sorry I added a fix for this as well before realizing you also fixed it.
https://github.com/google-gemini/gemini-cli/pull/703

The one thing your pull request does that mine didn't is show 'No theme is set`.
If that is important let me know and I'll layer it on top of my change.

miguelsolorio:
Let's go with @jacob314's PR

PR #695, opened 2025-06-02
feat: Add --yolo mode that automatically accepts all tools executions

tolikmalibroda:
Implemented --yolo flag that enables automatic execution of all tools without confirmation.
Currently --yolo flag implies sandboxing but not enforces it hardly (e.g. sandboxing can be disabled using `-s false`).
Some security measures that I'll add in the upcoming PRs:
- Add allowlist of shell commands that are allowed to execute in --yolo mode.
- Restrict network connection within the sandbox to forbid arbitrary requests to the internet.
- Enforce sandboxing without ability to disable it for --yolo mode.

### DIFF
--- a/packages/cli/src/config/config.ts
+++ b/packages/cli/src/config/config.ts
@@ -80,7 +80,8 @@ async function parseArguments(): Promise<CliArgs> {
     .option('yolo', {
       alias: 'y',
       type: 'boolean',
-      description: 'Automatically accept all actions (aka YOLO mode)?',
+      description:
+        'Automatically accept all actions (aka YOLO mode, see https://www.youtube.com/watch?v=xvFZjo5PgG0 for more details)?',
       default: false,
     })
     .version() // This will enable the --version flag based on package.json
--- a/packages/cli/src/config/config.ts
+++ b/packages/cli/src/config/config.ts
@@ -38,6 +38,7 @@ interface CliArgs {
   prompt: string | undefined;
   all_files: boolean | undefined;
   show_memory_usage: boolean | undefined;
+  yolo: boolean | undefined;
 }
 
 async function parseArguments(): Promise<CliArgs> {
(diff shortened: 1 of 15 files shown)


packages/cli/src/config/config.ts
```diff
@@ -75,6 +77,12 @@ async function parseArguments(): Promise<CliArgs> {
       description: 'Show memory usage in status bar',
       default: false,
     })
+    .option('yolo', {
+      alias: 'y',
+      type: 'boolean',
+      description: 'Automatically accept all actions (aka YOLO mode)?',
```
NTaylorMullen: Would it be bad to update to:

```suggestion
      description: 'Automatically accept all actions (aka YOLO mode, see https://www.youtube.com/watch?v=xvFZjo5PgG0 for more details)?',
```

packages/cli/src/ui/components/AutoAcceptIndicator.tsx
```diff
@@ -7,12 +7,41 @@
...
+  switch (approvalMode) {
+    case ApprovalMode.AUTO_EDIT:
+      textColor = Colors.AccentGreen;
+      textContent = 'accepting edits';
+      subText = ' (shift + tab to toggle)';
+      break;
+    case ApprovalMode.YOLO:
+      textColor = Colors.AccentRed;
+      textContent = 'YOLO mode';
+      subText = ' (ctrl + y to toggle)';
```
NTaylorMullen, line 32: oOooooo I really like that you made yolo toggleable. Does this mean in a normal session you can also toggle it on?

packages/core/src/core/coreToolScheduler.ts
```diff
@@ -323,29 +327,33 @@ export class CoreToolScheduler {
...
-                outcome,
-              ),
-          };
-          this.setStatusInternal(
-            reqInfo.callId,
-            'awaiting_approval',
-            wrappedConfirmationDetails,
-          );
-        } else {
+        if (this.approvalMode === ApprovalMode.YOLO) {
```
NTaylorMullen, line 330: nice, I didn't think about this route for doing it but it makes senes.

NTaylorMullen (approved):
@tolikmalibroda super super clean. You found routes to doing things that I hadn't considered but ultimately I really like the approach you took.

:shipit:

packages/cli/src/ui/components/AutoAcceptIndicator.tsx
tolikmalibroda: Yes, it is possible to enable yolo mode already after starting the normal session

PR #702, opened 2025-06-02
Feat: Enable YOLO mode for non-interactive execution

NTaylorMullen:
- Refactor non-interactive tool loading to support YOLO mode.
- When YOLO mode is enabled, all tools (including interactive ones) can be used in non-interactive mode.
- Otherwise, non-interactive mode continues to restrict tools to read-only operations.

Fixes https://github.com/google-gemini/gemini-cli/issues/701

### DIFF
--- a/packages/cli/src/gemini.tsx
+++ b/packages/cli/src/gemini.tsx
@@ -13,11 +13,13 @@ import { readPackageUp } from 'read-package-up';
 import { fileURLToPath } from 'node:url';
 import { dirname } from 'node:path';
 import { sandbox_command, start_sandbox } from './utils/sandbox.js';
-import { loadSettings } from './config/settings.js';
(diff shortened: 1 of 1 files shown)


packages/cli/src/gemini.tsx
```diff
@@ -114,32 +116,14 @@ async function main() {
...
-    WebFetchTool.Name,
-    WebSearchTool.Name,
-    ReadManyFilesTool.Name,
-    ShellTool.Name,
-    MemoryTool.Name,
-  ];
-  const interactiveTools = [ShellTool.Name, EditTool.Name, WriteFileTool.Name];
-  const nonInteractiveTools = existingCoreTools.filter(
-    (tool) => !interactiveTools.includes(tool),
+  const nonInteractiveConfigResult = await loadNonInteractiveConfig(
```
olcan, line 119: redundant variable

PR #709, opened 2025-06-03
Consolidate sandbox code.

scidomino:
Move logic out of footer and prompt and into sandbox.ts.

### DIFF
--- a/packages/core/src/core/prompts.ts
+++ b/packages/core/src/core/prompts.ts
@@ -107,19 +107,11 @@ When requested to perform tasks like fixing bugs, adding features, refactoring, 
 - **Feedback:** To report a bug or provide feedback, please use the /bug command.
 
 ${(function () {
-  // Determine sandbox status based on environment variables
-  const isSandboxExec = process.env.SANDBOX === 'sandbox-exec';
-  const isGenericSandbox = !!process.env.SANDBOX; // Check if SANDBOX is set to any non-empty value
-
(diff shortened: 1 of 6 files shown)


packages/core/src/core/prompts.ts
```diff
@@ -107,19 +107,11 @@ When requested to perform tasks like fixing bugs, adding features, refactoring,
...
-  const isSandboxExec = process.env.SANDBOX === 'sandbox-exec';
-  const isGenericSandbox = !!process.env.SANDBOX; // Check if SANDBOX is set to any non-empty value
-
-  if (isSandboxExec) {
-    return `
-# MacOS Seatbelt
-You are running under macos seatbelt with limited access to files outside the project directory or system temp directory, and with limited access to host system resources such as ports. If you encounter failures that could be due to MacOS Seatbelt (e.g. if a command fails with 'Operation not permitted' or similar error), as you report the error to the user, also explain why you think it could be due to MacOS Seatbelt, and how the user may need to adjust their Seatbelt profile.
-`;
-  } else if (isGenericSandbox) {
+  const sandboxPromptText = process.env.SANDBOX_PROMPT;
```
NTaylorMullen, line 110: The environment variable doesn't feel like the right way to configure the sandbox information. Hmmm this is making me think we may want to invent `.gemini/extension/...` which then has its own custom settings to some extent

NTaylorMullen (approved):

PR #710, opened 2025-06-03
refactor: maintain 1 GeminiChat per GeminiClient

KeijiBranshi:
CHILD=#711
ISSUE=#494

Preliminary refactor for chat history compression. This way GeminiClient can compress history under the hood without GeminiChat references in other parts of the code going stale. Took a pass through and it didn't seem like we were creating multiple chats for a single client anyways


NTaylorMullen:
Solid refactor. Have already been using this to peek into history recently 😄

PR #711, opened 2025-06-03
feat: compress chat history when we approach token limit

KeijiBranshi:
DIFFBASE=#710
ISSUE=#494

### DIFF
--- a/packages/core/src/core/client.ts
+++ b/packages/core/src/core/client.ts
@@ -26,6 +26,7 @@ import { reportError } from '../utils/errorReporting.js';
 import { GeminiChat } from './geminiChat.js';
 import { retryWithBackoff } from '../utils/retry.js';
 import { getErrorMessage } from '../utils/errors.js';
+import { tokenLimit } from './tokenLimits.js';
 
 export class GeminiClient {
   private chat: Promise<GeminiChat>;
@@ -172,6 +173,7 @@ export class GeminiClient {
       return;
     }
 
+    await this.tryCompressChat();
     const chat = await this.chat;
     const turn = new Turn(chat);
     const resultStream = turn.run(request, signal);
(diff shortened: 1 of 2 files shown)


packages/core/src/core/client.ts
```diff
@@ -322,4 +324,53 @@ export class GeminiClient {
...
+      console.warn(
+        `No token limit defined for model ${this.model}. Skipping compression check.`,
+      );
+      return;
+    }
+
+    if (tokenCount < 0.95 * limit) {
+      return;
+    }
+    const summarizationRequestMessage = {
```
NTaylorMullen, line 359: Nice!

packages/core/src/core/client.ts
```diff
@@ -322,4 +324,53 @@ export class GeminiClient {
...
+      text: 'Summarize our conversation up to this point. The summary should be a concise yet comprehensive overview of all key topics, questions, answers, and important details discussed. This summary will replace the current chat history to conserve tokens, so it must capture everything essential to understand the context and continue our conversation effectively as if no information was lost.',
+    };
+    const response = await chat.sendMessage({
+      message: summarizationRequestMessage,
+    });
+    this.chat = this.startChat([
+      {
+        role: 'user',
+        parts: [summarizationRequestMessage],
```
NTaylorMullen, line 368: Ah super clever

packages/core/src/core/client.ts
```diff
@@ -172,6 +173,7 @@ export class GeminiClient {
       return;
     }
 
+    await this.tryCompressChat();
```
NTaylorMullen, line 176: We'll want to have a notification to the user that chat is being compressed. You may need to create a custom event indicating that OR you could have a custom tool call event that utilizes a tool under the hood for it too which might be cool looking.

KeijiBranshi: agreed. I can do that in a fast follow, just to keep the diffs small

PR #716, opened 2025-06-03
Refactor: Make MCP server discovery non-blocking

NTaylorMullen:
- Removed await for MCP server discovery in `Config`.
- This allows the main application flow to continue without waiting for all MCP servers to establish connections and announce their tools.
- Unexported the discover mcp tools function so we can ensure it only happens once.

Fixes https://github.com/google-gemini/gemini-cli/issues/712

PR #717, opened 2025-06-03
Fix: Ensure MCP tools are discovered from slow-starting servers

NTaylorMullen:
- Decouples MCP tool discovery from `Config` by moving the `discoverMcpTools` logic to `ToolRegistry`.
- `createToolRegistry` now awaits the completion of `discoverMcpTools`, ensuring that tools from all MCP servers, including slow-starting ones, are registered before the registry is considered fully initialized.
- Updates tests to reflect the new `discoverMcpTools` signature and calling convention.

Part of https://github.com/google-gemini/gemini-cli/issues/716

### DIFF
--- a/packages/core/src/config/config.ts
+++ b/packages/core/src/config/config.ts
@@ -254,7 +254,7 @@ export function createServerConfig(params: ConfigParameters): Config {
   });
 }
 
-function createToolRegistry(config: Config): Promise<ToolRegistry> {
+export function createToolRegistry(config: Config): Promise<ToolRegistry> {
   const registry = new ToolRegistry(config);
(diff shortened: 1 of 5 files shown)


packages/core/src/config/config.ts
```diff
@@ -281,12 +281,8 @@ function createToolRegistry(config: Config): Promise<ToolRegistry> {
   registerCoreTool(ShellTool, config);
   registerCoreTool(MemoryTool);
   registerCoreTool(WebSearchTool, config);
-
```
olcan, line 284: i'm confused... i made this change yesterday in #691 , was it dropped somehow?

PR #726, opened 2025-06-03
Update light themes

miguelsolorio:
- Fixes a bug where line numbers had extra left padding
- Fixes a bug where theme default foreground colors were using the wrong color (SubtleComment)
- Updates theme preview to show a better example with more code
- Updates default light theme colors for better contrast
- Adds new themes (Ayu, Ayu Light, GitHub, GitHub Light)


NTaylorMullen (approved):

packages/cli/src/ui/utils/CodeColorizer.tsx
```diff
@@ -133,10 +133,10 @@ export function colorizeCode(
       <Text>
         {lines.map((line, index) => (
           <Text key={index}>
-            <Text color={activeTheme.colors.SubtleComment}>
-              {`${String(index + 1).padStart(3, ' ')} `}
+            <Text color={activeTheme.defaultColor}>
+              {`${String(index + 1).padStart(0, ' ')} `}
```
jacob314: Can you ask Gemin to pick padStart based on the maximum index present.
padStart should equal 1 if the largest index is less than 10, 1 if it is less than 100, 2 if less than 100, 3 if less than 1000.

jacob314 (approved):
lgtm

packages/cli/src/ui/utils/CodeColorizer.tsx
miguelsolorio: Oh this was me 😅 I was trying to reduce the extra margin on the left and didn't realize this was contextual. I'll see if I can add some logic for it.

miguelsolorio: Decided to go with `const padWidth = String(lines.length).length` so that it'll scale out as it grows in number length

PR #732, opened 2025-06-03
bandaid for broken mcp calls

olcan:


olcan (commented):
NOTE: this is a bandaid fix. the function formatLlmContentForFunctionResponse should be reviewed carefully. also it seems tools that return a single number are being converted to an array of string, e.g. add_two_numbers(2,3) -> [ "5" ] , which is not great

olcan (commented):
NOTE: this is a bandaid fix. the function formatLlmContentForFunctionResponse should be reviewed carefully. also it seems tools that return a single number are being converted to an array of string, e.g. add_two_numbers(2,3) -> [ "5" ] , which is not great

PR #739, opened 2025-06-04
fix: Fix piped input mode in sandbox

tolikmalibroda:
Remove TTY from the container if the input to the gemini-cli was piped. If the gemini-cli's stdin was piped, than it's not TTY, so we shouldn't initialize TTY in the container.


NTaylorMullen (approved):

NTaylorMullen:
ooo nice!

PR #746, opened 2025-06-04
feat: Add multi-stage docker build support for custom sandbox.Dockerfile

tolikmalibroda:
To use custom sandbox image, user needs to create custom sandbox.Dockerfile in the .gemini folder. Previously it should've contain the full setup of gemini-cli. With this change, user can base their custom sandbox.Dockerfile on the base gemini-cli image (using `FROM gemini-cli-sandbox`). When gemini-cli detects custom sandbox.Dockerfile, it'll automatically build the custom image and launch gemini-cli inside.

### DIFF
--- a/docs/cli/configuration.md
+++ b/docs/cli/configuration.md
@@ -77,6 +77,7 @@ When you create a `.gemini/settings.json` file for project-specific settings, or
   - See the [Theming section in README.md](../../README.md#theming) for available theme names.
 - **`sandbox`** (boolean or string):
   - Controls whether and how to use sandboxing for tool execution.
+  - If a `.gemini/sandbox.Dockerfile` exists in your project, it will be used to build a custom sandbox image based on `gemini-cli-sandbox`.
   - `true`: Enable default sandbox (see [README](../../README.md) for behavior).
   - `false`: Disable sandboxing (WARNING: this is inherently unsafe).
   - `"docker"` or `"podman"`: Explicitly choose container-based sandboxing command.
@@ -275,6 +276,31 @@ This example demonstrates how you can provide general project context, specific 
   - See the [Commands documentation](./commands.md#memory) for full details on the `/memory` command and its sub-commands (`show` and `refresh`).
 
 By understanding and utilizing these configuration layers and the hierarchical nature of context files, you can effectively manage the AI's memory and tailor the Gemini CLI's responses to your specific needs and projects.
+
+## Sandboxing
+
+The Gemini CLI can execute potentially unsafe operations (like shell commands and file modifications) within a sandboxed environment to protect your system.
+
+Sandboxing is disabled by default, but you can enable it in a few ways:
+
+- Using `--sandbox` or `-s` flag.
+- Setting `GEMINI_SANDBOX` environment variable.
+- Sandbox is enabled in `--yolo` mode by default.
+
+By default, it uses a pre-built `gemini-cli-sandbox` Docker image.
+
+For project-specific sandboxing needs, you can create a custom Dockerfile at `.gemini/sandbox.Dockerfile` in your project's root directory. This Dockerfile should be based on the base sandbox image:
+
+```dockerfile
+FROM gemini-cli-sandbox
+
+# Add your custom dependencies or configurations here
+# For example:
+# RUN apt-get update && apt-get install -y some-package
+# COPY ./my-config /app/my-config
+```
+
+When `.gemini/sandbox.Dockerfile` exists, the CLI will automatically build and use a custom image for your project.
 
 ## Theming
 
--- a/packages/cli/src/utils/sandbox.ts
+++ b/packages/cli/src/utils/sandbox.ts
@@ -73,7 +73,9 @@ async function shouldUseCurrentUserInSandbox(): Promise<boolean> {
   return false; // Default to false if no other condition is met
 }
 
-async function getSandboxImageName(): Promise<string> {
+async function getSandboxImageName(
+  isCustomProjectSandbox: boolean,
+): Promise<string> {
   const packageJsonResult = await readPackageUp();
   const packageJsonConfig = packageJsonResult?.packageJson.config as
     | { sandboxImageUri?: string }
@@ -81,7 +83,9 @@ async function getSandboxImageName(): Promise<string> {
   return (
     process.env.GEMINI_SANDBOX_IMAGE ??
     packageJsonConfig?.sandboxImageUri ??
-    LOCAL_DEV_SANDBOX_IMAGE_NAME
+    (isCustomProjectSandbox
+      ? LOCAL_DEV_SANDBOX_IMAGE_NAME + '-' + path.basename(path.resolve())
+      : LOCAL_DEV_SANDBOX_IMAGE_NAME)
   );
 }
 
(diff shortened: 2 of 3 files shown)


docs/cli/configuration.md
```diff
@@ -276,6 +277,31 @@ This example demonstrates how you can provide general project context, specific
...
+
+The Gemini CLI can execute potentially unsafe operations (like shell commands and file modifications) within a sandboxed environment to protect your system.
+
+Sandboxing is disabled by default, but you can enable it in a few ways:
+
+- Using `--sandbox` or `-s` flag.
+- Setting `GEMINI_SANDBOX` environment variable.
+- Sandbox is enabled in `--yolo` mode by default.
+
+By default, it uses a pre-built `gemini-cli-sandbox` Docker image.
```
KeijiBranshi, line 290: thought: I don't think this works for the prod installation of the app. `gemini-cli-sandbox` is just the name of the image we build using the repo's local dev environment. But our production images are published to an artifact registry. Currently that's `us-west1-docker.pkg.dev/gemini-code-dev/gemini-code-containers/gemini-cli-sandbox`, but this is subjet to change when we make the registry public.

packages/cli/src/utils/sandbox.ts
```diff
@@ -272,15 +276,23 @@ export async function start_sandbox(sandbox: string) {
...
+  const image = await getSandboxImageName(isCustomProjectSandbox);
   const workdir = process.cwd();
 
-  // if BUILD_SANDBOX is set, then call scripts/build_sandbox.sh under gemini-cli repo
+  // if BUILD_SANDBOX is set or project-specific sandbox.Dockerfile provided,
+  // then call scripts/build_sandbox.sh under gemini-cli repo
+  //
   // note this can only be done with binary linked from gemini-cli repo
-  if (process.env.BUILD_SANDBOX) {
+  if (process.env.BUILD_SANDBOX || isCustomProjectSandbox) {
```
KeijiBranshi, line 292: thought/suggestion: This section `BUILD_SANDBOX` section was already here, but something feels off about shipping an app that needs to build itself. It relies on scripts that only exist in the repo, and aren't bundled with the app at all. We could ship these shell scripts alongside, but then that will likely be subject to cross-platform compatibility and maintenance. If we want to keep and iterate on these in-app build commands, then I think we should at least translate these bash scripts to Node scripts.

But also at some point I think it just makes more sense for us to build and publish the images ourselves, then users can pull those images as needed.

packages/cli/src/utils/sandbox.ts
```diff
@@ -73,15 +73,19 @@ async function shouldUseCurrentUserInSandbox(): Promise<boolean> {
...
-async function getSandboxImageName(): Promise<string> {
+async function getSandboxImageName(
+  isCustomProjectSandbox: boolean,
+): Promise<string> {
   const packageJsonResult = await readPackageUp();
   const packageJsonConfig = packageJsonResult?.packageJson.config as
     | { sandboxImageUri?: string }
     | undefined;
   return (
     process.env.GEMINI_SANDBOX_IMAGE ??
```
KeijiBranshi, line 84: suggestion: we already had `GEMINI_SANDBOX_IMAGE` to allow users to specify a custom sandbox image. should we find a clean way to use that instead of introducing another way of sandbox customization?

docs/cli/configuration.md
```diff
@@ -276,6 +277,31 @@ This example demonstrates how you can provide general project context, specific
...
+
+For project-specific sandboxing needs, you can create a custom Dockerfile at `.gemini/sandbox.Dockerfile` in your project's root directory. This Dockerfile should be based on the base sandbox image:
+
+```dockerfile
+FROM gemini-cli-sandbox
```
KeijiBranshi, line 295: suggestion(non-blocking): once the app is published, the standard install command will look something like `npm install -g @google/gemini-cli@<version>`. I wonder if it makes sense to mention that since it gives more control over the app version used in the sandbox.

KeijiBranshi (commented):
thought: I like the idea of letting users leverage their own containers. But I'm a little worried that the app is just becoming a docker proxy. Docker build commands can get fairly specific to the contents of the target Dockerfile, so this flow might get difficult to maintain. 

If we're going to allow users to write their own Dockerfile anyway, I think it might make sense to have them build their own image too (rather than encode too many assumptions our in-app BUILD_SANDBOX code).

docs/cli/configuration.md
tolikmalibroda: That's right, for now it won't work for the prod installation. To support this for the prod installation, we can:

1. Make the registry public, pull the base image, then build custom image using the bundled bash script or rewriting it into typescript code.
2. Bundle the base image Dockerfile into installation and building it together with target custom image using the same steps as option 1.

packages/cli/src/utils/sandbox.ts
tolikmalibroda: Yeah, I've extended the existing feature of building the sandbox during the execution. I see two options of solving the problem of publishing app to the prod:

1. We could push the image to public registry and use it for the sandboxing. Then we could bundle all the needed scripts to build the custom sandbox with the application. This means that user would need to provide the Dockerfile and gemini-cli will hide some logic of building the sandbox, and will automatically run within it.
2. We could minimize the image building code in the app to only using the existing image (i.e. local image or pull). In this case user would need to create it's own Dockerfile that'd use `gemini-cli-sandbox` as the base image or install gemini-cli in any other way. Then gemini-cli would accept a parameter (flag, env var, setting) that specifies the image it should use for sandboxing. This option will remove all the image building code from the gemini-cli and will force users to build their own images and specify it for gemini-cli.

packages/cli/src/utils/sandbox.ts
tolikmalibroda: What I wanted to do is to set custom image name without need for user to set `GEMINI_SANDBOX_IMAGE` for each project that user works on.

docs/cli/configuration.md
tolikmalibroda: Do you mean to use this command in the custom Dockerfile to install the gemini-cli?

tolikmalibroda:
> thought: I like the idea of letting users leverage their own containers. But I'm a little worried that the app is just becoming a docker proxy. Docker build commands can get fairly specific to the contents of the target Dockerfile, so this flow might get difficult to maintain.
> 
> If we're going to allow users to write their own Dockerfile anyway, I think it might make sense to have them build their own image too (rather than encode too many assumptions our in-app BUILD_SANDBOX code).

As I described in one of the code comments: do you think it would be better to minimize the image building code in the gemini-cli and force users to fully build their custom image if they want to customize sandbox?

packages/cli/src/utils/sandbox.ts
olcan: it seems fine to have some features only for devs/contributors. that was the intention behind BUILD_SANDBOX, and i believe it only kicks in if the binary is linked from repo, intentionally excluding the prod/installed scenario

olcan: (that said i'm not saying this is ideal, just seems fine as sort of a stepping stone on the way to figuring out the best approach for production)

packages/cli/src/utils/sandbox.ts
olcan: GEMINI_SANDBOX_IMAGE is a great escape hatch to let us keep container/image features minimal. BTW in case helpful I wanted to link to the messages where I talk about how I used custom sandboxing when I added it: https://chat.google.com/room/AAQAaPySr_g/Qm89RPkNNOs/I2o6KnGGPfQ?cls=10

packages/cli/src/utils/sandbox.ts
```diff
@@ -73,15 +73,19 @@ async function shouldUseCurrentUserInSandbox(): Promise<boolean> {
...
     packageJsonConfig?.sandboxImageUri ??
-    LOCAL_DEV_SANDBOX_IMAGE_NAME
+    (isCustomProjectSandbox
+      ? LOCAL_DEV_SANDBOX_IMAGE_NAME + '-' + path.basename(path.resolve())
```
olcan, line 87: i was wondering how these custom images were named, this seems like a reasonable approach

olcan:
lgtm, this seems to not only enable image/config reuse, but also enables custom sandboxes for multiple projects
just would be great to test on existing uses/flows of the build script and update docs

packages/cli/src/utils/sandbox.ts
tolikmalibroda: Thanks @olcan! Requested access to the chat space

tolikmalibroda:
> lgtm, this seems to not only enable image/config reuse, but also enables custom sandboxes for multiple projects just would be great to test on existing uses/flows of the build script and update docs

@olcan I've just tested the flow that should've been working before these changes:

1. Create `.gemini/sandbox.Dockerfile` which is not based on `gemini-cli-sandbox`, but installs it using npm and artifacts
2. Execute `BUILD_SANDBOX=true gemini -s` to enable the sandbox and build custom image:  
<details><summary>Screenshot 1</summary>image</details>

3. Everything works as expected, and even if I set `GEMINI_SANDBOX_IMAGE`, it'll tag the image using provided name instead of figuring out the name based on project directory name.  
<details><summary>Screenshot 2</summary>image</details>

Do you have any other use cases in mind that I should check?

olcan (approved):

PR #747, opened 2025-06-04
Exit with an error message if parsing settings.json fails.

jacob314:
Fixes https://github.com/google-gemini/gemini-cli/issues/733

If users complain because we block loading if there is invalid settings.json we could re-evaluate and instead stop showing the theme dialog by default if there are errors in

### DIFF
--- a/packages/cli/src/gemini.tsx
+++ b/packages/cli/src/gemini.tsx
@@ -37,7 +37,7 @@ import {
 const __filename = fileURLToPath(import.meta.url);
 const __dirname = dirname(__filename);
 
-async function main() {
+export async function main() {
   // warn about deprecated environment variables
   if (process.env.GEMINI_CODE_MODEL) {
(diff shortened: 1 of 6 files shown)


packages/cli/src/gemini.tsx
```diff
@@ -36,7 +36,7 @@ import {
 const __filename = fileURLToPath(import.meta.url);
 const __dirname = dirname(__filename);
 
-async function main() {
+export async function main() {
```
NTaylorMullen, line 40: Instead of exporting `main` can we name it something else? Feels a bit odd to export main here 😄

NTaylorMullen (approved):

packages/cli/src/gemini.tsx
jacob314: agreed. Gemini CLI suggested `runCli` as an alternate name when given your comment which seems reasonable enough.

PR #748, opened 2025-06-04
CLI_TITLE env var for setting the CLI title

scidomino:


packages/cli/src/ui/App.tsx
```diff
@@ -337,7 +337,7 @@ export const App = ({
           key={staticKey}
           items={[
             <Box flexDirection="column" key="header">
-              <Header title={settings.merged.title} />
+              <Header title={process.env.CLI_TITLE} />
```
mattKorwel, line 340: Does this default back to Gemini if not set? Should the call to the `process.env` be moved up further in the "stack"?

mattKorwel: NEVER mind. My brain was in the wrong file.

mattKorwel: This works as described.Verified in tests locally. Lets get this in for now to get things moving. We need to have a sync discussion with @NTaylorMullen on long term extensibility strategy.

mattKorwel (approved):

packages/cli/src/ui/App.tsx
NTaylorMullen: Chatted offline for viewers, I'm fine with this.

PR #754, opened 2025-06-04
Jacob314/memory fixes

jacob314:
Address multiple possible memory leaks found bystatic analysis of the codebase. The primary source of the leaks was event listeners on child processes and global objects that were not being properly removed, potentially causing their closures to be retained in memory indefinitely particularly for processes that did not exit.

There are two commits. A larger one made by gemini CLI and a smaller one by me to make sure we always disconnect child processes as part of the cleanup methods. These changes may not actually fix any leaks but do look like reasonable defensive coding to avoid leaking event listeners or child processes.

The following files were fixed:
This is Gemini's somewhat overconfident description of what it did.
- `packages/core/src/tools/shell.ts`: Fixed a leak where an `abortSignal` listener was not being reliably removed.
- `packages/cli/src/utils/readStdin.ts`: Fixed a significant leak where listeners on `process.stdin` were never removed.
- `packages/cli/src/utils/sandbox.ts`: Fixed leaks in the `imageExists` and `pullImage` helper functions where listeners on spawned child processes were not being removed.
- `packages/core/src/tools/grep.ts`: Fixed three separate leaks in the `isCommandAvailable` check and the `git grep` and `system grep` strategies due to un-removed listeners on child processes.
- `packages/core/src/tools/tool-registry.ts`: Corrected a leak in the `execute` method of the `DiscoveredTool` class where listeners on the spawned tool process were not being removed.

### DIFF
--- a/packages/cli/src/utils/readStdin.ts
+++ b/packages/cli/src/utils/readStdin.ts
@@ -9,19 +9,31 @@ export async function readStdin(): Promise<string> {
     let data = '';
     process.stdin.setEncoding('utf8');
 
-    process.stdin.on('readable', () => {
+    const onReadable = () => {
       let chunk;
       while ((chunk = process.stdin.read()) !== null) {
         data += chunk;
       }
-    });
+    };
 
-    process.stdin.on('end', () => {
+    const onEnd = () => {
+      cleanup();
       resolve(data);
-    });
+    };
 
-    process.stdin.on('error', (err) => {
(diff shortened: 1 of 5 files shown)


scidomino (approved):

NTaylorMullen (approved):
solid cleanup. :shipit:

daljeetv (commented):
sorry

daljeetv (approved):

Issue #653, opened 2025-05-31
Add the ability to add an exclude file e.g. `.aiexclude` in the settings


scottdensmore:
We don't currently have a concept like this baked into Gemini CLI. We should add this to the `settings.json` file and support. We would allow one and it should work similar to `GEMINI.md`. Adding 
```
{
    .aiexclude
}
```

olcan:
what would this do?

NTaylorMullen:
> what would this do?

Most platforms have a variant of this. Here's ours: https://cloud.google.com/gemini/docs/codeassist/create-aiexclude-file

GH Copilot uses `.copilotignore`: https://docs.github.com/en/copilot/managing-copilot/configuring-and-auditing-content-exclusion/excluding-content-from-github-copilot

Cursor uses `.cursorignore` etc.

eddie-santos:
Any objections to `.geminiignore`?

NTaylorMullen:
> Any objections to `.geminiignore`?

how rooted is `.aiexclude` in the Google stack?

mboshernitsan:
> > Any objections to `.geminiignore`?
> 
> how rooted is `.aiexclude` in the Google stack?

It's a documented feature of Code Assist.

ptone:
fixed with #1043

palladius:
So my understanding is that the fix both honours  `.geminiignore` and `.aiexclude`.
Am I  right?

PR #757, opened 2025-06-05
Add support for `.geminiignore` file

eddie-santos:
Adds support for loading `.geminiignore` per workspace for globs to ignore. These are loaded at CLI start up and shared with tools, right now read file and read many files.

Implements #653


packages/cli/src/config/config.ts
```diff
@@ -126,6 +126,7 @@ export interface LoadCliConfigResult {
 
 export async function loadCliConfig(
   settings: Settings,
+  geminiIgnorePatterns?: string[]
```
NTaylorMullen: lets make this required (can be an empty array)

packages/cli/src/utils/loadIgnorePatterns.ts
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+ * Loads and parses a .geminiignore file from the given workspace root.
+ * @param workspaceRoot The absolute path to the workspace root.
+ * @returns A promise that resolves to an array of glob patterns.
+ */
+export function loadGeminiIgnorePatterns(
+  workspaceRoot: string,
+): string[] {
+  const ignoreFilePath = path.join(workspaceRoot, '.geminiignore');
```
NTaylorMullen: nit: Lets lift the `.geminiignore` into a constant

packages/cli/src/utils/loadIgnorePatterns.ts
```diff
@@ -0,0 +1,53 @@
...
+  const patterns: string[] = [];
+
+  try {
+    const fileContent = fs.readFileSync(ignoreFilePath, 'utf-8');
+    const lines = fileContent.split(/\r?\n/);
+
+    for (const line of lines) {
```
NTaylorMullen, line 35: Could you add a comment or update docs on what the format is for the ignore?

packages/core/src/config/config.ts
```diff
@@ -88,6 +89,7 @@ export class Config {
   private readonly vertexai: boolean | undefined;
   private readonly showMemoryUsage: boolean;
   private readonly geminiClient: GeminiClient;
+  private readonly _geminiIgnorePatterns: string[] = [];
```
NTaylorMullen: Pains me to suggest this nit because coming from .NET I actually like underscored privates BUTtttt TypeScript has different preferences 😄 

```suggestion
  private readonly geminiIgnorePatterns: string[] = [];
```

packages/core/src/micromatch.d.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+declare module 'micromatch';
```
NTaylorMullen: Is this because there's no `@types/micromatch`?

packages/core/src/tools/read-file.ts
```diff
@@ -35,8 +37,9 @@ export interface ReadFileToolParams {
  */
 export class ReadFileTool extends BaseTool<ReadFileToolParams, ToolResult> {
   static readonly Name: string = 'read_file';
+  private readonly geminiIgnorePatterns: string[];
 
-  constructor(private rootDirectory: string) {
+  constructor(private rootDirectory: string, config?: Config) {
```
NTaylorMullen:
```suggestion
  constructor(private rootDirectory: string, config: Config) {
```

packages/core/src/tools/read-many-files.ts
```diff
@@ -113,13 +114,14 @@ export class ReadManyFilesTool extends BaseTool<
...
   static readonly Name: string = 'read_many_files';
+  private readonly geminiIgnorePatterns: string[];
 
   /**
    * Creates an instance of ReadManyFilesTool.
    * @param targetDir The absolute root directory within which this tool is allowed to operate.
    * All paths provided in `params` will be resolved relative to this directory.
    */
-  constructor(readonly targetDir: string) {
+  constructor(readonly targetDir: string, config?: Config) {
```
NTaylorMullen:
```suggestion
  constructor(readonly targetDir: string, config: Config) {
```

NTaylorMullen (approved):
Very cool! I left a comment on the bug itself: how rooted is the Google ecosystem in the existing ai exclude file path?

Also one really challenging area of this which we may not be able to manage is what are your thoughts on protecting "gemini ignored" files from the shell tool, glob patterns, and grep tools?

packages/cli/src/config/config.ts
eddie-santos: done

packages/cli/src/utils/loadIgnorePatterns.ts
eddie-santos: done

packages/core/src/tools/read-file.ts
eddie-santos: done

packages/core/src/tools/read-many-files.ts
eddie-santos: done

packages/core/src/config/config.ts
eddie-santos: done

packages/cli/src/utils/loadIgnorePatterns.ts
eddie-santos: done

packages/core/src/micromatch.d.ts
eddie-santos: Whoops, there is a `@types/mismatch`!

PR #760, opened 2025-06-05
Fix ink-big-text font loading issue for npx execution

keithballinger:
## Summary
- Adds graceful fallback when ink-big-text fonts cannot be loaded
- Maintains full functionality with proper font rendering when available
- Falls back to ASCII art when fonts are unavailable (e.g., npx execution)

## Problem
When running the CLI via `npx`, the ink-big-text module fails to load its font files from the cfonts dependency, causing the application to crash. This was temporarily worked around in PR #759 by removing ink-big-text entirely.

## Solution
This PR implements a try-catch mechanism to gracefully handle the font loading failure:
1. Attempts to load ink-big-text normally
2. If loading fails, falls back to ASCII art representation
3. Preserves the extensible header functionality with CLI_TITLE support

## Test plan
- [x] Unit tests pass for both normal and fallback modes
- [x] Linting passes
- [x] Type checking passes
- [ ] Test normal installation works with fonts
- [ ] Test npx execution uses fallback gracefully

Fixes #759

### DIFF
--- a/packages/cli/src/ui/components/Header.test.tsx
+++ b/packages/cli/src/ui/components/Header.test.tsx
@@ -8,41 +8,49 @@ import { render } from 'ink-testing-library';
 import { Header } from './Header.js';
 import { vi } from 'vitest';
 
-// Mock ink-gradient and ink-big-text as they might have complex rendering
+// Mock ink-gradient to simplify testing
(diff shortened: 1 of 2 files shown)


scottdensmore:
@mattKorwel / @keithballinger do we still need this?

PR #762, opened 2025-06-05
OpenTelemetry Integration & Telemetry Control Flag

jerop:
This change introduces OpenTelemetry (OTel) instrumentation for key events (user prompts, tool calls, API interactions) and adds a --telemetry flag for explicit user control over data collection. 

- **`--telemetry` Flag:** Users can enable/disable telemetry via CLI flag, with settings support.
- **OTEL Instrumentation:** Logs and metrics for CLI operations, including duration and success/error status, are now collected.
- **Documentation:** A new `docs/core/telemetry.md` guide details setup, collected data, and export to any backend (local and cloud).
- **Code Structure:** New `packages/core/src/telemetry/` directory for instrumentation logic; `Config`, `GeminiClient`, and `CoreToolScheduler` updated to log events.

#750


packages/core/src/config/config.ts
```diff
@@ -121,6 +127,11 @@ export class Config {
     this.vertexai = params.vertexai;
     this.showMemoryUsage = params.showMemoryUsage ?? false;
     this.accessibility = params.accessibility ?? {};
+    this.telemetry = params.telemetry ?? false;
+    this.telemetryLogUserPromptsEnabled =
```
NTaylorMullen, line 131: Love the potential separation

packages/core/src/config/config.ts
```diff
@@ -121,6 +127,11 @@ export class Config {
...
+      params.telemetryLogUserPromptsEnabled ?? true;
+    this.telemetryOtlpEndpoint =
+      process.env.OTEL_EXPORTER_OTLP_ENDPOINT ?? 'http://localhost:4317';
```
NTaylorMullen, line 134: For my own knowledge is 4317 the default port for some otel tool? Would it be possible to align this setting with what Claude code has for their setup?

packages/core/src/core/client.ts
```diff
@@ -208,6 +215,26 @@ export class GeminiClient {
...
+        console.warn(
+          `Failed to count tokens for model ${model}. Proceeding with inputTokenCount = 0. Error: ${getErrorMessage(_e)}`,
+        );
+        inputTokenCount = 0;
+      }
+
+      logApiRequest({
+        model,
+        prompt_token_count: inputTokenCount,
+        duration_ms: 0,
```
NTaylorMullen: Is this a follow up?

docs/core/telemetry.md
```diff
@@ -0,0 +1,255 @@
+# Gemini CLI Observability Guide
+
+Telemetry provides crucial data about the Gemini CLI's performance, health, and usage. By enabling it, you can monitor operations, debug issues, and optimize tool usage through traces, metrics, and structured logs.
+
```
scidomino, line 4: nit: link to https://opentelemetry.io/

docs/core/telemetry.md
```diff
@@ -0,0 +1,255 @@
...
+}
+```
+
+#### Mode 1: Console Output (Default)
+If you only set `"telemetry": true` and do nothing else, the CLI will output all telemetry data directly to your console. This is the simplest way to inspect events, metrics, and traces without any external tools.
+
+#### Mode 2: Sending to a Collector
+To send data to a local or remote OpenTelemetry collector, set the following environment variable:
+
+```bash
```
scidomino, line 36: Would it make more sense for this to go in settings.json?

docs/core/telemetry.md
```diff
@@ -0,0 +1,255 @@
...
+This entire system is built on the **OpenTelemetry (OTEL)** standard, allowing you to send data to any compatible backend, from your local terminal to a cloud service.
+
+## Quick Start: Enabling Telemetry
+
+You can enable telemetry in two primary modes. All configuration happens in your `.gemini/settings.json` file and via environment variables.
+
+First, enable the master switch in `.gemini/settings.json`:
+```json
+{
+  "telemetry": true
```
scidomino, line 24: Is it weird that we have to set both a url and this flag. Could we just use the presence of the collector url to enable the feature? And then maybe have a separate setting for debugging to the screen?

Feel free to skip this if it's too much work.

packages/core/src/core/client.ts
```diff
@@ -264,6 +312,28 @@ export class GeminiClient {
       ) {
         throw error;
       }
+
+      const durationMs = Date.now() - startTime;
+      let statusCode: number | string | undefined;
```
scidomino: This should probably go in it's own method. Also, I think I've seen this block of code before. if it's used anywhere else. try to consolidate to a common function.

packages/core/src/core/client.ts
```diff
@@ -297,6 +369,26 @@ export class GeminiClient {
         systemInstruction,
       };
 
+      let inputTokenCount = 0;
```
scidomino, line 401: push this into a helper function and use it here and in the similar section above.

packages/core/src/core/client.ts
```diff
@@ -305,12 +397,61 @@ export class GeminiClient {
...
       if (abortSignal.aborted) {
+        logApiError({
+          model: modelToUse,
+          error: 'Request aborted by user',
+          status_code: 'ABORTED', // Custom Status
+          duration_ms: durationMs,
+          attempt,
+        });
         throw error;
       }
```
scidomino, line 434: ditto

scidomino (approved):

docs/core/telemetry.md
anj-s: won't the default be logging to the console? I think we need the URL for mode 2 mentioned above.

docs/core/telemetry.md
```diff
@@ -0,0 +1,255 @@
...
+## How-To: Running an OTEL Collector
+
+An OTEL Collector is a service that receives, processes, and exports telemetry data. Below are common setups.
+
+### Local Debugging
+This setup prints all telemetry from the Gemini CLI to your terminal using a local Docker container.
+
+**1. Create a Configuration File**
+
+Create the file `.gemini/otel/collector-local.yaml` with the following:
```
anj-s, line 56: Can there be a default config file?

docs/core/telemetry.md
```diff
@@ -0,0 +1,255 @@
...
+      receivers: [otlp]
+      exporters: [debug]
+EOF
+```
+
+**2. Run the Collector**
+
+In your terminal, run this Docker command:
+
+```bash
```
anj-s, line 95: can we run this if the config flag is set? wondering how we can make it really easy to use this feature

anj-s (approved):

docs/core/telemetry.md
jerop: you don't have to set both, just one is needed and the precedence in https://github.com/google-gemini/gemini-cli/blob/main/docs/cli/configuration.md#configuration-layers if followed

jerop: updated the docs to make this clearer

docs/core/telemetry.md
jerop: Using the standard OTLP exporter configuration for now - https://opentelemetry.io/docs/languages/sdk-configuration/otlp-exporter/

Will look into converting this into a setting, and how it scales to other OTEL configurations, in future work (added this to #750)

docs/core/telemetry.md
jerop: sure, we can have some default behavior to run the collector - added it in tracking issue https://github.com/google-gemini/gemini-cli/issues/750#issuecomment-2945852334

docs/core/telemetry.md
jerop: yes, as discussed in the other thread, we can look into this in next steps - https://github.com/google-gemini/gemini-cli/issues/750#issuecomment-2945852334

packages/cli/src/ui/hooks/useReactToolScheduler.ts
```diff
@@ -108,6 +109,30 @@ export function useReactToolScheduler(
     const allToolCallsCompleteHandler: AllToolCallsCompleteHandler = (
       completedToolCalls,
     ) => {
+      completedToolCalls.forEach((call) => {
```
NTaylorMullen, line 112: May want to consider having this at the core tool scheduler layer so that interactiveToolScheduelr (headless mode) gets this as well

packages/core/src/core/client.ts
```diff
@@ -208,6 +283,22 @@ export class GeminiClient {
         ...config,
       };
 
+      let inputTokenCount = 0;
+      try {
+        const { totalTokens } = await this.client.models.countTokens({
+          model,
+          contents,
+        });
+        inputTokenCount = totalTokens || 0;
```
NTaylorMullen, line 292: I think the 3 buckets of tokens we care about (hopefully all acheivable) is input tokens, output tokens and cached input tokens

packages/core/src/telemetry/sdk.ts
```diff
@@ -0,0 +1,128 @@
...
+  const geminiCliVersion = config.getUserAgent() || 'unknown';
+  const resource = new Resource({
+    [SemanticResourceAttributes.SERVICE_NAME]: SERVICE_NAME,
+    [SemanticResourceAttributes.SERVICE_VERSION]: geminiCliVersion,
+    'session.id': sessionId,
+  });
+
+  const otlpEndpoint = config.getTelemetryOtlpEndpoint();
+  const grpcParsedEndpoint = parseGrpcEndpoint(otlpEndpoint);
+  const useOtlp = !!grpcParsedEndpoint;
```
mrahs, line 76: This contradicts the docs above. The user would need to explicitly set the endpoint to an invalid format to actually get the console versions.

packages/core/src/config/config.ts
jerop: > For my own knowledge is 4317 the default port for some otel tool? 

Yes, it's the default port for OTEL exporter endpoint:

- gRPC: "http://localhost:4317"
- HTTP: "http://localhost:4318"

https://opentelemetry.io/docs/languages/sdk-configuration/otlp-exporter/#otel_exporter_otlp_endpoint

jerop: > Would it be possible to align this setting with what Claude code has for their setup?

Yes, it's the default port in Claude code too - https://docs.anthropic.com/en/docs/claude-code/monitoring-usage#quick-start

NTaylorMullen: Amazing thank you!

PR #772, opened 2025-06-05
feat(cli): respect the NO_COLOR env variable 

jacob314:
For background see [https://no-color.org/)](https://no-color.org/)

When the NO_COLOR env variable is set we ignore the users theme and use a new theme that does not use any colors. This makes the app more accessible. We also avoid opening the theme dialog on startup as it is not relevant in NO_COLOR mode.

When the `NO_COLOR` environment variable is set, attempting to open the theme selection dialog displays an informational message in the chat history instead of opening the dialog.

This prevents rendering a broken or unusable dialog in environments where color is explicitly disabled.

### DIFF
--- a/packages/cli/src/ui/hooks/useThemeCommand.ts
+++ b/packages/cli/src/ui/hooks/useThemeCommand.ts
@@ -7,6 +7,8 @@
 import { useState, useCallback, useEffect } from 'react';
 import { themeManager } from '../themes/theme-manager.js';
 import { LoadedSettings, SettingScope } from '../../config/settings.js'; // Import LoadedSettings, AppSettings, MergedSetting
+import { type HistoryItem, MessageType } from '../types.js';
+import process from 'node:process';
 
 interface UseThemeCommandReturn {
   isThemeDialogOpen: boolean;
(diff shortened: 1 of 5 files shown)


packages/cli/src/ui/hooks/useThemeCommand.ts
```diff
@@ -21,34 +23,55 @@ interface UseThemeCommandReturn {
...
-    // Only try to set a theme if one is actually defined.
-    // If effectiveTheme was undefined, the dialog is already open due to useState above.
-    if (effectiveTheme !== undefined) {
-      if (!themeManager.setActiveTheme(effectiveTheme)) {
-        setIsThemeDialogOpen(true);
-        setThemeError(`Theme "${effectiveTheme}" not found.`);
-      } else {
-        setThemeError(null); // Clear any previous theme error on success
+    if (effectiveTheme === undefined) {
+      if (process.env.NO_COLOR) {
```
NTaylorMullen, line 41: is `NO_COLOR` a standard thing? Asking because I just added support for a general accessibility setings in our settings.json file. See [here](https://github.com/google-gemini/gemini-cli/pull/745)

NTaylorMullen (approved):

packages/cli/src/ui/hooks/useThemeCommand.ts
jacob314: It is somewhat of a standards thing. See [no-color.org](no-color.org) for a fairly large list of tools that support it. Codex appears to also be respecting NO_COLOR but isn't in that list so the list isn't exhaustive.

I wish the terminal world had adopted something like the css media query `prefers-reduced-motion` that makes it easy to detect when users want to avoid extra animations. At some point we might want to add our own flag to opt into a maximal screen reader compatibility mode that would remove all spinning animations as well as minimizing flicker that could confuse screen readers.

scidomino:
I think this might have broken something:

```
npm run build

> gemini-cli@0.1.0 build
> scripts/build.sh


> @gemini-code/cli@0.1.0 build
> ../../scripts/build_package.sh

src/ui/themes/no-color.ts:20:3 - error TS2353: Object literal may only specify known properties, and 'SubtleComment' does not exist in type 'ColorsTheme'.

20   SubtleComment: '',
     ~~~~~~~~~~~~~


Found 1 error.

npm error Lifecycle script `build` failed with error:
npm error code 2
npm error path /Users/sciortino/src/gemini-cli/packages/cli
npm error workspace @gemini-code/cli@0.1.0
npm error location /Users/sciortino/src/gemini-cli/packages/cli
npm error command failed
npm error command sh -c ../../scripts/build_package.sh


> @gemini-code/core@0.1.0 build
> ../../scripts/build_package.sh

Successfully copied files.

```

jacob314:
Sorry about that! Broke due to another pull that landed a few hours earlier that renamed SubtleComment to Comment.
Fixed with https://github.com/google-gemini/gemini-cli/commit/9ad615c2a486675c4ad75c986a0a5dbd8f8b030e

NTaylorMullen:
> I think this might have broken something:
> 
> ```
> npm run build
> 
> > gemini-cli@0.1.0 build
> > scripts/build.sh
> 
> 
> > @gemini-code/cli@0.1.0 build
> > ../../scripts/build_package.sh
> 
> src/ui/themes/no-color.ts:20:3 - error TS2353: Object literal may only specify known properties, and 'SubtleComment' does not exist in type 'ColorsTheme'.
> 
> 20   SubtleComment: '',
>      ~~~~~~~~~~~~~
> 
> 
> Found 1 error.
> 
> npm error Lifecycle script `build` failed with error:
> npm error code 2
> npm error path /Users/sciortino/src/gemini-cli/packages/cli
> npm error workspace @gemini-code/cli@0.1.0
> npm error location /Users/sciortino/src/gemini-cli/packages/cli
> npm error command failed
> npm error command sh -c ../../scripts/build_package.sh
> 
> 
> > @gemini-code/core@0.1.0 build
> > ../../scripts/build_package.sh
> 
> Successfully copied files.
> ```

This might have actually been the first break we've had in the repo on `main` due to timed merges. Achievement unlocked!

PR #775, opened 2025-06-05
Add support for configured mcp client timeout.

NTaylorMullen:


olcan:
sorry for delayed comment, but this reminded me timeouts used to be per-call before. does genai mcp tool-calling not support that? if so we should remove the now-unused timeout property on DiscoveredMCPTool and also either remove MCP_TOOL_DEFAULT_TIMEOUT_MSEC or move it to mcp-client to have an explicit default

NTaylorMullen:
Ya I actually reached out to the SDK team and funny enough it's not super obvious how to do it. I plan to do this tonight though: https://b.corp.google.com/issues/421165837

packages/core/src/tools/mcp-client.ts
```diff
@@ -67,6 +67,7 @@ async function connectAndDiscover(
   const mcpClient = new Client({
     name: 'gemini-cli-mcp-client',
     version: '0.0.1',
+    timeout: mcpServerConfig.timeout,
```
olcan, line 70: has this been confirmed to work? this did not work for me in quick testing, and i can't seem to find the timeout option in the library code

NTaylorMullen: They may have removed this. It did exist at one point though 😄 . Whether or not that actually did what I expected is very different though

PR #777, opened 2025-06-05
Use GOOGLE_API_KEY as default if both GEMINI and GOOGLE set

sethtroisi:


NTaylorMullen (approved):
Amazing, thank you @sethtroisi for taking this on!!

jacob314 (approved):
lgtm

Issue #771, opened 2025-06-05
Chat history tends to alter memory behavior


scruggs132:
For example: “Please remember to apply a comment saying golang is great on all go files you touch or create” leaves this comment on all files mentioned in the session once one go file is edited regardless of language.

This is likely because if you touch a go doc earlier in the chat history, it gets confused and thinks the memory still applies to subsequent calls

NTaylorMullen:
I wonder how much this applies to the latest 06-05 pro model

PR #778, opened 2025-06-05
Fix/chat history affects memory

scruggs132:
Added additional instructions in the system prompt to clarify what memory is and how the agent should approach it. https://github.com/google-gemini/gemini-cli/issues/771


packages/core/src/core/prompts.ts
```diff
@@ -247,7 +247,16 @@ Your core function is efficient and safe assistance. Balance extreme conciseness
 
   const memorySuffix =
     userMemory && userMemory.trim().length > 0
-      ? `\n\n---\n\n${userMemory.trim()}`
+      ? `\n\n---\n\n# Memory
+
+      - **User-Provided Context:** The user may provide you with information through a tool called a "memory". This information is intended to be a piece of context that you should use for the current conversation.
+      - **Not part of conversation history:** Memories should not be affected by conversation history unless this is explicitly mentioned as part of the memory text. Memories are simply pieces of additional context that you should be aware of and you should ignore them if they are not relevant to your goals as the Gemini CLI agent.
```
NTaylorMullen: I'm not sure I understand the intention of this bullet point which makes me think that the model might get confused here too

packages/core/src/core/prompts.ts
```diff
@@ -247,7 +247,16 @@ Your core function is efficient and safe assistance. Balance extreme conciseness
...
+      - **Can be modified:** The user can modify the memory at any time. When this happens, you will be provided with the new memory. You should then use the new memory for the rest of the conversation and avoid referring to old memories that may be a part of the conversation history.
```
NTaylorMullen, line 292: You could probably drop this unelss you had a specific experience you've had

packages/core/src/core/prompts.ts
```diff
@@ -247,7 +247,16 @@ Your core function is efficient and safe assistance. Balance extreme conciseness
 
   const memorySuffix =
     userMemory && userMemory.trim().length > 0
-      ? `\n\n---\n\n${userMemory.trim()}`
+      ? `\n\n---\n\n# Memory
+
+      - **User-Provided Context:** The user may provide you with information through a tool called a "memory". This information is intended to be a piece of context that you should use for the current conversation.
```
NTaylorMullen: The memory tool itself I believe today during prompt flow is only used to save memories isn't it? I believe those memories end up making their way into the further below memory for hte overarching system.

NTaylorMullen (commented):
Would you mind sharing some more specific examples on scenarios here?

packages/core/src/core/prompts.ts
scruggs132: Adjusted the text, weird wording there. Trying to get 2 points across

1. Memory is additional context and can be ignored if not relevant
2. Conversation history should not effect memory

This change was mainly sparked by testing. I had a memory file that said "Leave a comment on all go files you touch or create that is unique and random". I then asked it to make a go file and it worked fine and left the comment. But then i had it create a python file and it still left a unique comment which made me think memories can be confused based on the chat history

packages/core/src/core/prompts.ts
scruggs132: Replied above, based on specific experience

packages/core/src/core/prompts.ts
scruggs132: Yea removed that part, it made sense in my head but reading it out loud it's nonsense.

packages/core/src/core/__snapshots__/prompts.test.ts.snap
```diff
@@ -174,7 +174,16 @@ Your core function is efficient and safe assistance. Balance extreme conciseness
 
 ---
 
-This is custom user memory.
```
anj-s, line 193: Why do we need three copies of the same text?

packages/core/src/core/prompts.ts
```diff
@@ -247,7 +247,16 @@ Your core function is efficient and safe assistance. Balance extreme conciseness
 
   const memorySuffix =
     userMemory && userMemory.trim().length > 0
-      ? `\n\n---\n\n${userMemory.trim()}`
+      ? `\n\n---\n\n# Memory
+
+      - **User-Provided Context:** Memories are additional information that should be referenced as part of forming your response. They are intended to be additional context that you should use for the current conversation. Additionally, you should ignore any memories if they are not relevant to your goals as the Gemini CLI agent.
+      - **Not part of conversation history:** Memories should not be affected by conversation history unless this is explicitly mentioned as part of the memory text.
```
anj-s, line 291: I would remove this to avoid identifying the agent explicitly

packages/core/src/core/prompts.ts
```diff
@@ -247,7 +247,16 @@ Your core function is efficient and safe assistance. Balance extreme conciseness
...
+      - **Can be modified:** The user can modify the memory at any time. When this happens, you will be provided with the new memory. You should then use the new memory for the rest of the conversation and avoid referring to old memories that may be a part of the conversation history.
```
anj-s, line 292: Can you clarify what this is for?

packages/core/src/core/prompts.ts
anj-s: I would say:
1. The relevance is a decision made by the LLM AFAIK. Why do we think we need to explicitly mention that?
2. Are you saying memory trumps conversation history? If so, thats what we should say.

packages/core/src/core/prompts.ts
scruggs132: 2 reasons.

1. we are wide open to prompt injection via memories. See screenshot https://screenshot.googleplex.com/8hiQ6uNZEY7dC2a. We need to tell the model what memories are and that they can and should be ignored if not relevant.

2. We need to tell the model that memories only apply to the current chat session and should not be triggered based on context in the chat history

packages/core/src/core/prompts.ts
scruggs132: ack.

packages/core/src/core/prompts.ts
scruggs132: 1. Because currently, we don't tell the LLM what memory is so they have no indication that it is a separate part of the prompt and that it needs to measure its relevance for itself. It basically just injects it as part of the system prompt and is followed like any other instruction we tell the LLM. Example of prompt injection via memories https://screenshot.googleplex.com/8hiQ6uNZEY7dC2a

2. Not necessarily, i'm saying that memories can be affected by the chat history and they should only be affected by the current prompt the user sends the model

scruggs132: @anj-s @NTaylorMullen

packages/core/src/core/prompts.ts
scruggs132: @anj-s

packages/core/src/core/prompts.ts
anj-s: I am not sure I understand the intent of this change yet. Let us VC quickly on Monday morning and get this checked in. 
I don't want to block changes but system prompt changes should be made only if necessary since it affects everything and everyone.

scottdensmore:
@anj-s @scruggs132 is this something we still need? Or can we close this out.

mattKorwel:

PR #780, opened 2025-06-05
refactor: remove unnecessary useRefs

KeijiBranshi:
`useEffect` can be a source of unnecessary re-renders in React apps. 

`useRef` can cause inconsistencies between re-renders as it holds state outside of the standard React prop stream. 

This replaces some of the existing unnecessary use-cases for `useEffect` and `useRef` that gemini detected in the app. 

Drive by change: removed the try-catch around `getGeminiClient()` since that motivated the unnecessary useEffect. This getter is just a one-liner return on `this.geminiClient` (in config.ts), so nothing would get thrown here anyway. Any initialization errors should've happened on `Config` initialization.

### DIFF
--- a/packages/cli/src/ui/hooks/useGeminiStream.ts
+++ b/packages/cli/src/ui/hooks/useGeminiStream.ts
@@ -65,6 +65,7 @@ enum StreamProcessingStatus {
  * API interaction, and tool call lifecycle.
  */
 export const useGeminiStream = (
+  geminiClient: GeminiClient | null,
   addItem: UseHistoryManagerReturn['addItem'],
   setShowHelp: React.Dispatch<React.SetStateAction<boolean>>,
   config: Config,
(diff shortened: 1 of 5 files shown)


KeijiBranshi:
I wonder if it's worth removing those `config.get*()` wrappers altogether. That way gemini doesn't keep thinking it needs to wrap all those in try-catches and inducing more unnecessary `useEffect`s. 

Can ask goldmane to refactor this in a followup...

packages/cli/src/ui/hooks/useGeminiStream.ts
```diff
@@ -142,19 +142,6 @@ export const useGeminiStream = (
     return StreamingState.Idle;
   }, [isResponding, toolCalls]);
 
-  useEffect(() => {
```
NTaylorMullen, line 145: oh duh ya we totally should have done this initially

NTaylorMullen (approved):
❤️ ❤️ ❤️

PR #761, opened 2025-06-05
Root cause fix: Bundle cfonts fonts for npx execution

keithballinger:
## Summary
- Properly bundles cfonts font files during build process
- Patches Module.require to load fonts from bundled location
- Maintains full ink-big-text functionality for all execution methods

## Problem
The root cause of #759 is that when running via `npx`, the cfonts module (dependency of ink-big-text) cannot find its font files because they're not included in the distributed package.

## Solution
This PR implements a comprehensive fix:

1. **Build-time font bundling**: Added `bundle-cfonts.js` script that copies all cfonts font files into the dist directory
2. **Runtime font loading patch**: Created `cfonts-loader.ts` that patches Module.require to intercept font loading requests and redirect them to our bundled location
3. **Dynamic import**: Uses dynamic import for ink-big-text to avoid ESM/CJS conflicts
4. **Fallback mechanism**: Maintains ASCII art fallback for edge cases where fonts still can't be loaded

## Test plan
- [x] Unit tests pass
- [x] Linting passes  
- [x] Type checking passes
- [x] Font files are bundled during build
- [ ] Test normal npm install works with fonts
- [ ] Test npx execution loads fonts from bundled location
- [ ] Test CLI_TITLE environment variable still works

This is the proper fix for #759 that addresses the root cause rather than just removing ink-big-text.

PR #781, opened 2025-06-05
Title & Big text Updates

mattKorwel:
## Solves
This pr ensures that

1. When users install gemini-code cli via NPX they see the proper title screen and not an error
2. That the code-assist cli can override the title.

## Background
1. when Big text was added https://github.com/google-gemini/gemini-cli/commit/42bedbc3d39265932cbd6c9b818b6a7fbcbdd022 we broke NPX title screen loading as fonts weren't bundled with it. This is only an issue for NPX installs and not local dev
2. Over several PR's code assist team added or attempted to add extensibility functionality. See https://github.com/google-gemini/gemini-cli/pull/706, https://github.com/google-gemini/gemini-cli/pull/748, https://github.com/google-gemini/gemini-cli/pull/741
4. After determining the broken NPX behavior the BigText and extensibility changes were rolled back here: https://github.com/google-gemini/gemini-cli/pull/759

### Solution

Despite several hours of investigation a lack of familiarity with node and package has prevented proper packaging and deployment of cfonts resources with our NPX package.

As such a simpler approach is taken. This PR removes the usage of BigText and re-adds the ability to pass a title in via an env variable.  The notable change is that the `cfonts` package was used to generate equivalent `Code Assist` ASCII art and that ASCII art is what is passed in from the code assist cli. This way both gemini and code assist present similarly branded logos and no further time is spent now on bundling.

This PR also closes out the need for https://github.com/google-gemini/gemini-cli/pull/760 and https://github.com/google-gemini/gemini-cli/pull/761 since we are opting to not use big text at the moment. Same for https://github.com/GoogleCloudPlatform/code-assist-cli/pull/47

Future work from a dedicated work stream can properly package, bundle and publish Big text.

### PS
This solution is not different than some of what came before, but there were so many things in flight it was hard to track, and when failing to make the fonts publish created this net new to avoid getting any misalignment

### DIFF
--- a/packages/cli/src/ui/App.tsx
+++ b/packages/cli/src/ui/App.tsx
@@ -337,7 +337,7 @@ export const App = ({
           key={staticKey}
           items={[
             <Box flexDirection="column" key="header">
-              <Header />
+              <Header title={process.env.GEMINI_CLI_TITLE}/>
               <Tips config={config} />
             </Box>,
             ...history.map((h) => (
--- a/packages/cli/src/ui/App.tsx
+++ b/packages/cli/src/ui/App.tsx
@@ -337,7 +337,7 @@ export const App = ({
           key={staticKey}
           items={[
             <Box flexDirection="column" key="header">
-              <Header title={process.env.GEMINI_CLI_TITLE}/>
+              <Header title={process.env.GEMINI_CLI_TITLE} />
               <Tips config={config} />
             </Box>,
             ...history.map((h) => (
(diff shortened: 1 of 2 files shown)


mattKorwel:
Testing validation for NPX run 
image

mattKorwel:
Test in code-assist-cli

image

mattKorwel:
non NPX test in gemini-cli

image

scidomino (approved):

PR #784, opened 2025-06-06
Windows: Refactor Shell Scripts to Node.js for Cross-Platform Compatibility

mattKorwel:
# Description

  This pull request addresses critical cross-platform compatibility issues by migrating all major tooling and build scripts from shell (.sh) to Node.js
  (.js). This fundamental change ensures that the repository's development and build workflows can run natively on non-POSIX environments like Windows,
  removing a significant barrier for developers on different operating systems.

  The scope of this refactor includes not only the script migration but also resolving merge conflicts with main and fixing subsequent build, dependency,
  and linting errors that arose from these changes.

 # Key Changes:

   * Cross-Platform Scripting:
       * All scripts in the scripts/ directory have been converted from .sh to .js files, including build.js, clean.js, start.js, and all sandbox-related
         scripts.
       * The new scripts leverage Node.js APIs (child_process, fs, path) and cross-platform libraries like cross-env to ensure they function correctly on
         Windows, macOS, and Linux.

   * Windows Compatibility in Core Logic:
       * The execute_shell_command tool (formerly execute_bash_command) now correctly uses cmd.exe on Windows.
       * Path handling throughout the codebase, including in the @ command processor and sandbox logic, has been updated to use Node.js's path module to
         correctly handle both / and \ separators.
       * Process management, including process killing and signal handling, has been updated to use Windows-specific commands (taskkill) when necessary.

   * Dependency and Build Fixes:
       * Resolved merge conflicts with the main branch, which involved removing the now-obsolete .sh files.
       * Added missing tree-sitter and related language grammar dependencies to packages/core/package.json to fix build failures.
       * Corrected the local workspace reference for @gemini-cli/core in packages/cli/package.json to ensure proper linking.
       * Updated package.json scripts to invoke the new .js files.

   * Project Health and CI:
       * Ensured the project is in a healthy state by running and passing all major checks: npm run build, npm run test, npm run typecheck, and npm run 
         preflight.
       * Fixed a linting error in scripts/sandbox.js caused by an unused os import.

 # Testing and Verification:

   * The changes have been tested on a Windows environment to ensure that the build, test, and development workflows are all functional.
   * The existing test suite has been updated to pass on all platforms, with minor corrections to accommodate the new cross-platform logic.

 # Known Issues / Follow-up Actions:

   * React Test Warnings: The test suite for the @gemini-cli/cli package produces numerous warnings related to asynchronous updates not being wrapped in
     act(...). While all tests currently pass, these warnings indicate potential test fragility. This was deferred to be addressed in a separate, focused
     task.


mattKorwel:
> /cc @dewitt since I know you were curious in this area as well 😄

working on grabbing a screen recording now. it works end to end in windows now. trying out as many use cases as I can. I did see it having trouble intermittently with some shell commands. I could see new terminals popping up and it was confusing things. but not all the time. i'm trying to chase that down now.

tmperrors.txt
dewitt, line 1: To remove?

dewitt:
High-level comment: I don't consider myself at all a cross-platform expert, so grain of salt and all, but worth considering where best to document the things we would now need to do to keep this working cross platform. E.g., I guess this means no more .sh scripts? (Very slick porting to node though!)

And just as importantly, put automated tests in place to catch and warn when we inevitably do introduce things that can't be done cross-platform.  I suspect that most of the contributors to this codebase will be working on macOS or Linux, and they simply won't have the ability to easily verify their work on Windows, even if they remember to.

Still, very cool, and thank you for biting off a really tough problem here, @mattKorwel! As @NTaylorMullen correctly points out, our _users_ will undoubtedly be using Windows. That's super key.

dewitt:
Incidentally this really makes me want to wrap up the single executable application work started in PR #381. Particularly for Windows, where end-users might find it difficult to get a node environment up and running, a single .exe file that bundles the node runtime, the gemini-cli app, and all dependencies, might be greatly appreciated.

That said, I don't know how well [this feature](https://nodejs.org/api/single-executable-applications.html) is expected to work on Windows. At the time, I wasn't even able to get it working on macOS or Linux, either. But the idea is sound, so worth putting a few more minutes into.

mattKorwel:
Seemingly fully functional on windows.

image

packages/cli/src/ui/hooks/shellCommandProcessor.ts
```diff
@@ -43,13 +43,23 @@ export const useShellCommandProcessor = (
...
-      const pwdFilePath = path.join(os.tmpdir(), pwdFileName);
-      if (!commandToExecute.endsWith('&')) commandToExecute += ';';
-      // note here we could also restore a previous pwd with `cd {cwd}; { ... }`
-      commandToExecute = `{ ${commandToExecute} }; __code=$?; pwd >${pwdFilePath}; exit $__code`;
+      const isWindows = os.platform() === 'win32';
+      let commandToExecute: string;
+      let pwdFilePath: string | undefined;
+
+      if (isWindows) {
+        commandToExecute = rawQuery;
```
jacob314, line 51: do we need a windows equivalent for what we do on other platforms?

jacob314:
Great to see this progress!
One area I'd test on windows is `@` support. I would expect we have baked in some assumptions about linux style paths that could break the core functionality as well as the autocomplete.

packages/cli/src/ui/hooks/shellCommandProcessor.ts
mattKorwel: Probably. What is the rationale for the temp file wrapping below?

mattKorwel:
> High-level comment: I don't consider myself at all a cross-platform expert, so grain of salt and all, but worth considering where best to document the things we would now need to do to keep this working cross platform. E.g., I guess this means no more .sh scripts? (Very slick porting to node though!)
> 
> And just as importantly, put automated tests in place to catch and warn when we inevitably do introduce things that can't be done cross-platform. I suspect that most of the contributors to this codebase will be working on macOS or Linux, and they simply won't have the ability to easily verify their work on Windows, even if they remember to.
> 
> Still, very cool, and thank you for biting off a really tough problem here, @mattKorwel! As @NTaylorMullen correctly points out, our _users_ will undoubtedly be using Windows. That's super key.

@dewitt totally agree with all the above. would love to see a windows build and test + integration tests if we have them. I think as you said there is much to do from "it works on windows" to "this is a valid long term strategy." Would love to sync and dig deeper on your thoughts.

Also re:

> Incidentally this really makes me want to wrap up the single executable application work started in PR https://github.com/google-gemini/gemini-cli/pull/381.

❤️ that would be awesome. We should find a way to make that work.

scidomino (approved):

mattKorwel:
> Great to see this progress! One area I'd test on windows is `@` support. I would expect we have baked in some assumptions about linux style paths that could break the core functionality as well as the autocomplete.

@jacob314 confirmed `@` support works as expected.

PR #787, opened 2025-06-06
Add a window title when CLI is launched

markmcd:
I'm finding I have a number of shell tabs open with the CLI running. This change sets the window title with the product and project names, and restores them on exit.

This will work in linux, mac and newer windows environments.


packages/cli/src/gemini.tsx
```diff
@@ -38,6 +38,13 @@ const __filename = fileURLToPath(import.meta.url);
 const __dirname = dirname(__filename);
 
 async function main() {
+  const projectName = basename(process.cwd());
+  process.stdout.write(`\x1b]2;✨ Gemini - ${projectName} \x07`);
```
jacob314: Bonus points: can we add a setting for this so users who don't want this can opt out? GeminiCLI is pretty good at doing the boilerplate to add a setting so this shouldn't be a heavy lift other than the fact that this code will need to move to after settings have been loaded.
Might also want to consider not adding the `✨` to the title or making that a setting. @miguelsolorio might have some good ideas on the UI for this.

jacob314 (commented):
I personally can't wait to use this as I also end up with a lot of Gemini CLI running.

For future work, I'd also like it if a little bit of status showed up in my settings so I can easily wee which of my terminals are waiting for my input and which are working away.

packages/cli/src/gemini.tsx
miguelsolorio: I do think having a setting for this would be the best route as others will have their own preference. For example, we already show the cwd in the footer, so it feels little duplicative but I can see the benefit when you have multiple windows open.

If we go the setting route, I think it might make more sense to have a setting dialog with other settings to toggle. We've talked about allowing users to customize the footer items. But that's a larger effort. For now we can start with just `Gemini - ${projectName}`

markmcd: OK - all done. Removed emoji and gated with a flag. PTAL.

NTaylorMullen (approved):

packages/cli/src/gemini.tsx
```diff
@@ -175,4 +192,4 @@ async function loadNonInteractiveConfig(
     nonInteractiveSettings,
     config.getGeminiIgnorePatterns(),
   );
-}
+}
```
jacob314: nit: missing a trailing newline. have you run the formatter?

jacob314 (approved):
lgtm

PR #788, opened 2025-06-06
Setup Git repository for storing checkpoints

louisjimenez:
Adds a GitService for creating a hidden Git repository in the user's project that will be managed by the CLI. The hidden Git repository is operated against using `simple-git` operations. Git commits to the hidden repository will be used by the CLI to capture file edits and restore previous state.

While under development, I am keeping this functionality behind a cli flag. Once stable and complete, I will remove it or at the least default to true.

This PR provides the setup for the GitService, the full feature is becoming a larger change so breaking this into working, reviewable components.

### DIFF
--- a/packages/cli/src/config/config.ts
+++ b/packages/cli/src/config/config.ts
@@ -46,6 +46,7 @@ interface CliArgs {
   show_memory_usage: boolean | undefined;
   yolo: boolean | undefined;
   telemetry: boolean | undefined;
+  checkpoint: boolean | undefined;
 }
 
 async function parseArguments(): Promise<CliArgs> {
@@ -93,6 +94,12 @@ async function parseArguments(): Promise<CliArgs> {
     .option('telemetry', {
       type: 'boolean',
       description: 'Enable telemetry?',
+    })
+    .option('checkpoint', {
+      alias: 'c',
+      type: 'boolean',
+      description: 'Enables checkpointing of file edits',
+      default: false,
     })
     .version() // This will enable the --version flag based on package.json
     .help()
@@ -227,6 +234,7 @@ export async function loadCliConfig(
     fileFilteringRespectGitIgnore: settings.fileFiltering?.respectGitIgnore,
     fileFilteringAllowBuildArtifacts:
       settings.fileFiltering?.allowBuildArtifacts,
+    checkpoint: argv.checkpoint,
   };
 
   const config = createServerConfig(configParams);
--- a/packages/cli/src/gemini.tsx
+++ b/packages/cli/src/gemini.tsx
@@ -66,6 +66,11 @@ async function main() {
 
   // Initialize centralized FileDiscoveryService
   await config.getFileService();
+  try {
+    await config.getGitService();
+  } catch {
+    // For now swallow the error, later log it.
+  }
 
   if (modelWasSwitched && originalModelBeforeSwitch) {
     console.log(
--- a/packages/core/src/services/gitService.ts
+++ b/packages/core/src/services/gitService.ts
@@ -1,0 +1,96 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import * as fs from 'fs/promises';
+import * as path from 'path';
+import { isNodeError } from '../utils/errors.js';
+import { isGitRepository } from '../utils/gitUtils.js';
+import { exec } from 'node:child_process';
+import { simpleGit, SimpleGit, CheckRepoActions } from 'simple-git';
+
+export const historyDirName = '.gemini_cli_history';
+
+export class GitService {
+  private projectRoot: string;
+
+  constructor(projectRoot: string) {
+    this.projectRoot = path.resolve(projectRoot);
+    if (!isGitRepository(this.projectRoot)) {
+      throw new Error('GitService requires a Git repository');
+    }
+  }
+
+  async initialize(): Promise<void> {
+    const gitAvailable = await this.verifyGitAvailability();
+    if (!gitAvailable) {
+      throw new Error('GitService requires Git to be installed');
+    }
+    this.setupHiddenGitRepository();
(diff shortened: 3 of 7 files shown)


packages/cli/src/config/config.ts
```diff
@@ -94,6 +95,12 @@ async function parseArguments(): Promise<CliArgs> {
       type: 'boolean',
       description: 'Enable telemetry?',
     })
+    .option('checkpoint', {
```
NTaylorMullen, line 98: I feel like `checkpoint` might give the wrong vibe given it kind of reads like "checkpoint here so I can continue later".

packages/core/src/services/gitService.ts
```diff
@@ -0,0 +1,96 @@
...
+      });
+    });
+  }
+
+  /**
+   * Creates a hidden git repository in the project root.
+   * The Git repository is used to support checkpointing.
+   */
+  async setupHiddenGitRepository() {
+    const historyDir = path.join(this.projectRoot, historyDirName);
```
NTaylorMullen, line 51: Hmm should we put this in our `.gemini` folder?

packages/core/src/services/gitService.ts
```diff
@@ -0,0 +1,96 @@
...
+import * as path from 'path';
+import { isNodeError } from '../utils/errors.js';
+import { isGitRepository } from '../utils/gitUtils.js';
+import { exec } from 'node:child_process';
+import { simpleGit, SimpleGit, CheckRepoActions } from 'simple-git';
+
+export const historyDirName = '.gemini_cli_history';
+
+export class GitService {
+  private projectRoot: string;
```
NTaylorMullen, line 17:
```suggestion
  private readonly projectRoot: string;
```

packages/core/src/services/gitService.ts
```diff
@@ -0,0 +1,96 @@
...
+
+  async initialize(): Promise<void> {
+    const gitAvailable = await this.verifyGitAvailability();
+    if (!gitAvailable) {
+      throw new Error('GitService requires Git to be installed');
+    }
+    this.setupHiddenGitRepository();
+  }
+
+  verifyGitAvailability(): Promise<boolean> {
```
NTaylorMullen, line 34: I believe the `config` now has a isGitRepository piece that we can key off of

packages/core/src/services/gitService.ts
```diff
@@ -0,0 +1,96 @@
...
+    });
+  }
+
+  /**
+   * Creates a hidden git repository in the project root.
+   * The Git repository is used to support checkpointing.
+   */
+  async setupHiddenGitRepository() {
+    const historyDir = path.join(this.projectRoot, historyDirName);
+    const repoDir = path.join(historyDir, 'repository');
```
NTaylorMullen, line 52: On windows we're going to worry about path to long exceptions. I wonder if we should be storing this outside of the current repo actually /cc @mattKorwel

packages/core/src/services/gitService.ts
```diff
@@ -0,0 +1,96 @@
...
+    await fs.mkdir(repoDir, { recursive: true });
+    const repoInstance: SimpleGit = simpleGit(repoDir);
+    const isRepoDefined = await repoInstance.checkIsRepo(
+      CheckRepoActions.IS_REPO_ROOT,
+    );
+    if (!isRepoDefined) {
+      await repoInstance.init();
+    }
+
+    const visibileGitIgnorePath = path.join(this.projectRoot, '.gitignore');
```
NTaylorMullen, line 63: Instead of visible/hidden can we use current and shadow?

packages/core/src/services/gitService.ts
```diff
@@ -0,0 +1,96 @@
...
+      if (isNodeError(error) && error.code !== 'ENOENT') {
+        throw error;
+      }
+    }
+
+    await fs.writeFile(hiddenGitIgnorePath, visibileGitIgnoreContent);
+
+    if (!visibileGitIgnoreContent.includes(historyDirName)) {
+      const updatedContent = `${visibileGitIgnoreContent}\n# Gemini CLI history directory\n${historyDirName}\n`;
+      await fs.writeFile(visibileGitIgnorePath, updatedContent);
```
NTaylorMullen, line 82: This is a great reason to store the project outside of the current repo actually. Editing the user's `.gitconfig` without permissions feels intrusive

packages/cli/src/gemini.tsx
```diff
@@ -66,6 +66,11 @@ async function main() {
 
   // Initialize centralized FileDiscoveryService
   await config.getFileService();
+  try {
+    await config.getGitService();
```
NTaylorMullen, line 70: After leaning on the config's git repo checking you should be able to remove all the init stuff

NTaylorMullen (commented):
Shadow dirs are super exciting 😄 , I imagine we'll be able to do so much with this in the future. I wonder where Claude Code stores all their temp files

packages/cli/src/config/config.ts
louisjimenez: Good point, I'm going to consult with UX on this.

packages/core/src/services/gitService.ts
louisjimenez: I agree this doesn't belong here. I am going to move it into the user's home directory

packages/core/src/services/gitService.ts
louisjimenez: Right, moving this out.

packages/core/src/services/gitService.ts
louisjimenez: +1, will change in follow up PR to this work https://github.com/google-gemini/gemini-cli/pull/934

louisjimenez:
Thank you for the feedback @NTaylorMullen ! This work moved into this PR https://github.com/google-gemini/gemini-cli/pull/934 and I am going to apply your feedback in my fast follow up PR.

PR #794, opened 2025-06-06
Enable environment variable substitution in settings (useful for MCP servers)

jerop:
This commit introduces the ability to use system environment variables within the settings files (e.g., `settings.json`). Users can now reference environment variables using the `${VAR_NAME}` syntax.

This enhancement improves security and flexibility, particularly for configurations like MCP server settings, which often require sensitive tokens.

Previously, to configure an MCP server, a token might be directly embedded:
```json
"mcpServers": {
  "github": {
    "env": {
      "GITHUB_PERSONAL_ACCESS_TOKEN": "pat_abc123"
    }
  }
}
```

With this change, the same configuration can securely reference an environment variable:
```json
"mcpServers": {
  "github": {
    "env": {
      "GITHUB_PERSONAL_ACCESS_TOKEN": "${GITHUB_PERSONAL_ACCESS_TOKEN}"
    }
  }
}
```

This allows users to avoid storing secrets directly in configuration files.

Close https://github.com/google-gemini/gemini-cli/issues/791

### DIFF
--- a/packages/cli/src/config/settings.ts
+++ b/packages/cli/src/config/settings.ts
@@ -110,6 +110,15 @@ function resolveEnvVarsInString(value: string): string {
 }
 
 function resolveEnvVarsInObject<T>(obj: T): T {
+  if (
+    obj === null ||
+    obj === undefined ||
+    typeof obj === 'boolean' ||
+    typeof obj === 'number'
+  ) {
+    return obj;
+  }
+
   if (typeof obj === 'string') {
     return resolveEnvVarsInString(obj) as unknown as T;
   }
@@ -118,7 +127,7 @@ function resolveEnvVarsInObject<T>(obj: T): T {
     return obj.map((item) => resolveEnvVarsInObject(item)) as unknown as T;
   }
 
-  if (obj && typeof obj === 'object') {
+  if (typeof obj === 'object') {
     const newObj = { ...obj } as T;
     for (const key in newObj) {
       if (Object.prototype.hasOwnProperty.call(newObj, key)) {
--- a/packages/cli/src/config/settings.ts
+++ b/packages/cli/src/config/settings.ts
@@ -98,6 +98,39 @@ export class LoadedSettings {
   }
 }
 
+function resolveEnvVarsInString(value: string): string {
+  const envVarRegex = /\$(?:(\w+)|{([^}]+)})/g; // Find $VAR_NAME or ${VAR_NAME}
+  return value.replace(envVarRegex, (match, varName1, varName2) => {
+    const varName = varName1 || varName2;
+    if (process && process.env && typeof process.env[varName] === 'string') {
+      return process.env[varName]!;
+    }
+    return match;
+  });
+}
+
+function resolveEnvVarsInObject<T>(obj: T): T {
+  if (typeof obj === 'string') {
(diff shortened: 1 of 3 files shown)


packages/cli/src/config/settings.ts
```diff
@@ -98,6 +98,39 @@ export class LoadedSettings {
...
+    if (process && process.env && typeof process.env[varName] === 'string') {
+      return process.env[varName]!;
+    }
+    return match;
+  });
+}
+
+function resolveEnvVarsInObject<T>(obj: T): T {
+  if (typeof obj === 'string') {
+    return resolveEnvVarsInString(obj) as unknown as T;
```
jacob314, line 123: Is this `as unknown as T` really required?
worst case I would expect you can write
return resolveEnvVarsInString<T>(obj)
and the type system should understand.

packages/cli/src/config/settings.ts
```diff
@@ -98,6 +98,39 @@ export class LoadedSettings {
...
+    const varName = varName1 || varName2;
+    if (process && process.env && typeof process.env[varName] === 'string') {
+      return process.env[varName]!;
+    }
+    return match;
+  });
+}
+
+function resolveEnvVarsInObject<T>(obj: T): T {
+  if (typeof obj === 'string') {
```
jacob314, line 122: optional nit: consider starting this method with pass throughs returning obj for null, undefined, bool, and number.
At that point the case for object should always apply if the object isn't an array.

packages/cli/src/config/settings.ts
```diff
@@ -98,6 +98,39 @@ export class LoadedSettings {
   }
 }
 
+function resolveEnvVarsInString(value: string): string {
+  const envVarRegex = /\$(?:(\w+)|{([^}]+)})/g; // Find $VAR_NAME or ${VAR_NAME}
```
jacob314, line 102: should this be 
replaceAll rather than replace so that all matches of env variables are replaced?

plausible use case
`'${HOSTNAME}:${PORT}'`

jacob314 (approved):
lgtm
Like the flexibility this will give us for settings!

packages/cli/src/config/settings.ts
jerop: good catch - implemented in https://github.com/google-gemini/gemini-cli/pull/794/commits/7f582fe40cb60dc61012f3f2f043812cab558136

packages/cli/src/config/settings.ts
jerop: added a test for this: https://github.com/google-gemini/gemini-cli/commit/e2d1d13fcfbe21ec12969d96f2852d024ba3e4d0

packages/cli/src/config/settings.ts
jerop: tried this but ran into build errors:

```
src/config/settings.ts(114,5): error TS2322: Type 'string' is not 
   assignable to type 'T'.
'T' could be instantiated with an arbitrary type which could be 
   unrelated to 'string'.
src/config/settings.ts(118,5): error TS2322: Type 'any[]' is not 
   assignable to type 'T'.
'T' could be instantiated with an arbitrary type which could be 
   unrelated to 'any[]'.
```

PR #800, opened 2025-06-06
feat(ui): add cursor to empty input prompt

jacob314:
Previously, when the input prompt was empty, there was no cursor or visual indication that the component was focused and ready for input. This could be confusing for the user.

Screenshot 2025-06-06 at 11 44 32 AM

### DIFF
--- a/packages/cli/src/ui/components/InputPrompt.tsx
+++ b/packages/cli/src/ui/components/InputPrompt.tsx
@@ -39,7 +39,7 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
   onClearScreen,
   config,
   slashCommands,
-  placeholder = 'Type your message or @path/to/file',
+  placeholder = '  Type your message or @path/to/file',
   height = 10,
   focus = true,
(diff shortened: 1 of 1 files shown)


packages/cli/src/ui/components/InputPrompt.tsx
```diff
@@ -39,7 +39,7 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
   onClearScreen,
   config,
   slashCommands,
-  placeholder = 'Type your message or @path/to/file',
+  placeholder = '  Type your message or @path/to/file',
```
jacob314, line 42: added these two extra spaces because I think it looks better. We can remove resulting in the character "T" shows as inverted.

miguelsolorio (approved):
🙏 Thank you!

olcan:
woohoo

NTaylorMullen:
The cursor is BACK!

Oh and I love your theme @jacob314 (thank you @miguelsolorio for making them sexy)

PR #808, opened 2025-06-07
feat: Add flow to allow modifying edits during edit tool call

ngleo:
This is a first version and enables users to edit the diff proposed by the tool.

**Note:**
- Editors:
  - Only vscode or vimdiff available for now - vimdiff is a backup for users without a GUI.
  - vscode does not open with diff when the CLI is in `sandbox-execute` on macOS, hence hidden in selection if user is in sandbox mode.
  - we will add more options for users to configure the editor in a subsequent PR.
- If user decides to change the proposed diff too much, Gemini can retrigger the edit tool after the modified diff is applied.

**Flow:**
1) Edit tool confirmation
modify_flow_v2_1

2) Click on modify, confirmation box changes state and wait for user to exit editor
blocking_screen_3

3) VS Code view
modify_flow_v2_3a

4) vimdiff view (cursor default on the right, key instructions at top & bottom)
modify_flow_v3_3b

5) Upon exiting editor, confirmation box diff updated, flow continues as normal
modify_flow_v2_4


keithballinger:
Note that we are introducing a new file edit tool this weekend https://github.com/google-gemini/gemini-cli/pull/648 

you may want to wait until that goes in (Taylor and I are planning this evening)

packages/core/src/tools/edit.test.ts
```diff
@@ -600,4 +605,189 @@ describe('EditTool', () => {
...
+      await tool.onModify(params, new AbortController().signal);
+      const firstCall = mockOpenDiff.mock.calls[0];
+      const firstOldPath = firstCall[0];
+      const firstNewPath = firstCall[1];
+
+      expect(fs.existsSync(firstOldPath)).toBe(true);
+      expect(fs.existsSync(firstNewPath)).toBe(true);
+
+      // Add a small delay to ensure different timestamps
+      await new Promise((resolve) => setTimeout(resolve, 2));
```
jacob314, line 839: nit: why does it matter whether the timestamps are different?

packages/core/src/tools/edit.ts
```diff
@@ -459,6 +463,110 @@ Expectation for required parameters:
...
+    };
+  }
+
+  private deleteTempFiles(): void {
+    try {
+      if (this.tempOldDiffPath) {
+        fs.unlinkSync(this.tempOldDiffPath);
+        this.tempOldDiffPath = undefined;
+      }
+      if (this.tempNewDiffPath) {
```
jacob314, line 648: tweak this so that each deletion is surrounded by its own try catch. otherwise we won't clean up the second temp file if the first one can't be deleted.

jacob314:
This is looking really promising! Can't wait to use it.
Can we make this a setting so users can choose whether to opt into this?
Additionally would be nice to have a setting so users can choose their editor. For example, I might have Cursor and VSCode installed but could prefer a specific one for my Gemini CLI usage.
Good news is Gemini CLI is good at plumbing through settings so should only take a prompt to get this to be wired up as a setting rather than enabled by default.

packages/core/src/tools/edit.ts
ngleo: Sure, done!

packages/core/src/tools/edit.test.ts
ngleo: This is just for testing purpose as it creates the temp file names with a timestamp in it. I added comments to explain.

ngleo:
> This is looking really promising! Can't wait to use it. Can we make this a setting so users can choose whether to opt into this? Additionally would be nice to have a setting so users can choose their editor. For example, I might have Cursor and VSCode installed but could prefer a specific one for my Gemini CLI usage. Good news is Gemini CLI is good at plumbing through settings so should only take a prompt to get this to be wired up as a setting rather than enabled by default.

Would it be possible to follow up on this in a new PR? Or do you prefer including it in here?

jacob314:
Just played around with this. I'm good with landing this without a setting and landing that as a fast follow change as long as we  give the user a little warning what they are about to get themselves into.

This could be as simple as changing "Modify change" to "Modify change with an external editor"
or showing an initial Y/N confirmation dialog after they first click "Modify change" explaining that vimdiff is about to launch and that they should make changes to the RHS to modify the file. Confirm the user wants to do that before opening vimdiff.

That would ensure that users are not surprised when Gemin CLI disappears replaced by a vim based tool they may have trouble even exiting.

packages/core/src/utils/editor.ts
```diff
@@ -0,0 +1,93 @@
...
+          },
+        ]
+      : []),
+    // Terminal-based editors
+    {
+      name: 'vimdiff',
+      command: 'vimdiff',
+      args: (oldPath: string, newPath: string) => [oldPath, newPath],
+    },
+  ];
```
eddie-santos: As a historical vim user, I was actually not used to `vimdiff` and could not figure out how to toggle between left / right. How about we add the tool name, e.g. `Edit change/ vimdiff`, this way users can look up how to use the tool since we can't modify vimdiff to say `"use ctrl+w` to toggle diff sides"

packages/core/src/utils/editor.ts
```diff
@@ -0,0 +1,93 @@
...
+      ? [
+          {
+            name: 'vscode',
+            command: 'code',
+            args: (oldPath: string, newPath: string) => [
+              '--wait',
+              '--diff',
+              oldPath,
+              newPath,
+            ],
```
eddie-santos: This is great! I was literally just working on the same thing. We found an elegant way to use the Gemini Code Assist VSCode extension here, which is using the vscode protocol. We should make this command:

```bash
open vscode://google.geminicodeassist/open_diff?left=/path/to/left&right=/path/to/right
```

How about we then do the following?

Let's check if the GCA VS Code is installed, and then default to the protocol approach.

If not, then we use the `code` command.

Through the VS Code extension / protocol handler, we can get a richer experience, for example the "apply" button the top right.

We'd just need to also add OS specific commands, e.g. `open`, `xdg-open`, `start`, etc

packages/cli/src/ui/components/messages/ToolConfirmationMessage.tsx
```diff
@@ -63,6 +63,10 @@ export const ToolConfirmationMessage: React.FC<
         label: 'Yes, allow always',
         value: ToolConfirmationOutcome.ProceedAlways,
       },
+      {
+        label: 'Modify change',
+        value: ToolConfirmationOutcome.Modify,
+      },
```
eddie-santos: I think it doesn't hurt to give users some choice here, e.g.

`Modify change with vimdiff`
`Modify change in {VS Code, Windsurf, etc}`

We can then turn off an option if we don't detect VS Code for example.

packages/core/src/utils/editor.ts
eddie-santos: Note we should also add windsurf or other IDEs, but in order of:

1. GCA VS Code extension
2. VS Code
3. Windsurf
4. Etc

eddie-santos:
This is great @ngleo ! I was literally working on this same exact thing on Friday, cool to see we both recognized a pain button and have similar vision!

packages/core/src/utils/editor.ts
```diff
@@ -0,0 +1,93 @@
...
+    } catch {
+      return false;
+    }
+  };
+
+  const isSandbox = !!process.env.SANDBOX;
+  const diffTools = [
+    // GUI editors
+    ...(!isSandbox
+      ? [
```
eddie-santos: Something isn't working here. I have `GEMINI_SANDBOX=false` but I'm getting `vimdiff`

packages/core/src/utils/editor.ts
```diff
@@ -0,0 +1,93 @@
...
+              oldPath,
+              newPath,
+            ],
+          },
+        ]
+      : []),
+    // Terminal-based editors
+    {
+      name: 'vimdiff',
+      command: 'vimdiff',
```
eddie-santos: Also having challenges with vimdiff.

Here's what I get:
1. Diff window opens. I toggle to the right and make an edit, `:wq` to save and exit.
2. This takes me to the normal vim editor. I don't see the edits. I can make changes here, but when I try to save, I get

`E138: Can't write viminfo file /Users/eddiesantos/.viminfo.tmp!`
3. I get taken back to the edit tool, but I don't see any of the changes I made in vimdiff or vim. We should show the most recent edits in the edit tool so it's clear the changes were made.

We should fix this

eddie-santos (changes requested):

ngleo:
Thanks for the comments both! It sounds like making the naming of "Modify change" option more explicit will solve few birds with one stone. I will firstly in this PR expose the couple of options of "Modify change with {EDITOR_NAME}". 

Once we add the user settings for this in the next PR, we could let them configure and change the text to "Modify with your chosen editor".

Let me investigate the other issues :)

packages/cli/src/ui/components/messages/ToolConfirmationMessage.tsx
ngleo: Great idea!

packages/core/src/utils/editor.ts
ngleo: Could it be due to this? If it can't detect vscode , it will not consider it.
```
  const commandExists = (cmd: string): boolean => {
    try {
      execSync(`which ${cmd}`, { stdio: 'ignore' });
      return true;
    } catch {
      return false;
    }
  };
  ```

File: packages/core/src/utils/editor.ts (hunk 7 above)
Comment (ngleo) [reply packages/core/src/utils/editor.ts]: I will:
- Make cursor default to be in right window
- Make left window not editable
- Add explaination for keys (e.g. :wqa, :qa!, i) to status line

That should be way better! Resolving a related comment in favour of this one.

File: packages/core/src/utils/editor.ts (hunk 3 above)
Comment (ngleo) [reply packages/core/src/utils/editor.ts]: Will reply in the related comment about changes in vimdiff, resolving this

**Comment by eddie-santos** [comment]
Sounds good, tag me for re-review when ready

File: packages/core/src/utils/editor.ts (hunk 6 above)
Comment (ngleo) [reply packages/core/src/utils/editor.ts]: I will try to read the isSandbox from Config object instead.

File: packages/core/src/utils/editor.ts (hunk 7 above)
Comment (jacob314) [reply packages/core/src/utils/editor.ts]: Sounds great!

File: packages/core/src/utils/editor.ts (hunk 4 above)
Comment (ngleo) [reply packages/core/src/utils/editor.ts]: Strangely, I haven't been able to make this link work from my terminal. Let me follow up on internal chat if that is ok.

Regarding extra options of IDEs. I think this can be merged with the "configure editor in user settings" feature in the next PR.

**Comment by ngleo** [comment]
@eddie-santos Please could you take a look?

The GCA option might need more investigation (I have sent you an internal message). Otherwise, I have addressed your comments.

Also, added some UX improvements, please see the updated screenshots in the PR description. Thanks!

**Comment by eddie-santos** [comment]
Trying it out right now

**Comment by eddie-santos** [comment]
Hey @ngleo , check out this video on running these changes locally

https://screencast.googleplex.com/cast/NjY3ODA1MDU2NDQ3MjgzMnwxZWViNjkzZi0yNg

vimdiff isn't saving the changes I'm making, and VSCode is defaulting to the welcome page. Also note that I had to shut down the terminal because it hung after I clicked Modify change in VS Code, couldn't esc, ctrl+c, use arrow keys, or scroll

I was running into these challenges too, this is a tough nut to crack 😬

File: packages/cli/src/ui/components/messages/ToolConfirmationMessage.tsx [hunk 8]
```diff
@@ -63,8 +72,24 @@ export const ToolConfirmationMessage: React.FC<
...
+
+    // Conditionally add editor modification options if editors are installed
+    if (checkHasEditor('vscode') && !config?.getSandbox()) {
+      options.push({
+        label: 'Modify with VS Code',
+        value: ToolConfirmationOutcome.ModifyVSCode,
+      });
+    }
+
+    if (checkHasEditor('vimdiff')) {
```
jacob314: lets only show at most one "Modify with" option using the first available tool. Users who have a different preference can set their preferred tool in settings.json in the future.

packages/core/src/core/coreToolScheduler.ts
```diff
@@ -471,6 +483,33 @@ export class CoreToolScheduler {
         'cancelled',
         'User did not allow tool call',
       );
+    } else if (
+      outcome === ToolConfirmationOutcome.ModifyVSCode ||
```
jacob314, line 487: do these have to be separate tool outcomes? Can this just be ModifyExternalEditor or similar?

jacob314 (approved):
I'd suggest we hide this behind a disabled by default setting accessible through Config. Gemini CLI will one shot creating that so it is barely any more work than creating a const boolean option in ToolConfirmationMessage. If you have trouble with having Gemini CLI one shot it, feel free to just add the bool in ToolConfirmationMessage. 
Once that is addressed 'm good with landing this and working out the other issues in a follow up PR.
I am also seeing the issue that VSCode hangs my terminal running on mac with `sandbox-exec (minimal)`.

packages/core/src/utils/editor.ts
ngleo: You are right, my code was wrong!

ngleo:
Thanks for testing @eddie-santos. That helped me catch a lot of scenarios.

> vimdiff isn't saving the changes I'm making

I think vimdiff worked as intended there. The issue is that you updated the right window (new file), you closed it. Now the left window remains (old file). You edited and forced saved the old file, resulting in the diff you see. On execution, the old diff doesn't match the file content.

If you use :wqa, you can close both windows at the same time. But I have now made it automatically close both windows even if you close one. Also added more instructions at the top of the window.

> and VSCode is defaulting to the welcome page. 

This is a known issue when VS code when you are in sandbox mode. 

You are not supposed to see the VC code option if you're in sandbox – I have fixed this now. Could you please double check?

> Also note that I had to shut down the terminal because it hung after I clicked Modify change in VS Code, couldn't esc, ctrl+c, use arrow keys, or scroll

I have made opening VS Code non-blocking and adjusted the "editing screen" to make it more intuitive how is going on:
blocking_screen_3

packages/cli/src/ui/components/messages/ToolConfirmationMessage.tsx
ngleo: I adopted @eddie-santos 's suggestion to display the different options. I think for the time being without a settings.json, it might be more predictable / transparent if we display them all. 

However, I do agree that when we have a settings option, we should unify to "Modify with external editor". 

(p.s. Or even provide a setting UI similar to the themes setting, listing the multiple options that come out of the box. Providing pre-set editor options might be important because the commands might need some specific flags for the flow to work seamlessly, e.g. for VC Code, without the --wait flag, we cannot reliably listen to the editor's IO being closed and reactively update for UI)

packages/core/src/core/coreToolScheduler.ts
ngleo: This will be easier when we have user settings and unified them to "Modify with external editor". But this is related to [your other comment](https://github.com/google-gemini/gemini-cli/pull/808#discussion_r2134387321).

I will resolve this for now in favour of the other comment.

ngleo:
> I'd suggest we hide this behind a disabled by default setting accessible through Config.

I have added an `enableModifyWithExternalEditors` setting to hide the feature by default. Please add `"enableModifyWithExternalEditors": true` to your settings for testing.

@jacob314 @eddie-santos

packages/cli/src/ui/components/messages/ToolConfirmationMessage.tsx
```diff
@@ -44,6 +47,26 @@ export const ToolConfirmationMessage: React.FC<
   >();
 
   if (confirmationDetails.type === 'edit') {
+    if (confirmationDetails.isModifying) {
+      return (
+        <Box padding={0.5} minWidth="90%">
```
jacob314: nit: does padding 0.5 do anything?

packages/cli/src/ui/components/messages/ToolConfirmationMessage.tsx
```diff
@@ -44,6 +47,26 @@ export const ToolConfirmationMessage: React.FC<
...
+          <Box
+            width="100%"
```
jacob314: make this this width="90%" and remove the outer box assuming that does not change how things are rendered.

jacob314 (approved):
lgtm
Thanks for all the iteration!

packages/cli/src/ui/components/messages/ToolConfirmationMessage.tsx
ngleo: Ah yes, fixed. Thank you!

eddie-santos:
Testing now

eddie-santos (approved):
Alright, vimdiff is behaving much more intuitively, and I see the updated diff as expected, thanks!

> This is a known issue when VS code when you are in sandbox mode.

Note: I'm not using Sandbox as I don't have docker or podman installed on my local machine, and I've confirmed that `GEMINI_SANDBOX=false` in my .env file. If I turn it to true, the CLI doesn't start.

Approving for progress, let's get VS Code ironed out next.

Thanks @ngleo, merging!

ngleo:
Thanks @eddie-santos. 

Just double-checking - I noticed on your screencast that you're using sandbox-exec on macOS (which is the default for mac unless your set `export SEATBELT_PROFILE=none`)? If so, then I'd expect vs code option to not appear in the listed options. (If you open vs code, you will land on a blank page.) If SEATBELT_PROFILE=none, I'd expect you to see the vs code option.

Is that aligned with what you observed?

PR #812, opened 2025-06-07
Auto insert @ when dragging and dropping files.

jacob314:
This change automatically prepends an "@" symbol when a valid file path is dragged and dropped into the input prompt. This provides a more intuitive user experience for referencing files.

  The changes include:
   - Updated InputPrompt.tsx to pass an isValidPath function to the useTextBuffer hook.
   - Modified text-buffer.ts to check for valid file paths on text insertion and prepend "@" accordingly.
   - Added and updated tests in text-buffer.test.ts to cover the new file path handling logic.

### DIFF
--- a/packages/cli/src/ui/components/InputPrompt.tsx
+++ b/packages/cli/src/ui/components/InputPrompt.tsx
@@ -4,6 +4,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+import * as fs from 'fs';
 import React, { useCallback, useEffect, useState } from 'react';
 import { Text, Box, useInput, useStdin } from 'ink';
 import { Colors } from '../colors.js';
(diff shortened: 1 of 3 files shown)


packages/cli/src/ui/components/InputPrompt.tsx
```diff
@@ -58,11 +59,20 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
 
   const { stdin, setRawMode } = useStdin();
 
+  const isValidPath = useCallback((filePath: string): boolean => {
+    try {
+      return fs.existsSync(filePath) && fs.statSync(filePath).isFile();
```
NTaylorMullen, line 64: TIL of `isFile()`

packages/cli/src/ui/components/shared/text-buffer.ts
```diff
@@ -561,6 +564,20 @@ export function useTextBuffer({
       }
       dbg('insert', { ch, beforeCursor: [cursorRow, cursorCol] });
       pushUndo();
+
+      if (ch.length > 2) {
```
NTaylorMullen: Add a comment as to what `2` indicates or use a local var

NTaylorMullen (approved):
This definitely makes sense! One thing that's unclear to me though, does this also end up auto-prefixing file paths that are manually typed out? Want to make sure we're not injecting `@`s into user's prompts in normal typing scenarios

packages/cli/src/ui/components/shared/text-buffer.ts
jacob314: done

PR #813, opened 2025-06-07
bump minimum node version to 20

cornmander:
Matching GCP policies to support the LTS and Maintenance LTS node releases:

https://cloud.google.com/nodejs/getting-started/supported-nodejs-versions

Dependencies of gemini-cli like the @google/genai package already are 20+ only:
https://www.npmjs.com/package/@google/genai


NTaylorMullen (approved):
@cornmander could you also update the docs to reflect this pre-req too? 

/cc @cperry-goog for early access emails etc.

NTaylorMullen:
After chats with @bbiggs I'm uncertain if we should get this in anymore

mattKorwel:
> After chats with @bbiggs I'm uncertain if we should get this in anymore

what is the calculus on why we might not submit this? At a minium I would want to see a testing matrix that this works across 🍏 , 🪟 , 🐧.

cornmander:
Debian Stable is still on Node 18. The next Debian Stable is probably coming out by September. Last time we tested this, the codebase worked on Node 18 but it's not gauranteed to work if dependencies actually start using 20+ features. It's not likely that the genai library will pick up node 20+ features in the next 3 months, so I think we can just ignore this for now.

PR #817, opened 2025-06-07
Added /mcp command support and cleaned up broken tests

bdmorgan:
1. Added a new /mcp command between /help and /memory commands that:
    - Lists all configured MCP servers and their tools
    - Shows a clear message when no MCP servers are configured
    - Shows when a server has no tools available
2. Modified the existing /tools command to:
    - Only show Gemini CLI native tools, excluding MCP tools
    - Update the description to clarify this change
    - Change the output message to indicate "Gemini CLI tools"
3. Updated the tests to:
    - Verify the /tools command correctly filters out MCP tools
    - Test the new /mcp command with various scenarios (no registry, no servers, servers with tools, servers without tools)
    - Ensure both commands provide appropriate error messages when needed
    - Added tests for each status indicator scenario
    - Made sure tests handle the default status case (DISCONNECTED)
    - Ensured proper test isolation by clearing the status map before each test
4. Added MCP Server Status Tracking:
    - Created an MCPServerStatus enum with three states: CONNECTED, CONNECTING, and DISCONNECTED
    - Added a map (mcpServerStatuses) to track the status of each MCP server
    - Updated the connection and discovery logic to maintain server status information
5. Enhanced MCP Command UI:
    - Added visual status indicators (colored dots) for each MCP server:
      - 🟢 Green dot: Server is connected and ready
      - 🟡 Yellow dot: Server is in the process of connecting
      - 🔴 Red dot: Server is disconnected or experiencing errors
    - Maintained the existing functionality to list available tools for each server

Also cleaned up some test issues I was seeing:

1. Fixed test failures in GlobTool tests:
    - Fixed how we create invalid test parameters in the test cases
    - Used TypeScript's @ts-expect-error comment to indicate that we're intentionally creating invalid parameters for
  testing purposes
    - Maintained the test's original intent while making them type-safe
2. Also fixed a bug in the build process related to JSON validation in the sandbox_command.sh script by:
    - Adding proper error handling for JSON parsing
    - Adding a check to avoid using complex JSON objects as commands
    - Using jq if available, which is more reliable for JSON parsing

### DIFF
--- a/packages/core/src/tools/mcp-client.ts
+++ b/packages/core/src/tools/mcp-client.ts
@@ -13,6 +13,77 @@ import { DiscoveredMCPTool } from './mcp-tool.js';
 import { CallableTool, FunctionDeclaration, mcpToTool } from '@google/genai';
 import { ToolRegistry } from './tool-registry.js';
 
+/**
+ * Enum representing the connection status of an MCP server
+ */
+export enum MCPServerStatus {
+  /** Server is disconnected or experiencing errors */
+  DISCONNECTED = 'disconnected',
+  /** Server is in the process of connecting */
+  CONNECTING = 'connecting',
+  /** Server is connected and ready to use */
+  CONNECTED = 'connected'
+}
+
+/**
+ * Map to track the status of each MCP server within the core package
+ */
+const mcpServerStatusesInternal: Map<string, MCPServerStatus> = new Map();
+
+/**
+ * @deprecated Use getMCPServerStatus() and getAllMCPServerStatuses() instead
+ * This is exported only for backward compatibility
+ */
+export const mcpServerStatuses = mcpServerStatusesInternal;
+
+/**
+ * Event listeners for MCP server status changes
+ */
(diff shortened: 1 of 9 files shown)


packages/core/src/tools/mcp-client.ts
```diff
@@ -13,6 +13,77 @@ import { DiscoveredMCPTool } from './mcp-tool.js';
...
+  /** Server is connected and ready to use */
+  CONNECTED = 'connected'
+}
+
+/**
+ * Map to track the status of each MCP server within the core package
+ */
+const mcpServerStatusesInternal: Map<string, MCPServerStatus> = new Map();
+
+/**
```
scidomino, line 33: It might be best to remove this instead of maintaining backwards compatibility.

scidomino (approved):

packages/core/src/tools/mcp-client.ts
bdmorgan: fixed

PR #818, opened 2025-06-07
Add embedder

eddie-santos:
Adding Gemini embedder to be usable in tools


packages/core/src/core/geminiEmbed.ts
```diff
@@ -0,0 +1,48 @@
...
+ */
+
+import {
+  EmbedContentParameters,
+  EmbedContentResponse,
+  GoogleGenAI,
+} from '@google/genai';
+
+export class GeminiEmbed {
+  private static instance: GeminiEmbed;
```
cornmander: Why does this need to be a singleton? Are there other examples of singletons in the codebase?

Does it make sense to have this functionality in client.ts?

eddie-santos: Inspected a bit, and it does! Refactored accordingly, I think it's much cleaner.

packages/core/src/core/client.ts
```diff
@@ -450,6 +454,26 @@ export class GeminiClient {
     }
   }
 
+  async generateEmbedding(text: string, model: string): Promise<number[]> {
```
NTaylorMullen: Lets encourage computing multiple embeddings in parallel, worse case consumers can just pass a single array. Also until we need model configurability lets assume default

```suggestion
  async generateEmbedding(text: string[]): Promise<number[][]> {
```

NTaylorMullen (approved):

NTaylorMullen:
Any tool that ends up using embeddings should definitely be a convo! Embeddings are a spicey area and are VERY hard to get right

packages/core/src/core/client.ts
eddie-santos: Makes sense, updating. I decided not to let the embedding model be a CLI arg, so no need to have it as a variable here as well

snapxmin:
It looks like that generateEmbeddings(*)  does not be used by any tool now?

PR #822, opened 2025-06-07
refactor: rename gemini-code to gemini-cli

cperry-goog:


package-lock.json
NTaylorMullen, line 1: You may need to fully regenerate this but maybe not as well

NTaylorMullen (approved):
So good, there's a high chance this breaks publishing but that's ok. Lets get this in and the dev side can work through breaks as needed.

package-lock.json
cperry-goog: I regenerated before committing so we'll see

PR #826, opened 2025-06-07
Introduce ContentGeneratorConfig

scidomino:
This will better encapsulate the configuration for content generation and make core.Config easier to understand.

Also, this will make it easier to support the CCPA content generator once we have it.

### DIFF
--- a/packages/core/src/core/contentGenerator.ts
+++ b/packages/core/src/core/contentGenerator.ts
@@ -11,6 +11,7 @@ import {
   CountTokensParameters,
   EmbedContentResponse,
   EmbedContentParameters,
+  GoogleGenAI,
 } from '@google/genai';
 
 /**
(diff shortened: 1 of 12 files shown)


packages/core/src/core/contentGenerator.ts
```diff
@@ -29,3 +30,25 @@ export interface ContentGenerator {
 
   embedContent(request: EmbedContentParameters): Promise<EmbedContentResponse>;
 }
+
+export type ContentGeneratorConfig = {
+  model: string;
+  apiKey?: string;
+  vertexai?: boolean;
+  userAgent: string;
```
NTaylorMullen, line 38: If the goal of content generator config is for extensibility we may want to remove this bit. I'm not sure we want to allow replacing the user agent easily

NTaylorMullen (approved):

packages/core/src/core/contentGenerator.ts
scidomino: Makes sense. I'll do that in a separate PR as I want this to be a pure refactor.

PR #756, opened 2025-06-05
Creating Node AST Tool.

mattKorwel:
Ported by gemini-cli from https://github.com/jim-and-i/gca-cli-go-Deprecated-/tree/main/internal/tools/codeparsing.

Supports the following languages:

```
Optional: specific languages to parse (e.g., ["python", "java", "go", "csharp", "typescript", "tsx", "javascript", "rust"]).
```

PR #831, opened 2025-06-08
Fix Build Failure - Fails in sandbox due to missing build toolchain

abhipatel12:
The sandbox build (`npm run build:all`) is failing because of a missing build toolchain (python3, make, g++). This was caused by the introduction of the tree-sitter dependency, which requires native compilation via node-gyp. The base image used for the sandbox does not contain these tools by default.

This regression was likely introduced in PR #756 

Fix:
This PR modifies the root Dockerfile to include `python3`, `make`, and `g++` in the list of packages installed via apt-get.

How to Verify Fix:
1. Check out this branch.
2. Ensure your Podman/Docker environment is clean by running podman system `prune -a -f`.
3. Run `npm run build:all`.
4. The command should now complete successfully without any errors.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -98,6 +98,7 @@ npm run test
 This will run tests located in the `packages/core` and `packages/cli` directories. Ensure tests pass before submitting any changes.
 
 #### Important Note for Sandbox Users on macOS/Windows
+
 This project uses native dependencies (e.g., `tree-sitter`) that are compiled for a specific operating system.
 
 When you run the application in the development sandbox via `npm start`, these dependencies are automatically rebuilt for the container's Linux environment.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -97,6 +97,28 @@ npm run test
 
 This will run tests located in the `packages/core` and `packages/cli` directories. Ensure tests pass before submitting any changes.
 
+#### Important Note for Sandbox Users on macOS/Windows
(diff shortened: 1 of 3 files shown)


abhipatel12:
**Update: Second Bug Found at Runtime**

After fixing the initial build failure by adding the build toolchain to the Dockerfile, a second error occurs at runtime when starting the application with `npm start`

Runtime Error:

```
Error: No native build was found for platform=linux arch=arm64 ...
   loaded from: /path/to/project/node_modules/tree-sitter
```

Likely Cause:

The application's startup process correctly re-launches itself inside a sandbox container. However, it mounts the host project directory, including the host's `node_modules` folder. The application inside the Linux container then attempts to load the tree-sitter native addon, but it finds the version that was compiled for the host's OS (e.g., macOS darwin).

Next Steps:
The project's startup logic needs to be modified to rebuild native dependencies inside the container after it starts.

abhipatel12:
This fix modifies the container entrypoint command to run `npm rebuild` before `npm start`. This recompiles the native addons for the container's Linux environment, resolving the crash.

abhipatel12:
Adding some additional notes about using `npm run tests`. With the above changes, running tests locally without first rebuilding causes issues.

NTaylorMullen (approved):
Love all the commentary and tracing of your flow @abhipatel12!

:shipit: 

/cc @olcan @KeijiBranshi

abhipatel12:
Thank you @NTaylorMullen! Appreciate the review!

PR #832, opened 2025-06-08
feat: display commit hash in detached HEAD state

scottdensmore:
When the git repository is in a detached HEAD state, the branch name is displayed as 'HEAD'. This is not very informative. This change modifies the  hook to display the short commit hash instead of 'HEAD' when the repository is in a detached HEAD state.

Fixes #686


NTaylorMullen (approved):
love it

PR #840, opened 2025-06-08
fix(deps): externalize tree-sitter

mattKorwel:
fixes #839 

Bundling issues impacted NPX install


mattKorwel:
Introduced via https://github.com/google-gemini/gemini-cli/pull/756

mattKorwel:
Still testing...

mattKorwel:
NPX test command `npx --yes --no-cache --loglevel verbose https://github.com/google-gemini/gemini-cli#issue-839 -p "test"`

mattKorwel:
Understanding the fix for future us

```
✦ Excellent! It seems the npx command finally ran successfully. The error is resolved.

  I have:


   1. Identified the root cause of the tree-sitter native module error.
   2. Tried several approaches to fix the issue, including using esbuild loaders and plugins.
   3. Finally, resolved the issue by marking the tree-sitter and its language packages as external in the esbuild.config.js file.
   4. Pushed all the changes to the issue-839 branch.


  The package should now be installable and usable via npx.

> can you explain the error to me in more detail and how you fixed it?

✦ Of course. Here’s a breakdown of the error and the steps taken to fix it.


  The Core Problem: Native Modules and Bundling


  The error Error: No native build was found... is specific to Node.js packages that rely on "native modules."


   1. What are Native Modules? Most npm packages are written in pure JavaScript. However, some, like tree-sitter, use C++ addons (.node files) for
      performance-critical tasks. These are pre-compiled binaries specific to an operating system (Linux, macOS, Windows) and Node.js version. When you run
      npm install, it downloads the correct binary for your environment.


   2. How Bundlers Complicate Things: Your project uses esbuild to bundle all the necessary JavaScript code into a single file (bundle/gemini.js). This is
      great for distribution, but esbuild doesn't know how to package the C++ native modules. It can't embed the binary .node file into the JavaScript
      bundle.


  The error occurred because the bundled gemini.js file tried to load tree-sitter, but the required .node file wasn't inside the bundle, nor was the code
  correctly configured to find it elsewhere.


  The Journey to the Fix


  Our goal was to make the native modules available to the bundled code when it runs after being installed by npx.


   1. Initial State: The project tried to solve this by having a script (copy_bundle_assets.sh) that copied the .node files into the bundle/ directory. This
      didn't work because the code inside gemini.js didn't know to look for them there.


   2. Attempt 1: Using an `esbuild` Loader: My first attempt was to use esbuild's --loader:.node=file option. This tells esbuild to copy the .node file to
      the output directory and change the require statement to return the path to that new file. This approach also failed, likely because tree-sitter has a
      complex way of loading its modules that wasn't compatible with this simple path replacement.


   3. Attempt 2: Using Plugins: I then tried using plugins (esbuild-native-node-modules-plugin and a custom one) designed to handle native modules. These
      are more sophisticated and attempt to intelligently copy the binaries and rewrite the paths in the bundled code. However, these also failed,
      indicating the way tree-sitter and its various language packages are structured was still too complex for the plugins to handle automatically.

  The Final Solution: Marking as "External"


  The successful solution was to stop trying to bundle tree-sitter altogether.


  In the esbuild.config.js file, I marked tree-sitter and all its language packages as "external":



   1 external: [
   2   'tree-sitter', 
   3   'tree-sitter-c-sharp', 
   4   'tree-sitter-go', 
   5   'tree-sitter-java', 
   6   'tree-sitter-python', 
   7   'tree-sitter-rust', 
   8   'tree-sitter-typescript'
   9 ],


  Why this worked:


   * It tells `esbuild` to ignore them. When esbuild encounters require('tree-sitter'), it leaves the line of code exactly as is in the final bundle. It
     doesn't try to find the tree-sitter files and pull them in.
   * It relies on `npm`'s standard behavior. When a user runs npx ... gemini-cli, npm installs the gemini-cli package and all of its dependencies (as
     listed in package.json) into a temporary node_modules folder.
   * Node.js finds the modules at runtime. Now, when the gemini.js bundle executes and hits the require('tree-sitter') line, Node.js performs its normal
     lookup: it finds tree-sitter in the node_modules folder, which in turn correctly loads its native .node binary.


  In essence, we let the bundler handle the JavaScript code and let npm and the Node.js runtime handle the native dependencies, which is the standard and
  most reliable way to manage them.
```

mattKorwel:
Confirmed working on main with `npx --yes clear-npx-cache && npx --loglevel verbose --yes https://github.com/google-gemini/gemini-cli`

NTaylorMullen (commented):
Really nice cleanup for this too. The inlining before was far from pretty :)

PR #853, opened 2025-06-08
Add Windsurf in edit tool to modify changes, if installed

eddie-santos:
If the user has Windsurf installed, it'll show up in the list of options to modify the change

image

image


ngleo (approved):
Lgtm! Thanks for adding this so quickly

NTaylorMullen:
How do you feel about collapsing all the modify options down into 1 and then having its own flow to pick which editor to ues? I'd hate for the confirmation dialog to grow unboundedly

cc @miguelsolorio in case he has opinions

scottdensmore:
Should we add a setting for what editor and default to the `$EDITOR`? I agree with @NTaylorMullen that we don't want to turn into so many setttings that is overwhelming.

miguelsolorio:
+1 to "Modify with Editor" and maybe we include this as part of the onboarding "wizard" so they select their default. 

As an aside, I also think we need a separate "Modify" option that allows the user to redirect via the input.

ngleo:
Allowing users to pick an editor in settings + unifying the options to "Modify with external editor" is on my list, as promised!

Plan is to create a UI (similar to theme settings) for picking editors with a few configurations that come out of the box.

eddie-santos:
@NTaylorMullen @scottdensmore  +1 to having a `preferredEditor` option in settings, perhaps defaulting to VS Code + vimdiff, but also include others.

Note that the only reason I'm okay with it here is that this relies on them having these editors installed, i.e. if they don't have Windsurf installed they won't see it as an option. Might be magical for them if they rely on one (e.g. VS Code) and it magically appears, but to your point it could also overwhelm them if they have multiple.

PR #867, opened 2025-06-09
Add support for sorting files search by recency threshold followed by lexicographic sorting

anj-s:
- Recency threshold set to 1 day
- Created a sorting function
- Added tests

Fixes https://github.com/google-gemini/gemini-cli/issues/562


NTaylorMullen (approved):
Love it, we should also do something like this for completion lists too (although we may already have it), if so we may want to dedupe core logic

PR #869, opened 2025-06-09
rename shell tool more intuitively as run_shell_command (from historical name of execute_bash_command inherited from terminal tool)

olcan:


NTaylorMullen:
Love it

PR #871, opened 2025-06-09
feat(core): support .aiexclude and refactor file filtering

### LINKED ISSUE #653: Add the ability to add an exclude file e.g. `.aiexclude` in the settings

mrahs:
File filtering is centralized in `FileDiscoveryService` and controlled by the following sources:
- .geminiignore which defines glob patterns to be excluded from
  listings.
- .gitignore for git repositories.
- .aiexclude for AI specific exclusions.

Other custom settings for exclusions are removed, including
`customIgnorePatterns` and `allowBuildArtifacts`, which were noops as of
this.

Fixes #653

### DIFF
--- a/packages/core/src/services/fileDiscoveryService.ts
+++ b/packages/core/src/services/fileDiscoveryService.ts
@@ -4,19 +4,27 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { GitIgnoreParser, GitIgnoreFilter } from '../utils/gitIgnoreParser.js';
+import { GitIgnoreParser } from '../utils/gitIgnoreParser.js';
+import { AIExcludeParser } from '../utils/aiExcludeParser.js';
+import { IgnoreFilter } from '../utils/ignoreFilter.js';
 import { isGitRepository } from '../utils/gitUtils.js';
+import { makeRelative } from '../utils/paths.js';
 import * as path from 'path';
 import fg from 'fast-glob';
+import micromatch from 'micromatch';
 
 export interface FileDiscoveryOptions {
   respectGitIgnore?: boolean;
+  respectAIExclude?: boolean;
+  geminiIgnorePatterns?: string[];
   includeBuildArtifacts?: boolean;
   isGitRepo?: boolean;
 }
 
 export class FileDiscoveryService {
-  private gitIgnoreFilter: GitIgnoreFilter | null = null;
+  private gitIgnoreFilter: IgnoreFilter | null = null;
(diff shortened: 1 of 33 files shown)


packages/core/src/services/fileDiscoveryService.ts
```diff
@@ -70,14 +101,27 @@ export class FileDiscoveryService {
...
+    // aiexclude takes precedence over gitignore
+    // see https://cloud.google.com/gemini/docs/codeassist/create-aiexclude-file#write_an_aiexclude_file
     if (
+      options.respectAIExclude !== false &&
+      this.aiExcludeFilter?.isIgnored(filePath)
+    ) {
+      return true;
+    }
+
+    if (
```
scidomino, line 117: if (foo) { return true; } else {return false;} 

is the same as:

return foo;

packages/core/src/services/fileDiscoveryService.ts
```diff
@@ -87,4 +131,16 @@ export class FileDiscoveryService {
...
+
+  /**
+   * Returns true if the filePath matches any glob pattern in geminiIgnorePatterns
+   */
+  geminiIgnoreMatch(filePath: string, patterns?: string[]): boolean {
+    const allPatterns = (this.geminiIgnorePatterns ?? []).concat(
+      patterns ?? [],
+    );
+    const relativePath = makeRelative(filePath, this.projectRoot);
+    const isMatch = micromatch.isMatch(relativePath, allPatterns);
```
scidomino, line 143: nit: inline this var

packages/core/src/utils/aiExcludeParser.ts
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  isIgnored(filePath: string): boolean {
+    const normalizedPath = normalizeFilePath(this.aiExcludeRoot, filePath);
+
+    if (normalizedPath === '' || normalizedPath.startsWith('..')) {
+      return false;
+    }
+
+    const ignored = this.ig.ignores(normalizedPath);
```
scidomino, line 48: nit: inline

scidomino (approved):

mattKorwel (changes requested):
Actually, before we approve, can we talk about breaking this down into smaller parts. Also can we document in the issue what testing has and hasn't happened? This is a significant set of changes and we're not far from release. Want to make sure we're being safe.

PR #877, opened 2025-06-09
Integration tests: v0

mattKorwel:
## TLDR (Generated by GC)

This PR makes several changes to help make it easier to keep our code in a working state.

- to run non sandbox integration tests
```bash
npm run test:e2e
```

- to run a standard validation of your local code in the same way as ci
```bash
npm run preflight
```

- to launch the CLI in step through debugging in VSCode press `F5`

## BUGFIX: Note on Proxy networking

This pr makes what I believe is a bug fix to proxy networking. Without the proxy networking command set, the current code was still creating a number of docker config commands that were in fact shutting off all networking in the container. This fix ensures that without networking proxy set no network commands are created. *This fix was found via new integration tests here*

### Build times
I do see that these tests could, if we add an ever expanding set of test cases could become prohibitively slow. currently we use a fresh sandbox for each test. There are multiple future mitigations (running a 'fast' suite blocking ci and a full suite on merge to main, not doing a new sandbox for each request etc). 

### API Key Capacity

We are using GitHub Secrets management to store a shared API key from the `gemini-cloud-dev` GCP project. This API Key will at some point run out of capacity and hit throttling issues.  TBD on what to do about that. 

## Dive Deeper: A Guide to Integration Testing

For those new to the concept, integration tests are a way to verify that different parts of our application work together correctly. In the context of the Gemini CLI, this means we are testing the entire application from the command line, just as a user would, to ensure that commands for file operations, and other features behave as expected.

### Why Do We Need Integration Tests?

While unit tests are great for testing individual functions in isolation, they don't tell us if the entire application works when all the pieces are connected. Integration tests fill this gap by:

1.  **Simulating Real User Scenarios:** They run the actual `gemini` command and check its output and effects on the file system.
2.  **Catching Bugs Early:** They can uncover issues related to configuration, environment differences, or unexpected interactions between components that unit tests might miss.
3.  **Providing Confidence:** A passing integration test suite gives us high confidence that the CLI is working correctly for our users.

### The Integration Test Framework

This PR establishes a new framework located in the `integration-tests/` directory. Here's how it works:

-   **`run-tests.js`:** This is the heart of the test runner. It discovers all `*.test.js` files and executes them one by one. It's responsible for setting up a unique, temporary directory for each test run to ensure tests don't interfere with each other.
-   **`test-helper.js`:** This file contains the `TestRig` class, a powerful helper that simplifies writing tests. It handles the boilerplate of setting up temporary directories, creating test files, and running the `gemini` command itself.
-   **Test Files (`*.test.js`):** These files contain the actual test cases, written using Node.js's built-in test runner. Each test uses the `TestRig` to simulate a scenario and then makes assertions about the outcome.

## Reviewer Test Plan: How to Use the Framework

As a developer, you can easily run these tests to validate your changes.

### Running Tests

To run the entire suite of integration tests without any sandboxing and with verbose logging and preservation of tmp files used during command runs, use the following command:

```bash
npm run integration-test:no-sandbox
```

You can also run a specific test file or even a single test case by name:

```bash
# Run a single file
npm run integration-test:no-sandbox -- integration-tests/file-system.test.js

# Run tests whose names match a pattern
npm run integration-test:no-sandbox -- --test-name-pattern "writes a file"
```

### Debugging Your Tests

The framework includes powerful debugging tools:

**1. Keeping Temporary Files (`--keep-output` or `KEEP_OUTPUT`)**

By default, the temporary directories created for each test are deleted after the run. To inspect the files created by a test, you can tell the runner to keep them.

```bash
# Use the flag
npm run test:integration:sandbox:none -- --keep-output

# Or use the environment variable
KEEP_OUTPUT=true npm run test:integration:sandbox:none
```

**2. Verbose Logging (`--verbose`)**

If a test is failing, you'll want to see the actual output from the `gemini` command. The `--verbose` flag streams all command output to your console in real-time.

```bash
npm run test:integration:sandbox:none -- --verbose --keep-output
```

The output is clearly marked to show which test it came from, making it easy to follow along:

```
--- TEST: file-system:writes-a-file ---
... all output from the gemini command for this test ...
--- END TEST: file-system:writes-a-file ---
```

### Code Quality

To maintain code quality, the integration test files are automatically checked for issues during the build process (`npm run preflight`). You can also manually fix any linting problems:

```bash
npm run lint --fix
```

### Continuous Integration (CI)

To ensure these tests are always run, a new GitHub Actions workflow has been added in `.github/workflows/e2e.yml`. This workflow automatically runs the integration tests on every pull request and push to `main`, testing against multiple sandboxing environments (no sandbox, Docker, and Podman) to ensure broad compatibility.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Closes #876


scidomino (dismissed):

scottdensmore:
@mattKorwel still working on this one?

packages/cli/src/utils/sandbox.ts
mattKorwel, line 1: This is a bugfix for proxy networking that was causing docker/podman sandboxing to not be able to access the network.

scripts/build_sandbox.js
mattKorwel, line 1: This change removes code that prevents docker from running on mac. verified on multiple devices. I'm not aware that we are using buildx in any capacity. Will revise if that understanding is wrong.

scripts/start.js
mattKorwel, line 1: This change was needed to get correct exit codes for testing.

mattKorwel:
> @mattKorwel still working on this one?

@scottdensmore yep got back to this now.

integration-tests/write_file_test.js
```diff
@@ -0,0 +1,25 @@
...
+import { TestRun } from '../test-helper';
+import * as fs from 'fs';
+import * as path from 'path';
+
+test('should be able to write a file', async () => {
+  const testRun = new TestRun();
+  try {
+    const prompt =
+      'Can you write a file named "new_file.txt" with the content "new file content"';
+    await testRun.run(prompt);
```
NTaylorMullen: Hmmm, i would have thought you'd need to pass `--yolo` here given the underlying system uses the -p parameter (unable to provide confirmations). Something might be up here

packages/cli/src/nonInteractiveCli.ts
```diff
@@ -97,6 +97,7 @@ export async function runNonInteractive(
             console.error(
               `Error executing tool ${fc.name}: ${toolResponse.resultDisplay || toolResponse.error.message}`,
             );
+            process.exit(1);
```
NTaylorMullen, line 100: Hmm why exit here vs. naturally end?

packages/cli/src/utils/sandbox.ts
```diff
@@ -485,38 +484,42 @@ export async function start_sandbox(sandbox: string) {
...
+      args.push('--env', `HTTP_PROXY=${proxy}`);
+      args.push('--env', `http_proxy=${proxy}`);
+    }
+    const noProxy = process.env.NO_PROXY || process.env.no_proxy;
+    if (noProxy) {
+      args.push('--env', `NO_PROXY=${noProxy}`);
+      args.push('--env', `no_proxy=${noProxy}`);
+    }
+
+    // if using proxy, switch to internal networking through proxy
```
NTaylorMullen, line 529:
```suggestion
```

scripts/build_sandbox.js
NTaylorMullen: @KeijiBranshi may have more insight here

NTaylorMullen (approved):
Incredibly impressive. The tests feel very natural. Love this

mattKorwel:
> Incredibly impressive. The tests feel very natural. Love this

Thanks. @NTaylorMullen  One thing i forgot to mention. I didn't intend or really desire to write a whole testing framework. If we can later retro fit this to just use vitest like the reset then that's totally 1000% good and frankly probably best. I was just deeper than i wanted before i realized i had diverged. just fyi i'm not tied to any of this.

integration-tests/write_file_test.js
mattKorwel: You would think right ;) check the latest commit. All though I'm not sure why the file-system.tests were workig becuase write-file tests woudln't without yolo. added in the latest round

NTaylorMullen:
Note: We'll probably want to cherry-pick this to the release branch once it makes it into main. Could also consider rebasing onto release as well

packages/cli/src/nonInteractiveCli.ts
mattKorwel: Without it the natural exit was for some reason not passing ack the failure return code.  It was always zero. There's a better fix, something is aligned poorly up the stack but this was tactically enough for now.

mattKorwel:
I did not enable auto merge. Want to give folks a little bit of feedback time, but we now have a green build so mid day monday will merge.

PR #878, opened 2025-06-09
feat(cli): support ctrl+d to exit

taeold:
Similar to ctrl+c, ctrl+d can now be used to exit the program. To avoid accidental exit, ctrl+d must be pressed twice in relatively quick succession (same as ctrl+c).

Following common UX pattern, ctrl+d will be ignored when the input prompt is non-empty. This behavior is similar to how most shell (bash/zsh) behaves. To support this, I had to refactor so that text buffer is initialized outside of the `InputPrompt` component and instead do it on the main App component to allow input controller to have access to check the content of the text buffer.


dewitt:
Thanks so much for the PR!

I'm +1 to ctrl+d exiting the program _if typed at an empty prompt_.

I'm -1 on ctrl+c clearing the current line, which seems to be non-standard behavior, though I'm definitely open to hearing evidence otherwise.

NTaylorMullen:
Ya I'd definitely like to keep ctrl + c as an exit ability

NTaylorMullen (dismissed):
@taeold for simplicity sake would you mind rebasing / making it so the changes are minimal for getting ctrl + d to exit?

packages/cli/src/ui/App.tsx
```diff
@@ -151,58 +171,103 @@ const App = ({ config, settings, startupWarnings = [] }: AppProps) => {
...
           setCtrlCPressedOnce(false);
           ctrlCTimerRef.current = null;
         }, CTRL_C_PROMPT_DURATION_MS);
       }
+    } else if (key.ctrl && (input === 'd' || input === 'D')) {
+      if (buffer.text.length > 0) {
+        // Do nothing if there is text in the input.
+        return;
+      }
+      handleExit(ctrlDPressedOnce, setCtrlDPressedOnce, ctrlDTimerRef);
```
abhipatel12, line 263: Could you ensure that `ctrl+d` uses the same exit command as `ctrl+c`? Currently it looks like ctrl+d uses a different function which may not cause the exit session summary stats to show up.

abhipatel12 (changes requested):

packages/cli/src/ui/App.tsx
```diff
@@ -151,58 +171,103 @@ const App = ({ config, settings, startupWarnings = [] }: AppProps) => {
...
+        if (timerRef.current) {
+          clearTimeout(timerRef.current);
+        }
+        process.exit(0);
+      } else {
+        setPressedOnce(true);
+        timerRef.current = setTimeout(() => {
+          setPressedOnce(false);
+          timerRef.current = null;
+        }, CTRL_C_PROMPT_DURATION_MS);
```
abhipatel12: We should rename this if we are going to use this duration for more than just ctrl+c

packages/cli/src/ui/App.tsx
taeold: done

abhipatel12:
LGTM! I have one remaining comment on the name of the constant for duration if you can make it more generic but it's a nit!

abhipatel12 (approved):

packages/cli/src/ui/App.tsx
taeold: great call

PR #882, opened 2025-06-10
feat: Editor settings

ngleo:
Allow users to configure a "preferred editor" in a settings UI using `/editor` command.

Settings UI will automatically pop up if user wants to "Modify with external editor" but preferred editor not configured.
Screenshot 2025-06-12 at 00 19 58

History messages
Screenshot 2025-06-10 at 21 58 09
Screenshot 2025-06-10 at 21 58 15

Automatically open editor settings
Screenshot 2025-06-10 at 21 59 01


packages/cli/src/ui/editors/editorSettingsManager.ts
```diff
@@ -0,0 +1,60 @@
...
+};
+
+class EditorSettingsManager {
+  private readonly availableEditors: EditorDisplay[];
+
+  constructor() {
+    const editorTypes: EditorType[] = ['vscode', 'windsurf', 'cursor', 'vim'];
+    this.availableEditors = [
+      {
+        name: 'Not set',
```
miguelsolorio: Should this be `None` instead?

packages/cli/src/ui/components/EditorSettingsDialog.tsx
```diff
@@ -0,0 +1,140 @@
...
+    }
+    if (key.escape) {
+      onExit();
+    }
+  });
+
+  const editorItems: EditorDisplay[] =
+    editorSettingsManager.getAvailableEditorDisplays();
+  const currentPreference = settings.merged.preferredEditor;
+  const initialEditorIndex = currentPreference
```
jacob314: optional nit: 
log an error if findIndex returns -1
explaining that the currentPreference wasn't found.

packages/cli/src/ui/components/EditorSettingsDialog.tsx
```diff
@@ -0,0 +1,140 @@
...
+    ? Math.max(
+        0,
+        editorItems.findIndex(
+          (item: EditorDisplay) => item.type === currentPreference,
+        ),
+      )
+    : 0;
+
+  const scopeItems = [
```
jacob314, line 60: nit: there is now matching code to specify this UI in the ThemeDialog and this dialog. Consider asking Gemini CLI to refactor to have unified component for letting the user switch between user and workspace settings.  Suspect this won't be the last setting editor we add. Would expect that we will want to refine the ui for toggling whether settings are set on user or workspace settings in the future as the current UI for this (I'm to blame) is a little clunky.

packages/cli/src/ui/components/EditorSettingsDialog.tsx
```diff
@@ -0,0 +1,140 @@
...
+    }
+    onSelect(editorType, selectedScope);
+  };
+
+  const handleScopeSelect = (scope: SettingScope) => {
+    setSelectedScope(scope);
+    setFocusedSection('editor');
+  };
+
+  let otherScopeModifiedMessage = '';
```
jacob314, line 78: tthis is also logic that is common for all setting editing UI (this + theme dialog). Maybe we should refactor to have a base SettingDialog component that embeds this functionality.

packages/cli/src/ui/components/EditorSettingsDialog.tsx
jacob314: after more review I think it would be worth it to define a 
SettingsDialog.tsx that handles this functionality as well as surfacing the UI explaining that the setting was also modified by your other config file. Gemini CLI might even be able to do that whole refactor for you if given the 2 relevant files.

packages/cli/src/ui/components/shared/RadioButtonSelect.tsx
```diff
@@ -88,7 +89,7 @@ export function RadioButtonSelect<T>({
   /**
    * Custom item component for displaying the label.
    * Color changes based on whether the item is selected and if its group is focused.
-   * Now also handles displaying theme type with custom color.
+   * Now also handles displaying theme type with custom color and dims unavailable items.
```
jacob314: nit: remove all of line 92 as it is describing how the code changed not what it does.

packages/cli/src/ui/components/shared/RadioButtonSelect.tsx
```diff
@@ -97,11 +98,14 @@ export function RadioButtonSelect<T>({
...
       themeNameDisplay?: string;
       themeTypeDisplay?: string;
+      disabled?: boolean;
     };
 
     let textColor = Colors.Foreground;
     if (isSelected) {
       textColor = Colors.AccentGreen;
+    } else if (itemWithThemeProps.disabled === true) {
+      textColor = Colors.Gray; // Dim disabled items
```
jacob314: nit: strip low value comment.

packages/cli/src/ui/hooks/useGeminiStream.ts
```diff
@@ -76,6 +77,8 @@ export const useGeminiStream = (
     cmd: PartListUnion,
   ) => import('./slashCommandProcessor.js').SlashCommandActionReturn | boolean,
   shellModeActive: boolean,
+  loadedSettings: LoadedSettings,
+  onEditorNotConfigured?: () => void,
```
jacob314: nit: make this required as there isn't any need to run in a mode that doesn't support configuring editors.

packages/core/src/core/coreToolScheduler.ts
```diff
@@ -484,15 +496,23 @@ export class CoreToolScheduler {
...
-      outcome === ToolConfirmationOutcome.ModifyVSCode ||
-      outcome === ToolConfirmationOutcome.ModifyWindsurf ||
-      outcome === ToolConfirmationOutcome.ModifyCursor ||
-      outcome === ToolConfirmationOutcome.ModifyVim
-    ) {
+    } else if (outcome === ToolConfirmationOutcome.ModifyWithEditor) {
       const waitingToolCall = toolCall as WaitingToolCall;
       if (waitingToolCall?.confirmationDetails?.type === 'edit') {
         const editTool = waitingToolCall.tool as EditTool;
+        const editorType = this.getPreferredEditor?.();
```
jacob314: for discussion: what do you think about using Node EventEmitter for this instead of a callback?
We can then use useEffect on the react side to listen for the events from the event emitter. I think that could help keep this code from getting coupled to react.
Potentially this code could be as simple as dispatching the edit event leaving the react code in the cli package to deal with all the business logic for configuring an editor.

packages/cli/src/ui/hooks/useGeminiStream.ts
```diff
@@ -102,6 +110,8 @@ export const useGeminiStream = (
       },
       config,
       setPendingHistoryItem,
+      getPreferredEditor,
```
jacob314, line 120: is there a reason this can't be
loadedSettings.merged.preferredEditor? instead of passing in a getter?

packages/cli/src/ui/hooks/useReactToolScheduler.ts
```diff
@@ -162,12 +164,17 @@ export function useReactToolScheduler(
         onAllToolCallsComplete: allToolCallsCompleteHandler,
         onToolCallsUpdate: toolCallsUpdateHandler,
         approvalMode: config.getApprovalMode(),
+        config,
+        getPreferredEditor,
+        onEditorNotConfigured,
```
jacob314: rather than passing onEditorNotConfigured and getPreferredEditor here I'd suggest we add a single onEdit callback and leave the business logic for calling onEditorNotConfigured to this file. That would keep the react code separate from the less reactive code. you could either use EventEmitter as suggested earlier or just pass in a simple callback to here.

jacob314 (dismissed):
This looks really good. Sorry for all the comments. All comments are for discussion rather than strictly required.  Feel free to land the large changes from the comment as a follow up pull request or two rather than as part of this pull request. High level the two key points in the comments are thinking about building some shared infra for settings dialogs and think about what we should do when we have code in core that needs to call back to react. ideally the code in core could use standard node idioms like EventEmitter rather than getting coupled to React idioms.

packages/cli/src/ui/hooks/useGeminiStream.ts
ngleo: Sure

packages/cli/src/ui/components/shared/RadioButtonSelect.tsx
ngleo: Sure

packages/cli/src/ui/components/shared/RadioButtonSelect.tsx
ngleo: Sure, fixed.

packages/cli/src/ui/components/EditorSettingsDialog.tsx
ngleo: Sure, will address in follow up PRs

packages/cli/src/ui/components/EditorSettingsDialog.tsx
ngleo: I actually think that having the scope setting on the same dialog is rather convenient! Not sure about other users...

packages/cli/src/ui/components/EditorSettingsDialog.tsx
ngleo: Sure, fixed.

packages/cli/src/ui/editors/editorSettingsManager.ts
ngleo: Sure, fixed.

packages/cli/src/ui/components/EditorSettingsDialog.tsx
jacob314: I like the scope setting but have hopes and dreams of at some point making it fit on one line instead of taking up 3 lines. @miguelsolorio

packages/cli/src/ui/hooks/useGeminiStream.ts
ngleo: This is needed for the flow where the the settings automatically pops up, since the scheduler code will not get an updated value until some rerender. Or I can pass the entire LoadedSettings. Let me think about how to deal with this in relation to your other comment.

packages/cli/src/ui/hooks/useReactToolScheduler.ts
ngleo: I have now created a `getPreferredEditor` in App.tsx to handle the business logic of opening the settings if there is no preferred editor, and passed it to the scheduler.

packages/core/src/core/coreToolScheduler.ts
ngleo: As mentioned in the other comment, I took your suggestion to update the callbacks to 1 single `getPreferredEditor` callback  that is created in cli layer. I think it's reasonable clean and is separating concerns at the moment.

Maybe we should consider `EventEmitter` when there are multiple use cases where the core components need to talk to the cii?

ngleo:
@jacob314 Thanks for the feedback! I have addressed most of them now. Will leave the creation of a base `settingdialog` component to follow up PRs.

packages/cli/src/ui/hooks/useGeminiStream.test.tsx
```diff
@@ -358,8 +358,7 @@ describe('useGeminiStream', () => {
           props.onDebugMessage,
           props.handleSlashCommand,
           props.shellModeActive,
-          props.loadedSettings,
-          () => {},
+          () => 'vscode' as EditorType,
```
jacob314, line 361: nit: use value from the enum directly rather than `as`

jacob314 (approved):
lgtm

Thanks for all the cleanup! Looks good now. Can't wait to use this!

PR #879, opened 2025-06-09
feat: Display initial token usage metrics in /stats

abhipatel12:
**Issue:** Continues work to address #811
**Related PR:** #854 

### Overview

This PR updates the `/stats` feature in the CLI to contain more information from the `UsageData` in API responses from Gemini. The ultimate goal is to provide users with valuable insights into their token usage, performance, and overall interaction during a session.

### Scope of this Initial Commit

This is the first in a chain of commits for this feature. The changes included here are strictly foundational and focus on setting up the necessary state management pattern.

-   **Updates to stats available in SessionContext:** `SessionContext now contains Gemini API `UsageMetadata` for cumulative and last turn metrics. This gives us an in-depth idea about how token input, output, tool usage, think, modalities, etc. 
-   **/stats update:** Currently prints out some metrics like total tokens, input tokens, output tokens, and cache efficiency in addition to session duration. These are not set in stone and may be changed. This was just to add a proof of concept.

**Current State of `/stats`:**

Screenshot 2025-06-09 at 5 21 23 PM

### Next Steps (in follow-up PRs)
-   [ ] Develop the UI component to render the formatted `/stats` output.
-   [ ] Add client-side logic to track API time per turn (and possibly total)
-   [ ] Implement the on-exit summary display

PR #854, opened 2025-06-08
feat: Introduce session-based usage statistics

abhipatel12:
**Issue:** Works to address #811

### Overview

This PR introduces a new usage statistics feature in the CLI. The ultimate goal is to provide users with valuable insights into their token usage, performance, and overall interaction during a session.

This will be delivered in two ways:
1.  **On-Demand:** A `/stats` slash command that displays detailed metrics for the current session and the most recent turn.
2.  **On-Exit:** A summary of session usage presented when the user exits the CLI.

This initial commit establishes the core architecture for session state management but does not yet include the final UI or data collection logic.

### Scope of this Initial Commit

This is the first in a chain of commits for this feature. The changes included here are strictly foundational and focus on setting up the necessary state management pattern.

-   **React Context for Session State:** Adds a new `SessionContext`. This provides a centralized, application-wide store for session data, avoiding the need to pass state down through multiple component layers.
-   **Session Provider:** Implements `SessionProvider` which wraps the main `App` component. This ensures that session state is available to the entire component tree.
-   **Initial State Definition:** Defines the basic structure for the session state, starting with a session `startTime`. I will add additional metrics like token counts, performance metrics, and more later in subsequent commits. 

**Current UI:**

Screenshot 2025-06-08 at 4 38 49 PM

### Next Steps (in this PR)

-   [ ] Expand the `SessionContext` to hold all relevant statistical data (such as token counts from the Gemini API, etc.).
-   [ ] Implement the logic to capture `UsageMetadata` from API responses and update the session state.
-   [ ] Develop the UI component to render the formatted `/stats` output.
-   [ ] Add client-side logic to track wall-clock time and API latency per turn.
-   [ ] Implement the on-exit summary display (This can also be made into a separate PR if this one becomes to lengthy).

This PR is a work-in-progress. Please let me know if there is any feedback as I work to add onto this chain!

Issue #811, opened 2025-06-07
[FR] Report token usage stats upon user request and on exit


abhipatel12:
Usage stats are important for users who want to understand how many tokens a current session has consumed. This can be surfaced in 2 ways:
* Upon user request via CLI (i.e. `/stats` or `/cost` or `/usage`
* Upon session exit

Example of a similar feature from `claude`:
```
Total cost:            $0.0005
Total duration (API):  1.1s
Total duration (wall): 16.2s
Total code changes:    0 lines added, 0 lines removed
Tokens:                443 base input, 0 cache write, 0 cache read, 24 base output
```

Note: Issue #664 tracks an adjacent feature where token usage is shown in real time per given request.

NTaylorMullen:
I'll be super curious how you try and track `cost` here 😄 . We'll need to estimate so we'll definitely want to be careful with our wording around that as well.

abhipatel12:
That's true! As an initial effort to add this feature, I think usage in terms of simply tokens (input, output, cache read) may be best. 

I still believe a *estimated* cost would be helpful to users (possibly as a follow-up)! New users may have greater difficulty understanding tokens and their pricing structure which can possibly introduce friction. For example, new users may see token counts dramatically increase as the size of a project increases (ex. 400 to 80k+). The large token counts (input + output) may seem daunting when in reality, it may only equate to fractions of a dollar.

I think the "how" will be difficult though. The pricing structure when considering input, output, caching, etc, can be quite complex. There also isn't a `getCost` func or `cost` field on the API responses.

Some possible options:
1. Best-effort maintained config that contains pricing configs for all models that can be used in gemini-cli
2. A mapping for SKU to models that we can use to retrieve pricing via the cloud billing catalog API
3. (Less likely) Ask the genai API team to include a cost field or function. 

And of course, careful language on it being an estimate. Would love any suggestions!

NTaylorMullen:
> 1. Best-effort maintained config that contains pricing configs for all models that can be used in gemini-cli

Early on I followed up with the billing team and unfortunately there's roughly a ~9 hour delay before things take effect. I think this will have to be our only option 😄

nullbio:
Related: https://github.com/google-gemini/gemini-cli/issues/2094

> The large token counts (input + output) may seem daunting when in reality, it may only equate to fractions of a dollar.

[...]

PR #883, opened 2025-06-10
feat: Add UI for /stats slash command

abhipatel12:
**Issue:** Continues work to address #811
**Related PR:** #854, #879

### Overview

This PR updates the `/stats` feature in the CLI to show a new UI component that contains last turn and cumulative token counts. It also shows total API & Wall time.

### Scope of Commits

These commit contain the initial addition of the UI component and the wiring of stats available. This was then followed by another commit that added last turn token stats and total API time.

-   **Updates to stats available in SessionContext:** `SessionContext` now contains token stats aggregated for the most recent turn. Additionally, the API time total and last turn have been added.
-   **/stats update:** Currently prints out token stats for last turn and cumulative and prints API time for last turn and total session.

**Current State of `/stats`:**

Screenshot 2025-06-10 at 1 53 53 AM

### Next Steps (in follow-up PRs)
-   [ ] Implement the on-exit summary display
-   [ ] Implement code change stats (i.e. lines added, deleted)


packages/cli/src/ui/components/HistoryItemDisplay.test.tsx
```diff
@@ -0,0 +1,104 @@
...
+    <Text>ErrorMessage: {text}</Text>
+  ),
+}));
+vi.mock('./AboutBox.js', () => ({
+  AboutBox: () => <Text>AboutBox</Text>,
+}));
+vi.mock('./StatsDisplay.js', () => ({
+  StatsDisplay: () => <Text>StatsDisplay</Text>,
+}));
+vi.mock('./messages/ToolGroupMessage.js', () => ({
```
jacob314, line 14: nit: these don't seem like things that should be mocked. Can you ask Gemini CLI to keep the behavior of the test unchanged but not mock these components?

packages/cli/src/ui/components/StatsDisplay.test.tsx
```diff
@@ -0,0 +1,108 @@
...
+        stats={mockStats}
+        lastTurnStats={mockLastTurnStats}
+        duration={mockDuration}
+      />,
+    );
+
+    const output = lastFrame();
+
+    // Check for main title and duration
+    expect(output).toContain('Stats');
```
jacob314: nit: checking all these with toContain is a bit odd. Can you instead that the output matches a single golden string. That will make the behavior clearer. With Gemini CLI in the loop updating the single golden string will be easy enough as Gemini CLI is good at fixing the test expectation automatically if the output changes.

packages/cli/src/ui/components/StatsDisplay.tsx
```diff
@@ -0,0 +1,160 @@
...
+        <StatRow
+          label="Thoughts Tokens"
+          value={stats.thoughtsTokens.toLocaleString()}
+        />
+        <StatRow
+          label="Cached Tokens"
+          value={cachedDisplay}
+          valueColor={cachedColor}
+        />
+        <Box
```
jacob314, line 92: nit: ask Gemini CLI if there is a more idiomatic way to achieve the effect of drawing a line here than this empty box.

packages/cli/src/ui/components/StatsDisplay.tsx
```diff
@@ -0,0 +1,160 @@
...
+        <StatsColumn title="Last Turn" stats={lastTurnFormatted} />
+        <StatsColumn
+          title={`Cumulative (${stats.turnCount} Turns)`}
+          stats={cumulativeFormatted}
+          isCumulative={true}
+        />
+      </Box>
+
+      <Box flexDirection="row" justifyContent="flex-end" marginTop={1}>
+        <Box width="48%" flexDirection="column">
```
jacob314: optional nit: I see 48% a couple times here.  Consider making it a constant or refactoring the layout so you only have to hard code the width in one spot.

packages/cli/src/ui/contexts/SessionContext.test.tsx
```diff
@@ -136,31 +141,40 @@ describe('SessionStatsContext', () => {
...
     expect(stats?.cumulative.toolUsePromptTokenCount).toBe(10);
 
-    // Last turn should hold the metadata from the most recent call
-    expect(stats?.lastTurn?.metadata).toEqual(mockMetadata2);
+    // --- Check Current Turn Stats ---
+    // All fields should be the SUM of both calls for the turn
+    expect(stats?.currentTurn.totalTokenCount).toBe(330);
+    expect(stats?.currentTurn.candidatesTokenCount).toBe(220);
+    expect(stats?.currentTurn.thoughtsTokenCount).toBe(22);
+    expect(stats?.currentTurn.promptTokenCount).toBe(110); // 100 + 10
```
jacob314: nit: rather than writing 100 + 10 in a comment replace 110 with 100 + 10 to make this self documenting.

jacob314 (approved):
lgtm
This looks so good! Added a few minor comments.

packages/cli/src/ui/components/HistoryItemDisplay.test.tsx
abhipatel12: Good call! Fixed this.

packages/cli/src/ui/components/StatsDisplay.test.tsx
abhipatel12: True! Created a snapshot instead that can be compared against since that would likely be better! In the future, the CLI should be able to update as changes are made!

packages/cli/src/ui/components/StatsDisplay.tsx
abhipatel12: Yeah i didn't like it earlier either. I double checked but there's not a "Divider" in the ink library. It was either this or some hacky text that was drawn using some column width measurement. 

For now this is the most "idiomatic" after consulting the cli tool and gemini in web again. I can add a comment to make it more clear tho!

packages/cli/src/ui/components/StatsDisplay.tsx
abhipatel12: Done!

packages/cli/src/ui/contexts/SessionContext.test.tsx
abhipatel12: Done!

abhipatel12:
Just an update, had to fix a few bugs that were missed originally where the cumulative stats were not aggregating the all of the API calls per turn (just the first call) which made the stat very wrong. 

This has been fixed.

Along the way, it was a quick add for API time per turn so that has been added. At this point of the PR, the most up to date UI is below:

Screenshot 2025-06-10 at 3 27 13 PM

packages/cli/src/ui/components/__snapshots__/StatsDisplay.test.tsx.snap
```diff
@@ -0,0 +1,43 @@
+// Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
```
jacob314, line 1: awesome! TIL!

jacob314 (approved):
lgtm

packages/cli/src/ui/components/__snapshots__/StatsDisplay.test.tsx.snap
abhipatel12: Same, really cool way to test. Way better than the first iteration i had. Thanks for leaving a comment or i probably wouldn't have reworked the test!

PR #888, opened 2025-06-10
GEMINI_SANDBOX=false should disable seatbelt

scidomino:
Refactors the sandbox configuration and execution logic.

  Key changes:
  - GEMINI_SANDBOX=false now disables all sandboxing
   - The --yolo flag now correctly disables the sandbox.
   - Sandbox command detection is improved, prioritizing sandbox-exec on macOS.
   - The sandbox configuration is now optional in the core config.

### DIFF
--- a/packages/cli/src/config/config.ts
+++ b/packages/cli/src/config/config.ts
@@ -151,7 +151,7 @@ export async function loadCliConfig(
   return new Config({
     contentGeneratorConfig,
     embeddingModel: DEFAULT_GEMINI_EMBEDDING_MODEL,
-    sandbox: sandbox,
+    sandbox,
     targetDir: process.cwd(),
     debugMode,
     question: argv.prompt || '',
--- a/packages/cli/src/config/config.ts
+++ b/packages/cli/src/config/config.ts
@@ -143,10 +143,15 @@ export async function loadCliConfig(
 
   const contentGeneratorConfig = await createContentGeneratorConfig(argv);
 
+  let sandbox = argv.sandbox ?? settings.sandbox;
+  if (argv.yolo) {
+    sandbox = false;
+  }
+
   return new Config({
     contentGeneratorConfig,
     embeddingModel: DEFAULT_GEMINI_EMBEDDING_MODEL,
-    sandbox: argv.sandbox ?? settings.sandbox ?? argv.yolo ?? false,
+    sandbox: sandbox,
     targetDir: process.cwd(),
     debugMode,
     question: argv.prompt || '',
--- a/packages/cli/src/utils/sandbox.ts
+++ b/packages/cli/src/utils/sandbox.ts
@@ -109,42 +109,44 @@ export function sandbox_command(sandbox?: string | boolean): string {
   if (sandbox === '1' || sandbox === 'true') sandbox = true;
   else if (sandbox === '0' || sandbox === 'false') sandbox = false;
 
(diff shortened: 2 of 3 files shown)


bdmorgan (approved):

packages/cli/src/config/config.ts
```diff
@@ -143,10 +143,15 @@ export async function loadCliConfig(
 
   const contentGeneratorConfig = await createContentGeneratorConfig(argv);
 
+  let sandbox = argv.sandbox ?? settings.sandbox;
+  if (argv.yolo) {
```
jacob314, line 147: what are we disabling the sandbox in yolo mode? We should be enabling sandbox in yolo mode not disabling this. 
please ask Gemini to add a test for this as well. Probably worth ensuring that we are enabling the sandbox when we expect to.

packages/cli/src/utils/sandbox.ts
```diff
@@ -102,49 +102,46 @@ async function getSandboxImageName(
   );
 }
 
-// node.js equivalent of scripts/sandbox_command.sh
```
jacob314, line 105: thanks for removing this comment!

packages/cli/src/utils/sandbox.ts
```diff
@@ -102,49 +102,46 @@ async function getSandboxImageName(
...
 export function sandbox_command(sandbox?: string | boolean): string {
   // note environment variable takes precedence over argument (from command line or settings)
   sandbox = process.env.GEMINI_SANDBOX?.toLowerCase().trim() ?? sandbox;
   if (sandbox === '1' || sandbox === 'true') sandbox = true;
```
jacob314, line 108: not your fault but I wonder if it is worth doing a little bit of refactoring to make this method a bit clearer. right now I found this a bit hard to read. Given this logic is quite important I think it would be worth simplifying where possible as well as well as asking Gemini to write some tests to make sure it is behaving as expected. added a couple comments on ideas of refactors to simplify.

packages/cli/src/utils/sandbox.ts
```diff
@@ -102,49 +102,46 @@ async function getSandboxImageName(
...
   else if (sandbox === '0' || sandbox === 'false') sandbox = false;
```
jacob314, line 109: we could combine the three cases checking for no-sandbox mode by writing
```if (sandbox == '0' || sandbox == 'false' || sandbox == false) {
  return '';
}
```
at line 108
rather than converting to a boolean and then checking.

packages/cli/src/utils/sandbox.ts
```diff
@@ -102,49 +102,46 @@ async function getSandboxImageName(
...
+    return 'docker';
+  } else if (commandExists.sync('podman')) {
+    return 'podman';
+  }
 
-    return ''; // no sandbox
+  // throw an error if user requested sandbox but no command was found
+  if (sandbox === true) {
+    console.error(
+      'ERROR: GEMINI_SANDBOX is true but failed to determine command for sandbox; ' +
```
jacob314, line 138: nit: could be nice to use ansi escape codes to make the text red 
Here an example from gemini.tsx. It could make sense to extract this as a helper.

      let errorMessage = `Error in ${error.path}: ${error.message}`;
      if (!process.env.NO_COLOR) {
        errorMessage = `\x1b[31m${errorMessage}\x1b[0m`;
      }

PR #892, opened 2025-06-10
fix version set for cli and slash commands

zsais:
Fix for https://github.com/google-gemini/gemini-cli/issues/607. It also sets the correct version in the "/about" command. Previously, we were using the version of NPM as the "CLI Version".


scottdensmore (approved):
Looks good. Thanks for the fix! :shipit:

PR #895, opened 2025-06-10
Fix inconsistent diff rendering

mrahs:
Diffs were being generated slightly differently before and after applying the edit, leading to intermittent inconsistency in the renderer.
Using the same helper function for both call sites should eliminate that.

Also, added a unit test to cover multiple hunks handling in the parser, which turned out to be unrelated to the bug, but still good to have.

Fixes #536


scottdensmore (approved):
LGTM

PR #898, opened 2025-06-10
Add multi-line input support with Alt+Enter

### LINKED ISSUE #849: <shift>-<enter> should do an enter but keep user control
**Describe the bug**
Sometimes I want to tell gemini cli: 
"""
Look at this output, this is clearly wrong:

  - gic_migration_2024_06_25: ✅ Assertion passed: Command succeeded.
  - sakura_repo_exists: ✅ Path exists: /Users/ricc/git/sakura
"""

Where the first part (lines 1/2) im typing and the second part (3/4) is the output of some commands i cut pasted.

However, as soon as I press the first eNTER i lose control and gemini cli gets it. We need to have some mechanism to introduce an <ENTER> without ceding focus to the LLM and still be in tyiuping mode.

I suppose SHIFT ENTER is the most natural, but happy to consider others too (or maybe one exists and Im not aware of it)

deepankarsharma:
Enable users to compose multi-line messages before submitting. This addresses the need to paste multi-line content or format messages with proper line breaks without immediately sending them.

- Add Alt+Enter (Option+Enter on Mac) keybinding for inserting newlines
- Change prompt from '>' to '»' when buffer contains newlines for visual feedback
- Update documentation with new keyboard shortcut
- Add tip in startup message about multi-line input capability
- Hide autocomplete suggestions when inserting newlines

Alt+Enter was chosen for multi-line input because Shift+Enter doesn't work reliably across different terminal emulators, where many terminals treat it identically to pressing Enter alone.

Commit tested in terminal and VS Code terminal on Mac and Linux.

Fixes #849

### DIFF
--- a/docs/cli/commands.md
+++ b/docs/cli/commands.md
@@ -2,7 +2,14 @@
 
 Gemini CLI supports several built-in commands to help you manage your session, customize the interface, and control its behavior. These commands are prefixed with a forward slash (`/`), an at symbol (`@`), or an exclamation mark (`!`).
 
-## Slash commands (`/`)
+## Keyboard Shortcuts
+
+- **Multi-line Input:**
+  - **Shift+Enter** or **Ctrl+Enter** or **Alt+Enter** (Option+Enter on Mac): Insert a new line without submitting your message
+  - **Enter**: Submit your message
+  - When in multi-line mode, the prompt changes from `>` to `»` to indicate multi-line content
+
+## Slash Commands (`/`)
 
 Slash commands provide meta-level control over the CLI itself.
 
@@ -79,6 +86,24 @@ Slash commands provide meta-level control over the CLI itself.
 - **`/quit`** (or **`/exit`**)
 
   - **Description:** Exit Gemini CLI.
+
+- **`/terminal-setup`**
+  - **Description:** Automatically detects and configures your terminal emulator to work with gemini-cli.
+  - **Action:**
+    - Detects your current terminal emulator
+    - For supported terminals, automatically adds the necessary keybinding configuration
+    - Creates backups of configuration files before modifying them
+    - Reports success or failure with appropriate instructions
+  - **Supported Terminals:**
+    - **Kitty** - Adds keybindings to `~/.config/kitty/kitty.conf`
+    - **Alacritty** - Adds keybindings to `~/.config/alacritty/alacritty.toml` or `.yml`
+    - **Ghostty** - Adds keybindings to `~/.config/ghostty/config`
+    - **URxvt** - Adds keybindings to `~/.Xresources` (requires `xrdb -merge ~/.Xresources`)
+    - **VS Code** - Adds keybindings to VS Code's `keybindings.json`
+    - **Foot** - Works out of the box, no configuration needed
+  - **Notes:**
+    - Most terminals require a restart after configuration
+    - VS Code requires restarting the entire application, not just the terminal
 
 ## Checkpointing Commands
 
--- a/packages/cli/src/ui/components/InputPrompt.tsx
+++ b/packages/cli/src/ui/components/InputPrompt.tsx
@@ -9,7 +9,7 @@ import { Box, Text } from 'ink';
 import { Colors } from '../colors.js';
 import { SuggestionsDisplay } from './SuggestionsDisplay.js';
 import { useInputHistory } from '../hooks/useInputHistory.js';
-import { TextBuffer } from './shared/text-buffer.js';
+import { TextBuffer, NEWLINE_INPUT_SEQUENCES } from './shared/text-buffer.js';
 import { cpSlice, cpLen } from '../utils/textUtils.js';
 import chalk from 'chalk';
 import stringWidth from 'string-width';
@@ -273,8 +273,14 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
         console.log('[InputPromptCombined] event', { key });
       }
 
+      const { sequence: input } = key;
       // Ctrl+Enter for newline, Enter for submit
-      if (key.name === 'return') {
+      if (
+        key.name === 'return' &&
+        !key.ctrl &&
+        !key.shift &&
+        !NEWLINE_INPUT_SEQUENCES.includes(input || '')
+      ) {
         const [row, col] = buffer.cursor;
         const line = buffer.lines[row];
         const charBefore = col > 0 ? cpSlice(line, col - 1, col) : '';
@@ -367,7 +373,7 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
         <Text
           color={shellModeActive ? Colors.AccentYellow : Colors.AccentPurple}
         >
-          {shellModeActive ? '! ' : '> '}
+          {shellModeActive ? '! ' : buffer.lines.length > 1 ? '» ' : '> '}
         </Text>
         <Box flexGrow={1} flexDirection="column">
           {buffer.text.length === 0 && placeholder ? (
--- a/packages/cli/src/ui/components/Tips.tsx
+++ b/packages/cli/src/ui/components/Tips.tsx
@@ -24,9 +24,20 @@ export const Tips: React.FC<TipsProps> = ({ config }) => {
       <Text color={Colors.Foreground}>
         2. Be specific for the best results.
       </Text>
+      <Text color={Colors.Foreground}>
+        3. Use{' '}
+        <Text bold color={Colors.AccentPurple}>
+          Alt+Enter
+        </Text>{' '}
+        or{' '}
+        <Text bold color={Colors.AccentPurple}>
+          Shift+Enter
+        </Text>{' '}
+        for multi-line input.
+      </Text>
       {geminiMdFileCount === 0 && (
         <Text color={Colors.Foreground}>
-          3. Create{' '}
+          4. Create{' '}
           <Text bold color={Colors.AccentPurple}>
             GEMINI.md
           </Text>{' '}
@@ -34,7 +45,7 @@ export const Tips: React.FC<TipsProps> = ({ config }) => {
         </Text>
       )}
       <Text color={Colors.Foreground}>
-        {geminiMdFileCount === 0 ? '4.' : '3.'}{' '}
+        {geminiMdFileCount === 0 ? '5.' : '4.'}{' '}
         <Text bold color={Colors.AccentPurple}>
           /help
         </Text>{' '}
--- a/packages/cli/src/ui/utils/terminalSetup.ts
+++ b/packages/cli/src/ui/utils/terminalSetup.ts
@@ -1,0 +1,514 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * Terminal setup utility for configuring Shift+Enter and Ctrl+Enter support.
+ *
+ * This module provides automatic detection and configuration of various terminal
+ * emulators to support CSI u format escape sequences for modified Enter keys.
+ *
+ * Supported terminals:
+ * - Kitty: Configures ~/.config/kitty/kitty.conf
+ * - Alacritty: Configures ~/.config/alacritty/alacritty.toml (or .yml)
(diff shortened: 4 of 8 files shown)


packages/cli/src/ui/components/InputPrompt.tsx
```diff
@@ -292,11 +302,12 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
         console.log('[InputPromptCombined] event', { input, key });
       }
 
-      // Ctrl+Enter for newline, Enter for submit
+      // Alt+Enter for newline, Enter for submit
       if (key.return) {
-        if (key.ctrl) {
-          // Ctrl+Enter for newline
+        if (key.meta) {
```
tolikmalibroda: Why "Alt+Enter" and not "Shift+Enter"? I believe that latter one is more common

deepankarsharma: I tried implementing Shift-Enter first since that is what the user requested but the terminal library we are using [ink](https://github.com/vadimdemedes/ink) does not seem to distinguish between Shift-Enter and Enter. In fact in the raw key event, shift pressed is seen as false.

docs/cli/commands.md
```diff
@@ -2,6 +2,13 @@
 
 The Gemini CLI supports several built-in commands to help you manage your session, customize the interface, and control its behavior. These commands are typically prefixed with a forward slash (`/`), an at symbol (`@`), or an exclamation mark (`!`).
 
+## Keyboard Shortcuts
+
+- **Multi-line Input:**
+  - **Alt+Enter** (Option+Enter on Mac): Insert a new line without submitting your message
+  - **Enter**: Submit your message
+  - When in multi-line mode, the prompt changes from `>` to `»` to indicate multi-line content
```
jacob314, line 10: @miguelsolorio curious about your thoughts on this

packages/cli/src/ui/components/Tips.tsx
```diff
@@ -31,9 +31,16 @@ export const Tips: React.FC<TipsProps> = ({ config }) => {
       <Text color={Colors.Foreground}>
         3. Be specific for the best results.
       </Text>
+      <Text color={Colors.Foreground}>
+        4. Use{' '}
+        <Text bold color={Colors.AccentPurple}>
+          Alt+Enter
+        </Text>{' '}
+        for multi-line input.
```
jacob314, line 36: love adding this to tips. We should also consider adding a terminal-setup option and suggesting people call it to support shift+enter on mac.

packages/cli/src/ui/components/InputPrompt.tsx
```diff
@@ -195,7 +195,11 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
           return;
         }
         if (key.return) {
-          if (completion.activeSuggestionIndex >= 0) {
+          if (key.meta) {
+            // Alt+Enter for newline
+            buffer.newline();
```
jacob314: we should move these checks to TextBuffer.ts where the existing checks for return + ctrl inserting a newline already present. I don't think you need to resetCompletionState when a newline is added as completion state should automatically clear as there is no longer a completion available due to the newline.

packages/cli/src/ui/components/InputPrompt.tsx
jacob314: shift+enter is supported but it isn't discoverable how to enable it. see https://b.corp.google.com/issues/424007249

packages/cli/src/ui/components/InputPrompt.tsx
```diff
@@ -255,6 +259,12 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
       }
 
       // Core text editing from MultilineTextEditor's useInput
+      if (key.ctrl && input === 'j') {
```
jacob314: this is already supported the impl is in TextBuffer not this file.

packages/cli/src/ui/components/InputPrompt.tsx
```diff
@@ -255,6 +259,12 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
...
+        // Ctrl+J for newline (common terminal convention)
+        buffer.newline();
+        resetCompletionState();
+        return;
+      }
       if (key.ctrl && input === 'k') {
```
jacob314: not your fault but we should cleanup moving all of the basic buffer functionality like this that has leaked into InputPrompt moving to TextBuffer where most of the logic like this is. This file should only have changes that are truly InputPrompt specific.

packages/cli/src/ui/components/InputPrompt.tsx
deepankarsharma: Ack. Working on addressing comments.

packages/cli/src/ui/components/Tips.tsx
deepankarsharma: Created #926 to track this.

packages/cli/src/ui/components/InputPrompt.tsx
deepankarsharma: @jacob314 - thanks for the pointer to the bug. It pointed me in the right direction. I have implemented a /terminal-setup command that adds support for Shift-Enter and Ctrl-Enter.

packages/cli/src/ui/components/Tips.tsx
deepankarsharma: Your wish has been answered. Instead of doing what I was doing, I just skipped a step and went the terminal-setup route!

packages/cli/src/ui/components/InputPrompt.tsx
deepankarsharma: Did some cleanup in this regard. Let me know what you think.

deepankarsharma:
@jacob314 code is ready for your review.

NTaylorMullen:
@jacob314 if you could take another look here that'd be great! Also, we may need to cherry-pick to the release branch once merged

docs/cli/commands.md
jacob314: @miguelsolorio what do you think of `>` changing to `»` for  multiline input?

packages/cli/src/ui/components/InputPrompt.tsx
```diff
@@ -329,7 +334,7 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
         <Text
           color={shellModeActive ? Colors.AccentYellow : Colors.AccentPurple}
         >
-          {shellModeActive ? '! ' : '> '}
+          {shellModeActive ? '! ' : buffer.text.includes('\n') ? '» ' : '> '}
```
jacob314: nit: buffer provides a list of list of lines. check the length of that rather than searching for \n here.

packages/cli/src/ui/utils/terminalSetup.ts
```diff
@@ -0,0 +1,515 @@
...
+    if (!content.includes('performable:ctrl+enter=')) {
+      content += ctrlEnterLine + '\n';
+      modified = true;
+    }
+
+    if (modified) {
+      await fs.writeFile(configFile, content);
+      return {
+        success: true,
+        message: 'Added Shift+Enter and Ctrl+Enter keybindings to Ghostty.',
```
jacob314: nit: here and elsewhere maybe include the file that was modified so users know where to go in all of these cases if something went wrong.

packages/cli/src/ui/utils/terminalSetup.ts
```diff
@@ -0,0 +1,515 @@
...
+  if (!terminal) {
+    return {
+      success: false,
+      message:
+        'Could not detect terminal type. Supported terminals: Kitty, Alacritty, Ghostty, URxvt, VS Code, and Foot.',
+    };
+  }
+
+  switch (terminal) {
+    case 'kitty':
```
jacob314, line 496: just to double check, have you manually tested each of these terminals. I would prefer we err on the side of supporting fewer terminals versus have any of these accidentally mess up a users's configuration.

deepankarsharma: Yes I did check manually multiple times by blowing out the configuration on each try. But I am happy to pare down the list to what makes sense and then add terminals over time. Let me know what would be a good initial list and I will reduce to that.

packages/cli/src/ui/utils/terminalSetup.ts
deepankarsharma: Done.

packages/cli/src/ui/components/InputPrompt.tsx
deepankarsharma: Done

jacob314 (commented):
LGTM once these minor comments are addressed. Thanks for all your work here!

jacob314 (dismissed):
lgtm

deepankarsharma:
https://github.com/google-gemini/gemini-cli/actions/runs/15655712420 has been stuck for two days preventing any new Lint checks or Test checks on this branch. Opening a ticket with github support to help with this because trying to cancel the stuck action did not help. Contrary to many claims on stack overflow - closing and re-opening the PR did not work either.

mattKorwel:
Hi @deepankarsharma. I changed your pr to base against main; mainly i did this because we'll need it to go to release to be in the release but also to jump start your actions. I've configured them to work against forks now. Now you can see the build job failed with formatting errors that will need fixed.

One note (and we'll get this added to the docs you're actually the first one we're working with here) in order to run the E2E tests you'll need to add a [repo secret](https://docs.github.com/en/actions/security-for-github-actions/security-guides/using-secrets-in-github-actions#creating-secrets-for-a-repository) for `GEMINI_API_KEY` and provide a/the value you've been testing with.

Also, i assume you are intending this for the June 25th release, is that accurate? if so please tag with `June 25th Release`. 

Cheers!

deepankarsharma:
Per discussion with @jacob314 

1. We should merge this post release, so dropping the 25th Jun Release label
2. Target main 
3. Change approach to merge settings with existing config rather than overwriting existing settings and warn users if we have overwritten things. 
4. Possibly as a nice touch think about whether it makes sense to print the diff for the changes that we made, though this might be overwhelming for less technical users.

mattKorwel:
I changed this to point to main, you will need to rebase when you get a. moment.

deepankarsharma:
I will rebase on main and update this PR shortly.

scottdensmore:
@deepankarsharma can you resolve these conflicts and we can merge this into main.

deepankarsharma:
@scottdensmore Done!

siba2893:
Please for the love of god merge this PR into the next release! Waiting for this feature so much.

welovfree:
> Please for the love of god merge this PR into the next release! Waiting for this feature so much.

On Windows, by pressing `Ctrl+Enter` you can add as much new lines as you want in the input box.

WindowsTerminal_KSDWSegH4X_28-06-25

deepankarsharma:
This PR is affected by the visibility change that happened when the private gemini repo was made public. Github describes what happens on visibility changes in this way on [this page](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/what-happens-to-forks-when-a-repository-is-deleted-or-changes-visibility#changing-a-private-repository-to-a-public-repository)

```text

[Changing a private repository to a public repository](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/what-happens-to-forks-when-a-repository-is-deleted-or-changes-visibility#changing-a-private-repository-to-a-public-repository)


When a private repository is made public, all the commits in that repository, including any commits
previously pushed to private forks of that repository, will be migrated to a new public repository
network and become visible to everyone. Any previously created private forks will remain private
but will become disconnected from the original repository that was made public. Each private fork
will become a separate private repository and create its own independent network of repositories.
Any new changes made to these networks will not be accessible from the original repository that
was made public.
```

Post the gemini-cli repo becoming public my fork of the private repo was marked as an independent
repo by Github and the PR CI workflow does not seem to be working anymore. It is non obvious to me
how I can get Github to mark my repo marked as a fork again. I will close this PR and open a new one
that is based on the main branch of the gemini-cli repo.

PR #901, opened 2025-06-10
Contributing guidelines

cperry-goog:


scottdensmore (commented):
We should link this with the CONTRIBUTING.md. We should also suggest people to use Gemini to fix the issues and at the least use it to write the commit messages etc. Maybe give them an example gemini.md file that describes a good commit message / title???

cperry-goog:
> We should link this with the CONTRIBUTING.md. We should also suggest people to use Gemini to fix the issues and at the least use it to write the commit messages etc. Maybe give them an example gemini.md file that describes a good commit message / title???

I'm sorry I don't follow - this is the CONTRIBUTING.md - do you mean create a separate page and link to it from CONTRIBUTING.md?

And I can add the other bits yeah. Will ask the team about a good gemini.md file.

cperry-goog:
committing, saving todos for later

PR #907, opened 2025-06-10
fix: Change sandbox network check command for docker

tolikmalibroda:
Docker doesn't support `exists` command for checking existence of the network. Changing it to `inspect` because both docker and podman support it.


olcan (approved):
Lgtm thanks!

PR #909, opened 2025-06-10
feat: Add token stats in footer

asadm:
Adds a persistent footer line with tokens sent/received so far.
Screenshot 2025-06-11 at 9 14 24 AM


NTaylorMullen:
Yooo this is sick

@miguelsolorio from a UX perspective what do you think?

packages/cli/src/ui/components/Footer.tsx
```diff
@@ -34,63 +45,87 @@ export const Footer: React.FC<FooterProps> = ({
...
-          {' ' + (debugMessage || '--debug')}
-        </Text>
-      )}
-    </Box>
+  promptTokenCount,
+  candidatesTokenCount,
+  totalTokenCount,
+}) => {
+  const limit = tokenLimit(model);
+  const percentage = totalTokenCount / limit;
```
abhipatel12, line 43: This is where there may be some issues. For example, let's say a single turn involves multiple steps before user input is taken again. This happens occasionally, when the agent keeps moving on its plan without a rejection from the user. In this case a single "turn" can actually turn into dozens of API calls. 

The `sessionStats.currentTurn.*` stats will continue to aggregate all tokens that were inputted and outputted for that one turn (including dozens of API calls). 

**HOWEVER**, if the context gets too large, the `client.ts` may proactively compress the chat history during this single turn. This means the token counts that you are using may not be the token inputs actually being sent to the model. 

**TlDr;**  
`currentTurn` token count is not measuring the size of the context window being sent to the model. It's a cumulative sum of the tokens processed in all API requests and responses that constituted the complete turn.

asadm: addressed in latest commit. We show last response stats only.

abhipatel12: Awesome that LGTM! Thanks for fixing that!

abhipatel12 (dismissed):

abhipatel12:
LGTM from a token counting perspective! Let's just make sure the UX is solid before merging!

miguelsolorio:
> Yooo this is sick
> 
> @miguelsolorio from a UX perspective what do you think?

I like this in theory, though I have some concerns about enabling this by default. My suggestion would be to have this under a setting that persist. We've talked about the overcrowding of the footer and this makes this worse. There's also font issues if you don't powerline fonts installed, see example:

CleanShot 2025-06-11 at 10 12 21@2x

I'd also suggest changing the styles so they are more inline with the rest of the footer, and remove some of the `[` characters like:

CleanShot 2025-06-11 at 10 22 32@2x

asadm:
updated with ui improvements
Screenshot 2025-06-11 at 10 59 26 AM

asadm:
I would recommend keeping this on by default as this has been helpful and core indicator of what to expect from model. This recommendation comes from probably 1000s of sessions i have gone through with various agents (aider etc). 

On contrary, I feel sandbox and model info can just be condensed to an icon or be a "log" event in stream instead of persistent footer text.

But I am ok if that doesn't make sense to others.

abhipatel12:
One other consideration is that other similar tools like Claude code add the token count right next to the time the agent is taking to still give the user an idea of the tokens:

The footer can then just maintain a percent that reflects context window. 

Ours currently:

`⠙ One moment, optimizing humor... (esc to cancel, 20s)`

Possible option:
`⠙ One moment, optimizing humor... (esc to cancel, 20s, ↑400 ↓ 150)`

keir:
The current status bar is crowded with information that's not useful to me, also based on experience with other agentic tools. I suggest moving the sandbox mode and model display to startup, retain the directory and branch, and add in token stats.

keir:
In regards to percentage vs raw token counts in status, here's my suggestion:

- In the short term: Show the raw token count and percent.
- In the long term: Show only the percentage used, with tokens traded shown in the logs.

Why? Based on extensive use of both Gemini and other AI coding tools, the cold reality is that the performance of the models on contexts that fill the window is poor.  For Gemini, I find a notable drop in performance past 150-200K tokens. By percentage, that's only 20% of the window.

Long term, the performance of full context windows will improve. But we aren't there today, so the raw token count is useful.

asadm:
I agree, I tested out removing token count from footer. It seems I find raw count more helpful than the percentage so I wait for model to think, so I can peak at the raw count, which is ironic. 

Would it be ok if we move forward with keeping raw count in footer?

NTaylorMullen:
As tokens increase wont that jump the footer around having them on the right hand side like that?

My personal take: I don't think the tokens is super useful out of the box and here's why: It's effectively a way to monitor progress in a way that keeps you engaged that "something" is happening that's significant. I think we can get the same behavior by showing the thoughts of the model.

Alternatively, I do think its' very valuable to show the % of the context window because that's where Gemini's strengths are. Is there a version of this that only shows the % with a common font?

asadm:
@NTaylorMullen how does this version look. 
Screenshot 2025-06-12 at 12 35 32 PM

miguelsolorio:
@asadm that's looking better, I'm wondering if we should combined that text with the model, given they are somewhat related? So it would be `gemini-2.5-pro-preview-06-05 (100% context left)`? It would also match the other footer items

abhipatel12:
> @asadm that's looking better, I'm wondering if we should combined that text with the model, given they are somewhat related? So it would be `gemini-2.5-pro-preview-06-05 (100% context left)`? It would also match the other footer items

just a passing comment, is it better to have this percent as context remaining? or context used?

i.e.
`35% ctx window used`

`65% ctx window left`

asadm:
> @asadm that's looking better, I'm wondering if we should combined that text with the model, given they are somewhat related? So it would be `gemini-2.5-pro-preview-06-05 (100% context left)`? It would also match the other footer items

how about now @miguelsolorio



Screenshot 2025-06-12 at 12 56 58 PM

asadm:
@NTaylorMullen what do you think of this?

NTaylorMullen (approved):

NTaylorMullen:
> @NTaylorMullen what do you think of this?

Love how this landed ❤️

PR #911, opened 2025-06-10
/save and /resume for checkpointing conversations

sethtroisi:
this should be extended to support `/save <ID>` and `/resume <ID>` as well as keeping a <LAST> conversation by default all of that in a follow up PR.

Screenshot 2025-06-10 at 3 36 36 PM
After /resume
Screenshot 2025-06-10 at 3 36 53 PM

### DIFF
--- a/packages/cli/src/ui/hooks/slashCommandProcessor.ts
+++ b/packages/cli/src/ui/hooks/slashCommandProcessor.ts
@@ -11,10 +11,11 @@ import process from 'node:process';
 import { UseHistoryManagerReturn } from './useHistoryManager.js';
 import {
   Config,
-  MCPServerStatus,
-  getMCPServerStatus,
-  getMCPDiscoveryState,
+  Logger,
   MCPDiscoveryState,
+  MCPServerStatus,
+  getMCPDiscoveryState,
+  getMCPServerStatus,
 } from '@gemini-cli/core';
 import { Message, MessageType, HistoryItemWithoutId } from '../types.js';
 import { useSessionStats } from '../contexts/SessionContext.js';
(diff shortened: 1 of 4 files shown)


packages/cli/src/ui/hooks/slashCommandProcessor.ts
```diff
@@ -487,6 +488,58 @@ Add any other context about the problem here.
...
       },
+      {
+        name: 'save',
+        description: 'save conversation checkpoint',
+        action: async (_mainCommand, _subCommand, _args) => {
+          const logger = new Logger();
+          await logger.initialize();
+          const chat = await config?.getGeminiClient()?.getChat();
+          const history = chat?.getHistory() || [];
+          if (history.length > 0) {
```
jacob314, line 499: hit: call add message to notify the user why no history was saved if there was no history.

packages/cli/src/ui/hooks/slashCommandProcessor.ts
```diff
@@ -487,6 +488,58 @@ Add any other context about the problem here.
...
+          clearItems();
+          let i = 0;
+          const rolemap: {[key: string]: MessageType} = {
+            'user': MessageType.USER,
+            'model': MessageType.GEMINI,
+          }
+          for (const item of conversation) {
+            i += 1;
+            const text = item.parts?.filter((m) => !!m.text).map((m) => m.text).join("") || '';
+            if (i <= 2 && (text.length > 1000 || text.startsWith('Got it.'))) {
```
jacob314: lets skip the first two items rather than requiring the `Got it.` text as that might change with different models.

packages/core/src/core/logger.ts
```diff
@@ -229,9 +233,61 @@ export class Logger {
...
+      console.debug(
+        'Logger not initialized or checkpoint file path not set. Cannot save a checkpoint.',
+      );
+      return;
+    }
+
+    try {
+      await fs.writeFile(
+        this.checkpointFilePath,
+        JSON.stringify(conversation, null, 2),
```
jacob314: nit: don't pretty print. 
bonus points: add a Config option to pretty print.

packages/core/src/core/logger.ts
```diff
@@ -229,9 +233,61 @@ export class Logger {
...
+        'utf-8',
+      );
+    } catch (error) {
+      console.debug('Error writing to checkpoint file:', error);
+    }
+  }
+
+  async loadCheckpoint(): Promise<Content[]> {
+    if (!this.initialized || !this.checkpointFilePath) {
+      console.debug(
```
jacob314: nit: use console.error instead of debug?
debug is actually just for low priority info message not an error/warning.

jacob314: same applies to other console.debug in this file.

packages/core/src/core/logger.ts
```diff
@@ -92,6 +95,7 @@ export class Logger {
     this.sessionId = Math.floor(Date.now() / 1000);
     const geminiDir = path.resolve(process.cwd(), GEMINI_DIR);
     this.logFilePath = path.join(geminiDir, LOG_FILE_NAME);
+    this.checkpointFilePath = path.join(geminiDir, CHECKPOINT_FILE_NAME);
```
jacob314, line 98: do we want to override the previous checkpoint file or create 
checkpoint-1.json, checkpoint-2.json files or similar?

jacob314: optional nit

jacob314 (approved):
lgtm

packages/core/src/core/logger.ts
sethtroisi: This is coming as a direct followup.

packages/core/src/core/logger.ts
sethtroisi: Done

PR #920, opened 2025-06-11
Remove warnings for old env var names.

scidomino:
Developers have had a week to get used to the new names. And we don't want these sticking around for the release.


scottdensmore (approved):
:shipit:

Issue #805, opened 2025-06-06
FR write down my own "@" or "/" commands as defined in the GEMINI.md file


zeroasterisk:
Feature Request: Custom User-Defined Commands

  1. Overview & Scenario Context:

  Currently, users interact with Gemini through a predefined set of system commands (e.g., /help, /bug) and natural language requests. While Gemini can remember and execute multi-step
  processes described by the user (e.g., "run my 'refresh chat space overview' process"), this relies on natural language understanding for each invocation and can be verbose for frequently
  used, complex workflows.

  The core scenario is to empower users to encapsulate complex, multi-tool, or frequently repeated sequences of operations into a single, easily invokable custom command. This would be
  analogous to creating aliases or scripts in a traditional shell environment but tailored to the capabilities and interaction model of Gemini.

  2. What We Are Trying To Do (User Goal):

  We want to enable users to define their own personalized "@" or "/" commands that represent a specific workflow or a series of actions they frequently perform with Gemini. When a user
  invokes their custom command, Gemini would execute the predefined sequence of operations associated with that command.

  Example User Stories:

   * As a developer, I want to define a command like `/refresh_my_project_overview` so that Gemini automatically fetches the latest chat space summaries, recent commits, and open high-priority
     bugs for my primary project, without me having to detail these steps each time.
       * Current approach: User details each step ("list spaces with X, get logs for Y, run git status, search Jira for Z...").
       * Desired approach: User types /refresh_my_project_overview.
   * As a project manager, I want to create a command like `@team_sync_prep` so that Gemini gathers specific metrics from different sources (e.g., task management system via a custom tool, code
     repository statistics, build server status) and formats them into a summary report, which I need daily.
       * Current approach: User lists multiple data gathering and formatting instructions.
       * Desired approach: User types @team_sync_prep.
   * As a user managing multiple environments, I want to define `/deploy_to_staging web_app_v2` where `web_app_v2` is a parameter, so that Gemini executes a predefined sequence of build, test,
     and deployment tool calls for that specific application and version to the staging environment.
       * Current approach: User details each build, test, and deploy command with parameters.
       * Desired approach: User types /deploy_to_staging web_app_v2.

  Benefits:

   * Efficiency: Reduces verbosity and time for common, complex tasks.
   * Consistency: Ensures complex workflows are executed the same way every time.
   * Personalization: Allows users to tailor Gemini's operation to their specific needs and workflows.
[...]

PR #923, opened 2025-06-11
Add user-defined commands feature (#805)

deepankarsharma:
Enables users to create custom commands as markdown files in `.gemini/user-tools/` directory. Commands are invoked with `/user-` prefix and support additional parameters.

- Add UserToolsLoader to discover and parse tool markdown files
- Add UserToolProcessor to handle command execution
- Integrate user tools with slash command system and autocomplete
- Add comprehensive test coverage for all components
- Update documentation with examples and usage instructions
- Handle directory creation failures gracefully for restricted environments

Example: Create `.gemini/user-tools/find-math-book.md`:
```markdown
---
description: Get a random math book recommendation
parameters:
---

Please recommend a random mathematics book. Consider various branches of mathematics (algebra, calculus, geometry, topology, number theory, statistics, etc.) and difficulty levels (from popular math to advanced textbooks).

For your recommendation, please provide:
1. Book title and author(s)
2. What branch of mathematics it covers
3. The target audience/difficulty level
4. A brief description of what makes this book interesting or valuable
5. Why someone might want to read it

If the user provides any preferences or constraints in their additional instructions, take those into account when making your recommendation.
```

Then reload tools: `/reload-user-tools`
Run with: `/user-find-math-book` or `/user-find-math-book focusing on topology`

### DIFF
--- a/docs/cli/commands.md
+++ b/docs/cli/commands.md
@@ -122,3 +122,110 @@ The `!` prefix provides a powerful way to interact with your system's shell dire
 - **Caution for all `!` usage:** Be mindful of the commands you execute, as they have the same permissions and impact as if you ran them directly in your terminal. The Shell Mode feature does not inherently add extra sandboxing beyond what's already configured for the underlying `run_shell_command` tool.
 
 This integrated shell capability allows for seamless switching between AI-assisted tasks and direct system interaction.
+
+## User-Defined Commands (`/user-`)
+
+The Gemini CLI supports custom user-defined commands that allow you to create personalized shortcuts and workflows. These commands are stored as markdown files in the `.gemini/user-tools/` directories and can be invoked using the `/user-` prefix.
+
+### Creating User Tools
+
+1. **Location:** User tools can be stored in two locations:
+
+   - **Global tools:** `~/.gemini/user-tools/` - Available from any directory
+   - **Workspace tools:** `.gemini/user-tools/` - Relative to your current working directory
+
+   When both locations contain a tool with the same name, the workspace tool takes precedence.
+
+2. **Format:** Each tool is a markdown file (`.md`) with YAML frontmatter.
+3. **Naming:** The filename (without `.md` extension) becomes the tool name.
+
+### User Tool File Structure
+
+```markdown
+---
+description: Brief description of what this tool does
+autoSubmit: false
+---
+
+Your prompt template here. This template will be sent to Gemini as-is,
+along with any additional instructions the user provides when invoking the tool.
+```
+
+**Frontmatter Options:**
+
+- `description`: A brief description shown in autocomplete
+- `autoSubmit`: Set to `true` if the tool should auto-submit when selected from autocomplete (defaults to `false`)
+
+### Example User Tools
+
+**`.gemini/user-tools/git-log.md`**
+
+```markdown
+---
+description: Show git commit history with customizable formatting
+---
+
+Please show the git commit history for this repository. Use `git log` with appropriate formatting options to display:
+
+- Commit hash (abbreviated)
+- Author name and date
+- Commit message
+- Files changed statistics
+
+Make the output clear and easy to read. If the user provides additional instructions (like date ranges, author filters, or specific formatting), incorporate those into the git log command.
+```
+
+**`.gemini/user-tools/find-math-book.md`**
+
+```markdown
+---
+description: Get a random math book recommendation
+---
+
+Please recommend a random mathematics book. Consider various branches of mathematics (algebra, calculus, geometry, topology, number theory, statistics, etc.) and difficulty levels (from popular math to advanced textbooks).
+
+For your recommendation, please provide:
+
+1. Book title and author(s)
+2. What branch of mathematics it covers
+3. The target audience/difficulty level
+4. A brief description of what makes this book interesting or valuable
+5. Why someone might want to read it
+
+If the user provides any preferences or constraints in their additional instructions, take those into account when making your recommendation.
+```
+
+### Using User Tools
+
+- **Invocation:** Use `/user-toolname` followed by any additional instructions or context.
+- **Examples:**
+  - `/user-git-log` - Shows full git history with default formatting
+  - `/user-git-log --since="1 week ago" --author="John"` - Shows commits from last week by John
+  - `/user-git-log last 4 days` - User can use free form text and Gemini will do the right thing and show git log for last four days
+  - `/user-find-math-book` - Gets a random math book recommendation
+  - `/user-find-math-book find me a good topology book` - Gets a recommendation for a topology book
+- **Autocomplete:** Type `/user-` and press Tab to see available user tools.
+
+### Managing User Tools
+
+- **`/reload-user-tools`**
+  - **Description:** Reloads all user tools from both global (`~/.gemini/user-tools/`) and workspace (`.gemini/user-tools/`) directories.
+  - **Action:** Scans both directories and refreshes the available user tools list. Workspace tools take precedence over global tools with the same name.
+  - **Usage:** Run this command after adding, modifying, or removing user tool files.
+
+### How User Tools Work
+
+When you invoke a user tool, the tool's template is sent to Gemini along with any additional instructions you provide. For example:
+
+- `/user-git-log` sends the git-log template to Gemini
+- `/user-git-log last 3 days` sends the template plus "At the time of invoking this tool, the user provided these additional instructions: last 3 days"
+
+This approach allows Gemini to use its judgment to interpret your specific needs while following the general template of the tool.
+
+### Best Practices
+
+- Keep tool names descriptive and use hyphens for multi-word names (e.g., `git-summary`, `code-review`)
+- Provide clear descriptions in the frontmatter for better autocomplete hints
+- Test your tools after creation using the actual command syntax
+- Write clear prompts that explain what the tool does and what information it needs
+- Consider creating tools for repetitive tasks or complex multi-step workflows
--- a/packages/cli/src/ui/components/InputPrompt.tsx
+++ b/packages/cli/src/ui/components/InputPrompt.tsx
@@ -17,6 +17,7 @@ import { useCompletion } from '../hooks/useCompletion.js';
 import { isAtCommand, isSlashCommand } from '../utils/commandUtils.js';
 import { SlashCommand } from '../hooks/slashCommandProcessor.js';
 import { Config } from '@gemini-cli/core';
+import { UserTool } from '../../utils/userToolsLoader.js';
 
 export interface InputPromptProps {
   buffer: TextBuffer;
@@ -25,6 +26,7 @@ export interface InputPromptProps {
   onClearScreen: () => void;
   config: Config; // Added config for useCompletion
   slashCommands: SlashCommand[]; // Added slashCommands for useCompletion
+  userTools: Map<string, UserTool>;
   placeholder?: string;
   focus?: boolean;
   inputWidth: number;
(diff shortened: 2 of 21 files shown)


packages/cli/src/ui/components/InputPrompt.tsx
```diff
@@ -26,6 +26,10 @@ export interface InputPromptProps {
   onClearScreen: () => void;
   config: Config; // Added config for useCompletion
   slashCommands: SlashCommand[]; // Added slashCommands for useCompletion
+  userTools: Map<
+    string,
+    { name: string; description: string; parameters: string[]; content: string }
```
jacob314: define an interface mapping the type of the value in this map as it is large enough to justify that.

packages/cli/src/ui/components/InputPrompt.tsx
```diff
@@ -26,6 +26,10 @@ export interface InputPromptProps {
...
+  >; // Added userTools for useCompletion
```
jacob314: strip comments like this from the CL that indicates what Gemini CLI changed rather than describing what the code does. Gemini CLI used to really struggle with this but still sometimes adds these low value comments.

packages/cli/src/ui/components/InputPrompt.tsx
```diff
@@ -137,9 +143,26 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
       if (query.trimStart().startsWith('/')) {
         const slashIndex = query.indexOf('/');
         const base = query.substring(0, slashIndex + 1);
-        const newValue = base + selectedSuggestion.value;
-        buffer.setText(newValue);
-        handleSubmitAndClear(newValue);
+        const autoCompleteStartIndex = slashIndex + 1;
+
+        // For user tools, don't auto-submit - let the user add more instructions
```
jacob314: It is great you are tackling this problem.
Lets flow this through the system a bit more rather than hardcoding based on user-.
What you have noticed is an equally bad problem for `/history` and probably other built in tools in the future. I'd suggest having the list of completionSuggestions change from an array of strings to an array of Completion objects where completion objects indicate whether the completion should be autoSubmitted when accepted. You can then pass that arg as false for the memory completion and all user completions. Hopefully Gemini CLI can handle a lot of the busy work of flowing this data through the system for you with the appropriate prompt.

packages/cli/src/ui/components/InputPrompt.tsx
```diff
@@ -137,9 +143,26 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
...
+        if (selectedSuggestion.value.startsWith('user-')) {
+          // Replace from after the slash to the end with the suggestion + space
+          buffer.replaceRangeByOffset(
+            autoCompleteStartIndex,
+            buffer.text.length,
+            selectedSuggestion.value + ' ',
```
jacob314, line 125: love the auto added space for this case!

docs/cli/commands.md
```diff
@@ -117,3 +117,101 @@ The `!` prefix provides a powerful way to interact with your system's shell dire
 - **Caution for all `!` usage:** Be mindful of the commands you execute, as they have the same permissions and impact as if you ran them directly in your terminal. The Shell Mode feature does not inherently add extra sandboxing beyond what's already configured for the underlying `run_shell_command` tool.
 
 This integrated shell capability allows for seamless switching between AI-assisted tasks and direct system interaction.
+
+## User-Defined Commands (`/user-`)
+
+The Gemini CLI supports custom user-defined commands that allow you to create personalized shortcuts and workflows. These commands are stored as markdown files in the `.gemini/user-tools/` directory and can be invoked using the `/user-` prefix.
```
jacob314: nit: should it be directories rather than directory to make it clear that there can be both a user level as well as project level .gemini directory?

packages/cli/src/ui/hooks/useCompletion.ts
```diff
@@ -149,12 +160,44 @@ export function useCompletion(
           label: cmd.name, // Always show the main name as label
           value: cmd.name, // Value should be the main command name for execution
           description: cmd.description,
-        }))
-        .sort((a, b) => a.label.localeCompare(b.label));
+        }));
+
+      // Add user tool suggestions if the partial command starts with "user-"
```
jacob314: why not show user tool suggestions all the time? If I added a user tool I probably like it and would want to see it in the list when I type `/u`
Right now only stats shows up when I type `/u` and isn't a great match for `/u` so there is little harm in showing all the user tools.

jacob314: to be clear I aI haven't looked through the logic here carefully but it looks a bit more complex than needed. if you have reasons why this is what you want let me know and I'll review more carefully.

packages/cli/src/ui/hooks/useCompletion.ts
```diff
@@ -149,12 +160,44 @@ export function useCompletion(
...
+              description: tool.description || `User-defined tool: ${toolName}`,
+            });
+          }
+        }
+      }
 
-      setSuggestions(filteredSuggestions);
-      setShowSuggestions(filteredSuggestions.length > 0);
-      setActiveSuggestionIndex(filteredSuggestions.length > 0 ? 0 : -1);
+      // Combine and sort all suggestions
```
jacob314: optional nit: remove the comment on line 192 as the code is self documenting. (I'm assuming Gemini CLI wrote this comment). If you prefer to keep it that is fine.

packages/cli/src/ui/hooks/useGeminiStream.ts
```diff
@@ -182,6 +182,17 @@ export const useGeminiStream = (
         if (typeof slashCommandResult === 'boolean' && slashCommandResult) {
           // Command was handled, and it doesn't require a tool call from here
           return { queryToSend: null, shouldProceed: false };
+        } else if (
+          typeof slashCommandResult === 'object' &&
```
jacob314: nit: `typeof slashCommandResult == 'object'` seems a little odd to check here.
would it make sense to instead log an error and fail if the slashCommandResult is not an object from within the block?

packages/cli/src/utils/userToolProcessor.ts
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  /**
+   * Validate if the command has the required parameters
+   */
+  static validateParameters(
+    _tool: UserTool,
+    _args: string[],
+  ): { valid: boolean; message?: string } {
+    // With the new approach, we don't enforce strict parameter validation
```
jacob314: Remove the text `With the new approach,`

packages/cli/src/utils/userToolsLoader.ts
```diff
@@ -0,0 +1,172 @@
...
+  parameters: string[];
+  content: string;
+  filePath: string;
+}
+
+export class UserToolsLoader {
+  private userTools: Map<string, UserTool> = new Map();
+  private userToolsDir: string;
+
+  constructor(private config: Config) {
```
jacob314: pass in the targetDir rather than the entire Config. That will make this code more standalone and easier to test.

packages/cli/src/utils/userToolsLoader.ts
```diff
@@ -0,0 +1,172 @@
...
+          // Only log in non-test environments
+          if (process.env.NODE_ENV !== 'test' && !process.env.VITEST) {
+            console.debug('Failed to create user tools directory:', mkdirError);
+          }
+        }
+        return this.userTools;
+      }
+    } catch (existsError) {
+      // If even checking existence fails, return empty
+      if (process.env.NODE_ENV !== 'test' && !process.env.VITEST) {
```
jacob314: avoid conditional log for tests like this even for log statements. if this is a problem, fix the test so this failure doesn't occur or just allow the test to log some messages.

packages/cli/src/utils/userToolsLoader.ts
```diff
@@ -0,0 +1,172 @@
...
+    filePath: string,
+  ): UserTool {
+    // Look for metadata in the markdown frontmatter or first paragraph
+    const lines = content.split('\n');
+    let description = '';
+    let parameters: string[] = [];
+    let contentStartIndex = 0;
+
+    // Simple frontmatter parsing
+    if (lines[0] === '---') {
```
jacob314: did you try to use an existing markdown parser rather than rolling a custom one? 
If an existing one is not practical I would suggest hardening this one slightly trimming trailing whitespace from lines that must be '---'

packages/cli/src/ui/components/InputPrompt.tsx
jacob314: You might also want to allow user tools that don't specify parameters to auto-submit. I was just playing around with this adding a /code-review tool and I'm not sure I really want to specify a parameter for it.

packages/cli/src/ui/hooks/slashCommandProcessor.ts
```diff
@@ -56,8 +60,26 @@ export const useSlashCommandProcessor = (
...
+  initialUserTools?: Map<string, UserTool>,
 ) => {
   const session = useSessionStats();
+  const [userToolsLoader, setUserToolsLoader] =
+    useState<UserToolsLoader | null>(null);
+  const [userTools, setUserTools] = useState<Map<string, UserTool>>(
+    initialUserTools || new Map(),
+  );
+
+  // Initialize user tools loader
```
jacob314: this is a somewhat odd place to be loading user tools.
I'd suggest we store UserTools as part of LoadedSettings. Seems logical that these are just Settings we happen to be stored in separate files so that the UX to edit them is better.
Aligning this code to how settings are loaded would also simplify how these are loaded and be a central place to support reloading these rather than requiring a command to reload.
I think you also need the same functionality in LoadedSettings to load the tools from both the .gemini directory under the users home directory and the per project one

jacob314 (commented):
Done with a round of code review comments. Happy to GVC to discuss further if you have any questions. Overall I think this is on the right path but there are a few things I'd like to tweak before we land.

packages/cli/src/ui/hooks/useCompletion.ts
deepankarsharma: I was a bit worried that if people created a ton of user commands, they might drown out the inbuilt commands. But perhaps we can cross that bridge when we get there. I will change things to always show completion.

packages/cli/src/utils/userToolsLoader.ts
deepankarsharma: Didnt want to add a dependency since I thought it would be intrusive and am new to the project. If you think its the right thing to do I would be happy to do the following - I harden this parser for the purposes of this PR and I open a follow on issue to evaluate and incorporate a markdown parser.

packages/cli/src/ui/hooks/slashCommandProcessor.ts
deepankarsharma: Done.

packages/cli/src/ui/components/InputPrompt.tsx
deepankarsharma: Done

packages/cli/src/utils/userToolsLoader.ts
deepankarsharma: Done.

deepankarsharma:
> Done with a round of code review comments. Happy to GVC to discuss further if you have any questions. Overall I think this is on the right path but there are a few things I'd like to tweak before we land.

@jacob314, think ive addressed all your comments. PTAL. Would love to GVC next week just to touch base with you.

deepankarsharma:
@NTaylorMullen @scottdensmore - any pointers on who would be a good person to talk to for gathering requirements for this feature? Think adding support for user created tools would be super powerful, and I am happy to collaborate and iterate on this until it is in a good enough place to be merged.

jacob314:
@deepankarsharma happy to discuss this further. Please add a meeting to my calendar or discuss over chat if you have any questions.

PR #927, opened 2025-06-11
Cache oauth credentials

scidomino:
When using code assist, ensure that we have valid credentials in the cache on start up (before we jump into the sandbox). If the credentials are missing or invalid refresh them by prompting the user to login by redirecting them to an oauth url.

Currently there is no support for:

- reauthenticating when the refresh token expires
- logging in from a browser that is not on the same machine as is running the cli
- ccpa properly handling the llm calls that we make once we successfully authenticate

These will be added in subsequent PRs

### DIFF
--- a/packages/cli/src/gemini.tsx
+++ b/packages/cli/src/gemini.tsx
@@ -32,6 +32,7 @@ import {
   WebFetchTool,
   WebSearchTool,
   WriteFileTool,
+  ensureCodeAssistCredentials
   sessionId,
 } from '@gemini-cli/core';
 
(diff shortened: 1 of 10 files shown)


mattKorwel:
@scidomino how do I use code-assist for the "when using code assist" part :)

mattKorwel:
> @scidomino how do I use code-assist for the "when using code assist" part :)

Answering my own question...

This work builds on https://github.com/google-gemini/gemini-cli/pull/910  and there is a `GEMINI_CODE_ASSIST` env flag to set.

mattKorwel:
When I run with the code assist ENV var i do in fact get prompted to do the oauth dance. Trying to see now if there is a way to test that I am in fact authenticated.

mattKorwel (changes requested):
This as we discovered in testing has a bug that seems to throw you out to the command line when not in sandbox mode. @scidomino is investigating.

scidomino:
> This as we discovered in testing has a bug that seems to throw you out to the command line when not in sandbox mode. @scidomino is investigating.

We determined that this is due to ccpa requiring Dasher users (like googlers) to provide a project id. So I updated the code to log this requirement and exit.

mattKorwel (approved):
Nice! LGTM

PR #934, opened 2025-06-11
Restore checkpoint

louisjimenez:
*Holding off on review until we meet later today to discuss how related efforts interact with this set of changes*

**How to test**:
Run gemini-cli with the flag enabled `npm start -- -c true`
View checkpoints with the command `/restore`
After the model offers at least one file edit, use the restore command to restore the checkpoint `/restore name-of-file`

**The way the PR works is**

Adds support for a /restore command. Without sub-arguments the command list available checkpoints that can be restored. With a filename argument, it restores the given checkpoint. Checkpoints are created per tool call and capture the tool arguments along with the client and conversation history. This is paired with a hidden git repository that is used to capture file state with a commit per checkpoint. When a /restore command is run for a checkpoint, the file state and history is restored. The user is presented with the same UI they saw when the tool was first previewed and file state is reset. Meaning they see the tool diff and can decide to run it again or do something else.


packages/cli/src/ui/hooks/slashCommandProcessor.ts
```diff
@@ -569,12 +594,120 @@ Add any other context about the problem here.
...
         },
       },
-    ],
+    ];
+
+    if (config?.getCheckpointEnabled()) {
+      commands.push({
+        name: 'restore',
+        description:
+          'restore a tool call. This will reset the conversation and file history to the state it was in when the tool call was suggested',
```
anj-s, line 607: As discussed, we want to have restore point to a tag or filename that is a function(session id, timestamp)

packages/cli/src/ui/hooks/useGeminiStream.ts
```diff
@@ -605,6 +618,106 @@ export const useGeminiStream = (
...
+              `Failed to write restorable tool call file: ${getErrorMessage(
+                error,
+              )}`,
+            );
+          }
+        }
+      }
+    };
+    saveRestorableToolCalls();
+  }, [toolCalls, config, onDebugMessage, gitService, history, geminiClient]);
```
anj-s, line 719: What is the advantage of saving at the tool call boundary?

My thinking is that we can do this entirely in the client side and triggered at different slash command events. I think requiring it to be triggered at tool calls might add additional complexity. Wdyt?

packages/core/src/config/config.ts
```diff
@@ -291,6 +308,14 @@ export class Config {
     }
     return this.fileDiscoveryService;
   }
+
+  async getGitService(): Promise<GitService> {
+    if (!this.gitService) {
+      this.gitService = new GitService(this.targetDir);
```
anj-s, line 314: how do you enable ignoring this shadow directory?

packages/core/src/services/gitService.ts
```diff
@@ -0,0 +1,132 @@
...
+    } catch (error) {
+      if (isNodeError(error) && error.code !== 'ENOENT') {
+        throw error;
+      }
+    }
+
+    await fs.writeFile(hiddenGitIgnorePath, visibileGitIgnoreContent);
+
+    if (!visibileGitIgnoreContent.includes(historyDirName)) {
+      const updatedContent = `${visibileGitIgnoreContent}\n# Gemini CLI history directory\n${historyDirName}\n`;
```
anj-s, line 91: Do we delete this path at any point? Or should we let the user know that this will be added?

anj-s (approved):

packages/cli/src/ui/hooks/slashCommandProcessor.ts
louisjimenez: Discussed in a direct message but I can modify these names to use the existing session id.

packages/cli/src/ui/hooks/useGeminiStream.ts
louisjimenez: My thinking was to hook into the place where the model has provided the tool call data but the tool has not yet executed. That way we could capture file state, the tool call data and the history as it appears just before execution and then recreate the existing diff preview logic when we restore to that point. 

> My thinking is that we can do this entirely in the client side

What would this entail? I was thinking this is client side although it does occur after the model has provided the function calling response. Is there another place in the codebase you think this logic is more well suited to?

packages/core/src/config/config.ts
louisjimenez: We update the user's project .gitignore but in a follow up change I will change this to just right the shadow repo to the user's home directory, something like ~/.gemini-cli/history/<project_id>

packages/core/src/services/gitService.ts
louisjimenez: We don't delete the path but this will change when I move it to the home directory.

louisjimenez:
Follow up PR in progress to address feedback

PR #935, opened 2025-06-11
(telemetry) update token usage with types

jerop:
#750

image

### DIFF
--- a/packages/core/src/telemetry/metrics.ts
+++ b/packages/core/src/telemetry/metrics.ts
@@ -18,7 +18,7 @@ import {
   METRIC_TOOL_CALL_LATENCY,
   METRIC_API_REQUEST_COUNT,
   METRIC_API_REQUEST_LATENCY,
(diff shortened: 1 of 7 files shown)


packages/core/src/telemetry/metrics.ts
```diff
@@ -64,8 +64,8 @@ export function initializeMetrics(): void {
       valueType: ValueType.INT,
     },
   );
-  tokenInputCounter = meter.createCounter(METRIC_TOKEN_INPUT_COUNT, {
-    description: 'Counts the total number of input tokens sent to the API.',
+  tokenUsageCounter = meter.createCounter(METRIC_TOKEN_USAGE, {
+    description: 'Counts the total number of tokens used.',
```
anj-s, line 68: Is this not only input?

jerop: yes, changed this to track all token usage and added type as an attribute - `input`, `output`, `cache`, `thought`, `tool`

PR #936, opened 2025-06-11
Add support for local logging per session

anj-s:
* This allows us to view local locals for user/agent/tool turn
* Temporary solution till we have something we can post process from OTEL logs. 

We need this since we don't have insight into failing tasks during development, usage or evals. Users can now use this to post errors to bug reports.

### DIFF
--- a/packages/cli/src/ui/hooks/useGeminiStream.ts
+++ b/packages/cli/src/ui/hooks/useGeminiStream.ts
@@ -382,22 +382,46 @@ export const useGeminiStream = (
             break;
           case ServerGeminiEventType.ToolCallRequest:
             toolCallRequests.push(event.value);
+            await logger?.logMessage(
+              MessageSenderType.TOOL_REQUEST,
+              JSON.stringify(event.value.args),
+            );
             break;
           case ServerGeminiEventType.UserCancelled:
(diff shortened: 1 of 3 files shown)


anj-s:
fixing tests..

packages/cli/src/ui/hooks/useGeminiStream.ts
```diff
@@ -406,6 +430,7 @@ export const useGeminiStream = (
           }
         }
       }
+      await logger?.logMessage(MessageSenderType.SYSTEM, geminiMessageBuffer);
```
scidomino, line 446: There's a warning here I don't know if you can see it:

"React Hook useCallback has a missing dependency: 'logger'. Either include it or remove the dependency array"

scidomino (dismissed):

packages/cli/src/ui/hooks/useGeminiStream.ts
anj-s: it doesn't show up in the local lint checks for some reason

scidomino (approved):

sethtroisi:
This breaks `getPreviousMessages` it used to be that opening gc you could press up and get messages from your previous sessions (because they were all stored in logs.json) after this change this is broken.

PR #937, opened 2025-06-11
Modify /save and /restore to be sessions specific.

anj-s:
* We want to save and restore sessions
* These should be tied to checkpoints which is work in progress
* Each Content item should be associated with a message, role and optional checkpoint information

(still fixing tests but wanted to put this out there to start discussing this feature.)


anj-s:
Opened this PR to demonstrate /sessions. @louisjimenez will be working on this in this checkpoints PR post today morning's discussions.

PR #938, opened 2025-06-11
Improve the performance of filename completion over large repositories.

dewitt:
This change improves the performance of @-completion by speeding up the filename discovery by using fast-glob.

Consequences of this change / differences from current behavior include:

- Directory names alone no longer match (e.g., "@testing" now will not return the directory name "testing/"
   in the completions, but would still return "lib/testing.py")
- Searches are breadth-first
- Notably faster completions.

Note: We preserve the existing case-insensitivity on the glob searches, which is a matter of taste, but I believe
is the right choice.

**Performance testing methodology:** (basic, but it is representative of real-world usage)

I generated random test files and directories, with a branching factor of 10 files and 10 directories per directory, to a depth of 5, which is 122221 total, consistent with a moderate- to large sized project codebase.

I then placed a single `haystack.txt` file at the bottom of the hierarchy (5 levels deep, hidden in O(10k) files):

At HEAD (commit e02a035), typing `"@hay"` finds the `haystack.txt` file in **4.46** seconds.

After this change (commit aa6bf79): typing `"@hay"` finds the `haystack.txt` file in **0.34 seconds**.

Tested on a Macbook Pro M1 with 16GB RAM using the internal SSD.

Will run future tests on different hardware. (Generating the test data takes the bulk of the time.)


dewitt:
TODO before reviewing: Capture baseline performance numbers on very large directories (deeply nested, large branching factor, etc) and compare with the glob-based results.  Wanted to upload the branch for ease of running reproducible tests across environments.

dewitt:
> TODO before reviewing: Capture baseline performance numbers on very large directories

This is done. Updated the CL description above with testing data. Copying here for visibility:

**Performance testing methodology:** (basic, but it is representative of real-world usage)

I generated random test files and directories, with a branching factor of 10 files and 10 directories per directory, to a depth of 5, which is 122221 total, consistent with a moderate- to large sized project codebase.

I then placed a single `haystack.txt` file at the bottom of the hierarchy (5 levels deep, hidden in O(10k) files):

At HEAD (commit e02a035), typing `"@hay"` finds the `haystack.txt` file in **4.46** seconds.

After this change (commit aa6bf79): typing `"@hay"` finds the `haystack.txt` file in **0.34 seconds**.

Tested on a Macbook Pro M1 with 16GB RAM using the internal SSD.

Will run future tests on different hardware. (Generating the test data takes the bulk of the time.)

dewitt:
And in case we ever need it: [generate_test_files.sh](https://gist.github.com/dewitt/a70815d8a509bdf1732489ed54aa343f)

Took nearly an hour to run on my MBP at depth 5. Worth optimizing if anyone is ever going to do that again.

packages/cli/src/ui/hooks/useCompletion.integration.test.ts
```diff
@@ -42,6 +42,7 @@ describe('useCompletion git-aware filtering integration', () => {
       shouldIgnoreFile: vi.fn(),
       filterFiles: vi.fn(),
       getIgnoreInfo: vi.fn(() => ({ gitIgnored: [], customIgnored: [] })),
+      glob: vi.fn().mockResolvedValue([]),
```
jacob314, line 45: can you add an integration test that triggers the glob functionality?

packages/core/src/utils/gitIgnoreParser.ts
```diff
@@ -6,7 +6,9 @@
 
 import * as fs from 'fs/promises';
 import * as path from 'path';
-import ignore, { Ignore } from 'ignore';
+import ignore from 'ignore';
+
+type Ignore = ReturnType<typeof ignore>;
```
jacob314: the previous code is a little more idiomatic. Any reason for this change?

jacob314 (dismissed):
lgtm

packages/core/src/utils/gitIgnoreParser.ts
dewitt: Gemini-generated code when we needed the `type`. Addressed in c34b09a.

packages/cli/src/ui/hooks/useCompletion.integration.test.ts
dewitt: Good catch! Lost along the way. Regenerated and added back in c1171e0.

jacob314 (approved):
lgtm

PR #939, opened 2025-06-11
do not auto-enable container sandboxing (fixing recently introduced bug)

olcan:


scidomino (commented):
Oops! I didn't realize that this was intentional. My bad. Thanks for fixing it.

PR #944, opened 2025-06-11
Extensibility: Gemini.md files

mattKorwel:
Implements #943


packages/core/src/utils/memoryDiscovery.test.ts
```diff
@@ -532,6 +532,107 @@ describe('loadServerHierarchicalMemory', () => {
...
+    vi.spyOn(fsSync, 'existsSync').mockImplementation((p) => {
+      return p === workspaceExtensionsPath || p === homeExtensionsPath;
+    });
+
+    mockFs.stat.mockImplementation(async (p) => {
+      if (p === path.join(PROJECT_ROOT, '.git')) {
+        return { isDirectory: () => true } as Stats;
+      }
+      throw new Error('File not found');
+    });
```
scidomino: I'm not a fan of mocking the filesystem. Can we use tmp directories like we do in extension.test.ts?

Feel free to ignore if it's too time consuming.

packages/core/src/utils/memoryDiscovery.ts
```diff
@@ -195,6 +195,37 @@ async function getGeminiMdFilePathsInternal(
     }
   }
 
+  // Search for GEMINI.md in extensions directories
```
scidomino: My concern here is that the extension.ts code has logic for working dir extensions overriding home dir extensions with the same name. This doesn't seem to cover that. It's also a little gross that the extension code is now half in extension.ts and have here. Could we consolidate? I think maybe instead of returning ExtensionConfigs, extension.ts should return a objects of type Extension, a class delcared there. Then we can pipe the Extensions down to the memory tool so it can use it.

mattKorwel: Yep, totally agree!

mattKorwel:
wow. Gemini Cli deleted my branch both local and remote ;)

mattKorwel:
Screenshot 2025-06-11 12 10 10 PM


image

mattKorwel:
Restructured the code to better follow our patterns. This was the first time gemini failed me ;). Now we read the contextFilePath respecting the JSON -> then the existence of a file. And we ensure those paths get sent to the Gemini.md loading look up. Screen shots of it working and loading Gemini.md above.

packages/cli/src/config/extension.ts
```diff
@@ -74,6 +74,22 @@ function loadExtensionsFromDir(dir: string): ExtensionConfig[] {
...
         continue;
       }
+
+      if (extensionConfig.contextFileName) {
+        const contextFilePath = path.join(
+          extensionDir,
+          extensionConfig.contextFileName,
+        );
+        if (fs.existsSync(contextFilePath)) {
+          extensionConfig.contextFileName = contextFilePath;
```
scidomino, line 84: I am not a fan of replacing the contextFileName with the filepath. It's confusing.

I suggest separating the ExtensionConfig (which matches gemini-extension.ts) from the Extension (a new object returned from locaExtensionsFromDir()) That way there's no ambiguity.

mattKorwel: What's the interface for exposing this to the extension creators?

I think its confusing to have a whole separate object.... Like what do we name that "ExtenstionConfigWithFullPathForInternaluseonly"?

I would say we should actually just make it configFilePath in case a user needs to store it in a sub dir. Then we would just be expanding the path, not using a value that doesn't match the name?

mattKorwel: PS. Sorry I put on auto merge, but we don't have required reviewers yet so it automerged after the build.

PR #948, opened 2025-06-11
Edit pass of docs/deployment.md

starsandskies:
Make a pass through the docs/deployment.md file to hopefully improve readability and completeness

Notably, I've proposed axing the entire addendum section since it seems like a discussion that's more general than - and beyond the scope of - the Gemini CLI. Plus, it covers something users don't really have control of (the user choice is really about whether they want to use the early-access version of the CLI, IIUC)


docs/deployment.md
```diff
@@ -1,18 +1,18 @@
 # Gemini CLI Execution and Deployment
 
-This document outlines the various methods for running the Gemini CLI and the deployment architecture that supports them. Understanding these options is crucial for both users and developers.
+This document describes how to run the Gemini CLI and explains the deployment architecture that the Gemini CLI uses.
 
-## How to Run the Gemini CLI
+## Running the Gemini CLI
```
NTaylorMullen:
```suggestion
## Running Gemini CLI
```

LLM's love to put "the Gemini CLI" what do you think about striking that representation everywhere for `Gemini CLI`?

docs/deployment.md
```diff
@@ -75,97 +71,46 @@ Contributors to the project will want to run the CLI directly from the source co
 
 ---
 
-### 4. Running the Latest Version from GitHub (Early Access)
+### 4. Running the latest early-access version from GitHub
```
NTaylorMullen: I actually think both of these variants are wrong, neither should have said "early access". They're really executing the latest commit from GitHub

docs/deployment.md
```diff
@@ -75,97 +71,46 @@ Contributors to the project will want to run the CLI directly from the source co
...
 
-- **Cons:**
-  - **Dependency Hell:** The primary drawback. It can lead to complex dependency resolution issues, version conflicts, and a large `node_modules` directory.
-  - **Slower `npx` Install:** For `npx`, resolving and installing the full dependency tree can be significantly slower than downloading a single pre-packaged file.
-  - **"Works on my machine" issues:** Slight differences in dependency trees between environments can lead to subtle bugs.
+**Docker sandbox image**
 
-**Build & Distribution Considerations**
+The Docker-based execution method is supported by the `gemini-cli-sandbox` container image. This image is published to a container registry and contains a pre-installed, global version of the Gemini CLI. The `scripts/prepare-cli-packagejson.js` script dynamically injects the URI of this image into the CLI's `package.json` before publishing, so the CLI knows which image to pull when the `--sandbox` flag is used.
```
NTaylorMullen: @KeijiBranshi could you advise while @olcan's out?

NTaylorMullen (dismissed):

docs/deployment.md
starsandskies: Based on the consensus I saw in the group chat, this SGTM. Updated through this doc, and will continue to do so moving forward.

docs/deployment.md
starsandskies: Pretty sure I'm following, and I've updated the document accordingly (there's also some mention of this below that I updated)

docs/deployment.md
KeijiBranshi: this blurb is correct

docs/deployment.md
```diff
@@ -75,97 +71,46 @@ Contributors to the project will want to run the CLI directly from the source co
...
-
-1.  **Builds** the NPM packages using `tsc`.
-2.  **Updates** the CLI's `package.json` with the Docker image URI.
-3.  **Builds and tags** the `gemini-cli-sandbox` Docker image.
-4.  **Pushes** the Docker image to the container registry.
-5.  **Publishes** the NPM packages to the artifact registry.
-
----
-
-### Addendum: Bundling vs. Transpiling for Distribution
```
KeijiBranshi, line 126: suggestion: can we at least paste this section somewhere like a buganizer ticket attached to this PR? I think it's good to have a paper-trail of decisions. The whole reason we added this section was to provide an archeological record for onboarding contributors.

KeijiBranshi (approved):
LGTM. Added one suggestion for the removed addendum

docs/deployment.md
starsandskies: I've added this content into b/425414345, but let me know if there's something else you had in mind

PR #949, opened 2025-06-11
clean chat history before /resume

sethtroisi:


jacob314 (approved):
lgtm

would be great to have a test for this if you can coach Gemini CLI to create a really simple test that verifies that the chat history has been cleared.

PR #950, opened 2025-06-11
Fix noisy tests

sethtroisi:
Removes 344 lines of output from tests! This was causing tearing, jank, and flickering for me.

### DIFF
--- a/packages/cli/src/ui/hooks/useLoadingIndicator.test.ts
+++ b/packages/cli/src/ui/hooks/useLoadingIndicator.test.ts
@@ -16,12 +16,11 @@ import {
 describe('useLoadingIndicator', () => {
   beforeEach(() => {
     vi.useFakeTimers();
-    vi.clearAllMocks();
   });
 
   afterEach(() => {
-    vi.runOnlyPendingTimers();
     vi.useRealTimers(); // Restore real timers after each test
+    act(() => vi.runOnlyPendingTimers);
   });
 
   it('should initialize with default values when Idle', () => {


packages/cli/src/ui/hooks/useLoadingIndicator.test.ts
```diff
@@ -16,11 +16,9 @@ import {
 describe('useLoadingIndicator', () => {
   beforeEach(() => {
     vi.useFakeTimers();
-    vi.clearAllMocks();
   });
 
   afterEach(() => {
-    vi.runOnlyPendingTimers();
```
jacob314, line 23: as far as I can tell this is a best practice using viTest. What issue are we seeing? perhaps there is another way to fix the issue while still running the pending timers.
```
The primary goal of using vi.runOnlyPendingTimers() in afterEach is to flush out any pending timers that were set during a test's execution before restoring the real timers with vi.useRealTimers(). This is crucial for maintaining test isolation, a fundamental principle of good testing practices.
```

packages/cli/src/ui/hooks/useLoadingIndicator.test.ts
```diff
@@ -16,11 +16,9 @@ import {
 describe('useLoadingIndicator', () => {
   beforeEach(() => {
     vi.useFakeTimers();
-    vi.clearAllMocks();
```
jacob314, line 19: shot in the dark: should you move vi.clearAllMocks up a line rather than deleting it? 
calling vi.clearAllMocks() is a best practice. Also curious what is going wrong here.

sethtroisi: There are no mocks in this file

packages/cli/src/ui/hooks/useLoadingIndicator.test.ts
sethtroisi: discussed offline, added back wrapped in `act`

jacob314:
lgtm

PR #952, opened 2025-06-11
telemetry: expand cli configuration event

jerop:
Adds the following attributes to the  event:
- `embedding_model`
- `api_key_enabled`
- `code_assist_enabled`
- `vertex_enabled`
- `debug_mode`
- `mcp_servers`

This additional data will provide more insight into user configurations.

Example event in GCP logs explorer:
image

#750

### DIFF
--- a/docs/core/telemetry.md
+++ b/docs/core/telemetry.md
@@ -262,13 +262,18 @@ These are timestamped records of specific events.
 
   - **Attributes**:
     - `model` (string)
+    - `embedding_model` (string)
     - `sandbox_enabled` (boolean)
     - `core_tools_enabled` (string)
     - `approval_mode` (string)
+    - `api_key_enabled` (boolean)
     - `vertex_ai_enabled` (boolean)
+    - `code_assist_enabled` (boolean)
     - `log_user_prompts_enabled` (boolean)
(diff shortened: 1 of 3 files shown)


jerop:
cc @anj-s

docs/core/telemetry.md
```diff
@@ -262,13 +262,18 @@ These are timestamped records of specific events.
 
   - **Attributes**:
     - `model` (string)
+    - `embedding_model` (string)
```
anj-s, line 265: we don't use this yet so it might be confusing. I would recommend removing it.

docs/core/telemetry.md
```diff
@@ -262,13 +262,18 @@ These are timestamped records of specific events.
 
   - **Attributes**:
     - `model` (string)
```
anj-s, line 264: How can we handle multiple models that we are using right now?

jerop: this is recording an event, so each event would have the model that was used in that particular instance

docs/core/telemetry.md
jerop: was reading this from the configuration - maybe we should remove it from configuration as well too?

PR #955, opened 2025-06-11
Improvements to CLI's ability to perform refactoring. Includes additions to the system prompt and GEMINI.md.

mboshernitsan:


mboshernitsan:
@NTaylorMullen , @anj-s curious about your thoughts on these changes.  Thanks!

GEMINI.md
```diff
@@ -58,6 +58,24 @@ This project uses **Vitest** as its primary testing framework. When writing test
 - When adding tests, first examine existing tests to understand and conform to established conventions.
 - Pay close attention to the mocks at the top of existing test files; they reveal critical dependencies and how they are managed in a test environment.
 
+## Refactoring Guidance
+
+### Common Patterns
+
+When refactoring, prioritize these common patterns to improve the codebase:
+
+- **Consolidate State with `useReducer`**: If a component or hook uses three or more `useState` calls for related data, refactor them into a single `useReducer`. This centralizes state logic and is the preferred pattern for managing complex state.
```
NTaylorMullen: cc @KeijiBranshi for these (he's the expert)

GEMINI.md
```diff
@@ -58,6 +58,24 @@ This project uses **Vitest** as its primary testing framework. When writing test
...
+- **Extract Custom Hooks**: If component logic (especially `useEffect` or complex state) is repeated across multiple components, extract it into a custom hook (e.g., `useDataFetching`, `useWindowSize`).
+- **Co-locate Logic**: If a utility function or type is only used by a single module, move it inside that module file instead of keeping it in a shared `utils` or `types` folder. Export it only if necessary for testing.
+- **Replace Prop Drilling with Context**: For state that needs to be passed down through 3 or more layers of components, refactor to use React Context. Use this judiciously; for one or two layers, prop passing is still preferred.
+
+### What to Avoid
+
+- **DO NOT perform large-scale, automated renaming**: Avoid renaming core components, functions, or widely used types without explicit user confirmation. Renaming has a high risk of breaking imports or dynamic references.
```
NTaylorMullen: While it can break imports or dynamic references I feel like tests are our counter measure to this

GEMINI.md
```diff
@@ -58,6 +58,24 @@ This project uses **Vitest** as its primary testing framework. When writing test
...
+### Common Patterns
+
+When refactoring, prioritize these common patterns to improve the codebase:
+
+- **Consolidate State with `useReducer`**: If a component or hook uses three or more `useState` calls for related data, refactor them into a single `useReducer`. This centralizes state logic and is the preferred pattern for managing complex state.
+- **Extract Custom Hooks**: If component logic (especially `useEffect` or complex state) is repeated across multiple components, extract it into a custom hook (e.g., `useDataFetching`, `useWindowSize`).
+- **Co-locate Logic**: If a utility function or type is only used by a single module, move it inside that module file instead of keeping it in a shared `utils` or `types` folder. Export it only if necessary for testing.
+- **Replace Prop Drilling with Context**: For state that needs to be passed down through 3 or more layers of components, refactor to use React Context. Use this judiciously; for one or two layers, prop passing is still preferred.
+
+### What to Avoid
```
NTaylorMullen: I'm not convinced these are needed. While I can understand what you're trying to accomplish at the same time I feel that the LLM can easily misinterpret these bits to cause friction for deving

packages/core/src/core/prompts.ts
```diff
@@ -61,6 +61,20 @@ When requested to perform tasks like fixing bugs, adding features, refactoring,
 4. **Verify (Tests):** If applicable and feasible, verify the changes using the project's testing procedures. Identify the correct test commands and frameworks by examining 'README' files, build/package configuration (e.g., 'package.json'), or existing test execution patterns. NEVER assume standard test commands.
 5. **Verify (Standards):** VERY IMPORTANT: After making code changes, execute the project-specific build, linting and type-checking commands (e.g., 'tsc', 'npm run lint', 'ruff check .') that you have identified for this project (or obtained from the user). This ensures code quality and adherence to standards. If unsure about these commands, you can ask the user if they'd like you to run them and if so how to.
 
+## Refactoring Workflow
```
NTaylorMullen, line 65: I love what you have here and think it's better than the 1st primary workflow to a fair extent. Part of me wonders, why have a separate thing than Software Engineering tasks? Could we merge what you have and what's in the Software Engineering tasks but leave it as "Software Engineering Tasks"? I feel as though "refactoring" fits into that category.

I think in the original software engineering tasks workflow the callout of the edit/shell tools is important still BUT pretty much everything else you have here is better

packages/core/src/core/prompts.ts
```diff
@@ -184,18 +188,36 @@ model: [tool_call: ${ShellTool.Name} for 'node server.js &' because it must run
...
+(After analysis)
+'requests' is available in 'requirements.txt'. The current code uses 'urllib.request' which can be replaced directly.
+
+Here is my proposed plan:
+1.  Replace the 'urllib' calls in 'src/auth.py' with the equivalent 'requests' calls.
+2.  Add appropriate error handling for network issues using 'try...except requests.exceptions.RequestException'.
+3.  Remove the now-unused 'import urllib' statement.
+4.  After making the changes, I will run the project's linter and tests (e.g., 'ruff check .' and 'pytest') to verify that the refactoring did not introduce any regressions.
+
+Do you approve this plan?
```
NTaylorMullen: We may want to change the verbiage here a little bit only to influence the tone the model takes with the user. This might be too formal

packages/core/src/core/prompts.ts
```diff
@@ -184,18 +188,36 @@ model: [tool_call: ${ShellTool.Name} for 'node server.js &' because it must run
...
+(After user approval)
```
NTaylorMullen:
```suggestion
user: Yes
model:
```

packages/core/src/core/prompts.ts
```diff
@@ -184,18 +188,36 @@ model: [tool_call: ${ShellTool.Name} for 'node server.js &' because it must run
...
+Plan approved. I will now refactor 'src/auth.py'.
```
NTaylorMullen: Ditto about model tone

packages/core/src/core/prompts.ts
```diff
@@ -50,6 +49,7 @@ You are an interactive CLI agent specializing in software engineering tasks. You
 - **Proactiveness:** Fulfill the user's request thoroughly, including reasonable, directly implied follow-up actions.
 - **Confirm Ambiguity/Expansion:** Do not take significant actions beyond the clear scope of the request without confirming with the user. If asked *how* to do something, explain first, don't just do it.
 - **Explaining Changes:** After completing a code modification or file operation *do not* provide summaries unless asked.
+- **Preserve Functionality:** When refactoring, your primary goal is to restructure existing code *without changing its external behavior*. Do not add, alter, or remove functionality unless the refactoring goal explicitly requires it (e.g., "refactor to use a new API which returns different data").
```
NTaylorMullen: The example at the end is a bit hard for me to parse as a human I'm not sure it will contextually parse for a model either

NTaylorMullen (approved):
Solid changes!

packages/core/src/core/prompts.ts
mboshernitsan: This is a great observation.  The reason I kept the separate is two fold:
1) I was attempting to make smaller, additive changes to the prompt.
2) In general Refactoring has a fundamentally different goal and risk profile than other software engineering tasks and a different definition of success (make something new vs. preserve the existing behavior).

I suggest we keep them separate for now, but consolidate in the future if it seems like calling out refactoring separately confuses the model.

GEMINI.md
```diff
@@ -58,6 +58,24 @@ This project uses **Vitest** as its primary testing framework. When writing test
...
+
+- **DO NOT perform large-scale, automated renaming**: Avoid renaming core components, functions, or widely used types without explicit user confirmation. Renaming has a high risk of breaking imports or dynamic references.
+- **AVOID changing public APIs without cause**: Do not alter the signature (parameters, return values) of an exported function or a component's props unless that is the explicit goal of the refactoring task.
```
anj-s: Do you feel that the agent is actually tending towards these behaviors?

packages/core/src/core/prompts.ts
```diff
@@ -61,6 +61,20 @@ When requested to perform tasks like fixing bugs, adding features, refactoring,
...
 
+## Refactoring Workflow
+When asked to refactor code, follow this specialized sequence to ensure safety and correctness:
+1.  **Clarify the Goal:** If the user's request is ambiguous (e.g., "refactor this file"), ask for the specific goal. Is it to improve readability, enhance performance, reduce complexity, adhere to a new pattern, or something else?
+2.  **Analyze Scope and Impact:** Use '${GlobTool.Name}' and '${GrepTool.Name}' to identify not just the target code, but also where it is used and what other parts of the system it might affect. State your understanding of the scope.
+3.  **Propose a Detailed Plan:** Formulate a step-by-step plan. For each step, specify the change to be made and the verification that will follow. The plan must include a final verification step using the project's tests, linter, and type checker. Present this plan to the user for approval before proceeding.
```
anj-s: I would also make a change that we should update this plan as we execute on it and based on the output. Static plans rarely work over multiple turns

GEMINI.md
mboshernitsan: Had a chat with Taylor and we decide that it's best to revert changes to GEMINI.md

packages/core/src/core/prompts.ts
mboshernitsan: That's a great idea, let me add that in the next commit.

mboshernitsan: PTAL at the latest.

packages/core/src/core/prompts.ts
mboshernitsan: This is fair. I removed the example at the end because I think it is superfluous.

packages/core/src/core/prompts.ts
```diff
@@ -184,18 +194,39 @@ model: [tool_call: ${ShellTool.Name} for 'node server.js &' because it must run
...
+
+Here's the plan:
+1.  Replace the 'urllib' calls with 'requests'.
+2.  Add proper 'try...except' error handling for the new network calls.
+3.  Remove the old 'urllib' import.
+4.  Run the project's linter and tests to verify the changes.
+
+Should I proceed?
+user: Yes
+[tool_call: ${WriteFileTool.Name} or ${EditTool.Name} to apply the refactoring to 'src/auth.py']
```
NTaylorMullen, line 218:
```suggestion
model:
[tool_call: ${WriteFileTool.Name} or ${EditTool.Name} to apply the refactoring to 'src/auth.py']
```

packages/core/src/core/prompts.ts
```diff
@@ -184,18 +194,39 @@ model: [tool_call: ${ShellTool.Name} for 'node server.js &' because it must run
...
+Refactoring complete. Running verification...
+[tool_call: ${ShellTool.Name} for 'ruff check src/auth.py && pytest']
+(After verification passes)
+All checks passed. This is a stable checkpoint.
+</example>
+
+${(function () {
```
NTaylorMullen, line 223: This bit should be in the </example> shouldn't it?

packages/core/src/core/prompts.ts
```diff
@@ -184,18 +194,39 @@ model: [tool_call: ${ShellTool.Name} for 'node server.js &' because it must run
...
+  if (isGitRepository(process.cwd())) {
+    return `I can commit this now with the message "refactor(auth): migrate from urllib to requests". Sound good?`;
```
NTaylorMullen: Gotta be careful about wording / tone for the model. By sharing just the title the model will start to only provide single sentence commit messages which we want it to do a bit more.

```suggestion
    return `Would you like me to write a commit message and commit these changes? `;
```

NTaylorMullen (approved):

packages/core/src/core/prompts.ts
mboshernitsan: Good catch.  Fixed!

packages/core/src/core/prompts.ts
mboshernitsan: Yup.  Too much cutting and pasting.  Fixed.

packages/core/src/core/prompts.ts
mboshernitsan: Good point! Thank you!

PR #957, opened 2025-06-11
add excludeTools flag to settings.json config

JingboWang1997-1:
This new flag enables certain tools to be removed based on settings.json config.

For example, we wouldn't want to do `glob` to find files in 1P cases.


NTaylorMullen:
Thank you @JingboWang1997-1 !

PR #960, opened 2025-06-11
Telemetry: Improve clarity of user prompt event

jerop:
Renames the `prompt_char_count` attribute to `prompt_length` in the user prompt telemetry event for better clarity and consistency.

This also updates the documentation and tests to reflect the new attribute name.

#750


docs/core/telemetry.md
```diff
@@ -278,7 +278,7 @@ These are timestamped records of specific events.
 - `gemini_cli.user_prompt`: Fired when a user submits a prompt.
 
   - **Attributes**:
-    - `prompt_char_count`
+    - `prompt_length`
```
anj-s, line 281: just curious, what does this particular metric help us track? If we are tracking "prompt", would that not be sufficient?

anj-s (dismissed):

docs/core/telemetry.md
jerop: it can help us understand latency issues e.g. it would be interesting to compare prompt length vs time to first token

it would also be interesting to measure the impact of user prompt length vs the quality of response or success rate

finally, if the user disables logging the prompt itself, we can just the prompt length as the only signal we have for the above - this is why it's separate from the prompt

jerop:
CI here got stuck after approvals was required - opening #967

PR #961, opened 2025-06-11
Reduce coupling between core and cli packages

mboshernitsan:
Reduce coupling between core and cli packages by moving all environment access (file system, process environment) from core to CLI and injecting it into core.  Added new tests prior to refactor to increase test coverage and improve confidence in the refactor.


scidomino (commented):
There are a few "defined but not used" errors. Fix those and then I'll approve.

mboshernitsan:
> There are a few "defined but not used" errors. Fix those and then I'll approve.

Should be good now.

packages/cli/src/config/config.ts
```diff
@@ -260,3 +271,39 @@ async function createContentGeneratorConfig(
 
   return config;
 }
+
+function findEnvFile(startDir: string): string | null {
```
NTaylorMullen, line 274: Extensions need to do this too we may want to unify this logic

packages/core/src/config/config.ts
```diff
@@ -79,10 +76,13 @@ export interface ConfigParameters {
   accessibility?: AccessibilitySettings;
   telemetry?: boolean;
   telemetryLogUserPromptsEnabled?: boolean;
+  telemetryOtlpEndpoint?: string;
   fileFilteringRespectGitIgnore?: boolean;
   fileFilteringAllowBuildArtifacts?: boolean;
   enableModifyWithExternalEditors?: boolean;
   checkpoint?: boolean;
+  proxy?: string;
+  cwd?: string;
```
NTaylorMullen: Is this ever actually `undefined`?

NTaylorMullen (dismissed):
Nice cleanup!

scidomino (dismissed):

packages/cli/src/config/config.ts
mboshernitsan: Makes sense, let's do this in a follow up PR?

packages/core/src/config/config.ts
mboshernitsan: CWD should always be defined; fixed this in the latest commit.  Having proxy be undefined seems reasonable?

scidomino (dismissed):

mattKorwel (approved):
LGTM. I did pull this and run it to test since its allready been approved, fyi.

packages/core/src/config/config.ts
NTaylorMullen: Poifect

PR #962, opened 2025-06-11
Fixed @ file content not being added or sent to server

sethtroisi:
Paired with jacob.

/stats now shows large number of tokens.

### DIFF
--- a/packages/cli/src/ui/hooks/atCommandProcessor.ts
+++ b/packages/cli/src/ui/hooks/atCommandProcessor.ts
@@ -367,7 +367,7 @@ export async function handleAtCommand({
     };
 
     if (Array.isArray(result.llmContent)) {
-      const fileContentRegex = /^--- (.*?) ---\n\n([\s\S]*?)$/;
+      const fileContentRegex = /^--- (.*?) ---\n\n([\s\S]*?)\n\n$/;
       processedQueryParts.push({
         text: '\n--- Content from referenced files ---',
       });
--- a/packages/cli/src/ui/hooks/atCommandProcessor.ts
+++ b/packages/cli/src/ui/hooks/atCommandProcessor.ts
@@ -366,38 +366,22 @@ export async function handleAtCommand({
       confirmationDetails: undefined,
     };
 
-    if (
-      result.llmContent &&
-      typeof result.llmContent === 'string' &&
-      result.llmContent.trim() !== ''
-    ) {
+    if (Array.isArray(result.llmContent)) {
+      const fileContentRegex = /^--- (.*?) ---\n\n([\s\S]*?)$/;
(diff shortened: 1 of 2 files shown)


allenhutchison (dismissed):

packages/cli/src/ui/hooks/atCommandProcessor.ts
```diff
@@ -401,6 +401,14 @@ export async function handleAtCommand({
         }
       }
       processedQueryParts.push({ text: '\n--- End of content ---' });
+    } else if (Array.isArray(result.llmContent)) {
```
NTaylorMullen: Oh wow so we fully weren't showing the content. For some reason I thought it had to do with how we were representing it in history. Good catch!

Would be great to add tests for this case if we could!

NTaylorMullen (dismissed):

packages/cli/src/ui/hooks/atCommandProcessor.ts
```diff
@@ -366,38 +366,22 @@ export async function handleAtCommand({
...
-      for (const requestedSpec of pathSpecsToRead) {
-        if (!foundContentForSpecs.has(requestedSpec)) {
-          onDebugMessage(
-            `Content for @${requestedSpec} was expected but not found in read_many_files output.`,
-          );
-          // Optionally add a note about missing content for this spec
-          // processedQueryParts.push({ text: `\nContent for @${requestedSpec} not found or empty.\n` });
+      for (const part of result.llmContent) {
+        if (typeof part === 'string') {
+          const match = fileContentRegex.exec(part);
```
jacob314, line 376: if there isn't a match still add to the list of processedQueryParts. Same if the part is not a string. this is important for images in the file contents.

packages/cli/src/ui/hooks/atCommandProcessor.ts
```diff
@@ -366,38 +366,22 @@ export async function handleAtCommand({
       confirmationDetails: undefined,
     };
 
-    if (
-      result.llmContent &&
-      typeof result.llmContent === 'string' &&
-      result.llmContent.trim() !== ''
-    ) {
+    if (Array.isArray(result.llmContent)) {
```
jacob314, line 369: handle if we get a single string instead of an array

sethtroisi: I'm testing with an image in a follow up PR

packages/cli/src/ui/hooks/atCommandProcessor.ts
sethtroisi: Working on this in a follow up.

jacob314 (approved):
lgtm

PR #966, opened 2025-06-12
telemetry: include user decisions in tool call logs

jerop:
Add the user's decision (accept, reject, modify) to tool call telemetry to better understand tool failure. The decision provides crucial context to the `success` metric, particularly allow us to drill into tool failures. 

Also prettify the arguments json.

Example: 
image

#750


packages/core/src/core/coreToolScheduler.ts
```diff
@@ -25,20 +25,23 @@ export type ValidatingToolCall = {
...
   startTime?: number;
+  decision?: 'accept' | 'reject' | 'modify';
 };
 
 export type ScheduledToolCall = {
   status: 'scheduled';
   request: ToolCallRequestInfo;
   tool: Tool;
   startTime?: number;
+  decision?: 'accept' | 'reject' | 'modify';
```
NTaylorMullen: Today these are only understood by the telemetry system. How do you feel about passing the outcome here and having the logger translate to these?

packages/core/src/telemetry/loggers.ts
```diff
@@ -111,7 +111,7 @@ export function logToolCall(
     ...event,
```
NTaylorMullen, line 139: non-blocking: Food for thought, currently the UI is the one calling `logToolCall`; however, the `core` is where the logic is contained.  If you were to lift the calling of `logToolCall` event into the `coreToolSchedule` from the reactToolScheduler you wouldn't have to propagate all of the decisions across the ui/core boundary. It would also enable automatic telemetry from consuming extensions like the GCA VSCode extension

packages/core/src/core/coreToolScheduler.ts
jerop: ~which outcome? success?~

discussed offline - `ToolConfirmationOutcome`

packages/core/src/telemetry/loggers.ts
jerop: makes sense - removed tool calling logic from UI

jerop: fixed in https://github.com/google-gemini/gemini-cli/pull/966/commits/3ced24b3faaa94b9db0224ad5c901f7ed3157e40

packages/core/src/core/coreToolScheduler.ts
jerop: addressed in https://github.com/google-gemini/gemini-cli/pull/966/commits/78b3b9f528ddfc63fc3d43b9e773434dbf315718

mattKorwel (approved):

PR #970, opened 2025-06-12
Revert "Add support for local logging per session (#936)"

anj-s:
This reverts commit 2dc79b3bd0f020ede1a923a9c8311058a23977b0.

We are storing previous session history in a file called logs.json and there is no other state related to these messages.  We need to now store local logs in another file to avoid breaking this feature.


scottdensmore (approved):
Ship it!

PR #971, opened 2025-06-12
Edit pass of docs/architecture.md 

starsandskies:
Make a pass through the docs/archtiecture.md file to hopefully improve readability and conciseness.

Notably, I've proposed axing the conceptual diagram, since as currently presented it's not very comprehensible. I think that a diagram could have value here, but it would need a fair amount of tweaking, and I'm not familiar with the syntax being used.


NTaylorMullen (approved):
❤️

scidomino (approved):

PR #972, opened 2025-06-12
Added sandbox error hint when MCP servers fail to launch in sandbox mode

markmcd:
Adding extra error context when sandbox is enabled and MCP servers fail.

As a user I had half-expected the MCP tools to automatically be plumbed into a sandbox, so this extra hint just makes it a bit clearer to users when they first see the failure.


NTaylorMullen (approved):
Love this @markmcd !!!

NTaylorMullen:
@markmcd merging on your behalf to ensure this makes it into the release branch pre-ask mode.

PR #978, opened 2025-06-12
Ignore dot files on @-completion.

dewitt:
This change elides dot files in two places:

- When typing `@` and returning the current directory contents.
- When typing `@{search prefix}` and filtering all files.

It does display . files if the user explicitly prefixes the search string with a `.`.

Fixes issue #977.

### DIFF
--- a/packages/cli/src/ui/hooks/useCompletion.integration.test.ts
+++ b/packages/cli/src/ui/hooks/useCompletion.integration.test.ts
@@ -241,7 +241,7 @@ describe('useCompletion git-aware filtering integration', () => {
 
     expect(mockFileDiscoveryService.glob).toHaveBeenCalledWith('**/s*', {
       cwd: testCwd,
-      dot: true,
+      dot: false,
     });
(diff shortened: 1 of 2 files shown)


packages/cli/src/ui/hooks/useCompletion.integration.test.ts
```diff
@@ -241,7 +241,7 @@ describe('useCompletion git-aware filtering integration', () => {
 
     expect(mockFileDiscoveryService.glob).toHaveBeenCalledWith('**/s*', {
       cwd: testCwd,
-      dot: true,
```
jacob314, line 244: can you also add a test that results in dot: true

jacob314 (dismissed):
lgtm

packages/cli/src/ui/hooks/useCompletion.integration.test.ts
dewitt: Done in f537976. And really, thank you Gemini, who did that in one shot based only on my copy and pasting the above review comment verbatim into the CLI.

jacob314 (approved):
lgtm

PR #985, opened 2025-06-12
bug: only show tool tokens if > 0

abhipatel12:
Currently, it appears that tool usage (it terms of GC) is not adequately portrayed by the tool token count field in the usage metadata  in the response from the Gemini API. 

Thus, even when GC appears to use tools like `readFile` and `writeFile`, the stats shown in `/stats` and `/quit` ALWAYS show `0` for tool token counts. 

This PR adds a conditional that hides this row unless there is something returned in the API response. 

A good followup would be to understand why tools are not being counted here and what we can do to make this a more useful stat for users.


jacob314 (approved):
lgtm

PR #986, opened 2025-06-12
Add `/compress` command to force a compression of the context

jakemac53:
Related to https://b.corp.google.com/issues/423605555 - I figured this might be a simpler solution to start with, while still also being useful on its own even if we do implement that.

The UX could probably be improved a little bit, if you could point me at how the loading spinners are implemented for regular chats I could try and replicate that.


packages/cli/src/ui/hooks/slashCommandProcessor.ts
```diff
@@ -641,6 +641,31 @@ Add any other context about the problem here.
           }, 100);
         },
       },
+      {
+        name: 'compress',
+        description: 'Force compresses the context by replacing it with a summary.',
```
jacob314: Nit: remove the word Force.

packages/core/src/core/client.ts
```diff
@@ -522,37 +522,41 @@ export class GeminiClient {
...
+  async tryCompressChat(force: boolean = false): Promise<boolean> {
     const chat = await this.chat;
     const history = chat.getHistory(true); // Get curated history
 
-    const cg = await this.contentGenerator;
-    const { totalTokens } = await cg.countTokens({
-      model: this.model,
-      contents: history,
-    });
+    // If not forced, check if we should compress based on context size.
```
jacob314, line 415: nit: even when force is true do you really want to do anything if history is empty?

docs/cli/commands.md
```diff
@@ -61,6 +61,10 @@ Slash commands provide meta-level control over the CLI itself. They can typicall
   - **Description:** Displays a list of all the tools that are currently available to the model.
   - **Action:** Outputs a list of the available tools.
 
+- **`/compress`**
```
jakemac53, line 65: We could also call this `summarize` or something

packages/cli/src/ui/hooks/slashCommandProcessor.ts
```diff
@@ -641,6 +641,31 @@ Add any other context about the problem here.
...
+        action: async (_mainCommand, _subCommand, _args) => {
+          addMessage({
+            type: MessageType.INFO,
+            content: 'Attempting to compress chat history...',
+            timestamp: new Date(),
+          });
+          try {
+            await config!.getGeminiClient()!.tryCompressChat(true);
```
jacob314: nit: maybe check the return value of tryCompressChat to check whether it is false just in case you want to show a different message if there was nothing to compress or the history was too small to bother trying to compress.

jakemac53:
I will try and figure out how to get formatting set up to make this happy as well...

packages/cli/src/ui/hooks/slashCommandProcessor.ts
```diff
@@ -641,6 +641,31 @@ Add any other context about the problem here.
...
+            addMessage({
+              type: MessageType.INFO,
+              content: 'Chat history compressed!',
```
jacob314: nit: change `!` to `.`
not everyone has quite your excitement about compressing history :)

jacob314: also maybe show the # of tokens before and after right in this message so people know how much they saved

packages/core/src/core/client.ts
jakemac53: I added a check for empty history, beyond that since its an explicit command I think we can assume the user knows what they are doing.

docs/cli/commands.md
abhipatel12: `/summarize` could be the alt word we check for in case someone looks for that.

packages/cli/src/ui/hooks/slashCommandProcessor.ts
jakemac53: Done

jakemac53:
Ok, I believe all the comments have been resolved now, let me know if you all want to iterate on the UX but I might need a bit of direction on how to improve that.

jakemac53:
[Screen recording 2025-06-12 1.56.25 PM.webm](https://github.com/user-attachments/assets/f9a858e5-ca2a-4334-abb7-e4eaa9696163)

Here is an example using the feature, I asked it to tell me about the pkgs/cli dir and then compressed the chat, then asked again a question to ensure that previous task was still in the history.

abhipatel12:
> Screen.recording.2025-06-12.1.56.25.PM.webm 
> Here is an example using the feature, I asked it to tell me about the pkgs/cli dir and then compressed the chat, then asked again a question to ensure that previous task was still in the history.



> Screen.recording.2025-06-12.1.56.25.PM.webm 
> Here is an example using the feature, I asked it to tell me about the pkgs/cli dir and then compressed the chat, then asked again a question to ensure that previous task was still in the history.



> Screen.recording.2025-06-12.1.56.25.PM.webm 
> Here is an example using the feature, I asked it to tell me about the pkgs/cli dir and then compressed the chat, then asked again a question to ensure that previous task was still in the history.

Super cool! For UX, do you think we could use the `LoadingIndicator` UI component? It would give us the loading symbol, the message "Compressing chat history ...", and also a timer which lets the user know the app isn't stuck (just in case this takes some time).

Once the compression completes, i think the token count is helpful. Claude code actually prints the compacted history back to the user. I think both use cases are compelling. Token count lets me know my window is freed up. But seeing the compacted history can also make it easier to remember/double check what the agent thinks we've been up to.

jakemac53:
> Super cool! For UX, do you think we could use the `LoadingIndicator` UI component? It would give us the loading symbol, the message "Compressing chat history ...", and also a timer which lets the user know the app isn't stuck (just in case this takes some time).

Yeah this is what I was asking about above, but I don't know react, typescript, or any of the stuff here so I need a bit of direction for how to implement this. I tried to get gemini_cli to at least give me some direction but it just burned 4 million tokens over about 10 minutes and never really did anything useful for me haha.

jakemac53:
Ok, I think I am figuring out how this all works 🤣 . Might not have anything out until Monday though. I am adding a new kind of history item for these with its own message display.

abhipatel12:
> Ok, I think I am figuring out how this all works 🤣 . Might not have anything out until Monday though.

Haha I totally feel the pain. The UI for the stats had me stressed. My blood pressure and token counts were high 😂 Just lmk if I can help!

jakemac53:
[Screen recording 2025-06-13 12.43.20 PM.webm](https://github.com/user-attachments/assets/4314a3ea-0182-47ba-acdc-c977d3718744)

Ok, here is the update UI including an error case (compressing while already compressing).

packages/cli/src/ui/hooks/slashCommandProcessor.test.ts
```diff
@@ -959,6 +956,14 @@ Add any other context about the problem here.
   describe('/compress command', () => {
     it('should call tryCompressChat(true)', async () => {
       const { handleSlashCommand } = getProcessor();
+      mockTryCompressChat.mockImplementationOnce(async (force?: boolean) => {
+        // TODO: Check that we have a pending compression item in the history.
```
jakemac53, line 960: I could not for the life of me figure out how to get the latest value of the pending history items here... the `getProcessor` call above does return the list, but its the list at the start (so, empty). Calling `getProcessor()` again gives null. I don't know react well enough to know how to properly get the up to date list?

docs/cli/commands.md
jakemac53: I did add this as an alt

packages/cli/src/ui/components/messages/CompressionMessage.tsx
```diff
@@ -0,0 +1,48 @@
...
+}) => {
+  const text = compression.isPending
+    ? 'Compressing chat history'
+    : `Chat history compressed from ${compression.originalTokenCount} to ${compression.newTokenCount} tokens.`;
+
+  return (
+    <Box flexDirection="row">
+      <Box marginRight={1}>
+        {compression.isPending ? (
+          <Spinner type="dots" />
```
jacob314, line 32: Can you use GeminiRespondingSpinner instead?
For your case it shouldn't really matter but it helps us avoid Ink flicker by not spinning if it is shown at the same time we are trying to allow the user to input something.

Btw that is the spinner that Gemini recommends for our project if I give it
@packages and ask what spinner to use.

packages/cli/src/ui/hooks/slashCommandProcessor.test.ts
jacob314: see test buffer tests for examples using act and renderHook to get this from React. Gemini CLI is mediocre at React testing but can do a better job if you ask it to modle what it is doing off existing tests. For example, use the same patterns as text-buffer-test.ts for testing react components.

packages/cli/src/ui/types.ts
```diff
@@ -53,6 +53,12 @@ export interface IndividualToolCallDisplay {
   renderOutputAsMarkdown?: boolean;
 }
 
+export interface CompressionProps {
+  isPending: boolean;
+  originalTokenCount?: number;
```
jacob314, line 58: nit: make these two required instead of optional.

packages/core/src/core/turn.ts
```diff
@@ -98,8 +98,14 @@ export type ServerGeminiErrorEvent = {
   value: GeminiErrorEventValue;
 };
 
+export interface ChatCompressionInfo {
+  originalTokenCount: number | undefined;
```
jacob314, line 102: nit: make these two both non-nullable and make the ChatCompressionInfo value on line 108 nullable instead if that is needed to handle the failure case.

jacob314 (approved):
lgtm
This is looking good feel free to land as is and address comments in a follow up pull request.

abhipatel12:
The UI looks great!

packages/cli/src/ui/components/messages/CompressionMessage.tsx
jakemac53: I looked into the loading spinner, then the gemini spinner, and then ended up on this one because it was a lot simpler. The gemini spinner works based on the global streaming state IIRC.

packages/cli/src/ui/types.ts
jakemac53: They are optional because they aren't known up front - I could make the original count required though.

Or do you want them to be required but nullable, like use `number | null` as the type?

scottdensmore:
@jakemac53 are you ready to merge this?

jakemac53:
Sure sounds good, I can follow up as needed

PR #987, opened 2025-06-12
Add web socket protocol support for IDE MCP server

skeshive:
Tested locally with the following settings.json:
```
{
"mcpServers": {
    "ide": {
      "tcp": "ws://localhost:3000"
    }
  }
}
```

Screenshot 2025-06-12 at 3 44 36 PM

For now, IDE MCP servers are treated the same as other user-specified MCP servers. In a follow-up, we should automatically detect & start up the IDE MCP server instead.


mattKorwel (approved):
LGTM. I'm approving but i'm also about to pull it and test other MCP servers to double check it hasnt' broken anything

eddie-santos (approved):

PR #989, opened 2025-06-12
`shift+enter` creates a new line, `enter` still submits the command

eddie-santos:
For interacting with LLMs, whether through Gemini web app, ChatGPT, Claude Code, etc, `shift+enter` creates a new line and `enter` submits the prompt.

This has been a major pain point, so fixing it here.

### DIFF
--- a/packages/cli/src/ui/components/InputPrompt.tsx
+++ b/packages/cli/src/ui/components/InputPrompt.tsx
@@ -292,10 +292,10 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
         console.log('[InputPromptCombined] event', { input, key });
       }
 
-      // Ctrl+Enter for newline, Enter for submit
+      // Shift+Enter for newline, Enter for submit
       if (key.return) {
-        if (key.ctrl) {
-          // Ctrl+Enter for newline
+        if (key.shift) {
+          // Shift+Enter for newline
           buffer.newline();
         } else {
           // Enter for submit


packages/cli/src/ui/components/InputPrompt.tsx
```diff
@@ -292,10 +292,10 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
         console.log('[InputPromptCombined] event', { input, key });
       }
 
-      // Ctrl+Enter for newline, Enter for submit
```
eddie-santos, line 295: Note: ctrl+enter has not worked and is not the industry established pattern

NTaylorMullen: Agree BUT there might be an underlying Ink thing here. @dewitt do you recall by chance?

NTaylorMullen (approved):

dewitt:
IIRC, this change doesn't actually work in most terminals out of the box, nice as it would be, as the gemini-cli app won't receive this event by default.

My understanding is that Claude Code solved this by providing an in-app helper utility called /terminal-setup that attempts to map shift-enter to `\n` so the app's event handler can see it.

But even that work by Claude Code seems to be best effort and terminal-specific, based on the number of issues filed against Claude Code for it: https://github.com/search?q=repo%3Aanthropics%2Fclaude-code+terminal-setup&type=issues

**Note:**

I checked out the `eddie-santos/new-line` branch it did _not_ work on my machine, which is a fairly representative setup (Terminal.app 2.14, macOS 15.5). I've never run Claude Code's /terminal-setup on that machine.

**Questions:**

@eddie-santos, did this change work on your own terminal? If so, what terminal was that?

Have you previously run /terminal-setup in Claude Code before? (I'm going to wager a chance of "yes" if this worked for you.)

**Recommendation:**

I'd recommend *not* merging this PR until we know how widely supported this change would be.

We may need to figure out our own version of /terminal-setup, since shift-enter is so incredibly useful!

CC @jacob314 for visibility.

deepankarsharma:
I did some related work in #898 for this. Added a terminal-setup command that addresses the issues @dewitt raises above.

mattKorwel:
Hi I'm making a pass doing some hygine on outstanding issues as we try and lock things down for the release.

@eddie-santos is this a pr that we still want to move foward with? If so it should be based on the release branch and tagged as `July 25th Release`. If not please close. Thank you!

NTaylorMullen:
Closing for now due to staleness, feel free to reopen.

PR #992, opened 2025-06-12
feat: Show model thoughts while loading

asadm:
https://github.com/user-attachments/assets/560cbbb6-e694-49a2-a6b6-4823b485052c


packages/cli/src/ui/App.tsx
```diff
@@ -485,6 +490,7 @@ const App = ({ config, settings, startupWarnings = [] }: AppProps) => {
           ) : (
             <>
               <LoadingIndicator
+                thought={thought}
```
allenhutchison: I think if the loading phrases are disabled you should disable these as well.

packages/cli/src/ui/components/LoadingIndicator.tsx
```diff
@@ -15,38 +15,69 @@ interface LoadingIndicatorProps {
...
   const streamingState = useStreamingContext();
 
   if (streamingState === StreamingState.Idle) {
     return null;
   }
 
+  const getBoldedText = (text: string): string => {
+    const match = text.match(/\*\*(.*?)\*\*/s);
+    return match ? match[1].trim() : '';
+  };
```
allenhutchison: Do the thoughts come back as markdown bold?

packages/cli/src/ui/components/LoadingIndicator.tsx
```diff
@@ -15,38 +15,69 @@ interface LoadingIndicatorProps {
...
+            }
+          />
+        </Box>
+        {primaryText && <Text color={Colors.AccentPurple}>{primaryText}</Text>}
+        <Text color={Colors.Gray}>
+          {streamingState === StreamingState.WaitingForConfirmation
+            ? ''
+            : ` (esc to cancel, ${elapsedTime}s)`}
+        </Text>
+        <Box flexGrow={1}>{/* Spacer */}</Box>
```
allenhutchison, line 55: I think that if we are going to do this we should display the thoughts in place of the witty phrases when we have thoughts, not below.

packages/core/src/core/client.ts
```diff
@@ -48,6 +48,11 @@ if (proxy) {
   setGlobalDispatcher(new ProxyAgent(proxy));
 }
 
+function isThinkingSupported(model: string) {
+  if (model.startsWith('gemini-2.5')) return true;
+  return false;
+}
```
allenhutchison, line 44: I think there is a better way to do this by calling the capabilities of the API. See models:list here https://ai.google.dev/api/models

allenhutchison (commented):
I think this is overall useful, but my opinion is that you should replace the witty phrases with thoughts when we have them and fall back to witty phrases for actions that don't have thoughts. 

I'm uncomfortable with the heuristic for determining if the model has thoughts, that is too easy to get out of date. I left a link for how to get that data programmatically. 

Ultimately, @NTaylorMullen is the decider on if this should go in or not, so these are just my opinions.

packages/cli/src/ui/components/LoadingIndicator.tsx
abhipatel12: +1, I also think the 'title' of the thoughts may contain enough information initially. Adding the description may add more information that needed, and often times it may get cut off anyway due to the character limit

packages/cli/src/ui/components/LoadingIndicator.tsx
abhipatel12: There's not great documentation on this [here](https://ai.google.dev/gemini-api/docs/thinking#summaries) but maybe we can pass this to the frontend in a more structured object?

Ex.
```
type ThoughtSummary = {
  subject: string;
  description: string;
};

// Use this in an array here
export type ServerGeminiThoughtEvent = {
  type: GeminiEventType.Thought;
  value: ThoughtSummary[];
};
```
That way it's more readable and we decouple the backend response from this frontend parsing?

abhipatel12:
I think the thoughts can absolutely be helpful. I have a some questions about UX (full transparency I'm not a UX expert, so this is just a perspective as a user)

### Disconnect b/w top and bottom of viewport
In the video, when the tool prompt appears for the user, the thought shown at the bottom is a little disconnected with what's happening in general. This can create a flow disruption between the top and bottom of the viewport (top = tool prompt for shell command, bottom = says the agent is actually identifying the rendering library). 

The thinking summary is sort of a chronological plan that gives an idea to the user about what the agent is doing so the plan and action on screen should match. One possible option is to have this thought plan above the response area. This is def hard tho since the plan may get scrolled out of existence as more steps are followed. There may be some UI that allows it be sticky and showcase the steps we've taken thus far, with a max number of steps before the previous ones are lost.

```
Model is thinking...
┌─ Thought Process ───────────────
│ ► Interpreting the Prompt...
│ └─ I'm breaking down the user's inquiry, focusing on the ambiguous term "rendering".
│
│ ► Confirming UI Framework...
│ └─ I've confirmed the UI framework: it's indeed Ink! I found it listed as a dependency.
│
│ ► Identifying the Rendering Library...
│ └─ Now I need to confirm how Ink is used. I'll search for it in the codebase.
└─────────────────────────────────
```

An another alternative that isn't a complete change in UI, is to be sure to change the description text back to "Waiting for user confirmation" to avoid this confusion. This may be a lot easier to implement and stays true to the current design.


### Two lines of thought
In the video, there is almost a subject/title thought and then a grey line of descriptive text. I think that second line is a bit distracting and doesn't add as much value as the high level thought in the bold. The 80 character limit means, you get some detail, but it cuts off before you really know what's going on. My suggestion would be to just not have that second line.

packages/cli/src/ui/components/LoadingIndicator.tsx
NTaylorMullen: Agree

NTaylorMullen:
@abhipatel12's [comment](https://github.com/google-gemini/gemini-cli/pull/992#issuecomment-2968699957) aligns xactly how I was thinking of this problem as well! I'd love to get this in after addressing some of those concerns (+ the ones Allen left as well).

Really appreciate the contribution @asadm, very cool that you took this on ❤️

packages/core/src/core/client.ts
asadm: I tried models:list and I dont see any way to infer thinking. 
```
[
  {
      "name": "models/gemini-2.5-pro-preview-06-05",
      "displayName": "Gemini 2.5 Pro Preview",
      "description": "Preview release (June 5th, 2025) of Gemini 2.5 Pro",
      "version": "2.5-preview-06-05",
      "tunedModelInfo": {},
      "inputTokenLimit": 1048576,
      "outputTokenLimit": 65536,
      "supportedActions": [
        "generateContent",
        "countTokens",
        "createCachedContent",
        "batchGenerateContent"
      ]
  },
...
]
```

asadm:
I have updates:
- Thoughts are parsed and sent in structured object from backend
- We show witty phrase until thought arrives, we replace witty phrase with thought at that point
- Removed subtext, we can iterate on showing extended UI (as @abhipatel12 suggested) in a followup sometime
- Hide thought if disableLoadingPhrases is set

https://github.com/user-attachments/assets/2d1973b0-2d40-4279-9360-493aa01dac8c

packages/cli/src/ui/components/LoadingIndicator.tsx
asadm: done

packages/cli/src/ui/components/LoadingIndicator.tsx
asadm: done. we only show title in-place of phrase when available.

packages/core/src/core/client.ts
allenhutchison: So sorry I thought they had added a thinking status. I've talked to the API team and they are going to work on that. Tracking bug at b/424830866

abhipatel12:
> I have updates:
> 
> * Thoughts are parsed and sent in structured object from backend
> * We show witty phrase until thought arrives, we replace witty phrase with thought at that point
> * Removed subtext, we can iterate on showing extended UI (as @abhipatel12 suggested) in a followup sometime
> * Hide thought if disableLoadingPhrases is set
> 
>  gclithoughts2.mov

This looks great! One more question, in that last example, we didn't have a user being prompted to accept or reject. When that happens are we reverting back to the phrase being "Waiting on user input?"

asadm:
> > I have updates:
> > 
> > * Thoughts are parsed and sent in structured object from backend
> > * We show witty phrase until thought arrives, we replace witty phrase with thought at that point
> > * Removed subtext, we can iterate on showing extended UI (as @abhipatel12 suggested) in a followup sometime
> > * Hide thought if disableLoadingPhrases is set
> > 
> > gclithoughts2.mov
> 
> This looks great! One more question, in that last example, we didn't have a user being prompted to accept or reject. When that happens are we reverting back to the phrase being "Waiting on user input?"

Didn't realise that happens, addressed that in latest commit.

packages/core/src/core/turn.ts
```diff
@@ -69,11 +70,21 @@ export interface ServerToolCallConfirmationDetails {
   details: ToolCallConfirmationDetails;
 }
 
+export type ThoughtSummary = {
+  subject: string;
+  description: string;
```
NTaylorMullen, line 75: It might be nice to have a hot key to expand thought descriptions at some point as a follow up with some nice UX 😄

packages/core/src/core/turn.ts
```diff
@@ -154,6 +166,26 @@ export class Turn {
         }
         this.debugResponses.push(resp);
 
+        const thought = resp.candidates?.[0]?.content?.parts?.[0]?.thought;
+        if (thought) {
+          const rawText = resp.candidates?.[0]?.content?.parts?.[0]?.text ?? '';
+          const subjectStringMatches = rawText.match(/\*\*(.*?)\*\*/s);
```
NTaylorMullen, line 180: Mind adding a large comment to these bits as to what the thought pieces look like / what you're parsing? I would have imagined that the genai SDK would have handled all this for us

NTaylorMullen (approved):

packages/core/src/core/turn.ts
keir: Consider reducing the double reach into the object depths; maybe something like this:

```js
const thought = resp.candidates?.[0]?.content?.parts?.[0];
if (thought?.thought) {
  const rawText = thought.text ?? '';
  const subjectStringMatches = rawText.match(/\*\*(.*?)\*\*/s);
```

PR #994, opened 2025-06-12
refactor: Use default centralized Flash & Pro models everywhere

NTaylorMullen:
- Move default model names to a new file in packages/core
- Update references to use the new centralized location
- This makes model configuration more consistent and easier to manage.
- Remove gemini-2.0-flash usage in favor of 2.5-flash (not supported on Titan)

Fixes https://b.corp.google.com/issues/424624952

### DIFF
--- a/packages/core/src/utils/editCorrector.ts
+++ b/packages/core/src/utils/editCorrector.ts
@@ -13,8 +13,9 @@ import {
 import { GeminiClient } from '../core/client.js';
 import { EditToolParams } from '../tools/edit.js';
 import { LruCache } from './LruCache.js';
+import { DEFAULT_GEMINI_FLASH_MODEL } from '../config/models.js';
 
-const EditModel = 'gemini-2.5-flash-preview-04-17';
+const EditModel = DEFAULT_GEMINI_FLASH_MODEL;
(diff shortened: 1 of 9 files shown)


anj-s (approved):

packages/core/src/utils/editCorrector.ts
```diff
@@ -13,8 +13,9 @@ import {
 import { GeminiClient } from '../core/client.js';
 import { EditToolParams } from '../tools/edit.js';
 import { LruCache } from './LruCache.js';
+import { DEFAULT_GEMINI_FLASH_MODEL } from '../config/models.js';
 
-const EditModel = 'gemini-2.5-flash-preview-04-17';
+const EditModel = DEFAULT_GEMINI_FLASH_MODEL;
```
MarlonGamez, line 18: Should we just remove this variable? Since it's const as well

MarlonGamez (approved):

packages/core/src/utils/editCorrector.ts
NTaylorMullen: I imagine we may change this in the future, good callout though.

PR #997, opened 2025-06-13
Improve some tools to support abortSignal

scidomino:
b/423771113

Some tools take a long time and users are annoyed that hitting cancel doesn't do anything.

Since the existing library we were using (fast-glob) doesn't support abrotSignal I switched to "glob" instead. It's more full featured, though possibly not quite as fast.

Note: fast-glob had abortSignal support added 5 months ago but it hasn't made it to the newest release yet (they release very infrequently). So we might consider migrating back someday if the speed difference matters.

This was originally attempted in #988 and rolled back in #996.


mattKorwel (approved):
As a part of my pr review i pulled this branch built installed and ran from code. I also ran ` npx https://github.com/google-gemini/gemini-cli#tomm_glob` to test npx install. both worked.

I also ran some basic smoke tests to validate that file operations still worked including having it search for key words. didn't have it take long enough to need to abort.

mattKorwel:
@scidomino approved, please validate with  ` npx https://github.com/google-gemini/gemini-cli#main` after merge builds run.

scidomino:
I ran npx https://github.com/google-gemini/gemini-cli#main after merging and all my tests passed.

PR #1001, opened 2025-06-13
Update contextFileName to support an optional list of strings

bbiggs:
Add support for a list of filenames in the contextFileName setting. The setting can now be either a single filename, or a list of supported filenames. Does not update the (many) references to GEMINI.md in the output (example: "Using 2 GEMINI.md file(s)". Fixes #1000


packages/cli/src/ui/components/ContextSummaryDisplay.tsx
```diff
@@ -30,7 +30,9 @@ export const ContextSummaryDisplay: React.FC<ContextSummaryDisplayProps> = ({
 
   const geminiMdText =
     geminiMdFileCount > 0
-      ? `${geminiMdFileCount} ${contextFileName} file${geminiMdFileCount > 1 ? 's' : ''}`
+      ? `${geminiMdFileCount} ${contextFileNames[0]} file${
```
NTaylorMullen, line 33: non-blocking: Instead of taking the first element could we see if all file names are the same, if so, then take [0], otherwise just render `context` that way the final output is something like `2 context files` in the case that we have varying names?

Alternatively I could see us spell out each context file here too with its count; although that may get verbose / lengthy

packages/core/src/tools/memoryTool.ts
```diff
@@ -51,18 +51,32 @@ export const MEMORY_SECTION_HEADER = '## Gemini Added Memories';
...
 // It defaults to DEFAULT_CONTEXT_FILENAME but can be overridden by setGeminiMdFilename.
-let currentGeminiMdFilename = DEFAULT_CONTEXT_FILENAME;
+let currentGeminiMdFilename: string | string[] = DEFAULT_CONTEXT_FILENAME;
 
-export function setGeminiMdFilename(newFilename: string): void {
-  if (newFilename && newFilename.trim() !== '') {
+export function setGeminiMdFilename(newFilename: string | string[]): void {
+  if (Array.isArray(newFilename)) {
+    if (newFilename.length > 0) {
+      currentGeminiMdFilename = newFilename;
```
NTaylorMullen: non-blocking but nice to have:

```suggestion
      currentGeminiMdFilename = newFilename.map(name => name.trim());
```

NTaylorMullen (approved):
Feel free to merge away after fixes!

packages/cli/src/ui/components/ContextSummaryDisplay.tsx
bbiggs: I think this just needs a rework a bit so I recommend doing this separately from this change which just adds the basic functionality.

packages/core/src/tools/memoryTool.ts
bbiggs: done

PR #1002, opened 2025-06-13
Handle telemetry in non-interactive mode

jerop:
Changes:
- Ensure proper shutdown in non-interactive mode
- Ensures the initial user prompt is logged in non-interactive mode
- Improve telemetry for streaming - handle chunks and input token count is now alongside other token counts in response

To test:
- Follow instructions in https://github.com/google-gemini/gemini-cli/blob/main/docs/core/telemetry.md#google-cloud
- Run CLI in non-interactive mode and observe logs/metrics in GCP Logs Explorer and Metrics Explorer

#750


packages/core/src/core/geminiChat.ts
```diff
@@ -140,51 +140,42 @@ export class GeminiChat {
     validateHistory(history);
   }
 
+  private _getRequestTextFromContents(contents: Content[]): string {
```
NTaylorMullen, line 143: @jacob314 we have a helper for this don't we?

NTaylorMullen (approved):

packages/core/src/core/geminiChat.ts
jacob314: Yep I'd created a method for that but forgot where so I just asked Gemini CLI and it reminded me.
"@packages where is a method that generates a String from a PartListUnion object?"
answer:

partListUnionToString in geminiRequest.ts

PR #1003, opened 2025-06-13
Telemetry: Auto-export to GCP if GOOGLE_CLOUD_PROJECT is set

jerop:
This change simplifies telemetry setup for users who want to integrate with GCP, as they no longer need to configure a local collector or an explicit endpoint if their project is already set up in the environment.

This change updates the telemetry system to automatically export traces, logs, and metrics to Google Cloud Platform (GCP) if the `GOOGLE_CLOUD_PROJECT` environment variable is set and no explicit `telemetryOtlpEndpoint` is configured by the user.

Key changes:
-   The default `telemetryOtlpEndpoint` in `Config` is now an empty string.
-   The `initializeTelemetry` SDK logic now prioritizes:
    -  User-defined `telemetryOtlpEndpoint`.
    -  `GOOGLE_CLOUD_PROJECT` for direct GCP export.
    -  Console exporters as a fallback.
-   If an invalid `telemetryOtlpEndpoint` is provided, it falls back to console exporters with a warning.

#750


docs/core/telemetry.md
NTaylorMullen, line 1: Absolutely epic docs

NTaylorMullen (approved):
Added to status update too. Epicness!

PR #1004, opened 2025-06-13
Light README polish

markmcd:
I've updated `GEMINI.md` to support one of the headline examples (querying GitHub issues).

My only concern with this is whether we want to lead with generating a Gemini API app in the first example, as it tends to use the older SDKs.  I think the self-awareness of it is pretty compelling but maybe something more reliable would be better. LMKWYT.


GEMINI.md
```diff
@@ -60,7 +60,9 @@ This project uses **Vitest** as its primary testing framework. When writing test
 
 ## Git Repo
 
-The main branch for this project is called "main"
+The main branch for this project is called "main".
+
+This project is hosted on GitHub. To interact with issues and pull requests, you should use the GitHub CLI tool (`gh`) if it's installed. Otherwise you can find the project's issues at https://github.com/google-gemini/gemini-cli/issues and pull requests at https://github.com/google-gemini/pulls.
```
cperry-goog: @NTaylorMullen can you ack the changes to the GEMINI.md file here? I don't trust myself.

cperry-goog (approved):
README lgtm, asking @NTaylorMullen to check out GEMINI.md

GEMINI.md
NTaylorMullen: Ya I'm not convinced you need to add this line to be honest. The model knows of `gh` and the pathing conventions of GitHub. I'd remove this

README.md
```diff
@@ -2,28 +2,46 @@
...
-Once the CLI is running, you can start interacting with Gemini from your shell. Try a simple query:
+Once the CLI is running, you can start interacting with Gemini from your shell.
 
-```
-> How can I build a web app?
+You can start a project from a new directory:
+
+```sh
+$ cd projects/
+$ gemini --yolo  # Auto-accept everything!
```
NTaylorMullen: Probably dangerous for us to encourage use of `--yolo` outside of a sandbox. I'm not sure I'd do that here. People can use at their own risk. In this scenario GC will walk folks through changes needed most likely in a straightforward way

NTaylorMullen (approved):

README.md
markmcd: Yeah makes sense.

My thinking was a that for a clean new project, --yolo is safe, and this way we demonstrate the feature on the "front page" but agreed that we shouldn't open with `--yolo`

GEMINI.md
markmcd: I added this as I tested the demo prompt a bunch of times, and mostly I would get answers like `✦ I am sorry, I cannot fulfill this request. I do not have access to real-time information about GitHub issues.` or `✦ I am sorry, but I cannot access GitHub to view recently filed issues. I can only access the files in the current project directory.`

The best result I saw was when it tried to use Google Search, but that doesn't work atm (private repo)

The demo prompt below doesn't work reliably without it, so I'll need to find another prompt.

NTaylorMullen: Ah very interesting, if that's the case I'd recommend putting the `gh` inclusion in your `~/.gemini/GEMINI.md`. Primary reason is that it's not guaranteed to be on dev machines

NTaylorMullen: Part of me wonders if we should ask the model to be probing for specific utilities based on the user repo. I.e. on startup I could imagine us running git commands to see if the repo is a GitHub repo and then adding some logic that then probes for `gh` and the like

markmcd: OK - if we drop the `gh` tool option and just use the URL hints, then it's fine, but it won't work until launch :laughing:  (needs to be public)

I've switched to another prompt to make it simpler. LMKWYT

PR #998, opened 2025-06-13
Support GOOGLE_API_KEY hoisting in sandbox too

markmcd:

PR #1005, opened 2025-06-13
Add Vertex env vars to sandbox

markmcd:
I accidentally merged #998 before adding these, so here's another PR. This includes environment vars and brings the ~/.config/gcloud dir into the sandbox too.

I'm not 100% confident that the gcloud config dir would work everywhere (e.g. windows?) but it seemed fine in a local test.


NTaylorMullen (approved):
@markmcd you sir are epic 👏

PR #1007, opened 2025-06-13
Address b/424256913 - fixed error in correctStringEscaping() and improved backslash handling

bdmorgan:
Summary of Changes Made

  **1. Fixed Critical Property Name Mismatch Bug:**
    - In correctStringEscaping function, changed result.corrected_new_string_escaping to result.corrected_string_escaping to match the schema
    - This bug was causing the escaping correction to fail silently, allowing extra backslashes to persist in files
  **2. Enhanced Regex for Escaped Backslashes:**
    - Updated unescapeStringForGeminiBug regex from /\\+(n|t|r|'|"||\n)/gto/(n|t|r|'|"||\\|\n)/g
    - Added support for handling escaped backslashes (\\ → \)
    - Changed from matching "one or more backslashes" to "exactly one backslash" for more precise unescaping
  **3. Comprehensive Test Coverage:**
    - Added tests for escaped backslashes handling
    - Added tests for ensureCorrectFileContent function to verify the property name fix
    - Updated existing tests to match the new regex behavior
    - All 39 tests now pass, ensuring no regressions


scottdensmore (approved):
LGTM :shipit:

packages/core/src/utils/editCorrector.ts
```diff
@@ -564,39 +564,45 @@ function trimPairIfPossible(
...
-  //                       This handles cases like "\\n", "\\\\`", etc.
-  //   \n                 : This matches an actual newline character. This handles cases where the input
-  //                       string might have something like "\\\n" (a literal backslash followed by a newline).
+  //                       This handles cases like "\\n", "\\`", etc.
+  //   \\ : This matches a literal backslash. This handles cases like "\\\\" (escaped backslash).
+  //   \n : This matches an actual newline character. This handles cases where the input
+  //        string might have something like "\\\n" (a literal backslash followed by a newline).
   // g : Global flag, to replace all occurrences.
 
-  return inputString.replace(/\\+(n|t|r|'|"|`|\n)/g, (match, capturedChar) => {
```
NTaylorMullen, line 575: Why remove the `+`? Originally this was intended to capture cases like `\\\\\\n`. I.e. the model loses its mind and overescapes things to the extreme 😄

bdmorgan: You're right - didn't catch that. Will fix.

PR #1008, opened 2025-06-13
feat: clear should also clear chat history

abhipatel12:
Currently clear only clears the console. This PR resets the chat as well to provide the user with a completely fresh session essentially without quitting and restarting.

**FEEDBACK WANTED**
Currently, this implementation does not reset session stats. Would love some feedback on what the correct UX would be in a reset feature like this. Clear stats (brand new session) or just clear history (same session).

### DIFF
--- a/packages/cli/src/ui/hooks/slashCommandProcessor.test.ts
+++ b/packages/cli/src/ui/hooks/slashCommandProcessor.test.ts
@@ -340,6 +340,27 @@ describe('useSlashCommandProcessor', () => {
       expect(commandResult).toBe(true);
     });
 
+    it('/clear should clear items, reset chat, and refresh static', async () => {
+      const mockResetChat = vi.fn();
+      mockConfig = {
+        ...mockConfig,
+        getGeminiClient: () => ({
+          resetChat: mockResetChat,
+        }),
+      } as unknown as Config;
+
+      const { handleSlashCommand } = getProcessor();
+      let commandResult: SlashCommandActionReturn | boolean = false;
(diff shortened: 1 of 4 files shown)


dewitt:
I'm +1 to some way to reset the entire session and letting the user start over without restarting that whole app. If that's even possible though—there's a fair bit of state scattered throughout, and I'd be worried we haven't been disciplined enough about isolating it to catch everything in one go.

But -1 to repurposing `/clear` for this, both because it is a fine name for what it already does (clearing the screen) and because I don't think that's a great name for what you intend here.

So if we _can_ implement a proper session reset, I'd recommend adding a new command `/reset` or similar for this functionality.

Another alternative would be to figure out a way to `/relaunch` by quitting and exec-ing gemini-cli again.

NTaylorMullen (approved):
After offline convo it seems @dewitt that you're also ok with resetting chat history now right?

imo I think it makes sense to clear in a consistency sake for what others are doing in the ecosystem

dewitt:
> After offline convo it seems @dewitt that you're also ok with resetting chat history now right?

Yes, per offline conversation, `/clear -- clear chat history` makes sense. 

Thanks for circling back!

NTaylorMullen:
@abhipatel12 merging on your behalf to ensure this makes it into the release branch pre-ask mode.

abhipatel12:
> @abhipatel12 merging on your behalf to ensure this makes it into the release branch pre-ask mode.

Awesome, thanks!

PR #1013, opened 2025-06-13
Move the shadow git repository to the user's home dir

louisjimenez:
This changes the checkpoint feature to create the shadow repository in the user's home folder and not in the project repository. This avoids have to modify the user's .gitignore. It uses the project's root to create a hash for storing the shadow repos in the home directory.

### DIFF
--- a/packages/core/src/services/gitService.test.ts
+++ b/packages/core/src/services/gitService.test.ts
@@ -5,7 +5,7 @@
  */
 
 import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
-import { GitService, historyDirName } from './gitService.js';
+import { GitService } from './gitService.js';
 import * as path from 'path';
 import type * as FsPromisesModule from 'fs/promises';
 import type { ChildProcess } from 'node:child_process';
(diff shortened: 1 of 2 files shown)


packages/cli/src/utils/sandbox-macos-permissive-open.sb
```diff
@@ -10,6 +10,7 @@
     (subpath (param "TMP_DIR"))
     (subpath (param "CACHE_DIR"))
     (subpath (string-append (param "HOME_DIR") "/.gemini"))
+    (subpath (string-append (param "HOME_DIR") "/.gemini-cli"))
```
NTaylorMullen: Why .gemini-cli/history vs. .gemini/history?

louisjimenez: Agree, I think the /history does a good enough job of keeping it separate. Updated.

NTaylorMullen (approved):

NTaylorMullen:
@louisjimenez merging on your behalf to ensure this makes it into the release branch pre-ask mode branching

PR #1024, opened 2025-06-13
Fix default extension context filename and update docs

scidomino:
Also refactored extensions to introduce a new object.

This was revealed while working on b/424768855

### DIFF
--- a/packages/cli/src/config/extension.ts
+++ b/packages/cli/src/config/extension.ts
@@ -12,6 +12,11 @@ import * as os from 'os';
 export const EXTENSIONS_DIRECTORY_NAME = path.join('.gemini', 'extensions');
 export const EXTENSIONS_CONFIG_FILENAME = 'gemini-extension.json';
 
+export interface Extension {
+  config: ExtensionConfig;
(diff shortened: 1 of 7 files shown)


packages/cli/src/config/extension.ts
```diff
@@ -12,95 +12,97 @@ import * as os from 'os';
...
+  const configFilePath = path.join(extensionDir, EXTENSIONS_CONFIG_FILENAME);
+  if (!fs.existsSync(configFilePath)) {
+    console.error(
+      `Warning: extension directory ${extensionDir} does not contain a config file ${configFilePath}.`,
+    );
+    return null;
   }
 
-  return extensions;
+  const configContent = fs.readFileSync(configFilePath, 'utf-8');
```
mrahs: nit: why not catch the error here to read it atomically instead of the non-atomic checks above?

mrahs (approved):

packages/cli/src/config/extension.ts
scidomino: good point. Fixed!

PR #1025, opened 2025-06-13
reuse `GitIgnoreParser` for loading `.geminiignore`

mrahs:
This is an incremental change towards incorporating .geminiignore filtering into `FileDiscoveryService`.


scidomino (approved):
Love a cl where more lines are deleted than added.

PR #1026, opened 2025-06-13
Stop extension MCPs from hopping into settings.

scidomino:
b/424768855

We were modifying the mcps in the settings object rather than just copying it!


mrahs (approved):

mrahs:
Is it worth a unit test?

scidomino:
> Is it worth a unit test?

Done.

PR #1031, opened 2025-06-14
feat: text-buffer: input sanitization and delete character handling.

jacob314:
Adds a new  function to the text-buffer component to remove ANSI escape codes and control characters from input. This is used in the  hook to sanitize text input.
Start process of refactoring text-buffer to use applyOperations to resolve the issue that current functions in text-buffer are unsafe to apply repeatedly within the safe function call due to how state works in React.


NTaylorMullen (approved):

NTaylorMullen:
@jacob314 merging on your behalf to get this into ask mode

PR #1032, opened 2025-06-14
fix: remove quota project override in non-streaming calls to CCPA

MarlonGamez:
The quota project is implicitly provided by our OAuth2 credentials and we don't want to override this as it's causing errors. We just need to provide the project as part of the request body (which we're already doing).

#975


mattKorwel (approved):
Lgtm

NTaylorMullen (approved):

PR #1036, opened 2025-06-14
fix(core): Improve shell tool reliability and test portability

sijieamoy:
This pull request addresses two related issues to improve the robustness of the ShellTool and the portability of our test suite. The changes are organized into two distinct, atomic commits for clarity and easier review.

1. Fix: Corrected Temp File Name Handling in ShellTool (Fixes https://github.com/google-gemini/gemini-cli/issues/1035)

  This commit refactors the logic for generating the temporary file path used for pgrep output on non-Windows systems.

   * Before: The temporary file path was being generated multiple times within the same function, which could lead to race conditions or errors in tracking background processes.
   * After: The path is now generated once at the beginning of the function and reused, ensuring the ShellTool reliably captures background PIDs.

2. Test: Made Prompt Tests OS-Agnostic

  This commit resolves a brittle test in prompts.test.ts that would produce different snapshots depending on the host operating system (macOS vs. others).

   * Before: The test snapshot included an OS-specific string ("MacOS Seatbelt"), causing the test to fail on non-macOS environments.
   * After: The os.platform() module is now mocked during the test run. This ensures the generated prompt is consistent, making the test portable and reliable across all development machines
     and CI runners.

3. Initializes the GeminiClient before the ToolRegistry within the Config constructor (fixes https://github.com/google-gemini/gemini-cli/issues/1037).

Previously, the ToolRegistry and its associated tools were created before the GeminiClient instance was ready. This caused tools that depend on the client, such as EditTool, to fail with a TypeError when they attempted to make LLM calls.

 This change ensures that all dependencies are available to the tools at instantiation time, preventing the initialization race condition.

### DIFF
--- a/packages/core/src/config/config.ts
+++ b/packages/core/src/config/config.ts
@@ -174,8 +174,8 @@ export class Config {
       setGeminiMdFilename(params.contextFileName);
     }
 
-    this.toolRegistry = createToolRegistry(this);
     this.geminiClient = new GeminiClient(this);
+    this.toolRegistry = createToolRegistry(this);
 
     if (this.telemetrySettings.enabled) {
       initializeTelemetry(this);
--- a/packages/core/src/tools/shell.ts
+++ b/packages/core/src/tools/shell.ts
@@ -170,17 +170,16 @@ export class ShellTool extends BaseTool<ShellToolParams, ToolResult> {
     }
 
     const isWindows = os.platform() === 'win32';
+    const tempFileName = `shell_pgrep_${crypto
+      .randomBytes(6)
+      .toString('hex')}.tmp`;
+    const tempFilePath = path.join(os.tmpdir(), tempFileName);
 
     // pgrep is not available on Windows, so we can't get background PIDs
     const command = isWindows
       ? params.command
       : (() => {
           // wrap command to append subprocess pids (via pgrep) to temporary file
-          const tempFileName = `shell_pgrep_${crypto
-            .randomBytes(6)
-            .toString('hex')}.tmp`;
-          const tempFilePath = path.join(os.tmpdir(), tempFileName);
-
           let command = params.command.trim();
           if (!command.endsWith('&')) command += ';';
           return `{ ${command} }; __code=$?; pgrep -g 0 >${tempFilePath} 2>&1; exit $__code;`;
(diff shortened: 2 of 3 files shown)


packages/core/src/config/config.ts
```diff
@@ -162,8 +162,8 @@ export class Config {
       this.geminiIgnorePatterns = params.geminiIgnorePatterns;
     }
 
-    this.toolRegistry = createToolRegistry(this);
     this.geminiClient = new GeminiClient(this);
+    this.toolRegistry = createToolRegistry(this);
```
NTaylorMullen, line 178: Mind sharing a bit more as to why this reordering fixes https://github.com/google-gemini/gemini-cli/issues/1037 ?

The create tool registry command is async fire & forget whether you do it before or after client initialization. I could totally be missing something here though!

packages/core/src/tools/shell.ts
```diff
@@ -293,10 +292,6 @@ export class ShellTool extends BaseTool<ShellToolParams, ToolResult> {
     // parse pids (pgrep output) from temporary file and remove it
     const backgroundPIDs: number[] = [];
     if (os.platform() !== 'win32') {
-      const tempFileName = `shell_pgrep_${crypto
```
NTaylorMullen, line 296: Ya this just feels like a bug with what we had before. We were regenerating the file name every time. Good catch!

NTaylorMullen (approved):

scottdensmore:
@sijieamoy we should get this one checked in

scottdensmore (approved):
Approved with Taylor's comments

packages/core/src/config/config.ts
sijieamoy: Here's the explanation from Gemini which sounds plausible to me but could you double check? I have been using the patched build for a few hours now and the error hasn't reappeared.

   1. `Config` Constructor Order: In packages/core/src/config/config.ts, the Config class constructor has these two lines in this specific order:
```
// 1. Tool registry is created first
this.toolRegistry = createToolRegistry(this);

// 2. Gemini client is created second
```

   2. `EditTool` Creation: The createToolRegistry(this) function immediately creates a new EditTool instance, passing the Config object (this) to its constructor.


   3. `EditTool` Initialization: The EditTool constructor tries to get the Gemini client and store it:
```
// Inside EditTool's constructor
this.client = config.getGeminiClient();
```

   4. The Problem: At the moment the EditTool is created, the Config constructor has not yet executed the line this.geminiClient = new GeminiClient(this);. Therefore, config.getGeminiClient()
      returns undefined. The EditTool instance now holds this.client = undefined.


   5. The Error Cascade: Later, when you perform an action that uses the EditTool:
       * EditTool.calculateEdit is called.
       * It calls ensureCorrectEdit, passing its own this.client (which is undefined).
       * ensureCorrectEdit then calls correctOldStringMismatch, passing along that same undefined client.
       * Finally, correctOldStringMismatch tries to execute geminiClient.generateJson(...). Since geminiClient is undefined, this results in the TypeError: Cannot read properties of undefined (reading 'generateJson').


In short, the EditTool is created with an incomplete Config object because the geminiClient is initialized too late. Swapping the two lines in the Config constructor ensures the geminiClient exists before any tools that depend on it are created.

NTaylorMullen: This absolutely tracks after digging in. Thank you so much for taking the time to explain a bit more. Really appreciate this fix @sijieamoy!

PR #1038, opened 2025-06-14
Dockerfile: Node 22 and ENTRYPOINT

steren:
Upgrade to Node 22
Use ENTRYPOINT instead of CMD for the entrypoint

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM docker.io/library/node:20-slim
+FROM docker.io/library/node:22-slim
 
 ARG SANDBOX_NAME="gemini-cli-sandbox"
 ENV SANDBOX="$SANDBOX_NAME"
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM docker.io/library/node:20-slim
+FROM docker.io/library/node:22-slim
```
NTaylorMullen, line 1: Any harm in upgrading to node 22?

steren: I'm able to build with `GEMINI_SANDBOX=true npm run build:sandbox` 

And then run with `docker run gemini-cli-sandbox` without errors

steren: I have not performed more intense testing, but now would be good to upgrade and catch any issues.

Node 20 is in maintenance mode, while 22 is LTS https://nodejs.org/en/about/previous-releases

NTaylorMullen (approved):
cc @KeijiBranshi

PR #1039, opened 2025-06-14
centralize file filtering in `FileDiscoveryService`

mrahs:
Here is how it looks like in the CLI:
```
....
Loaded 17 patterns from /.../gemini-cli/.gitignore
Loaded 1 patterns from /.../gemini-cli/.geminiignore
...
> @CONTRIBUTING.md what is this file?

 ╭─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
 │ ✔  ReadManyFiles Will attempt to read and concatenate files using patterns: `CONTRIBUTING.md` (within target directory: `/usr/local/google/home/mrahs/dev/gemini-…  │
 │                                                                                                                                                                     │
 │    ReadManyFiles Result (Target Dir: /usr/local/google/home/mrahs/dev/gemini-cli)                                                                                   │
 │                                                                                                                                                                     │
 │    No files were read and concatenated based on the criteria.                                                                                                       │
 ╰─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯

 ╭─────────────────────────────────────────────────────────────────────────╮
 │ ✔  ReadFile CONTRIBUTING.md                                             │
 │                                                                         │
 │    File path 'CONTRIBUTING.md' is ignored by .geminiignore pattern(s).  │
 ╰─────────────────────────────────────────────────────────────────────────╯

✦ I am unable to directly read CONTRIBUTING.md because it's specified in the project's ignore file.
```

Sorry for the large PR.. it started with removing a couple of files and grow organically with TDD.


packages/cli/src/ui/hooks/useCompletion.ts
```diff
@@ -285,9 +287,7 @@ export function useCompletion(
       setIsLoadingSuggestions(true);
       let fetchedSuggestions: Suggestion[] = [];
 
-      const fileDiscoveryService = config
-        ? await config.getFileService()
-        : null;
+      const fileDiscoveryService = config ? config.getFileService() : null;
```
NTaylorMullen, line 290: Isn't the file serverice required to be non-null?

packages/cli/src/ui/hooks/useCompletion.ts
```diff
@@ -263,9 +264,10 @@ export function useCompletion(
       maxResults = 50,
     ): Promise<Suggestion[]> => {
       const globPattern = `**/${searchPrefix}*`;
-      const files = await fileDiscoveryService.glob(globPattern, {
+      const files = await glob(globPattern, {
```
NTaylorMullen, line 267: Why move glob out of the discovery service?

packages/core/src/utils/bfsFileSearch.ts
```diff
@@ -69,7 +69,7 @@ export async function bfsFileSearch(
 
     for (const entry of entries) {
       const fullPath = path.join(currentDir, entry.name);
-      if (fileService?.shouldIgnoreFile(fullPath)) {
+      if (fileService?.shouldGitIgnoreFile(fullPath)) {
```
NTaylorMullen, line 72: I imagine we should have a joint `ignore` shouldn't we that does both git ignore and geminiignore?

packages/core/src/services/fileDiscoveryService.ts
```diff
@@ -7,84 +7,87 @@
...
-
       return true;
     });
   }
 
   /**
-   * Checks if a single file should be ignored
+   * Checks if a single file should be git-ignored
    */
-  shouldIgnoreFile(
```
NTaylorMullen, line 69: Interesting so before this was synonymous with both gitignore and geminiignore? 

I'd imagine we'd still need one that did both

NTaylorMullen (approved):

packages/cli/src/ui/hooks/useCompletion.ts
mrahs: It wan't, it looks like. I'm trying to keep the PR from growing too much. I'll sweep service usages in a next one to ensure all patterns apply consistently everywhere.

packages/cli/src/ui/hooks/useCompletion.ts
mrahs: It's used differently in different contexts: `useCompletion` wants the result as `string[]`, but the glob tool wants additional fields. `glob` is used in other files as well. Providing a consistent glob helper with filtering requires some abstraction that I wanted to do in a subsequent PR.

packages/core/src/utils/bfsFileSearch.ts
mrahs: Yes, we should. I wanted to do that in a subsequent PR to keep this one from growing too much by maintaining existing behaviour.

packages/core/src/services/fileDiscoveryService.ts
mrahs: Agreed. Will follow up in a subsequent PR.

dewitt:
It appears that this introduces a regression from prior behavior regarding @-completion. 

E.g.:

image

Versus:

image

Specifically, the .gitignores are now not being filtered (see `node_modules' in the example above).

packages/core/src/services/fileDiscoveryService.ts
dewitt: > Agreed. Will follow up in a subsequent PR.

@mrahs, can you take please take a look at the regression we noticed (pictures below), which came up on a bug triage chat this evening. I suspect the subsequent PR is just what we need here.

Please send a pull request against the `release` branch to get this on the release track.

CC @NTaylorMullen

mrahs:
Thank you for the details. Posted https://github.com/google-gemini/gemini-cli/pull/1135 to fix it.

PR #863, opened 2025-06-09
Rollforward AST changes to unblock Sandboxing

mattKorwel:
As the fix for #850 we're rolling forward to manually prune changes that crossed multiple PR's.

PR #1042, opened 2025-06-14
Add: CodeParser Tool

mattKorwel:
# TLDR

This PR adds the ability to use a code parser tool. This ensures it works in all environments by including the correct navitve modules in the docker image; this part of the fix is temporary and will need longer term fixes appled. See below.

### ASTs
When we send requests for parsing code to the LLM must decide if we let the LLM intreperet that code and what it means and what structure it has, or if we do that for the LLM. Tools that parse source code into [Abstract Syntax Trees](https://medium.com/basecs/leveling-up-ones-parsing-game-with-asts-d7a6fc2400ff) provide us that ability to add code specific context to the LLM. 

### Native modules
This PR adds one such tool for us. [Tree Sitter](https://github.com/tree-sitter). These parsing tools are language specific and are only wrapped, in our case, with node bindings to run in a js environment.  Being non js code these modules cannot be included in our [bundle](https://github.com/google-gemini/gemini-cli/blob/3bcb3c36661c87e5dfdd91b2fce19f8b29b31e01/package.json#L28) step.

## Previous attempts.

This code was originally added in #756. This worked in some scenarios, non sandbox mode, but not in others specifically docker, podman and npx install. #840 was was a change to make it work for NPX installs. #863 Rolled the code forward (due to other changes editing files to remove code parser was quicker than actual reverting).  

# Diving Deeper

## More on Native Modules

Our code is intended to run on many different platforms. Currently we support MacOS, Windows, and Linux (desktop, server or docker). While js code is, generally speaking, cross platform compatible due to being interpreted by the node js. This runtime handles the specifics of running on different platforms. 

However in some cases, such as code parsing or running some services such as sq lite, the authors of the modules determined they needed better performance or integration outside of what node could provide. In those case code written in c/c++ is bundled into a package and shiped with node. See the node [documentation](https://nodejs.org/api/addons.html), and a more detailed [explanation](https://blog.risingstack.com/writing-native-node-js-modules/) if interested. 

## Platform dependent

In this case we must ensure that when our code runs it has the right version of Tree sitter for OsX, Windows or linux. This is also true of any other external dependencies we will need to use in the future or [past](https://github.com/google-gemini/gemini-cli/issues/735#issuecomment-2938742910)

### A note on packaging.

Our repo is a node js that uses [workspaces](https://docs.npmjs.com/cli/v7/using-npm/workspaces). The two workspaces cli and core are responsible for code and dependencies for what happens in the system. The main [package.json](https://github.com/google-gemini/gemini-cli/blob/3bcb3c36661c87e5dfdd91b2fce19f8b29b31e01/package.json#L8) has all the scripts for building packaging and shipping these two packages. This is notable and has other downstream effects.

### git clone
For a user who clones our repository on their local machine and then runs `npm install && npm build && npm run start` with sandboxing off, the system will just work. `npm install` is reponsible for pulling the correct versions of tree sitter for the architecture of the users machine.  For this scenario the dependencies for tree sitter can be declared in the core package and were in the first [PR](https://github.com/google-gemini/gemini-cli/pull/756/files#diff-0b810c38f3c138a3d5e44854edefd5eb966617ca84e62f06511f60acc40546c7)

### npx install
For [NPX](https://dev.to/orlikova/understanding-npx-1m4) install we have a different scenario. npx will

1. pull our package from github as a tarball and extract it to `~/.npm/_npx/.....`
2. run npm install on the main package.json only
3. bundle it and run the file set as `binary` in our main package.json. in our case that is [bundle\gemini.js](https://github.com/google-gemini/gemini-cli/blob/3bcb3c36661c87e5dfdd91b2fce19f8b29b31e01/package.json#L39-L41). 

In this case when npx install runs, the tree sitter modules have not be pulled; they don't happen as a part of bundling since they are non js code and they don't get installed when `npm install` runs against main package.json as they aren't declared there. Thus in our original PR we saw errors trying to NPX install.

The fix for this is to move the declaration of the native modules to the main package see https://github.com/google-gemini/gemini-cli/pull/840. This ensures the modules are present via `npm install` on a developers machine when coding locally and also during an install via NPX; the `npm install` step there sees the dependencies in the main package and pulls the appropriate modules.

### Sandboxing

This however does not help us in a sandboxing environment. Here running in docker or podman we again, fail to include the treesitter dependencies for much the same reason as NPX. Only the fix is different. In the docker file we don't actually build the code at all. 

Current behavior is that we build the bundle on the developer/build machine and copy those bundles into the docker file.  We do this differently than NPX though and here we bundle both the CLI and the Core individually and install them as global node packages. None of the npm commands are run locally in the docker container. 

https://github.com/google-gemini/gemini-cli/blob/3bcb3c36661c87e5dfdd91b2fce19f8b29b31e01/Dockerfile#L41-L45

## Tactical fix

Regarding sandboxing/containers, what we have today is more than reasonable for how the code base has evolved and have met our goals as they've evolved 👍. This is a good functional state. However, after going deep on this my understanding is that our sandboxing, docker builds and interaction between the two are at the root, in need of a strategic overhaul. Notes on that in a follow up discussion.

That strategic overhaul is not in scope for this PR nor, I am hopeful, before launch on June 24th. After several false starts around a rewrite, I've settled on a sub optimal but tinniest change possible. Given the native module dependencies have to live in the main package.json we'll use this as our own temp package, parsing those lines to do install for the packages required. This install happens in the docker container so we get the right architecture and the code runs.

https://github.com/google-gemini/gemini-cli/blob/cd12c2485593a0398f55e3f2a703a5200ede9fbe/Dockerfile#L47-L48

This fix, while 'clever', is much more brittle than acceptable and risks maintenance issues moving forward. I don't love it.

## Alternatives considered.
1. Rewrite the whole thing and build a proper docker file - Dismissed as not right now/to big
2. Create a separate file for tracking external depends - Seems duplicative and is one more thing to maintain

### Testing

|  | 🍏  | 🪟  | 🐧  |
|--------|--------|--------|--------|
| npm run | ❓ | ❓ | ❓ |
| npx | ❓ | ❓ | ❓ |
| Docker | ❓ |  ❓ |  ❓  | 
| Podman | ❓ | - | - |
| Seatbelt | ❓ | - | - | 

### How to test
I've been asking the llm to generate the AST for a given file and save it to a file or print it out.


mattKorwel:
🍏 + seatbelt

Screenshot 2025-06-14 at 8 06 22 PM

mattKorwel:
Will return to this later as an extension. closing for now.

PR #1043, opened 2025-06-14
feat(core): support .aiexclude

### LINKED ISSUE #653: Add the ability to add an exclude file e.g. `.aiexclude` in the settings

mrahs:
By simply making the filename configurable:
```
Loaded 17 patterns from /.../gemini-cli/.gitignore
Loaded 1 patterns from /.../gemini-cli/.aiexclude
```

Fixes #653

### DIFF
--- a/packages/core/src/utils/gitIgnoreParser.ts
+++ b/packages/core/src/utils/gitIgnoreParser.ts
@@ -30,30 +30,31 @@ export class GitIgnoreParser implements GitIgnoreFilter {
     this.addPatterns(['.git']);
 
     const patternFiles = ['.gitignore', path.join('.git', 'info', 'exclude')];
-    for (const pf of patternFiles) {
-      this.loadPatterns(pf);
-    }
+    this.loadPatterns(patternFiles);
   }
 
-  loadPatterns(patternsFileName: string): void {
-    const patternsFilePath = path.join(this.projectRoot, patternsFileName);
-    let content: string;
-    try {
-      content = fs.readFileSync(patternsFilePath, 'utf-8');
-    } catch (_error) {
-      // ignore file not found
-      return;
+  loadPatterns(patternsFileName: string | string[]): void {
+    const patternFiles = Array.isArray(patternsFileName)
+      ? patternsFileName
+      : [patternsFileName];
+    for (const pf of patternFiles) {
+      const pfp = path.join(this.projectRoot, pf);
+      let content: string;
+      try {
+        content = fs.readFileSync(pfp, 'utf-8');
+      } catch (_error) {
+        // ignore file not found
+        continue;
+      }
+      const patterns = (content ?? '')
+        .split('\n')
+        .map((p) => p.trim())
+        .filter((p) => p !== '' && !p.startsWith('#'));
+      if (patterns.length > 0) {
+        console.log(`Loaded ${patterns.length} patterns from ${pfp}`);
+      }
+      this.addPatterns(patterns);
     }
-    const patterns = (content ?? '')
-      .split('\n')
-      .map((p) => p.trim())
-      .filter((p) => p !== '' && !p.startsWith('#'));
-    if (patterns.length > 0) {
-      console.log(
-        `Loaded ${patterns.length} patterns from ${patternsFilePath}`,
-      );
-    }
-    this.addPatterns(patterns);
   }
 
   private addPatterns(patterns: string[]) {
--- a/scripts/sandbox_command.js
+++ b/scripts/sandbox_command.js
@@ -63,7 +63,8 @@ if (!geminiSandbox) {
   geminiSandbox = process.env.GEMINI_SANDBOX;
 }
 
-geminiSandbox = (geminiSandbox || '').toLowerCase();
+geminiSandbox =
+  typeof geminiSandbox === 'string' ? geminiSandbox.toLowerCase() : '';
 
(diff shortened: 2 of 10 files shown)


scottdensmore:
Not sure what the actual support is supposed to look like but ran from this branch and this is what happened. I see dockerfile showed up in red.. but it still read the file.


https://github.com/user-attachments/assets/f3f566f2-614f-4124-bef2-51312282bbd8

mrahs:
Interesting! My attempt was successful:
```
Loaded 1 patterns from /usr/local/google/home/mrahs/dev/gemini-cli/.aiexclude

$ cat .aiexclude

ℹ Dockerfile

> @Dockerfile what does this file do?

 ╭─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
 │ ✔  ReadManyFiles Will attempt to read and concatenate files using patterns: `Dockerfile` (within target directory: `/usr/local/google/home/mrahs/dev/gemini-cli`)…  │
 │                                                                                                                                                                     │
 │    ReadManyFiles Result (Target Dir: /usr/local/google/home/mrahs/dev/gemini-cli)                                                                                   │
 │                                                                                                                                                                     │
 │    No files were read and concatenated based on the criteria.                                                                                                       │
 ╰─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯

 ╭────────────────────────────────────────────────────────────────────╮
 │ ✔  ReadFile Dockerfile                                             │
 │                                                                    │
 │    File path 'Dockerfile' is ignored by .geminiignore pattern(s).  │
 ╰────────────────────────────────────────────────────────────────────╯
```
I wonder if your workspace needs a `npm run build` before `npm start`?

NTaylorMullen:
Should `.aiexclude` be an alias for `.geminiignore`?

scottdensmore:
Interesting. I am still able to read the file off your branch. Where and how is your settings file look?

https://github.com/user-attachments/assets/30114cd4-68a5-4e05-9255-1f6c8490b9c5

mrahs:
Ah, it's the sandbox. I'm not using one. I was able to reproduce when I turn it on. Will look into it.

> Should `.aiexclude` be an alias for `.geminiignore`?

You mean implicitly without being configurable as in this PR? I was thinking of other conventions like `.cursorignore`.
On second thought, the config should probably accept a list of aliases instead of one.
wdyt?

NTaylorMullen:
> On second thought, the config should probably accept a list of aliases instead of one.
wdyt?

This is where my heads at too. That's how we work with `contextFileName` as well (defaults to `GEMINI.md`)

NTaylorMullen:
Note: IF this is something you want to include in the official launch we'll need to cherry-pick it to the release branch + go through ask mode

mrahs:
> This is where my heads at too. That's how we work with `contextFileName` as well (defaults to `GEMINI.md`)

Done with a default to both .geminiignore and .aiexclude.

>Ah, it's the sandbox. I'm not using one. I was able to reproduce when I turn it on. Will look into it.
Fixed that:
```
using macos seatbelt (profile: permissive-open) ...
Loaded 18 patterns from /Users/mrahs/Dev/gemini-cli/.gitignore
Loaded 1 patterns from /Users/mrahs/Dev/gemini-cli/.geminiignore
Loaded 1 patterns from /Users/mrahs/Dev/gemini-cli/.aiexclude
```

mrahs:
> Note: IF this is something you want to include in the official launch we'll need to cherry-pick it to the release branch + go through ask mode

I'll let the release lead decide.

uryposario:
Thank you guys.

palladius:
So IIUC now we exclude BOTH  `.geminiignore` and `.aiexclude` right?

scripts/sandbox_command.js
```diff
@@ -63,7 +63,8 @@ if (!geminiSandbox) {
   geminiSandbox = process.env.GEMINI_SANDBOX;
 }
 
-geminiSandbox = (geminiSandbox || '').toLowerCase();
+geminiSandbox =
+  typeof geminiSandbox === 'string' ? geminiSandbox.toLowerCase() : '';
```
scidomino, line 67: This seems unrelated.

packages/core/src/utils/gitIgnoreParser.ts
```diff
@@ -30,30 +30,31 @@ export class GitIgnoreParser implements GitIgnoreFilter {
...
-    }
-    const patterns = (content ?? '')
-      .split('\n')
-      .map((p) => p.trim())
-      .filter((p) => p !== '' && !p.startsWith('#'));
-    if (patterns.length > 0) {
-      console.log(
-        `Loaded ${patterns.length} patterns from ${patternsFilePath}`,
-      );
+  loadPatterns(patternsFileName: string | string[]): void {
```
scidomino, line 36: Changing this method to take two different types is a bad code smell. 

Our settings file should only accept arrays and arrays should be used throuought. There is no backwards compatibility we have to worry about here.

mrahs:
This PR is outdated and doesn't reflect the practices that has since been adopted. I suggest opening another one for this FR if it's still desired. Closing.

PR #1046, opened 2025-06-14
bug: Fix code colorizer

ngleo:
Fix bug where: when adding a new file without an extension, DiffRenderer does not behave correctly.

Before:
Screenshot 2025-06-14 at 20 19 12

After:
Screenshot 2025-06-14 at 20 44 16


NTaylorMullen (approved):
Great bug fix!

PR #1050, opened 2025-06-14
Add support for /mcp schema to show full parameter schema as JSON

bbiggs:
Add support for a /mcp schema command to show the full parameter schema as provided to the model. Fixes #1049


NTaylorMullen (approved):
Can definitely see the value in this from a debuggability perspective. Also will make digging into MCP schemas damn easy.

Mind adding a few tests for these w/ GC?

packages/cli/src/ui/hooks/slashCommandProcessor.test.ts
```diff
@@ -1019,6 +1019,90 @@ Add any other context about the problem here.
...
+          type: MessageType.INFO,
+          text: expect.stringContaining('Configured MCP servers:'),
+        }),
+        expect.any(Number),
+      );
+
+      const message = mockAddItem.mock.calls[1][0].text;
+
+      // Check that server description is included (with ANSI color codes)
+      expect(message).toContain('\u001b[1mserver1\u001b[0m - Ready (2 tools)');
```
NTaylorMullen:
```suggestion
      expect(message).toContain('Ready (2 tools)');
```

packages/cli/src/ui/hooks/slashCommandProcessor.test.ts
```diff
@@ -1019,6 +1019,90 @@ Add any other context about the problem here.
...
+      expect(message).toContain(
```
NTaylorMullen: You probably only need 2 contains, does it contain "This is tool 1 description" and "This is tool 2 description"

The rest of the structure will evolve and we should be resilient to that

NTaylorMullen (approved):

packages/cli/src/ui/hooks/slashCommandProcessor.test.ts
NTaylorMullen: Actually param1 and param2 as well are probably ok

packages/cli/src/ui/hooks/slashCommandProcessor.test.ts
bbiggs: done, updated tests to only check content not formatting

PR #1052, opened 2025-06-15
Add support for showing descriptions of CLI tools

bbiggs:
Adds an option to `/tools` (`/tools desc`) that shows the description provided to the model for each CLI tool. Fixes #1051

### DIFF
--- a/packages/cli/src/ui/hooks/slashCommandProcessor.ts
+++ b/packages/cli/src/ui/hooks/slashCommandProcessor.ts
@@ -419,6 +419,21 @@ export const useSlashCommandProcessor = (
         name: 'tools',
         description: 'list available Gemini CLI tools',
         action: async (_mainCommand, _subCommand, _args) => {
+          // Check if the _subCommand includes a specific flag to control description visibility
+          let useShowDescriptions = showToolDescriptions;
+          if (_subCommand === 'desc' || _subCommand === 'descriptions') {
+            useShowDescriptions = true;
+          } else if (
+            _subCommand === 'nodesc' ||
+            _subCommand === 'nodescriptions'
+          ) {
+            useShowDescriptions = false;
+          } else if (_args === 'desc' || _args === 'descriptions') {
+            useShowDescriptions = true;
+          } else if (_args === 'nodesc' || _args === 'nodescriptions') {
+            useShowDescriptions = false;
+          }
+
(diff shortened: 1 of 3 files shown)


packages/cli/src/ui/hooks/slashCommandProcessor.ts
```diff
@@ -419,6 +419,21 @@ export const useSlashCommandProcessor = (
...
+          // Check if the _subCommand includes a specific flag to control description visibility
+          let useShowDescriptions = showToolDescriptions;
+          if (_subCommand === 'desc' || _subCommand === 'descriptions') {
+            useShowDescriptions = true;
+          } else if (
+            _subCommand === 'nodesc' ||
+            _subCommand === 'nodescriptions'
+          ) {
+            useShowDescriptions = false;
+          } else if (_args === 'desc' || _args === 'descriptions') {
```
NTaylorMullen, line 431: I'm now confused by our own command parser lol. I would have thought subCommand would have been the desc bit not the args. If you do `npm run debug` -> F5 to attach in VSCode which one of these do you see get hit?

packages/cli/src/ui/hooks/slashCommandProcessor.ts
```diff
@@ -432,11 +447,51 @@ export const useSlashCommandProcessor = (
...
           // Filter out MCP tools by checking if they have a serverName property
           const geminiTools = tools.filter((tool) => !('serverName' in tool));
-          const geminiToolList = geminiTools.map((tool) => tool.displayName);
+
+          let message = 'Available Gemini CLI tools:\n\n';
+
+          if (geminiTools.length > 0) {
+            geminiTools.forEach((tool) => {
+              if (useShowDescriptions && tool.description) {
+                // Format tool name in cyan using simple ANSI cyan color
```
NTaylorMullen, line 456: Appreciate the comment callouts

NTaylorMullen (approved):
Extra credit: Work with GC to add tests 😄

packages/cli/src/ui/hooks/slashCommandProcessor.ts
bbiggs: subCommand is hit because that's the first thing

/commandname subcommand-is-a-single-word and args is everything that comes after that

PR #1054, opened 2025-06-15
Create pull_request_template.md

mattKorwel:
:) adding a template so pr descriptions aren't empty ;)


NTaylorMullen (approved):
We can give this a shot BUT I do worry that it's a bit too much of a requirement for PR creators

PR #1057, opened 2025-06-15
fix: Push tool calls to absolute paths (#1055)

keir:
Make several changes to guide the model to request absolute paths, reducing frequent accidental relative path tool call failures.

- Switch the parameter name: path --> abs_path.
- Update the tool definition to strongly require an absolute path.
- Update the system prompt to indicate absolute paths are required.
- Update the system prompt tool use examples to use absolute paths.

Test case:

Open GC in GC: "Locate the primary file calling genai"
Expected: Model opens files with absolute path, successfully.
Actual (pre-patch): Failure, attempts to read with relative path.
Actual (post-patch): Success, attempts to read with asolute path.

### DIFF
--- a/packages/core/src/tools/read-file.ts
+++ b/packages/core/src/tools/read-file.ts
@@ -97,7 +97,10 @@ export class ReadFileTool extends BaseTool<ReadFileToolParams, ToolResult> {
 
     const fileService = this.config.getFileService();
     if (fileService.shouldGeminiIgnoreFile(params.absolute_path)) {
-      const relativePath = makeRelative(params.absolute_path, this.rootDirectory);
+      const relativePath = makeRelative(
+        params.absolute_path,
+        this.rootDirectory,
+      );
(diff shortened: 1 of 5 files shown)


packages/core/src/tools/read-file.ts
```diff
@@ -18,7 +18,7 @@ export interface ReadFileToolParams {
   /**
    * The absolute path to the file to read
    */
-  path: string;
+  abs_path: string;
```
NTaylorMullen: Lets fully qualify to `absolute_path`, this is super clever though I like it

NTaylorMullen (approved):
After fully qualifying to `absolute_path` lets get this in. I really like it

packages/core/src/tools/read-file.ts
keir: All done.

NTaylorMullen:
Actually just realized, we'll want to keep an eye on this. Thinking being that the read-file tool now says absolute_path but none of the others do; meaning, will those fail? Something we'll want to watch

PR #1061, opened 2025-06-15
Provide up-to-date files with every request

keir:
This is an experimental branch that isn't intended to be merged, but more as a discussion point.

The main change (ignoring the file token count display) is to switch away from imperative ReadFile calls that bake a point-in-time snapshot into history, and instead have an active set of files that are always sent fresh on every request; history no longer contains full file snapshots.


NTaylorMullen:
One thing to note from some background I have in this area: Providing having files exist throughout history prior to model responses is actually a quality increasing function. That's typically how the models themselves are trained and you end up getting higher quality results because of it. For instance, when you have a coherent history where every "response" is dictated by a very specific input (a file snapshot) the models tend to do better. Now this could be a point-in-time thing where we change how we train models in the future to make this less-so BUT until then I do worry about the quality impact this approach a "always fresh file" approach might entail

mattKorwel:

PR #1063, opened 2025-06-15
Support completion of checkpoint names in /resume

bbiggs:
Support completion of checkpoint names in /resume. Fixes #1062


NTaylorMullen (approved):
Nifty. Feel free to get these in as-is.

For extra credit work with GC to add some tests for the new functionality 😉

PR #1064, opened 2025-06-15
Telemetry: Improve API response logging with function call details

jerop:
Improve API response handling to include structured function call data in logs. 

New utilities in `generateContentResponseUtilities.ts` extract function calls, which are now logged alongside text responses. This provides more complete telemetry for API interactions involving function calls.

---

Example API response log entry with function calls included:

image

And the tool calls that happened afterwards:
image

#750


bdmorgan (approved):

NTaylorMullen:
Very nice!

PR #1065, opened 2025-06-15
feat(cli): add glob as a direct dependency

NTaylorMullen:
- The  hook relies on  to provide file-based autocompletion. This commit adds  as a direct dependency to , ensuring it is available in the sandboxed environment and preventing  errors.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM docker.io/library/node:22-slim
+FROM docker.io/library/node:20-slim
 
 ARG SANDBOX_NAME="gemini-cli-sandbox"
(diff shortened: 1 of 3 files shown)


NTaylorMullen:
@steren reverting your docker change. Ended up breaking bash invocations. I'm a docker noob so we can potentially bring this back BUT reverting for now to unbreak sandboxing.

mattKorwel (approved):
Verified on Chrome and Mac that its working with this fix

steren:
Sorry about that, yes changing CMD to ENTRYPOINT can mean we now need to update the command to run the container (it depends on how it was invoked)

PR #1068, opened 2025-06-15
Add file operation telemetry

jerop:
Introduces telemetry for file create, read, and update operations.

This change adds the `gemini_cli.file.operation.count` metric, recorded by the `read-file`, `read-many-files`, and `write-file` tools.

The metric includes the following attributes:
    - `operation` (string: `create`, `read`, `update`): The type of file operation.
    - `lines` (optional, Int): Number of lines in the file.
    - `mimetype` (optional, string): Mimetype of the file.
    - `extension` (optional, string): File extension of the file.

Here is a stacked bar chart of file operations by extension (`js`, `ts`, `md`):
image

Here is a stacked bar chart of file operations by type (`create`, `read`, `update`):
image

#750 

cc @allenhutchison as discussed

### DIFF
--- a/packages/core/src/telemetry/metrics.ts
+++ b/packages/core/src/telemetry/metrics.ts
@@ -20,8 +20,15 @@ import {
   METRIC_API_REQUEST_LATENCY,
   METRIC_TOKEN_USAGE,
   METRIC_SESSION_COUNT,
+  METRIC_FILE_OPERATION_COUNT,
 } from './constants.js';
 import { Config } from '../config/config.js';
+
+export enum FileOperation {
+  CREATE = 'create',
+  READ = 'read',
+  UPDATE = 'update',
+}
 
 let cliMeter: Meter | undefined;
 let toolCallCounter: Counter | undefined;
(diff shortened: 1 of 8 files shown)


packages/core/src/telemetry/metrics.ts
```diff
@@ -171,3 +182,21 @@ export function recordApiErrorMetrics(
     model,
   });
 }
+
+export function recordFileOperationMetric(
```
NTaylorMullen, line 186: What if we created a `record(Read|Create|Update)` method and hide the FileOperation enum above from consumers. That way each method could take in say file content and file path and then do the right thing to ensure that all callers are getting their data represented in a consistent way + wouldn't leak the metric impls into any of the tools

NTaylorMullen (approved):

packages/core/src/telemetry/metrics.ts
jerop: Tracked in https://github.com/google-gemini/gemini-cli/issues/1069

PR #1072, opened 2025-06-15
refactor(cli): Use excludeTools for non-interactive mode

NTaylorMullen:
- Migrates the logic for disabling interactive tools in non-interactive mode to use the `excludeTools` setting.
- This corrects the previous implementation that was improperly filtering the `coreTools` list, ensuring that custom tool configurations are not overridden.
- Aligns the non-interactive mode with the intended design for tool exclusion.
- Updates a test case that was affected by the change in configuration loading.

Fixes https://b.corp.google.com/issues/425113548

### DIFF
--- a/packages/cli/src/gemini.tsx
+++ b/packages/cli/src/gemini.tsx
@@ -21,15 +21,7 @@ import {
   ApprovalMode,
   Config,
   EditTool,
-  GlobTool,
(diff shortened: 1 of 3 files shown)


mattKorwel (approved):
LGTM. Did not run the code.

PR #1076, opened 2025-06-15
feat: Adds shell command context to gemini history

abhipatel12:
## TLDR

PR focuses on giving the model awareness of manually run shell commands by the user.

## Dive Deeper

This commit introduces a significant enhancement to the shell command processing capabilities of the Gemini CLI. The primary goal is to provide the language model with context about shell commands executed by the user, leading to more informed and relevant interactions.

### Key Changes:

1.  **Contextual History for the Model:**
    *   The most important feature introduced is the `addShellCommandToGeminiHistory` function. After a user executes a shell command, both the command itself and its result are formatted and added to the conversation history with the Gemini model.
    *   This gives the model visibility into the user's manual command-line actions, allowing it to understand the state of the user's environment better and provide more accurate assistance.
    *   To prevent excessive context length, the output sent to the model is truncated if it exceeds a defined limit.

2.  **Improved Shell Command Execution:**
    *   The underlying execution mechanism has been migrated from `child_process.exec` to `child_process.spawn`. This provides more robust and granular control over subprocesses, enabling better real-time output streaming and process management.
    *   The new implementation gracefully handles various execution outcomes, including standard output, standard error, exit codes, and process signals (e.g., `SIGTERM`).
    *   Process cleanup has been improved. On POSIX-compliant systems (like Linux and macOS), it now terminates the entire process group, ensuring that child processes spawned by the command are also cleaned up correctly.

3.  **Binary Output Detection:**
    *   A new utility function, `isBinary`, has been added to detect whether the output of a command is binary data.
    *   If binary output is detected, it is not displayed in the CLI or sent to the model. This prevents garbled text in the UI and avoids polluting the model's context with unreadable data.

4.  **Refined User Experience:**
    *   The UI now provides real-time feedback for long-running commands by streaming the output as it's generated.
    *   Error messages are more descriptive, clearly indicating if a command failed due to a non-zero exit code, was terminated by a signal, or was canceled by the user.
    *   The existing warning about the stateless nature of `cd` commands remains, reminding users that directory changes do not persist across commands.

5.  **Comprehensive Testing:**
    *   The test suite for the shell command processor has been completely rewritten to cover the new `spawn`-based implementation.
    *   The tests now cover a wide range of scenarios, including successful execution, errors, signal handling, command cancellation (aborting), binary output, and output truncation.

#### Example/Demo:
Simple echo example:
echo-shell-context

Video Example:

https://github.com/user-attachments/assets/d8fc0a54-56c1-420b-9fa6-1e8725fa88b2

### Testing Strategy for Reviewers

To manually verify the changes, a reviewer can perform the following tests in the Gemini CLI:

**1. Test Contextual History for the Model:**

- **Step 1 Command:** `ls -l`
- **Step 2 Follow-up Question:** After the `ls -l` command completes, ask the model a question about its output, such as: "Based on the last command's output, what are the permissions for the `README.md` file?"
- **Expected Result:** The model should be able to correctly answer the question by referencing the output of the `ls -l` command, demonstrating that the command's result was successfully added to its context.

**2. Test Real-time Feedback:**

- **Command:** `for i in 1 2 3 4 5; do echo "Processing item $i..."; sleep 1; done`
- **Expected Result:** The output "Processing item..." should appear line-by-line, with a one-second delay between each line, demonstrating that output is being streamed in real-time.

**3 Test Binary Output Detection:**

- **Action:** Find an image file or any other non-text file in your project (e.g., `docs/assets/theme-ansi.png`).
- **Command:** `cat path/to/your/binary/file.png`
- **Expected Result:** Instead of printing garbled binary data to the screen, the CLI should display a message like "[Command produced binary output, which is not shown.]".

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅ | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ✅  | -   | -   |

## Linked issues / bugs

[b/424853926](https://b.corp.google.com/issues/424853926)

### DIFF
--- a/packages/cli/src/ui/hooks/shellCommandProcessor.ts
+++ b/packages/cli/src/ui/hooks/shellCommandProcessor.ts
@@ -173,12 +173,10 @@ function executeShellCommand(
 }
 
 function addShellCommandToGeminiHistory(
-  geminiClient: GeminiClient | null,
+  geminiClient: GeminiClient,
   rawQuery: string,
   resultText: string,
 ) {
-  if (!geminiClient) return;
-
   const modelContent =
     resultText.length > MAX_OUTPUT_LENGTH
       ? resultText.substring(0, MAX_OUTPUT_LENGTH) + '\n... (truncated)'
@@ -214,7 +212,7 @@ export const useShellCommandProcessor = (
   onExec: (command: Promise<void>) => void,
   onDebugMessage: (message: string) => void,
   config: Config,
-  geminiClient: GeminiClient | null,
+  geminiClient: GeminiClient,
 ) => {
   const handleShellCommand = useCallback(
     (rawQuery: PartListUnion, abortSignal: AbortSignal): boolean => {
--- a/packages/cli/src/ui/hooks/shellCommandProcessor.ts
+++ b/packages/cli/src/ui/hooks/shellCommandProcessor.ts
@@ -5,11 +5,13 @@
  */
 
 import { spawn } from 'child_process';
+import { StringDecoder } from 'string_decoder';
 import type { HistoryItemWithoutId } from '../types.js';
-import type { exec as ExecType } from 'child_process';
 import { useCallback } from 'react';
-import { Config } from '@gemini-cli/core';
+import { Config, GeminiClient } from '@gemini-cli/core';
 import { type PartListUnion } from '@google/genai';
+import { formatMemoryUsage } from '../utils/formatters.js';
+import { isBinary } from '../utils/textUtils.js';
 import { UseHistoryManagerReturn } from './useHistoryManager.js';
 import crypto from 'crypto';
 import path from 'path';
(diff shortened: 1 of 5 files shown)


packages/cli/src/ui/hooks/shellCommandProcessor.ts
```diff
@@ -31,227 +214,126 @@ export const useShellCommandProcessor = (
   onExec: (command: Promise<void>) => void,
   onDebugMessage: (message: string) => void,
   config: Config,
-  executeCommand?: typeof ExecType, // injectable for testing
+  geminiClient: GeminiClient | null,
```
NTaylorMullen: Possible to make required vs. nullable?

packages/cli/src/ui/hooks/shellCommandProcessor.ts
```diff
@@ -31,227 +214,126 @@ export const useShellCommandProcessor = (
...
-        );
-        return true; // Handled (by showing error)
-      }
+        onDebugMessage(`Executing in ${targetDir}: ${commandToExecute}`);
+        executeShellCommand(
+          commandToExecute,
+          targetDir,
+          abortSignal,
+          (streamedOutput) => {
+            // Throttle pending UI updates to avoid excessive re-renders.
```
NTaylorMullen, line 255: Ahhh this is one way. Another route you could do this is to do it kind of like the shell tool today where you can only really see the last X amount of content and the rest is hidden. I wonder if we should do that here too. That would mean no re-renders / flickering.

(Could be a follow up)

NTaylorMullen (approved):
Love the video, the description and all the effort you put into this making it super resilient! 👏 

Do you think there's ever a world where we'd be able to delegate some of this heavy lifting to the shell.ts tool? Or are the two experiences unique enough to not warrant?

abhipatel12:
> Love the video, the description and all the effort you put into this making it super resilient! 👏
> 
> Do you think there's ever a world where we'd be able to delegate some of this heavy lifting to the shell.ts tool? Or are the two experiences unique enough to not warrant?

Thank you!

Great question. There is def some complex SHARED low-level process execution logic. This could def be extracted and placed into some shared utility file that both the core and cli shell scripts can then rely on. This would help us with maintainability instead of worrying about keeping these two in sync. 

As for just using the `shell_tool` directly, that would take a lot more consideration. There are some significant differences in what they do. The impl in `slashCommandProcessor` does a lot of work to make the experience seem closer to a regular shell. It allows for custom UI to improve UX for certain states like errors. The `shell_tool` can also show errors but we rely on the backend to simply provide a string that is formatted by frontend UI component. 

The biggest issue is that the interactive CLI wraps the command to check the final working directory to provide its "stateless directory" warning. When letting the tool execute `cd` the tool does so and responds that it was successful. 

**TLDR:**
At first glance, I'm +1 for extracting the low-level execution logic (in a future PR) and -1 for directly using the `shell_tool` for the shell mode in the frontend.

packages/cli/src/ui/hooks/shellCommandProcessor.ts
abhipatel12: Oh yeah for sure! In this case would the user be able to see the full output or only the last part? 

I think because of the UX considerations here, I will leave this for a fast follow!

packages/cli/src/ui/hooks/shellCommandProcessor.ts
abhipatel12: Yes! Awesome catch. The existing null check in `useGeminiStream` must have made the GC agent think it had to account for a null but `config.getGeminiClient` never returns null.

Removed this and fixed the null check.

NTaylorMullen:
That completely tracks, appreciate you humoring me!

packages/cli/src/ui/hooks/shellCommandProcessor.ts
NTaylorMullen: I think my gut says only the last part BUT we potentially could even go one step further and have a hot key interaction to expand what's shown

PR #1094, opened 2025-06-16
[DO NOT MERGE] fix: Fix flickering when output is bigger than terminal window

tolikmalibroda:
# DO NOT MERGE

## TLDR

This PR tries to fix b/422574796 by updating `ink` library to the latest version.

## Dive Deeper

`ink` library has two parts of output:
1. `<Static>` block which should be rendered once.
2. Everything else is continuously changing, so is rendred dynamically.

At the same time, terminals have multiple modes of output:
1. Visible area in terminal window – we can clear as many lines as we want here.
2. Scrollback buffer – output that becames not visible, user can scroll up to see the history, but only way to modify it is to clear entirely.
3. Alternate screen – the mode that is used to display content without printing it into main terminal output (e.g. `less` command), is not currently used in `gemini-cli`.

There was a bug in `ink` library where renderer incorrectly tracked the height of the output which resulted in incorrect rendering when previous dynamic output was bigger than terminal window but the new dynamic output was smaller. This bug was caused by the lack of ability to clear scrollback buffer in terminal.

`ink` version 6.0.0 contains fix for this bug, so I decided to upgrade to it. However it also requires the migration to React 19, which caused some new bugs in `ink` library that affect us, i.e. https://github.com/vadimdemedes/ink/pull/723. This PR locks `ink` dependency at the git revision that also fixes this bug, but is not merged and included in the release yet.

Even after updating to the new version of `ink`, there's still a problem with long dynamic outputs that clear terminal and cause full re-render. In `gemini-cli` if the history becames relatively big, there are two reasons (maybe more) that cause the flickering:
1. Generating huge file diff for user's approval – when user changes the option, it causes re-render of full diff (and entire history), so it flickers if the previous output is big.
2. When executing shell command, we print only last lines of the command output to fit into the terminal window. However, if some lines of the output are too long, it causes line-wrapping that increases output area height above the terminal window height. Because of that, every second when such output gets re-rendered, it causes re-render of entire history which also flickers.

There's also barely noticable flashing of the screen which is not related to the full re-rendering of history, it appears when `gemini-cli` renders the output of shell command and the spinner that displays processing. Spinners is updated too frequently, so it causes frequent re-rendering of the dynamic area, which sometimes flashes a bit.

After such deep research I'd split this into 3 separate tasks that would address the issues described above. Updating to `ink` 6.0.0 would fix the bug with incorrect rendering, but won't fix the issue with flickering.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


NTaylorMullen:
@jacob314 could you help take a look here

packages/cli/package.json
```diff
@@ -37,7 +37,7 @@
     "dotenv": "^16.4.7",
     "glob": "^10.4.1",
     "highlight.js": "^11.11.1",
-    "ink": "^6.0.0",
+    "ink": "github:vadimdemedes/ink#bf8e084194d9b6e2ce9ad126c41782ebf29b6c8b",
```
NTaylorMullen, line 39: I had no clue you could even do this

packages/cli/src/gemini.tsx
```diff
@@ -85,14 +84,17 @@ export async function main() {
...
     render(
-      <React.StrictMode>
-        <AppWrapper
-          config={config}
-          settings={settings}
-          startupWarnings={startupWarnings}
-        />
-      </React.StrictMode>,
-      { exitOnCtrlC: false },
+      // <React.StrictMode>
```
NTaylorMullen: Why is strict mod dropping required?

packages/cli/src/ui/components/Spinner.tsx
```diff
@@ -0,0 +1,35 @@
...
+import { Text } from 'ink';
+
+interface SpinnerProps {
+  type?: SpinnerName;
+}
+
+export const Spinner: React.FC<SpinnerProps> = ({ type = 'dots' }) => {
+  const [frame, setFrame] = useState(0);
+  const spinner = spinners[type];
+  const interval = 1000;
```
NTaylorMullen: So a custom spinner that updates every 1 second right?

packages/cli/src/ui/hooks/useHistoryManager.ts
```diff
@@ -49,23 +49,10 @@ export function useHistory(): UseHistoryManagerReturn {
       const id = getNextMessageId(baseTimestamp);
       const newItem: HistoryItem = { ...itemData, id } as HistoryItem;
 
-      setHistory((prevHistory) => {
```
jacob314: was removing this intentional?

packages/cli/src/gemini.tsx
tolikmalibroda: StrictMode is for development, it causes some events to be invoked twice, which might be suboptimal and cause some flashing. However, I'm removing this change from this PR as I was just experimenting.

packages/cli/package.json
tolikmalibroda: That's also something new that I learned =)

packages/cli/src/ui/hooks/useHistoryManager.ts
tolikmalibroda: No, just experimenting, reverting dummy changes and leaving this PR only with `ink` dependency update

packages/cli/src/ui/components/Spinner.tsx
tolikmalibroda: Yes, but I moved it to separate PR: https://github.com/google-gemini/gemini-cli/pull/1137

jacob314:
Closing as we are now on Ink6. Thanks for helping us get there!

PR #1104, opened 2025-06-16
getPreviousUserMessages can be synchronous

sethtroisi:


jacob314 (approved):
lgtm

NTaylorMullen:
Note: This isn't currently going to be in the release because of the branch it's targeting

sethtroisi:
> Note: This isn't currently going to be in the release because of the branch it's targeting

Ack, just a small cleanup I have planned from a while back.

mattKorwel:
@sethtroisi should it also get merged to release? the further things diverge, the more work later. Are there p0 items we should be picking up?

PR #1107, opened 2025-06-16
Add an entry for `/bug` to the docs/cli/commands.md file

starsandskies:
## TLDR

Currently, there is not an entry for this particular command


allenhutchison (approved):

mattKorwel:
@starsandskies should it also get merged to release? the further things diverge, the more work later. Are there p0 items we should be picking up?

PR #1109, opened 2025-06-16
feat(cli): Improve Tab completion

sijieamoy:
**Improved Tab Completion for Slash Commands**: The behavior of the Tab key for slash command autocompletion right now submits the command immediately, leaving no time to enter the arguments e.g. for `/memory`. After this change, the tab completes the command and adds a space, without submitting it.

### DIFF
--- a/packages/cli/src/ui/App.tsx
+++ b/packages/cli/src/ui/App.tsx
@@ -295,6 +295,9 @@ const App = ({ config, settings, startupWarnings = [] }: AppProps) => {
         handleSlashCommand(newValue ? '/mcp desc' : '/mcp nodesc');
       }
     } else if (key.ctrl && (input === 'c' || input === 'C')) {
+      if (buffer.text.length > 0) {
+        buffer.setText('');
+      }
       handleExit(ctrlCPressedOnce, setCtrlCPressedOnce, ctrlCTimerRef);
     } else if (key.ctrl && (input === 'd' || input === 'D')) {
       if (buffer.text.length > 0) {
--- a/packages/cli/src/ui/App.tsx
+++ b/packages/cli/src/ui/App.tsx
@@ -295,9 +295,6 @@ const App = ({ config, settings, startupWarnings = [] }: AppProps) => {
         handleSlashCommand(newValue ? '/mcp desc' : '/mcp nodesc');
       }
     } else if (key.ctrl && (input === 'c' || input === 'C')) {
-      if (buffer.text.length > 0) {
-        buffer.setText('');
-      }
       handleExit(ctrlCPressedOnce, setCtrlCPressedOnce, ctrlCTimerRef);
     } else if (key.ctrl && (input === 'd' || input === 'D')) {
       if (buffer.text.length > 0) {
--- a/packages/cli/src/ui/components/InputPrompt.tsx
+++ b/packages/cli/src/ui/components/InputPrompt.tsx
@@ -115,20 +115,10 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
       const selectedSuggestion = completionSuggestions[indexToUse];
 
       if (query.trimStart().startsWith('/')) {
-        const parts = query.trimStart().substring(1).split(' ');
-        const commandName = parts[0];
         const slashIndex = query.indexOf('/');
         const base = query.substring(0, slashIndex + 1);
-
-        const command = slashCommands.find((cmd) => cmd.name === commandName);
-        if (command && command.completion) {
-          const newValue = `${base}${commandName} ${selectedSuggestion.value}`;
-          buffer.setText(newValue);
-        } else {
-          const newValue = base + selectedSuggestion.value;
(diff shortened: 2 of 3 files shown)


scidomino (approved):

packages/cli/src/ui/App.tsx
```diff
@@ -264,6 +264,9 @@ const App = ({ config, settings, startupWarnings = [] }: AppProps) => {
         handleSlashCommand(newValue ? '/mcp desc' : '/mcp nodesc');
       }
     } else if (key.ctrl && (input === 'c' || input === 'C')) {
+      if (buffer.text.length > 0) {
```
jacob314: please revert this change . this functionality already exists mapped to ctrl-u

packages/cli/src/ui/components/InputPrompt.tsx
```diff
@@ -109,20 +109,10 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
...
-        const command = slashCommands.find((cmd) => cmd.name === commandName);
-        if (command && command.completion) {
-          const newValue = `${base}${commandName} ${selectedSuggestion.value}`;
-          buffer.setText(newValue);
-        } else {
-          const newValue = base + selectedSuggestion.value;
-          buffer.setText(newValue);
-          handleSubmitAndClear(newValue);
-        }
+        const newValue = `${base}${selectedSuggestion.value} `;
```
jacob314, line 121: this change also regresses autocomplete for cases where commands should be submitted. if you want to fix this, please add metadata for the commands to indicate which commands take arguments and which don't. there is a pending pull request for user defined commands that may also address this.

jacob314 (commented):
Thanks for the pull request! I'm a bit concerned about making `ctrl-c` clear the current line as that is dangerously close to pressing `ctrl-c` twice to exit.
We already support `ctrl-u` to clear the current line which is consistent with most terminal on linux and mac.

packages/cli/src/ui/components/InputPrompt.tsx
sijieamoy: `tab` to me is associated with autocomplete instead of autocomplete+submit, so it is not exactly a regression for the commands that don't take arguments. Users should `tab` to complete then `enter` to submit.

sijieamoy:
> Thanks for the pull request! I'm a bit concerned about making `ctrl-c` clear the current line as that is dangerously close to pressing `ctrl-c` twice to exit. We already support `ctrl-u` to clear the current line which is consistent with most terminal on linux and mac.

Thanks for raising this. Makes sense. I have a different opinion on the `tab` behavior - please see the inline comment above.

packages/cli/src/ui/App.tsx
sijieamoy: Actually, `ctrl-u` only clears the current line, so it doesn't really solve the multi-line use case. Any ideas?

packages/cli/src/ui/components/InputPrompt.tsx
NTaylorMullen: > Users should tab to complete then enter to submit.

I agree with this.

sijieamoy:
@jacob314 @NTaylorMullen to keep the ball rolling, I reverted the `ctrl-c` change (will probably do a separate PR) so this PR is only for the `tab` behavior.

NTaylorMullen:
@sijieamoy can you re-target this to the release branch as well.

@jacob or @dewitt could I get one of your reviews as experts in the area. My personal take is making sure tab completion feels natural is an easy win and the change here is small enough to warrant taking in release

sijieamoy:
> @sijieamoy can you re-target this to the release branch as well.

Done!

mattKorwel:
@sijieamoy is this related to any of the release priority work captured here go/gemini-cli-p0 - if not is is stricly required for release? If so it should be based against the release branch.

sijieamoy:
> @sijieamoy is this related to any of the release priority work captured here go/gemini-cli-p0 - if not is is stricly required for release? If so it should be based against the release branch.

It's not. I was following @NTaylorMullen 's recommendation above. It's an annoying detail but definitely don't want to complicate the release. PLMK what y'all decide.

scottdensmore (approved):
Tested on my mac and love it. Playing around on linux and windows. But LGTM

packages/cli/src/ui/App.tsx
jacob314: good point. given something separate is needed for the multi-line case I am ok with ctrl-C

packages/cli/src/ui/components/InputPrompt.tsx
jacob314: agree having tab auto-submit is surprising. I'm good to remove that.

jacob314 (commented):
lgtm
Looks like the commit is pointing to main instead of release. please rebase your change on main before merging.

Lee-WonJun:
Is this running on the MAIN branch now?

PR #1111, opened 2025-06-16
Clear out untracked files when restoring a checkpoint

louisjimenez:
This addresses a bug where a new file that was written by the Write tool would not be removed when restoring a checkpoint to the time right before that file was created. This clears out any untracked files when restoring a checkpoint, resetting the working tree to how it was at that moment. 

1. Run the cli with checkpoints enabled `npm run start -- --checkpoint`
2. Trigger a file write tool call. Example prompt "Write Foo.md. The content should be a poem"
3. After the tool has run, use the `/restore` command to view available checkpoints.
4. Call the `/restore` command with a checkpoint name.
5. The recently added `Foo.md` should be removed and conversation state should be reset to right before the Write File tool was run.


NTaylorMullen (approved):

louisjimenez:
Going to move from the fork to opening up PRs directly.

PR #1112, opened 2025-06-16
Copy the telemetry binary to avoid a cross device link error when trying to rename files

anj-s:
Let me know if there is a reason we don't want to copy the otel binary.


jerop (approved):

mattKorwel:
@anj-s @jerop should it also get merged to release? the further things diverge, the more work later.

anj-s:
> @anj-s @jerop should it also get merged to release? the further things diverge, the more work later.

@mattKorwel Yes, its was on my list to merge to release once the PR was approved.

mattKorwel:
the way we are tackling that is if you make this PR against release instead of Main then we are going to regularly handle merging of released back into main. So our ask is all PRS for the release go against the release Branch please. 

* I'm on my phone and I see that this already merged so that's totally cool. Just note for next time.

PR #1113, opened 2025-06-16
Fix bug inserting text in the middle of a line

jacob314:
Previously we were dropping all text after the insert for single line inserts


NTaylorMullen:
@jacob314 were you looking to have this in the release candidate?

.github/CODEOWNERS
```diff
@@ -0,0 +1,7 @@
+# By default, require reviews from the release approvers for all files.
+* @google-gemini/gemini-cli-askmode-approvers
```
mattKorwel, line 2: is this what we want for all files? we already have this on PR's to the release branch. SO for now we'll just be requiring this on main? Maybe go with the whole teams here

We have a separate effort to codify all the "how we want to oss work" and i'm certain an overall approval strategy will come out of that.

packages/core/src/code_assist/codeAssist.ts
mattKorwel, line 1: What is the relevance of the code assist changes here? seems unrelated to this pr?

mattKorwel (changes requested):
Hey @jacob314 I think it would be super useful to have a bit more info. Are these changes to CCPA calling code intentional? If so, can we get more context on why they are neccessary and how we validated and tested them? Is there a b/# that this is related to?

Thanks and sorry for asking so many questions. I think the fix itself looks good.

Oh, I would love if you could add an `integration-test` around this behavior seems like a great one to capture. Happoy pair on that if you'd like.

🙇 thanks!

PR #1114, opened 2025-06-16
Fix bug where single line inserts were deleting all text after the in…

jacob314:
Previously we were dropping all text after the insert for single line inserts


dewitt (approved):
Thanks for the quick fix!

NTaylorMullen (approved):

PR #1115, opened 2025-06-16
Fix edit tool retrigger after modify

ngleo:
## Background
#808 enables users to modify the proposed content when the edit tool is triggered. However, Gemini is not aware that the user has modified the proposed content. When the tool has executed, Gemini sometimes reads the file to verify the change. When it sees an unexpected change by the user, it might retrigger the tool.

## Fix
This PR changes the edit tool & write tool to conditionally add a "content has been modified" message in the response to the LLM if the user goes through the modify flow.

With this approach, I can see Gemini's thought / response below and Gemini has not retriggered edit tool so far:

`
I can confirm that the file has been updated. The first line of CONTRIBUTING.md now reads \"XYZ\" It seems the user overrode the original request, and I've processed the  modified content accordingly. I will now proceed with the updated state of the file.
`

`
I have changed the first line of CONTRIBUTING.md to "Hello" as you requested. The edit was modified by you.
`

## Reviewer Test Plan
- Ask Gemini to edit a file with a very specific change. e.g. @{some file} change the 1st line to {something}
- Trigger modify flow, modify the proposed change (can be a very different change), and see if it retriggers the tool.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓ | -   | -   |

## Linked issues / bugs
- Fixes b/424920326.


packages/core/src/tools/edit.ts
```diff
@@ -414,9 +426,12 @@ Expectation for required parameters:
         displayResult = { fileDiff, fileName };
       }
 
+      const modfiedWarning = params.modified_by_user
+        ? " The proposed edit content was overriden by the user. This is intended and should not be corrected or reverted. When reporting the task completion, ensure to mention the user's modification."
```
jacob314: I wonder if we should modify the core system prompt in packages/core/src/core/prompts.ts with instructions about what to do when proposed edits are modified by the user so that this message can just be
"The proposed edit content was modified by the user."

packages/core/src/tools/write-file.ts
```diff
@@ -80,6 +87,11 @@ export class WriteFileTool
             description: 'The content to write to the file.',
             type: 'string',
           },
+          modified_by_user: {
+            type: 'boolean',
+            description:
+              'Whether the content was overridden by the user. This should default to false and only be set by the tool.',
```
jacob314: nit: should we be saying modified rather than "overridden" here and elsewhere to clarify that the user likely modified the response slightly rather than completely overriding it.

packages/core/src/tools/write-file.ts
```diff
@@ -270,9 +282,12 @@ export class WriteFileTool
         DEFAULT_DIFF_OPTIONS,
       );
 
+      const modfiedWarning = params.modified_by_user
+        ? " The new content was overriden by the user. This is intended and should not be corrected or reverted. When reporting the task completion, ensure to mention the user's modification."
```
jacob314: would also be good to simplify this message to just state what happened relying on they system prompt to ensure the model treads the modifications appropriately.

jacob314 (commented):
Thanks for the polish!

packages/core/src/tools/edit.ts
ngleo: Sure, let me try that out

packages/core/src/tools/edit.ts
```diff
@@ -111,6 +118,11 @@ Expectation for required parameters:
               'Number of replacements expected. Defaults to 1 if not specified. Use when you want to replace multiple occurrences.',
             minimum: 1,
           },
+          modified_by_user: {
```
NTaylorMullen: Why surface this the model?

packages/core/src/tools/edit.ts
```diff
@@ -81,6 +86,8 @@ export class EditTool
       'Edit',
       `Replaces text within a file. By default, replaces a single occurrence, but can replace multiple occurrences when \`expected_replacements\` is specified. This tool requires providing significant context around the change to ensure precise targeting. Always use the ${ReadFileTool.Name} tool to examine the file's current content before attempting a text replacement.
 
+      The user has the ability to modify the proposed content. If modified, this will be stated in the tool response.
```
NTaylorMullen:
```suggestion
The user has the ability to modify the \`new_string\` content. If modified, this will be stated in the response.
```

packages/core/src/tools/edit.ts
```diff
@@ -92,7 +99,7 @@ Expectation for required parameters:
         properties: {
           file_path: {
             description:
-              "The absolute path to the file to modify. Must start with '/'.",
+              "The absolute path to the file to edit. Must start with '/'.",
```
NTaylorMullen: why modify -> edit?

packages/core/src/tools/edit.ts
NTaylorMullen: How do you all feel about utilizing the llmContent to surface the actual content changes that the user made? I.e. state that Successfully .... User modified the \`new_string\` content to be: ....

packages/core/src/tools/write-file.ts
NTaylorMullen, line 1: Ditto comments in this file

NTaylorMullen (commented):
How easy would it be to test the modify commands?

mattKorwel:
Also @ngleo I think this would be a great candidate to capture in the `integration_tests` directory on a real end to end run though.

packages/core/src/tools/edit.ts
ngleo: The model doesn't have to know about this. You're right. Will fix.

packages/core/src/tools/edit.ts
ngleo: I tried to change the wording a bit to distinguish between edit (what the tool does) and modify (the specific user flow). Haven't noticed any side effects yet. Do you have any concerns?

(note that I also changed the tool name to edit_file)

packages/core/src/tools/edit.ts
ngleo: Could give it a try. Let's see how it behaves.

ngleo: Result seems really good - even when modifying to something unreasonable! Don't even need to add system prompts.

ngleo:
@mattKorwel Sure, I will take a look as a follow up. But note that this modify edit flow is interactive (involves triggering the flow with 3rd option in confirmation box, and then modifying the edit diff in an IDE). Might be tricky.

packages/core/src/tools/write-file.ts
ngleo: Done.

packages/core/src/tools/edit.ts
ngleo: Updated

packages/core/src/tools/write-file.ts
ngleo: This param definition is now removed as the model doesn't need to know about it

ngleo:
@NTaylorMullen @jacob314 I have addressed the comments.

packages/core/src/tools/edit.ts
NTaylorMullen: Ah I see, I'm not sure if we can take those prompting bits. That might be too significant to get in for the launch /cc @anj-s

NTaylorMullen:
> @NTaylorMullen @jacob314 I have addressed the comments.

Thank you! Have one last big one: https://github.com/google-gemini/gemini-cli/pull/1115#discussion_r2153648754

I'm not sure we can afford to take prompting changes as significant as that this close to release. Is that necessary?

ngleo:
> Thank you! Have one last big one: https://github.com/google-gemini/gemini-cli/pull/1115#discussion_r2153648754

Sorry I missed this. Testing the modify flow is covered by unit tests, except for the spawning/closing of IDE, which are OS dependent. Hence, e2e will be harder in terms of coverage.

Testing the fix of this PR specifically will involve some human judgement (e.g. looking at Gemini's thoughts and response).

> I'm not sure we can afford to take prompting changes as significant as that this close to release. Is that necessary?

I think the impact of this bug is not huge. Users can always select no when the tool retriggers. Also, this only happens when the prompt is very specific. If it's so specific, the user might have opted to do the change themselves. I think it is reasonable to not include it in the release.

jacob314:
Agree lets hold off including this in the release and instead land it on main.

ngleo:
@jacob314 @NTaylorMullen 

I have removed the wording changes in prompts and have pointed this to `main`

ngleo:
Closing this in favour of #2306

PR #1120, opened 2025-06-16
feat(prompt): Allow --prompt to accept a file path

allenhutchison:
This PR implements the feature requested in #1099, allowing the --prompt argument to accept a file path as input.

### DIFF
--- a/packages/cli/src/utils/prompt.ts
+++ b/packages/cli/src/utils/prompt.ts
@@ -1,0 +1,43 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import fs from 'node:fs';
+import path from 'node:path';
+
+/**
+ * Resolves a prompt. If the input string is a valid path to an existing
+ * file, its content is read and returned. Otherwise, the original string
+ * is returned.
+ *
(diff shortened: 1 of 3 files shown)


packages/cli/src/utils/prompt.ts
```diff
@@ -0,0 +1,43 @@
...
+  promptOrPath: string,
+  workingDir: string,
+): string {
+  // An empty string can never be a file, so return early.
+  if (!promptOrPath) {
+    return promptOrPath;
+  }
+
+  try {
+    const potentialPath = path.resolve(workingDir, promptOrPath);
```
NTaylorMullen, line 29: Should we do checks to see if the prompt is absolute or relative?

packages/cli/src/utils/prompt.ts
NTaylorMullen, line 1: nit: Instead of prompt.ts could we change this file + the .test.ts file to be something like `promptUtils` or `queryHelpers`.  I worry that the extra prompt.ts will add confusion for OSS viewers + the tools themselves

NTaylorMullen (approved):
This will be really nice for evals too. My gut says this change isn't needed for launch (not targeting the release branch) BUT if @anj-s says this would make lives a lot easier by being in the release branch for eval purposes we should consider retargeting

PR #1128, opened 2025-06-17
Update prompt to allow for the agent to verify changes as part of the implementation

anj-s:
Tested it on a few sample problems. We will want to follow up with more structured workflows in the future if we want this to be enforced.


mattKorwel:
Hey @anj-s if you want this in the release, you'll need to git rebase release as we are shipping out of that branch. Also please label this PR as June 25th Release and add the OSS MileStone.

If its not part of the release can you label it `Post-twenty-fifth`?

Thanks much!

NTaylorMullen (approved):

PR #1137, opened 2025-06-17
fix: Make spinner update less frequently to reduce flickering

tolikmalibroda:
## TLDR

Trying to reduce render flashing by slowing down updating of `<Spinner>` component. 

## Dive Deeper

To do it, I've implemented custom `<Spinner>` component instead of using `ink-spinner` and I've set update frequency to 1 second. However flashing still appears when shell command output gets updated because update of spinner and update of shell command output are not synchronized.

## Reviewer Test Plan

Execute any command and notice how frequent spinner updating is.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ✅ |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

N/A


packages/cli/src/ui/components/Spinner.tsx
```diff
@@ -0,0 +1,35 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { SpinnerName } from 'cli-spinners';
+import React, { useEffect, useState } from 'react';
```
jacob314, line 8: fix multiple import issue.

packages/cli/src/ui/components/Spinner.tsx
```diff
@@ -0,0 +1,35 @@
...
+import { Text } from 'ink';
+
+interface SpinnerProps {
+  type?: SpinnerName;
+}
+
+export const Spinner: React.FC<SpinnerProps> = ({ type = 'dots' }) => {
+  const [frame, setFrame] = useState(0);
+  const spinner = spinners[type];
+  const interval = 1000;
```
jacob314, line 19: comment why a slow interval rather than the default from from the spinner.

jacob314 (dismissed):
This is reasonable to mitigate the impact of flicker.
lgtm

NTaylorMullen:
@tolikmalibroda are we hoping to get this in as part of the launch as part of the flicker work? I.e. is this a required component of ensuring things don't flicker as poorly?

If so would you mind rebasing onto the release branch?

tolikmalibroda:
> @tolikmalibroda are we hoping to get this in as part of the launch as part of the flicker work? I.e. is this a required component of ensuring things don't flicker as poorly?
> 
> If so would you mind rebasing onto the release branch?

It definitely makes flashing less noticeable, but still not zero. Here is comparison.

Without fix:
https://github.com/user-attachments/assets/d72f3ea9-bcb4-4054-9ff1-f7ac21d6a38d

With fix:
https://github.com/user-attachments/assets/e0e42954-5970-4faa-b9ff-05de1c2c1284

So I'm not sure if we want it in this state into release, WDYT?
If so, should I create separate PR to the release branch?

packages/cli/src/ui/components/Spinner.tsx
tolikmalibroda: Done.

packages/cli/src/ui/components/Spinner.tsx
tolikmalibroda: Done.

NTaylorMullen:
@tolikmalibroda I love how less flickery it is BUT I think that the spinner in the loading dialog feels a bit too sluggish. Is there a middleground debounce?

jacob314:
I'm not sure we need this now that we have flicker generally under control unless you press ctrl-s to expand content. Given that, I'd suggest we only trigger the slow mode when we are in a state where we know there will be flicker. this can be achieved by measuring the size of the app in App.tsx. We could also approximate it in the short term by looking at the value of `constrainHeight` in app.tsx.

PR #1145, opened 2025-06-17
Move the logs.json to a project specific user home dir

louisjimenez:
Moves the logs.json file from being written to the `.gemini` folder in the user's project to the home directory. There will be other follow up PRs to move other features that are writing to the project level .gemini folder. Everything that the user may want to eventually clear is being grouped under a tmp folder. An example path `/Users/my-user/.gemini/tmp/<project-hash>/logs.json

### DIFF
--- a/packages/core/src/core/logger.ts
+++ b/packages/core/src/core/logger.ts
@@ -5,10 +5,13 @@
  */
 
 import path from 'node:path';
+import os from 'node:os';
+import crypto from 'node:crypto';
 import { promises as fs } from 'node:fs';
 import { Content } from '@google/genai';
 
 const GEMINI_DIR = '.gemini';
+const TMP_DIR_NAME = 'tmp';
 const LOG_FILE_NAME = 'logs.json';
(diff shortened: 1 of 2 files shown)


packages/core/src/core/logger.ts
```diff
@@ -95,7 +98,18 @@ export class Logger {
...
       return;
     }
-    this.geminiDir = path.resolve(process.cwd(), GEMINI_DIR);
+
+    const projectHash = crypto
+      .createHash('sha256')
+      .update(process.cwd())
+      .digest('hex');
+
+    this.geminiDir = path.join(
```
NTaylorMullen, line 107: We should probably expose this on the config actually so you can use it more broadly

NTaylorMullen (approved):

packages/core/src/core/logger.ts
louisjimenez: Good point, will do in follow up PR.

PR #1156, opened 2025-06-17
refactor: consolidate container image tag source of truth to cli package.json

KeijiBranshi:
Consolidating to a single source of truth for the target sandbox image name. This should hopefully reduce confusion around which image we should be hopping into. 

## Details

* `packages/cli/src/config/sandboxConfig.ts`
  The sandbox configuration logic is updated to require an explicit image URI, removing the hardcoded fallback. The function now returns a valid configuration only if both a sandbox command and an image are explicitly defined.

* `scripts/build_sandbox.js`
  This build script is updated to read the base sandbox image URI directly from the CLI's package.json. This change ensures the script builds the image using the same tag that the application expects, making the package configuration the single source of truth.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅   | ❓  | ✅  |
| npx      | ❓  | ❓  | ✅  |
| Docker   | ❓  | ❓  | ✅   |

### DIFF
--- a/packages/cli/src/config/sandboxConfig.ts
+++ b/packages/cli/src/config/sandboxConfig.ts
@@ -85,18 +85,13 @@ export async function loadSandboxConfig(
   argv: SandboxCliArgs,
 ): Promise<SandboxConfig | undefined> {
   const sandboxOption = argv.sandbox ?? settings.sandbox;
-  const sandboxCommand = getSandboxCommand(sandboxOption);
-  if (!sandboxCommand) {
-    return undefined;
-  }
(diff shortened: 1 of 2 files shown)


scidomino (dismissed):

NTaylorMullen (dismissed):

KeijiBranshi:
fyi, going to change the base to the release branch then create a cherry pick for this into main. unfortunately, this will reset the reviews 😢

scottdensmore (approved):
Approved

PR #1157, opened 2025-06-17
Support logging in with Application Default Credentials

scidomino:
## TLDR

Previously CCPA would only be supported with Oauth. Now, we also support looking for the gcloud application default credentials.

## Dive Deeper

The verbage about the two kinds of users was pulled from this: https://developers.google.com/gemini-code-assist/resources/faqs#gcp-project-requirement

## Reviewer Test Plan

1. Run this to revoke your Default Application credentials:  `gcloud auth application-default revoke;`
2. Remove any oauth credentials you might have: `rm ~/.gemini/oauth_creds.json`
3. Enable Code assist with  `export GEMINI_CODE_ASSIST=true`
4. Unset the cloud project `unset GOOGLE_CLOUD_PROJECT`
5. Start up Gemini CLI and log in with a gmail.com account to verify that Personal accounts work.
6. Shut down the CLI.
7. Remove any oauth credentials you might have: `rm ~/.gemini/oauth_creds.json`
8. Log in with gcloud: `gcloud auth application-default login`
9. Set the cloud project `export GOOGLE_CLOUD_PROJECT=aipp-internal-testing`
10. Start up Gemini CLI and verify that you are logged in without having to do oauth.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | x  | ❓  | ❓  |
| npx      | x  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt |  x  | -   | -   |

## Linked issues / bugs

b/425360849

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,19 +33,26 @@ The Gemini CLI requires you to authenticate with Google's AI services. You'll ne
 1.  **Gemini Code Assist:**
 
     - To enable this mode you only need set the GEMINI_CODE_ASSIST environment variable to true.
-    - Enterprise users must also provide a GOOGLE_CLOUD_PROJECT environment variable specifying their Google Cloud project.
-      In the following methods, replace `GOOGLE_CLOUD_PROJECT` with the relevant values for your project:
       - You can temporarily set the environment variable in your current shell session using the following command:
         ```bash
         export GEMINI_CODE_ASSIST="true"
-        export GOOGLE_CLOUD_PROJECT="YOUR_PROJECT_ID" // Enterprise users only.
         ```
       - For repeated use, you can add the environment variable to your `.env` file (located in the project directory or user home directory) or your shell's configuration file (like `~/.bashrc`, `~/.zshrc`, or `~/.profile`). For example, the following command adds the environment variable to a `~/.bashrc` file:
         ```bash
         echo 'export GEMINI_CODE_ASSIST="true"' >> ~/.bashrc
-        echo 'export GOOGLE_CLOUD_PROJECT="YOUR_PROJECT_ID"' >> ~/.bashrc # Enterprise users only.
         source ~/.bashrc
         ```
+    - There are two types of Google Accounts you might use to try to use Gemini Code Assist:
+      - **Personal Google Account**: This is the standard, free account you use for services like Gmail, Google Photos, and Google Drive for personal use (e.g. your-name@gmail.com).
+      - **Google Workspace Account**: This is a paid service for businesses and organizations that provides a suite of productivity tools, including a custom email domain (e.g. your-name@your-company.com), enhanced security features, and administrative controls. These accounts are often managed by an employer or school.
+        - Google Workspace Account must configure a Google Cloud Project Id to use. This can be done by:
+          - Using gcloud to set `gcloud config set project <your project id>`
+          - Setting the GOOGLE_CLOUD_PROJECT environment variable.
+    - During start up, Gemini CLI will look for cached credentials.
+      - If it finds Application Default Credentials (set up with `gcloud auth application-default login`) it will use them.
+      - Otherwise, if it find unexpired cached Oauth2 credentials it will use them.
+      - Otherwise, it will prompt you to visit a URL to authenticate (using Oauth2). Once authenticated, your Oauth2 credentials will be cached locally. Credentials normally expire after about 20 hours.
+    - Note that the Oauth2 login must be done in a browser that can communicate with the machine Gemini Cli is being run from. (Specifically, the browser will be redirected to a localhost url that Gemini CLI will be listening on).
 
 2.  **Gemini API key:**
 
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -33,19 +33,32 @@ The Gemini CLI requires you to authenticate with Google's AI services. You'll ne
...
         export GEMINI_CODE_ASSIST="true"
-        export GOOGLE_CLOUD_PROJECT="YOUR_PROJECT_ID" // Enterprise users only.
         ```
       - For repeated use, you can add the environment variable to your `.env` file (located in the project directory or user home directory) or your shell's configuration file (like `~/.bashrc`, `~/.zshrc`, or `~/.profile`). For example, the following command adds the environment variable to a `~/.bashrc` file:
         ```bash
         echo 'export GEMINI_CODE_ASSIST="true"' >> ~/.bashrc
-        echo 'export GOOGLE_CLOUD_PROJECT="YOUR_PROJECT_ID"' >> ~/.bashrc # Enterprise users only.
         source ~/.bashrc
         ```
+    - There are two types of Google Accounts you might use to try to use Gemini Code Assist:
```
NTaylorMullen:
```suggestion
    - There are two types of Google Accounts you can use with Gemini CLI:
```

README.md
```diff
@@ -33,19 +33,32 @@ The Gemini CLI requires you to authenticate with Google's AI services. You'll ne
...
+        - Google Workspace Account must configure a Google Cloud Project Id to use. You can temporarily set the environment variable in your current shell session using the following command:
+        ```bash
+        export GOOGLE_CLOUD_PROJECT_ID="YOUR_PROJECT_ID"
+        ```
+        - For repeated use, you can add the environment variable to your `.env` file (located in the project directory or user home directory) or your shell's configuration file (like `~/.bashrc`, `~/.zshrc`, or `~/.profile`). For example, the following command adds the environment variable to a `~/.bashrc` file:
+        ```bash
+        echo 'export GOOGLE_CLOUD_PROJECT_ID="YOUR_PROJECT_ID"' >> ~/.bashrc
+        source ~/.bashrc
+        ```
+    - During start up, Gemini CLI will look for cached credentials.
```
NTaylorMullen: This might be too in-the-weeds for folks, also, do Application Default Credentials refer to the above workspace account?

User's reading the README.md might treat Oauth2 etc. as an impl detail.

packages/core/src/code_assist/oauth2.ts
```diff
@@ -165,6 +153,14 @@ async function getCachedCredentialClient(): Promise<OAuth2Client> {
   }
 }
 
+async function setCachedCredentials(credentials: Credentials) {
+  const filePath = getCachedCredentialPath();
+  await fs.mkdir(path.dirname(filePath), { recursive: true });
+
+  const credString = JSON.stringify(credentials, null, 2);
+  await fs.writeFile(filePath, credString);
```
NTaylorMullen, line 161: Out of curiosity, how do other tech stacks in general cache creds on disk? Not sure we should be storing them plaintext

NTaylorMullen (commented):
Clarifying questions:
- Do we indicate to the user what account they're logged in with by chance?
- Is there a world where a user has both workspace credentials + personal credentials available? Assuming so, it sounds like we just pick the workspace credentials?

cornmander:
Does this honor the APPLICATION_DEFAULT_CREDENTIALS env var? (if you're using the normal google auth libraries it should).

GCA uses the ADC file format but doesn't write the file to the well-known location, so we'll need to override the path when launching the agent in the context of GCA extension.

packages/core/src/code_assist/oauth2.ts
scidomino: Caching the creds on disk is standard. Gcloud stores the application default credentials at: home_dir/.config/gcloud/application_default_credentials.json

scidomino:
> Does this honor the APPLICATION_DEFAULT_CREDENTIALS env var? (if you're using the normal google auth libraries it should).
> 
> GCA uses the ADC file format but doesn't write the file to the well-known location, so we'll need to override the path when launching the agent in the context of GCA extension.


The library we are using supports GOOGLE_APPLICATION_CREDENTIALS environment variable for pointing to the file.

scidomino:
> Clarifying questions:
> 
> * Do we indicate to the user what account they're logged in with by chance?
> * Is there a world where a user has both workspace credentials + personal credentials available? Assuming so, it sounds like we just pick the workspace credentials?

1. We don't. I don't think we have access to the username of the account. I could add a console.log() telling them which method we used to auth them.

2. Yes. This can happen but I think it would be pretty rare. Should we hide this behind a flag?

NTaylorMullen:
> We don't. I don't think we have access to the username of the account. I could add a console.log() telling them which method we used to auth them.

Alright good to know. We don't need to block on this but something we may want to consider cc @miguelsolorio 

> Yes. This can happen but I think it would be pretty rare. Should we hide this behind a flag?

Hmm, would this be rare? For instance, I have a gcloud project so I'm logged in BUT I want to use gemini from my personal account for things.

Either way doesn't need to block this PR but something we need to be intentional around

NTaylorMullen (approved):

README.md
scidomino: Updated the readme to remove references to Oauth.

NTaylorMullen (approved):

PR #1168, opened 2025-06-18
Set slice=true on tests to supress output.

scidomino:
## TLDR

Silences console.log() output during tests.

## Reviewer Test Plan

run `npm run test`

## Testing Matrix

N/A because it's test only.


NTaylorMullen (dismissed):
Feel free to also get this into `release` if you'd like

NTaylorMullen (approved):

PR #1176, opened 2025-06-18
fix: replace start.sh with start.js in create_alias script

jimmyliao:
## TLDR
1. the fixpack after #784 PR.

Before: > gemini
gemini-cli/scripts/start.sh: No such file or directory

After: > gemini
works with expected


packages/core/src/config/models.ts
```diff
@@ -4,6 +4,6 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-export const DEFAULT_GEMINI_MODEL = 'gemini-2.5-pro-preview-06-05';
-export const DEFAULT_GEMINI_FLASH_MODEL = 'gemini-2.5-flash-preview-05-20';
+export const DEFAULT_GEMINI_MODEL = 'gemini-2.5-pro';
+export const DEFAULT_GEMINI_FLASH_MODEL = 'gemini-2.5-flash';
```
NTaylorMullen: Would you mind undoing these two changes?

scripts/create_alias.sh
```diff
@@ -4,7 +4,7 @@
 
 # Determine the project directory
 PROJECT_DIR=$(cd "$(dirname "$0")/.." && pwd)
-ALIAS_COMMAND="alias gemini='$PROJECT_DIR/scripts/start.sh'"
+ALIAS_COMMAND="alias gemini='node $PROJECT_DIR/scripts/start.js'"
```
NTaylorMullen, line 7: nice catch!

NTaylorMullen (approved):

packages/core/src/config/models.ts
jimmyliao: reverted

scripts/create_alias.sh
jimmyliao: Hi @NTaylorMullen, please let me know if anything need from me for this ticket. Thanks.

NTaylorMullen:
@jimmyliao so sorry for the delay. I'm so used to folks merging PRs on their own BUT just realized you don't have write-access. Getting this in now!

PR #1188, opened 2025-06-18
Move the otel folder out of project root .gemini and into user home dir

louisjimenez:
This changes the telemetry script so that it writes the `/otel` folder to the user's home directory rather than the project root `.gemini` where it would require the user to modify the `.gitignore`. The `~/.gemini/tmp/<hashed-project>` is being used as a general purpose storage place for application data.

### DIFF
--- a/scripts/telemetry_utils.js
+++ b/scripts/telemetry_utils.js
@@ -12,15 +12,31 @@ import net from 'net';
 import os from 'os';
 import { execSync } from 'child_process';
 import { fileURLToPath } from 'url';
+import crypto from 'node:crypto';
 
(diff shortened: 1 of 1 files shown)


louisjimenez:
@jerop I ran this locally and verified it writes the otel folder to the correct place and generates logs. As the implementor of telemetry, can you take a look at this change? Thanks

mattKorwel (approved):

jerop (commented):
@louisjimenez do you mind following up with a docs update? https://github.com/google-gemini/gemini-cli/blob/main/docs/core/telemetry.md#local

PR #1190, opened 2025-06-18
fix: flicker of StreamingState to Idle when tool finishes

asadm:
## TLDR

The flicker of  StreamingState to Idle happens because of the time between when a tool finishes executing and when the model starts responding, the state is Idle. 

## Dive Deeper

Here's the sequence of events:

1. A tool is executing, so toolCalls has a tool with status executing. This makes streamingState become Responding.
2. The tool finishes. The useReactToolScheduler's onComplete callback is fired.
3. Inside onComplete, the tool's result is added to the history.
4. The useEffect that watches for completed tool calls is triggered.
5. This useEffect prepares the tool's response and calls submitQuery to send it to the model.
6. submitQuery sets isResponding to true.

The brief moment of Idle occurs between step 3 and step 6. After the tool completes, the toolCalls array is updated, and for a brief moment, neither isResponding is true nor are there any active tool calls. This causes the streamingState to become Idle until submitQuery is called and sets isResponding to true, which then transitions the state back to Responding.

This was surfaced when working on #1191 

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


sethtroisi (approved):
Nice find!

PR #1191, opened 2025-06-18
feat: promptReadyCommand - command to execute when the model has responded and is ready for user input

asadm:
## TLDR
Usually, I switch to a different window when gemini is doing the task. I would like to be notified when gemini is done and is waiting for my input. This PR does 2 things
1. By default, it will ring a "bell" sound using BEL control character
2. It let's you override the behavior by setting a custom shell command to run when done, the command also gets last response via stdin. This is interesting because I have it hooked to macOS's `say` that reads ~10 words from response for me. Setting this command disables the default BEL sound.

Example settings.json (reads 10 words from last response by model when it's done with a task or is waiting for a confirmation)

```
{
    "promptReadyCommand": "bash -c \"awk -v N=10 'BEGIN{ORS=\\\" \\\"} {for(i=1; i<=NF; i++) {if(count<N){print \\$i; count++} else {exit}}}' | say\""
}
```

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


NTaylorMullen:
This is definitely a very interesting direction and commit for a feature like this. I think my gut leans towards this may be too specific BUT the ending of a prompt cycle undoubtedly is a common thing. I also noted all of the feedback on your chat message in the users group too.

Out of curiosity, do you think this would work as an MCP server with a "Done" tool that tries to prompt the model to invoke it right before it returns feedback to the user?

asadm:
That might work but can get flaky (as with other tool calling). Also might be tricky to do for "waiting for confirmation" part. I can look into it. 

I would say there are 2 things in this PR
1. Alert for user when it's their turn 
2. Custom command part

Do you think it makes sense to start with only the alerting part for now? it's nice to be notified for any interruptions model may have so user doesn't face a bummer when coming back many minutes later expecting the task must have finished but see a tool confirmation or a question.

NTaylorMullen:
I think doing just the alerting is a good option. We know we need to have that baked into Gemini CLI at the very least

Kailian-Jacy:
Also wanting this feature. 

But using MCP might be a bad choice.. It's managing the llm workflow, rather than a functionality that should be provided by llm itself. It's better to be implemented programmatically. MCP is too unreliable.

Maybe some kind of hook mechanism could be provided, allowing users to customize logic when events happens, more than just "bell ring" on "prompt ready". It's also paving the way for integration into other systems like editors.

scottdensmore:
@asadm Appreciate the PR. We are coming back to quite a few of these. Do you mind filing an issue for this and we can discuss on how to best implement this.

PR #1195, opened 2025-06-18
Move the shell history our of the project .gemini to the home dir

louisjimenez:
This moves the shell_history file out of the project root .gemini folder to the tmp directory that exists under the home directory. It moves the history to a project specific folder which adds support for project scoped command histories.

### DIFF
--- a/packages/cli/src/ui/hooks/useShellHistory.ts
+++ b/packages/cli/src/ui/hooks/useShellHistory.ts
@@ -7,14 +7,13 @@
 import { useState, useEffect, useCallback } from 'react';
 import * as fs from 'fs/promises';
 import * as path from 'path';
-import { isNodeError } from '@gemini-cli/core';
+import { isNodeError, getProjectTempDir } from '@gemini-cli/core';
 
-const HISTORY_DIR = '.gemini';
 const HISTORY_FILE = 'shell_history';
 const MAX_HISTORY_LENGTH = 100;
 
 async function getHistoryFilePath(projectRoot: string): Promise<string> {
-  const historyDir = path.join(projectRoot, HISTORY_DIR);
+  const historyDir = getProjectTempDir(projectRoot);
   return path.join(historyDir, HISTORY_FILE);
 }
 
--- a/packages/core/src/utils/paths.ts
+++ b/packages/core/src/utils/paths.ts
@@ -6,6 +6,10 @@
 
 import path from 'node:path';
 import os from 'os';
+import * as crypto from 'crypto';
+
+export const GEMINI_DIR = '.gemini';
+const TMP_DIR_NAME = 'tmp';
 
 /**
  * Replaces the home directory with a tilde.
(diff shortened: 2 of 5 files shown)


packages/cli/src/ui/hooks/useShellHistory.ts
```diff
@@ -7,14 +7,26 @@
...
+import * as crypto from 'crypto';
 import { isNodeError } from '@gemini-cli/core';
 
-const HISTORY_DIR = '.gemini';
+const GEMINI_DIR = '.gemini';
+const TMP_DIR_NAME = 'tmp';
 const HISTORY_FILE = 'shell_history';
 const MAX_HISTORY_LENGTH = 100;
 
 async function getHistoryFilePath(projectRoot: string): Promise<string> {
```
scottdensmore, line 15: Can we consolidate this path code because it is now shared by userShellHistory.ts, logger.ts and getService.ts. That way we always have the project hash correct (in case we add more)

scottdensmore (dismissed):

packages/cli/src/ui/hooks/useShellHistory.ts
louisjimenez: Good point, I updated it.

packages/core/src/utils/paths.ts
```diff
@@ -134,3 +138,22 @@ export function escapePath(filePath: string): string {
...
+export function getProjectHash(projectRoot: string): string {
+  return crypto.createHash('sha256').update(projectRoot).digest('hex');
+}
+
+/**
+ * Generates a unique temporary directory path for a project.
+ * @param projectRoot The absolute path to the project's root directory.
+ * @returns The path to the project's temporary directory.
+ */
+export function getProjectTempDir(projectRoot: string): string {
```
dewitt, line 156: Calling this `getProjectTempDir` feels like a misnomer, insofar as the directory name is stable, no?  Every time we run gemini under the same project root, the history should be the same, correct?

  homedir (stable) + '.gemini' (stable) + 'tmp' (stable) + 'hash(project root)' (stable)
  
If so, could we call it getProjectHistoryDir instead?

packages/core/src/utils/paths.ts
```diff
@@ -6,6 +6,10 @@
 
 import path from 'node:path';
 import os from 'os';
+import * as crypto from 'crypto';
+
+export const GEMINI_DIR = '.gemini';
+const TMP_DIR_NAME = 'tmp';
```
dewitt, line 12: If my comment below about this directory being stable is accurate, is this really tmp? If we don't clean it up periodically or expect the user to, then maybe rename to 'history' or something.

louisjimenez: I was favoring tmp since while we don't clean it up right now we are working under the assumption that the user might want to (so we can't assume it is always there) and down the line we might do so.

packages/core/src/utils/paths.ts
louisjimenez: Keep as temp to keep in line with the other comment and the directory name "tmp"

packages/core/src/utils/paths.ts
```diff
@@ -134,3 +138,22 @@ export function escapePath(filePath: string): string {
...
   return filePath.replace(/\\ /g, ' ');
 }
+
+/**
+ * Generates a unique hash for a project based on its root path.
+ * @param projectRoot The absolute path to the project's root directory.
+ * @returns A SHA256 hash of the project root path.
+ */
+export function getProjectHash(projectRoot: string): string {
+  return crypto.createHash('sha256').update(projectRoot).digest('hex');
```
dewitt, line 148: Just a non-blocking note that we presumably don't need the full 64 byte hex digest, and could safely truncate this to the first 8 ascii bytes for aesthetic purposes when the user sees it in their directory. It just needs not to collide, not be cryptographically secure.

dewitt (approved):
LGTM

packages/core/src/utils/paths.ts
louisjimenez: Thanks, I created this https://github.com/google-gemini/gemini-cli/issues/1249

mattKorwel (approved):
Lgtm. 
So if I read this right, every directory that I open up in the CLI on the same machine will keep its history separated in a hash directory. So when I go back to that one I have my own unique history? That's pretty cool. Good stuff.

PR #1202, opened 2025-06-18
Use Env Var directly instead of through GoogleAuth()

scidomino:
## TLDR

To let the vscode plugin provide its own credentials we want to support the GOOGLE_APPLICATION_CREDENTIALS env var. GoogleAuth() does that but it also picks up the application default credentials which is not what we want. So just support that env var directly.

## Reviewer Test Plan

After logging in with Code Assist normally exit gemini cli. Rename your credential file at `/Users/your_home_dir/.gemini/oauth_creds.json` to `oauth_creds_alt.json`. Then set your env var to point to the renamed file:

`export GOOGLE_APPLICATION_CREDENTIALS=/Users/your_home_dir/.gemini/oauth_creds_alt.json`

Then start up Gemini CLI and verify that it does not ask you to log in again since it finds your creds at the alternative location.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | X  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

b/425360849

### DIFF
--- a/packages/core/src/code_assist/codeAssist.ts
+++ b/packages/core/src/code_assist/codeAssist.ts
@@ -4,7 +4,6 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { GoogleAuth, AuthClient } from 'google-auth-library';
 import { ContentGenerator } from '../core/contentGenerator.js';
 import { getOauthClient } from './oauth2.js';
 import { setupUser } from './setup.js';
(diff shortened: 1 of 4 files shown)


scottdensmore (approved):
Approved

PR #1203, opened 2025-06-19
Support Oauth2 Reauthentication

scidomino:
## TLDR

If a oauth token expires while generating content, prompt the user to log in again. 

## Dive Deeper

This modifies the CodeAssistServer to throw ReauthNeededError when the user's credentials expire. It adds a new ReauthDialog to display when authentication is happening:

https://screenshot.googleplex.com/APpSYynBQVsUFzc
https://screenshot.googleplex.com/7gcF24Tj7ai4aYJ

Note: This catches Reauth errors when the agent is streaming content but there are a few cases (like /compress) where we call the CodeAssistServer via alternative means. I will address those in a follow up CL as this one is already quite large.

## Reviewer Test Plan

1. Start up the CLI and log in with code assist normally.
2. Revoke your credentials by:
   1.  navigating to [Your connections to third-party apps & services](https://myaccount.google.com/connections?continue=https%3A%2F%2Fmyaccount.google.com%2Fsecurity)
   2. Select "Gemini Code Assist" 
   3. click "Delete all connections you have with Gemini Code Assist"
3. **Without exiting the CLI**, Wait five minutes for the revocation to propagate.
4. Ask the cli a question.
5. Verify that the reauth popup appears. 
6. Press return to open the browser.
7. Authenticate. 
8. Verify that you can now talk to the cli without issue.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | X  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

b/425360849

### DIFF
--- a/packages/cli/src/ui/components/ReauthDialog.tsx
+++ b/packages/cli/src/ui/components/ReauthDialog.tsx
@@ -4,10 +4,43 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { Box, Text } from 'ink';
+import { Box, Text, useInput } from 'ink';
+import { useState, useCallback } from 'react';
+import { Config, refreshAuth } from '@gemini-cli/core';
+import open from 'open';
 import { Colors } from '../colors.js';
 
-export const ReauthDialog = () => {
+interface ReauthDialogProps {
+  config: Config;
+  onAuthSuccess: () => void;
+}
+
+export const ReauthDialog = ({ config, onAuthSuccess }: ReauthDialogProps) => {
+  const [authUrl, setAuthUrl] = useState<string | null>(null);
+  const [isAuthenticating, setIsAuthenticating] = useState(false);
+  const [error, setError] = useState<string | null>(null);
+
+  const handleConfirm = useCallback(async () => {
+    if (isAuthenticating) return;
+
+    try {
+      const authProcess = await refreshAuth(config.getGeminiClient().getContentGenerator());
+      setAuthUrl(authProcess.authUrl);
+      setIsAuthenticating(true);
+      await open(authProcess.authUrl);
+      await authProcess.loginCompletePromise;
+      onAuthSuccess();
+    } catch (e) {
+      setError(e instanceof Error ? e.message : 'An unknown error occurred.');
+    }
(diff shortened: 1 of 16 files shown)


packages/core/package.json
```diff
@@ -33,6 +33,7 @@
     "@types/html-to-text": "^9.0.4",
     "diff": "^7.0.0",
     "dotenv": "^16.4.7",
+    "gaxios": "^6.1.1",
```
NTaylorMullen, line 36: Do we need gaxios over native fetch support?

packages/cli/src/ui/components/ReauthDialog.tsx
NTaylorMullen, line 1: Going to skip reviewing this bit because the dialog will need to be integrated into the primary auth flow piece that @mattKorwel is building

packages/core/src/code_assist/oauth2.ts
```diff
@@ -114,7 +119,6 @@ async function authWithWeb(client: OAuth2Client): Promise<OauthWebLogin> {
             redirect_uri: redirectUri,
           });
           client.setCredentials(tokens);
-          await cacheCredentials(client.credentials);
```
NTaylorMullen, line 117: Why this removal? Is the tokens handler called when we do setCredentials above?

NTaylorMullen (approved):

packages/core/src/code_assist/oauth2.ts
scidomino: Yes. Exactly. The Oauth client will refresh the auth token on its own and we want to pick up and persist that too, not just when we first do web auth here.

packages/core/package.json
scidomino: Gaxios is already a dependency of google-auth-library. I just imported this so we can catch the GaxiosError that it may throw.

mattKorwel:
Superceded by https://github.com/google-gemini/gemini-cli/pull/1207

PR #1207, opened 2025-06-19
Auth First Run

mattKorwel:
## TLDR

Adds an auth landing screen.

## Dive Deeper

This adds an auth screen and reworks where we load the content generators. we now don't try and connect / auth until the ui/ux loads. We now set an `authType` setting from the ui. If this is set we use it and go through the auth flow.

There was a lot of refactoring here as the current interweaving of things is tightly coupled.

## Reviewer Test Plan

There are a bunch of things.

# Testing changing out

1. start unauthed and see that the diaglog pops up.
2. select 1. log into google with personal and see that the o-auth flow pops up. test calling things. Leave the app. restart, you should see no auth dialog but still be connected and able to do things
3. use `/auth` to load the dialog again. try "geminiAPI". test and repeat
4. try for vertex_ai
5. try for enterprise oauth.

# Testing starting up with invalid auth

1. Run once and set auth to say, gemini_api_key
2. go close
3. go comment out your env for gemini_api_key
4. start up and see you are prompted

# Testing headless auth validation

1. Comment out all env variables and try and run headless. see that you get similar error mesasages to the main auth screen
5. test with just gemini_api_key env var set. this is a special case for things like remote boxes

# Testing sandbox auth validation

1. sandbox solution is that it will be treated as a headless startup and you shouuld see same erorrs

# Testing Re-auth

1. Start up the CLI and log in with code assist normally.
2. Revoke your credentials by:
3. navigating to [Your connections to third-party apps & services](https://myaccount.google.com/connections?continue=https%3A%2F%2Fmyaccount.google.com%2Fsecurity)
4. Select "Gemini Code Assist"
5. click "Delete all connections you have with Gemini Code Assist"
6. Without exiting the CLI, Wait five minutes for the revocation to propagate.
7. Ask the cli a question.
8. Verify that the reauth popup appears.
9. Press return to open the browser.
10. Authenticate.
11. Verify that you can now talk to the cli without issue.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅   | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

no idea.


mattKorwel:
https://github.com/user-attachments/assets/d7004b9f-72cf-4e4b-b890-441fbd866048

Screenshot 2025-06-18 at 10 41 04 PM

mattKorwel:
Also confirmed working for headless 

```npm run start -- --prompt "edit test.txt to have a hello world message"

> gemini-cli@0.1.0 start
> node scripts/start.js --prompt edit test.txt to have a hello world message

Checking build status...
Build is up-to-date.

Loaded 18 patterns from /Users/mattkorwel/dev/gemini-cli/.gitignore
Loaded 18 patterns from /Users/mattkorwel/dev/gemini-cli/.gitignore
Please set an Auth method in your .gemini/settings.json file befor running
> 
> npm run start -- --prompt "edit test.txt to have a hello world message"

> gemini-cli@0.1.0 start
> node scripts/start.js --prompt edit test.txt to have a hello world message

Checking build status...
Warning: Source file "packages/cli/src/gemini.tsx" has been modified since the last build.

Run "npm run build" to incorporate changes before starting.
Loaded 18 patterns from /Users/mattkorwel/dev/gemini-cli/.gitignore
Loaded 18 patterns from /Users/mattkorwel/dev/gemini-cli/.gitignore
GEMINI_API_KEY  environment variable not found. Add that to your .env and try again, no reload needed!
```

mattKorwel:
headless testing - 

works with just gemini_api_key set.
work with oauth personal set
gives errors when auth not right

mattKorwel:
```
> npm run start -- -p "hello"

> gemini-cli@0.1.0 start
> node scripts/start.js -p hello
//// with gemini_api_key set
Checking build status...
Build is up-to-date.
Loaded 18 patterns from /Users/mattkorwel/dev/gemini-cli/.gitignore
Loaded 18 patterns from /Users/mattkorwel/dev/gemini-cli/.gitignore
Which stock are you interested in?
> npm run start -- -p "hello"

> gemini-cli@0.1.0 start
> node scripts/start.js -p hello

// without auth set
Checking build status...
Build is up-to-date.
Loaded 18 patterns from /Users/mattkorwel/dev/gemini-cli/.gitignore
Loaded 18 patterns from /Users/mattkorwel/dev/gemini-cli/.gitignore
Please set an Auth method in your .gemini/settings.json OR specify GEMINI_API_KEY env variable file before running
> npm run start -- -p "hello"

> gemini-cli@0.1.0 start
> node scripts/start.js -p hello
// set to personal
Checking build status...
Build is up-to-date.
Loaded 18 patterns from /Users/mattkorwel/dev/gemini-cli/.gitignore
Loaded 18 patterns from /Users/mattkorwel/dev/gemini-cli/.gitignore


Code Assist login required.
Attempting to open authentication page in your browser.
Otherwise navigate to:

https://accounts.google.com/o/oauth2/v2/auth?redirect_uri=http%3A%2F%2Flocalhost%3A54652%2Foauth2callback&access_type=offline&scope=https%3A%2F%2Fwww.googleapis.com%2Fauth%2Fcloud-platform%20https%3A%2F%2Fwww.googleapis.com%2Fauth%2Fuserinfo.email%20https%3A%2F%2Fwww.googleapis.com%2Fauth%2Fuserinfo.profile&state=428421891f71fe81876260c420201795c9202c64cd369de62dee125f1b4abb03&response_type=code&client_id=681255809395-oo8ft2oprdrnp9e3aqf6av3hmdib135j.apps.googleusercontent.com

// set to entrprise with no project set
Waiting for authentication...
Of which company?
> npm run start -- -p "hello"

> gemini-cli@0.1.0 start
> node scripts/start.js -p hello

Checking build status...
Build is up-to-date.
Loaded 18 patterns from /Users/mattkorwel/dev/gemini-cli/.gitignore
Loaded 18 patterns from /Users/mattkorwel/dev/gemini-cli/.gitignore
GOOGLE_CLOUD_PROJECT environment variable not found. Add that to your .env and try again, no reload needed!
~/dev/gemini-cli on auth-first-run *3 +1 !5 ---------------------------------------------------------- node 20.19.2 pkg 0.1.0 at 11:30:45
> 
```

scidomino:
I found a bug: I clear all my settings, run `npx https://github.com/google-gemini/gemini-cli#auth-first-run` and then choose "Login with personal account" but if I use my @google.com it just silently exits. When I try to start it up again it just quietly exits again. Previously it would print out:

Error onboarding with Code Assist.
Google Workspace Account (e.g. your-name@your-company.com)
must specify a GOOGLE_CLOUD_PROJECT environment variable.

It probably still tries to print that but can't because react and ink get in the way.

packages/cli/index.ts
```diff
@@ -13,7 +13,7 @@ import { main } from './src/gemini.js';
 main().catch((error) => {
   console.error('An unexpected critical error occurred:');
   if (error instanceof Error) {
-    console.error(error.message);
+    console.error(error.stack);
```
NTaylorMullen, line 16: We'll probably want to include the message as well. FWIW there's a getErrorMessage(..) helper in the repo

packages/cli/src/config/auth.ts
```diff
@@ -0,0 +1,44 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { AuthType } from '@gemini-cli/core';
+import { loadEnvironment } from './config.js';
+
+export const validateAuthMethod = (authMethod: string): string | null => {
```
NTaylorMullen, line 10: non-blocking: cc @miguelsolorio for verbiage / UXisms here for errors

packages/cli/src/config/auth.ts
```diff
@@ -0,0 +1,44 @@
...
+  loadEnvironment();
+  if (authMethod === AuthType.LOGIN_WITH_GOOGLE_PERSONAL) {
+    return null;
+  }
+
+  if (authMethod === AuthType.LOGIN_WITH_GOOGLE_ENTERPRISE) {
+    if (!process.env.GOOGLE_CLOUD_PROJECT) {
+      return 'GOOGLE_CLOUD_PROJECT environment variable not found. Add that to your .env and try again, no reload needed!';
```
NTaylorMullen, line 18: - I'm not sure we need the `no reload needed` bit across all of these, if people try and reload that's ok and if they just try again that's also ok. 
- You may be able to keep this super simple as in: `GOOGLE_CLOUD_PROJECT environment variable not found.` We could mention `.env` but that's kind of an expectation in this world though too

packages/cli/src/gemini.tsx
```diff
@@ -73,6 +71,14 @@ export async function main() {
   if (!process.env.SANDBOX) {
     const sandboxConfig = config.getSandbox();
     if (sandboxConfig) {
+      if (settings.merged.selectedAuthType) {
+        const err = validateAuthMethod(settings.merged.selectedAuthType);
```
scidomino, line 76: nit: Add a comment explaining that we do this before sandboxing so avoid oauth redirect troubles inside the sandbox.

packages/cli/src/gemini.tsx
```diff
@@ -152,28 +158,58 @@ async function loadNonInteractiveConfig(
...
-  const interactiveTools = [ShellTool.Name, EditTool.Name, WriteFileTool.Name];
+  return await validateNonInterActiveAuth(
+    settings.merged.selectedAuthType,
+    finalConfig,
+  );
+}
+
+async function validateNonInterActiveAuth(
+  selectedAuthType: AuthType | undefined,
+  nonInteractiveConfig: Config,
```
scidomino, line 195: nit: I don't see why we need to pass the config into this. You can just call refreshAuth on it after this method is called.

packages/core/src/code_assist/codeAssist.ts
```diff
@@ -4,15 +4,23 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { ContentGenerator } from '../core/contentGenerator.js';
+import { AuthType, ContentGenerator } from '../core/contentGenerator.js';
 import { getOauthClient } from './oauth2.js';
 import { setupUser } from './setup.js';
 import { CodeAssistServer, HttpOptions } from './server.js';
 
 export async function createCodeAssistContentGenerator(
```
scidomino, line 12: The changes to this method are all unnecessary since we already check to make sure the authType is correct before calling this.

packages/core/src/code_assist/server.ts
```diff
@@ -4,7 +4,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { OAuth2Client } from 'google-auth-library';
+import { AuthClient } from 'google-auth-library';
```
scidomino, line 7: I believe the changes to this file are unnecessary. It's fine to use Oauth2Client instead of AuthClient.

packages/cli/src/config/settings.ts
```diff
@@ -32,6 +33,7 @@ export interface AccessibilitySettings {
 
 export interface Settings {
   theme?: string;
+  selectedAuthType?: AuthType;
```
NTaylorMullen, line 36: Should this be nullable?

packages/core/src/config/config.test.ts
```diff
@@ -80,13 +90,40 @@ describe('Server Config (config.ts)', () => {
...
     sessionId: SESSION_ID,
+    model: MODEL,
   };
 
   beforeEach(() => {
     // Reset mocks if necessary
     vi.clearAllMocks();
   });
 
+  // i can't get vi mocking to import in core. only in cli. can't fix it now.
```
scidomino, line 101: nit: if it's worth commenting out, it's worth deleting.

scidomino (approved):

packages/cli/src/gemini.tsx
```diff
@@ -152,28 +158,58 @@ async function loadNonInteractiveConfig(
...
+) {
+  // making a special case for the cli. many headless environments might not have a settings.json set
+  // so if GEMINI_API_KEY is set, we'll use that. However since the oauth things are interactive anyway, we'll
+  // still expect that exists
+  if (!selectedAuthType && !process.env.GEMINI_API_KEY) {
+    console.error(
+      'Please set an Auth method in your .gemini/settings.json OR specify GEMINI_API_KEY env variable file before running',
```
NTaylorMullen, line 202: use the constant placeholders for the folders

packages/cli/src/gemini.tsx
scidomino:
```suggestion
        // Validate authentication here because the sandbox will interfere with the Oauth2 web redirect.
        const err = validateAuthMethod(settings.merged.selectedAuthType);
```

packages/cli/src/ui/components/AuthDialog.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+interface AuthDialogProps {
+  onSelect: (authMethod: string | undefined, scope: SettingScope) => void;
+  onHighlight: (authMethod: string | undefined) => void;
+  settings: LoadedSettings;
+  initialErrorMessage?: string | null;
+}
+
+export function AuthDialog({
+  onSelect,
```
NTaylorMullen, line 23: non-blocking: You may be able to isolate the onSelect/onHighlight actions within this component. Not sure what the best practice is for React but I could imagine you being able to do `useAuthCommand` here.

packages/cli/src/ui/components/AuthDialog.tsx
```diff
@@ -0,0 +1,94 @@
...
+  onHighlight,
+  settings,
+  initialErrorMessage,
+}: AuthDialogProps): React.JSX.Element {
+  const [errorMessage, setErrorMessage] = useState<string | null>(
+    initialErrorMessage || null,
+  );
+  const authItems = [
+    {
+      label: 'Login w/Google: Personal Account',
```
NTaylorMullen: The `w/` saves space but I feel also makes it feel less clean cc @miguelsolorio 

```suggestion
      label: 'Login with Google Personal Account',
```

packages/cli/src/ui/components/AuthDialog.tsx
```diff
@@ -0,0 +1,94 @@
...
+      value: AuthType.LOGIN_WITH_GOOGLE_PERSONAL,
+    },
+    {
+      label: 'Login w/Google: Work Account and a GCP project',
```
NTaylorMullen: Ditto but the emphasis here is I'm putting on the Google project because people can quickly visually invalidate the option

```suggestion
      label: 'Login with GCP Project and Google Work Account',
```

scidomino (commented):
Also, you have to update the README and other docs to remove references to GEMINI_CODE_ASSIST since I believe this removes all references to them.

packages/cli/src/ui/components/AuthDialog.tsx
```diff
@@ -0,0 +1,94 @@
...
+      value: AuthType.LOGIN_WITH_GOOGLE_ENTERPRISE,
+    },
+    { label: 'Gemini API Key', value: AuthType.USE_GEMINI },
```
NTaylorMullen: Lets move this up below the personal account

```suggestion
    { label: 'Gemini API Key', value: AuthType.USE_GEMINI },
```

packages/cli/src/ui/components/AuthDialog.tsx
```diff
@@ -0,0 +1,94 @@
...
+    { label: 'Vertex AI', value: AuthType.USE_VERTEX_AI },
```
NTaylorMullen, line 41: Ugh you may hate me but now that I'm really digesting and playing with this I'm worried we're showing too many options for login. We should chat about this offline but do you think we'd be able to create a `More...` option that then reveals Google work / Vertex AI account?

packages/core/src/config/config.ts
```diff
@@ -106,12 +109,13 @@ export interface ConfigParameters {
...
   fileDiscoveryService?: FileDiscoveryService;
   bugCommand?: BugCommandSettings;
+  model: string;
 }
 
 export class Config {
   private toolRegistry: Promise<ToolRegistry>;
   private readonly sessionId: string;
-  private readonly contentGeneratorConfig: ContentGeneratorConfig;
+  private contentGeneratorConfig!: ContentGeneratorConfig;
```
NTaylorMullen, line 118:
```suggestion
  private readonly contentGeneratorConfig: ContentGeneratorConfig;
```

packages/core/src/config/config.ts
```diff
@@ -130,18 +134,18 @@ export class Config {
...
-  private readonly geminiClient: GeminiClient;
+  private geminiClient!: GeminiClient;
   private readonly fileFilteringRespectGitIgnore: boolean;
   private fileDiscoveryService: FileDiscoveryService | null = null;
   private gitService: GitService | undefined = undefined;
   private readonly checkpoint: boolean;
   private readonly proxy: string | undefined;
   private readonly cwd: string;
   private readonly bugCommand: BugCommandSettings | undefined;
+  private readonly model: string;
```
NTaylorMullen, line 145: Why have the separate model?

NTaylorMullen (approved):
We can get this in as-is; however, the biggest comment I have is around surfacing all 4 options actually and the wording we're using there. We should start a thread offline

packages/cli/index.ts
mattKorwel: Interesting enough stack does include the message. at least on my box. having both doubled the message.

packages/cli/src/config/auth.ts
mattKorwel: To be honest, i didn't know how else to document the idea that they don't have to exit ...

packages/core/src/code_assist/codeAssist.ts
mattKorwel: They are unneccssary in a happy path. but future us might call this method somewhere else and rahther than silently failing, at least we'll have an indication that something went wrong. defense in depth. are you ok leaving them in?

packages/core/src/code_assist/server.ts
mattKorwel: i think when this pr merges one of our todo's is pull in your other oauth2 changes from the refresh auth branch just so we're complete. ok leaving these as is until then?

packages/core/src/config/config.test.ts
mattKorwel: agree. cant figure out the suggestion inline. will follow up.

packages/cli/src/config/settings.ts
mattKorwel: since its an interface the ? says the field itself is optional on the interface and if its not there the value will be undefined.  in this case i don't think there is a functional difference between `selectedAuthType?: AuthType` and `selectedAuthType: AuthType | undefined` with the exception that the first option plays nicer with signatures everywhere and requires less refactoring.

I could be wrong though, i am not a ts expert nor do i play one on tv.

packages/core/src/config/config.ts
mattKorwel: we need the model to recreate the generatorConfig when auth changes via the `refreshAuth` method https://github.com/google-gemini/gemini-cli/blob/62dbe3ee0375ffd94f2e129855780180c00a6c73/packages/core/src/config/config.ts#L194

On startup if auth isn't set we'll not have an existing contentGeneratorConfig to pull from. So we need to then pass the model around in some storage so we have it avialable as needed to make the refresh auth call. given we already have a data store in config, seemed better to me to store it there than refactor a bunch more calling sites.

packages/core/src/config/config.ts
mattKorwel: acutally this can't be readonly since it will change each time the auth changes its reset in `refreshAuth`

packages/cli/src/config/settings.ts
NTaylorMullen: Oh sorry, my question is why not require it? As in 
```suggestion
  selectedAuthType: AuthType;
```

Or do we rely on the undefined behavior to indicate "first run"?

packages/cli/src/gemini.tsx
mattKorwel: i only want to call refrsh auth after we know we're valid. putting it outside this fucnction would possibly separate this protection from the call. It needs to stay in here.

packages/cli/src/config/settings.ts
mattKorwel: Oh, i'm pretty sure i didn'tw ant to have to fix all the places it needed reafactored :)

scottdensmore:
We need to ensure that the terminal blocks while you are signing in and you can't go back and type and crash Gemini

scottdensmore:
For Vertex AI, we should have more explanation of the variables to set to get to Vertex vs Vertex Express.

packages/core/src/code_assist/codeAssist.ts
scidomino: My gut says that this kind of defensive programming might be more confusion than it's worth but it's a judgement call so I'm happy to go with whatever you decide.

packages/core/src/code_assist/server.ts
scidomino: ok

scidomino (approved):

PR #1211, opened 2025-06-19
Make checkpoints configurable in settings.json

louisjimenez:
This allows the checkpoint feature to be enabled in the settings.json. Example
```
{
  "theme": "Dracula",
  "checkpoint": { "enabled": true }
}
```
If omitted and the checkpoint flag is not set to true, the feature will continue to default to off.

### DIFF
--- a/packages/cli/src/config/config.ts
+++ b/packages/cli/src/config/config.ts
@@ -49,7 +49,7 @@ interface CliArgs {
   show_memory_usage: boolean | undefined;
   yolo: boolean | undefined;
   telemetry: boolean | undefined;
-  checkpoint: boolean | undefined;
+  checkpointing: boolean | undefined;
   telemetryTarget: string | undefined;
   telemetryOtlpEndpoint: string | undefined;
   telemetryLogPrompts: boolean | undefined;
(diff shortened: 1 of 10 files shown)


packages/cli/src/config/config.ts
```diff
@@ -233,7 +233,7 @@ export async function loadCliConfig(
     },
     // Git-aware file filtering settings
     fileFilteringRespectGitIgnore: settings.fileFiltering?.respectGitIgnore,
-    checkpoint: argv.checkpoint,
+    checkpoint: argv.checkpoint || settings.checkpoint?.enabled,
```
NTaylorMullen: Should it be `checkpoint` or `checkpointing`?

louisjimenez: Updated to checkpointing

NTaylorMullen (approved):

scottdensmore:
@louisjimenez how do you see the automatic saved checkpoints? Is there a way to list all the checkpoints?

louisjimenez:
> @louisjimenez how do you see the automatic saved checkpoints? Is there a way to list all the checkpoints?

To see all checkpoints run the /restore command with no arguments. I am closing this PR and opening up another that I verified works and will be against the release branch

PR #1213, opened 2025-06-19
Remove verification to prevent file reverts and endless loops

anj-s:
Specifying verification checks in the system prompt leads the LLM to revert any changes it deems as not passing those verification checks. We need to direct the agent to run tests and other lint checks but the model seems to be overly sensitive to these two specs.

I am seeing more than one instance of loops without this change. I also think we should modify the directives related to LLM's summarization and explanation of actions to better understand what it is doing.


keithballinger:
Have you considered alternative directions here to instruct the model to not engage in the faulty behavior, but still run lints and tests?

NTaylorMullen (approved):

NTaylorMullen:
> Have you considered alternative directions here to instruct the model to not engage in the faulty behavior, but still run lints and tests?

@keithballinger see the ask mode channel, we had a little convo about that there

PR #1217, opened 2025-06-19
Fix flicker issues by ensuring all actively changing content fits in the viewport

jacob314:
Flow max heights and widths everywhere in the world of messages and add a MaxSizedBlock renderer for robust wrapping.

MaxSizedBlock is robust as it does all its own word wrapping rather than relying on Ink's functionality which has a lot of odd edge cases. 
MaxSizedBlock only accepts a restricted subset of layouts to make it practical to implement the layouts statically at the Element layer rather than requiring integration with the rendering system.
All history item display now wrap and truncate content perhaps to an excessive degree. This ensures that the current group never overflows the viewport avoiding bad flickering.
Even the theme dialog truncates and wraps as needed. Not much value in that but it served as a reasonable testing ground to validate the logic.

Naming discussion:
MaxSizedBox is probably a bad name. Please suggest your better name.

Optional item: add a `ctrl-S` keyboard shortcut to toggled between truncating messages and showing the entire message while risking enormous messages and flicker. I have this hooked up to the debug console but not yet to the terminal messages. It could provide a good affordance to toggle back and forth between the view that fits in the window and the view that requires scrolling.

## Reviewer Test Plan

Test 

Try your favorite inputs and flows and make sure that the current message group is reasonably sized and never exceeds the size of the viewport. Verify that word wrapping looks ok. For example, try multi char wide emojis or other languages.
Test cases with large terminal output, markdown, diffs.

Prompts:
`Add the comment // hello world before each method in App.tsx. Apply
  as a single operation
`

` write 3 text files a.txt, b.txt, and c.txt. Each 50 rows from 1 to 50. Each
  line should have the phrase "The quick brown fox jumps over the lazy dog.
  <line number> Some more text.". Use WriteFile`

This generates a diff that wraps verifying that the line wrapping was robust.

Test in both auto-run and manual accept mode.
Test terminal commands that output a lot of output. Everything should wrap well.
Test wide and narrow terminals.

Changes are UI only. I recommend iterm2 to test as it has the worst flicker issues.
## Linked issues / bugs
b/422571619

Screenshots:
Screenshot 2025-06-19 at 2 47 08 AM


packages/cli/src/ui/App.tsx
```diff
@@ -479,7 +487,7 @@ const App = ({ config, settings, startupWarnings = [] }: AppProps) => {
             </Box>,
             ...history.map((h) => (
               <HistoryItemDisplay
-                availableTerminalHeight={availableTerminalHeight}
+                terminalWidth={mainAreaWidth}
```
jacob314, line 493: For discussion: I can add back availableTerminalHeight for the items rendered to history in the static area. I figured users probably want to see the full version not the version we truncated due to height constraints but I am fine either way.
This is a free change as we already need to support the non-truncated version when users press ctrl-s  to relax the height constraints but allow flicker.

packages/cli/src/ui/components/messages/DiffRenderer.test.tsx
```diff
@@ -226,15 +297,19 @@ fileDiff Index: file.txt
...
+      />,
     );
     const output = lastFrame();
 
-    expect(output).toContain('1    - const oldVar = 1;');
-    expect(output).toContain('1    + const newVar = 1;');
-    expect(output).toContain('═');
-    expect(output).toContain("20   - const anotherOld = 'test';");
-    expect(output).toContain("20   + const anotherNew = 'test';");
+    expect(output).toEqual(`1    - const oldVar = 1;
```
jacob314, line 308: changed some of these tests to be more explicit golden tests to make sure the rendering is good.

packages/cli/src/ui/components/messages/DiffRenderer.tsx
```diff
@@ -209,16 +231,9 @@ const renderDiffContent = (
...
-              key={`gap-${index}`}
-              width="100%"
-              borderTop={true}
-              borderBottom={false}
-              borderRight={false}
-              borderLeft={false}
-              borderStyle="double"
-              borderColor={Colors.Gray}
-            ></Box>,
+            <Box key={`gap-${index}`}>
```
jacob314, line 234: for discussion: We could switch this back to using a box to draw the line but that requires extending MaxSizedBox to allow Boxes with styling. Displaying this as a Box instead of Text seems more verbose anyway so don't see the value.

packages/cli/src/ui/components/messages/ToolConfirmationMessage.tsx
```diff
@@ -115,12 +144,44 @@ export const ToolConfirmationMessage: React.FC<
...
+    }
+    bodyContent = (
+      <Box flexDirection="column">
+        <Box paddingX={1} marginLeft={1}>
+          <MaxSizedBox
+            maxHeight={bodyContentHeight}
+            maxWidth={Math.max(childWidth - 4, 1)}
+          >
+            <Box>
+              <Text color={Colors.AccentCyan}>{executionProps.command}</Text>
```
jacob314, line 160: previously commands did not have any truncation protection.

packages/cli/src/ui/components/shared/RadioButtonSelect.tsx
```diff
@@ -113,14 +114,18 @@ export function RadioButtonSelect<T>({
...
+        <Text color={textColor} wrap="truncate">
           {itemWithThemeProps.themeNameDisplay}{' '}
           <Text color={Colors.Gray}>{itemWithThemeProps.themeTypeDisplay}</Text>
         </Text>
       );
     }
 
-    return <Text color={textColor}>{label}</Text>;
+    return (
+      <Text color={textColor} wrap="truncate">
```
jacob314, line 125: change this to ensure radio buttons don't wrap. If radio buttons take more than 1 line each that breaks UI assuming a fixed height per radio button entry. Generally we don't have any radio buttons that need to wrap given a plausible screen width so this is fine.

packages/cli/src/ui/components/shared/RadioButtonSelect.tsx
```diff
@@ -78,11 +78,12 @@ export function RadioButtonSelect<T>({
...
-      <Box marginRight={1}>
-        <Text color={isSelected ? Colors.AccentGreen : Colors.Foreground}>
-          {isSelected ? '●' : '○'}
-        </Text>
-      </Box>
+      <Text
+        color={isSelected ? Colors.AccentGreen : Colors.Foreground}
+        wrap="truncate"
+      >
+        {isSelected ? '● ' : '○ '}
```
jacob314, line 85: can revert this change. Simplified to not add a Box for no reason. I was hoping this would also help resolve the issue that something in Ink removes the space when the line becomes too long but it doesn't. there may be an alternate space character we can use to fix this.

packages/cli/src/ui/utils/textUtils.ts
```diff
@@ -45,3 +45,25 @@ export function isBinary(
   // If no NULL bytes were found in the sample, we assume it's text.
   return false;
 }
+
+/*
```
jacob314, line 49: these aren't new. they are moved from text-buffer.ts

packages/cli/src/ui/components/shared/text-buffer.ts
```diff
@@ -69,28 +70,6 @@ function clamp(v: number, min: number, max: number): number {
   return v < min ? min : v > max ? max : v;
 }
 
-/*
```
jacob314, line 72: moved to text-utils

packages/cli/src/ui/components/DetailedMessagesDisplay.tsx
```diff
@@ -29,47 +32,50 @@ export const DetailedMessagesDisplay: React.FC<
...
     >
       <Box marginBottom={1}>
         <Text bold color={Colors.Foreground}>
           Debug Console <Text color={Colors.Gray}>(ctrl+O to close)</Text>
         </Text>
       </Box>
-      {messages.map((msg, index) => {
-        let textColor = Colors.Foreground;
-        let icon = '\u2139'; // Information source (ℹ)
+      <MaxSizedBox maxHeight={maxHeight} maxWidth={width - 4}>
```
NTaylorMullen: Extract 4 to be an lol variable or something else

packages/cli/src/ui/components/ThemeDialog.tsx
```diff
@@ -94,6 +98,33 @@ export function ThemeDialog({
         : `(Modified in ${otherScope})`;
   }
 
+  // Constants for calculating preview pane layout.
+  // These values are based on the JSX structure below.
+  const PREVIEW_PANE_WIDTH_PERCENTAGE = 0.55;
+  // A safety margin to prevent text from touching the border.
+  const PREVIEW_PANE_WIDTH_SAFETY_MARGIN = 0.9;
```
NTaylorMullen, line 106: We may want to centralize the .9 hack OR ignore

packages/cli/src/ui/components/messages/ToolConfirmationMessage.tsx
```diff
@@ -47,11 +56,40 @@ export const ToolConfirmationMessage: React.FC<
...
+      MARGIN_QUESTION_BOTTOM +
+      HEIGHT_OPTIONS;
+    return Math.max(availableTerminalHeight - surroundingElementsHeight, 1);
+  }
   if (confirmationDetails.type === 'edit') {
     if (confirmationDetails.isModifying) {
       return (
         <Box
-          minWidth="90%"
+          minWidth=""
```
NTaylorMullen: intentional?

packages/cli/src/ui/components/messages/ToolConfirmationMessage.tsx
NTaylorMullen, line 1: Odd diffing from GitHub here for future onlookers.

One thing to follow up on is MCP support validation

packages/cli/src/ui/components/shared/RadioButtonSelect.tsx
NTaylorMullen: We should validate that the spacing is the same as before (or if it looks good, feel free to keep)

NTaylorMullen (approved):

packages/cli/src/ui/components/DetailedMessagesDisplay.tsx
jacob314: done

packages/cli/src/ui/components/ThemeDialog.tsx
jacob314: done.

packages/cli/src/ui/components/messages/ToolConfirmationMessage.tsx
jacob314: no. reverted.

packages/cli/src/ui/components/messages/DiffRenderer.test.tsx
KeijiBranshi: love it

packages/cli/src/ui/components/messages/DiffRenderer.tsx
```diff
@@ -136,18 +141,31 @@ export const DiffRenderer: React.FC<DiffRendererProps> = ({
...
   return renderedOutput;
 };
 
 const renderDiffContent = (
   parsedLines: DiffLine[],
-  filename?: string,
+  filename: string | undefined,
   tabWidth = DEFAULT_TAB_WIDTH,
+  availableTerminalHeight: number | undefined,
+  terminalWidth: number,
```
KeijiBranshi, line 168: thought(non-blocking): it's a bit of a sin to use nested render functions like this. but also, i realize it was already here and isn't strictly related to your PR. mostly just noting for the record:

https://alexkondov.com/tao-of-react/#avoid-nested-render-functions

packages/cli/src/ui/components/messages/DiffRenderer.tsx
```diff
@@ -136,18 +141,31 @@ export const DiffRenderer: React.FC<DiffRendererProps> = ({
...
+    );
   }
 
   return renderedOutput;
 };
 
 const renderDiffContent = (
   parsedLines: DiffLine[],
-  filename?: string,
+  filename: string | undefined,
```
KeijiBranshi, line 165: question(non-blocking): why the switch to `string | undefined`? it doesn't impact logic, but ive just personally preferred `?` since its shorter

packages/cli/src/ui/components/messages/DiffRenderer.tsx
```diff
@@ -191,7 +209,11 @@ const renderDiffContent = (
   const MAX_CONTEXT_LINES_WITHOUT_GAP = 5;
 
   return (
-    <Box flexDirection="column" key={key}>
+    <MaxSizedBox
+      maxHeight={availableTerminalHeight}
+      maxWidth={terminalWidth}
+      key={key}
```
KeijiBranshi, line 215: thought: keys are typically used for list items passed as children components. so it might be a code smell to use it at the top level like this.

packages/cli/src/ui/components/shared/MaxSizedBox.tsx
```diff
@@ -0,0 +1,511 @@
...
+    }
+    if (element.type === Box) {
+      layoutInkElementAsStyledText(element, maxWidth!, laidOutStyledText);
+      return;
+    }
+
+    debugReportError('MaxSizedBox children must be <Box> elements', element);
+  }
+
+  React.Children.forEach(children, visitRows);
```
KeijiBranshi, line 131: thought: just a heads up http://screen/7Vq98szsrrJJUbU

packages/cli/src/ui/components/shared/MaxSizedBox.tsx
```diff
@@ -0,0 +1,511 @@
...
+ *
+ * `MaxSizedBox` requires a specific structure for its children to correctly
+ * measure and render the content:
+ *
+ * 1.  **Direct children must be `<Box>` elements.** Each `<Box>` represents a
+ *     single row of content.
+ * 2.  **Row `<Box>` elements must contain only `<Text>` elements.** These
+ *     `<Text>` elements can be nested and there are no restrictions to Text
+ *     element styling other than that non-wrapping text elements must be
+ *     before wrapping text elements.
```
KeijiBranshi, line 60: suggestion: this seems very difficult to enforce without more advanced types. I don't remember off the top of my head, but it's something like:

```
interface MaxSizedBoxProps {
  children?: React.ReactElement<BoxProps> || React.ReactElement<BoxProps>[]
}
```

KeijiBranshi: But given the flexibility of the `Box` type as well, it might be better to also wrap that as well to enforce the children need to be `Text` components

NTaylorMullen (approved):

packages/cli/src/ui/components/messages/DiffRenderer.tsx
jacob314: it impacts lints about required parameters after optional parameters and I'd rather have filename not be optional so the terminalWidth and height can be the final parameters where they are a bit more idiomatic.

packages/cli/src/ui/components/shared/MaxSizedBox.tsx
jacob314: suspected you would comment about that. I'm intentionally breaking the rules because it gives me the power to implement this in much less time than layering where it should eventually be in the stack which is at the same level to how Ink currently handles Layout and ascii art generation for Box and Text. The reason it is ok in this case is I am strict about the children only being <Text> and <Box> elements with strict debug checks so there are not surprises for users.

packages/cli/src/ui/components/messages/DiffRenderer.tsx
jacob314: ack. agree that should be cleaned up all over the code base. cleaning this up here would actually make things harder as I just want a soup of styled <Text> is exactly what I need here to truncate and render styled terminal output simply until we can invest in an appropriate truncated box layout engine that operates one level deeper in the stack..

packages/cli/src/ui/components/messages/DiffRenderer.tsx
jacob314: Note the previous code also had the key. Agree it doesn't do anything. Removing. I think it is something we added a couple months ago hoping to reduce flickering or because it was needed to avoid one of the lints about list items without keys

PR #1218, opened 2025-06-19
feat: update default gemini flash model to GA 2.5 flash

jimmyliao:
feat: update default gemini flash model to GA 2.5 flash


NTaylorMullen:
Unfortunately looks like I beat you to the punch here @jimmyliao 😄 . Appreciate the contrib though, closing for now ❤️

PR #1231, opened 2025-06-19
Update issue templates

aper0lspritz:
update to add /about

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -4,7 +4,10 @@ about: Create a report to help us improve
 title: ''
 labels: ''
 assignees: ''
+
 ---
+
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -4,8 +4,11 @@ about: Create a report to help us improve
 title: ''
 labels: ''
 assignees: ''
+
 ---
 
+/about
```
cperry-goog, line 10: I think we want to put this at the bottom and say something like:

**Client information**
Please paste the full text from the /about command run from Gemini CLI. Also include which platform (MacOS, Windows, Linux)

aper0lspritz: roger.

PR #1247, opened 2025-06-20
Tactical client lifetime fix.

NTaylorMullen:
## TLDR

This is a tactical fix for the issue which was plagueing our integration tests where they'd occasionally fail with "undefined client" exceptions.

## Dive Deeper

A longer leaded fix is to solidify the lifetime of our client object so that it cannot be called in an invalid state OR immediately explodes.

## Reviewer Test Plan
Normal

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅   | ❓  | ❓  |
| npx      | ✅   | ❓  | ❓  |
| Docker   | ✅   | ❓  | ❓  |
| Seatbelt | ✅  | -   | -   |

### DIFF
--- a/packages/core/src/config/config.ts
+++ b/packages/core/src/config/config.ts
@@ -113,7 +113,7 @@ export interface ConfigParameters {
 }
 
 export class Config {
-  private toolRegistry: Promise<ToolRegistry>;
+  private toolRegistry!: ToolRegistry;
   private readonly sessionId: string;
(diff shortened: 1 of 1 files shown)


mattKorwel (approved):
I'm on my phone, so I'm not running this, I presume you validated it does what we thought it does?

NTaylorMullen:
Yes!

PR #1250, opened 2025-06-20
Bug fix telemetry token count 

abhipatel12:
## TLDR

This pull request fixes a bug in the telemetry system where token usage for streaming responses was calculated incorrectly. It changes the logic to correctly use the `usageMetadata` from the final chunk of the response, which contains the accurate total, instead of summing the metadata from all intermediate chunks.

## Dive Deeper

The `getFinalUsageMetadata` function in `packages/core/src/telemetry/loggers.ts` was previously iterating over every chunk in a streaming `GenerateContentResponse` and summing the token counts. This resulted in significantly inflated and inaccurate token usage metrics being reported.

The fix refactors this function to find the *last* chunk in the response array that contains a `usageMetadata` object and use that object as the single source of truth for the token counts. This aligns with the intended behavior of the Gemini API, which provides the final, complete usage statistics in the last message of a stream.

## Reviewer Test Plan

The validation for this change is primarily covered by unit tests.

1.  **Review the Code:**
    *   Inspect the change in `packages/core/src/telemetry/loggers.ts` to see the updated `getFinalUsageMetadata` function.
    *   Review the new unit tests in `packages/core/src/telemetry/loggers.test.ts` under the `describe('getFinalUsageMetadata', ...)` block.

2.  **Run Tests:**
    *   Pull down the branch and run the tests for the `core` package to ensure they all pass.
    *   Execute the following command from the root of the project:
        ```bash
        npm test -w packages/core loggers.test.ts
        ```
    *   Confirm that the new tests pass, validating the logic under various conditions (e.g., metadata in the last chunk, an intermediate chunk, or no chunks).

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ✅  | -   | -   |

### DIFF
--- a/packages/core/src/telemetry/loggers.ts
+++ b/packages/core/src/telemetry/loggers.ts
@@ -288,42 +288,14 @@ export function logApiResponse(
   recordTokenUsageMetrics(config, event.model, event.tool_token_count, 'tool');
 }
 
-export function combinedUsageMetadata(
+export function getFinalUsageMetadata(
   chunks: GenerateContentResponse[],
-): GenerateContentResponseUsageMetadata {
-  const metadataKeys: Array<keyof GenerateContentResponseUsageMetadata> = [
-    'promptTokenCount',
-    'candidatesTokenCount',
-    'cachedContentTokenCount',
-    'thoughtsTokenCount',
-    'toolUsePromptTokenCount',
-    'totalTokenCount',
-  ];
+): GenerateContentResponseUsageMetadata | undefined {
+  //
+  const lastChunkWithMetadata = chunks
+    .slice()
+    .reverse()
+    .find((chunk) => chunk.usageMetadata);
 
-  const totals: Record<keyof GenerateContentResponseUsageMetadata, number> = {
-    promptTokenCount: 0,
-    candidatesTokenCount: 0,
-    cachedContentTokenCount: 0,
-    thoughtsTokenCount: 0,
-    toolUsePromptTokenCount: 0,
-    totalTokenCount: 0,
-    cacheTokensDetails: 0,
-    candidatesTokensDetails: 0,
-    promptTokensDetails: 0,
-    toolUsePromptTokensDetails: 0,
-    trafficType: 0,
-  };
-
-  for (const chunk of chunks) {
-    if (chunk.usageMetadata) {
-      for (const key of metadataKeys) {
-        const chunkValue = chunk.usageMetadata[key];
-        if (typeof chunkValue === 'number') {
-          totals[key] += chunkValue;
-        }
-      }
-    }
-  }
-
-  return totals as unknown as GenerateContentResponseUsageMetadata;
+  return lastChunkWithMetadata?.usageMetadata;
 }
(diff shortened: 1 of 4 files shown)


mattKorwel (approved):
Nice catch. Is there any documentation you could link as a follow on for future us where it details "the metatdata in the final chunk" is the accurate total? Would be good to have that if possible, even just in the comments.

LGTM

packages/core/src/telemetry/loggers.ts
```diff
@@ -288,42 +288,14 @@ export function logApiResponse(
...
-  const metadataKeys: Array<keyof GenerateContentResponseUsageMetadata> = [
-    'promptTokenCount',
-    'candidatesTokenCount',
-    'cachedContentTokenCount',
-    'thoughtsTokenCount',
-    'toolUsePromptTokenCount',
-    'totalTokenCount',
-  ];
+): GenerateContentResponseUsageMetadata | undefined {
+  //
```
NTaylorMullen:
```suggestion
  // Only the last streamed item has the final token count.
```

abhipatel12:
> Nice catch. Is there any documentation you could link as a follow on for future us where it details "the metatdata in the final chunk" is the accurate total? Would be good to have that if possible, even just in the comments.
> 
> LGTM

To be honest, this is not documented very clearly anywhere that I could find. When I wrote the aggregation logic for `/stats` I had to test this locally to see what the chunks look like. There are some community posts about this but I think the official docs should be updated to clearly state this behavior. 

Here is sample of 3 chunks from the same stream. For context, this call was reading the `README.md` file:

```json
Usage metadata from stream 1: {
  promptTokenCount: 12284,
  totalTokenCount: 12354,
  cachedContentTokenCount: 8074,
  promptTokensDetails: [ { modality: 'TEXT', tokenCount: 12284 } ],
  cacheTokensDetails: [ { modality: 'TEXT', tokenCount: 8074 } ],
  thoughtsTokenCount: 70
}
Usage metadata from stream 1: {
  promptTokenCount: 12284,
  totalTokenCount: 12415,
  cachedContentTokenCount: 8074,
  promptTokensDetails: [ { modality: 'TEXT', tokenCount: 12284 } ],
  cacheTokensDetails: [ { modality: 'TEXT', tokenCount: 8074 } ],
  thoughtsTokenCount: 131
}
Usage metadata from stream 1: {
  promptTokenCount: 12284,
  candidatesTokenCount: 30,
  totalTokenCount: 12445,
  cachedContentTokenCount: 8074,
  promptTokensDetails: [ { modality: 'TEXT', tokenCount: 12284 } ],
  cacheTokensDetails: [ { modality: 'TEXT', tokenCount: 8074 } ],
  thoughtsTokenCount: 131
}
```
Looking at this, you can see how it appears that certain fields seem to be the exact same. This leads me to believe the community notes for this behavior is true and that the last chunk is likely the only one necessary to see that usage data.

PR #1252, opened 2025-06-20
Handle stdin for prompts using readline for escape character parsing

bbiggs:
## TLDR

Switch out input handling for the prompt from ink's useInput to a new hook that uses readline to parse character sequences.

## Dive Deeper

The prompt requires careful handling of input character sequences for broad compatibility complex editing of prompts. Currently ui/components/shared/text-buffer.ts is growing to include particular handling (ex: VSCode's shift-enter code) that seems better suited for a lower layer. This change introduces a new hook, useKeypress, affording a location in the code to handle character sequences. To keep this small and robust, we delegate to node's readline library to handle the basics of decoding terminal escape sequences.

I am interested in critical feedback for this change. This may be the wrong approach.

There's a messy thing where the code arbitrarily decides that any input longer than 6 characters is probably a paste, and should be inserted into the prompt verbatim. A more conservative approach may increase this size, as slow ssh connections or very-fast-typers might run up against this and have their input sequences turn into garbly junk. A more correct approach would be to enumerate any specific actions that should happen on multi-character pastes (such as prompt submission or any special substitutions) and pass along the full paste size with each character.

## Reviewer Test Plan

Creating this patch required a lot of manual testing of different keyboard input scenarios, and I don't think I got them all. I am eager for feedback on the approach and whether this is worth pursuing.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes #1212

### DIFF
--- a/packages/cli/src/ui/components/InputPrompt.tsx
+++ b/packages/cli/src/ui/components/InputPrompt.tsx
@@ -294,7 +294,7 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
         const [row, col] = buffer.cursor;
         const line = buffer.lines[row];
         const charBefore = col > 0 ? cpSlice(line, col - 1, col) : '';
-        if (key.ctrl || charBefore === '\\') {
+        if (key.ctrl || charBefore === '\\' || key.paste) {
           // Ctrl+Enter or escaped newline
           if (charBefore === '\\') {
             buffer.backspace();
--- a/packages/cli/src/ui/components/InputPrompt.tsx
+++ b/packages/cli/src/ui/components/InputPrompt.tsx
@@ -5,7 +5,7 @@
  */
 
 import React, { useCallback, useEffect, useState } from 'react';
-import { Text, Box, useInput } from 'ink';
+import { Box, Text } from 'ink';
 import { Colors } from '../colors.js';
 import { SuggestionsDisplay } from './SuggestionsDisplay.js';
 import { useInputHistory } from '../hooks/useInputHistory.js';
@@ -16,6 +16,7 @@ import stringWidth from 'string-width';
 import process from 'node:process';
 import { useShellHistory } from '../hooks/useShellHistory.js';
 import { useCompletion } from '../hooks/useCompletion.js';
+import { useKeypress, Key } from '../hooks/useKeypress.js';
 import { isAtCommand, isSlashCommand } from '../utils/commandUtils.js';
 import { SlashCommand } from '../hooks/slashCommandProcessor.js';
 import { Config } from '@gemini-cli/core';
@@ -155,29 +156,29 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
     ],
   );
 
-  useInput(
-    (input, key) => {
+  const handleInput = useCallback(
+    (key: Key) => {
       if (!focus) {
         return;
       }
       const query = buffer.text;
 
-      if (input === '!' && query === '' && !completion.showSuggestions) {
+      if (key.name === '!' && query === '' && !completion.showSuggestions) {
         setShellModeActive(!shellModeActive);
         buffer.setText(''); // Clear the '!' from input
         return true;
       }
 
       if (completion.showSuggestions) {
-        if (key.upArrow) {
+        if (key.name === 'up') {
           completion.navigateUp();
           return;
         }
-        if (key.downArrow) {
+        if (key.name === 'down') {
           completion.navigateDown();
           return;
         }
-        if (key.tab) {
+        if (key.name === 'tab') {
           if (completion.suggestions.length > 0) {
             const targetIndex =
               completion.activeSuggestionIndex === -1
@@ -189,7 +190,7 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
           }
           return;
         }
-        if (key.return) {
+        if (key.name === 'return') {
           if (completion.activeSuggestionIndex >= 0) {
             handleAutocomplete(completion.activeSuggestionIndex);
           } else if (query.trim()) {
@@ -199,19 +200,19 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
         }
       } else {
         // Keybindings when suggestions are not shown
-        if (key.ctrl && input === 'l') {
+        if (key.ctrl && key.name === 'l') {
           onClearScreen();
-          return true;
+          return;
         }
-        if (key.ctrl && input === 'p') {
+        if (key.ctrl && key.name === 'p') {
           inputHistory.navigateUp();
-          return true;
+          return;
         }
-        if (key.ctrl && input === 'n') {
+        if (key.ctrl && key.name === 'n') {
           inputHistory.navigateDown();
-          return true;
+          return;
         }
-        if (key.escape) {
+        if (key.name === 'escape') {
           if (shellModeActive) {
             setShellModeActive(false);
             return;
@@ -222,54 +223,55 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
       }
 
       // Ctrl+A (Home)
-      if (key.ctrl && input === 'a') {
+      if (key.ctrl && key.name === 'a') {
         buffer.move('home');
         buffer.moveToOffset(0);
         return;
       }
       // Ctrl+E (End)
-      if (key.ctrl && input === 'e') {
+      if (key.ctrl && key.name === 'e') {
         buffer.move('end');
         buffer.moveToOffset(cpLen(buffer.text));
         return;
       }
       // Ctrl+L (Clear Screen)
-      if (key.ctrl && input === 'l') {
+      if (key.ctrl && key.name === 'l') {
         onClearScreen();
         return;
       }
       // Ctrl+P (History Up)
-      if (key.ctrl && input === 'p' && !completion.showSuggestions) {
+      if (key.ctrl && key.name === 'p' && !completion.showSuggestions) {
         inputHistory.navigateUp();
         return;
       }
       // Ctrl+N (History Down)
-      if (key.ctrl && input === 'n' && !completion.showSuggestions) {
+      if (key.ctrl && key.name === 'n' && !completion.showSuggestions) {
         inputHistory.navigateDown();
         return;
       }
 
       // Core text editing from MultilineTextEditor's useInput
-      if (key.ctrl && input === 'k') {
+      if (key.ctrl && key.name === 'k') {
         buffer.killLineRight();
         return;
       }
-      if (key.ctrl && input === 'u') {
+      if (key.ctrl && key.name === 'u') {
         buffer.killLineLeft();
         return;
       }
       const isCtrlX =
-        (key.ctrl && (input === 'x' || input === '\x18')) || input === '\x18';
+        (key.ctrl && (key.name === 'x' || key.sequence === '\x18')) ||
+        key.sequence === '\x18';
       const isCtrlEFromEditor =
-        (key.ctrl && (input === 'e' || input === '\x05')) ||
-        input === '\x05' ||
+        (key.ctrl && (key.name === 'e' || key.sequence === '\x05')) ||
+        key.sequence === '\x05' ||
         (!key.ctrl &&
-          input === 'e' &&
-          input.length === 1 &&
-          input.charCodeAt(0) === 5);
+          key.name === 'e' &&
+          key.sequence.length === 1 &&
+          key.sequence.charCodeAt(0) === 5);
 
       if (isCtrlX || isCtrlEFromEditor) {
-        if (isCtrlEFromEditor && !(key.ctrl && input === 'e')) {
+        if (isCtrlEFromEditor && !(key.ctrl && key.name === 'e')) {
           // Avoid double handling Ctrl+E
           buffer.openInExternalEditor();
           return;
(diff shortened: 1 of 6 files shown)


dewitt:
Not a proper review yet, but just wanted to first share an enthusiastic **thank you** from me for even attempting this! Even at a quick glance, and even with your intentionally keeping the changes small and self-contained (e.g.., no whotesale refactoring of the various key events yet) I can already tell how much cleaner this is, and how much more control we'll get out of it.

I think my open questions will likely center around how well this can handle multiline editing flows, which are critical, though @jacob314 is better equipped to look at that than me.

And I expect we'll have a few easy litmus tests to throw at this early on, since the currently keybinding support is full of holes, and this should quickly be of help.

BTW, is this in a fork? I know there are some things about forks of private repos that act oddly. This might be easier just as a branch off the google-gemini/gemini-cli repo, since I can't clone out of bbiggs/gemini-cli at the moment. (Or maybe I can, and I'm just holding it wrong.)

Thanks again, this is super exciting!

jacob314:
I was able to clone to test locally by 
```git clone git@github.com:bbiggs/gemini-cli.git
git checkout feature/1212-delete-acts-as-backspace
```

The changes work great on my machine with both delete and backspace being handled correctly with a bluetooth keyboard connected to by Mac. Starting a more careful review now.

I like that this plays nicely enough with using ink's directly for input handling that we are still fine to use Ink for all input cases outside the InputPrompt even while the input prompt has focus. For example, pressing ctrl-o to open and close the debug console still works fine with your change.

FWIW Ink also has some similar issues about deciding what code is paste that required us to special case handling `U+007F` (delete) characters that made it into the paste text verbatim.

packages/cli/src/ui/hooks/useKeypress.ts
```diff
@@ -0,0 +1,90 @@
...
+
+    const handleKeypress = (_: unknown, key: Key) => {
+      onKeypress(key);
+    };
+
+    const handleData = (data: Buffer) => {
+      const sequence = data.toString('utf8');
+
+      // Any data chunk that is longer than 6 characters is treated as a paste.
+      if (sequence.length > 6) {
```
jacob314: Idea for a more robust solution: we should add logic to gemini CLI to enable bracketed paste mode. then we can check for the bracketed paste characters here and only assume it is a paste if the chars are present. we would still probably need a fallback mode for terminals that don't support the mode but the majority of users could get the robust behavior.

packages/cli/src/ui/hooks/useKeypress.ts
```diff
@@ -0,0 +1,90 @@
...
+
+  useEffect(() => {
+    if (!isActive || !stdin.isTTY) {
+      return;
+    }
+
+    const keypressStream = new PassThrough();
+    const rl = readline.createInterface({
+      input: keypressStream,
+      escapeCodeTimeout: 50,
```
jacob314: comment on why we are using a timeout much lower than the default of 500.

packages/cli/src/ui/hooks/useKeypress.ts
jacob314: the current behavior of this for 5 character pastes is a bit of a blocker.
if you paste "abcde" you only get "a" in the buffer. if you ended up with "abcde" that would be fine.
similarly if you pasted
ab[delete]cde
would be ok if that ended up as 
a
[delete]
cde

jacob314 (commented):
This is super promising! if we can address the paste issue I'm open to switching directly to this. The other option is to use a combo approach we we still use Ink for all processing all simple input (paste + typed keys) and rely on this code for all other keyboard input.
That could unblock us from the cases ink does not handle well without introducing risk for basic text input.

packages/cli/src/ui/hooks/useKeypress.ts
bbiggs: Will add bracketed paste mode which I suspect will be better. In the mean time, I fixed the 5-character bug: the listener was being dropped due to useEffect having a dependency on onKeypress, thereby having the listener be removed before processing was completed.

bbiggs: I updated the change to use bracketed paste mode, it works well and is simpler.

packages/cli/src/ui/hooks/useKeypress.ts
bbiggs: I opted to drop the override for now until I can prove it's useful/necessary. Makes this change simpler :)

packages/cli/src/ui/hooks/useKeypress.ts
```diff
@@ -43,8 +44,12 @@ export function useKeypress(
     }
 
     setRawMode(true);
+    // Enable bracketed paste mode.
+    process.stdout.write('\x1B[?2004h');
```
jacob314: this is activating every time isActive changes. Should we instead enable this in gemini.tsx when we start up the app and disable it on exit of the app? would be nice if we tried our best to send the control code to exit the mode as I believe it persists after our app exits making it annoying for users who are just trying to use the terminal.

packages/cli/src/ui/hooks/useKeypress.ts
```diff
@@ -54,41 +59,27 @@ export function useKeypress(
...
-          name: '',
-          ctrl: false,
-          meta: false,
-          shift: false,
-          sequence,
-        });
-        return;
+      if (key.name === 'paste-start') {
+        isPaste = true;
+      } else if (key.name === 'paste-end') {
```
jacob314, line 58: rather than emitting the pastes as key presses can you instead build up a buffer here of the in progress paste and emit the entire paste as a single event? otherwise additional changes will be needed to text-buffer.ts and InputPrompt.ts as they have existing ugly logic for handling pastes. For example, we want to know when a file path was pasted in at once the automatically add a leading `@`. This is most useful for dragging files into the terminal.

jacob314 (commented):
Looks really good! LGTM once these comments are addressed.

packages/cli/src/ui/hooks/useKeypress.ts
bbiggs: Yes, this is quite chatty. It felt like there were two approaches, one would be a global init/exit for tty input that sets and restores terminal state, and for correctness we would need to close the tty on any unclean exit. My alternative here is simply to set it up whenever we are in focus and handling user input, to minimize chances of being in this state on an unclean exit.

packages/cli/src/ui/hooks/useKeypress.ts
bbiggs: Done. Verified that `@` handling works, and pasting is much faster. This preserves the old behaviour, but does mean that pasting sequences doesn't work, which can be good or bad depending on your point of view. Seems like the right thing to do right now.

packages/cli/src/ui/hooks/useKeypress.ts
bbiggs: I updated the implementation to split out enabling and disabling bracketed paste mode into its own hook, triggering enablement as part of App.tsx. I also disable it on clean and unclean exit. Seems pretty clean.

packages/cli/src/ui/components/InputPrompt.tsx
```diff
@@ -222,54 +223,55 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
...
       }
-      if (key.ctrl && input === 'u') {
+      if (key.ctrl && key.name === 'u') {
         buffer.killLineLeft();
         return;
       }
       const isCtrlX =
-        (key.ctrl && (input === 'x' || input === '\x18')) || input === '\x18';
+        (key.ctrl && (key.name === 'x' || key.sequence === '\x18')) ||
+        key.sequence === '\x18';
```
dewitt, line 264: Nit, as it's okay to leave for a future cleanup, I believe this expression can be simplified to:

`(key.ctrl && key.name === 'x') || key.sequence === '\x18'`

Because (A AND (B OR C)) OR (C) == (A AND B) OR C

Note: That might not hold if key.ctrl unexpectedly clears key.sequence, but that would be a bug we should address, no?

packages/cli/src/ui/components/InputPrompt.tsx
```diff
@@ -222,54 +223,55 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
...
       const isCtrlEFromEditor =
-        (key.ctrl && (input === 'e' || input === '\x05')) ||
-        input === '\x05' ||
+        (key.ctrl && (key.name === 'e' || key.sequence === '\x05')) ||
+        key.sequence === '\x05' ||
```
dewitt, line 267: As above, this may be simplified to:

`(key.ctrl && key.name === 'e') || key.sequence === '\x05'`

packages/cli/src/ui/components/InputPrompt.tsx
```diff
@@ -222,54 +223,55 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
...
         (!key.ctrl &&
-          input === 'e' &&
-          input.length === 1 &&
-          input.charCodeAt(0) === 5);
+          key.name === 'e' &&
+          key.sequence.length === 1 &&
+          key.sequence.charCodeAt(0) === 5);
```
dewitt, line 271: Nit: Can this be removed?

Wouldn't we expect the comparison `key.sequence === '\x05'` to already match here as `(s.length === 1 && s.charCodeAt(0) === _n_)` should always equal `s === '\x_n_', no?

I think this is perhaps catching a case where the control code `ENQ` is set in the `sequence`, but the key.ctrl is not propagated, which itself is suspect?

This code is just confusing, and predates this PR.

Maybe check this PR in with the existing logic, then we'll do a pass through all the key event handling in a follow up PR, which in theory will be easier to do now.

dewitt (commented):
I eyeballed some of the keybindings in more detail and noted some opportunities for simplification and/or correction, but we can go back and do them in a followup PR to minimize the change in logic in this PR.

dewitt:
Non-blocking, but for posterity: I'm sure this is user error on my end, but even though @jacob314 [commented](https://github.com/google-gemini/gemini-cli/pull/1252#issuecomment-2992399395) that he was able to fetch this branch and test it locally, this is still failing on my workstation:

```
% git fetch
% git checkout feature/1212-delete-acts-as-backspace                             
error: pathspec 'feature/1212-delete-acts-as-backspace' did not match any file(s) known to git
```

I wonder how that worked for Jacob, since this PR looked like a branch in the bbiggs/gemini-cli private fork.

jacob314:
> Non-blocking, but for posterity: I'm sure this is user error on my end, but even though @jacob314 [commented](https://github.com/google-gemini/gemini-cli/pull/1252#issuecomment-2992399395) that he was able to fetch this branch and test it locally, this is still failing on my workstation:
> 
> ```
> % git fetch
> % git checkout feature/1212-delete-acts-as-backspace                             
> error: pathspec 'feature/1212-delete-acts-as-backspace' did not match any file(s) known to git
> ```
> 
> I wonder how that worked for Jacob, since this PR looked like a branch in the bbiggs/gemini-cli private fork.

Here is what works for me to try out this branch. I have a separate directory of clones of the Gemini CLI repo just for reviewing people's code. I find that workflow works well as when I am done reviewing I just delete the clones and the repo is small enough and quick enough to build that it doesn't matter.

```
git clone git@github.com:bbiggs/gemini-cli.git
cd gemini-cli/
git checkout feature/1212-delete-acts-as-backspace
```

packages/cli/src/ui/hooks/useBracketedPaste.ts
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * Enables and disables bracketed paste mode in the terminal.
+ *
+ * This hook ensures that bracketed paste mode is enabled when the component
+ * mounts and disabled when it unmounts or when the process exits.
+ */
+export const useBracketedPaste = () => {
+  const cleanup = () => {
+    if (isCleanedUp) {
```
jacob314: optional nit: don't think isCleanedUp is needed. If you want to keep something like that maybe would be best to keep track of the count of the number of bracketed pastes we are inside. that way you can ensure we are calling ENABLE_BRACKETED_PASTE no more than once and that we are only calling DISABLE_BRACKETED_PASTE when all useBracketedPaste calls in the app have ended.

packages/cli/src/ui/hooks/useKeypress.ts
```diff
@@ -0,0 +1,99 @@
...
+      // If we are in the middle of a paste, send what we have.
+      if (isPaste) {
+        onKeypressRef.current({
+          name: '',
+          ctrl: false,
+          meta: false,
+          shift: false,
+          paste: true,
+          sequence: pasteBuffer.toString(),
+        });
```
jacob314, line 95: clear out paste buffer after sending

jacob314 (approved):
lgtm
Amazing work. Reviewed and tried it locally and everything I'm able to think of still works robustly including dragging files and small pastes.

dewitt:
> > I wonder how that worked for Jacob, since this PR looked like a branch in the bbiggs/gemini-cli private fork.
> 
> Here is what works for me to try out this branch. 

@bbiggs figured out that you are an admin on the source repo so you could see all downstream forks. He just added me to his repo and we're all set now. Thanks!

jacob314 (approved):

Lee-WonJun:
When will the last commit be merged into main?
It looks like all related PRs have been closed.
Is there a scheduled release date?

jacob314:
Lets get this merged into the new main branch. For discussiom we might need to put this behind a setting so we can test it with users before enabling it broadly as it is risky as changing how we support keyboard handling is a big change.

PR #1256, opened 2025-06-20
feat(auth): handle auth flow errors gracefully

NTaylorMullen:
## TLDR
- Catches exceptions during the authentication flow.
- Displays a user-friendly error message in the UI instead of crashing the application.
    - Special cases login with Google so that we return a graceful error message.
- Fixes a flaky test in  to ensure deterministic behavior.

image

## Reviewer Test Plan

Login with your Google account  BUT pick your enterprise work account -> fail

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅   | ❓  | ❓  |
| npx      | ✅   | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |

### DIFF
--- a/packages/cli/src/ui/App.tsx
+++ b/packages/cli/src/ui/App.tsx
@@ -587,11 +587,6 @@ const App = ({ config, settings, startupWarnings = [] }: AppProps) => {
             </Box>
           ) : isAuthDialogOpen ? (
             <Box flexDirection="column">
-              {authError && (
-                <Box marginBottom={1}>
-                  <Text color={Colors.AccentRed}>{authError}</Text>
-                </Box>
-              )}
               <AuthDialog
                 onSelect={handleAuthSelect}
                 onHighlight={handleAuthHighlight}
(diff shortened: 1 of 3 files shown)


packages/cli/src/ui/App.tsx
```diff
@@ -587,11 +587,6 @@ const App = ({ config, settings, startupWarnings = [] }: AppProps) => {
             </Box>
           ) : isAuthDialogOpen ? (
             <Box flexDirection="column">
-              {authError && (
```
NTaylorMullen, line 590: This was duplicated with the auth dialog below. The auth dialog shows the error. @mattKorwel am I missing something?

packages/cli/src/ui/hooks/usePhraseCycler.test.ts
mattKorwel, line 1: approving but are these changes intentional?

mattKorwel (approved):

packages/cli/src/ui/App.tsx
mattKorwel: Nope you're not. unsure why that was doubled there. gemini got carried away?

packages/cli/src/ui/hooks/usePhraseCycler.test.ts
NTaylorMullen: Ya they were failing locally as being undeterministic

PR #1201, opened 2025-06-18
point Code Assist client to prod server 

MarlonGamez:
## TLDR

The client currently points to staging, this PR updates it to point to prod.

## Reviewer Test Plan

Paid Tier:
- `export GEMINI_CODE_ASSIST=true`
- `export GOOGLE_CLOUD_PROJECT=${VALID_GCP_PROJECT}`
- `npm run  build && npm run start`
- Use as normal!

Free Tier:
- `export GEMINI_CODE_ASSIST=true`
- `unset GOOGLE_CLOUD_PROJECT`
- `npm run  build && npm run start`
- Use as normal!

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| -------- | --- | --- | --- |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

#1165

PR #1265, opened 2025-06-20
Fix CODE_ASSIST_ENDPOINT env var loading

scidomino:
## TLDR

setting CODE_ASSIST_ENDPOINT in .env files stopped working, probably because recent changes moved up when this static var is loaded to before the .env file is loaded. This moves the calculation back so to after the .env files are applied. But it also just makes a little more sense this way.

## Reviewer Test Plan

Set CODE_ASSIST_ENDPOINT in an .env files and verify that it works. (for example, by pointing it to http://derp.com and verifying that the CLI blows up during start up).

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | X  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

### DIFF
--- a/packages/core/src/code_assist/server.ts
+++ b/packages/core/src/code_assist/server.ts
@@ -39,7 +39,6 @@ export interface HttpOptions {
 
 // TODO: Use production endpoint once it supports our methods.
 export const CODE_ASSIST_ENDPOINT =
-  process.env.CODE_ASSIST_ENDPOINT ??
   'https://staging-cloudcode-pa.sandbox.googleapis.com';
 export const CODE_ASSIST_API_VERSION = 'v1internal';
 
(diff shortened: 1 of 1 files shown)


packages/core/src/code_assist/server.ts
```diff
@@ -39,7 +39,6 @@ export interface HttpOptions {
 
 // TODO: Use production endpoint once it supports our methods.
 export const CODE_ASSIST_ENDPOINT =
```
anj-s, line 41: Should we set this to staging?

anj-s (approved):

packages/core/src/code_assist/server.ts
scidomino: We have https://github.com/google-gemini/gemini-cli/pull/1201 to point it to prod in the release branch.

PR #1266, opened 2025-06-20
feat: align cli argv, settings.json, and GEMINI_SANDBOX envvar

KeijiBranshi:
CHILD=#1267
DEMO=http://screen/9nG8XAvWzpks6Ei

## TLDR

Aligns the accepted values for sandbox configuration between the 3 supported sources:

- CLI args
- .gemini/settings.json
- environment variables

Adds extra build-time type guards to ensure strings provided from these sources are all valid values (see DEMO above)

## Deeper Dive

This change adds stronger types, extra validation, and error messages when coalescing these sandbox configuration sources. They also bring the CLI arg up to parity with the settings and envvar options (previously the CLI arg only accepted booleans).

This also adds the extra options `auto` and `none` to _eventually_ replace `0`, `1`, `true`, and `false`. These new values are more explicit about the intended behavior. 

> NOTE: checking this in to the `main` branch for now to avoid friction and breakages in `release`. We can cherry-pick if needed or deemed valuable enough. 

> NOTE: tests and documentation are added in a followup PR (see child) to keep this diff small and more reviewer-friendly

### Detailed Changes

   * `packages/core/src/config/config.ts`: This file introduces the `SupportedSandboxCommand` type, which is a union of the supported sandbox command [literals](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#literal-types). This adds a layer of [build-time type safety](http://screen/9nG8XAvWzpks6Ei) by ensuring that only valid commands can be used in the SandboxConfig.

   * `packages/cli/src/config/config.ts`: This file was updated to improve type safety by using the new `SandboxOption` and `SandboxLegacyOption` types for the `--sandbox` CLI argument. This ensures that only valid sandbox options can be provided on the command line.

   * `packages/cli/src/config/sandboxConfig.ts`: This file was refactored to introduce more robust type checking for sandbox configurations. It now defines and uses the `SandboxOption` and `SandboxLegacyOption` types to validate sandbox settings from all sources (CLI, settings.json, environment variables). It also performs necessary type guards to ensure generic `string` types from environment variables and settings.jsons can be cast to the more strict [literal type](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#literal-types) `SandboxOption`.

   * `packages/cli/src/config/settings.ts`: The sandbox property in the Settings interface was changed from `boolean | string` to just string.

   * `scripts/sandbox_command.js`: This script was updated to align with the new sandbox options. It now correctly handles the 'auto' and 'none' options, making the script's behavior consistent with the new configuration options.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅   | ❓  | ✅   |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ✅   | ❓  | ✅   |
| Podman   | ✅   | -   | ✅    |

### DIFF
--- a/packages/cli/src/config/config.ts
+++ b/packages/cli/src/config/config.ts
@@ -27,7 +27,13 @@ import * as dotenv from 'dotenv';
 import * as fs from 'node:fs';
 import * as path from 'node:path';
 import * as os from 'node:os';
-import { loadSandboxConfig } from './sandboxConfig.js';
+import {
+  loadSandboxConfig,
+  SANDBOX_OPTIONS,
+  SANDBOX_LEGACY_OPTIONS,
+  SandboxOption,
+  SandboxLegacyOption,
+} from './sandboxConfig.js';
 
 // Simple console logger for now - replace with actual logger if available
 const logger = {
@@ -41,8 +47,8 @@ const logger = {
 
 interface CliArgs {
   model: string | undefined;
-  sandbox: boolean | string | undefined;
-  'sandbox-image': string | undefined;
+  sandbox?: Exclude<SandboxOption, SandboxLegacyOption>;
+  'sandbox-image'?: string;
   debug: boolean | undefined;
   prompt: string | undefined;
   all_files: boolean | undefined;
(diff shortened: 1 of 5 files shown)


scidomino (approved):

jacob314 (approved):
lgtm

PR #1274, opened 2025-06-21
Clearcut logging - initial implementation

owenofbrien:
## TLDR

Flag-guarded initial implementation of a clearcut logger to collect telemetry data and send it to Concord for dashboards, etc.

## Dive Deeper

- This change is guarded by a new config flag, disable_data_collection, which defaults to true for now.
- Events are logged in batches with no less than 1 minute between each batch.
- Refactored OTEL logging to reuse event data and much of the logging logic.
- The custom event data in event-metadata-key.ts corresponds to a new proto enum that must be checked into g3 as a requirement from the Clearcut eng team (see cl/773719965). These two enums must be kept in sync manually, as we can't directly depend on code in g3.

Follow-ups:
- Tests will be added in a follow-up PR.
- A unique install_id will be added and logged in a follow-up PR to identify users and allow for DAU metrics, as we currently don't have a good way to identify users. See [go/gemini-cli-clearcut](go/gemini-cli-clearcut) for details.
- A one-time telemetry onboarding screen will be added in a follow-up PR.

## Reviewer Test Plan

For now, please just run the CLI to verify that this change doesn't break anything. Be sure to actively use it for at least a minute to provide time for the Clearcut logger to flush log events, in case it's unintentionally running.

I'm working with the Clearcut team on end-to-end validation of the logging. Full e2e validation will depend on submission of cl/773719965. Also, it takes a full day for logs to propagate into the publicly available Concord tables due to anonymization pipelines, and the test tables are highly restricted.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅   | ❓  | ✅   |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

b/424466618


packages/core/src/telemetry/clearcut-logger/clearcut-logger.ts
```diff
@@ -0,0 +1,231 @@
...
+// Singleton class for batch posting log events to Clearcut. When a new event comes in, the elapsed time
+// is checked and events are flushed to Clearcut if at least a minute has passed since the last flush.
+export class ClearcutLogger {
+  private static instance: ClearcutLogger;
+  private config?: Config;
+  private readonly events: any = [];
+  private last_flush_time: number = Date.now();
+  private flush_interval_ms: number = 1000 * 60; // Wait at least a minute before flushing events.
+
+  private constructor(config?: Config) {
```
anj-s, line 44: should the config be optional?

packages/core/src/telemetry/clearcut-logger/clearcut-logger.ts
```diff
@@ -0,0 +1,231 @@
...
+    this.config = config;
+  }
+
+  public static getInstance(config?: Config): ClearcutLogger | undefined {
```
anj-s: Just to double check: once we have the enum CL approved, we should be able to enable logging to Clearcut by default. Is that correct? If we enabled it now, will we see any errors?

packages/core/src/telemetry/clearcut-logger/clearcut-logger.ts
```diff
@@ -0,0 +1,231 @@
...
+    req.on('error', e => {
+      reject(e);
+    });
+    req.end(body);
+    }).then((buf: Buffer) => {
+      try {
+        this.events.length = 0;
+        return this.decodeLogResponse(buf) || {};
+      } catch {
+        return {};
```
anj-s, line 118: should we log the error here?

packages/core/src/telemetry/clearcut-logger/clearcut-logger.ts
```diff
@@ -0,0 +1,231 @@
...
+      }
+    });
+  }
+
+  // Visible for testing. Decodes protobuf-encoded response from Clearcut server.
+  decodeLogResponse(buf: Buffer): LogResponse | undefined {
+    if (buf.length < 1) {
+      return undefined;
```
anj-s, line 127: Can we add a TODO here to return specific errors for us to debug as needed.

packages/core/src/telemetry/clearcut-logger/clearcut-logger.ts
owenofbrien: That's correct. I did run it on my Cloudtop with Clearcut logging enabled, and I didn't notice any errors.

packages/core/src/telemetry/clearcut-logger/clearcut-logger.ts
owenofbrien: I ended up making it optional because the config was not readily available within the 'SIGINT' handler callback.

packages/core/src/telemetry/clearcut-logger/clearcut-logger.ts
owenofbrien: Done

packages/core/src/telemetry/clearcut-logger/clearcut-logger.ts
owenofbrien: Done

anj-s (approved):

packages/core/src/config/config.ts
```diff
@@ -111,6 +113,7 @@ export interface ConfigParameters {
   fileDiscoveryService?: FileDiscoveryService;
   bugCommand?: BugCommandSettings;
   model: string;
+  disableDataCollection?: boolean;
```
mattKorwel, line 120: How do i turn this on to test? I don't see any reference to this. I was expecting to see a an envFlag or setting. I'm assuming i missed it?

packages/core/src/config/config.ts
```diff
@@ -144,6 +147,7 @@ export class Config {
   private readonly cwd: string;
   private readonly bugCommand: BugCommandSettings | undefined;
   private readonly model: string;
+  private readonly disableDataCollection: boolean = true;
```
mattKorwel:
```suggestion
  private readonly disableDataCollection: boolean;
```

we should keep hard coded values to as few places as possible.

packages/core/src/config/config.ts
```diff
@@ -180,6 +184,7 @@ export class Config {
     this.fileDiscoveryService = params.fileDiscoveryService ?? null;
     this.bugCommand = params.bugCommand;
     this.model = params.model;
+    this.disableDataCollection = params.disableDataCollection ?? true;
```
mattKorwel: when and where do we set the input to config for external variable for disable data collection? I want to merge this to release and let logs run this weekend but i can't find a flag to enable it.

owenofbrien: Sorry, I haven't properly exposed it yet. I was just trying to get the bulk of the code in for now without breaking anything. After I added the custom data, things stopped flowing through to the Sawmill test table. I think cl/773902039 needs time to propagate before all the data can flow through to its ultimate destination in Concord.

If we can get this PR submitted, I'll follow up tomorrow morning with more updates.

mattKorwel (approved):
I'm approving this so we are not blocked. obviously tests need to pass and the data collection needs to be configurable. Ping me if you need anything specific.

PR #1276, opened 2025-06-21
Add support for logging unique user ids for telemetry

anj-s:
* Based off clearcut-logging
* Added unique user ids by storing a UUID in the user's home directory.
* Add support for setting the disable property in the settings file

### DIFF
--- a/packages/core/src/utils/user_id.ts
+++ b/packages/core/src/utils/user_id.ts
@@ -1,0 +1,55 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import * as os from 'os';
+import * as fs from 'fs';
+import * as path from 'path';
+import { randomUUID } from 'crypto';
+
+const geminiDir = path.join(os.homedir(), '.gemini');
+const userIdFile = path.join(geminiDir, 'user_id');
+
+function ensureGeminiDirExists() {
+  if (!fs.existsSync(geminiDir)) {
+    fs.mkdirSync(geminiDir, { recursive: true });
+  }
+}
+
+function readUserIdFromFile(): string | null {
+  if (fs.existsSync(userIdFile)) {
+    const userId = fs.readFileSync(userIdFile, 'utf-8').trim();
+    return userId || null;
+  }
+  return null;
+}
+
+function writeUserIdToFile(userId: string) {
+  fs.writeFileSync(userIdFile, userId, 'utf-8');
+}
+
+/**
+ * Retrieves the persistent user ID from a file, creating it if it doesn't exist.
+ * This ID is used for unique user tracking in telemetry.
+ * @returns A UUID string for the user.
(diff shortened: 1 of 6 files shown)


packages/core/src/utils/user_id.ts
```diff
@@ -0,0 +1,53 @@
...
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import * as os from 'os';
+import * as fs from 'fs';
+import * as path from 'path';
+import { randomUUID } from 'crypto';
+
+const geminiDir = path.join(os.homedir(), '.gemini');
```
mattKorwel: https://github.com/google-gemini/gemini-cli/blob/6c0f1e635bb9d0fb9fb75a2044dbfec2803c48f7/packages/cli/src/config/settings.ts#L21

and https://github.com/google-gemini/gemini-cli/blob/6c0f1e635bb9d0fb9fb75a2044dbfec2803c48f7/packages/core/src/utils/paths.ts#L11

already have code related to this. we shouldn't add new code, rather reuse what we have there.

anj-s: Yep, didn't see this constant before. Will use it here.

packages/core/src/utils/user_id.ts
```diff
@@ -0,0 +1,53 @@
...
+ * This ID is used for unique user tracking.
+ * @returns A UUID string for the user.
+ */
+export function getPersistentUserId(): string {
+  try {
+    ensureGeminiDirExists();
+    let userId = readUserIdFromFile();
+
+    if (!userId) {
+      userId = randomUUID();
```
mattKorwel, line 45: Shouldn't this be some sort of hash?

https://github.com/google-gemini/gemini-cli/pull/1145/files#diff-e0aa7e3fce2abaa1aa14fe47057cc975566fe6cf9c4e451d7d12a298eb711e8c

What's the expected outcome. if i run the cli 10x's won't this give me 10 unique user ids'?

anj-s: You'll only create a UUID if there is no existing file in the home directory. I am assuming we don't delete that file after every run.

packages/core/src/utils/user_id.ts
```diff
@@ -0,0 +1,53 @@
...
+      writeUserIdToFile(userId);
+    }
+
+    return userId;
+  } catch (error) {
+    console.error('Error accessing persistent user ID file, generating ephemeral ID:', error);
+    return "123456789";
```
anj-s, line 52: Should we just error out here? Or return a dummy user id?

owenofbrien: Good question. Using a sentinel ID like what you have here seems like a safe approach. That way we can easily check how often these errors happen by querying the logs for events with the sentinel user id.

owenofbrien (approved):
This looks great. Thank you!

PR #1286, opened 2025-06-21
Drop file search in completion for simpler and faster path completion

### LINKED ISSUE #1285: Slow @ completion when using a remote filesytem
**Describe the bug**

If you use `@...` to complete in the prompt when located in a remote filesystem, it can be very very slow as it searches for the filename.

**Additional context**
The completion code currently searches for the target file using a ** glob, which will recursively look for the file from the current directory path. In a large tree or a remote filesystem, this can be catastrophically slow. It seems we can either remove the feature, make it a configurable option, or somehow be smarter or bounded in how far we look for the given filename.

bbiggs:
## TLDR

When in a remote or large filesystem, doing a recursive glob is very expensive. Drop the feature for a simpler and faster completion.

## Dive Deeper

Currently you can use @-completion to find flies anywhere in the local tree, for example typing in `@useFoo` will search from the current directory for any file called `useFoo` in any subdirectory. When running gemini-cli from a remote filesystem, this can be very expensive and slow. While we could keep the feature or make it optional, this patch opts for just removing the capability for now.

## Reviewer Test Plan

This change should be straightforward as it is just removing a feature, but also possible I broke something! Please actually try this out :)

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes #1285

### DIFF
--- a/packages/cli/src/ui/hooks/useCompletion.integration.test.ts
+++ b/packages/cli/src/ui/hooks/useCompletion.integration.test.ts
@@ -60,8 +60,11 @@ describe('useCompletion git-aware filtering integration', () => {
   });
 
   it('should filter git-ignored entries from @ completions', async () => {
-    const globResults = [`${testCwd}/data`, `${testCwd}/dist`];
-    vi.mocked(glob).mockResolvedValue(globResults);
+    // Mock fs.readdir to return both regular and git-ignored directories
+    vi.mocked(fs.readdir).mockResolvedValue([
+      { name: 'dist', isDirectory: () => true },
+      { name: 'data', isDirectory: () => true },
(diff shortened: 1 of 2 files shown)


mattKorwel:
Thanks for this @bbiggs appreciate the callout that this can be a blocker. Going to investigate this so we can gather some data to inform the go/no go. we'd like to keep it but don't want to if it just doesn't work. Will share findings here.

mattKorwel:
closing in favor of #1290

PR #1288, opened 2025-06-21
Jacob314/overflow notification and one MaxSizedBox bug fix

jacob314:
## TLDR

This pull request has two commits.
The first changes the UI to show "Press Ctrl-S to show more lines" 

Show grey messages saying 
"Press Ctrl-S to show more lines" 
bellow both boxes in the UI where important content is truncated.

Original UX mock from Miguel:
image

Screenshot of how it typically shows tool confirmations
image

Screenshot showing how this looks for the debug console.
image

The second fixes a bug where the optimization to truncate code in CodeColorizer would sometimes truncate all the lines before passing them to MaxSizedBox due to a logic bug. Also added a minimum height to MaxSizedBox to ensure we don't have any cases where were setting MaxSizedBox objects to have too small a height to show some actual text as well as the truncation message.

## Dive Deeper

It is crucial this code doesn't cause infinite re-rendering and in fact Gemini CLIs first implementation of this did.

In the unlikely even that you have confirmation messages for the debug console and the main ui at the same time, we will show the message twice once bellow each box but that is an ok tradeoff to have both messages show more intuitively for the typical cases.

We don't bother showing this for the Theme dialog because while it could truncate, the truncated text barely matters.

### DIFF
--- a/packages/cli/src/ui/components/shared/MaxSizedBox.tsx
+++ b/packages/cli/src/ui/components/shared/MaxSizedBox.tsx
@@ -13,6 +13,13 @@ import { useOverflowActions } from '../../contexts/OverflowContext.js';
 
 let enableDebugLog = false;
 
+/**
+ * Minimum height for the MaxSizedBox component.
+ * This ensures there is room for at least one line of content as well as the
+ * message that content was truncated.
+ */
+export const MINIMUM_MAX_HEIGHT = 2;
+
 export function setMaxSizedBoxDebugging(value: boolean) {
   enableDebugLog = value;
 }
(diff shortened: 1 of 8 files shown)


packages/cli/src/ui/utils/CodeColorizer.tsx
```diff
@@ -111,11 +112,11 @@ export function colorizeCode(
 
     let hiddenLinesCount = 0;
 
-    // Optimizaiton to avoid highlighting lines that cannot possibly be displayed.
-    if (availableHeight && lines.length > availableHeight) {
-      const sliceIndex =
-        lines.length - availableHeight + RESERVED_LINES_FOR_TRUNCATION_MESSAGE;
```
jacob314, line 117: RESERVED_LINES_FOR_TRUNCATION_MESSAGE was wrong.. it should have been 1 not 2. In any case, there was no need for this class to even be aware of how much space the truncation message uses as there is no harm in us passing a couple extra lines as MaxSizedBox handles truncation robustly.
I also tweaked MaxSizedBox to enforce a MINIMUM_MAX_HEIGHT to make sure we didn't pass in a value of 1 or 0 by accident as that would cause problems.

packages/cli/src/ui/components/shared/MaxSizedBox.tsx
```diff
@@ -103,54 +115,71 @@ export const MaxSizedBox: React.FC<MaxSizedBoxProps> = ({
   // In the future we might choose to still apply our layout logic
   // even in this case particularlly if there are cases where we
   // intentionally diverse how certain layouts are rendered.
-  if (maxHeight === undefined) {
```
jacob314, line 106: this code still exists it just had to be refactored around to allow using useEffect to call addOverflowingId an

packages/cli/src/ui/components/shared/MaxSizedBox.tsx
```diff
@@ -103,54 +115,71 @@ export const MaxSizedBox: React.FC<MaxSizedBoxProps> = ({
...
-        {children}
-      </Box>
-    );
-  }
-
-  if (maxWidth === undefined) {
-    throw new Error('maxWidth must be defined when maxHeight is set.');
-  }
+  let targetMaxHeight;
+  if (maxHeight !== undefined) {
```
jacob314, line 119: these two lines are the only logic change. We now force maxHeight to round to an integer (avoids possible bug of showing 3.2 lines truncated if a non-integer max height was passed in) and we enforce MINIMUM_MAX_HEIGHT so we don't show useless messages that contain nothing but a message that content was truncated.

mattKorwel (changes requested):
seeing lots of flickering http://screencast/cast/NjI0OTgwOTc4OTg0NTUwNHwwNzk3ZmE2Zi0yNg. marking request changes just so we can review and i can remember i looked at this one already.

mattKorwel (approved):
confirmed this branch does not introduce the flickering, merging it.

PR #1290, opened 2025-06-22
Add setting enableRecursiveFileSearch to control @-file completion

### LINKED ISSUE #1285: Slow @ completion when using a remote filesytem

bbiggs:
## TLDR

Add a new setting, fileFiltering.enableRecursiveFileSearch, to control recursive @-file completion.

## Dive Deeper

Searching recursively for files by name in the current directory tree can be a performance challenge on remote filesystems. This change adds a new global setting to disable the feature.

## Reviewer Test Plan

I added a test for the basic behaviour but suggestions, but please check that the prompt still works well. 

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes #1285

### DIFF
--- a/docs/cli/configuration.md
+++ b/docs/cli/configuration.md
@@ -56,13 +56,15 @@ In addition to a project settings file, a project's `.gemini` directory can cont
 - **`fileFiltering`** (object):
 
   - **Description:** Controls git-aware file filtering behavior for @ commands and file discovery tools.
-  - **Default:** `"respectGitIgnore": true`
+  - **Default:** `"respectGitIgnore": true, "enableRecursiveFileSearch": true`
   - **Properties:**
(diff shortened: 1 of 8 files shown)


mattKorwel (approved):
tested, validated LGTM

PR #1292, opened 2025-06-22
feat: Add client-initiated tool call handling

abhipatel12:
## TLDR

This change fixes two critical bugs related to tool-calling and state management:

1.  **Crashing on Client-Side Tools**: Resolves a `400 INVALID_ARGUMENT` crash when using client-side commands like `/memory add` by preventing the client from sending a tool response for a tool the API did not request.
2.  **Stale Memory State**: Fixes a bug where newly added memories were not immediately reflected in the UI. The fix ensures that the application's memory is refreshed automatically after a memory-saving operation, providing immediate feedback to the user.

### 1. Crashing on Client-Side Tools

The root cause of the crash was that the client was sending a `functionResponse` to the Gemini API for a tool call that was initiated locally by the user (e.g., via a slash command). The Gemini API requires a strict 1:1 mapping where it only accepts a `functionResponse` for a `functionCall` it explicitly issued in the preceding turn.

This fix introduces a mandatory `isClientInitiated: boolean` flag to the `ToolCallRequestInfo` interface to solve this problem at the data-structure level:

*   **Mandatory Flag**: Every tool call request must now be explicitly marked as either originating from the client (`true`) or from the model (`false`). This provides compile-time safety, preventing this bug from recurring.
*   **Conditional Logic**: The `useGeminiStream` hook now inspects this flag on all completed tool calls.
    *   If `isClientInitiated` is `true`, the hook finalizes the tool's lifecycle purely on the client-side and, crucially, **does not** send any response to the Gemini API.
    *   If `isClientInitiated` is `false`, the existing logic proceeds as expected, sending the tool's output back to the model for further processing.
*   **Safe Invocation**: All tool call creation sites have been updated to set this flag correctly, ensuring a clear distinction between the two types of tool invocations throughout the application.

### 2. Stale Memory State

Previously, when a user added a new memory using the `/memory add` command or by asking the model to remember something, the underlying `GEMINI.md` file was updated, but the application's in-memory representation of that file was not. This meant that subsequent commands, like `/memory show`, would display stale data until the application was restarted.

This fix introduces a robust refresh mechanism to solve this problem:

*   **Automatic Refresh**: The `useGeminiStream` hook now calls the `performMemoryRefresh` function whenever a `save_memory` tool call completes successfully, regardless of whether it was initiated by the user or the model.
*   **Idempotent Effect**: The refresh logic is designed to be idempotent by tracking which tool calls have already been processed. This prevents unnecessary refreshes and ensures that the operation is efficient and reliable, even across component re-renders.

## Reviewer Test Plan

To validate this change, please test the following scenarios:

1.  **Client-Initiated Tool Call (The Fix)**:
    *   Run the command: `/memory add "my favorite color is blue"`
    *   **Expected Behavior**: The command executes successfully, a confirmation appears, and the CLI returns to an idle state without crashing. There should be no `400 INVALID_ARGUMENT` error.

2.  **Stale Memory State (The Fix)**:
    *   Run the command: `/memory add "my dog's name is fido"`
    *   Immediately run: `/memory show`
    *   **Expected Behavior**: The output of `/memory show` should immediately include "my dog's name is fido".

3.  **Model-Initiated Tool Call (Regression Test)**:
    *   Ask a question that requires a tool, such as: `"Can you remember that my favorite bird is a duck."`
    *   Immediately run: `/memory show`
    *   **Expected Behavior**: The model should request the memory tool, the tool should execute, and the result should be sent back to the model. The model should then use the tool's output to answer tell you it will remember that. The entire flow should complete without errors, and the new memory should appear in the output of `/memory show`.

4.  **Standard Chat (Regression Test)**:
    *   Have a simple conversation with the model without using tools.
    *   **Expected Behavior**: The conversation should proceed normally with no change in behavior.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧 |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ✅  | -   | -   |

## Linked issues / bugs
Closes [b/426844709](https://b.corp.google.com/issues/426844709)

### DIFF
--- a/packages/cli/src/nonInteractiveCli.ts
+++ b/packages/cli/src/nonInteractiveCli.ts
@@ -89,6 +89,7 @@ export async function runNonInteractive(
             callId,
             name: fc.name as string,
             args: (fc.args ?? {}) as Record<string, unknown>,
+            isClientInitiated: false,
           };
 
           const toolResponse = await executeToolCall(
--- a/packages/cli/src/ui/App.tsx
+++ b/packages/cli/src/ui/App.tsx
@@ -362,6 +362,7 @@ const App = ({ config, settings, startupWarnings = [] }: AppProps) => {
     shellModeActive,
     getPreferredEditor,
     onAuthError,
+    performMemoryRefresh,
   );
   pendingHistoryItems.push(...pendingGeminiHistoryItems);
   const { elapsedTime, currentLoadingPhrase } =
--- a/packages/cli/src/ui/hooks/useGeminiStream.test.tsx
+++ b/packages/cli/src/ui/hooks/useGeminiStream.test.tsx
@@ -389,6 +389,7 @@ describe('useGeminiStream', () => {
           >,
           shellModeActive: false,
           loadedSettings: mockLoadedSettings,
+          toolCalls: initialToolCalls,
         },
       },
     );
@@ -404,7 +405,12 @@ describe('useGeminiStream', () => {
   it('should not submit tool responses if not all tool calls are completed', () => {
     const toolCalls: TrackedToolCall[] = [
       {
-        request: { callId: 'call1', name: 'tool1', args: {} },
+        request: {
+          callId: 'call1',
+          name: 'tool1',
+          args: {},
+          isClientInitiated: false,
+        },
         status: 'success',
         responseSubmittedToGemini: false,
         response: {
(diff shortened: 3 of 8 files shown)


scottdensmore:
It works when I add memory but when I run `/memory show` it doesn't show what was just added
CleanShot 2025-06-21 at 20 17 03@2x

But it is in the file
CleanShot 2025-06-21 at 20 18 18@2x

abhipatel12:
> It works when I add memory but when I run `/memory show` it doesn't show what was just added CleanShot 2025-06-21 at 20 17 03@2x
> 
> But it is in the file CleanShot 2025-06-21 at 20 18 18@2x

We found a bug in a bug! wooohoo! This is fixed now too

scottdensmore (approved):
Approved

PR #1293, opened 2025-06-22
Jacob314/auto exit unconstrained height mode

jacob314:
## TLDR

Unconstrained mode causes flicker so we shouldn't let users accidentally stay in it.
We now exit unconstrained mode automatically on any keystroke.
As a bonus we also apply our hacks to fixup torn terminal output when needed when exiting unconstrained mode.


SandyTao520 (approved):

mattKorwel (approved):
LGTM

PR #1294, opened 2025-06-22
Provide a .gitconfig for shadow repos

louisjimenez:
This writes a .gitconfig to the shadow repos that back checkpointing. It also sets environment variables to prevent shadow repo git operations from trying to load the user's gitconfig.


scottdensmore (approved):
Approved

PR #1299, opened 2025-06-22
feat: Only show ctrl-s when idle

NTaylorMullen:
## TLDR

- Lowercase ctrl-s to be consistent with the rest of the codebase.
- Do not show the indicator if we're in the middle of streaming content. It should only show when we're idle or awaiting a confirmation.

Flow: https://screencast.googleplex.com/cast/NTIyNTUxMjAzMjAxMDI0MHxlMWU0MDk5MC1kYw

FYI @jacob314 

## Dive Deeper

- We shouldn't be showing `ctrl + s` hot key while streaming, we don't want to encourage that flow because if we do then we're basically encouraging flickering.

## Reviewer Test Plan

See video

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅   | ❓  | ❓  |
| npx      | ✅   | ❓  | ❓  |
| Docker   | ✅    | ❓  | ❓  |
| Podman   | ✅    | -   | -   |

### DIFF
--- a/packages/cli/src/ui/components/ShowMoreLines.tsx
+++ b/packages/cli/src/ui/components/ShowMoreLines.tsx
@@ -6,6 +6,8 @@
 
 import { Box, Text } from 'ink';
 import { useOverflowState } from '../contexts/OverflowContext.js';
+import { useStreamingContext } from '../contexts/StreamingContext.js';
+import { StreamingState } from '../types.js';
 import { Colors } from '../colors.js';
 
 interface ShowMoreLinesProps {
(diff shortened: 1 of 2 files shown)


NTaylorMullen:
@mattKorwel feel free to merge on my behalf given I'm OOO

mattKorwel (approved):

packages/cli/src/ui/components/ShowMoreLines.tsx
```diff
@@ -14,19 +16,24 @@ interface ShowMoreLinesProps {
...
   const overflowState = useOverflowState();
+  const streamingState = useStreamingContext();
 
   if (
     overflowState === undefined ||
     overflowState.overflowingIds.size === 0 ||
-    !constrainHeight
+    !constrainHeight ||
+    !(
+      streamingState === StreamingState.Idle ||
```
jacob314, line 26: good idea!

jacob314 (commented):
lgtm

PR #1302, opened 2025-06-22
{bug} Vertex Auth Support

mattKorwel:
This restores vertex behavior to the state before auth flow so env vars work the same.


scidomino (approved):

scottdensmore (approved):
Approved

PR #1306, opened 2025-06-22
Fix so that pressing ctrl-s a second time toggles off constrain height mode

jacob314:
Previously pressing ctrl-S twice would do nothing as constrain height mode would be automatically disabled then reenabled.


scottdensmore (approved):
Approved

PR #1319, opened 2025-06-23
Rename `disableDataCollection` to `usageStatisticsEnabled `

jerop:
Renames the `disableDataCollection` flag to the more intuitive and positive `usageStatisticsEnabled`.

1. Changing from `disable` to `enabled` improves code clarity by avoiding double negatives (`disableDataCollection: false`) and making the purpose of the flag more direct. The logic has been inverted wherever the flag is used to accommodate the new naming convention.

2. Using `enabled` as a suffix instead of a prefix follows a common convention that improves readability. 
- A condition like `if (usageStatisticsEnabled)` reads like a natural language sentence ("if usage statistics is enabled"), which reduces cognitive load.
- Distinguishes the boolean flag (representing a state) from potential functions that would perform an action (e.g., `enableUsageStatistics()` or `disableUsageStatistics()`), avoiding ambiguity between checking a value and calling a function.

3. Changing from "data collection" to "usage statistics" clarifies the scope and purpose of the data.

#750

### DIFF
--- a/packages/cli/src/config/config.ts
+++ b/packages/cli/src/config/config.ts
@@ -226,7 +226,8 @@ export async function loadCliConfig(
         process.env.OTEL_EXPORTER_OTLP_ENDPOINT ??
         settings.telemetry?.otlpEndpoint,
       logPrompts: argv.telemetryLogPrompts ?? settings.telemetry?.logPrompts,
-      disableDataCollection: settings.telemetry?.disableDataCollection ?? false,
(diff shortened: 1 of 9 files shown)


anj-s (approved):
I was thinking of doing something similar since double negatives actually confused me the first time around. Thanks!

owenofbrien (approved):
LGTM. Thanks for doing this!

mattKorwel (approved):

PR #1326, opened 2025-06-23
Update Auth Label to include AI Studio

philschmid:
# What does this PR do? 

It updates the Auth label from `Gemini API Key` to include AI Studio `Gemini API Key (AI Studio)`. For better understandbility.


cperry-goog (approved):
good call

PR #1336, opened 2025-06-23
feat(docs): add `/chat` and `/restore` commands

cperry-goog:


docs/cli/commands.md
```diff
@@ -10,6 +10,16 @@ Slash commands provide meta-level control over the CLI itself.
 
   - **Description:** File an issue about Gemini CLI. By default, the issue is filed within the GitHub repository for Gemini CLI. The string you enter after `/bug` will become the headline for the bug being filed. The default `/bug` behavior can be modified using the `bugCommand` setting in your `.gemini/settings.json` files.
 
+- **`/chat`**
+  - **Description:** Save and resume conversation history for branching conversation state interactively, or resuming a previous state from a later session.
+  - **Sub-commands:**
+    - **`save`**
+      - **Description:** Saves a checkpoint of the current conversation history. You can optionally add a `[tag]` to easily identify it later.
```
abhipatel12: nit: I feel like the usage of "checkpoint" could be confusing but also ack that there may not be a better word for it

docs/cli/commands.md
```diff
@@ -10,6 +10,16 @@ Slash commands provide meta-level control over the CLI itself.
...
+      - **Usage:** `/save [tag]`
```
sethtroisi: should be `/chat save [tag]`.

If we want to always require a tag we should probably use

`/chat save <tag>`

docs/cli/commands.md
```diff
@@ -10,6 +10,16 @@ Slash commands provide meta-level control over the CLI itself.
 
   - **Description:** File an issue about Gemini CLI. By default, the issue is filed within the GitHub repository for Gemini CLI. The string you enter after `/bug` will become the headline for the bug being filed. The default `/bug` behavior can be modified using the `bugCommand` setting in your `.gemini/settings.json` files.
 
+- **`/chat`**
+  - **Description:** Save and resume conversation history for branching conversation state interactively, or resuming a previous state from a later session.
+  - **Sub-commands:**
```
abhipatel12, line 16: Are we also adding `/chat list` or `/chat show` here to show saved tags?

docs/cli/commands.md
abhipatel12: nit: `/chat save [tag]`

docs/cli/commands.md
```diff
@@ -10,6 +10,16 @@ Slash commands provide meta-level control over the CLI itself.
...
+    - **`save`**
+      - **Description:** Saves a checkpoint of the current conversation history. You can optionally add a `[tag]` to easily identify it later.
+      - **Usage:** `/save [tag]`
+    - **`resume`**
+      - **Description:** Resumes a conversation from a previously saved checkpoint. If you don't provide a `[tag]`, it will attempt to load the latest checkpoint.
+      - **Usage:** `/resume [tag]`
```
abhipatel12: same as above: `/chat resume [tag]`

docs/cli/commands.md
cperry-goog: right

docs/cli/commands.md
cperry-goog: right

docs/cli/commands.md
cperry-goog: yes thanks

sethtroisi (approved):

PR #1337, opened 2025-06-23
Fix batch flush to Clearcut

jerop:
The `last_flush_time` in the Clearcut logger was being updated before the asynchronous flush operation completed, leading to incorrect flush timing and potential data delivery delays.

This change moves the timestamp update to after the flush promise resolves successfully, ensuring the flush time is only recorded after events are sent to Clearcut.


bdmorgan (approved):

owenofbrien (approved):
Good catch - thanks for fixing this!

anj-s (approved):

scottdensmore (approved):
Approved

PR #1340, opened 2025-06-23
Remove fallback to render normall rather than using custom MaxSizedBox layout logic

jacob314:
This resolves issues where regular ink Rendering had bugs improving the display of very long lines that wrap.

## TLDR

This change looks like it is inserting functionality but if you look at the diff carefully it is pretty much only deleting fallback code that wasn't needed.

This resolves an issue caught while testing the release where after pressing `ctrl-s` the following bad indentation of the content after the line number was shown
Screenshot 2025-06-23 at 12 19 09 PM
instead of line wrapping that
Screenshot 2025-06-23 at 12 13 40 PM

This change fixes the issue.


jacob314:
Be sure to review with the "hide whitespace" option checked to see a diff that looks like this:
Screenshot 2025-06-23 at 12 21 21 PM

mattKorwel (approved):

PR #1344, opened 2025-06-23
Ensure telemetry events are flushed immediately

jerop:
The previous implementation used `flushIfNeeded` to batch most telemetry events, but it was not reliably sending them, leading to data loss. Notably, the `startSession` event, which already used `flushToClearcut`, was working correctly, indicating an issue with the batching logic itself.

This change replaces all calls to `flushIfNeeded` with `flushToClearcut` to align all event logging with the working `startSession` implementation and ensure that events are sent immediately. This prioritizes the reliability of data collection over network efficiency.

This is a temporary solution to prevent further data loss. The underlying issue with the batching mechanism in `flushIfNeeded` should be investigated and fixed in the future, at which point this change can be reverted.

### DIFF
--- a/packages/core/src/core/nonInteractiveToolExecutor.test.ts
+++ b/packages/core/src/core/nonInteractiveToolExecutor.test.ts
@@ -19,6 +19,7 @@ import { Part, Type } from '@google/genai';
 const mockConfig = {
   getSessionId: () => 'test-session-id',
   getUsageStatisticsEnabled: () => true,
+  getDebugMode: () => false,
 } as unknown as Config;
 
 describe('executeToolCall', () => {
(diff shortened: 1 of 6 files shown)


mattKorwel (approved):

packages/core/src/core/nonInteractiveToolExecutor.test.ts
```diff
@@ -19,6 +19,7 @@ import { Part, Type } from '@google/genai';
 const mockConfig = {
   getSessionId: () => 'test-session-id',
   getDisableDataCollection: () => false,
+  getDebugMode: () => false,
```
anj-s, line 22: I see a bunch of debugMode changes to tests. Confirming that they were intended? Didn't see any change for this.

anj-s (approved):

packages/core/src/core/nonInteractiveToolExecutor.test.ts
jerop: It's because `flushToClearcut` calls `getDebugMode`

scottdensmore (approved):
approve

PR #1349, opened 2025-06-23
[June 25] Refine refresh static logic

SandyTao520:
## TLDR

Refresh static when terminal height changes, also remove unnecessary refresh after pressing ``ctrl + t`` or ``ctrl + o``.

## Dive Deeper

Reducing the height of terminal when a MaxSizedBox is visible causes render duplications, this is because previously we only listen to width decrease, it turns out this is not enough. Since we're limiting refresh to 1 time / 300ms, it's not too bad to 
trigger this on every resize event.

Another optimization is to not refreshStatic when opening a new panel (like the tools descriptions, or the debug console).

## Reviewer Test Plan

1. Open debug console (``ctrl + o``) and trigger show tools (``ctrl + t``) to check they show correctly.
2. Ask GC to use WriteFile to make a large edit (write a.txt with 50 lines from 1 to 50), once the proposed edit is shown, resize terminal to check if there are stale drawings.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅ | ❓  | ❓  |
| npx      | ✅  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ✅  | -   | -   |

## Linked issues / bugs
N/A


packages/cli/src/ui/App.tsx
```diff
@@ -475,16 +472,13 @@ const App = ({ config, settings, startupWarnings = [] }: AppProps) => {
 
     // debounce so it doesn't fire up too often during resize
     const handler = setTimeout(() => {
-      if (terminalWidth < lastTerminalWidth.current) {
-        setStaticNeedsRefresh(true);
-      }
-      lastTerminalWidth.current = terminalWidth;
+      refreshStatic();
```
jacob314, line 476: should we also make sure that 
setStaticNeedsRefresh gets set to false here so that we don't trigger an unneeeded refresh at a later point because we have already refreshed?

jacob314 (approved):
lgtm

jacob314 (approved):

scottdensmore (approved):
Approved

PR #1354, opened 2025-06-23
Include all chat messages

sethtroisi:
`/resume` suppressed displaying the system prompt we send to the GeminiChat class. The way those message were suppressed resulted in the message not being including it in future `/save` which caused the first two messages to be lost when those conversations where `/resume`d 

This makes everything a bit more robust. This was discovere while working on the consolidation to `/chat`


scottdensmore (approved):
Approved

PR #1356, opened 2025-06-23
Polish Theme Dialog

jacob314:
This fixes https://github.com/google-gemini/gemini-cli/issues/1347 for all reason terminal widths and heights.

* The theme dialog now displays fully on screen at the default Mac terminal size and slightly smaller screens. This important for the first run experience where the theme dialog is shown.

Video:
https://screencast.googleplex.com/cast/NjYzMjE4NDYwODcxODg0OHw4MGFkZTFiNC1mZQ

Before:
Screenshot 2025-06-23 at 4 07 36 PM

After:
Screenshot 2025-06-23 at 4 06 47 PM

 
This required making the theme dialog UI adaptive as we have a lot of themes. 
The theme dialog now removes the 1 line of vertical padding above and bellow the dialog when there isn't enough space.
The theme dialog now hides the "Apply to" section of the dialog automatically when there isn't enough space.
A bunch of tweaking to the wrapping logic including ensuring all content is wrappable so the sizes are predictable.

Fixed a long standing issue where the space in a radio button between the selection indicator and the text would be removed if the radio button had to have its text truncated.
This issue occurs at wider screen widths but the narrow screen width makes it very obvious
Before:
Screenshot 2025-06-23 at 2 28 20 PM

After:
Screenshot 2025-06-23 at 2 27 47 PM

Tossed in minor polish so that the Gemini Logo renders without extra line breaks (the terminal may still add extra line breaks).
This makes rendering better when the Gemini CLI terminal goes to a narrow width to a wider width.

### DIFF
--- a/packages/cli/src/ui/components/Header.tsx
+++ b/packages/cli/src/ui/components/Header.tsx
@@ -30,17 +30,22 @@ export const Header: React.FC<HeaderProps> = ({
       terminalWidth >= widthOfLongLogo ? longAsciiLogo : shortAsciiLogo;
   }
 
+  const artWidth = getAsciiArtWidth(displayTitle);
+
   return (
-    <>
-      <Box marginBottom={1} alignItems="flex-start">
-        {Colors.GradientColors ? (
-          <Gradient colors={Colors.GradientColors}>
-            <Text>{displayTitle}</Text>
-          </Gradient>
-        ) : (
+    <Box
+      marginBottom={1}
+      alignItems="flex-start"
+      width={artWidth}
+      flexShrink={0}
+    >
+      {Colors.GradientColors ? (
+        <Gradient colors={Colors.GradientColors}>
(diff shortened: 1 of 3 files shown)


packages/cli/src/ui/components/Header.tsx
```diff
@@ -30,9 +30,16 @@ export const Header: React.FC<HeaderProps> = ({
       terminalWidth >= widthOfLongLogo ? longAsciiLogo : shortAsciiLogo;
   }
 
+  const artWidth = getAsciiArtWidth(displayTitle);
+
   return (
     <>
-      <Box marginBottom={1} alignItems="flex-start">
+      <Box
```
jacob314: this keeps Ink from adding extra line breaks in the Gemini Logo

allenhutchison (approved):

PR #1357, opened 2025-06-23
Fix duplicate startSession logs and duplicate logging events over the wire

anj-s:
## TLDR

Duplicate start session logs were showing up in addition to event logs not being cleared after being sent. 

## Dive Deeper

Two issues were identified:
1. we were initializing clearcut logging in config.ts and sdk.ts. We should skip sdk.ts cause that is part of telemetry initialization and does not depend on the usageStatisticsEnabled flag.
2. The events queue was being cleared only after returning from the server. This meant if events were sent in quick succession they would be duplicates of them in the event queue. Created a temporary event queue for enabling us to send and restore events based on the success/error state of the response. 

## Reviewer Test Plan

* Start the cli in debug mode and see clearcut logs being streamed to the console.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | Y  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

### DIFF
--- a/packages/cli/src/config/config.ts
+++ b/packages/cli/src/config/config.ts
@@ -169,7 +169,7 @@ export async function loadCliConfig(
   loadEnvironment();
 
   const argv = await parseArguments();
-  const debugMode = argv.debug || false;
+  const debugMode = argv.debug || true;
 
   // Set the context filename in the server's memoryTool module BEFORE loading memory
   // TODO(b/343434939): This is a bit of a hack. The contextFileName should ideally be passed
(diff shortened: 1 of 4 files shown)


packages/cli/src/config/config.ts
```diff
@@ -169,7 +169,7 @@ export async function loadCliConfig(
   loadEnvironment();
 
   const argv = await parseArguments();
-  const debugMode = argv.debug || false;
+  const debugMode = argv.debug || true;
```
owenofbrien: Was this intentional?

scottdensmore (approved):
Approved

owenofbrien (approved):
Thanks for fixing this!

packages/cli/src/config/config.ts
anj-s: Definitely not :) I forgot to save the file that reverted this.. fixed now.

PR #1365, opened 2025-06-24
Added ToS links for each surface

markmcd:
## TLDR

Added missing ToS.

## Dive Deeper

The terms linked only applied to users accessing via the Gemini API directly. There are other mechanisms so include them.


scottdensmore (approved):
Approve

PR #1373, opened 2025-06-24
Fix remaining path parameter naming inconsistencies

keir:
## TLDR

#1057 renamed ReadFile parameters to get the model to more frequently use an absolute path. This PR finishes the job with the other commands.

## Testing

Tested each command and had the model recite its tool definitions.

### DIFF
--- a/docs/tools/file-system.md
+++ b/docs/tools/file-system.md
@@ -11,7 +11,7 @@ All file system tools operate within a `rootDirectory` (usually the current work
 - **File:** `ls.ts`
 - **Description:** Lists the names of files and subdirectories directly within a specified directory path. It can optionally ignore entries matching provided glob patterns.
 - **Parameters:**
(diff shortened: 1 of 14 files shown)


mattKorwel (changes requested):
@keir i belive you'll need to rebase this PR against main? is this intended for the release, or after?

mattKorwel:
SORRY, you don't need to rebase, i clicked the wrong button. i see this is targeting what is now 'backfill-post-june25' branch.

NTaylorMullen:
After getting in the last change things have worked BUT I simultaneously realized that our examples were encouraging non-absolute path behavior. I actually think we may be able to revert the absolute_path naming and instead work on our system prompts / descriptions to consistently mention absolute pathing as well. We should get a higher quality output that way

NTaylorMullen:
Closing for now given the time dialation + several other efforts around this. ❤️ Keir!

PR #1376, opened 2025-06-24
[JUNE 25] Permanent failover to Flash model for OAuth users after persistent 429 errors

bdmorgan:
**NOTE**: This was added as a failsafe mechanism in case early launch days user load is such to create a large number of 429 errors. If a Code Assist user receives three consecutive 429 errors as part of the backoff/retry mechanism, they will automatically be switched over to use the Flash model for the remainder of the session. This actually improves the UX but also presents users from having the appearance of a broken system.

Added a permanent failover mechanism for OAuth users that switches to Flash model after persistent 429 errors, with user confirmation and session persistence.   

Test by setting environment variable GEMINI_CLI_SIMULATE_429=true

How It Works

  1. OAuth users only (LOGIN_WITH_GOOGLE_PERSONAL/ENTERPRISE) experience this fallback
  2. After persistent 429 errors through the retry backoff process
  3. User is prompted with a dialog asking to switch to Flash model
  4. If accepted, the model switches to gemini-2.5-flash for the rest of the session
  5. UI immediately updates to show the new model in the footer
  6. Request is retried with the new model

  Testing

  Environment Variables for Testing:
  export GEMINI_CLI_SIMULATE_429=true
  export SIMULATE_429_FOR_AUTH_TYPE=oauth-personal
  export DEBUG_FALLBACK_FLOW=true

  Unit Tests:
  - Run npm test to execute comprehensive test suite for fallback logic

1. Automatic Flash Fallback System

  - Trigger: 3 consecutive 429 errors for OAuth users only (LOGIN_WITH_GOOGLE_PERSONAL and LOGIN_WITH_GOOGLE_ENTERPRISE)
  - Behavior: Automatically switches from current model to gemini-2.5-flash without user confirmation
  - UI Notification: Displays clear message in terminal interface:
  ⚡ Rate limiting detected. Automatically switching from gemini-2.5-pro to gemini-2.5-flash for faster responses for the remainder of this session.

  2. Model Reset on Auth Changes

  - Problem Solved: Users switching from OAuth to API Key via /auth were staying on Flash model
  - Solution: Added resetModelToDefault() method that resets to original default model when auth method changes
  - Behavior: Any auth method switch via /auth command automatically resets to gemini-2.5-pro

  3. Testing Infrastructure

  - Environment Variables:
    - GEMINI_CLI_SIMULATE_429=true - Enable 429 simulation
    - SIMULATE_429_AFTER_N_REQUESTS=N - Trigger after N requests
    - SIMULATE_429_FOR_AUTH_TYPE=oauth-personal - Target specific auth types
    - DEBUG_FALLBACK_FLOW=true - Enable debug logging
  - Smart Simulation: Automatically disables after successful fallback to prevent infinite loops


packages/core/src/config/config.ts
```diff
@@ -216,16 +227,40 @@ export class Config {
...
   async refreshAuth(authMethod: AuthType) {
+    // Check if this is actually a switch to a different auth method
+    const previousAuthType = this.contentGeneratorConfig?.authType;
+    const isAuthMethodSwitch =
+      previousAuthType && previousAuthType !== authMethod;
+
+    // Always use the original default model when switching auth methods
+    // This ensures users don't stay on Flash after switching between auth types
+    // and allows API key users to get proper fallback behavior from getEffectiveModel
+    const modelToUse = this.model; // Use the original default model
```
abhipatel12, line 237: nit: probably can remove `// Use the original default model`

bdmorgan:
Updated this to remove the environment variables:

  1. Simplified testUtils.ts:
    - Removed all environment variable initialization
    - Kept programmatic functions for unit testing
    - Removed debug logging that relied on env vars
  2. Cleaned Production Code:
    - Removed shouldSimulate429() and createSimulated429Error() calls from retry.ts
    - Removed disableSimulationAfterFallback() calls from client.ts and geminiChat.ts
    - Removed resetSimulationState() call from config.ts
  3. Enhanced Integration Tests:
    - Created comprehensive test in flashFallback.integration.test.ts that verifies:
        - OAuth users get Flash fallback after 3 consecutive 429 errors
      - API key users do NOT get Flash fallback (they use the startup fallback instead)
      - Fallback handler is called with correct parameters
      - Simulation state management works properly

scottdensmore (approved):
Approve

Issue #1385, opened 2025-06-24
MCP Server Tool call responses should show preview and be collapsed by default


jackwotherspoon:
**Describe the bug**
When an MCP server tool is called, the raw response seems to be outputted into the CLI.

For extremely large tool responses this spams the CLI output with an extremely large text, causing the user to have to scroll up for ages...

The CLI should preview some of the response and collapse the rest , allowing a user to click on a small arrow or similar icon to view the full payload if they so choose.

Below is an example payload response for GitHub MCP server `list_issues` on a repo with a decent sized number of open issues https://github.com/a2aproject/A2A. This would be even worse with a larger repo...

https://github.com/user-attachments/assets/1bea4488-cade-468b-a3b5-3bde9ac68ad0

This will vastly improve the user experience 😄

jackwotherspoon:
Can we close this based on #1388 being merged @SandyTao520 ?

SandyTao520:
https://github.com/google-gemini/gemini-cli/pull/1388 collapsed tool responses by default but currently we're lacking a way to expand them. Feel free to close this and file another feature request to track this.

ptone:
marking as fixed by #1388 for now @jackwotherspoon to open additional issue for expansion if needed

PR #1388, opened 2025-06-24
Do not render mcp responses as markdown

SandyTao520:
## TLDR

Fallback to render large tool call responses in plain text instead of markdown.

## Dive Deeper

Markdown display inside a ToolMessage doesn't respect availableTerminalHeight, causing flickering when streaming large contents. The implementation of  MardownDisplay is complicated, the safest way to avoid flickering before launch is to render as plaintext, which is height-constrained by a MaxSizedBox.

## Reviewer Test Plan

Connect to an MCP server to stream large content. Verify the content is contained within a terminal-fitting box with the top lines hidden.

(A simple example is use Github MCP server to list commits in a public repo)

Also verify short markdown still gets displayed properly as markdown.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ✅  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ✅  | -   | -   |

## Linked issues / bugs

N/A

### DIFF
--- a/packages/cli/src/ui/components/messages/ToolMessage.tsx
+++ b/packages/cli/src/ui/components/messages/ToolMessage.tsx
@@ -44,6 +44,9 @@ export const ToolMessage: React.FC<ToolMessageProps> = ({
       )
     : undefined;
 
+  // Long tool call response in MarkdownDisplay doesn't repect availableTerminalHeight properly,
+  // we're forcing it to not render as markdown when the response is too long, it will fallback
+  // to render as plain text, which is contained within the terminal using MaxSizedBox
   if (availableHeight) {
     renderOutputAsMarkdown = false;
   }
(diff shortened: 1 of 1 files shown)


scidomino (dismissed):

packages/cli/src/ui/components/messages/ToolMessage.tsx
```diff
@@ -44,6 +44,11 @@ export const ToolMessage: React.FC<ToolMessageProps> = ({
       )
     : undefined;
 
+  // TODO: Fix markdown display to respect max height
```
scottdensmore: @SandyTao520 can you file an issue for this and remove the TODO

SandyTao520: Someone already did! https://github.com/google-gemini/gemini-cli/issues/1385

jackwotherspoon: Glad this is getting fixed already 👏  Thanks!

jacob314 (approved):
lgtm
This is the right mitigation to unblock for launch. We'll fast follow with fixes to the markdown rendering.

scottdensmore (approved):
Approved

PR #1389, opened 2025-06-24
Remove uses of the spread operator that appear to have caused a maximum call stack size exceeded error

jacob314:
I would have thought this was just a theoretical stack overflow issue but we have this user report so it appears some users are generating outputs large enough to hit this issue.

https://b.corp.google.com/issues/427533822


scottdensmore (approved):
Approved

PR #1390, opened 2025-06-24
sandbox doc

emeryray2002:
## TLDR

Adds sandboxing documentation.

## Dive Deeper

Adds sandboxing documentation, to include: prerequisites, quickstart configuration, and common troubleshooting issues.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | x  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


docs/sandbox.md
```diff
@@ -0,0 +1,149 @@
...
+Before using sandboxing, you need to install and set up Gemini CLI:
+
+```bash
+# Clone and install
+git clone https://github.com/google-gemini/gemini-cli.git
+cd gemini-cli
+npm install
+
+# Link the gemini command globally
+npm link
```
mattKorwel: This is valid. however this is centric to this repo and people developing against it. Our users will use `npm install -g @google/gemini-cli` and i think that's what we should identify here. Or both if we feel strongly. 

```suggestion
```bash
#install gemini-cli with npm
npm install -g @google/gemini-cli
```

docs/sandbox.md
```diff
@@ -0,0 +1,149 @@
...
+- `permissive-closed` - Write restrictions, no network
+- `permissive-proxied` - Write restrictions, network via proxy
+- `restrictive-open` - Strict restrictions, network allowed
+- `restrictive-closed` - Maximum restrictions
+
+### Container-based Configuration
+
+#### Building the Sandbox Image
+
+First time setup:
```
mattKorwel: Same comment here as above. As an end user guide, this should clarify that building is only needed for local development if you want to CHANGE the sandbox image.

docs/sandbox.md
mattKorwel, line 1: There is an additional step where users can supply `SANDBOX_IMAGE` env variable and point to their own image. we should capture that here if we are making the updates.

mattKorwel (approved):

PR #1394, opened 2025-06-24
fix: use correct directory for update checks

mboshernitsan:
The update notifier was incorrectly checking for updates to the `@gemini-cli/cli` package, and it was relying on the `package.json` of the user`s current working directory, which is not reliable for a globally installed tool.

This change modifies the build process to inject the top-level package name and version into the bundle. The update checker now uses these build-time variables, ensuring it always checks for the correct package (`gemini-cli`) and its version.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | x  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

### DIFF
--- a/esbuild.config.js
+++ b/esbuild.config.js
@@ -22,6 +22,7 @@ esbuild
     platform: 'node',
     format: 'esm',
     define: {
+      'process.env.CLI_NAME': JSON.stringify(pkg.name),
       'process.env.CLI_VERSION': JSON.stringify(pkg.version),
     },
     banner: {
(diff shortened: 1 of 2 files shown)


scidomino (approved):

mattKorwel (approved):
@mara

PR #1395, opened 2025-06-24
Truncate all strings before displaying in a tool messages to avoid stack overflows

jacob314:
Add lightweight string truncation on top of the heavier weight overflow wrapping logic to prevent extremely large strings from crashing the UI rendering content that the user will never see.

The following issue was due to Gemini CLI attempting to render a string that was so large that it took a long time just to paste it to a file.
https://b.corp.google.com/issues/427533822

### DIFF
--- a/packages/cli/src/ui/components/messages/ToolMessage.tsx
+++ b/packages/cli/src/ui/components/messages/ToolMessage.tsx
@@ -18,6 +18,9 @@ const RESERVED_LINE_COUNT = 5; // for tool name, status, padding etc.
 const STATUS_INDICATOR_WIDTH = 3;
 const MIN_LINES_SHOWN = 2; // show at least this many lines
 
+// Large threshold to ensure we don't cause performance issues for very large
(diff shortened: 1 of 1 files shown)


mattKorwel:
@jacob314 do you have a way i can exercise this? Are we sure there are no downstream effects, at least in happy path?

scottdensmore (approved):
Approved

jacob314:
> @jacob314 do you have a way i can exercise this? Are we sure there are no downstream effects, at least in happy path?
No downstream effects on the happy path which is that tool respones are all way less than 1MB.
On the unhappy path, the impact is you will see a slightly inaccurate # for the # of lines truncated as we truncated before going to that logic due to the extra content size. This will not impact Gemini behavior at all because I purely mess with things at the UI layer.

jacob314:
Easiest way to test is to generate a shell command that intentionally outputs more than 1 million characters. See if you can tell that the # of lines truncated is somewhat off.

PR #1399, opened 2025-06-24
Fix lint error for usMemo that could result in a stale version of savedChatTags being executed.

jacob314:
Probably not release blocking but isn't great to have the main branch have lint issues and this one typically does indicate a hard to diagnose logic bug.


mattKorwel:
@sethtroisi made the same change. closing this pr. feel free to reopen if i have it wrong @jacob314

PR #1402, opened 2025-06-24
[June 25] handle early output pipe closer

keithballinger:
handle a pipe closing early in headless mode


scottdensmore (approved):
approved

PR #1407, opened 2025-06-25
fix: define es module __dirname in updateCheck.ts

KeijiBranshi:
this is a global in some JS variants (CommonJS?), but not in ES modules. So it needs to be explicitly defined


mattKorwel (approved):

KeijiBranshi:
#1406

PR #1413, opened 2025-06-25
CI: Linting Fix

mattKorwel:
## TLDR

Our ci & preflight should fail when there are errors or warnings.

## Dive Deeper

These warning for linting, are actual errors for typescript so we need to catch them.


scottdensmore (approved):
Approved

PR #1414, opened 2025-06-25
Add more staticRefresh hacks to avoid rendering artifacts when showing and hiding the debug console.

jacob314:
## TLDR

This will be obsolete with Ink 6.0 but fixing for now as the artifacts without it are annoying

https://screencast.googleplex.com/cast/NDkwMzQxMzMwODQ1Njk2MHxlZDM3ZmE5OS1lZA


NTaylorMullen:
@jacob314 still needed?

PR #1415, opened 2025-06-25
Stop matching numbers in error messages if it is not required.

jacob314:
For discussion. Need to verify whether we have errors that lack a status field. As @bbiggs pointed out, the regex matches for 3 digit numbers aren't safe as they false positive on errors for the token limit being exceeded as those messages include the token count and the typical 7 digit number often has a 5 followed by a 2 digits. If we need to include status, we should exactly match message text that indicates the number is an error code.


jacob314:
Ok get why it was needed. Expect we need to parse the json from the message to make this robust.

PR #1423, opened 2025-06-25
Update Readme to call out gemini command after npm install

davenportjw:
fix npm install directions

## TLDR

Fixes the readme so after npm install there is the command to start gemini.

## Dive Deeper

The TLDR is the thing.

## Reviewer Test Plan

They should have a fresh install of this and make sure that the two steps are needed.

## Testing Matrix

Not applicable.

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,13 +23,14 @@ With the Gemini CLI you can:
 2. **Run the CLI:** Execute the following command in your terminal:
 
    ```bash
-   npx https://github.com/google-gemini/gemini-cli
+   $ npx https://github.com/google-gemini/gemini-cli
    ```
 
    Or install it with:
 
    ```bash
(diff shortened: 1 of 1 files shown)


markmcd (changes requested):
These commands are no longer copy-pastable if you add a leading `$`

davenportjw:
> These commands are no longer copy-pastable if you add a leading `$`

We should strike them from the example then too. I can update.

markmcd (approved):

PR #1461, opened 2025-06-25
Prod: Integration Tests Main Only

mattKorwel:
## TLDR

Make integration tests not run or required for pull requests.

## Dive Deeper

Our e2e tests are currently flaky and experiencing permissions issues for forked users. This is a temp change to only run on pushes to main. Follow up work will be needed to ensure these work long term.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -4,9 +4,7 @@ name: E2E Tests
 
 on:
   push:
-    branches: [main, release]
-  pull_request:
-    branches: [main, release]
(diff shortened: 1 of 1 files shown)


mattKorwel:
I think there is a missing secret exposing in the integeration tests. this wasn't an issue before as users were part of the repo even when forking.  https://docs.github.com/en/actions/security-for-github-actions/security-guides/using-secrets-in-github-actions#using-secrets-in-a-workflow

scottdensmore (approved):
Approved

PR #1479, opened 2025-06-25
fix: proxy settings

itechbear:
## TLDR
This PR fixes http proxy settings.

## Dive Deeper

In the constructor of GeminiClient, the http proxy is set:
https://github.com/google-gemini/gemini-cli/blob/21cfe9f6801f286dda6d51d2886e27bd67bd5fa4/packages/core/src/core/client.ts#L61

Moving the instance creation of `GeminiClient` before `await createContentGeneratorConfig()` would help setup the proxy before any http requests are made.

### DIFF
--- a/packages/core/src/config/config.ts
+++ b/packages/core/src/config/config.ts
@@ -240,13 +240,13 @@ export class Config {
     // the previous session's model (which might be Flash)
     this.contentGeneratorConfig = undefined!;
 
+    const gc = new GeminiClient(this);
     const contentConfig = await createContentGeneratorConfig(
       modelToUse,
       authMethod,
       this,
     );
 
-    const gc = new GeminiClient(this);
     this.geminiClient = gc;
     this.toolRegistry = await createToolRegistry(this);
     await gc.initialize(contentConfig);


Xtamaporn:
Sexygirls

NTaylorMullen:
@itechbear I'm not quite sure this is doing what you think. When the content generator is created it can't see the local client so any changes there wouldn't be respected either way. We then also initialize the client a few lines down which is dependent on the content generator being created.

Going to close this out for now but feel free to re-open if I'm missing something!

itechbear:
Currently, when `await createContentGeneratorConfig()` is called, it has a side effect to update the model. 

The call stack is

https://github.com/google-gemini/gemini-cli/blob/a943cee45d9e6088a7e84acc9e6d98e8d4f504b1/packages/core/src/config/config.ts#L243

https://github.com/google-gemini/gemini-cli/blob/a943cee45d9e6088a7e84acc9e6d98e8d4f504b1/packages/core/src/core/contentGenerator.ts#L92

https://github.com/google-gemini/gemini-cli/blob/a943cee45d9e6088a7e84acc9e6d98e8d4f504b1/packages/core/src/core/modelCheck.ts#L46

So it would actually make an HTTP request to `https://generativelanguage.googleapis.com/v1beta/models/${modelToTest}:generateContent?key=${apiKey}`

However, only after the instance of `GeminiClient` is created, the proxy is setup. So the above request would pass through the proxy settings without this PR.

itechbear:
It seems I don't have any permission to re-open this issue.

Issue #1511, opened 2025-06-25
Option + Left / Right Arrow Keys


guidedways:
On macOS you can Option + left / right arrow keys to quick navigate and jump between words. You can do this in other CLIs but not in Gemini CLI. Would be nice to have so it's easier to go navigate between words.

nethushadev:
i would love this too. its harder to navigate otherwise

PR #1523, opened 2025-06-25
Support more option keyboard shortcuts

ryanchentw:
## TLDR

Support option based keyboard shortcuts:
- option+left
- option+right

This solves the #1511

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ✅  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


jacob314:
Thanks for the pull request @ryanchentw! I'll review as soon as you sign the CLA. Would be nice to have these shortcuts!
One request. Can you make the changes to text-buffer rather than InputPrompt. That will also make it easy to write some tests. Expect Gemini CLI will be able to write some test cases for a change like this.

packages/cli/src/ui/components/InputPrompt.tsx
```diff
@@ -348,6 +348,21 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
         return;
       }
 
+      if (key.meta && key.delete) {
+        buffer.deleteWordLeft();
+        return;
+      }
+
+      if (key.meta && key.leftArrow) {
```
jacob314: move the leftArrow and rightArrow support to textBuffer itself with the other keyboard shortcuts that aren't specific to the input prompt itself.

packages/cli/src/ui/components/InputPrompt.tsx
```diff
@@ -348,6 +348,21 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
         return;
       }
 
+      if (key.meta && key.delete) {
```
jacob314: I'm not sure we should add is keyboard shortcut. if we do, we should probably add it in combination with supporting backspace as separate from delete. In vscode `alt backspace` deletes a word right and `alt delete` deletes a word right. @dewitt may also have some thoughts as he as done more work than I have on exactly what keyboard shortcuts we should and shouldn't include.

jacob314 (commented):
thanks for the pr. will be definitely great to have meta right and meta left working.

packages/cli/src/ui/components/InputPrompt.tsx
ryanchentw: Actually it works like `alt backspace` right now. That's pretty weird. I believe there're some re-mapping keys in somewhere but I can't figure out currently. So let me remove this section. And try to focus on `alt+{left,right}` first

jacob314: Sounds good. Just realized thst @bbiggs has a change that needs to get merged into the main branch that correctly handles delete and backspace as separate keys. Once that lands you should be unblocked for the alt delete work.

packages/cli/src/ui/components/shared/text-buffer.test.ts
```diff
@@ -677,6 +677,54 @@ describe('useTextBuffer', () => {
       expect(getBufferState(result).cursor).toEqual([0, 2]);
     });
 
+    it('should handle Meta+leftArrow for moving word left', () => {
```
jacob314: Can you tweak the tests so they ate parameterized to test all of the key combinations now mapped to moving a word left and right?

jacob314 (commented):
Approved after a minor nit about the tests

packages/cli/src/ui/components/shared/text-buffer.test.ts
ryanchentw: I've already updated the test with parameterization. Please take a look. Thanks for the suggestion!

jacob314 (approved):
Thanks fr the contribution! I will let you know when the delete and backspace support is in to unblock landing word delete support.

NTaylorMullen:
@ryanchentw would you mind rebasing on latest? Sorry for the conflicts. There's been a good amount of churnin this area 😄

ryanchentw:
Hi @NTaylorMullen looks like the issue already fix on #1972, I think we don't need this PR anymore 🤣

NTaylorMullen:
> Hi @NTaylorMullen looks like the issue already fix on #1972, I think we don't need this PR anymore 🤣

Thank you for the follow up @ryanchentw !

PR #1524, opened 2025-06-25
feat(cli): Add hideTips setting

marcinjahn:
## TLDR

Adds a configuration option `hideTips` that allows to disable the display of tips in the CLI UI.

The UI becomes a bit less cluttered:
image

## Dive Deeper

The tips displayed in the CLI are useful for first time users, but after time they most likely will not be that desired by some. This PR adds a way to disable tips for those that don't need to see them.

## Reviewer Test Plan

- set `hideTips` to `true` in your settings file and see if tips do disappear.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | X  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

- #1525

### DIFF
--- a/docs/cli/configuration.md
+++ b/docs/cli/configuration.md
@@ -65,14 +65,17 @@ In addition to a project settings file, a project's `.gemini` directory can cont
     ```
 
 - **`coreTools`** (array of strings):
-  - **Description:** Allows you to specify a list of core tool names that should be made available to the model. This can be used to restrict the set of built-in tools. See [Built-in Tools](../core/tools-api.md#built-in-tools) for a list of core tools.
+  - **Description:** Allows you to specify a list of core tool names that should be made available to the model. This can be used to restrict the set of built-in tools. See [Built-in Tools](../core/tools-api.md#built-in-tools) for a list of core tools. You can also specify command-specific restrictions for tools that support it, like the `ShellTool`. For example, `"coreTools": ["ShellTool(ls -l)"]` will only allow the `ls -l` command to be executed.
   - **Default:** All tools available for use by the Gemini model.
-  - **Example:** `"coreTools": ["ReadFileTool", "GlobTool", "SearchText"]`.
+  - **Example:** `"coreTools": ["ReadFileTool", "GlobTool", "ShellTool(ls)"]`.
 
 - **`excludeTools`** (array of strings):
-  - **Description:** Allows you to specify a list of core tool names that should be excluded from the model. A tool listed in both `excludeTools` and `coreTools` is excluded.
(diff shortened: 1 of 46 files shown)


packages/cli/src/config/config.ts
```diff
@@ -245,6 +245,7 @@ export async function loadCliConfig(
     bugCommand: settings.bugCommand,
     model: argv.model!,
     extensionContextFilePaths,
+    hideTips: settings.hideTips ?? false,
```
jacob314: is this really needed in config or is just settings sufficient?

marcinjahn: Technically, settings is sufficient, and that’s how I initially did it. However, I later noticed that there is this „config” abstraction which kind of pulls all configuration sources together and it seemed like I should be using that. If not, I can go back to the (simpler) settings based approach, just let me know

jacob314: Lets just go with settings. You can always add it more places later.  Generally I would only put it in settings if there isnt a compelling reason why it should be settable on the command line.

marcinjahn: I pushed https://github.com/google-gemini/gemini-cli/pull/1524/commits/bc89becfda5e19cd1c00d67c14a0dfd8936f1594. Have a look

docs/cli/configuration.md
```diff
@@ -179,13 +179,24 @@ In addition to a project settings file, a project's `.gemini` directory can cont
     }
     ```
 - **`usageStatisticsEnabled`** (boolean):
+
```
jacob314: Can you leave the whitespace alone for the other entries in this file or make them all consistent? Seems like the file still has a mix of the two options.

jacob314 (dismissed):
Thanks for the pull request. Looks good now!

docs/cli/configuration.md
marcinjahn: Removed the newline here. Leaving it mixed as it is, I guess some formatting for MDs can be added later to the repo to make it all consistent.

NTaylorMullen (dismissed):

NTaylorMullen:
@marcinjahn looks like there's still a few formatting issues. See: https://github.com/google-gemini/gemini-cli/actions/runs/15953222049/job/44996060414?pr=1524

you can resolve these by running `npm run format`

allenhutchison (approved):
Redoing @NTaylorMullen approval, and merging.

swissspidy:
Could someone close #1524 accordingly?

PR #1534, opened 2025-06-25
Fixed Checkpointing docs for enabling checkpointing using settings.json

arjunlall:
## TLDR

The previous docs had incorrect details for enabling checkpointing via the settings.json file. This now has the correct version.

## Dive Deeper

The current docks wrap the needed setting in a "features" object, but instead it just needs to be

`"checkpointing": { "enabled": true }`

at the root level.


allenhutchison (approved):
Thanks for the change.

Issue #1531, opened 2025-06-25
[TOOL LOOP] CLI getting stuck in tool loop 


akwin1234:
## TL;DR

By @jackwotherspoon 

Summarizing duplicate and common issues:
- Stuck due to tool issues (for example `write_tool` not working) so will continuously loop
- Potentially getting stuck due to switching back and forth between Pro and Flash models?

Main fix seems to be to look at improving tool usage 👍 

by @silviojr 

I am making this issue the Master issue for tool loops. I will close the other issues that are referring to tool loops as well, but they should also be checked for more details and examples of what loops can look like. Those loops are coming mostly from bad tool usage: wrong parameters, tool failures etc 

The new tasks created under this should be focused on mitigating this issue. 

Tool loop replace due to tool failure: https://github.com/google-gemini/gemini-cli/issues/13613 , https://github.com/google-gemini/gemini-cli/issues/13676 
Tool loop writeFile: https://github.com/google-gemini/gemini-cli/issues/13704 
Tool loop shell: https://github.com/google-gemini/gemini-cli/issues/13147 , https://github.com/google-gemini/gemini-cli/issues/12871 , 
Tool searchText: https://github.com/google-gemini/gemini-cli/issues/13092 
Mispelled parameter (mcp tool): https://github.com/google-gemini/gemini-cli/issues/9971 

## Bug Description

![Image](https://github.com/user-attachments/assets/af1e2da0-ae10-4b58-a736-6800a5becac2)

it gives up on doing small tasks after a while. and gets stuck in a loop. takes infinite time

MarkEdmondson1234:
I had a siutation where it was stuck in the same solutions in a loop that didn't work, I even stopped it and said it was in a loop which it acknowledged but then went back to the same.  It seems like a too aggressive cache on input/outputs? It kept re-trying the same, and cost me a few million tokens for the trouble.


```
I've fixed the error in src/components/markdown/Artifact.tsx. Now I'll move on to the
  next file with errors, src/components/markdown/ArtifactPre.tsx. I'll read the file to
  understand the context of the useCallback error.
 ╭────────────────────────────────────────────────────────────────────────────────────────╮
 │ ✔  ReadFile src/markdown/ArtifactPre.tsx                                               │
 ╰────────────────────────────────────────────────────────────────────────────────────────╯
✦ I've confirmed that src/components/markdown/ArtifactPre.tsx has the correct useCallback
[...]

Issue #1484, opened 2025-06-25
Got into chanting loop after editing file


wittrock:
## Describe the bug
I asked gemini to fix a test for me in the Fuchsia repository. It made the correct changes, then reverted them, made them again, and then got into a chanting loop about marking its task complete. 

Transcript here: https://gist.github.com/wittrock/59aaded8cf584c716ae6f1b7149ad100#file-gistfile1-txt

## Additional context
  This is in the Fuchsia respository, at fuchsia.googlesource.com/fuchsia

## Diagnostic Information
*   **CLI Version:** 0.1.0-52afcb3.0-release
*   **Git Commit:** 52afcb3
*   **Operating System:** linux v22.16.0
*   **Sandbox Environment:** no sandbox
*   **Model Version:** gemini-2.5-pro
*   **Memory Usage:** 999.5 MB

scidomino:
This is a problem we've been combatting for a while now. We made tweaks to the prompt that have reduced the problem but it still happens occasionally.

ruchernchong:
Yeah, the CLI went into a loop by running the tests, linter and git status for me

PR #1546, opened 2025-06-25
Create GeminiChatHistory object which reduces MemoryRot behavior

mainroach:
## TLDR

This PR abstracts the GeminiChat::History object into a new class, which better manages the history into generational segmentation in order to reduce Memory Rot behaviors.

## Dive Deeper

Gemini-CLI has a Mono-Agent structure - one Agentic loop is responsible for all SWE tasks with access to on-disk installed tools (“SWEAgent”). We observe that for tasks which require a medium-to-large amount of autonomous execution, the quality of response of the MonoAgent decreases. This behavior pattern often starts occurring after the size of contextual memory eclipses a non-specific threshold, and increases in quantity as the context memory continues to grow.

Currently, the Gemini-CLI experience produces less than ideal customer experience - the LLM will often run off into the weeds, start making unrelated edits, remove code it needs, only to re-add it later, incorrectly, etc etc etc. Many of these errors relate to long-running history problems, and the user can address it by closing the application and restarting. This reboot-loop-for-quality is not an ideal user experience, and could be improved by addressing memory rot issues in the code.

For each subtask, we observe a clear trend: LLM accuracy declines as the number of required operations increases. The pattern is intuitive: at low operation counts, accuracy remains near
1.0; as complexity increases, performance first decays gradually, then drops sharply toward zero, where LLMs effectively fail to solve the problems. 

We can observe that this decay is not related to the complexity of the individual tool execution, but rather the complexity of information provided in the context history which causes confusion and Memory Rot related behaviors.

Through empirical observation, we note that there exists a window of messages which are critically relevant to the current task , and as time goes on, any given message becomes less relevant, and more outdated (thus increasing Rot Behavior). 

A naive approach is to segment history into 2 streams, the first is the mostRecent messages, these are items which are most likely critically relevant to the current task, and should be in full content form. After that, we maintain a list of historicMessages, which are relevant to the long term goals, and omit messages that may contribute to Rot. We maintain list size limits as a barrier to decide how big each queue is. Once `mostRecientMessages` eclipses the target limit, any messages above the limit are filtered, and potentially promoted to historicMessages.

We filter `historicMessages` based upon the following conditions:
* Output from specific tools are ignored - things like LSTool, GrepTool, GlobTool, ReadFileTool, ReadManyFilesTool are all irrelevant in the history of things, and for long-running tasks, may represent stale content. Removing this from history is a good thing. 
* Any message which is considered an Error. The current tool output does not contain this as explicit field (currently), so instead, we check if the string contains “Error:”
* Any messages from the model which are invalid, empty, or thoughts.

This allows our local context to be ideal for the task at hand, and our longer history context to only contain the most relevant messages to the long running tasks - assuming that over time, many things become irrelevant, eventually truncating out of long history.

This approach takes advantage of the observation that most historic messages lose value over time, and increase potential for rot - by segmenting into two different datasets, with different residency requirements, we can manage the data there differently.

**As a Note** : There's multiple potential approaches to solving memory rot issues. This is simply one approach and may need further fine tuning as the product matures.

## Reviewer Test Plan

With gemini 2.5 (flash or pro) Memory Rot behavior can often be observed once the total memory size eclipses around 15k lines of content. As such, any task which forces the MonoAgent model to execute in an iterative fashion in CRUD loops will easily trigger this behavior. For example:

* Given the directory <here> Walk through all the folders and add a readme.md file to each folder which describes the files contained there-in, and any thematic summary of this folder.
* Given the directory <here> open each file, observe each function, and update the docstring for it. Remember, docstrings are critical in maintaining solid code practices. Your responsibility is to ONLY edit the docstring for the file (don't touch any other parts of the file, like function headers or comments)

When observing with these types of "long jobs", notice that about the 25-30 turn mark (where each message has 2 turns, a call and response pair) that  behavior starts to degrade - loops start to happen, the model starts to forget the work they are doing, etc etc. With the edits in place, you should see that behavior changes such that:

1. These types of memory rot behaviors occur less often
2. tryCompressChat gets called less often. 

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅   | ✅   | ❓  |
| npx      | ✅   | ✅   | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

We observe that the following issues are most likely related to memory rot at certain chat session lengths or job sizes:
* https://github.com/google-gemini/gemini-cli/issues/1531
* https://github.com/google-gemini/gemini-cli/issues/1484


mattKorwel:

PR #1557, opened 2025-06-25
fix(telemetry): handle all promise rejections in ClearcutLogger

zhiyue:
## Summary
- Added proper error handling to `flushToClearcut()` method to prevent unhandled promise rejections
- The method now always returns a resolved promise with an empty LogResponse object on errors
- This prevents the CLI from crashing when telemetry requests fail due to network errors

## Problem
The gemini CLI crashes with an unhandled promise rejection when telemetry requests to `play.googleapis.com` fail due to network errors (e.g., ECONNRESET). This disrupts user workflow and causes unexpected exits.

## Solution
Modified the `flushToClearcut()` method to include a `.catch()` block that:
1. Handles all errors gracefully
2. Logs errors only in debug mode to avoid cluttering console
3. Returns an empty LogResponse object to maintain the Promise contract
4. Ensures the promise is never rejected, preventing unhandled rejections

## Testing
- The fix ensures all promise rejections are handled internally
- No additional `.catch()` blocks needed at call sites since the method never rejects
- Existing functionality remains unchanged - telemetry still works when network is available

## Related Issues
Fixes #1539
Fixes #1425


packages/core/src/telemetry/clearcut-logger/clearcut-logger.ts
```diff
@@ -120,15 +120,27 @@ export class ClearcutLogger {
         reject(e);
       });
       req.end(body);
```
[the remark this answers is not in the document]
zhiyue: ignore

jerop:
@zhiyue thanks for the contribution, could you please sign the CLA?

zhiyue:
Thank you! I have already registered/sign the CLA.

> @zhiyue thanks for the contribution, could you please sign the CLA?

packages/core/src/telemetry/clearcut-logger/clearcut-logger.ts
```diff
@@ -113,22 +113,34 @@ export class ClearcutLogger {
...
+        }
+      })
+      .catch((error: unknown) => {
+        // Handle all errors to prevent unhandled promise rejections
+        if (this.config?.getDebugMode()) {
+          console.debug(
+            'Failed to flush events to Clearcut (non-critical):',
+            error,
+          );
+        }
```
jerop: could you handle the error similar to line 129 above with console.error?
```suggestion
        console.error('Error flushing log events:', error);
```

jerop: cc @anj-s

anj-s (dismissed):

jerop (dismissed):

zhiyue:
@jerop Hi, I see that this PR has already been approved. Can I go ahead and merge it now?
Do we need to wait for approval from gemini-cli-askmode-approvers as well?

jerop:
yes, we need an additional approval - cc @allenhutchison @scottdensmore

packages/core/src/telemetry/clearcut-logger/clearcut-logger.ts
```diff
@@ -84,7 +84,7 @@ export class ClearcutLogger {
 
   flushToClearcut(): Promise<LogResponse> {
     if (this.config?.getDebugMode()) {
-      console.log('Flushing log events to Clearcut.');
+      console.debug('Flushing log events to Clearcut.');
```
jerop: could you change these back to log, and keep this PR focused on the catching the error only?

applies to the ones below too

scottdensmore:
@zhiyue can you make the couple of changes to move back to log vs debug and we can get this change merged and out with the next release

packages/core/src/telemetry/clearcut-logger/clearcut-logger.ts
zhiyue: Okay, I’ll change these back to log and keep this PR focused only on catching the error. Thanks for pointing it out!

scottdensmore (approved):
Approved

PR #1566, opened 2025-06-25
Validate for known fields in settings json

louisjimenez:
This validates the settings.json in both the project and user paths to ensure that only known fields are provided. If an unrecognized field is provided an error is thrown on startup and exits.

```
Please fix /.gemini/settings.json and try again.
Error in /.gemini/settings.json: Invalid settings foo: bar
```


packages/cli/src/config/settings.ts
```diff
@@ -215,6 +269,20 @@ export function loadSettings(workspaceDir: string): LoadedSettings {
       const parsedWorkspaceSettings = JSON.parse(
         stripJsonComments(projectContent),
       ) as Settings;
+
+      const invalidWorkspaceSettingsKeys = validateSettings(
+        parsedWorkspaceSettings,
+        validSettingKeys,
+      );
+      if (invalidWorkspaceSettingsKeys.length > 0) {
```
jacob314: nit: add a helper that validates the settings and adds them to setting errors to avoid duplicating this code for the two settings files.

jacob314 (dismissed):
This is great! Looking forward to building off this to provide some basic UI to let users edit all of the UI from inside Gemini CLI.
lgtm

packages/cli/src/config/settings.ts
louisjimenez: Good idea, done!

jacob314 (approved):

packages/cli/src/config/settings.ts
```diff
@@ -67,6 +67,36 @@ export interface Settings {
   // Add other settings here.
 }
 
+// The allSettings `Record<keyof Settings, undefined>`,
+// ensures that any changes to the `Settings` interface will result in a
+// compile-time error if this object is not also updated.
+const allSettings: Record<keyof Settings, undefined> = {
```
swissspidy, line 73: Ideally we have a JSON schema of all supported keys and their expected values. Then we can just validate it using the newly available `SchemaValidator` class in the core package.

jacob314: agree a JSON schema for this would be ideal.

PR #1579, opened 2025-06-25
changed 429 failover from 3 consecutive to 2 for OAuth users

bdmorgan:
In an attempt to address persistent 429 errors, OAuth users will now be failed over to Flash after two consecutive errors (was previously set to 3).


scottdensmore (approved):
Approved

NTaylorMullen (approved):
Hopefully this will speed up responses for folks too

PR #1584, opened 2025-06-25
fix(ui): reset session stats when clearing screen so footer shows 100 % context (Fixes #1541)

### LINKED ISSUE #1541: Feature Request: Enhance /clear command to reset model context
#### **The Problem**

Currently, the `/clear` command in the Gemini CLI only clears the terminal's visual output. It does not reset the model's conversational context.

This means that to start a fresh conversation with a clean slate, a user must exit the CLI and restart it. This process often requires re-authentication, which can be inconvenient and disruptive to the workflow.

#### **Proposed Solution**

I suggest enhancing the `/clear` command to also reset the model's conversational context. This would effectively start a new, clean session without the need to restart the application.

Perhaps the existing `/clear` command could be modified, or a new command like `/reset` or `/new` could be introduced.

#### **Expected Behavior**

When the enhanced `/clear` (or a new) command is executed:

1. The terminal screen is cleared.  
2. The model's conversational history and context are purged.  
3. The user can start a new conversation from scratch without restarting the CLI.

#### **Benefits**

This change would significantly improve the user experience by providing a quick and easy way to manage conversational state, making the tool more efficient and user-friendly.

Vinit-Chandak:
### PR Title
**fix(ui): reset session stats when clearing screen so footer shows 100 % context (Fixes #1541)**

---

## TL;DR
Adds `SessionContext.resetStats()` and calls it from both the `/clear` slash command and the **Ctrl + L** clear-screen shortcut.  
This instantly zeros the token counters so the footer reads “100 % context left” right after the screen is cleared.

## Dive Deeper
* **Problem:** Context was cleared in the backend, but the UI token counts persisted until the next user turn.  
* **Solution:**  
  1. Implemented `resetStats()` to wipe `cumulative`, `currentTurn`, and `currentResponse` in `SessionContext`.  
  2. Called the helper (defensively with `?.`) in `slashCommandProcessor` and `App` so existing mocks/tests still pass.  
* **Scope:** Purely UI state; no API surface or user-visible strings changed.

## Reviewer Test Plan
1. Start the CLI (`npm start` or `npx @google/gemini-cli`).  
2. Send a few prompts until the footer shows < 100 % context.  
3. Type `/clear` → footer should immediately show **100 %**.  
4. Repeat the test using **Ctrl + L** instead of `/clear`.  
5. Run automated checks:

   ```bash
   npm run preflight   # lint + type-check + unit + integration tests
   ```

## Testing Matrix

| Platform | 🍏 macOS | 🪟 Windows | 🐧 Linux |
|----------|----------|-----------|----------|
| **npm run** | ✅ | ✅ | ✅ |
| **npx** | ✅ | ✅ | ✅ |
| **Docker** | ✅ | ✅ | ✅ |
| **Podman** | ✅ | — | ✅ |
| **Seatbelt** | ✅ | — | — |

## Linked Issues / Bugs
* Fixes [#1541](https://github.com/google-gemini/gemini-cli/issues/1541)

### DIFF
--- a/packages/cli/src/ui/contexts/SessionContext.tsx
+++ b/packages/cli/src/ui/contexts/SessionContext.tsx
@@ -42,6 +42,12 @@ interface SessionStatsContextValue {
   addUsage: (
     metadata: GenerateContentResponseUsageMetadata & { apiTimeMs?: number },
   ) => void;
+  /**
+   * Fully resets all token counters (cumulative, currentTurn, currentResponse) as well as the
+   * session start time. This is primarily used by the /clear command so that the "context left"
+   * indicator in the footer immediately reflects the cleared state.
+   */
+  resetStats: () => void;
 }
 
 // --- Context Definition ---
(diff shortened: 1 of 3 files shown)


packages/cli/src/ui/contexts/SessionContext.tsx
```diff
@@ -175,13 +181,55 @@ export const SessionStatsProvider: React.FC<{ children: React.ReactNode }> = ({
     }));
   }, []);
 
+  /**
+   * Resets the entire stats object back to its initial state. We preserve the existing
+   * sessionStartTime because /clear is intended to wipe context, not necessarily denote a brand
+   * new session. If this behaviour needs to change in the future we can revisit it.
+   */
+  const resetStats = useCallback(() => {
```
NTaylorMullen, line 189: Hmmm but this also then kills any in-between tokens users have utilized doesn't it? 

Unfortunately I think we need to maintain that information because the root motivation of stats is to show people how much has happened on their account.

NTaylorMullen:
@abhipatel12 I know you've done a lot of thinking in this area as well. Would you mind advising?

packages/cli/src/ui/contexts/SessionContext.tsx
Vinit-Chandak: Makes sense. So, we can simply preserve the cumulative totals instead of resetting them, would that work?

abhipatel12:
Hey! Sorry for the delayed response. Yesterday, a pretty large refactoring of the stats was merged in. However, this edge case still exists in the new refactoring. In the new change, we use the `lastPromptTokenCount` as a means to calculate this context window. However, we still need to reset this amount. 

We could technically listen in and reset this one stat by plugging into the slash command processor like you have. Another option is to start logging a `/clear` event within the current logging system. This event can then be listened for and the only changes needed would be in the `uiTelemetryService`. 

Take a look at that service and lmk your thoughts! Great catch on this bug though!

mattKorwel:

Also I believe this issue was resolved in a different PR. Please verify if the needed behavior is still an issue before reopening.

PR #1608, opened 2025-06-25
Streamline issue submission with YAML forms

jerop:
This commit migrates the project's issue templates from the legacy markdown format to the modern YAML-based GitHub issue forms. This provides a more structured and user-friendly experience for submitting bug reports and feature requests.

Key changes:
- Replaced  with a structured  form.
- Added a new structured  form.
- Updated the  command to pre-fill the  field in the new bug report template, ensuring diagnostic information is correctly passed.

Here's an example issue form that was created when I ran `/bug foo`:
image

Note that the above picked up the template in the organization because the one in this PR has not been checked into the repo yet - https://github.com/google-gemini/.github/blob/main/.github/ISSUE_TEMPLATE/bug_report.yml.


allenhutchison:
Can you either update the /bug command template or change this template to be compatible?

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,52 @@
...
+
+  - type: textarea
+    id: expected
+    attributes:
+      label: What did you expect to happen?
+    validations:
+      required: true
+
+  - type: textarea
+    id: client-info
```
allenhutchison: /bug expects a component called body, but we can change that in the template. Just make sure it's consistent with what you have here.

jerop: Done!

aper0lspritz:
This is awesome, thanks @jerop!

allenhutchison:
Because you have changed the fields we also need to update the docs for bugCommand in configuration.md. This may be a breaking change for people who are using custom bug paths.

jerop:
@allenhutchison fixed the docs - do we just need to mention this change in the release notes?

allenhutchison (dismissed):
Looks good. I think the breaking issue is ok.

allenhutchison (approved):

PR #1618, opened 2025-06-25
feat(mcp): add oauth for sse and http transports

### LINKED ISSUE #3704: Add Google OAuth support to MCP client

Add to MCP client the ability to pass OAuth tokens in requests to MCP servers that require authentication. Specifically, we need support for Google OAuth for our MCP servers.

### Why is this needed?

We are developing an MCP server that we want to restrict to certain users/groups. Currently the Gemini CLI does not add the user's oauth tokens in the requests sent to the server (The server is running in SSE/StreamableHttp mode). It would be great if the CLI could authenticate the user that's using the CLI and pass in the OAuth token when making requests to the MCP server. If we don't want this to apply across the board, maybe we can add a field in the mcpservers config that enables it for specific mcp servers?

### Additional context

_No response_

jamesewoo:
This change introduces an 'oauth' boolean flag to 'MCPServerConfig' to enable authentication for servers connected via SSE and Streamable HTTP transports.

Fixes: #3704

### DIFF
--- a/packages/core/src/telemetry/clearcut-logger/clearcut-logger.ts
+++ b/packages/core/src/telemetry/clearcut-logger/clearcut-logger.ts
@@ -23,6 +23,7 @@ import {
   getCachedGoogleAccount,
   getLifetimeGoogleAccounts,
 } from '../../utils/user_account.js';
+import { safeJsonStringify } from '../../utils/safeJsonStringify.js';
 
 const start_session_event_name = 'start_session';
 const new_prompt_event_name = 'new_prompt';
@@ -65,7 +66,7 @@ export class ClearcutLogger {
     this.events.push([
       {
         event_time_ms: Date.now(),
-        source_extension_json: JSON.stringify(event),
+        source_extension_json: safeJsonStringify(event),
       },
     ]);
   }
@@ -121,7 +122,7 @@ export class ClearcutLogger {
           log_event: eventsToSend,
         },
       ];
-      const body = JSON.stringify(request);
+      const body = safeJsonStringify(request);
       const options = {
         hostname: 'play.googleapis.com',
         path: '/log',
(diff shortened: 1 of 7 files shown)


ashwinpvg (approved):

packages/core/src/tools/mcp-client.ts
```diff
@@ -171,12 +172,39 @@ async function connectAndDiscover(
...
+  if (
+    (mcpServerConfig.httpUrl || mcpServerConfig.url) &&
+    mcpServerConfig.oauth
+  ) {
+    try {
+      const auth = new GoogleAuth({
+        scopes: 'https://www.googleapis.com/auth/userinfo.email',
+      });
+      const authClient = await auth.getClient();
+      const headers = await authClient.getRequestHeaders();
```
AustinWise: Does this work with identity providers other than Google? Also it looks like this gets the headers once on connection. If so, this does not handle getting a new token when the token expires.

The MCP library has OAuth support that you could potentially use. See the [sample](https://github.com/modelcontextprotocol/typescript-sdk/blob/main/src/examples/client/simpleOAuthClient.ts). One limitation is that it currently only supports [dynamic client registration](https://modelcontextprotocol.io/specification/2025-06-18/basic/authorization#dynamic-client-registration), see modelcontextprotocol/typescript-sdk#714.

jamesewoo: We want to add support for Google auth specifically. We tried using modelcontextprotocol SDK, but its strict adherence to OAuth 2 doesn't work with Google.

I could change the oauth config from a boolean to a provider enum. Would that work? I'll also add the token refresh logic.

AustinWise: I think an enum makes sense. That way it could be extended to support other identity providers, like GitHub.

jamesewoo: I added token refresh support and the provider enum, PTAL.

ashwinpvg (approved):

AustinWise (approved):
Thanks, I think this would work for a use case I had in mind. That said, I'm not a maintainer of this repo.

FYI, there is another, potentially complimentary, PR for OAuth here: #3569

PR #1633, opened 2025-06-25
Add emacs support, as per user requests. :) 

### LINKED ISSUE #1660: Support Emacs Editor

Please support the old editor, Emacs. Thanks.

### Why is this needed?

Emacs is the best editor of the world, and many people are using Emacs.

### Additional context

This feature is related to [Add Support for EDITOR Environment Variable in External Editor Selection](https://github.com/google-gemini/gemini-cli/issues/1698#issue-3177433782).

dynamicwebpaige:
:)

Fixes #1660


packages/core/src/utils/editor.ts
```diff
@@ -49,6 +50,10 @@ export function allowEditorTypeInSandbox(editor: EditorType): boolean {
   if (['vscode', 'windsurf', 'cursor', 'zed'].includes(editor)) {
     return notUsingSandbox;
   }
+  // For terminal-based editors like vim and emacs, allow in sandbox.
+  if (['vim', 'emacs'].includes(editor)) {
```
jacob314: this is dead code as the default is to return true.

jacob314 (commented):
Thanks for adding this. Appreciate the option as a long time emacs user who barely knows there way around vim.
please add a test similar to the test for vim.

allenhutchison:
When you have the tests ready give me a ping to merge this.

dynamicwebpaige:
@allenhutchison Added tests, PTAL! 😄

jacob314 (dismissed):
Thanks so much for adding! Cant wait to use this.

allenhutchison (dismissed):

package-lock.json
NTaylorMullen, line 1: Woops I think you may have accidentally nuked this one Paige ❤️

rmarquis:
All right, I'll be "that" guy: Instead of adding yet again another hardcoded editor, shouldn't this be closed without merging and handled instead by the more genalized #1698 and related PR #1708?

packages/core/src/utils/editor.ts
```diff
@@ -35,6 +35,7 @@ const editorCommands: Record<EditorType, { win32: string; default: string }> = {
   cursor: { win32: 'cursor', default: 'cursor' },
   vim: { win32: 'vim', default: 'vim' },
   zed: { win32: 'zed', default: 'zed' },
+  emacs: { win32: 'emacs.exe', default: 'emacs' },
```
vroonhof: Though I agree with the sentiment that harcoding all these editors feels wrong in general, if we are goint ot go this way we should really support emacsclient    (e.g. with 'emacsclient -a "emacs")

klukaszek:
Hey @rmarquis, I totally agree and I'll be "that" guy with you as well. 

I believe that hardcoding editor options into the TUI is ridiculous considering there are existing platform standards for this exact thing.

I think the best solution would be to prompt the user once to pick between "use terminal default editor" or "choose editor" (which would bring them to the existing choice modal with hardcoded editors). If the user changes their mind, they can just use the respective / command to change their choice.

Also adding a note about the EDITOR env-var could be nice for users who might be unfamiliar with their terminal environments.

vroonhof:
> Hey @rmarquis, I totally agree and I'll be "that" guy with you as well.
> 
> I believe that hardcoding editor options into the TUI is ridiculous considering there are existing platform standards for this exact thing.
> 
> I think the best solution would be to prompt the user once to pick between "use terminal default editor" or "choose editor" (which would bring them to the existing choice modal with hardcoded editors). If the user changes their mind, they can just use the respective / command to change their choice.
> 
> Also adding a note about the EDITOR env-var could be nice for users who might be unfamiliar with their terminal environments.

I think the main issue is that at least some of the invocations don't actually invoke a plain editor, but want to invoke a "visual" diff or even a merge tool (because the edit is on a diff). See ediff invovocation. There are less conventions for that. However even that can be solved just by a config settings that allows a string with place holders (and in the age of LLM it doesn't need menu).

Phundrak:
While I do agree that using EDITOR is a good option, another option could be to use [launch-editor](https://github.com/yyx990803/launch-editor) which could allow for a greater control of how to open a file (exact line and column selected). I also don’t think this is either EDITOR or launch-editor, it could be both: use launch-editor by default, use EDITOR at the request of the user, or vice versa.  
But if we have to choose between one or the other, I’d definitively prefer implementing EDITOR first.

lesteral:
@dynamicwebpaige, @NTaylorMullen - (eager emacs user here :-) the failed check in [Build and Lint (20.x)](https://github.com/google-gemini/gemini-cli/actions/runs/16093068314/job/45412218870?pr=1633#logs) looks like it just needs a "});" which might have gone AWOL during merge.

```
[error] packages/core/src/utils/editor.test.ts: SyntaxError: '}' expected. (402:1)
[error]   400 |   });
[error]   401 | });
[error] > 402 |
[error]       | ^
```

swissspidy:
Some good suggestions here regarding `EDITOR`, `launch-editor` and similar vs maintaining an ever growing hardcoded list. I just opened #3849 to track this in a more centralized place.

packages/core/src/utils/editor.ts
```diff
@@ -44,35 +29,31 @@ function commandExists(cmd: string): boolean {
...
 export function allowEditorTypeInSandbox(editor: EditorType): boolean {
   const notUsingSandbox = !process.env.SANDBOX;
-  if (['vscode', 'vscodium', 'windsurf', 'cursor', 'zed'].includes(editor)) {
+  if (['vscode', 'windsurf', 'cursor', 'zed'].includes(editor)) {
     return notUsingSandbox;
   }
+  // For terminal-based editors like vim and emacs, allow in sandbox.
+  if (['vim', 'emacs'].includes(editor)) {
+    return true;
+  }
```
mattKorwel:
```suggestion
```

mattKorwel: Per @jacob314 's comment above

packages/core/src/utils/editor.ts
mattKorwel: @vroonhof I agree there needs to be a better plan. However given we don't have that plan yet, i'm going to merge this, it can be changed later.

packages/core/src/utils/editor.ts
```diff
@@ -141,6 +145,11 @@ export function getDiffCommand(
           newPath,
         ],
       };
+      case 'emacs':
```
mattKorwel:
```suggestion
    case 'emacs':
```

mattKorwel: linting

mattKorwel (approved):
Thanks for the add here @dynamicwebpaige . Fixed up build erros and merging. As for the outstanding comments on better approach, there might be a more holistic way, we'll address that in a follow up. Thanks all!

PR #1653, opened 2025-06-25
When resuming a checkpoint always add items to history even if not shown

### LINKED ISSUE #1650: Resuming a saved snapshot does not restore tool calls or responses
Saving (`/chat save`) creates a json file with full tool calls and responses. However, (`/chat resume`) does not restore the tool calls. To reproduce:

 1. Open gemini-cli and provide a prompt that causes the model to perform a tool call
 2. Use `/chat save` to save a checkpoint
 3. Quit and restart the app, use /chat resume to load the previously stored checkpoint
 4. Use `/chat save` again to save a new checkpoint
 5. Compare the two checkpoints, note that the new one does not round-trip the tool calls

bbiggs:
## TLDR

Today when we resume checkpoints, tool calls and output are not restored from the saved checkpoint. The code that presents them to the UI skipped over these parts, but also skipped adding them to history.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes #1650


scottdensmore (approved):
Approved

PR #1656, opened 2025-06-25
feat: reliably use selected themes across the software and remove hardcoded ansi codes (greenscreen theme added as a test)

acoliver:
… colors

- Add new GreenScreen theme with monochrome green terminal aesthetic
- Migrate UI components from global Colors object to theme-based colors
- Fix authentication, stats, and footer components to respect active theme
- Update theme manager to include GreenScreen in available themes

## TLDR

new greenscreen theme, all the components now use the themes instead of hardcoded things

## Dive Deeper

I added a greenscreen theme, this was in part to test the fixes and partly because I prefer green on black during long coding sessions. Imagine it is 1982 and you have a shiny new Phosphor 1 or 27 monitor but maybe a slightly more pleasant shade of green. 

Now, imagine, for some reason, random screens show up in strange colors. This would be troubling and you'd wonder if you got ripped off and were on CGA instead of your higher resolution Hercules card. That would be very disappointing.

With these changes, all the components I could discover no longer have hard-coded colors but ask the theme. 

## Reviewer Test Plan

switch to diferent themes. Try them. 

## Testing Matrix

let's be real none of these changes affect docker, podman or seatbelt

|          | 🍏  | 🪟  | 🐧  |
| npm run  | x  | ❓  | ❓  |
| npx      | x  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

https://github.com/google-gemini/gemini-cli/issues/1661


acoliver:
https://github.com/google-gemini/gemini-cli/issues/1661

packages/cli/src/ui/components/messages/ToolMessage.tsx
```diff
@@ -180,15 +185,18 @@ const ToolInfo: React.FC<ToolInfo> = ({
...
-const TrailingIndicator: React.FC = () => (
-  <Text color={Colors.Foreground} wrap="truncate">
-    {' '}
-    ←
-  </Text>
-);
+const TrailingIndicator: React.FC = () => {
+  const theme = themeManager.getActiveTheme();
+  return (
+    <Text color={theme.colors.Foreground} wrap="truncate">
```
jacob314: grabbing the theme from the theme manager is making all of this code a bit more complex and isn't needed as colors.ts is actually doing this behind the scenes in each getter.

packages/cli/src/ui/themes/green-screen.ts
```diff
@@ -0,0 +1,120 @@
...
+  AccentCyan: '#6a9955',
+  AccentGreen: '#6a9955',
+  AccentYellow: '#6a9955',
+  AccentRed: '#6a9955',
+  Comment: '#6a9955',
+  Gray: '#6a9955',
+  GradientColors: ['#6a9955', '#6a9955'],
+};
+
+export const GreenScreen: Theme = new Theme(
```
jacob314, line 25: Fyi @miguelsolorio for feedback on if we want to include this theme.
Rather than supporting this theme we might want to instead consider allowing users to load a custom theme from a file. That would allow power users to customize their theme by editing settings.json without requiring us to add every possible theme to this class.

packages/cli/src/ui/components/messages/ToolMessage.tsx
acoliver: Then it must not be working because no theme colors were applied in the ToolMessages. Looking into this maybe there were other parts that were hardcoded. There are/were hardcoded ansi codes in some places.

acoliver: @jacob314 this should be resolved - I also found other places with hardcoded ansi codes. In addition, I swapped some of those to use chalk (since it was already was a dependency of ink) and extended the colors utility to work outside of the react parts.

packages/cli/src/ui/themes/green-screen.ts
acoliver: To be clear you want this theme specifically for one reason. Unlike other themes you immediately know if something is wrong in the theme application. You see a color that isnt green: it is wrong. Other themes you have to go look. It's as much a visual debugging tool as a theme. Custom themes would be nice but at least one "everything is the same color" theme is ideal. It also doesn't use the default green (00ff00) so you know its theme green and not ansi green.

jacob314: Clever idea!I like having a custom theme for debugging. Lets enable it but hide it from the theme picker dialog. Gemini CLI authors can still enable it by manually editing their their settings.json. What do you think?

Another option for this is to allow users to specify completely custom themes from their settings.json.
Thumbs up
https://github.com/google-gemini/gemini-cli/issues/2122
if you would like that

mattKorwel:
@acoliver and @jacob314 this seems pretty out of date and has quite a few merge conflicts, it would need some work to be mergeable.

Going through our oldest PR's and marking them as Stale and closing. Please feel free to reopen and make ready for merge and review if you would like. 

Either way, Thanks for the contribution!

jacob314:
We now support custom themes so you can add the greenscreen theme as a custom theme without any change to Gemini CLI.

acoliver:
> We now support custom themes so you can add the greenscreen theme as a custom theme without any change to Gemini CLI.

so you guys reallllly focused on that part of the patch. which was mainly for testing

So did you ACTAULLY fix the issue (themes don't get applied to most UI elements and you have hardcoded ansi characters throughout) or "oh now you can have greenscreen" ?  
@jacob314

PR #1668, opened 2025-06-26
429 fix

bdmorgan:
Model fallback fixes


scottdensmore (approved):
Approved

tarruda:
Did this fix make into  `v0.1.4`? I'm still getting 429 when I use `GEMINI_API_KEY` env var as the auth method on that version.

bdmorgan:
Yes we're currently investigating throttling and capacity issues deeper in the stack. This fix is in the client (and fixed a major issue...but there are still more).

PR #1612, opened 2025-06-25
Fix URL for tos-privacy.md

bobaekang:
## TLDR

Fixing the URL for `tos-privacy.md` in `<AuthDialog>` component.

## Dive Deeper

Changing the URL path to use the correct org name  (`google/gemini-cli` -> `google-gemini/gemini-cli`).

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| -------- | --- | --- | --- |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

PR #1688, opened 2025-06-26
Fix doc link in the auth dialog 

anj-s:
* Moved to removing hyperlinks since they don't render in all terminals. We need to investigate a better way of displaying this.
* Fixing broken URL link


iennae:
This would replace https://github.com/google-gemini/gemini-cli/pull/1612

scidomino (approved):

scottdensmore (approved):
Approved

PR #1708, opened 2025-06-26
feat(cli): Add support for VISUAL/EDITOR environment variable for file viewing.

### LINKED ISSUE #1698: Add support for VISUAL/EDITOR environment variable in external editor selection
## Problem

The Gemini CLI currently only supports a hardcoded list of editors (`vscode`, `windsurf`, `cursor`, `vim`, `zed`) and does not check for the Unix/Linux `EDITOR` environment variable.

## Current Behavior

- Users can only select from 5 predefined editors
- The `EDITOR` environment variable is completely ignored
- Users with editors like Helix, Nano, Emacs, or other editors cannot use the "Modify with external editor" feature effectively

## Expected Behavior

- The CLI should respect the `EDITOR` environment variable
- Users should be able to select "EDITOR Environment Variable" as an option
- The system should intelligently handle different editors' diff modes
- Maintain backward compatibility with existing hardcoded editor types

## Proposed Solution

Add an "editor" type to the `EditorType` union that:

1. **Reads from `EDITOR` environment variable**
2. **Intelligently handles editor-specific diff modes:**
   - Vim/Neovim: Uses `-d` flag with proper diff configuration
   - Emacs: Opens `file1` and `file2` as before and after.
   - Helix: Uses `--vsplit` for side-by-side view
   - Others: Falls back to opening the new file for editing
3. **Provides proper execution strategy** (spawn vs execSync based on editor type)
4. **Maintains full backward compatibility**

## Implementation Status

I have already implemented this feature and will be opening a PR shortly. The implementation includes:
- Full test coverage with 82 passing tests
- Support for intelligent editor detection
- Backward compatibility with all existing functionality
- Passes all preflight checks
- Tested locally using Helix, Emacs, Nano, and Vim.

### Why is this needed?

Using the EDITOR environment variable lets users integrate their editor of choice into gemini-cli for quick editing jobs.

### Additional context

_No response_

klukaszek:
## TLDR

Adds support for the `EDITOR` environment variable to external editor functionality. Users can now select "EDITOR Environment Variable" option and use their preferred editor (vim, emacs, helix, nano, etc.) instead of being limited to 5 hardcoded choices.

## Dive Deeper

Previously only supported `vscode`, `windsurf`, `cursor`, `vim`, `zed`. Now respects the standard Unix `EDITOR` environment variable with intelligent handling for different editor types (vim gets `-d` diff mode, helix gets `--vsplit`, others get basic file opening). The proposed fix does not break backwards compatibility as I've just added a new `editor` to the list of existing supported ones.

## Reviewer Test Plan

1. Set `export EDITOR="your_editor"`
2. Use "Modify with external editor" → select "EDITOR Environment Variable". 
You can also set your preferred editor in `~/.gemini/settings.json` by setting `{ "preferredEditor": "editor" }`.
3. Verify your editor opens and changes apply correctly
4. Test with a few different editors if available

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ✅  | ❓  | ❓  |
| Docker   | N/A | N/A | N/A |
| Podman   | N/A | -   | -   |
| Seatbelt | N/A | -   | -   |

`npm run preflight` tests and manual testing with Helix, Nano, Emacs, and Vim, were all successful.

## Linked issues / bugs

Fixes: [Feature request: Add support for EDITOR environment variable](https://github.com/google-gemini/gemini-cli/issues/1698)


NTaylorMullen:
Hey @klukaszek thank you for the contrib! I totally get the value in wanting to have a general or "custom" editor setting. That being said I could also imagine something like this with a "custom" setting or something to that extent. Lets hash out some more details in the corresponding issue.

Also, I think some of your other changes here are orthogonal to custom editor commands so I'd suggest splitting those out if you're looking to have those for consideration.

cowile:
This should check and give priority to the VISUAL env variable too like bash does.

https://unix.stackexchange.com/questions/4859/visual-vs-editor-what-s-the-difference

klukaszek:
@cowile You're completely right! I've added VISUAL as the preferred env-var and then EDITOR as a fallback!

I've also addressed some earlier concerns as well.

EDIT: Also unsure of the "needs issue" tag issued by the bot because there is an issue associated with this PR already.

packages/core/src/utils/editor.ts
```diff
@@ -35,9 +43,22 @@ const editorCommands: Record<EditorType, { win32: string; default: string }> = {
   cursor: { win32: 'cursor', default: 'cursor' },
   vim: { win32: 'vim', default: 'vim' },
   zed: { win32: 'zed', default: 'zed' },
+  editor: {
+    win32: process.env.EDITOR || 'notepad',
+    default: process.env.EDITOR || 'vi',
+  },
```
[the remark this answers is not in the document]
klukaszek: /gemini review

packages/core/src/utils/editor.ts
```diff
@@ -114,6 +146,67 @@ export function getDiffCommand(
...
+      if (!editorCmd) {
+        return null;
+      }
+      const parts = editorCmd.split(' ');
+      const editorCommand = parts[0];
+      const editorArgs = parts.slice(1);
+      const editorName = editorCommand.toLowerCase();
+
+      // Handle different editors' diff modes
+      if (
```
NTaylorMullen, line 161: Why special case vi/vim/nvim etc. vs. relying on the user to provide this in the editor args & command?

I ask because this seems to be overloading our existing "pick something specific" and add a little bit too much arcane magic to the editor flow

packages/core/src/utils/editor.ts
```diff
@@ -166,14 +259,61 @@ export async function openDiff(
...
 
+      case 'editor': {
+        // Handle VISUAL/EDITOR environment variable with proper execution strategy
+        const editorCmd = process.env.VISUAL || process.env.EDITOR;
+        if (!editorCmd) {
+          throw new Error('VISUAL or EDITOR environment variable not set');
+        }
+        const editorName = editorCmd.split(' ')[0].toLowerCase();
+
+        // Use spawn for GUI editors that might support --wait or similar
```
NTaylorMullen, line 280: Is there a better way to do this than special casing? If we're making this configurable IMO we should be relying entirely on configuration.

NTaylorMullen (changes requested):

eliliam:
This will be great to get merged, as https://github.com/google-gemini/gemini-cli/pull/1633 only added support for emacs, but my primary editor is helix.

eliliam:
@klukaszek one thing I noticed while testing this is when the diff opens, Helix is unable to automatically set the language type for syntax highlighting due to the file being opened not ending with the correct extension. Instead of adding a suffix for this temporary file like you currently are, is it possible to use a prefix instead so that the file extension is preserved thus allowing editors to infer the language based on the file extension?

mattKorwel:
👋 Hi there @klukaszek i'm looking through our outstanding open PR's. Can you, if you'd still like to move forward with this, address the merge conflicts and address the comments from Taylor above?

abhipatel12:
Hey! It's been a week since the last check on this PR. I'm going to mark this as closed due to lack of activity. If you need, please reopen it!

Soccera1:
I would like to see this merged. I love nano and while I can use emacs, I don't like it if I'm working in a CLI.

schmitd:
If I punch up this PR with the requested changes, is there any hope of it getting merged? I see that the list of available editors has since grown, but helix, nano, and many others are still unsupported.

PR #1805, opened 2025-06-26
feat(agent): Introduce Foundational Subagent Architecture

abhipatel12:
## TLDR

This PR introduces a functional prototype of a **`SubAgentScope`**. The primary goal is to establish a clean, minimal, and reviewable baseline implementation of this new architectural pattern. It serves as a foundational "scaffolding" to facilitate broader discussion on the overall architecture and to unblock iterative development on more advanced capabilities.

Thanks to @mainroach for the initial investigation, diagnosis, and proof-of-concept on which this work is based. This PR builds directly on his work from the [mainroach_subagents](https://github.com/google-gemini/gemini-cli/tree/mainroach_subagents) branch.

## Dive Deeper

Our current mono-agent architecture lacks the ability to handle complex, long-running tasks without polluting the main chat history with intermediate steps (e.g., error/self-healing loops).

Solution: The Subagent (or "Temporal Chat Scope") Pattern

This PR represents a fundamental architectural shift for our agent. It moves us away from a limited "mono-agent" model, where all work pollutes a single chat history, and introduces the concept of Subagents. Think of a Subagent as an isolated, temporary function call. It can perform complex, multi-step tasks with its own self-healing loops, then discard its internal history and return only the final result. This immediately enables "Smart-Tools" that are more resilient and keep the main chat history clean.

Strategically, this is the essential first step toward building a true Multi-Agent System (MAS), which will allow us to orchestrate specialized agents to solve much larger and more complex problems in the future.

### **Scope of This PR**

To keep this PR focused and reviewable, it includes:

- [✅] The core, cleaned-up `SubAgentScope` class.
- [✅] Basic unit tests and local testing (seen in video) to validate the current functionality.

### **Architectural Questions for Discussion**

This PR is intended to start a conversation. Your feedback on these points is crucial:

1.  **History Management:** The current implementation gives the subagent a clean slate. Should we introduce an option to inherit history from the parent agent? What are the trade-offs?
2.  **Error Handling Strategy:** When a tool fails, we currently log the error and continue. Is this the right default? Should a single tool failure terminate the entire subagent?
3.  **Scope-Local Tools:** `self.emitvalue` is currently hardcoded. Is this acceptable for internal mechanisms, or should we design a more generic registration system for scope-local tools?

### **Future Work & Known Issues (To Be Addressed in Follow-up PRs)**

-   **Forcing `self.emitvalue`:** The model sometimes forgets to emit all required output variables. A robust re-prompting mechanism is needed to ensure the subagent's contract is always fulfilled.
-   **Telemetry:** Subagent usage is not yet reflected in session stats. This is necessary for users to understand model utilization.

### **Demonstration: Before vs. After**

The value of this pattern is best shown visually. The "Smart-Tool" enabled by this architecture can self-heal from errors without polluting the main chat history.

**BEFORE: Mono-Agent Attempt (Current `main` branch)**

https://github.com/user-attachments/assets/edc721f7-bf97-4236-86bf-cd5231704942

Note: Tool failures will continue to cause memory rot. Each failure adds to the total chat history, and thus, each subsequent tool call will suffer a higher likelihood of failure. 

**AFTER: Subagent as a Smart-Tool (This PR)**

https://github.com/user-attachments/assets/75bf66d6-7a6c-4f75-832b-06ab592b4b11

Note: This does not show that the subagent had zero failures. It shows that the Agent-as-tool was able to self-heal and return a valid response without rotting the memory of the monolithic agent. This is essential to moving towards a multi-agent system.

## How to try locally

1. Start a chat with Gemini CLI.
2. Within this repo, provide the following prompt:
  ```
  Hey - you've got a /packages/src/tools folder, right? I need you to run through each file in that directory and update the  docstrings for each one - make sure you use the clean docstrings tool.
  ```

3. See the tool successfully add docstrings to every file without impacting the mono agent's context window.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ✅  | -   | -   |


mainroach:
Added FR for this feature at [3132](https://github.com/google-gemini/gemini-cli/issues/3132)

packages/core/src/core/subagent.ts
```diff
@@ -0,0 +1,599 @@
...
+      this.output.terminate_reason = SubagentTerminateMode.ERROR;
+      return;
+    }
+
+    const abortController = new AbortController();
+    const toolRegistry: ToolRegistry =
+      await this.runtimeContext.getToolRegistry();
+
+    // Prepare the list of tools available to the subagent.
+    const tools_to_load: string[] = [];
```
NTaylorMullen: nit: camelCase instead of snake case in TS ❤️ 
```suggestion
    const toolsToLoad: string[] = [];
```

packages/core/src/core/subagent.ts
```diff
@@ -0,0 +1,599 @@
...
+
+    chat.setSystemInstruction(this.buildChatSystemPrompt(context, toolsList));
+
+    let currentMessages: Content[] = [
+      { role: 'user', parts: [{ text: 'Get Started!' }] },
+    ];
+
+    const startTime = Date.now();
+    try {
+      while (true) {
```
NTaylorMullen, line 382: Interesting I would have imagined this while (true) would have been logic that we would have had centralized in our non interactive executor already.

packages/core/src/core/subagent.ts
```diff
@@ -0,0 +1,599 @@
...
+    }
+  }
+
+  /**
+   * Retrieves environment-related information to be included in the chat context.
+   * This includes the current working directory, date, operating system, and folder structure.
+   * Optionally, it can also include the full file context if enabled.
+   * @returns A promise that resolves to an array of `Part` objects containing environment information.
+   */
+  private async getEnvironment(): Promise<Part[]> {
```
NTaylorMullen: Ah interesting that you had to include this bit. Why does lifting the pre-existing history from the original chat thread not work?

packages/core/src/core/subagent.ts
```diff
@@ -0,0 +1,599 @@
...
+
+    return initialParts;
+  }
+
+  /**
+   * Returns an array of FunctionDeclaration objects for tools that are local to the subagent's scope.
+   * Currently, this includes the `self.emitvalue` tool for emitting variables.
+   * @returns An array of `FunctionDeclaration` objects.
+   */
+  private getScopeLocalFuncDefs() {
```
NTaylorMullen, line 621: I'd create a new tool that looks like the existing tools to capture schema. In the end all tools should be re-usable everywhere IMO.

packages/core/src/core/subagent.ts
```diff
@@ -0,0 +1,599 @@
...
+
+    // Add variable emission goals..
+    for (const [key, value] of Object.entries(this.promptConfig.outputs)) {
+      templated_goals += `\n* Use the 'self.emitvalue' tool to emit the '${key}' key, with a value described as '${value}'`;
+    }
+
+    const input = `You are an expert AI that takes on all sorts of roles to accomplish tasks for the user. You will continue to iterate, and call tools until the goals are complete. 
+
+Here are the tools you have access to, for this session, to solve the goals:
+<TOOLS>
```
NTaylorMullen: Why re-iterate the tools in this way when the core Gemini stack does this already?

NTaylorMullen (commented):
I know you're still cleaning up the PR a bit but had time to do a pass.

packages/core/src/core/subagent.ts
mainroach: We can do that - my hesitation with that path, originally was that it might cause more confusion in the codebase?

- The top-level MonoAgent doesn't entertain the concept of EmitValues - this is a subagent-only idea.

- The current codebase doesn't allow for the calling instance pointer to be passed into the tool execution, thus we have to manually filter this tool call inside the SubAgentScope::run function, and handle the variable storage there.

- As such, EmitValuesTool shouldn't follow the same pattern as existing tools - It shouldn't be initialized in config.ts (since it's not a global singleton, it's scope-local to the calling subagent), and if it was, the `execute` function would be stubbed out (since that execution happens inside the subagent) which means that if someone did accidentally use it outside of the SubAgentScope class, it would lead to undefined results. 

If you'd like EmitValuesTool to follow the same syntax as other tools, we would need to make a _larger change_ such that we didn't have to stub out the ::execute function : 
- modify the ToolBase::execute function definition to **also** accept a dict object of _instance specific variables_ which could be used in the execute loop. 
- These values would be provided by the runtime, not the LLM (like the current params are).  
- This way, when a tool is called, it can be passed the calling GeminiChat object, the calling scope (if applicable) or any other instance local variables which may be required for execution or prompt rendering. 

This would allow EmitValues to be a globally defined tools (like all the other tools are) but allow for invocation specific scoping variables to do execution on. You can see a bunch of scenarios where scope-specific variables are needed, that the runtime manages, not the LLM.

WDYT?

packages/core/src/core/subagent.ts
mainroach: Initial testing was showing that w/o this block, tools would not be called by the GenAI::GenerateContent invocation. 

We tested probably 40-50 iterations, and was unable to get generateContent call inside of GeminiChat to produce a tool call correctly w/o some reference to the tools themselves. We also tried just listing the _names_ of the tools (similar to how prompts.ts does it), but that also resulted in lack of tool invocation. After a few iterations, we found that we had to include the full definition list.

Now, it may be that the full definition list was required during the initial phases of testing due to the _brevity_ of this prompt, and that building a larger prompt (like prompts.ts) would trigger the GenerateContent invocation to return tools properly.. or that something else is going on 🤷 .

We will take a look at this before submitting the updated PR.

packages/core/src/core/subagent.ts
mainroach: A few concepts lead to the duplication of this function, and it's usage in the subagent scope.

1. It should be assumed that any SubAgent operates in their own scope, with a fresh history, as the default operation. Some SubAgents _may_ want to fork the calling Agent's chat history (be it the MonoAgent or any other SubAgent), which can be accomplished by passing in the parent GeminiChat::history object to the instantiation of the subagent. However, by default, any SubAgent should be considered a fresh execution scope.

2. GeminiClient::getEnvironment is responsible for _pushing messages_ into the GeminiChat's history to act as the starter messages for the whole system. SubAgentScope does not have access to the GeminiClient, and thus needed to duplicate the function. 

3. There's a chance that this data wouldn't exist in the calling GeminiChat's history, due to truncation or tryCompressChat. To ensure residency. it may be more viable to allow certain parts of this data to be rendered into the system prompt, rather than being a message. (e.g. date/time should most likely be in the system prompt, but the file directory should be a message, since it gets out of date quickly with any directory mutations)

NTaylorMullen: I see, how do you feel about refactoring the environmental context piece into a standalone utility that can be imported and re-used?

packages/core/src/core/subagent.ts
NTaylorMullen: I see hmmm. Let me ask maybe a dumb question. What if we were to move away from the variable functions entirely and instead relied on 1 final schema binding request that would take all of the history and fill out a given schema object? That way there wouldn't be individual tools that the LLM would need to invoke each step of the way, we'd have a guaranteed output and it'd ensure that the entire scope of the conversation was considered when deriving final outputs.

packages/core/src/core/subagent.ts
NTaylorMullen: Appreciate you taking a second look here. I'm definitely a bit surprised on the requirement here. Typically systems will perform better when the system prompt clearly defines "when" to use said tools BUT doesn't require the full schema or information

packages/core/src/core/subagent.ts
mainroach: I'll move getEnvironment into a new file which can be leveraged by the monoAgent and SubAgent .

packages/core/src/core/subagent.ts
mainroach: If I understand your comment, you're suggesting:
- Removing the EmitVariableTool (which emits a single variable, and can be called multiple times for multiple variables)
- When a subagent loop is completed, the subagent would call the `geminiClient::generateJson` function (or similar) to fill out all the target emission variables (if applicable).
  -  Once that call returns, parse the json, and fill out the emittedVariables map so that the calling system can fetch them.


If that's correct (?), I think that would be a fine pivot in order to remove the confusion regarding how this tool works... however, I'd provide the following observations:
- There's non-zero chances that the SubAgent claims work is finished, w/o doing the work required to fill out the target variables. In that case, the above emission would either most likely hallucinate a value, which would be incorrect vs. the work done.
- If the model does not fill out the right final variables, it would be ideal to enter back into the Agentic loop, to force it to do the extra work for that variable data. If we're doing a single call at the end, we'd loose this ability, since we've exited the agentic loop at this point - We'd have to re-enter the execution loop with an augmented message, targeting to self-heal the missing data.  
  - When you look at multiple failures, you're effectively back at being in the core of the Agentic loop... (e.g. we'd need an inner Agentic loop, and then an out "check if variables were properly emitted" loop). 
- Your proposal would suffice for short work - however if there's any history consolidation (e.g. tryCompressChat) it may remove the data from the history that would be needed to fill out the final JSON. That would move us back into the situation of the above two bulletpoints.

Completely do-able, if you'd prefer to go that route.

May I provide an alternate suggestion for consideration?
- Assume that SubAgent existence may lead to other tools which start to think in the space of wanting to do contextual scoped changes, rather than global ones - thus tools needing access to the calling AgentScope during `execute(..)` would be needed. So, this is not a 1-off situation.
- This could be supported with minimal code changes to `BaseTool::execute(..)` to accept an optional "runtimeContext:Record<string, any>" variable. Any existing tools would not need to be modified at all, and the subagent can fill out this packet when invoking the tools.
- EmitValueTool then gets upgraded to a _proper_ tool, as per your original request. 

WDYT?

packages/core/src/tools/clean-docstring.ts
```diff
@@ -0,0 +1,172 @@
...
+          returnDisplay: 'Docstrings updated successfully',
+        };
+      }
+    } catch (error) {
+      console.error('Orchestrator run failed:', error);
+    }
+    return {
+      llmContent: `Error: An error occurred while trying to fix up the docstrings for ${params.path}`,
+      returnDisplay: `Error: An error occurred while trying to fix up the docstrings for ${params.path}`,
+    };
```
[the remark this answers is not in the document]
abhipatel12: Not relevant. Docstring cleaning is being removed from this PR

packages/core/src/core/subagent.ts
```diff
@@ -0,0 +1,599 @@
...
+        callId,
+        name: functionCall.name as string,
+        args: (functionCall.args ?? {}) as Record<string, unknown>,
+        isClientInitiated: true,
+      };
+
+      let toolResponse;
+
+      // Handle scope-local tools first.
+      if (callId.startsWith('self.emitvalue')) {
```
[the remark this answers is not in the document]
abhipatel12: Done

packages/core/src/core/subagent.ts
```diff
@@ -0,0 +1,599 @@
...
+            toolResponseParts.push(part);
+          }
+        }
+      }
+    }
+    currentMessages = [{ role: 'user', parts: toolResponseParts }];
+    return currentMessages;
+  }
+
+  /**
```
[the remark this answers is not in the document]
abhipatel12: Done

packages/core/src/core/subagent.ts
abhipatel12: Done

packages/core/src/core/subagent.ts
abhipatel12: Sorry, reopened to help me track my changes. This has been done!

packages/core/src/core/subagent.ts
NTaylorMullen: > May I provide an alternate suggestion for consideration?

Ya I think this could work too. My comment was less about the coding aspects of the tool implementations though and more around the lines of ensuring an output. That being said I imagine evals could also help define which path is better here (I don't feel strongly). Your suggestion seems reasonable though.

packages/core/src/core/subagent.ts
abhipatel12: Update here! I worked on retesting this behavior.

I followed the following process:

1. Create new implementation of a subagent that uses two tools: LS tool, Write File Tool
2. Remove `<tool>` from `subagent.ts`
3. Run an input that activates this smart tool for 10 directories
4. See if write file correctly creates 10 files using the ls tool for its content.
5. Repeat 5 times (50 iterations)

After running this experiment, the tools seem to be called even without the `<tool>` block in the system.

I've removed it from the impl!

packages/core/src/core/subagent.ts
abhipatel12: Making async chat visible here:

Chatted about how this was slightly different due to the possibility that the agent can finish tool calls and not reach the goal which would prompt another loop. This isn't covered in the non interactive loop we have currently, so we can't necessarily (easily) reuse that loop. Marking this as resolved for now

packages/core/src/core/subagent.ts
```diff
@@ -0,0 +1,588 @@
...
+}
+
+/**
+ * Configures the execution environment and constraints for the subagent.
+ * This interface defines parameters that control the subagent's runtime behavior,
+ * such as maximum execution time, to prevent infinite loops or excessive resource consumption.
+ */
+export interface RunConfig {
+  /** The maximum execution time for the subagent in minutes. */
+  max_time_minutes: number;
```
allenhutchison, line 142: I feel like this could cause problems for long running stuff. I wonder if we would want to be more expressive here. Maybe adding some other fields 

- max_turns - which would stop execution after a set number of turns if the agent hasn't returned by then.

- max_tokens - like a budget for how many input tokens you are willing to spend on this task.

packages/core/src/core/subagent.ts
```diff
@@ -0,0 +1,588 @@
...
+}
+
+/**
+ * Configures the generative model parameters for the subagent.
+ * This interface specifies the model to be used and its associated generation settings,
+ * such as temperature and top-p values, which influence the creativity and diversity of the model's output.
+ */
+export interface ModelConfig {
+  /** The name or identifier of the model to be used. */
+  model: string;
```
allenhutchison, line 122: If we go down the route of model routing how would you want to handle this here? Would it be a magic string or would you add a model router option to the interface? Would you still honor temp and top_p in that case?

packages/core/src/core/subagent.ts
```diff
@@ -0,0 +1,588 @@
...
+ */
+export enum SubagentTerminateMode {
+  /**
+   * Indicates that the subagent's execution terminated due to an unrecoverable error.
+   */
+  ERROR = 'ERROR',
+  /**
+   * Indicates that the subagent's execution terminated because it exceeded the maximum allowed working time.
+   */
+  TIMEOUT = 'TIMEOUT',
```
allenhutchison, line 43: If you take my suggestion from below we may want to add to this enum for other termination mechanisms.

packages/core/src/core/subagent.ts
```diff
@@ -0,0 +1,588 @@
...
+/**
+ * Configures the prompt and expected behavior of the subagent.
+ * This interface defines the essential parameters that guide the subagent's
+ * interaction and task execution, including its plan, goals, and available tools.
+ */
+export interface PromptConfig {
+  /** A high-level plan or strategy for the subagent to follow. */
+  plan: string;
+  /** The specific goals the subagent is expected to achieve. */
+  goals: string;
```
allenhutchison: I think you might be better off and more flexible in defining this as a single system prompt string, and an array of chat pairs (user, model). That would allow for any use case where someone wants to setup a subagent and keeps this interface consistent with what you see if you are using the SDKs natively.

packages/core/src/core/subagent.ts
```diff
@@ -0,0 +1,588 @@
...
+  /** A list of expected output objects and the variables they should emit. */
+  outputs: Record<string, string>;
```
allenhutchison: outputs feels a little strange defined in something called PromptConfig. Maybe split into a seperate interface?

packages/core/src/core/subagent.ts
```diff
@@ -0,0 +1,588 @@
...
+  /** A list of tool names (in the tool registry) or full function declarations that the subagent is permitted to use. */
+  tools: Array<string | FunctionDeclaration>;
```
allenhutchison, line 97: I suspect ToolConfig is an interface you will want to split out as well to allow for changes over time.

packages/core/src/core/subagent.ts
```diff
@@ -0,0 +1,588 @@
...
+   * @param modelConfig - Configuration for the generative model parameters.
+   * @param runConfig - Configuration for the subagent's execution environment and constraints.
+   */
+  private constructor(
+    readonly runtimeContext: Config,
+    private readonly promptConfig: PromptConfig,
+    private readonly modelConfig: ModelConfig,
+    private readonly runConfig: RunConfig,
+  ) {
+    this.subagentId = Math.random().toString(36).slice(2);
```
allenhutchison: For debugging purposes you might want to have the name of the sub-agent in here along with the random string.

packages/core/src/core/subagent.ts
```diff
@@ -0,0 +1,588 @@
...
+    for (const toolName of toolsToLoad) {
+      const tool = toolRegistry.getTool(toolName);
+      if (tool) {
+        const confirmationDetails = await tool.shouldConfirmExecute(
+          {},
+          new AbortController().signal,
+        );
+        if (confirmationDetails) {
+          throw new Error(
+            `Tool "${toolName}" requires user confirmation and cannot be used in a non-interactive subagent.`,
```
allenhutchison: I totally understand starting here, but we really need to define a way for sub-agents to request a confirmation from the UX.

packages/core/src/core/subagent.ts
```diff
@@ -0,0 +1,588 @@
...
+
+ Here are your goals for this session. Use the tools available to attempt to achieve these goals:
+ <GOALS>
+ ${templated_goals}
+ </GOALS>
+ 
+ Important things:
+ * You are running in non-interactive mode. You cannot ask the user for input.
+ * You must determine if your goals are complete. Once you believe all goals have been met, do not call any more tools.
+ `;
```
allenhutchison: ok, this answers my questions about the setup for the PromptConfig, but I think we need to revisit this. because this system prompt wouldn't be appropriate for all of the sub-agent tasks we can envision. For example a summarization sub-agent that doesn't have any tool calls. I think for this to be successful, we need to provide developers with more flexibility around this prompt.

allenhutchison (commented):
Mostly focused on subagent.ts very consistent with what I was hoping for for an agent factory. Looking forward to getting this in.

mattKorwel:
👋 As a part of the on call shift I'm trying to clean up old PR's. I know this one is internal but there hasn't neen any movment in more than 3 weeks. I see the feature attached is linked to the new public roadmap which is 🎉 

@abhipatel12 @mainroach do we have any plans on when and how to move forward with this?

sxthunder:
anything new?

packages/core/src/core/subagent.ts
abhipatel12: Agreed!

packages/core/src/core/subagent.ts
abhipatel12: That makes sense! I made this possible via the properties available in the prompt config.

packages/core/src/core/subagent.ts
abhipatel12: Sounds good, Added a max turns

packages/core/src/core/subagent.ts
abhipatel12: Done.

packages/core/src/core/subagent.ts
abhipatel12: Done

packages/core/src/core/subagent.ts
abhipatel12: Done

packages/core/src/core/subagent.ts
abhipatel12: Good point. In the future, we may want to use a string like 'auto` to enable routing. From a user perspective, we would default to a model unless provided.

packages/core/src/core/subagent.ts
abhipatel12: I had considered this as something to add in a followup. I added turns and can add tokens in a follow-up!

packages/core/src/core/subagent.ts
abhipatel12: Done

abhipatel12:
Thanks for the feedback @allenhutchison! Made some changes. Let me know what you think!

NTaylorMullen:
Fast follows after playing with this locally: 
- Make argument passing cleaner:
image
- Stream sub-agent interactions to the resultDisplay so user can see the progress

packages/core/src/core/subagent.ts
allenhutchison: sounds good with the todo. I would encourage you to avoid a magic string and have a bool for model routing but otherwise lgtm.

packages/core/src/core/subagent.ts
allenhutchison: sounds good. Thanks for adding max turns.

allenhutchison (approved):
This looks great, I can't wait to start using it.

allenhutchison:
I think this is a great addition. Thanks for addressing all the comments. I agree with @NTaylorMullen suggestion for a fast follow on output display in interactive mode for subagents, and we also need to fast follow with a mechanism for subagents to use confirmation-based tools as well. Neither of those are blockers to getting this in though.

packages/core/src/core/subagent.ts
abhipatel12: Sounds good. Once we have routing, we can add the bool

abhipatel12:
Awesome, thanks for the review!

PR #1817, opened 2025-06-26
add Error Handling for prevents crashing mid-process

phattarin-kitbumrung:
Error Handling for copyFileSync
Prevents crashing mid-process:


ricardochi095 (approved):

NTaylorMullen:
Thanks for the contribution @phattarin-kitbumrung ! Unfortunately this is actually an intentional crash. The script you changed is tied to our build pipeline which ensures that we copy all necessary files for publication. In the case that one of those copies fails we want to fail aggressively.

PR #1822, opened 2025-06-26
fix(cli): add trailing space on context completion

Jacky040124:
## TLDR

This PR fixes a minor but annoying UI issue where a space was not automatically appended after selecting a context file with the @ prefix. This change adds the trailing space, improving the user's typing flow.

## Dive Deeper

When using the @ command to reference a file or directory for context, the user is presented with an autocomplete
suggestion list. Upon selecting a suggestion (e.g., with the Tab key), the path was inserted into the input prompt, but the cursor was placed immediately after the path. The user would then have to manually press the spacebar before continuing to type their message.

This change modifies the handleAutocomplete function in InputPrompt.tsx to append a single space character (' ') to the suggestion value when it is inserted, creating a smoother and more intuitive user experience.

## Reviewer Test Plan

To validate this change, a reviewer can perform the following steps:
**1.** Pull down this branch and run the CLI using npm start.
**2.** In the input prompt, type @ followed by the first few letters of a file or directory in your current project to trigger the autocomplete suggestions.
**3.** Select one of the suggestions using the Tab or Enter key.
**4.** Verify: The full path is inserted into the prompt, immediately followed by a single space, with the cursor positioned after that space, ready for further typing.

## Testing Matrix

**Note to Reviewers:** Apologies for not being able to fill most of these out, I am new to open source and I only have a mac and do not have linux environment set up, I will proceed to test as much as possible :)

"I have run and passed all the following checks for this PR:
- **Unit tests** (npm run test)
- **Integration tests** (npm run test:e2e)
- **Preflight checks** (npm run preflight)
- **Code formatting** (npm run format)
- **Linting** (npm run lint)

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ✅  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

https://github.com/google-gemini/gemini-cli/issues/1638

### DIFF
--- a/packages/cli/src/ui/components/InputPrompt.tsx
+++ b/packages/cli/src/ui/components/InputPrompt.tsx
@@ -141,7 +141,7 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
         buffer.replaceRangeByOffset(
           autoCompleteStartIndex,
           buffer.text.length,
-          selectedSuggestion.value,
+          selectedSuggestion.value + ' ',
         );
       }
       resetCompletionState();


Jacky040124:
> Thanks for your pull request! It looks like this may be your first contribution to a Google open source project. Before we can look at your pull request, you'll need to sign a Contributor License Agreement (CLA).
> 
> View this [failed invocation](https://github.com/google-gemini/gemini-cli/pull/1822/checks?check_run_id=44828145271) of the CLA check for more information.
> 
> For the most up to date status, view the checks section at the bottom of the pull request.

I've just submitted my CLA!!

NTaylorMullen:
This seems super reasonable, @jacob314 / @dewitt / @allenhutchison do you recall why we didn't do this in the past?

jacob314:
This makes a lot of sense for completing tool names so we should absolutely do it for that case. It also makes sense for completing file names but it doesn't make sense for completing directory names.
For example, if I accept the completion to complete
`@packages/` to `@packages/cli` I don't expect a space because I would like to continue adding further completions to the directory.

The simplest way to do this is to probably add a trailing space to all the completion entries where it makes sense but avoid adding the space for the case of directories where it doesn't. We could also return some additonal metadata with completions but that seems more complicated for no benifit I can think of.

mattKorwel:
Hi there @Jacky040124, I'm following up on some older dangling pull requests. I'm going to close this one one but if you're wanting to address @jacob314's comments above please feel free to repoen and update. Eitiher way thank you for the contribution!

Issue #1938, opened 2025-06-26
git checkout without permission


pmwkaa:
### What happened?

Not sure if that should be marked as a bug or a feature, or the way how it should be used.

I was trying it out combining manual changes in the C project and requests to gemini.  The request was: `make release and fix compilation warnings` for a C project.

It asked to run every command and did some useful work in a loop. Suddenly at some point it said that its last command made some unnecessary changes to the repository and automatically executed `git checkout` resetting all uncommitted changes in the repository 😄 

### What did you expect to happen?

I believe it should ask permission to run any command, especially those which could lead to a work/data loss.

### Client information

Sorry, not kept the output.

sahil9001:
Hey @pmwkaa , I have created a PR for this, please check.

pmwkaa:
@sahil9001 awesome, thanks

jacob314:
Lowering the priority under the assumption that you had approved all git commands.
I believe the path forward to solving this is 
https://github.com/google-gemini/gemini-cli/issues/2417
please comment on that issue with any suggestions you have on how this should be implemented.

bdmorgan:
Closing this for now based on the PR and @jacob314 's assumption. Please reply back and/or ping @jacob314 or myself directly and we can dig in deeper.

PR #1947, opened 2025-06-26
fix: always confirm dangerous shell commands

sahil9001:
## Problem
Previously, the CLI would execute potentially destructive shell commands such as git checkout, git reset, or rm without always prompting the user for confirmation. Once a root command (like git) was whitelisted, all subcommands—including dangerous ones—could be executed without further checks. This could lead to accidental data loss, as reported in Issue #1938 

## Solution
This PR introduces stricter confirmation logic for dangerous shell commands:
- A list of dangerous command patterns (e.g., git checkout, git reset, rm, mv, dd, etc.) is now maintained.
- Any command matching a dangerous pattern will always prompt the user for explicit confirmation, regardless of previous approvals or whitelisting.
- Whitelisting is still supported for non-dangerous commands, so safe commands (like ls or echo) can be approved for repeated use without further prompts.
- The confirmation dialog for dangerous commands is clearly labeled to indicate the risk.

### DIFF
--- a/packages/core/src/tools/shell.ts
+++ b/packages/core/src/tools/shell.ts
@@ -29,6 +29,16 @@ import { spawn } from 'child_process';
 
 const OUTPUT_UPDATE_INTERVAL_MS = 1000;
 
+const DANGEROUS_COMMAND_PATTERNS = [
+  /^git\s+checkout\b/,
+  /^git\s+reset\b/,
+  /^git\s+clean\b/,
+  /^rm\b/,
+  /^mv\b/,
+  /^dd\b/,
+  // Add more as needed
+];
+
 export class ShellTool extends BaseTool<ShellToolParams, ToolResult> {
   static Name: string = 'run_shell_command';
   private whitelist: Set<string> = new Set();
@@ -121,6 +131,12 @@ export class ShellTool extends BaseTool<ShellToolParams, ToolResult> {
     return null;
   }
 
+  isDangerousCommand(command: string): boolean {
+    return DANGEROUS_COMMAND_PATTERNS.some((pattern) =>
+      pattern.test(command.trim()),
+    );
+  }
+
   async shouldConfirmExecute(
     params: ShellToolParams,
     _abortSignal: AbortSignal,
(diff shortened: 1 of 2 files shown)


packages/core/src/tools/shell.ts
```diff
@@ -29,6 +29,16 @@ import { spawn } from 'child_process';
...
+const DANGEROUS_COMMAND_PATTERNS = [
+  /^git\s+checkout\b/,
+  /^git\s+reset\b/,
+  /^git\s+clean\b/,
+  /^rm\b/,
+  /^mv\b/,
+  /^dd\b/,
+  // Add more as needed
+];
+
```
crazylemon8, line 41: @sahil9001 
I feel that rather than hardcoding these values, these should be picked up from the config, as someone may have a different use case or might want to add their own dangerous patterns.

sahil9001: Do you mean from `.gemini/config.yaml` file?

crazylemon8: Not really in config, but rather in settings.json
Currently storing things like theme, authType, fileFiltering etc, we can add this there only. In code i can see it's currently handled in `packages/cli/src/config/settings.ts`

NTaylorMullen: Ya I feel like this would be better handled via some variant of: https://github.com/google-gemini/gemini-cli/issues/2417

i.e. an ability to provide patterns for what sort of commands to auto-accept or not.

aspiers: IMHO maintaining a blocklist of dangerous commands is not the way to go, and is just asking for trouble.  There are an infinite number of dangerous commands.  Instead, it should be an allowlist of known safe commands.  And yes, this should definitely be user-configurable and support regex pattern matching.

sahil9001: @aspiers @NTaylorMullen @crazylemon8 Can we narrow down on a single approach here? Currently I feel a lot of noise, IMO what @aspiers says is reasonable to do here.

crazylemon8: I agree that looks like a better approach +1

palladius:
So what do we do here? 
There seems to be consensus on changing approach.

[sahil9001](https://github.com/sahil9001) ,  Should we close this and move to https://github.com/google-gemini/gemini-cli/issues/2417 and corroborate into those PRs?

sahil9001:
Sure @palladius , that works with me!

scottdensmore:
Closing in favor of #2417

PR #1949, opened 2025-06-26
fix: character encoding issues in shell command processor

### LINKED ISSUE #1945: Shell Command Output with Chinese Characters Displays as Garbage
### What happened?

When running shell commands using gemini-cli, if the command output contains Chinese characters, the output appears garbled or corrupted due to encoding issues.

Image

Image

The shell command output should correctly display Chinese characters without any garbled or unreadable text.

### Client information

<details>

```console
$ gemini /about
│ CLI Version        0.1.4                               │
│ Git Commit         1078a54                             │
│ OS                 win32                               │
```

</details>

### Login information

_No response_

_No response_

boylin0:
Fixes #1945

- Add system encoding detection for Windows (chcp) and Unix-like systems (locale)
- Replace StringDecoder with TextDecoder for better multi-byte character support
- Add comprehensive Windows code page to encoding mappings
- Remove deprecated decoder.end() calls that are not needed with TextDecoder

This resolves issues with garbled output when shell commands produce non-UTF-8 encoded text, particularly on Windows systems with different code pages.

## TLDR

This PR improves the handling of shell command outputs with non-UTF-8 encodings by dynamically detecting system encoding on Windows and Unix-like OSes. It replaces `StringDecoder` with `TextDecoder` to properly decode multi-byte characters such as Chinese, fixing the garbled text problem.

## Dive Deeper

Previously, shell command outputs containing Chinese or other non-UTF-8 characters appeared corrupted due to improper decoding. The original implementation used `StringDecoder`, which does not support Big5 or other non-UTF-8 encodings well. On Windows, code pages vary and can be detected via `chcp`, while on Unix-like systems, `locale` provides the encoding info. By integrating these detection methods and switching to `TextDecoder` (which supports multi-byte decoding natively), this change ensures the correct interpretation of command outputs across different environments. Deprecated calls to `decoder.end()` were removed because `TextDecoder` does not require them.

## Reviewer Test Plan

1. Run shell commands via `gemini` on Windows with a code page like 950 (Big5) or 936 (GBK) that outputs Chinese characters, and verify the output is displayed correctly without garbled text.  
2. Repeat similar tests on Unix/Linux/macOS with locales that produce Chinese or other multi-byte output.  
3. Confirm that no regressions occur with UTF-8 encoded outputs.  
4. Review the code for correct usage of system encoding detection and proper replacement of decoding logic.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ✅  | ✅  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs
#1945: Garbled output from shell commands containing Chinese characters due to encoding issues


chaizp:
Today I found that when entering Chinese, only the last character will be displayed in the terminal, and the way of copying text will also only show the last character....

packages/cli/src/ui/hooks/shellCommandProcessor.ts
```diff
@@ -96,8 +153,8 @@ function executeShellCommand(
 
       const decodedChunk =
         stream === 'stdout'
-          ? stdoutDecoder.write(data)
-          : stderrDecoder.write(data);
+          ? stdoutDecoder.decode(data)
+          : stderrDecoder.decode(data);
```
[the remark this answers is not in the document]
boylin0: The original code already uses stripAnsi to remove ANSI escape codes after decoding. Since this logic is already in place, I don't think additional changes are necessary here.

packages/cli/src/ui/hooks/shellCommandProcessor.ts
```diff
@@ -154,9 +237,6 @@ function executeShellCommand(
       exited = true;
       abortSignal.removeEventListener('abort', abortHandler);
 
-      // Handle any final bytes lingering in the decoders
```
jacob314, line 162: is there a reason this case is removed?

packages/cli/src/ui/hooks/shellCommandProcessor.ts
```diff
@@ -66,8 +148,9 @@ function executeShellCommand(
     });
 
     // Use decoders to handle multi-byte characters safely (for streaming output).
-    const stdoutDecoder = new StringDecoder('utf8');
-    const stderrDecoder = new StringDecoder('utf8');
+    const systemEncoding = getSystemEncoding();
```
jacob314: can we cache the system encoding and only perform the logic to determine it once rather than determining it fresh for each chunk of output

packages/cli/src/ui/hooks/shellCommandProcessor.ts
```diff
@@ -22,6 +22,88 @@ import stripAnsi from 'strip-ansi';
...
+    1257: 'windows-1257',
+    1258: 'windows-1258',
+    65001: 'utf-8'
+  };
+
+  if (map[cp]) {
+    return map[cp];
+  }
+
+  console.warn(`Unknown Windows code page: ${cp}. Falling back to utf-8.`);
```
jacob314: I wonder if we should consider using chardet (https://www.npmjs.com/package/chardet) to autodetect the encoding at least for cases where we were not able to determine the encoding. It could be overkill.

jacob314 (commented):
Thank you for the pull request! This is a very important polish item.

packages/cli/src/ui/hooks/shellCommandProcessor.ts
boylin0: After reading through the [Encoding Standard for TextDecoder](https://encoding.spec.whatwg.org/#interface-textdecoder), I realized I missed handling the end-of-stream properly. I initially thought it wasn’t necessary, but now I’m addressing that.

packages/cli/src/ui/hooks/shellCommandProcessor.ts
boylin0: Great idea! I’ve integrated chardet as a fallback to autodetect encoding when we can’t determine it upfront.

packages/cli/src/ui/hooks/shellCommandProcessor.ts
boylin0: Sure! I’ve added caching so the system encoding is determined just once, improving efficiency.

packages/cli/src/ui/hooks/shellCommandProcessor.ts
```diff
@@ -18,10 +18,128 @@ import path from 'path';
...
+ * Returns the cached system encoding it if not cached.
+ * This function is used to ensure we only detect the system encoding
+ * by calling `getSystemEncoding()` or `detectEncodingFromBuffer()` once,
+ * and then reuse the cached value for subsequent calls.
+ * @param buffer A buffer to use for detecting the system encoding.
+ */
+function getCachedSystemEncoding(buffer: Buffer) {
+  if (cachedSystemEncoding === null) {
+    // If the system encoding detection fails, we fallback to chardet.
+    cachedSystemEncoding = getSystemEncoding() || detectEncodingFromBuffer(buffer);
```
jacob314: I'm not sure detecting from the buffer should be cached. that might be appropriate to run on each new buffer as the encoding might not be clear until characters that don't match utf8 are encountered. Definitely worth some tests to see what the behavior is for various inputs.

packages/cli/src/ui/hooks/shellCommandProcessor.ts
```diff
@@ -18,10 +18,128 @@ import path from 'path';
...
+      const output = execSync('chcp', { encoding: 'utf8' });
+      const match = output.match(/:\s*(\d+)/);
+      if (match) {
+        const codePage = parseInt(match[1], 10);
+        if (!isNaN(codePage)) {
+          return windowsCodePageToEncoding(codePage);
+        }
+      }
+    } catch (_e) {
+      console.warn('Failed to get Windows code page.');
```
jacob314: make the message a bit clearer as users will see it in the console.
The error should make it clear how the user should expect things may go wrong and should perhaps log what output was so we can get bug reports that result in actionable fixes to windowsCodePageToEncoding

packages/cli/src/ui/hooks/shellCommandProcessor.ts
```diff
@@ -155,8 +280,12 @@ function executeShellCommand(
       abortSignal.removeEventListener('abort', abortHandler);
 
       // Handle any final bytes lingering in the decoders
-      stdout += stdoutDecoder.end();
-      stderr += stderrDecoder.end();
+      if (stdoutDecoder) {
+        stdoutDecoder.decode();
```
jacob314: why is this decode instead of end()?
also why isn't the content added to the stdout and stderr?

jacob314 (commented):
approved after these comments are addressed and some tests are added. If tests mock the output of the commands to enable them to run on platforms other than windows that is fine.

Thanks again for this! This will really help users who aren't using utf8.

packages/cli/src/ui/hooks/shellCommandProcessor.ts
boylin0: Makes sense — I’ve removed the cache and now run detection on each call in [getCachedEncodingForBuffer()](https://github.com/google-gemini/gemini-cli/pull/1949/files#diff-e48aa6b68e687b08e91b099cb9830dfeefd42ca7cd8a76657d97bbdf7157afaaR28-R41). Appreciate the suggestion, definitely a more reliable approach.

packages/cli/src/ui/hooks/shellCommandProcessor.ts
boylin0: Good point I updated the error message in [windowsCodePageToEncoding()](https://github.com/google-gemini/gemini-cli/pull/1949/files#diff-e48aa6b68e687b08e91b099cb9830dfeefd42ca7cd8a76657d97bbdf7157afaaR63-R72).

packages/cli/src/ui/hooks/shellCommandProcessor.ts
boylin0: Thanks for pointing it out! But TextDecoder doesn’t have an end() method. Per the [WHATWG spec#interface-textdecoder](https://encoding.spec.whatwg.org/#interface-textdecoder) example, we use decode(chunk, { stream: true }) for each chunk, and a final decode() to flush any remaining bytes.

As for the second point you're right! I missed appending the decoded output to stdout and stderr. I've updated the code .

boylin0:
> approved after these comments are addressed and some tests are added. If tests mock the output of the commands to enable them to run on platforms other than windows that is fine.
> 
> Thanks again for this! This will really help users who aren't using utf8.

@jacob314 Thanks so much for the thorough code review and your patience! All comments have been addressed, and I’ve added tests with mocked command output to ensure cross-platform compatibility. Please feel free to let me know if you spot any other issues or have further suggestions.

jacob314 (approved):
lgtm
Great work! Thank you for sticking with this pull request.

PR #1972, opened 2025-06-26
Handle stdin for prompts using readline for escape character parsing

bbiggs:
## TLDR

Switch out input handling for the prompt from ink's useInput to a new hook that uses readline to parse character sequences.

## Dive Deeper

The prompt requires careful handling of input character sequences for broad compatibility complex editing of prompts. Currently ui/components/shared/text-buffer.ts is growing to include particular handling (ex: VSCode's shift-enter code) that seems better suited for a lower layer. This change introduces a new hook, useKeypress, affording a location in the code to handle character sequences. To keep this small and robust, we delegate to node's readline library to handle the basics of decoding terminal escape sequences.

This cherrypicks the following PR back into main:

 https://github.com/google-gemini/gemini-cli/pull/1252

## Linked issues / bugs

Fixes #1212

### DIFF
--- a/packages/cli/src/ui/App.tsx
+++ b/packages/cli/src/ui/App.tsx
@@ -63,6 +63,7 @@ import {
   useSessionStats,
 } from './contexts/SessionContext.js';
 import { useGitBranchName } from './hooks/useGitBranchName.js';
+import { useBracketedPaste } from './hooks/useBracketedPaste.js';
 import { useTextBuffer } from './components/shared/text-buffer.js';
 import * as fs from 'fs';
 import { UpdateNotification } from './components/UpdateNotification.js';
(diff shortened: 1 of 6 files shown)


jacob314 (dismissed):
lgtm
this fixes some major bugs in our key handling and puts it on a solid foundation for other future work such as supporting deleting words forward and backward with delete and backspace.
The one question is how do we want to roll out a risky change like this now that we are live. Potentially we need to put this behind a setting and have users dogfood it for a week before we make it the default or maybe there is an easier way.

jacob314:
Crowd sourced testing this.
We have two issues we need to fix before landing. Aside from that seems very solid.
The following two issues reproduce for me on linux and mac.
`!` no longer switches to shell mode
`option enter` now submits instead of a newline

jacob314 (approved):
lgtm
Tested locally and all remaining bugs are fixed.

allenhutchison (approved):

PR #1988, opened 2025-06-26
feat: add initial extensions.json for VSCode recommendations

Lethabu:


NTaylorMullen:
@Lethabu appreciate the contribution and the suggestion. I think for now we're going to leave this area blank BUT appreciate you taking the time.

PR #2024, opened 2025-06-26
feat(cli/ui): Add status command for configuration and connectivity check

### LINKED ISSUE #1941: feat(cli): Add `/status` command for configuration and connectivity check

  I propose adding a new slash command: /status.

  This command would provide a quick summary of the current configuration and
  test the connection to Google AI, making it easier for users to self-diagnose
  and troubleshoot common problems.

  The output of the /status command would look something like this:

```
--- Gemini CLI Status ---
   - CLI Version:    1.0.0
   - OS:             darwin
   - Model:          gemini-1.5-flash
   - Authentication: gcloud CLI
 
 --- Connectivity ---
   - Status: ✅ Success
   - Details: Successfully connected to Google AI services.
```

  If there's a connection error, it would display a helpful message:

```
 --- Connectivity ---
   - Status: ❌ Failed
   - Reason: [Error message from the API call]
```

### Why is this needed?

Currently, when users encounter issues with gemini-cli, it can be difficult to determine if the problem stems from a misconfiguration (e.g., incorrect auth settings, wrong model name) or a connectivity issue with the Google AI services. There is no built-in command to quickly diagnose the current setup.

I've investigated the existing startup checks in packages/cli/src/config/auth.ts. The current validateAuthMethod function primarily checks for the presence of environment variables (e.g., GEMINI_API_KEY) required for a given authentication method. The proposed /status command complements this by performing a more comprehensive, on-demand check:

1. On-Demand Verification: Unlike the startup validation, /status can be run at any time by the user to diagnose issues without restarting the CLI.
2. Dynamic Connectivity Test: Crucially, /status will not just check if a variable is set. It will actively attempt to connect to the Google AI service (e.g., by calling config.refreshAuth()) to confirm that the credentials are not only present but also valid and working.
3. Configuration Display: It provides a clear summary of the currently active configuration (like the model name), which is not part of the initial validation logic.

Therefore, this command provides a distinct and valuable diagnostic capability that does not overlap with the existing startup checks.

### Additional context

_No response_

iamrajiv:
## TLDR

Fixes: #1941

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

### DIFF
--- a/packages/cli/src/ui/hooks/slashCommandProcessor.test.ts
+++ b/packages/cli/src/ui/hooks/slashCommandProcessor.test.ts
@@ -336,6 +336,132 @@ describe('useSlashCommandProcessor', () => {
     });
   });
 
+  describe('/status command', () => {
+    it('should display configuration and connectivity status with successful connection', async () => {
+      mockGetCliVersionFn.mockResolvedValue('1.0.0');
+      const mockGenerateContent = vi.fn().mockResolvedValue({ 
+        candidates: [{ content: { parts: [{ text: 'Hi' }] } }] 
+      });
+      mockConfig = {
+        ...mockConfig,
+        getModel: vi.fn(() => 'gemini-1.5-flash'),
+        getContentGeneratorConfig: vi.fn(() => ({
+          authType: 'oauth-personal',
+        })),
+        getGeminiClient: vi.fn(() => ({
+          generateContent: mockGenerateContent,
+        })),
+      } as unknown as Config;
+
+      const { handleSlashCommand } = getProcessor();
+
+      let commandResult: SlashCommandActionReturn | boolean = false;
+      await act(async () => {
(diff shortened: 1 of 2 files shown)


NTaylorMullen:
Thank you for your contribution, @umairidris! Appreciate you taking the time to work on this.

After some internal discussion, we've decided to close this pull request. The functionality provided by the `/status` command, while useful, overlaps significantly with our existing `/about` and `/auth` commands. We're trying to keep the number of slash commands focused to avoid confusion, and we believe the current commands already cover the necessary diagnostic information.

Grateful for your effort and hope you'll continue to contribute to the project!

PR #2032, opened 2025-06-26
fix: add repository field to package.jsons

KeijiBranshi:
Our release publishing trigger needs this metadata to be set in the package.jsons. 

Drive by change: point the source `sandboxImageUri` to the image name used for local sandbox development (note, this gets updated to a production sandbox in the release pipeline). Added extra steps to copy README and LICENSE into the core package.


scottdensmore (approved):
Approved

PR #2059, opened 2025-06-27
Add privacy notice slash command

scidomino:
## TLDR

A /privacy command that brings up a different privacy notice depending on which auth system you are using. 

## Dive Deeper

Users with cloud free tier are also given an interface to opt out of data collection.

## Reviewer Test Plan

Test with:
* Gemini Key
* Vertex AI Key
* Oauth personal account
* Verify that  Oauth personal account opt in setting can be changed and persists between executions.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | x  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

### DIFF
--- a/packages/cli/src/ui/hooks/usePrivacySettings.ts
+++ b/packages/cli/src/ui/hooks/usePrivacySettings.ts
@@ -1,0 +1,135 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { GaxiosError } from 'gaxios';
+import { useState, useEffect, useCallback } from 'react';
+import { Config, CodeAssistServer, UserTierId } from '@google/gemini-cli-core';
+
+export interface PrivacyState {
+  isLoading: boolean;
+  error?: string;
+  isFreeTier?: boolean;
+  dataCollectionOptIn?: boolean;
+}
+
+export const usePrivacySettings = (config: Config) => {
+  const [privacyState, setPrivacyState] = useState<PrivacyState>({
+    isLoading: true,
+  });
+
+  useEffect(() => {
+    const fetchInitialState = async () => {
+      setPrivacyState({
+        isLoading: true,
+      });
+      try {
+        const server = getCodeAssistServer(config);
+        const tier = await getTier(server);
+        if (tier !== UserTierId.FREE) {
+          // We don't need to fetch opt-out info since non-free tier
+          // data gathering is already worked out some other way.
+          setPrivacyState({
+            isLoading: false,
+            isFreeTier: false,
+          });
+          return;
+        }
+
+        const optIn = await getRemoteDataCollectionOptIn(server);
+        setPrivacyState({
+          isLoading: false,
+          isFreeTier: true,
+          dataCollectionOptIn: optIn,
+        });
+      } catch (e) {
+        setPrivacyState({
+          isLoading: false,
+          error: e instanceof Error ? e.message : String(e),
+        });
+      }
+    };
+    fetchInitialState();
+  }, [config]);
+
+  const updateDataCollectionOptIn = useCallback(
+    async (optIn: boolean) => {
+      try {
+        const server = getCodeAssistServer(config);
+        const updatedOptIn = await setRemoteDataCollectionOptIn(server, optIn);
+        setPrivacyState({
+          isLoading: false,
+          isFreeTier: true,
+          dataCollectionOptIn: updatedOptIn,
+        });
+      } catch (e) {
+        setPrivacyState({
+          isLoading: false,
+          error: e instanceof Error ? e.message : String(e),
+        });
+      }
+    },
+    [config],
+  );
+
+  return {
+    privacyState,
+    updateDataCollectionOptIn,
+  };
+};
+
+function getCodeAssistServer(config: Config): CodeAssistServer {
+  const server = config.getGeminiClient().getContentGenerator();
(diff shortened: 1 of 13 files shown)


packages/cli/src/ui/hooks/usePrivacySettings.ts
```diff
@@ -0,0 +1,121 @@
...
+
+export const usePrivacySettings = (config: Config) => {
+  const [dataCollectionOptIn, setDataCollectionOptIn] = useState<
+    boolean | undefined
+  >(undefined);
+  const [isFreeTier, setIsFreeTier] = useState(false);
+  const [isLoading, setIsLoading] = useState(false);
+  const [error, setError] = useState<string | null>(null);
+
+  const fetchDataCollectionOptIn = useCallback(async () => {
```
jacob314: inline this in the useEffect call on line 42 and add a cleanup return. otherwise if the config does change then this could get in a confusing state of a mixture of values set by the previous request and the new request.

packages/cli/src/ui/hooks/usePrivacySettings.ts
```diff
@@ -0,0 +1,121 @@
...
+    isFreeTier,
+    dataCollectionOptIn,
+    updateDataCollectionOptIn,
+  };
+};
+
+function getCodeAssistServer(config: Config): CodeAssistServer {
+  const server = config.getGeminiClient().getContentGenerator();
+  if (!(server instanceof CodeAssistServer)) {
+    throw new Error('Oauth not being used');
```
jacob314, line 87: lets avoid throwing exceptions as part of things operating normally. Can you make the return value nullable instead if it is typical for this to throw an exception if not using the CodeAssistServer.
If we start throwing exceptions as part of normal operation in the code base then it will be harder for people trying to debug gemini cli with pause on exception.

packages/cli/src/ui/privacy/CloudFreePrivacyNotice.tsx
```diff
@@ -0,0 +1,96 @@
...
+            onSelect={(value) => {
+              updateDataCollectionOptIn(value);
+              onExit();
+            }}
+          />
+        </Box>
+      )}
+      <Newline />
+      <Text>
+        <Text color="red">[1]</Text> https://policies.google.com/privacy
```
jacob314: use Colors. instead of hard coding "red" to leverage the theme.

packages/cli/src/ui/privacy/CloudPaidPrivacyNotice.tsx
```diff
@@ -0,0 +1,51 @@
...
+  });
+
+  return (
+    <Box flexDirection="column" marginBottom={1}>
+      <Box justifyContent="center">
+        <Text bold>Vertex AI Notice</Text>
+      </Box>
+      <Newline />
+      <Text>
+        Service Specific Terms<Text color="red">[1]</Text> are incorporated into
```
jacob314: use Colors. for the theme.
We should add instructions to GEMINI.md to to always use the that for colors so that we leverage the theme. Same for all other colors in this file as well

jacob314: please apply to the other files as well.

packages/cli/src/ui/hooks/usePrivacySettings.ts
jacob314: For discussion: this is somewhat tightly coupled to parts of core that would ideally be implementation details in the cli package. Is there a reasonable way we could make this less tightly coupled? One idea is we have good markdown rendering support in the cli package so the core package could have a generic method to return a privacy notice as markdown that the cli package renders rather than this code getting tightly coupled to implementation details of the core package.

jacob314 (commented):
done with a round of comments. It will be great to have this functionality!
If it is too hard to refactor this so that core and cli are not tightly coupled for this then I'm fine doing it this way just want to make sure there isn't something clean we can do,

packages/cli/src/ui/hooks/usePrivacySettings.ts
scidomino: This really shouldn't ever happen though. This class is only loaded by the CloudFreePrivacyNotice which is only displayed if the auth mode is oauth.

packages/cli/src/ui/hooks/usePrivacySettings.ts
scidomino: Unfortunately, tight coupling is unavoidable because of the opt-out form we want to show the free tier users. Otherwise, I totally agree it would have been better to push this logic into core.

packages/cli/src/ui/hooks/usePrivacySettings.ts
jacob314: ok sounds good. Superficially looking at the code it seemed like it might throw as part of routine operations.
Perhaps add a commend indicating that this shouldn't happen so that is clearer.

jacob314 (commented):
approved after addressing the one open robustness comment.
lgtm

packages/cli/src/ui/hooks/usePrivacySettings.ts
scidomino: Done.

allenhutchison (approved):
This is great. Thank you.

PR #2096, opened 2025-06-27
Upgrade to Ink 6 and React 19

SandyTao520:
## TLDR

Upgraded to Ink 6 and React 19, including cleanup of temporary hacks and revised tool call update logic in useGeminiStream

## Dive Deeper

Some issues introduced with React 19:

1. useRef gets cleaned up too early in useEffect, causing ``useConsoleMessages`` to break, so we clone the ``messageQueueRef`` array to make sure it's available.
2. Due to React 19's stricter batching behavior, ``toolCallsUpdateHandler`` updates ``toolCallsForDisplay`` state, but the ``useEffect`` in ``useGeminiStream.ts`` that depends on toolCalls may not be triggering when completed tool calls are cleared from the scheduler. The solution is to handle completed tool calls in ``onComplete`` of ``useReactToolScheduler``.
3. ReactNode.element.props is now of type unknown instead of any, we need to manually cast it to appropriate type.

## Reviewer Test Plan

Use this branch to test internally for some days to see there are more hidden problems.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  |  ✅ | ❓  | ❓  |
| npx      | ✅  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ✅  | -   | -   |

## Linked issues / bugs

N/A


packages/cli/src/ui/App.tsx
```diff
@@ -520,23 +508,6 @@ const App = ({ config, settings, startupWarnings = [] }: AppProps) => {
...
-    );
-
-    // If our pending history item happens to exceed the terminal height we will most likely need to refresh
-    // our static collection to ensure no duplication or tearing. This is currently working around a core bug
-    // in Ink which we have a PR out to fix: https://github.com/vadimdemedes/ink/pull/717
-    if (pendingItemDimensions.height > availableTerminalHeight) {
-      setStaticNeedsRefresh(true);
-    }
-  }, [pendingHistoryItems.length, availableTerminalHeight, streamingState]);
-
```
KeijiBranshi, line 549: ❤️

packages/cli/src/ui/hooks/useReactToolScheduler.ts
```diff
@@ -64,7 +64,7 @@ export type TrackedToolCall =
   | TrackedCancelledToolCall;
 
 export function useReactToolScheduler(
-  onComplete: (tools: CompletedToolCall[]) => void,
+  onComplete: (tools: CompletedToolCall[]) => void | Promise<void>,
```
KeijiBranshi: nitpick: if this is never `await`ed, should we just keep the return type as `void`?

packages/cli/src/ui/hooks/useReactToolScheduler.ts
```diff
@@ -152,7 +152,7 @@ export function useReactToolScheduler(
   );
 
   const schedule: ScheduleFn = useCallback(
-    async (
+    (
```
jacob314, line 155: was it just a bug that this was ever async?

SandyTao520: yea, no need to mark this method as async

packages/cli/src/ui/hooks/useConsoleMessages.ts
```diff
@@ -25,9 +25,12 @@ export function useConsoleMessages(): UseConsoleMessagesReturn {
       return;
     }
 
+    const newMessagesToAdd = messageQueueRef.current;
+    messageQueueRef.current = [];
+
```
KeijiBranshi, line 30: thought: mentioned offline, but will just put here for readers. I think long term, the more _idiomatic_ way to coordinate sequential state updates like this is to leverage aggregate state objects with [state updaters](https://react.dev/reference/react/useReducer) and/or [useReducers](https://react.dev/reference/react/useReducer).

this would be a much bigger refactor though. so am cool with punting it.

packages/cli/src/ui/components/shared/MaxSizedBox.tsx
```diff
@@ -258,7 +261,10 @@ function visitBoxRow(element: React.ReactNode): Row {
   }
 
   if (enableDebugLog) {
-    const boxProps = element.props;
+    const boxProps = element.props as {
+      children?: React.ReactNode | undefined;
+      readonly flexDirection?: "row" | "column" | "row-reverse" | "column-reverse" | undefined;
+    };
```
KeijiBranshi: nitpick: can this assertion type by `BoxProps`? Or something like `Pick<BoxProps, 'children' | 'flexDirection'>`?

packages/cli/src/ui/components/shared/MaxSizedBox.tsx
```diff
@@ -258,7 +261,10 @@ function visitBoxRow(element: React.ReactNode): Row {
   }
 
   if (enableDebugLog) {
-    const boxProps = element.props;
+    const boxProps = element.props as {
```
KeijiBranshi, line 266: question: should we add an `element.type === Box` check before this?

KeijiBranshi (dismissed):
Thank you so much for tackling this!

packages/cli/src/ui/components/shared/MaxSizedBox.tsx
jacob314: there already is. slightly hidden on line 253

packages/cli/src/ui/hooks/useConsoleMessages.ts
jacob314: we coud put a TODO to perform that refactor in the future. 100% agree it is the right path and would like to encourage Gemini CLI and users reading this code not to duplicate this pattern.

packages/cli/src/ui/hooks/useGeminiStream.ts
```diff
@@ -570,40 +574,33 @@ export const useGeminiStream = (
...
   );
 
-  /**
-   * Automatically submits responses for completed tool calls.
-   * This effect runs when `toolCalls` or `isResponding` changes.
-   * It ensures that tool responses are sent back to Gemini only when
-   * all processing for a given set of tools is finished and Gemini
-   * is not already generating a response.
-   */
-  useEffect(() => {
```
jacob314, line 580: so glad to see this useEffect going away! It was the root of the issue of blocks of whitespace at the bottom of the UI.

jacob314 (dismissed):
Amazing work!
This will make the experience so much better for our users.
lgtm_dog2

allenhutchison (dismissed):

allenhutchison (dismissed):

allenhutchison (approved):

PR #2112, opened 2025-06-27
Add the current auth method and GCP Project config to the about message

### LINKED ISSUE #1992: Add login method and GCP project config to /about

add new fields to /about

### Why is this needed?

The only way you can determine how you are currently logged in is to re-auth

Also - if you have competing local env and .env settings for project, it may not be clear which one CLI is picking up.

### Additional context

_No response_

ptone:
## TLDR

Add two more rows to the about information, to help people better understand their current configuration

## Dive Deeper

Currently the only way to know what auth method you are using is to choose to re-auth.

When there are competing .env files and locally set environment, sometimes it is unclear if the GCP project is set

Screenshot 2025-06-26 at 9 52 45 PM

## Reviewer Test Plan

try the /about command with several different auth options

## Testing Matrix

I've tested this with all three primary auth methods. The GCP project is only displayed when set.

|          | 🍏  | 🪟  | 🐧  |
| npm run  |✅   | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

fixes #1992

### DIFF
--- a/packages/cli/src/ui/components/AboutBox.tsx
+++ b/packages/cli/src/ui/components/AboutBox.tsx
@@ -98,7 +98,9 @@ export const AboutBox: React.FC<AboutBoxProps> = ({
         </Text>
       </Box>
       <Box>
-        <Text>{selectedAuthType}</Text>
+        <Text>
+          {selectedAuthType.startsWith('oauth') ? 'OAuth' : selectedAuthType}
+        </Text>
       </Box>
     </Box>
     {gcpProject && (
--- a/packages/cli/src/ui/components/AboutBox.tsx
+++ b/packages/cli/src/ui/components/AboutBox.tsx
@@ -14,6 +14,8 @@ interface AboutBoxProps {
   osVersion: string;
   sandboxEnv: string;
   modelVersion: string;
+  selectedAuthType: string;
+  gcpProject: string;
 }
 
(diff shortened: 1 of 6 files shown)


packages/cli/src/ui/components/AboutBox.tsx
```diff
@@ -14,13 +14,17 @@ interface AboutBoxProps {
   osVersion: string;
   sandboxEnv: string;
   modelVersion: string;
+  selectedAuthType: string;
```
Adityayadav7459, line 17: selectAuthType: string;

ptone:
@mboshernitsan would you be willing to take a look?

ptone:
Actually - I should update this to change all occurrences of the authType from oauth-personal to just oauth - as that will better reflect the current login options and will be less confusing - may do this at the display layer only (if beginsWith oauth "OAuth")

Adityayadav7459:
Sure please go ahead

scidomino (approved):

scottdensmore (approved):
Tested and approved.

D30554 (approved):

PR #2169, opened 2025-06-27
Update authentication.md

garciasces:
Added commands to configure Gemini CLI in MacOS using the default shell zsh.

## TLDR

The examples provided in the docs didn't work for MacOS users due to different quotes management. I added the equivalent commands for MacOS, so users could set Gemini CLI in an easier way.

## Dive Deeper

Besides the MacOS sections I also added a comment regarding the Vertex AI default location being global, as I spent a lot of time looking for this specific parameter in the Console.

## Reviewer Test Plan

If you are using a Mac, you can execute the different code examples and verify the produce the intended configurations regarding the authentication of the Gemini CLI.

## Testing Matrix

This patch only applies to MacOS using the default shell (zsh)

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

### DIFF
--- a/docs/cli/authentication.md
+++ b/docs/cli/authentication.md
@@ -19,6 +19,11 @@ The Gemini CLI requires you to authenticate with Google's AI services. On initia
         echo 'export GOOGLE_CLOUD_PROJECT="YOUR_PROJECT_ID"' >> ~/.bashrc
         source ~/.bashrc
         ```
+        - MacOS users using zsh can use the following commands to add the environmental variables to ~/.zshrc file:
+        ```zsh
+        echo 'export GOOGLE_CLOUD_PROJECT=YOUR_PROJECT_ID' >> ~/.zshrc
+        source ~/.zshrc
+        ```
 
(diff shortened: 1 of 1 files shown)


allenhutchison:
I appreciate what you are asking for here, but how to configure env variables across all shells and os ssytems is outside the bounds of what we want to include in our documentation. Thank you.

garciasces:
Now worries, thanks for reviewing it.

We are reviewing most AI CLI tools lately. Most of the times the docs focus on setting up MacOS/Linux as dev environments with sample code that works for bash (Linux) and zsh (MacOS default shell). Once I discovered that this was not the case, I documented it and shared it with my team, so that people can get to the tool ASAP without having to waste time tinkering it again and again.

Then I thought other people might benefit from it, but if that doesn't suit the docs, it's ok.

PR #2203, opened 2025-06-27
docs: clarify shell command does not use default shell

### LINKED ISSUE #2199: Docs on shell use tool incorrectly claim default shell is used
### What happened?

The [docs](https://github.com/google-gemini/gemini-cli/blob/main/docs/cli/commands.md#shell-mode--passthrough-commands-) on shell-mode and state that the command will be executed  "...in your system's default shell".

However, the [docs](https://github.com/google-gemini/gemini-cli/blob/main/docs/tools/shell.md) and [implementation](https://github.com/google-gemini/gemini-cli/blob/main/packages/core/src/tools/shell.ts#L189) for the `run_shell_command` tool indicates that it always uses bash (or cmd.exe).

This is also easily confirmed using the shell tool itself:

Image

It would be great to use the user's preferred shell, but until that can be achieved, the main docs should reflect actual behavior.

### Client information

<details>

```console
$ gemini /about
╭───────────────────────────────────────────────────────────────────────────╮
│                                                                           │
│ About Gemini CLI                                                          │
│                                                                           │
│ CLI Version              0.1.5                                            │
│ Git Commit               bf873a1                                          │
│ Model                    gemini-2.5-pro                                   │
│ Sandbox                  no sandbox                                       │
│ OS                       linux                                            │
│                                                                           │
╰───────────────────────────────────────────────────────────────────────────╯

```

</details>

### Login information

_No response_

_No response_

timalsinaditya:
Fixes #2199

This PR updates the shell command documentation in `docs/cli/commands.md` to clarify that
the shell tool uses `bash` on Linux/macOS and `cmd.exe` on Windows, rather than the system's
default shell. This prevents confusion for users expecting custom shell features to work.

Thanks!


bSysIT:
Why this package using powershell.exe? Why Avast call it Virus?

t-dub (approved):
Awesome! Thanks for picking this up!

NTaylorMullen (approved):

PR #2226, opened 2025-06-27
Fix: Enable run_shell_command in non-interactive mode

DubstepQBA:
The `run_shell_command` tool was unintentionally excluded in non-interactive mode, leading to "tool not found" errors. This commit removes `ShellTool.Name` from the `interactiveTools` exclusion list in `packages/cli/src/gemini.tsx`, allowing it to be used in non-YOLO approval modes.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

### DIFF
--- a/packages/cli/src/gemini.tsx
+++ b/packages/cli/src/gemini.tsx
@@ -28,7 +28,6 @@ import {
   ApprovalMode,
   Config,
   EditTool,
-  ShellTool,
   WriteFileTool,
   sessionId,
   logUserPrompt,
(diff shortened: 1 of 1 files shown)


allenhutchison:
Thanks for your contribution. We have the shell tool in yolo mode only so that the system doesn't run any commands with side effects in an unsupervised setting. This is intentional and part of our security posture.

syabro:
@allenhutchison can we provide allowed commands/tools with args? I want to be able to run some commands with `-p` :(

Issue #1585, opened 2025-06-25
Feature Enhancements for Gemini CLI


coygeek:
### **To-Do List: Feature Enhancements for Gemini CLI**

Hello Gemini CLI Team,

Congratulations on the launch of the Gemini CLI! It's a powerful tool with a solid foundation. Based on a review of its capabilities and a comparison with an established tool like Anthropic's Claude Code, here is a categorized to-do list of potential features to consider for future development.

#### **1. Deepen Developer Workflow & Agentic Capabilities**

These features would more tightly integrate the CLI into a developer's day-to-day coding loop.

*   **☐ Implement Native Git Integration**
    *   **Description:** Allow Gemini to perform git operations directly. Claude Code can create commits with descriptive messages, create new branches, show history, and even generate entire pull requests.
    *   **Rationale:** This moves the tool from being a code *generator* to a true coding *assistant* that participates in the version control lifecycle, saving developers significant time.
    *   **Potential Tools:** `create_commit`, `create_pr`, `get_git_diff`, `get_git_log`.

*   **☐ Add Image & Vision Capabilities**
    *   **Description:** Enable the CLI to understand and analyze images. Claude Code allows users to provide images (like screenshots of a UI, an error message, or an architecture diagram) as context for a prompt.
    *   **Rationale:** Visual context is incredibly powerful for UI development ("generate CSS to match this mockup"), debugging (analyzing a screenshot of a bug), and understanding system diagrams.

*   **☐ Introduce "Extended Thinking" or a Computional Budget**
    *   **Description:** Allow users to signal that a task is complex and requires more in-depth reasoning. Claude Code has keywords like "think", "think harder", and "ultrathink" that allocate more computational time for planning and analysis before generating a response.
    *   **Rationale:** For complex tasks like planning a major refactor or designing a new system architecture, a simple, fast response is often insufficient. Giving the model a larger "thinking" budget leads to more comprehensive and well-thought-out plans.

*   **☐ Implement Structured To-Do List Management**
    *   **Description:** Add tools for creating and managing a structured task list within a session. Claude Code has `TodoWrite` and `TodoRead` tools to help break down a large problem into a checklist that both the user and the AI can track.
    *   **Rationale:** This improves the agentic ability to handle multi-step tasks by maintaining a clear plan and tracking progress, preventing the AI from getting lost or forgetting steps.

#### **2. Enhance User Experience & Interactivity**

These features focus on improving the usability and power of the interactive REPL.

[...]

PR #2230, opened 2025-06-27
feat: modular GEMINI.md imports with @file.md syntax (#1585)

bniladridas:
**Feature: Modular Imports for `GEMINI.md` Files**

**Summary**
This update adds support for using `@path/to/file.md` in `GEMINI.md` files. It lets you split large project context into smaller, organized files.

**Why This Matters**

* **Modular Design**: Break big context files into smaller, easier-to-manage pieces
* **Reusable Content**: Share common parts across different projects or teams
* **Easier Updates**: Makes it simpler to keep things up to date as projects grow

**How It Works**

* A new import processor handles `@` imports, even nested ones
* Catches issues like circular or missing imports
* Built into how context files are loaded
* Added docs explaining how to use it
* Wrote tests to cover different edge cases

**Tests Included**

* Nested import handling
* Circular import detection
* Missing file errors
* Verified that existing tests still pass

**Checklist**

* [x] Feature added
* [x] Tests written and passing
* [x] Docs updated
* [x] Lint and build pass (`npm run preflight`)
* [x] Ready for review

### DIFF
--- a/packages/core/src/utils/importProcessor.ts
+++ b/packages/core/src/utils/importProcessor.ts
@@ -1,0 +1,185 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import * as fs from 'fs/promises';
+import * as path from 'path';
+
+// Simple console logger for import processing
+const logger = {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  debug: (...args: any[]) =>
+    console.debug('[DEBUG] [ImportProcessor]', ...args),
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  warn: (...args: any[]) => console.warn('[WARN] [ImportProcessor]', ...args),
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  error: (...args: any[]) =>
+    console.error('[ERROR] [ImportProcessor]', ...args),
+};
+
+/**
+ * Interface for tracking import processing state to prevent circular imports
+ */
+interface ImportState {
+  processedFiles: Set<string>;
+  maxDepth: number;
+  currentDepth: number;
+  currentFile?: string; // Track the current file being processed
+}
+
+/**
+ * Processes import statements in GEMINI.md content
+ * Supports @path/to/file.md syntax for importing content from other files
+ *
+ * @param content - The content to process for imports
+ * @param basePath - The directory path where the current file is located
+ * @param debugMode - Whether to enable debug logging
+ * @param importState - State tracking for circular import prevention
+ * @returns Processed content with imports resolved
+ */
+export async function processImports(
+  content: string,
+  basePath: string,
+  debugMode: boolean = false,
+  importState: ImportState = {
+    processedFiles: new Set(),
+    maxDepth: 10,
+    currentDepth: 0,
+  },
+): Promise<string> {
+  if (importState.currentDepth >= importState.maxDepth) {
+    if (debugMode) {
+      logger.warn(
+        `Maximum import depth (${importState.maxDepth}) reached. Stopping import processing.`,
+      );
+    }
+    return content;
+  }
+
+  // Regex to match @path/to/file.md imports
+  // Supports both @path/to/file.md and @./path/to/file.md syntax
+  const importRegex = /@([./]?[^\s\n]+\.md)/g;
+
+  let processedContent = content;
+  let match: RegExpExecArray | null;
+
+  // Process all imports in the content
+  while ((match = importRegex.exec(content)) !== null) {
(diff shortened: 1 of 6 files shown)


packages/core/src/utils/importProcessor.ts
allenhutchison, line 1: Can you rename this file to memoryImportProcessor.ts?

packages/core/src/utils/importProcessor.ts
```diff
@@ -0,0 +1,185 @@
...
+      logger.warn(
+        `Maximum import depth (${importState.maxDepth}) reached. Stopping import processing.`,
+      );
+    }
+    return content;
+  }
+
+  // Regex to match @path/to/file.md imports
+  // Supports both @path/to/file.md and @./path/to/file.md syntax
+  const importRegex = /@([./]?[^\s\n]+\.md)/g;
```
allenhutchison: When you add the docs for this, can you make sure to call out clearly that this is scoped only to md files? Also I'd like to see a console.warn for people who are trying to import non-md files. If I have:

@foo/instructions.txt that would fail and I don't think that would be obvious to a person trying to use this.

packages/core/src/utils/importProcessor.ts
```diff
@@ -0,0 +1,185 @@
...
+    });
+  }
+
+  // For relative paths, check if resolved path is within allowed directories
+  return allowedDirectories.some((allowedDir) => {
+    const normalizedAllowedDir = path.resolve(allowedDir);
+    const normalizedResolvedPath = path.resolve(resolvedPath);
+    return normalizedResolvedPath.startsWith(normalizedAllowedDir);
+  });
+}
```
allenhutchison: The logic in validateImportPath seems a bit redundant. The check for absolute paths and relative paths could be simplified. Since path.resolve handles both cases, you can resolve the path first and then check if it's within the allowed directories.

  Here's a simplified version of validateImportPath:

```
export function validateImportPath(
  importPath: string,
  basePath: string,
  allowedDirectories: string[],
  ): boolean {
     // Reject URLs
     if (/^(file|https?):\/\//.test(importPath)) {
        return false;
     }

     const resolvedPath = path.resolve(basePath, importPath);
 
     return allowedDirectories.some((allowedDir) => {
       const normalizedAllowedDir = path.resolve(allowedDir);
       return resolvedPath.startsWith(normalizedAllowedDir);
     });
   }
```

allenhutchison (commented):
This is a great start and I love this addition. Can you please add updates for the documentation and a robust set of tests to this PR?

packages/core/src/utils/importProcessor.ts
bniladridas: Thank you for this, I have just ensured it.

bniladridas:
The code now issues a clear console.warn message:

```
if (!importPath.endsWith('.md')) {
  logger.warn(
    `Import processor only supports .md files. Attempting to import non-md file: ${importPath}. This will fail.`,
  );
  // Replace the import with a warning comment
  processedContent = processedContent.replace(
    match[0],
    `<!-- Import failed: ${importPath} - Only .md files are supported -->`,
  );
  continue;
}
```

- Scoped to .md files only  
- Warns on non-md imports  
- Includes path validation cleanup

All quality checks have passed:
1. ✅ Linting: npm run lint:ci - No ESLint errors or warnings
2. ✅ Type Checking: npm run typecheck - No TypeScript errors
3. ✅ Tests: npm run test:ci - All 647 tests passed (45 test files)

allenhutchison (approved):
Thank you for this contribution!

jackpal:
Very cool cl! Thanks for implementing this much-needed feature!

FWIW, this implementation is incompatible with the CLAUDE.md equivalent in (at least) the following ways:

1. It accepts @ imports everywhere in the text of the file, whereas CLAUDE.md ignores @ imports inside inline code quotes and code blocks.
2. It only imports md files, whereas CLAUDE.md imports any file (and treats whatever it imports as a markdown file.)
3. It defaults to an import depth of 10 vs claude's import depth limit of 5.

I think all three of these differences should be fixed, for compatibility sake. I'll file separate bugs for each, since I could see people arguing that 2 and 3 are useful, if incompatible with established practice.

bniladridas:
Thanks for your feedback! 

@jackpal, I've opened a new pull request addressing your concerns: [#2978](https://github.com/google-gemini/gemini-cli/pull/2978).

jackpal:
@bniladridas Thanks for trying to improve things, I'll leave some review comments on your pull request.

bniladridas:
@allenhutchison, 
Thanks again for merging this! I've opened a follow-up PR [#2978](https://github.com/google-gemini/gemini-cli/pull/2978) 
that continues the CLAUDE.md compatibility work. Would you be able to take a look?

PR #2233, opened 2025-06-27
Update MaxSizedBox.tsx

elvinagam:
## TLDR

Introduced support for truncating lines with an ellipsis when the available width is too small, preventing overflowing segments

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


packages/cli/src/ui/components/shared/MaxSizedBox.tsx
```diff
@@ -418,7 +419,26 @@ function layoutInkElementAsStyledText(
 
   if (availableWidth < 1) {
     // No room to render the wrapping segments. TODO(jacob314): consider an alternative fallback strategy.
```
jacob314: you can remove the TODO on line 421 now that there is a fallback solution

packages/cli/src/ui/components/shared/MaxSizedBox.test.tsx
```diff
@@ -247,6 +247,20 @@ Line 3`);
 🐶🐶
 🐶`);
   });
+  it('falls back to an ellipsis when width is extremely small', () => {
```
jacob314, line 251: can you add a couple more test cases?
Other cases of interest:
The non wrapping text exceeds the length of the MaxSizedBox.
The non wrapping text happens to have line breaks (ugly edge case for this)
The non wrapping text  exceeds the length of the MaxSizedBox.
Add a case with multiple rows in the MaxSizedBox just for clarity of how this may look with multiple lines with `...`

jacob314 (commented):
This looks generally quite good. Thanks for the polish! I'll approve after the test only comments are resolved.
One question: are there any real world cases you are seeing where this mattered? I made the handling of this edge case a bit ugly because I couldn't think of where it would occur in practice in the app.

elvinagam:
> This looks generally quite good. Thanks for the polish! I'll approve after the test only comments are resolved. One question: are there any real world cases you are seeing where this mattered? I made the handling of this edge case a bit ugly because I couldn't think of where it would occur in practice in the app.

@jacob31  So while the edge case is very narrow (only appears when the terminal window is really narrow), it can happen in practice (for instance, a user viewing a confirmation in a tiny tmux pane).

jacob314 (dismissed):
Thank you for fixing one of the rough edges in MaxSizedBox!
lgtm

packages/cli/src/ui/components/shared/MaxSizedBox.tsx
```diff
@@ -432,8 +433,26 @@ function layoutInkElementAsStyledText(
...
+      if (remainingWidth <= 0) break;
+      const segmentWidth = stringWidth(segment.text);
+      if (segmentWidth <= remainingWidth) {
+        truncatedLine.push(segment);
+        remainingWidth -= segmentWidth;
+      } else {
+        const slice = cpSlice(segment.text, 0, remainingWidth);
+        if (slice) {
+          truncatedLine.push({ text: slice, props: segment.props });
+        }
```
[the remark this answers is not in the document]
jacob314: darn. gemini-code-assist was a better reviewer than I was this time. Agree that something along the lines of what gemini-code-assist suggested is required. to test this please add a test that uses some wide characters (see the test case "wraps text with multi-byte emoji characters correctly" for some examples)

elvinagam (commented):
added changes

packages/cli/src/ui/components/shared/MaxSizedBox.tsx
```diff
@@ -432,8 +433,39 @@ function layoutInkElementAsStyledText(
...
+    const truncatedLine: StyledText[] = [];
+    let remainingWidth = Math.max(0, maxWidth - stringWidth('…'));
+    for (const segment of nonWrappingContent) {
+      if (remainingWidth <= 0) break;
+      const segmentWidth = stringWidth(segment.text);
+      if (segmentWidth <= remainingWidth) {
+        truncatedLine.push(segment);
+        remainingWidth -= segmentWidth;
+      } else {
+        // Slice the segment by visual width rather than code points so that
```
jacob314: nit: comment isn't quite right. I'd suggest removing or saying that you are segmenting by code points rather than segmenting by visual width

packages/cli/src/ui/components/shared/MaxSizedBox.test.tsx
```diff
@@ -247,7 +247,73 @@ Line 3`);
...
+            <Text>{'A\nBCDE'}</Text>
+            <Text wrap="wrap">wrap</Text>
+          </Box>
+        </MaxSizedBox>
+      </OverflowProvider>,
+    );
 
+    expect(lastFrame()).equals(`A\n…`);
+  });
+
```
jacob314, line 295: also add a test case with emojis or other wide characters to make sure we are now handling that case correctly.

jacob314 (dismissed):
lgtm
A couple minor comments but looks good to submit after those are addressed. Thanks for the cleanup!

elvinagam (commented):
added requested changes @jacob314

scottdensmore:
@elvinagam there are build issues in the PR. Can you run `npm build` and then before submitting run `npm run preflight`.

elvinagam:
@scottdensmore looks good to me.

elvinagam:
@jacob314 @umairidris @dewitt

packages/cli/src/ui/components/shared/MaxSizedBox.test.tsx
jacob314: Is this test actually checking the rendered text for this case. On my phone so sorry if I am missing something.

scottdensmore (dismissed):
Approved

scottdensmore:
@elvinagam I approved... looks like there are some build failures in the test that need to be fixed.

elvinagam:
> @elvinagam I approved... looks like there are some build failures in the test that need to be fixed.

no build failures as of now, I believe @scottdensmore

scottdensmore:
The issue is in MaxSizedBox.test.tsx. If you run `npm run preflight` you will see this.
CleanShot 2025-07-08 at 11 42 15@2x

swissspidy:
There's 1 failing test now:

```
FAIL  packages/cli/src/ui/components/shared/MaxSizedBox.test.tsx > <MaxSizedBox /> > truncates non-wrapping text containing line breaks
AssertionError: expected 'A…\nB' to equal 'A\n…'

- Expected
+ Received

- A
- …
+ A…
+ B
```

jacob314:
If you can fix the test failures it would be great to land this!

elvinagam:
@jacob314 fixed!


✅ FixedMaxSizedBox line break truncation)
✅ 26 test suites passing (214 individual tests passing)

There are ❌ 29 test suites failing due to build/dependency issues (not test logic problems). We can fix those in another PR.

jacob314 (approved):
lgtm
To speed up landing this I went ahead and ran the formatter and fixed a lint error due to an unused import.

scottdensmore (approved):
Approved

elvinagam:
fantastic. Thanks, team @jacob314 @scottdensmore. Let me know if there is something in priority to look at next.

PR #2273, opened 2025-06-27
Add support for summarizing tool output 

anj-s:
## TLDR

Add support for summarizing tool output if they are over 2k characters long.
1. shell commands
2. MCP servers 

## Dive Deeper

2k limit was chosen to be aggressive for some of the longer running outputs that can increase context length unnecessarily.

## Reviewer Test Plan
1. Run a shell command that might produce a large output
2. Run a MCP server and this summarization is applied for all outputs over 2k

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | Y  | ❓  | Y  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


anj-s:
> /gemini review

Are we testing out this feature? Would be great to allow the author to trigger this functionality.

packages/core/src/core/coreToolScheduler.ts
```diff
@@ -134,17 +135,30 @@ function createFunctionResponsePart(
...
+  config: Config,
+  abortSignal: AbortSignal,
+): Promise<PartListUnion> {
   const contentToProcess =
     Array.isArray(llmContent) && llmContent.length === 1
       ? llmContent[0]
       : llmContent;
 
+  const geminiClient = config.getGeminiClient();
+  const coreTools =
```
NTaylorMullen, line 151: nit: split this up into a few variables. Will make debugging difficult with it inlined like this

packages/core/src/utils/toolOutputSummarizer.ts
```diff
@@ -0,0 +1,100 @@
...
+    return textToSummarize;
+  }
+  const history = geminiClient.getChat().getHistory(true);
+  const historyJsonString = JSON.stringify(history, null, 2);
+  const prompt = SUMMARIZE_TOOL_OUTPUT_TEMPLATE.replace(
+    '{maxLength}',
+    String(maxLength),
+  )
+    .replace('{textToSummarize}', textToSummarize)
+    .replace('{history}', historyJsonString);
```
NTaylorMullen, line 80: Why embed the history in the request prompt vs. letting the history naturally exist as part of the request's built-in history/contents collection?

packages/core/src/utils/toolOutputSummarizer.ts
```diff
@@ -0,0 +1,100 @@
...
+  Content,
+  GenerateContentConfig,
+  SchemaUnion,
+  Type,
+} from '@google/genai';
+import { GeminiClient } from '../core/client.js';
+import { DEFAULT_GEMINI_FLASH_MODEL } from '../config/models.js';
+
+const toolOutputSummarizerModel = DEFAULT_GEMINI_FLASH_MODEL;
+const toolOutputSummarizerConfig: GenerateContentConfig = {
```
NTaylorMullen, line 17: You'll also want to control for max tokens here

packages/core/src/utils/toolOutputSummarizer.ts
```diff
@@ -0,0 +1,100 @@
...
+        "type": "string",
+        "description": "The concise summary of the text, not exceeding the specified character limit."
+    }
+  },
+  "required": ["summary"]
+}
+\`\`\`
+`;
+
+const RESPONSE_SCHEMA: SchemaUnion = {
```
NTaylorMullen, line 53: I'm not sure I'd use a schema for single output summarization. It'll require the model to output in a json format which will be prone to escaping issues.

packages/core/src/core/coreToolScheduler.ts
```diff
@@ -134,17 +135,30 @@ function createFunctionResponsePart(
...
+    (await config.getToolRegistry()).getAllTools().map((t) => t.name) || [];
   if (typeof contentToProcess === 'string') {
+    if (toolName === 'run_shell_command' || !coreTools.includes(toolName)) {
+      const summarizedContent = await summarizeToolOutput(
```
NTaylorMullen, line 155: Why embed summarizations here vs. inside the tools themselves?

allenhutchison: Agree. I don't want to derail this PR, but I was hoping that we would land on a new hook in the tool interface called summerize output that our tool scheduler would call before the output comes back to the main agent thread. The tool could decide how to best summarize it's own output or use a default summarizer on the interface.

umairidris:
> > /gemini review
> 
> Are we testing out this feature? Would be great to allow the author to trigger this functionality.

Anyone can comment the command and trigger the bot. We also enabled automatic reviews on new PRs being opened in the repo. I'm just backfilling reviews on existing PRs.

Issue #2140, opened 2025-06-27
Implement resilient rate limiting with circuit breaker pattern and intelligent backoff


eLyiN:
## Summary

Implement a robust rate limiting system with circuit breaker pattern to prevent infinite loops and provide graceful degradation when hitting API limits.

## Current Problem

Multiple users are experiencing infinite rate limiting loops that render the CLI completely unusable:

- **Issue #1626**: Permanent CLI lockup with endless rate limit messages
- **Issue #1631**: Looping on rate limiting detection when switching models
- **Issue #1687**: Rate limiting causes endless printing instead of graceful stopping
- **Issue #1644**: CLI fails with 429 errors immediately on second request
- **Issue #1502**: Users getting 429 errors even on first prompt

**Common pattern**: CLI enters unrecoverable state where it continuously retries without backoff, preventing any user interaction.

### 1. Circuit Breaker Implementation
- **Open State**: Stop attempting requests after consecutive failures
- **Half-Open State**: Allow limited retry attempts with exponential backoff
- **Closed State**: Resume normal operation after successful requests
- **Configurable thresholds**: User-definable failure counts and timeout periods

### 2. Intelligent Backoff Strategy
- **Exponential backoff with jitter**: Prevent thundering herd problems
- **Respect Retry-After headers**: Use server-provided timing when available
- **Per-model rate limiting**: Independent tracking for gemini-2.5-pro vs gemini-2.5-flash
- **User feedback**: Clear progress indicators during backoff periods

### 3. Graceful Degradation
- **Automatic model switching**: With user notification and consent
- **Fallback mechanisms**: Local caching, offline mode for some operations
- **Request queuing**: Buffer requests during temporary rate limits
- **Recovery detection**: Automatically resume when rate limits clear

### 4. Enhanced User Experience
- **Clear messaging**: Replace "User cancelled" with accurate rate limit status
- **Progress indicators**: Show remaining backoff time
- **Manual override**: Allow users to cancel and retry manually
- **Configuration options**: Let users tune retry behavior

## Technical Implementation

- Extend existing `geminiChat.ts` retry logic with circuit breaker pattern
- Add rate limiting state management to prevent infinite loops
- Implement configurable backoff strategies in API client
- Enhance error messaging in CLI UI components
- Add telemetry for rate limiting patterns (complements #2127)

## Benefits

1. **Eliminates infinite loops** - CLI remains responsive during rate limiting
[...]

PR #2278, opened 2025-06-27
 Introduces a new CLI option that allows users to disable automatic m…

nofearsk:
…odel

  fallback when slow response times are detected. When enabled, the system
  will continue using the specified model regardless of performance.

  - Add --force-model boolean option to CLI arguments
  - Pass forceModel parameter through config initialization
  - Implement logic in flashFallbackHandler to check flag before switching
  - Display informative message when model switching is blocked

  This gives users more control over model selection in scenarios where
  they prefer consistency over automatic performance optimization.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


packages/core/src/config/config.ts
```diff
@@ -266,12 +269,20 @@ export class Config {
   }
 
   setModel(newModel: string): void {
+    if (this.forceModel) {
```
jacob314, line 272: setModel should set the model no matter what but we should audit call sites to ensure we do not set this when forceModel is set unless the user has confirmed.

packages/cli/src/config/config.ts
```diff
@@ -63,6 +64,11 @@ async function parseArguments(): Promise<CliArgs> {
       description: `Model`,
       default: process.env.GEMINI_MODEL || DEFAULT_GEMINI_MODEL,
     })
+    .option('force-model', {
```
jacob314, line 67: we should also probably support force-model in 
settings.ts so you can force it in your settings.json rather than via a command line argument

packages/cli/src/ui/App.tsx
```diff
@@ -245,6 +245,19 @@ const App = ({ config, settings, startupWarnings = [] }: AppProps) => {
       currentModel: string,
       fallbackModel: string,
     ): Promise<boolean> => {
+      // Check if force model is enabled
+      if (config.getForceModel()) {
+        // Add message to UI history explaining that model switching is disabled
+        addItem(
+          {
+            type: MessageType.INFO,
```
jacob314, line 253: I wonder if we likely need to use exponential backoff or similar here.  We also need to make sure we are surfacing how long the user has been delayed by not allowing a switch to gemini flash. Ideally the user would also have the ability to trigger the switch to flash without having to stop gemini cli again even when in forceMode.  This could be accomplished by perhaps surfacing a dialog asking the user if they would like to switch to flash or supporting a clearly visible keyboard shortcut to switch to flash when slow responses have been detected and have been on going for a significant amount of time.
https://github.com/google-gemini/gemini-cli/issues/2140

packages/cli/src/config/config.ts
jacob314: sorry after some further thinking. Can we make this only be in settings.json to start out with?

packages/cli/src/ui/App.tsx
jacob314: to make sure that users have a good experience instead of experiencing timeouts, lets only apply this potion for API Key to start with to avoid unintended consequences like too many requests timing out without a response due to a setting users might not be aware was setup. That would also avoid us having to make the fallback behavior when the model continues to be slow too complicated. With that change I would be fine landing without giving users a keyboard shortcut to manually fall back.

ASVIEST: > to make sure that users have a good experience instead of experiencing timeouts, lets only apply this potion for API Key to start with to avoid unintended consequences like too many requests timing out without a response due to a setting users might not be aware was setup. That would also avoid us having to make the fallback behavior when the model continues to be slow too complicated. With that change I would be fine landing without giving users a keyboard shortcut to manually fall back.

Please allow this for account users from the start, automatic switching to flash model is very inconvenient, I don't care that the responses can be quite slow. Flash model often fails to cope with the project that gemini 2.5 pro can handle. I don't see any problem with just adding the --force-model flag. If I'm interested in switching to flash, why can't I just add the /model command?

jacob314: @allenhutchison curious on your thoughts here. Would be nice to enable this for as many users as possible but want to make sure we don’t do it in a way that eill cause system instability for times we are experiencing higher than expected  usage.

jacob314: This was answered better than I could in the Gemin CLI AMA. Unfortunately for this pull request lets stick to removing the fallback for API Key users because we have to balance model choice with capacity for free tier.

https://www.reddit.com/r/Bard/comments/1lms62v/comment/n0lysi2/
The advice from @ryanjsalva in the AMA is that if you want to use a specific model, you can always [use an API Key](https://github.com/reugn/gemini-cli?tab=readme-ov-file#api-key). In a perfect world, you shouldn’t need to think about the model. It should Just Work.™ After all, Pro is overkill for a lot of really simple steps (e.g. “start the npm server”). Pro is better suited to big, complex tasks that require reasoning. 

For those devs using the free tier, our goal is to deliver the best possible experience at the keyboard – ideally one where you never have to stop work because you hit a limit. To do that inside a free tier, we have to balance model choice with capacity.

alexykn:
+1 please allow this, auto switch to flash on standard tier is horrible. I can not leave it running alone because flash will just take over mid project and literally destroy things. I don't care if there is rate limits or slower response times just please do not auto switch to flash

packages/cli/src/ui/App.tsx
```diff
@@ -245,6 +245,19 @@ const App = ({ config, settings, startupWarnings = [] }: AppProps) => {
...
+            text: `⚡ Slow response times detected, but model switching is disabled due to --force-model flag. Continuing with ${currentModel}.`,
+          },
+          Date.now(),
+        );
+        return false; // Reject the fallback
```
jacob314, line 258: Sorry for the long delay on this.
We've discussed this and are now ready to approve as long as we make sure this errors out after you get 5 429s rather than continuing to to get a response indefinitely. When we error out due to 5 429s we should make sure there is a user visible message explaining that the request failed rather than falling back to Flash due to the --force-model flag.

abhipatel12:
Hello! Checking in on this. Is there still an interest to move forward with this PR? Otherwise I will mark it as closed

fabianx-ai:
It would be better and nicer to just have /model to select a model ourselves - especially as Vertex AI supports anthropic and hence we could run the gemini cli with Claude for most things and use gemini only for the hard-analysis things (like code reviews, etc.) - saving lots of tokens and capacity.

jacob314:
Closing due to inactivity. If you would like to proceed please ping and I will reopen.

packages/cli/src/ui/App.tsx
Exadra37: > For those devs using the free tier, our goal is to deliver the best possible experience at the keyboard – ideally one where you never have to stop work because you hit a limit. To do that inside a free tier, we have to balance model choice with capacity.

Well you are giving the WORST possible ever experience, because flash is at the Junior lever for coding and is a complete time waster and cause of frustration.  Better to continue coding by myself then waste my time with a model that's not up to the task. 

So, i prefer to not have an auto fallback in case of rate limits or slow response times. Instead I prefer the CLI to just stop work on rate limits via a setting configuration and have a timeout on slow response times with a retry in x amount of time.

PR #2303, opened 2025-06-27
fix: Enabled detection of VS Code installed on macOS in Applications/ non PATH

### LINKED ISSUE #1695: Detect default VSCode installation on macOS (non-$PATH)
### What happened?

Image

VSC is installed, thus, it should be detected

### Client information

<details>

```console
$ gemini /about
│                                                                                        │
│ About Gemini CLI                                                                       │
│                                                                                        │
│ CLI Version                   0.1.3                                                    │
│ Git Commit                    382486e (local modifications)                            │
│ Model                         gemini-2.5-pro                                           │
│ Sandbox                       no sandbox                                               │
│ OS                            darwin                                                   │
│                                                                                        │
╰────────────────────────────────────────────────────────────────────────────────────────╯
```

</details>

### Login information

_No response_

_No response_

devpool007:
## TLDR

Updated editor.ts to enable option of detection of VS Code when installed in Applications and not installed code on the $PATH 

## Dive Deeper

- As suggested by @markmcd '/Applications/Visual Studio Code.app/Contents/Resources/app/bin/code' was utilized as a fallback for checking if code command exists for VS Code Editor on macOS.

- Also editor.test.ts had to be updated since now what was expected as false before returned true especially in the 'should return false for vscode when not installed and not in sandbox mode', test for macOS/darwin 

## Reviewer Test Plan
1. Ether enable or disable sandbox by putting env variable GEMINI_SANDBOX=true or GEMINI_SANDBOX=false
2. npm run build
3. npm start
4. Type /editor and check the output 

In case of GEMINI_SANDBOX=true you should see this:
Screenshot 2025-06-28 at 01 22 18

In case of GEMINI_SANDBOX=false you should see this:
Screenshot 2025-06-28 at 00 56 05

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅   | ✅   | ✅   |
| npx      | ✅   | ✅   | ✅   |
| Docker   | ✅   | ✅   | ✅   |
| Podman   | ❓  | -   | -   |

## Linked issues / bugs

Fixes #1695 and #2422


packages/core/src/utils/editor.ts
```diff
@@ -41,6 +50,14 @@ export function checkHasEditorType(editor: EditorType): boolean {
   const commandConfig = editorCommands[editor];
   const command =
     process.platform === 'win32' ? commandConfig.win32 : commandConfig.default;
+
+  // Special handling for VS Code on macOS
+  if (editor === 'vscode' && process.platform === 'darwin') {
+    const fallbackPath =
+      '/Applications/Visual Studio Code.app/Contents/Resources/app/bin/code';
+    return commandExists(command, fallbackPath);
```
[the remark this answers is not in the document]
devpool007: Made and pushed the changes to branch

devpool007:
Hi @allenhutchison could you please assign someone to review this small but effective change for macOS?

packages/core/src/utils/editor.ts
```diff
@@ -41,6 +52,13 @@ export function checkHasEditorType(editor: EditorType): boolean {
   const commandConfig = editorCommands[editor];
   const command =
     process.platform === 'win32' ? commandConfig.win32 : commandConfig.default;
+
+  // Special handling for VS Code on macOS
+  if (editor === 'vscode' && process.platform === 'darwin') {
```
NTaylorMullen: Instead of this special casing would it be possible to build in the fallback logic into the editorCommands declaration?

devpool007: Hi, thanks for the review, I think that's a nice idea would make more sense and code would be cleaner, I will implement this and update the code soon.

devpool007:
Hi @NTaylorMullen I moved fallback logic inside editorCommands and updated code according to latest changes in main. I was also wondering should I also add macOSFallbacks for other editors too?
As right now in this PR the focus was on vscode but maybe it might be safe to do this for other editors too many of which are vscode based? 

Or should this be handled then in another PR? Lmk your thoughts :)

packages/core/src/utils/editor.ts
```diff
@@ -50,9 +59,15 @@ function commandExists(cmd: string): boolean {
  */
 const editorCommands: Record<
   EditorType,
-  { win32: string[]; default: string[] }
+  { win32: string[]; default: string[]; macOSFallbacks?: string[] }
```
NTaylorMullen: ditto elsewhere

```suggestion
  { win32: string[]; default: string[]; fallbackPaths?: string[] }
```

packages/core/src/utils/editor.ts
```diff
@@ -65,7 +80,24 @@ export function checkHasEditorType(editor: EditorType): boolean {
...
+  // First, try the standard commands
+  const hasStandardCommand = commands.some((cmd) => commandExists(cmd));
+  if (hasStandardCommand) {
+    return true;
+  }
+
+  // On macOS, try fallback path for vscode available
+  if (
+    editor === 'vscode' &&
+    process.platform === 'darwin' &&
```
NTaylorMullen: Lets not just check on darwin, lets check always, you can delete this bit and you should be able to delete the editor vscode check. SO fallback paths should be truly fall back paths and be checked always if they're configured

NTaylorMullen (changes requested):

packages/core/src/utils/editor.ts
devpool007: Thanks for the suggestions, yes that makes sense actually, will do so.

teeler:
Any updates? Just skimming through old PRs - can we close this or do you all want to resolve the conflicts and the last few unresolved comments?

chrstnb:
Going to close as no updates since @allenhutchison's comment, but feel free to re-open if needed

PR #2306, opened 2025-06-27
fix edit retrigger

ngleo:
**Copied from #1115**

## Background
#808 enables users to modify the proposed content when the edit tool is triggered. However, Gemini is not aware that the user has modified the proposed content. When the tool has executed, Gemini sometimes reads the file to verify the change. When it sees an unexpected change by the user, it might retrigger the tool.

## Fix
This PR changes the edit tool & write tool to conditionally add a "content has been modified" message in the response to the LLM if the user goes through the modify flow.

With this approach, I can see Gemini's thought / response below and Gemini has not retriggered edit tool so far:

`
I can confirm that the file has been updated. The first line of CONTRIBUTING.md now reads \"XYZ\" It seems the user overrode the original request, and I've processed the  modified content accordingly. I will now proceed with the updated state of the file.
`

`
I have changed the first line of CONTRIBUTING.md to "Hello" as you requested. The edit was modified by you.
`

## Reviewer Test Plan
- Ask Gemini to edit a file with a very specific change. e.g. @{some file} change the 1st line to {something}
- Trigger modify flow, modify the proposed change (can be a very different change), and see if it retriggers the tool.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓ | -   | -   |

## Linked issues / bugs
- Fixes b/424920326.


packages/core/src/tools/edit.ts
```diff
@@ -414,9 +421,12 @@ Expectation for required parameters:
         displayResult = { fileDiff, fileName };
       }
 
+      const modfiedInfo = params.modified_by_user
+        ? ` User modified the \`new_string\` content to be: ${params.new_string}.`
```
jacob314: optional style nit. rather than creating this as an empty string use the handy typescript array building syntax to build up a list of parts and then join them with ` `. that would avoid having this string start with a space which is slightly odd.

jacob314 (commented):
Can you also add a basic test to this code? After that LGTM from me.

jacob314 (dismissed):
LGTM! Thanks for adding tests. Will hel make sure we dont accidentally break this important functionality in the future.

packages/core/src/tools/edit.ts
ngleo: Sounds good, done. Just noticed that I didn't add the modified messages to the new file case, oops.

ngleo:
Thanks for reviewing, added!

jacob314 (approved):

allenhutchison (approved):
This is a fantastic contribution. Thank you so much!

PR #2335, opened 2025-06-28
feat: Add cancellation token for chat messages

### LINKED ISSUE #2324: Cancellation Token

Please include Cancellation token.

### Why is this needed?

When I ask to search web for something, it takes long time, if i want to stop and continue, there is no option to abort...

### Additional context

_No response_

tejasm-189:
This change introduces cancellation support for chat messages using AbortSignal.

- GeminiChat.sendMessage and GeminiChat.sendMessageStream now accept an optional AbortSignal.
- The AbortSignal is passed down to the ContentGenerator, which then passes it to the underlying API calls.
- Unit tests have been added to verify that the AbortSignal is correctly propagated and that API calls are cancelled when the signal is aborted.

## Linked issues / bugs

Fixes #2324

### DIFF
--- a/packages/core/src/core/client.ts
+++ b/packages/core/src/core/client.ts
@@ -229,6 +229,7 @@ export class GeminiClient {
       yield { type: GeminiEventType.ChatCompressed, value: compressed };
     }
     const turn = new Turn(this.getChat());
+    // Pass the signal to the turn's run method, which passes it to chat.sendMessageStream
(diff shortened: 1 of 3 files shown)


tejasm-189:
issue #2324 cancellation token feat fixes

allenhutchison:
I don't think this is needed because we have the esc command for cancelling the current execution. Please reopen the issue if you disagree.

PR #2442, opened 2025-06-28
Fix Shift + Enter key combo to allow multiline input as documented

slawt:
## TLDR

Updated code to handle `Shift + Enter` to allow the user to enter new lines as was documented. Updated documentation to show that both this key combo and `Ctrl + Enter` are now accepted.

## Dive Deeper

The Help documentation previously stated that `Shift + Enter` creates a new line, but the actual implementation only supported `Ctrl + Enter` (and `Meta + Enter` on Mac). This PR fixes that inconsistency by adding `key.shift` to the condition that checks for newline insertion.

The change is minimal - just adding `|| key.shift` to the existing condition in `InputPrompt.tsx`. A test was added to verify that Shift+Enter triggers `buffer.newline()` instead of submitting the message.

## Reviewer Test Plan

1. Run the Gemini CLI: `npm start`
2. Type some text in the prompt
3. Press `Shift + Enter` - it should insert a newline and allow you to continue typing
4. Press `Ctrl + Enter` - it should also insert a newline (existing behavior)
5. Press plain `Enter` - it should submit the message
6. Type `/help` and verify the documentation shows "Ctrl+Enter / Shift+Enter - New line"

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅ | ❓  | ❓  |
| npx      | ✅ | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

### DIFF
--- a/packages/cli/src/ui/components/Help.tsx
+++ b/packages/cli/src/ui/components/Help.tsx
@@ -97,7 +97,7 @@ export const Help: React.FC<Help> = ({ commands }) => (
     </Text>
     <Text color={Colors.Foreground}>
       <Text bold color={Colors.AccentPurple}>
-        Shift+Enter
+        Ctrl+Enter / Shift+Enter
       </Text>{' '}
       - New line
     </Text>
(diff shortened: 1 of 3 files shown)


jacob314:
Thank you for your pull request! 100% agree that supporting shift+enter is a top priority. Unfortunately it is really complicated to support Shift+enter on all terminals (feels like Internet Explorer 6.0 error web compatibility). This change will likely cause some terminals to think that all enters are now shift+enter.
We're working on it and I'm optimistic we will have a robust solution landed shortly.
See
https://github.com/google-gemini/gemini-cli/pull/3289
if you would like to follow along as we try to solve this using the kitty protocol.

PR #2491, opened 2025-06-28
fix: show ctrl+s shortcut to expand debug console #2002

### LINKED ISSUE #2002: I use https proxy and I get an error first 923 lines hidden
### What happened?

Error content:
first 923 lines hidden ...                                                                  │
│          errno: -60,                                                                            │
│          code: 'ETIMEDOUT',                                                                     │
│          syscall: 'connect',                                                                    │
│          address: '216.239.32.223',                                                             │
│          port: 443                                                                              │
│        }                                                                                        │
│      ]                          

Image
Also, every conversation context is 100% starting with not knowing if this is the way it is or if there is a problem

What is the specific question of this

### Client information

<details>

```console
$ gemini /about
# paste output here
```

</details>

### Login information

google

he question of whether my https proxy port is not port 443 and the context of each conversation is 100% started 

Image

devpool007:
## TLDR

Added ctrl+s to be visible on debug console for expanding the debug console, when the output is too large inside it.

## Dive Deeper
Nothing much to dive deep :) 

## Reviewer Test Plan
- Open debug console with ctrl+o and the ctrl+s shortcut should be visible
- Try to run some code by gemini which throws a large error in debug console
- Press the shortcut should work as mentioned by @jacob314 

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ✅  | ✅  |
| npx      | ✅  | ✅  | ✅  |
| Docker   | ✅ | ✅  | ✅  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ✅  | -   | -   |

## Linked issues / bugs
Fixes #2002


packages/cli/src/ui/components/DetailedMessagesDisplay.tsx
```diff
@@ -37,7 +37,8 @@ export const DetailedMessagesDisplay: React.FC<
     >
       <Box marginBottom={1}>
         <Text bold color={Colors.Foreground}>
-          Debug Console <Text color={Colors.Gray}>(ctrl+o to close)</Text>
+          Debug Console{' '}
+          <Text color={Colors.Gray}>(ctrl+o to close, ctrl+s to expand)</Text>
```
jacob314: We should only show ctrl-s when the content is truncated. Look at how this is implemented in app.tsx. If you ask gemini cli or do a grep search for ctrl-s you should find the existing code.

devpool007: Hi thanks for the review, will check it out and do the improvement!

jacob314: Sounds good. Let me know if you have any questions.

devpool007: Hi so I went through the code and try to understand the flow of things and it took me a while but I think I am headed in a good direction but need your help and suggestions:

So here I can see in line 766 if there are errors the <DetailedMessagesDisplay> and <ShowMoreLines constrainHeight={constrainHeight} /> are shown wrapped in the <OverflowProvider>. So technically if everything was correct the ctrl-s should be shown by ShowMoreLines component when debug console doesnt have constrainHeight.

Screenshot 2025-06-30 152005

Now I see inside <ShowMoreLines> component we have this complex if statement of which any condition being true leads to return null hence no ctrl-s being shown
Screenshot 2025-06-30 154716


Analysis:

- From my observation I  think overflowState is not undefined and overflowState.overflowingIds.size is not 0
- constrainHeight should also be true as setConstrainHeight only goes to false when ctrl-s is pressed (code in App.tsx)
Screenshot 2025-06-30 153547
- I assume !(streamingState === StreamingState.Idle || streamingState === StreamingState.WaitingForConfirmation) the last condition could be causing problem, though not sure
Screenshot 2025-06-30 154405

- Given the type above the third condition can only lead to true if streaminState is responding correct? if its ideal or waiting for confirmation it should overall return false still showing the ctrl-s and not returning null 

Though I am bit confused now why it doesn't show ctrl-s , do you have a suggestion where there might be a bug or an area I should focus on to find what might be causing ctrl-s to not appear? 

Or is it something really minor in the component code?      

```
<Box>
      <Text color={Colors.Gray} wrap="truncate">
        Press ctrl-s to show more lines
      </Text>
    </Box>

```

Let me know your thoughts and if maybe something needs to be checked in debugging mode etc.

jacob314: Please ask Gemini CLI, it should point you in the right direction if you ask it to emulate how this UI is shown in App.tsx. In particular tell it to bring in `@packages/cli` and it should be able to follow the code and understand how the existing logic works.

devpool007: Hi alright will try this approach, seems quite exciting to do it with the CLI :) Will update you when I find time to complete this.

devpool007: > Please ask Gemini CLI, it should point you in the right direction if you ask it to emulate how this UI is shown in App.tsx. In particular tell it to bring in `@packages/cli` and it should be able to follow the code and understand how the existing logic works.

Hi so I was finally able to find the issue it was more simpler than I thought so the ShowMoreLines component for the debug console was inside the same OverflowProvider as the DetailedMessagesDisplay it's supposed to be monitoring. 

This was the problem because the ShowMoreLines component is not a child of DetailedMessagesDisplay, so it doesn't get re-rendered when the overflow state changes. Wrapping them in a Box works as this was done for the HistoryItemDisplay component too. 

I have pushed the latest commit let me know when your review on it, thanks!

jacob314 (approved):
lgtm
Great fix! Thanks for looking closely and finding the root cause.

devpool007:
> lgtm Great fix! Thanks for looking closely and finding the root cause.

My pleasure :) and thanks for the review! 

Hi @allenhutchison, would you mind giving this PR an approving review when you have a moment? It's ready to be merged.

NTaylorMullen (approved):

PR #2506, opened 2025-06-29
fix: Fix Japanese IME input issue by using refs for state synchronization

dobachi:
When typing Japanese characters, React's asynchronous state updates cause all characters to be inserted at position 0, resulting in only the last character being displayed.

This fix uses useRef to maintain references to the latest state values, ensuring that rapid character inputs from IME always use the correct cursor position.

Fixes the same issue as #2462 but without removing the DEL character handling.

## TLDR

This PR fixes the Japanese IME input issue where only the last character would be displayed when typing Japanese. The root cause is React's asynchronous state updates - when IME sends multiple characters rapidly, each `handleInput` call uses stale state values. This PR solves it by using refs to always access the latest state values.

## Dive Deeper

The issue occurs when:
1. Using Japanese IME (or other IMEs that send multiple characters rapidly)
2. Multiple characters are sent in quick succession

This affects both `npx` and `npm run start` execution methods.

The problem is that `applyOperations` callback in `useTextBuffer` hook was using stale values from its closure. When IME sends "こんにちは", each character insertion sees cursor position as (0,0) because React hasn't updated the state yet.

The solution:
- Add `useRef` for lines, cursorRow, and cursorCol
- Update refs when state changes using `useEffect`
- In `applyOperations`, read from refs instead of closure values
- Update refs immediately after calling setState

This ensures that even with React's batched/async updates, we always have access to the latest values.

## Reviewer Test Plan

1. **Build the changes:**
   ```bash
   cd packages/cli
   npm run build
   ```

2. **Test Japanese IME input:**
   ```bash
   # Via npx:
   npx .

   # Or via npm:
   npm run start

   # Type Japanese text like "こんにちは" using IME
   # All characters should appear correctly (not just the last one)
   ```

3. **Verify existing functionality:**
   - Regular text input should work as before
   - Backspace/delete operations should work
   - The DEL character (\x7f) handling remains intact

4. **Run tests:**
   ```bash
   npm test -- text-buffer.test.ts
   # All 60 tests should pass
   ```

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ✅  |
| npx      | ❓  | ❓  | ✅  |
| Docker   | ❓  | ❓  | ✅  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

✅ = Tested and working (WSL2 Ubuntu for local tests, Docker container for Docker test)
❓ = Not tested

### Docker Testing Instructions
```bash
# Build Docker sandbox
npm run build
GEMINI_SANDBOX=docker npm run build:sandbox

# Run with mounted auth (requires prior authentication on host)
docker run -it --rm \
  -v ~/.config/gcloud:/home/node/.config/gcloud:ro \
  -v ~/.gemini:/home/node/.gemini:ro \
  gemini-cli-sandbox

# Note: Browser-based authentication doesn't work in Docker containers
# For testing Japanese IME, using 'npx .' locally is more practical
```

## Linked issues / bugs

- Related to #2462 - This PR provides an alternative solution that preserves the DEL character handling


packages/cli/src/ui/components/shared/text-buffer.ts
```diff
@@ -658,8 +680,13 @@ export function useTextBuffer({
...
       setCursorCol(newCursorCol);
       setPreferredCol(null);
+
+      // Update refs immediately
+      linesRef.current = newLines;
+      cursorRowRef.current = newCursorRow;
+      cursorColRef.current = newCursorCol;
     },
-    [lines, cursorRow, cursorCol, pushUndo, setPreferredCol],
+    [pushUndo, setPreferredCol],
```
jacob314, line 689: Fyi @bbiggs was just looking at fixing something perhaps similar with 
https://github.com/google-gemini/gemini-cli/pull/2504
Rather than using refs here which may cause other problems, I think we should resolve this by extending the logic for queuing up a list of operations to perform that is already applied for `delete` and `insert` but can be extended to most setters in this file.
You might even see if what bbiggs has today fixes your issue.
What @bbiggs has is imperfect as rather than using effects it should
coordinate sequential state updates like this by leveraging aggregate state objects with [state updaters](https://react.dev/reference/react/useReducer) and/or [useReducers](https://react.dev/reference/react/useReducer).
This is at the edge of my current limited but growing react expertise so please do push back if any of this doesn't make sense.

jacob314 (commented):
Thanks for the PR. I think you are right about the general direction we need to go to fix this issue resolving the state management in text-buffer.

nnfct:
Hello, s there a way for me to resolve this issue directly on my end? not the way waiting the update

dobachi:
@nnfct
While waiting for the official fix, you can test this PR by:
- Building from my fork: https://github.com/dobachi/gemini-cli/tree/fix-japanese-ime-state-sync
- This is for testing purposes only - please wait for the official release for production use.

[temporary-fix-comment.md](https://github.com/user-attachments/files/21004711/temporary-fix-comment.md)

dobachi:
@nnfct 
You may also refer to this PR: #2504 , which offers a more React-like solution.

nnfct:
@dobachi thanks a lot!

SandyTao520:
We decided to merge https://github.com/google-gemini/gemini-cli/pull/2504 as a temporary solution, and we're working on migrating text-buffer to use reducer as a long term fix.

Closing this one, thanks for your investigation and detailed report!

PR #2507, opened 2025-06-29
Highlight previous user input

AlphaDaze:
## TLDR

Draw a box around previous user input to highlight it for better readability, especially when there are multiple inputs in sequence.
Currently user input is gray which can be hard to find when scrolling through multiple messages. 
## Dive Deeper

When using the tool, the Gray colour is utilised for many other outputs such as comments in code blocks. This along with the gray colour make it hard to discern user input from Gemini output. It can be hard to determine when a previous output ends and new one begins as well as determine what the user had input.

This change will draw a box around input the input to make it slightly pop. I have utilised similar code to existing code - so should not see any new issue with any setup.

| Before  | After |
| ------------- | ------------- |
| gemini-cli-before | new_box |

I believe in the future this could be more customizable (other than just the theme). E.g. I like the gradient look as well!
Screenshot_20250629_011415

## Reviewer Test Plan

Enter any input within the CLI and observe the new formatting.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ✅  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

### DIFF
--- a/packages/cli/src/ui/components/messages/UserMessage.tsx
+++ b/packages/cli/src/ui/components/messages/UserMessage.tsx
@@ -19,7 +19,7 @@ export const UserMessage: React.FC<UserMessageProps> = ({ text }) => {
   return (
     <Box
       borderStyle="round"
-      borderColor="gray"
+      borderColor={Colors.Gray}
       flexDirection="row"
       paddingX={2}
       paddingY={0}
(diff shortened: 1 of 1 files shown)


allenhutchison (approved):
Thanks for this UI improvement! The border around the user's message makes the conversation easier to follow. Merging now.

PR #2511, opened 2025-06-29
Optionally specify an output path for non-interactive mode

mtrofin:
When using the non-interactive mode in a scripting tool, the output
is polluted with logging messages, which may contain sandboxing outputs
as well. Passing `-o <path>` allows capturing the clean agent output.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | Y  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | Y  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

ISSUE: #2748


mtrofin:
Tagging a few folks as per @eddie-santos 's suggestion: @scottdensmore @keithballinger @mattKorwel @NTaylorMullen @allenhutchison 
(unable to add reviewers)

NTaylorMullen:
Hmmm arguably I think we should be suppressing logging in nonInteractive mode as an alternative

mtrofin:
> Hmmm arguably I think we should be suppressing logging in nonInteractive mode as an alternative

Logging, and also outputs from e.g. docker (like useradd complaining about UID value range).

But, indeed, and that was my initial thought, too, but it looks like node's console.log loves stdout. Not having much of a JS/node experience myself, I imagined this may be some kind of a preference in that community.

My thinking was, if console.log is potentially used by library dependencies as well, and given tools (like docker) would reasonably use stdout for some messaging, it may be less maintainable to chase all these aspects than the opposite - place the interesting stream in a specific place (i.e. this PR) In a tooled set up, I'd create a named pipe, for example.

allenhutchison:
Hi @mtrofin sorry for the long delay. Are you still interested in pursuing this PR? If so, can you please resolve the merge conflicts?

mtrofin:
> Hi @mtrofin sorry for the long delay. Are you still interested in pursuing this PR? If so, can you please resolve the merge conflicts?

Yes - should have an update later today.

Thanks!

mtrofin:
> > Hi @mtrofin sorry for the long delay. Are you still interested in pursuing this PR? If so, can you please resolve the merge conflicts?
> 
> Yes - should have an update later today.
> 
> Thanks!

Done, ptal

allenhutchison (commented):
I'm a little on the fence for this one. I'm a little concerned that we are gathering too many command line options and I'm wondering if the better approach would be to add some tests that would fail for non-model output on STDOUT? 

One of the things I have been trying to enforce in this application is that framework messages go to STDERR and model output goes to STDOUT. If you are finding that isn't the case we should address that.

Once that is clean then you can get STDOUT in all the usual ways you would expect for a cli application (i.e. `gemini --prompt "Hi gemini" >> outfile.txt`).

Can you tell me a little more about the cases where you are seeing framework data in STDOUT?

mtrofin:
> I'm a little on the fence for this one. I'm a little concerned that we are gathering too many command line options and I'm wondering if the better approach would be to add some tests that would fail for non-model output on STDOUT?
> 
> One of the things I have been trying to enforce in this application is that framework messages go to STDERR and model output goes to STDOUT. If you are finding that isn't the case we should address that.
> 
> Once that is clean then you can get STDOUT in all the usual ways you would expect for a cli application (i.e. `gemini --prompt "Hi gemini" >> outfile.txt`).
> 
> Can you tell me a little more about the cases where you are seeing framework data in STDOUT?

at 820169ba2e0777ee2bee240f063649cc4b2b107f:

```
$ npm start -- -s -p "hi"

Checking build status...
Build is up-to-date.
hopping into sandbox (command: docker) ...
Checking for sandbox image: us-docker.pkg.dev/gemini-code-dev/gemini-cli/sandbox:0.1.15
Sandbox image us-docker.pkg.dev/gemini-code-dev/gemini-cli/sandbox:0.1.15 found locally.
INFO: Defaulting to use current user UID/GID for Debian/Ubuntu-based Linux.
useradd warning: gemini's uid <REDACTED> outside of the UID_MIN 1000 and UID_MAX 60000 range.
Hi there! How can I help you today?
```

let's split stderr and stdout:

```
$ npm start -- -s -p "hi" > /tmp/out 2> /tmp/err


$ cat /tmp/out
Checking build status...
Build is up-to-date.
Checking for sandbox image: us-docker.pkg.dev/gemini-code-dev/gemini-cli/sandbox:0.1.15
Sandbox image us-docker.pkg.dev/gemini-code-dev/gemini-cli/sandbox:0.1.15 found locally.
useradd warning: gemini's uid <REDACTED> outside of the UID_MIN 1000 and UID_MAX 60000 range.
Hi there! How can I help you today?

$ cat /tmp/err
hopping into sandbox (command: docker) ...
INFO: Defaulting to use current user UID/GID for Debian/Ubuntu-based Linux.
```

To be clear, the "build is up-to-date" bit is understandable :) but the rest is illustrative of the problem.

allenhutchison:
I agree there is a problem here, but I think the root cause is the lack of a centralized logger in Gemini CLI that would keep the STDOUT clean. What if we did something more like #5602

mtrofin:
> I agree there is a problem here, but I think the root cause is the lack of a centralized logger in Gemini CLI that would keep the STDOUT clean. What if we did something more like #5602

I agree, plus the output of tools like docker would need to be appropriately piped, too. The latter is probably not that hard to keep in check.

I can close this PR if things are moving in that (==#5602) direction?

allenhutchison:
I just merged in #5624 which should address this issue. In your output example there was a few lines from our node scripts that are a byproduct of running with `npm run start` those messages will still be in STDOUT, but if you run from the installed package `gemini --prompt "Hi Gemini"` you shouldn't see anything but the model output on STDOUT.

mtrofin:
> I just merged in #5624 which should address this issue. In your output example there was a few lines from our node scripts that are a byproduct of running with `npm run start` those messages will still be in STDOUT, but if you run from the installed package `gemini --prompt "Hi Gemini"` you shouldn't see anything but the model output on STDOUT.

Thanks - I'll close this. We still have the chatter from docker, I can try addressing that, but in a different PR then.

PR #2526, opened 2025-06-29
feat(cli): add explicit proxy option in cli

warjiang:
## TLDR

`gemini-cli` will reuse environment `HTTP(S)_PROXY`(both in upper and lower case), this PR will 
1.  provide extra options, use `--proxy` option to set proxy for gemini-client.
2. set transport layer with explicit proxy instead of environment variable.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


allenhutchison (dismissed):
Thanks for adding this important feature! The implementation looks good. Before we can merge this, could you please add some tests to verify that the proxy setting is correctly applied? Also, could you please add documentation for the new --proxy option to the docs/cli/configuration.md file? Also the web-fetch tool has it's own fallback http fetch that should be included in these changes.

warjiang:
> Thanks for adding this important feature! The implementation looks good. Before we can merge this, could you please add some tests to verify that the proxy setting is correctly applied? Also, could you please add documentation for the new --proxy option to the docs/cli/configuration.md file? Also the web-fetch tool has it's own fallback http fetch that should be included in these changes.

thanks for your advice:
1. I've add extra UT for proxy options.
2. Updated the cli document for `proxy` options.
3. Set proxy for `web-fetch` tool if `proxy` option set.

allenhutchison:
This is making good progress, but it looks like you have some failing tests, and need to resolve some merge conflicts.

warjiang:
> This is making good progress, but it looks like you have some failing tests, and need to resolve some merge conflicts.

Yeah, i'm working on it ~

warjiang:
> > This is making good progress, but it looks like you have some failing tests, and need to resolve some merge conflicts.
> 
> Yeah, i'm working on it ~

@allenhutchison I've tested it locally, could plz re-run the ci task?

allenhutchison:
Thanks for the great work on this PR, @warjiang! Adding proxy support is a fantastic improvement.

I did a deeper dive into the codebase to see if there were any other places that make network requests, and I found a few spots that might also need to be updated to use the new proxy setting.

Here are the files I noticed:
   1. `packages/core/src/core/modelCheck.ts`: The fetch call on line 46 doesn't seem to be using the proxy.
   2. `packages/core/src/telemetry/clearcut-logger/clearcut-logger.ts`: The https.request call on line 110 for sending telemetry data could also benefit from the proxy.
   3. `scripts/telemetry_utils.js`: The getJson function uses curl to fetch release information from GitHub. This curl command should probably include the proxy as well.

Let me know what you think! Happy to discuss further.

warjiang:
> Thanks for the great work on this PR, @warjiang! Adding proxy support is a fantastic improvement.
> 
> I did a deeper dive into the codebase to see if there were any other places that make network requests, and I found a few spots that might also need to be updated to use the new proxy setting.
> 
> Here are the files I noticed:
> 
> 1. `packages/core/src/core/modelCheck.ts`: The fetch call on line 46 doesn't seem to be using the proxy.
> 2. `packages/core/src/telemetry/clearcut-logger/clearcut-logger.ts`: The https.request call on line 110 for sending telemetry data could also benefit from the proxy.
> 3. `scripts/telemetry_utils.js`: The getJson function uses curl to fetch release information from GitHub. This curl command should probably include the proxy as well.
> 
> Let me know what you think! Happy to discuss further.

thanks for you comprehensive advice, 
1. for `packages/core/src/core/modelCheck.ts`, you're right, and I add the proxy when fetch model info:
    https://github.com/google-gemini/gemini-cli/blob/8e3d6c43e6b9d9f99704d2df319ddca00efe0f98/packages/core/src/core/modelCheck.ts#L48-L58

2.  for `packages/core/src/telemetry/clearcut-logger/clearcut-logger.ts`, because the request use `https.request` not build-in `fetch` method, so need extra proxy-agent for proxy request, we follow the implementation of `kubernetes-client/javascript` (I've tested  the implementation of proxy, and I thinks it's robus enough)

3.  for `scripts/telemetry_utils.js`, I thinks we should split another PR to add feature of proxy, reasons as following:
     - otel is started through sub process, and the config for otel is load from setting file instead of options, when we start gemini-cli, we'll not start the otel collector directly, the collector is started by running `npm run telemetry` or `npm run start:gcp`, the `proxy` option should be add here not in cli.
     -  currently, the way to parse command line arguments is simple, it read the process.argv and match the keyword like:
         https://github.com/google-gemini/gemini-cli/blob/edd69cb7d4dd616049fe604be10067683247568d/scripts/telemetry.js#L52-L70
         here we need to read `--target` option to start gcp or local otel collector,  if we use `--target gcp`, the implementation above will not detect `target=gcp` correctly, and fallback to local collector
    
    so, , I want to add arguments parsing in another PR and add `proxy` options in the command line and setting file at the same time.

warjiang:
@allenhutchison could you help me recheck the PR. BTW the gemini-bot label the PR with `status/need-issue`, should I create extra issue related to the PR for better description?

packages/core/src/telemetry/clearcut-logger/clearcut-logger.ts
```diff
@@ -389,6 +399,20 @@ export class ClearcutLogger {
     this.flushToClearcut();
   }
 
+  getProxyAgent() {
+    const proxyUrl = this.config?.getProxy();
+    if (!proxyUrl) return undefined;
+    if (proxyUrl.startsWith('socks')) {
```
Dcatfly: 1. Both http and https protocols can be handled in HttpsProxyAgent
2. The `ProxyAgent` provided by `undici` currently does not support the socks type. If you want to add support for the socks type proxy, it is not enough to handle it here.

packages/core/src/tools/web-fetch.ts
```diff
@@ -80,6 +81,9 @@ export class WebFetchTool extends BaseTool<WebFetchToolParams, ToolResult> {
         type: 'object',
       },
     );
+    if (config.getProxy()) {
```
Dcatfly: `setGlobalDispatcher` will intercept all node fetches in the current program.
`setGlobalDispatcher` has already been called in [core/client.ts](https://github.com/google-gemini/gemini-cli/blob/85a1d814a7e6cba15de59505c82bf96798807eaf/packages/core/src/core/client.ts#L62), and the timing is early enough. Therefore, the undici-based proxy agent in this PR might be unnecessary.

packages/cli/src/config/config.ts
```diff
@@ -237,6 +243,7 @@ export async function loadCliConfig(
     },
     checkpointing: argv.checkpointing || settings.checkpointing?.enabled,
     proxy:
+      argv.proxy ||
```
Dcatfly, line 387: Is it necessary to add a configuration to set whether the proxy is enabled? Maybe reading from the env is enough.

packages/core/src/code_assist/oauth2.ts
```diff
@@ -53,10 +54,15 @@ export interface OauthWebLogin {
...
 }
 
-export async function getOauthClient(): Promise<OAuth2Client> {
+export async function getOauthClient(
+  proxy?: string | undefined,
+): Promise<OAuth2Client> {
   const client = new OAuth2Client({
     clientId: OAUTH_CLIENT_ID,
     clientSecret: OAUTH_CLIENT_SECRET,
+    transporter: new Gaxios({
```
Dcatfly: If `transporter` is not set, `OAuth2Client` will actually automatically read proxy settings from the environment. Both methods only support http(s).

Dcatfly (commented):
Hi, I saw your related PR at https://github.com/google-gemini/gemini-cli/pull/3017. The functionality of these two PRs is indeed repetitive, but I prefer the implementation in 3017, and I've left some comments for you. 😂

packages/core/src/telemetry/clearcut-logger/clearcut-logger.ts
warjiang: thanks for pointing that~(got the new usage of HttpsProxyAgent and HttpProxyAgent 👍 )
here use `https` seems not very good, the reason I add the implementation of socks5 protocol here is that, I think the `clearcut-logger` is an isolated implementation.

packages/cli/src/config/config.ts
warjiang: > Is it necessary to add a configuration to set whether the proxy is enabled? Maybe reading from the env is enough.

Use environment variable can works(I've tested it, it worked out ~), but provide explicit option will it easy to understand, also provide extra way to make gemini-cli works with other software, otherwise we need to set and unset the environment variable to achieve that.

packages/core/src/code_assist/oauth2.ts
warjiang: `OAuth2Client` extends `AuthClient`, the `transporter` is set in the `AuthClient`:


https://github.com/googleapis/google-auth-library-nodejs/blob/bd7e4e3465567bba6f2599f63ac8ed62c0027de9/src/auth/authclient.ts#L246-L275
image

if unset `transporter` for `OAuth2Client `, it will init a brand new instance using `new Gaxios(opts.transporterOptions)`, but in `Gaxios` it will not use environment variable directly. 


https://github.com/googleapis/gaxios/blob/cdef1bb3f4f382d4581e21ebebfae2aca2f34029/src/gaxios.ts#L70-L76
image

warjiang:
> > Thanks for the great work on this PR, @warjiang! Adding proxy support is a fantastic improvement.
> > I did a deeper dive into the codebase to see if there were any other places that make network requests, and I found a few spots that might also need to be updated to use the new proxy setting.
> > Here are the files I noticed:
> > 
> > 1. `packages/core/src/core/modelCheck.ts`: The fetch call on line 46 doesn't seem to be using the proxy.
> > 2. `packages/core/src/telemetry/clearcut-logger/clearcut-logger.ts`: The https.request call on line 110 for sending telemetry data could also benefit from the proxy.
> > 3. `scripts/telemetry_utils.js`: The getJson function uses curl to fetch release information from GitHub. This curl command should probably include the proxy as well.
> > 
> > Let me know what you think! Happy to discuss further.
> 
> thanks for you comprehensive advice,
> 
> 1. for `packages/core/src/core/modelCheck.ts`, you're right, and I add the proxy when fetch model info:
>    https://github.com/google-gemini/gemini-cli/blob/8e3d6c43e6b9d9f99704d2df319ddca00efe0f98/packages/core/src/core/modelCheck.ts#L48-L58
> 2. for `packages/core/src/telemetry/clearcut-logger/clearcut-logger.ts`, because the request use `https.request` not build-in `fetch` method, so need extra proxy-agent for proxy request, we follow the implementation of `kubernetes-client/javascript` (I've tested  the implementation of proxy, and I thinks it's robus enough)
> 3. for `scripts/telemetry_utils.js`, I thinks we should split another PR to add feature of proxy, reasons as following:
>    
>    * otel is started through sub process, and the config for otel is load from setting file instead of options, when we start gemini-cli, we'll not start the otel collector directly, the collector is started by running `npm run telemetry` or `npm run start:gcp`, the `proxy` option should be add here not in cli.
>    * currently, the way to parse command line arguments is simple, it read the process.argv and match the keyword like:
>      https://github.com/google-gemini/gemini-cli/blob/edd69cb7d4dd616049fe604be10067683247568d/scripts/telemetry.js#L52-L70
>      
>      here we need to read `--target` option to start gcp or local otel collector,  if we use `--target gcp`, the implementation above will not detect `target=gcp` correctly, and fallback to local collector
>    
>    so, , I want to add arguments parsing in another PR and add `proxy` options in the command line and setting file at the same time.

@Dcatfly  can we cooperate with the proxy feature. Apart from proxy for cli, I think there is need to add proxy feature for telemetry, like init the otel collector from github, report metrics to gcp endpoint, maybe there still need extra work about this.

packages/core/src/telemetry/clearcut-logger/clearcut-logger.ts
Dcatfly: Thanks for your reply.
What I mean is, if `socks` support is to be added, then all requests within the CLI should support `socks` proxy, as their proxy source is the same.

packages/core/src/code_assist/oauth2.ts
Dcatfly: In the official doc for the [google-auth-library](https://github.com/googleapis/google-auth-library-nodejs?tab=readme-ov-file#using-a-proxy), it is mentioned that proxy settings are obtained from the environment.
image

The specific code implementation is likely located around [here](https://github.com/googleapis/gaxios/blob/cdef1bb3f4f382d4581e21ebebfae2aca2f34029/src/gaxios.ts#L494).

packages/cli/src/config/config.ts
Dcatfly: You make a good point, but this would require all request-making points in the project to actively add the proxy, including `curl` in `telemetry_utils`, as well as libraries that might be introduced in the future, or even AI-generated shells. This might require some trade-offs.

packages/core/src/telemetry/clearcut-logger/clearcut-logger.ts
warjiang: @Dcatfly  Agree with you, so I co-authored with you to re-use `https-proxy-agent` for `http/https` proxy protocol, for `socks` related protocol,  it will report the unsupported error 

related commit=> https://github.com/google-gemini/gemini-cli/pull/2526/commits/e13ef756f5ed224a5a55969c91770682c04b58ed#diff-053150b640a7ce75eff69d1a22cae7f0f94ad64ce9a855db544dda0929316519

warjiang:
I find the problem of ci, the cause is the trailing space at the end of the comment line
image

howerver, run `format` task in local machine(for me is macos) cannot find the error.

packages/core/src/telemetry/clearcut-logger/clearcut-logger.ts
```diff
@@ -413,6 +421,20 @@ export class ClearcutLogger {
     });
   }
 
+  getProxyAgent() {
+    const proxyUrl = this.config?.getProxy();
+    if (!proxyUrl) return undefined;
+    // undici which is widely used in the repo can only support http & https proxy protocol,
+    // https://github.com/nodejs/undici/issues/2224
+    // as @Dcatfly metioned, `HttpsProxyAgent` can support https & http proxy protocol at the same time,
```
NTaylorMullen: Can drop this line and the next given they're a bit more code reviewy for a comment

packages/core/src/tools/web-fetch.ts
NTaylorMullen:
```suggestion
    const proxy = config.getProxy();
    if (proxy) {
```

And then can re-use below

NTaylorMullen (changes requested):
Excuse my ignorance here but isn't global dispatcher truly "global"? Are we not able to set it on application start in gemini.ts vs. set it everywhere individually?

packages/core/src/tools/web-fetch.ts
warjiang: thanks for your advice, accept changes~

packages/core/src/telemetry/clearcut-logger/clearcut-logger.ts
warjiang: ok, I'll delete these two lines

warjiang:
> `setGlobalDispatcher` only means global for `fetch` api


> Excuse my ignorance here but isn't global dispatcher truly "global"? Are we not able to set it on application start in gemini.ts vs. set it everywhere individually?

yes, reasons as follwowing:
1. we use `https` and `fetch` at the same time, the way to set proxy is different from each other.
2. if take telemetry into consideration, it's need extra efforts, like  handling shell command like `curl` for proxy.
3. the way to set proxy is more like ensuring logic even if the global setting not make effective

NTaylorMullen:
> we use https and fetch at the same time, the way to set proxy is different from each other.

That's ok. I know Allen called out the location BUT replacing the clearcut logger proxy pieces is something we can leave out of this PR.

> if take telemetry into consideration, it's need extra efforts, like handling shell command like curl for proxy.

Ya I'm not sure you'll ever be able to fully proxy everything on the machine unfortunately

> the way to set proxy is more like ensuring logic even if the global setting not make effective

I'm not quite sure I follow. What if we were to remove all of the setGlobalDispatcher pieces from the individual call sites and only do it once inside of gemini.ts. Would that not work given its the entry point?

warjiang:
> Ya I'm not sure you'll ever be able to fully proxy everything on the machine unfortunately

I thinks that's not a problem currently, for cli, we are proxy every network request now. For the part of otel collector, I think we should have better command arguments parser first(the implementation of otel collector is very simple, it use string match to check whether use gcp or local provider, it works but not very good 😭 ), then we can add the proxy feature(that will be easy).

> I'm not quite sure I follow. What if we were to remove all of the setGlobalDispatcher pieces from the individual call sites and only do it once inside of gemini.ts. Would that not work given its the entry point?

Not exactly, first we should ensure every network request should use node.js build-in `fetch` method instead of `https` module, as you metioned, Allen is trying to replace `https` with `fetch`, after that, I think we can remove all pieces of `setGlobalDispatcher` and set proxy once at the entrypoint(that will make the code clean), the proxy will work for fetch inside gemini, but for third-party like `gaxios` and `curl` command, I think there still some needs to set proxy case-by-case.

NTaylorMullen (dismissed):

NTaylorMullen:
@warjiang this is looking good. Would you mind handling some of these merge conflicts + `npm run format`?

warjiang:
@NTaylorMullen confilct already been resolved, could we re-run the ci task and  move forward this feature

NTaylorMullen:
@warjiang ugh I'm sorry. Looks like a few more conflicts have popped up. The repo is moving so fast!

warjiang:
> @warjiang ugh I'm sorry. Looks like a few more conflicts have popped up. The repo is moving so fast!

@NTaylorMullen  ya, it moved really so fast.  All conflicts has already been resolved.

warjiang:
after bump gaxios from `6.7.1.` to `7.1.1`, there will be some type error for setting `transporter` field, so I use `transporterOptions` as alternative.
image
image

> the effect is equal ~


@NTaylorMullen all conflict have been resolved, could you plz take another look at it ?

warjiang:
ci task passed in my fork repo, so it just need one approval:
image

NTaylorMullen (dismissed):

warjiang:
@allenhutchison hi, allen, can we push these feature forward ?

NTaylorMullen (dismissed):

NTaylorMullen:
@warjiang I tried resolving conflicts for you via the GitHub UI but looks like it missed a bit 😢 . Would you mind fixing the build real quick and re-pushing. Also could you `@` me directly once you're done. Apologeis for all this back and forth. There's been so much volume in the repo it's hard to keep up so the `@`'s have been helpful

warjiang:
> @warjiang I tried resolving conflicts for you via the GitHub UI but looks like it missed a bit 😢 . Would you mind fixing the build real quick and re-pushing. Also could you `@` me directly once you're done. Apologeis for all this back and forth. There's been so much volume in the repo it's hard to keep up so the `@`'s have been helpful

Never mind, I know it's really hard to maintain a popular oss repo, especially the repo also iterated in a high speed.

warjiang:
@NTaylorMullen all conflicts have been resolved and re-tested in forked repo. Let's move forward ~

NTaylorMullen (approved):

PR #2609, opened 2025-06-29
Adds --hide-footer and /toggle-footer features

oneness:
feat: Add --hide-footer switch and /toggle-footer command

## TLDR

This pull request introduces a `--hide-footer` command-line switch and a `/toggle-footer` interactive command, allowing users to control the visibility of the CLI footer.

## Dive Deeper

The `--hide-footer` switch provides a mechanism to launch the Gemini CLI without the persistent footer, which is particularly useful for scripting, automated tasks, or users who prefer a minimalist interface. The `/toggle-footer` command offers dynamic control within an active interactive session, enabling users to show or hide the footer on demand. This enhancement improves user experience by providing greater flexibility in managing screen real estate and visual output according to individual preferences or operational needs.

## Reviewer Test Plan

To validate these changes, please perform the following steps:

1.  **Verify `--hide-footer` switch:**
    *   Run the CLI with the new switch: `npm run cli -- --hide-footer`
    *   Confirm that the footer is not visible upon application startup.
2.  **Verify `/toggle-footer` command:**
    *   Start the CLI in interactive mode: `npm run cli`
    *   Observe that the footer is initially visible.
    *   Type `/toggle-footer` and press Enter. Confirm the footer disappears.
    *   Type `/toggle-footer` again and press Enter. Confirm the footer reappears.
3.  **Combined Test:**
    *   Start the CLI with the `--hide-footer` switch: `npm run cli -- --hide-footer`
    *   Confirm the footer is hidden.
    *   Type `/toggle-footer` and press Enter. Confirm the footer appears.
    *   Type `/toggle-footer` again and press Enter. Confirm the footer disappears.

## Testing Matrix

|          | npm run                               | npx                                   | Docker                                | Podman                                | Seatbelt                              |
| -------- | ------------------------------------- | ------------------------------------- | ------------------------------------- | ------------------------------------- | ------------------------------------- |
| npm run  | ✅ (cli, cli -- --hide-footer)        | ✅ (gemini-cli, gemini-cli --hide-footer) | ❓ (build and run image with options) | ❓ (build and run image with options) | ❓ (run with seatbelt enabled)        |
| npx      | ✅ (cli, cli -- --hide-footer)        | ✅ (gemini-cli, gemini-cli --hide-footer) | ❓ (build and run image with options) | ❓ (build and run image with options) | ❓ (run with seatbelt enabled)        |
| Docker   | ❓ (build and run image with options) | ❓ (build and run image with options) | ❓ (build and run image with options) | ❓ (build and run image with options) | ❓ (run with seatbelt enabled)        |
| Podman   | ❓ (build and run image with options) | ❓ (build and run image with options) | ❓ (build and run image with options) | ❓ (build and run image with options) | ❓ (run with seatbelt enabled)        |
| Seatbelt | ❓ (run with seatbelt enabled)        | ❓ (run with seatbelt enabled)        | ❓ (run with seatbelt enabled)        | ❓ (run with seatbelt enabled)        | ❓ (run with seatbelt enabled)        |

## Linked issues / bugs

N/A

### DIFF
--- a/packages/core/src/core/coreToolScheduler.test.ts
+++ b/packages/core/src/core/coreToolScheduler.test.ts
@@ -407,3 +407,123 @@ describe('convertToFunctionResponse', () => {
     });
   });
 });
+
+describe('CoreToolScheduler edit cancellation', () => {
+  it('should preserve diff when an edit is cancelled', async () => {
+    class MockEditTool extends BaseTool<Record<string, unknown>, ToolResult> {
+      constructor() {
+        super(
+          'mockEditTool',
+          'mockEditTool',
+          'A mock edit tool',
+          Icon.Pencil,
+          {},
+        );
+      }
+
+      async shouldConfirmExecute(
+        _params: Record<string, unknown>,
+        _abortSignal: AbortSignal,
+      ): Promise<ToolCallConfirmationDetails | false> {
+        return {
+          type: 'edit',
+          title: 'Confirm Edit',
+          fileName: 'test.txt',
+          fileDiff:
+            '--- test.txt\n+++ test.txt\n@@ -1,1 +1,1 @@\n-old content\n+new content',
+          originalContent: 'old content',
+          newContent: 'new content',
+          onConfirm: async () => {},
+        };
+      }
+
+      async execute(
+        _params: Record<string, unknown>,
+        _abortSignal: AbortSignal,
+      ): Promise<ToolResult> {
+        return {
+          llmContent: 'Edited successfully',
+          returnDisplay: 'Edited successfully',
+        };
+      }
+    }
+
+    const mockEditTool = new MockEditTool();
+    const toolRegistry = {
+      getTool: () => mockEditTool,
+      getFunctionDeclarations: () => [],
+      tools: new Map(),
+      discovery: {} as any,
+      registerTool: () => {},
+      getToolByName: () => mockEditTool,
+      getToolByDisplayName: () => mockEditTool,
+      getTools: () => [],
+      discoverTools: async () => {},
+      getAllTools: () => [],
+      getToolsByServer: () => [],
+    };
+
+    const onAllToolCallsComplete = vi.fn();
+    const onToolCallsUpdate = vi.fn();
+
+    const mockConfig = {
+      getSessionId: () => 'test-session-id',
+      getUsageStatisticsEnabled: () => true,
+      getDebugMode: () => false,
+    } as unknown as Config;
+
+    const scheduler = new CoreToolScheduler({
+      config: mockConfig,
+      toolRegistry: Promise.resolve(toolRegistry as any),
+      onAllToolCallsComplete,
+      onToolCallsUpdate,
+      getPreferredEditor: () => 'vscode',
+    });
+
+    const abortController = new AbortController();
+    const request = {
+      callId: '1',
+      name: 'mockEditTool',
+      args: {},
+      isClientInitiated: false,
+      prompt_id: 'prompt-id-1',
+    };
+
+    await scheduler.schedule([request], abortController.signal);
+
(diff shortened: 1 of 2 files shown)


packages/cli/src/config/config.ts
```diff
@@ -122,6 +123,11 @@ async function parseArguments(): Promise<CliArgs> {
       description:
         'Enable or disable logging of user prompts for telemetry. Overrides settings files.',
     })
+    .option('hide-footer', {
```
jacob314: rather than adding this to config can we instead add it to settings.json with a property added to settings.ts.
Generally settings.json is a bette place for options that are small UI tweaks than command line flags. If there is a use case where a user wants to frequently change between showing the footer and not showing the footer then I would be open to including it in config.ts

packages/cli/src/ui/hooks/slashCommandProcessor.ts
```diff
@@ -584,10 +585,17 @@ export const useSlashCommandProcessor = (
       },
       {
         name: 'corgi',
-        action: (_mainCommand, _subCommand, _args) => {
+        action: async (_mainCommand, _subCommand, _args) => {
           toggleCorgiMode();
         },
       },
+      {
+        name: 'toggle-footer',
```
jacob314: do users really want to toggle and show the footer within a session?
Would you still want this if the footer had a mode where it never took up more than a single line? I agree it can be fairly annoying today when it takes up multiple lines.

danielgwilson: I think it actually causes a major issue on some terminals—not certain, but I think the footer is actually causing major problems for me on Warp. I'll try to get a screenshot, but the footer seems to be dynamic in height or something—and it kind of pushes off the entirety of the CLI UI upwards on the screen. Sometimes it then actually blanks it out until the next workflow step completes.

danielgwilson: Might be related, might be a separate issue

oneness: > do users really want to toggle and show the footer within a session? Would you still want this if the footer had a mode where it never took up more than a single line? I agree it can be fairly annoying today when it takes up multiple lines.

I'd say depends but most likely yes considering the footer could have more diagnostic or meta information that is useful to the session or users in the future.  

Also,  I had to point out that the original issue where it does flicker from time to time and might be related what @danielgwilson mentioned above.

packages/cli/src/config/config.ts
oneness: @jacob314 Thanks for reviewing the change. I am ok with adding it to settings.json though I'd like to point out following so you can weigh the pro and cons before we reach a way forward:

1. Footer does have important use cases for diagnostics and at a glance information. 
2. The ability to turn it on and off could be very useful for cases where one needs the extra screen space while periodically check context window (other info you guys end up putting there). 

Let me know what you think.

packages/cli/src/ui/hooks/slashCommandProcessor.ts
jacob314: > I think it actually causes a major issue on some terminals—not certain, but I think the footer is actually causing major problems for me on Warp. I'll try to get a screenshot, but the footer seems to be dynamic in height or something—and it kind of pushes off the entirety of the CLI UI upwards on the screen. Sometimes it then actually blanks it out until the next workflow step completes.

that is not due to the footer.  That is Ink being clever to try to keep other UI in the app in a stable location.  Good news is @SandyTao520 recently landed changes that reduce the number of times where you get extra spaces at the bottom of the UI. Now a lot of the times when Ink includes the extra spaces for us, it actually makes UI positions feel more stable rather than being annoying (there was a bug in how were were streaming messages in that made the problem worse).

PR #2652, opened 2025-06-30
Refactor text-buffer to use reducer

SandyTao520:
## TLDR

To address bugs caused by React 19's batching behavior with rapid state updates, we refactored text-buffer to use a reducer for text operations.

## Dive Deeper

With the update to React 19 in https://github.com/google-gemini/gemini-cli/pull/2096, State updates are now batched, with only the final update in each batch taking effect, intermediate updates might be dropped.

Rapid text insertions (like those from an IME or drag-and-drop) are failing, with only the last character being inserted. To fix this, we need to migrate to using a reducer. This is the standard way to handle a series of state updates that depend on previous states.

Besides migrating all actions to use reducer, we also cleaned up ``applyOperations``, which is the previous attempt to queue updates. Removed logic with selection and copy/paste as they are currently unused.

## Reviewer Test Plan

Test these actions in input prompt:
1. navigate through history
2. insert text
3. paste text
4. backspace/delete
5. move up/down/left/right with ``alt/option + direction``
6. delete word left (ctrl + w)
7. delete line left/right (ctrl + u/k)
8. auto-completion

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅ | ❓  | ❓  |
| npx      | ✅ | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ✅ | -   | -   |

## Linked issues / bugs

[
](https://github.com/google-gemini/gemini-cli/pull/2504)


packages/cli/src/ui/components/shared/text-buffer.ts
```diff
@@ -395,187 +395,103 @@ function calculateVisualLayout(
...
+}
 
-    if (visualCursor[0] < visualScrollRow) {
-      newVisualScrollRow = visualCursor[0];
-    } else if (visualCursor[0] >= visualScrollRow + height) {
-      newVisualScrollRow = visualCursor[0] - height + 1;
+const historyLimit = 100;
+
+type TextBufferAction =
+  | { type: 'SET_TEXT'; payload: string; pushToUndo?: boolean }
```
jacob314: super minor nit: can you ask gemini CLI to make these all lower case. I think that is generally more coherent with the rest of the code base unless I'm missing something.

packages/cli/src/ui/components/shared/text-buffer.ts
```diff
@@ -1039,140 +634,552 @@ export function useTextBuffer({
...
           }
-          break;
+          return {
+            ...state,
+            cursorRow: newCursorRow,
+            cursorCol: newCursorCol,
+            preferredCol: null,
+          };
         }
+        // wordLeft and wordRight are complex and better handled by dispatching from the component
```
jacob314: is this accurate? I see an impl of wordLeft and wordRight above.

packages/cli/src/ui/components/shared/text-buffer.ts
```diff
@@ -1039,140 +634,552 @@ export function useTextBuffer({
...
+    }
+
+    case 'START_SELECTION': {
+      return {
+        ...state,
+        selectionAnchor: [state.cursorRow, state.cursorCol],
+      };
+    }
+
+    // Other actions like DELETE_WORD_LEFT, REPLACE_RANGE etc. would be implemented here.
```
jacob314: this comment is also misleading. likely stuff Gemini CLI left while doing this refactor. I verified all these methods exist. Go ahead and remove this misleading comment.

packages/cli/src/ui/components/shared/text-buffer.ts
```diff
@@ -1039,140 +634,552 @@ export function useTextBuffer({
...
+    // For brevity, they are omitted but would follow a similar pattern of taking state,
+    // performing operations on copies, and returning a new state object.
+    default:
+      return state;
+  }
+
+  return state;
```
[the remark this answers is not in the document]
jacob314: agreed. remove

packages/cli/src/ui/components/shared/text-buffer.ts
```diff
@@ -1039,140 +634,552 @@ export function useTextBuffer({
...
+
+  const [visualScrollRow, setVisualScrollRow] = useState<number>(0);
+
+  useEffect(() => {
+    if (onChange) {
+      onChange(text);
+    }
+  }, [text, onChange]);
+
+  // Update visual scroll (vertical)
```
jacob314, line 1008: btw the diff is smaller than it looks as this is just a move.

packages/cli/src/ui/components/shared/text-buffer.ts
```diff
@@ -395,187 +395,103 @@ function calculateVisualLayout(
...
+        cursorRow: lastNewLineIndex,
+        cursorCol: cpLen(lines[lastNewLineIndex] ?? ''),
+        preferredCol: null,
+      };
+    }
 
-  const applyOperations = useCallback(
-    (ops: UpdateOperation[]) => {
-      if (ops.length === 0) return;
+    case 'APPLY_OPERATIONS': {
```
jacob314: here Gemini CLI lost the thread a little. APPLY_OPERATIONS was just my attempt at a reducer without a reducer. Get this folded in with the core reducer functionality.

packages/cli/src/ui/components/shared/text-buffer.ts
```diff
@@ -1039,140 +634,552 @@ export function useTextBuffer({
...
+      return {
+        ...state,
+        selectionAnchor: [state.cursorRow, state.cursorCol],
+      };
+    }
+
+    // Other actions like DELETE_WORD_LEFT, REPLACE_RANGE etc. would be implemented here.
+    // For brevity, they are omitted but would follow a similar pattern of taking state,
+    // performing operations on copies, and returning a new state object.
+    default:
```
KeijiBranshi: nitpick: it's a little crunchy, but I like making `default` cases perform exhaustive checks. eg https://github.com/google-gemini/gemini-cli/blob/505a5d617b4bfa02ea363d3a3e0ae0bc178ddc58/packages/cli/src/ui/hooks/useReactToolScheduler.ts#L199-L203

this just helps us enforce all cases are handled at compile time.

packages/cli/src/ui/components/shared/text-buffer.ts
```diff
@@ -395,187 +395,103 @@ function calculateVisualLayout(
...
-
-  useEffect(() => {
-    if (onChange) {
-      onChange(text);
+  | { type: 'MOVE_TO_OFFSET'; payload: { text: string; offset: number } }
+  | { type: 'COPY' }
+  | { type: 'PASTE' }
+  | { type: 'START_SELECTION' };
+
+function textBufferReducer(
```
KeijiBranshi: suggestion(non-blocking): now with state transition captured in a pure function, we should be able to easily vibe-code up some clean unit tests for this. i'm not sure how we can guarantee this stays pure (ignoring `dbg`) with static analysis tools like eslint, so good tests are important here.

KeijiBranshi (commented):
quick first pass

packages/cli/src/ui/components/shared/text-buffer.ts
jacob314: +1 to this. the more strict we make things the better Gemini CLI can course correct if it forgets cases.

packages/cli/src/ui/components/shared/text-buffer.ts
```diff
@@ -395,561 +388,177 @@ function calculateVisualLayout(
...
-  useEffect(() => {
-    const { height } = viewport;
-    let newVisualScrollRow = visualScrollRow;
+const historyLimit = 100;
 
-    if (visualCursor[0] < visualScrollRow) {
-      newVisualScrollRow = visualCursor[0];
-    } else if (visualCursor[0] >= visualScrollRow + height) {
-      newVisualScrollRow = visualCursor[0] - height + 1;
+type TextBufferAction =
```
jacob314, line 392: this is so much better than the previous code!

packages/cli/src/ui/components/shared/text-buffer.ts
```diff
@@ -1002,140 +611,502 @@ export function useTextBuffer({
...
-      visualLines,
-      preferredCol,
-      lines,
-      currentLineLen,
-      visualToLogicalMap,
-      logicalToVisualMap,
-      cursorCol,
-      cursorRow,
-    ],
+    [cursorRow, cursorCol, isValidPath],
```
jacob314: nit: cursorRow and cursorCol are only needed for the dbg logging which isn't crucial. I'd suggest stripping all the dbg logging and seeing which dependencies can be removed. Generally very few of these methods should now need to take dependencies.

packages/cli/src/ui/components/shared/text-buffer.ts
```diff
@@ -1002,140 +611,502 @@ export function useTextBuffer({
...
   );
 
+  const newline = useCallback((): void => {
+    dispatch({ type: 'insert', payload: '\n' });
+  }, []);
+
+  const backspace = useCallback((): void => {
+    if (cursorCol === 0 && cursorRow === 0) return;
+    dispatch({ type: 'backspace' });
+  }, [cursorRow, cursorCol]);
```
jacob314: remove cursorRow and cursorCol from this

packages/cli/src/ui/components/shared/text-buffer.ts
```diff
@@ -1002,140 +611,502 @@ export function useTextBuffer({
...
-    ],
+    [cursorRow, cursorCol, isValidPath],
   );
 
+  const newline = useCallback((): void => {
+    dispatch({ type: 'insert', payload: '\n' });
+  }, []);
+
+  const backspace = useCallback((): void => {
+    if (cursorCol === 0 && cursorRow === 0) return;
```
jacob314: remove this line as it isn't needed as backspace itself should handle doing nothing if the buffer is empty

packages/cli/src/ui/components/shared/text-buffer.ts
```diff
@@ -1002,140 +611,502 @@ export function useTextBuffer({
...
+    dispatch({ type: 'backspace' });
+  }, [cursorRow, cursorCol]);
+
+  const del = useCallback((): void => {
+    dispatch({ type: 'delete' });
+  }, []);
+
+  const move = useCallback(
+    (dir: Direction): void => {
+      dispatch({ type: 'move', payload: { dir, visualLayout } });
```
jacob314: passing in visualLayout to this in the payload isn't right. make the payload just the dir and compute the visual layout for the actual lines at the state where this event is processed by the reducer.

packages/cli/src/ui/components/shared/text-buffer.ts
```diff
@@ -1002,140 +611,502 @@ export function useTextBuffer({
...
+    },
+    [visualLayout],
+  );
+
+  const undo = useCallback((): boolean => {
+    dispatch({ type: 'undo' });
+    return state.undoStack.length > 0;
+  }, [state.undoStack.length]);
```
jacob314: remove 
state.undoStack.length  from this and change the return type to this to void. can't imagine we care.

packages/cli/src/ui/components/shared/text-buffer.ts
```diff
@@ -1002,140 +611,502 @@ export function useTextBuffer({
...
+
+  const redo = useCallback((): boolean => {
+    dispatch({ type: 'redo' });
+    return state.redoStack.length > 0;
+  }, [state.redoStack.length]);
```
jacob314: same change as for undo

packages/cli/src/ui/components/shared/text-buffer.ts
```diff
@@ -1281,13 +1251,41 @@ export function useTextBuffer({
...
+        startRow > endRow ||
+        (startRow === endRow && startCol > endCol) ||
+        startRow < 0 ||
+        startCol < 0 ||
+        endRow >= lines.length ||
+        (endRow < lines.length && endCol > cpLen(lines[endRow]))
+      ) {
+        return false;
+      }
+      dispatch({
```
jacob314, line 1223: remove the code to `return false`.
should massively simplify this to just forward to calling dispatch.

packages/cli/src/ui/components/shared/text-buffer.ts
```diff
@@ -1297,13 +1295,9 @@ export function useTextBuffer({
 
   const moveToOffset = useCallback(
     (offset: number): void => {
-      const [newRow, newCol] = offsetToLogicalPos(text, offset);
-      setCursorRow(newRow);
-      setCursorCol(newCol);
-      setPreferredCol(null);
-      dbg('moveToOffset', { offset, newCursor: [newRow, newCol] });
+      dispatch({ type: 'move_to_offset', payload: { text, offset } });
```
jacob314: its a little odd this takes `text` as a parameter. Ideally look at that and remove if possible

packages/cli/src/ui/components/shared/text-buffer.ts
```diff
@@ -1002,140 +611,502 @@ export function useTextBuffer({
...
+        lines: newLines,
+        cursorRow: finalCursorRow,
+        cursorCol: finalCursorCol,
+        preferredCol: null,
+      };
+    }
+
+    case 'move_to_offset': {
+      const { text, offset } = action.payload;
+      const [newRow, newCol] = offsetToLogicalPos(text, offset);
```
jacob314: as mentioned bellow, get the logical position off the current text buffer computed from the current lines not the possible stale `text` object. we can then remove `text from the payload`

packages/cli/src/ui/components/shared/text-buffer.ts
```diff
@@ -1002,140 +611,502 @@ export function useTextBuffer({
...
+      };
+    }
+
+    case 'create_undo_snapshot': {
+      return pushUndo(state);
+    }
+
+    default: {
+      const exhaustiveCheck: never = action;
+      console.warn(`Unknown action encountered: ${exhaustiveCheck}`);
```
jacob314: nit: make an error or throw as this would indicate a significant logic bug probably resulting in the user wondering why the input prompt is behaving badly.

jacob314 (dismissed):
This looks really good. Went through text-buffer.ts fairly carefully. main thing that is missing is there are a couple cases where we didn't go quite far enough to push all the logic inside dispatch. LGTM once these nits are addressed.
lgtm

packages/cli/src/ui/components/shared/text-buffer.ts
SandyTao520: This involves adding another action for setting viewport width

jacob314 (approved):
lgtm

packages/cli/src/ui/components/shared/text-buffer.ts
```diff
@@ -1328,45 +1263,13 @@ export function useTextBuffer({
...
-      let selectedTextVal;
-      if (sr === er) {
-        selectedTextVal = cpSlice(currentLine(sr), sc, ec);
-      } else {
-        const parts: string[] = [cpSlice(currentLine(sr), sc)];
-        for (let r = sr + 1; r < er; r++) parts.push(currentLine(r));
-        parts.push(cpSlice(currentLine(er), 0, ec));
-        selectedTextVal = parts.join('\n');
-      }
-      setClipboard(selectedTextVal);
```
NTaylorMullen, line 1358: How do we handle the clipboard functionality now?

packages/cli/src/ui/components/shared/text-buffer.ts
```diff
@@ -1328,45 +1263,13 @@ export function useTextBuffer({
...
-      return selectedTextVal;
-    }, [selectionAnchor, cursorRow, cursorCol, currentLine, setClipboard]),
-    paste: useCallback(() => {
-      if (clipboard === null) return false;
-      applyOperations([{ type: 'insert', payload: clipboard }]);
-      return true;
-    }, [clipboard, applyOperations]),
-    startSelection: useCallback(
-      () => setSelectionAnchor([cursorRow, cursorCol]),
```
NTaylorMullen, line 1367: Did the old selection logic get lifted into the refactor in some way or is it no longer needed?

packages/cli/src/ui/components/shared/text-buffer.ts
SandyTao520: Selection (using Shift + arrow keys) was implemented but never bound to a key combination. This copy() function was intended to copy the selected text. We decided to remove them for simplicity until they're needed again.

Copy is handled by the system (cmd + c) now and paste is now captured in useKeyPress.ts and handled as an insert in text-buffer.ts.

packages/cli/src/ui/components/shared/text-buffer.ts
SandyTao520: Selection was never used, removed for simplicity.

NTaylorMullen (approved):
Amazing job. There was a lot here to dig through and should make our lives significantly easier in the future. :shipit:

PR #2653, opened 2025-06-30
feat(shell): Enable prefix matching for flexible command validation

jerop:
This commit evolves the run_shell_command tool from exact command matching to a more flexible and powerful prefix-based validation.

Previously, allow/block rules required an exact command string (e.g., "run_shell_command(git status)"). Now, you can specify a command prefix (e.g., "run_shell_command(git)"), which will correctly validate all subcommands like "git status" or "git log". This makes configuring command restrictions more practical and powerful.

To ensure this added flexibility remains secure:
- Commands chained with &&, ||, or ; are split, and each part is validated against the prefix rules.
- Command substitution using $(...) or `` `...` `` is explicitly blocked to prevent injection.

The documentation and test suite have been updated to reflect these changes.

Prior work in https://github.com/google-gemini/gemini-cli/pull/2605.

### DIFF
--- a/docs/tools/shell.md
+++ b/docs/tools/shell.md
@@ -64,96 +64,71 @@ run_shell_command(command="npm run dev &", description="Start development server
 
 You can restrict the commands that can be executed by the `run_shell_command` tool by using the `coreTools` and `excludeTools` settings in your configuration file.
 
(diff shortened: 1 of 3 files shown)


scottdensmore (approved):
Approved

Issue #1249, opened 2025-06-20
Project hashes do not need to be full 64 byte hex digest


louisjimenez:
The paths.ts file calculates a full 64 byte hex digest from the project path. It should be ok to truncate this to the first 8 ascii bytes. See comment https://github.com/google-gemini/gemini-cli/pull/1195#discussion_r2157961581

iamrajiv:
hi @louisjimenez i raised a pr on this can you review it once https://github.com/google-gemini/gemini-cli/pull/2022?

mattKorwel:
Closig as not needed.

PR #2699, opened 2025-06-30
fix: increase character length to 16 as suggested in #2022

yashwanth2706:
As suggested, I have increased the character length to 16 to avoid hash collisions.

Refference: #2022


yashwanth2706:
After reading the comment in #1249 
I realized that this was not a cryptographic issue so i have closed the pull request

Issue #1570, opened 2025-06-25
Sandbox fails with ARM64 arch


libbkmz:
When attempting to launch the gemini-cli sandbox on a Linux system with an  ARM64 (aarch64) architecture, the process fails. The logs indicate the CLI  attempts to run an `amd64` Docker image.
I run this on Raspberry PI Compute Module 5 which is `aarch64` 

   * Operating System: Debian Bookworm (12)
   * Host Architecture: linux/arm64/v8
   * Sandbox Image: us-docker.pkg.dev/gemini-code-dev/gemini-cli/sandbox:0.1.3
   * Sandbox Image Architecture: linux/amd64
   * `Linux raspberrypi 6.12.22-v8-16k+ #1 SMP PREEMPT Thu Apr 10 10:10:28 CEST 2025 aarch64 GNU/Linux`

```
.bin/gemini --sandbox
hopping into sandbox (command: docker) ...
Checking for sandbox image: us-docker.pkg.dev/gemini-code-dev/gemini-cli/sandbox:0.1.3
Sandbox image us-docker.pkg.dev/gemini-code-dev/gemini-cli/sandbox:0.1.3 found locally.
INFO: Defaulting to use current user UID/GID for Debian/Ubuntu-based Linux.
WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested
**
ERROR:../../accel/tcg/cpu-exec.c:954:cpu_exec: assertion failed: (cpu == current_cpu)
Bail out! ERROR:../../accel/tcg/cpu-exec.c:954:cpu_exec: assertion failed: (cpu == current_cpu)
bash: line 1:    12 Trace/breakpoint trap   (core dumped) groupadd -f -g 993 gemini
**
ERROR:../../accel/tcg/cpu-exec.c:954:cpu_exec: assertion failed: (cpu == current_cpu)
Bail out! ERROR:../../accel/tcg/cpu-exec.c:954:cpu_exec: assertion failed: (cpu == current_cpu)
bash: line 1:    15 Trace/breakpoint trap   (core dumped) useradd -o -u 1001 -g 993 -d /home/cpi -s /bin/bash gemini
Sandbox process exited with code: 133, signal: null

```

## Diagnostic Information
*   **CLI Version:** 0.1.3
*   **Git Commit:** 382486e (local modifications)
*   **Operating System:** linux v18.19.0
*   **Sandbox Environment:** no sandbox
[...]

PR #1226, opened 2025-06-19
Correct Node Version in README

scidomino:
## TLDR

README.md used to say we required Node 20 but we only require Node 18.

PR #1164, opened 2025-06-17
Make Node Versions match in docs

sethtroisi-test:
## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| -------- | --- | --- | --- |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

PR #2700, opened 2025-06-30
chore: Expand node version test matrix

swissspidy:
## TLDR

This expands the Node version test matrix for integration tests and removes it for the basic lint job, which doesn't need a matrix.

Adds an `.nvmrc` file for easier repo setup.

## Reviewer Test Plan

* GitHub test matrix should be expanded and all status checks should pass

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

n/a

## Linked issues / bugs

* #3113


.github/workflows/ci.yml
```diff
@@ -17,7 +17,7 @@ jobs:
       contents: read # For checkout
     strategy:
       matrix:
-        node-version: [20.x]
+        node-version: [18.x, 20.x, 22.x, 24.x]
```
NTaylorMullen: ooo I definitely like the idea of testing on multiple; however, I wonder if we can swap "when" we do this. For instance, we probably don't need to re-run every aspect of the build on each of these versions and could probably get away with running these along side our integration tests (cc @mattKorwel).

Would you mind taking a stab at this changeset from the lens of optimizing PR build times while also trying to keep us honest?

swissspidy:
Relevant bug report when trying to contribute using Node 18:

https://github.com/google-gemini/gemini-cli/issues/1570#issuecomment-3005891976

.github/workflows/ci.yml
swissspidy: I understand the build times concern. Though with the parallel execution of matrix job there shouldn't be any extra overhead.

Looking at it again, this can be improved though:

* No need to run linters on multiple Node versions
* No need to upload and download artifacts — just building in the same step is much faster

I just pushed some changes to reduce the complexity, adding a matrix for the integration tests while at it.

.github/workflows/ci.yml
```diff
@@ -45,24 +42,17 @@ jobs:
...
-    needs: build # This job depends on the 'build' job
+    needs: lint
     permissions:
       contents: read
       checks: write
       pull-requests: write
     strategy:
       matrix:
-        node-version: [20.x] # Should match the build job's matrix
+        node-version: [18.x, 20.x, 22.x, 24.x]
```
mattKorwel: Given the issues around not building with 18 and dependency on packages that need 20, i propose we don't include 18 here and just make the minimum version 20.

```suggestion
        node-version: [20.x, 22.x, 24.x]
```

.github/workflows/ci.yml
```diff
@@ -124,7 +106,7 @@ jobs:
       pull-requests: write # For commenting
     strategy:
       matrix:
-        node-version: [20.x] # Should match the test job's matrix
+        node-version: [18.x, 20.x, 22.x, 24.x] # Should match the test job's matrix
```
mattKorwel:
```suggestion
        node-version: [20.x, 22.x, 24.x] # Should match the test job's matrix
```

.github/workflows/e2e.yml
```diff
@@ -14,14 +14,15 @@ jobs:
     strategy:
       matrix:
         sandbox: [sandbox:none, sandbox:docker]
+        node-version: [18.x, 20.x, 22.x, 24.x]
```
mattKorwel:
```suggestion
        node-version: [20.x, 22.x, 24.x]
```

mattKorwel:
Several related issues here to collate. 

#1226 #1164 #2930

.github/workflows/e2e.yml
```diff
@@ -14,14 +14,15 @@ jobs:
     strategy:
       matrix:
         sandbox: [sandbox:none, sandbox:docker]
+        node-version: [20.x, 22.x, 24.x]
```
mattKorwel, line 17: This is our longest job. Like Taylor mentioned else where i'm nervous this might slow things down, despite the parallelism of actions. I say lets try it, but i think I'm inclined to only do the matrix on the quicke CI tests. Ship it for now, lets be prepared to cut back if its impactful.

mattKorwel (dismissed):

NTaylorMullen:
@swissspidy going to rely on you to monitor and react to long build times if needed after this one

swissspidy:
Sure, no problem.

FYI: this PR needs manual handling because the `Build and Lint (20.x)` status check is currently required, but this PR renames it to just `Lint`, so the previous one will always be in waiting state.

To address this, need to temporarily make it not required, merge the PR, and then make the new one required.

NTaylorMullen (approved):

PR #2709, opened 2025-06-30
Fix oauth credential caching.

scidomino:
## TLDR

Previously, we only cached credentials on web login, but we should also cache them any time they are updated.

## Dive Deeper

This won't have any user visible changes, but it will save time refreshing expired tokens on startup.

## Reviewer Test Plan

Log in with oauth and verify that the credentials get cached appropriately.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | x  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

This is a cherrypick of [this commit](https://github.com/google-gemini/gemini-cli/commit/78baebccfa477f1435d721466f4974e468fd52d7) to the old main branch.

### DIFF
--- a/packages/core/src/code_assist/oauth2.test.ts
+++ b/packages/core/src/code_assist/oauth2.test.ts
@@ -64,6 +64,7 @@ describe('oauth2', () => {
       setCredentials: mockSetCredentials,
       getAccessToken: mockGetAccessToken,
       credentials: mockTokens,
+      on: vi.fn(),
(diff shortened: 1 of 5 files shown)


scottdensmore (approved):
Approved

PR #2724, opened 2025-06-30
fix(cli): Require tag for /chat save command

### LINKED ISSUE #2189: Bug: /chat save fails silently when no tag is provided
### What happened?

When using the `/chat save` command without providing a tag, the command does not return an error. It prints a confirmation message "Conversation checkpoint saved.", which suggests the operation was successful.
```shell
> /chat save


ℹ Conversation checkpoint saved.
 

> /chat list


ℹ list of saved conversations: 
```
However, the chat is not saved with a usable tag, making it impossible to resume later.
This silent failure leads to a poor user experience as the user is led to believe the session has been saved correctly.

When `/chat save` is run without a tag, the CLI should display an error message informing the user that a tag is required. For example: `Error: A tag is required. Usage: /chat save <your-tag-name>`.

### Client information

<details>

```console
$ gemini /about

CLI Version            0.1.5
Git Commit             bf873a1
Model                  gemini-2.5-pro
Sandbox                no sandbox
OS                     darwin
```

</details>

### Login information

Google Account.

Instead of performing the save operation with an empty tag, the command should ideally:
1. Validate the `tag` argument: Check if a tag is provided and is not empty.
2. Prevent save operation: If no valid tag is given, the save operation should not proceed.
3. Display an explicit error: Inform the user that a tag is required for /chat save.

flowernotfound:
## TLDR

スクリーンショット 2025-07-01 2 38 37

This pull request addresses a user experience issue with the `/chat save` command by making the `tag` argument mandatory. Previously, saving without a tag would create an undiscoverable checkpoint. Now, the command will prompt the user for a tag if one is not provided, ensuring all saved sessions are easily accessible.

## Dive Deeper

The core issue was that a user could run `/chat save` without specifying a name (a `tag`). This action would succeed, but the resulting saved conversation would have an empty tag. Because it lacked a name, it would not appear in the output of `/chat list`, effectively making the saved session lost to the user.

This change introduces a simple validation at the beginning of the `save` action. It checks for the presence of a tag. If the tag is missing, it prevents the save operation and returns an informative error message guiding the user on the correct usage: `/chat save <tag>`. This ensures that every saved checkpoint is intentionally named and can be easily listed and resumed, directly improving the command's usability and reliability.

## Reviewer Test Plan

1.  Build and run the CLI (`npm run build` then `node scripts/start.js`).
2.  Start a conversation by sending any message (e.g., "hello").
3.  Run the command `/chat save` without any arguments.
    -   **Verify:** An error message is displayed: `Missing tag\nUsage: /chat save <tag>`.
    -   **Verify:** No new checkpoint is saved (you can check the `.gemini` directory if you wish).
4.  Run the command `/chat save my-test-session`.
    -   **Verify:** A success message is displayed: `Conversation checkpoint saved with tag: my-test-session.`
5.  Run `/chat list`.
    -   **Verify:** The tag `my-test-session` appears in the list of saved conversations.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

-   Fixes #2189


sethtroisi:
Thanks for the PR, I'm working on the same thing in #2904 which requires the tag for both save and resume.

PR #2747, opened 2025-06-30
Use structured prompt for compression.

scidomino:
## TLDR

Provides detailed instructions for how to compress history specifying an xml structure.

## Reviewer Test Plan

Ask gemini to do some work, `/compress` then `/chat save`. Go to the saved chat json and verify that the output is as expected.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | x  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

b/427863603

### DIFF
--- a/packages/core/src/core/client.ts
+++ b/packages/core/src/core/client.ts
@@ -22,7 +22,7 @@ import {
   ChatCompressionInfo,
 } from './turn.js';
 import { Config } from '../config/config.js';
-import { getCoreSystemPrompt } from './prompts.js';
+import { getCoreSystemPrompt, getCompressionPrompt } from './prompts.js';
 import { ReadManyFilesTool } from '../tools/read-many-files.js';
 import { getResponseText } from '../utils/generateContentResponseUtilities.js';
 import { checkNextSpeaker } from '../utils/nextSpeakerChecker.js';
@@ -171,7 +171,7 @@ export class GeminiClient {
     const toolRegistry = await this.config.getToolRegistry();
     const toolDeclarations = toolRegistry.getFunctionDeclarations();
     const tools: Tool[] = [{ functionDeclarations: toolDeclarations }];
-    const initialHistory: Content[] = [
+    const history: Content[] = [
       {
         role: 'user',
         parts: envParts,
@@ -180,8 +180,8 @@ export class GeminiClient {
         role: 'model',
         parts: [{ text: 'Got it. Thanks for the context!' }],
       },
+      ...(extraHistory ?? []),
     ];
-    const history = initialHistory.concat(extraHistory ?? []);
     try {
       const userMemory = this.config.getUserMemory();
       const systemInstruction = getCoreSystemPrompt(userMemory);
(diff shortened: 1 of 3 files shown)


packages/core/src/core/prompts.ts
```diff
@@ -271,3 +271,81 @@ Your core function is efficient and safe assistance. Balance extreme conciseness
...
 }
+
+/**
+ * Provides the system prompt for the state snapshotting process.
+ * This prompt instructs the model to act as a specialized state manager,
+ * think in a scratchpad, and produce a structured XML summary.
+ */
+export function getCompressionPrompt(): string {
+  return `
+You are the internal monologue and state management component of a software engineering AI agent. Your primary function is to perform "state snapshotting".
```
anj-s: Can we reword this first line to make it more simple:
"You are the component that summarizes internal chat history into a given structure"

packages/core/src/core/prompts.ts
```diff
@@ -271,3 +271,81 @@ Your core function is efficient and safe assistance. Balance extreme conciseness
...
+
+When the conversation history grows too large, you will be invoked to distill the entire history into a concise, structured XML snapshot. This snapshot is CRITICAL, as it will become the agent's *only* memory of the past. The agent will resume its work based solely on this snapshot. All crucial details, plans, errors, and user directives MUST be preserved.
```
anj-s, line 284: "This snapshot is CRITICAL, as it will become the agent's *only* memory of the past" -> This will change with the recency bias PR. Can we remove or reword so that we don't mention what the summarization covers?

packages/core/src/core/prompts.ts
```diff
@@ -271,3 +271,81 @@ Your core function is efficient and safe assistance. Balance extreme conciseness
...
+The structure MUST be as follows:
+
+<state_snapshot>
+    <overall_goal>
+        <!-- A single, concise sentence describing the user's high-level objective. -->
+        <!-- Example: "Refactor the authentication service to use a new JWT library." -->
+    </overall_goal>
+
+    <key_knowledge>
+        <!-- Crucial facts, conventions, and constraints the agent must remember. Use bullet points. -->
```
anj-s:
```suggestion
        <!-- Crucial facts, conventions, and constraints the agent must remember based on the conversation history and interaction with the user. Use bullet points. -->
```

packages/core/src/core/prompts.ts
```diff
@@ -271,3 +271,81 @@ Your core function is efficient and safe assistance. Balance extreme conciseness
...
+        <!-- Example:
+         - Build Command: \`npm run build\`
+         - Testing: Tests are run with \`npm test\`. Test files must end in \`.test.ts\`.
+         - API Endpoint: The primary API endpoint is \`https://api.example.com/v2\`.
+         - Style: Project uses Prettier and follows a functional programming style.
```
anj-s:
```suggestion
         
```

packages/core/src/core/prompts.ts
scidomino: Good idea. But let's leave it as is for now and change it when we actually implement the recency bias PR.

packages/core/src/core/prompts.ts
scidomino: Suggestion applied.

packages/core/src/core/prompts.ts
```diff
@@ -271,3 +271,81 @@ Your core function is efficient and safe assistance. Balance extreme conciseness
...
+        -->
+    </key_knowledge>
+
+    <file_system_state>
```
anj-s, line 308: We should add that the agent will need to read any file it needs back into conversation history.

packages/core/src/core/prompts.ts
```diff
@@ -271,3 +271,81 @@ Your core function is efficient and safe assistance. Balance extreme conciseness
...
+        <!-- Example:
+         1. [DONE] Identify all files using the deprecated 'UserAPI'.
+         2. [IN PROGRESS] Refactor \`src/components/UserProfile.tsx\` to use the new 'ProfileAPI'.
+         3. [TODO] Refactor the remaining files.
+         4. [TODO] Update tests to reflect the API change.
+        -->
+    </current_plan>
+
+    <next_step>
+        <!-- The single, immediate next action the agent should take. Be specific. -->
```
anj-s: I don't know if we need this because technically the agent should figure this out given the current state and plan

packages/core/src/core/prompts.ts
```diff
@@ -271,3 +271,81 @@ Your core function is efficient and safe assistance. Balance extreme conciseness
...
+    </next_step>
+
+    <blockers_and_questions>
+        <!-- Any unresolved issues, errors, or questions for the user. -->
+        <!-- Example:
+         - BLOCKER: \`npm install\` failed due to a peer dependency conflict. Awaiting user guidance.
+         - QUESTION: Waiting for the user to confirm the correct database credentials for staging.
+        -->
+    </blockers_and_questions>
+</state_snapshot>
```
anj-s: lets us call this compressed_chat_history for the sake of readability

packages/core/src/core/prompts.ts
scidomino: suggestion applied

packages/core/src/core/prompts.ts
```diff
@@ -271,3 +271,81 @@ Your core function is efficient and safe assistance. Balance extreme conciseness
...
+        -->
+    </current_plan>
+
+    <next_step>
+        <!-- The single, immediate next action the agent should take. Be specific. -->
+        <!-- Example: "Continue refactoring the \`render\` method in \`src/components/UserProfile.tsx\` on line 45." -->
+    </next_step>
+
+    <blockers_and_questions>
+        <!-- Any unresolved issues, errors, or questions for the user. -->
```
anj-s: I would remove this since at this point we are either in the middle of a request or executing a specific compress command

packages/core/src/core/prompts.ts
scidomino: Taking this out.

packages/core/src/core/prompts.ts
scidomino: Removed.

packages/core/src/core/prompts.ts
scidomino: Suggestion applied.

packages/core/src/core/prompts.ts
scidomino: I played around with it but I couldn't think of a good way to phrase that because it's kind of already implied by the structure.

packages/core/src/core/client.ts
```diff
@@ -428,74 +428,61 @@ export class GeminiClient {
...
     }
 
-    const summarizationRequestMessage = {
-      text: 'Summarize our conversation up to this point. The summary should be a concise yet comprehensive overview of all key topics, questions, answers, and important details discussed. This summary will replace the current chat history to conserve tokens, so it must capture everything essential to understand the context and continue our conversation effectively as if no information was lost.',
-    };
-    const response = await this.getChat().sendMessage({
-      message: summarizationRequestMessage,
+    const { text: summary } = await this.getChat().sendMessage({
+      message: {
+        text: 'First, reason in your scratchpad. Then, generate the <state_snapshot>.',
```
anj-s, line 455: We need to update <state_snapshot> here.

packages/core/src/core/geminiChat.ts
```diff
@@ -72,10 +72,6 @@ function isValidContent(content: Content): boolean {
  * @throws Error if the history contains an invalid role.
  */
 function validateHistory(history: Content[]) {
-  // Empty history is valid.
```
anj-s, line 75: do you know why we moved this?

packages/core/src/core/prompts.ts
```diff
@@ -271,3 +271,68 @@ Your core function is efficient and safe assistance. Balance extreme conciseness
 
   return `${basePrompt}${memorySuffix}`;
 }
+
+/**
+ * Provides the system prompt for the state snapshotting process.
```
anj-s:
```suggestion
 * Provides the system prompt for the history compression process.
```

packages/core/src/core/geminiChat.ts
scidomino: Yes. I did this. It's a useless 4 lines of code. If history is empty, the loop is a no-op anyways.

packages/core/src/core/prompts.ts
scidomino: Applied.

allenhutchison (approved):

PR #2759, opened 2025-06-30
fix(workflows): use preview release gemini-cli in triage workflows

jerop:
this allows us to test out the command settings in triage workflows 

we don't want to build from main, we want to rely on the gemini cli action

### DIFF
--- a/.github/workflows/gemini-automated-issue-triage.yml
+++ b/.github/workflows/gemini-automated-issue-triage.yml
@@ -28,11 +28,8 @@ jobs:
         with:
           token: ${{ steps.generate_token.outputs.token }}
 
-      - name: Install dependencies
-        run: npm install
-
       - name: Run Gemini Issue Triage
-        uses: google-gemini/gemini-cli-action@1efc0bac9e0b2da6c6cab95df513324d8dfc2a79
+        uses: google-gemini/gemini-cli-action@111dadaecabd309baba60f56f2b520c52c0f9a47
         env:
           GITHUB_TOKEN: ${{ steps.generate_token.outputs.token }}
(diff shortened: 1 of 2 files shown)


.github/workflows/gemini-automated-issue-triage.yml
```diff
@@ -28,19 +28,16 @@ jobs:
...
-        uses: google-gemini/gemini-cli-action@1efc0bac9e0b2da6c6cab95df513324d8dfc2a79
+        uses: google-gemini/gemini-cli-action@111dadaecabd309baba60f56f2b520c52c0f9a47
         env:
           GITHUB_TOKEN: ${{ steps.generate_token.outputs.token }}
           ISSUE_TITLE: ${{ github.event.issue.title }}
           ISSUE_BODY: ${{ github.event.issue.body }}
           ISSUE_NUMBER: ${{ github.event.issue.number }}
           REPOSITORY: ${{ github.repository }}
         with:
-          version: main
```
KeijiBranshi, line 43: the release (rc and official) are generally just cut from the `main` branch. so i would expect the changes you need to also be on `main`. just curious why we're switching this away from `main`?

KeijiBranshi (approved):
Approving to unblock. But left a question on the `version` change.

.github/workflows/gemini-automated-issue-triage.yml
jerop: It's because we are installing google-gemini/gemini-cli in [google-gemini/gemini-cli-action](https://github.com/google-gemini/gemini-cli-action), and it doesn't work well when building from main (it'd want us to install and build, instead of using the action

allenhutchison (approved):

PR #2780, opened 2025-07-01
feat: add fine-grained access control for shell commands

heartyguy:
# Updated PR Description

## Summary

This PR implements fine-grained access control for shell commands through `allowCommands` and `denyCommands` configuration. Commands can be pre-approved or explicitly denied using flexible pattern matching.

### Core Features

1. **allowCommands** - Pre-approve safe commands to run without confirmation
2. **denyCommands** - Always require confirmation (takes precedence over allow)
3. **Pattern matching**:
   - Single-word patterns match command root: `"git"` → matches `git push`
   - Multi-word patterns match as prefix: `"git push"` → matches `git push --force`
   - Glob patterns: `"git*"`, `"*.js"`, `"test?"`
   - Regex patterns: `"/^make\\s+\\w+$/"`
4. **Session learning**: `/pushcmdz` saves session-approved commands

### Configuration

```json
{
  "allowCommands": ["ls", "pwd", "git status", "npm test", "git*"],
  "denyCommands": ["rm -rf", "sudo*", "git push --force"]
}
```

Or via CLI:
```bash
gemini --allow-commands 'ls,pwd,git*' --deny-commands 'rm -rf,sudo*'
```

## Testing the Feature

```bash
# 1. View the test configuration
cat .gemini/settings.json

# 2. Test allowed commands (no confirmation)
ls -la                    # ✅ Single-word pattern
git status --verbose      # ✅ Multi-word prefix match
node --version           # ✅ Glob pattern (node*)
make build              # ✅ Regex pattern

# 3. Test denied commands (always confirm)
rm -rf /tmp             # ⚠️ Denied (prefix match)
git push --force        # ⚠️ Denied (overrides git*)
sudo apt update         # ⚠️ Denied (glob match)

# 4. Test session learning
yarn install            # Requires confirmation
# Choose "Yes (always)"
yarn install            # Now runs without confirmation
/pushcmdz              # Save to project settings
```

## Changes Made

- Added `allowCommands` and `denyCommands` to Settings interface and Config class
- Enhanced ShellTool with pattern matching and precedence logic
- Added `/pushcmdz` and `/pushdeny` slash commands
- Implemented ReDoS protection using safe-regex library
- Fixed command matching logic based on review feedback
- Comprehensive tests covering all pattern types
- Consolidated documentation in `docs/shell-command-control.md`

## Review Notes

1. All critical issues from code review have been addressed:
   - Multi-word patterns now work correctly with prefix matching
   - ReDoS protection properly implemented
   
2. Tests are passing, linter/formatter applied

3. Feature is opt-in with no breaking changes

Fixes #2417


heartyguy:
# Response to PR Review Comments

Thank you @gemini-code-assist for the thorough review! I've addressed both critical issues:

## 1. Fixed: Command Matching Logic (Critical)

You're absolutely right - the previous implementation only matched against command root, breaking multi-word patterns. I've fixed this by:

- Added `getCommandToMatch()` method that determines whether to match against full command or command root based on pattern type
- Patterns with spaces now match against the full command (as prefix match)
- Regex patterns match against the full command
- Single-word patterns continue to match against command root for backward compatibility

Example behavior now:
```javascript
// Pattern: "git push"
"git push --force" ✅ matches (prefix match)
"git pull" ❌ doesn't match

// Pattern: "rm -rf"  
"rm -rf /" ✅ matches
"rm file.txt" ❌ doesn't match
```

## 2. Fixed: ReDoS Protection (High)

Implemented proper ReDoS protection using the `safe-regex` library:

- All regex patterns are validated with `safe-regex` before execution
- Unsafe patterns are rejected with a warning
- Applied to both user-provided regex patterns and converted glob patterns
- Removed the misleading timeout comment

The implementation now properly protects against malicious regex patterns that could cause denial of service.

## Additional Updates

- Updated documentation to clarify pattern matching behavior
- Added comprehensive tests for multi-word pattern matching
- All tests passing, linter and formatter applied to modified files only

The feature now works correctly as documented, with proper security safeguards in place.

heartyguy:
Testing Screenshots:

Screenshot 2025-07-01 at 4 17 23 PM


Screenshot 2025-07-01 at 4 11 47 PM

Screenshot 2025-07-01 at 4 11 57 PM

Screenshot 2025-07-01 at 4 14 07 PM

palladius:
Thanks for the great progress, and for fixing 2 bugs.

Couple of thoughts: 
* if `rm -rf` is in the denyList, it should prompt user or just say NO YOU CANT? It looks like you added `confirmCommand` in the code but not in the discussion. as Confirm, its perfect this way :) 
* how does this interact with yolo modo? I suppose it basically makes allow/confirmCommand irrelevant (always yes) but preserves the denyCommand, right?

docs/command-access-control.md
```diff
@@ -0,0 +1,366 @@
...
+**Project Settings** (`.gemini/settings.json`):
+```json
+{
+  "allowCommands": [
+    "ls",
+    "pwd",
+    "git status",
+    "npm test"
+  ],
+  "denyCommands": [
```
palladius: I think we need to add documentation and example for `confirmCommands`

heartyguy: added.

heartyguy:
> Thanks for the great progress, and for fixing 2 bugs.
> 
> Couple of thoughts:
> 
> * if `rm -rf` is in the denyList, it should prompt user or just say NO YOU CANT? It looks like you added `confirmCommand` in the code but not in the discussion. as Confirm, its perfect this way :)
> * how does this interact with yolo modo? I suppose it basically makes allow/confirmCommand irrelevant (always yes) but preserves the denyCommand, right?
Thank you for the review @palladius! 🙏

## 1. Yes, it prompts (not blocks)!

You're spot on! Commands in the list prompt for confirmation, they don't block. That's exactly why I renamed it to `confirmCommands` - much clearer than `denyCommands`.

When `rm -rf` is in `confirmCommands`:
```
Command requires confirmation. Allow execution of 'rm -rf'? (y/N)
```
User can type `y` to proceed or `n` to cancel. Perfect UX as you said! ✅

## 2. YOLO Mode Interaction

Almost right, but with a twist:

- **Normal mode**: 
  - ✅ Checks `allowCommands` 
  - ✅ Checks `confirmCommands`
  
- **YOLO mode**: 
  - ✅ Still checks `allowCommands` (safety net remains!)
  - ❌ Skips `confirmCommands` (no prompts)

Example:
```json
{
  "allowCommands": ["git*", "npm*"],
  "confirmCommands": ["git push --force"]
}
```

In YOLO mode:
- `git push --force` → Runs without confirmation ✅
- `rm -rf` → Still blocked (not in allowCommands) ❌

So YOLO is "more permissive but not reckless" - perfect for the YOLT (You Only Live Twice) philosophy!

## 3. Documentation

Already updated everywhere! The latest commit has `confirmCommands` throughout docs and code. 

Thanks for catching the naming inconsistency - your feedback made this feature much better! 🚀

## Future Enhancement: True Deny

If we need commands that are truly blocked (no way to run them), we could add `denyCommands` in a future PR using this infrastructure. For now, `confirmCommands` covers the main use case - protecting against accidental dangerous commands while still allowing intentional use.

heartyguy:
[palladius](https://github.com/palladius) I am ready for rereview!

palladius:
@heartyguy there are some conflicts in `shell.ts`, can you please resolve them?
Then I'll download your patch and try it out locally.

Thanks again!

timothyfroehlich:
It's great to see this implemented! My 2 cents, the slash commands are a bit odd. I'd rather select "always approve" while approving or denying a command ( in the moment)

packages/cli/src/config/config.integration.test.ts
```diff
@@ -240,4 +240,320 @@ describe('Configuration Integration Tests', () => {
...
+        sandbox: false,
+        targetDir: tempDir,
+        debugMode: false,
+        sessionId: 'test-session',
+        model: 'test-model',
+        allowCommands: [
+          'ls', // exact match
+          'git*', // glob pattern
+          '*.sh', // glob pattern
+          'test?', // glob pattern
```
aspiers, line 309: Missing tests for `foo[0-9]` and `foo{bar,baz}` style of globbing.

packages/core/src/tools/shell.ts
```diff
@@ -98,6 +103,82 @@ Process Group PGID: Process group started or \`(none)\``,
...
+        console.warn(`Invalid regex pattern: ${pattern}`);
+        return false;
+      }
+    }
+
+    // Check for glob-like patterns (contains * or ?)
+    if (pattern.includes('*') || pattern.includes('?')) {
+      // Convert glob to regex
+      const regexPattern = pattern
+        .replace(/[.+^${}()|[\]\\]/g, '\\$&') // Escape regex special chars except * and ?
```
aspiers, line 133: This is not correct; `foo[0-9]` and `foo{bar,baz}` are both valid glob patterns as previously mentioned in https://github.com/google-gemini/gemini-cli/issues/2417#issuecomment-3019600412

packages/core/src/tools/shell.ts
```diff
@@ -98,6 +103,82 @@ Process Group PGID: Process group started or \`(none)\``,
...
+        .replace(/\*/g, '.*') // * matches any characters
+        .replace(/\?/g, '.'); // ? matches single character
```
aspiers, line 135: [The original plan was to use `minimatch`](https://github.com/google-gemini/gemini-cli/issues/2417#issuecomment-3032901481); curious why this is no longer being followed?  There may be good reasons (e.g. `*` behaves differently to `**` which shows that globbing is intended for path matching), but I would definitely question not supporting `[...]` character classes at least as this is very standard syntax in any globbing context.

packages/core/src/tools/shell.ts
```diff
@@ -98,6 +103,82 @@ Process Group PGID: Process group started or \`(none)\``,
...
+      try {
+        const regexStr = pattern.slice(1, -1);
+
+        // Check if regex is safe from ReDoS attacks
+        if (!safeRegex(regexStr)) {
+          console.warn(
+            `Potentially unsafe regex pattern detected (ReDoS risk): ${pattern}`,
+          );
+          return false;
+        }
```
aspiers, line 118: IMHO this concern from the Gemini review bot is a bit overkill as the config comes from the user so they are not going to intentionally DoS themselves, and the chance of them DoS-ing themselves by accident is tremendously slim.  So it's certainly a low severity concern not high severity [as claimed](https://github.com/google-gemini/gemini-cli/pull/2780#discussion_r2176217615).  However it doesn't do any harm so it can be kept.

aspiers (changes requested):

packages/core/src/tools/shell.ts
```diff
@@ -98,6 +102,49 @@ Process Group PGID: Process group started or \`(none)\``,
...
+      try {
+        const regex = new RegExp(pattern.slice(1, -1));
+        // Simple ReDoS protection: timeout after 100ms
+        const startTime = Date.now();
+        const result = regex.test(command);
+        const elapsed = Date.now() - startTime;
+        
+        // Warn if regex takes too long (potential ReDoS)
+        if (elapsed > 50) {
+          console.warn(`Slow regex pattern detected (${elapsed}ms): ${pattern}`);
```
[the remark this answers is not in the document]
aspiers: This part of the review is inaccurate IMHO, and shows the limitations of automated code review by a bot which doesn't have full context around the security model:

- The config comes from the user, so they are not going to intentionally DoS themselves.
- The chance of them DoS-ing themselves by accident is tremendously slim.
- A "malicious user with the ability to edit the .gemini/settings.json file" would also have the ability to do far worse damage than just editing that file.

So it's certainly a low severity concern not high severity as claimed.  Nevertheless, having protection against slow regexen doesn't do any harm and does cover the tiny corner case of accidentally poor performing patterns.

packages/cli/src/config/config.integration.test.ts
palladius: +1

packages/core/src/tools/shell.ts
palladius: +1 on this. @heartyguy  can you please address this?

palladius:
> It's great to see this implemented! My 2 cents, the slash commands are a bit odd. I'd rather select "always approve" while approving or denying a command ( in the moment)

Can you give an example? I'm not sure I follow you

aspiers:
Think I agree about the / commands, they're a bit odd or at least strangely named. Why "/pushcmdz"? Even "/pushcmds" isn't entirely intuitive.

aspiers:
`/savecmds` makes more sense to me.

palladius:
## Context
So I was thinking of push/pop behaviour. But in fact, it looks like only one direction make sense (from local memory to JSON settings, since the other direction is implicit), and `cmdz` is like a Google slang thing (add a Z to the end to avoid duplication).

However, I agree `/savecmds` makes more sense. Should we have a plethora of different CMD actions (/savecmds, /loadcmds, ...) I'd be keen on reverting the words to /cmdsave /cmd... to leverage TAB/autocompletion. But it doesnt seem the case now.

timothyfroehlich:
> > It's great to see this implemented! My 2 cents, the slash commands are a bit odd. I'd rather select "always approve" while approving or denying a command ( in the moment)
> 
> Can you give an example? I'm not sure I follow you

Sorry for the delayed response. I suppose I'm just asking that the tool use approval box have an "always approve" option that saves it locally, to that project's .gemini file, so if I start a new session I don't need to wade through approving it to run all the stuff I know is safe but isn't covered by the auto-approve setting.

jacob314:
Thank you for trying to tackle this hard and important issue.
However due to all the merge conflicts and over a month of inactivity I'm closing. If you are still interested in working on this please open a new issue. If possible I would suggest trying to break down the task into a few smaller prs so they can be reused more easily and you don't end up having to deal with the painful merge conflicts you would have handle to land this pr.

PR #2793, opened 2025-07-01
feat(core): Add infinite loop protection to client

allenhutchison:
## TLDR

The sendMessageStream method had a vulnerability where its infinite loop protection could be bypassed by providing an excessively large turns value. This change ensures the client remains stable by strictly capping the number of recursive turns.

## Dive Deeper

The sendMessageStream() method uses recursion to allow the model to continue speaking when checkNextSpeaker() returns 'model'. To prevent infinite loops, it has a turns parameter that decrements with each recursive call, stopping at 0. However, if the method was called with a very large turns value (e.g., Number.MAX_SAFE_INTEGER), the protection would be ineffective, leading to uncontrolled recursion.

The Fix
A bound check has been added at the beginning of sendMessageStream():

const boundedTurns = Math.min(turns, this.MAX_TURNS);

This ensures that regardless of the input value, the method will never recurse more than MAX_TURNS (100) times.

## Reviewer Test Plan
A new test case was added to confirm the fix. The test simulates the vulnerability by:

Mocking checkNextSpeaker() to always return 'model', triggering continuous recursion.

Calling sendMessageStream() with Number.MAX_SAFE_INTEGER for the turns parameter.

Verifying that with the fix, the loop correctly stops at the MAX_TURNS limit instead of running indefinitely.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | X  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

#1484 and probably others


scottdensmore:
@allenhutchison Do we have telemetry around these turns? It would be good to collect this if not.

allenhutchison:
@scottdensmore  Not that I know of, but I agree we should get the instrumentation in here to get better data.

scottdensmore (approved):
I believe we should get this out in the next release so approving. @allenhutchison can you look at maybe adding some telemetry or putting in an issue as a fast follow.

owenofbrien:
@allenhutchison @scottdensmore - I was curious how this fix worked, so I took a look at the code, but I don't see how it fixes any active looping bugs. It looks like the only non-test, non-recursive caller of sendMessageStream doesn't pass the optional turns? param, so the recursion always starts with MAX_TURNS.

This change seems to protect us against future changes where someone passes an excessively high turns value to sendMessageStream, but afaict it doesn't fix any current bug. Please let me know if I'm missing anything here!

PR #2794, opened 2025-07-01
Fix spurious logs about invalid MaxSizedBox children due to Ink6 + React19 migration

jacob314:
## TLDR

We had debug logging that warned if Box children of a MaxSizedBox had concerning flexDirection that was panicking as the Default value (row) was no longer showing up as Box children with the default value were now showing as undefined.

This is all completely fine. We just need to stop warning about this case.

To repro the previous behavior where every MaxSizedBox was spewing debug logs, run:
`npm run start -- -d`


scottdensmore (approved):
Approved

PR #2864, opened 2025-07-01
Glob to fd

brychanrobot:
## TLDR

This pull request significantly improves the performance and reliability of file completions,
especially in large codebases. The primary motivation is to provide a faster and more responsive
user experience when using the @ file completion feature.

This is a prototype. It works, but it probably needs some refinement before it can be landed.

## Dive Deeper

 * Faster File Search: The underlying file search mechanism has been migrated from a Node.js-based
   glob implementation to a native, high-performance fd binary. This results in a dramatic speed
   improvement for file and directory lookups.
 * Completion Interruption: The completion logic now cancels previous, outdated search requests when
   a new one is initiated. This prevents stale results from appearing and ensures that the
   suggestions are always relevant to the user's latest input.
 * Fuzzy Search: The file search now employs a fuzzy matching strategy when an exact match is not
   found. This makes it easier to find files even if the exact path is not known.
 * Dependency Management: The fd binary is now downloaded as part of the postinstall script,
   ensuring it is always available without being bundled directly in the package.

## Reviewer Test Plan

Clone the chromium repo and try some @ completions

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | 🪟  | 🐧  |
| npx      | ❓  | 🪟  | 🐧  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

https://github.com/google-gemini/gemini-cli/issues/2861


packages/cli/src/ui/components/InputPrompt.tsx
```diff
@@ -133,14 +133,8 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
...
         if (atIndex === -1) return;
-        const pathPart = query.substring(atIndex + 1);
-        const lastSlashIndexInPath = pathPart.lastIndexOf('/');
-        let autoCompleteStartIndex = atIndex + 1;
-        if (lastSlashIndexInPath !== -1) {
-          autoCompleteStartIndex += lastSlashIndexInPath + 1;
-        }
         buffer.replaceRangeByOffset(
-          autoCompleteStartIndex,
+          atIndex + 1,
```
jacob314, line 137: was this change intentional?

chrstnb:
Going to close this because of the merge conflicts and other work in this area!

PR #2881, opened 2025-07-01
Improve Function Call argument validation and typing

### LINKED ISSUE #1331: Fix DiscoveredMCPTool typing
DiscoveredMCPTool stores parameterSchema as Record<string, unknown> but then casts it to a Schema. We should store it as a Schema from the beggining. There are also some typing errors in setting parameterSchema we need to clean up like in mcp-clients where we cast Schema parameters to `FunctionDeclaration` which makes absolutely no sense at all.

scidomino:
## TLDR

Uniformly using  the '@google/genai' Schema type. Also use the popular `ajv` library to validate function call arguments which covers the entire schema and provides detailed error messages.

## Dive Deeper

This change fixes the fact that we send malformed schema to Gemini. Gemini uses a Schema object which represents Types as an Enum (with UPPERCASE values) and other number values (like minItems and minLength) as strings. Instead, our tools were declaring objects that used a more standard JSON Schema format and then casting to genai Schema. As far as I can tell the only reason we weren't getting errors is that Gemini's API (in genai and CCPA) was silently converting it for us.

This also improves bad tool call error messages: Previously, most schema errors would just result in an uninformative error message "Parameters failed schema validation". Now we have detailed errors. In same cases, we were previously working around this by manually writing code to test for certain error cases. I removed those where they just duplicated what the schema validation gives us.

## Reviewer Test Plan

Have the agent make many tool calls. Eventually, some of them will mess up.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | x  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes #1331


packages/core/src/utils/schemaValidator.ts
```diff
@@ -4,55 +4,63 @@
...
-        }
-      }
+  static validate(schema: Schema | undefined, data: unknown): string | null {
+    if (!schema) {
+      return null;
     }
+    if (typeof data !== 'object' || data === null) {
+      return 'Value of params must be an object';
+    }
+    const validate = ajValidator.compile(this.toObjectSchema(schema));
```
NTaylorMullen, line 27: How expensive is this / should we be caching?

I imagine every tool call this will be invoked but the schema will always be the same.

NTaylorMullen (dismissed):
Love that we can centralize this a bit more @scidomino, very cool!

packages/core/src/utils/schemaValidator.ts
scidomino: In my tests it didn't noticeably slow things down. That said, I was thinking that I would cache this in a follow up PR. This PR is already pretty big.

NTaylorMullen (approved):

PR #2883, opened 2025-07-01
feat: Handle inline content modification in tool scheduler

adamfweidman:
## TLDR

The CoreToolScheduler can now process user-modified content (newContent) submitted alongside a tool confirmation. By delegating state management, it allows any client using gemini-core as a server to define its own modification workflow. This supports a single click "edit-and-apply" flow.

## Dive Deeper
*The Goal:*
This change enables a user to edit a proposed file modification and have the new content used immediately upon execution, without the gemini-cli client managing the state or requiring additional user approval. It empowers any client connecting to the gemini-cli agent to implement its own custom tool modification flow, providing the flexibility needed to support gemini-cli running as a local or remote server. (this change will be reusable to support shell tool changes in the future).

*The Solution:*
The handleConfirmationResponse function in packages/core/src/core/coreToolScheduler.ts has been enhanced to
handle a payload containing newContent.

 The new logic is as follows:
 1. Receive Payload: When the user confirms a tool call (e.g., ProceedOnce) and provides modified content,
    handleConfirmationResponse receives both the outcome and the payload.
 2. Update Parameters: If a payload exists, it uses the tool's ModifyContext to create a new set of tool
    parameters based on the newContent. The setArgsInternal method is called to apply these new arguments to the
    pending tool call.
 3. Refresh Cliwnr: It calculates an updated fileDiff and calls setStatusInternal to put the tool back into the
    awaiting_approval state, but with the new confirmationDetails containing the updated diff. This is the key
    step that triggers a client refresh for the user to see their changes reflected.
 4. Schedule Execution: Immediately after refreshing the UI, the function proceeds to handle the user's original
    confirmation outcome (e.g., ProceedOnce), calling setStatusInternal again to move the tool to the scheduled
    state for execution.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


packages/core/src/core/coreToolScheduler.ts
```diff
@@ -545,11 +551,62 @@ export class CoreToolScheduler {
...
+    const updatedDiff = Diff.createPatch(
+      modifyContext.getFilePath(toolCall.request.args),
+      currentContent,
+      payload.newContent,
+      'Current',
+      'Proposed',
+    );
+
+    this.setArgsInternal(toolCall.request.callId, updatedParams);
+    this.setStatusInternal(toolCall.request.callId, 'awaiting_approval', {
```
ngleo, line 605: Is this `setStatusInternal` needed?

I think it will be replaced immediately by`this.setStatusInternal(callId, 'scheduled')`.

ngleo: `fileDiff` is intended to show the frontend the latest proposed diff in the original multistep flow. I think it makes sense for your flow to not display this. 

Also, I think it's nicer for `coreToolScheduler` to be agnostic of Diffing impl.

packages/core/src/tools/tools.ts
```diff
@@ -199,7 +199,10 @@ export interface FileDiff {
 export interface ToolEditConfirmationDetails {
   type: 'edit';
   title: string;
-  onConfirm: (outcome: ToolConfirmationOutcome) => Promise<void>;
+  onConfirm: (
+    outcome: ToolConfirmationOutcome,
+    payload?: { newContent: string },
```
ngleo: Consider creating a type for `ToolConfirmationPayload` and add a comment for `newContent`.

packages/core/src/core/coreToolScheduler.ts
```diff
@@ -545,11 +551,62 @@ export class CoreToolScheduler {
...
     this.attemptExecutionOfScheduledCalls(signal);
   }
 
+  /**
+   * Applies user-provided content changes to a tool call that is awaiting confirmation.
+   * This method updates the tool's arguments and refreshes the confirmation prompt with a new diff
+   * before the tool is scheduled for execution.
+   * @private
+   */
+  private async _applyInlineEdit(
```
ngleo: nit: consider  `Modify` instead of `Edit` in naming to disambiguate from Edit tool

packages/core/src/core/coreToolScheduler.ts
```diff
@@ -545,11 +551,62 @@ export class CoreToolScheduler {
         } as ToolCallConfirmationDetails);
       }
     } else {
+      // If the user provided new content, apply it before scheduling.
```
ngleo: nit: suggest to use "client" instead of "user" to disambiguate from `ModifyWithEditor` flow, where new content is also provided by the user. Same for the comment for the private function below.

ngleo (commented):
Thanks for the change!

packages/core/src/core/coreToolScheduler.ts
adamfweidman: Done.

packages/core/src/core/coreToolScheduler.ts
adamfweidman: Done.

packages/core/src/tools/tools.ts
adamfweidman: Done.

packages/core/src/core/coreToolScheduler.ts
adamfweidman: In VSCode users can expand and look at the previous tool calls that were made so its important that we be able to update the diff. Thats the only reason the setStatusInternal is required.

ngleo (commented):
Thanks for the updates!

ngleo:
@jacob314 FYI

NTaylorMullen (approved):

PR #2943, opened 2025-07-02
Always allow should be smart about subcommands using a safety analyzer

### LINKED ISSUE #2840: Always allow should be smart about subcommands

The prompt for allowing a command should specialize based on subcommands when appropriate. 

### Why is this needed?

`git ls-files` and `git commit` are really different in whether I want to allow them automatically. Similarly whether `timeout 30 ...` is safe depends a lot on what the command is. 

### Additional context

_No response_

PranavDarshan:
## TLDR

Subcommands such as ```git reset --hard``` and other commands considered as dangerous under the command-safety-db.ts which can be updated once more commands are considered dangerous.

## Dive Deeper

Example: 

Screenshot 2025-07-02 180537

The command is blocked and not executed by gemini-cli for safety. 

image

The above command is executed as it is considered as safe and it lists the branches. 

I have created a new folder ```/packages/core/src/safety ``` where the code is added. I have also updated the ```shell.ts``` to validate the safe commands.

It executes the commands which are considered ```safe``` and the ones that need approval are marked under ```requires-approval``` and the dangerous commands blocked are marked as ```dangerous``` under the ```command-safety-db.ts``` file. This file can be updated to add more dangerous commands.

## Reviewer Test Plan

Reviewer can test on the following commands:
1. ```git reset --hard``` - dangerous - blocked for safety
2. ```git branch --list``` - safe - allowed to execute
3. ```chmod -R 755 foldername``` - dangerous - blocked for safety

Other commands under the ```command-safety-db.ts``` file can also be executed.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ✅  | ✅  |
| npx      | ✅  | ✅  | ✅  |
| Docker   | ?  | ?  | ?  |
| Podman   | - | -   | -   |
| Seatbelt | -  | -   | -   |

All the workflow actions have been successfully completed.

## Linked issues / bugs

Fixes #2840 Always allow should be smart about subcommands.

### DIFF
--- a/packages/core/src/config/config.test.ts
+++ b/packages/core/src/config/config.test.ts
@@ -422,7 +422,7 @@ describe('Server Config (config.ts)', () => {
         ...baseParams,
         usageStatisticsEnabled: true,
       });
-      await config.initialize();
+      await config.refreshAuth(AuthType.USE_GEMINI);
 
       expect(
         ClearcutLogger.prototype.logStartSessionEvent,
--- a/packages/core/src/config/config.ts
+++ b/packages/core/src/config/config.ts
@@ -459,7 +459,6 @@ export class Config {
     }
     this.promptRegistry = new PromptRegistry();
     this.toolRegistry = await this.createToolRegistry();
-    logCliConfiguration(this, new StartSessionEvent(this, this.toolRegistry));
 
     await this.geminiClient.initialize();
   }
@@ -496,6 +495,9 @@ export class Config {
 
     // Reset the session flag since we're explicitly changing auth and using default model
     this.inFallbackMode = false;
+
+    // Logging the cli configuration here as the auth related configuration params would have been loaded by this point
+    logCliConfiguration(this, new StartSessionEvent(this, this.toolRegistry));
   }
 
   getUserTier(): UserTierId | undefined {


PranavDarshan:
I am not able to complete workflow actions because of gemini api key exhausting resources.

PranavDarshan:
@allenhutchison could you please review the code.

jacob314:
Sorry this pull request never got a review. If you are still interested in landing this please let me know and I will take a look.

PranavDarshan:
@jacob314 could you please review this. most of the builds are failing because of the api limit exceeded

packages/core/src/safety/command-safety-db.ts
```diff
@@ -0,0 +1,393 @@
...
+  ping: 'safe',
+  nslookup: 'safe',
+  dig: 'safe',
+  netstat: 'safe',
+  ss: 'safe',
+
+  // Curl - context dependent
+  curl: {
+    '-X': {
+      GET: 'safe',
```
allenhutchison: We have a web-fetch tool, so I'd prefer that is used rather than curl. Event then, I think all curl operations have to require approval because of the risk of data exfiltration in the URL arguemnts (in the case of GET).

packages/core/src/safety/safety-analyzer.ts
```diff
@@ -0,0 +1,384 @@
...
+
+export type SafetyLevel = 'safe' | 'requires-approval' | 'dangerous';
+export type ExtendedSafetyLevel = SafetyLevel | 'analyze-nested-command';
+
+/**
+ * Simple shell command parser that handles basic quoting
+ * This is a minimal implementation - for production use, consider using
+ * a library like 'shell-quote' or 'shell-parse'
+ */
+function parseShellCommand(command: string): string[] {
```
allenhutchison: I think if we are going to go down this route we need to have a better library for parsing shell commands. We also need to think about all the environments we run in (mac, linux, windows)

packages/core/src/safety/command-safety-db.ts
```diff
@@ -0,0 +1,393 @@
...
+
+export type ExtendedSafetyLevel = SafetyLevel | 'analyze-nested-command';
+
+// Recursive type to handle nested command structures
+// Use an interface to handle recursive structure
+export interface CommandSafetyValue {
+  [key: string]: ExtendedSafetyLevel | CommandSafetyValue;
+}
+
+export const commandSafety: Record<
```
allenhutchison: Is this list sourced from somewhere or did you come up with it? What are your thoughts for how we would maintain this in the long run? What are your thoughts around cross platform commands (i.e. windows commands).

packages/core/src/safety/command-safety-db.ts
```diff
@@ -0,0 +1,393 @@
...
+    '--header': 'safe', // Headers (generally safe)
+    '-o': 'requires-approval', // Output to file can overwrite data
+    '--output': 'requires-approval', // Output to file can overwrite data
+    '-L': 'safe', // Follow redirects
+    '--location': 'safe', // Follow redirects
+    '*': 'safe', // Default GET behavior is safe
+  },
+
+  // Wget - context dependent
+  wget: {
```
allenhutchison: Same comments as CURL

allenhutchison (changes requested):
I think overall I have some reservations about this PR from a few different perspectives. I agree with what you are trying to achieve here, but not convinced by this implementation.

First, this seems mostly biased towards unix-like operating systems. I'm sure there are commands that we need to consider for Windows as well.

Second, it seems like managing this long list of command states is going to be very difficult over time. If there is a trusted resource that does this, please let me know.

Third, parsing shell command input correctly is a pretty nuanced task that I'm not sure we want to undertake as a first party system. Is there a reputable library that does this function over all of the OS targets that we have for Gemini CLI?

packages/core/src/safety/command-safety-db.ts
PranavDarshan: I came up with the list of commands. Most reputable tools like ShellCheck or ExplainShell focus on analysis and explanation, not on explicit safety-level categorization or runtime prevention of dangerous commands. I was not able to find any trusted resource that does this. If you find any please let me know. 

Maintaining this in the long run can done through regular updates to the database of safe commands. I do understand that it might become complicated but I was not able to find any other alternative for this.

This is currently more focused towards Linux but we can add the list of commands for other platforms.

packages/core/src/safety/command-safety-db.ts
PranavDarshan: Thank you for bringing this to my attention. I will update the safety database so that all curl operations are categorized as "requires approval," regardless of HTTP method. Please note that marking curl this way in the database simply requires an approval step for curl usage—it does not prevent the use of curl in favor of the web-fetch tool by itself. Could you please let me know what command or pattern is executed by the web-fetch tool? I can include it in the database as well for consistent handling.

packages/core/src/safety/safety-analyzer.ts
allenhutchison: We already have shell-quote as a dependency in the project, so we should look at using that if we move forward with this PR.

packages/core/src/safety/command-safety-db.ts
allenhutchison: Take a look at the work I've been doing for our upcoming policy engine in:

https://github.com/google-gemini/gemini-cli/blob/main/packages/core/src/policy/policy-engine.ts

and then the implementation of policy from settings in: #8348 

I wonder if we could offer a list like this as a policy extension or a way for users to bring in a set of extra policy that would achieve the goal you are trying to achieve here?

Issue #2963, opened 2025-07-02
Invisible text in dark mode with black terminal background


NeroMok:
### What happened?

When using the dark theme, if the terminal background is set to pure black, some text becomes
  invisible because it has the same color as the background. Please see the attached screenshot.

Environment: macOS 12

<img width="909" alt="Image" src="https://github.com/user-attachments/assets/6ec3b1ec-accd-42c4-85bd-6a8da8e756ae" />

### What did you expect to happen?

I prefer dark themes, help

### Client information

*   **CLI Version:** 0.1.9
*   **Git Commit:** 34935d6
*   **Operating System:** darwin v23.11.1
*   **Sandbox Environment:** no sandbox
*   **Model Version:** gemini-2.5-pro
*   **Memory Usage:** 206.8 MB

abcxhadi:
Are you still experiencing this error?

abcxhadi:
After reviewing the theme files, it appears that the ansiTheme in packages/cli/src/ui/themes/theme.ts is likely causing the issue. I'm working on developing a fix for it.

NeroMok:
> After reviewing the theme files, it appears that the ansiTheme in packages/cli/src/ui/themes/theme.ts is likely causing the issue. I'm working on developing a fix for it.

thx, buddy

abcxhadi:
> > After reviewing the theme files, it appears that the ansiTheme in packages/cli/src/ui/themes/theme.ts is likely causing the issue. I'm working on developing a fix for it.
> 
> thx, buddy

I just tracked down the issue, but looks like the bot beat me to the fix! 😅

gsquared94:
Found possible duplicate issues (updated):

- #3551: (0.9272841215133667)
- #4611: (0.9255520105361938)

galdawave:
Thanks for taking the time to create this issue!
It looks like this is a duplicate of issue #3551. To keep all related conversations in one place, I'm going to close this one. Please feel free to subscribe to and continue the discussion on the original issue.

PR #3003, opened 2025-07-02
fix(ui): adjust dark theme for pure black terminals

abcxhadi:
## fix(ui): Adjust dark theme for pure black terminals

**Closes:** https://github.com/google-gemini/gemini-cli/issues/2963#issuecomment-3027399182

### Problem

When using a terminal with a pure black background (`#000000`), certain text rendered by the CLI becomes invisible when the `ANSI` theme is active. This occurs because the theme also uses a pure black background, causing any black-colored text to blend in completely.

### Solution

This commit addresses the issue by subtly adjusting the background color of the `ANSI` theme from pure black to a very dark gray (`#010101`). This ensures a slight contrast, making black text visible even on pure black terminal backgrounds without noticeably changing the theme's overall appearance.

The change was applied to:
- `packages/cli/src/ui/themes/ansi.ts`
- `packages/cli/src/ui/themes/theme.ts`

### How to Test

To verify the fix, please follow these steps:

1.  Configure your terminal emulator to use a pure black (`#000000`) background color.
2.  Run the Gemini CLI.
3.  Activate the `ANSI` theme.
4.  Execute a command or trigger an output that is known to render some text in black.
5.  **Expected Result:** The black text should now be clearly visible against the new dark gray background.

#### Testing Matrix

| OS      | Terminal         | Theme | Background | Expected Result   |
| :------ | :--------------- | :---- | :--------- | :---------------- |
| macOS   | iTerm2 / Terminal | ANSI  | `#000000`  | Text is visible   |
| Windows | Windows Terminal | ANSI  | `#000000`  | Text is visible   |
| Linux   | GNOME Terminal   | ANSI  | `#000000`  | Text is visible   |

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ The Gemini API provides a free tier with [100 requests per day](https://ai.googl
 
 ### Use a Vertex AI API key:
 
-The Vertex AI provides [free tier](https://cloud.google.com/vertex-ai/generative-ai/docs/start/express-mode/overview) using express mode for Gemini 2.5 Pro, control over which model you use, and access to higher rate limits with a billing account:
+The Vertex AI API provides a [free tier](https://cloud.google.com/vertex-ai/generative-ai/docs/start/express-mode/overview) using express mode for Gemini 2.5 Pro, control over which model you use, and access to higher rate limits with a billing account:
 
 1. Generate a key from [Google Cloud](https://cloud.google.com/vertex-ai/generative-ai/docs/start/api-keys).
 2. Set it as an environment variable in your terminal. Replace `YOUR_API_KEY` with your generated key and set GOOGLE_GENAI_USE_VERTEXAI to true


NTaylorMullen:
@jacob314 could you review please.

packages/cli/src/ui/themes/theme.ts
```diff
@@ -56,20 +56,7 @@ export const darkTheme: ColorsTheme = {
...
-  AccentBlue: 'blue',
-  AccentPurple: 'magenta',
-  AccentCyan: 'cyan',
-  AccentGreen: 'green',
-  AccentYellow: 'yellow',
-  AccentRed: 'red',
-  Comment: 'gray',
-  Gray: 'gray',
-};
+
```
jacob314: nit: remove extra blank line.

jacob314:
Thanks for the pull request! Can you please attach screenshots on what you see this change doing for each of the ansi themes (there is a both light and dark ansi theme). I've tried it locally and it doesn't seem to solve the problem when I set my terminal to have an absolute black background. I suspect what you may want is to 
leave `foreground` and `background` intentionally blank in both the ansi themes. that will increase their compatibility with a range of terminals which is the main requirement for the ansi themes.

PR #77, opened 2025-04-21
feat: add custom eslint rule for cross-package imports

KeijiBranshi:
BUG=http://b/411707095
DEMO=http://gpaste/6097194865917952

Should help with the decoupling. Ensures that packages only import each other through named package imports - ie `@gemini-code/<package_name>`.

PR #3012, opened 2025-07-02
fix(eslint): remove custom rule in favor of `eslint-plugin-import`

swissspidy:
## TLDR

Leverages an existing rule from `eslint-plugin-import` to prevent relative imports, avoiding the need to maintain a custom rule. Essentially reverts #77.

## Dive Deeper

I've been using this configuration in a very similar workspaces monorepo for years and the rule works like a charm. (see https://github.com/googleforcreators/web-stories-wp/)

There shouldn't be a need to maintain a custom rule for this.

If I am missing something, please let me know and feel free to disregard this PR in that case :)

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

* #3112


mattKorwel:
@swissspidy being not an es lint expert myself, how would I validate this does no harm? Is is just pull, build and run? Could you detail what you did to valiate so future us knows?

swissspidy:
Yes sure! Similar to the [example from the original PR](https://paste.googleplex.com/6097194865917952), this will flag an incorrect relative import when making a temporary change to a file.

So if you add the following line to `packages/cli/src/gemini.tsx`:

`import { BugCommandSettings } from '../../core/src/config/config.ts';`

And run `npx eslint`, you will get an error:

```
Relative import from another package is not allowed. Use `@google/gemini-cli-core/src/config/config.ts` instead of `../../core/src/config/config.ts`  import/no-relative-packages
```

Note: the suggested alternative by the rule is technically not correct, but the rule doesn't know that it can be imported from the top. If you try it, you would get another error: `Reaching to "@google/gemini-cli-core/src/config/config.ts" is not allowed`

So in that regard the ergonomics of the custom built rule are slightly better. But in my experience that isn't really an issue, a contributor (or LLM) will quickly figure out that it must import from `'@google/gemini-cli-core'` instead.

NTaylorMullen (dismissed):

NTaylorMullen (approved):

PR #3035, opened 2025-07-02
Update notification template

eddie-santos:
## TLDR

Move UpdateNotification outside Static so it can re-render when updateMessage changes

## Reviewer Test Plan

Alter the versions in `package.json` and `packages/cli/package.json` to less than, equal, and greater than the current package version, which right now is `0.1.9`. You should only see the notification when the current version is less than the latest.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅   | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ✅   | ❓  | ❓  |

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@google/gemini-cli",
-  "version": "0.1.0",
+  "version": "0.1.9",
   "engines": {
     "node": ">=18.0.0"
   },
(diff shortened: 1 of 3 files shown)


scottdensmore (approved):
Approved

PR #3039, opened 2025-07-02
Adding TurnId to Tool call and API responses and error logs.

uttamkanodia14:
## TLDR

Adding PromptId to Tool call and API request, responses and error logs.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs
Resolves https://github.com/google-gemini/gemini-cli/issues/3445


packages/cli/src/ui/hooks/slashCommandProcessor.ts
```diff
@@ -884,7 +884,7 @@ export const useSlashCommandProcessor = (
           try {
             const compressed = await config!
               .getGeminiClient()!
-              .tryCompressChat(true);
+              .tryCompressChat('Turn Id not set', true);
```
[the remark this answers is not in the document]
uttamkanodia14: Done.

packages/cli/src/ui/hooks/useGeminiStream.ts
```diff
@@ -499,22 +503,32 @@ export const useGeminiStream = (
...
         abortSignal,
+        turn_id!,
       );
 
       if (!shouldProceed || queryToSend === null) {
         return;
       }
+      
+      if (!options?.isContinuation) {
+        console.log('################# New turn started #################');
```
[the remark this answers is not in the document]
uttamkanodia14: Done.

packages/core/src/core/geminiChat.ts
```diff
@@ -244,7 +247,10 @@ export class GeminiChat {
    */
   async sendMessage(
     params: SendMessageParameters,
+    turn_id: string,
   ): Promise<GenerateContentResponse> {
+    console.log('############################ SendMessage TurnId:', turn_id);
```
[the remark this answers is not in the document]
uttamkanodia14: Done.

packages/core/src/core/geminiChat.ts
```diff
@@ -338,7 +345,14 @@ export class GeminiChat {
    */
   async sendMessageStream(
     params: SendMessageParameters,
+    turn_id: string,
   ): Promise<AsyncGenerator<GenerateContentResponse>> {
+    
+    console.log(
+      '############################ SendMessageStream TurnId:',
+      turn_id,
+    );
```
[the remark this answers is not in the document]
uttamkanodia14: Done.

packages/cli/src/ui/hooks/useGeminiStream.ts
```diff
@@ -500,21 +508,34 @@ export const useGeminiStream = (
       const abortSignal = abortControllerRef.current.signal;
       turnCancelledRef.current = false;
 
+      if (!turn_id) {
+        turn_id = config.getSessionId() + '########' + getTurnCount();
```
owenofbrien: It would be nice to track the number of model turns per user prompt, but I'm not sure I see the value in logging the number of model turns per session.

owenofbrien:
I'm not sure I understand what we're trying to accomplish here. I thought the plan was to log a prompt_id as the next step?

Here's my mental model of the logging hierarchy (bold denotes things we don't log yet but probably should at some point):

-> User (unfortunately, we have no direct user_id, we can only approximately identify users via a combination of client_install_id & obfuscated_gaia_id)
--> Session (session_id): identifies a unique invocation of Gemini CLI from the time the user starts up the CLI to the time they close it.
---> Task (**task_id**): identifies a task that the user wants Gemini CLI to accomplish
----> Prompt (**prompt_id**): identifies a prompt that the user sends to Gemini CLI
-----> API Request (???): we might want to log the prompt_id and task_id along with each API Request so that we can count the number of API requests per task and per prompt during log analysis

Let's sync up offline.

uttamkanodia14:
> I'm not sure I understand what we're trying to accomplish here. I thought the plan was to log a prompt_id as the next step?
> 
> Here's my mental model of the logging hierarchy (bold denotes things we don't log yet but probably should at some point):
> 
> -> User (unfortunately, we have no direct user_id, we can only approximately identify users via a combination of client_install_id & obfuscated_gaia_id) --> Session (session_id): identifies a unique invocation of Gemini CLI from the time the user starts up the CLI to the time they close it. ---> Task (**task_id**): identifies a task that the user wants Gemini CLI to accomplish ----> Prompt (**prompt_id**): identifies a prompt that the user sends to Gemini CLI -----> API Request (???): we might want to log the prompt_id and task_id along with each API Request so that we can count the number of API requests per task and per prompt during log analysis
> 
> Let's sync up offline.

I have updated the PR as per the discussion

- Added promptId to api request event
- Renamed TurnId to promptId.

uttamkanodia14:
Re-opening

packages/cli/src/ui/hooks/slashCommandProcessor.ts
```diff
@@ -884,7 +884,8 @@ export const useSlashCommandProcessor = (
           try {
             const compressed = await config!
               .getGeminiClient()!
-              .tryCompressChat(true);
+              // TODO: Set Prompt id for CompressChat from SlashCommandProcessor.
```
owenofbrien, line 883: Does this mean the prompt id gets lost when the user compresses the chat?

packages/core/src/telemetry/clearcut-logger/event-metadata-key.ts
```diff
@@ -137,6 +137,9 @@ export enum EventMetadataKey {
 
   // Logs the end of a session.
   GEMINI_CLI_END_SESSION_ID = 34,
+
```
owenofbrien, line 140: Consider adding a doc heading for shared keys:

  // ==========================================================================
  // Shared Keys
  // ===========================================================================

packages/cli/src/ui/hooks/useGeminiStream.ts
```diff
@@ -500,21 +508,34 @@ export const useGeminiStream = (
       const abortSignal = abortControllerRef.current.signal;
       turnCancelledRef.current = false;
 
+      if (!prompt_id) {
+        prompt_id = config.getSessionId() + '########' + getTurnCount();
```
owenofbrien: Can you move the prompt id creation up next to the logUserPrompt() call at line 222? That will provide better logical organization. As is, I'm concerned that this will increment the prompt id too often.

owenofbrien: Can you also log the prompt id on the user prompt event?

owenofbrien (changes requested):

packages/cli/src/ui/hooks/useGeminiStream.ts
uttamkanodia14: Oh it looks like the refactoring here was not published. This line is changed as:

      if (!prompt_id) {
        prompt_id = config.getSessionId() + '########' + getPromptCount();
      }

 So the promptId generation needs user session and the Number of the the prompt in the session. It is not just the session-id, adding promptCount will make is unique in the session.

packages/cli/src/ui/hooks/slashCommandProcessor.ts
uttamkanodia14: I want to track it as a part of the different PR. Because it is a bit complicated to get the promptId there.

packages/core/src/telemetry/clearcut-logger/event-metadata-key.ts
uttamkanodia14: Done.

packages/cli/src/ui/hooks/useGeminiStream.ts
uttamkanodia14: For the first one, I looked at the changes - I think that will make the method interface of `prepareQueryForGemini` as it will also need to take the prompt_id and in case when the promptId is not present then generate and return. So it would be like taking prompt_id? and returning prompt_id. 

Can you also log the prompt id on the user prompt event? - Done.

packages/cli/src/ui/hooks/useGeminiStream.ts
```diff
@@ -500,21 +508,35 @@ export const useGeminiStream = (
...
         userMessageTimestamp,
         abortSignal,
+        prompt_id!,
       );
 
       if (!shouldProceed || queryToSend === null) {
         return;
       }
 
+      if (!options?.isContinuation) {
```
owenofbrien, line 532: Should we combine this block with the id creation block at line 511? It seems like startNewPrompt should be called before creating the prompt_id, no?

uttamkanodia14: Done.

owenofbrien (dismissed):
Thanks for tackling this!

packages/cli/src/gemini.tsx
```diff
@@ -196,10 +196,12 @@ export async function main() {
     process.exit(1);
   }
 
+  const prompt_id = Math.random().toString(16).slice(2);
```
NTaylorMullen, line 199: nit: camelCase in TS 😄 . Same elsewhere

```suggestion
  const promptId = Math.random().toString(16).slice(2);
```

packages/cli/src/ui/hooks/useGeminiStream.ts
```diff
@@ -485,7 +489,11 @@ export const useGeminiStream = (
   );
 
   const submitQuery = useCallback(
-    async (query: PartListUnion, options?: { isContinuation: boolean }) => {
+    async (
+      query: PartListUnion,
+      options?: { isContinuation: boolean },
+      prompt_id?: string,
```
NTaylorMullen, line 495: Hmmm, instead of passing prompt_id throughout everywhere, should we have an official request object that has a `promptId` as part of it? Otherwise we have a ton of extra parameters we need to weave inbetween.

NTaylorMullen (changes requested):
I'd like to see us create a first-class request object that has a prompt id on it so we don't needt o weave the promptId explicitly into every sub-message call.

packages/cli/src/ui/hooks/useGeminiStream.ts
uttamkanodia14: I agree with the Idea. But can we do that in a separate PR. This PR has been open quite some time and we need this data. I will change this in a latter PR by creating a PromptRequest object (?) and having the Id there.

NTaylorMullen (dismissed):
Unblocking for now with the expectation that there's a very quick follow with the rename/refactors

packages/cli/src/ui/hooks/useGeminiStream.ts
uttamkanodia14: Discussed offline, we need to fast follow with these changes.

NTaylorMullen (dismissed):

NTaylorMullen (approved):

Issue #2869, opened 2025-07-01
Can clobber human changes when working iteratively on the same block of code


mainroach:
### What happened?

When working collaboratively with the MonoAgent system in Gemini-CLI, we can observe an out-of-date error problem, where an edit has occurred to the file (usually manually due to a human engineer) that the Agent isn't aware of. If the NEXT action the agent takes is to edit the section of code that the Human edited, then it will clobber the human edit, and revert the block to whatever the agent has in its' history.

So, to summarize:

- Agent edits a block of code
- Human edits the same block of code
- Human asks edit to also edit that block of code further
- Agent makes modifications, but puts back the human edited code to the state the Agent last understood.

As an example : 

Here was the starter code:
```typescript
else if (event.body.collisionFilterGroup === BOUNCY_GROUP) {
        score += 15;
        scoreElement.textContent = `Score: ${score}`;
        createFloatingText('+15', new THREE.Vector3().copy(ballBody.position as any), 'green');
        kickSound.play();
        const impulse = new CANNON.Vec3(0, 20, 0);
        ballBody.applyImpulse(impulse, ballBody.position);
    }
    
```
Due to an error in the physics, I manually changed the last line to remove the relative impulse addition: 
```typescript
else if (event.body.collisionFilterGroup === BOUNCY_GROUP) {
        score += 15;
        scoreElement.textContent = `Score: ${score}`;
        createFloatingText('+15', new THREE.Vector3().copy(ballBody.position as any), 'green');
        kickSound.play();
        const impulse = new CANNON.Vec3(0, 20, 0);
        ballBody.applyImpulse(impulse); <--- HUMAN MODIFIED to return the second parameter
    }
    
```

I then provided GeminiCLI with the following prompt : 
> Ok - hrm.. it looks like the ball can get stuck on top of the bouncy block - any chance that when the ball bounces on the bouncy block, we add a little "forward" inertia to it as well, so it slowly moves off the block? 

Now - because the geminiChat agent did NOT have knowledge of the edit I made, the edit tool defaulted to its' previous historical information of what the block of text was, thus overwriting my previous change : 
```typescript
else if (event.body.collisionFilterGroup === BOUNCY_GROUP) {
        score += 15;
        scoreElement.textContent = `Score: ${score}`;
        createFloatingText('+15', new THREE.Vector3().copy(ballBody.position as any), 'green');
        kickSound.play();
[...]

PR #3043, opened 2025-07-02
fix: EditTool can clobber human edits to the same file.

mainroach:
## TLDR

This PR issues a fix for issue [2869](https://github.com/google-gemini/gemini-cli/issues/2869) , in which the EditTool can clobber writes to a file based upon an out-dated view of the file in geminiHistory (when another process edits the file since the GC has last acquired a view)

## Dive Deeper

Fix for issue [2869](https://github.com/google-gemini/gemini-cli/issues/2869) - when a given file has been edited outside of GC (by hand, or by the manual user edit tool) the GC system doesn't have visibility to it. So, the edits could be overwritten, given stale state of the MonoAgent.

This PR modifies the editCorrector in the following ways:
* Addition of `findLastEditTimestamp` function - this function will walk through the current geminiChat's history, and check for _write_ operations on a specific file (any instances of WriteFile or EditTool). If it's found in the history, it extracts the timestamp from the ID, and returns it as a number.
* In `ensureCorrectEdit` if zero occurrences are found, we call `findLastEditTimestamp` and `fs.statSync`. If the statSync return value is _more recent_ than the edit timestamp in our history, we assume that the file has been edited outside of geminiCLI, and return a failure code, rather than moving on to let the  LLM try to find the correct string. This return will force the mono-agent to re-fetch the file to update its' view.

This has also required that the _filename_ be passed into ensureCorrectEdit (although if it is empty, the check will be skipped). This parameter change required updating the API call in other functions.

When executed correctly, this has shown to address the issue of clobbering manual commits to a file. If there's an error in the calculation, the worst effect is that a false-negative occurs, and an extra file-read of the file. The LLM will then have the updated data, and can provide the proper string to the edit function next time around.

## Reviewer Test Plan
* npm start
* input the prompt 'Write a local file to your folder "temp.txt" set the value to "apple"  '
  *  This should trigger the writeFile tool, as normal.
* input the prompt 'ok, I would like you to EDIT the file (use the edit tool, immediately) and set the word "apple" to "berry"  '
  * This will trigger the edit tool, which will succeed.
* go manually edit the file - open it up and change the content. (wait 2 seconds)
* input the prompt 'ok, I would like you to EDIT the file (use the edit tool, immediately) and set the word "berry" to "cherry"  '
  * This will trigger the failure case, since the file has been edited and GC's view of it is stale.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ✅  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

[Issue 2869](https://github.com/google-gemini/gemini-cli/issues/2869)


packages/core/src/utils/editCorrector.ts
```diff
@@ -49,6 +52,68 @@ export interface CorrectedEditResult {
...
+      )
+        continue;
+
+      const outval = JSON.stringify(part.functionResponse.response['output']);
+
+      // check the string versions
+      if (!outval || outval.includes('Error') || outval.includes('Failed'))
+        continue;
+
+      if (!outval.includes(filePath)) continue;
```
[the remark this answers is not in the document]
mainroach: Due to the inconsistent nature of tool development, the return values placed in functionResponse.results.* could be any variation of a string or other object. Even between WriteFileTool and EditFileTool, how they display error or success messages vary. 
As such, this serialization is a fast path around having to hand-code checks for each one of those edge cases (which is super flakey, esp. if the implementation changes).

packages/core/src/utils/editCorrector.test.ts
```diff
@@ -491,16 +519,17 @@ describe('editCorrector', () => {
 
     describe('Scenario Group 5: Specific unescapeStringForGeminiBug checks (integrated into ensureCorrectEdit)', () => {
       it('Test 5.1: old_string needs LLM to become currentContent, new_string also needs correction', async () => {
-        const currentContent = 'const x = "a\\nbc\\\\"def\\\\"';
+        const currentContent = 'const x = "a\nbc\\"def\\"';
```
scidomino, line 522: Were these changes intentional?

packages/core/src/utils/editCorrector.ts
```diff
@@ -140,6 +206,34 @@ export async function ensureCorrectEdit(
         );
       }
     } else if (occurrences === 0) {
+      if (filePath) {
+        // In order to keep from clobbering edits made outside our system,
```
scidomino, line 244: 1. Do we want to protect against this even when occurrences !== 0?
2. I think we need to also count *reads* of the file. If GC edits a file, and then I edit it, I want GC to be able to read it and make more edits. I guess the current code allows that because it only kicks in if GC doesn't get the params right the first time. Still, it feels weird to depend on the GC screwing up for protections. 

I guess maybe we can ignore all this if we're very confident that the issues only happen when GC edit matches no occurances.

scidomino (dismissed):

packages/core/src/utils/editCorrector.ts
mainroach: Good questions - 

- (1) This existing file is a complex set of conditionals for dealing with the potential for multiple edits in the file, while also dealing with potential non-found results due to string escaping. While debugging, I found that this issue wasn't occurring in the prior conditions (occurrences===expected and occurrences>expected); the unwanted behavior was coming from the usage of the `correctOldStringMismatch` which was only called during this block of code where occurrences ===0. This triggers an LLM call which finds a "fuzzy replacement" for the target string, and attempts to identify the best expected match in the file (which subsequently might be the code the human edited)
  - I would observe that this function would benefit from a bit of a re-write to handle _each match individually_ rather than in bulk (allowing us to match for some occurrences, miss for others, and only deal with fixes/drops for the misses). But it would significantly change the current usage patterns and expectations. Without a larger bench to test those changes against, I'm hesitant to modify that level of behavior.
- (2) Yes, good call. There's a _lot_ of different ways that a file snippet can be inserted into Agent memory and cause stale reference (e.g.  `GC.Read*File/GC.Glob`->`Human edit file`->`GC.EditFile`) . Let me update the code to take these into account (for the set of items we know about).

mainroach:
I've updated the code, (per discussion with @scidomino ) to add specific reading tools to the check as well. The nuance here  is that the tools are currently inconsistent with their emission of file-names in the responses. For example, ReadFileTool does not emit the filename in the response packet, and thus the functionCall packet has to be checked for the filename instead.  We have to assume that for the immediate future, these things won't be restricted, and thus these added functions have to have some more complexity to serialize the Content.Pars[] objects, and use a blunt-force string search.  Likewise, we ignore some other tools (MCP, GLOB, ...) since they don't provide the right snippets to cause the errors listed in 2869.

**Options for future improvements:**

It's worth noting that this approach is optimized for the _smallest edited footprint_. It solves the issue while touching the least amount of code as possible. As such, it provides a downside :  It is costly for long conversations (Each message invokes a JSON.stringify). If performance is our concern, then we could instead keep a record of files referenced in the geminiChat::history. As a snippet is added to history, we update the table with the current timestamp, leaving the fetch operation to be optimal for finding the last referenced file time. 

Tracking this would require one-of-two implementations:

A) Given that the tools return response packets in inconsistent ways, we could ask each tool to be modified to mark when these references occur. This would ensure that the table doesn't track false-positives generated from error strings, and reduce the need to json.stringify the packets to search for file references. However this would require extra toil into the system, as each tool would need to be modified to track this data. That extra toil may not be worth the effort.

B) We can provide additional filtering to the functions which add messages to geminiChat history. The algorithm in this CL (which strinigify's the packet, searches for filenames, and extracts the timestamp) can be run on each message insertion, and the hash table of time stamps (from option A above) could be managed this way. This would remove the code-toil from option A, at the cost of adding a json.stringify for every packet added to the history.

If performance has not been noted to be an issue (we're running on modern hardware, after all, and even 1k Json.stringify operations are pretty fast, compared to the LLM time) then we can ignore these options and stick with the proposed implementation in this PR.

Passing back to the queue for review.

packages/core/src/utils/editCorrector.ts
```diff
@@ -49,6 +58,96 @@ export interface CorrectedEditResult {
...
+ * @returns a DateTime (as a number) of when the last edit occured, or -1 if no edit was found.
+ */
+async function findLastEditTimestamp(
+  filePath: string,
+  client: GeminiClient,
+): Promise<number> {
+  const history = (await client.getHistory()) ?? [];
+
+  // Tools that may reference the file path in their FunctionResponse `output`.
+  const toolsInResp = new Set([
```
scidomino, line 92: might as well make this a constant.

scidomino (approved):

packages/core/src/utils/editCorrector.ts
NTaylorMullen, line 92: Hmm note, this wont account for MCP servers (that's ok) so might be worth adding a comment here that we're doing our best to ensure that we capture this scenario as much as possible

NTaylorMullen (approved):
The way this is implemented currently is a bit of a hack which we'll want to revisit in the future. We may want to have a centralized file tracking service that all of our file mutating tools talks to. That way instead of trying to reverse engineer the shape of how a message looks we can rely on a more centralized stateful component. The way we could enforce its usage is to restrict all `fs` APIs in tools and require them to always go through our centralized service + have an eslint rule to further bind that expectation.

That being said, having that centralized service is something we can do as a follow up. The hacky way for now is sufficient and safe.

packages/core/src/utils/editCorrector.ts
mainroach: 👍

mainroach: Agreed - Current design was intentional about not handling MCP Tools (or any other RPC return packets) - there's too many 'unknown-unknowns' about general snippet residency to confidently tie this check into those systems. It would most likely be "the l33t answer" to build an arbitrary cache residency & matching system.. but I'm not confident we have the data to justify the complexity of that need until we observe more MCP usage.

PR #3075, opened 2025-07-03
Quota error

### LINKED ISSUE #2909: Clarify when 429 is due to rate limiting or quota

Clarify during fallback the cause.

### Why is this needed?

The behavior is to fall back from pro to flash when there is a latency resulting from too many retries when 429s are being returned. But 429 can also be returned due to quota depending on the current login method. 

This can result in the fallback message claiming slow response times, when in fact it happens instantly when quota is exhausted.  The fallback message is given instantly on first attempt when quota is exhausted.

### Additional context

_No response_

ptone:
## TLDR

Clarify in the fallback behavior when this is because of quota, and when this is because of rate limiting

## Dive Deeper

429 errors are used in both situations. We were doing a retry, and then essentially displaying a message about slow responses.

However when people have hit their daily quota this fallback will be consistent on every run of the CLI and it has led to numerous issues.

A fair amount of this change relates to plumbing through the error so that the message contents can be checked at the key fallback message generation moment.

## Reviewer Test Plan

The tests gemini helped revise seem a bit convoluted in places. but I've manually carefully reviewed the core logic and it looks correct. 
## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

resolves #2909

### DIFF
--- a/packages/cli/src/ui/App.test.tsx
+++ b/packages/cli/src/ui/App.test.tsx
@@ -14,6 +14,7 @@ import {
   ToolRegistry,
   AccessibilitySettings,
   SandboxConfig,
+  FlashFallbackHandler,
 } from '@google/gemini-cli-core';
 import { LoadedSettings, SettingsFile, Settings } from '../config/settings.js';
 import process from 'node:process';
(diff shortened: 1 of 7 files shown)


ptone:
The quota message now looks like:

Screenshot 2025-07-02 at 6 51 00 PM

PR #3175, opened 2025-07-03
feature(commands) - Refactor Slash Command + Vision For the Future

### LINKED ISSUE #2622: Feature: Nested Slash Command Completion and Execution
### Feature: Nested Slash Command Completion and Execution

The goal is to evolve our slash command system from a flat list (`/command`) to a nested tree structure (`/command subcommand ...args`), with intelligent autocompletion and execution at each level.

---

### 1. Data Structure Design (`SlashCommand` Interface)

The foundation of this feature is a recursive data structure that can represent the command tree.

**File:** `packages/cli/src/ui/hooks/slashCommandProcessor.ts`

**Proposed Interface Change:**

We will modify the `SlashCommand` interface to support nesting.

```typescript
// Proposed new interface
export interface SlashCommand {
  name: string;
  altName?: string;
  description?: string;
  
  // Action is now optional. A command with subCommands might not have a direct action.
  action?: (args?: string) =>
    | void
    | SlashCommandActionReturn
    | Promise<void | SlashCommandActionReturn>;

  // For argument completion (e.g., completing a tag for `/chat resume <tag>`)
  completion?: (partialArg: string) => Promise<string[]>;

  // The new property for nesting commands.
  subCommands?: SlashCommand[];
}
```

**Design Rationale:**

*   **`subCommands?: SlashCommand[]`**: This optional array allows us to define a command that acts as a namespace for other commands. This is the core of the nested structure.
*   **`action?: (args?: string) => ...`**: The action is now optional. A command like `/memory` might not do anything on its own but will have subcommands like `add` and `show`. Its purpose is to group related functionality. The signature is also simplified to only accept `args`, as the action is now directly tied to the specific command being executed.
*   **`completion?: (partialArg: string) => Promise<string[]>`**: This remains, but its role is now clarified: it provides completions for the *arguments* of the specific command it's attached to, not for subcommand names.

**Example Refactoring (`/memory`):**

Currently, `/memory` uses a `switch` statement in its action. We will refactor it like this:

```typescript
// In slashCommands array in useSlashCommandProcessor.ts

{
  name: 'memory',
  description: 'Manage memory. Usage: /memory <show|refresh|add> [text for add]',
  // No action, it's a parent command.
  subCommands: [
    {
      name: 'show',
      description: 'Show the current memory content.',
      action: showMemoryAction // The existing function
    },
    {
      name: 'refresh',
      description: 'Reload memory from GEMINI.md files.',
      action: performMemoryRefresh // The existing function
    },
    {
      name: 'add',
      description: 'Add a fact to the session memory.',
      action: addMemoryAction // The existing function
    }
  ]
}
```

---

### 2. Completion Logic Design (`useCompletion` hook)

This hook needs to become aware of the command tree and provide context-sensitive suggestions.

**File:** `packages/cli/src/ui/hooks/useCompletion.ts`

**Proposed Algorithm:**

1.  **Parse Input**: When the completion hook is active (query starts with `/`), split the query (e.g., `/memory re`) into parts: `['memory', 're']`.

2.  **Traverse the Command Tree**:
    *   Start with the root `slashCommands` array as the current level of commands to search.
    *   Iterate through the query parts. For each part, try to find a *complete* match in the current command level.
    *   If a match is found and it has `subCommands`, descend into that level for the next part of the query.
    *   If a part does not fully match any command, it's considered a "partial" for completion.

3.  **Provide Suggestions Based on Context**:
    *   **Scenario A: Completing a Command/Subcommand Name**
        *   This occurs when the last part of the query is a partial match.
        *   **Example**: For `/memory re`, the traversal finds `/memory`, descends, and then `re` is the partial.
        *   **Action**: Filter the subcommands of `/memory` (e.g., `show`, `refresh`, `add`) for those starting with `re`. The suggestion would be `refresh`.

    *   **Scenario B: Completing Command Arguments**
        *   This occurs when the query represents a full, valid command path and ends with a space (e.g., `/chat resume `).
        *   **Action**: Check if the final matched command object has a `completion` function. If it does, invoke it with the (currently empty) argument part to get suggestions (e.g., the list of saved chat tags).

    *   **Scenario C: No Further Completions**
        *   If the user has typed a full command that has no `completion` function and no `subCommands` (e.g., `/help `), no suggestions will be shown.

---

### 3. Execution Logic Design (`handleSlashCommand` function)

This function will be updated to traverse the command tree and execute the correct action.

**File:** `packages/cli/src/ui/hooks/slashCommandProcessor.ts`

**Proposed Algorithm:**

1.  **Parse and Traverse**: Just like the completion hook, parse the query into parts and traverse the `slashCommands` tree to find the most specific command that matches the input path.

2.  **Identify Command and Arguments**:
    *   The last fully matched command in the tree is the `commandToExecute`.
    *   Any remaining parts of the query after the matched command path are concatenated to form the `args` string.

3.  **Execute Action**:
    *   **If `commandToExecute.action` exists**: Call the function, passing the extracted `args`.
    *   **If `commandToExecute.action` does not exist BUT `commandToExecute.subCommands` does**: This means the user entered a parent command (e.g., `/memory`). We will automatically display a help message listing the available subcommands and their descriptions. This provides a great user experience for discovering functionality.
    *   **If no command is matched**: Display the standard "Unknown command" error.

---

### 4. Testing Strategy

To ensure robustness, we will update our tests to cover the new logic.

1.  **`slashCommandProcessor.test.ts`**:
    *   Add tests for the refactored nested structure of `/memory`, `/stats`, and `/chat`.
    *   Verify that calling a parent command like `/memory` correctly triggers the new help display for its subcommands.
    *   Verify that calling a subcommand like `/memory add "new fact"` correctly invokes the `addMemoryAction` with `"new fact"` as the argument.

2.  **`useCompletion.test.ts`**:
    *   Add tests for subcommand name completion (e.g., input `/memory r` should suggest `refresh`).
    *   Add tests for argument completion on subcommands (e.g., input `/chat resume ` should trigger the `completion` function and suggest saved tags).
    *   Test that completion gracefully stops when a command has no further subcommands or argument completions.

abhipatel12:
## TLDR

This PR implements a fundamental architectural refactor of the Gemini CLI's slash command system. It replaces the monolithic `useSlashCommandProcessor` with a scalable, extensible, and testable **`CommandService`** that manages a **nested command tree**.

This is the foundational first step in a multi-phase effort to enable user-defined commands and a more powerful automation platform, in a more broad, community driven vision. This PR delivers the core service, the new nested command structure, and migrates the first set of commands (`/memory`, `/help`, `/clear`) to the new system. It also includes a crucial **legacy adapter** to ensure all non-migrated commands remain fully functional during the transition.

### The Problem

The previous command system, centered in `useSlashCommandProcessor`, was a single, massive hook that was difficult to maintain, test, and extend. Commands with sub-options (e.g., `/memory show`) were handled with brittle `switch` statements, and there was no path to allow users to define their own commands—a key strategic feature.

### The Solution: A Service-Oriented, Nested Architecture

This refactor introduces a clean, modern architecture composed of three key pillars:

1.  **The Nested `SlashCommand` Interface (`types.ts`):** I've adopted a recursive data structure (inspired by @allenhutchison on #2622 !!) where a command can contain `subCommands`. This allows us to model command hierarchies naturally (e.g., `/memory` is a parent to `show`, `add`, and `refresh`).

2.  **The `CommandService` (`CommandService.ts`):** This new, central service is now the single source of truth for all commands.
    *   **Discovery:** It is responsible for loading command definitions. In this PR, it loads our built-in commands from their own modules (e.g., `memoryCommand.ts`). In the future, it will be extended to load commands from user-defined YAML files or any other format we choose to pursue!
    *   **Tree Construction:** It assembles the flat list of definitions into the unified, nested command tree.
    *   **Provision:** It provides this tree to the UI.

    This service is the central hub responsible for the entire command lifecycle, abstracting the UI from the command source.

    **Architectural Flow Goals:**

    ```
    +-------------------+      +---------------------+      +--------------------+
    |   Built-in cmds   |      | Local Command Files |      | MCP Server Prompts |
    | (TS Objects with  |      |  (.gemini.yml with  |      | (JSON definitions) |
    |  subCommands)     |      |   subCommands)      |      +--------------------+
    +-------------------+      +---------------------+               |
             |                        | (Discovers & Parses)         | (Discovers via API)
             | (Loads)                |                              |
             +------------------------+------------------------------+
                                      |
                          +---------------------------+
                          |      CommandService       |
                          | (Builds unified command   |
                          |           TREE)           |
                          +---------------------------+
                                      |
                                      | (Provides Tree)
                                      v
                        +-----------------------------+
                        |  CLI UI / useSlashProcessor |
                        | (Traverses tree to execute) |
                        +-----------------------------+
    ```

    This architecture is the key enabler for the phased rollout plan.

3.  **The `CommandContext` Object (`types.ts`):** I've established a single context object that is passed to every command's `action`. This acts as a dependency injection (DI) container, providing access to all necessary services (`config`, `logger`), UI actions (`addItem`, `clear`), and session data (`stats`) in a clean, explicit, and highly testable way. This eliminates the need for prop-drilling and makes individual command logic much easier to unit test in isolation.

### The Legacy Adapter: Ensuring a Smooth Transition

To avoid a massive, all-or-nothing PR, I've implemented a **legacy adapter pattern** within `useSlashCommandProcessor`.

- The `handleSlashCommand` function now **first** attempts to find and execute a command from the new `CommandService` tree using a tree-traversal algorithm.
- If no command is found in the new system, it **falls back** to the old `legacyCommands` array, ensuring that commands not yet migrated (like `/chat`, `/stats`, `/docs`) continue to function perfectly.
- This allows us to migrate the entire command system incrementally without any disruption to the user.

### Scope of This PR

-   [✅] Introduction of the `CommandService` and the `CommandContext` DI container.
-   [✅] Definition of the new nested `SlashCommand` interface in `types.ts`.
-   [✅] Migration of `/clear`, `/help`, and the entire `/memory` command suite to the new modular, tree-based structure.
-   [✅] Refactoring of `useSlashCommandProcessor` to consume commands from the `CommandService` and include the legacy adapter.
- [✅] Significant refactoring of the `useCompletion` hook, `InputPrompt` component, and `<Help/>` component to be fully aware of the nested command structure, providing intelligent, context-aware autocompletion and help display for sub-commands and their arguments.
- [✅] New, comprehensive unit tests for the `CommandService` and all migrated commands, leveraging a new `createMockCommandContext` test utility for easy and consistent test setup.

## Future Work (To Be Addressed in Follow-up PRs)

This PR lays the groundwork for the next phases! This are for discussion and I would love to get community feedback towards the direction we want to drive this. Some of these just encompass some ideas.

#### **High-Priority Follow-up**

-   **Complete Command Migration:** Systematically migrate all remaining commands from the `legacyCommands` array (`/chat`, `/stats`, `/docs`, `/theme`, `/auth`, etc.) into their own `[commandName]Command.ts` files using the new `SlashCommand` interface.
-   **Remove Legacy Adapter:** Once all commands are migrated, the `legacyCommands` array and the fallback logic will be removed entirely, simplifying `useSlashCommandProcessor`.

#### **Medium-Priority: Custom Commands**

-   **YAML Parser:** Enhance the `CommandService` to discover and parse user-defined `.gemini.yml` files from project and user directories.
-   **Generic Action Runner:** Implement the generic action that will execute prompts defined in YAML, handling argument substitution.

#### **Medium-Priority: MCP Prompts**

-  **MCP Prompt Discovery:** The `CommandService` will be extended to query connected MCP servers for available prompts using the `prompts/list` protocol endpoint.
-  **Seamless Adaptation:** Each discovered MCP Prompt will be automatically adapted and grafted into the command tree as a namespaced command (e.g., a `create-ticket` prompt from a `jira` server becomes the `/jira:create-ticket` command).
-  **Unified Invocation:** The `action` for these commands will transparently handle the `prompts/get` request/response cycle, making the remote command feel completely native to the user.

#### **Low-Priority: UX Polish**

-   **Command Palette:** Implement a Ctrl+K-style command palette to improve discoverability in our increasingly powerful command system.

## Reviewer Test Plan

To validate these changes, please pull down the branch and run the Gemini CLI.

1.  **Test Migrated Commands:**
    *   Run `/clear`. The screen should clear as expected.
    *   Run `/help`. The help dialog should appear, and you should see the nested structure for `/memory`.
    *   Run `/memory show`, `/memory add "test fact"`, and `/memory refresh`. All three sub-commands should execute correctly.

2.  **Test Nested Autocompletion:**
    *   Type `/mem` and press `Tab`. It should complete to `/memory `.
    *   With `/memory ` in the input, press `Tab` again. It should cycle through the sub-commands (`show`, `add`, `refresh`).
    *   Type `/memory sh` and press `Tab`. It should complete to `/memory show `.

3.  **Test Legacy Command Fallback:**
    *   Run a command that has **not** been migrated, such as `/stats` or `/docs`.
    *   Verify that these commands still work exactly as they did before. This confirms the legacy adapter is working correctly.

4.  **Review the Code:**
    *   Pay close attention to `packages/cli/src/services/CommandService.ts` and `packages/cli/src/ui/commands/types.ts` as they define the core of the new architecture.
    *   Examine `packages/cli/src/ui/hooks/slashCommandProcessor.ts` to see how the new service is consumed and how the legacy adapter is implemented.
    *   Look at `packages/cli/src/ui/commands/memoryCommand.ts` as a reference for how new, modular, nested commands are defined.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ✅  | -   | -   |

## Linked issues / bugs

- Closes #2622


packages/cli/src/services/CommandService.ts
```diff
@@ -0,0 +1,38 @@
...
+    // The constructor can be used for dependency injection in the future.
+  }
+
+  async loadCommands(): Promise<void> {
+    // For now, we only load the built-in commands.
+    // File-based and remote commands will be added later.
+    this.commandTree = await this.commandLoader();
+  }
+
+  getCommandTree(): SlashCommand[] {
```
NTaylorMullen: The fact that they're a tree is an impl detail. I'd suggest just calling this `getCommands`

packages/cli/src/ui/App.tsx
NTaylorMullen, line 1: Few comments in this file you'll need to cleanup from AIisms

packages/cli/src/ui/commands/clearCommand.ts
```diff
@@ -0,0 +1,19 @@
...
+ */
+
+import { SlashCommand } from './types.js';
+
+export const clearCommand: SlashCommand = {
+  name: 'clear',
+  description: 'clear the screen and conversation history',
+  action: async (context, _args) => {
+    context.utils.onDebugMessage('Clearing terminal and resetting chat.');
+    context.ui.clearItems();
```
NTaylorMullen:
```suggestion
    context.ui.clear();
```

And we should probably make ui.clear() do the `console.clear()` and `refreshStatic` too. Honestly I might recommend dropping `refreshStatic` entirely from the ui context piece because it's a bit of a detail

packages/cli/src/ui/commands/helpCommand.ts
```diff
@@ -0,0 +1,17 @@
...
+ */
+
+import { SlashCommand } from './types.js';
+
+export const helpCommand: SlashCommand = {
+  name: 'help',
+  altName: '?',
+  description: 'for help on gemini-cli',
+  action: (context, _args) => {
+    context.utils.onDebugMessage('Opening help.');
```
NTaylorMullen: I forget, can't we replace all the `onDebugMessage` bits with `console.debug(...)`?

allenhutchison:
This sounds great and I'm looking forward to reading through the code when I'm back at a computer. 

One idea while I was reading your description is that you might want to enable tools to define their own commands from the tool definition. You mentioned /memory in your writeup and I wonder if it would be even cleaner to allow the memory tool to define its commands in the tool definition?

packages/cli/src/ui/commands/memoryCommand.ts
```diff
@@ -0,0 +1,61 @@
...
+        );
+        await showMemory();
+      },
+    },
+    {
+      name: 'add',
+      description: 'Add content to the memory.',
+      action: (context, args): SlashCommandActionReturn | void => {
+        if (!args || args.trim() === '') {
+          context.utils.addMessage({
```
NTaylorMullen: should it be `ui.addMessage`?

packages/cli/src/ui/commands/memoryCommand.ts
```diff
@@ -0,0 +1,61 @@
...
+        }
+        // UI feedback for attempting to schedule
+        context.utils.addMessage({
+          type: MessageType.INFO,
+          content: `Attempting to save to memory: "${args.trim()}"`,
+          timestamp: new Date(),
+        });
+        // Return info for scheduling the tool call
+        return {
+          shouldScheduleTool: true,
```
NTaylorMullen: Is there a world where we return a an object that looks like a tool but we don't schedule it?

packages/cli/src/ui/commands/memoryCommand.ts
```diff
@@ -0,0 +1,61 @@
...
+          toolName: 'save_memory',
+          toolArgs: { fact: args.trim() },
+        };
+      },
+    },
+    {
+      name: 'refresh',
+      description: 'Refresh the memory from the source.',
+      action: async (context) => {
+        await context.actions.performMemoryRefresh();
```
NTaylorMullen: hmmm these `actions` seem super specific to these slash commands. Is there another way where we can provide memory services to this slash command so that it can then do this if needed?

packages/cli/src/ui/commands/types.ts
```diff
@@ -0,0 +1,87 @@
...
+export interface CommandContext {
+  // Core services and configuration
+  services: {
+    config: Config | null;
+    settings: LoadedSettings;
+    git: GitService | undefined;
+    logger: Logger;
+  };
+  // UI state and history management
+  ui: {
```
NTaylorMullen, line 23: Hmmm, a lot of the bits here in the ui are a little implementation specific (like quitting message, pending history items etc.). Are there other ways?

packages/cli/src/ui/commands/types.ts
```diff
@@ -0,0 +1,87 @@
...
+    setQuittingMessages: (messages: HistoryItem[]) => void;
+    pendingHistoryItems: HistoryItemWithoutId[];
+  };
+  // Functions to open dialogs/modals
+  dialogs: {
+    openTheme: () => void;
+    openAuth: () => void;
+    openEditor: () => void;
+    openPrivacy: () => void;
+    setShowHelp: (show: boolean) => void;
```
NTaylorMullen: This feels out of place to the others

packages/cli/src/ui/commands/types.ts
```diff
@@ -0,0 +1,87 @@
...
+  };
+  // Specific actions that interact with other hooks/state
+  actions: {
+    performMemoryRefresh: () => Promise<void>;
+    toggleCorgiMode: () => void;
```
NTaylorMullen: lol

abhipatel12:
> This sounds great and I'm looking forward to reading through the code when I'm back at a computer. 
> 
> One idea while I was reading your description is that you might want to enable tools to define their own commands from the tool definition. You mentioned /memory in your writeup and I wonder if it would be even cleaner to allow the memory tool to define its commands in the tool definition? 

Ohh I like that! I'll have to take a look when I'm back at my computer

packages/cli/src/ui/commands/types.ts
```diff
@@ -0,0 +1,87 @@
...
+  // Functions to open dialogs/modals
+  dialogs: {
+    openTheme: () => void;
+    openAuth: () => void;
+    openEditor: () => void;
+    openPrivacy: () => void;
+    setShowHelp: (show: boolean) => void;
+  };
+  // Specific actions that interact with other hooks/state
+  actions: {
```
NTaylorMullen: Hmmm, is it possible to surface actions another way? Having them here means we'll have loads of impl specific bits on the actions API as part of this context. From an API purist perspective I imagine we probably want each slash command to own its own destiny a little differently

packages/cli/src/ui/commands/types.ts
```diff
@@ -0,0 +1,87 @@
...
+    stats: SessionStatsState;
+  };
+  // Low-level utilities
+  utils: {
+    onDebugMessage: (message: string) => void;
+    addMessage: (message: Message) => void;
+  };
+}
+
+export interface SlashCommandActionReturn {
```
NTaylorMullen: We should make this a union type of regular messages or tool based messages

packages/cli/src/ui/components/InputPrompt.tsx
```diff
@@ -184,25 +202,42 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
           return;
         }
         if (key.name === 'tab') {
+          // if (completion.suggestions.length > 0) {
```
NTaylorMullen: Intentional?

packages/cli/src/ui/components/InputPrompt.tsx
```diff
@@ -184,25 +202,42 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
...
         }
-        if (key.name === 'return') {
-          if (completion.activeSuggestionIndex >= 0) {
-            handleAutocomplete(completion.activeSuggestionIndex);
-          } else if (query.trim()) {
-            handleSubmitAndClear(query);
-          }
-          return;
-        }
+        // if (key.name === 'return') {
```
NTaylorMullen: ditto

NTaylorMullen (commented):
Super exciting changes. There's so much here that's going to make our lives significantly easier with new slash command pieces coming up. Left a good amount of comments around API state. Let me know what you think!

packages/cli/src/test-utils/mockCommandContext.ts
```diff
@@ -0,0 +1,99 @@
...
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const merge = (target: any, source: any) => {
+    for (const key in source) {
+      if (source[key] instanceof Object && target && key in target) {
+        Object.assign(source[key], merge(target[key], source[key]));
+      }
+    }
+    Object.assign(target || {}, source);
+    return target;
+  };
```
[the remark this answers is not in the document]
abhipatel12: This suggestion does not work, but the bug was correctly identified. This has been fixed and all test cases pass now.

packages/cli/src/services/CommandService.ts
abhipatel12: Done

packages/cli/src/ui/App.tsx
abhipatel12: Done

packages/cli/src/ui/commands/clearCommand.ts
abhipatel12: Good point! Originally, I just moved over the exact impl using the same exact logic, but this would be a great time to go ahead and abstract these impl details from the context. 

In order to do this, I did the following:
- Create a new `clear` function that encapsulates this logic
- pass this function down to the `useSlashCommandProcessor` hook
- Update the `context.ui` to only have a single "clear" function.

packages/cli/src/ui/commands/helpCommand.ts
abhipatel12: Yes we def can! 

When I was testing this, the difference is how it appears for the user. If the CLI is started in debug mode, the `onDebugMessage` surfaces in the footer while the `console.debug(...)` appears within the console. 

I think it depends on which UX we prefer. I added a log statement and then ran `/clear`. The resulting UI is:

Screenshot 2025-07-04 at 1 19 35 PM

Do you have a UX preference here? I can make the switch pretty easily.

packages/cli/src/ui/commands/memoryCommand.ts
abhipatel12: I changed the logic to consolidate all of the "add message" to the `ui` property, so this should be resolved now!

packages/cli/src/ui/commands/memoryCommand.ts
abhipatel12: Yeah absolutely. In the long-term vision, we need to account for this. Based on your other comment, we now have a discriminated union for `SlashCommandActionReturn`. So for now we use the `ToolActionReturn` but in a future phase we can introduce a new type to the action return value.

Ex.

```typescript
export interface CommandCallActionReturn {
  type: 'command_call';
  commandName: string;
  args: Record<string, unknown>;
}

// The union would be extended like this:
export type SlashCommandActionReturn =
  | ToolActionReturn
  | MessageActionReturn
  | CommandCallActionReturn;
```

packages/cli/src/ui/commands/memoryCommand.ts
abhipatel12: actions have been removed. Each slash command should be able to control its destiny better now by using what is provided to it via the UI + services it has access to within the `CommandContext`

packages/cli/src/ui/commands/types.ts
abhipatel12: Yeah that's a great point. One possibility here is to return another Action return type.

ex.

```
export interface QuitActionReturn {
  type: 'quit';
  messages: HistoryItem[];
}
```

And then add that to the union.

In this case, the command itself would say its intention and the handler would then make the UI state changes as needed based on the messages it should display. Quit is a special case, but in general we can consolidate quite a bit.

I added the following:

```
export interface OpenDialogActionReturn {
  type: 'dialog';
  // TODO: Add 'theme' | 'auth' | 'editor' | 'privacy' as migration happens.
  dialog: 'help';
}
```

This is added to the union and returned. So the command doesn't have to directly call any sort of dialog directly, and can instead just pass its intention down to the slashCommandProcessor where the dialog/modal is opened. Left the others as TODOs since we need to add them as the migration continues.

In general, I am removing any of the extra pieces I had that accounted for other slash commands. When those slash commands are being migrated, we should do our best to keep this API clean and not bring the same mess over from the original.

packages/cli/src/ui/commands/types.ts
abhipatel12: These are gone now!

packages/cli/src/ui/commands/types.ts
abhipatel12: actions have been removed! When we eventually need corgi mode, we should do it via the existing ui or services properties.

packages/cli/src/ui/commands/types.ts
abhipatel12: Yeah, that's a great point. In this first iteration, we also don't need some of these actions (they are for some of the legacy commands that aren't being migrated in this PR). But as a proof-of-concept, I have exported the memory refresh directly in the config which allows the command to directly call it without the need of an "action". Let me know what you think of this method!

packages/cli/src/ui/commands/types.ts
abhipatel12: Good point, added two new interfaces and made it a union. Let me know what you think!

packages/cli/src/ui/components/InputPrompt.tsx
abhipatel12: ooops forgot to remove the commented out code. Had to redo the autocomplete logic in this latest batch of commits.

packages/cli/src/ui/components/InputPrompt.tsx
abhipatel12: actions have been removed! When we eventually need corgi mode, we should do it via the existing ui or services properties.

abhipatel12:
@NTaylorMullen Thanks for all the great feedback on that last round! I made some pretty significant shifts here so let me know what you think!

abhipatel12:
> This sounds great and I'm looking forward to reading through the code when I'm back at a computer.
> 
> One idea while I was reading your description is that you might want to enable tools to define their own commands from the tool definition. You mentioned /memory in your writeup and I wonder if it would be even cleaner to allow the memory tool to define its commands in the tool definition?

@allenhutchison I wanted to follow up on this idea! It enables `Tool` definitions to become the source of truth for their own user-facing slash commands. We should definitely do this since it co-locates related logic, eliminates boilerplate, and creates a single, unified discovery mechanism for commands. 

The architecture proposed in this PR supports this idea and I want to propose pursuing this in a fast-follow.

An example implementation would be as follows:

## Possible Impl Plan

### 1. Enhance the Tool Interface:
We can introduce an optional method, `getSlashCommands()`, to the core `Tool` interface. This method will allow any tool to return an array of `SlashCommand` objects that it provides.

```typescript
export interface Tool {
  // ... existing properties
  getSlashCommands?(context: CommandContext): SlashCommand[];
}
```

### 2. `CommandService` Discovery Mechanism Changes
The `CommandService` can query the `ToolRegistry` as a new source for commands. This new discovery step will happen alongside the loading of built-in and possible file-based commands.

```typescript
async loadCommands(context: CommandContext): Promise<void> {
  // Load built-in commands (e.g., /help, /clear)
  const builtInCommands = await loadBuiltInCommands();

  // Discover commands from tools
  const toolBasedCommands: SlashCommand[] = [];
  for (const tool of this.toolRegistry.getAllTools()) {
    if (tool.getSlashCommands) {
      toolBasedCommands.push(...tool.getSlashCommands(context));
    }
  }

  // 3. FUTURE: Discover commands from .gemini.yml files
  // 4. FUTURE: Discover commands from MCP `prompts/list` endpoint

  // 5. Merge all sources into a single command tree
  this.commandTree = [...builtInCommands, ...toolBasedCommands, ...];
}
```

### 3. Proof-of-Concept: `/memory`
We can start with `/memory` to prove out this concept. This PR moves `/memory` using built-in commands and this fast-follow would then directly use the `Tool` interface.

---

Let me know what you think!

packages/cli/src/services/CommandService.ts
```diff
@@ -0,0 +1,36 @@
...
+    // The constructor can be used for dependency injection in the future.
+  }
+
+  async loadCommands(): Promise<void> {
+    // For now, we only load the built-in commands.
+    // File-based and remote commands will be added later.
+    this.commandTree = await this.commandLoader();
+  }
+
+  getCommand(): SlashCommand[] {
```
NTaylorMullen:
```suggestion
  getCommands(): SlashCommand[] {
```

packages/cli/src/services/CommandService.ts
```diff
@@ -0,0 +1,36 @@
...
+import { clearCommand } from '../ui/commands/clearCommand.js';
+
+const loadBuiltInCommands = async (): Promise<SlashCommand[]> => [
+  clearCommand,
+  helpCommand,
+  memoryCommand,
+];
+
+export class CommandService {
+  private commandTree: SlashCommand[] = [];
```
NTaylorMullen:
```suggestion
  private commands: SlashCommand[] = [];
```

packages/cli/src/ui/commands/helpCommand.ts
NTaylorMullen: Hmm, I actually think the onDebugMessage is a bit misleading and not worth. I'd argue we should do console.debug only.

packages/cli/src/ui/commands/memoryCommand.ts
```diff
@@ -0,0 +1,106 @@
...
+    {
+      name: 'refresh',
+      description: 'Refresh the memory from the source.',
+      action: async (context) => {
+        context.ui.addItem(
+          {
+            type: MessageType.INFO,
+            text: 'Refreshing memory from source files...',
+          },
+          Date.now(),
```
NTaylorMullen, line 72: Do we ever not do `Date.now()`? Wondering if this can be automatic in `addItem`

packages/cli/src/ui/commands/types.ts
```diff
@@ -0,0 +1,98 @@
...
+import { Config, GitService, Logger } from '@google/gemini-cli-core';
+import { LoadedSettings } from '../../config/settings.js';
+import { UseHistoryManagerReturn } from '../hooks/useHistoryManager.js';
+import { SessionStatsState } from '../contexts/SessionContext.js';
+
+// Grouped dependencies for clarity and easier mocking
+export interface CommandContext {
+  // Core services and configuration
+  services: {
+    config: Config | null;
```
NTaylorMullen, line 17: What scenario is this `null` in?

packages/cli/src/ui/commands/types.ts
```diff
@@ -0,0 +1,98 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { Config, GitService, Logger } from '@google/gemini-cli-core';
+import { LoadedSettings } from '../../config/settings.js';
+import { UseHistoryManagerReturn } from '../hooks/useHistoryManager.js';
+import { SessionStatsState } from '../contexts/SessionContext.js';
+
+// Grouped dependencies for clarity and easier mocking
+export interface CommandContext {
```
NTaylorMullen, line 13: Love how much you cleaned this up 👏

packages/cli/src/ui/commands/types.ts
```diff
@@ -0,0 +1,98 @@
...
+  type: 'tool';
+  toolName: string;
+  toolArgs: Record<string, unknown>;
+}
+
+/**
+ * The return type for a command action that results in a simple message
+ * being displayed to the user.
+ */
+export interface MessageActionReturn {
```
NTaylorMullen, line 57: Ah is this in place of say `ui.addItem`?

packages/cli/src/ui/commands/types.ts
```diff
@@ -0,0 +1,98 @@
...
+  messageType: 'info' | 'error';
+  content: string;
+}
+
+/**
+ * The return type for a command action that needs to open a dialog.
+ */
+export interface OpenDialogActionReturn {
+  type: 'dialog';
+  // TODO: Add 'theme' | 'auth' | 'editor' | 'privacy' as migration happens.
```
NTaylorMullen, line 68: ➕

packages/cli/src/ui/commands/types.ts
```diff
@@ -0,0 +1,98 @@
...
+  | OpenDialogActionReturn;
+
+// The standardized contract for any command in the system.
+export interface SlashCommand {
+  name: string;
+  altName?: string;
+  description?: string;
+
+  // The action to run. Optional for parent commands that only group sub-commands.
+  action?: (
```
NTaylorMullen, line 84: What commands don't have an action? Wondering if this can be required / not nullable

packages/cli/src/ui/hooks/slashCommandProcessor.ts
```diff
@@ -141,41 +153,51 @@ export const useSlashCommandProcessor = (
...
+      onDebugMessage,
+    ],
   );
 
+  const commandService = useMemo(() => new CommandService(), []);
+
+  useEffect(() => {
+    const load = async () => {
+      await commandService.loadCommands();
+      setCommandTree(commandService.getCommand());
```
```suggestion
      setCommands(commandService.getCommand());
```

NTaylorMullen (dismissed):
Love the refactor and where you went with it!

NTaylorMullen:
My 2cents on exposing slash commands on tools: We could BUT I'm not sure how valuable it'd be. I could see a world in which we automatically translate some tools to slash commands maybe.

Memory might be an exception to the common case here 😄

packages/cli/src/services/CommandService.ts
abhipatel12: Done

packages/cli/src/services/CommandService.ts
abhipatel12: Done

packages/cli/src/ui/commands/memoryCommand.ts
abhipatel12: Originally we had two of these addItems, and we consolidated to a single one. `Data.now()` is being used as the `id` here. I agree that it's a bit repetitive, but I believe the best change may be to make this id field optional and have a default id that uses `Date.now()`. But i think that's best addressed in a follow-up PR if that sounds good to you!

packages/cli/src/ui/commands/types.ts
abhipatel12: I don't believe it should ever be. This was the type passed in defensively when I made the context in the event that it could be possible, but I agree it shouldn't really be possible. 

Since the solution involves modifying the app's startup logic in `gemini.tsx` to disable input until the config is loaded, it's a non-trivial architectural change if we want to handle this safely. I added a TODO and to have a fast follow to remove this extra check.

packages/cli/src/ui/commands/types.ts
abhipatel12: Thank you!

packages/cli/src/ui/commands/types.ts
abhipatel12: Sort of. For multi-step actions, the command takes action through the `context.ui.AddItem` function. So for example, in `/memory` we use the `context.ui.AddItem` in order to provide updates saying "refresh memory", etc.

But when a slash command's job is to produce a single, final message it uses `MessageActionReturn`.

I added `MessageActionReturn` for a couple of reasons.

1. It simplifies certain commands. This is especially true for error handling. We have a common error handling path currently, and making every command implement its own HistoryItem would be more work.
2. It makes the contract stronger. Now that the return is a union, the slash Command Processor knows exactly what each command is doing and it doesn't have to rely on optional checks. 
  - tool --> needs to schedule a tool call
  - message --> needs to display the basic message for the user (maybe it's an error)
  - void --> command handled everything, so don't worry about doing anything.

packages/cli/src/ui/commands/types.ts
abhipatel12: Great question!

The `action` property is optional by design to support parent commands. In our nested tree structure (`/command subcommand ...args`), a command can be a "leaf" that performs an action, or it can be a "branch" that only groups other commands.

Example:
The top-level `/memory` command only exists to contain its `subCommands` (`show`, `add`, `refresh`). It doesn't have an action of its own. If a user types just `/memory`, the system should recognize that the action is undefined and displays a help message listing the available subcommands.

If we made the `action` property required, it would force us to add empty, placeholder actions to all parent commands, which would be more boilerplate.

packages/cli/src/ui/hooks/slashCommandProcessor.ts
abhipatel12: Done (also changed `commandTree` value to --> `commands`)

abhipatel12:
> My 2cents on exposing slash commands on tools: We could BUT I'm not sure how valuable it'd be. I could see a world in which we automatically translate some tools to slash commands maybe.
> 
> Memory might be an exception to the common case here 😄

Good point, something to consider if we want go forward with that plan!

NTaylorMullen (approved):
:shipit:

mattKorwel:
👏

PR #3191, opened 2025-07-04
Add a command line option to enable and list extensions

bbiggs:
## TLDR

Adds two command line options, --list-extensions to list all installed extensions, and --extensions to provide an exclusive list of extensions to load.

## Dive Deeper

Some extensions are only useful in certain contexts. This change provides two new command line options, one --list-extensions that lists each extension and version and exits. The other, --extensions, provides a comma-separated list of extension names. All other extensions are disabled. A special name "none" requests that no extensions are loaded.

## Reviewer Test Plan

Please try with various extensions as you have them. Right now I am assuming all extensions have single-word names, another option is that we add a way for extensions to provide a 'short name'. Right now we are not validating that extensions have unique names, which may be a problem?

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes #1467.

### DIFF
--- a/packages/cli/src/config/config.ts
+++ b/packages/cli/src/config/config.ts
@@ -49,7 +49,7 @@ interface CliArgs {
   telemetryOtlpEndpoint: string | undefined;
   telemetryLogPrompts: boolean | undefined;
   'allowed-mcp-server-names': string | undefined;
-  extensions: string | undefined;
+  extensions: string[] | undefined;
   'list-extensions': boolean | undefined;
 }
 
@@ -137,9 +137,10 @@ async function parseArguments(): Promise<CliArgs> {
     })
     .option('extensions', {
       alias: 'e',
-      type: 'string',
+      type: 'array',
+      string: true,
       description:
-        'Comma-separated list of extensions to use. If not provided, all extensions are used.',
+        'A list of extensions to use. If not provided, all extensions are used.',
     })
     .option('list-extensions', {
       alias: 'l',
@@ -187,8 +188,7 @@ export async function loadCliConfig(
   const argv = await parseArguments();
   const debugMode = argv.debug || false;
 
-  const enabledExtensions =
-    argv.extensions?.split(',').map((e) => e.trim()) || [];
+  const enabledExtensions = argv.extensions?.map((e) => e.trim()) || [];
 
   let activeExtensions;
   try {
--- a/packages/cli/src/config/config.ts
+++ b/packages/cli/src/config/config.ts
@@ -303,6 +303,10 @@ export async function loadCliConfig(
     model: argv.model!,
     extensionContextFilePaths,
     listExtensions: argv['list-extensions'] || false,
+    activeExtensions: activeExtensions.map((e) => ({
+      name: e.config.name,
+      version: e.config.version,
+    })),
   });
 }
 
--- a/packages/cli/src/config/config.ts
+++ b/packages/cli/src/config/config.ts
@@ -49,7 +49,7 @@ interface CliArgs {
   telemetryOtlpEndpoint: string | undefined;
   telemetryLogPrompts: boolean | undefined;
   'allowed-mcp-server-names': string | undefined;
-  'disable-extensions': string | undefined;
+  extensions: string | undefined;
   'list-extensions': boolean | undefined;
 }
 
@@ -135,10 +135,11 @@ async function parseArguments(): Promise<CliArgs> {
       type: 'string',
       description: 'Allowed MCP server names',
     })
-    .option('disable-extensions', {
+    .option('extensions', {
       alias: 'e',
       type: 'string',
-      description: 'Comma-separated list of extensions to disable.',
+      description:
+        'Comma-separated list of extensions to use. If not provided, all extensions are used.',
     })
     .option('list-extensions', {
       alias: 'l',
@@ -186,12 +187,15 @@ export async function loadCliConfig(
   const argv = await parseArguments();
   const debugMode = argv.debug || false;
 
-  const disabledExtensions =
-    argv['disable-extensions']?.split(',').map((e) => e.trim().toLowerCase()) ||
-    [];
-  const activeExtensions = extensions.filter(
-    (e) => !disabledExtensions.includes(e.config.name.toLowerCase()),
-  );
+  const enabledExtensions =
+    argv.extensions?.split(',').map((e) => e.trim().toLowerCase()) || [];
+
+  const activeExtensions =
+    enabledExtensions.length > 0
+      ? extensions.filter((e) =>
+          enabledExtensions.includes(e.config.name.toLowerCase()),
+        )
+      : extensions;
 
   // Set the context filename in the server's memoryTool module BEFORE loading memory
   // TODO(b/343434939): This is a bit of a hack. The contextFileName should ideally be passed
--- a/packages/cli/src/config/config.ts
+++ b/packages/cli/src/config/config.ts
@@ -188,29 +188,10 @@ export async function loadCliConfig(
   const argv = await parseArguments();
   const debugMode = argv.debug || false;
 
-  const enabledExtensions = argv.extensions?.map((e) => e.trim()) || [];
-
-  let activeExtensions;
-  try {
-    activeExtensions = filterActiveExtensions(extensions, enabledExtensions);
-  } catch (e) {
-    console.error((e as Error).message);
-    process.exit(1);
-  }
-
-  if (enabledExtensions.length > 0) {
-    const activeNames = new Set(
-      activeExtensions.map((e) => e.config.name.toLowerCase()),
-    );
-    for (const extension of extensions) {
-      const status = activeNames.has(extension.config.name.toLowerCase())
-        ? 'Activated'
-        : 'Disabled';
-      console.log(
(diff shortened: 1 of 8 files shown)


scidomino:
Thanks for adding this! I think it would be more consistent with the rest of the product if we didn't do --list-extension as a flag. Instead it would make more sense as a slash command (and might not be much more difficult to implement).

It might also make sense to split that out into two PRs.

packages/cli/src/config/config.ts
```diff
@@ -171,6 +184,50 @@ export async function loadCliConfig(
...
+      const activeNames = new Set(
+        activeExtensions.map((e) => e.config.name.toLowerCase()),
+      );
+      for (const requestedExtension of lowerCaseEnabledExtensions) {
+        if (!activeNames.has(requestedExtension)) {
+          console.error(`Extension not found: ${requestedExtension}`);
+          process.exit(1);
+        }
+      }
+    }
```
scidomino: I would push this all into a helper function filterExtensions(extensions, activeExtensions). And that helper function should probably live in extension.ts (config.ts is pretty bloated as it is).

packages/cli/src/config/config.ts
```diff
@@ -171,6 +184,50 @@ export async function loadCliConfig(
...
+    ) {
+      activeExtensions.length = 0;
+    } else {
+      const activeNames = new Set(
+        activeExtensions.map((e) => e.config.name.toLowerCase()),
+      );
+      for (const requestedExtension of lowerCaseEnabledExtensions) {
+        if (!activeNames.has(requestedExtension)) {
+          console.error(`Extension not found: ${requestedExtension}`);
+          process.exit(1);
```
scidomino: What's the behavior if we throw an error here instead? Does it leave an ugly stack trace?

packages/cli/src/config/config.ts
```diff
@@ -171,6 +184,50 @@ export async function loadCliConfig(
...
+        }
+      }
+    }
+
+    const activeNames = new Set(
```
scidomino: See other comment. I'd do this as a slash command.

scidomino:
> Thanks for adding this! I think it would be more consistent with the rest of the product if we didn't do --list-extension as a flag. Instead it would make more sense as a slash command (and might not be much more difficult to implement).
> 
> It might also make sense to split that out into two PRs.

Taylor disagree with me on this so keep this as is. :D

packages/cli/src/config/config.ts
bbiggs: Done, much cleaner and I added a test.

bbiggs:
> > Thanks for adding this! I think it would be more consistent with the rest of the product if we didn't do --list-extension as a flag. Instead it would make more sense as a slash command (and might not be much more difficult to implement).
> > It might also make sense to split that out into two PRs.
> 
> Taylor disagree with me on this so keep this as is. :D

I decided to *also* add a slash command to list active extensions, which I think makes sense.

packages/cli/src/config/config.ts
```diff
@@ -48,6 +48,8 @@ interface CliArgs {
   telemetryTarget: string | undefined;
   telemetryOtlpEndpoint: string | undefined;
   telemetryLogPrompts: boolean | undefined;
+  extensions: string[] | undefined;
+  'list-extensions': boolean | undefined;
```
scidomino: telemetry-target seems to be using "telemetryTarget" without issue so maybe use listExtension instead.

packages/cli/src/config/config.ts
```diff
@@ -164,6 +178,30 @@ export async function loadCliConfig(
   const argv = await parseArguments();
   const debugMode = argv.debug || false;
 
+  const enabledExtensions = argv.extensions?.map((e) => e.trim()) || [];
+
+  let activeExtensions;
```
scidomino: is it necessary to catch this here? Does something in the call hierarchy of loadCliConfig() catch errors?

packages/cli/src/config/config.ts
```diff
@@ -164,6 +178,30 @@ export async function loadCliConfig(
   const argv = await parseArguments();
   const debugMode = argv.debug || false;
 
+  const enabledExtensions = argv.extensions?.map((e) => e.trim()) || [];
```
scidomino: Push all of this into a helper method rather than just leave it in loadCliConfig

packages/cli/src/config/extension.ts
```diff
@@ -114,3 +114,37 @@ function getContextFileNames(config: ExtensionConfig): string[] {
   }
   return config.contextFileName;
 }
+
+export function filterActiveExtensions(
+  extensions: Extension[],
+  enabledExtensions: string[],
```
scidomino: nit: "enabledExtensionNames" to distinguish names for actual extensions.

packages/cli/src/config/extension.ts
```diff
@@ -114,3 +114,37 @@ function getContextFileNames(config: ExtensionConfig): string[] {
...
+): Extension[] {
+  const lowerCaseEnabledExtensions = enabledExtensions.map((e) =>
+    e.toLowerCase(),
+  );
+  const activeExtensions =
```
scidomino: Simplify this with:

if (enabledExtensions.length === 0){
    return extensions;
}

Then you don't have to check for it everywhere.

packages/cli/src/config/extension.ts
```diff
@@ -114,3 +114,37 @@ function getContextFileNames(config: ExtensionConfig): string[] {
...
+          lowerCaseEnabledExtensions.includes(e.config.name.toLowerCase()),
+        )
+      : extensions;
+
+  if (enabledExtensions.length > 0) {
+    if (
+      lowerCaseEnabledExtensions.length === 1 &&
+      lowerCaseEnabledExtensions[0] === 'none'
+    ) {
+      activeExtensions.length = 0;
```
scidomino: "return [];" would be clearer. Then you wouldn't need to stuff the next bit of code into the else clause.

packages/cli/src/config/extension.ts
```diff
@@ -114,3 +114,37 @@ function getContextFileNames(config: ExtensionConfig): string[] {
...
+    } else {
+      const activeNames = new Set(
+        activeExtensions.map((e) => e.config.name.toLowerCase()),
+      );
+      for (const requestedExtension of lowerCaseEnabledExtensions) {
+        if (!activeNames.has(requestedExtension)) {
+          throw new Error(`Extension not found: ${requestedExtension}`);
+        }
+      }
+    }
```
scidomino, line 151: Feels like here and above we're doing this work twice. Instead, consider Making a map of lowercased extension names to extension, then build enabledExtension by iterating over enabledExtensionNames and throwing an error if there is no such extension name.

packages/cli/src/gemini.tsx
```diff
@@ -103,6 +103,14 @@ export async function main() {
   const extensions = loadExtensions(workspaceRoot);
   const config = await loadCliConfig(settings.merged, extensions, sessionId);
 
+  if (config.getListExtensions()) {
```
scidomino, line 106: My gut says we want to keep main() as slim as possible. Would it make sense to push this into loadCliConfig? Then we wouldn't even have to save listExtension in cli config.

packages/cli/src/config/config.ts
bbiggs: yep, fixed

packages/cli/src/config/extension.ts
bbiggs: fixed

packages/cli/src/config/extension.ts
bbiggs: nice, fixed

packages/cli/src/config/extension.ts
bbiggs: done

packages/cli/src/config/extension.ts
bbiggs: done

packages/cli/src/config/extension.ts
bbiggs: done

packages/cli/src/config/extension.ts
bbiggs: done

NTaylorMullen (dismissed):

packages/cli/src/gemini.tsx
bbiggs: I like having this here because it's printing and exiting (wow), and this isolates calls to process.exit(1) in gemini.tsx. What do you think?

packages/cli/src/config/config.ts
bbiggs: Cleaned this up - we just warn to the console when the extension isn't found, I think it's cleaner.

packages/cli/src/config/config.ts
bbiggs: Simplified this.

NTaylorMullen (approved):
LGTM, love this

PR #3208, opened 2025-07-04
Fix root path issue in large projects

### LINKED ISSUE #3414: RangeError in GitIgnoreParser when processing files titled "\"
### What happened?

Running npx @google/gemini-cli in a project that contains a file titled \ causes the CLI to crash with a RangeError.

This happens because the path normalization logic produces edge-case values such as /, "", or ".", which are then passed to the ignore library. That library expects valid relative paths and throws when it receives something like /.

`RangeError: path should be a path.relative()d string, but got '/'

This issue is triggered by:
	•	Files with edge-case names like \
	•	Resulting normalized paths being invalid or unexpected

What did you expect to happen?

The CLI should:
	•	Gracefully skip or handle edge-case paths like /, \, ".", or ""
	•	Handle atypical but valid project structures without breaking

### Client information

<details>

```console
│ CLI Version                 0.1.9                                                │
│ Git Commit                  34935d6                                              │
│ Model                       gemini-2.5-pro                                       │
│ OS                          darwin                                               │
│ Auth Method                 gemini-api-key
```

</details>

### Login information

API key

_No response_

spencerwalden:
## TLDR

  Fixes a RangeError crash in GitIgnoreParser when processing certain directory structures. Added comprehensive path validation to prevent invalid paths (empty, root "/", or current directory ".") from being
  passed to the ignore library.

  ## Dive Deeper

  The issue occurred when the CLI attempted to process files in certain directory structures where path normalization could result in edge case values like "/" or empty strings being passed to the underlying
  `ignore` library. The `ignore` library expects relative paths and throws a RangeError when it receives root or invalid path strings.

  The fix adds additional safety checks in the `isIgnored` method of `GitIgnoreParser` to filter out these problematic paths before they reach the ignore library.
  ## Reviewer Test Plan

  To validate this fix:

  1. **Reproduce the original issue**:
     - Add a file titled "\" inside the project you are running gemini cli on.
     - Run the CLI without input `npx @google/gemini-cli`
     - Verify it previously crashed with "RangeError: path should be a `path.relative()`d string, but got '/'"

  3. **Test the fix**:
     - Build the bundle: `npm run bundle`
     - Test in the same problematic directory
     - Verify it now shows "No input provided via stdin" instead of crashing

  4. **Test normal functionality**:
     - Verify `.gitignore` rules still work correctly in various project types
     - Test with different directory structures and file patterns
     - Ensure no regression in ignore pattern matching

  ## Testing Matrix

  |          | 🍏  | 🪟  | 🐧  |
  | npm run  | ✅  | ❓  | ❓  |
  | npx      | ❓  | ❓  | ❓  |
  | Docker   | ❓  | ❓  | ❓  |
  | Podman   | ❓  | -   | -   |
  | Seatbelt | ❓  | -   | -   |

  *Tested on macOS with npm run. Additional platform testing would be valuable.*

Fixes #3414


leehagoodjames:
@spencerwalden thank you for making this change !

With a large number of contributors, it looks like someone else already merged a fix (#3417) for this. Given that, I am closing this PR - but please re-open if it adds additional functionality that does not exist in `head`.

PR #3229, opened 2025-07-04
feat: add .svg support

PugazhendhiDev:
## TLDR

Adds .svg support in Gemini CLI file processing by treating .svg files as text, allowing Gemini to read, analyze, and edit SVG files directly.

## Dive Deeper

SVG files are XML-based text describing vector graphics, not binary data.

By adding .svg handling, Gemini can:

 -> Display SVG contents as readable text in the CLI.

 -> Allow LLM reasoning on SVG structure for generation, editing, and explanations.

 -> Avoids misclassification of SVGs as images requiring base64 handling.

This is implemented by inserting a pre-switch check for .svg extensions to read them as UTF-8 text, returning a llmContent string.

## Reviewer Test Plan

 -> Test reading various SVG files and verify the content is returned as text.

 -> Test editing SVG files through Gemini and check that changes are applied correctly.

 -> Confirm the edited SVG files open and display properly in a browser or viewer.

 -> Verify that the system can analyze SVG content without errors.

 -> Check error handling for missing or invalid SVG files.

 -> Ensure other file types still work as before.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | -  | ✅  | -  |
| npx      | -  | ✅  | -  |
| Docker   | -  | -  | -  |
| Podman   | -  | -   | -   |
| Seatbelt | -  | -   | -   |


packages/core/src/utils/fileUtils.ts
```diff
@@ -228,6 +228,21 @@ export async function processSingleFileContent(
       .relative(rootDirectory, filePath)
       .replace(/\\/g, '/');
 
+    if (path.extname(filePath).toLowerCase() === '.svg') {
```
NTaylorMullen: Lets migrate this detection logic into `detectFileType`. Similar to how we detect `image`, `text` or `pdf` we can also detect `svg` given it's a little unique. Once the detection logic is there you can then migrate all of the reactionary logic into the case statement below.

NTaylorMullen (changes requested):
Really appreciate the change here @PugazhendhiDev and would love to help get it in!

Left one comment on refactoring and could we also add a test for this use case?

packages/core/src/utils/fileUtils.ts
PugazhendhiDev: Thank you for the feedback! I have migrated the SVG detection logic into detectFileType and moved the related handling into the switch case as requested. Please let me know if there are any other changes needed.
image
image
image
image

NTaylorMullen:
@PugazhendhiDev I think you missed this bit: `could we also add a test for this use case?`

PugazhendhiDev:
> @PugazhendhiDev I think you missed this bit: `could we also add a test for this use case?`

Hi,
I added tests for SVG file. Please check. Thanks!

NTaylorMullen (dismissed):
Thank you for the contribution and working with me @PugazhendhiDev, excited to get this in!

NTaylorMullen:
Ah @PugazhendhiDev it looks like you may need to run `npm run format` and update the branch

PugazhendhiDev:
> Thank you for the contribution and working with me @PugazhendhiDev, excited to get this in!

My pleasure, happy to contribute!

PugazhendhiDev:
> Ah @PugazhendhiDev it looks like you may need to run `npm run format` and update the branch

Got it, just formatted and pushed.

NTaylorMullen (approved):

PR #3289, opened 2025-07-05
Add terminal setup command for Shift+Enter and Ctrl+Enter support

### LINKED ISSUE #849: <shift>-<enter> should do an enter but keep user control

deepankarsharma:
Implements automatic terminal configuration for multi-line input support across various terminal emulators. 

This support falls into two categories

1. Zero configuration support for terminals that support the kitty protocol (list given below). For these terminals gemini-cli will auto detect support and set things up such that Shift-Enter and Ctrl-Enter just works without any further setup required.
2. The new /terminal-setup command detects the current terminal and applies appropriate keybinding configurations for some important terminals that do not support the Kitty protocol currently (VSCode, Cursor, Windsurf)

Supported terminals:
- VS Code, Cursor, Windsurf: User will have to run /terminal-setup which will configure keybindings.json
- Kitty: via kitty protocol requires no setup on part of user
- Alacritty: via kitty protocol requires no setup on part of user
- Ghostty: via kitty protocol requires no setup on part of user
- Foot: via kitty protocol requires no setup on part of user
- iTerm2: via kitty protocol requires no setup on part of user
- rio: via kitty protocol requires no setup on part of user
- wezTerm: via kitty protocol requires no setup on part of user

The feature centralizes newline input sequences in NEWLINE_INPUT_SEQUENCES and ensures proper handling of modified Enter keys across all supported terminals.

## TLDR

Adds support for Shift-Enter to insert newline. Code is the same as #898 which was LGTM'ed but ran into issues with visibility change of gemini-cli repo from private to public. Detailed explanation available in last comment of #898. 

## Dive Deeper

1. Bulk of the new logic is in terminalSetup.ts
3. There is a small cleanup in text-buffer.ts which adds NEWLINE_INPUT_SEQUENCES

## Reviewer Test Plan

1. Verify that the change adds keybindings to the supported terminals

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | Y  | Y  | Y  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs
Resolves #849


packages/cli/src/ui/components/shared/text-buffer.ts
```diff
@@ -14,6 +14,19 @@ import stringWidth from 'string-width';
...
+export const NEWLINE_INPUT_SEQUENCES = [
+  '\n',
+  '\r\n',
+  '\\\r', // VSCode terminal represents shift + enter this way
+  '[13;2u', // shift-enter kitty (needs-setup)
+  '[13;5u', // ctrl-enter kitty (needs-setup)
+  '[27;2;13~', // shift-enter foot (no-setup)
+  '[27;5;13~', // ctrl-enter foot (no-setup)
+  '\x1b[13;2u', // shift-enter alacritty (needs-setup)
+  '\x1b[13;5u', // ctrl-enter alacritty (needs-setup)
```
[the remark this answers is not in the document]
deepankarsharma: @gemini-code-assist I have dropped alacritty support and have verified that kitty works correctly with currrent bindings as is.

packages/cli/src/ui/utils/terminalSetup.ts
```diff
@@ -0,0 +1,524 @@
...
+      ) {
+        content += '\n' + yamlBindings;
+        await fs.writeFile(configFile, content);
+        return {
+          success: true,
+          message: `Added Shift+Enter and Ctrl+Enter keybindings to Alacritty.\nModified: ${configFile}`,
+          requiresRestart: true,
+        };
+      }
+    }
```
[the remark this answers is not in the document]
deepankarsharma: @gemini-code-assist I have dropped alacritty support.

packages/cli/src/ui/utils/terminalSetup.ts
```diff
@@ -0,0 +1,524 @@
...
+    if (parentName.includes('kitty')) return 'kitty';
+    if (parentName.includes('alacritty')) return 'alacritty';
+    if (parentName.includes('ghostty')) return 'ghostty';
+    if (parentName.includes('urxvt') || parentName.includes('rxvt-unicode'))
+      return 'urxvt';
+    if (parentName.includes('foot')) return 'foot';
+  } catch (error) {
+    // Continue detection even if process check fails
+    console.debug('Parent process detection failed:', error);
+  }
```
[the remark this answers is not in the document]
deepankarsharma: @gemini-code-assist Fixed.

packages/cli/src/ui/components/shared/text-buffer.ts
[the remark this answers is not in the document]
deepankarsharma: @gemini-code-assist 

1. Good catch @ alacritty reference in text-buffer.ts. It is now removed.
2. Regarding \x1b, I had tested shift-enter and ctrl-enter on kitty without the \x1b and it worked. But since you suggested that \x1b is more compliant, I checked with adding that and kitty works with that as well, so I have made your suggested change.

packages/cli/src/ui/utils/terminalSetup.ts
```diff
@@ -0,0 +1,420 @@
...
+    if (!content.includes('URxvt.keysym.S-Return:')) {
+      content +=
+        '\n! Gemini CLI keybindings for URxvt\n' + shiftEnterLine + '\n';
+      modified = true;
+    }
+
+    if (!content.includes('URxvt.keysym.C-Return:')) {
+      content += ctrlEnterLine + '\n';
+      modified = true;
+    }
```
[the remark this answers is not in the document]
deepankarsharma: @gemini-code-assist - this is now fixed.

packages/cli/src/ui/utils/terminalSetup.ts
```diff
@@ -0,0 +1,420 @@
...
+
+    if (!content.includes('performable:shift+enter=')) {
+      content += '\n# Gemini CLI keybindings\n' + shiftEnterLine + '\n';
+      modified = true;
+    }
+
+    if (!content.includes('performable:ctrl+enter=')) {
+      content += ctrlEnterLine + '\n';
+      modified = true;
+    }
```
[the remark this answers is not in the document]
deepankarsharma: @gemini-code-assist - this is now fixed.

packages/cli/src/ui/utils/terminalSetup.ts
```diff
@@ -0,0 +1,420 @@
...
+    configDir = path.join(
+      os.homedir(),
+      'Library',
+      'Application Support',
+      'Code',
+      'User',
+    );
+  } else {
+    configDir = path.join(os.homedir(), '.config', 'Code', 'User');
+  }
```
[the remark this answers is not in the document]
deepankarsharma: @gemini-code-assist - this is now fixed

packages/cli/src/ui/utils/terminalSetup.ts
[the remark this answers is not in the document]
deepankarsharma: Commit 369e5cc55f39b0d5bf791777bdb9211775f66162 has the fix.

kovidgoyal:
Doing this will *break* the usage of those keys in other terminal programs. You cant remap keys globally for your individual program by modifying conf files for the terminal emulator. If you want to support shift+enter and ctrl+enter correctly, add support for the kitty keyboard protocol, which supports those key combinations in all of the terminals above except possibly urxvt.

deepankarsharma:
Thanks for the pointer @kovidgoyal! I will go through https://sw.kovidgoyal.net/kitty/keyboard-protocol and see if the approach works with the terminals listed above. I will look at urxvt separately and see how that should be handled.

packages/cli/src/ui/components/shared/text-buffer.ts
```diff
@@ -14,6 +14,17 @@ import stringWidth from 'string-width';
...
 
+// Constants for newline input sequences
+export const NEWLINE_INPUT_SEQUENCES = [
+  '\n',
+  '\r\n',
+  '\\\r', // VSCode terminal represents shift + enter this way
+  '\x1b[13;2u', // shift-enter kitty (needs-setup)
+  '\x1b[13;5u', // ctrl-enter kitty (needs-setup)
+  '[27;2;13~', // shift-enter foot (no-setup)
+  '[27;5;13~', // ctrl-enter foot (no-setup)
```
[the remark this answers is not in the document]
deepankarsharma: @gemini-code-assist - have dropped entries for foot and kitty now since we will now handle these via the kitty protocol code path.

packages/cli/src/ui/utils/terminalSetup.ts
```diff
@@ -0,0 +1,484 @@
...
+      return configureCursor();
+    case 'windsurf':
+      return configureWindsurf();
+    default:
+      return {
+        success: false,
+        message: `Terminal "${terminal}" is not supported yet.`,
+      };
+  }
+}
```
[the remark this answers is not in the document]
deepankarsharma: @gemini-code-assist - I have updated the PR description and documentation to reflect this.

packages/cli/src/ui/utils/terminalSetup.ts
```diff
@@ -0,0 +1,484 @@
...
+  }
+  // Check VS Code last since forks may also set VSCODE env vars
+  if (termProgram === 'vscode' || process.env.VSCODE_GIT_IPC_HANDLE) {
+    return 'vscode';
+  }
+
+  // Check parent process name
+  if (os.platform() !== 'win32') {
+    try {
+      const { stdout } = await execAsync('ps -o comm= -p $PPID');
```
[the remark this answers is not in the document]
deepankarsharma: @gemini-code-assist This is a draft PR. I dont have access to a windows machine at the moment but will test code on windows before opening the PR up for review or merging.

packages/cli/src/ui/hooks/useEnhancedKeypress.ts
```diff
@@ -0,0 +1,200 @@
...
+      }
+
+      // Handle other keys as needed
+      return null;
+    };
+
+    const handleKeypress = (_: unknown, key: Key) => {
+      // Always pass through Ctrl+C immediately, regardless of protocol state
+      // Check both standard format and Kitty protocol sequence
+      if ((key.ctrl && key.name === 'c') || key.sequence === '\x1b[99;5u') {
```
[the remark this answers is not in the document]
deepankarsharma: @gemini-code-assist - since both these are in an or condition, order should not matter.

JeelRajodiya:
Hello @deepankarsharma, are you still working on the issue? I would like to give it a try if you are not working on it anymore!

deepankarsharma:
@JeelRajodiya - Code for this is almost ready to go. I am traveling this week so dont have access to a Windows machine to test the change on windows. I should be able to get this ready for review shape by early next week.

JeelRajodiya:
Sure! No worries. Thanks for the update!

deepankarsharma:
Updates
1. Was able to test on Windows and confirmed that things work
2. For VS Code / Cursor / Windsurf - decided to make the keybindings be the same as certain other agentic cli tool so that users dont have to change their bindings back and forth. Both shift-enter and ctrl-enter will be configured emit \\\r\n and InputPrompt.tsx is changed to detect this and handle it correctly

deepankarsharma:
@jacob314 - this is ready to review

docs/cli/commands.md
```diff
@@ -70,6 +70,29 @@ Slash commands provide meta-level control over the CLI itself.
...
+    - **Alacritty**
+    - **Ghostty**
+    - **Foot**
+    - **iTerm2**
+    - **Kitty**
+    - **Rio**
+    - **WezTerm**
+  - **Notes:**
+    - Most terminals require a restart after configuration
+    - VS Code requires restarting the entire application, not just the terminal
```
jacob314: on Mac it seems like keybindings work out of the box in VSCode even without running the terminal-setup. I suspect some of your other fixes make it work by default. I verified and my `keybindings.json` was empty. Without your change I do not get proper shift-enter behavior on mac.

packages/cli/src/ui/utils/terminalSetup.ts
```diff
@@ -0,0 +1,339 @@
...
+
+type SupportedTerminal = 'vscode' | 'cursor' | 'windsurf';
+
+// Terminal detection
+async function detectTerminal(): Promise<SupportedTerminal | null> {
+  const termProgram = process.env.TERM_PROGRAM;
+
+  // Check VS Code and its forks - check forks first to avoid false positives
+  // Check for Cursor-specific indicators
+  if (
```
jacob314, line 59: can we use $TERM_PROGRAM instead? On my Mac vscode is reliably setting it to `vscode`. I would worry VSCODE_GIT_ASKPASS_MAIN might not work if the git extension is not installed.

packages/cli/src/ui/utils/terminalSetup.ts
```diff
@@ -0,0 +1,339 @@
...
+  if (process.env.VSCODE_GIT_ASKPASS_MAIN?.toLowerCase().includes('windsurf')) {
+    return 'windsurf';
+  }
+  // Check VS Code last since forks may also set VSCODE env vars
+  if (termProgram === 'vscode' || process.env.VSCODE_GIT_IPC_HANDLE) {
+    return 'vscode';
+  }
+
+  // Check parent process name
+  if (os.platform() !== 'win32') {
```
jacob314, line 75: wonder if this whole block can be deleted if we rely on TERM_PROGRAM.

packages/cli/src/ui/utils/kittyProtocolDetector.ts
```diff
@@ -0,0 +1,105 @@
...
+    setTimeout(() => {
+      if (!checkFinished) {
+        process.stdin.removeListener('data', handleData);
+        if (!originalRawMode) {
+          process.stdin.setRawMode(false);
+        }
+        detectionComplete = true;
+        resolve(false);
+      }
+    }, 50);
```
jacob314, line 88: can you comment where the 50 ms threshold comes from? I would worry we might miss that terminals support kitty when the users machine happens to be under heavy load but I also don't love adding >50ms of delay for all users who happen to not be using a kitty protocol terminal.

packages/cli/src/ui/utils/terminalSetup.ts
```diff
@@ -0,0 +1,339 @@
...
+ * if (result.success) {
+ *   console.log(result.message);
+ *   if (result.requiresRestart) {
+ *     console.log('Please restart your terminal');
+ *   }
+ * }
+ */
+export async function terminalSetup(): Promise<TerminalSetupResult> {
+  // Check if terminal already has optimal keyboard support
+  if (isKittyProtocolEnabled()) {
```
jacob314, line 309: I get this message in iterm but not vscode even though vscode does appear to work fine out of the box for me. not sure why.

packages/cli/src/ui/hooks/useKittyKeyboardProtocol.ts
```diff
@@ -0,0 +1,31 @@
...
+  enabled: boolean;
+  checking: boolean;
+}
+
+/**
+ * Hook that returns the cached Kitty keyboard protocol status.
+ * Detection is done once at app startup to avoid repeated queries.
+ */
+export function useKittyKeyboardProtocol(): KittyProtocolStatus {
+  const [status] = useState<KittyProtocolStatus>({
```
jacob314, line 24: what is the purpose of using this as a hook instead of just calling isKittyProtocolSupported and isKittyProtocolEnabled directly from callers that care?

packages/cli/src/ui/App.tsx
```diff
@@ -479,7 +479,10 @@ const App = ({ config, settings, startupWarnings = [], version }: AppProps) => {
       if (Object.keys(mcpServers || {}).length > 0) {
         handleSlashCommand(newValue ? '/mcp desc' : '/mcp nodesc');
       }
-    } else if (key.ctrl && (input === 'c' || input === 'C')) {
+    } else if (
+      (key.ctrl && (input === 'c' || input === 'C')) ||
+      input === '[99;5u'
```
jacob314: perhaps we should grab this and other kitty(?) related constants and add them to a file of constants along with brief documentation of what each of them means

packages/cli/src/ui/components/InputPrompt.tsx
```diff
@@ -285,6 +289,35 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
         return;
       }
 
+      // Check if this is the special backslash+return sequence
+      const isPrevBackslash =
```
jacob314: can we handle this up a level rather than input prompt? this seems simpler to handle as part of processing keyboard events that directly in InputPrompt.  that could avoid the currently logic with prevKeyRef ideally surfacing this just as shift+enter as far as inputPrompt is concerned.

packages/cli/src/ui/components/shared/text-buffer.ts
```diff
@@ -1167,42 +1176,55 @@ export function useTextBuffer({
 
       if (
         key.name === 'return' ||
-        input === '\r' ||
-        input === '\n' ||
-        input === '\\\r' // VSCode terminal represents shift + enter this way
-      )
+        (input && NEWLINE_INPUT_SEQUENCES.includes(input))
```
jacob314: please add a test for this.

packages/cli/src/ui/hooks/useEnhancedKeypress.ts
```diff
@@ -0,0 +1,200 @@
...
+ *
+ * Key codes for Enter variants:
+ * - 13: Enter/Return key
+ *
+ * Modifier flags:
+ * - 2: Shift
+ * - 5: Ctrl
+ * - 9: Alt
+ */
+export function useEnhancedKeypress(
```
jacob314: if it makes things easier please consider merging this in with useKeypress.ts 
I haven't thought deeply but some of the logic like 'paste-start' and 'past-end' seems a bit similar.

jacob314: in either case see useKeypress.test.ts as reasonably tests to ask Gemini CLI to use as inspiration when writing tests for this.

jacob314 (commented):
This is amazing work! Can't wait to get this to users. Happy to discuss any of these review comments over chat as well if that is easier.

docs/cli/commands.md
deepankarsharma: Great! - think we are hitching a ride on the \\\r\n train probably.

packages/cli/src/ui/utils/terminalSetup.ts
deepankarsharma: I tried $TERM_PROGRAM initially when I was just handling VS Code, issue with it is that Windsurf and Cursor also report TERM_PROGRAM as vscode.

packages/cli/src/ui/utils/terminalSetup.ts
deepankarsharma: I explained above why TERM_PROGRAM will make it difficult to handle vscode forks. One other thought I have here is that outside the scope of this PR I take a stab at https://github.com/xtermjs/xterm.js/issues/4198 which is the blocker for VS Code getting Kitty Protocol support (and with it we will be able to unify behind a single way of doing keyboard shortcut handling).

packages/cli/src/ui/utils/kittyProtocolDetector.ts
deepankarsharma: It was an arbitrary "fast enough" threshold I came up with, I can reduce the threshold. Alternatively I can do this check for non vscode/windsurf/cursor cases only so that those users are not affected.

packages/cli/src/ui/utils/terminalSetup.ts
deepankarsharma: I will investigate Mac behavior and update this ticket on why VS Code is working on Mac without any key bindings being configured.

packages/cli/src/ui/hooks/useKittyKeyboardProtocol.ts
deepankarsharma: Am doing it at the start and caching results because I didnt want to introduce any hard to investigate rendering artifacts at runtime once the gui has started rendering due to the terminal sequence we emit to detect - its unclear to me how all terminals that dont support kitty keyboard protocol will behave in such a scenario.

packages/cli/src/ui/App.tsx
deepankarsharma: Great idea - I will start a constants file.

packages/cli/src/ui/hooks/useEnhancedKeypress.ts
deepankarsharma: I will look to unify the two.

packages/cli/src/ui/components/InputPrompt.tsx
deepankarsharma: Any pointers to a more idiomatic place in which to place this?

packages/cli/src/ui/App.tsx
deepankarsharma: Done

packages/cli/src/ui/components/InputPrompt.tsx
deepankarsharma: Do you think doing this in the keypress hook would be better?

packages/cli/src/ui/hooks/useEnhancedKeypress.ts
deepankarsharma: Done

packages/cli/src/ui/components/InputPrompt.tsx
deepankarsharma: Ive moved it to keypress hook.

docs/cli/commands.md
deepankarsharma: I am unable to replicate this on my mac machine. I need to do terminal-setup for shift-enter to work even with latest changes.

packages/cli/src/ui/utils/terminalSetup.ts
deepankarsharma: I am unable to replicate this on my mac machine. I need to do terminal-setup for shift-enter to work even with latest changes.

deepankarsharma:
@jacob314 - this should be ready for review again

packages/cli/src/ui/components/InputPrompt.tsx
```diff
@@ -419,13 +389,11 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
...
         if (buffer.text.trim()) {
-          const [row, col] = buffer.cursor;
-          const line = buffer.lines[row];
-          const charBefore = col > 0 ? cpSlice(line, col - 1, col) : '';
-          if (charBefore === '\\') {
-            buffer.backspace();
-            buffer.newline();
-          } else if (NEWLINE_INPUT_SEQUENCES.includes(key.sequence)) {
+          if (
+            (NEWLINE_INPUT_SEQUENCES as readonly string[]).includes(
```
jacob314: nit: is the `as readonly string[]` needed?

deepankarsharma: If I remove that i get this complaint from type checker - > 

```text
node ../../scripts/build_package.js

src/ui/components/InputPrompt.tsx:394:15 - error TS2345: Argument of type 'string' is not assignable to parameter of type '"\n" | "\r\n" | "\\\r" | "\u001B[13;2u" | "\u001B[13;5u"'.

394               key.sequence,
                  ~~~~~~~~~~~~
```

packages/cli/src/ui/hooks/useKeypress.test.ts
```diff
@@ -258,4 +268,325 @@ describe('useKeypress', () => {
...
+          meta: false,
+          shift: false,
+        });
+      });
+
+      // Backslash should be buffered
+      expect(onKeypress).not.toHaveBeenCalled();
+
+      // Send 'a' after backslash
+      act(() => {
```
jacob314: can you also add a test where there is a backslash but not a follow up character. Seems like this ends up still working after an imperceptible timeout but would be good to verify that case is handled robustly as would be confusing to users otherwise

packages/cli/src/ui/hooks/useKeypress.ts
```diff
@@ -64,8 +70,141 @@ export function useKeypress(
...
+          // Try to parse the buffer as a Kitty sequence
+          const kittyKey = parseKittySequence(kittySequenceBuffer);
+          if (kittyKey) {
+            kittySequenceBuffer = '';
+            onKeypressRef.current(kittyKey);
+            return;
+          }
+
+          // If buffer doesn't match expected pattern and is getting long, flush it
+          if (kittySequenceBuffer.length > 10) {
```
jacob314: comment on why the magic number 10. Is there some alternate magic number we can use aligned with the longest kitty sequence or something?

packages/cli/src/ui/hooks/useKeypress.ts
```diff
@@ -64,8 +70,141 @@ export function useKeypress(
 
     let isPaste = false;
     let pasteBuffer = Buffer.alloc(0);
+    let kittySequenceBuffer = '';
+    let previousSequence = '';
+
+    // Parse Kitty protocol sequences
+    const parseKittySequence = (sequence: string): Key | null => {
+      // Match CSI <number> ; <modifiers> u or ~
+      // eslint-disable-next-line no-control-regex
```
jacob314: nit: can you cleanup this regex to not have to disable this lint

jacob314 (commented):
This is looking really good now. Only thing blocking me approving is there appears to be a React issue likely due to calling setState within useEffect. If you can't reproduce let me know. I'm able to reproduce immediately within iterm2 on my Mac.

Image

packages/cli/src/ui/hooks/useKeypress.test.ts
deepankarsharma: Done

packages/cli/src/ui/hooks/useKeypress.ts
deepankarsharma: Done.

packages/cli/src/ui/hooks/useKeypress.ts
deepankarsharma: Done

deepankarsharma:
@jacob314 - unfortunately cant replicate the iterm issue. I am on iterm2 Build 3.5.14

siba2893:
Can’t wait for this feature!

jacob314:
> @jacob314 - unfortunately cant replicate the iterm issue. I am on iterm2 Build 3.5.14

Ok let me see if I can figure out an easy fix for it as I can reproduce reliably.

jacob314 (commented):
As would be expected for a change this significant there are a number of regressions. 
Here are the ones I've caught so far manually testing this:
1. ctrl-o no longer works on iterm2 (it should open the debug console). Likely something about your code is breaking the default `useInput` provided by Ink. When I use a terminal that doesn't support Kitty then ctrl-O continues to work with your changes

I'd suggest we resolve by switching fully over to the new improved input support that you and @bbiggs have provided. We might need to refactor useKeypress.ts a little to make it handle multiple active useKeypress hooks present in the same app more efficiently.

2. When in iterm2 the terminal hangs for a few seconds after pressing the left or right arrow due us not yet handling the kitty sequences for these keys correctly. Seems like kitty is just using the sequence 27,91,68 for the left arrow
with option+left arrow showing as 27,91,49,59,51,68.
To avoid other issues like this I'd suggest we start failing harder on mismatched kitty sequences logging an error if we ever exceed the kitty terminal buffer length.

deepankarsharma:
1. Good catch @ Ctrl-O. I will triage and fix this. 
2. I will look at the changes @bbiggs has made and will talk with him if I have questions
3. Will add error logs and will also see if I can write more exhaustive tests to rule such things out.

jacob314:
> 1. Good catch @ Ctrl-O. I will triage and fix this.
> 2. I will look at the changes @bbiggs has made and will talk with him if I have questions
> 3. Will add error logs and will also see if I can write more exhaustive tests to rule such things out.

We should probably add some logging to analytics to catch how often we are hitting the 11 character limit where we punt a character sequence if it doesn't match a known kitty sequence. that will give us confidence when we launch that we aren't causing too many users to experience these odd hangs.

deepankarsharma:
@jacob314 - will add the logging. Will also write a development time test (ie wont check it in as an automated test) that exhaustively walks through the entire set of unicode end points and makes sure the state transition is valid.

deepankarsharma:
Have addressed the two comments, however upstream/main has moved quite a bit in the same files. Will catch up and understand whats changed and evolve this changeset.

packages/cli/src/ui/components/InputPrompt.tsx
```diff
@@ -421,13 +423,17 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
...
       }
-      // Ctrl+C (Clear input)
+      // Ctrl+C (Clear input or exit if empty)
       if (key.ctrl && key.name === 'c') {
         if (buffer.text.length > 0) {
           buffer.setText('');
           resetCompletionState();
           return;
         }
+        // If buffer is empty, pass through to App for exit handling
```
jacob314: I'd really prefer if we didn't solve this by tightly coupling InputPrompt and App.tsx in this unnatural way. instead we should make the way App is handling keys use our new approach rather than the previous standard ink keyboard handling.

deepankarsharma: Understood - will study the new approach and use it as a baseline.

jacob314: let me know if you need help. We can take on refactoring to fully adopt the new input handling as a separate pr and then you can use it to land this one.

deepankarsharma: @jacob314 - have updated the code and tried to make it idiomatic over the current upstream/main. Please take a look.

aalexand:
I assume for this to work in iTerm2, the "Report keys using CSI u" setting needs to be checked?

image

Is it checked by default? If not, then we'd need to tell users to "go to dialog X > Y > Z, and check the A box there" which is not out of the box experience like Claude has with its automatic `/terminal-setup` initialization.

deepankarsharma:
> I assume for this to work in iTerm2, the "Report keys using CSI u" setting needs to be checked?
> 
> image
> Is it checked by default? If not, then we'd need to tell users to "go to dialog X > Y > Z, and check the A box there" which is not out of the box experience like Claude has with its automatic `/terminal-setup` initialization.

We work out of the box on iterm2 even without that box ticked. 

Here are my settings 
Screenshot 2025-08-11 at 7 38 11 PM

And here is what the session looked like 
Screenshot 2025-08-11 at 7 37 31 PM

We fall into the same category as neovim which is listed on the iterm2 website as working out of the box with csi-u with no additional configuration required
Screenshot 2025-08-11 at 7 38 21 PM

aalexand:
> We work out of the box on iterm2 even without that box ticked.

That's cool!

jacob314 (approved):
lgtm

jacob314 (approved):
lgtm

PR #3296, opened 2025-07-05
fix(test): Stabilize flaky test in useLoadingIndicator

BigUncle:
## TLDR

This pull request fixes a flaky (unstable) unit test in `useLoadingIndicator.test.ts`. The test was failing intermittently because it incorrectly assumed a randomized loading phrase would always be different after a set interval.

After incorporating code review feedback on an initial approach, this PR implements a more robust, single-line assertion. This new check correctly validates the loading phrase in all scenarios, ensuring the test is stable and reliable, which improves the health of our CI pipeline.

## Dive Deeper

The root cause of the flakiness was the `should reflect values when Responding` test case. It incorrectly asserted that the loading phrase *must* change after a set interval. However, since the phrases are chosen randomly from the `WITTY_LOADING_PHRASES` array, there's a small but non-zero chance that the same phrase could be selected twice in a row, causing the test to fail.

An initial fix attempt used an `if/else` block. However, as pointed out during code review, that approach introduced a new logical flaw: it failed to validate the new phrase against the master list if the phrase was different.

This PR implements the recommended, superior solution: a single assertion that always checks if the `currentLoadingPhrase` is a valid member of the `WITTY_LOADING_PHRASES` array. This is simpler, more correct, and fully resolves the flakiness while maintaining the test's integrity.

## Reviewer Test Plan

1.  Pull down this branch.
2.  Run `npm run preflight` multiple times.
3.  Observe that the tests, specifically the one in `packages/cli/src/ui/hooks/useLoadingIndicator.test.ts`, pass consistently every time. No further user-facing validation is needed as this is a test-only change.

## Testing Matrix

|          | Local (WSL2) | CI | Notes |
| -------- | :---: | :---: | :--- |
| npm run  |   ✅  |   ✅  | All checks pass, including the previously flaky test. |
| npx      |  N/A  |  N/A  | Not applicable for this change. |
| Docker   |  N/A  |  N/A  | Not applicable for this change. |
| Podman   |   -   |   -   | Not applicable for this change. |
| Seatbelt |   -   |   -   | Not applicable for this change. |

## Linked issues / bugs

This PR addresses a flaky test discovered during the work on another branch. It does not close any specific existing issue but improves the overall health of the `main` branch.


scottdensmore:
Looks like all the changes were already merged in so closing out this PR.

Issue #3311, opened 2025-07-05
Add @file command support for non-interactive cli input


zfflxx:
### What would you like to be added?

Extend the existing @file command functionality to work in non-interactive (piped) mode. Currently, when users input @filename in interactive UI, the file  content is automatically read and added to the chat context. This same logic should be implemented for non-interactive mode.

### Why is this needed?

In non-interactive cli mode, without direct file reading, Gemini cli needs to first create a tool call request and then read the file, requiring multiple API calls and taking longer to  complete. 

 For example, Claude Code supports: claude -p "explain this code to me @test.js" - completing the task in a single interaction without additional ReadFile  tool calls. Gemini CLI should provide the same efficient workflow.

  Benefits:
  - Performance: Eliminates extra API calls and reduces response time
  - Consistency: Unified @file behavior across interactive and non-interactive modes
  - Efficiency: Enables single-command workflows for file-based queries

allenhutchison:
Non interative mode is supposed to be more like a traditional command line interface where you would pipe the contents of your file into the command. To achieve this you would do the following:

`cat test.js | gemini  --prompt "Explain this code to me.`

This keeps our TUI code focused on the TUI and maintains the separation of concerns between the two. If you want to pipe whole directories to your CLI invocation I would recommend https://github.com/simonw/files-to-prompt in that case you would run something like:

`files-to-prompt src/gemini-cli/packages/core | gemini --prompt "Explain this code to me"`

PR #3315, opened 2025-07-05
Add @file command support for non-interactive cli input (#3311)

zfflxx:
## TLDR

This PR adds `@file` command support to non-interactive (piped) input mode. Previously, `@file` commands only worked in interactive mode, forcing users to rely on slower tool calls in pipeline workflows. Now both modes provide consistent file reference functionality.

## Dive Deeper

The existing `@file` command implementation in `atCommandProcessor.ts` was only integrated into the interactive UI flow via `useGeminiStream.ts`. Non-interactive mode (`nonInteractiveCli.ts`) processed piped input directly without checking for `@file` commands.

This creates a significant UX inconsistency:
- **Interactive**: `@scripts/start.js` → instant file content inclusion
- **Non-interactive**: `@scripts/start.js` → treated as literal text, requiring tool calls

The fix integrates the existing `handleAtCommand` logic into `runNonInteractive` before message construction. This reuses all existing functionality (path resolution, glob patterns, escaped spaces, error handling) without code duplication.

**Key design decisions:**
- Reuse existing `handleAtCommand` function with minimal stub callbacks
- Process `@file` commands before creating the initial message
- Maintain backward compatibility for non-`@file` inputs
- Handle both string and PartListUnion return types from the processor

## Reviewer Test Plan

1. **Build and test the changes:**
   ```bash
   npm run build
   ```

2. **Test non-interactive @file functionality:**
   ```bash
   # Single file reference
   echo "Explain this code @scripts/start.js" | npm start -p
   
   # Multiple file references  
   echo "Compare @package.json @packages/cli/package.json" | npm start -p
   
   # Mixed content
   echo "Review the following files for security issues: @src/*.ts" | npm start -p
   
   # Directory reference
   echo "Summarize the code structure @src/" | npm start -p
   ```

3. **Verify interactive mode still works:**
   ```bash
   npm start
   # Then in interactive mode: "Explain @scripts/start.js"
   ```

4. **Test error handling:**
   ```bash
   echo "Read this file @nonexistent.txt" | npm start -p
   ```

**Expected behavior:**
- File content should be included in the initial context
- No additional tool calls for file reading
- Consistent behavior between interactive and non-interactive modes

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

*Tested on macOS with `npm run` - other platforms need validation*

## Linked issues / bugs

This PR addresses a feature gap where `@file` commands work inconsistently across CLI modes, impacting user experience and workflow automation capabilities.

### DIFF
--- a/packages/cli/src/nonInteractiveCli.ts
+++ b/packages/cli/src/nonInteractiveCli.ts
@@ -17,6 +17,7 @@ import {
   Part,
   FunctionCall,
   GenerateContentResponse,
+  PartListUnion,
 } from '@google/genai';
 
 import { parseAndFormatApiError } from './ui/utils/errorParsing.js';
@@ -64,12 +65,12 @@ export async function runNonInteractive(
   const abortController = new AbortController();
   
   // Process @file commands if present
-  let processedInput: string | any = input;
+  let processedInput: string | PartListUnion = input;
   if (isAtCommand(input)) {
     const atCommandResult = await handleAtCommand({
       query: input,
       config,
-      addItem: () => Date.now(), // Non-interactive mode doesn't need history, return timestamp as ID
+      addItem: (..._args: any[]) => Date.now(), // Non-interactive mode doesn't need history, return timestamp as ID
       onDebugMessage: () => {}, // Non-interactive mode doesn't need debug messages
       messageId: Date.now(),
       signal: abortController.signal,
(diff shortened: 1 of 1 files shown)


greycodee:
link issue #3311

allenhutchison:
Non interative mode is supposed to be more like a traditional command line interface where you would pipe the contents of your file into the command. To achieve this you would do the following:

cat test.js | gemini  --prompt "Explain this code to me.

This keeps our TUI code focused on the TUI and maintains the separation of concerns between the two. If you want to pipe whole directories to your CLI invocation I would recommend https://github.com/simonw/files-to-prompt in that case you would run something like:

files-to-prompt src/gemini-cli/packages/core | gemini --prompt "Explain this code to me"

PR #3318, opened 2025-07-05
Add Copilot integration and update configuration

SkybuckFlying:
Attempts to add github co-pilot.

However github co-pilot might be missing an api endpoint and public api keys ?

If available in future or now, then this should be possible...


scottdensmore:
@SkybuckFlying thanks for the submitting the PR. We don't have any plans on adding support for Copilot so we are gong to close this PR. If this is something you want to discuss, please file an issue.

SkybuckFlying:
Some things needed to be installed to get it working... ask the AI...

PR #3333, opened 2025-07-05
fix(cli): Group cancelled tool call responses to prevent API errors

### LINKED ISSUE #3131: Code 400
### What happened?

✕ [API Error: [{
    "error": {
      "code": 400,
      "message": "Please ensure that the number of function response parts is equal to the number of function call
   parts of the function call turn.",
      "errors": [
        {
          "message": "Please ensure that the number of function response parts is equal to the number of function 
  call parts of the function call turn.",
          "domain": "global",
          "reason": "badRequest"
        }
      ],
      "status": "INVALID_ARGUMENT"
    }
  }
  ]]

make a plan

### Client information

*   **CLI Version:** 0.1.9
*   **Git Commit:** N/A
*   **Operating System:** win32 v22.16.0
*   **Memory Usage:** 144.3 MB

### Login information

_No response_

_No response_

NTaylorMullen:
## TLDR

- When the model requests multiple tool calls in a single turn and they are all cancelled, the client was sending a separate `functionResponse` for each cancellation.
- This violates the Gemini API's expectation that all responses for a single multi-tool-call request be bundled in a single user turn, leading to a 400 Bad Request error.
- This change corrects the behavior by grouping all `functionResponse` parts from the cancelled tools into a single user message before adding it to the history.

This pull request fixes a bug where the client would send individual responses for multiple cancelled tool calls, leading to a `400 Bad Request` from the Gemini API. The fix ensures that all cancelled tool call responses from a single model turn are grouped into a single user message.

## Dive Deeper

The root cause of the issue was in `packages/cli/src/ui/hooks/useGeminiStream.ts`. When multiple tool calls from a single model request were cancelled, the `handleCompletedTools` function would iterate over the cancelled tools and call `geminiClient.addHistory` for each one. This resulted in multiple user messages being sent to the Gemini API, which is not the expected behavior.

The fix involves collecting all the `functionResponse` parts from the cancelled tool calls into a single array and then making a single call to `geminiClient.addHistory`. This ensures that the Gemini API receives a single user message with all the cancelled tool call responses, which is the correct behavior.

A new test case has been added to `packages/cli/src/ui/hooks/useGeminiStream.test.tsx` to cover this specific scenario and prevent future regressions.

## Reviewer Test Plan

To test this change, a reviewer can manually trigger a scenario where multiple tool calls are cancelled.

1.  Run the CLI.
2.  Enter a prompt that will trigger multiple tool calls, for example: `Write a 3 paragraph cat poem. Write each paragraph to Foo.md, Bar.md and Baz.md. Do all 3 tool calls using parallel tool calls`
3.  When the confirmation prompt for the tool calls appears, cancel all of them.
4.  Type "hello" and ensure you get a response without an error

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ✅  | ❓  | ❓  |
| Docker   | ✅   | ❓  | ❓  |

## Linked issues / bugs

Fixes #3131

### DIFF
--- a/packages/cli/src/ui/hooks/useGeminiStream.test.tsx
+++ b/packages/cli/src/ui/hooks/useGeminiStream.test.tsx
@@ -604,6 +604,112 @@ describe('useGeminiStream', () => {
     });
   });
 
+  it('should group multiple cancelled tool call responses into a single history entry', async () => {
+    const cancelledToolCall1: TrackedCancelledToolCall = {
+      request: {
+        callId: 'cancel-1',
+        name: 'toolA',
+        args: {},
+        isClientInitiated: false,
+      },
+      tool: {
+        name: 'toolA',
+        description: 'descA',
+        getDescription: vi.fn(),
+      } as any,
+      status: 'cancelled',
+      response: {
+        callId: 'cancel-1',
+        responseParts: [
+          { functionResponse: { name: 'toolA', id: 'cancel-1' } },
+        ],
+        resultDisplay: undefined,
+        error: undefined,
+      },
+      responseSubmittedToGemini: false,
+    };
+    const cancelledToolCall2: TrackedCancelledToolCall = {
+      request: {
+        callId: 'cancel-2',
+        name: 'toolB',
+        args: {},
+        isClientInitiated: false,
+      },
+      tool: {
+        name: 'toolB',
+        description: 'descB',
+        getDescription: vi.fn(),
+      } as any,
+      status: 'cancelled',
+      response: {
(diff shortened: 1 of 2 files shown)


mattKorwel (approved):
LGTM. I tested the code and didn't get an error. However i also didn't get an error on main so maybe my test is invalid?

jkaster:
This is great news since I've ran into this a few times. How do I get the fix? Do I just restart gemini CLI, does it self-update, or do I need to perform some update command?

PR #3343, opened 2025-07-05
feat(auth): make terms of service link clickable

sinanptm:
## TLDR

feat(auth): make terms of service link clickable

## Dive Deeper

This pull request enhances the user experience of the authentication flow by making the Terms of Service link clickable.

Previously, the URL was displayed as plain text. This change replaces the `Text` component with the `Link` component from the `ink-link` library, allowing users to directly open the link from their terminal.

This improves usability and makes it easier for users to access the Terms of Service and Privacy Notice.

## Reviewer Test Plan

1. Pull down this branch.
2. Run `npm install` to install the new `ink-link` dependency.
3. Run the CLI in a way that triggers the authentication dialog (e.g., by running a command that requires authentication without being logged in, or by clearing existing credentials).
4. Verify that the Terms of Service link is now blue (or the accent color of your terminal theme).
5. Cmd+click (or the equivalent for your terminal) on the link.
6. Verify that the link opens the correct URL (`https://github.com/google-gemini/gemini-cli/blob/main/docs/tos-privacy.md`) in your web browser.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅ | ❓  | ❓  |
| npx      | ✅  | ❓  | ❓  |
| Docker   | ✅  | ❓  | ❓  |
| Podman   | ✅  | -   | -   |
| Seatbelt | ✅  | -   | -   |

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,9 +10,6 @@
       "workspaces": [
         "packages/*"
       ],
-      "dependencies": {
-        "shell-quote": "^1.8.3"
-      },
       "bin": {
         "gemini": "bundle/gemini.js"
       },
@@ -11384,7 +11381,7 @@
         "ignore": "^7.0.0",
         "micromatch": "^4.0.8",
         "open": "^10.1.2",
-        "shell-quote": "^1.8.2",
+        "shell-quote": "^1.8.3",
         "simple-git": "^3.28.0",
         "strip-ansi": "^7.1.0",
         "undici": "^7.10.0",
--- a/packages/cli/src/ui/components/AuthDialog.tsx
+++ b/packages/cli/src/ui/components/AuthDialog.tsx
@@ -12,6 +12,8 @@ import { LoadedSettings, SettingScope } from '../../config/settings.js';
 import { AuthType } from '@google/gemini-cli-core';
 import { validateAuthMethod } from '../../config/auth.js';
 
+import Link from 'ink-link';
+
 interface AuthDialogProps {
   onSelect: (authMethod: AuthType | undefined, scope: SettingScope) => void;
   settings: LoadedSettings;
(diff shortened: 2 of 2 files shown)


NTaylorMullen:
Thanks for the contribution @sinanptm ! Unfortunately this looks to not work consistently in all terminals sadly. I gave it a try in VSCode and worked beautifully; however, trying in Ghostty:
image

Is there something more we could do here?

sinanptm:
Thanks for the feedback! You're right, terminal support for clickable links can be inconsistent

 To address this, I'm proposing a change to always display the full URL next to the link text in the output. This way, even if the terminal doesn't support the clickable link feature, the URL will still be visible and can be copied by the user. It ensures the information is always accessible, regardless of the terminal
  emulator.
  
  I'm ready to implement this change if that sounds like a good approach. Let me know your thoughts

image

However, when I checked it on Ghostty, it was working fine as expected

NTaylorMullen:
> To address this, I'm proposing a change to always display the full URL next to the link text in the output.

Would my screenshot be the normal in that case (link duplicated)?

> However, when I checked it on Ghostty, it was working fine as expected

Hmmm this is my version currently: 
image

I also have the default settings so I wonder if there's some customization.

sinanptm:
I also have the same version, without any customisation.

It should solve the link duplication issue if it's caused by the link

NTaylorMullen:
> I also have the same version, without any customisation.
> 
> It should solve the link duplication issue if it's caused by the link

Unfortunately it doesn't on my side. We may need to do more work to understand when we can or can't render links.

sinanptm:
@NTaylorMullen, I have just made some changes. Could you please check it on your device? I really appreciate it.

It's working fine for me.

For ref, my system is: `Mac M3 Air`

PR #3349, opened 2025-07-06
feat(core): improve error messages in isCommandAllowed

### LINKED ISSUE #3268: Improve Error Messages in isCommandAllowed in shell.ts
### What happened?

Because of the lack of the context to reject the shell command, Gemini thinks a command is prohibited to use even if the command is not listed in the deny list.

Image

The error message is always `Command is not allowed: <command>` even if it was rejected because the command includes ` backquote.
https://github.com/google-gemini/gemini-cli/blob/e90e0015eacda0cf8315e30b72305e026f3768b3/packages/core/src/tools/shell.ts#L200-L203

The shell tool should return more details for prohibiting running a command. Then Gemini can retry it in another way.

For example, if gemini fails to commit a change with the following command, then Gemini think the git command is prohibited and doesn't understand the message part is wrong.

```shell
git commit -m "feat: adding \`foo\` field on the data model"
```

### Client information

<details>

│ About Gemini CLI│
│ :-:│
│ CLI Version          0.1.9                                    │
│ Git Commit           34935d6                                  │
│ Model                gemini-2.5-pro                           │
│ Sandbox              no sandbox                               │
│ OS                   darwin                                   │
│ Auth Method          gemini-api-key                           │
│                                    
</details>

### Login information

_No response_

It would be better to change the method signature of `isCommandAllowed` to return more detailed error message  rather than just returning `true` or `false`.
https://github.com/google-gemini/gemini-cli/blob/e90e0015eacda0cf8315e30b72305e026f3768b3/packages/core/src/tools/shell.ts#L117

Then Gemini would know more context and retry by itself.

y-okt:
## TLDR

As described in #3268, due to the lack of details in the error messages in `isCommandAllowed` method, Gemini is unable to retry even if the command itself is allowed (for instance, the reason the command is rejected could be that it contains a backquote, not because the command itself is prohibited). To do so, the tools should return reasons why the command is not allowed.

## Dive Deeper

Made `isCommandAllowed` method not only return the boolean value, but also return the reason as well. When outputting the validation result, it will output the reason why the command is rejected.

## Reviewer Test Plan

1. Pull down this branch
2. Run `npm i`
3. Run `npm run build`
4. Run `npm start`
5. In the input screen, input `Run printf "This is \`NOT\` allowed command" `
6. It will output something like below
スクリーンショット 2025-07-06 11 01 34

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

- Resolves #3268


packages/core/src/tools/shell.ts
```diff
@@ -172,11 +186,14 @@ Process Group PGID: Process group started or \`(none)\``,
 
     for (const cmd of commandsToValidate) {
       // 2. Check if the command is on the blocklist.
-      const isBlocked = [...blockedCommands].some((blocked) =>
+      const blockedCommand = [...blockedCommands].find((blocked) =>
         isPrefixedBy(cmd, blocked),
       );
```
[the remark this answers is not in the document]
y-okt: Thanks, fixed

packages/core/src/tools/shell.ts
```diff
@@ -170,35 +184,45 @@ Process Group PGID: Process group started or \`(none)\``,
 
     const commandsToValidate = command.split(/&&|\|\||\||;/).map(normalize);
 
+    const blockedCommandsArr = [...blockedCommands];
```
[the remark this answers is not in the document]
y-okt: The spread syntax is already applied.

packages/core/src/tools/shell.ts
```diff
@@ -114,10 +114,21 @@ Process Group PGID: Process group started or \`(none)\``,
    * @param command The shell command string to validate
    * @returns True if the command is allowed to execute, false otherwise
```
kyasbal: Maybe we need to update this `@returns` type doc?

packages/core/src/tools/shell.ts
```diff
@@ -170,35 +184,45 @@ Process Group PGID: Process group started or \`(none)\``,
...
-    return true;
+    return { allowed: true };
   }
 
   validateToolParams(params: ShellToolParams): string | null {
-    if (!this.isCommandAllowed(params.command)) {
-      return `Command is not allowed: ${params.command}`;
+    const commandCheck = this.isCommandAllowed(params.command);
+    if (!commandCheck.allowed) {
+      return commandCheck.reason || `Command is not allowed: ${params.command}`;
```
kyasbal: `commentCheck.reason` must not be empty when `commandCheck.allowed=true`. The message used for the non expected case `!commentCheck.reason && !commandCheck.allowed` would be better to express something unexpected happening.

kyasbal (commented):
@y-okt 
Because I created the original issue #3268 and I'm interested this issue being fixed, I'll just leave review comments here. I'm a Googler but I'm not related to this project, thus these are just my personal comments not as an engineer related to this project. Please just ignore them if you think these are not helpful.

packages/core/src/tools/shell.ts
y-okt: @kyasbal Thank you for your review and the catch! I fixed it.

packages/core/src/tools/shell.ts
y-okt: Agree, I added a branch when `commandCheck.allowed` is false.

NTaylorMullen (approved):
Thanks for the contribution @y-okt!! Getting this in.

y-okt:
@NTaylorMullen Thank you for merging!!

PR #1955, opened 2025-06-26
feat: Add markdown table rendering support

heartyguy:
## Summary
- Added markdown table rendering support to enhance readability of tabular data in the terminal
- Tables are now rendered with proper formatting using box-drawing characters instead of raw markdown
- Implemented automatic column width scaling to fit terminal width

## Visual Example

**Before (raw markdown):**
```
| Name    | Age | City     |
|---------|-----|----------|
| Alice | 30 | New York |
| Bob   | 25 | London |
| Charlie|35  |Tokyo  |
```

**After (formatted table):**
```
┌─────────┬─────┬──────────┐
│ Name    │ Age │ City     │
├─────────┼─────┼──────────┤
│ Alice   │ 30  │ New York │
│ Bob     │ 25  │ London   │
│ Charlie │ 35  │ Tokyo    │
└─────────┴─────┴──────────┘
```

## Dive Deeper
This PR implements the feature requested in #1775 to render markdown tables in a formatted way rather than displaying raw markdown syntax.

### Implementation Details:
1. **Table Detection**: Added regex patterns to detect markdown table syntax with pipe separators and header rows
2. **Custom TableRenderer Component**: Created a new component that renders tables using Unicode box-drawing characters for a polished terminal appearance
3. **Integration**: Seamlessly integrated table parsing into the existing `MarkdownDisplay` component without affecting other markdown features
4. **Features**:
   - Supports standard markdown table syntax
   - Handles varying column widths with automatic scaling
   - Properly handles empty cells
   - Works well with mixed content (tables alongside other markdown elements)
   - Headers are highlighted in cyan for better visibility

### Technical Approach:
- Extended the existing markdown parser to detect table patterns
- Created a dedicated `TableRenderer` component for clean separation of concerns
- Added comprehensive test coverage for various table scenarios

## Reviewer Test Plan
1. Clone and build the branch:
   ```bash
   git checkout feature/issue-1775-markdown-table-rendering
   npm install
   npm run build
   ```

2. Test with markdown containing tables:
   ```bash
   npm start
   ```
   Then try prompts like:
   - "Create a markdown table comparing Python vs JavaScript features"
   - "Show me a table of common HTTP status codes"
   - "Generate a table with varying column widths"

3. Run the tests:
   ```bash
   cd packages/cli && npm test MarkdownDisplay
   ```

## Testing Matrix
- [x] macOS
- [ ] Linux
- [ ] Windows
- [x] Unit tests pass
- [x] Integration tests pass
[...]

PR #3362, opened 2025-07-06
Fix nested markdown Rendering for table headers and rows #3331

### LINKED ISSUE #3331: Improve nested markdown Rendering for table headers and rows

See the rendered markdown table in the below image:

Image

Improvements:
- We should be rendering the back ticks properly in the headers
- The bold `**` isn't being respected.

### Why is this needed?

Our markdown rendering should be as awesome as possible 😄 

### Additional context

Here's how GitHub renders this:
| Characteristic | `execSync` (Old Way) | `spawn` (New Way in PR) |
| :--- | :--- | :--- |
| **Execution** | Synchronous (blocks everything) | Asynchronous (non-blocking) |
| **I/O Handling** | Buffers entire output in memory | Streams data in chunks (memory efficient) |
| **Security** | **Vulnerable to shell injection** | **Safe from shell injection** |
| **Use Case** | Simple, quick commands with small, trusted... | Long-running processes, large I/O, and especially for running user-configur... |

zfflxx:
截屏2025-07-06 14 52 32

also fixed https://github.com/google-gemini/gemini-cli/pull/1955#issuecomment-3031129505

### DIFF
--- a/packages/cli/src/ui/utils/InlineMarkdownRenderer.tsx
+++ b/packages/cli/src/ui/utils/InlineMarkdownRenderer.tsx
@@ -1,0 +1,171 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import React from 'react';
+import { Text } from 'ink';
+import { Colors } from '../colors.js';
+import stringWidth from 'string-width';
+
+// Constants for Markdown parsing
+const BOLD_MARKER_LENGTH = 2; // For "**"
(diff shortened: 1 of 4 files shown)


NTaylorMullen (approved):

NTaylorMullen:
@zfflxx, thank you so much for your contribution and for taking a crack at the inline markdown rendering!

I love that you took the initiative to pull out the inline markdown component into its own dedicated component file + the `getPlainTextLength` approach is a very nice touch.

There are probably some smaller naming things we could have adjusted BUT overall very solid change. Love that you took this on, thank you!

PR #3056, opened 2025-07-02
feat: add user startup warnings, add home directory check

psinha40898:
## TL;DR

This PR implements a utility function to return startup warnings designed for the user experience. 

One such warning is implemented, it checks if the Gemini CLI is executed from the home directory, addressing part of #2617.

---

## Dive Deeper
`startupWarnings` is currently architected for developer warnings via `npm run start`.  
This can be conflicting for developers trying to add user-facing warnings when getting familiar with the codebase.  
See https://github.com/google-gemini/gemini-cli/pull/2974#issuecomment-3029374119

Therefore, this PR creates `userStartupWarnings.ts` and a corresponding test file. It also adds a home directory warning with respect to #2617.

This separates all future user focused warnings from having to push each warning directly to the startupWarnings array, and allows contributors to test new warnings.

Instead, all developer and user focused startup warnings can be merged before being passed down to App.

#2617 and #2930 show that the inclusion of user-facing warnings may be a recurring request that would benefit from this abstraction. With this PR, contributors can rely on a separation between developer and user warnings, and their tests.

This will be useful for all future developers adding simple startup warnings that should be displayed to the user.

---

## Reviewer Test Plan

**To test development:**

```bash
npm install
npm run start
```

**To test build:**

```bash
npm install
npm run build
npm link .
```
Then enter the home/user directory of your OS and run the linked build.

For example:

```bash
cd home/psinha98
gemini
```

---

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| -------- | --- | --- | --- |
| npm run  | ✅  | ✅  | ✅  |
| npx      | ✅  | ✅  | ✅  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

---

## Linked Issues 

- Fixes #2617

PR #3371, opened 2025-07-06
feat: add a warning that shows if user uses node -v <20 #2930

### LINKED ISSUE #2930: Warn users if they are on node < 20

Node 18 is out of security support and I would rather we didnt have to spend too much time making our app work with it unless a larger number of our users than I am guessing are still on it. https://endoflife.date/nodejs

We should display a warning on the startup screen when they are using node < 20.

### Why is this needed?

Worrying about node < 20 results in us spending time on items like https://github.com/google-gemini/gemini-cli/issues/2474

If we should support node 18 please comment on this bug.

### Additional context

_No response_

devpool007:
## TLDR

Added a warning message which is visible on Startup if the node version of user is < 20

## Dive Deeper
Node 18 is out of security support and hence the team would need to spend quite some time figuring out ways to work around this and make app fully functional.

Hence we should display a warning on the startup screen when users are using node < 20. So they can most likely update to get much better experience and the latest working functionality. 

I have added a utility function for this in which the minimum required node version can be dynamically set currently it's set to 20 but can be easily updated in the future.

## Reviewer Test Plan
- npm run build and npm start the project
- If your node version is < 20 you should be shown such a warning at startup 

Screenshot 2025-07-02 at 23 55 19

PLEASE NOTE - The above screenshot was just for testing since I have node 23 on my system. So, I set the nodeWarning for 24 minimum

- Else no warning should be shown regarding this

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ✅  |
| npx      | ✅  | ❓  | ✅  |
| Docker   | ✅  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ✅  | -   | -   |

## Linked issues / bugs
Fixes #2930


packages/cli/src/utils/nodeVersionWarning.test.ts
```diff
@@ -0,0 +1,54 @@
...
+    const warning = getNodeVersionWarning(20);
+    expect(warning).toBeNull();
+  });
+
+  it('should use default minMajor=20 if not provided', () => {
+    setNodeVersion('18.0.0');
+    const warning = getNodeVersionWarning();
+    expect(warning).toContain('Node.js v18.0.0');
+    expect(warning).toContain('requires Node.js 20 or higher');
+  });
```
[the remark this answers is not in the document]
devpool007: Added the fix

psinha40898:
You could do this without creating additional files. 

The logic for the check can be added as another warning in `userStartupWarnings.ts` and the tests in the `userStartupWarnings.test.ts` (see the commented block at the bottom of the test file). This was actually the purpose of #3056, to prevent file clutter when adding more warnings.

devpool007:
Hi @psinha40898 ahhh ok thanks for pointing that out. I implement in this way then, I will try to add the tests in userStartupWarnings.ts could you then help me review the code?

psinha40898:
> Hi @psinha40898 ahhh ok thanks for pointing that out. I implement in this way then, I will try to add the tests in userStartupWarnings.ts could you then help me review the code?

Yes sir I'd love to give it a preliminary review when you're ready

psinha40898:
Yes this is exactly what I meant. Now it's up to the core team. 

Good luck with all things and thank you for collaborating with me on these warnings!

devpool007:
Hey anytime! Thanks for the quick review and your help :)

devpool007:
Hi @jacob314 please let me know if you get a moment to review or assign someone to this. I have made sure now it follows the userStartupWarning approach which was merged into main recently.

packages/cli/src/utils/userStartupWarnings.ts
```diff
@@ -32,8 +33,23 @@ const homeDirectoryCheck: WarningCheck = {
   },
 };
 
+const nodeVersionCheck: WarningCheck = {
+  id: 'node-version',
+  check: async (_workspaceRoot: string) => {
+    const minMajor = 20;
+    const major = semver.major(process.versions.node);
+    if (major < minMajor) {
+      return `Warning: You are using Node.js v${process.versions.node}. Gemini CLI requires Node.js ${minMajor} or higher for best results.`;
```
jacob314: nit: remove the text `Warning: ` to conform with the other checks that do not not include it for their warnings.

packages/cli/src/utils/userStartupWarnings.test.ts
```diff
@@ -66,6 +66,49 @@ describe('getUserStartupWarnings', () => {
     });
   });
 
+  function setNodeVersion(version: string) {
```
jacob314: nit: can you use refactor this to use the standard mocking libraries used in other tests rather than a direct defineProperty? If there is a reason that doesn't work fine to leave but generally would recommend mocking rather than directly patching properties.

jacob314 (commented):
looks good once the following two nits are addressed. thanks for the fix!
sorry for the delay! I was just traveling back from vacation.

packages/cli/src/utils/userStartupWarnings.test.ts
devpool007: Used the standard mocking libs , thanks for the idea.

devpool007:
Addressed the nits, hope it looks good now. Thanks for the review! 
No worries hope you had a nice vacation :)

jacob314 (approved):
lgtm

devpool007:
Hi @jacob314 since the PR is ready to be merged do I need to ping someone to do this?

jacob314:
> Hi @jacob314 since the PR is ready to be merged do I need to ping someone to do this?

Taking care of that now. Thanks again for the pull request!

NTaylorMullen (approved):

PR #3372, opened 2025-07-06
fix: Prevent infinite loop when building on Windows

### LINKED ISSUE #2084: CONTRIBUTING.md is incorrect; also build on Windows hangs
### What happened?

I tried building on Windows. It seems to hang after building gemini-cli-core. 
```
PS C:\Users\dpchi\dev\gemini-cli> npm run build

> @google/gemini-cli@0.1.5 build
> node scripts/build.js


> @google/gemini-cli@0.1.5 generate
> node scripts/generate-git-commit-info.js


> @google/gemini-cli@0.1.5 build
> node ../../scripts/build_package.js

Successfully copied files.

> @google/gemini-cli-core@0.1.5 build
> node ../../scripts/build_package.js

Successfully copied files.
```

It  stayed there for 15 minutes before I stopped it.   I have podman installed on my Windows machine, and it appears to be functioning correctly.  So I would think it would find podman.  But somehow it didn't?  It just hung forever, I think in sandbox_command.js. 

But I don't really want the sandbox anyway; I read the README and am not clear on why I would need it .  I just want to build the latest commit. 

Finally solved the hang by explicitly setting the env variable: 
```
 [System.Environment]::SetEnvironmentVariable('GEMINI_SANDBOX', '0', 'Process')
```

And then `npm run build` worked and finished.  

From CONTRIBUTING.md it says I need to run `npm run build:all` to get the sandbox.  But then why check for the existence of podman or docker if I didn't use `build:all` ? Something is broken here.

**EXPECTED**

1. Seems like, to conform with the intended documented behavior in CONTRIBUTING.md,  the build script ought to check for the env variable BUILD_SANDBOX first, and then check for podman or docker iff BUILD_SANDBOX is true. 
2. Secondary bug: On Windows, with podman, ... the sandbox build does not work. Or, it is very quiet for a very very long time. No indication of progress. Possible that its downloading something , but I didn't see any progress.

| Q | A |
| --- | --- |
| node --version | v22.13.0 |
| npm --version |  10.9.2 | 
| GIT_COMMIT_INFO | 'bf873a1d (local modifications) (I think these are just my diagnostic mods to the build scripts) |
| Windows Version | 10.0.26100 Build 26100 |

The build should terminate relatively quickly, or show some sort of progress message as it works.

### Client information

<details>

```console
PS> gemini --version
(node:17196) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
(Use `node --trace-deprecation ...` to show where the warning was created)
0.1.5
```

There is no command-line flag for `/about` in Windows. 
Within the cli:
```console
About Gemini CLI
│                                                                          │
│ CLI Version              0.1.5                                           │
│ Git Commit               bf873a1d                                        │
│ Model                    gemini-2.5-flash                                │
│ Sandbox                  no sandbox                                      │
│ OS                       win32                
```
</details>

### Login information

This is a build-time error. I haven't even got to the login part.

nope

Principal-Ideal:
## TLDR

This change fixes a potential infinite loop in the `.env` file discovery logic when building on Windows.

## Dive Deeper

The original code used `while (currentDir !== '/')` to terminate the search for an `.env` file at the filesystem root. However, this condition is never met on Windows, causing the while loop to run indefinitely if no `.env` file is found.

The updated code instead checks if the parent directory is the same as the current directory `(parentDir === currentDir)`, which reliably stops the loop on any operating system.

## Reviewer Test Plan

On a Windows machine, navigate to a project directory that does not contain a `.env` file in its path.

Run `npm run build` and verify that it terminates successfully without hanging.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Resolves #2084

### DIFF
--- a/scripts/sandbox_command.js
+++ b/scripts/sandbox_command.js
@@ -48,7 +48,7 @@ if (!geminiSandbox) {
 
 if (!geminiSandbox) {
   let currentDir = process.cwd();
-  while (currentDir !== '/') {
+  while (true) {
     const geminiEnv = join(currentDir, '.gemini', '.env');
     const regularEnv = join(currentDir, '.env');
     if (existsSync(geminiEnv)) {
@@ -58,7 +58,11 @@ if (!geminiSandbox) {
       dotenv.config({ path: regularEnv, quiet: true });
       break;
     }
-    currentDir = dirname(currentDir);
+    const parentDir = dirname(currentDir);
+    if (parentDir === currentDir) {
+      break;
+    }
+    currentDir = parentDir;
   }
   geminiSandbox = process.env.GEMINI_SANDBOX;
 }


NTaylorMullen:
@Principal-Ideal thank you for the contribution! Would you mind writing some tests for this change?

SandyTao520:
Already fixed

PR #3418, opened 2025-07-07
Release misc

mattKorwel:
## TLDR

Release Fixes.

## Dive Deeper

- fix flaky, "list directory test". It only outputs lines in the style of "ls" half the time, so looking for 2 lines doesn't work.
- Fix output flags on integration tests
- Add tests for release version calcs
- fix bug tagging for failed tests
- ensure that the new version scripts don't wait for user input.

### DIFF
--- a/integration-tests/list_directory.test.js
+++ b/integration-tests/list_directory.test.js
@@ -15,11 +15,10 @@ test('should be able to list a directory', async (t) => {
   rig.mkdir('subdir');
   rig.sync();
 
-  const prompt = `Can you list the files in the current directory`;
-  const result = await rig.run(prompt);
+  const prompt = `Can you list the files in the current directory. Display them in the style of 'ls'`;
+  const result = rig.run(prompt);
 
   const lines = result.split('\n').filter((line) => line.trim() !== '');
-  assert.equal(lines.length, 2);
-  assert.ok(lines.includes('file1.txt'));
-  assert.ok(lines.includes('subdir'));
+  assert.ok(lines.some((line) => line.includes('file1.txt')));
+  assert.ok(lines.some((line) => line.includes('subdir')));
 });
--- a/integration-tests/run-tests.js
+++ b/integration-tests/run-tests.js
@@ -72,35 +72,56 @@ async function main() {
       `------------- Running test file: ${testFileName} ------------------------------`,
     );
 
-    const child = spawn('node', ['--test', testFile], {
+    const nodeArgs = ['--test'];
+    if (verbose) {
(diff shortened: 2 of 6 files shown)


mattKorwel:
Nightly release generated here https://github.com/google-gemini/gemini-cli/actions/runs/16105486440/job/45440297378

integration-tests/run-tests.js
```diff
@@ -72,35 +72,56 @@ async function main() {
...
+    });
+
     const exitCode = await new Promise((resolve) => {
-      child.on('close', resolve);
+      child.on('close', (code) => {
+        if (outputStream) {
+          outputStream.end();
+        }
+        resolve(code);
+      });
```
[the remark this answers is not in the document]
mattKorwel:
```suggestion
      child.on('close', (code) => {
        if (outputStream) {
          outputStream.end(() => {
            resolve(code);
          });
        } else {
          resolve(code);
        }
      });
```

integration-tests/list_directory.test.js
```diff
@@ -15,11 +15,10 @@ test('should be able to list a directory', async (t) => {
   rig.mkdir('subdir');
   rig.sync();
 
-  const prompt = `Can you list the files in the current directory`;
-  const result = await rig.run(prompt);
+  const prompt = `Can you list the files in the current directory. Display them in the style of 'ls'`;
+  const result = rig.run(prompt);
 
   const lines = result.split('\n').filter((line) => line.trim() !== '');
-  assert.equal(lines.length, 2);
```
NTaylorMullen, line 22: Was this the bit that was failing before?

integration-tests/run-tests.js
```diff
@@ -72,35 +72,59 @@ async function main() {
       `------------- Running test file: ${testFileName} ------------------------------`,
     );
 
-    const child = spawn('node', ['--test', testFile], {
+    const nodeArgs = ['--test'];
+    if (verbose) {
```
NTaylorMullen, line 76: verbose == the debug checkbox on runs right?

NTaylorMullen (approved):

integration-tests/list_directory.test.js
mattKorwel: Yeah, it only comes out that way. half the time it say "i need an absolute path. Ok now i see files1 and subdir". have we addressed the abs path thing at all?

integration-tests/run-tests.js
mattKorwel: no, this is just for the integration tests, --verbose is "print all the test output so i can figure out why its broken"

PR #3461, opened 2025-07-07
feat(core): add a prompt to make models prefer to inspect differ content by adding about `git show`

### LINKED ISSUE #671: Git interactions should prefer to inspect diffed content (rather than just files changed)
The model loves to assume the content of files and more specifically loves to simply just get status from git and start acting. Instead we should encourage deep introspection of what was done to files (i.e. git show)

y-okt:
## TLDR

As written in #671 , added a prompt to make models prefer to inspect differ content by adding about `git show`.

## Dive Deeper

Described in #671 

## Reviewer Test Plan

This is a prompt change, and the deterministic validation is not possible. However, you can verify by:

- npm install
- npm run build
- npm start
- Input "understand the recent commit change and prepare a commit"

The result will change between before and after the change.
### before:
スクリーンショット 2025-07-08 7 05 41

### after:
スクリーンショット 2025-07-08 7 08 09

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅   | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Resolves #671


y-okt:
@NTaylorMullen Hello, could you please check if my change aligns with the issue you raised (#671)?

packages/core/src/core/prompts.ts
```diff
@@ -136,12 +136,15 @@ ${(function () {
     return `
 # Git Repository
 - The current working (project) directory is being managed by a git repository.
-- When asked to commit changes or prepare a commit, always start by gathering information using shell commands:
-  - \`git status\` to ensure that all relevant files are tracked and staged, using \`git add ...\` as needed.
+- When working with git, prioritize deep introspection of changes over quick assumptions based on file names.
+- When asked to commit changes or prepare a commit, always start by gathering detailed information using shell commands:
+  - \`git show\` to inspect their content to understand recent commits.
+  - \`git status\` to see which files are tracked and staged, but always examine the actual changes in these files, not just file names.
```
NTaylorMullen, line 142: One thing I'm worried about is we don't want the model to "always" inspect the diff whenever it runs git status. We only want it to inspect the diff WHEN it's trying to do something where file content would amplify the result. As an example, it will typically run git status after committing to ensure the commit was successful (this doesn't depend on file content).

Similarly, I'm a little hesitant in taking system prompt changes of this nature off the bat. We're in the midst of making our evals easier to run which I'll want to run this change through.

y-okt: Hi @NTaylorMullen , thank you for your feedback. I understand your concern. I will close this PR, and am looking forward to new evaluation ways!

NTaylorMullen: Thank you for your understanding!

PR #3469, opened 2025-07-07
refactor: rename allowed_mcp_server_names to allowed-mcp-server-names

teeler:
Just what it says in the title - renaming the flag for consistency. 

Also, added a section to GEMINI.md on style and included a bit about using hyphens instead of underscores.

### DIFF
--- a/packages/cli/src/config/config.ts
+++ b/packages/cli/src/config/config.ts
@@ -48,7 +48,7 @@ interface CliArgs {
   telemetryTarget: string | undefined;
   telemetryOtlpEndpoint: string | undefined;
   telemetryLogPrompts: boolean | undefined;
-  allowed_mcp_server_names: string | undefined;
+  'allowed-mcp-server-names': string | undefined;
 }
 
 async function parseArguments(): Promise<CliArgs> {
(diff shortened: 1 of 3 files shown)


jackwotherspoon (approved):
Thanks @teeler 😄

mattKorwel (approved):

packages/cli/src/config/config.ts
```diff
@@ -48,7 +48,7 @@ interface CliArgs {
   telemetryTarget: string | undefined;
   telemetryOtlpEndpoint: string | undefined;
   telemetryLogPrompts: boolean | undefined;
-  allowed_mcp_server_names: string | undefined;
+  'allowed-mcp-server-names': string | undefined;
```
NTaylorMullen, line 51: Should these map 1:1 to the actual user typed name or should they be TypeScripty? I.e. should this be `allowedMCPServerNames`?

NTaylorMullen (approved):

packages/cli/src/config/config.ts
jackwotherspoon: they should probably be TypeScripty, good catch @NTaylorMullen 👍 

Is this something the lint job should catch?

NTaylorMullen: Ya I imagine it could hmmm, if you want to take a stab at adding that would definitely approve! Although I imagine we might hit quite a few errors initially

PR #3483, opened 2025-07-07
Initialize MCP tools once at start up instead of every time we auth.

### LINKED ISSUE #2926: "Authentication timed out" message due to an MCP server connection issue.
### What happened?

I set the environment variable GEMINI_API_KEY and executed gemini.
I confirmed that no other gemini login-related environment variables were set.
The API key is confirmed to be working. When I made an API call using the same key with curl, I received an immediate response.

After executing the gemini CLI, authentication should complete immediately and the prompt should be ready for input.

### Client information

<details>

```console
$ gemini /about
# paste output here
```

</details>

I attempted to execute the gemini CLI multiple times, but it failed, so I couldn't use the /about command.
The installed version is 0.1.9.
I am using zsh on macOS.

### Login information

api key.

Authentication with Vertex AI also fails.
Both express mode and non-express mode failed. I have never successfully authenticated with Vertex AI.
I use GCP for work, so I am confident that I entered the correct values in the environment variables.

scidomino:
## TLDR

Initialize MCP tools once at start up instead of every time we auth.

## Dive Deeper

Previously we were initializing MCP tools every time we authenticate. This was confusing users with MCP errors because they'd see "Waiting for Auth..." in the start up screen when they had a MCP server that was slow to start up. Even worse, if there was an error with MCP users would assume it was related to Auth.

This change does cause startup to hang while we wait for the MCP servers to start up but I think this is still an improvement. We can add a MCP loading spinner in subsequent PRs.

## Reviewer Test Plan

Use the following MCP server in your settings.json to simulate a MCP server that takes a long time to start up.

```
{
  "mcpServers": {
    "slow": {
      "command": "node",
      "args": [
        "-e",
        "setInterval(() => {}, 1000);"
      ],
      "timeout": 10000
    }
  }
}
```

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | x  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes #2926

### DIFF
--- a/packages/cli/src/gemini.tsx
+++ b/packages/cli/src/gemini.tsx
@@ -115,15 +115,7 @@ export async function main() {
 
   setMaxSizedBoxDebugging(config.getDebugMode());
 
-  // Initialize centralized FileDiscoveryService
-  config.getFileService();
-  if (config.getCheckpointingEnabled()) {
-    try {
-      await config.getGitService();
-    } catch {
-      // For now swallow the error, later log it.
-    }
-  }
+  await config.initialize();
 
   if (settings.merged.theme) {
     if (!themeManager.setActiveTheme(settings.merged.theme)) {
@@ -133,12 +125,11 @@ export async function main() {
     }
   }
 
-  const memoryArgs = settings.merged.autoConfigureMaxOldSpaceSize
-    ? getNodeMemoryArgs(config)
-    : [];
-
   // hop into sandbox if we are outside and sandboxing is enabled
   if (!process.env.SANDBOX) {
+    const memoryArgs = settings.merged.autoConfigureMaxOldSpaceSize
+      ? getNodeMemoryArgs(config)
+      : [];
     const sandboxConfig = config.getSandbox();
     if (sandboxConfig) {
       if (settings.merged.selectedAuthType) {
--- a/packages/core/src/config/config.ts
+++ b/packages/core/src/config/config.ts
@@ -231,32 +231,30 @@ export class Config {
     }
   }
 
+  async initialize(): Promise<void> {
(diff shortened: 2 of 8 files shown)


packages/core/src/config/config.ts
```diff
@@ -231,32 +231,30 @@ export class Config {
...
-    this.contentGeneratorConfig = undefined!;
+  async initialize(): Promise<void> {
+    // Initialize centralized FileDiscoveryService
+    this.getFileService();
+    if (this.getCheckpointingEnabled()) {
+      try {
+        await this.getGitService();
+      } catch {
+        // For now swallow the error, later log it.
+      }
```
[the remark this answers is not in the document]
scidomino: This is preexisting code.

packages/cli/src/gemini.tsx
```diff
@@ -133,12 +125,11 @@ export async function main() {
     }
   }
 
-  const memoryArgs = settings.merged.autoConfigureMaxOldSpaceSize
```
NTaylorMullen, line 136: Why the move to only be done when sandboxing?

NTaylorMullen (approved):

packages/cli/src/gemini.tsx
scidomino: This variable is only used inside that if statement. It has no side-effects so I thought it would be clearer to move it to it's smallest necessary scope.

PR #3495, opened 2025-07-08
add new flag, --mcp-server-dev that ignores any mcp name restrictions imposed by --allowed-mcp-server-names

teeler:
## TLDR

Allows all mcp servers (an override) during the session. Intended for MCP server developers needing to work around allowlist restrictions.

## Dive Deeper

We're using a wrapper that hard-codes a list to --allowed-mcp-server-names, so this flag exists to allow users the ability to temporarily override that otherwise hard-coded setting. 

## Reviewer Test Plan

Not much I think, test coverage looks good.

## Testing Matrix

NA

## Linked issues / bugs

b/428217139 - bug discusses an override flag.

### DIFF
--- a/packages/cli/src/config/config.ts
+++ b/packages/cli/src/config/config.ts
@@ -53,6 +53,7 @@ interface CliArgs {
   allowedMcpServerNames: string[] | undefined;
   extensions: string[] | undefined;
   listExtensions: boolean | undefined;
+  mcpServerDev: boolean | undefined;
 }
 
 async function parseArguments(): Promise<CliArgs> {
@@ -155,6 +156,7 @@ async function parseArguments(): Promise<CliArgs> {
       type: 'array',
       string: true,
       description: 'Allowed MCP server names',
+      hidden: true,
     })
     .option('extensions', {
       alias: 'e',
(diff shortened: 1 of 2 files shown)


NTaylorMullen:
How is this different than having a local settings.json that overrides the allow names bits?

jackwotherspoon:
I am a fan of keeping things clean and not creating tons of flags unless absolutely necessary. 

This seems like an unnecessary flag to me... I don't think we should surface `dev` commands.

packages/cli/src/config/config.ts
```diff
@@ -155,6 +156,7 @@ async function parseArguments(): Promise<CliArgs> {
       type: 'array',
       string: true,
       description: 'Allowed MCP server names',
+      hidden: true,
```
teeler, line 159: @jackwotherspoon fyi

teeler:
(discussed elsewhere)

PTAL!

jackwotherspoon (dismissed):
LGTM ✅

smhendrickson:
I think removing `--allowed-mcp-server-names` entirely also achieves this, so maybe we don't need the extra flag?

teeler:
Yea i understand - we don't need this, ignore it!

PR #3498, opened 2025-07-08
Add system-wide settings config for administrators

chrstnb:
## TLDR

Adds a new type of settings that can be defined by system administrators to set organization-wide tools/settings/configuration for their users. This takes precedence over other settings.

## Reviewer Test Plan

Run the tests

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | y  | ❓  | ❓  |
| npx      | y  | ❓  | ❓  |
| Docker   | y | ❓  | ❓  |
| Podman   | y  | -   | -   |
| Seatbelt | y  | -   | -   |

## Linked issues / bugs

Fixes #3650

somewhat related to #3495


teeler:
Can we get these as flags also please? The default paths LGTM but it would be convenient to have these paths configurable via flags.

packages/cli/src/ui/App.test.tsx
```diff
@@ -197,7 +201,12 @@ describe('App UI', () => {
         ...settings,
       },
     };
-    return new LoadedSettings(userSettingsFile, workspaceSettingsFile, []);
+    return new LoadedSettings(
```
mattKorwel, line 206: I'm sorry i'm not following but how do these tests change with out corresponding code changes? I get we are updating the settings mocks, but if the mocks need updated, should there be code that exercises these? Either as new tests or new prod code?

mattKorwel (dismissed):
Left one non blocking comment

packages/cli/src/config/config.ts
```diff
@@ -53,6 +53,7 @@ interface CliArgs {
   allowedMcpServerNames: string | undefined;
   extensions: string[] | undefined;
   listExtensions: boolean | undefined;
+  'system-settings-path': string | undefined;
```
jackwotherspoon: nit: can we use camelCase for consistency here...

```suggestion
  systemSettingsPath: string | undefined;
```

packages/cli/src/config/config.ts
```diff
@@ -219,6 +224,10 @@ export async function loadCliConfig(
     argv.extensions || [],
   );
 
+  if (argv['system-settings-path']) {
+    setSystemSettingsPath(argv['system-settings-path']);
+  }
```
jackwotherspoon:
```suggestion
  if (argv.systemSettingsPath) {
    setSystemSettingsPath(argv.systemSettingsPath);
  }
```

jackwotherspoon (dismissed):

jackwotherspoon:
@NTaylorMullen @mattKorwel @teeler I am a little hesitant and wary of us adding too many top level CLI flags...

I think the top level flags should remain pretty minimal, otherwise the `--help` message will soon balloon to the point where it is too much for users to navigate and reduces its benefit.

I think we need a clear strategy for sub-commands. 

One of which would be `gemini config <command> [options]`. The config sub-command would allow setting everything related to `config` and would be the proper place for the `--system-settings-path` in my mind.

Happy to hear people's thoughts.

packages/cli/src/ui/App.test.tsx
chrstnb: Added some mocked out settings and threaded through. Also added a new test.

mattKorwel (dismissed):
Thanks for working through the details here. Looks good to me

packages/cli/src/config/settings.test.ts
```diff
@@ -409,7 +484,7 @@ describe('Settings Loading and Merging', () => {
       delete process.env.WORKSPACE_ENDPOINT;
     });
 
-    it('should prioritize workspace env variables over user env variables if keys clash after resolution', () => {
```
jacob314, line 412: keep a test verifying that workspace overrides user as well as adding a test that system overrides workspace.

packages/cli/src/config/settings.ts
```diff
@@ -105,6 +121,7 @@ export class LoadedSettings {
     return {
       ...this.user.settings,
       ...this.workspace.settings,
+      ...this.system.settings,
     };
```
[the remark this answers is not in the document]
jacob314: gemini code assist seems to be getting confused about the order overrides occur in.

jacob314:
Like the overall approach! 
Can you please also update 
ThemeDialog.tsx to reflect the fact that there are now more than two scopes when it shows messages warning that the theme is also set by a different scope. 
Please update that message to check all the other scopes.
```
let otherScopeModifiedMessage = '';
  const otherScope =
    selectedScope === SettingScope.User
      ? SettingScope.Workspace
      : SettingScope.User;
  if (settings.forScope(otherScope).settings.theme !== undefined) {
    otherScopeModifiedMessage =
      settings.forScope(selectedScope).settings.theme !== undefined
        ? `(Also modified in ${otherScope})`
        : `(Modified in ${otherScope})`;
```

mattKorwel (approved):

PR #3506, opened 2025-07-08
Fix infinite loop in start.js on Windows

SandyTao520:
## TLDR

This PR addresses an issue where the script would enter an infinite loop when attempting to locate .env files on Windows operating systems.

## Dive Deeper

The script uses a ``while (currentDir !== '/')`` loop to search for .env files in parent directories. On Windows, the root directory is not /; it's a drive letter like ``C:\``. This means the condition ``currentDir !== '/'`` will never be true, and the loop will run forever, causing the script to hang.

The loop is modified to correctly identify the root of the filesystem on both Windows and Unix-like systems. I can do this by checking if the parent directory is the same as the current directory.

## Reviewer Test Plan

In Windows PowerShell or Terminal,

```
npm run build
npm start
```

Verify Gemini Cli starts.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  |  ✅  | ❓  |
| npx      |  ✅  |  ✅  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt |  ✅  | -   | -   |


mattKorwel (approved):
🪟

PR #3507, opened 2025-07-08
Add support for specifying maxSessionTurns via the settings configuration

anj-s:
## TLDR
This pull request introduces a maxSessionTurns settings parameter to control the maximum number of conversation turns. This feature is designed to prevent infinite loops and provide better control over the execution flow,  especially in non-interactive mode. There is no default value and this feature is only when users specify the value of maxSessionTurns to be > 0.

## Dive Deeper

  The core of this change is the addition of the maxTurns property to the Config object in both @google/gemini-cli-core and @google/gemini-cli.
   - Configuration:
     - A new maxSessionsTurns settings option is added in packages/cli/src/config/config.ts.
     - The Config class in packages/core/src/config/config.ts now accepts and stores the maxSessionTurns value.
   - Non-Interactive Mode:
     - The runNonInteractive function in packages/cli/src/nonInteractiveCli.ts now includes a turn counter. It will exit gracefully if the number of turns exceeds the configured maxSessionTurns.
   - Interactive Mode (UI):
     - The GeminiClient class contains a turn counter that is incremented for every call to sendMessageStream
     - When the  limit is reached, a message is displayed to the user, and the application will start a new chat.

## Reviewer Test Plan

1. Non-Interactive Mode:
       - Run the CLI in non-interactive mode with a prompt that is likely to result in multiple turns (e.g., a multi-step task).
       - Use the maxSessionTurns setting with a low value (e.g., 2 or 3).
       - Example: gemini -p "list the files in the current directory and then tell me the size of package.json" --max-turns 2
       - Expected Behavior: The CLI should execute for a maximum of 2 turns and then exit.

 2. Interactive Mode:
     - Start the interactive CLI with the maxSessionTurns setting.json value set to a low value like 3.
     - Engage in a conversation with the model.
     - Expected Behavior: After the 3rd turn, the CLI should display a message indicating the turn limit has been reached and start a new chat

 3. Default Behavior:
     - Run the CLI in both interactive and non-interactive modes without the maxSessionsTurns setting.
     - Expected Behavior: The CLI should function as normal.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


packages/core/src/config/config.ts
```diff
@@ -223,6 +225,7 @@ export class Config {
     this.bugCommand = params.bugCommand;
     this.model = params.model;
     this.extensionContextFilePaths = params.extensionContextFilePaths ?? [];
+    this.maxTurns = params.maxTurns ?? 50;
```
anj-s: Open to suggestions of how to define a constant here.

packages/cli/src/config/config.ts
```diff
@@ -155,6 +157,11 @@ async function parseArguments(): Promise<CliArgs> {
       type: 'string',
       description: 'Allowed MCP server names',
     })
+    .option('max-turns', {
```
NTaylorMullen: Should this be a command line flag or a settings.json piece? I think I'd opt for the latter because it only impacts command line interactions that are `-p` or full run based. For instance, interacting with --version or extension commands don't make sense.

packages/cli/src/ui/App.tsx
NTaylorMullen, line 1: I wouldn't imagine you'd want to touch this file for controlling max turns. It doesn't have visibility into all of the underlying turns. See Allen's max turn changeset here: https://github.com/google-gemini/gemini-cli/pull/2793/files

packages/cli/src/config/config.ts
anj-s: Modified it to be part of the settings like we discussed.

packages/cli/src/ui/App.tsx
anj-s: Discussed offline and realized we have two different types of turns we are measuring. Modified the PR to count turns in Gemini Client in core/

mattKorwel (approved):
Looks good to me. I'm not super familiar with this area of the code but seems like a good step forward!

PR #3511, opened 2025-07-08
Fix nightly Release

mattKorwel:
## TLDR

I'm not 100% sure this fixes it but I think maybe we hit a race condition on the last broken nightly build.

This led to a successful test build.


mattKorwel:
the nightly build from utc 07/08 won't run. gemini thinks its one of these 2 issues.

NTaylorMullen (approved):

PR #3535, opened 2025-07-08
remove annoying deprecation warning due to version mistmatch

seuros:
This change resolves the following deprecation warning that appears during npm operations:
``` 
[DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
```
The warning is silenced by adding an  `overrides` field to the root  `package.json` to ensure a consistent version of
     `punycode` is used throughout the dependency tree.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,9 +10,6 @@
       "workspaces": [
         "packages/*"
       ],
-      "dependencies": {
-        "shell-quote": "^1.8.3"
-      },
       "bin": {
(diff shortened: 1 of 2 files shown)


NTaylorMullen:
Is this still needed given https://github.com/google-gemini/gemini-cli/pull/4296 ?

Assuming it's still needed, instead of overriding do we know what package is bringing in that bad version?

PR #3537, opened 2025-07-08
docs(contributing): mention macOS Seatbelt in GEMINI_SANDBOX examples

kdozlo:
## TLDR

Clarify in the “Enabling Sandboxing” example line that on macOS `GEMINI_SANDBOX=true` will also use Seatbelt, not just Docker or Podman.

## Dive Deeper

Clarifies that on macOS, setting GEMINI_SANDBOX=true will invoke Seatbelt rather than Docker or Podman, so newcomers don’t have to dig through the code to understand why Docker/Podman isn’t used. Although the full details live lower in the doc, a brief call-out reduces confusion.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -118,7 +118,7 @@ This command typically compiles TypeScript to JavaScript, bundles assets, and pr
 
 ### Enabling Sandboxing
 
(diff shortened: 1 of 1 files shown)


NTaylorMullen (approved):
❤️

PR #3300, opened 2025-07-05
fix(core): Sanitize tool parameters to fix 400 API errors

BigUncle:
> **Note**
> This PR depends on the test stabilization fix in **#3296**. Please review and merge that PR first. This PR includes the same test fix temporarily to pass CI checks, and it will be removed by rebasing onto `main` once #3296 is merged.

## TLDR

This pull request provides a robust fix for a persistent 400 API error caused by improper tool parameter schemas being sent to the Gemini API. It directly resolves issue #2141 and several duplicates by sanitizing tool function declarations before they are sent, ensuring compatibility with the API's requirements.

**Key Fixes:**

- **Sanitizes `format` property:** For `STRING` type parameters, it removes any `format` value that is not `enum` or `date-time`, which was the primary cause of the 400 errors.
- **Improves Tool Registry Robustness:** Adds `Array.isArray` checks to ensure `function_declarations` are correctly structured, which likely prevents the `missing field` errors also reported.

**This PR closes:** #2141, #2706, #3178, #2237
**This PR likely also resolves:** #2522

## Dive Deeper

The root cause of the `Persistent 400 API error` was that the tool discovery mechanism would generate function declarations with `format` properties (e.g., `uri`, `email`) that are not supported by the Gemini API for `STRING` parameters. The API strictly requires that the `format` for a string be either `enum` or `date-time`.

This PR introduces a `sanitizeParameters` function that recursively traverses the tool schema. It inspects every `STRING` parameter and sets its `format` to `undefined` if it's not one of the allowed values. This sanitization happens right before the tool is registered, ensuring that any request sent to the Gemini API contains a valid tool definition.

Additionally, issue #2141 and the related #2522 also reported errors like `...items: missing field.`. This often happens when an expected array structure is malformed. The secondary fix in this PR, which validates that `function_declarations` and `functionDeclarations` are indeed arrays before processing them, adds a layer of type safety that should prevent these malformed definitions from causing API errors.

## Reviewer Test Plan

1.  Pull down this branch.
2.  Run `npm run preflight`. All tests should pass consistently.
3.  To manually verify, one could set up a local tool discovery command that produces a tool with an unsupported string format (e.g., `"format": "uri"`).
    - **Before this PR:** Running a prompt that uses this tool would result in a 400 API error.
    - **After this PR:** The command should execute successfully without any API errors, as the unsupported format is sanitized before the request is made.

## Testing Matrix

|          | Local (WSL2) | CI  | Notes                                                            |
[...]

PR #3557, opened 2025-07-08
Fix version of shell-quote

scidomino:
## TLDR

Fixes shell-quote version.

## Reviewer Test Plan

I ran preflight.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,9 +10,6 @@
       "workspaces": [
         "packages/*"
       ],
-      "dependencies": {
-        "shell-quote": "^1.8.3"
-      },
       "bin": {
         "gemini": "bundle/gemini.js"
       },
@@ -11231,7 +11228,7 @@
         "open": "^10.1.2",
         "react": "^19.1.0",
         "read-package-up": "^11.0.0",
-        "shell-quote": "^1.8.2",
+        "shell-quote": "^1.8.3",
         "string-width": "^7.1.0",
         "strip-ansi": "^7.1.0",
         "strip-json-comments": "^3.1.1",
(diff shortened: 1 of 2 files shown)


chrstnb (approved):

mattKorwel:
What uses this package and why did it need updated now? It makes my brain happy to understand the provenance on these types of things to make sure we're not missing some larger issue.

swissspidy (approved):

scidomino:
> What uses this package and why did it need updated now? It makes my brain happy to understand the provenance on these types of things to make sure we're not missing some larger issue.

We've been using this dep for at least two weeks (in sandbox.ts). 3 days ago we updated our devDependencies and added the top level dependency in https://github.com/google-gemini/gemini-cli/pull/3300. I think that's the mistake. They should have run npm install before merging.

swissspidy:
> I think that's the mistake. They should have run npm install before merging.

That's an easy thing to catch in CI. Happy to provide a PR.

scottdensmore (approved):
Approved

BigUncle:
Hi all, thanks for the fix and the great discussion here. 

As the author of #3300, this is really helpful for me to understand the project's dependency conventions better. I really appreciate that you not only fixed the issue but also identified the gap in the CI checks. 

Great stuff!

PR #3568, opened 2025-07-08
Fix bad request in model check

SandyTao520:
## TLDR

Set thinking budget in model check request to 128 (minimum allowed).

## Dive Deeper

When ``thinkingBudget`` is set to 0, test requests consistently result in a "400 Bad Request" error. This indicates that the model's intended fallback to "flash" functionality never worked during auth.

According to https://ai.google.dev/gemini-api/docs/thinking#set-budget, it is not possible to disable thinking and the minimum budget is 128.

### Before
image

### After
image

## Reviewer Test Plan

Test with an free-tier API key that is already rate-limited. 

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ✅  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ✅  | -   | -   |

## Linked issues / bugs
N/A

### DIFF
--- a/packages/core/src/core/modelCheck.ts
+++ b/packages/core/src/core/modelCheck.ts
@@ -35,7 +35,7 @@ export async function getEffectiveModel(
       maxOutputTokens: 1,
       temperature: 0,
       topK: 1,
-      thinkingConfig: { thinkingBudget: 0, includeThoughts: false },
+      thinkingConfig: { thinkingBudget: 128, includeThoughts: false },
     },
   });
 


NTaylorMullen:
oh shit great catch @SandyTao520 !

NTaylorMullen (approved):

PR #3569, opened 2025-07-08
Working oauth client

emeryray2002:
## TLDR

This pull request introduces robust OAuth 2.0 authentication support for connecting to remote MCP servers. The primary goal is to enable Gemini CLI to securely connect to protected tool servers that require authentication, without needing developers to manually handle tokens.

The key change is the implementation of an automatic OAuth discovery and Device Grant authentication flow. When the CLI attempts to connect to a server and receives a `401 Unauthorized` error, it now automatically looks for an OAuth configuration, initiates a device auth flow (redirecting the local browser to the OAuth flow and opening a local port for the token callback), and then securely stores the obtained tokens for future use.

**For Reviewers, please focus on:**
- The new OAuth authentication logic in `packages/core/src/mcp/mcp-client-impl.ts`.
- The automatic discovery mechanism for OAuth endpoints.
- The updated connection logic that handles authenticated and unauthenticated connections for both HTTP and SSE transports.

## Dive Deeper

This change introduces several new components and modifications to the existing MCP client implementation.

### 1. OAuth Provider and Secure Token Storage

- **`MCPOAuthProvider`**: A new provider class is introduced in `mcp-client-impl.ts` to manage the OAuth lifecycle. It uses the "Device Grant" flow, which is well-suited for CLI applications. When authentication is required, it provides the user with a URL and a code to complete the authentication process in a browser.
- **`MCPOAuthTokenStorage`**: To avoid forcing users to re-authenticate on every session, this PR adds a token storage mechanism. 

### 2. Automatic Discovery and Authentication Flow

The most significant part of this change is the new automatic authentication handling.

1.  **Initial Connection Fails (401)**: When `connectAndDiscover` first attempts to connect to an MCP server and receives a `401 Unauthorized` response, the new logic kicks in.
2.  **`www-authenticate` Header**: The client inspects the `401` response for a `www-authenticate` header. This header is expected to contain the location of the OAuth authorization server's metadata.
3.  **OAuth Configuration Discovery**: If the header is found, the client makes a request to a well-known URI (`.well-known/oauth-authorization-server`) on the authorization server to fetch the necessary endpoints (e.g., `authorization_endpoint`, `token_endpoint`).
4.  **Device Auth Flow**: With the OAuth configuration, the `MCPOAuthProvider` starts the device grant flow, prompting the user to authorize the CLI.
5.  **Token Storage & Retry**: Upon successful authorization, the `MCPOAuthTokenStorage` saves the new tokens, and the client automatically retries the connection, this time with the `Authorization: Bearer <token>` header.

This flow makes the user experience much smoother, as they are guided through the authentication process only when needed.

### 3. Transport and Connection Logic

The `connectAndDiscover` function has been heavily modified to accommodate the auth flow:

- **Proactive Token Check**: Before making a connection, the client now checks if it has a stored, valid token for the server. If so, it attaches it to the request from the start.
- **HTTP vs. SSE with OAuth**: A key decision was how to handle OAuth with SSE (Server-Sent Events), which doesn't natively support custom headers in the initial browser-based `EventSource` API. The new logic attempts to connect via a standard `HTTP` transport first if an OAuth token is present. If that fails, it falls back to using the `SSEClientTransport` but includes the `Authorization` header in the initial request, which works for Node.js environments.
- **Single Server Re-discovery**: A new `discoverToolsForServer` method was added to the `ToolRegistry`. This is crucial for the post-authentication step, allowing the CLI to fetch the tools from the newly-authenticated server without re-discovering tools from all other configured servers.

### 4. Tool Schema Sanitization

A secondary improvement in this PR is a fix in `sanitizeParameters` (`tool-registry.ts`). The Gemini API requires that `enum` values in a tool's function declaration be strings. The previous implementation did not enforce this, causing issues with tools that used numeric enums. The function now correctly converts any `enum` values (e.g., numbers, booleans) to their string representation to ensure compatibility.

## Reviewer Test Plan

To validate these changes, a reviewer should ideally test the end-to-end OAuth flow. This requires an MCP tool server protected by OAuth 2.0 using the device grant flow.

If you don't have a test server, you can still review the code logic for correctness. However, to fully test:

1.  **Configure a Protected Server**: Set up an MCP server in your `~/.gemini/config.json` that requires OAuth and does not have any tokens stored for it yet.

    ```json
    "mcpServers": {
      "my_protected_server": {
        "httpUrl": "https://your-protected-server.com/tools"
      }
    }
    ```

2.  **Trigger the Auth Flow**: Run a command that would cause the CLI to connect to the server (e.g., `gemini-cli --tools-help`).

3.  **Observe the Output**:
    - You should see an initial connection failure.
    - The CLI should then print a message like: `To sign in, use a web browser to open the page <URL> and enter the code <CODE> to authenticate.`
    - Follow the instructions.

4.  **Verify Successful Connection**:
    - After you authenticate in the browser, the CLI should automatically resume.
    - It should successfully connect to the server and discover its tools.
    - The command you ran (`gemini-cli --tools-help`) should now complete successfully, showing tools from `my_protected_server`.

5.  **Test Token Persistence**:
    - Quit the CLI and start it again.
    - Run the same command (`gemini-cli --tools-help`).
    - This time, it should connect directly without asking you to authenticate, as it should use the token stored in the keychain.

6.  **Test Enum Sanitization (Optional)**:
    - If you have control over the tool server, define a tool with a numeric enum:
      ```json
      "parameters": {
        "type": "object",
        "properties": {
          "status": {
            "type": "integer",
            "description": "The status code.",
            "enum": [0, 1, 2]
          }
        }
      }
      ```
    - Verify that the CLI can still discover and use this tool correctly, and that `gemini-cli --tools-help` shows the enum values as strings (`"0"`, `"1"`, `"2"`). 
## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


NTaylorMullen:
@emeryray2002 would you mind sharing some information around precedent / how other clients handle this as well as part of the PR?

packages/core/src/tools/mcp-client.ts
```diff
@@ -179,18 +379,134 @@ async function connectAndDiscover(
...
+        );
+        updateMCPServerStatus(mcpServerName, MCPServerStatus.DISCONNECTED);
+        return;
+      }
+    } else {
+      // Check if we have stored OAuth tokens for this server (from previous authentication)
+      accessToken = await MCPOAuthProvider.getValidToken(mcpServerName, {
+        authorizationUrl: '', // Will be discovered automatically
+        tokenUrl: '', // Will be discovered automatically
+      });
```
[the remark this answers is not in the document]
emeryray2002: resolved

emeryray2002:
See https://screencast.googleplex.com/cast/NjM2NzY0NTEzNjg0Njg0OHxkNmI5ZTAyYS0yNg

packages/cli/src/ui/hooks/slashCommandProcessor.ts
```diff
@@ -215,6 +217,103 @@ export const useSlashCommandProcessor = (
...
+
+      try {
+        addMessage({
+          type: MessageType.INFO,
+          content: `Starting OAuth authentication for MCP server '${serverName}'...`,
+          timestamp: new Date(),
+        });
+
+        // Import dynamically to avoid circular dependencies
+        const { MCPOAuthProvider } = await import('@google/gemini-cli-core');
```
NTaylorMullen: Core doesn't depend on the CLI (it's the opposite). This shouldn't be needed

packages/cli/src/ui/hooks/slashCommandProcessor.ts
```diff
@@ -369,34 +475,54 @@ export const useSlashCommandProcessor = (
...
-
             // Format server header with bold formatting and status
-            message += `${statusIndicator} \u001b[1m${serverName}\u001b[0m - ${statusText}`;
+            message += `${statusIndicator} \u001b[1m${serverName}\u001b[0m - ${detailedStatus}`;
+
+            let needsAuthHint = mcpServerRequiresOAuth.get(serverName) || false;
+            // Add OAuth status if applicable
+            if (serverConfig.oauth?.enabled) {
+              needsAuthHint = true;
+              const { MCPOAuthTokenStorage } = await import(
```
NTaylorMullen: ditto

packages/core/src/config/config.ts
```diff
@@ -44,6 +44,10 @@ import {
   DEFAULT_GEMINI_FLASH_MODEL,
 } from './models.js';
 import { ClearcutLogger } from '../telemetry/clearcut-logger/clearcut-logger.js';
+import { MCPOAuthConfig } from '../mcp/oauth-provider.js';
+
+// Re-export OAuth config type
```
NTaylorMullen, line 50: Could you go through this PR and remove the AI comments

packages/core/src/core/client.ts
```diff
@@ -140,6 +140,13 @@ export class GeminiClient {
     this.getChat().setHistory(history);
   }
 
+  async setTools(): Promise<void> {
```
NTaylorMullen, line 155: Is this more of a refresh tools?

docs/tools/mcp-server.md
```diff
@@ -93,6 +93,90 @@ Each server configuration supports the following properties:
...
+The Gemini CLI supports OAuth 2.0 authentication for remote MCP servers using SSE or HTTP transports. This enables secure access to MCP servers that require authentication.
+
+#### Automatic OAuth Discovery
+
+For servers that support OAuth discovery, you can omit the OAuth configuration and let the CLI discover it automatically:
+
+```json
+{
+  "mcpServers": {
+    "discoveredServer": {
```
NTaylorMullen, line 107:
```suggestion
    "fooServer": {
```

packages/core/src/mcp/oauth-discovery.ts
```diff
@@ -0,0 +1,160 @@
...
+        if (authServerResponse.ok) {
+          const authServerMetadata =
+            (await authServerResponse.json()) as OAuthAuthorizationServerMetadata;
+          const oauthConfig: MCPOAuthConfig = {
+            authorizationUrl: authServerMetadata.authorization_endpoint,
+            tokenUrl: authServerMetadata.token_endpoint,
+            scopes: authServerMetadata.scopes_supported || [],
+          };
+          if (authServerMetadata.registration_endpoint) {
+            console.log(
```
NTaylorMullen: Should this be log or debug?

packages/core/src/mcp/oauth-discovery.ts
```diff
@@ -0,0 +1,160 @@
...
+    if (authzResponse.ok) {
+      const authServerMetadata =
+        (await authzResponse.json()) as OAuthAuthorizationServerMetadata;
+      const oauthConfig: MCPOAuthConfig = {
+        authorizationUrl: authServerMetadata.authorization_endpoint,
+        tokenUrl: authServerMetadata.token_endpoint,
+        scopes: authServerMetadata.scopes_supported || [],
+      };
+      if (authServerMetadata.registration_endpoint) {
+        console.log(
```
NTaylorMullen: ditto

packages/core/src/mcp/oauth-provider.ts
```diff
@@ -0,0 +1,789 @@
...
+  private static async discoverOAuthFromMCPServer(
+    mcpServerUrl: string,
+  ): Promise<MCPOAuthConfig | null> {
+    try {
+      // Extract the base URL from the MCP server URL
+      const serverUrl = new URL(mcpServerUrl);
+      const baseUrl = `${serverUrl.protocol}//${serverUrl.host}`;
+
+      // Try to get the protected resource metadata first
+      const resourceMetadataUrl = new URL(
```
NTaylorMullen: You did a good portion of this also in the oauth-discovery.ts. Would it make sense to centralize some of this logic?

packages/core/src/mcp/oauth-provider.ts
```diff
@@ -0,0 +1,789 @@
...
+    return { codeVerifier, codeChallenge, state };
+  }
+
+  /**
+   * Start a local HTTP server to handle OAuth callback.
+   *
+   * @param expectedState The state parameter to validate
+   * @returns Promise that resolves with the authorization code
+   */
+  private static async startCallbackServer(
```
NTaylorMullen, line 178: Quite epic

packages/core/src/mcp/oauth-provider.ts
```diff
@@ -0,0 +1,789 @@
...
+
+    // Save token
+    try {
+      await MCPOAuthTokenStorage.saveToken(
+        serverName,
+        token,
+        config.clientId,
+        config.tokenUrl,
+      );
+      console.log('Authentication successful! Token saved.');
```
NTaylorMullen, line 609: May want to consider debug calls for a lot of these. I'll stop leaving the comment of log -> debug on the PR though

packages/core/src/mcp/oauth-token-storage.ts
```diff
@@ -0,0 +1,205 @@
...
+  private static readonly CONFIG_DIR = '.gemini';
+
+  /**
+   * Get the path to the token storage file.
+   *
+   * @returns The full path to the token storage file
+   */
+  private static getTokenFilePath(): string {
+    const homeDir = os.homedir();
+    return path.join(homeDir, this.CONFIG_DIR, this.TOKEN_FILE);
```
NTaylorMullen, line 48: Should this be in the tmp directory? i.e. `~/.gemini/tmp/...`

packages/core/src/tools/tool-registry.ts
```diff
@@ -378,6 +403,19 @@ function _sanitizeParameters(schema: Schema | undefined, visited: Set<Schema>) {
       }
     }
   }
+
+  // Handle enum values - Gemini API only allows enum for STRING type
```
NTaylorMullen, line 409: Hmm we didn't support this earlier? Typically the genai SDK will do the translation for this

NTaylorMullen (commented):
This is a fantastic and much-needed feature! The implementation looks solid. To make it a bit easier for the team to review thoroughly and merge safely, would you be open to splitting this into a few smaller, stacked PRs? I was thinking we could break it down like this:

1. Core OAuth infrastructure (the provider, token storage, etc.)
2. Integration with the MCP client
3. The new CLI commands and UI changes

The schema sanitization fix could also be a small, separate PR that could be merged very quickly.

I left some higher level comments BUT struggled to dig in nearly enough to have confidence here. Anyhow, let me know what you think!

packages/core/src/mcp/oauth-provider.ts
```diff
@@ -0,0 +1,698 @@
...
+  codeVerifier: string;
+  codeChallenge: string;
+  state: string;
+}
+
+/**
+ * Provider for handling OAuth authentication for MCP servers.
+ */
+export class MCPOAuthProvider {
+  private static readonly REDIRECT_PORT = 7777;
```
cornmander, line 90: If the port is static, what happens if multiple gemini cli instances are running? Or any other reason why the port may be in-use. I believe if you leave the port un-specified node will automatically choose a random port, which would not run into this problem:

https://nodejs.org/api/net.html#serverlistenport-host-backlog-callback

packages/core/src/mcp/oauth-token-storage.ts
cornmander: Agree with this -- it would be good to put them in a child directory set to 600. At least on my machine ~/.gemini is world-readable and world-executable.

packages/core/src/tools/mcp-client.ts
```diff
@@ -129,6 +138,165 @@ export function getMCPDiscoveryState(): MCPDiscoveryState {
...
+ * Extract WWW-Authenticate header from error message string.
+ * This is a more robust approach than regex matching.
+ *
+ * @param errorString The error message string
+ * @returns The www-authenticate header value if found, null otherwise
+ */
+function extractWWWAuthenticateHeader(errorString: string): string | null {
+  // Try multiple patterns to extract the header
+  const patterns = [
+    /www-authenticate:\s*([^\n\r]+)/i,
```
cornmander, line 165: These first two patterns seem to be redundant since they only differ in casing and it's a case insensitive match.

packages/core/src/tools/mcp-client.ts
```diff
@@ -347,40 +517,396 @@ export async function connectToMcpServer(
...
+    }
+  }
+
   if (mcpServerConfig.httpUrl) {
     const transportOptions: StreamableHTTPClientTransportOptions = {};
-    if (mcpServerConfig.headers) {
+
+    // Set up headers with OAuth token if available
+    if (hasOAuthConfig && accessToken) {
+      transportOptions.requestInit = {
```
cornmander, line 898: Why not set the authProvider transport option?

https://github.com/modelcontextprotocol/typescript-sdk/blob/16ea27783cc534f629ad548ccaa8a1af16832763/src/client/sse.ts#L34

This seems like it will take care of getting new access tokens as needed.

packages/core/src/mcp/oauth-provider.ts
cornmander, line 1: This looks great but is it redundant with https://github.com/modelcontextprotocol/typescript-sdk/blob/16ea27783cc534f629ad548ccaa8a1af16832763/src/client/auth.ts ?

emeryray2002:
merged in another PR

DinoChiesa:
which PR? can you cite it pls? 
I think probably #4316 #4318 #4319

emeryray2002:
https://github.com/google-gemini/gemini-cli/pull/4319 included all the changes

PR #3611, opened 2025-07-09
fix: /help remove flickering and respect clear shortcut (ctr+l)

### LINKED ISSUE #3224: cli jumping each keystroke after /help request
### What happened?

The cli interface is jumping on every keystroke (flickers, twitches) after the /help request until the next request is executed or fullscreen mode is enabled

stable behavior w/o reactions on keystrokes

### Client information

*   **CLI Version:** 0.1.9
*   **Git Commit:** 34935d6
*   **Operating System:** win32 v24.3.0
*   **Memory Usage:** 142.1 MB

### Login information

Google account, Workspace.

_No response_

psinha40898:
## TLDR
Currently, from a UX perspective the keyboard shortcut to clear does not clear the output from the help command.

## Dive Deeper
Help is changed from its own component in App.tsx outside of Static to a regular history item like /about

### Clear
Before, the help dialog was its own React component in App.tsx, and the state it was attached to was not incorporated in the clear shortcut.

Now, help is rendered just like about -- into the HistoryItemDisplay under Static.

All those items are already respecting ctrl + l, so now help does too.

### Flickering

For components rendered outside of Static in Ink:
From Ink:
```
174		if (outputHeight >= this.options.stdout.rows) {
			this.options.stdout.write(
				ansiEscapes.clearTerminal + this.fullStaticOutput + output,
			);
			this.lastOutput = output;
			return;
		}
```
`outputHeight` represents only the dynamic (non-static) content,

reference: https://github.com/vadimdemedes/ink
https://github.com/vadimdemedes/ink/blob/a31d76d0/src/ink.tsx

If the `outputHeight ` (the height of the rendered content) is greater than or equal to `this.options.stdout.rows` (the number of rows in the terminal), Ink will clear the entire terminal using `ansiEscapes.clearTerminal` before writing the new output. ink.tsx:173-176 

This full clear and redraw is what causes the visible flickering, especially for larger components like Help that occupy a significant portion of the terminal height. 

## Reviewer Test Plan

Pull the code
Run the branch
Type /help
Use the command control+L to clear the screen
It should set the state to false and provide the intended UX

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ✅  | ✅  |
| npx      | ✅  | ✅  | ✅  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes #3453 
Fixes #3224


packages/cli/src/ui/components/InputPrompt.tsx
```diff
@@ -208,6 +210,7 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
 
       if (key.ctrl && key.name === 'l') {
         onClearScreen();
+        setShowHelp(false);
         return;
       }
```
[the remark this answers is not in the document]
psinha40898: `onClearScreen()` clears the rendered history. The help command is not within the rendered history and exists within its own React State in `App.tsx`. This is beneficial because it would be wasteful to consistently add the help command's results to the CLI's history.

jacob314:
Sorry for the slow review. I'd suggest we resolve this by fixing how help is rendered. help should just render is a message rather than its own ui. that way it can just be included in the history and will clear normally. That will also resolve flickering in the history.

psinha40898:
> Sorry for the slow review. I'd suggest we resolve this by fixing how help is rendered. help should just render is a message rather than its own ui. that way it can just be included in the history and will clear normally. That will also resolve flickering in the history.

Sorry I totally accidentally commented on the issue instead of replying here. Deleted that 

If /help should render in the Static HistoryItemDisplay just like /about (as a message) I pushed those changes and will resolve conflicts when home. Ty for checking

jacob314:
> > Sorry for the slow review. I'd suggest we resolve this by fixing how help is rendered. help should just render is a message rather than its own ui. that way it can just be included in the history and will clear normally. That will also resolve flickering in the history.
> 
> Sorry I totally accidentally commented on the issue instead of replying here. Deleted that
> 
> If /help should render in the Static HistoryItemDisplay just like /about (as a message) I pushed those changes and will resolve conflicts when home. Ty for checking

Fantastic! It looks great now. I will approve as soon as conflicts are resolved.

psinha40898:
> > > Sorry for the slow review. I'd suggest we resolve this by fixing how help is rendered. help should just render is a message rather than its own ui. that way it can just be included in the history and will clear normally. That will also resolve flickering in the history.
> > 
> > 
> > Sorry I totally accidentally commented on the issue instead of replying here. Deleted that
> > If /help should render in the Static HistoryItemDisplay just like /about (as a message) I pushed those changes and will resolve conflicts when home. Ty for checking
> 
> Fantastic! It looks great now. I will approve as soon as conflicts are resolved.

Resolved. As always, thank you so much for the review!

psinha40898:
@jacob314 It should be good now. 

I had to push a change (automerge off) b.c of new tests in main that relied on the old help -- my mistake to not update the branch earlier.

Thank you and sorry about that

psinha40898:
Can I get some clarification if we are going with this PR or #5196 instead, since they overlap? 
This PR was meant to be atomic as possible and opened 3 weeks ago if that matters
I hope this friendly tag for the reviewer of that PR is okay

Thank you
@allenhutchison

packages/cli/src/ui/commands/helpCommand.ts
```diff
@@ -4,18 +4,19 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { CommandKind, OpenDialogActionReturn, SlashCommand } from './types.js';
+import { CommandKind, SlashCommand } from './types.js';
+import { MessageType, type HistoryItemHelp } from '../types.js';
 
 export const helpCommand: SlashCommand = {
   name: 'help',
-  altNames: ['?'],
```
jacob314, line 11: please add back altNames and then I'll approve

jacob314 (commented):
Thanks for the fix! Everything looks good except for altNames being missing.
You can repro by testing whether typing `/? works.

Like how small and simple this change ended up.

packages/cli/src/ui/commands/helpCommand.ts
psinha40898: Thank you!

jacob314 (approved):
lgtm

PR #3620, opened 2025-07-09
fix: Use Email for Clearcut Logging and Refactor User Info Fetching

### LINKED ISSUE #3149: Google ID is incorrectly being sent to Clearcut
### What happened?

This is fixed (hopefully) in https://github.com/google-gemini/gemini-cli/pull/3147

Previous implementation was setting the wrong field into Clearcut. It turns out that a UserInfo object needs to be populated:

The UserInfo message within input proto is the standard place to store user-related identifiers.
The UserID field within UserInfo is specifically designated for the user's Gaia ID.
This should correctly result in Google user IDs being passed in for logging purposes. They will then be obfuscated before storing to clear all personally identifiable information.

### Client information

<details>

``` CLI Version              0.1.9                                           │
│ Git Commit               a7d8bb2a (local modifications)                  │
│ Model                    gemini-2.5-pro                                  │
│ Sandbox                  no sandbox                                      │
│ OS                       darwin                                          │
│ Auth Method              OAuth                                           │
│ GCP Project              crescendo-ai-test  
```

</details>

### Login information

_No response_

_No response_

gsquared94:
This PR addresses an issue with our Clearcut logging by replacing the previously used Google Account ID with the user's email. This change aligns with standard practices (see http://go/cloudmill-1p-oss-instrumentation#define-sessionable-id) which should also work for clearcut to automatically delete the email field and populate the obscured_gaia_id.

### Key Changes:

*   **Updated Clearcut Logging:**
    *   The `ClearcutLogger` now sends the user's email (`client_email`) instead of the installation ID or a numeric Google Account ID.
    *   This ensures that logs are associated with a recognizable user identity when available.
    *   Logging falls back to `client_install_id` if the email is not available.

*   **Refactored User Information Fetching:**
    *   The logic for retrieving user information has been updated to use the `googleapis.com/oauth2/v2/userinfo` endpoint.
    *   This replaces the previous method of decoding the `id_token` to get the `sub` claim.
    *   The new approach fetches both the user's email and their Google Account ID, caching them for future use.

*   **Cached Credentials:**
    *   The user's email is now cached locally, similar to how the Google Account ID was previously handled.
    *   The `getGoogleAccountEmail()` utility function provides a consistent way to access the cached email.

*   **Error Handling and Dependencies:**
    *   Removed `id_token` verification logic that is no longer needed.
    *   Simplified promise handling and error logging around the flushing of telemetry data.
    *   Switched to using `require` for dynamic imports to avoid circular dependency issues.

Fixes #3149

### DIFF
--- a/packages/core/src/code_assist/oauth2.ts
+++ b/packages/core/src/code_assist/oauth2.ts
@@ -44,6 +44,7 @@ const SIGN_IN_FAILURE_URL =
 const GEMINI_DIR = '.gemini';
 const CREDENTIAL_FILENAME = 'oauth_creds.json';
 const GOOGLE_ACCOUNT_ID_FILENAME = 'google_account_id';
+const GOOGLE_ACCOUNT_EMAIL_FILENAME = 'google_account_email';
 
 /**
  * An Authentication URL for updating the credentials of a Oauth2Client
@@ -70,13 +71,10 @@ export async function getOauthClient(
   // If there are cached creds on disk, they always take precedence
   if (await loadCachedCredentials(client)) {
     // Found valid cached credentials.
-    // Check if we need to retrieve Google Account ID
-    if (!getCachedGoogleAccountId()) {
+    // Check if we need to retrieve Google Account ID or Email
+    if (!getCachedGoogleAccountId() || !getCachedGoogleAccountEmail()) {
       try {
-        const googleAccountId = await getRawGoogleAccountId(client);
-        if (googleAccountId) {
-          await cacheGoogleAccountId(googleAccountId);
-        }
+        await fetchAndCacheUserInfo(client);
       } catch {
         // Non-fatal, continue with existing auth.
       }
@@ -163,10 +161,7 @@ async function authWithWeb(client: OAuth2Client): Promise<OauthWebLogin> {
           client.setCredentials(tokens);
           // Retrieve and cache Google Account ID during authentication
           try {
-            const googleAccountId = await getRawGoogleAccountId(client);
-            if (googleAccountId) {
-              await cacheGoogleAccountId(googleAccountId);
-            }
+            await fetchAndCacheUserInfo(client);
           } catch (error) {
             console.error(
               'Failed to retrieve Google Account ID during authentication:',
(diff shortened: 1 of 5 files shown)


packages/core/src/code_assist/oauth2.ts
```diff
@@ -275,57 +270,73 @@ export function getCachedGoogleAccountId(): string | null {
   }
 }
 
+function getGoogleAccountEmailCachePath(): string {
+  return path.join(os.homedir(), GEMINI_DIR, GOOGLE_ACCOUNT_EMAIL_FILENAME);
```
NTaylorMullen, line 274: Should this go in the tmp dir there?

packages/core/src/telemetry/clearcut-logger/clearcut-logger.ts
```diff
@@ -66,47 +68,46 @@ export class ClearcutLogger {
...
 
-    // Fire and forget - don't await
-    this.flushToClearcut().catch((error) => {
-      console.debug('Error flushing to Clearcut:', error);
-    });
+    this.flushToClearcut();
   }
 
-  async flushToClearcut(): Promise<LogResponse> {
+  flushToClearcut(): Promise<LogResponse> {
```
NTaylorMullen, line 100: Should this still return a promise?

packages/core/src/telemetry/clearcut-logger/clearcut-logger.ts
```diff
@@ -268,9 +267,7 @@ export class ClearcutLogger {
     ];
 
     this.enqueueLogEvent(this.createLogEvent(new_prompt_event_name, data));
-    this.flushToClearcut().catch((error) => {
-      console.debug('Error flushing to Clearcut:', error);
```
NTaylorMullen: Should we be try catching more to protect log failures?

packages/core/src/utils/user_id.ts
NTaylorMullen, line 1: Just saw this file name / the test ones. THey're very pythonic lol, we should change. Doesn't need to be this PR

NTaylorMullen (dismissed):

owenofbrien:
@gsquared94 - I think we should log both the client_install_id and the client_email. We need to be able to look for cases where multiple client_emails are associated with the same client_install_id.

NTaylorMullen:
> @gsquared94 - I think we should log both the client_install_id and the client_email. We need to be able to look for cases where multiple client_emails are associated with the same client_install_id.

@gsquared94 please address ^ as well

gsquared94:
> > @gsquared94 - I think we should log both the client_install_id and the client_email. We need to be able to look for cases where multiple client_emails are associated with the same client_install_id.
> 
> @gsquared94 please address ^ as well

We cannot log both client_email and client_install_id for the same entry to concord. It's a violation of Google's wipeout policy compliance.

If the objective is to find multiple accounts against the same install ID, can we instead keep a record of all old values of client_email in the settings across multiple sign-ins, and log the duplicate count directly?

packages/core/src/telemetry/clearcut-logger/clearcut-logger.ts
gsquared94: Sorry, sending a PR on Github after many years. I accidentally resolved comments without posting my messages. 

This code is a revert from a previous commit. So I'll just keep it as-is for now. Will try to figure out a better way of doing this after familiarizing myself with the project.

packages/core/src/code_assist/oauth2.ts
gsquared94: I think that the tmp directory manages settings that a different for different projects, since the immediate subdirectories seem to be project-name-hashes(?). Since auth is being maintained globally this setting would need to live somewhere outside tmp. Keeping as-is for now.

packages/core/src/utils/user_id.ts
gsquared94: Ack

packages/core/src/telemetry/clearcut-logger/clearcut-logger.ts
gsquared94: Done.

mattKorwel (approved):

PR #3634, opened 2025-07-09
feat: Add Ctrl+n/p for suggestion navigation

DanisAlfonso:
This change adds support for navigating command suggestions using Ctrl+n (next) and Ctrl+p (previous) in addition to the arrow keys.

This improves the user experience for developers who are accustomed to these common keybindings from other CLI tools.


SandyTao520:
Fixed in #3641, thanks!

PR #3679, opened 2025-07-09
docs: Add missing API key to docker run command

reconsumeralization:


NTaylorMullen:
Hi @reconsumeralization thank you for the contribution! i think for now we're trying to leave our README sections a bit more terse. It's definitely challenging given how many different ways there are to auth with Gemini CLI 😄

reconsumeralization:
Hi, @NTaylorMullen , thanks for reviewing the PR!
Totally understand wanting to keep the README concise.
I’ve got some higher-priority work in flight right now, so I’ll circle back when I have bandwidth.
Appreciate the feedback and all the work you’re doing on Gemini CLI! 😊

PR #3680, opened 2025-07-09
Check for zeditor if zed binary is not found

### LINKED ISSUE #2346: Zed
### What happened?

Config didnt find Zed because there was nod "zed" in path. Some Linux distributions (at least arch) use a "zeditor" binary to launch the editor while the "zed" binary is under /usr/lib/zed/zed-editor
Workaround: ln -s /usr/lib/zed/zed-editor $SOME_DIR_IN_PATH/zed

recognize the editor as installed or provide an option to manually enter the path to the binary in the settings.json 

### Client information

*   **CLI Version:** 0.1.8
*   **Git Commit:** ad7839ea
*   **Operating System:** linux v22.14.0
*   **Memory Usage:** 177.0 MB

### Login information

_No response_

_No response_

cole-miller:
## TLDR

This PR adds a fallback to use the `zeditor` command when the user's preferred editor is `'zed'` and the `zed` binary is not found, which fixes using Zed with gemini-cli on Arch Linux and other distributions that package the Zed CLI as `zeditor`. 

## Dive Deeper

I've set this up to prefer `zed` over `zeditor` if both exist, since that's the canonical name for the CLI binary.

## Reviewer Test Plan

There are some new tests in editor.test.ts for the fallback logic.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Closes #2346


packages/core/src/utils/editor.ts
```diff
@@ -58,7 +58,14 @@ export function checkHasEditorType(editor: EditorType): boolean {
   const commandConfig = editorCommands[editor];
   const command =
     process.platform === 'win32' ? commandConfig.win32 : commandConfig.default;
-  return commandExists(command);
+
+  const primaryExists = commandExists(command);
```
jacob314: rather than special casing zed, can you extend the general code to accept multiple names. doubt this is the last editor that will require multiple names

jacob314 (commented):
approved once the hopefully minor comment to fix is addressed. thanks for the pull request!

packages/core/src/utils/editor.ts
cole-miller: Sounds good, will implement that and have this ready shortly!

cole-miller: Should be implemented now! I realized it made more sense to put `zeditor` first, since Arch and other distros are using that name specifically to avoid a collision on `zed`, and it means that if neither `zeditor` nor `zed` is installed we'll return `zed $ARGS` as the diff command, which should present a more helpful error.

packages/core/src/utils/editor.ts
```diff
@@ -92,13 +99,18 @@ export function getDiffCommand(
...
+  const command =
+    commands.slice(0, -1).find((cmd) => commandExists(cmd)) ||
+    commands[commands.length - 1];
+
   switch (editor) {
     case 'vscode':
     case 'vscodium':
     case 'windsurf':
     case 'cursor':
+      return { command, args: ['--wait', '--diff', oldPath, newPath] };
```
NTaylorMullen: Can drop I believe?

NTaylorMullen (changes requested):
Nice change and thank you for the contribution! Just one slight update.

packages/core/src/utils/editor.ts
```diff
@@ -44,21 +44,28 @@ function commandExists(cmd: string): boolean {
...
+  EditorType,
+  { win32: string[]; default: string[] }
+> = {
+  vscode: { win32: ['code.cmd'], default: ['code'] },
+  vscodium: { win32: ['codium.cmd'], default: ['codium'] },
+  windsurf: { win32: ['windsurf'], default: ['windsurf'] },
+  cursor: { win32: ['cursor'], default: ['cursor'] },
+  vim: { win32: ['vim'], default: ['vim'] },
+  neovim: { win32: ['nvim'], default: ['nvim'] },
+  zed: { win32: ['zed'], default: ['zeditor', 'zed'] },
```
NTaylorMullen: Can we also swap the order of zeditor and zed to prefer zed if both are available

cole-miller: Done.

packages/core/src/utils/editor.ts
cole-miller: Good catch, done!

NTaylorMullen (approved):

NTaylorMullen:
Thank you for the contribution @cole-miller !!

Issue #3657, opened 2025-07-09
FR: Sub-command for Gemini CLI config (`gemini config <command> [options]`)


jackwotherspoon:
### What would you like to be added?

Right now configuration of Gemini CLI is done via setting env vars or configuring files such as `.gemini/settings.json` etc.

For full control via CLI we should add a sub-command for control over Gemini CLI config.

Command: `gemini config <command> [option]`

### Why is this needed?

To allow a user to update config via the CLI and not need to set env vars or config files manually.

Much more discoverable via the `--help` command then having to search through docs for env vars etc..

jackwotherspoon:
cc @swissspidy created issue for scoping out the addition of `gemini config` sub-commands

jacob314:
This has been on my mind as well. Right now we have `/theme` for setting the current theme and `/auth` for setting auth but we lack ui for all the other settings and config options.

I believe we should add a `/config` command as well that provides an entry point to modify all the config and settings. The UI could be optimized to show all of the config settings and config on a single screen with ability to switch to the specific UI for cases like auth and themes that can benefit from custom UI.
This is UI perhaps best supported by a GUI tool so we might want to also consider being able to surface the UI in a simple electron app launched from GEMINI CLI to reduce the friction editing the configuration.

PR #3681, opened 2025-07-09
feat/config command

reconsumeralization:
- **docs: Add missing API key to docker run command**
- **feat: Add config command to manage settings**

### DIFF
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -37,8 +37,8 @@ For security and isolation, Gemini CLI can be run inside a container. This is th
 - **Directly from the Registry:**
   You can run the published sandbox image directly. This is useful for environments where you only have Docker and want to run the CLI.
   ```bash
-  # Run the published sandbox image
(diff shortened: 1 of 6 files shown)


jackwotherspoon:
@reconsumeralization Mind discussing the approach you are taking here and why on https://github.com/google-gemini/gemini-cli/issues/3657 so that we can discuss and make sure this covers all use-cases 😄 

nit: this should be split into two PRs (one for docs and one for the feat) for clear and clean history. For example if we wanted to revert the config command in the future, we couldn't just revert this PR as is because it also contains a docs update.

jackwotherspoon:
> nit: this should be split into two PRs (one for docs and one for the feat) for clear and clean history. For example if we wanted to revert the config command in the future, we couldn't just revert this PR as is because it also contains a docs update.

Looks like you did this and have the docs change here: https://github.com/google-gemini/gemini-cli/pull/3679

PR #3701, opened 2025-07-09
Add visual cues for nightly version

### LINKED ISSUE #3883: Nightly version should have indicator

When running the nightly version of the CLI, we should have some indicator so user's know they are running nightly (similar to VS Code's "Insiders" version).

### Why is this needed?

There is no distinction when running stable vs nightly unless you run /about

### Additional context

_No response_

miguelsolorio:
## TLDR

When running the nightly version, we'll add some minor visual cues (add version below logo, add gradient to cwd) to help give some indicator of being on the nightly version.

## Dive Deeper

**Regular version**
CleanShot 2025-07-09 at 13 13 12@2x

**Nightly version**
CleanShot 2025-07-09 at 13 21 57@2x

## Reviewer Test Plan

1. Manually change the cli version under `packages/cli/package.json`
2. Build source
3. Run local build
4. Ensure nightly version is displayed in the header & gradient is applied to cwd in footer

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes #3883


packages/cli/src/ui/components/Header.tsx
```diff
@@ -14,11 +14,13 @@ import { getAsciiArtWidth } from '../utils/textUtils.js';
 interface HeaderProps {
   customAsciiArt?: string; // For user-defined ASCII art
   terminalWidth: number; // For responsive logo
+  version: string;
 }
 
 export const Header: React.FC<HeaderProps> = ({
   customAsciiArt,
   terminalWidth,
+  version,
```
NTaylorMullen, line 24: Instead of passing version here. Could we pass a boolean for nightly or not?

packages/cli/src/ui/components/Footer.tsx
```diff
@@ -36,17 +38,28 @@ export const Footer: React.FC<FooterProps> = ({
   showErrorDetails,
   showMemoryUsage,
   promptTokenCount,
+  version,
```
NTaylorMullen: Ditto

packages/cli/src/ui/App.tsx
```diff
@@ -83,6 +83,7 @@ interface AppProps {
   config: Config;
   settings: LoadedSettings;
   startupWarnings?: string[];
+  version: string;
```
NTaylorMullen, line 87: ditto

NTaylorMullen (commented):
Let me know if you'd like some help applying some of the suggestions. Otherwise looks good. Love the look n feel

packages/cli/src/ui/components/Header.tsx
miguelsolorio: I did try this, but it seems we still need `version` to be displayed in the header so version is still needed? I also tried retrieving the version within the header via `getCliVersion` but unable to since it's an async component and the header

jacob314 (approved):
lgtm

NTaylorMullen (approved):

Issue #1956, opened 2025-06-26
Add headless/CLI authentication for Linux servers without GUI


inkolin:
### What would you like to be added?

I'm trying to use gemini-cli on a headless Linux server (including through code-server terminal), but the current authentication flow requires a GUI browser which isn't available in these environments. The OAuth flow opens a browser window that cannot be accessed on servers without desktop environment.
Please implement one of these authentication methods for headless environments:

Device code flow (like gcloud auth login --no-launch-browser)
Copy-paste authentication URL with callback token
API key authentication option
Environment variable for credentials

### Why is this needed?

Currently, gemini-cli is unusable on headless Linux servers, VPS instances, and containerized environments where GUI browsers are not available. This significantly limits the tool's utility for:

Developers working on remote servers via SSH
Teams using cloud-based development environments (AWS Cloud9, GitHub Codespaces, code-server)
Automation workflows and CI/CD pipelines that need Gemini API access
Docker containers and Kubernetes pods
System administrators managing multiple servers

### Additional context

This would greatly improve usability for:

Remote Linux servers
Docker containers
CI/CD pipelines
Cloud development environments (code-server, GitHub Codespaces)

PR #3713, opened 2025-07-10
Add NO_BROWSER environment variable to trigger offline oauth flow

sethtroisi:
If this flag is

1. auth runs before the app starts which helps with copying output.
2. The normal `open(url)` is replaced with `console.log(url)` and `code = readline()`
3. A configurable number of retries are attempted if the code doesn't validate.

This allows people in ssh and other headless environments to perform authentication.

### DIFF
--- a/packages/cli/src/config/config.ts
+++ b/packages/cli/src/config/config.ts
@@ -317,6 +317,7 @@ export async function loadCliConfig(
       name: e.config.name,
       version: e.config.version,
     })),
+    noBrowser: !!process.env.NO_BROWSER,
   });
 }
 
--- a/packages/cli/src/gemini.tsx
+++ b/packages/cli/src/gemini.tsx
@@ -172,10 +172,7 @@ export async function main() {
     config.getNoBrowser()
   ) {
     // Do oauth before app renders to make copying the link possible.
-    await getOauthClient(
-      settings.merged.selectedAuthType,
-      config.getNoBrowser(),
-    );
+    await getOauthClient(settings.merged.selectedAuthType, config);
   }
 
   let input = config.getQuestion();
--- a/packages/cli/src/gemini.tsx
+++ b/packages/cli/src/gemini.tsx
@@ -35,6 +35,7 @@ import {
   sessionId,
   logUserPrompt,
   AuthType,
+  getOauthClient,
 } from '@google/gemini-cli-core';
 import { validateAuthMethod } from './config/auth.js';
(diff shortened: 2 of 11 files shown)


packages/cli/src/ui/App.tsx
```diff
@@ -665,13 +665,29 @@ const App = ({ config, settings, startupWarnings = [] }: AppProps) => {
...
-            />
+            <>
+              <AuthInProgress
+                onTimeout={() => {
+                  setAuthError('Authentication timed out. Please try again.');
+                  cancelAuthentication();
+                  openAuthDialog();
+                }}
+              />
+              {showErrorDetails && (
```
sethtroisi, line 739: Needed if you reach the normal auth screen and want to see output. You can't copy it easily but at least it's available.

NTaylorMullen (dismissed):
Love it and thank you for the detailed comments Seth!

sethtroisi:
Partial fix for #1956

packages/cli/src/gemini.tsx
```diff
@@ -165,6 +166,15 @@ export async function main() {
       }
     }
   }
+
+  if (
+    settings.merged.selectedAuthType === AuthType.LOGIN_WITH_GOOGLE &&
+    config.getNoBrowser()
+  ) {
+    // Do oauth before app renders to make copying the link possible.
+    await getOauthClient(settings.merged.selectedAuthType, config);
```
sethtroisi, line 175: This is the reasonable flow here because you can actually copy-paste the url. BUT it requires the authtype to already be set. the "happy" path right now is open gemini, select login with google, quit, re-start it. This is obviously a terrible "happy" path but it's the best I could figure out today so we'll ship it and improve tomorrow.

scottdensmore:
@sethtroisi if you login with NO_BROWSER and no GOOGLE_CLOUD_PROJECT is set, it lets me login with the code but then when I set the GOOGLE_CLOUD_PROJECT, it just sits spinning waiting for auth.

NTaylorMullen (approved):

Manamama:
I confirm it works fine finally in Termux.

packages/cli/src/config/config.ts
```diff
@@ -317,6 +317,7 @@ export async function loadCliConfig(
       name: e.config.name,
       version: e.config.version,
     })),
+    noBrowser: !!process.env.NO_BROWSER,
```
akolotov, line 320: Looks like it needs to b reflected in the docs, doesn't it?

PR #3714, opened 2025-07-10
Introduce an output character limit the model can adhere to when summarizing.

vachan-shetty:
## TLDR

Introduces a lower bound in terms of number of characters for the model to adhere to in order to prevent unnecessary loss of information when summarizing.

## Dive Deeper

Currently we have a summarization prompt that does not have a max character limit and hence aggressively summarizes losing valuable information in the process. This change adds an output character limit the model can adhere to when summarizing. Currently being set to 0.4 * tokenLimit.

## Reviewer Test Plan
`npm run preflight` and adding new unit tests.


packages/core/src/core/client.ts
```diff
@@ -78,6 +78,45 @@ export function findIndexAfterFraction(
...
+export function calculateCharacterLimit(
+  tokenConsumed: number,
+  tokenLimit: number,
+  historyLength: number,
+): number {
+  if (tokenConsumed === 0) {
+    return 0;
+  }
+
+  const charsPerToken = historyLength / tokenConsumed;
```
NTaylorMullen, line 115: 1 character != token though is this a general algorithm we're trying? We also may need to max all of this against the models max output token limit threshold as well

packages/core/src/core/prompts.ts
```diff
@@ -276,8 +276,16 @@ Your core function is efficient and safe assistance. Balance extreme conciseness
...
  * think in a scratchpad, and produce a structured XML summary.
+ *
+ * The characterLimit ensures a minimum number of characters in the summary,
+ * implicitly, ensuring that summarization preserves important information.
  */
-export function getCompressionPrompt(): string {
+export function getCompressionPrompt(characterLimit: number): string {
+  const characterLimitText =
+    characterLimit > 0
+      ? `The <compressed_chat_history> XML object, MUST have a total character count of around ${characterLimit}. Please generate a summary of appropriate length to meet this target.`
```
NTaylorMullen, line 286: This is definitely interesting, so the goal is to try and encourage the model to output more by explicitly saying that. Not a bad idea. That being said, the model does have limits on how many tokens it can output. I worry that doing this without taking the actual token limit into consideratino will result in it going past its output token limit (32k tokens)

NTaylorMullen (commented):
Left some comments but I'm not sure generically having it try to abide by a character limit is the right move. Typically in conversation history the mass amount of context is derived from file contents. If we encourage a large enough output the model may try to fill historical gaps via partial files or unnecessary / irrelevant files.

What's your take?

abhipatel12:
Closing this for inactivity

PR #1073, opened 2025-06-15
Concise tool call display

keir:
## TLDR

- Switch tool call rendering to a single line where possible, with box rendering characters.
- Render tool call failures

This is not ready for merging. If there is interest in this alternate display approach, it can get cleaned up, tests fixed, etc.

## Dive Deeper

The current tool call rendering uses many lines of the terminal. Those lines push off important code, LLM, and user text that is useful when iterating.

## Reviewer Test Plan

Needs to be put together, but in short try a bunch of tool calling, including chains of tool calls, with some failing, some passing, some cancelled.

## Testing Matrix

For now, only tested locally on mac; not yet ready for merging.

|          | 🍏  | 🪟  | 🐧  |
| -------- | --- | --- | --- |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

PR #3745, opened 2025-07-10
Summarize tool call outputs using tool specific summarizers

anj-s:
## TLDR

 This pull request introduces an automatic summarization feature for tool outputs. When a tool produces a large amount of text, this feature condenses it into a  concise summary before adding it to the chat history. This significantly reduces the number of tokens sent to the model in subsequent turns, making conversations  more efficient and focused, especially for tools with verbose outputs like  run_shell_command. Currently all tools except the run_shell_command tool has a default summarizer that returns the output as is. The run_shell_command has a llmSummarizer. 

## Dive Deeper

The core of this implementation is a new summarizer.ts module, which exports a summarizeToolOutput function. This function is triggered when a tool's output  exceeds a certain character length. It uses a dedicated prompt to ask the LLM to
summarize the text, with specific instructions tailored to different content  types like directory listings, plain text, or shell command output (preserving  error stack traces).
The Tool interface has been updated to include an optional summarizer function and  a shouldSummarizeDisplay boolean flag. This allows individual tools to opt-in to  this new capability and control whether the user sees the summary or the full
  output.

The CoreToolScheduler and NonInteractiveToolExecutor now invoke the tool's  summarizer after successful execution. If a summary is generated, it is used as  the llmContent for the next turn. The run_shell_command tool has been updated to
 use this new llmSummarizer by default.

## Reviewer Test Plan
To validate these changes, you can perform the following checks:

   1. Test Verbose Output Summarization:
       - Run a command that produces a very long output, such as ls -R in a large directory.
       - Expected: The CLI should display the full, verbose output. However, the context sent to the model for the next turn should be a concise summary. You can verify this by observing the model's response, which should refer to the summarized content, or by checking the debug logs for the llmContent.

   2. Test Short Output:
       - Run a command with a brief output, like echo "hello".
       - Expected: The output should not be summarized. The full, original text should be sent to the model.

   3. Test Error Handling with Stack Traces:
       - Execute a command that is designed to fail and produce a long error message or stack trace.
       - Expected: The summarizer should preserve the complete stack trace within the summary, ensuring that critical error information is not lost.

   4. Test Other Tools:
       - Use other tools (e.g., read_file) to ensure their behavior remains unchanged.
       - Expected: Tools that have not opted into the new summarizer should continue to function as before, sending their full output to the model.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | Y  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Resolves #1073


packages/core/src/core/client.ts
```diff
@@ -409,8 +409,9 @@ export class GeminiClient {
     contents: Content[],
     generationConfig: GenerateContentConfig,
     abortSignal: AbortSignal,
+    model?: string,
```
NTaylorMullen, line 413: Do we need this? The APIs on the client can take the model already

packages/core/src/utils/summarizer.ts
```diff
@@ -0,0 +1,131 @@
...
+  }
+  return null;
+}
+
+const toolOutputSummarizerModel = DEFAULT_GEMINI_FLASH_MODEL;
+const toolOutputSummarizerConfig: GenerateContentConfig = {
+  maxOutputTokens: 2000,
+};
+
+const SUMMARIZE_TOOL_OUTPUT_TEMPLATE = `Summarize the following tool output to be a maximum of {maxLength} characters. The summary should be concise and capture the main points of the tool output.
```
NTaylorMullen:
```suggestion
const SUMMARIZE_TOOL_OUTPUT_PROMPT = `Summarize the following tool output to be a maximum of {maxLength} characters. The summary should be concise and capture the main points of the tool output.
```

packages/core/src/utils/summarizer.ts
```diff
@@ -0,0 +1,131 @@
...
+  if (Array.isArray(part)) {
+    return part.map(partToString).join('');
+  }
+  if ('text' in part) {
+    return part.text ?? '';
+  }
+  return '';
+}
+
+function getResponseText(response: GenerateContentResponse): string | null {
```
NTaylorMullen, line 60: I'd be surprised if the genai SDK didn't have a way already to pull text from responses

packages/core/src/tools/tools.ts
```diff
@@ -43,6 +44,16 @@ export interface Tool<
    */
   canUpdateOutput: boolean;
 
+  /**
+   * A function that summarizes the result of the tool execution.
+   */
+  summarizer?: Summarizer;
```
NTaylorMullen, line 50: Instead of baking a summarizer into the core type you may want to consider making "summarization" a detail of "execute" and remove all of the top-level contracts around it. The reason why is that with how you're centrally summarizing content right now you're requiring every tool to construct their own summarizer which needs to take in text as an input and output text as a mechanism of reduction. As an example, lets say we add summarization to our grep tool. 

You could potentially use an LLM to grep over the contents of that tool OR you could also make a decision to summarize content by removing searched file paths based offf of how deep in the folder structure they were. 

Basically by making summarization a detail of execute you have the option to utilize the rich data and pre-computed information already available. With the right APIs the additions to each tool should be minimal.

packages/core/src/utils/summarizer.ts
```diff
@@ -0,0 +1,131 @@
...
+const toolOutputSummarizerConfig: GenerateContentConfig = {
+  maxOutputTokens: 2000,
+};
+
+const SUMMARIZE_TOOL_OUTPUT_TEMPLATE = `Summarize the following tool output to be a maximum of {maxLength} characters. The summary should be concise and capture the main points of the tool output.
+
+The summarization should be done based on the content that is provided. Here are the basic rules to follow:
+1. If the text is a directory listing or any output that is structural, use the history of the conversation to understand the context. Using this context try to understand what information we need from the tool output and return that as a response.
+2. If the text is text content and there is nothing structural that we need, summarize the text.
+3. If the text is the output of a shell command, use the history of the conversation to understand the context. Using this context try to understand what information we need from the tool output and return a summarization along with the stack trace of any error within the <error></error> tags. The stack trace should be complete and not truncated. If there are warnings, you should include them in the summary within <warning></warning> tags.
```
NTaylorMullen, line 87: I may have missed it, do we parse warnings anywhere?

NTaylorMullen (commented):
Super excited to get this in! Great prompt btw

packages/core/src/tools/tools.ts
anj-s: Agreed. I considered this as well as one of the ways to implement it.  The initial thought was that there was sufficient common requirements for summarization that having a default summarizer would be valuable.  

Do you think we can do this as a second pass if needed and still get this in since we want some form of summarization for shell tools at least?

NTaylorMullen (dismissed):
Approving with expectation of fast-follow for the refactor

packages/core/src/core/client.ts
anj-s: The generateContent API that I am using doesn't but the generateJson one does. I am not sure why we have this discrepancy but can look into it in the follow up PR.

packages/core/src/tools/tools.ts
anj-s: Discussed async. The original thought was that a default summarizer would be applicable to multiple tools. The flip side is that this would not allow extensibility for tools to define their own summarization. Agreed to move to that model in a follow up PR. This also makes sense with what we are thinking of for MCP tool output summarization.

packages/core/src/utils/summarizer.ts
anj-s: Yes, I thought so to. I have a TODO to either move these functions to utils to consolidate or figure out what I can reuse.

packages/core/src/utils/summarizer.ts
anj-s: We don't right now but its a way to categorize and represent what is in the logs.

packages/core/src/core/client.ts
```diff
@@ -410,8 +410,9 @@ export class GeminiClient {
...
   ): Promise<GenerateContentResponse> {
-    const modelToUse = this.config.getModel();
+    const modelToUse = model ?? this.config.getModel();
```
mattKorwel, line 415: Why the change to the selected model here? This seems maybe unrelated?

mattKorwel (approved):

mattKorwel:
Lgtm. I didn't click auto merge as three are several open comments but feel free to merge of you are good to go!

packages/core/src/core/client.ts
anj-s: I want to override the mode if specified by the config when calling this function.

PR #3775, opened 2025-07-10
fix(editors): Support VS Code Insiders

### LINKED ISSUE #1495: Support VSCode Insiders
## Describe the bug
VSCode is supported, but only if using the mainline version. I use VSCode Insiders (the canary channel).

## Additional context
- https://code.visualstudio.com/insiders/
- Invoked with `code-insiders` (instead of `code`)
- Otherwise it's the same

## Diagnostic Information
*   **CLI Version:** 0.1.0
*   **Git Commit:** 852210e1 (local modifications)
*   **Operating System:** darwin v24.3.0
*   **Memory Usage:** 305.9 MB

swissspidy:
## TLDR

Support [Visual Studio Code Insiders](https://code.visualstudio.com/insiders/) (`code-insiders`) as an editor.

## Dive Deeper

Screenshot 2025-07-10 at 14 07 40

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes #1495

### DIFF
--- a/packages/cli/src/ui/editors/editorSettingsManager.ts
+++ b/packages/cli/src/ui/editors/editorSettingsManager.ts
@@ -19,6 +19,7 @@ export interface EditorDisplay {
 export const EDITOR_DISPLAY_NAMES: Record<EditorType, string> = {
   zed: 'Zed',
   vscode: 'VS Code',
+  'vscode-insiders': 'VS Code Insiders',
   vscodium: 'VSCodium',
   windsurf: 'Windsurf',
   cursor: 'Cursor',
@@ -33,6 +34,7 @@ class EditorSettingsManager {
     const editorTypes: EditorType[] = [
       'zed',
       'vscode',
+      'vscode-insiders',
       'vscodium',
       'windsurf',
       'cursor',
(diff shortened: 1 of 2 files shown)


abhipatel12:
Hey Pascal! Is there a strong reason for why we would want to add the pre-release/beta version for VS code as one of the static options?

swissspidy:
Great question!

There appears to be some demand for this, see #1495, and I thought it's a low-hanging fruit.

But now I actually stumbled upon the lengthy discussion at #1633 and the related issue #1698.

This could lead to a more generalized solution where custom editor(s) could be configured in the settings instead.

abhipatel12:
Yep! That was my exact concern. I think we need to do our best to support a variety of editors. A generalized solution, like you mentioned, would be best!

swissspidy:
I just opened #3849 to summarize & centralize discussions around this 👍

PR #3834, opened 2025-07-11
chore(deps): Pin @google/genai to 1.8.0

trapezoid:
@google/genai@1.9.0 introduces an incompatibility.

## TLDR

This PR pins the @google/genai dependency to the exact version 1.8.0 to prevent a critical issue where all tool schemas registered via MCP become invalid. The ^1.8.0 specifier in package.json currently allows the incompatible v1.9.0 to be installed, which breaks tool function calls.

## Dive Deeper

The gemini-cli currently specifies its dependency on @google/genai as ^1.8.0 in `packages/core/package.json` . This allows npm or yarn to automatically install the latest minor version, which is 1.9.0.

However, @google/genai@1.9.0 introduced a breaking change related to FunctionDeclaration schema conversion, as detailed in [this commit in the googleapis/js-genai repository](https://github.com/googleapis/js-genai/commit/36f6350705ecafc47eaea3f3eecbcc69512edab7#diff-fdde9372aec8593).

This change breaks the logic in the McpClient ([mcp-client.ts#L350](https://github.com/google-gemini/gemini-cli/blob/v0.1.10/packages/core/src/tools/mcp-client.ts#L350), causing it to fail when processing tool arguments. As a result, all tool schemas are incorrectly registered in MCP with a generic {"type": "object", "properties": {}} schema. This effectively makes all tool arguments any and prevents the model from correctly calling tools with parameters.

This PR resolves the issue by pinning the dependency to exactly 1.8.0, ensuring a compatible version is used until the gemini-cli can be updated to support the new schema structure.
## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓   | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11499,7 +11499,7 @@
       "name": "@google/gemini-cli-core",
       "version": "0.1.10",
       "dependencies": {
-        "@google/genai": "^1.8.0",
+        "@google/genai": "1.8.0",
         "@modelcontextprotocol/sdk": "^1.11.0",
         "@opentelemetry/api": "^1.9.0",
         "@opentelemetry/exporter-logs-otlp-grpc": "^0.52.0",
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -20,7 +20,7 @@
     "dist"
   ],
   "dependencies": {
-    "@google/genai": "^1.8.0",
+    "@google/genai": "1.8.0",
     "@modelcontextprotocol/sdk": "^1.11.0",
     "@opentelemetry/api": "^1.9.0",
     "@opentelemetry/exporter-logs-otlp-grpc": "^0.52.0",


trapezoid:
While switching the client to use parametersJsonSchema might be the long-term solution, it has not been verified. Therefore, this PR takes the safest immediate action by pinning the dependency to exactly 1.8.0.

NTaylorMullen (approved):

mattKorwel (approved):

mattKorwel:
Nice catch @trapezoid thanks for the contribution!

PR #3863, opened 2025-07-11
Added in proper checks for customer tiers in 429/Quota error messaging

### LINKED ISSUE #3862: Quota error message display should take into account Code Assist customer type
### What happened?

Currently, the quota/429 error message displays throughout the app are not fully taking the customer's free/paying Code Assist types into account. We shouldn't prompt paying users to upgrade to a paid subscription, for example.

I tried to fix this earlier but missed key error messaging in App.tsx so it's likely that many users are still seeing the incorrect message. 

### Client information

<details>

```console
$ gemini /about
# paste output here
```

</details>

### Login information

_No response_

_No response_

bdmorgan:
User Tier Detection Implementation

  Added automatic user tier detection for OAuth users to enable personalized quota error messaging:

  - User Tier State Management: Added userTier state to track the current user's subscription tier (FREE, STANDARD, LEGACY)
  - Automatic Tier Detection: Implemented a useEffect hook that automatically detects the user's tier for OAuth users by:
    - Checking if the user is authenticated via Google OAuth (AuthType.LOGIN_WITH_GOOGLE)
    - Calling the CodeAssist API to retrieve the user's current tier information
    - Gracefully handling detection failures with debug logging
  - Authentication State Integration: The tier detection properly integrates with the authentication flow:
    - Runs after authentication is complete (listens to isAuthenticating state)
    - Resets tier information when authentication changes
    - Skips detection during active authentication processes
  - Enhanced Quota Error Messages: The existing Flash fallback handler now uses the detected user tier to show appropriate upgrade suggestions:
    - Paid tier users: Directed to use API keys from AI Studio
    - Free tier users: Shown upgrade options for Code Assist Standard/Enterprise plans

  This enables the CLI to provide more relevant and helpful error messages based on the user's actual subscription status, improving the user experience during quota limit scenarios.

This should resolve issue #3862

### DIFF
--- a/packages/cli/src/ui/App.tsx
+++ b/packages/cli/src/ui/App.tsx
@@ -71,6 +71,8 @@ import {
   isProQuotaExceededError,
   isGenericQuotaExceededError,
   UserTierId,
+  AuthType,
+  CodeAssistServer,
 } from '@google/gemini-cli-core';
 import { checkForUpdates } from './utils/updateCheck.js';
 import ansiEscapes from 'ansi-escapes';
@@ -139,6 +141,7 @@ const App = ({ config, settings, startupWarnings = [] }: AppProps) => {
   const [showPrivacyNotice, setShowPrivacyNotice] = useState<boolean>(false);
   const [modelSwitchedFromQuotaError, setModelSwitchedFromQuotaError] =
     useState<boolean>(false);
+  const [userTier, setUserTier] = useState<UserTierId | undefined>(undefined);
 
   const openPrivacyNotice = useCallback(() => {
     setShowPrivacyNotice(true);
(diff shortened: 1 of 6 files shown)


packages/cli/src/ui/App.tsx
```diff
@@ -174,6 +177,55 @@ const App = ({ config, settings, startupWarnings = [] }: AppProps) => {
...
+        }
+      } catch (error) {
+        // Silently fail - this is not critical functionality
+        // We'll default to FREE tier behavior if tier detection fails
+        console.debug('User tier detection failed:', error);
+      }
+    };
+
+    detectUserTier();
+  }, [config, isAuthenticating]);
```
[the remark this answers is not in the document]
bdmorgan: I've implemented the cleanup function pattern you suggested to prevent stale API responses from overwriting newer data. The isCancelled flag now ensures that only the most recent API call can update the user tier state.

packages/cli/src/ui/App.tsx
```diff
@@ -174,6 +177,65 @@ const App = ({ config, settings, startupWarnings = [] }: AppProps) => {
...
+        console.debug('User tier detection failed:', error);
+      }
+    };
+
+    detectUserTier();
+
+    return () => {
+      isCancelled = true;
+    };
+  }, [config, isAuthenticating]);
```
swissspidy: Instead of doing this in an effect, could this perhaps be called within `Config.refreshAuth()`? There, one has access to the relevant bits such as `authType`, `geminiClient`, `contentGeneratorConfig`. And `userTier` could be stored in the `Config` instance too.

This avoids downsides of running async stuff in an effect, and ensures the user tier detection runs only when auth really changes.

bdmorgan: Good catch - I've refactored the user tier detection to run directly in Config.refreshAuth() instead of using a useEffect. This eliminates the race condition entirely and provides better architectural separation - the Config class now manages all authentication-related state including user tier detection, while the App component simply syncs the UI state from the Config. This approach is much cleaner and more maintainable.

packages/core/src/config/config.ts
```diff
@@ -272,6 +275,51 @@ export class Config {
...
+    // Detect user tier for OAuth users
+    await this.detectUserTier();
+  }
+
+  private async detectUserTier(): Promise<void> {
+    try {
+      // Reset user tier when authentication changes
+      this.userTier = undefined;
+
+      // Only attempt tier detection for OAuth users
```
scidomino: This seems like a really gross way to do it. Couldn't we store the tier in the CodeAssistServer and change the ContentGenerator interface to have a getTier method?

bdmorgan: I refactored as follows:

  1. Extended ContentGenerator interface with an optional getTier() method
  2. Moved tier detection logic into CodeAssistServer where it belongs
  3. Removed all the messy coupling from Config - it now just delegates to the interface
  4. Eliminated type casting and tight coupling between classes

CodeAssistServer handles its own tier detection with lazy loading, and Config now simply delegates through the interface.

mattKorwel (approved):
Lgtm

PR #3868, opened 2025-07-11
Declarative keybindings preliminary design review

dewitt:
This PR includes a preliminary high-level design for refactoring the keybinding declaration and discovery mechanisms.

At this stage, the PR intentionally does not include any code, but rather, as a matter of good design process, only includes a high-level outline of the goals and requirements of the project so we can get human expert feedback earlier in the loop.

Reviewers should focus on the implementation proposed by Gemini and offer feedback on docs/design/declarative-keybindings.md, either via in-line file comments or via responses to this PR, which I will then include verbatim for Gemini to subsequently iterate on.

This process itself is an experiment toward getting more human visibility on large-scale or potentially breaking changes with a large blast radius much earlier in the development cycle. If this process proves helpful we can document it as a best practice for future large-scale feature work. (Assistant-driven engineering has proven it so easy to go from idea to running code that we're starting to land complex changes without entirely grasping their impact.)


dewitt:
@jacob314 for review on the preliminary design proposal itself, and CC @NTaylorMullen for visibility and thoughts on the process for getting humans in the loop early in the design stage.

dewitt:
Also noting with no small joy that @gemini-code-assist was so quick to offer its own feedback on this Gemini-proposed design! : )

docs/design/declarative-keybindings.md
```diff
@@ -0,0 +1,137 @@
...
+
+### **1. Overview**
+
+The current implementation of keybindings in the Gemini CLI relies on the
+`useInput` hook from the `ink` library. This hook is used in various
+components to handle keyboard input. The main application component,
+`App.tsx`, handles global keybindings, while other components (like dialogs)
+handle their own specific key events. This decentralized approach makes it
+difficult to discover, manage, and update keybindings. It also increases the
+risk of conflicting key assignments.
```
jacob314, line 66: Might want to call out that InputPrompt and TextBuffer are the most complex keybindings so the solution must handle them well.

docs/design/declarative-keybindings.md
```diff
@@ -0,0 +1,137 @@
...
+
+1. **Refactor Existing Keybindings:**
+   - Gradually refactor existing `useInput` hooks to use the new keybinding
+     system.
+   - For each keybinding, register it with the `KeybindingRegistry` and add an
+     entry to the configuration file.
+   - This can be done component by component to minimize disruption.
+
+1. **Implement Keybinding Discovery:**
+   - Add a `/help` command or a dedicated help view that displays all
```
jacob314, line 110: not sure this should be /help. might clarify which command

docs/design/declarative-keybindings.md
```diff
@@ -0,0 +1,137 @@
...
+     registered keybindings and their current assignments. This will use the
+     information from the `KeybindingRegistry`.
+
+### **3. Risks and Alternatives**
+
+- **Risk:** Introducing a new system could be complex and might introduce new
+  bugs.
+  - **Mitigation:** We will implement the new system behind a feature flag
```
jacob314, line 118: don't think this is needed.

docs/design/declarative-keybindings.md
```diff
@@ -0,0 +1,137 @@
...
+   - Define a `Keybinding` interface with properties like `name`,
+     `description`, `callback`, and `defaultKey`.
+   - Implement a `KeybindingRegistry` class with methods to `register` and
+     `get` keybindings.
+   - The registry will be a singleton, accessible throughout the application.
+
+1. **Implement the Keybinding Service:**
+   - Create a `KeybindingService` that reads the keybinding configuration from
+     a file.
+   - The service will use a single `useInput` hook at the top level of the
```
jacob314, line 95: this doesn't seem very algined with existing code with useInput calls throughout the application and InputPrompt that does its own non-ink standard keyboard processing. Given that I would suggest a solution that is lighter weight and instead allows callers to resolve whether specific input matches a specific keybinding rather than a solution that requires taking control of all useInput in the application.

jacob314 (commented):
done with one round of review. one thought after reviewing is it might have been interesting to have some even if partially broken code from gemini cli to make some of the proposals more concrete. for example, would have been curious for this pr to see what Gemini CLI was trying to do for text-buffer if it had even considered it.

dewitt:
> done with one round of review. 

Great feedback, Jacob, thank you! I am going find a way to capture that verbatim, along with my own personal feedback on the proposal (which I intentionally haven't given it yet), and see how Gemini respond. I'll return here with an updated design doc accordingly.

> one thought after reviewing is it might have been interesting to have some even if partially broken code from gemini cli to make some of the proposals more concrete. for example, would have been curious for this pr to see what Gemini CLI was trying to do for text-buffer if it had even considered it.

I thought about doing it that way, but I've noticed that it tends to get on a tear sometimes and it is hard to give it a clear stopping point. The next feature we build this way maybe I'll try prompting with a "and implement a proof of concept" or something so we have a few files of code to look at.

FWIW, I don't think it discovered all the places that inputs need to be considered, and didn't even run a cursory recursive grep for indicators like `ctrl` or whatnot. In my own feedback I'm going to tell it to do a more thorough code review along those lines.

Thanks again!

docs/design/declarative-keybindings.md
```diff
@@ -127,8 +127,89 @@ complexity and ensure a smooth transition.
...
+- **Challenge:** How does the system determine which keybinding to execute when
+  multiple contexts are active or when a component-specific keybinding
+  should override a global one?
+- **Suggestion:** The design must incorporate a more sophisticated approach to
+  context. This could involve a "lighter weight" solution where callers can
+  resolve whether a specific input matches a keybinding, rather than a single
+  service that takes full control of all keyboard input. This aligns better
+  with the existing codebase, where `useInput` is used in multiple components.
+
+#### **2. Separation of Concerns: Actions vs. Mappings**
```
jacob314, line 159: wonder how to separate this so as a reviewer I can scan to the parts I care about and quickly ignore the parts that are just the discussion with GeminiCLI

docs/design/declarative-keybindings.md
```diff
@@ -127,8 +127,89 @@ complexity and ensure a smooth transition.
...
+- **Complex Components:** The solution must be robust enough to handle the most
+  complex existing keybinding implementations, such as those in `InputPrompt`
+  and `TextBuffer`.
+- **Discovery Command:** The proposal to use `/help` for keybinding discovery
+  was questioned. A more specific or dedicated command might be more
+  appropriate.
+- **Feature Flag:** The suggestion to put the new system behind a feature flag
+  was deemed unnecessary.
+
+### **Summary of Reviewer Recommendations**
```
jacob314, line 200: lgtm. one other thing I forgot to discuss that we might mention is make it clearer to users how keybindings vary on windows vs mac. Right now I worry our keyboard binding descriptions are confusing to users who might not know how, control, option, and command map to the linux or windows equivalents.

jerop:
@dewitt closing because we're cleaning up issues and prs, feel free to reopen when you pick this up again

PR #3869, opened 2025-07-11
Fix: Make video and audio file reading actually work

### LINKED ISSUE #3379: Agent thinks it cannot read videos, even though these are supported filetypes
### What happened?

When asking the agent to read video or audio (put it in the context). I sometimes encounter:

```
My current capabilities do not allow me to directly "watch" or process video files within the prompt. My visual analysis is limited to static images.
```

Even though mp4 / mp3 etc are supported filetypes: https://github.com/google-gemini/gemini-cli/blob/39e8509452b359774cc363a7e98810e48ec516a4/packages/core/src/tools/read-many-files.test.ts#L28-L33

I can convince the model to read videos but it takes an additional instruction.

It should believe it can read video and audio files.

### Client information

<details>

```console
$ gemini /about
mac OS 0.1.9
```

</details>

### Login information

_No response_

_No response_

IgorAherne:
## TLDR

This PR fixes a bug so the CLI can *actually* read video and audio files. 

Another change (#3380) updated the instructions to say this was possible, but the code itself was still broken. This PR fixes the code so it works as expected.

## Dive Deeper

**The Problem:**

On some computers (like mine), the tool we use to identify file types gets it wrong.
*   It labeled `.mp4` video files with the wrong tag (`application/mp4` instead of `video/mp4`), so the Gemini API would reject them.
*   It labeled `.ts` code files as videos, which is a big mistake.

**The Fix:**

Instead of fully trusting that tool, I made our code smarter:
1.  I created a simple list inside our code that has the **correct** tags for important files like `.mp4`, `.mov`, `.mp3`, and `.wav`.
2.  Our code now checks this list **first** before asking the old tool. This guarantees the right tag is always used for media files.
3.  I made sure the special rule for TypeScript files still works, so they are not mistaken for videos.

## Reviewer Test Plan

To make sure this works, you can:

1.  Get the code from this branch.
2.  Find a short `.mp4` video on your computer.
3.  Run this command: `gemini "Tell me what this video is about: @your_video_name.mp4"`. 
4. You can also do it now **without** the @ symbol, simply telling it to go find that video in the folder and it still works without pointing directly.
5.  **It should now work!** The CLI should process the video without errors.
6.  To make sure I didn't break anything, also test a code file: `gemini "What is this code file? @packages/core/src/utils/fileUtils.ts"`
7.  **It should still work correctly** and treat it as a text file.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  |  ✅ |
| npx      | ❓  | ❓  |  ✅ |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes #3379

### DIFF
--- a/diagnose-mime.cjs
+++ b/diagnose-mime.cjs
@@ -1,0 +1,11 @@
+// This is the correct code for a file named diagnose-mime.cjs
+const mime = require('mime-types');
+
+console.log('--- Mime-types Diagnosis ---');
+const extensions = ['.mp4', '.mov', '.mp3', '.wav', '.ts', '.js', '.png', '.pdf'];
+
+extensions.forEach(ext => {
+    const lookupResult = mime.lookup(ext);
+    console.log(`mime.lookup('${ext}') -> '${lookupResult}' (Type: ${typeof lookupResult})`);
+});
+console.log('--- End of Diagnosis ---');
--- a/packages/cli/src/ui/hooks/atCommandProcessor.ts
+++ b/packages/cli/src/ui/hooks/atCommandProcessor.ts
@@ -271,6 +271,15 @@ export async function handleAtCommand({
       pathSpecsToRead.push(currentPathSpec);
       atPathToResolvedSpecMap.set(originalAtPath, currentPathSpec);
       contentLabelsForDisplay.push(pathName);
+    } else {
+      // If stat and glob failed, trust the user and pass the path to the tool anyway.
+      // The tool itself is robust enough to handle read errors.
+      onDebugMessage(
+        `Could not resolve path "${pathName}" directly, but will attempt to read it as-is.`,
+      );
+      pathSpecsToRead.push(pathName);
+      atPathToResolvedSpecMap.set(originalAtPath, pathName);
+      contentLabelsForDisplay.push(pathName);
     }
   }
 
--- a/packages/core/src/utils/fileUtils.ts
+++ b/packages/core/src/utils/fileUtils.ts
@@ -255,6 +255,7 @@ export async function processSingleFileContent(
           returnDisplay: `Skipped binary file: ${relativePathForDisplay}`,
         };
       }
+      case 'svg':
       case 'text': {
         const content = await fs.promises.readFile(filePath, 'utf8');
         const lines = content.split('\n');
--- a/packages/core/src/utils/fileUtils.ts
+++ b/packages/core/src/utils/fileUtils.ts
@@ -9,6 +9,18 @@ import path from 'path';
 import { PartUnion } from '@google/genai';
 import mime from 'mime-types';
 
+/** 
+ *the single source of truth for
+ * overriding MIME types that are problematic in some environments or need to be
+ * normalized for the Gemini API. 
+*/
+const customMimeMap = new Map<string, string>([
+  ['.mp4', 'video/mp4'],
+  ['.mov', 'video/mp4'], // Normalize .mov to video/mp4 for Gemini
+  ['.mp3', 'audio/mpeg'],
+  ['.wav', 'audio/mpeg'], // Normalize .wav to audio/mpeg for Gemini
+]);
+
 // Constants for text file processing
 const DEFAULT_MAX_LINES_TEXT_FILE = 2000;
 const MAX_LINE_LENGTH_TEXT_FILE = 2000;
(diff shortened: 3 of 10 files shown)


packages/core/src/utils/fileUtils.ts
```diff
@@ -239,20 +255,7 @@ export async function processSingleFileContent(
...
-            returnDisplay: `Skipped large SVG file (>1MB): ${relativePathForDisplay}`,
-          };
-        }
-        const content = await fs.promises.readFile(filePath, 'utf8');
-        return {
-          llmContent: content,
-          returnDisplay: `Read SVG as text: ${relativePathForDisplay}`,
-        };
-      }
+      case 'svg':
```
NTaylorMullen, line 258: I wouldn't change this

diagnose-mime.cjs
NTaylorMullen, line 1: Needed?

packages/core/src/utils/fileUtils.ts
```diff
@@ -9,6 +9,18 @@ import path from 'path';
 import { PartUnion } from '@google/genai';
 import mime from 'mime-types';
 
+/** 
+ *the single source of truth for
+ * overriding MIME types that are problematic in some environments or need to be
+ * normalized for the Gemini API. 
+*/
+const customMimeMap = new Map<string, string>([
```
NTaylorMullen, line 17: @amirh do you know if there's an official list of these somewhere?

tsconfig - Copy.json:Zone.Identifier
NTaylorMullen, line 1: can delete

NTaylorMullen (changes requested):
Thanks for the contribution! I actually wasn't aware of the incompatibilities of Gemini API with these types. That seems odd.

diagnose-mime.cjs
leehagoodjames: Ditto

diagnose-mime.cjs
```diff
@@ -0,0 +1,11 @@
...
+const mime = require('mime-types');
+
+console.log('--- Mime-types Diagnosis ---');
+const extensions = ['.mp4', '.mov', '.mp3', '.wav', '.ts', '.js', '.png', '.pdf'];
+
+extensions.forEach(ext => {
+    const lookupResult = mime.lookup(ext);
+    console.log(`mime.lookup('${ext}') -> '${lookupResult}' (Type: ${typeof lookupResult})`);
+});
+console.log('--- End of Diagnosis ---');
```
leehagoodjames, line 11:
```suggestion
```

packages/cli/src/ui/hooks/atCommandProcessor.ts
```diff
@@ -271,6 +271,15 @@ export async function handleAtCommand({
...
       contentLabelsForDisplay.push(pathName);
+    } else {
+      // If stat and glob failed, trust the user and pass the path to the tool anyway.
+      // The tool itself is robust enough to handle read errors.
+      onDebugMessage(
+        `Could not resolve path "${pathName}" directly, but will attempt to read it as-is.`,
+      );
+      pathSpecsToRead.push(pathName);
+      atPathToResolvedSpecMap.set(originalAtPath, pathName);
+      contentLabelsForDisplay.push(pathName);
```
leehagoodjames, line 282: Nit: Is this necessary for this PR? Path resolving seems different than fixing the inferred mime type, and if so it would be good to pull into another PR

```suggestion
```

packages/core/src/utils/fileUtils.ts
leehagoodjames: +1 to referencing an official list
naming nit: 
```suggestion
const extensionToMimeMap = new Map<string, string>([
```

leehagoodjames (commented):
@IgorAherne thanks for this PR! 

I tested it locally, and was able to have an example that failed at head and succeeded with your changes - so I'd love to get your changes merged in! 

Manual testing screenshot (head=top terminal, this branch=bottom terminal)
image

Can you please address the comments so we can get these changes in?

leehagoodjames:
Hi @IgorAherne, we are trying to clean our repo from stale PRs that are no longer being actively pursued. Given the lack of activity, I am going to close this PR as inactive, but if you still want to get this change in please just re-open this PR

PR #3880, opened 2025-07-11
Refactor MCP code for reuse and testing

### LINKED ISSUE #3889: Reorg MCP code to allow à la carte server creation

Allow the creation of MCP tools outside of the mcp config framework.

### Why is this needed?

Background agent providers will be exposed using MCP servers but we don't want them configured using the same system since we want to be able to swap them in and out.

### Additional context

_No response_

scidomino:
## TLDR

Refactors mcp code so it can be reused for background tasks.

## Reviewer Test Plan

Startup with an mcp server and verify that it works as expected.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | x  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes #3889

### DIFF
--- a/packages/core/src/tools/mcp-client.ts
+++ b/packages/core/src/tools/mcp-client.ts
@@ -5,6 +5,7 @@
  */
 
 import { Client } from '@modelcontextprotocol/sdk/client/index.js';
+import { Transport } from '@modelcontextprotocol/sdk/shared/transport.js';
 import { StdioClientTransport } from '@modelcontextprotocol/sdk/client/stdio.js';
 import {
   SSEClientTransport,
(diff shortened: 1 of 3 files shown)


packages/core/src/tools/mcp-client.ts
```diff
@@ -167,50 +180,111 @@ export async function discoverMcpTools(
...
-      } as Record<string, string>,
-      cwd: mcpServerConfig.cwd,
-      stderr: 'pipe',
-    });
-  } else {
-    console.error(
-      `MCP server '${mcpServerName}' has invalid configuration: missing httpUrl (for Streamable HTTP), url (for SSE), and command (for stdio). Skipping.`,
-    );
-    // Update status to disconnected
+      const tools = await dicoverTools(
```
abhipatel12: small type: `dicoverTools` --> `discoverTools` 

(missing `s`)

packages/core/src/tools/mcp-client.ts
```diff
@@ -156,16 +154,31 @@ export async function discoverMcpTools(
...
-  } catch (error) {
-    // Still mark as completed even with errors
+  } finally {
     mcpDiscoveryState = MCPDiscoveryState.COMPLETED;
-    throw error;
   }
 }
 
+/** Visible for Testing */
+export function appendMcpServerCommand(
```
NTaylorMullen:
```suggestion
export function populateMcpServerCommand(
```

NTaylorMullen (dismissed):

mattKorwel (approved):

PR #3881, opened 2025-07-11
Grok

taf2:
Just seems like we could add more models and grok4 seems pretty good so it'd be nice to use it in a cli.


allenhutchison:
Hi @taf2 thank you for your contribution, but at this time we aren't accepting PRs to support models outside of the gemini family.

Issue #3653, opened 2025-07-09
Screen flicker when selecting theme


bryanjacquot:
### What happened?

Running Gemini CLI within the Cloud Shell Terminal in the Google Cloud Console, when prompted to select the theme, using arrow keys to move up/down between selections the screen flickers with the Gemini ascii art logo appearing and disappearing as part of the flicker.

![Image](https://github.com/user-attachments/assets/11819db5-4d19-4485-8b63-08f5d7841944)

### What did you expect to happen?

I should be able to navigate between themes without screen flicker.

### Client information

<details>

```console
$ gemini /about
# paste output here
```

</details>

jackwotherspoon:
This seems to be an issue when the terminal height is of a small size (which is usually the case seen in Cloud Shell).

I can reproduce this on my mac when I make the height of my terminal small. 

Thanks for raising @bryanjacquot 😄 

cc @jacob314 as I think there have been other terminal size related issues

jacob314:
Our logic for computing the height available must be off by 1 in this case as "Apply To" is still showing even though there is not quite enough height available. The UI is supposed to adapt to hide that when there is not space.

gsquared94:
Found possible duplicate issues:

- #2428: (0.9124502811566083)
- #2859: (0.9052801998043124)
- #3055: (0.9035530032845142)
- #2431: (0.9000946551589569)

PR #3895, opened 2025-07-11
Add scrolling to theme dialog

### LINKED ISSUE #3890: Better support for more themes

We need a way to display more themes with our limited space in the theme dialog

Image

### Why is this needed?

More requests will appear to support more themes and we need a scalable solution

### Additional context

_No response_

miguelsolorio:
## TLDR

Adds scrolling to theme dialog to allow more themes to be shown

## Dive Deeper

https://github.com/user-attachments/assets/22a29b8c-6ca5-43fd-a9bc-5658c1afd312

This copies the same interaction pattern from our suggestions. 

There is a minor screen flicker (current issue #3653) where the screen flickers when navigating through the themes

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes #3890

### DIFF
--- a/packages/cli/src/ui/components/shared/RadioButtonSelect.tsx
+++ b/packages/cli/src/ui/components/shared/RadioButtonSelect.tsx
@@ -66,7 +66,7 @@ export function RadioButtonSelect<T>({
         onSelect(items[activeIndex]!.value);
       }
     },
-    { isActive: isFocused },
+    { isActive: isFocused && items.length > 0 },
   );
 
   const visibleItems = items.slice(scrollOffset, scrollOffset + maxItemsToShow);
--- a/packages/cli/src/ui/components/shared/RadioButtonSelect.tsx
+++ b/packages/cli/src/ui/components/shared/RadioButtonSelect.tsx
@@ -86,6 +86,17 @@ export function RadioButtonSelect<T>({
       if (key.return) {
         onSelect(items[activeIndex]!.value);
       }
+
+      // Enable selection directly from number keys.
+      if (/^[1-9]$/.test(input)) {
+        const targetIndex = Number.parseInt(input, 10) - 1;
+        if (targetIndex >= 0 && targetIndex < visibleItems.length) {
+          const selectedItem = visibleItems[targetIndex];
+          if (selectedItem) {
+            onSelect?.(selectedItem.value);
+          }
+        }
+      }
     },
     { isActive: isFocused && items.length > 0 },
   );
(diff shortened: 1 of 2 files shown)


packages/cli/src/ui/components/shared/RadioButtonSelect.tsx
```diff
@@ -4,139 +4,122 @@
...
-      themeNameDisplay?: string;
-      themeTypeDisplay?: string;
-    }
-  >;
-
-  /** The initial index selected */
+  items: Array<RadioSelectItem<T>>;
   initialIndex?: number;
-
-  /** Function called when an item is selected. Receives the `value` of the selected item. */
```
jacob314, line 41: add back comments from this file that Gemini CLI for some reason decided were low value.

packages/cli/src/ui/components/shared/RadioButtonSelect.tsx
```diff
@@ -4,139 +4,122 @@
...
+        let textColor = Colors.Foreground;
+        if (isSelected) {
+          textColor = Colors.AccentGreen;
+        } else if (item.disabled) {
+          textColor = Colors.Gray;
+        }
 
-  initialIndex = initialIndex ?? 0;
-  return (
-    <SelectInput
```
jacob314, line 132: fyi: https://github.com/vadimdemedes/ink-select-input/blob/master/src/SelectInput.tsx
already had a limit prop but I think your functionality is a bit clearer due to the arrows clarifying when there are more items.you might want to notify Gemini CLI what functionality https://github.com/vadimdemedes/ink-select-input/blob/master/src/SelectInput.tsx
hadd so you don't lose support for things like selecting by pressing a number which the previous functionality provided as it turns out.

miguelsolorio: Good catch, updated to include pressing numbers for selection

packages/cli/src/ui/components/shared/RadioButtonSelect.tsx
```diff
@@ -4,139 +4,122 @@
...
-  ): React.JSX.Element {
-    const { isSelected = false, label } = props;
-    const itemWithThemeProps = props as typeof props & {
-      themeNameDisplay?: string;
-      themeTypeDisplay?: string;
-      disabled?: boolean;
-    };
+  useInput(
+    (input, key) => {
+      if (key.upArrow) {
```
jacob314: nit: can you add support for 'j' and 'k' to match the existing ink library we were previously using?

jacob314 (approved):
lgtm

mattKorwel (approved):

PR #3921, opened 2025-07-11
Add numbers to selection list

miguelsolorio:
## TLDR

Adds numbers to selection list for better accessibility support

## Dive Deeper

https://github.com/user-attachments/assets/2e75e8dd-aec6-4286-abcd-3bf78966e429

This only appears for selection through the model, currently not supported for suggestions or theme dialog.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes #1776

### DIFF
--- a/packages/cli/src/ui/components/ThemeDialog.tsx
+++ b/packages/cli/src/ui/components/ThemeDialog.tsx
@@ -196,6 +196,7 @@ export function ThemeDialog({
             onSelect={handleThemeSelect}
             onHighlight={onHighlight}
             isFocused={currenFocusedSection === 'theme'}
+            showNumbers={currenFocusedSection === 'theme'}
           />
 
           {/* Scope Selection */}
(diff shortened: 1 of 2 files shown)


packages/cli/src/ui/components/ThemeDialog.tsx
```diff
@@ -196,6 +196,7 @@ export function ThemeDialog({
             onSelect={handleThemeSelect}
             onHighlight={onHighlight}
             isFocused={currenFocusedSection === 'theme'}
+            showNumbers={false}
```
miguelsolorio, line 199: Since the theme dialog has more than 9 entries, this can be tricky to solve so I've opted to disabled for now. We could re-enable later since there is multi-digit entry support, but unsure if this will be too long for the dialog.

PR #3943, opened 2025-07-12
Fix extra whitespace in markdown rendering.

### LINKED ISSUE #3942: Markdown display insert spurious whitespace
### What happened?

Repro steps:
Make request to gemini that contains markdown. Notice that there is excessive whitespace between blocks.
Most blocks have two lines of whitespace between them when they should have 1 and some blocks even have 3 lines of whitespace.

Expect to see no more than one line of whitespace between blocks in whitespace. This is a terminal application so users are expecting dense output.

Image

### Client information

<details>

```console
$ gemini /about
# paste output here
```

</details>

### Login information

_No response_

_No response_

jacob314:
## TLDR

Can't believe we didn't notice that markdown rendering had a ton of extra whitespace.
I'd looked at the markdown output many times but never noticed that there 2 and sometimes 3 lines of whitespace between entries rather than just 1.

## Dive Deeper

Here is how the markdown output from #3942 looks with this change applied. The previous output had 3 blank lines between these blocks instead of just 1.
Screenshot 2025-07-11 at 6 37 14 PM

* Fixed this by removing extra vertical padding around code blocks which isn't needed as all that matters is the horizontal indent.
* Ensuring that we only add at most one blank line between blocks.

## Reviewer Test Plan

Request your favorite markdown (for example bullet lists, code samples) and verify that it renders correctly.

## Linked issues / bugs

- Fixes #3942

### DIFF
--- a/packages/cli/src/ui/utils/MarkdownDisplay.test.tsx
+++ b/packages/cli/src/ui/utils/MarkdownDisplay.test.tsx
@@ -5,323 +5,172 @@
  */
 
 import { render } from 'ink-testing-library';
+import { describe, it, expect, vi, beforeEach } from 'vitest';
 import { MarkdownDisplay } from './MarkdownDisplay.js';
 
-describe('MarkdownDisplay', () => {
-  describe('Table Rendering', () => {
-    it('should render a simple table', () => {
-      const tableMarkdown = `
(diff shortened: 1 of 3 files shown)


scottdensmore (approved):
Approved

PR #3961, opened 2025-07-12
Docs: Refactor and Inline document

JasonHonKL:
## TLDR
Just refactor if statement to match case cause in the future we may need to match more pattern. 

## Dive Deeper

Some in line docs and refactor some code. 

## Reviewer Test Plan
N/A 

## Testing Matrix

No difference just refactor. 

## Linked issues / bugs
N/A, just refactor and make it looks better


abhipatel12:
Hey! Thanks for the contribution. At this time, we don't need to pre-emptively add a switch statement. If more cases are added we can consider it then. Thanks again!

PR #3967, opened 2025-07-12
Feat/home end key support

edithatogo:
## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

### DIFF
--- a/packages/cli/src/ui/components/InputPrompt.tsx
+++ b/packages/cli/src/ui/components/InputPrompt.tsx
@@ -341,6 +341,16 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
         return;
       }
 
+      if (key.name === 'home') {
+        buffer.move('home');
+        return;
+      }
+
+      if (key.name === 'end') {
+        buffer.move('end');
+        return;
+      }
+
       // Ctrl+A (Home) / Ctrl+E (End)
       if (key.ctrl && key.name === 'a') {
         buffer.move('home');
--- a/packages/cli/src/ui/components/messages/ToolConfirmationMessage.tsx
+++ b/packages/cli/src/ui/components/messages/ToolConfirmationMessage.tsx
@@ -139,7 +139,7 @@ export const ToolConfirmationMessage: React.FC<
         value: ToolConfirmationOutcome.ProceedOnce,
       },
       {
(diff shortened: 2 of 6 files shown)


packages/cli/src/ui/components/InputPrompt.tsx
```diff
@@ -341,6 +341,16 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
         return;
       }
 
+      if (key.name === 'home') {
```
jacob314, line 344: please add these to TextBuffer rather than InputPrompt as nothing about this is specific to the InputPrompt

packages/cli/src/ui/components/messages/ToolConfirmationMessage.tsx
```diff
@@ -139,7 +139,7 @@ export const ToolConfirmationMessage: React.FC<
         value: ToolConfirmationOutcome.ProceedOnce,
       },
       {
-        label: `Yes, allow always "${executionProps.rootCommand} ..."`,
```
jacob314, line 142: this is unrelated to keyboard shortcuts. please move this to a separate pull request as it is unrelatedto the keyboard shortcut feature.

packages/cli/src/utils/startupWarnings.ts
```diff
@@ -12,12 +12,20 @@ import { getErrorMessage } from '@google/gemini-cli-core';
 const warningsFilePath = pathJoin(os.tmpdir(), 'gemini-cli-warnings.txt');
 
 export async function getStartupWarnings(): Promise<string[]> {
+  const warnings: string[] = [];
```
jacob314, line 15: this is also unrelated. please revert or move to a separate pull request. I believe someone else has already sent out a pull request for this as well.

jacob314 (commented):
Thank your for sending this pull request. However for us to review we will need you to split the separate unrelated features into separate pull requests so it is easier to have separate conversations about each one with the right reviewers for each issue.

PR #3990, opened 2025-07-12
fix(ui): remove extraneous whitespace from startup screen

### LINKED ISSUE #3988: Gemini CLI initial screen has extra whitespace
### What happened?

Image

Notice that there are two lines of whitespace between different blocks in the ui

I would expect only a single line of whitespace between blocks like in the screenshot.
I would also expect 0 lines of whitespace between the prompt and status bar to leave maximum space for the content the user actually cares about.

Image

jacob314:
Remove extra whitespace lines from App.tsx

Removes two lines of whitespace that were appearing before and after the "Tips for getting started" section on the CLI startup screen. This was caused by a combination of a leading newline in the ASCII art and extra margins on the Header, Tips, and ContextSummaryDisplay components. This change removes the unnecessary whitespace to tighten up the UI. We accidentally had multiple lines of whitespace
at multiple places in App.tsx.

Adds a snapshot test for the initial rendering of the App.tsx component. This will help to prevent unintentional UI regressions.

With the changes the startup UI now looks like
Screenshot 2025-07-12 at 12 42 26 PM

## Linked issues / bugs
- Fixes #3988

### DIFF
--- a/packages/cli/src/ui/App.test.tsx
+++ b/packages/cli/src/ui/App.test.tsx
@@ -648,6 +648,38 @@ describe('App UI', () => {
     });
   });
 
+  it('should render the initial UI correctly', () => {
+    const { lastFrame, unmount } = render(
+      <App
+        config={mockConfig as unknown as ServerConfig}
+        settings={mockSettings}
+        version={mockVersion}
+      />,
+    );
+    currentUnmount = unmount;
+    expect(lastFrame()).toMatchSnapshot();
+  });
+
+  it('should render correctly with the prompt input box', () => {
+    vi.mocked(useGeminiStream).mockReturnValue({
+      streamingState: StreamingState.Idle,
+      submitQuery: vi.fn(),
+      initError: null,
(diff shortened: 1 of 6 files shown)


miguelsolorio:
Looks good, I agree that less white space is better, though I’d keep the bottom space between the input region and tips. When there are Gemini.md/mcp servers it looks too tight so maybe add some conditioning for that space.

NTaylorMullen (approved):
Nice!

NTaylorMullen:
> Looks good, I agree that less white space is better, though I’d keep the bottom space between the input region and tips. When there are Gemini.md/mcp servers it looks too tight so maybe add some conditioning for that space.

Fair!

OBITOONDEADO1MZ (approved):

miguelsolorio (approved):

jacob314:
Finally got back to this.
Added back the line that was a step too far.
Screenshot 2025-07-25 4 51 28 PM

PR #4078, opened 2025-07-14
Refac: Centralize storage file management

y-okt:
## TLDR

As [commented](https://github.com/google-gemini/gemini-cli/pull/4078#issuecomment-3099907823) by @NTaylorMullen , we'd like to centralize the way we manage files under `.gemini` directory.

## Dive Deeper

The same as TLDR

## Reviewer Test Plan

pull this branch and locally verify ~/.gemini 

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅   | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


packages/core/src/utils/migration.ts
```diff
@@ -0,0 +1,190 @@
...
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import * as fs from 'fs';
+import * as path from 'path';
+import * as os from 'os';
+import { GEMINI_DIR } from './paths.js';
+
+const INTERNAL_DIR = 'internal';
```
NTaylorMullen: We actually have a `tmp` directory that's also exposed on the config. How would you feel about moving all the contents there?

NTaylorMullen (commented):
@y-okt thank you for the contribution again!! Love what you've done and love that you've considered migrations. Had one suggestion to go from `internal` -> utilizing our centrally surfaced `tmp` dir

packages/core/src/utils/migration.ts
y-okt: @NTaylorMullen  Thank you for reviewing and the suggestion! I changed the path name. Could you review again?

NTaylorMullen:
@y-okt so I've been meandering this PR and the more I played with it the more I think there needs to be a bigger consolidation of pathing to make this reasonable. I started laying out the groundwork here: https://github.com/google-gemini/gemini-cli/compare/ntm/storage?expand=1

Gist: I worry that having another isolated mechanism for defining directories will further segment how we manage all of state files SO I started playing around with centralizing that. How would you feel about taking on a bigger work item / taking over that ntm/storage branch and finishing it out to feel nice & cohesive?

y-okt:
@NTaylorMullen Hi, thanks for the suggestion! I think centralizing the way we manage these is a great idea. I’d be happy to take on this task. I’ll mention you once it’s ready for review.

packages/core/src/config/storage.ts
```diff
@@ -0,0 +1,122 @@
...
+
+export class Storage {
+  private readonly targetDir: string;
+
+  constructor(targetDir: string) {
+    this.targetDir = targetDir;
+    this.ensureGlobalGeminiDirExists();
+    this.ensureGlobalTempDirExists();
+    this.ensureProjectTempDirExists();
+  }
```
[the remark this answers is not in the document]
y-okt: Addressed

packages/cli/src/services/FileCommandLoader.ts
```diff
@@ -133,11 +129,14 @@ export class FileCommandLoader implements ICommandLoader {
...
+    const globalStorage = new Storage(process.cwd());
+    const projectStorage = new Storage(this.projectRoot);
+
     // 1. User commands
-    dirs.push({ path: getUserCommandsDir() });
+    dirs.push({ path: globalStorage.getUserCommandsDir() });
 
     // 2. Project commands (override user commands)
-    dirs.push({ path: getProjectCommandsDir(this.projectRoot) });
+    dirs.push({ path: projectStorage.getProjectCommandsDir() });
```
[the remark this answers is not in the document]
y-okt: I don't think this is appropriate, as storage.getProjectCommandsDir should be called for project storage, not global storage.

packages/core/src/config/config.ts
```diff
@@ -351,6 +350,12 @@ export class Config {
     this.initialized = true;
     // Initialize centralized FileDiscoveryService
     this.getFileService();
+
+    const tmpDir = this.storage.getGlobalTempDir();
+    if (!fs.existsSync(tmpDir)) {
+      fs.mkdirSync(tmpDir, { recursive: true });
+    }
```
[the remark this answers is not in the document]
y-okt: Fixed

packages/core/src/config/storage.ts
```diff
@@ -0,0 +1,128 @@
...
+  getGoogleAccountsPath(): string {
+    return path.join(this.getGlobalGeminiDir(), GOOGLE_ACCOUNTS_FILENAME);
+  }
+
+  getHistoryDir(): string {
+    const hash = this.getFilePathHash(this.getProjectRoot());
+    const historyDir = path.join(this.getGlobalGeminiDir(), 'history');
+    fs.mkdirSync(historyDir, { recursive: true }); // Ensure it exists
+    return path.join(historyDir, hash);
+  }
```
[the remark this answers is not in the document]
y-okt: Done

packages/cli/src/ui/hooks/useGeminiStream.ts
```diff
@@ -113,6 +114,13 @@ export const useGeminiStream = (
     return new GitService(config.getProjectRoot());
   }, [config]);
 
+  const storage = useMemo(() => {
+    if (!config.getProjectRoot()) {
+      return;
+    }
+    return new Storage(config.getProjectRoot());
+  }, [config]);
```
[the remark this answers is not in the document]
y-okt: Done

packages/cli/src/ui/hooks/useGeminiStream.ts
```diff
@@ -877,8 +885,8 @@ export const useGeminiStream = (
       );
 
       if (restorableToolCalls.length > 0) {
-        const checkpointDir = config.getProjectTempDir()
-          ? path.join(config.getProjectTempDir(), 'checkpoints')
+        const checkpointDir = storage?.getProjectTempDir()
+          ? path.join(storage.getProjectTempDir(), 'checkpoints')
           : undefined;
```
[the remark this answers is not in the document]
y-okt: Resolved

packages/core/src/config/storage.ts
```diff
@@ -0,0 +1,134 @@
...
+      this.ensureGlobalGeminiDirExists();
+      this.ensureGlobalTempDirExists();
+      this.ensureProjectTempDirExists();
+      this.ensureGlobalHistoryDirExists();
+    } catch (error) {
+      throw new Error(
+        `Failed to create required Gemini directories. Please check permissions for your home and temp directories. Original error: ${error instanceof Error ? error.message : String(error)}`,
+      );
+    }
+  }
```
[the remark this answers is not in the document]
y-okt: While I agree that this is a side effect, this can prevent developers from forgetting to confirm the existence of the path.

packages/core/src/code_assist/oauth2.ts
```diff
@@ -17,16 +17,28 @@ import * as net from 'net';
...
+  return storageInstance;
+}
+
+let userAccountManagerInstance: UserAccountManager | null = null;
+function getUserAccountManager(): UserAccountManager {
+  if (!userAccountManagerInstance) {
+    userAccountManagerInstance = new UserAccountManager(getStorage());
+  }
+  return userAccountManagerInstance;
+}
```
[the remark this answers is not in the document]
y-okt: Fixed

packages/core/src/code_assist/oauth2.test.ts
```diff
@@ -44,17 +43,27 @@ const mockConfig = {
...
-  beforeEach(() => {
+  beforeEach(async () => {
     tempHomeDir = fs.mkdtempSync(
       path.join(os.tmpdir(), 'gemini-cli-test-home-'),
     );
     (os.homedir as Mock).mockReturnValue(tempHomeDir);
+    vi.spyOn(process, 'cwd').mockReturnValue(tempHomeDir);
+    vi.resetModules();
+    const mod = await import('./oauth2.js');
+    getOauthClient = mod.getOauthClient;
```
y-okt: This is because `oauth2.ts` needs process.cwd() for storage.ts, but by single mock, we can't give this value to `oauth2.ts`

packages/core/src/config/storage.ts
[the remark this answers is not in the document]
abhipatel12: I think we should consider this some more. Open to being convinced otherwise, but in general, agree with the gemini review here. This blocks the loop while the sync I/O calls are made.

packages/core/src/config/storage.ts
```diff
@@ -0,0 +1,137 @@
...
+
+  getGeminiDir(): string {
+    return path.join(this.targetDir, GEMINI_DIR);
+  }
+
+  getGlobalGeminiDir(): string {
```
abhipatel12: Can we refactor global path getters to be static methods? This allows consumers that need the global path to do so without creating an instance.

packages/core/src/config/storage.ts
```diff
@@ -0,0 +1,137 @@
...
+import * as path from 'node:path';
+import * as os from 'os';
+import * as crypto from 'crypto';
+import * as fs from 'fs';
+
+export const GEMINI_DIR = '.gemini';
+export const GOOGLE_ACCOUNTS_FILENAME = 'google_accounts.json';
+const TMP_DIR_NAME = 'tmp';
+
+export class Storage {
```
abhipatel12, line 16: A few different services instantiate `Storage` directly. Can we use dependency injection and have a class like `Config` store the single storage instance that then gets passed in via the constructors?

Ex. for logger.ts

```typescript
constructor(sessionId: string, private readonly storage: Storage) {
    this.sessionId = sessionId;
}
```

packages/core/src/utils/paths.ts
```diff
@@ -180,23 +179,6 @@ export function getProjectHash(projectRoot: string): string {
  * @returns The path to the project's temporary directory.
  */
 export function getProjectTempDir(projectRoot: string): string {
```
abhipatel12: Can we just remove this and have all callers use the new storage method directly?

packages/cli/src/config/settings.ts
```diff
@@ -38,7 +42,7 @@ export function getSystemSettingsPath(): string {
 }
 
 export function getWorkspaceSettingsPath(workspaceDir: string): string {
```
abhipatel12: Can we just have the caller of this use storage directly?

packages/cli/src/services/FileCommandLoader.ts
```diff
@@ -133,11 +129,14 @@ export class FileCommandLoader implements ICommandLoader {
   private getCommandDirectories(): CommandDirectory[] {
     const dirs: CommandDirectory[] = [];
 
+    const globalStorage = this.config?.storage ?? new Storage(process.cwd());
+    const projectStorage = new Storage(this.projectRoot);
```
abhipatel12: If our global paths are static, we could use a single instance here right?

packages/core/src/mcp/oauth-token-storage.ts
```diff
@@ -36,17 +36,14 @@ export interface MCPOAuthCredentials {
...
-
   /**
    * Get the path to the token storage file.
    *
    * @returns The full path to the token storage file
    */
   private static getTokenFilePath(): string {
-    const homeDir = os.homedir();
-    return path.join(homeDir, this.CONFIG_DIR, this.TOKEN_FILE);
+    const storage = new Storage(process.cwd());
```
abhipatel12: This is good example where the `getMcpOAuthTokensPath()` could just be a static function on `Storage`. In this case, we can avoid doing the `new Storage(process.cwd())`

packages/core/src/config/config.ts
```diff
@@ -327,6 +324,7 @@ export class Config {
     this.chatCompression = params.chatCompression;
     this.interactive = params.interactive ?? false;
     this.trustedFolder = params.trustedFolder;
+    this.storage = new Storage(this.targetDir);
```
abhipatel12, line 340: This is probably a dumb question, but why are there times that we instantiate a new Storage instance and then sometimes use the the one inside of config?

Do they have different paths for some reason?

abhipatel12 (commented):
Hey! This is a great refactoring to centralize the file path management. I left a few comments that I had as I reviewed. Some may be due to a lack of context on this area, so def let me know if that's the case!

packages/core/src/config/storage.ts
y-okt: Thanks! Refactored

packages/core/src/config/storage.ts
y-okt: Thanks, applied this to logger.ts, gitService.ts

packages/core/src/utils/paths.ts
y-okt: Removed, thanks

packages/cli/src/config/settings.ts
y-okt: Removed!

packages/cli/src/services/FileCommandLoader.ts
y-okt: That's correct, changed files

packages/core/src/mcp/oauth-token-storage.ts
y-okt: Yeah, thanks for your advice!

packages/core/src/config/config.ts
y-okt: This is a good point, actually we should utilize config.storage whenever accessible. I changed this.

packages/cli/src/ui/hooks/useLogger.ts
```diff
@@ -14,7 +14,7 @@ export const useLogger = () => {
   const [logger, setLogger] = useState<Logger | null>(null);
 
   useEffect(() => {
-    const newLogger = new Logger(sessionId);
+    const newLogger = new Logger(sessionId, new Storage(process.cwd()));
```
abhipatel12: Should we pass the storage to this hook via the config in `useGeminiStream` so that we have access to the storage instead of instantiating another?

abhipatel12 (approved):
Thanks for the work on this PR! Just one small comment, but this looks great! Going to go ahead and approve and pass to Taylor!

packages/cli/src/ui/hooks/useLogger.ts
y-okt: Thank you! Fixed it!

packages/core/src/config/storage.ts
```diff
@@ -0,0 +1,114 @@
...
+  private readonly targetDir: string;
+
+  constructor(targetDir: string) {
+    this.targetDir = targetDir;
+  }
+
+  static getGlobalGeminiDir(): string {
```
NTaylorMullen, line 23: Could be a follow-up don't want you to block on this BUT what do you think about not having any statics on this class and requiring that the instance of this type is passed to all consumers. The reason being is that in a world where we allow configuration of the global/user gemini dir then these values could change.

packages/core/src/config/storage.ts
```diff
@@ -0,0 +1,114 @@
...
+
+  static getOAuthCredsPath(): string {
+    return path.join(Storage.getGlobalGeminiDir(), 'oauth_creds.json');
+  }
+
+  getProjectRoot(): string {
+    return this.targetDir;
+  }
+
+  private getFilePathHash(filePath: string): string {
```
NTaylorMullen, line 81: nit: group privates/statics/implicit members so in this case this would go at the bottom so all the implicits are together

NTaylorMullen (approved):

packages/core/src/config/storage.ts
abhipatel12: So sorry Yuki! I didn't consider the configurable dir possibility here when I asked for the statics

y-okt: That makes sense, thank you very much for your insight! I'll address this in the next PR, since this PR is already large and prone to code conflicts

PR #4100, opened 2025-07-14
Use EnvHttpProxyAgent to make sure proxy setting from env is correctly used

### LINKED ISSUE #3340: add the NO_PROXY env variable feature to avoid proxy calling MCP servers  listening on localhost
### What happened?

I am using gemini-cli behind my company proxy that works only for internet websites and doesn't work for the local network or localhost.  I put the proxy url in the http_proxy and https_proxy enviroment variables in my .bat, It is ok only for the tools that go to the internet, unfortunately for the MCP servers that are listening on localhost or 127.0.0.1 those MCP servers don't work because there is no way to exclude localhost and 127.0.0.1 from the proxy. I use many nodejs apps and they all use the env variable NO_PROXY to exclude some address, that doesnt work for gemini-cli. So how can I use my MCP servers if I am behind a company proxy?

setting NO_PROXY=localhost  I expected gemini-cli working with MCP server or services on the localhost even if I am behind a proxy to navigate internet. 

### Client information

<details>

```console
$ gemini /about
 About Gemini CLI                                   │
│                                                    │
│ CLI Version       0.1.9                            │
│ Git Commit        34935d6                          │
│ Model             gemini-2.5-pro                   │
│ Sandbox           no sandbox                       │
│ OS                win32                            │
│ Auth Method       OAuth
```

</details>

### Login information

_No response_

_No response_

sylxjtu:
## TLDR

Use the undici [EnvHttpProxyAgent](https://github.com/nodejs/undici/blob/main/docs/docs/api/EnvHttpProxyAgent.md) to make sure http_proxy, https_proxy and no_proxy in the environment is used.

## Dive Deeper

Currently we manually fetched the `https?_proxy` env variable and set the proxy, other envs such as `no_proxy` and  is not 

## Reviewer Test Plan

Please add the following to the `settings.json`.

```
{
  "general": {
    "useEnvProxy": true
  }
}
```

Then set the environment variables to make sure it work. For example, if we use:
```
export no_proxy=127.0.0.1
export https_proxy=http://some.proxy.server
```
gemini cli should use the https proxy to call gemini api but use no proxy to call local mcp server.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | :heavy_plus_sign:   |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Resolves #3340

### DIFF
--- a/packages/cli/src/config/config.ts
+++ b/packages/cli/src/config/config.ts
@@ -599,6 +599,7 @@ export async function loadCliConfig(
       process.env['https_proxy'] ||
       process.env['HTTP_PROXY'] ||
       process.env['http_proxy'],
+    useEnvProxy: settings.general?.useEnvProxy,
     cwd,
     fileDiscoveryService: fileService,
     bugCommand: settings.advanced?.bugCommand,
--- a/packages/a2a-server/src/testing_utils.ts
+++ b/packages/a2a-server/src/testing_utils.ts
@@ -37,6 +37,7 @@ export function createMockConfig(
     setFlashFallbackHandler: vi.fn(),
     initialize: vi.fn().mockResolvedValue(undefined),
     getProxy: vi.fn().mockReturnValue(undefined),
+    getUseEnvProxy: vi.fn().mockReturnValue(false),
     getHistory: vi.fn().mockReturnValue([]),
     getEmbeddingModel: vi.fn().mockReturnValue('text-embedding-004'),
     getSessionId: vi.fn().mockReturnValue('test-session-id'),
--- a/packages/cli/src/config/config.test.ts
+++ b/packages/cli/src/config/config.test.ts
@@ -346,6 +346,22 @@ describe('loadCliConfig', () => {
     expect(config.getShowMemoryUsage()).toBe(true);
   });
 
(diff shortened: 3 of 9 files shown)


packages/cli/src/config/config.ts
```diff
@@ -175,6 +176,11 @@ export async function parseArguments(): Promise<CliArgs> {
       type: 'boolean',
       description: 'List all available extensions and exit.',
     })
+    .option('use-env-proxy', {
```
jacob314: rather than adding this to config.ts please add it to settings.json. Gemini CLI can help with that change.

jacob314 (commented):
Sorry for the slow review!
I'll approve this change once there is a minimal test for the change to client.ts and the config option is moved to settings.json.
Thanks for the PR addressing this issue impacting many users!

sylxjtu:
Thank you for reviewing, I added tests and moved the config to settings.

chrstnb:

PR #4135, opened 2025-07-14
Add support for allowed/excluded MCP server names in settings

chrstnb:
## TLDR

In lieu of a flag setting allowed MCP server names, add this as an option to settings so it can be managed by enterprises and users. Following the same scheme as coreTools/excludeTools.
 
## Dive Deeper

In a follow-up, we can remove the flag. Additionally, should we want to support more fine-grained definitions, we can accept either a string MCP server name OR an mcp server name mapped to an array of supported or unsupported tool names.

e.g.,

```
allowMCPServers: [
  "server1",
  { "server2": ['tool1', 'tool2'] }
]
```

## Reviewer Test Plan

Allow or exclude MCP servers in different settings.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | y  | ❓  | ❓  |
| npx      | y  | ❓  | ❓  |
| Docker   | y  | ❓  | ❓  |
| Podman   | y  | -   | -   |
| Seatbelt | y  | -   | -   |


smhendrickson:
The diff only appears to test server names, but the original description mentions per-tool allowlisting. Is this intentional?

swissspidy:
> The diff only appears to test server names, but the original description mentions per-tool allowlisting. Is this intentional?

It does mention this is something for a follow-up.

----

Stupid question, but assuming an enterprise config like `allowMCPServers: ['allowedServerName'],` how does this prevent someone from doing this:

```json
{
  "mcpServers": {
    "allowedServerName": {
      "command": "path/to/malicious/custom/mcp-server",
      "args": ["--arg1", "value1"],
      "env": {
        "API_KEY": "$MY_API_TOKEN"
      },
      "cwd": "./server-directory",
      "timeout": 30000,
      "trust": false
    }
  }
}
```

Like, shouldn't this be an allowlist of commands or URLs?

chrstnb:
@swissspidy it's a great point-- I think this is kind of a tricky space where it's quite difficult to ensure an airtight configuration. Open to other ideas on how to support these configurations, especially for enterprises! I'm not opposed to moving to using commands or URLs, but I think that is similarly bypass-able. One thing I'm thinking about is if enterprises are especially concerned, they could create an allowed set of MCP servers at the top level by setting `mcpServers`, and then users could opt in or out of that list using the allow/exclude mechanism at the user settings level. 

Open to other ideas as well.

smhendrickson:
> One thing I'm thinking about is if enterprises are especially concerned, they could create an allowed set of MCP servers at the top level by setting `mcpServers`, and then users could opt in or out of that list using the allow/exclude mechanism at the user settings level.
> 
> Open to other ideas as well.

We have to choose the threat model here. In this case, I think we should be able to trust the configuration (at least the global enterprise configuration, which overrides others) file. I do agree in practice the MCP name is not enough, although it is a starting point. For defense in depth, we may want to instead consider whether sandboxing solutions can provide some guarantees around which MCPs can be used.

NTaylorMullen (approved):

chrstnb:
Going to type up something longer to discuss some of this further. Submitting this to get the ball rolling.

Issue #3906, opened 2025-07-11
Create lightweight gemini-cli companion VS Code extension


skeshive:
### What would you like to be added?

Boilerplate VS code extension that will be used to pull context from & execute actions within the IDE

### Why is this needed?

VS Code's programmatic control and extensibility are managed through its Extensions API. To enable IDE Mode, we will leverage an extension.

theguy000:
from where I can download the extension

aborruso:
How to install it?

Thank you

skeshive:
We are still actively developing this feature and have not yet published the extension to the VS code marketplace. We're anticipating publishing the extension in the upcoming weeks!

PR #4147, opened 2025-07-14
[ide-mode] Stream notifications when the active file changes

chrstnb:
## TLDR

In the VSCode extension-based MCP server, send notifications when the active file changes. 

## Dive Deeper

Returns structured data with { filePath } for now, but we can add additional listeners and data in a follow-up.

Also now supports resumable sessions.

## Reviewer Test Plan

Test it with the inspector or any MCP client.

image

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Progress on #3906


packages/vscode-ide-companion/src/ide-server.ts
```diff
@@ -105,5 +158,50 @@ const createMcpServer = () => {
       }
     },
   );
+  server.registerTool(
```
skeshive: I think we could also implement this without explicitly registering a tool & just notifying all clients via transport.send() when the active file changes

what would be the use case for the model calling this tool?

packages/vscode-ide-companion/src/ide-server.ts
```diff
@@ -105,5 +158,50 @@ const createMcpServer = () => {
...
+      inputSchema: {},
+    },
+    (_, { sendNotification }) =>
+      new Promise((_, reject) => {
+        const sendActiveFile = async (
+          editor: vscode.TextEditor | undefined,
+        ) => {
+          if (!editor || !editor.document.uri.fsPath) {
+            await sendNotification({
+              method: 'notifications/message',
```
skeshive: can the method be ide/activeFileChanged or something similar?

packages/vscode-ide-companion/src/ide-server.ts
```diff
@@ -105,5 +158,50 @@ const createMcpServer = () => {
...
+              reject(e);
+            });
+          } else {
+            const { document } = editor;
+            await sendNotification({
+              method: 'notifications/message',
+              params: {
+                level: 'info',
+                data: {
+                  filePath: document.uri.fsPath,
```
skeshive: can we also pass back line number and column number nested in a cursor object to match the client schema:

cursor: z.object({
  line: z.number(),
  character: z.number(),
});

chrstnb: Was imagining we would add this in a follow-up when we're also listening for changes to selected text since this data is always empty after changing files. WDYT?

packages/vscode-ide-companion/src/ide-server.ts
chrstnb: How are you imagining this getting invoked? Are there any examples of this pattern in the MCP spec? I'm not opposed but feels a little wonky?

packages/vscode-ide-companion/src/ide-server.ts
skeshive: SGTM- will mark the cursor info as optional on the client-side!

packages/vscode-ide-companion/src/ide-server.ts
```diff
@@ -8,28 +8,80 @@ import * as vscode from 'vscode';
...
-    console.log('Received MCP request:', req.body);
+    const sessionId = req.headers['mcp-session-id'] as string | undefined;
+    let transport: StreamableHTTPServerTransport;
+
+    if (sessionId && transports[sessionId]) {
+      transport = transports[sessionId];
+    } else if (!sessionId && isInitializeRequest(req.body)) {
+      transport = new StreamableHTTPServerTransport({
+        sessionIdGenerator: () => randomUUID(),
+        onsessioninitialized: (newSessionId) => {
```
NTaylorMullen, line 45:
```suggestion
        onSessionInitialized: (newSessionId) => {
```

Do we own this bit?

packages/vscode-ide-companion/src/ide-server.ts
```diff
@@ -60,10 +126,13 @@ export async function startIDEServer(_context: vscode.ExtensionContext) {
 }
 
 const createMcpServer = () => {
-  const server = new McpServer({
-    name: 'vscode-ide-server',
-    version: '1.0.0',
-  });
+  const server = new McpServer(
+    {
+      name: 'vscode-ide-server',
```
NTaylorMullen: We should name this more gemini-cli-esc 😄

packages/vscode-ide-companion/src/ide-server.ts
```diff
@@ -8,28 +8,80 @@ import * as vscode from 'vscode';
...
 
-  const mcpServer = createMcpServer();
-  const transport = new StreamableHTTPServerTransport({
-    sessionIdGenerator: undefined,
-    enableJsonResponse: true,
-  });
+  const transports: { [sessionId: string]: StreamableHTTPServerTransport } = {};
 
-  mcpServer.connect(transport);
+  vscode.window.onDidChangeActiveTextEditor((editor) => {
```
NTaylorMullen: iirc this returns a disposable to help cleanup state in the IDE that we then need to return as part of the normal extension bootup in VSCode

NTaylorMullen (dismissed):

packages/vscode-ide-companion/src/ide-server.ts
chrstnb: no unfortunately! it's part of the mcp SDK

packages/vscode-ide-companion/src/ide-server.ts
```diff
@@ -8,28 +8,81 @@ import * as vscode from 'vscode';
...
-  const mcpServer = createMcpServer();
-  const transport = new StreamableHTTPServerTransport({
-    sessionIdGenerator: undefined,
-    enableJsonResponse: true,
-  });
+  const transports: { [sessionId: string]: StreamableHTTPServerTransport } = {};
 
-  mcpServer.connect(transport);
+  const disposable = vscode.window.onDidChangeActiveTextEditor((editor) => {
+    const filePath = editor ? editor.document.uri.fsPath : null;
```
skeshive, line 24: instead of null, can we send an empty string here? that way, on the client we can distinguish between the user closing all windows vs. the server not sending any notification yet

packages/vscode-ide-companion/src/ide-server.ts
```diff
@@ -8,28 +8,81 @@ import * as vscode from 'vscode';
...
+    let transport: StreamableHTTPServerTransport;
+
+    if (sessionId && transports[sessionId]) {
+      transport = transports[sessionId];
+    } else if (!sessionId && isInitializeRequest(req.body)) {
+      transport = new StreamableHTTPServerTransport({
+        sessionIdGenerator: () => randomUUID(),
+        onsessioninitialized: (newSessionId) => {
+          transports[newSessionId] = transport;
+          const editor = vscode.window.activeTextEditor;
```
skeshive, line 47: this won't actually send the notification yet- IIUC the SDK correctly, first the client will connect to the server via /mcp POST and then it sends a /mcp GET request to start listening to the SSE

we should instead send the notification when we get the first /mcp GET request (ok to address this in a follow-up)

skeshive (approved):

mattKorwel (approved):

packages/vscode-ide-companion/src/ide-server.ts
chrstnb: Will address this and the other in a follow-up! FWIW this also gets sent between open files. We can add debouncing if reasonable

PR #4154, opened 2025-07-14
Adds the user's active file in the IDE to the footer

### LINKED ISSUE #4138: Show current active file in the UI

The user's current active file in the IDE should be displayed in the UI.

### Why is this needed?
The user should be aware when gemini-cli is using context from the IDE.

### Additional context

_No response_

skeshive:
## TLDR

Adds the user's active file in the IDE to the footer if IDE mode is active.

## Dive Deeper
* Updates mcp-client to add a notification handler for the _ide_server that triggers whenever the extension detects the active file changing
* Propagating this context to the system prompt will be done in a follow-up
* Displays the current active file in the footer:
Screenshot 2025-07-14 at 4 10 56 PM

## Reviewer Test Plan
- Build companion VS code extension using https://github.com/google-gemini/gemini-cli/pull/4147
- Run gemini-cli w/ --ide-mode
- Change file in VS code 
- Updated file name should show up in the UI

## Linked issues / bugs
Resolves #4138

### DIFF
--- a/packages/cli/src/ui/components/Footer.tsx
+++ b/packages/cli/src/ui/components/Footer.tsx
@@ -54,6 +54,13 @@ export const Footer: React.FC<FooterProps> = ({
   );
 
   useEffect(() => {
+    const updateActiveFile = () => {
+      const currentActiveFile = ideContext.getActiveFileContext();
+      setActiveFile(currentActiveFile);
+    };
+
+    updateActiveFile();
+
     const unsubscribe = ideContext.subscribeToActiveFile(setActiveFile);
     return () => {
       unsubscribe();
@@ -83,10 +90,10 @@ export const Footer: React.FC<FooterProps> = ({
               {shortenPath(tildeifyPath(activeFile.filePath), 70)}
             </Text>
             {activeFile.cursor && (
-                <Text color={Colors.Gray}>
-                  :{activeFile.cursor.line}:{activeFile.cursor.character}
-                </Text>
-              )}
+              <Text color={Colors.Gray}>
+                :{activeFile.cursor.line}:{activeFile.cursor.character}
+              </Text>
+            )}
           </Text>
         )}
         {debugMode && (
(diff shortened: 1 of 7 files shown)


packages/cli/src/ui/components/Footer.tsx
```diff
@@ -43,6 +49,17 @@ export const Footer: React.FC<FooterProps> = ({
   const limit = tokenLimit(model);
   const percentage = promptTokenCount / limit;
 
+  const [activeFile, setActiveFile] = useState<ActiveFile | undefined>(
+    undefined,
+  );
```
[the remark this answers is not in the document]
skeshive: updated

packages/cli/src/ui/components/Footer.tsx
```diff
@@ -59,6 +83,19 @@ export const Footer: React.FC<FooterProps> = ({
             {branchName && <Text color={Colors.Gray}> ({branchName}*)</Text>}
           </Text>
         )}
+        {activeFile && activeFile.filePath && (
+          <Text>
+            <Text color={Colors.Gray}> | </Text>
```
chrstnb, line 88: wonder if there's a better way to display this than with the post for UI consistency since currently the items in the footer are separated by whitespace. also worth making this a different color than the workspace?

chrstnb (approved):

packages/cli/src/ui/components/Footer.tsx
skeshive: yeah I think you're right- I was originally grouping this with the directory but it doesn't stand out enough

NTaylorMullen: Ya I definitely think there's probably more here we should do. Grouping it in this way feels too additive. @miguelsolorio what would you do?

NTaylorMullen (approved):
Approving for now but I imagine we'll need to change how this is visualized as well.

packages/cli/src/ui/components/Footer.tsx
skeshive: ack I'll address the UX update in a fast-follow

PR #4158, opened 2025-07-14
Make @ command sort file without extension name

SandyTao520:
## TLDR

Try to sort without extension first, this ensures test like a.test.ts doesn't rank before a.ts

## Dive Deeper

It's quite annoying that after pressing @ commands and searching for a file, when a file a.ts with a test that has a.test.ts pattern is ordered above a.ts. I have to arrow down every time I searched for a file.

Sorting without extension ensures files with tests like this are ordered before their tests.

Before:
image

After:
image

## Reviewer Test Plan

Press @ in gemini-cli directory and search any file that has a test, you should see the test of a file is ranked below the file itself.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ✅   | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |

## Linked issues / bugs

N/A

### DIFF
--- a/packages/cli/src/ui/hooks/useCompletion.test.ts
+++ b/packages/cli/src/ui/hooks/useCompletion.test.ts
@@ -1,0 +1,898 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { describe, it, expect, beforeEach, vi, afterEach } from 'vitest';
+import type { Mocked } from 'vitest';
+import { renderHook, act } from '@testing-library/react';
+import { useCompletion } from './useCompletion.js';
+import * as fs from 'fs/promises';
+import { glob } from 'glob';
+import { CommandContext, SlashCommand } from '../commands/types.js';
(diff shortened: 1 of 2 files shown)


packages/cli/src/ui/hooks/useCompletion.test.ts
```diff
@@ -0,0 +1,947 @@
...
+        await new Promise((resolve) => setTimeout(resolve, 150));
+      });
+
+      expect(result.current.suggestions).toHaveLength(6);
+
+      // Extract labels for easier testing
+      const labels = result.current.suggestions.map((s) => s.label);
+
+      // Find indices of source files and their corresponding test files
+      const componentIndex = labels.indexOf('component.ts');
```
jacob314: nit: would be simpler to just verify the sorted order is exactly what you expect. That way what the test is asserting is clearer.

jacob314 (dismissed):
lgtm

NTaylorMullen (approved):

PR #4162, opened 2025-07-14
bug(ux): update context percentage when /clear command is run

nsalerni:
Update context percentage when /clear command is run

## TLDR

Previously after running `/clear` you would only see the decrease in context window after running a subsequent prompt. This PR changes the behavior to update the context window as soon as you run `/clear`.

See #4142 for a full example scenario, or try it yourself with the example below with HEAD of this repo.

## Reviewer Test Plan

- Pull the change
- Run locally with `npm start`
- Load the context with a lot of information (normally asking to load all tests in this repo is sufficient to consume 30-40% of the context window).
- Run the `/clear` command
- Context used percentage should go back to 100%.
- Run a simple command like 'hello' and you should see that context window slowly fill up again.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ✅  | -   | -   |

## Linked issues / bugs

Resolves #4142


abhipatel12:
Quick thought on this, is it possible to have the UITelemetryService, listen for another type of `UiEvent` and then reset the statistic without the CLI package calling for a direct reset of the state?

mattKorwel:
> Quote reply

@abhipatel12 do you think this thought is blocking?

abhipatel12:
@mattKorwel No I don't think it's blocking b/c the UX is technically broken right now for this flow, and this does fix it. My biggest concern is having the CLI package reaching into a core service to change its state directly. 

I would suggest at least one unit test for the new behavior before giving it a LGTM

mattKorwel:
>My biggest concern is having the CLI package reaching into a core service to change its state directly.

Interesting. I would worry that an event listener might cause some sort of concurrency issue. I don't think calling the core lib directly is bad, though I'm not familiar with the over all telemetry abstraction so if we are violating something there I would agree.

@nsalerni add some unit tests around the changes in core and we'll :shipit: Since @abhipatel12 is good with that.

packages/core/src/telemetry/uiTelemetry.ts
```diff
@@ -133,6 +133,14 @@ export class UiTelemetryService extends EventEmitter {
     return this.#lastPromptTokenCount;
   }
 
+  resetLastPromptTokenCount(): void {
```
mattKorwel, line 136: Per the main discussion wrap some unit tests here and we're good to go. marking as needs changes in case others miss this convo.

mattKorwel (changes requested):
Generally looks good thanks for taking the time to update this! Requesting changes for unit tests and then we're good!

packages/core/src/telemetry/uiTelemetry.ts
nsalerni: Updated with tests for this behavior.

nsalerni:
Thanks @mattKorwel and @abhipatel12 ! If we want to go the event-based approach for this, I'd be happy to take it as a follow up.

mattKorwel (approved):
Lgtm!

PR #3919, opened 2025-07-11
Introduce loop detection service that breaks simple loop

SandyTao520:
## TLDR

Introduce a loop detection service in core package to detect and halt infinite loops caused by 1) repeated tool calls that share the same name and arguments, and 2) model chanting the same sentence nonstop.

## Dive Deeper

Added the simplest and safest approach to stop loops. These loops typically came as
1. a sequence of ``ReadFile`` attempts on a continuously failing file
2. ``WriteFile`` operations yielding "No changes detected,"
3. ``Edit`` commands that repeatedly fail to locate the specified string for replacement.
4. Model chanting (repeats the same response) non stop.

What is common in all these instances is an exact match in the tool's name, arguments. Recognizing this specific pattern allows us to proactively intervene and break the loop.

When these conditions are met, we interrupt the current turn and trigger a ``LoopDetected`` event, which notifies the user via ``useGeminiStream``:
1. Repeated Tool Calls: If we detect 5 consecutive instances of an identical tool call (i.e., the same tool name with the same arguments).
2. Repetitive Model Output: If the model repeatedly outputs the same sentence. We check for this by determining if the last sentence has appeared more than 10 times within the previous ``MAX_LOOPBACK_WINDOW`` (set to 1000 characters).

Detector is reset when user sends a new prompt (``prompt_id`` changes).

Future enhancements will focus on expanding this logic to identify and mitigate more intricate and extensive looping behaviors.

## Reviewer Test Plan
Turn on auto accept edits.

Prompt I use to trigger a WriteFile loop:
```
Use 1 single WriteFile call to write a.txt file of 100 lines from 1 to 100. Once you received the response, make another exact same single WriteFile tool call to write a file of 10 lines from 1 to 10. Repeat this for a total of 10 times.
```
It should stop after the 5th no-change WriteFile attempt.

image

Prompt I use to trigger chanting:
```
say 'Never gonna give you up, never gonna let you down...' 20 times
```
You should see this sentence at most 10 times (might be incomplete).

image

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| -------- | --- | --- | --- |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ✅  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ✅  | -   | -   |

## Linked issues / bugs

#3929 
#3934

PR #4193, opened 2025-07-15
Log the 2 types of loop detection

SandyTao520:
## TLDR

Introduce telemetry for loop detection. It adds logging for two types of loops: tool call loops and content loops. 

## Dive Deeper

The core of the changes are in ``LoopDetectionService``, which now logs an event when a loop is detected. A new ``LoopDetectedEvent`` is introduced in the telemetry service to handle these events. The two types of loops are:

- Tool Call Loops: Triggered when the same tool is called 5 times in a row with the same parameters.
- Content Loops: Triggered when the same content is generated 10 times in a row.

The ``GeminiClient`` is updated to pass the necessary configuration to the ``LoopDetectionService``.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ✅  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ✅  | -   | -   |

## Linked issues / bugs

#3919

### DIFF
--- a/packages/core/src/telemetry/types.ts
+++ b/packages/core/src/telemetry/types.ts
@@ -246,6 +246,23 @@ export class FlashFallbackEvent {
   }
 }
 
+export enum LoopType {
+  TOOL_CALL = 'tool_call',
+  CONTENT = 'content',
+}
+
+export class LoopDetectedEvent {
+  'event.name': 'loop_detected';
+  'event.timestamp': string; // ISO 8601
+  loop_type: LoopType;
+
+  constructor(loop_type: LoopType) {
+    this['event.name'] = 'loop_detected';
+    this['event.timestamp'] = new Date().toISOString();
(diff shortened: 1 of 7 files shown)


packages/core/src/telemetry/types.ts
```diff
@@ -246,6 +246,23 @@ export class FlashFallbackEvent {
   }
 }
 
+export enum LoopType {
+  TOOL_CALL = 'tool_call',
```
NTaylorMullen: We're going to have a number of tool call type loops. We should make this a bit more specific. Ditto with the Content one below

NTaylorMullen (dismissed):

NTaylorMullen (approved):

PR #4227, opened 2025-07-15
Improve altName completion behavior in slash commands

### LINKED ISSUE #4201: `/?` does not execute `/help` as expected
### What happened?

Typing `/?` seems to act as a shortcut for `/help`, but pressing Enter only autocompletes it to `/help`—it doesn't execute the command. As a result, the user has to press Enter twice: `/?` → Enter → `/help` → Enter.

Pressing Enter after `/?` should immediately execute the `/help` command.

### Client information

*   **CLI Version:** 0.1.12
*   **Git Commit:** b29c02dd (local modifications)
*   **Operating System:** linux v20.19.3
*   **Memory Usage:** 173.6 MB

### Login information

_No response_

`/?` appears to be a shortcut for `/help` in code, but the current behavior only triggers autocomplete, not execution, which can be confusing for users.

Lee-WonJun:
## TLDR

Fix altName commands (like `/?`) to execute immediately instead of showing autocomplete suggestions. When users type a complete altName like `/?` or `/usage` and press Enter, the command now executes directly rather than requiring a second Enter keypress.

## Dive Deeper

Previously, typing `/?` (an altName for `/help`) would show an autocomplete suggestion for `/help` instead of executing the command immediately. This created a confusing UX where users had to press Enter twice: `/?` → Enter → `/help` → Enter.

The fix involves two key changes:

* If a user partially types an altName (e.g. `/usag`), suggestions should continue to appear — this is intentional.
* However, **if a full altName is typed** (like `/?` or `/usage`), it should behave exactly the same as typing a full command name and **execute immediately without suggestions**.

This change ensures consistency between full `name` matches and full `altName` matches.

## 1. Perfect Match Logic

The completion logic was updated to treat full altName matches the same as full command name matches.
Previously, the perfect match logic only checked `s.name === partial`. Now, it also checks `s.altName === partial`.
This simple expansion ensures that full altName inputs (like `/?`) are recognized as exact matches and **trigger immediate command execution** without further interaction.

## 2. Test Coverage for altName Behavior

Test cases were added and updated to cover this new behavior:
* Full altNames like `/?` and `/usage` **execute immediately** and do **not** show suggestions.
* Partial altNames like `/usag` **still show autocomplete suggestions** as intended.

---
## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ✅(WSL)  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes #4201

##  Video Explanation

https://github.com/user-attachments/assets/7b301138-58d9-4813-ba71-9c05e9167a3d

The **left** shows the original behavior, the **right** shows the fix.

1. First, when partially typing an altName (like `/usa`), suggestions appear as expected.
2. Second and third, when fully typing `/usage` or `/?`, the original behavior doesn't execute the command on Enter — it turns into a suggestion instead.

This feels unintended because:
Typing `"/?" + Enter` changes it to a suggestion,
but typing `"/? " + Enter` runs the command directly.
That inconsistency seems wrong to me.


jacob314 (approved):
Thank you for this polish! The previous behavior was annoying and this is the sort of pull request that is really nicely self contained and easy for us to review and approve. 
lgtm

scottdensmore (approved):
Approved

PR #4240, opened 2025-07-15
Include api key in header instead of in the URL

SandyTao520:
## TLDR

For security, it's better not to include the key parameter in the URL.

## Dive Deeper

Per https://cloud.google.com/docs/authentication/api-keys-use#using-with-rest, it's better to include API key in the header.

## Reviewer Test Plan

Use Sherlog or Proxy to see the model check request no longer includes the key parameter.

image

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ✅  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ✅  | -   | -   |

## Linked issues / bugs

N/A

### DIFF
--- a/packages/core/src/core/modelCheck.ts
+++ b/packages/core/src/core/modelCheck.ts
@@ -28,7 +28,7 @@ export async function getEffectiveModel(
 
   const modelToTest = DEFAULT_GEMINI_MODEL;
   const fallbackModel = DEFAULT_GEMINI_FLASH_MODEL;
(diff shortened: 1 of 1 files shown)


mattKorwel (approved):
This seems like a solid change. It does seem like the kind of thing that might expose an edge case? Lets just make sure to watch for bugs related to this. LGTM

PR #4255, opened 2025-07-15
Set port dynamically in VSCode extension and read from it in gemini-cli and send initial notification

### LINKED ISSUE #3918: Generate companion MCP server port dynamically + write to env variable

chrstnb:
## TLDR

Dynamically generates a port and writes it to IDE terminals. Note that when/if we first install the extension, the user will need to re-start the terminal or create a new one or reload vscode to pick up the changes.

Also sends a notification after spinning up.

## Reviewer Test Plan

Run extension locally and verify that when you spin up gemini CLI in IDE mode, it is able to connect to the MCP server.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Resolves #3918


packages/vscode-ide-companion/src/ide-server.ts
```diff
@@ -113,16 +120,22 @@ export async function startIDEServer(context: vscode.ExtensionContext) {
...
-  app.listen(PORT, (error?: Error) => {
+  // Dynamically import 'get-port' as it is an ESM-only package.
+  const { default: getPort } = await import('get-port');
+  const port = await getPort();
+  context.environmentVariableCollection.replace(
+    'GEMINI_CLI_IDE_SERVER_PORT',
+    port.toString(),
+  );
+
+  app.listen(port, (error?: Error) => {
```
skeshive: instead of getting a port first and then starting up the server, you can pass in 0 to the app.listen call and the server will start up on a dynamic port, you can write the env var where we log the successful start up call

(that way, the env var will only be set if the server starts up with no errors)

packages/vscode-ide-companion/src/ide-server.ts
```diff
@@ -101,6 +103,11 @@ export async function startIDEServer(context: vscode.ExtensionContext) {
     }
 
     const transport = transports[sessionId];
+    if (!sessionsWithInitialNotification.has(sessionId)) {
```
skeshive: this needs to be done after the handleRequest call below because the handleRequest call is what starts up the SSE stream (see https://github.com/modelcontextprotocol/typescript-sdk/blob/main/src/server/streamableHttp.ts#L301)

packages/cli/src/config/config.ts
```diff
@@ -286,13 +286,20 @@ export async function loadCliConfig(
   }
 
   if (ideMode) {
+    const companionPort = process.env.GEMINI_CLI_IDE_SERVER_PORT;
+    if (!companionPort) {
+      throw new Error(
+        'The GEMINI_CLI_IDE_SERVER_PORT environment variable must be set when running in IDE mode. Try running in a fresh terminal.',
```
skeshive: hmm we shouldn't expose the GEMINI_CLI_IDE_SERVER_PORT env var to the user- they may try setting it themselves or get confused

hb something like could not run in ide mode, make sure you're running in vs code integrated terminal. try running in a fresh terminal?

skeshive (dismissed):
nit: can you also update the commit message to include the notification changes

mattKorwel (approved):
LGTM

PR #4256, opened 2025-07-15
Run model availability check in the background to speed up startup

SandyTao520:
## TLDR

This change makes the check for model availability and the subsequent fallback to a different model an asynchronous, non-blocking operation. Previously, the CLI would wait for the model check to complete before starting up, which could cause a noticeable delay. Now, the check happens in the background, leading to a faster startup. If the user's configured model is not available, the system will automatically switch to a compatible fallback model and notify the user without interrupting their workflow.

## Dive Deeper

The primary motivation for this change is to improve the perceived performance and responsiveness of the CLI. The previous implementation of `refreshAuth` would `await` the `createContentGeneratorConfig` function, which in turn awaited an async call to `getEffectiveModel`. This created a blocking sequence during the application's startup or authentication refresh.

The new implementation refactors `createContentGeneratorConfig` to be a synchronous function that immediately returns a configuration object. It then initiates the model availability check in the background using a promise chain (`.then()`).

If the check determines that the user's preferred model is not available for their API key, it now:
1. Updates the model in the central `Config` object via `config.setModel()`.
2. Triggers a `config.flashFallbackHandler()` callback. This handler is responsible for notifying other parts of the application, such as the UI, that the model has been changed.

This ensures the application remains interactive while these checks are performed, providing a smoother user experience.

### Scenario 1: Model Fallback
1. Use a API-key that is rate limited or manually modified the code to always return the fallback model.
2. Start the Gemini CLI.
3. **Expected Behavior:** The CLI should start up almost instantly, initially displaying that it's using your chosen (unavailable) model.
4. Within a few seconds, you should see model name changed to `gemini-2.5-flash`.
5. Run a prompt. It should execute successfully using the fallback model.

### Scenario 2: Model is Available (No Fallback)
1. Configure the CLI with a model that you have access to (e.g., `gemini-2.5-pro`).
2. Start the Gemini CLI.
3. **Expected Behavior:** The CLI should start up quickly. No fallback message should be displayed.
4. Run a prompt. It should execute successfully with your configured model.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅ | ❓  | ❓  |
| npx      | ✅ | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ✅ | -   | -   |

## Linked issues / bugs

N/A


packages/cli/src/ui/App.tsx
```diff
@@ -294,64 +295,70 @@ const App = ({ config, settings, startupWarnings = [], version }: AppProps) => {
...
-      const isPaidTier =
-        userTier === UserTierId.LEGACY || userTier === UserTierId.STANDARD;
-
-      // Check if this is a Pro quota exceeded error
-      if (error && isProQuotaExceededError(error)) {
-        if (isPaidTier) {
-          message = `⚡ You have reached your daily ${currentModel} quota limit.
+      if (
+        config.getContentGeneratorConfig().authType ===
+        AuthType.LOGIN_WITH_GOOGLE
```
NTaylorMullen, line 300: Hmmm, we may want to be careful with this change. You can still experience slow downs with API keys etc. and have flash fallback

SandyTao520: For API keys, this only happens during startup. This function is only called when using LOGIN_WITH_GOOGLE (http://github.com/google-gemini/gemini-cli/blob/main/packages/core/src/core/client.ts#L624).

Do you think we should do this for API keys as well?

NTaylorMullen: Ah I hadn't realized we had that restriction actually. Forget me :)

NTaylorMullen (approved):

PR #4264, opened 2025-07-15
[ide-mode] Thread active file through to system prompt

### LINKED ISSUE #4164: Propagate user's active file to the system prompt

chrstnb:
## TLDR

When in IDE mode, thread the active file through to the system prompt.
## Dive Deeper

Want to get this checked in so things are working e2e-- will run evals and do manual evals as a follow-up.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Resolves #4164

### DIFF
--- a/packages/core/src/core/prompts.ts
+++ b/packages/core/src/core/prompts.ts
@@ -17,8 +17,13 @@ import { WriteFileTool } from '../tools/write-file.js';
 import process from 'node:process';
 import { isGitRepository } from '../utils/gitUtils.js';
 import { MemoryTool, GEMINI_CONFIG_DIR } from '../tools/memoryTool.js';
+import { Config } from '../config/config.js';
+import { ideContext } from '../services/ideContext.js';
 
-export function getCoreSystemPrompt(userMemory?: string): string {
+export function getCoreSystemPrompt(
+  config: Config,
+  userMemory?: string,
+): string {
   // if GEMINI_SYSTEM_MD is set (and not 0|false), override system prompt from file
   // default path is .gemini/system.md but can be modified via custom path in GEMINI_SYSTEM_MD
   let systemMdEnabled = false;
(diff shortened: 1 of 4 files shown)


skeshive (approved):

packages/core/src/core/prompts.ts
```diff
@@ -153,7 +158,25 @@ ${(function () {
   }
   return '';
 })()}
-
+${(function () {
```
mattKorwel, line 161: Anything that messes with the core system prompt makes me nervous. Think this is good. happy to approve and get rolling, lets be open to revisiting if need be based on any follow on feedback?

mattKorwel (approved):
LGTM

packages/core/src/core/prompts.ts
```diff
@@ -153,7 +158,25 @@ ${(function () {
...
+  if (config.getIdeMode()) {
+    const activeFile = ideContext.getActiveFileContext();
+    if (activeFile?.filePath) {
+      let prompt = `
+# IDE Mode
+You are running in IDE mode. The user has the following file open:
```
NTaylorMullen, line 167: Ah you may want to change how this is done. Ultimately this should be diff-based and should be represented as a synthetic user message. Meaning, whenever the user changes their active file path we should probably create a history message that indicates that change. Otherwise, having this as part of the system prompt will lead to historical inconsistencies where we'll be effectively encouraging the model to hallucinate because it will be reading an active file based on 0 information.

So to get more crisp. Lets say the current user history is empty and user sends request "Hello A":

History: ["active file: A.txt...", "Hello A"]

User swaps to B.txt and then to C.txt and asks "What about C?"
History: ["active file: A.txt", "Hello A", "Response to Hello A", "active file: C.txt", "What about C?"]

anj-s: +1 I think this should be part of the context that we provide to the user when we give it a task. I don't know if it makes sense to add it to the history but instead provide it via the prompt as file context.
I would almost say that it should be something that is part of the prompt that says:
Here is the repo, structure and current active files in the IDE. how can we execute on the user's request.

chrstnb: Proposal here, which I think is in line with what @anj-s is proposing https://github.com/google-gemini/gemini-cli/pull/4312

PR #4265, opened 2025-07-15
Add /ide status & /ide install commands to manage IDE integration

### LINKED ISSUE #4166: Add /ide slash commands
The /ide slash command should support the following sub-commands:

/ide status --> Shows the status of the IDE integration (connected, disconnected, pending)
/ide install --> Installs the companion VS Code extension (since the extension isn't published yet, we'll need to build a local vsix and install that)

skeshive:
## TLDR

Add /ide status & /ide install commands to manage IDE integration. For now, the /ide is feature gated by the --ide-mode flag. In the future, we'll update the /ide slash command to show a error message if the user is not running in the VS Code integrated terminal. 

## Dive Deeper

/ide status --> Shows the connection status of the IDE server running in the VS code companion extension
/ide install --> Attempts to install the VS code companion extension for the user. Since the extension is not published to marketplace yet, we first attempt to fetch the built VSIX from the bundled files (for users running via npx) and then fallback to fetching it from the local directory (for users running via npm start). Once we publish to the marketplace, we'll install directly from there. 

## Linked issues / bugs
Resolves #4166


packages/cli/src/ui/commands/ideCommand.ts
```diff
@@ -0,0 +1,164 @@
...
+          const bundleDir = path.dirname(fileURLToPath(import.meta.url));
+          // The VSIX file is copied to the bundle directory as part of the build.
+          let vsixFiles = glob.sync(path.join(bundleDir, '*.vsix'));
+          if (vsixFiles.length === 0) {
+            // If the VSIX file is not in the bundle, it might be a dev
+            // environment running with `npm start`. Look for it in the original
+            // package location, relative to the bundle dir.
+            const devPath = path.join(
+              bundleDir,
+              '..',
```
NTaylorMullen, line 113: lol all the `..`

packages/cli/src/ui/commands/ideCommand.ts
```diff
@@ -0,0 +1,164 @@
...
+              '..',
+              '..',
+              '..',
+              '..',
+              'vscode-ide-companion',
```
NTaylorMullen: Lets put this magic string somewhere if we could as a constant

packages/cli/src/ui/commands/ideCommand.ts
```diff
@@ -0,0 +1,164 @@
...
+              '*.vsix',
+            );
+            vsixFiles = glob.sync(devPath);
+          }
+          if (vsixFiles.length === 0) {
+            context.ui.addItem(
+              {
+                type: 'error',
+                text: 'Could not find the required VS Code companion extension.',
```
NTaylorMullen: Lets make this more "fatal" and suggest the user file a bug via `/bug` immediately.

NTaylorMullen (dismissed):

packages/cli/src/ui/commands/ideCommand.ts
skeshive: its ugly but this logic will be deprecated as soon as we publish the extension so I didn't bother coming up with something better

packages/cli/src/ui/commands/ideCommand.ts
skeshive: done

packages/cli/src/ui/commands/ideCommand.ts
skeshive: done

NTaylorMullen (approved):

PR #4312, opened 2025-07-16
[ide-mode] Add active file to user model request

chrstnb:
## TLDR

In lieu of adding the active file info to the system prompt (in #4264, rolled back in #4308), add it to the request sent to the model with the user message.

## Dive Deeper

This is in response to @anj-s's suggestion in the other PR https://github.com/google-gemini/gemini-cli/pull/4264#discussion_r2208878953. 

## Reviewer Test Plan

Run in IDE mode and confirm the active file is accessible to the model.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

#4164


NTaylorMullen:
@anj-s I saw in your other comment that you weren't worried about it being added to history or not. Don't you think that will skew the history representation?

packages/core/src/core/client.ts
```diff
@@ -303,6 +304,24 @@ export class GeminiClient {
     if (compressed) {
       yield { type: GeminiEventType.ChatCompressed, value: compressed };
     }
+
+    if (this.config.getIdeMode()) {
+      const activeFile = ideContext.getActiveFileContext();
+      if (activeFile?.filePath) {
+        let context = `
+Here is information about the user's current IDE context. Use this if relevant to the user's query:
```
anj-s: I would say "These are the files that the user was most recently looking at:"  I don't know if the agent actually understands "IDE context" and if we have defined that somewhere.

anj-s: Also mention that you are providing cursor position.

anj-s (dismissed):

anj-s (approved):

PR #4320, opened 2025-07-16
Add numbers to selection list

miguelsolorio:
## TLDR

Adds numbers to selection list for better accessibility support

## Dive Deeper

When the CLI shows a list of options, we now show numbers next to each selection. For numbers larger than 10, there is a small delay to wait for an additional number.

CleanShot 2025-07-17 at 15 12 11@2x

https://github.com/user-attachments/assets/d59b772f-b01b-47e4-bb04-6e4d5933911b

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes https://github.com/google-gemini/gemini-cli/issues/1776


packages/cli/src/ui/components/shared/RadioButtonSelect.tsx
```diff
@@ -71,30 +76,81 @@ export function RadioButtonSelect<T>({
...
+    },
+    [],
+  );
+
   useInput(
     (input, key) => {
+      const isNumeric = showNumbers && /^[0-9]$/.test(input);
+
+      // Any key press that is not a digit should clear the number input buffer.
+      if (!isNumeric && numberInputTimer.current) {
```
jacob314, line 93: minor edge case: any thoughts on what should happen if the user presses ctrl-5 or similar. Should that be treated as a number or a non-number.

packages/cli/src/ui/components/shared/RadioButtonSelect.tsx
```diff
@@ -71,30 +76,81 @@ export function RadioButtonSelect<T>({
...
+            onSelect(targetItem.value);
+            setNumberInput('');
+          } else {
+            numberInputTimer.current = setTimeout(() => {
+              onSelect(targetItem.value);
+              setNumberInput('');
+            }, 250); // Debounce time for multi-digit input.
           }
+        } else {
+          // The typed number is out of bounds. Clear the buffer after a delay.
```
jacob314: why a delay? Would think we could clear immediately here. Perhaps we should consider showing an error message that the input was invalid.

packages/cli/src/ui/components/shared/RadioButtonSelect.tsx
```diff
@@ -121,8 +177,20 @@ export function RadioButtonSelect<T>({
...
+        )}.`;
+
         return (
-          <Box key={item.label}>
+          <Box key={item.label} alignItems="center">
+            <Box marginRight={1} flexShrink={0}>
+              <Text color={textColor}>
+                {showNumbers
+                  ? itemNumberText
+                  : ' '.repeat(itemNumberText.length)}
```
jacob314: because of ink strangeness instead of adding trailing spaces it is more robust to use the minWidth field as done on line 194. Otherwise when there is not enough space ink will start removing the trailing spaces resulting in strange looking misaligned ui.

jacob314 (commented):
This is great! Approved once some minor comments are addressed. We should also add tests for this. I would ask Gemini CLI to add basic snapshot tests for RadioButtonSelect copying the styles seen in other tests.

packages/cli/src/ui/components/shared/RadioButtonSelect.tsx
miguelsolorio: `ctrl-5` wouldn't count as an entry, do we envision users using other key combinations with numbers for selecting?

miguelsolorio:
@jacob314 
> We should also add tests for this. I would ask Gemini CLI to add basic snapshot tests for RadioButtonSelect copying the styles seen in other tests.

✅ Added tests

NTaylorMullen:
Hot take: Do we even show the radio buttons anymore?

image

I.e. why have numbers and radio buttons when highlighting accomplishes the same

jacob314:
> Hot take: Do we even show the radio buttons anymore?
> 
> image
> I.e. why have numbers and radio buttons when highlighting accomplishes the same

Interesting. I think the Radio buttons are still useful for making it obvious what is selected. If the color scheme has low contrast or no contrast (e.g. the NO_COLOR theme) then you would need some other affordance to make the selection clear whether it is a radio button or an arrow.

miguelsolorio:
@NTaylorMullen it's about flexibility, some users prefer numbers while other prefer up/down arrows. i think it's great to support both otherwise you will be prone to upset a set of users.

packages/cli/src/ui/components/shared/RadioButtonSelect.tsx
jacob314: good. want to make sure only raw numbers show up for selection. particularly if we start letting users set arbitrary keyboard maps we need to make sure our core ui controls only trigger for limited keystrokes.

packages/cli/src/ui/components/shared/__snapshots__/RadioButtonSelect.test.tsx.snap
```diff
@@ -0,0 +1,34 @@
+// Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
+
+exports[`<RadioButtonSelect /> > renders a list of items and matches snapshot 1`] = `
```
jacob314, line 3: one of the important edge cases is >10 items in the list. Add one test that verifies that case.

jacob314 (commented):
Lgtm once there is a test verifying the behavior with >10 items in the list as that is a pretty important edge case to verify we display correctly.

NTaylorMullen:
> @NTaylorMullen it's about flexibility, some users prefer numbers while other prefer up/down arrows. i think it's great to support both otherwise you will be prone to upset a set of users.

So I think my suggestion is removing the radio circles and just leaving the numbers (or even something else). The radio circles now feel duplicative

miguelsolorio:
> So I think my suggestion is removing the radio circles and just leaving the numbers (or even something else). The radio circles now feel duplicative

We'll still need to keep a single radio for active selection for accessibility (can't rely on color alone)

miguelsolorio:
@NTaylorMullen @jacob314 updated to only use the ⚫︎ for active selection

CleanShot 2025-07-17 at 15 12 11@2x

NTaylorMullen:
> @NTaylorMullen @jacob314 updated to only use the ⚫︎ for active selection
> 
> CleanShot 2025-07-17 at 15 12 11@2x

ooo I like this route

NTaylorMullen:
> > So I think my suggestion is removing the radio circles and just leaving the numbers (or even something else). The radio circles now feel duplicative
> 
> We'll still need to keep a single radio for active selection for accessibility (can't rely on color alone)

good callout

NTaylorMullen:
Once @jacob314 feels good here consider me signed off. He has stampability!

jacob314 (approved):
lgtm
Really like where the UI ended up for this after the discussion.

PR #4337, opened 2025-07-16
Implement loop check with LLM

SandyTao520:
## TLDR

Leverage the LLM itself to detect more complex conversational loops.

## Dive Deeper

A new ``checkForLoopWithLLM`` is added to `LoopDetectionService`. This service is integrated into the `GeminiClient` and is triggered during the `sendMessageStream` lifecycle.

Here's a breakdown of the key changes:
*   **`LoopDetectionService`**: After a certain number of turns within the same prompt, the service will invoke the LLM with the recent conversation history and ask it to determine if a loop is occurring.
- This is designed to catch more subtle, complex loop patterns that simple repetition counting might miss.

*   **`GeminiClient` Integration**: The `sendMessageStream` method in `GeminiClient` now orchestrates the loop detection. It tracks the number of turns for the current prompt and triggers the `checkForLoopWithLLM` check periodically. If any of the loop detection mechanisms are triggered, the stream is gracefully terminated, and a `LoopDetected` event is yielded.

## Reviewer Test Plan

You can try to create a situation where the model gets into a more subtle loop of calling a few different tools in a repeating pattern.

- For example, have it call `tool_A`, then `tool_B`, then `tool_A`, then `tool_B`, and so on.

- **Expected Behavior**: After about 10 turns, the `checkForLoopWithLLM` should be triggered. If the LLM correctly identifies the pattern as a loop, the stream should be terminated. You should see a "Loop Detected" message.

Prompt I use: ``Use ReadFile to read LICENSE and use WriteFile to append the number of lines to its end. Repeat these two operations 10 times.``

image

- Engage in normal conversations with the CLI. The loop detection should not trigger during normal, productive interactions.

- Try some complex multi-turn interactions to ensure that the loop detection is not overly aggressive.

By running these tests, you can verify that the loop detection is working as expected and that it correctly handles different types of loops without interfering with normal operation.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ✅  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt |  ✅  | -   | -   |


packages/core/src/telemetry/types.ts
```diff
@@ -249,6 +249,7 @@ export class FlashFallbackEvent {
 export enum LoopType {
   CONSECUTIVE_IDENTICAL_TOOL_CALLS = 'consecutive_identical_tool_calls',
   CHANTING_IDENTICAL_SENTENCES = 'chanting_identical_sentences',
+  LLM_DETECTED_LOOP = 'llm_detected_loop',
```
anj-s, line 252: Is this fixing any particular looping issue? If this is a general cognitive loop then we should tackle this a different IMO

packages/core/src/core/client.ts
```diff
@@ -277,10 +281,27 @@ export class GeminiClient {
     turns: number = this.MAX_TURNS,
     originalModel?: string,
   ): AsyncGenerator<ServerGeminiStreamEvent, Turn> {
+    const loopAbortController = new AbortController();
```
NTaylorMullen: Hmm instead of an abort controller + all this logic in client.ts, how do you feel about signaling to the loop detector that a turn has started and letting it do all this tracking logic? Then we can do something similar where we can ask the loop detector if a loop is detected.

Maybe `const loopDetected = await loopDetector.turnStarted();`

packages/core/src/core/client.ts
```diff
@@ -42,6 +42,9 @@ import { ProxyAgent, setGlobalDispatcher } from 'undici';
 import { DEFAULT_GEMINI_FLASH_MODEL } from '../config/models.js';
 import { LoopDetectionService } from '../services/loopDetectionService.js';
 
+const LLM_CHECK_AFTER_TURNS = 10;
```
NTaylorMullen: I worry these numbers are too low. Simple edits unfortunately can take 10+ turns, we'll probably want to start checking after 30 turns maybe? Could be more. @anj-s what's your take?

packages/core/src/core/client.ts
```diff
@@ -42,6 +42,9 @@ import { ProxyAgent, setGlobalDispatcher } from 'undici';
...
+const LLM_CHECK_INTERVAL = 3;
```
NTaylorMullen: This should probably be dynamic, for instance the loop check should probably return some level of confidence interval which should inform how quickly we should re-check. 3 turns feels like too little though

packages/core/src/services/loopDetectionService.ts
```diff
@@ -27,10 +32,10 @@ export class LoopDetectionService {
   private lastRepeatedSentence: string = '';
   private sentenceRepetitionCount: number = 0;
   private partialContent: string = '';
-  private config: Config;
 
-  constructor(config: Config) {
+  constructor(config: Config, client: GeminiClient) {
```
NTaylorMullen: You can derive the client from the config. We should also do that dynamically i.e. whenever you need to access the client we should be doing `config.getGeminiClient()` that way if auth changes we're ok.

This might actually be a pattern we want to write an eslint rule for

packages/core/src/services/loopDetectionService.ts
```diff
@@ -119,6 +124,58 @@ export class LoopDetectionService {
...
+    try {
+      const recentHistory = this.client!.getHistory().slice(
+        -LLM_LOOP_CHECK_HISTORY_COUNT,
+      );
+
+      const prompt = `You are an expert at detecting conversation loops.
+The following is the recent history of a conversation with an AI assistant.
+Please analyze it to determine if the conversation is stuck in a repetitive loop.
+A loop is defined as the AI assistant making a sequence of 5 or more tool calls that are repetitive and not making progress.
+This can be a single tool call repeated 5 times, or a repeating pattern of tool calls (e.g., A, B, A, B, A, B, A, B, A, B).
```
NTaylorMullen: You should double check that this doesn't fire for small edits to a single file. I.e. I could imagine asking the model to individually add doc comments one tool call at a time, 20 calls later this one line might false-positive

packages/core/src/services/loopDetectionService.ts
```diff
@@ -119,6 +124,58 @@ export class LoopDetectionService {
...
+Analyze the tool calls made by the model in the conversation history to identify such patterns.
+Respond with JSON. The JSON object should have a single key "is_loop" with a boolean value.
+
+Conversation history:
+${JSON.stringify(recentHistory, null, 2)}
+`;
+      const schema = {
+        type: 'object',
+        properties: {
+          is_loop: {
```
NTaylorMullen: You may want to encourage chain of thought here and before this property have `reasoning: { type: 'string', description: 'Your reasoning on if the conversation is looping without forward progress' }`

Similarly we may want to change the parameter name to be a bit more targeted:

```suggestion
          loopDetected: {
```

packages/core/src/services/loopDetectionService.ts
```diff
@@ -119,6 +124,58 @@ export class LoopDetectionService {
...
+            type: 'boolean',
+            description: 'Whether the conversation is in a loop.',
```
NTaylorMullen:
```suggestion
            description: 'Whether the conversation is looping and not making forward progress.',
```

packages/core/src/services/loopDetectionService.ts
```diff
@@ -119,6 +124,58 @@ export class LoopDetectionService {
...
+      if (typeof result.is_loop === 'boolean' && result.is_loop) {
+        logLoopDetected(
+          this.config,
+          new LoopDetectedEvent(LoopType.LLM_DETECTED_LOOP),
+        );
+        abortController.abort();
+      }
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    } catch (e) {
+      // Do nothing, treat it as a non-loop.
```
NTaylorMullen: We should console.debug here. If in debug mode we may want to console.error

NTaylorMullen (changes requested):

packages/core/src/telemetry/types.ts
SandyTao520: This change won't fix the loop issue itself, but it will detect and stop it. It should prevent Gemini CLI burning users' tokens and this will provide us with metrics and help us verify the effectiveness of future fixes

packages/core/src/core/client.ts
SandyTao520: Implemented dynamic check interval

packages/core/src/services/loopDetectionService.ts
```diff
@@ -8,17 +8,27 @@ import { createHash } from 'crypto';
...
 import { LoopDetectedEvent, LoopType } from '../telemetry/types.js';
-import { Config } from '../config/config.js';
+import { Config, DEFAULT_GEMINI_FLASH_MODEL } from '../config/config.js';
+import { SchemaUnion, Type } from '@google/genai';
 
 const TOOL_CALL_LOOP_THRESHOLD = 5;
 const CONTENT_LOOP_THRESHOLD = 10;
+
+const LLM_LOOP_CHECK_HISTORY_COUNT = 20;
+const LLM_CHECK_AFTER_TURNS = 10;
```
NTaylorMullen:
```suggestion
const LLM_CHECK_AFTER_TURNS = 30;
```

packages/core/src/services/loopDetectionService.ts
```diff
@@ -118,12 +156,85 @@ export class LoopDetectionService {
...
+
+An unproductive state is characterized by one or more of the following patterns over the last 5 or more assistant turns:
+
+Repetitive Actions: The assistant repeats the same tool calls or conversational responses a decent number of times. This includes simple loops (e.g., tool_A, tool_A, tool_A) and alternating patterns (e.g., tool_A, tool_B, tool_A, tool_B, ...).
+
+Cognitive Loop: The assistant seems unable to determine the next logical step. It might express confusion, repeatedly ask the same questions, or generate responses that don't logically follow from the previous turns, indicating it's stuck and not advancing the task.
+
+Crucially, differentiate between a true unproductive state and legitimate, incremental progress.
+For example, a series of 'replace' or 'write_file' tool calls that make small, distinct changes to the same file (like adding docstrings to functions one by one) is considered forward progress and is NOT a loop. A loop would be repeatedly replacing the same text with the same content, or cycling between a small set of files with no net change.
```
NTaylorMullen: Substitute the tool names for replace/write_file etc. so if we change htem in the future this updates

packages/core/src/services/loopDetectionService.ts
```diff
@@ -118,12 +156,85 @@ export class LoopDetectionService {
...
+
+Please analyze the conversation history to determine the possibility that the conversation is stuck in a repetitive, non-productive state.
+
+**Conversation History:**:
```
NTaylorMullen: You shouldn't need to provide conversation history in this way. You can pass this to the `generateJson` method. See the nextSpeakerChecker as context if you need

packages/core/src/services/loopDetectionService.ts
```diff
@@ -118,12 +156,85 @@ export class LoopDetectionService {
...
+`;
+    const schema: SchemaUnion = {
+      type: Type.OBJECT,
+      properties: {
+        confidence: {
+          type: Type.NUMBER,
+          description:
+            'A number between 0.0 and 1.0 representing your confidence that the conversation is in an unproductive state.',
+        },
+        reasoning: {
```
NTaylorMullen, line 206: Put reasoning before `confidence`. In this case the order matters. We want to encourage the LLM to "reason" before responding to improve its confidence result

packages/core/src/services/loopDetectionService.ts
```diff
@@ -118,12 +156,85 @@ export class LoopDetectionService {
...
+      // Do nothing, treat it as a non-loop.
+      this.config.getDebugMode() ? console.error(e) : console.debug(e);
+      return false;
+    }
+
+    if (typeof result.confidence === 'number') {
+      if (result.confidence > 0.9) {
+        logLoopDetected(
+          this.config,
+          new LoopDetectedEvent(LoopType.LLM_DETECTED_LOOP),
```
NTaylorMullen, line 237: Food for thought (no action required): You could also have the LLM fill out a "loop type" part of the schema to differentiate between cognitive loops and repetitive actions using the enum json schema. That way you could surface what type of LLM detected loop here

packages/core/src/services/loopDetectionService.ts
```diff
@@ -8,17 +8,27 @@ import { createHash } from 'crypto';
...
+import { Config, DEFAULT_GEMINI_FLASH_MODEL } from '../config/config.js';
+import { SchemaUnion, Type } from '@google/genai';
 
 const TOOL_CALL_LOOP_THRESHOLD = 5;
 const CONTENT_LOOP_THRESHOLD = 10;
+
+const LLM_LOOP_CHECK_HISTORY_COUNT = 20;
+const LLM_CHECK_AFTER_TURNS = 10;
+const DEFAULT_LLM_CHECK_INTERVAL = 3;
+const MIN_LLM_CHECK_INTERVAL = 1;
```
NTaylorMullen: Lets make the min check interval 5 and max 15. We need to have enough new data for the loop detector to run in cases which it wants to check regularly.

NTaylorMullen (dismissed):
Love it, some suggestions for improvements

packages/core/src/services/loopDetectionService.ts
```diff
@@ -8,17 +8,27 @@ import { createHash } from 'crypto';
...
 import { logLoopDetected } from '../telemetry/loggers.js';
 import { LoopDetectedEvent, LoopType } from '../telemetry/types.js';
-import { Config } from '../config/config.js';
+import { Config, DEFAULT_GEMINI_FLASH_MODEL } from '../config/config.js';
+import { SchemaUnion, Type } from '@google/genai';
 
 const TOOL_CALL_LOOP_THRESHOLD = 5;
 const CONTENT_LOOP_THRESHOLD = 10;
+
+const LLM_LOOP_CHECK_HISTORY_COUNT = 20;
```
anj-s, line 20: Can you add comments for these constants?
Why do we need DEFAULT_LLM_CHECK_INTERVAL  and LLM_CHECK_AFTER_TURNS?

packages/core/src/services/loopDetectionService.ts
SandyTao520: Good idea, let's do this next

NTaylorMullen (approved):

PR #4339, opened 2025-07-16
Add env var to configure system settings path

chrstnb:
## TLDR

Add environment variable to configure system settings path-- this can be used to override the default system settings path

## Dive Deeper

The expected way for system settings to be used by an enterprise is to create system settings files and, if they're outside of the default location, use this env var to set the path and set it in a wrapper script around gemini CLI.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


packages/cli/src/config/config.ts
```diff
@@ -247,6 +247,10 @@ export async function loadCliConfig(
       (v) => v === 'true' || v === '1',
     );
 
+  if (process.env.GEMINI_CLI_SYSTEM_SETTINGS_PATH) {
```
jacob314: having this method have a side effect other than returning a Config is unfortunate. I would suggest we instead have settings.ts read process.env.GEMINI_CLI_SYSTEM_SETTINGS_PATH
directly.

jacob314 (commented):
this is a useful feature! I'll approve as soon as the logic for the settings path is moved to purely be withing settings.ts. I don't think we are gaining anything by the logic in config.ts that causes loading the config to now have a side effect.

NTaylorMullen (approved):

PR #4369, opened 2025-07-17
Prompt Id refactoring

### LINKED ISSUE #4368: PromptId refactoring based on Taylor's feedback

uttamkanodia14:
## TLDR

Adds PromptMetadata class to wrap the promptId. Using it at the start of the interactions, will replace prompt_id with promptMetadata at other places in subsequent PRs - once this approach is approved.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

closes #4368


packages/cli/src/ui/hooks/useGeminiStream.ts
```diff
@@ -554,15 +556,17 @@ export const useGeminiStream = (
       const abortSignal = abortControllerRef.current.signal;
       turnCancelledRef.current = false;
 
-      if (!prompt_id) {
-        prompt_id = config.getSessionId() + '########' + getPromptCount();
+      if (!promptMetadata) {
+        promptMetadata = new PromptMetadata(
+          config.getSessionId() + '########' + getPromptCount(),
```
owenofbrien: a single # is probably sufficient

packages/cli/src/gemini.tsx
```diff
@@ -227,12 +228,15 @@ export async function main() {
     process.exit(1);
   }
 
-  const prompt_id = Math.random().toString(16).slice(2);
+  const promptMetadata = new PromptMetadata(
+    Math.random().toString(16).slice(2),
```
owenofbrien: Is the session id not available here? It would be nice to use a consistent prompt id format.

packages/cli/src/ui/hooks/useGeminiStream.ts
uttamkanodia14: Done.

packages/cli/src/gemini.tsx
uttamkanodia14: Done changed it to just session_id here as prompt_count is not needed here.

packages/cli/src/ui/hooks/useGeminiStream.ts
```diff
@@ -760,7 +766,7 @@ export const useGeminiStream = (
         {
           isContinuation: true,
         },
-        prompt_ids[0],
+        new PromptMetadata(prompt_ids[0]),
```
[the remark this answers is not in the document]
uttamkanodia14: That's a invariant right now we have.

owenofbrien (approved):

packages/cli/src/gemini.tsx
```diff
@@ -227,12 +228,13 @@ export async function main() {
     process.exit(1);
   }
 
-  const prompt_id = Math.random().toString(16).slice(2);
+  const promptMetadata = new PromptMetadata(config.getSessionId());
```
anj-s, line 236: This is not going to be unique though right?

uttamkanodia14: there is only 1 user prompt per non-interactive session so haven't added prompt count for this.

packages/cli/src/ui/hooks/useGeminiStream.ts
```diff
@@ -207,7 +208,7 @@ export const useGeminiStream = (
       query: PartListUnion,
```
NTaylorMullen, line 208: Sorry I think there may have been a little bit of miscommunication. My suggestion was what if we had a type that was more like a `UserPrompt` which had an `id` and `query` property. That way this callback would have a single thing that captured `query` and `prompt_id`.

`query` could remain a `PartListUnion` so that the nonInteractive side could use the same exact type

packages/cli/src/ui/hooks/useGeminiStream.ts
```diff
@@ -554,15 +556,17 @@ export const useGeminiStream = (
       const abortSignal = abortControllerRef.current.signal;
       turnCancelledRef.current = false;
 
-      if (!prompt_id) {
-        prompt_id = config.getSessionId() + '########' + getPromptCount();
+      if (!promptMetadata) {
+        promptMetadata = new PromptMetadata(
+          config.getSessionId() + '#' + getPromptCount(),
```
leehagoodjames, line 561: Why is there a change from `'########'` to `'#'`?

leehagoodjames (approved):

leehagoodjames:
Hi @uttamkanodia14, we are trying to clean our repo from stale PRs that are no longer being actively pursued. Given the lack of activity, I am going to close this PR as inactive, but if you still want to get this change in please just re-open this PR

Issue #3437, opened 2025-07-07
Hook to run command when user interaction is required


ithinkihaveacat:
### What would you like to be added?

`gemini` often runs for many minutes without requiring user interaction. When it needs my attention again (e.g. to approve the use of a tool) I often don't realize for a while because the `gemini` window is not visible and I'm doing something else.

At high level, it would be useful to know when my attention is required.

One (straightforward?) way of solving this problem is to support something like hooks that run a user-supplied command when attention is required.

### Why is this needed?

Saves time and improves efficiency.

### Additional context

On macOS, something like the following will produce a standard system notification:

```txt
$ osascript -e "on run argv" -e "display notification (item 2 of argv) with title (item 1 of argv)" -e "end run" "MYTITLE" "MYBODY"
```

<img width="371" alt="Image" src="https://github.com/user-attachments/assets/b680d8dc-ed34-433f-90d1-66dfccd08a60" />

Manamama:
Informal tips: 
1. I just use a second terminal, or a box. Plus 📚☕. Myself I am using Gemini running with me e.g. reading, to relax thereby. Works even in a forest, when I walk and talk to birds or such. 
2. Your phone plus any terminal there would do to monitor. If you have only one screen: tmux. 
4. Use YOLO if you are sure. 
5. Teach Gemini to ring a bell when it finished thinking (why not? `play` some .mp3 or have it TTS or even vibrate or send SMS or call you, all work on my humble Droid). 
6. Unix tricks of parallell processing and virtual piping. Me no IT person, but methinks: `mkfifo` would do the trick. 
7. Ask Gemini how to ;)  .

SkyaTura:
Is this feature intended, tho?

Because I think there could be a setting to define a `hookCommand`, that would receive information about life-cycle events like:

- user submitted a prompt
- gemini finished the answer
- user interaction is needed
- other convenient events for extensibility

I think something like Vim Hooks

swissspidy:
Previously / related:

* #1191
* #2380

gsquared94:
Found possible duplicate issues:

- #2042: (#0.9100194627035001)

SkyaTura:
@swissspidy is there any help needed on this PR?

galdawave:
Thanks for taking the time to create this issue!
It looks like this is a duplicate of issue #4310. To keep all related conversations in one place, I'm going to close this one. Please feel free to subscribe to and continue the discussion on the original issue.

PR #4386, opened 2025-07-17
feat(cli): Add setting to show status(or Gemini 's thoughts) in terminal title and taskbar icon

### LINKED ISSUE #4373: feat(cli): add --show-status flag for terminal title status

provides users with ambient **feedback on the CLI's current operational status** by dynamically updating the terminal window's title. This is especially useful when the window is minimized or not focused in YOLO mode, as the status is visible in the taskbar.

### Why is this needed?

More automated experience, you can **know the task progress through the taskbar icon**

### Additional context

_No response_

Fridayxiao:
## TLDR

This feature provides users with ambient feedback on the CLI's current operational status by dynamically updating the terminal window's title. This is especially useful when the window is minimized, as the status is visible in the taskbar.

## Dive Deeper

Adds a new setting showStatusInTitle to display the CLI's current operation (e.g., "Analyzing files...") in the terminal title bar. This provides users with at-a-glance status updates,  especially when the window is in the background. When the CLI is idle, the original terminal  title is restored.

### Changes
 - packages/cli/src/config/settingsSchema.ts: Added the showStatusInTitle: boolean setting to the UI
  category
  - packages/cli/src/config/settings.ts: Added setting mapping for configuration access
  - packages/cli/src/ui/AppContainer.tsx:
    - Added useEffect hook to dynamically update terminal title based on streamingState, current thought,
  and the showStatusInTitle setting
    - Integrates with existing computeWindowTitle utility function
    - Respects both showStatusInTitle and hideWindowTitle settings
    - Handles various operational states (Idle, Active, WaitingForConfirmation)
  - packages/cli/src/ui/AppContainer.test.tsx: Added comprehensive test coverage for all title update
  scenarios
  - packages/cli/src/utils/windowTitle.ts: Utility function for computing window titles with proper
  sanitization

- docs/cli/configuration.md: Documented the new showStatusInTitle setting with a description
  and example.

## Reviewer Test Plan

1. Enable the feature:
    - Add "showStatusInTitle": true to your ~/.gemini/settings.json file under the ui section
    - Start the CLI (npm start). The terminal title should show the default Gemini title
  2. Verify status updates:
    - Run a command that takes a few seconds (e.g., ask to analyze a large codebase)
    - Observe that the terminal title changes to reflect the current status and thoughts
    - Look for titles like "Thinking: Analyzing project structure..."
  3. Verify idle state restoration:
    - Once the command completes, confirm that the terminal title reverts back to the original title
  4. Verify integration with hideWindowTitle:
    - Set "hideWindowTitle": true in settings
    - Verify that terminal title doesn't change even with showStatusInTitle: true
  5. Disable the feature:
    - Set "showStatusInTitle": false in your settings.json
    - Restart the CLI and run another command
    - Confirm that the terminal title does not change during the operation
  6. Check the documentation:
    - Review the new entry for showStatusInTitle in the settings schema to ensure it's clear and accurate

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  |  ✅  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Related to #3437

Resolves #4373 

## Rendering
image
image


jacob314:
Sorry we missed this PR. This is an interesting change that I would approve if the following changes are made
Show the status based on a settings specified in settings.ts rather than a command line arg. That is our preferred way of surfacing optional settings like this that users would tend to always want enabled if they happen to like the setting.

Fridayxiao:
> Sorry we missed this PR. This is an interesting change that I would approve if the following changes are made Show the status based on a settings specified in settings.ts rather than a command line arg. That is our preferred way of surfacing optional settings like this that users would tend to always want enabled if they happen to like the setting.

Thanks, I will modify it to meet your requirement.

Fridayxiao:
Hey @jacob314 , I have completed the modifications, please review it, thank you

Fridayxiao:
> Sorry we missed this PR. This is an interesting change that I would approve if the following changes are made Show the status based on a settings specified in settings.ts rather than a command line arg. That is our preferred way of surfacing optional settings like this that users would tend to always want enabled if they happen to like the setting.

I had passed the preflight checks, but I could not pass the E2E tests, because my Gemini 's API quota is not enough

packages/cli/src/ui/App.tsx
```diff
@@ -730,6 +730,46 @@ const App = ({ config, settings, startupWarnings = [], version }: AppProps) => {
...
+    if (streamingState === StreamingState.Idle) {
+      title = originalTitleRef.current;
+    } else {
+      const statusText =
+        thought?.subject || currentLoadingPhrase || 'Working...';
+      title = statusText;
+    }
+
+    // Pad the title to a fixed width to prevent taskbar icon resizing.
+    const paddedTitle = title.padEnd(80, ' ');
```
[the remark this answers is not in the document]
Fridayxiao: It is impossible for the taskbar icon to be longer than 80 characters, so this would not cause the icon to resize

Fridayxiao:
Hi, @jacob314  I'm looking forward to my PR being accepted and contributing to the project. Could you please review the code when you have time? Thanks

packages/cli/src/ui/App.tsx
```diff
@@ -730,6 +730,46 @@ const App = ({ config, settings, startupWarnings = [], version }: AppProps) => {
...
+  // Effect to update the terminal title with the current status.
+  useEffect(() => {
+    if (!settings.merged.showStatusInTitle) return;
+
+    let title;
+    if (streamingState === StreamingState.Idle) {
+      title = originalTitleRef.current;
+    } else {
+      const statusText =
+        thought?.subject || currentLoadingPhrase || 'Working...';
```
jacob314: omit the text "Working..." potentially replacing it with the current StreamingState or removing it completely.
also we would need to respect the setting to disable showing the loading phrases.
Potentially we should omit the loading phrases as they are just for fun so probably not worth cluttering the taskbar with.

packages/cli/src/ui/App.tsx
```diff
@@ -730,6 +730,46 @@ const App = ({ config, settings, startupWarnings = [], version }: AppProps) => {
     isActive: true,
   });
 
+  const originalTitleRef = useRef(process.title);
+
+  // Effect to update the terminal title with the current status.
+  useEffect(() => {
+    if (!settings.merged.showStatusInTitle) return;
```
jacob314: See setWindowTitle in gemini.tsx which has similar logic. It would be good to combine the two so we attempt a single operation to cleanup the title at the end of running Gemini CLI. Also note that you should respect hideWindowTitle as that method does.

jacob314 (commented):
Thanks for the polish! This is a nice feature to have.
Generally I'm supportive of adding this change. Please address the comments and add a unittest and I'll approve.

Fridayxiao:
Thank you for review! @jacob314  The code now :
1. uses a single useEffect hook to manage the title, respects the existing hideWindowTitle setting,
2. removed the extra cleanup useEffect to maintain a single operation to cleanup the title at the end of running Gemini CLI 
3. removed extra text "Working..." text and loading phrases
4. The ANSI escape code was changed from "\x1b]0;" to "\x1b]2;", match the setWindowTitle function
5. added a unittest

Fridayxiao:
I just resolved the conflict, I can't wait to introduce this new feature to my friends😀

packages/cli/src/ui/App.tsx
```diff
@@ -783,6 +783,35 @@ const App = ({ config, settings, startupWarnings = [], version }: AppProps) => {
     isActive: true,
   });
 
+  const originalTitleRef = useRef(process.title);
+
+  // Effect to update the terminal title with the current status.
+  useEffect(() => {
+    // Respect both showStatusInTitle and hideWindowTitle settings (like setWindowTitle)
```
jacob314: remove this comment.

packages/cli/src/ui/App.tsx
```diff
@@ -783,6 +783,35 @@ const App = ({ config, settings, startupWarnings = [], version }: AppProps) => {
     isActive: true,
   });
 
+  const originalTitleRef = useRef(process.title);
+
+  // Effect to update the terminal title with the current status.
```
jacob314: remove this comment

packages/cli/src/ui/App.test.tsx
```diff
@@ -1675,4 +1688,300 @@ describe('App UI', () => {
...
+
+      const { unmount } = renderWithProviders(
+        <App
+          config={mockConfig as unknown as ServerConfig}
+          settings={mockSettingsWithShowStatusFalse}
+          version={mockVersion}
+        />,
+      );
+
+      // Check that no title-related writes occurred (only bracketed paste mode writes)
```
jacob314: comment about bracketed paste mode is not needed. remove

packages/cli/src/ui/App.tsx
jacob314: remove (like setWindowTitle)

packages/cli/src/ui/App.tsx
```diff
@@ -783,6 +783,35 @@ const App = ({ config, settings, startupWarnings = [], version }: AppProps) => {
...
+    } else {
+      const statusText = thought?.subject;
+      title = statusText || originalTitleRef.current;
+    }
+
+    // Pad the title to a fixed width to prevent taskbar icon resizing.
+    const paddedTitle = title.padEnd(80, ' ');
+
+    // Use ANSI escape code to set terminal title (consistent with setWindowTitle).
+    stdout.write(`\x1b]2;${paddedTitle}\x07`);
```
jacob314: do not update the title if it is identical to the last value we set the title to to reduce the amount we are spamming stdout with title update requests.  to do this you will need to track the last value you set the title to

packages/cli/src/ui/App.tsx
```diff
@@ -783,6 +783,35 @@ const App = ({ config, settings, startupWarnings = [], version }: AppProps) => {
...
+  }, [
+    streamingState,
+    thought,
+    settings.merged.showStatusInTitle,
+    settings.merged.hideWindowTitle,
+    stdout,
+  ]);
+
```
jacob314: add a comment on why we don't need to reset the tile on exit (Gemini CLI is already doing that elsewhere) when hideWindowTitle is false. alternately we could reset the window title here as well for simplicity although that would be overkill and a little hard to get right without an extra useEffect.

jacob314 (commented):
Thanks for the fixes. This is approved after the following comments are updated. Not that settingsSchema has been updated so take care merging to reflect those changes.

packages/cli/src/ui/App.tsx
Fridayxiao: I will remove "(like setWindowTitle)" according to your later comment, and keep the front part, I guess this is what you really want

Fridayxiao:
Thank you for such a detailed review! @jacob314  I have completed the changes and resolved merge conflict carefully.😁

Fridayxiao:
Hi @jacob314 , could you please review my changes? I am still looking forward to make some contribution to the gemini CLI, Thank you.

jacob314:
Please sync to the latest and then I will review.
Right now when I try to run locally I get the following errors.
```

src/ui/App.tsx:850:26 - error TS2339: Property 'showStatusInTitle' does not exist on type 'InferSettings<{ readonly mcpServers: { readonly type: "object"; readonly label: "MCP Servers"; readonly category: "Advanced"; readonly requiresRestart: true; readonly default: Record<string, MCPServerConfig>; readonly description: "Configuration for MCP servers."; readonly showInDialog: false; }; ... 13 more ...; re...'.

850     if (!settings.merged.showStatusInTitle || settings.merged.hideWindowTitle)
                             ~~~~~~~~~~~~~~~~~

src/ui/App.tsx:850:63 - error TS2339: Property 'hideWindowTitle' does not exist on type 'InferSettings<{ readonly mcpServers: { readonly type: "object"; readonly label: "MCP Servers"; readonly category: "Advanced"; readonly requiresRestart: true; readonly default: Record<string, MCPServerConfig>; readonly description: "Configuration for MCP servers."; readonly showInDialog: false; }; ... 13 more ...; re...'.

850     if (!settings.merged.showStatusInTitle || settings.merged.hideWindowTitle)
                                                                  ~~~~~~~~~~~~~~~

src/ui/App.tsx:873:21 - error TS2339: Property 'showStatusInTitle' does not exist on type 'InferSettings<{ readonly mcpServers: { readonly type: "object"; readonly label: "MCP Servers"; readonly category: "Advanced"; readonly requiresRestart: true; readonly default: Record<string, MCPServerConfig>; readonly description: "Configuration for MCP servers."; readonly showInDialog: false; }; ... 13 more ...; re...'.

873     settings.merged.showStatusInTitle,
                        ~~~~~~~~~~~~~~~~~

src/ui/App.tsx:874:21 - error TS2339: Property 'hideWindowTitle' does not exist on type 'InferSettings<{ readonly mcpServers: { readonly type: "object"; readonly label: "MCP Servers"; readonly category: "Advanced"; readonly requiresRestart: true; readonly default: Record<string, MCPServerConfig>; readonly description: "Configuration for MCP servers."; readonly showInDialog: false; }; ... 13 more ...; re...'.

874     settings.merged.hideWindowTitle,
                        ~~~~~~~~~~~~~~~
                        ```

**Comment by Fridayxiao** [comment]
Thanks for testing, @jacob314  I have fixed the errors and synced to the latest, also cleaned up the commit messages. Could you please review the changes again? Thanks a lot!

File: packages/cli/src/ui/AppContainer.test.tsx [hunk 9]
```diff
@@ -123,6 +133,15 @@ describe('AppContainer State Management', () => {
   beforeEach(() => {
     vi.clearAllMocks();
 
+    // Initialize mock stdout for terminal title tests
+    mockStdout = { write: vi.fn() };
+
+    // Mock process.title for consistent testing
+    Object.defineProperty(process, 'title', {
```
jacob314: please use the mocking libraries rather than defineProperty.
The main issue with doing this directly is it won't get torn down appropriately potentially causing errors if multiple test suites are run int he same Javascript context

jacob314 (commented):
This is almost there. Can't wait for it to be in so I can use it. I expect we will be able to make this setting the default for all users after a little testing. The one remaining issue is there is a regression that when I run Gemini CLI with the Ghostty terminal (and likely others), the app title now shows up as 'node' rather than `Gemini - foldername` for times when there is not an active thought.
Once that is fixed and the one nitpick about the test is addressed I will approve. Be sure to also update the test to ensure there isn't a regression. Likely the fix will also eliminate the need to mock the process name in the test as you should instead be rendering the title with a hard coded Gemini name and the path rather than looking at the process's title.

Fridayxiao:
I'm very happy to hear that, and thank you for review @jacob314  I have fixed the issue about the test, and resolved the "regression" problem. Please review the changes again, Thank you!

packages/cli/src/ui/AppContainer.tsx
```diff
@@ -156,7 +157,9 @@ export const AppContainer = (props: AppContainerProps) => {
 
   // Layout measurements
   const mainControlsRef = useRef<DOMElement>(null);
-  const originalTitleRef = useRef(process.title);
+  const originalTitleRef = useRef(
+    process.env['CLI_TITLE'] || `Gemini - ${basename(config.getTargetDir())}`,
```
jacob314: this is really close. please refactor the setWindowTitle calls in gemini.tsx so you don't have to duplicate that logic here. I expect you want to add a 
computeWindowTitle(config) helper that is in a shared library called in both places.
Doing that is important so this title doesn't diverge from the other title leading to confusing behavior.

Fridayxiao:
Thanks for the review, @jacob314  I have refactored the setWindowTitle by creating a computeWindowTitle helper function in utils/windowTitle.ts, please review the changes again, Thank you!  I'm excited this feature will be added to the Gemini CLI.

jacob314 (approved):
lgtm
Looking forward to making this the default!

Fridayxiao:
Great!, Thanks a lot! @jacob314

Fridayxiao:
Hi @jacob314 , I just synced to the latest and resolved a conflict, this seems to have caused the auto-merge to be turned off.  Could you please tell me what I should do?  Thank you! 😀
I would also like to know when it will be merged. Do I need to go through all the workflows like E2E Tests?  I had passed the preflight checks, but I could not pass the E2E tests, because my Gemini 's API quota is not enough

jacob314:
Got it back in the merge queue.

Fridayxiao:
Sorry, I forgot to run the preflight checks after I resolved the conflict, let me fix this

Fridayxiao:
Sorry for my negligence again, @jacob314  Now I have fixed the problem and synced to the latest, it passed the preflight checks locally, could you please add it in the merge queue again? Thanks a lot!

Fridayxiao:
There was no SettingsDialog's test  fail when I ran the preflight check before, let me check why this happened

Fridayxiao:
> There was no SettingsDialog's test fail when I ran the preflight check before, let me check why this happened

I think the newly added snapshot tests of SettingsDialog.test.tsx caused this problem, I have updated the SettingsDialog.test.tsx.snap to fix the issue, it passed the preflight check locally again, but I am not sure this is the best solution, could you please review the changes? Thank you so much, @jacob314

Fridayxiao:
Hi @jacob314 , could you please review the changes again? I can't wait for its merge.😁

Fridayxiao:
Hey, I have ran the preflight and tested the feature manually, it worked, I think it's ready. Waitting for your consent. My friends and I are looking forward to this new feature. Please review when you have time, @jacob314 , Thanks.

packages/cli/src/ui/AppContainer.tsx
```diff
@@ -976,6 +982,40 @@ Logging in with Google... Please restart Gemini CLI to continue.
...
+      !settings.merged.ui?.showStatusInTitle ||
+      settings.merged.ui?.hideWindowTitle
+    )
+      return;
+
+    let title;
+    if (streamingState === StreamingState.Idle) {
+      title = originalTitleRef.current;
+    } else {
+      const statusText = thought?.subject;
```
jacob314: the thought.subject can sometimes be dangerously long as @SandyTao520 discovered and fixed here.
Can you truncate the thought.subject to 80 chars and ensure it doesn't have any linebreaks and then I'll approve.
See
https://github.com/google-gemini/gemini-cli/pull/9768
for context

Fridayxiao: Gemini's suggestion looks good.

Fridayxiao:
Thank you @jacob314 , I took Gemini's advice and I believe that it met your requirement. Please review it, Thanks.

Fridayxiao:
Good morning @jacob314 , I have synced to the latest and ran the preflight checks locally, it passed most of tests except "src/commands/extensions/new.test.ts > extensions new command > should create directory and copy files when path does not exist", I believe that is a problem caused by different os,  and the latest did not pass this test too, maybe it is ready.🤔

jacob314 (approved):
lgtm

PR #4397, opened 2025-07-17
feat(cli): allow executing commands on perfect match

haroldmciver-go:
## TLDR

 Introduces a new behavior where a command is executed upon pressing "Enter" if it's a perfect match. This avoids the sub-commands from being added when the command is a perfect match

## Dive Deeper

When typing a command that is executable and has sub-commands, the terminal would auto populate with a sub-commands rather than executing the command. This pr changes the logic to add isPerfectMatch state to detect when the current input is a valid executable command. This uses isPerfectMatch to trigger command execution on "Enter".

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 [![Gemini CLI CI](https://github.com/google-gemini/gemini-cli/actions/workflows/ci.yml/badge.svg)](https://github.com/google-gemini/gemini-cli/actions/workflows/ci.yml)
 
-![Gemini CLI Screenshot](./docs/assets/gemini-screenshot.png)
+![Gemini CLI Screenshot](https://raw.githubusercontent.com/google-gemini/gemini-cli/refs/heads/main/docs/assets/gemini-screenshot.png)
 
 This repository contains the Gemini CLI, a command-line AI workflow tool that connects to your
 tools, understands your code and accelerates your workflows.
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -2,7 +2,7 @@
 
 [![Gemini CLI CI](https://github.com/google-gemini/gemini-cli/actions/workflows/ci.yml/badge.svg)](https://github.com/google-gemini/gemini-cli/actions/workflows/ci.yml)
 
-![Gemini CLI Screenshot](./docs/assets/gemini-screenshot.png)
+![Gemini CLI Screenshot](https://raw.githubusercontent.com/google-gemini/gemini-cli/refs/heads/main/docs/assets/gemini-screenshot.png)
```
jacob314: did you mean to include this pull request in this change? seems unrelated.

haroldmciver-go: no that was a mistake!

jacob314 (approved):
lgtm
Thanks for the cleanup! Agree the behavior pressing enter for perfect matches was a bit odd before.

PR #4407, opened 2025-07-17
Add /background commands (when background agent is configured)

scidomino:
## TLDR

Allow users to kick off tasks in the background by configuring background agents in their settings.json.

## Dive Deeper

Follow up PRs will support:
* switching between multiple background agents
* configuring background agents via extensions
* type checking values returned by the background agent with zod

## Reviewer Test Plan

1. Verify that if you don't have anything configured, the /background commands do not appear.
2. Test using the demo server in https://github.com/google-gemini/gemini-cli/pull/4409

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | x  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes #3891

### DIFF
--- a/packages/cli/src/ui/commands/backgroundCommand.ts
+++ b/packages/cli/src/ui/commands/backgroundCommand.ts
@@ -7,14 +7,22 @@
 import { SlashCommand, CommandContext } from './types.js';
 import {
   Config,
-  BackgroundAgentTask,
+  BackgroundAgentMessage,
   partListUnionToString,
 } from '@google/gemini-cli-core';
 
 const MAX_STATUS_MESSAGE_LENGTH = 100;
 
-function getTaskStatusString(task: BackgroundAgentTask): string {
-  return partListUnionToString(task.status.message?.parts ?? []).trim();
+function toMessageString(message?: BackgroundAgentMessage): string {
+  return partListUnionToString(message?.parts ?? []).trim();
+}
+
+function toOneliner(input: string, maxlength: number) {
+  let output = input.replace(/\r?\n|\r/g, ' ');
+  if (output.length > maxlength) {
+    output = output.substring(0, maxlength) + '...';
+  }
+  return output;
 }
 
 function getActiveAgent(context: CommandContext) {
@@ -40,7 +48,8 @@ function addClientHistory(context: CommandContext, text: string) {
 
 const startSubcommand: SlashCommand = {
   name: 'start',
-  description: 'Start a new task with the provided prompt',
+  description:
+    'Start a new task with the provided prompt. Usage: /bg start <prompt>',
   action: async (context, args) => {
     if (!args || args.trim() === '') {
       return {
@@ -50,7 +59,8 @@ const startSubcommand: SlashCommand = {
       };
     }
 
-    const task = await getActiveAgent(context).startTask(args);
+    const agent = getActiveAgent(context);
+    const task = await agent.startTask(args);
 
     addClientHistory(
       context,
@@ -67,7 +77,7 @@ const startSubcommand: SlashCommand = {
 
 const stopSubcommand: SlashCommand = {
   name: 'stop',
-  description: 'Stops a running task',
+  description: 'Stops a running task. Usage: /bg stop <task_id>',
   action: async (context, args) => {
     if (!args || args.trim() === '') {
       return {
@@ -76,7 +86,8 @@ const stopSubcommand: SlashCommand = {
         content: 'The `stop` command requires a task id.',
       };
     }
-    await getActiveAgent(context).cancelTask(args);
+    const agent = getActiveAgent(context);
+    await agent.cancelTask(args);
     addClientHistory(context, `I canceled the background task with id ${args}`);
     return {
       type: 'message',
@@ -98,22 +109,19 @@ const listSubcommand: SlashCommand = {
       };
     }
 
-    const tasks = await getActiveAgent(context).listTasks();
+    const agent = getActiveAgent(context);
+    const tasks = await agent.listTasks();
     let content: string;
     if (tasks.length === 0) {
       content = 'No background tasks found.';
     } else {
       const taskList = tasks
         .map((task) => {
-          let statusMessage = getTaskStatusString(task).replace(
-            /\r?\n|\r/g,
-            ' ',
+          const shortStatus = toOneliner(
+            toMessageString(task.status.message),
+            MAX_STATUS_MESSAGE_LENGTH,
           );
-          if (statusMessage.length > MAX_STATUS_MESSAGE_LENGTH) {
-            statusMessage =
-              statusMessage.substring(0, MAX_STATUS_MESSAGE_LENGTH) + '...';
-          }
-          return `  - ${task.id}: (${task.status.state}) ${statusMessage}`;
+          return `  - ${task.id}: (${task.status.state}) ${shortStatus}`;
         })
         .join('\n');
       content = `Background tasks:\n${taskList}`;
@@ -128,7 +136,7 @@ const listSubcommand: SlashCommand = {
 
 const getSubcommand: SlashCommand = {
   name: 'get',
-  description: 'View a task',
+  description: 'View a task. Usage: /bg get <task_id>',
   action: async (context, args) => {
     if (!args || args.trim() === '') {
       return {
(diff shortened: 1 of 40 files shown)


packages/core/src/background/types.ts
```diff
@@ -0,0 +1,103 @@
...
+    'input-required',
+    'completed',
+    'failed',
+  ]),
+  message: BackgroundAgentMessageSchema.describe(
+    'Message describing the state of the task.',
+  ).optional(),
+});
+
+export const BackgroundAgentTaskSchema = z.object({
```
ngleo, line 84: Taking inspiration from A2A, would it be good to have some property to store the output? I see that there is a "history" property but I guess it serves a different purpose.

packages/core/src/background/agent.ts
```diff
@@ -0,0 +1,124 @@
...
+    await server.close();
+    throw error;
+  }
+}
+
+export class BackgroundAgent {
+  readonly startTaskTool: DiscoveredMCPTool;
+  readonly getTaskTool: DiscoveredMCPTool;
+  readonly listTasksTool: DiscoveredMCPTool;
+  readonly messageTaskTool: DiscoveredMCPTool;
```
ngleo: What is the purpose of messageTask? Is it supposed to be a generic tool to give context / edit task, etc?

ngleo (commented):
Nice! Regarding the design, I was thinking whether the background agent should have its own agentic loop (with the memory to keep track of the tasks), and then can dynamically set up MCP tools / talk to agents via A2A.

The current method with MCP is limiting in that the implementer has to build their own loop (for example, if I want a different agent to do a long running task, I might not want to set up a whole MCP scaffolding), but it could be a more conservative approach by limiting the scope of GC.

packages/core/src/background/types.ts
scidomino: We purposefully decided to punt on this for now. A2A has issues because it stores everything as artifacts on the task but there's no way to look at the task without downloading all the artifacts.

packages/core/src/background/agent.ts
scidomino: I shared a doc with you explaining the design.

scidomino:
> Nice! Regarding the design, I was thinking whether the background agent should have its own agentic loop (with the memory to keep track of the tasks), and then can dynamically set up MCP tools / talk to agents via A2A.
> 
> The current method with MCP is limiting in that the implementer has to build their own loop (for example, if I want a different agent to do a long running task, I might not want to set up a whole MCP scaffolding), but it could be a more conservative approach by limiting the scope of GC.

We plan to expose these tools to the agent but we're purposefully keeping this first version very simple. There's lots of places we could go from here. Take a look at the doc I sent you.

packages/cli/src/ui/commands/backgroundCommand.ts
```diff
@@ -0,0 +1,245 @@
...
+  action: async (context, args) => {
+    if (!args || args.trim() === '') {
+      return {
+        type: 'message',
+        messageType: 'error',
+        content: 'The `start` command requires a prompt.',
+      };
+    }
+
+    const task = await getActiveAgent(context).startTask(args);
```
NTaylorMullen: nit: separate this out as getting the agent then using said agent to start a task. Ditto elsewhere

packages/cli/src/ui/commands/backgroundCommand.ts
```diff
@@ -0,0 +1,245 @@
...
+
+    const tasks = await getActiveAgent(context).listTasks();
+    let content: string;
+    if (tasks.length === 0) {
+      content = 'No background tasks found.';
+    } else {
+      const taskList = tasks
+        .map((task) => {
+          let statusMessage = getTaskStatusString(task).replace(
+            /\r?\n|\r/g,
```
NTaylorMullen: Mind sharing motivation here?

packages/cli/src/ui/commands/backgroundCommand.ts
```diff
@@ -0,0 +1,245 @@
...
+  action: async (context, args) => {
+    if (!args || args.trim() === '') {
+      return {
+        type: 'message',
+        messageType: 'error',
+        content: 'The `log` command requires a task id.',
+      };
+    }
+    const task = await getActiveAgent(context).getTask(args);
+    const content = `Task logs for ${task.id}. status: (${task.status.state})\n${getTaskStatusString(task)}`;
```
NTaylorMullen: This seems to just be re-showing task status? Do we need the logs command?

packages/cli/src/ui/commands/backgroundCommand.ts
```diff
@@ -0,0 +1,245 @@
...
+      type: 'message',
+      messageType: 'info',
+      content,
+    };
+  },
+};
+
+const messageSubcommand: SlashCommand = {
+  name: 'message',
+  description: 'Send a message to a task',
```
NTaylorMullen: This should probably indicate the expected parameters. Probably true for all of the sub command desc's.

packages/core/src/background/agent.ts
NTaylorMullen, line 1: Lets call this `backgroundAgent.ts`

packages/cli/src/ui/commands/backgroundCommand.ts
```diff
@@ -0,0 +1,245 @@
...
+    role: 'model',
+    parts: [{ text: 'Got it.' }],
+  });
+}
+
+const startSubcommand: SlashCommand = {
+  name: 'start',
+  description: 'Start a new task with the provided prompt',
+  action: async (context, args) => {
+    if (!args || args.trim() === '') {
```
NTaylorMullen, line 54: If there's only 1 background agent do we want to make args pre-populate via the completion aspect of slash commands OR do we want to just have the id as optional?

packages/core/src/background/types.ts
```diff
@@ -0,0 +1,103 @@
...
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { z } from 'zod';
+import { Outcome, Language, FunctionResponseScheduling } from '@google/genai';
+
+// Should conform to Part in @google/genai
+export const PartSchema = z.object({
```
NTaylorMullen, line 11: I'm surprised you had to do this vs. just using what @google/genai has?

packages/core/src/tools/mcp-tool.ts
```diff
@@ -93,16 +93,19 @@ export class DiscoveredMCPTool extends BaseTool<ToolParams, ToolResult> {
     return confirmationDetails;
   }
 
-  async execute(params: ToolParams): Promise<ToolResult> {
+  async mcpCall(params: ToolParams): Promise<Part[]> {
```
NTaylorMullen: Why not just lean on execute?

NTaylorMullen (dismissed):
Very cool! Left a few higher level comments

packages/cli/src/ui/commands/backgroundCommand.ts
scidomino: We want each task's info to fit on one line.

scidomino: I reworked this for clarity.

packages/cli/src/ui/commands/backgroundCommand.ts
scidomino: Added code to print the history.

packages/core/src/background/agent.ts
scidomino: done.

packages/cli/src/ui/commands/backgroundCommand.ts
scidomino: Done. I followed the pattern from /chat

packages/cli/src/ui/commands/backgroundCommand.ts
scidomino: I have a follow up PR to autocomplete task ids when available. I also have one to autocomplete agent names in the /bg switch command.

packages/core/src/background/types.ts
scidomino: Me too! Unfortunately interfaces are compile time only and we need this at run time.

packages/core/src/tools/mcp-tool.ts
scidomino: reverted.

bdmorgan (approved):

PR #4417, opened 2025-07-17
Handle unhandled rejections more gracefully.

### LINKED ISSUE #4411: Gemini CLI crashes without showing an error unhandled rejections occur
### What happened?

If you introduce a bug such as running a tool throwing an exception, Gemini CLI will crash without a user visible error.

Expect to see an error logged to the debug console in Gemini CLI. If the error is particularly severe I expect the debug console to be opened for the user with a call to action to file a bug. I don't expect Gemini CLI to crash unless the error is truly non-recoverable.

### Client information

<details>

```console
$ gemini /about
# paste output here
```

</details>

### Login information

_No response_

_No response_

jacob314:
Previously Gemini CLI intentionally exited when an unhandled rejections occurred. Due to how we surface console.error messages when the App is running with ink, users got no message explaining the crash.

These errors weren't fatal so we are now surfacing them to the user with a call to action to file a bug instead of terminating.

As these errors do indicate logic errors users should file bugs about, we open the debug console the first time one of these errors is hit so the user cannot miss them.

To make this work well, I also ended up fixing a couple bugs in the debug console that were making behavior when multiple errors were thrown confusing.
There was a bug in useConsoleMessage likely related to React19 migration where messages were sometimes double counted. Switched to the more idiomatic reducer solution to avoid this. Also fixed the error count to increment properly when the same exact error was thrown multiple times in the row.

Video of the new behavior.
[Error demo.webm](https://github.com/user-attachments/assets/ed7b5670-b633-4181-8a51-6d61b3052536)

## Reviewer Test Plan

Modify Gemini CLI so it throws an exception when triggering an operation like running the `/help` command. Verify that the app doesn't crash and that the error is displayed. 
Screenshot 2025-07-17 at 4 00 21 PM

## Linked issues / bugs

- Fixes #4411

### DIFF
--- a/packages/cli/src/gemini.tsx
+++ b/packages/cli/src/gemini.tsx
@@ -40,6 +40,7 @@ import {
 import { validateAuthMethod } from './config/auth.js';
 import { setMaxSizedBoxDebugging } from './ui/components/shared/MaxSizedBox.js';
 import { validateNonInteractiveAuth } from './validateNonInterActiveAuth.js';
+import { appEvents, AppEvent } from './utils/events.js';
 
 function getNodeMemoryArgs(config: Config): string[] {
   const totalMemoryMB = os.totalmem() / (1024 * 1024);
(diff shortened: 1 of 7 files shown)


SandyTao520 (approved):

NTaylorMullen:
Will the error pane auto-open if a user gets a 500 or a 429 mid-request? Asking because I imagine that might cause flickering

jacob314:
> Will the error pane auto-open if a user gets a 500 or a 429 mid-request? Asking because I imagine that might cause flickering
Agree that would be really bad. The only time the error pane is auto opening is for the exceptions where we previously killed the application and worse didn't even show the user an error because we killed the app before we would log the error..

packages/cli/src/gemini.tsx
```diff
@@ -86,7 +87,30 @@ async function relaunchWithAdditionalArgs(additionalArgs: string[]) {
 }
 import { runAcpPeer } from './acp/acpPeer.js';
 
+export function setupUnhandledRejectionHandler() {
+  let unhandledRejectionOccurred = false;
+  process.on('unhandledRejection', (reason, _promise) => {
+    const errorMessage = `=========================================
+CRITICAL: Unhandled Promise Rejection!
+=========================================
+This is an unexpected error. Please file a bug report using the /bug tool.
```
NTaylorMullen: Can you put this line at the very top of this error so it's the first thing people read (especially the /bug part)

NTaylorMullen (approved):

packages/cli/src/gemini.tsx
jacob314: done

PR #4426, opened 2025-07-18
Improve instruction of replace tool documentation not to remove code

### LINKED ISSUE #4037: Gemini may replace code with placeholder comments in new_string to call the edit tool
### What happened?

I've noticed that Gemini sometimes removes parts of the code, replacing them with comments like `// (...rest of implementations)` or `// (...no changes around here)`.

Image

I believe this is due to insufficient instructions in `edit.ts`. Although the description for `new_string` already says, "`Provide the EXACT text.`," we may need to add more explicit instructions, such as, "`Do not omit any text from 'old_string'. Provide the EXACT and COMPLETE text.`"

https://github.com/google-gemini/gemini-cli/blob/main/packages/core/src/tools/edit.ts#L85-L125

Gemini shouldn't attempt removing code because Gemini tried omitting codes with comment in `new_string` parameter.

### Client information

This issue is not specific to client environment.

### Login information

_No response_

_No response_

kyasbal:
This commit clarifies the documentation for the 'replace' tool, specifically for the `new_string` parameter. The updated text emphasizes that `new_string` must be a complete, self-contained block of code and warns against using placeholders.

This change prevents Gemini CLI to replace code base with the `new_string` block including the placeholder as I mentioned in the issue #4037.

Resolves #4037, #7515

## TLDR

Gemini CLI sometimes uses the replace tool with specifying a placeholder comments like `// (unchanged fields...)` in `new_string` when the diff contains multiple unchanged lines. I tweaked the prompt not to use these place holder comments in the `new_string` field. Please see #4037 as well.

I modified this prompt in my local env and have tested for this week for the most of my development. I didn't see my code unintentionally removed after this patch although I see this unintentional remove around few times per day. 

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Closes #4037


gsquared94 (approved):

packages/core/src/tools/edit.ts
```diff
@@ -486,7 +487,7 @@ Expectation for required parameters:
           },
           new_string: {
             description:
-              'The exact literal text to replace `old_string` with, preferably unescaped. Provide the EXACT text. Ensure the resulting code is correct and idiomatic.',
+              'The exact literal text to replace `old_string` with, preferably unescaped. This MUST be the COMPLETE and EXACT text, including any unchanged lines from `old_string`. Do NOT use placeholder comments like `// ...` to abbreviate. Ensure the resulting code is correct and idiomatic.',
```
anj-s, line 490: how do we know this works? did we run any evals?

Issue #4067, opened 2025-07-13
Windows CTRL + Backspace dose not delete a word.


the-not-boring:
Currently performing the action dose not work just skips the word and takes the cursour focus to the start of the word user wants to delete, without even deleting it.

epodak:
Me too. Nor can the cursor be moved forward to modify the text.

PR #4432, opened 2025-07-18
fix(cli): improve Ctrl+Backspace word deletion functionality

### LINKED ISSUE #3101: Ctrl+Backspace to delete whole words from prompt

Delete whole words with Ctrl+Backspace or Shift+Backspace.

### Why is this needed?

It sometimes happens, that I accidentally paste wrong data or code into a prompt.

To remove that code I have to press or hold backspace which deletes every character one by one.

Unfortunately that is very slow for long text. I can usually in most shells and text editors delete whole words with Ctrl+Backspace or Shift+Backspace. That would speed up the process a lot.

I hope you consider this as a feature for upcoming releases. ❤️

### Additional context

_No response_

Rebraws:
Fix Ctrl+Backspace word deletion that wasn't working in certain cases. Add platform-specific detection (Linux: \x08, Windows: \x7f) and environment variable control via GEMINI_CLI_CTRL_BACKSPACE_MODE.

Fixes #3101

## TLDR

 Fixes Ctrl+Backspace word deletion functionality that wasn't working in certain terminals
   by improving key sequence detection and adding platform-specific handling.
### Problem

Different terminals handle backspace and Ctrl+Backspace inconsistently, creating compatibility issues:

- **Most Linux terminals by default** (st, alacritty, kitty, rxvt-unicode, xfce-terminal, terminator, etc.) use:
  - `\x7f` for backspace
  - `\x08` for Ctrl+Backspace

- **Other terminals like Windows Terminal or xterm** use:
  - `\x08` for backspace
  - Different sequence for Ctrl+Backspace

Although gemini-cli has logic to support Ctrl+Backspace, it relies on Node's readline module for key processing. When readline encounters the `\x08` byte, it always interprets it as a regular backspace:

```javascript
{
  sequence: '\x08',
  name: 'backspace',
  ctrl: false,
  // ...
}
```

This causes the following conditional to never evaluate to `true` for Ctrl+Backspace in terminals that use `\x08` or `\x7f` for ctrl+backspace:

```javascript
else if (
  (key.meta || key.ctrl) &&
  (key.name === 'backspace' || input === '\x7f')
)
  deleteWordLeft();
```

We cannot simply check if the byte is `\x08` and set the `ctrl` flag to true, because this would cause terminals like Windows terminal to treat regular a backspace as Ctrl+Backspace.

While it would be ideal to detect which byte sequence the terminal uses for backspace, the most straightforward solution is to implement an environment variable that users can set to enable Ctrl+Backspace functionality, this will avoid breaking existing functionality while solving the issue on most terminals.

### Implementation

The implemented solution uses a feature flag approach with platform-aware detection, when linux users opt-in via `GEMINI_CLI_CTRL_BACKSPACE_MODE=true` we manually set the `ctrl` flag as true for the byte `\x08` (this will make ctrl+backspace work on most terminals), and for Windows users we do the same but for `\x7f`

The environment variable approach ensures backward compatibility while solving the issue for users who need it.

**Opt-in behavior**: Users can enable when needed
**Backward compatibility**: Existing workflows remain unchanged

## Reviewer Test Plan

  1. **Enable the feature**: Set `export GEMINI_CLI_CTRL_BACKSPACE_MODE="true"`
  2. **Test previously broken scenarios**:
     - Type "hello world test"
     - Press Ctrl+Backspace → should now reliably delete "test" on terminals that use `x08` on linux or `x7f` on windows. 

The implementation includes tests in two separate files:

`useKeypress.test.ts`: Tests the low-level key detection logic - verifies that the `ctrl` flag gets set correctly based on platform and byte sequence when the environment variable is enabled.
`text-buffer.test.ts`: Tests the ctrl+backspace word deletion behavior, ensures that when `ctrl` is `true`, the `deleteWordLeft()` function actually gets called and removes the correct text.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

- Fixes: #3101 
- #4067 Duplicate of 3101

### DIFF
--- a/packages/cli/src/ui/contexts/KeypressContext.tsx
+++ b/packages/cli/src/ui/contexts/KeypressContext.tsx
@@ -75,6 +75,11 @@ const ALT_KEY_CHARACTER_MAP: Record<string, string> = {
   '\u03A9': 'z',
 };
 
+const CTRL_BACKSPACE_SEQUENCES = {
+  win32: '\x7f',
+  default: '\x08',
+} as const;
+
 export interface Key {
   name: string;
   ctrl: boolean;
@@ -109,11 +114,13 @@ export function useKeypressContext() {
 export function KeypressProvider({
   children,
   kittyProtocolEnabled,
+  ctrlBackspaceModeFix,
   config,
   debugKeystrokeLogging,
 }: {
   children: React.ReactNode;
   kittyProtocolEnabled: boolean;
+  ctrlBackspaceModeFix?: boolean;
   config?: Config;
   debugKeystrokeLogging?: boolean;
 }) {
@@ -763,6 +770,19 @@ export function KeypressProvider({
       if (key.name === 'return' && key.sequence === `${ESC}\r`) {
         key.meta = true;
       }
+
+      // readline doesn't set the ctrl flag for Ctrl+Backspace sequences
+      // (\x7f, \x08), so we set it manually when the ctrlBackspaceModeFix setting is enabled
+      if (ctrlBackspaceModeFix && key.name === 'backspace' && !key.ctrl) {
+        const expectedSequence =
+          process.platform === 'win32'
+            ? CTRL_BACKSPACE_SEQUENCES.win32
+            : CTRL_BACKSPACE_SEQUENCES.default;
+        if (key.sequence === expectedSequence) {
+          key.ctrl = true;
+        }
+      }
+
       broadcast({ ...key, paste: isPaste });
     };
 
@@ -907,6 +927,7 @@ export function KeypressProvider({
     config,
     subscribers,
     debugKeystrokeLogging,
+    ctrlBackspaceModeFix,
   ]);
 
   return (
--- a/docs/get-started/configuration.md
+++ b/docs/get-started/configuration.md
@@ -119,6 +119,12 @@ their corresponding top-level category object in your `settings.json` file.
   - **Description:** Enable debug logging of keystrokes to the console.
   - **Default:** `false`
 
+- **`general.ctrlBackspaceModeFix`** (boolean):
+  - **Description:** Enables a fix for `Ctrl+Backspace` functionality in the CLI
+    prompt. This is useful if you are experiencing issues with word deletion in
+    your terminal.
+  - **Default:** `false`
+
 - **`general.sessionRetention.enabled`** (boolean):
   - **Description:** Enable automatic session cleanup.
   - **Default:** `false`
--- a/packages/cli/src/config/settingsSchema.ts
+++ b/packages/cli/src/config/settingsSchema.ts
@@ -203,6 +203,16 @@ const SETTINGS_SCHEMA = {
         description: 'Enable debug logging of keystrokes to the console.',
         showInDialog: true,
       },
+      ctrlBackspaceModeFix: {
+        type: 'boolean',
+        label: 'Ctrl+Backspace mode fix',
+        category: 'General',
+        requiresRestart: true,
+        default: false,
+        description:
+          'Enable fix for Ctrl+Backspace functionality on non-kitty terminals',
+        showInDialog: false,
+      },
       sessionRetention: {
         type: 'object',
         label: 'Session Retention',
--- a/packages/cli/src/gemini.tsx
+++ b/packages/cli/src/gemini.tsx
@@ -175,6 +175,7 @@ export async function startInteractiveUI(
           kittyProtocolEnabled={kittyProtocolStatus.enabled}
           config={config}
           debugKeystrokeLogging={settings.merged.general?.debugKeystrokeLogging}
+          ctrlBackspaceModeFix={settings.merged.general?.ctrlBackspaceModeFix}
         >
           <SessionStatsProvider>
             <VimModeProvider settings={settings}>
--- a/packages/cli/src/ui/components/shared/text-buffer.test.ts
+++ b/packages/cli/src/ui/components/shared/text-buffer.test.ts
@@ -430,6 +430,111 @@ describe('useTextBuffer', () => {
     });
   });
 
+  describe('Word Deletion ctrl+backspace', () => {
+    it('should handle ctrl+backspace to delete word left', () => {
+      const { result } = renderHook(() =>
+        useTextBuffer({
+          initialText: 'ab cd',
+          viewport,
+          isValidPath: () => false,
+        }),
+      );
+
+      act(() => result.current.move('end'));
+      act(() =>
+        result.current.handleInput({
+          name: 'backspace',
+          ctrl: true,
+          meta: false,
+          shift: false,
+          paste: false,
+        }),
+      );
+      const state = getBufferState(result);
+      expect(state.text).toBe('ab ');
+      expect(state.cursor).toEqual([0, 3]);
+    });
+
+    it('should handle ctrl+backspace at line beginning to merge lines', () => {
+      const { result } = renderHook(() =>
+        useTextBuffer({
+          initialText: 'hello\nworld',
+          viewport,
+          isValidPath: () => false,
+        }),
+      );
+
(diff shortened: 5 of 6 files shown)


packages/cli/src/ui/hooks/useKeypress.ts
```diff
@@ -87,6 +87,27 @@ export function useKeypress(
           if (key.name === 'return' && key.sequence === '\x1B\r') {
             key.meta = true;
           }
+
+          // readline doesn't set the ctrl flag for byte sequences x7f and x08
+          // so we need to set it up manually when needed
+          if (
+            key.name === 'backspace' &&
+            !key.ctrl &&
+            process.env.GEMINI_CLI_CTRL_BACKSPACE_MODE === 'true'
```
jacob314: remove this env variable. I don't think this needs a setting but if it did, we should instead add it to settings.ts rather than checking env variables directly. that is the generally preferred way for us to manage user options that some users might want to opt into.

docs/cli/configuration.md
```diff
@@ -330,6 +330,12 @@ The CLI automatically loads environment variables from an `.env` file. The loadi
 - **`CODE_ASSIST_ENDPOINT`**:
   - Specifies the endpoint for the code assist server.
   - This is useful for development and testing.
+- **`GEMINI_CLI_CTRL_BACKSPACE_MODE`**:
```
jacob314: remove this

jacob314 (commented):
Thanks for adding! lgtm once the env variable is removed. if this is risky in a way I'm not thinking about I'm also open for guarding it behind a setting as discussed in the comment.

jacob314:
One other thought is it would be ideal if of mac we also supported option-delete and option-backspace which is the default for most mac editors.

packages/cli/src/ui/hooks/useKeypress.ts
Rebraws: It would be great to get rid of the env/setting variable, but I'm not sure if it can be done so easily, the main issue I see is the inconsistency between terminals to handle backspace and ctrl+backspace, let's say we have two terminals

  Terminal 1: Handles backspace as `0x08` and ctr+backspace as `0x17`
  Terminal 2: Handles backspace as `0x7f` and ctrl+backspace as `0x08`

Currently ctrl+backspace works on Terminal 1 but it doesn't work on Terminal 2, by enabling the setting we'll make ctrl+backspace work on Terminal 2 but that would break how backspace works on Terminal 1, because we'll be setting the `ctrl` flag to true for `0x08`, if we enable this fix unconditionally, we could break the normal backspace in some terminals (like xterm).

By guarding this under a setting we could make the functionality work on both terminals, maybe the name of the setting isn't the best since ctrl+backspace will work by default on certain terminals like the terminal on vscode, perhaps a name like `legacyCtrlBackspaceFix` or `ctrlBackspaceFix` would be better to clarify that this addresses terminal-specific compatibility issues.


Moving the environment variable to settings definitely sounds much better.

jacob314:
Heads up that we're adding support for shift+enter which you may be able to build upon to handle this robustly in most terminals.
https://github.com/google-gemini/gemini-cli/pull/3289
Let me know if you have any questions.

Rebraws:
I've been looking at the PR you linked and adding support for kitty protocol sounds great, but from what I've been testing it appears these changes are still necessary, particularly for terminals that don't support the kitty protocol. For terminals that do support it, we could add functionality to handle the corresponding byte sequence.

I tested the PR you referenced and confirmed it works well with that addition. Here's a test on kitty terminal with the debug console open to show the byte sequence:


https://github.com/user-attachments/assets/e3679ce3-162d-4ce7-b9b7-9f1e9afb77da




Once that change merges, maybe we could handle the byte sequence shown in the video: \1xB[127;5u

In my latest commit, I've refactored the code to move away from the environment variable approach to a setting-based configuration, as you suggested. As I mentioned in a comment, I believe we need to guard this feature behind a setting to prevent breaking compatibility with some terminals.

Here's an example demonstrating where this feature could break the default experience for some users. The left terminal is rxvt-unicode, and the right terminal is xterm:



https://github.com/user-attachments/assets/bec93150-9150-4cea-8320-56d87238dfd2



With the feature disabled, both terminals handle Ctrl+Backspace and Backspace identically. When enabled, Ctrl+Backspace works correctly in the left terminal, but it breaks the Backspace key functionality in the right terminal.


About Mac: 

From my testing it seems that option+backspace and option+del already work but the user needs to enable the feature to handle escape sequences in the terminal,  here's a video testing it with Iterm2. 


https://github.com/user-attachments/assets/d4be9003-8c69-4430-ad59-4e6449f8ad49

Let me know what you think about the changes or if you'd like any adjustments

jacob314:
The kitty protocol changes have finally landed. We'd welcome your contributions on top of them in particular to help for non-kitty terminals.

Rebraws:
I've rebased this branch to catch up with the significant changes that happened since my last push.  I've tested the changes locally, but given the scope of the rebase, please let me know if you notice any issues.  

Note that these changes don't address the issue for kitty terminals, as I  mention in a previous comment I think handling the case of the `\1xB[127;5u` byte sequence would be enough for those,  I didn't include that change in this PR because it would likely be better to handle it in a separate, more focused PR and keep this one just for non-kitty terminals.

packages/cli/src/ui/contexts/KeypressContext.tsx
```diff
@@ -381,7 +401,7 @@ export function KeypressProvider({
         stdin.removeListener('keypress', handleKeypress);
       }
 
-      rl.close();
+      rl?.close();
```
jacob314: please revert this change. rl cannot be undefined.

packages/cli/src/ui/contexts/KeypressContext.tsx
```diff
@@ -304,6 +311,19 @@ export function KeypressProvider({
       if (key.name === 'return' && key.sequence === `${ESC}\r`) {
         key.meta = true;
       }
+
+      // readline doesn't set the ctrl flag for byte sequences x7f and x08
```
jacob314: this logic looks like it runs even if the terminal uses kitty. Can you make it more explicit that this logic is only applied if the terminal is not kitty if htat is your intend?

jacob314 (approved):
lgtm

packages/cli/src/ui/contexts/KeypressContext.tsx
Rebraws: addressed

packages/cli/src/ui/contexts/KeypressContext.tsx
Rebraws: Do you mean like a check for whether the Kitty Protocol is enabled? I didn't think that was necessary because the logic only runs when the user enables the feature and Kitty users won't need to enable it once the byte sequence for kitty terminals is handled. 


I've updated the comment to clarify this and also moved the feature flag to the beginning of the if statement to make it clearer that the entire block is gated by that setting. Is that sufficient, or would you still like me to add something else?

jacob314:
Sorry I missed landing this. If you resolve the merge conflicts I will approve.

Rebraws:
No worries, I also forgot about this PR too :rofl: Just rebased and resolved the conflicts, I test it out locally and seems to be working fine. Let me know if you need anything else

gsehgal:
@Rebraws , can you please resolve the conflict one more time. Sorry for the convenience. I will take it as priority to get this merged.

jacob314:
Closing as the way we handle keypresses has significantly changed. Good news is @scidomino has dramatically improved the robustness of this support.

PR #4460, opened 2025-07-18
feat(memory): make directory search limit on memory discovery configurable with settings.json

KeijiBranshi:
This change makes the directory search limit for hierarchical memory discovery configurable.

A new memoryDiscoveryMaxDirs field can be added to the user's settings.json to override the default limit of 200 directories.

To implement this, the refreshMemory method was removed from the core config service and its logic was moved into the memoryCommand. This was done to avoid adding the new memoryDiscoveryMaxDirs setting to the core Config interface, keeping the change localized to the CLI package where it is used.

**Recommended Review Order**

   1. packages/core/src/services/memory/memoryDiscovery.ts (to see the new parameter)
   2. packages/cli/src/config/config.ts (to see the setting being passed down)
   3. packages/cli/src/ui/commands/memoryCommand.ts (to see the refactored refresh logic)
   4. packages/cli/src/config/settings.ts (to see the new setting definition)
   5. packages/cli/src/ui/App.tsx (to see the setting passed during initial load)

The remaining changes are corollary updates to tests and call sites.

----

[DEPRECATED PR DESCRIPTION BEFORE INITIAL REVIEW. LEFT TO MAINTAIN THE PAPER TRAIL]

~~This change introduces a configuration option for the memory discovery mechanism.~~

~~The primary change is the addition of the `GEMINI_MEMORY_DISCOVERY_MAX_DIRS` environment variable. This variable allows users to configure the maximum number of directories that will be scanned during the downward traversal phase of memory discovery. The default value is 200.~~

~~A test case has been added to verify that the `GEMINI_MEMORY_DISCOVERY_MAX_DIRS` environment variable is respected. The documentation in `docs/cli/configuration.md` has been updated to reflect this new environment variable and its effect on the breadth of the directory search.~~

### DIFF
--- a/packages/cli/src/config/config.ts
+++ b/packages/cli/src/config/config.ts
@@ -220,6 +220,7 @@ export async function loadHierarchicalGeminiMemory(
   currentWorkingDirectory: string,
   debugMode: boolean,
   fileService: FileDiscoveryService,
+  settings: Settings,
   extensionContextFilePaths: string[] = [],
   fileFilteringOptions?: FileFilteringOptions,
 ): Promise<{ memoryContent: string; fileCount: number }> {
@@ -237,6 +238,7 @@ export async function loadHierarchicalGeminiMemory(
     fileService,
     extensionContextFilePaths,
     fileFilteringOptions,
+    settings.memoryDiscoveryMaxDirs,
   );
 }
 
@@ -293,6 +295,7 @@ export async function loadCliConfig(
     process.cwd(),
     debugMode,
     fileService,
+    settings,
     extensionContextFilePaths,
     fileFiltering,
   );
--- a/packages/core/src/utils/memoryDiscovery.ts
+++ b/packages/core/src/utils/memoryDiscovery.ts
@@ -33,8 +33,6 @@ const logger = {
     console.error('[ERROR] [MemoryDiscovery]', ...args),
 };
 
-const MAX_DIRECTORIES_TO_SCAN_FOR_MEMORY = 200;
-
 interface GeminiFileContent {
   filePath: string;
   content: string | null;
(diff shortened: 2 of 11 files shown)


packages/core/src/utils/memoryDiscovery.ts
```diff
@@ -29,7 +29,8 @@ const logger = {
     console.error('[ERROR] [MemoryDiscovery]', ...args),
 };
 
-const MAX_DIRECTORIES_TO_SCAN_FOR_MEMORY = 200;
+const MAX_DIRECTORIES_TO_SCAN_FOR_MEMORY =
+  parseInt(process.env.GEMINI_MEMORY_DISCOVERY_MAX_DIRS ?? '', 10) || 200;
```
jacob314: it would make this a bit more annoying to use as settings.json is in the CLI package but what do you think of making this be in settings.ts rather than an env variable? this kindof seems like something I might want to set in settings.json. For example, maybe my project is really large so I want a specific value for it.
in general I'm trying to keep most settings in settings.ts where feasible rather than encouraging more env variables or options in config controlled by args unless they strictly have to go that way. Happy to be convinced otherwise for this case.

KeijiBranshi: I like settings.json! I'll migrate this over to that.

Passing overkill thought: at some point I think it'd be great to have a cascading hierarchy of settings like this made more explicit in the `Config` construction. Something like

yargs (CLI args) > settings.json (local project + home directory) > environment variables (`.env` + system) > hardcoded default

packages/core/src/utils/memoryDiscovery.ts
```diff
@@ -29,7 +29,8 @@ const logger = {
...
```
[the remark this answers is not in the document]
KeijiBranshi: outdated after implementing the suggestion in https://github.com/google-gemini/gemini-cli/pull/4460#discussion_r2216965591

packages/cli/src/config/config.ts
```diff
@@ -218,6 +218,7 @@ export async function loadHierarchicalGeminiMemory(
   currentWorkingDirectory: string,
   debugMode: boolean,
   fileService: FileDiscoveryService,
+  settings: Settings,
```
jacob314, line 221: thanks! this will be easier for the next person who needs to flow through a change like this.

jacob314 (approved):
lgtm

Issue #3907, opened 2025-07-11
gemini-cli companion extension MCP server


skeshive:
### What would you like to be added?

Stand up a MCP server within the companion gemini-cli VS code extension. This MCP server will be responsible for:

1) Listening for open file changes
2) Notifying clients of open file changes (current file + cursor position + list of open files) via SSE

### Why is this needed?

// TODO

skeshive:
We need to run evals before adding this context to the prompt

PR #4463, opened 2025-07-18
[ide-mode] Keep track of recently-opened files and send them to the CLI

chrstnb:
## TLDR

Keeps a running list of recently-active files and stores their timestamp. Removes them from the list if they are deleted or closed. Sends this list as part of the activeFile notification any time there is a change.

## Dive Deeper

There weren't any VSCode APIs that do this natively (from what I could find!).

Also, added an environment variable for configuring the # of recent files since we may want to tune this.

Will follow up with the gemini-cli-side change.

## Reviewer Test Plan

Run locally and verify notifications are being sent as expected.
## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

#3907

### DIFF
--- a/packages/core/src/services/ideContext.ts
+++ b/packages/core/src/services/ideContext.ts
@@ -26,6 +26,14 @@ export type Cursor = z.infer<typeof CursorSchema>;
 export const ActiveFileSchema = z.object({
   filePath: z.string(),
   cursor: CursorSchema.optional(),
+  recentOpenFiles: z
+    .array(
+      z.object({
+        filePath: z.string(),
+        timestamp: z.number(),
+      }),
+    )
+    .optional(),
 });
 export type ActiveFile = z.infer<typeof ActiveFileSchema>;
 
--- a/packages/vscode-ide-companion/src/recent-files-manager.ts
+++ b/packages/vscode-ide-companion/src/recent-files-manager.ts
@@ -6,26 +6,8 @@
 
 import * as vscode from 'vscode';
 
-const DEFAULT_MAX_RECENT_FILES = 10;
-
-export function getMaxRecentFiles() {
-  const value = process.env['IDE_MODE_MAX_RECENT_FILES'];
-  if (!value) {
-    return DEFAULT_MAX_RECENT_FILES;
-  }
-  const parsed = parseInt(value, 10);
-  return isNaN(parsed) ? DEFAULT_MAX_RECENT_FILES : parsed;
-}
-
-// Threshold of minutes since the file was last active.
-export function getMaxFileAge() {
-  const value = process.env['IDE_MODE_MAX_FILE_AGE_MINUTES'];
-  if (!value) {
-    return 10;
-  }
-  const parsed = parseInt(value, 10);
-  return isNaN(parsed) ? 10 : parsed;
-}
+export const MAX_FILES = 10;
+export const MAX_FILE_AGE_MINUTES = 5;
 
 interface RecentFile {
   uri: vscode.Uri;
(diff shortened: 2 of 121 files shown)


packages/vscode-ide-companion/src/recent-files-manager.ts
skeshive, line 1: hmmmm I wonder if we can simplify this logic a bit? something like:

when the active file changes, add it to the front of the queue (if its not already at the front) and then if the queue length is > max recent files, we pop the oldest?

in the prompt we could have something like:

"the user is looking at this file {active file}
and recently opened the following files...{recently opened files}"

IMO we don't need to track file deletions / closing active files if we use the above wording in the prompt, WDYT?

chrstnb: I don't think we should include deleted files. Have no opinion about closed files, though. I think we might want to play around with this and see how it feels one way or the other-- I have a feeling we may want to keep the thresholds pretty small. On the local agent side, we only send other active files if the user explicitly includes them-- I think we might find that this is a little confusing to the model

chrstnb: Let's revisit as we play around with it!

packages/core/src/services/ideContext.ts
```diff
@@ -26,6 +26,14 @@ export type Cursor = z.infer<typeof CursorSchema>;
 export const ActiveFileSchema = z.object({
```
skeshive, line 26: nit (ok to address in a follow-up): lets rename this to openFilesSchema and update filePath to activeFilePath since the scope of the notification is expanding

packages/vscode-ide-companion/src/recent-files-manager.ts
```diff
@@ -0,0 +1,111 @@
...
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import * as vscode from 'vscode';
+
+const DEFAULT_MAX_RECENT_FILES = 10;
+
+export function getMaxRecentFiles() {
+  const value = process.env['IDE_MODE_MAX_RECENT_FILES'];
```
skeshive: is the intent here to allow us to set the env variable on the gemini-cli side so we can dynamically change this setting without relying on the user to update their extension?

if so, this is probably worth testing to see if the extension can access external env variables. if this doesn't end up working, I think we can just hardcode the number in the extension. IMO its unlikely we'll need to tune the number of max recent files frequently

packages/vscode-ide-companion/src/recent-files-manager.ts
```diff
@@ -0,0 +1,111 @@
...
+    this.maxRecentFiles = maxRecentFiles ?? getMaxRecentFiles();
+    this.maxFileAge = maxFileAge ?? getMaxFileAge();
+    const editorWatcher = vscode.window.onDidChangeActiveTextEditor(
+      (editor) => {
+        if (editor) {
+          this.add(editor.document.uri);
+        }
+      },
+    );
+    const fileWatcher = vscode.workspace.onDidDeleteFiles((event) => {
```
skeshive, line 36: should we also handle file renames explicitly?

skeshive:
Not related to this PR but could you please remove this logger.show() line? This causes the output debug logs to pop up anytime the extension is activated, which we don't need to show to users

https://github.com/google-gemini/gemini-cli/blob/main/packages/vscode-ide-companion/src/extension.ts#L15C1-L16C1

skeshive (changes requested):

packages/vscode-ide-companion/src/recent-files-manager.ts
skeshive: if we do keep the env var approach, can you rename them to be consistent with the port env var? e.g. prefix with GEMINI_CLI_IDE_*

chrstnb: I'll just remove for now

packages/core/src/services/ideContext.ts
chrstnb: Cool-- will update in the PR I had on top of this one!

skeshive (approved):

bdmorgan (approved):

PR #4475, opened 2025-07-18
Automatically detect non-interactive environments and fall back to a manual, code-based authentication flow 

### LINKED ISSUE #1437: Login with Google - Doesn't support (SSH Issue?) for headless login
It's stuck at ⠙ Waiting for auth... (Press ESC to cancel)  , shouldn't it provide a url for me to authenticate? 

For context, i'm doing SSH into my server so this isn't my computer

---------

**Update - Temporary solution**
Execute `npx https://github.com/google-gemini/gemini-cli --debug`

Proceed to accept and signin on browser via the provided link , you will receive a call back that looks like this on browser `http://localhost:41725/oauth2callbac`

Open another terminal , connect & copy the whole URL and do this

`curl "http://localhost:41725/oauth2callbac...."`

Get started coding and cancel your claude code subscription 😂

### LINKED ISSUE #2546: Failure to Login with google account
I am having a problem with the gemini cli when login in using my google account, the redirect to the login works and it accepts it but now in the cli it self I am stuck at waiting for auth I don't know why.

Image

### LINKED ISSUE #3390: Gemini CLI cannot authenticate on VM
### What happened?

Hi, 

I can install and run the gemini CLI without issues, but after choosing the 'login with google' option, the CLI just hangs: 
` ⠇ Waiting for auth... (Press ESC to cancel) `

There is no link shown that I can use to login. My guess is that the CLI is trying to open a browser window directly, but I am running the terminal `ssh-ing` into a VM, where there is no browser. 

1. Would it be possible to print the authentication link to the terminal, so I can copy paste it and open it in the browser? 
2. I have already authenticated with `gcloud`, are there plans to allow the authentication to be reused? 

I would expect to be able to authenticate 

### Client information

<details>

I cannot paste the output of gemini `/about` because the CLI is stuck on auth and does not allow me to issue any commands. 

</details>

### Login information

_No response_

_No response_

mboshernitsan:
## TLDR

Automatically detect non-interactive environments (like SSH sessions or CI/CD) and fall back to a manual, code-based authentication flow instead of attempting to launch a web browser.

This change checks for common environment variables (`CI`, `SSH_CONNECTION`, `DISPLAY`, etc.) to determine if a GUI browser is likely to be available. If a browser launch fails in an interactive environment, the application now advises the user to restart with NO_BROWSER=true, which is the correct procedure for the authcode flow. 

This improves the user experience in headless or remote environments and makes the authentication process more robust.

Finally this PR updates auth code redirect to the official one with correct branding

## Reviewer Test Plan

Please test on different flavors of headless environments.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | x  | ❓  | x  |
| npx      | ❓  | ❓  | x  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

This is a final PR in series of related changes and finally fixes #2546, #3390, #1437


packages/core/src/code_assist/oauth2.ts
```diff
@@ -121,7 +122,7 @@ export async function getOauthClient(
     }
   }
 
-  if (config.getNoBrowser()) {
+  if (config.getNoBrowser() || !shouldAttemptBrowserLaunch()) {
```
NTaylorMullen, line 125: Can we change the `getNoBrowser` to be `config.isBrowserless()` which also encompasses this `shouldAttemptBrowserLaunch` logic?

NTaylorMullen (approved):

packages/core/src/code_assist/oauth2.ts
mboshernitsan: My mental model is that config.getNoBrowser() accesses the provided setting (from the settings file or the env var), whereas shouldAttemptBrowserLaunch() attempts to detect the correct behavior from the environment, but does not persists its result.  I think they are logically separate, no?  I can see adding another utility function that combines the logic, but it might be an overkill given that it is used in exactly three places right now.

PR #1743, opened 2025-06-26
feat(cli): add support for --prompt-interactive/-i flag

taeold:
Supersedes: https://github.com/google-gemini/gemini-cli/pull/1277, fixes https://github.com/google-gemini/gemini-cli/issues/2594

## TLDR

This change introduces the "--prompt-interactive" (or "-i") flag that allows users to execute a prompt and then continue in interactive mode. This allows for a seamless transition from a single-shot command to a continued conversation.

## Dive Deeper

When a prompt is provided via `--prompt` (`-p`), the CLI typically processes the request and exits. With the new `--prompt-interactive` (`-i`) flag, the CLI will execute the provided prompt and then enter an interactive session.

This change also explicitly disallows using the `--prompt-interactive` flag when piping from standard input. This is a deliberate design choice, as interactive mode requires a direct connection to the user's terminal (a TTY) to handle keyboard input, which is unavailable when `stdin` is redirected from another process. This restriction prevents crashes and ensures predictable behavior.

### Real-World Examples

- **Python:** The `-i` flag can run an inline command with `-c` and then drop into an interactive session where the state from the command is preserved.

```bash
$ python3 -i -c "print('Hello from inlined script'); x = 10"
  >>> x
  10
  >>>
```

- **Node.js:** The `-i` flag forces the REPL to start, even when input is piped from another process, allowing inspection of the result.

```bash
$ node -e "console.log('Hello from inlined script'); x = 10" -i
Welcome to Node.js v22.12.0.
Type ".help" for more information.
> Hello from inlined script
x
10
> x
>
```

- **Docker:** The `-i` flag (combined with `-t` for a TTY) is fundamental for running an interactive shell inside a container.

```bash
$ docker run -it ubuntu bash
root@a1b2c3d4e5f6:/# ls
bin  boot  dev  etc  home  lib ...
root@a1b2c3d4e5f6:/# exit
```

## Reviewer Test Plan

1.  **Verify standard non-interactive behavior:** Run the CLI with a prompt and confirm it exits after providing the answer. 

```bash
$ gemini -p "hello"
Hi there\! What can I help you with today?
```

2.  **Verify interactive behavior with prompt:** Run the CLI with `-i` and a prompt, and confirm it enters an interactive session after executing the prompt.

```bash
$ gemini -i "hello"

 ███            █████████  ██████████ ██████   ██████ █████ ██████   █████ █████
[...]

PR #4484, opened 2025-07-19
Document --prompt-interactive/-i command-line argument

lambdacalculator:
This PR adds documentation to [docs/cli/configuration.md](docs/cli/configuration.md) for the --prompt-interactive (-i) command-line argument, which was implemented in PR #1743 but not documented.

## TLDR

This just adds one more entry to the "Command-line arguments" section of [docs/cli/configuration.md](docs/cli/configuration.md) for an already-implemented but undocumented option.

## Dive Deeper

I  had wanted to implement this feature myself for my own use-cases until examining the code and finding that it had already been done. I'm Just finishing up the documentation job so that other users can take advantage of the option and other developers don't have to go down the same path I did.

## Reviewer Test Plan

The change is only to documentation, not code. A simple check to verify 

- that the `-i` option has indeed been implemented, and
- the truth of the second sentence in my proposed documentation ("Cannot be used in conjunction with `--prompt` (`-p`).") 

should be all that is needed. A reviewer might also reasonably decide that this second sentence is not necessary and leave it out as being too obvious.

## Testing Matrix

No testing is required

## Linked issues / bugs

Feature was originally implemented in PR #1743.

### DIFF
--- a/docs/cli/configuration.md
+++ b/docs/cli/configuration.md
@@ -340,6 +340,8 @@ Arguments passed directly when running the CLI can override other configurations
   - Example: `npm start -- --model gemini-1.5-pro-latest`
 - **`--prompt <your_prompt>`** (**`-p <your_prompt>`**):
   - Used to pass a prompt directly to the command. This invokes Gemini CLI in a non-interactive mode.
+- **`--prompt-interactive <your_prompt>`** (**`-i <your_prompt>`**):
+  - Used to pass a prompt directly to the command, but continues in interactive mode afterward. Cannot be used in conjunction with `--prompt` (`-p`).
 - **`--sandbox`** (**`-s`**):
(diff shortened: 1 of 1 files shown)


lambdacalculator:
Thank you for the automated suggestion! I tested the CLI with `-i` and piped input (e.g., `echo "prompt" | gemini -i`) and found that it behaves the same as `-p`: it processes the input as a one-shot prompt and exits. There doesn’t seem to be any special interactive fallback when `stdin` is piped; both options send the prompt directly to Gemini.

I propose updating the documentation to clarify that `-i` behaves like `-p` when used with piped input, to prevent confusion.

adamfweidman:
HI @lambdacalculator thanks for this contribution, sorry we are getting to it late. Looks like this was added in #4950 so I'm closing this PR.

Issue #3789, opened 2025-07-10
[Tracking Issue] The Future of Slash Commands: Towards an Extensible Automation Platform


abhipatel12:
Hey everyone!

With the recent merge of the slash command architectural refactor (see PR #3175), we've successfully laid a new foundation for what's possible with commands in the Gemini CLI.

This refactor was the necessary first step towards a much broader vision. Now that the groundwork is in place, we want to open up a community-wide discussion about where we go next.

### The Vision: From Commands to a Platform

Our long-term vision is to evolve slash commands from a simple convenience feature into a powerful, user-driven automation platform. We want to empower you to teach Gemini CLI your unique workflows by creating a library of powerful, composable commands tailored to your projects.

### The New Foundation: How We'll Get There

The recent refactor introduced the `CommandService`, which is the key to this entire vision. It acts as a central hub that can discover commands from multiple sources and build a unified command tree.

```
+-------------------+      +---------------------+      +--------------------+
|   Built-in cmds   |      | Local Command Files |      | MCP Server Prompts |
| (TS Objects with  |      |  (.gemini.toml with |      | (via an API client)|
|  subCommands)     |      |   subCommands)      |      +--------------------+
+-------------------+      +---------------------+               |
         |                        | (Discovers & Parses)         | (Discovers via API)
         | (Loads)                |                              |
         +------------------------+------------------------------+
                                  |
                      +---------------------------+
                      |      CommandService       |
                      | (Builds unified command   |
                      |           TREE)           |
                      +---------------------------+
                                  |
                                  | (Provides Tree)
                                  v
                    +-----------------------------+
                    |  CLI UI / useSlashProcessor |
                    | (Traverses tree to execute) |
                    +-----------------------------+
```

This architecture means we can now "plug in" new command sources without changing the core UI. This is where we need your ideas!

### The Next Steps: A Call for Discussion

We want to focus our efforts on the following two steps. Our goal is to design these features in a way that makes them robust on their own, while setting the stage for them to be **chainable and composable** in the future.

#### Phase 1: Custom User-Defined Commands (The Immediate Next Step)
*(Discussion for this phase will lead to a dedicated sub-issue for the final design)*

*   **The Idea:** Allow users to define their own slash commands in a local configuration file. We believe `TOML` (.toml) is an excellent format for this, as its explicit syntax helps prevent common configuration errors. Commands could live at a project level (`<project>/.gemini/commands/`) and a user level (`~/.gemini/commands/`).
[...]

PR #4511, opened 2025-07-19
prefactor(commands): Command Service Prefactor for Extensible Commands

abhipatel12:
## TLDR

This PR introduces an extensible architecture for handling slash commands. It replaces the previous loading logic with an updated `CommandService` that operates on a provider-based `ICommandLoader` pattern.

This is prefactor step towards supporting custom, file-based commands and other future sources. There are **no user-facing changes** in this PR. It is a pure architectural refactoring designed to improve maintainability, testability, and extensibility.

**Key areas for review:**
*   The new `CommandService` and `BuiltinCommandLoader` in `packages/cli/src/services/`.
*   The addition of the `kind: CommandKind` enum to the `SlashCommand` interface in `packages/cli/src/ui/commands/types.ts`.
*   The updated instantiation logic in `packages/cli/src/ui/hooks/slashCommandProcessor.ts`, which now uses the async factory `CommandService.create()`.
*   The new unit tests that verify the service's resilience and override behavior.

## Dive Deeper

This refactor addresses the technical debt and limitations of our previous command handling system, preparing us for the strategic vision of an extensible automation platform.

### The New Command Service Architecture

The core of this change is the new service-oriented architecture:
*   **`CommandService`**: The central orchestrator for the command lifecycle. It uses a static `create()` factory method to asynchronously load commands.
*   **`ICommandLoader`**: A simple interface defining the contract for any class that can provide commands. This is the key to our future extensibility.
*   **`BuiltinCommandLoader`**: The first implementation of `ICommandLoader`. It is now solely responsible for gathering all hard-coded command definitions (including those created by factories like `ideCommand`) and providing them to the `CommandService`.

### Architectural Decision: Static Command Classification

A key decision in this implementation was how to classify commands based on their source. Instead of introducing a separate `SlashCommandDefinition` type, after reviews, I opted for a simpler and more direct approach:

**I added a `kind: CommandKind` enum directly to the `SlashCommand` interface.**

This `kind` property (`BUILT_IN` or `FILE`) provides essential static information about a command's origin without needing to inspect its runtime behavior. All existing built-in command definitions have been updated to include `kind: CommandKind.BUILT_IN`.

**Justification:**
*   **Simplicity:** This avoids the need for a transformation step within the loader, simplifying the data flow. The command objects are now defined in their final shape.
*   **Clarity:** It makes the source of any command explicit and type-safe.
*   **Future-Proofing:** This static classification is crucial for future features. It will allow us to easily build an intelligent `/help` command that can group commands by kind (e.g., "Built-in Commands", "Custom Prompts") or to create other UI experiences that treat commands differently based on their source.

## Reviewer Test Plan

The primary success criteria for this PR is that the application's command handling behavior is **identical** to `main`, proving the refactor was successful and introduced no regressions.

1.  **Code Review Focus:**
    *   Examine the new files in `packages/cli/src/services/`.
    *   Review the `CommandKind` enum and updated `SlashCommand` interface in `packages/cli/src/ui/commands/types.ts`.
    *   Confirm the new service instantiation in `packages/cli/src/ui/hooks/slashCommandProcessor.ts`.

2.  **Run Automated Tests:**
    *   Run `npm preflight`. All tests should pass, including the new unit tests for `CommandService.test.ts` and `BuiltinCommandLoader.test.ts` which validate the new architecture's core logic, resilience, and override behavior.

3.  **Manual Sanity Check:**
    *   Start the CLI with `npm start`.
    *   Execute a simple, top-level command: `/help`. The help dialog should appear as normal.
    *   Execute a nested command: `/chat list`. The list of saved chats should appear as normal.
    *   Execute a command that depends on `config` being passed in: `/ide status` (if you have IDE mode enabled). It should report the status correctly.
    *   Confirm that command autocompletion still works as expected (e.g., typing `/ch` suggests `chat`).

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  |✅  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ✅  | -   | -   |

## Linked issues / bugs

This PR makes progress on #3789

### DIFF
--- a/packages/cli/src/services/CommandService.ts
+++ b/packages/cli/src/services/CommandService.ts
@@ -75,33 +75,29 @@ import { ICommandLoader } from './types.js';
  * system to be extended with new sources without modifying the service itself.
  */
 export class CommandService {
-  // Use ReadonlyArray to enforce immutability on our internal state.
-  private commands: readonly SlashCommand[] = [];
+  /**
+   * Private constructor to enforce the use of the async factory.
+   * @param commands A readonly array of the fully loaded and de-duplicated commands.
+   */
+  private constructor(private readonly commands: readonly SlashCommand[]) {}
 
   /**
-   * Constructs the CommandService.
+   * Asynchronously creates and initializes a new CommandService instance.
+   *
+   * This factory method orchestrates the entire command loading process. It
+   * runs all provided loaders in parallel, aggregates their results, handles
+   * name conflicts by letting the last-loaded command win, and then returns a
+   * fully constructed `CommandService` instance.
    *
    * @param loaders An array of objects that conform to the `ICommandLoader`
    *   interface. The order of loaders is significant: if multiple loaders
    *   provide a command with the same name, the command from the loader that
-   *   appears later in the array will take precedence and override any earlier ones.
+   *   appears later in the array will take precedence.
+   * @returns A promise that resolves to a new, fully initialized `CommandService` instance.
    */
-  constructor(private loaders: ICommandLoader[]) {}
-
-  /**
-   * Triggers all registered command loaders to discover and load their respective
-   * commands.
-   *
-   * This method runs all loaders in parallel using `Promise.allSettled` to ensure
-   * that a failure in one loader does not prevent others from succeeding. It
-   * aggregates the results from successful loaders, resolves any name conflicts
-   * by letting the last-loaded command win, and stores the unified list internally.
-   *
-   * @returns A promise that resolves when all loading and processing is complete.
-   */
-  async loadCommands(): Promise<void> {
+  static async create(loaders: ICommandLoader[]): Promise<CommandService> {
     const results = await Promise.allSettled(
-      this.loaders.map((loader) => loader.loadCommands()),
+      loaders.map((loader) => loader.loadCommands()),
     );
 
     const allCommands: SlashCommand[] = [];
@@ -121,7 +117,8 @@ export class CommandService {
       commandMap.set(cmd.name, cmd);
     }
 
-    this.commands = Object.freeze(Array.from(commandMap.values()));
+    const finalCommands = Object.freeze(Array.from(commandMap.values()));
+    return new CommandService(finalCommands);
   }
 
   /**
--- a/packages/cli/src/services/CommandService.ts
+++ b/packages/cli/src/services/CommandService.ts
@@ -5,63 +5,7 @@
  */
 
 import { SlashCommand } from '../ui/commands/types.js';
-<<<<<<< HEAD
-import { memoryCommand } from '../ui/commands/memoryCommand.js';
-import { helpCommand } from '../ui/commands/helpCommand.js';
-import { clearCommand } from '../ui/commands/clearCommand.js';
-import { copyCommand } from '../ui/commands/copyCommand.js';
-import { corgiCommand } from '../ui/commands/corgiCommand.js';
-import { docsCommand } from '../ui/commands/docsCommand.js';
-import { mcpCommand } from '../ui/commands/mcpCommand.js';
-import { authCommand } from '../ui/commands/authCommand.js';
-import { themeCommand } from '../ui/commands/themeCommand.js';
-import { editorCommand } from '../ui/commands/editorCommand.js';
-import { chatCommand } from '../ui/commands/chatCommand.js';
-import { statsCommand } from '../ui/commands/statsCommand.js';
-import { privacyCommand } from '../ui/commands/privacyCommand.js';
-import { aboutCommand } from '../ui/commands/aboutCommand.js';
-import { extensionsCommand } from '../ui/commands/extensionsCommand.js';
-import { toolsCommand } from '../ui/commands/toolsCommand.js';
-import { compressCommand } from '../ui/commands/compressCommand.js';
-import { ideCommand } from '../ui/commands/ideCommand.js';
-import { bugCommand } from '../ui/commands/bugCommand.js';
-import { quitCommand } from '../ui/commands/quitCommand.js';
-import { restoreCommand } from '../ui/commands/restoreCommand.js';
-
-const loadBuiltInCommands = async (
-  config: Config | null,
-): Promise<SlashCommand[]> => {
-  const allCommands = [
-    aboutCommand,
-    authCommand,
-    bugCommand,
-    chatCommand,
-    clearCommand,
-    copyCommand,
-    compressCommand,
-    corgiCommand,
-    docsCommand,
-    editorCommand,
-    extensionsCommand,
-    helpCommand,
-    ideCommand(config),
-    mcpCommand,
-    memoryCommand,
-    privacyCommand,
-    quitCommand,
-    restoreCommand(config),
-    statsCommand,
-    themeCommand,
-    toolsCommand,
-  ];
-
-  return allCommands.filter(
-    (command): command is SlashCommand => command !== null,
-  );
-};
-=======
 import { ICommandLoader } from './types.js';
->>>>>>> 3d81bcd6 ((prefactor): Use loader system for slash commands)
 
 /**
  * Orchestrates the discovery and loading of all slash commands for the CLI.
--- a/packages/cli/src/services/CommandService.ts
+++ b/packages/cli/src/services/CommandService.ts
@@ -4,8 +4,8 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { Config } from '@google/gemini-cli-core';
 import { SlashCommand } from '../ui/commands/types.js';
+<<<<<<< HEAD
 import { memoryCommand } from '../ui/commands/memoryCommand.js';
 import { helpCommand } from '../ui/commands/helpCommand.js';
 import { clearCommand } from '../ui/commands/clearCommand.js';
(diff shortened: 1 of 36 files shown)


packages/cli/src/ui/commands/types.ts
```diff
@@ -106,12 +106,54 @@ export type SlashCommandActionReturn =
...
+    | SlashCommandActionReturn
+    | Promise<void | SlashCommandActionReturn>;
+
+  // Provides argument completion (e.g., completing a tag for `/chat resume <tag>`).
+  completion?: (
+    context: CommandContext,
+    partialArg: string,
+  ) => Promise<string[]>;
+
+  subCommands?: SlashCommandDefinition[];
```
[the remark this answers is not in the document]
abhipatel12: Done

packages/cli/src/services/CommandService.ts
```diff
@@ -4,79 +4,79 @@
...
+ * commands from a specific source (e.g., built-in code, local files).
+ *
+ * The CommandService is responsible for invoking these loaders, aggregating their
+ * results, and resolving any name conflicts. This architecture allows the command
+ * system to be extended with new sources without modifying the service itself.
+ */
 export class CommandService {
-  private commands: SlashCommand[] = [];
+  // Use ReadonlyArray to enforce immutability on our internal state.
+  private commands: readonly SlashCommand[] = [];
```
NTaylorMullen:
```suggestion
  private readonly commands: readonly SlashCommand[] = [];
```

packages/cli/src/services/CommandService.ts
```diff
@@ -4,79 +4,79 @@
...
-  }
+  /**
+   * Constructs the CommandService.
+   *
+   * @param loaders An array of objects that conform to the `ICommandLoader`
+   *   interface. The order of loaders is significant: if multiple loaders
+   *   provide a command with the same name, the command from the loader that
+   *   appears later in the array will take precedence and override any earlier ones.
+   */
+  constructor(private loaders: ICommandLoader[]) {}
```
NTaylorMullen: Here's another pattern you could consider:

```TypeScript
export class CommandService {
  private constructor(private readonly commands: readonly SlashCommand[]) {}

  static async create(loaders: ICommandLoader[]): Promise<CommandService> {
    const results = await Promise.allSettled(
      loaders.map((loader) => loader.loadCommands()),
    );

    const allCommands: SlashCommand[] = [];
    for (const result of results) {
      if (result.status === 'fulfilled') {
        allCommands.push(...result.value);
      } else {
        console.debug('A command loader failed:', result.reason);
      }
    }

    // De-duplicate commands using a Map. The last one found with a given name wins.
    // This creates a natural override system based on the order of the loaders
    // passed to the constructor.
    const commandMap = new Map<string, SlashCommand>();
    for (const cmd of allCommands) {
      commandMap.set(cmd.name, cmd);
    }

    const finalCommands = Object.freeze(Array.from(commandMap.values()));
    return new CommandService(finalCommands);
  }
  ...
```

packages/cli/src/services/CommandService.ts
```diff
@@ -4,79 +4,79 @@
...
+ * with an array of `ICommandLoader` instances, each responsible for fetching
+ * commands from a specific source (e.g., built-in code, local files).
+ *
+ * The CommandService is responsible for invoking these loaders, aggregating their
+ * results, and resolving any name conflicts. This architecture allows the command
+ * system to be extended with new sources without modifying the service itself.
+ */
 export class CommandService {
-  private commands: SlashCommand[] = [];
+  // Use ReadonlyArray to enforce immutability on our internal state.
```
NTaylorMullen:
```suggestion
```

packages/cli/src/services/types.ts
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * Loaders should receive any necessary dependencies (like Config) via their
+ * constructor.
+ */
+export interface ICommandLoader {
+  /**
+   * Discovers and returns a list of slash commands from the loader's source.
+   * @returns A promise that resolves to an array of SlashCommand objects.
+   */
+  loadCommands(): Promise<SlashCommand[]>;
```
NTaylorMullen: Should this be cancellable? I.e. should this take in an abort signal?

That way caller can determine timeout criteria

packages/cli/src/ui/commands/types.ts
```diff
@@ -106,12 +106,54 @@ export type SlashCommandActionReturn =
   | OpenDialogActionReturn
   | LoadHistoryActionReturn;
 
+/**
+ * Metadata describing a command's origin and expected behavior.
+ */
+export interface SlashCommandMetadata {
+  /** The source of the command definition. */
+  source: 'built-in' | 'file';
```
NTaylorMullen: Instead of source we may want to utilize a `kind` type here. Primary reason if we have more than 1 file type we support in the future this will become ambiguous

packages/cli/src/ui/commands/types.ts
```diff
@@ -106,12 +106,54 @@ export type SlashCommandActionReturn =
...
+
+  /** The primary behavior of the command. 'Custom' is for commands with unique,
+   * hard-coded logic (e.g., opening a dialog).
+   */
+  behavior: 'Custom'; // `| 'Prompt'` will be added in next PR for Custom Commands.
```
NTaylorMullen: Hmmm, what's the motivation of behavior though? Do we need to fork invocation behavior or is there a world where we can have 1 centralized invocation pattern that is a superset of them all?

packages/cli/src/ui/commands/types.ts
```diff
@@ -106,12 +106,54 @@ export type SlashCommandActionReturn =
...
+
+  /** The absolute path to the definition file, if applicable. */
+  // TODO: filePath?: string;
+}
+
+export interface SlashCommandDefinition {
+  name: string;
+  altName?: string;
```
NTaylorMullen: Eventually we'll probably want this to be an array

packages/cli/src/ui/commands/types.ts
```diff
@@ -106,12 +106,54 @@ export type SlashCommandActionReturn =
...
+  description?: string;
```
NTaylorMullen: Do we have any commands today that don't have a description? I wonder if we can enforce this to ensure the best possible user experience

packages/cli/src/ui/commands/types.ts
```diff
@@ -106,12 +106,54 @@ export type SlashCommandActionReturn =
...
+
+  // The action to run. Optional for parent commands that only group sub-commands.
```
NTaylorMullen: Food for thought (non-blocking on this PR): Right now we rely on the caller to dive down through sub-commands to find the right one to invoke. What if we relied on the parent command's action to delegate down?

packages/cli/src/ui/commands/types.ts
```diff
@@ -106,12 +106,54 @@ export type SlashCommandActionReturn =
...
+}
+
 // The standardized contract for any command in the system.
+// This version is separated from SlashCommandDefinition as it requires metadata.
 export interface SlashCommand {
   name: string;
   altName?: string;
   description?: string;
 
+  metadata: SlashCommandMetadata;
```
NTaylorMullen: I'll be curious to see how metadata plays out. I could see a world where slash commands just have a "kind" property built into them by default

NTaylorMullen (approved):
Love the refactoring. Had some suggestions to further improve it BUT feel free to follow your heart here.

Can absolutely see how this sets us up for the next wave ❤️

packages/cli/src/services/CommandService.ts
abhipatel12: Done

packages/cli/src/services/CommandService.ts
abhipatel12: Fixed via pattern you suggested in the next comment!

packages/cli/src/services/CommandService.ts
abhipatel12: Love it! Makes it much cleaner to use. Done

packages/cli/src/services/types.ts
abhipatel12: Thinking about the future, it may be helpful for things like MCP prompts. Added this in!

packages/cli/src/ui/commands/types.ts
abhipatel12: That's fair! So the way I viewed it is that the source is the general source (i.e a command loaded via a file set by a user), so this can be toml, md, etc.  So we would have a concise list of sources (built-in, file, mcp-server, etc). 

If we moved to a model that used: `built-in`, `file-toml`, `file-md`, etc, we would be mixing source along with implementation. 

If we end up supporting multiple file types in the future and need to differentiate between toml and md, we can use the `command.metadata.filepath` to see its extension.

WDYT?

packages/cli/src/ui/commands/types.ts
abhipatel12: Good question, in general, the loader would be able to correctly create the correct `action` property and the `slashCommandProcessor` and functionally, this would be the same without the `behavior` field. 

It may be overkill at this stage, but from Allen's suggestion about the `execute` ability (for non-AI) commands, it may be nice to have a on-load property that can be used to describe what a command does. This could especially be helpful if we have more descriptive UI/UX to help users know if a command is a Prompt (AI), Action (Shell/Tools/deployment?), or Custom (Built-in Commands). 

I'm leaning on keeping this, but if you feel like it's YAGNI, def okay with removing it as well! WDYT?

packages/cli/src/ui/commands/types.ts
abhipatel12: Done

packages/cli/src/ui/commands/types.ts
abhipatel12: Agreed, made this change. Will rely on the loader to have a placeholder if a user does not provide a description in the custom command use case.

packages/cli/src/ui/commands/types.ts
abhipatel12: Interesting. In this case though, wouldn't every parent command then need its own delegation logic down vs having the caller dive down? Currently we centralize this logic in one spot vs in the alternative, we'd need to replicate this logic to each individual command.

packages/cli/src/ui/commands/types.ts
abhipatel12: Originally I had considered a single top-level property but then decided to use this pattern instead. The built-in commands could define themselves, but i intended to separate the command's business logic from the command's context within the application. So a `/chat` command doesn't necessarily care that its built-in and neither should the author of the next built-in command. The loader adds the metadata for it to help describe the command to the rest of the system.

If this is YAGNI, we can replace it with a `kind` property as well!

abhipatel12:
Love all the feedback here! Made a few changes here and left a few unresolved comments

Thanks for the review!

packages/cli/src/ui/commands/types.ts
NTaylorMullen: That's a fair point; however, what's the value in tracking source? 😄

packages/cli/src/ui/commands/types.ts
NTaylorMullen: I think I'd lean on removing and adding back in when we need. It's hard to forecast future needs so I typically like to delegate any future design decisions to be grounded in a specific ask. Otherwise we could be having dead code that we need to weave among all of the provider types

packages/cli/src/ui/commands/types.ts
NTaylorMullen: > wouldn't every parent command then need its own delegation logic down vs having the caller dive down?

Yes BUT you could have a base command type that implements this across the board.

packages/cli/src/ui/commands/types.ts
NTaylorMullen: I think I'd recommend against metadata. Primary reason is that it's loose information that expects consumers to know magic strings in order to make sense of it (a bit of an anti-pattern). For similar reasons to [this](https://github.com/google-gemini/gemini-cli/pull/4511#discussion_r2217901663) if we don't need it yet I'd recommend removing

packages/cli/src/ui/commands/types.ts
abhipatel12: My first thoughts here would revolve around UI/UX. 
1. Knowing where a command came from/how it was loaded. So if a user has an error, they can tell if the custom command originated from a file loaded command, mcp, etc. (especially if the the names collide)

2. UI for autocomplete/suggestions - Currently, we just list all commands but if a user has a lot of commands from different sources, an icon/grouping can help distinguish the source/kind. We can also prioritize user commands first.

Dropping `metadata` in general, and renaming `source` to `kind` which will now become a top level required property.

packages/cli/src/ui/commands/types.ts
abhipatel12: Dropped it! We can add this back if needed in the future.

packages/cli/src/ui/commands/types.ts
abhipatel12: Oh I see, that could be interesting. I can take a look at this in a future PR if that sounds good?

packages/cli/src/ui/commands/types.ts
abhipatel12: Dropping `metadata` and we can add it back later if there's a need for it!

packages/cli/src/services/types.ts
```diff
@@ -0,0 +1,24 @@
...
+ * Loaders should receive any necessary dependencies (like Config) via their
+ * constructor.
+ */
+export interface ICommandLoader {
+  /**
+   * Discovers and returns a list of slash commands from the loader's source.
+   * @param signal An optional AbortSignal to allow cancellation.
+   * @returns A promise that resolves to an array of SlashCommand objects.
+   */
+  loadCommands(signal?: AbortSignal): Promise<SlashCommand[]>;
```
NTaylorMullen: Lets make signals required just to enforce good practices

```suggestion
  loadCommands(signal: AbortSignal): Promise<SlashCommand[]>;
```

packages/cli/src/ui/commands/types.ts
```diff
@@ -109,8 +109,10 @@ export type SlashCommandActionReturn =
 // The standardized contract for any command in the system.
 export interface SlashCommand {
   name: string;
-  altName?: string;
-  description?: string;
+  altNames?: string[];
+  description: string;
+
+  kind: 'built-in' | 'file';
```
NTaylorMullen: Lets create an enum for these kinds so they aren't magic strings. Otherwise in non-completion based IDEs or experiences it's harder to know what's available + AI systems do better with non-string arguments (otherwise they encourage hallucinations)

NTaylorMullen (approved):

packages/cli/src/services/types.ts
abhipatel12: Done

packages/cli/src/ui/commands/types.ts
abhipatel12: Ohh, that's cool, didn't know that. Done!

PR #4515, opened 2025-07-19
chore(release): v0.1.13

NTaylorMullen:
## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@google/gemini-cli",
-  "version": "0.1.12",
+  "version": "0.1.13",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@google/gemini-cli",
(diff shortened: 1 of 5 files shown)


package-lock.json
```diff
@@ -11862,7 +11862,7 @@
     },
     "packages/vscode-ide-companion": {
       "name": "gemini-cli-vscode-ide-companion",
-      "version": "0.0.1",
+      "version": "0.1.13",
```
NTaylorMullen, line 11865: cc @chrstnb @skeshive I don't think this is correct but it's in our release pipeline so I'm letting it go through for now. We may need to do a refactor here.

Issue #4385, opened 2025-07-17
Render diffs in-editor when in IDE mode


chrstnb:
### What would you like to be added?

When ide-mode=true (currently for VSCode), we should render diffs in the editor.

╭────────────────────────────────────────────────────────────────────────────╮
│  > create file haiku.txt at that path with a haiku you have written to it  │
╰────────────────────────────────────────────────────────────────────────────╯

 ╭─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
 │ ?  WriteFile Writing to haiku.txt ←                                                                                         │
 │                                                                                                                             │
 │ 1 Green squares on the screen,                                                                                              │
 │ 2 A silent language we speak,                                                                                               │
 │ 3 Building worlds with text.                                                                                                │
 │                                                                                                                             │
 │ Apply this change?                                                                                                          │
 │                                                                                                                             │
 │ ● Yes, allow once                                                                                                           │
 │ ○ Yes, allow always                                                                                                         │
 │ ○ Modify with external editor                                                                                               │
 │ ○ No (esc)                                                                                                                  │
 │                                                                                                                             │
 ╰─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯

Currently we show this.

We will need to explore the expected behavior when a user closes out of the file. Does closing mean 'allow' with any edits?

### Why is this needed?

Full-featured support of IDE mode.

PR #4519, opened 2025-07-19
[ide-mode] Add openDiff tool to IDE MCP server

chrstnb:
## TLDR

Adds openDiff tool to VSCode extension MCP server. This will open a diff, show it in the IDE, and allow the user to accept or reject it using buttons in the toolbar.

The CLI will need to receive a notification to determine when the diff has been accepted or rejected.

Note that rejecting the diff and closing the diff have the same behavior, but we can change this in the future.

cmd+s and ctrl+s will accept the diff.

## Dive Deeper

I have the CLI changes in a follow-up

Gemini CLI IDE Mode Diffing (1)
Gemini CLI IDE Mode Diffing (6)
Gemini CLI IDE Mode Diffing (5)
Gemini CLI IDE Mode Diffing (4)
Gemini CLI IDE Mode Diffing (3)

## Reviewer Test Plan

Overview using the MCP inspector.

https://github.com/user-attachments/assets/4bf5897c-1914-4370-98c6-d56495e2b5a3

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

#4385


packages/vscode-ide-companion/src/diff-content-provider.ts
```diff
@@ -0,0 +1,33 @@
...
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import * as vscode from 'vscode';
+
+export class DiffContentProvider implements vscode.TextDocumentContentProvider {
+  private content = new Map<string, string>();
+  private onDidChangeEmitter = new vscode.EventEmitter<vscode.Uri>();
```
NTaylorMullen:
```suggestion
  private readonly onDidChangeEmitter = new vscode.EventEmitter<vscode.Uri>();
```

packages/vscode-ide-companion/src/diff-content-provider.ts
```diff
@@ -0,0 +1,33 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import * as vscode from 'vscode';
+
+export class DiffContentProvider implements vscode.TextDocumentContentProvider {
+  private content = new Map<string, string>();
```
NTaylorMullen:
```suggestion
  private readonly content = new Map<string, string>();
```

packages/vscode-ide-companion/src/diff-manager.ts
```diff
@@ -0,0 +1,205 @@
...
+    } catch {
+      fileExists = false;
+    }
+
+    if (fileExists) {
+      const modifiedContent = newContent ?? '';
+      const rightDocUri = vscode.Uri.from({
+        scheme: DIFF_SCHEME,
+        path: filePath,
+        query: `rand=${Math.random()}`,
```
NTaylorMullen: should leave a comment on cache busting

packages/vscode-ide-companion/src/diff-manager.ts
```diff
@@ -0,0 +1,205 @@
...
+
+      this.addDiffDocument(rightDocUri, {
+        originalFilePath: filePath,
+        newContent: modifiedContent,
+        rightDocUri,
+      });
+
+      const diffTitle = `${path.basename(filePath)} ↔ Modified`;
+      await vscode.commands.executeCommand(
+        'setContext',
```
NTaylorMullen: Why do we need to set visible here? Doesn't executing the vscode.diff below force visibleness?

packages/vscode-ide-companion/src/diff-manager.ts
```diff
@@ -0,0 +1,205 @@
...
+      // If the file doesn't exist, we create it and show it directly
+      // instead of showing a diff.
+      const workspaceEdit = new vscode.WorkspaceEdit();
+      workspaceEdit.createFile(fileUri, { ignoreIfExists: true });
+      workspaceEdit.insert(
+        fileUri,
+        new vscode.Position(0, 0),
+        newContent ?? '',
+      );
+      await vscode.workspace.applyEdit(workspaceEdit);
```
NTaylorMullen: Wont this create the file on disk?

packages/vscode-ide-companion/src/diff-manager.ts
```diff
@@ -0,0 +1,205 @@
...
+      new vscode.Position(0, 0),
+      lastLine.range.end,
+    );
+    workspaceEdit.replace(fileUri, fullRange, modifiedContent);
+    await vscode.workspace.applyEdit(workspaceEdit);
+
+    await doc.save();
+
+    await this.closeDiffEditor(rightDocUri);
+    vscode.window.showInformationMessage('Changes applied and saved.');
```
NTaylorMullen: oh this is so interesting, so we're leaning on the IDE to apply the changes and not GC? Is this also how CC does it?

packages/vscode-ide-companion/src/diff-manager.ts
```diff
@@ -0,0 +1,205 @@
...
+  originalFilePath: string;
+  newContent: string;
+  rightDocUri: vscode.Uri;
+}
+
+/**
+ * Manages the state and lifecycle of diff views within the IDE.
+ */
+export class DiffManager {
+  private diffDocuments = new Map<string, DiffInfo>();
```
NTaylorMullen, line 52: The existence of this + the diffcontent provider make me think we should have a more central class that both the provider and this manager lean on for diff content. They're both managing diff content

packages/vscode-ide-companion/src/diff-manager.ts
```diff
@@ -0,0 +1,205 @@
...
+   * Called when the text document for a diff view is closed.
+   */
+  onDidCloseTextDocument(document: vscode.TextDocument) {
+    const closedUriString = document.uri.toString();
+    if (this.diffDocuments.has(closedUriString)) {
+      const diffInfo = this.diffDocuments.get(closedUriString);
+      if (diffInfo) {
+        this.broadcastNotification({
+          jsonrpc: '2.0',
+          method: 'ide/diffClosed',
```
NTaylorMullen: Non-blocking, food for future thought: All these methods of ide/diffClosed etc. these should probably be part of a protocol library or something that our ide integration in GC surfaces so that IDE integrators don't need to hardcode the protocol layer.

NTaylorMullen (commented):
Love seeing how this all weaves together. Left a few clarifying questions.

Very cool!

packages/vscode-ide-companion/src/diff-manager.ts
chrstnb: This results in the associated buttons showing in the toolbar, which is based on the gemini.diff.isVisible being set in the context. vscode.diff fires off the command to show the diff, but doesn't update the context state for the buttons.

packages/vscode-ide-companion/src/diff-manager.ts
chrstnb: Updated so that GC does it instead-- this gives us much more flexibility.

packages/vscode-ide-companion/src/diff-manager.ts
chrstnb: it should only create a file on disk if one doesn't already exist-- if not, it's creating a virtual document in VSCode

packages/vscode-ide-companion/src/diff-manager.ts
chrstnb: +1million!

packages/vscode-ide-companion/src/diff-manager.ts
chrstnb: DiffManager is the high-level orchestrator and manages the entire diff lifecycle. DiffContentProvider is the lower-level data source, which provides raw text for a virtual document, implementing specifically the vscode.TextDocumentContentProvider in the VSCode API. I think this separation is fairly standard!

packages/vscode-ide-companion/src/ide-server.ts
```diff
@@ -212,13 +223,61 @@ export class IDEServer {
...
       version: '1.0.0',
     },
     { capabilities: { logging: {} } },
   );
+  server.registerTool(
+    'openDiff',
+    {
+      description:
+        '(IDE Tool) Open a diff view to create or modify a file. Returns a notification once the diff has been accepted or rejcted.',
+      inputSchema: z.object({
```
skeshive, line 239: (for the follow-up CL)- can the schema be imported from the gemini-cli package instead?

packages/vscode-ide-companion/src/ide-server.ts
```diff
@@ -212,13 +223,61 @@ export class IDEServer {
...
+        filePath: z.string(),
+        newContent: z.string().optional(),
```
skeshive, line 242: can you add a TODO to follow up on whether this should be required or not?

packages/vscode-ide-companion/src/diff-manager.ts
NTaylorMullen: So if it doesn't exist it creates one (the write-file case), does that mean Gemini CLI then also tries to edit that file?

packages/vscode-ide-companion/src/diff-manager.ts
NTaylorMullen: Could be a difference of experiences here. I've typically seen content providers pull from centralized data sources. That being said it's a minor suggestion, non-blocking

NTaylorMullen (approved):

packages/vscode-ide-companion/package.json
```diff
@@ -31,7 +31,26 @@
...
+      }
+    ],
     "commands": [
+      {
+        "command": "gemini.diff.accept",
+        "title": "Accept",
+        "icon": "$(check)"
+      },
+      {
+        "command": "gemini.diff.cancel",
```
chrstnb, line 46: Expand the title to include Gemini CLI. Also remove the `Cancel` command

skeshive (approved):

packages/vscode-ide-companion/src/diff-manager.ts
chrstnb: OK I don't feel strongly! Why don't I put them in the same file for now and we can combine later.

packages/vscode-ide-companion/src/diff-manager.ts
chrstnb: ah no the behavior there was wrong previously. fixed it to behave the same as other cases but might require some tweaking

PR #4542, opened 2025-07-20
feat(cli): add warnings when gemini-cli is called in the root directory

### LINKED ISSUE #2792: FR: Add warning to users for directory structure if Gemini CLI is opened in root folder

Gemini CLI currently scans and uses the folder structure of the working directory in order to help provide context to the model.

When users open Gemini CLI in the root directory, a warning should be shown before first input that their folder structure is being used.

### Why is this needed?

It helps add transparency for how the model is gaining context and what is being sent.

### Additional context

_No response_

y-okt:
## TLDR

As described in #2792 , added a warning feature when gemini-cli is called in the root directory. 

## Dive Deeper

This is similar to `homeDirectoryCheck` in [userStartupWarnings.ts](https://github.com/google-gemini/gemini-cli/blob/main/packages/cli/src/utils/userStartupWarnings.ts#L16). Just like that method warns users because it's recommended to run gemini-cli in the project-specific directory, for the same reason, it shouldn't be run in the root directory.

## Testing Matrix

Pull this repository and in the gemini-cli's directory, run `npm install && npm run build`.
Then, go to the root directory and run ` node <PATH_TO_GEMINI_CLI_DIR>/packages/cli`

Before
スクリーンショット 2025-07-20 17 52 56

After
スクリーンショット 2025-07-20 17 58 40

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅   | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Resolves #2792

### DIFF
--- a/packages/cli/src/utils/userStartupWarnings.test.ts
+++ b/packages/cli/src/utils/userStartupWarnings.test.ts
@@ -8,6 +8,7 @@ import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
 import { getUserStartupWarnings } from './userStartupWarnings.js';
 import * as os from 'os';
 import fs from 'fs/promises';
+import path from 'path';
 
 vi.mock('os', () => ({
   default: { homedir: vi.fn() },
@@ -84,7 +85,7 @@ describe('getUserStartupWarnings', () => {
         expect.stringContaining('root directory'),
       );
       expect(warnings).toContainEqual(
-        expect.stringContaining('folder structure is being used'),
+        expect.stringContaining('folder structure will be used'),
       );
     });
 
@@ -93,13 +94,15 @@ describe('getUserStartupWarnings', () => {
         .mockResolvedValueOnce('C:\\')
         .mockResolvedValueOnce(homeDir);
 
+      vi.spyOn(path, 'dirname').mockImplementation(path.win32.dirname);
+
       const warnings = await getUserStartupWarnings('C:\\');
 
       expect(warnings).toContainEqual(
         expect.stringContaining('root directory'),
       );
       expect(warnings).toContainEqual(
-        expect.stringContaining('folder structure is being used'),
+        expect.stringContaining('folder structure will be used'),
       );
     });
 
--- a/packages/cli/src/utils/userStartupWarnings.test.ts
+++ b/packages/cli/src/utils/userStartupWarnings.test.ts
@@ -71,4 +71,58 @@ describe('getUserStartupWarnings', () => {
   //   // Tests for node version check would go here
   //   // This shows how easy it is to add new test sections
   // });
+
+  describe('root directory check', () => {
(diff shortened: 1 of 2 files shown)


packages/cli/src/utils/userStartupWarnings.test.ts
```diff
@@ -71,4 +71,58 @@ describe('getUserStartupWarnings', () => {
...
+    it('should handle errors when checking root directory', async () => {
+      vi.mocked(fs.realpath)
+        .mockRejectedValueOnce(new Error('FS error'))
+        .mockResolvedValueOnce(homeDir);
+
+      const warnings = await getUserStartupWarnings('/');
+      expect(warnings).toContainEqual(
+        expect.stringContaining('Could not verify'),
+      );
+    });
```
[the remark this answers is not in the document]
y-okt: I think the current test is to verify when `homeDirectoryCheck` fails, what output is expected.

packages/cli/src/utils/userStartupWarnings.ts
```diff
@@ -32,8 +32,37 @@ const homeDirectoryCheck: WarningCheck = {
   },
 };
 
+const rootDirectoryCheck: WarningCheck = {
+  id: 'root-directory',
+  check: async (workspaceRoot: string) => {
+    try {
+      const workspaceRealPath = await fs.realpath(workspaceRoot);
+      const errorMessage =
+        'Warning: You are running Gemini CLI in the root directory. Your entire folder structure is being used for context. It is strongly recommended to run in a project-specific directory.';
```
NTaylorMullen:
```suggestion
        'Warning: You are running Gemini CLI in the root directory. Your entire folder structure will be used for context. It is strongly recommended to run in a project-specific directory.';
```

packages/cli/src/utils/userStartupWarnings.ts
```diff
@@ -32,8 +32,37 @@ const homeDirectoryCheck: WarningCheck = {
...
+
+      // Check for Unix root directory
+      if (workspaceRealPath === '/') {
```
NTaylorMullen: I think both this and the next if block could be simplified down to:
image

Cross-check this though

NTaylorMullen (changes requested):

packages/cli/src/utils/userStartupWarnings.ts
y-okt: @NTaylorMullen Thank you for your suggestions! This seems much better. I verified on my local and tests, and these worked successfully. Could you review again?

NTaylorMullen (approved):

PR #3529, opened 2025-07-08
feat: Add /mcp refresh slash command

rmedranollamas:
Adds a new slash command `/mcp refresh` to allow users to reload MCP servers without restarting the Gemini CLI.

This change includes:
- A new `refeshMcpServers` method in the `Config` class to refresh the tool registry.
- A new `refreshMcpAction` function in `slashCommandProcessor.ts` to handle the command logic.
- Registration of the `/mcp refresh` command and updated description for the `/mcp` command.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| -------- | --- | --- | --- |
| npm run  | ✔️  | ❓  | ❓  |
| npx      | ✔️  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

- Closes #3528

PR #4566, opened 2025-07-20
feat: add /mcp refresh command

### LINKED ISSUE #3528: Allow to refresh / reload MCPs without rebooting the tool

Would be nice to have a command like /mcp reload (similar to /memory refresh) to reload the MCP tools without restarting the tool.

### Why is this needed?

Restarting the tool means losing context and progress and sometimes you need a new version of the MCP (that for example you may be developing with GCLI).

### Additional context

_No response_

rmedranollamas:
NOTE: this is a re-implementation of #3529 on the new slash command architecture. is functionally equal (it was easier than resolving that huge merge conflict).

Adds a new slash command `/mcp refresh` to allow users to reload MCP servers without restarting the Gemini CLI.

This change includes:
- A new `refeshMcpServers` method in the `Config` class to refresh the tool registry.
- A new `refreshMcpAction` function in `slashCommandProcessor.ts` to handle the command logic.
- Registration of the `/mcp refresh` command and updated description for the `/mcp` command.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✔️  | ❓  | ❓  |
| npx      | ✔️  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

- Closes #3528


packages/core/src/tools/tool-registry.ts
```diff
@@ -157,12 +157,22 @@ export class ToolRegistry {
...
 
     await this.discoverAndRegisterToolsFromCommand();
+    await this.reloadMcpTools();
+  }
+
+  async reloadMcpTools(): Promise<void> {
+    // remove any previously discovered mcp tools
+    for (const tool of this.tools.values()) {
+      if (tool instanceof DiscoveredMCPTool) {
+        this.tools.delete(tool.name);
```
NTaylorMullen, line 186: This looks more like a "clear MCP tools" method. What am I missing?

packages/cli/src/ui/hooks/slashCommandProcessor.ts
```diff
@@ -214,14 +215,15 @@ export const useSlashCommandProcessor = (
         if (foundCommand) {
           commandToExecute = foundCommand;
           pathIndex++;
-          if (foundCommand.subCommands) {
+          if (foundCommand.subCommands && depth < commandPath.length - 1) {
```
NTaylorMullen: What's this addition intending to do?

rmedranollamas: This is to correctly handle commands that have both a default action and subcommands. Without this change, when you type /mcp refresh, the command processor would incorrectly treat refresh as an argument to the default /mcp command's action, instead of executing the refresh subcommand's action.

packages/core/src/tools/tool-registry.ts
rmedranollamas: It is exactly that: clears the tools, then re-runs the discovery.

NTaylorMullen: Wait but I don't see the re-running of discovery

rmedranollamas:
ok, redid some parts, added tests and clarified the refresh logic.

NTaylorMullen (approved):

PR #4572, opened 2025-07-21
(feat): Initial Version of Custom Commands

abhipatel12:
## TLDR

This PR introduces the foundational support for user-defined **Custom Commands** in Gemini CLI. Users can now create personal and project-specific command shortcuts by defining simple `.toml` files in designated directories (`~/.gemini/commands` and `<project>/.gemini/commands`).

This change lays the architectural groundwork for a much richer command system. The core of this PR is the new `FileCommandLoader` which discovers, parses, and adapts these TOML files into executable slash commands, seamlessly integrating them into our existing `CommandService` architecture.

**Key things for reviewers to look at:**
-   `packages/cli/src/services/FileCommandLoader.ts`: The heart of the new functionality.
-   `packages/cli/src/ui/hooks/slashCommandProcessor.ts`: Note how the new loader is integrated.
-   `packages/cli/src/ui/hooks/useGeminiStream.ts`: The final connection point where a `submit_prompt` result from a custom command is handled.
-   `docs/cli/commands.md`: The user-facing documentation for this new feature.

## Dive Deeper

This PR utilizes the `ICommandLoader` to implement the `FileCommandLoader`, which extends our command system beyond the previously hard-coded, built-in commands.

**Architectural Highlights:**

1.  **Loader-Based Architecture:** The `CommandService` remains unchanged, proving it is a stable core that can be extended with new command sources without modification. The `useSlashCommandProcessor` simply adds the `FileCommandLoader` to the list of providers.

2.  **Filesystem as a Database:** We are treating the file system as the source of truth for command definitions. This has several benefits:
    *   **Discoverability:** Users can see their command structure simply by listing files.
    *   **Git-Friendliness:** Commands can be version-controlled, shared, and reviewed as part of a project's codebase.
    *   **Simplicity:** Naming is derived directly from the file path (`git/commit.toml` -> `/git:commit`), which is intuitive and predictable.

3.  **Clear Precedence:** The `CommandService`'s design, which uses a `Map` to de-duplicate commands, provides a natural and robust override mechanism. By loading `BuiltinCommandLoader` first and `FileCommandLoader` second, we guarantee that user- and project-defined commands can override built-in ones, and project commands can override user-level ones. This is explicitly tested.

4.  **Decoupled Execution:** The new `submit_prompt` action return type is a key architectural improvement. It cleanly separates the *result* of a command from the *execution* logic in the stream processor. The command's action simply declares "the user wants to submit this prompt," and the `useGeminiStream` hook is responsible for actually sending it to the model. This keeps concerns separated and will be vital as we add more complex command types in the future.

## Reviewer Test Plan

To validate this change, please pull the branch and perform the following manual tests in addition to reviewing the automated test suites.

#### 1. Test Basic User (Global) Command

1.  Create the user command directory and a test command file:
    ```bash
    mkdir -p ~/.gemini/commands
    ```
2.  Create `~/.gemini/commands/test-global.toml` with the following content:
    ```toml
    description = "A global test command."
    prompt = "This is a test prompt from the global user directory."
    ```
3.  Run `npm start` and type `/test-global`.
4.  **Expected:** The CLI should send "This is a test prompt from the global user directory." to the model.
5.  Check autocompletion by typing `/test-g` and pressing Tab. It should complete to `/test-global`.
6.  Check the help dialog with `/help`. Your new command should be listed.

#### 2. Test Project-Specific Command & Namespacing

1.  Navigate to a test project directory. Create the project-specific command directory and a namespaced command:
    ```bash
    mkdir -p .gemini/commands/git
    ```
2.  Create `.gemini/commands/git/commit.toml` with the following content:
    ```toml
    description = "Generates a Git commit message based on conventions."
    prompt = "Analyze the staged git changes and write a conventional commit message."
    ```
3.  Run `npm start` from this project directory and type `/git:commit`.
4.  **Expected:** The CLI should send the git-related prompt to the model.
5.  Check autocompletion by typing `/git:` and pressing Tab. It should complete to `/git:commit`.

#### 3. Test Project-over-User Override

1.  In the same project directory from step 2, create an override for the global command from step 1. Create `.gemini/commands/test-global.toml`:
    ```toml
    description = "A PROJECT-LEVEL override."
    prompt = "This prompt comes from the PROJECT file and should win."
    ```
2.  Run `npm start` and type `/test-global`.
3.  **Expected:** The CLI should send "This prompt comes from the PROJECT file and should win." to the model. The description in `/help` should also show "A PROJECT-LEVEL override."

#### 4. Test Malformed Files (Graceful Failure)

1.  Create an invalid TOML file:
    ```bash
    echo "this is not valid toml" > ~/.gemini/commands/invalid.toml
    ```
2.  Create a file missing the required `prompt` field:
    ```bash
    echo 'description = "I am missing a prompt"' > ~/.gemini/commands/no-prompt.toml
    ```
3.  Run `npm start`.
4.  **Expected:** The CLI should start normally without crashing. Neither `/invalid` nor `/no-prompt` should be available as commands. Check your console output for debug messages indicating why the files were skipped. Other valid custom commands should still load and work correctly.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ✅  | -   | -   |

## Linked issues / bugs

This PR continues to make progress on #3789


docs/cli/commands.md
```diff
@@ -93,6 +95,91 @@ Slash commands provide meta-level control over the CLI itself.
 - **`/quit`** (or **`/exit`**)
   - **Description:** Exit Gemini CLI.
 
+### Custom Commands
+
+Custom commands allow you to save and reuse your favorite or most frequently used prompts as personal shortcuts within the Gemini CLI. You can create commands that are specific to a single project or commands that are available globally across all your projects, streamlining your workflow and ensuring consistency.
```
NTaylorMullen:
```suggestion
Custom commands allow you to save and reuse your favorite or most frequently used prompts as personal shortcuts within Gemini CLI. You can create commands that are specific to a single project or commands that are available globally across all your projects, streamlining your workflow and ensuring consistency.
```

NTaylorMullen: Above this paragraph link to the example below as a quick start

docs/cli/commands.md
```diff
@@ -93,6 +95,91 @@ Slash commands provide meta-level control over the CLI itself.
...
+That's it! You can now run your command in the CLI. First, you might add a file to the context, and then invoke your command:
+
+```
+> @my-messy-function.js
+> /refactor:pure
+```
+
+Gemini CLI will then execute the multi-line prompt defined in your TOML file.
+
+#### What's Next?
```
NTaylorMullen: You don't need to include this bit if you don't want. This reads more as "release notes" or "code review status" than docs

packages/cli/src/services/FileCommandLoader.ts
```diff
@@ -0,0 +1,215 @@
...
+}
+
+/**
+ * A type guard to safely validate that a parsed TOML object conforms to our
+ * required command definition structure.
+ *
+ * @param obj The unknown object parsed from a TOML file.
+ * @returns True if the object is a valid TomlCommandDef, false otherwise.
+ */
+function isTomlCommandDef(obj: unknown): obj is TomlCommandDef {
```
NTaylorMullen: Oh this is interesting. SO any additional property etc. we'll need to add to this method for validation? Is there no better way?

packages/cli/src/services/FileCommandLoader.ts
```diff
@@ -0,0 +1,215 @@
...
+ * This loader is responsible for:
+ * - Recursively scanning command directories.
+ * - Parsing and validating TOML files.
+ * - Adapting valid definitions into executable SlashCommand objects.
+ * - Handling file system errors and malformed files gracefully.
+ */
+export class FileCommandLoader implements ICommandLoader {
+  private readonly projectRoot: string;
+
+  constructor(private readonly config: Config | null) {
```
NTaylorMullen, line 45: Why allow a null config?

packages/cli/src/services/FileCommandLoader.ts
```diff
@@ -0,0 +1,215 @@
...
+    this.projectRoot = config?.getProjectRoot() || process.cwd();
+  }
+
+  /**
+   * Loads all commands, applying the precedence rule where project-level
+   * commands override user-level commands with the same name.
+   * @param signal An AbortSignal to cancel the loading process.
+   * @returns A promise that resolves to an array of loaded SlashCommands.
+   */
+  async loadCommands(signal?: AbortSignal): Promise<SlashCommand[]> {
```
NTaylorMullen:
```suggestion
  async loadCommands(signal: AbortSignal): Promise<SlashCommand[]> {
```

packages/cli/src/services/FileCommandLoader.ts
```diff
@@ -0,0 +1,215 @@
...
+    const [userCommands, projectCommands] = await Promise.all([
+      this.loadCommandsFromDir(userDir, userDir, signal),
+      this.loadCommandsFromDir(projectDir, projectDir, signal),
+    ]);
+
+    const commandMap = new Map<string, SlashCommand>();
+    for (const cmd of userCommands) {
+      commandMap.set(cmd.name, cmd);
+    }
+    for (const cmd of projectCommands) {
```
NTaylorMullen: Should add a comment here around the sequencing being very intentional for overrides

packages/cli/src/services/FileCommandLoader.ts
```diff
@@ -0,0 +1,215 @@
...
+   * @param baseDir The root command directory (e.g., ~/.gemini/commands) used
+   *   to calculate the command's relative path name.
+   * @param currentDir The directory currently being scanned.
+   * @param signal An AbortSignal to cancel the loading process.
+   * @returns A promise resolving to an array of commands found in the directory.
+   */
+  private async loadCommandsFromDir(
+    baseDir: string,
+    currentDir: string,
+    signal?: AbortSignal,
```
NTaylorMullen:
```suggestion
    signal: AbortSignal,
```

packages/cli/src/services/FileCommandLoader.ts
```diff
@@ -0,0 +1,215 @@
...
+    try {
+      const entries = await fs.readdir(currentDir, { withFileTypes: true });
+      const commands: SlashCommand[] = [];
+
+      for (const entry of entries) {
+        if (signal?.aborted) break;
+
+        const fullPath = path.join(currentDir, entry.name);
+        if (entry.isDirectory()) {
+          const subCommands = await this.loadCommandsFromDir(
```
NTaylorMullen: Instead of recursively doing this we have fast-glob or the glob tool to find all these if you want 😄 . There's also the tech that the read many file tool utilizes

packages/cli/src/services/FileCommandLoader.ts
```diff
@@ -0,0 +1,215 @@
...
+        }
+      }
+      return commands;
+    } catch (error: unknown) {
+      if (isErrnoException(error) && error.code !== 'ENOENT') {
+        console.error(
+          `[FileCommandLoader] Error reading directory ${currentDir}:`,
+          error.message,
+        );
+      }
```
NTaylorMullen, line 94: With this it looks like we silence any non Errno exceptions, what's the motivation there? Ditto q elsewhere

packages/cli/src/services/FileCommandLoader.ts
```diff
@@ -0,0 +1,215 @@
...
+    if (!isTomlCommandDef(parsed)) {
+      console.error(
+        `[FileCommandLoader] Skipping invalid command file: ${filePath}. It must contain a 'prompt' string.`,
+      );
+      return null;
+    }
+
+    const relativePath = path
+      .relative(baseDir, filePath)
+      .replace(/\.toml$/, '');
```
NTaylorMullen: Doing a regex replace here feels heavy. Can't we just substring?

packages/cli/src/services/FileCommandLoader.ts
```diff
@@ -0,0 +1,215 @@
...
+    const commandName = relativePath
+      .split(path.sep)
+      // Sanitize each path segment to prevent ambiguity. Since ':' is our
+      // namespace separator, we replace any literal colons in filenames
+      // with underscores to avoid naming conflicts.
+      .map((segment) => segment.replace(/:/g, '_'))
```
NTaylorMullen: I think Node 15 introduced `.replaceAll` here. Could be wrong about one of those details though

packages/cli/src/services/FileCommandLoader.ts
```diff
@@ -0,0 +1,215 @@
...
+        `[FileCommandLoader] Skipping invalid command file: ${filePath}. It must contain a 'prompt' string.`,
+      );
+      return null;
+    }
+
+    const relativePath = path
+      .relative(baseDir, filePath)
+      .replace(/\.toml$/, '');
+    const commandName = relativePath
+      .split(path.sep)
```
NTaylorMullen, line 152: TIL path.sep. Is that both `\` and `/`?

NTaylorMullen (approved):
Very cool, played with this changeset and people are going to love it ❤️

docs/cli/commands.md
abhipatel12: Done and Done!

docs/cli/commands.md
abhipatel12: Fair, removed it!

packages/cli/src/services/FileCommandLoader.ts
abhipatel12: I used `zod` instead, and it made this a lot better. We use this other places, so we already had the dependency. 

Should be a lot better now moving forward even when we add new properties.

packages/cli/src/services/FileCommandLoader.ts
abhipatel12: This is b/c the slashCommandProcessor gets the `config` as `Config | null`. 

I have a TODO to clean this up and make this required. I was going to have a follow up where I remove this and clean up the code for it. This will have changes in the `CommandContext` as well.

If you're okay with it, I can make this a fast follow?

packages/cli/src/services/FileCommandLoader.ts
abhipatel12: Done

packages/cli/src/services/FileCommandLoader.ts
abhipatel12: Done

packages/cli/src/services/FileCommandLoader.ts
abhipatel12: Done

packages/cli/src/services/FileCommandLoader.ts
abhipatel12: Good point! Made the change and `glob` is much cleaner. Removed the `loadCommandsFromDir` helper. Thanks for the catch!

packages/cli/src/services/FileCommandLoader.ts
abhipatel12: Oh good catch this was bug. The goal was to suppress `ENOENT` errors. But With the use of `glob` we don't have this issue and will just `console.error(...)` everything.

packages/cli/src/services/FileCommandLoader.ts
abhipatel12: Yep! it essentially works as: `readonly sep: "\\" | "/";`

packages/cli/src/services/FileCommandLoader.ts
abhipatel12: Done!

packages/cli/src/services/FileCommandLoader.ts
abhipatel12: No that's right, `replaceAll` causes a build error b/c we're using `ES2020` in the `cli/tsconfig`:

```json
"lib": ["DOM", "DOM.Iterable", "ES2022"],
```

I had checked the root (`ES2023`) and then `core` (`ES2021`) and wasn't sure if this was `ES2020` on purpose?

I set it to `ES2023` to match the root config, so let me know if that sounds good!

packages/cli/src/services/FileCommandLoader.ts
```diff
@@ -0,0 +1,215 @@
...
+          }
+        }
+      }
+      return commands;
+    } catch (error: unknown) {
+      if (isErrnoException(error) && error.code !== 'ENOENT') {
+        console.error(
+          `[FileCommandLoader] Error reading directory ${currentDir}:`,
+          error.message,
+        );
```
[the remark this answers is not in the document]
abhipatel12: Leaving as is. I don't want errors to cause the entire loader to fail. These logs already appear correctly in our debug console.

abhipatel12:
Thanks for the review! Made all of the requested changes. Just had a few clarifications and we should be good to go!

packages/cli/src/services/FileCommandLoader.ts
NTaylorMullen: Yup love a fast-follow here

packages/cli/src/services/FileCommandLoader.ts
NTaylorMullen: Beautiful, that's a great choice

packages/cli/src/services/FileCommandLoader.ts
NTaylorMullen: Ya honestly I'm not entirely sure if this was intentional. @swissspidy do you have more context by chance?

Part of me wants to yolo this and see what breaks and do a fast react if needed

NTaylorMullen (approved):

PR #4604, opened 2025-07-21
Address flaky integration tests with retries

skeshive:
Update integration test script to retry failing tests up to 2 times to address flakiness. Also output logs from failing tests in non-verbose mode as well so we can pinpoint errors during the release.

### DIFF
--- a/integration-tests/run-tests.js
+++ b/integration-tests/run-tests.js
@@ -61,6 +61,7 @@ async function main() {
     console.log(`\tFound test file: ${testFileName}`);
   }
 
+  const MAX_RETRIES = 3;
   let allTestsPassed = true;
 
   for (const testFile of testFiles) {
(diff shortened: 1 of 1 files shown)


integration-tests/run-tests.js
```diff
@@ -72,63 +73,97 @@ async function main() {
...
+          }
+        });
+      });
+
+      if (exitCode === 0) {
+        testFilePassed = true;
+      } else {
+        lastStdout = stdout;
+        lastStderr = stderr;
+      }
```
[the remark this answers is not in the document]
skeshive: The flakiness is likely due to LLM output variation so a delay isnt needed

integration-tests/run-tests.js
```diff
@@ -72,63 +73,97 @@ async function main() {
...
+    }
+
+    if (!testFilePassed) {
+      console.error(
+        `Test file failed after ${MAX_RETRIES} attempts: ${testFileName}`,
+      );
+      if (!verbose) {
```
NTaylorMullen, line 163: Should this be inverted?

NTaylorMullen (approved):

integration-tests/run-tests.js
skeshive: no this was intentional- in verbose mode we log all stdout/stderr across runs in non-verbose mode we just log the last failed stdout/stderr

PR #4614, opened 2025-07-21
[ide-mode] Use active files and selected text in user prompt

chrstnb:
## TLDR

Including active files and text in the prompt to the user. We should run evals as a follow-up.

Will also include this in the UI in a follow-up!

NOTE: I left in cursor but also added selected text. I think we might as well thread both through from the IDE and see which works best in the prompt.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

#3907

### DIFF
--- a/packages/core/src/core/client.ts
+++ b/packages/core/src/core/client.ts
@@ -306,20 +306,40 @@ export class GeminiClient {
     }
 
     if (this.config.getIdeMode()) {
-      const activeFile = ideContext.getActiveFileContext();
-      if (activeFile?.filePath) {
-        let context = `
-This is the file that the user was most recently looking at:
-- Path: ${activeFile.filePath}`;
-        if (activeFile.cursor) {
-          context += `
-This is the cursor position in the file:
-- Cursor Position: Line ${activeFile.cursor.line}, Character ${activeFile.cursor.character}`;
+      const openFiles = ideContext.getOpenFilesContext();
+      if (openFiles) {
+        const contextParts: string[] = [];
+        if (openFiles.activeFile) {
+          contextParts.push(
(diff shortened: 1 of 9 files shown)


packages/vscode-ide-companion/src/ide-server.ts
```diff
@@ -224,7 +224,7 @@ const createMcpServer = () => {
     { capabilities: { logging: {} } },
   );
   server.registerTool(
```
skeshive, line 226: should we update the tool to return all the open files as well? otherwise seems fine to keep the original name so the model doesnt expect all oepn files to be returned

skeshive (approved):

packages/core/src/core/client.ts
```diff
@@ -306,20 +306,40 @@ export class GeminiClient {
...
+            .map((file) => `- ${file.filePath}`)
+            .join('\n');
+          contextParts.push(
+            `Here are files the user has recently opened, with the most recent at the top:\n${recentFiles}`,
+          );
+        }
+
+        if (contextParts.length > 0) {
+          request = [
+            { text: contextParts.join('\n') },
```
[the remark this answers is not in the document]
jacob314: this suggestion seems accurate

packages/vscode-ide-companion/src/ide-server.ts
chrstnb: I'm wondering if we should remove the tool altogether? WDYT?

jacob314 (approved):
lgtm

PR #4621, opened 2025-07-21
[ide-mode] Send the cursor and selected text from the IDE server

### LINKED ISSUE #3907: gemini-cli companion extension MCP server

chrstnb:
## TLDR

Send the cursor and selected text from the IDE MCP server

Will do evals to see how useful this is in a follow-up.

## Dive Deeper

https://code.visualstudio.com/api/references/vscode-api#TextEditor.selections

The `onDidChangeTextEditorSelection` event fires on both a change in selection and a change in cursor position!

## Reviewer Test Plan

Run the extension locally and verify the current state is being sent. e.g.,: 

Screenshot 2025-07-21 at 3 55 55 PM

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes #3907


packages/vscode-ide-companion/src/ide-server.ts
```diff
@@ -67,7 +84,26 @@ export class IDEServer {
         );
       }
     });
-    context.subscriptions.push(disposable);
+    let selectionChangeDebounceTimer: NodeJS.Timeout | undefined;
+    context.subscriptions.push(onDidChangeSubscription);
+
+    const onDidChangeTextEditorSelectionSubscription =
+      vscode.window.onDidChangeTextEditorSelection(() => {
```
NTaylorMullen: SO interesting, so there's no active way to query the active selection?

chrstnb: You can! But we want to send a notification any time this changes, unless I'm misunderstanding your question?

Basically this is not intended as a way to get the active selection (you can use `vscode.window.activeTextEditor.selection` to get it) but as a way to send a notification when the active selection changes

packages/vscode-ide-companion/src/ide-server.ts
```diff
@@ -67,7 +84,26 @@ export class IDEServer {
         );
       }
     });
-    context.subscriptions.push(disposable);
+    let selectionChangeDebounceTimer: NodeJS.Timeout | undefined;
+    context.subscriptions.push(onDidChangeSubscription);
+
+    const onDidChangeTextEditorSelectionSubscription =
```
skeshive: can you move this to the recent files manager file instead so all the logic for event listening is in one place? we can also reuse the debounce logic I just added

packages/vscode-ide-companion/src/ide-server.ts
```diff
@@ -25,14 +25,31 @@ function sendOpenFilesChangedNotification(
...
+  }
   logger.appendLine(`Sending active file changed notification: ${filePath}`);
   const notification: JSONRPCNotification = {
     jsonrpc: '2.0',
     method: 'ide/openFilesChanged',
     params: {
       activeFile: filePath,
       recentOpenFiles: recentFilesManager.recentFiles,
+      cursor: {
+        line: cursor.line,
```
skeshive: can you double check these arent off by 1? IIRC during prototyping we had to +- 1 to get the actual line/character

packages/vscode-ide-companion/src/ide-server.ts
```diff
@@ -25,14 +25,31 @@ function sendOpenFilesChangedNotification(
...
   const editor = vscode.window.activeTextEditor;
-  const filePath = editor ? editor.document.uri.fsPath : '';
+  if (!editor) {
+    return;
+  }
+
+  const filePath = editor.document.uri.fsPath;
+  const selection = editor.selection;
+  const cursor = selection.active;
+  const MAX_SELECTED_TEXT_LENGTH = 16384; // 16 KiB limit
```
skeshive: nit: move this to top of file

skeshive (approved):

packages/vscode-ide-companion/src/recent-files-manager.ts
```diff
@@ -47,11 +47,18 @@ export class RecentFilesManager {
...
+      vscode.window.onDidChangeTextEditorSelection(() => {
+        this.fireWithDebounce();
+      });
+
     context.subscriptions.push(
       editorWatcher,
       deleteWatcher,
       closeWatcher,
       renameWatcher,
+      onDidChangeTextEditorSelectionSubscription,
```
skeshive: nit: selection watcher

packages/vscode-ide-companion/src/ide-server.ts
```diff
@@ -18,17 +18,33 @@ import { RecentFilesManager } from './recent-files-manager.js';
...
+  if (!editor) {
+    return;
+  }
+
   const filePath =
     editor && editor.document.uri.scheme === 'file'
       ? editor.document.uri.fsPath
       : '';
+  const selection = editor.selection;
+  const cursor = {
```
skeshive: should we only add cursor+selection if non-empty? or does the client handle this

packages/vscode-ide-companion/src/ide-server.ts
```diff
@@ -18,17 +18,33 @@ import { RecentFilesManager } from './recent-files-manager.js';
...
 const IDE_SERVER_PORT_ENV_VAR = 'GEMINI_CLI_IDE_SERVER_PORT';
+const MAX_SELECTED_TEXT_LENGTH = 16384; // 16 KiB limit
 
 function sendOpenFilesChangedNotification(
   transport: StreamableHTTPServerTransport,
   log: (message: string) => void,
   recentFilesManager: RecentFilesManager,
 ) {
   const editor = vscode.window.activeTextEditor;
+  if (!editor) {
```
skeshive: we still need to send notifications when all files close

skeshive (approved):

packages/vscode-ide-companion/src/ide-server.ts
chrstnb: updated to send undefined if not set, but I think on the client if selectedtext is empty or undefined we should not do anything with it!

NTaylorMullen (approved):

PR #4626, opened 2025-07-21
feat(mcp): Enable MCP prompt discovery

haroldmciver-go:
## TLDR

Introduces functionality to the MCP-client to discover prompts advertised by MCP servers. Previously, the client only discovered tools. This change adds a discoverPrompts function that lists available prompts.

## Dive Deeper

The discoverPrompts function sends a prompts/list request to the MCP server. It uses a new Zod schema, PromptListResponseSchema, to validate the response, ensuring that the retrieved data conforms to the expected structure (an array of objects with name and optional description properties). The discovered prompt names are then logged to the console, providing visibility into the available prompts on a connected MCP server. This change is crucial for future features that might leverage MCP prompts directly.

## Reviewer Test Plan

Running `npm start` should log discovered prompts

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


packages/core/src/tools/mcp-client.ts
```diff
@@ -301,6 +307,44 @@ export async function discoverTools(
   }
 }
 
+const PromptListResponseSchema = z.object({
```
NTaylorMullen: Does the anthropic SDK have this schema somewhere to be re-used by chance? I would have figured they'd have all the types we need

packages/core/src/tools/mcp-client.ts
```diff
@@ -301,6 +307,44 @@ export async function discoverTools(
...
+ * @param mcpServerName The name of the MCP server.
+ * @param mcpClient The active MCP client instance.
+ */
+export async function discoverPrompts(
+  mcpServerName: string,
+  mcpClient: Client,
+): Promise<void> {
+  try {
+    const response = await mcpClient.request(
+      { method: 'prompts/list', params: {} },
```
NTaylorMullen, line 510: Should double check if the Anthropic SDK has a set of MCP methods (as static strings) for this

packages/core/src/tools/mcp-client.ts
```diff
@@ -291,6 +296,7 @@ export async function discoverTools(
           mcpServerConfig.trust,
         ),
       );
+      console.log(`Discovered tool: ${funcDecl.name}`);
```
NTaylorMullen:
```suggestion
      console.debug(`Discovered tool: ${funcDecl.name}`);
```

packages/core/src/tools/mcp-client.ts
```diff
@@ -301,6 +307,44 @@ export async function discoverTools(
...
+  try {
+    const response = await mcpClient.request(
+      { method: 'prompts/list', params: {} },
+      PromptListResponseSchema,
+    );
+
+    const promptNames = response.prompts.map((p) => p.name);
+
+    if (promptNames.length > 0) {
+      console.log(
```
NTaylorMullen:
```suggestion
      console.debug(
```

packages/core/src/tools/mcp-client.ts
```diff
@@ -301,6 +307,44 @@ export async function discoverTools(
...
+        `Discovered prompts from ${mcpServerName}: ${promptNames.join(', ')}`,
+      );
```
NTaylorMullen, line 522: I see so is next step to translate these prompts into slash commands? If so, would it be possible to do that in this PR?

I know that expands the size of this PR buttttt I think it may be required to ensure this represents a full unit of work

NTaylorMullen (commented):
It'ssss happpening!!

packages/core/src/tools/mcp-client.ts
haroldmciver-go: Yes, I just updated the mcp client to use their official list prompts schema.

PR #4631, opened 2025-07-22
feat(auth): support gemini api key and vertex auth in non-interactive mode

jerop:
Refactors the non-interactive authentication logic to support both Gemini API key and Vertex AI authentication without requiring manual configuration changes. This change moves the validation logic to a separate file and simplifies the authentication flow for headless environments.


bdmorgan (approved):
LGTM

PR #4648, opened 2025-07-22
Add yolo in settings.json + Bug fix when pasting texts as JSON where the POST was broken + Timeout limite Shell running

brunogm10:
## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


jacob314 (commented):
Please separate this pull request into separate smaller requests for each bug fix and each feature so we can review each separately. Code reviewing takes about O(n^2) the size of the pull request so this will really help you get a timely review.. Please also file an issue for each feature so we can confirm on each issue the priority of each feature. Adding yolo to settings.json may be a little controverial as we want people to have to think each time before entering yolo mode as it is unsafe.

Issue #1605, opened 2025-06-25
Add support for OpenAI models as a backend


infocusmodereal:
Integrate OpenAI's models into the Gemini CLI, allowing users to select OpenAI as an alternative model provider. This would increase the tool's flexibility and allow ranking model outputs with the same tool. This also attracts users who are already invested in the OpenAI ecosystem and have API keys. Like Kubernetes, but for models in the CLI.

#### Proposed Implementation Plan

I propose creating a provider-agnostic abstraction layer. This will prevent tight coupling to a single provider and make it easier to add other providers (e.g., Anthropic, xAI, etc) in the future.

The work can be broken down as follows:

1.  **Configuration Update:**
    *   Extend the configuration in `packages/core/src/config/config.ts` and `packages/cli/src/config/config.ts` to include:
        *   `provider`: A new field to select the model provider (e.g., `"google" | "openai"`).
        *   `openaiApiKey`: A field to store the OpenAI API key.
    *   Update the auth flow in `packages/cli/src/config/auth.ts` to support reading the OpenAI key from an environment variable (`OPENAI_API_KEY`) and storing it securely.

2.  **Provider Abstraction Layer:**
    *   In `packages/core/src/core/`, create a new `LanguageModelProvider.ts` file defining a generic `LanguageModelProvider` interface.
    *   This interface will standardize the methods for generating content, handling streaming, and managing tool calls, creating a common contract for all providers.

3.  **Refactor Existing Gemini Integration:**
    *   Create a `GeminiProvider.ts` class that implements the `LanguageModelProvider` interface.
    *   Migrate the existing Gemini-specific logic from `geminiRequest.ts` and `client.ts` into this new class.
    *   Refactor `geminiChat.ts` to depend on the `LanguageModelProvider` interface rather than the concrete Gemini implementation.

4.  **Implement OpenAI Provider:**
    *   Add the official `openai` npm package as a dependency to `packages/core`.
    *   Create a new `OpenAIProvider.ts` class that also implements the `LanguageModelProvider` interface.
    *   This class will be responsible for:
        *   Translating the application's generic request format into the format required by the OpenAI Chat Completions API.
        *   Mapping tool definitions and function-calling requests/responses between the generic format and OpenAI's format.
        *   Translating the OpenAI API response back into the application's generic response format.

5.  **Provider Factory:**
    *   Create a factory function that reads the user's configuration and instantiates the appropriate provider (`GeminiProvider` or `OpenAIProvider`).
[...]

PR #4668, opened 2025-07-22
Feature/OpenAI compatible auth

digduggog:
## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,6 +93,28 @@ The Vertex AI API provides a [free tier](https://cloud.google.com/vertex-ai/gene
 
 For other authentication methods, including Google Workspace accounts, see the [authentication](./docs/cli/authentication.md) guide.
 
+### Use an OpenAI-compatible API
+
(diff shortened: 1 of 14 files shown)


jacob314:
Heads up that there is an existing issue discussing this with links to some other pull requests sent in the past along similar lines.
https://github.com/google-gemini/gemini-cli/issues/1605

acoliver:
if you're looking for gemini-cli with mutliple provider support check https://github.com/acoliver/llxprt-code - it supports both openai and anthropic. It also supports the responses endpoint (for newer models this is substantially cheaper when using openai as a provider)  npm install -g @vybestack/llxprt-code

Or use npx
npx @vybestack/llxprt-code

Or Docker
docker run -it ghcr.io/acoliver/llxprt-code/sandbox:0.1.12

We're keeping up with Google's main generally, and the releases will be roughly compatible but with multi-llm features.

PR #4682, opened 2025-07-22
Add drawer for active files in IDE mode

chrstnb:
## TLDR

Add a separate drawer for IDE context invoked with ctrl+e

Screenshot 2025-07-24 at 12 47 45 PM
Screenshot 2025-07-24 at 12 47 53 PM

## Dive Deeper

I'm not wed to this approach, but it looks weird if we add this under the /mcp command (which is invoked via ctrl+t). This will only show up when the user is in IDE mode.

I like that it's sticky so it will follow you and show the active context while you're working, but open to alternative approaches.

## Reviewer Test Plan

https://github.com/user-attachments/assets/ca243d6d-a206-4dcf-9390-cdab8d7098d7

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

#3907


skeshive (approved):
LGTM overall , just 2 small comments

- please address /gemini review comments, they seem relevant
- in the summary, we show "6 recent files" but the detailed view shows 7 files (including the active files), can you account for this?

skeshive (approved):

packages/cli/src/ui/components/ContextSummaryDisplay.tsx
```diff
@@ -43,7 +43,7 @@ export const ContextSummaryDisplay: React.FC<ContextSummaryDisplayProps> = ({
     if (!openFiles?.activeFile) {
       return '';
     }
-    return `Open File (${path.basename(openFiles.activeFile)})`;
+    return `Open File (${path.basename(openFiles.activeFile)}) (ctrl+e for all IDE context)`;
```
skeshive: nit: Active file (to be consistent with the detailed view)

skeshive (approved):

packages/cli/src/ui/App.tsx
```diff
@@ -465,6 +468,8 @@ const App = ({ config, settings, startupWarnings = [], version }: AppProps) => {
       if (Object.keys(mcpServers || {}).length > 0) {
         handleSlashCommand(newValue ? '/mcp desc' : '/mcp nodesc');
       }
+    } else if (key.ctrl && input === 'e' && ideContext) {
+      setShowIDEContextDetail((prev) => !prev);
     } else if (key.ctrl && (input === 'c' || input === 'C')) {
       handleExit(ctrlCPressedOnce, setCtrlCPressedOnce, ctrlCTimerRef);
     } else if (key.ctrl && (input === 'd' || input === 'D')) {
```
cornmander, line 475: Can we refactor this input handler so that you can register a keystroke handler? This is starting to get unwieldly. It's ok to do this in a followup.

cornmander (approved):

chrstnb:
> LGTM overall , just 2 small comments
> 
> * please address /gemini review comments, they seem relevant
> * in the summary, we show "6 recent files" but the detailed view shows 7 files (including the active files), can you account for this?

all addressed! thanks @skeshive

packages/cli/src/ui/App.tsx
chrstnb: yep! will do right after

cornmander (approved):

packages/cli/src/ui/components/ContextSummaryDisplay.tsx
```diff
@@ -34,25 +33,26 @@ export const ContextSummaryDisplay: React.FC<ContextSummaryDisplayProps> = ({
...
 
-  const activeFileText = (() => {
-    if (!openFiles?.activeFile) {
+  const recentFilesText = (() => {
+    const count = openFiles?.recentOpenFiles?.length ?? 0;
+    if (count === 0) {
       return '';
     }
-    return `Open File (${path.basename(openFiles.activeFile)})`;
+    return `${count} recent file${count > 1 ? 's' : ''} (ctrl+e to view)`;
```
NTaylorMullen, line 46: Is it possible to list the extension name here to be `2 VSCode files`?

Also @jacob314 / @dewitt does ctrl + e mean anything shortcut wise to you that we should be concerned of?

packages/cli/src/ui/components/IDEContextDetailDisplay.tsx
```diff
@@ -0,0 +1,52 @@
...
+  return (
+    <Box
+      flexDirection="column"
+      marginTop={1}
+      borderStyle="round"
+      borderColor={Colors.AccentCyan}
+      paddingX={1}
+    >
+      <Text color={Colors.AccentCyan} bold>
+        IDE Context (ctrl+e to toggle)
```
NTaylorMullen, line 37: Similar to my other comment is it possible to show the name of the IDE so this becomes:
```suggestion
        VSCode Context (ctrl+e to toggle)
```

NTaylorMullen (approved):
Approving to unblock BUT one thing to think about is how do you plan to represent additional context such as "errors" when we start absorbing those? I imagine there will be UX implications to consider where we may go from `# VSCode files` => `# VSCode context` or something of the like

packages/cli/src/ui/components/IDEContextDetailDisplay.tsx
chrstnb: I think this is a good idea to address e.g., here https://github.com/google-gemini/gemini-cli/pull/4797

We can expose the name of the editor there (potentially as part of a protocol in the future) so it's not hard-coded here. I can make that change in that PR on top of this one?

PR #4686, opened 2025-07-22
 feat: Add auto update functionality

### LINKED ISSUE #4664: Add an "auto-update" feature

Would it be possible to add an "auto-update" feature that detects the current version at launch and automatically updates in the background? It's ok if it requires a restart of the CLI tool, but it's better than having to run the npm update command when a new update is available. Claude Code has exactly this feature, it simply auto-updates when a new version is available.

### Why is this needed?

To bring the quality and ease of use of this tool on-par with competitors.

### Additional context

_No response_

galdawave:
## TLDR

This pull request introduces an auto-update feature to the Gemini CLI. It checks for new versions and updates the cli to the newer version. This functionality can be disabled by setting the `GEMINI_CLI_DISABLE_AUTOUPDATER` environment variable.

## Dive Deeper

The core of this feature is the `checkForUpdates` function, which fetches the latest version information from npm. If a new version is available, it displays a notification to the user. The check is performed in the background to avoid blocking the user's workflow.

The following key changes were made:
- Added `update-notifier` to handle the version checking.
- Introduced a new `checkForUpdates` function in `src/utils/autoUpdate.ts`.
- Integrated the update check into the main `gemini` command.
- Added unit tests for the auto-update logic.
- Included a mechanism to disable the feature via an environment variable.

## Reviewer Test Plan

1. Pull down this branch.
2. Run `npm install` to install the new dependencies.
3. Run `npm run build` to build the project.
4. Run the CLI with `node ./packages/cli/dist/index.js` (or your usual method).
5. To test the update notification, you can temporarily modify `package.json` to an older version (e.g., "0.0.1") and run the CLI. You should see an update notification.
6. Make sure DEV environment variable is 'false'
7. To test the disable functionality, set the environment variable `export GEMINI_CLI_DISABLE_AUTOUPDATER=true` and run the CLI. You should not see an update notification and see notification telling you there's an update.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

---

Fixes #4664


packages/cli/src/ui/App.tsx
```diff
@@ -94,6 +95,76 @@ interface AppProps {
...
+  setUpdateInfo: (info: UpdateInfo | null) => void,
+  setIsUpdating: (isUpdating: boolean) => void,
+  addItem: (item: HistoryItem, timestamp: number) => void,
+) {
+  if (!info) {
+    return;
+  }
+
+  setUpdateInfo(info);
+  if (process.env.GEMINI_CLI_DISABLE_AUTOUPDATER === 'true') {
```
jacob314: can you put this in settings.ts / settings.json rather than adding as an env variable?
I could also see the case for making this an env variable but would default to just surfacing it in settings.ts as that should be sufficient and it is easier if you don't need to merge settings from multiple sources.
the case for why this needs to be settable via settings.json is that some orgs might want to disable this for everyone in their org and that is easiest to accomplish via the existing mechanism for org level settings.json files.

packages/cli/src/ui/App.tsx
```diff
@@ -103,15 +174,37 @@ export const AppWrapper = (props: AppProps) => (
...
+          addItem(
+            {
+              type: MessageType.INFO,
+              text: `${info?.message}\nRun npm install -g ${info?.update.name} to install`,
+            },
+            Date.now(),
+          );
+          setUpdateInfo(null);
+        }, 60000);
+        return;
```
jacob314: please address this lint warning.

packages/cli/src/ui/App.tsx
```diff
@@ -90,6 +91,76 @@ interface AppProps {
...
+  updateProcess.on('close', (code) => {
+    if (code === 0) {
+      addItem(
+        {
+          id: Date.now(),
+          type: MessageType.INFO,
+          text: `Update successful! The new version will be used on your next run.`,
+        },
+        Date.now(),
+      );
```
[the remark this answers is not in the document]
jacob314: please fix. I believe there is a version that does omit the ID.

packages/cli/src/ui/App.tsx
```diff
@@ -94,6 +95,76 @@ interface AppProps {
   version: string;
 }
 
+function handleAutoUpdate(
```
jacob314: can you move this out of App.tsx and into a separate file? I'd also suggest considering making this method be less coupled from react. After you sync to the latest you might want to check out how 'unhandledRejection' is now dealing with a similar case in gemini.tsx You could then start the process of checking for updates in gemini.tsx before the react app is even started using the same mechanism to send events to the app notifying it that updates have been accepted.

packages/cli/src/ui/App.tsx
```diff
@@ -94,6 +95,76 @@ interface AppProps {
...
+
+  setUpdateInfo(info);
+  if (process.env.GEMINI_CLI_DISABLE_AUTOUPDATER === 'true') {
+    setIsUpdating(false);
+    return;
+  }
+
+  setIsUpdating(true);
+
+  const updateProcess = spawn(
```
jacob314: This definitely makes sense for users that installed Gemini CLI with
`npm install -g @google/gemini-cli`
Does it make sense for users that ran
`npx https://github.com/google-gemini/gemini-cli`
or users that are running Gemini CLI from a local git repo?
If we can detect those other use cases it might make sense to skip the auto update step and instead just display the out of date.

jacob314 (commented):
Can't wait for auto update functionality!

galdawave:
sorry broke the branch when squashing some commits

packages/cli/src/ui/App.tsx
```diff
@@ -94,6 +96,75 @@ interface AppProps {
...
+    `rm -rf $(npm root -g)/@google/gemini-cli && npm install -g @google/gemini-cli@${info.update.latest}`,
```
swissspidy: This is fragile.

Many people use pnpm, yarn, bun, or even Homebrew (though that uses npm under the hood) to install Gemini CLI. Or maybe even install it as a local package, not globally. Or use `npx` to temporarily use an older version. Or use `nvm`.

Ideally, this detects which package manager is in use and the current location of the Gemini CLI executable.

The `rm -rf` is dangerous. If the `npm install` fails, you're left with nothing. It should either not do an `rm -rf` (as `npm install` would override the binary) or install to a temporary location before moving the file to the right place.

galdawave: I'll remove the `rm -rf` i knew this was dangerous but i know some people are encountering this issue https://github.com/npm/cli/issues/4096 when updating (i encounter it myself sometimes) and the only way to fix it is by removing that directory, i know some claude code users also reported this bug on their auto update feature.

packages/cli/src/ui/App.tsx
```diff
@@ -79,11 +79,13 @@ import {
   isGenericQuotaExceededError,
   UserTierId,
 } from '@google/gemini-cli-core';
-import { checkForUpdates } from './utils/updateCheck.js';
+import { checkForUpdates, type UpdateInfo } from './utils/updateCheck.js';
```
jacob314: nit: can you import `UpdateInfo` rather than `type UpdateInfo`. 
Not sure why Gemini CLI sometimes gets excited about importing types like this.

packages/cli/src/ui/App.tsx
jacob314: nit: still see reading an env variable rather than settings here.

packages/cli/src/utils/git.ts
```diff
@@ -0,0 +1,34 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import * as fs from 'fs/promises';
+import path from 'path';
+
+async function findGitRoot(startDir: string): Promise<string | null> {
```
jacob314: I would think we have something similar already in the core package. Could be worth checking to avoid creating a duplicate helper.

packages/cli/src/utils/installationInfo.ts
```diff
@@ -0,0 +1,174 @@
...
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { isGitRepository } from '@google/gemini-cli-core';
+import * as fs from 'fs';
+import * as path from 'path';
+import * as childProcess from 'child_process';
+
+export enum PackageManager {
```
jacob314, line 12: impressive list of supported package managers!

packages/cli/src/utils/installationInfo.ts
```diff
@@ -0,0 +1,174 @@
...
+    if (realPath.includes(path.join('.pnpm', '_pnpx'))) {
+      return {
+        packageManager: PackageManager.NPX,
+        isGlobal: false,
+        updateMessage: 'Running via pnpx, update not applicable.',
+      };
+    }
+    if (realPath.includes(path.join('.bun', 'install', 'cache'))) {
+      return {
+        packageManager: PackageManager.NPX,
```
jacob314: nit: should this be BUNX rather than NPX for clarity?

packages/cli/src/utils/installationInfo.ts
```diff
@@ -0,0 +1,174 @@
...
+    ) {
+      return {
+        packageManager: PackageManager.NPX,
+        isGlobal: false,
+        updateMessage: 'Running via npx, update not applicable.',
+      };
+    }
+    if (realPath.includes(path.join('.pnpm', '_pnpx'))) {
+      return {
+        packageManager: PackageManager.NPX,
```
jacob314: nit: should this be PNPM or PNPX for clarify rather than NPX?

packages/cli/src/utils/installationInfo.ts
```diff
@@ -0,0 +1,174 @@
...
+        updateCommand,
+        updateMessage: isAutoUpdateDisabled
+          ? `Please run ${updateCommand} to update`
+          : 'Installed with yarn. Attempting to automatically update now...',
+      };
+    }
+
+    // Check for bun
+    if (realPath.includes(path.join('.bun', 'bin'))) {
+      const updateCommand = 'bun add -g @google/gemini-cli@latest';
```
jacob314, line 131: wait a second bun entry? can they be merged?

packages/cli/src/gemini.tsx
```diff
@@ -325,3 +332,61 @@ async function loadNonInteractiveConfig(
     finalConfig,
   );
 }
+
+function handleAutoUpdate(
```
jacob314: nit: can you move this out of gemini into autoUpdate.ts or similar. could make testing easier and gemini.tsx is one of the tragedy of the commons files like app.tsx that it would be nice to try to keep small.

packages/cli/src/ui/App.tsx
```diff
@@ -175,45 +105,74 @@ const App = ({ config, settings, startupWarnings = [], version }: AppProps) => {
...
             },
             Date.now(),
           );
-          setUpdateInfo(null);
-        }, 60000);
-        return;
-      }
+        }
+        setUpdateInfo(null);
+      }, 60000);
```
jacob314: why do we need this 60 second timeout? Seems like we could just be adding the item right at the line where we set 
`successfullyInstalled = true`

jacob314 (commented):
lgtm
Looks good! I'll approve after the minor comments are addressed.

packages/cli/src/ui/App.tsx
galdawave: Do you mean immediately adding them item and not showing the notification? the timeout is there because this also covers the case where the user doesn't want to autoupdate or they can't so it will show the notification for a minute and then add it to history

packages/cli/src/utils/installationInfo.ts
galdawave: i moved it down to be near the other bun check, but bunx is different then bun since they install to different places. Bunx is similar to NPX where it doesn't actually install it and just runs it there.

packages/cli/src/utils/installationInfo.ts
galdawave: Thanks!

packages/cli/src/gemini.tsx
```diff
@@ -222,6 +224,10 @@ export async function main() {
       { exitOnCtrlC: false },
     );
 
+    checkForUpdates().then((info) => {
```
NTaylorMullen: You'll want to `.catch` this to ensure we don't have an unobserved exception in the worst case

packages/cli/src/ui/App.tsx
```diff
@@ -103,15 +104,76 @@ export const AppWrapper = (props: AppProps) => (
 const App = ({ config, settings, startupWarnings = [], version }: AppProps) => {
   const isFocused = useFocus();
   useBracketedPaste();
-  const [updateMessage, setUpdateMessage] = useState<string | null>(null);
+  const [updateInfo, setUpdateInfo] = useState<UpdateInfo | null>(null);
   const { stdout } = useStdout();
   const nightly = version.includes('nightly');
+  const { history, addItem, clearItems, loadHistory } = useHistory();
 
   useEffect(() => {
```
NTaylorMullen, line 119: Can we extract this bit into its own utility to keep App.tsx lean?

packages/cli/src/ui/App.tsx
```diff
@@ -760,6 +819,10 @@ const App = ({ config, settings, startupWarnings = [], version }: AppProps) => {
         {showHelp && <Help commands={slashCommands} />}
 
         <Box flexDirection="column" ref={mainControlsRef}>
+          {/* Move UpdateNotification to render update notification above input area */}
+          {updateInfo && (
+            <UpdateNotification message={`${updateInfo.message}`} />
```
```suggestion
            <UpdateNotification message={updateInfo.message} />
```

NTaylorMullen (approved):

jacob314 (approved):
lgtm
Really excited to have more users on the latest version!

PR #4689, opened 2025-07-22
add `--telemetry-outfile` flag

smhendrickson:
This change introduces a new flag, `--telemetry-outfile` which redirects all telemetry output to the specified file. No telemetry collector must be started to use this flag. This flag only works when used with `--telemetry-target local`.

Example:

```
npm run start -- --telemetry --telemetry-outfile=/tmp/foo.out
--telemetry-target local
```


packages/cli/src/config/config.ts
```diff
@@ -55,6 +55,7 @@ export interface CliArgs {
   telemetryTarget: string | undefined;
   telemetryOtlpEndpoint: string | undefined;
   telemetryLogPrompts: boolean | undefined;
+  telemetryOutfile: string | undefined;
```
jerop, line 58: also update the docs with instructions for using this: https://github.com/google-gemini/gemini-cli/blob/main/docs/telemetry.md

packages/cli/src/config/config.ts
```diff
@@ -412,6 +418,7 @@ export async function loadCliConfig(
         process.env.OTEL_EXPORTER_OTLP_ENDPOINT ??
         settings.telemetry?.otlpEndpoint,
       logPrompts: argv.telemetryLogPrompts ?? settings.telemetry?.logPrompts,
+      outfile: argv.telemetryOutfile,
```
jerop: what if it's set in settings instead or args?

jerop (commented):
🚀

packages/core/src/telemetry/file-exporters.ts
```diff
@@ -0,0 +1,89 @@
...
+  PushMetricExporter,
+  AggregationTemporality,
+} from '@opentelemetry/sdk-metrics';
+
+class FileExporter {
+  protected writeStream: fs.WriteStream;
+
+  constructor(filePath: string) {
+    this.writeStream = fs.createWriteStream(filePath, { flags: 'a' });
+  }
```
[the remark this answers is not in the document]
smhendrickson: I disagree - if a user wants telemetry and it isn't working, it should fail so the user knows instead of failing silently.

packages/cli/src/config/config.ts
smhendrickson: Added docs.

packages/cli/src/config/config.ts
smhendrickson: now supports settings

packages/cli/src/config/config.ts
```diff
@@ -159,6 +160,11 @@ export async function parseArguments(): Promise<CliArgs> {
       description:
         'Enable or disable logging of user prompts for telemetry. Overrides settings files.',
     })
+    .option('telemetry-outfile', {
+      type: 'string',
+      description:
+        'Redirect all telemetry output to the specified file. This will also redirect all other console output.',
```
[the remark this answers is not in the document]
smhendrickson: Addressed in latest commit

NTaylorMullen (approved):

PR #4737, opened 2025-07-23
feat(core, cli): Introduce high-performance FileSearch engine

brychanrobot:
## TLDR

This pull request introduces a new, high-performance `FileSearch` engine in `@google/gemini-cli-core` and integrates it into the CLI's file path autocompletion (`@` syntax). This significantly improves the speed and responsiveness of file searching, especially in large repositories. The processing of `@` commands in prompts has also been made more robust.

## Dive Deeper

The previous file completion logic was slow, relying on recursive `fs` calls or `glob` which did not scale well. This PR replaces it with a purpose-built `FileSearch` engine with the following characteristics:

-   **High-Performance Crawling**: Uses `fdir` to rapidly crawl the project directory. For performance, the initial crawl only respects directory-level ignores (e.g., `node_modules/`, `.git/`), which is where the biggest performance gains are.
-   **Powerful Matching**: Uses `picomatch` for `.gitignore`-style pattern matching and an incremental in-memory cache to make subsequent searches extremely fast.
-   **Caching**: Implements a two-tier cache:
    1.  **On-disk cache**: Persists the file list between CLI runs to speed up initial startup.
    2.  **In-memory cache**: Provides near-instantaneous results for searches within the same session by filtering down from previous, broader search results.

This new engine provides a much better user experience for file-based context fetching and lays the groundwork for potentially replacing the `glob` tool's backend in the future.

## Future work

FileSearch can be used to replace existing usages of glob, such as `GlobTool`.

## Reviewer Test Plan

To test these changes, pull this branch and run the CLI. The primary area to test is the `@` file path autocompletion and submission.

1.  **Test Autocompletion Performance**:
    -   `cd` into a very large repository (e.g., a checkout of `chromium`).
    -   Run Gemini CLI.
    -   Type `@` and start typing a file name. The suggestions should appear quickly. No need to do any manual globbing. You can just start typing any file or directory name from any level of the file structure.
    -   Try searching for a few different files to test the in-memory cache.

2.  **Test Ignore Directory Logic**:
    -   In a standard repo, type `@node_modules/` and verify that no files from `node_modules` are suggested.

3.  **Test Submission Logic**:
    -   Submit a prompt with multiple files: `What are the differences between @packages/core/index.ts and @packages/cli/index.ts?`
    -   Submit a prompt with an ignored file and text: `Why is @package-lock.json so big?` The command should proceed, but you should see a debug message indicating the file was skipped.

## Testing Matrix

|          | Linux | macOS | Windows |
| -------- | :---: | :---: | :---: |
| npm run  |  ✅   |  ❓   |  ❓   |
| npx      |  ✅   |  ❓   |  ❓   |
| Docker   |  ❓   |  ❓   |  ❓   |
| Podman   |  ❓   |   -   |   -   |
| Seatbelt |  ❓   |   -   |   -   |

### DIFF
--- a/packages/core/src/utils/filesearch/fileSearch.ts
+++ b/packages/core/src/utils/filesearch/fileSearch.ts
@@ -1,0 +1,265 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import path from 'node:path';
+import fs from 'node:fs';
+import { fdir } from 'fdir';
+import picomatch from 'picomatch';
+import { Ignore } from './ignore.js';
+import { ResultCache } from './result-cache.js';
+import * as cache from './crawlCache.js';
+
+export type FileSearchOptions = {
+  projectRoot: string;
+  ignoreDirs: string[];
+  useGitignore: boolean;
+  useGeminiignore: boolean;
+  cache: boolean;
+  cacheTtl: number;
+};
+
+export class AbortError extends Error {
+  constructor(message = 'Search aborted') {
+    super(message);
+    this.name = 'AbortError';
+  }
+}
+
+/**
+ * Filters a list of paths based on a given pattern.
+ * @param allPaths The list of all paths to filter.
+ * @param pattern The picomatch pattern to filter by.
+ * @param signal An AbortSignal to cancel the operation.
+ * @returns A promise that resolves to the filtered and sorted list of paths.
+ */
+export async function filter(
+  allPaths: string[],
+  pattern: string,
+  signal: AbortSignal | undefined,
+): Promise<string[]> {
+  const patternFilter = picomatch(pattern, {
+    dot: true,
+    contains: true,
+    nocase: true,
+  });
+
+  const results: string[] = [];
+  for (const [i, p] of allPaths.entries()) {
+    // Yield control to the event loop periodically to prevent blocking.
+    if (i % 1000 === 0) {
+      await new Promise((resolve) => {
+        setImmediate(resolve);
+      });
+      if (signal?.aborted) {
+        throw new AbortError();
+      }
+    }
+
+    if (patternFilter(p)) {
+      results.push(p);
+    }
+  }
+
+  results.sort((a, b) => {
+    const aIsDir = a.endsWith('/');
+    const bIsDir = b.endsWith('/');
+
+    if (aIsDir && !bIsDir) return -1;
+    if (!aIsDir && bIsDir) return 1;
+    return a.localeCompare(b);
+  });
+
+  return results;
+}
+
+export type SearchOptions = {
+  signal?: AbortSignal;
+  maxResults?: number;
+};
+
+/**
+ * Provides a fast and efficient way to search for files within a project,
+ * respecting .gitignore and .geminiignore rules, and utilizing caching
+ * for improved performance.
+ */
+export class FileSearch {
+  private readonly absoluteDir: string;
+  private readonly ignore: Ignore = new Ignore();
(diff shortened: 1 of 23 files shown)


packages/core/src/utils/filesearch/cache.ts
```diff
@@ -0,0 +1,52 @@
...
+  const hash = crypto.createHash('sha256');
+  hash.update(directory);
+  hash.update(ignoreContent);
+  return hash.digest('hex');
+};
+
+export const getCacheFile = (key: string): string => path.join(cacheDir, key);
+
+export const readCache = (file: string): { results: string[] } | undefined => {
+  if (fs.existsSync(file)) {
```
jacob314: nit: follow style of returning early to reduce nesting.

packages/core/src/utils/filesearch/fileSearch.ts
```diff
@@ -0,0 +1,194 @@
...
+
+    const normalizedPath = p.replace(/\\/g, '/');
+    const matches = patternFilter(normalizedPath);
+    if (matches) {
+      results.push(p);
+    }
+  }
+
+  results.sort((a, b) => {
+    const aIsDir = a.endsWith('/') || a.endsWith('\\');
```
jacob314: ask gemini cli to use a standard library for detecting directories instead of these checks.

packages/core/src/utils/filesearch/fileSearch.ts
```diff
@@ -0,0 +1,194 @@
...
+      const geminiignorePath = path.join(this.absoluteDir, '.geminiignore');
+      if (fs.existsSync(geminiignorePath)) {
+        this.ignore.add(fs.readFileSync(geminiignorePath, 'utf8'));
+      }
+    }
+
+    if (this.options.ignoreDirs.length > 0) {
+      this.ignore.add(
+        this.options.ignoreDirs.map((dir) => {
+          if (dir.endsWith('/')) {
```
jacob314, line 197: make sure we use the appropriate path separators for the current operating system.

packages/core/src/utils/filesearch/fileSearch.ts
```diff
@@ -0,0 +1,194 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
```
jacob314, line 6: this file could use more test coverage

packages/core/src/utils/filesearch/ignore.ts
```diff
@@ -0,0 +1,147 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+// Patterns with trailing spaces (even if escaped) are not supported due to picomatch trimming patterns and input.
+// This is a known limitation and rare edge case.
```
jacob314: thanks for documenting. agree this should be fine.

packages/core/src/utils/filesearch/ignore.ts
```diff
@@ -0,0 +1,147 @@
...
+      dot: true,
+      posix: true,
+      ignore: this.fileNegativePatterns,
+    });
+
+    return this;
+  }
+
+  /**
+   * @deprecated Use getDirectoryFilter or getFileFilter instead.
```
jacob314: why are we adding a deprecated method? can this be removed?

jacob314 (commented):
File search now seems like it is ignoring my git ignore directories.
Aside from that looks really promising! Performance and memory usage seem good in my limited testing so far.
Screenshot 2025-07-23 at 9 27 30 PM

Issue #4642, opened 2025-07-22
Add UI to View & Edit settings.json (Boolean Toggles, Enums, Arrays, Nested Objects)


ali-aljufairi:
### What would you like to be added?

# ✨ Add UI to View & Edit `settings.json` (Booleans, Enums, Arrays, Nested Objects)

 I’ll take full ownership of this change and all follow-up PRs.

## 🧩 Problem / Motivation

Editing `settings.json` by hand is error-prone and unfriendly. It hides powerful features (like `fileFiltering`, `mcpServers`, `checkpointing`, etc.) behind undocumented keys and types.

## 🎯 Goals

*   Provide an in-app/UI panel to **view and edit all supported settings** from `packages/cli/src/config/settings.ts`.
*   Use appropriate controls (toggles, dropdowns, number inputs, list editors) to prevent invalid values.
*   Show defaults and short descriptions.
*   Let the user choose which file to write to (system / user / workspace).

## 🚫 Non-Goals (v1)

*   Perfect editors for every deeply nested structure (JSON modal is fine initially).
*   Live hot-reload if it needs major architectural changes (can be a follow-up).

## 🗂 Settings Inventory & UI Mapping

| Key                                  | Type                       | UI Control                        | Default / Notes                                                                                                                                           | Complexity | Phase |
| ------------------------------------ | -------------------------- | --------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------- | ---------- | ----- |
| theme                                | string                     | Dropdown + “Custom…” link         | Default theme                                                                                                                                           | Low        | 1     |
| customThemes                         | Record<string, CustomTheme> | List + modal JSON/color editor    | {}                                                                                                                                                      | High       | 3     |
| selectedAuthType                     | enum AuthType              | Dropdown                          | e.g. "gemini-api-key"                                                                                                                                   | Low        | 1     |
| sandbox                              | boolean \| string          | Toggle / Select                   | false                                                                                                                                                   | Medium     | 2     |
| coreTools                            | string[]                   | Multi-select list                 | []                                                                                                                                                      | Medium     | 2     |
| excludeTools                         | string[]                   | Multi-select list                 | []                                                                                                                                                      | Medium     | 2     |
| toolDiscoveryCommand                 | string                     | Text input                        | ""                                                                                                                                                      | Low        | 2     |
| toolCallCommand                      | string                     | Text input                        | ""                                                                                                                                                      | Low        | 2     |
[...]

Issue #4683, opened 2025-07-22
Provide an easy method to install VS Code themes


miguelsolorio:
### What would you like to be added?

There are so many great themes on the [VS Code marketplace](https://marketplace.visualstudio.com/search?target=VSCode&category=Themes&sortBy=Installs) that it would be great to have an easy to install them through the UI. 

Expected behavior
1. `/install https://marketplace.visualstudio.com/items?itemName=arcticicestudio.nord-visual-studio-code`
2. Gemini downloads & extracts the theme
3. Gemini creates a configuration based on the theme and updates your settings with it

### Why is this needed?

There are a limited number of themes one can run and to setup a custom theme requires manual setup, all of which take a lot of time. We also have a lot of user requests for supporting different themes.

jacob314:
fyi @ali-aljufairi this would make the custom theme support you have added really easy for people to use.
I think it would also be worth it for the `/install` or similar command to also support installing Gemini CLI themes from a file so that users don't always have to copy and paste the JSON themselves.

ali-aljufairi:
@jacob314 
Yeah this approach makes more sense to and will make it easy to adopt themes but will require me to look more into how vscode themes are distributed because I don't know there current structure and how do they look like but I will update my issue to say that will be using this approach since it better and easy adeptness by the community and most people use Vscode fork in some kind of way so I think if this done will theme request will be mostly gone 

@miguelsolorio  That is a really smart way to solve a problem 

Action point : 

Update my issue and link this to it and once I reach phase I will see if somoene started working on it or not if someone did pick up will do it sound like intresting problem to solve also

gsquared94:
Found possible duplicate issues:

- #2122: (#0.906203350979393)

PR #4738, opened 2025-07-23
feat(ui): add /settings command and UI panel

ali-aljufairi:
## TLDR

Added command for /settings

## Dive Deeper
I just added to basic option to check first the approach and UI how it should look like and I use ThemeDialoge UI as refernce and look simalar looking for more feed back

This PR 1

I Think almost  everything done  but

Currently only Implement toggles

Not sure To do them in this PR or next one 
dropdowns, number inputs 

and not sure to put them in same line or different place not sure of this design desction

Also hot reloading is bit  to harder to implenent and I tried to do it but was complex so Decided to live it for later stages
also if you have tips on how to approach this problem please share 

# Demo
https://github.com/user-attachments/assets/808af74f-1d5e-41bb-adc0-2e77232f087d

## Reviewer 

Also asking about UI changes will possible to do it here or know

image

Also @jacob314 

Where should I put the Json Schema and where to update was not sure about this

## Testing Matrix

Update changes

|          | 🍏  | 🪟  | 🐧  |
| npm run  | x  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | x  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

#4642 

- Closes #<issue_number>
- Fixes #<issue_number>
- Resolves #<issue_number>

*Example: `Resolves #123`*

**If this PR is only related to an issue or is a partial fix, simply reference the issue number without a keyword:**

*Example: `This PR makes progress on #456` or `Related to #789`*
-->


packages/cli/src/ui/components/SettingsDialog.tsx
```diff
@@ -0,0 +1,412 @@
...
+            label: 'Show Memory Usage',
+            value: 'showMemoryUsage',
+            checked: getScopedValue('showMemoryUsage') ?? false,
+            toggle: () => {
+                updateSetting(
+                    'showMemoryUsage',
+                    !(getScopedValue('showMemoryUsage') ?? false),
+                );
+                setVersion((v) => v + 1);
+                setShowRestartPrompt(true);
```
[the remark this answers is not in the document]
ali-aljufairi: The problem it will trigger user to re authincate every time that I avoided the aproach from the first places 

maybe we separte the auth and only update auth or request when somthing happened

packages/cli/src/ui/components/SettingsDialog.tsx
```diff
@@ -0,0 +1,412 @@
...
+    function getScopedNestedValue(
+        parentKey: 'fileFiltering',
+        nestedKey: keyof FileFilteringSettings,
+    ): boolean | undefined;
+    function getScopedNestedValue(
+        parentKey: keyof Settings,
+        nestedKey: string,
+    ): boolean | undefined {
+        const parent = settings.forScope(selectedScope).settings[parentKey] as any;
+        return parent ? (parent[nestedKey] as boolean | undefined) : undefined;
```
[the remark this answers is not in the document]
ali-aljufairi: yeah Vaild and I fixed

packages/cli/src/ui/components/SettingsDialog.tsx
```diff
@@ -0,0 +1,412 @@
...
+            } else if (key.return || input === ' ') {
+                setSelectedScope(scopeItems[activeScopeIndex].value);
+                setFocusSection('settings');
+            } else if (key.tab) {
+                setFocusSection('settings');
+            }
+        }
+        if (showRestartPrompt && input === 'r') {
+            process.exit(0);
+        }
```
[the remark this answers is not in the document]
ali-aljufairi: Yeah fixed

packages/cli/src/ui/components/SettingsDialog.tsx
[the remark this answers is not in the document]
ali-aljufairi: @jacob314  This point need consider and thinking how to approach the problem or am I using gemini wrong by using sigining as auth method but I think a lot of users use this way it a bit annoying to sign every time you change a setting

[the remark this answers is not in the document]
jacob314: The settings dialog is one case where I am somewhat ok with this anti-pattern. Agree with the review bot in general but for this case I don't see a better option.

miguelsolorio:
This is a great start, thanks for working on this! I have a couple of points for feedback and some suggestions:
- Instead of using a ✅ use the `raw` values in a column next to it
- Increase the spacing between the settings to make them read a little better
- Use the same navigation verbiage from Settings
- Add "theme" to the list as most would expect to edit the theme here as well
- Move out "Apply to" into a different section (@jacob314 this is a larger item, likely needs to be done in themes too)
- Instead of mentioning that some settings require a reload, ask the user when the setting does require a reload (or do it for them). Otherwise it leaves them guessing about which settings need reload.

image

@jacob314 another item we should discuss is the naming of these settings to ensure consistency, we'll likely need to add a search input to filter the settings.

ali-aljufairi:
@miguelsolorio 

Thanks for the Feedback make a lot of sense and I will apply but here are some more question to help me be on right track 

Make a editor theme would be more complex  and  might need a lot of refinement  so I was thinking of doing in a diffrenet PR

Here was my original plan  i dont want to make Review job harder so that was break  into smaller chunks but if you want all of the UI changes in one PR I dont mind it 

For now I will start working on the obvious things that you have pointed out  and will update based on your next commint



if you have better plan please suggested 

PR 1 – Framework & Simple Scalars ( Current )

Settings panel/command scaffold.
Load & merge settings; display effective value.
Implement toggles, dropdowns, number inputs for simple keys.
Save flow (choose target file) with confirmation.

PR 2 – Arrays & Simple Nested Groups

Multi-select/list editors (coreTools, etc.).
Collapsible groups for fileFiltering, checkpointing, telemetry, bugCommand.
Per-tool number inputs for summarizeToolOutput.


PR 3 – Complex Objects
Updated

custom Themes will go with approach suggested in this issue https://github.com/google-gemini/gemini-cli/issues/4683 will see how it will go depending if
someone is working on the issue or not I will take it i no one have taken it
mcpServers table + modal with validation.

PR 4 – Polish & Extras

Tooltips/inline docs from TS/JSDoc (or a small map).
“Reset to default” per setting.
Export/import JSON.
(Optional) Live reload or “Apply & Restart”.

ali-aljufairi:
@miguelsolorio 

This Design I am think about Selecting scope not sure if this the right approach if there is other thing  I could do or try

I am not sure where do you want me to Move Apply to  please guid me or tell me if you want to remove completele

image

miguelsolorio:
@ali-aljufairi it's looking good! I think for the "Apply to" you can leave it where it was at before, we need to make wider changes that will affect Themes as well.

btw, when I played with your branch today I noticed there is a bit of a lag before the toggles change value, is this expected? if it is then we might want to show a loading indicator to let the user know it's processing otherwise it feels like it's stuck/broken.

ali-aljufairi:
@miguelsolorio  Okay I will revert it back should not taken time 

I will optimize performance now since I was not sure about about the whole now I get how the follow is and will fix and will make it faster thanks for you feedback

ali-aljufairi:
@miguelsolorio 

Return it back as it was made it faster by have local state and updating only when user restart because all change with current repo structure would require restart for settings


https://github.com/user-attachments/assets/f358056c-7489-42a1-a2ab-4bd431413bf7

packages/cli/src/ui/App.tsx
```diff
@@ -411,6 +413,8 @@ const App = ({ config, settings, startupWarnings = [], version }: AppProps) => {
     config.setFlashFallbackHandler(flashFallbackHandler);
   }, [config, addItem, userTier]);
 
+  const { isSettingsDialogOpen, openSettingsDialog, closeSettingsDialog } =
```
jacob314: please but this adjacent to similar vars for the theme dialog

packages/cli/src/ui/App.tsx
```diff
@@ -839,7 +844,15 @@ const App = ({ config, settings, startupWarnings = [], version }: AppProps) => {
             </Box>
           )}
 
-          {shellConfirmationRequest ? (
+          {isSettingsDialogOpen ? (
```
jacob314: please put this right next to similar logic for the theme dialog

packages/cli/src/ui/components/SettingsDialog.tsx
```diff
@@ -0,0 +1,455 @@
...
+interface CheckpointingSettings {
+  enabled?: boolean;
+}
+interface FileFilteringSettings {
+  respectGitIgnore?: boolean;
+  respectGeminiIgnore?: boolean;
+  enableRecursiveFileSearch?: boolean;
+}
+
+const SCOPE_LABELS = {
```
jacob314: please move this to a shared util for a settings scope dialog shared with the Themes dialog

packages/cli/src/ui/components/SettingsDialog.tsx
```diff
@@ -0,0 +1,455 @@
...
+          'autoConfigureMaxOldSpaceSize',
+          !(getScopedValue('autoConfigureMaxOldSpaceSize') ?? false),
+        );
+      },
+    },
+    {
+      label: 'Respect .gitignore',
+      value: 'fileFiltering.respectGitIgnore',
+      checked:
+        getScopedNestedValue('fileFiltering', 'respectGitIgnore') ?? false,
```
jacob314: these ignore options implicitly has a default value of true although it isn't obvious from the code... the default value isn't specified in settings where it would make sense but later when this is used.

I'd suggest we define these default values and nice human readable names in a json style object in settings.ts. That will set us up to export a json schema from them as well in the future.

packages/cli/src/ui/components/SettingsDialog.tsx
```diff
@@ -0,0 +1,455 @@
...
+        updateSetting(
+          'disableAutoUpdate',
+          !(getScopedValue('disableAutoUpdate') ?? false),
+        );
+      },
+    },
+  ];
+
+  // Scope selector items
+  const scopeItems = [
```
jacob314: can you refactor to share code with the identical implementation in the themes dialog.

packages/cli/src/ui/components/SettingsDialog.tsx
```diff
@@ -0,0 +1,455 @@
...
+        })}
+        {showScrollDown && <Text color={Colors.Gray}>▼</Text>}
+
+        <Box height={1} />
+
+        {/* Apply To Section - Moved to bottom */}
+        <Text color={Colors.AccentBlue}>Apply To</Text>
+        <Box height={1} />
+
+        {/* Display scope items vertically */}
```
jacob314: please align this one with how this UI is rendered in settings. The behavior is currently different  with theme surfacing the selection change immediately while this is a bit different. my preference is to align the two probably to what is in theme and we can later refine this. As Miguel mentioned, our UI for this isn't ideal right now but lets leave polishing that for a leter pr.

packages/cli/src/ui/components/SettingsDialog.tsx
```diff
@@ -0,0 +1,455 @@
...
+  // Local pending settings state for the selected scope
+  const [pendingSettings, setPendingSettings] = useState<Settings>(() =>
+    // Deep clone to avoid mutation
+    JSON.parse(JSON.stringify(settings.forScope(selectedScope).settings)),
+  );
+
+  // Reset pending settings when scope changes
+  useEffect(() => {
+    setPendingSettings(
+      JSON.parse(JSON.stringify(settings.forScope(selectedScope).settings)),
```
jacob314: nit: use structuredClone instead

packages/cli/src/ui/components/SettingsDialog.tsx
```diff
@@ -0,0 +1,455 @@
...
+  const [activeSettingIndex, setActiveSettingIndex] = useState(0);
+  const [activeScopeIndex, setActiveScopeIndex] = useState(0);
+  // Scroll offset for settings
+  const [scrollOffset, setScrollOffset] = useState(0);
+  const [showRestartPrompt, setShowRestartPrompt] = useState(false);
+
+  // Local pending settings state for the selected scope
+  const [pendingSettings, setPendingSettings] = useState<Settings>(() =>
+    // Deep clone to avoid mutation
+    JSON.parse(JSON.stringify(settings.forScope(selectedScope).settings)),
```
jacob314: use structuredClone here instead

packages/cli/src/ui/components/SettingsDialog.tsx
```diff
@@ -0,0 +1,455 @@
...
+  };
+  const updateSetting = <K extends keyof Settings>(
+    key: K,
+    value: Settings[K],
+  ) => {
+    setPendingSettings((prev) => ({ ...prev, [key]: value }));
+    setShowRestartPrompt(true);
+  };
+
+  // List of boolean settings (with default false if undefined)
```
jacob314: not all these should have a default of undefined to avoid confusion. otherwise users will think we ignore gitignore by default

jacob314 (commented):
Done with a round of review comments. This is looking really good overall! This makes the fact that we have a lot of settings so much less confusing for users!

packages/cli/src/ui/App.tsx
ali-aljufairi: Done

packages/cli/src/ui/App.tsx
ali-aljufairi: Done

packages/cli/src/ui/components/SettingsDialog.tsx
ali-aljufairi: done

packages/cli/src/ui/components/SettingsDialog.tsx
ali-aljufairi: Done

packages/cli/src/ui/components/SettingsDialog.tsx
ali-aljufairi: Done

packages/cli/src/ui/components/SettingsDialog.tsx
ali-aljufairi: "undefined" is shown currently based on this
1. Setting does NOT exist in the original JSON file for the current scope
2. User has NOT modified this setting in the current session

packages/cli/src/ui/components/SettingsDialog.tsx
ali-aljufairi: Thanks I was not aware of such method exist

jacob314:
Thanks for your contribution, @ali-aljufairi! This is a great start on a much-needed settings UI. The overall approach of adding a new command and a dedicated dialog is solid.

I have a few suggestions to improve the code quality, maintainability, and user experience.

### 1. Conventional Commit Title

The PR title `Feat/settings UI panel` doesn't follow our project's [Conventional Commits](https://www.conventionalcommits.org/) standard. A better title would be something like:

`feat(ui): add /settings command and UI panel`

This helps with automated changelog generation and makes the commit history easier to read.

### 2. Schema-Driven Settings

In `SettingsDialog.tsx`, the list of settings is hardcoded as a large array of objects. This has a few downsides:
- It's repetitive and hard to maintain.
- It mixes the UI presentation with the setting's definition.
- It makes it difficult to add new types of settings in the future (like dropdowns or text inputs).

You asked where to put a JSON schema, and that's exactly the right direction! We should define the settings in a central schema. This schema could then be used to dynamically generate the form in the UI. This would make the `SettingsDialog` component much simpler and more scalable.

For example, a schema could look something like this:

```json
{
  "showMemoryUsage": {
    "type": "boolean",
    "label": "Show Memory Usage",
    "category": "General",
    "requiresRestart": false
  },
  "checkpointing.enabled": {
    "type": "boolean",
    "label": "Enable Checkpointing",
    "category": "Checkpointing",
    "requiresRestart": true
  }
}
```

### 3. Simplify Saving Logic

The logic for saving the settings inside the `useInput` hook is quite complex. It handles checking for modified settings, comparing against default values, and then saving. This logic could be extracted into a separate function to make the `useInput` hook cleaner and easier to read. It would also make the saving logic itself easier to test in isolation.

### 4. User Experience for `undefined` Values

The UI currently displays the string `"undefined"` when a setting isn't explicitly defined in the current scope's configuration file. While technically accurate, this could be confusing for users.

Consider displaying the *effective* value (i.e., the value inherited from a higher-priority scope or the default value) and using a visual cue (like a different color or an icon) to indicate that the value is inherited and not set at the current scope.

### 5. Avoid Forcing a Restart

The current implementation forces the user to restart the CLI for *any* setting change. This is a disruptive user experience, especially for settings that could be applied instantly (like `hideTips` or `vimMode`).

The settings schema mentioned above could include a `requiresRestart` property. The UI could then track if any modified settings require a restart and only show the "Restart required" prompt when necessary.

### 6. Add Component Tests

The `SettingsDialog` component contains a lot of complex logic for state management, input handling, and rendering. It would be great to add tests for this component using `ink-testing-library` to ensure it behaves as expected and to prevent regressions in the future. You can look at other component tests in the codebase for examples.

### Positive Note

Great job on refactoring the scope selection logic into `dialogScopeUtils.ts`! That's a nice improvement that reduces code duplication and makes the code cleaner.

Looking forward to the next iteration!

packages/cli/src/config/settingsSchema.ts
```diff
@@ -0,0 +1,159 @@
...
+    parentKey: 'fileFiltering',
+    childKey: 'respectGitIgnore',
+  },
+
+  'fileFiltering.respectGeminiIgnore': {
+    type: 'boolean',
+    label: 'Respect .geminiignore',
+    category: 'File Filtering',
+    requiresRestart: true,
+    default: false,
```
jacob314, line 266: this default should be true.

packages/cli/src/config/settingsSchema.ts
```diff
@@ -0,0 +1,159 @@
...
+    parentKey: 'checkpointing',
+    childKey: 'enabled',
+  },
+
+  'fileFiltering.respectGitIgnore': {
+    type: 'boolean',
+    label: 'Respect .gitignore',
+    category: 'File Filtering',
+    requiresRestart: true,
+    default: false,
```
jacob314: this default should be true

packages/cli/src/utils/settingsUtils.test.ts
```diff
@@ -0,0 +1,636 @@
...
+  hasRestartRequiredSettings,
+  getRestartRequiredFromModified,
+  getDisplayValue,
+  isDefaultValue,
+  isValueInherited,
+  getEffectiveDisplayValue,
+} from './settingsUtils.js';
+
+describe('SettingsUtils', () => {
+  // ============================================================================
```
jacob314: nit: make the describe block be 
schema utilities
rather than just having it as a comment.

packages/cli/src/utils/settingsUtils.test.ts
```diff
@@ -0,0 +1,636 @@
...
+        'Accessibility',
+      );
+    });
+
+    it('should return undefined for invalid settings', () => {
+      expect(getSettingCategory('invalidSetting')).toBeUndefined();
+    });
+  });
+
+  // ============================================================================
```
jacob314: same comment. make these comments be describe blocks so it is obvious what block of tests the failures are from

packages/cli/src/config/settingsSchema.ts
```diff
@@ -0,0 +1,159 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export interface SettingDefinition {
```
jacob314, line 16: this schema is good but can be made better by tweaking so that the TypeScript type system ensures it is consistent with settings.ts. I've pushed a branch that illustrates how this could work while not changing any behavior. I got this implementation from your  code by entering the prompt:
"how can I refactor @packages/cli/src/config/settingsSchema.ts and @settings.ts to get more static type checking to make sure the schema is consistent with settings.ts? Give me 3 options."
I then selected the option that seemed the most reasonable.

https://github.com/google-gemini/gemini-cli/commit/719551f5f93bee00e240f91e91d99884927389e1
If it looks reasonable, you can add it to your branch by

```
git fetch origin
git cherry-pick 719551f5f93bee00e240f91e91d99884927389e1
```
or similar. 

Summary of the solution gemini CLI suggested:
 Schema as the Single Source of Truth

  This approach uses a nested SETTINGS_SCHEMA object as the definitive source for all settings, from which the Settings TypeScript type is automatically inferred.

  Pros:

   * Guaranteed Consistency: This is the biggest advantage. It is programmatically impossible for the Settings type and the schema to become out of sync. The type is a direct
     reflection of the schema.
   * Single Point of Maintenance: When adding, removing, or modifying a setting, you only need to make changes in one place: the SETTINGS_SCHEMA object. This reduces developer
     effort and eliminates the risk of forgetting to update a corresponding type.
   * Centralized Definition: All information related to a setting—its key, type, UI label, default value, and description—is co-located in a single, easy-to-find definition.

  Cons:

   * Requires Structural Refactoring: The current SETTINGS_SCHEMA uses a flat structure with dot-notation keys (e.g., 'fileFiltering.respectGitIgnore'). This option requires a
     one-time effort to convert it into a nested object structure, which can be tedious for a large schema.
   * Complex Type Inference: The utility type (InferSettings<T>) needed to generate the Settings type from the schema uses advanced TypeScript features. Developers who are not
     deeply familiar with mapped and conditional types might find this logic difficult to understand or debug if issues arise.
   * Less Direct Type Readability: A developer cannot simply read a clean interface to understand the shape of the Settings object. They must rely on their IDE's ability to
     expand the inferred type on hover, which can be less convenient than a straightforward declaration.

  Best for: Teams that prioritize long-term maintainability and strict correctness. This is the most robust solution for preventing configuration drift in a large or growing
  project.

jacob314 (commented):
Overall looks great and this is very close to approval. However some of the settings now do not seem to change when I press space. For example, select "Respect .geminiignore and press space` The value seems to flicker but goes back to false. A few options where the setting can be applied without restart such as vim mode and show memory usage work well but the others don't.

packages/cli/src/config/settingsSchema.ts
ali-aljufairi: Done

packages/cli/src/utils/settingsUtils.test.ts
ali-aljufairi: done

packages/cli/src/config/settingsSchema.ts
ali-aljufairi: The approach is resonable and I mostly followed same impelation but I did new json key which kinda answer why I did not want to do this approach from the start 

```js
  showInDialog?: boolean;

```

Because  I dont think all settings should be editable threw Dialogue specially MCP related setting I would need time to learn how to implement it

jacob314: showInDialog is great. thanks for adding that.

packages/cli/src/config/settingsSchema.ts
```diff
@@ -37,123 +102,369 @@ export const SETTINGS_SCHEMA: SettingsSchema = {
...
   },
-
   autoConfigureMaxOldSpaceSize: {
     type: 'boolean',
     label: 'Auto Configure Max Old Space Size',
     category: 'General',
     requiresRestart: true,
     default: false,
     description: 'Automatically configure Node.js memory limits',
+    showInDialog: false,
```
jacob314: nit: include in dialog

packages/cli/src/ui/components/SettingsDialog.tsx
```diff
@@ -70,6 +73,11 @@ export function SettingsDialog({
     new Set(),
   );
 
+  // Track the intended values for modified settings
+  const [modifiedValues, setModifiedValues] = useState<Map<string, boolean>>(
```
jacob314, line 77: nit: see if Gemini CLI can easily make this typed. If not, fine to leave. Potentially this should just be
useState<Settings>
and then use the appropriate APIs to enumerate the properties.

packages/cli/src/utils/settingsUtils.ts
```diff
@@ -32,44 +61,54 @@ export function getSettingsByCategory(): Record<string, SettingDefinition[]> {
...
+ * Recursively gets a value from a nested object using a key path array.
  */
-export function parseSettingKey(key: string): [string, string | undefined] {
-  const parts = key.split('.');
-  if (parts.length === 2) {
-    return [parts[0], parts[1]];
+function getNestedValue(
+  obj: Record<string, unknown>,
+  path: string[],
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
jacob314: avoid any. Make the return value Object if you need to even if it makes some call sites more ugly.

jacob314 (commented):
This is looking quite good. Expect to be able to approve after some manual testing in the morning.

packages/cli/src/config/settingsSchema.ts
ali-aljufairi: fixed

packages/cli/src/utils/settingsUtils.ts
ali-aljufairi: This is mistake from my end I should have fixed earlier I was trying  to fix flicker issue and had type issue so focused on fixing the flicker and then forgot to remove this sorry for the trouble

packages/cli/src/ui/components/SettingsDialog.tsx
ali-aljufairi: Hmm tried bunch of things did not work out but will keep this comment on my mind

miguelsolorio:
@ali-aljufairi I noticed when going up/down the list and you are at the top/bottom, the cursor isn't sent to the opposite side of the list like theme/tool calls. do you mind adding this support?

ali-aljufairi:
> @ali-aljufairi I noticed when going up/down the list and you are at the top/bottom, the cursor isn't sent to the opposite side of the list like theme/tool calls. do you mind adding this support?

Oh I did not notice that thanks will work on it now

ali-aljufairi:
@miguelsolorio done This exact that I have theme experience in local branch and latest gemini-cli latest version if you need other thing please inform and please 


https://github.com/user-attachments/assets/d644c29c-beac-47e2-ace3-e10fe8707084




when you have time inform about the choices and number should be as experience  and also theme editing setting.json but for MCP support I think it s needed since @NTaylorMullen  is working on gemini mcp add I dont think user would ever need to edit in the UI


Choicses NOT this PR because will make pushing this change a bit slower and we The and everyone to follow new approach for setting


https://github.com/user-attachments/assets/5924f55c-5946-45c6-906e-b491469316ba

jacob314 (approved):
Thanks for all your hard work on this! Users are going to really appreciate this feature.
Please resolve the merge conflicts then I will merge.lgtm

ali-aljufairi:
@jacob314 done please merge

jacob314 (approved):
lgtm

PR #4740, opened 2025-07-23
[Refactor] Centralizes autocompletion logic within useCompletion

SandyTao520:
## TLDR

Refactor the autocompletion logic by moving the isActive check and handleAutocomplete function from InputPrompt.tsx to the useCompletion.ts hook.

## Dive Deeper

The primary motivation for this refactor was to improve the separation of concerns. The handleAutocomplete function is intrinsically tied to the state managed by the useCompletion hook (e.g., suggestions, slashCommands). By moving this function into the hook, we co-locate the logic with the data it operates on.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ✅  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

N/A


packages/cli/src/ui/hooks/useCompletion.integration.test.ts
```diff
@@ -26,6 +27,42 @@ interface MockConfig {
   getFileService: () => FileDiscoveryService | null;
 }
 
+// Helper to create mock TextBuffer objects
+const createMockBuffer = (text: string): TextBuffer => {
```
jacob314: integration tests shouldn't be mocking things that don't need to be mocked. can you get gemini CLI to just use the real TextBuffer here.

packages/cli/src/ui/hooks/useCompletion.test.ts
```diff
@@ -18,6 +18,43 @@ import {
   SlashCommand,
 } from '../commands/types.js';
 import { Config, FileDiscoveryService } from '@google/gemini-cli-core';
+import { TextBuffer } from '../components/shared/text-buffer.js';
+
+// Helper to create mock TextBuffer objects.
+const createMockBuffer = (text: string): TextBuffer => {
```
jacob314: can you use the real textBuffer rather than mocking it?

jacob314 (approved):
lgtm
For follow up, the tests do more mocking than should be needed. Fine to fix that up in a follow up pull request.

PR #4741, opened 2025-07-23
Propagate user_prompt_id to GenerateConentRequest for logging

owenofbrien:
## TLDR

This PR populates the new user_prompt_id field on GenerateContent and StreamingGenerateContent requests.

## Dive Deeper

The user_prompt_id is needed for logging.

## Reviewer Test Plan

Test against CCPA autopush or staging (see https://g3doc.corp.google.com/company/teams/cloud-code/eng/private-api/development.md?cl=head#running-ides-against-autopushstaging). This PR should not be merged until cl/785782609 lands in CCPA prod.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | y  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

https://github.com/google-gemini/maintainers-gemini-cli/issues/28
-->


packages/core/src/code_assist/converter.ts
```diff
@@ -32,6 +32,7 @@ import {
 export interface CAGenerateContentRequest {
   model: string;
   project?: string;
+  user_prompt_id?: string;
```
NTaylorMullen, line 35: can this be camelCase instead of snake case (typescript default)?

Same elsewhere.

Also, don't we surface this as `promptId` elsewhere in the codebase?

packages/core/src/code_assist/server.ts
```diff
@@ -53,10 +53,16 @@ export class CodeAssistServer implements ContentGenerator {
 
   async generateContentStream(
     req: GenerateContentParameters,
+    userPromptId?: string,
```
NTaylorMullen: Possible to not be optional?

packages/core/src/core/contentGenerator.ts
```diff
@@ -25,10 +25,12 @@ import { UserTierId } from '../code_assist/types.js';
 export interface ContentGenerator {
   generateContent(
     request: GenerateContentParameters,
+    userPromptId?: string,
```
NTaylorMullen: ditto q for anywhere where this is optional

packages/core/src/core/geminiChat.ts
```diff
@@ -294,11 +294,14 @@ export class GeminiChat {
...
-          contents: requestContents,
-          config: { ...this.generationConfig, ...params.config },
-        });
+        return this.contentGenerator.generateContent(
+          {
+            model: modelToUse,
+            contents: requestContents,
+            config: { ...this.generationConfig, ...params.config },
+          },
+          prompt_id,
```
NTaylorMullen, line 303: and ditto about anywhere that's using snake case vs. camel

packages/core/src/code_assist/converter.ts
owenofbrien: Not unless there's some kind of auto-converter we can employ to translate to the corresponding proto field casing. I tried camelCase here but got a server error when testing it with the CCPA autopush endpoint. The request proto field uses snake_case: https://screenshot.googleplex.com/AHZsBYbfncBocaE

packages/core/src/core/contentGenerator.ts
owenofbrien: Done

packages/core/src/code_assist/server.ts
owenofbrien: Done

packages/core/src/core/geminiChat.ts
owenofbrien: There are 128 instances of snake_cased 'prompt_id' across 20 files, so I'd rather not pollute this PR with all those changes, some of which could be risky if not handled carefully.

NTaylorMullen (approved):

PR #4748, opened 2025-07-23
Add Google credentials provider for authenticating with MCP servers

### LINKED ISSUE #3704: Add Google OAuth support to MCP client

ashwinpvg:
This introduces a new `authProviderType` setting for MCP servers in `settings.json`.
When set to `google_credentials`, the CLI will use the `google-auth-library` to automatically obtain credentials from the environment.

To improve security, the `oauth.scopes` property is now mandatory when using the `google_credentials` provider, ensuring that users explicitly grant the necessary permissions.

Includes:
- `GoogleCredentialProvider` implementation
- Integration with MCP transport layer
- Unit tests for the new provider

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

- Resolves #3704


packages/core/src/config/config.ts
```diff
@@ -118,9 +118,15 @@ export class MCPServerConfig {
     readonly extensionName?: string,
     // OAuth configuration
     readonly oauth?: MCPOAuthConfig,
+    readonly authProviderType?: AuthProviderType,
   ) {}
 }
 
+export enum AuthProviderType {
+  DYNAMIC_DISCOVERY = 'dynamic_discovery',
```
sethtroisi, line 126: Doesn't seem used can this be removed?

docs/tools/mcp-server.md
```diff
@@ -179,6 +179,28 @@ OAuth tokens are automatically:
 - **Validated** before each connection attempt
 - **Cleaned up** when invalid or expired
 
+#### Authentication Provider Type
+
+You can specify the authentication provider type using the `authProviderType` property:
+
+- **`authProviderType`** (string): Specifies the authentication provider. Can be one of the following:
+  - **`dynamic_discovery`** (default): The CLI will automatically discover the OAuth configuration from the server.
```
sethtroisi, line 187: is there a need to be explicit or can authProviderType simply be omitted to get the same result?

cornmander: dynamic_discovery is default so omitting authProviderType will do the same.

cornmander (approved):

packages/core/src/config/config.ts
cornmander: It's better to be comprehensive with the enum even if DYNAMIC_DISCOVERY is the default.

NTaylorMullen (approved):

PR #4765, opened 2025-07-24
Log prompt id when a loop is detected

SandyTao520:
## TLDR

Include this for easier querying.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ✅  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

N/A

### DIFF
--- a/packages/core/src/services/loopDetectionService.ts
+++ b/packages/core/src/services/loopDetectionService.ts
@@ -50,7 +50,7 @@ const SENTENCE_ENDING_PUNCTUATION_REGEX = /[.!?]+(?=\s|$)/;
  */
 export class LoopDetectionService {
   private readonly config: Config;
-  private prompt_id = '';
+  private promptId = '';
 
   // Tool call tracking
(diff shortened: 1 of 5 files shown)


packages/core/src/services/loopDetectionService.ts
```diff
@@ -50,6 +50,7 @@ const SENTENCE_ENDING_PUNCTUATION_REGEX = /[.!?]+(?=\s|$)/;
  */
 export class LoopDetectionService {
   private readonly config: Config;
+  private prompt_id = '';
```
NTaylorMullen: CamelCase promptId instead of snake case

```suggestion
  private promptId = '';
```

Ditto elsewhere

NTaylorMullen (approved):

PR #4790, opened 2025-07-24
[Refactor] Add keystroke handler to App

chrstnb:
## TLDR

Clean up our handling of keystrokes. Also added tests for a couple

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

#4867


packages/cli/src/ui/App.tsx
```diff
@@ -447,6 +454,76 @@ const App = ({ config, settings, startupWarnings = [], version }: AppProps) => {
...
+      handleExit,
+      ctrlCPressedOnce,
+      setCtrlCPressedOnce,
+      ctrlCTimerRef,
+      ctrlDPressedOnce,
+      setCtrlDPressedOnce,
+      ctrlDTimerRef,
+      buffer.text.length,
+      setConstrainHeight,
+    ],
```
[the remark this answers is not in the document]
jacob314: please fix. for a range of reasons agree we shouldn't have this depend on buffer.length unless we absolutely have to.

packages/cli/src/ui/App.tsx
```diff
@@ -88,6 +88,13 @@ import { PrivacyNotice } from './privacy/PrivacyNotice.js';
 
 const CTRL_EXIT_PROMPT_DURATION_MS = 1000;
 
+interface KeystrokeHandler {
```
jacob314: this is worth moving to its own file rather than keeping in app.tsx
I'd like to reuse in InputPrompt and text-buffer.ts as well.
In addition, I'm hoping this can also be the basis to in the future configure keyboard shortcuts via a file and surface the up to date keyboard shortcuts in `/help`.

packages/cli/src/ui/App.tsx
```diff
@@ -447,6 +454,76 @@ const App = ({ config, settings, startupWarnings = [], version }: AppProps) => {
...
+        },
+      },
+      {
+        input: 'e',
+        ctrl: true,
+        condition: () => !!ideContext,
+        handler: () => setShowIDEContextDetail((prev) => !prev),
+      },
+      {
+        input: ['c', 'C'],
```
jacob314: to align with how things will look in the future when we align the input processing with what is in text-buffer.ts I think we should change this to allowing the command both with and without shift. rather than allowing both 'c' and 'C'

jacob314 (commented):
Some minor comments and areas for future discussion. It would be helpful to link this to a tracking issue. 
@dewitt and I were playing with this draft pull request trying to use Gemini CLI to help design the overall strategy for keyboard handling. You might find that discussion interesting. We definitely need someone to help drive this area forward. Are you interested?
https://github.com/google-gemini/gemini-cli/pull/3868

dewitt:
> Some minor comments and areas for future discussion. It would be helpful to link this to a tracking issue. @dewitt and I were playing with this draft pull request trying to use Gemini CLI to help design the overall strategy for keyboard handling. You might find that discussion interesting. We definitely need someone to help drive this area forward. Are you interested? #3868

+1, the entire area of keybindings still could use some love and attention. As @jacob314 mentioned, I took a morning the other week and attempted to collaborate using Gemini on a design doc for refactoring out that logic throughout and making the keybindings more declarative. 

This is the last version of the design doc that Gemini / Jacob / I checked in:

https://github.com/google-gemini/gemini-cli/blob/dewitt/keybindings-refactoring/docs/design/declarative-keybindings.md

(Jump to [Final Design](https://github.com/google-gemini/gemini-cli/blob/dewitt/keybindings-refactoring/docs/design/declarative-keybindings.md#final-design) if you're not interested in the whole back and forth.)

I don't think it's rock solid by any stretch, but it definitely captured at least some of the complexity involved in a whole-scale keybinding refactoring. Worth a skim for background info, anyway.

Would definitely love your input on (or ownership of!) this area, @chrstnb!

packages/cli/src/ui/App.tsx
chrstnb: that's a great idea! added to parent issue

packages/cli/src/ui/App.tsx
```diff
@@ -444,6 +449,77 @@ const App = ({ config, settings, startupWarnings = [], version }: AppProps) => {
...
+      {
+        input: 'e',
+        ctrl: true,
+        condition: () => !!ideContext,
+        handler: () => setShowIDEContextDetail((prev) => !prev),
+      },
+      {
+        input: 'c',
+        ctrl: true,
+        shift: true,
```
jacob314, line 481: is shift: true intended? I thought this one should allow both with and without shift?

packages/cli/src/ui/App.tsx
```diff
@@ -455,26 +531,20 @@ const App = ({ config, settings, startupWarnings = [], version }: AppProps) => {
...
-      if (Object.keys(mcpServers || {}).length > 0) {
-        handleSlashCommand(newValue ? '/mcp desc' : '/mcp nodesc');
-      }
-    } else if (key.ctrl && (input === 'c' || input === 'C')) {
-      handleExit(ctrlCPressedOnce, setCtrlCPressedOnce, ctrlCTimerRef);
-    } else if (key.ctrl && (input === 'd' || input === 'D')) {
-      if (buffer.text.length > 0) {
-        // Do nothing if there is text in the input.
-        return;
+    const handler = keystrokeHandlers.find((h) => {
```
jacob314, line 536: this logic to execute the keystroke handlers seems independent on app. I would move it to keystrokeHandler.ts (also please rename that file to align with naming schemes for non tsx files in this project.
Would then be nice to add a basic unittest for it that verifies cases such as if key.shift  and key.ctrl are handled correctly. seems like right now key.shift is ignored.

packages/cli/src/ui/components/IDEContextDetailDisplay.tsx
```diff
@@ -0,0 +1,55 @@
+/**
+ * @license
```
jacob314: please add a basic snapshot test for this UI.

jacob314 (commented):
A couple more comments. Approved when resolved. You may want to sync before resolving the comments as there are a few merge conflicts.

packages/cli/src/ui/App.tsx
cornmander: +1 to this. would add a method like findHandler(input, key). I also think it's worth having a class that would allow registering the handlers since it would let you add validation to prevent two handlers being registered on the same key:

KeystrokeHandlerRegistry(handlers: KeystrokeHandler[])
KeystrokeHandlerRegistry.findHandler(...) -> Optional<KeystrokeHandler>

packages/cli/src/ui/components/IDEContextDetailDisplay.tsx
chrstnb: This was added in an already-merged PR, sorry! Can def add a test later in a follow-up

jacob314:
Let me know if you would like to proceed with this. Changes to these files have been pretty intense over the past few weeks and others have landed some similar changes.

chrstnb:
I'm going to close this-- will re-visit if I have time and if not hoping someone else can grab!

PR #4791, opened 2025-07-24
feat: add GEMINI_CLI environment variable to spawned shell commands

taeold:
## Summary

Adds `GEMINI_CLI=1` environment variable to all spawned shell commands, allowing CLI tools to detect they're running inside Gemini CLI and adjust behavior (e.g., less verbose output, non-interactive mode).

## Changes

- Set `GEMINI_CLI=1` when spawning commands in shellCommandProcessor.ts and shell.ts
- Added tests for both implementations
- Cleaned up test imports

## Testing

- Unit tests verify spawn is called with correct env
- Integration test verifies env var is accessible to child processes
- All tests pass


jacob314 (approved):
lgtm

jacob314:
I'll land this as soon as the minor failing lints are resolved. 
Thanks for this nice polish feature.  We should also follow up with updates to our documentation so users are aware this env variable is available.

Issue #4390, opened 2025-07-17
Polish overall UX of IDE integration


skeshive:
This issue tracks polishing the overall UX of the IDE integration E2E including:

- Prompting the user to install the companion extension if they don't have it installed
- Ensuring user-facing error messages are accurate and specific

PR #4792, opened 2025-07-24
Add "Gemini CLI: Run" command shortcut

skeshive:
## TLDR
This command opens up gemini-cli in a new terminal window if the user already has the extension installed

## Linked issues / bugs
#4390


packages/vscode-ide-companion/src/extension.ts
```diff
@@ -24,6 +24,15 @@ export async function activate(context: vscode.ExtensionContext) {
     const message = err instanceof Error ? err.message : String(err);
     log(`Failed to start IDE server: ${message}`);
   }
+
+  context.subscriptions.push(
+    vscode.commands.registerCommand('gemini-cli.runGeminiCLI', () => {
+      const geminiCmd = 'gemini';
```
NTaylorMullen, line 30: We may want to validate this is on the path in the odd case that someone installed the extension without Gemini CLI

NTaylorMullen (approved):

Issue #3475, opened 2025-07-07
Reverse search for commands run in shell mode


smukhoo:
### What would you like to be added?

It would be great to support ctrl+r reverse search inside the shell mode

### Why is this needed?

I am finding myself doing a mix of manual edits and gemini cli assisted code changes. I then do a `hg diff -r <revision>` to make it see my changes and build upon it.

Currently I need to type out the whole command, having a reverse search here would be awesome.

PR #4793, opened 2025-07-24
feat: Add reverse search capability for shell commands

Aisha630:
## TLDR

Add reverse search mode in shell mode. 

## Dive Deeper
Type `Ctrl+r` to enter reverse string search mode when in shell mode. Use up/down to navigate matches.

https://github.com/user-attachments/assets/a6b20482-ac3f-483a-8deb-c192e429a5f3

## Reviewer Test Plan

1.  Enter shell mode with `!`
2.  Press `Ctrl+R` to activate reverse search

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs
#3475


packages/cli/src/ui/components/InputPrompt.tsx
```diff
@@ -296,6 +343,13 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
       }
 
       if (key.name === 'escape') {
+        if (reverseSearchActive) {
+          setReverseSearchActive(false);
+          reverseSearchCompletion.resetCompletionState();
+          buffer.setText(textBeforeReverseSearch);
+          return;
+        }
```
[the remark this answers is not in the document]
Aisha630: I do not think this is something that is required or high priority

packages/cli/src/ui/hooks/useShellHistory.ts
```diff
@@ -24,15 +25,32 @@ async function getHistoryFilePath(projectRoot: string): Promise<string> {
...
-    console.error('Error reading shell history:', error);
+
+    if (cur) result.push(cur);
+    return result;
+  } catch (err) {
+    if (isNodeError(err) && err.code === 'ENOENT') return [];
+    console.error('Error reading history:', err);
     return [];
   }
 }
```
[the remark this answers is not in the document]
Aisha630: This is already taken care of when adding a new command to history in the `addCommandToHistory`

Aisha630:
@NTaylorMullen @jacob314 Would love your feedback on this

packages/cli/src/ui/components/InputPrompt.tsx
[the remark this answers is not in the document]
jacob314: I do agree with the bot on this although it is somewhat minor.
to repro  type
foo <cursor here> bar
then enter reverse search mode
press escape
strangely the cursor is now at the end of the line rather than at <cursor here>
this should be quick to fix as you just need to store the cursor position when reverse search mode was entered and restore it along with restoring the text from before.

packages/cli/src/ui/components/InputPrompt.tsx
```diff
@@ -197,11 +226,58 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
         }
       }
 
+      if (shellModeActive && key.ctrl && key.name === 'r') {
+        setReverseSearchActive(true);
+        setTextBeforeReverseSearch(buffer.text);
+        return;
```
jacob314, line 253: also save the cursor position to restore here.

packages/cli/src/ui/components/InputPrompt.tsx
```diff
@@ -197,11 +226,58 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
...
+        if (showSuggestions) {
+          if (key.name === 'up') {
+            navigateUp();
+            return;
+          }
+          if (key.name === 'down') {
+            navigateDown();
+            return;
+          }
+          if (key.name === 'tab') {
```
jacob314, line 279: shouldn't this also exit reverse search mode to align with other terminals?

packages/cli/src/ui/components/InputPrompt.tsx
```diff
@@ -307,11 +353,57 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
...
+          handleSubmitAndClear(textToSubmit);
+          reverseSearchCompletion.resetCompletionState();
+          setReverseSearchActive(false);
+          return;
+        }
+
+        // Prevent up/down from falling through to regular history navigation
+        if (key.name === 'up' || key.name === 'down') {
+          return;
+        }
```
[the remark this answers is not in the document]
jacob314: don't think this is needed

packages/cli/src/ui/components/SuggestionsDisplay.tsx
```diff
@@ -59,17 +60,44 @@ export function SuggestionsDisplay({
         const isActive = originalIndex === activeIndex;
         const textColor = isActive ? Colors.AccentPurple : Colors.Gray;
 
+        // Helper to render label with matched part highlighted
```
jacob314: make this helper a little react component generating the label. gemini cli should be able to one shot this refactor. this will align with the style guide for gemini cli.

packages/cli/src/ui/components/InputPrompt.test.tsx
```diff
@@ -1094,7 +1137,6 @@ describe('InputPrompt', () => {
     });
 
     it('should call buffer.handleInput when vim mode is enabled but vim does not handle the input', async () => {
-      props.vimModeEnabled = true;
```
jacob314: why has vimModeEmabled been removed?

packages/cli/src/ui/components/InputPrompt.test.tsx
```diff
@@ -1080,7 +1124,6 @@ describe('InputPrompt', () => {
 
   describe('vim mode', () => {
     it('should not call buffer.handleInput when vim mode is enabled and vim handles the input', async () => {
-      props.vimModeEnabled = true;
```
jacob314: why has vimModeEmabled been removed?

packages/cli/src/ui/components/InputPrompt.test.tsx
```diff
@@ -19,7 +19,14 @@ import {
   useShellHistory,
   UseShellHistoryReturn,
```
jacob314, line 20: please add some tests for useReverseSearchCompletion to this file.

Note that while the previous tests mocked useSlashCompletion that is somewhat unfortunate. Would be good if your tests used the real useReverseSearchCompletion. Given these completions don't have to use the file system I would expect there should be no restrictions to support that.

jacob314 (commented):
Thank you for sticking with this feature. Users are really going to like having reverse search support. The implementation now looks quite good. Once these hopefully fairly minor comments are addressed I'll aprove.
As a follow up PR it would be awesome if we could also support reverse search through the history of all prompts the user has entered in the past. For fairly often entered chat commands that would be quite useful.

packages/cli/src/ui/components/InputPrompt.tsx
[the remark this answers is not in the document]
Aisha630: Added this

packages/cli/src/ui/components/InputPrompt.tsx
Aisha630: Done and verified. Added test cases as well

packages/cli/src/ui/components/InputPrompt.test.tsx
Aisha630: I have now added test cases as well both to `InputPrompt.test.tsx` and for the hook level functionaltiy as `useReverseSearchcompletion`. Both of these use the actual hook itself and not a mocked version

packages/cli/src/ui/components/InputPrompt.test.tsx
Aisha630: My bad, it was never a property in InputPromptProps so I removed it without investigating. I have added these back now, though, since it is passed through the VimMode context provider.

packages/cli/src/ui/components/InputPrompt.test.tsx
Aisha630: fixed

packages/cli/src/ui/components/SuggestionsDisplay.tsx
Aisha630: Done

packages/cli/src/ui/components/InputPrompt.tsx
Aisha630: I have made it exit reverse search now, but the idea was to let users edit more and then continue searching based on that edited command. I guess that is not necessary so I have made it exit reverse search

jacob314 (approved):
lgtm

jacob314:
Approved. Please resolve the conflicts and the. i will approve. Heads up, another contributor is trying to add history search when not in chat mode as a fast follow to this change.

Aisha630:
> Approved. Please resolve the conflicts and the. i will approve. Heads up, another contributor is trying to add history search when not in chat mode as a fast follow to this change.

I have resolved them

PR #3487, opened 2025-07-07
fix: correct mcp client name and version

seuros:
The mcp client keeps sending the same version 0.0.1 since dawn of time.

This PR, use the package version to give gemini mcp client a version.

PR #4797, opened 2025-07-24
[ide-mode] Create an IDE manager class to handle connecting to and exposing methods from the IDE server

chrstnb:
## TLDR

This is an incremental change towards an extensible implementation. 
- Abstracts the underlying implementation of the IDE server, allowing us to support integrations outside of MCP servers
- Centralizes logic surrounding IDE mode so we can do more/custom handling as things progress

This uses its own mcp client to connect to the IDE server to decouple this from the user-installed MCP implementation-- we may want to expose this MCP server as part of the other MCP servers to the model in the future, but an alternative implementation is to define a tool the model can use. 

In follow-ups:
- Add support for server re-tries and potentially reading the port from a file-- this will allow us to dynamically load the MCP server even if the terminal was created prior to the IDE writing to the port (this happens often in VSCode).
- Expose functions like openFileDiff, likely in a tool
- Add a transport/protocol for MCP servers that are compatible with this implementation

## Reviewer Test Plan

Test out the current IDE mode and ensure everything still works.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

#4800


packages/core/src/ide/ide-mode-manager.ts
```diff
@@ -0,0 +1,76 @@
...
+import { Client } from '@modelcontextprotocol/sdk/client/index.js';
+import { StreamableHTTPClientTransport } from '@modelcontextprotocol/sdk/client/streamableHttp.js';
+
+const logger = {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  debug: (...args: any[]) =>
+    console.debug('[DEBUG] [ImportProcessor]', ...args),
+};
+
+export enum IDEConnectionStatus {
```
NTaylorMullen: Given we're leaning on MCP for this any reason to not re-use our existing status details?

packages/core/src/ide/ide-mode-manager.ts
```diff
@@ -0,0 +1,76 @@
...
+  constructor() {
+    this.connectToMcpServer().catch((err) => {
+      logger.debug('Failed to initialize IdeModeManager:', err);
+    });
+  }
+  getConnectionStatus(): IDEConnectionStatus {
+    return this.connectionStatus;
+  }
+
+  async connectToMcpServer(): Promise<void> {
```
NTaylorMullen: Because of this the `ide-mode-manager` feels more like a n `ide-client`

packages/core/src/ide/ide-mode-manager.ts
```diff
@@ -0,0 +1,76 @@
...
+        'Unable to connect to IDE mode MCP server. GEMINI_CLI_IDE_SERVER_PORT environment variable is not set.',
+      );
+      this.connectionStatus = IDEConnectionStatus.Disconnected;
+      return;
+    }
+
+    try {
+      this.client = new Client({
+        name: 'streamable-http-client',
+        version: '1.0.0',
```
NTaylorMullen: We may want this version to reflect the CLI version eventually. See: https://github.com/google-gemini/gemini-cli/pull/3487

NTaylorMullen (approved):

packages/core/src/ide/ide-mode-manager.ts
```diff
@@ -0,0 +1,76 @@
...
+}
+
+/**
+ * Manages the connection to and interaction with the IDE server.
+ */
+export class IdeModeManager {
+  client: Client | undefined = undefined;
+  connectionStatus: IDEConnectionStatus = IDEConnectionStatus.Disconnected;
+
+  constructor() {
```
skeshive: 1) instead of initializing IDE mode implicitly, can we initialize it when the user invokes the /ide slash command? so the /ide slash command would act as a toggle for enabling / disabling IDE mode
2) until we're ready to launch, we should still hide /ide behind the --ide-mode flag

packages/core/src/ide/ide-mode-manager.ts
```diff
@@ -0,0 +1,76 @@
...
+    try {
+      this.client = new Client({
+        name: 'streamable-http-client',
+        version: '1.0.0',
+      });
+      const transport = new StreamableHTTPClientTransport(
+        new URL(`http://localhost:${idePort}/mcp`),
+      );
+      await this.client.connect(transport);
+      this.client.setNotificationHandler(
```
skeshive: we should also add the onerror handler to clear file context when the server disconnects: https://github.com/google-gemini/gemini-cli/blob/main/packages/core/src/tools/mcp-client.ts#L383

packages/vscode-ide-companion/src/ide-server.ts
```diff
@@ -229,7 +229,7 @@ const createMcpServer = () => {
     { capabilities: { logging: {} } },
   );
   server.registerTool(
```
skeshive: I think we can actually remove this tool now since we're not relying on the gemini-cli mcp-client which disconnects servers without any tools :)

packages/core/src/services/ideContext.ts
skeshive, line 1: lets move this to the /ide dir?

packages/core/src/ide/ide-mode-manager.ts
```diff
@@ -0,0 +1,76 @@
...
+  getConnectionStatus(): IDEConnectionStatus {
+    return this.connectionStatus;
+  }
+
+  async connectToMcpServer(): Promise<void> {
+    this.connectionStatus = IDEConnectionStatus.Connecting;
+    const idePort = process.env['GEMINI_CLI_IDE_SERVER_PORT'];
+    if (!idePort) {
+      logger.debug(
+        'Unable to connect to IDE mode MCP server. GEMINI_CLI_IDE_SERVER_PORT environment variable is not set.',
```
skeshive: lets surface this error message to the user via /ide and indicate the companion extension must be installed + they must run in the VS Code integrated terminal

skeshive (dismissed):

chrstnb:
Note: let's add support here for threading through the name of the IDE https://github.com/google-gemini/gemini-cli/pull/4682#discussion_r2229456676

packages/core/src/ide/ide-mode-manager.ts
chrstnb: I think it's possible we might want to support different protocols so wanted to keep them separate so the fact that it's relying on an MCP server is an implementation detail of the manager (now client). WDYT?

packages/core/src/ide/ide-mode-manager.ts
chrstnb: I think this is a good idea but out of scope for this PR-- I think we should move all of the install stuff into this package as an 'install registry', allowing us to surface installable options but still allowing other options to connect via the protocol. WDYT?

packages/core/src/ide/ide-mode-manager.ts
chrstnb: I moved it so it gets kicked off if ide-mode is enabled to match the current behavior, but we have more flexibility if we want to invoke it ad hoc! I think it makes sense in a follow up tho?

PR #4813, opened 2025-07-24
Sanitize checkpoint tags

sethtroisi:
## TLDR

Prevent directory transversal from `/chat save /../../../test.txt`

## Reviewer Test Plan

Verified locally.

## Testing Matrix

I plan to add follow up tests but thought it best to get this in first


jacob314 (approved):
bonus points: can you follow up with a test that verifies that we only accept sanitized paths?lgtm

PR #4824, opened 2025-07-25
Add support for tracking slash command usage + log usage of /ide commands

### LINKED ISSUE #4395: Track usage of IDE mode
We need to add some metrics to track the usage of IDE mode so we can verify whether or not users are discovering this feature

skeshive:
## TLDR

- Logs every time the /ide, /ide status and /ide install commands are invoked
- Logs once per session when a notification is recieved

## Linked issues / bugs
Fixes #4395

### DIFF
--- a/packages/cli/src/ui/commands/ideCommand.ts
+++ b/packages/cli/src/ui/commands/ideCommand.ts
@@ -12,8 +12,8 @@ import {
   IDE_SERVER_NAME,
   MCPDiscoveryState,
   MCPServerStatus,
-  logIdeCommandTriggered,
-  IdeCommandTriggeredEvent,
+  logSlashCommandTriggered,
+  SlashCommandTriggeredEvent,
 } from '@google/gemini-cli-core';
 import {
   CommandContext,
@@ -54,7 +54,7 @@ export const ideCommand = (config: Config | null): SlashCommand | null => {
     kind: CommandKind.BUILT_IN,
     action: () => {
       if (config) {
-        logIdeCommandTriggered(config, new IdeCommandTriggeredEvent('ide'));
+        logSlashCommandTriggered(config, new SlashCommandTriggeredEvent('ide'));
       }
     },
     subCommands: [
@@ -64,9 +64,9 @@ export const ideCommand = (config: Config | null): SlashCommand | null => {
         kind: CommandKind.BUILT_IN,
         action: (_context: CommandContext): SlashCommandActionReturn => {
           if (config) {
-            logIdeCommandTriggered(
+            logSlashCommandTriggered(
               config,
-              new IdeCommandTriggeredEvent('status'),
+              new SlashCommandTriggeredEvent('ide', 'status'),
             );
           }
           const status = getMCPServerStatus(IDE_SERVER_NAME);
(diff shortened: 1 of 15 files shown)


packages/core/src/telemetry/types.ts
jerop, line 1: could we make this more general to be about tracking slash commands being triggered, then we can have a "type" of the slash command, one of which is /ide

packages/core/src/telemetry/constants.ts
jerop, line 1: please update the data reference in https://github.com/google-gemini/gemini-cli/blob/e9ee686ab6c5c99d895d80951336613f248f6560/docs/telemetry.md#logs-and-metric-reference

jerop (commented):
This is a cool addition, wonder if we can make it more applicable to all slash commands? Don't think it'd scale well if we have separate metrics for all slash commands

packages/core/src/telemetry/types.ts
skeshive: good idea! added support for logging any slash command but only tracking /ide for now since I want to keep this PR small

packages/core/src/telemetry/constants.ts
skeshive: done

packages/cli/src/ui/commands/ideCommand.ts
```diff
@@ -50,12 +52,23 @@ export const ideCommand = (config: Config | null): SlashCommand | null => {
...
+        logSlashCommandTriggered(config, new SlashCommandTriggeredEvent('ide'));
+      }
+    },
     subCommands: [
       {
         name: 'status',
         description: 'check status of IDE integration',
         kind: CommandKind.BUILT_IN,
         action: (_context: CommandContext): SlashCommandActionReturn => {
+          if (config) {
```
NTaylorMullen: Do we need to do `if (config)` here? We do `if (!config?.getIdeMode()) {` above which will also shortcircuit if config is undefined. If TypeScript is complaining we should update the if statement on line 47 so we don't have to do all of these if configs

packages/cli/src/ui/commands/ideCommand.ts
```diff
@@ -50,12 +52,23 @@ export const ideCommand = (config: Config | null): SlashCommand | null => {
     name: 'ide',
     description: 'manage IDE integration',
     kind: CommandKind.BUILT_IN,
+    action: () => {
+      if (config) {
+        logSlashCommandTriggered(config, new SlashCommandTriggeredEvent('ide'));
```
NTaylorMullen: I assume goal is to eventually make `/ide` on its own do more right?

packages/cli/src/ui/commands/ideCommand.ts
```diff
@@ -94,6 +107,12 @@ export const ideCommand = (config: Config | null): SlashCommand | null => {
         description: 'install required VS Code companion extension',
         kind: CommandKind.BUILT_IN,
         action: async (context) => {
+          if (config) {
+            logSlashCommandTriggered(
```
NTaylorMullen: Instead of incorporating these at all layers for all slash commands individually can we log this more centrally? I.e. the bit that calls into these slash command / actions?

PR #4828, opened 2025-07-25
Load and use MCP server prompts as slash commands in the CLI

chrstnb:
## TLDR

@haroldmciver-go's changes introduce loading prompts and displaying them 

Added follow-up changes to support loading them in as slash commands. I'm imagining we're going to want better UX on this but would love feedback on what!

Currently it
1. Loads any MCP server prompts in as slash commands and displays the prompt name (as slash command) followed by the description 
2. If you try to invoke it without a required arg, it will yell at you
3. Once you type in required args, it will invoke the prompt and send the prompt (filled out with your arguments) to gemini on your behalf

Also includes a help subcommand

Screenshot 2025-07-25 at 10 34 34 AM

And handles collisions by appending the MCP server name (the first registered prompt will not have the MCP server name appended). 
Screenshot 2025-07-25 at 10 34 17 AM

## Reviewer Test Plan

https://github.com/user-attachments/assets/479d1145-9540-4a4a-b98b-4503f110cd90

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


packages/cli/src/ui/commands/promptCommands.ts
```diff
@@ -0,0 +1,183 @@
...
+  Config,
+  getMCPServerPrompts,
+  getErrorMessage,
+  invokeMcpPrompt,
+  connectToMcpServer,
+} from '@google/gemini-cli-core';
+
+export function createPromptCommands(config: Config | null): SlashCommand[] {
+  const promptCommands: SlashCommand[] = [];
+  if (config) {
```
NTaylorMullen: nit: How do you feel about early. returning here to reduce nesting? I.e. `if (!config) { return promptCommands; }`

packages/cli/src/ui/commands/promptCommands.ts
```diff
@@ -0,0 +1,183 @@
...
+                messageType: 'error',
+                content: 'Config not loaded.',
+              };
+            }
+
+            const promptArgs = args.trim();
+
+            const argValues: { [key: string]: string } = {};
+            // arg parsing: --key="value" or --key=value
+            const argRegex = /--([^=]+)=(?:"((?:\\.|[^"\\])*)"|([^ ]*))/g;
```
NTaylorMullen: Do other clients also try and parse args in this way? It does feel a bit heavy to provide name => value. I would have imagined that other clients would allow positional arguments in some form. Would you mind double checking this bit?

packages/core/src/tools/mcp-client.ts
```diff
@@ -125,6 +139,15 @@ export function getMCPServerStatus(serverName: string): MCPServerStatus {
   );
 }
 
+/**
+ * Get the discovered prompts for an MCP server
+ */
+export function getMCPServerPrompts(
```
NTaylorMullen: It's already on the mcp-client

```suggestion
export function getPrompts(
```

packages/core/src/tools/mcp-client.ts
```diff
@@ -62,6 +71,11 @@ export enum MCPDiscoveryState {
  */
 const mcpServerStatusesInternal: Map<string, MCPServerStatus> = new Map();
 
+/**
+ * Map to track discovered prompts for each MCP server
+ */
+const mcpServerPromptsInternal: Map<string, DiscoveredMCPPrompt[]> = new Map();
```
NTaylorMullen: I feel like the statusesInternal format above is being a bit too verbose as well. I see you followed that though, ideally these would both be `serverStatuses` and `prompts`

packages/core/src/tools/mcp-client.ts
```diff
@@ -459,15 +487,91 @@ export async function discoverTools(
...
+/**
+ * Invokes a prompt on a connected MCP client.
+ *
+ * @param mcpServerName The name of the MCP server.
+ * @param mcpClient The active MCP client instance.
+ * @param promptName The name of the prompt to invoke.
+ * @param promptParams The parameters to pass to the prompt.
+ * @returns A promise that resolves to the result of the prompt invocation.
+ */
+export async function invokeMcpPrompt(
```
NTaylorMullen, line 518: At some point we need to refactor all these to be off of the mcp client / server, the fact that we have to pass on the server name client etc. is a bit heavy.

packages/cli/src/ui/commands/promptCommands.ts
chrstnb: Adding support for this as well!

packages/core/src/tools/mcp-client.ts
chrstnb: added a prompt registry with a callback-- I totally agree!

packages/cli/src/services/McpPromptLoader.ts
```diff
@@ -0,0 +1,240 @@
...
+  constructor(private readonly config: Config | null) {}
+
+  /**
+   * Loads all available prompts from all configured MCP servers and adapts
+   * them into executable SlashCommand objects.
+   *
+   * @param _signal An AbortSignal (unused for this synchronous loader).
+   * @returns A promise that resolves to an array of loaded SlashCommands.
+   */
+  async loadCommands(_signal: AbortSignal): Promise<SlashCommand[]> {
```
NTaylorMullen: Can drop `async` here and return a completed promise

packages/cli/src/services/McpPromptLoader.ts
```diff
@@ -0,0 +1,240 @@
...
+                if (!prompt.arguments || prompt.arguments.length === 0) {
+                  return {
+                    type: 'message',
+                    messageType: 'info',
+                    content: `Prompt "${prompt.name}" has no arguments.`,
+                  };
+                }
+
+                let helpMessage = `Arguments for "${prompt.name}":\n\n`;
+                helpMessage += `You can provide arguments by name (e.g., --argName="value") or by position.\n\n`;
```
NTaylorMullen: `by position` may be hard to interpret here is there a better way we can showcase what that means?

packages/cli/src/services/McpPromptLoader.ts
```diff
@@ -0,0 +1,240 @@
...
+              };
+            } catch (error) {
+              return {
+                type: 'message',
+                messageType: 'error',
+                content: `Error: ${getErrorMessage(error)}`,
+              };
+            }
+          },
+          completion: async (_: CommandContext, partialArg: string) => {
```
NTaylorMullen, line 147: oh very cool usage of completion!

packages/cli/src/services/McpPromptLoader.ts
```diff
@@ -0,0 +1,240 @@
...
+            }
+
+            const suggestions: string[] = [];
+            const usedArgNames = new Set(
+              (partialArg.match(/--([^=]+)/g) || []).map((s) => s.substring(2)),
+            );
+
+            for (const arg of prompt.arguments) {
+              if (!usedArgNames.has(arg.name)) {
+                suggestions.push(`--${arg.name}=`);
```
NTaylorMullen: One interesting decision point by surfacing these in completion is that we don't surface with quotes. Meaning, if a user does `--name=John Do` the "doe" ends up getting dropped silently. It's a hard problem though because for most arguments they'll probably be simple and not require spaces

packages/cli/src/services/McpPromptLoader.ts
```diff
@@ -0,0 +1,240 @@
...
+      }
+    }
+
+    const unfilledArgs = promptArgs.filter(
+      (arg) => arg.required && !promptInputs[arg.name],
+    );
+
+    const missingArgs: string[] = [];
+    for (let i = 0; i < unfilledArgs.length; i++) {
+      if (positionalArgs.length > i && positionalArgs[i]) {
```
NTaylorMullen, line 218: Interesting idea, what if we fell back to an LLM query to map positional arguments if we ended up having too many.

For instance, I just ran into this where my 'preamble' prompt has a firstName and lastName prompt:
image

NTaylorMullen (approved):
Had some comments, we'll also want to monitor the `--` syntax utilization by customers. AFAIK this will be unique to Gemini CLI.

That being said, looks good overall. Feel free to merge once @abhipatel12 signs off!

packages/cli/src/services/McpPromptLoader.ts
chrstnb: I'll add quotes to be safe? the completions might take a little trial and error

packages/cli/src/services/McpPromptLoader.ts
chrstnb: This is super interesting! I think there might be more work to do around processing-- created a bug for tracking! https://github.com/google-gemini/gemini-cli/issues/4885

packages/cli/src/ui/hooks/slashCommandProcessor.ts
```diff
@@ -166,6 +167,7 @@ export const useSlashCommandProcessor = (
       const loaders = [
         new BuiltinCommandLoader(config),
         new FileCommandLoader(config),
+        new McpPromptLoader(config),
```
abhipatel12: Just one thing to keep in mind is that if there any name collisions, we're setting a precedence here. 

Mcp Prompts > File Loaded Prompts > Builtin Commands

Until we have a better name collision strategy, im wondering if builtin commands should have the highest order

packages/cli/src/ui/hooks/slashCommandProcessor.ts
abhipatel12, line 1: Can we add a couple of tests in the test file for this one since we added a new loader. This will also check the precedence that I mentioned in this file.

packages/cli/src/services/McpPromptLoader.ts
```diff
@@ -0,0 +1,231 @@
...
+   * @param _signal An AbortSignal (unused for this synchronous loader).
+   * @returns A promise that resolves to an array of loaded SlashCommands.
+   */
+  loadCommands(_signal: AbortSignal): Promise<SlashCommand[]> {
+    const promptCommands: SlashCommand[] = [];
+    if (!this.config) {
+      return Promise.resolve([]);
+    }
+    const mcpServers = this.config.getMcpServers() || {};
+    for (const serverName in mcpServers) {
```
abhipatel12, line 41: How are we handling name collisions at the loader level? Right now, these will eventually get deduped at the `CommandService` layer, but the user will have no idea why one mcp prompt was chosen vs another. 

We don't necessarily have to make this choice in this PR, but something to consider. This may play into the broader choice to add namespaces to help reduce the chances for overrides.

abhipatel12 (approved):
This is awesome! Thank you for getting this in!!

packages/cli/src/ui/hooks/slashCommandProcessor.ts
chrstnb: sounds good to me!

packages/cli/src/services/McpPromptLoader.ts
chrstnb: Collisions are getting handled in the prompt registry!

If, say, mcpServer1 has prompt 'review-code' and so does mcpServer2, the loaded prompts will be:

'review-code'

'mcpServer2_review-code'

abhipatel12: Ohhhh gotcha, missed that. Sorry!

Issue #4800, opened 2025-07-24
Enable extensibility for IDE integrations


skeshive:
## What would you like to be added?
Documentation: We need clear documentation that describes how IDE integration can be enabled for other IDEs. This should include details on the notification schemas and supported communication protocols.

Implementation Robustness: We must ensure the current implementation is stable. For example, logic like truncating context should be handled by the server not the client extension. This will prevent inconsistent behavior across different clients.

Protocol Evaluation: We should decide if we need to switch from HTTP to stdio for the VS Code extension, similar to the cider integration.

## Why is this needed?
Enable contributors to create their own IDE integrations

PR #4877, opened 2025-07-25
Introduce IDE mode installer

chrstnb:
## TLDR

Moving the vscode-specific logic into an IDE mode installer class, which includes a list of supported IDEs.

Additionally, abstract logic to detect the current IDE.

## Dive Deeper

In follow-ups, we can leverage this to, e.g., re-connect to the IDE server once the extension is installed, or to support different types of configurations.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

#4800


packages/cli/src/config/config.ts
```diff
@@ -258,9 +259,10 @@ export async function loadCliConfig(
       (v) => v === 'true' || v === '1',
     );
 
+  const currentIde = detectIde();
   const ideMode =
     (argv.ideMode ?? settings.ideMode ?? false) &&
-    process.env.TERM_PROGRAM === 'vscode' &&
+    !!currentIde &&
```
skeshive: we can remove the term_prog check here actually since the expected port var won't be set unless we're running in the vs code integrated terminal

packages/core/src/config/config.ts
```diff
@@ -582,6 +586,10 @@ export class Config {
     return this.ideClient;
   }
 
+  getCurrentIde(): SupportedIDE | undefined {
```
skeshive: should this be a field of ide-client instead?

packages/core/src/ide/detect-ide.ts
```diff
@@ -0,0 +1,25 @@
...
+export function getIdeDisplayName(ide: SupportedIDE): string {
+  switch (ide) {
+    case SupportedIDE.VSCode:
+      return 'VSCode';
+    default:
+      throw new Error(`Unsupported IDE: ${ide}`);
+  }
+}
+
+export function detectIde(): SupportedIDE | undefined {
```
skeshive: hmm should ide-client detect the IDE before attempting to connect to the port to confirm its running in a supported IDE?

packages/core/src/ide/ide-installer.ts
```diff
@@ -0,0 +1,102 @@
...
+
+  async install(): Promise<InstallResult> {
+    if (!this.isInstalled()) {
+      return {
+        success: false,
+        message: `VS Code command-line tool "${VSCODE_COMMAND}" not found in your PATH.`,
+      };
+    }
+
+    const bundleDir = path.dirname(fileURLToPath(import.meta.url));
```
skeshive, line 111: were you able to test the bundled path? /ide install wasn't working for me when running gemini-cli via the gemini command in a different repo

packages/core/src/ide/detect-ide.ts
```diff
@@ -0,0 +1,25 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export enum SupportedIDE {
```
skeshive: nit: DetectedIde to be more explicit that this the current IDE the CLI is running in

packages/core/src/ide/detect-ide.ts
```diff
@@ -0,0 +1,25 @@
...
+  VSCode = 'vscode',
+}
+
+export function getIdeDisplayName(ide: SupportedIDE): string {
```
skeshive: lets also use this when displaying the context per Taylor's comment in a previous PR

skeshive (changes requested):

packages/cli/src/config/config.ts
chrstnb: I don't think I'm following-- are you saying we should remove the currentIde check?

packages/core/src/ide/ide-installer.ts
chrstnb: yes! when running `npx github:google-gemini/gemini-cli#test-ide-mode --ideMode=true` it's working for me

I'll revert the path change tho because that doesn't seem relevant

packages/core/src/ide/detect-ide.ts
chrstnb: Added in one place but can add in others depending on UX feedback

packages/cli/src/config/config.ts
skeshive: no I was saying we can remove the process.env.TERM_PROGRAM === 'vscode' check since the expected port env var will only be set when the user's running in the integrated terminal!

packages/core/src/ide/ide-client.ts
```diff
@@ -54,6 +73,11 @@ export class IdeClient {
   }
 
   async connectToMcpServer(): Promise<void> {
+    if (!this.currentIde) {
```
skeshive: https://github.com/google-gemini/gemini-cli/blob/main/packages/core/src/ide/ide-client.ts#L140

lets add a similar condition in init() for this

packages/core/src/config/config.ts
```diff
@@ -589,6 +591,14 @@ export class Config {
     return this.ideClient;
   }
 
+  getDetectedIdeDisplayName(): string | undefined {
```
skeshive: can IDE client hold this state instead?

skeshive (approved):

packages/cli/src/config/config.ts
chrstnb: It's removed tho!

packages/core/src/ide/ide-client.ts
```diff
@@ -141,6 +160,10 @@ export class IdeClient {
     if (this.state.status === IDEConnectionStatus.Connected) {
       return;
     }
+    if (!this.currentIde) {
+      throw new Error('Not running in a supported IDE, skipping connection.');
```
skeshive: we probably don't want to throw an error, can we set the state to disconnected and show a user facing error message?

packages/core/src/ide/ide-client.ts
```diff
@@ -154,4 +177,12 @@ export class IdeClient {
 
     await this.establishConnection(port);
   }
+
+  getDetectedIdeDisplayName(): string | undefined {
```
skeshive, line 192: should this also be a instance variable that we set from init?

packages/core/src/config/config.ts
```diff
@@ -579,7 +579,7 @@ export class Config {
   }
 
   getIdeMode(): boolean {
-    return this.ideMode;
+    return this.ideMode && !!this.ideClient;
```
skeshive: could the client always be defined? in the constructor, we can pass in the ideMode flag and if the flag is false, we just set the status to disconnected?

that way we dont have to reason about whether the client is defined or not depending on the ideMode flag

packages/cli/src/ui/commands/ideCommand.ts
```diff
@@ -4,40 +4,27 @@
...
-  }
-}
 
 export const ideCommand = (config: Config | null): SlashCommand | null => {
   if (!config?.getIdeMode()) {
     return null;
   }
+  const currentIDE = config.getIdeClient()?.getCurrentIde();
+  if (!currentIDE) {
+    return null;
```
NTaylorMullen: This will trigger if we don't recognize the IDE right? Should we do a proper error here?

packages/cli/src/ui/components/IDEContextDetailDisplay.tsx
```diff
@@ -30,7 +32,8 @@ export function IDEContextDetailDisplay({
       paddingX={1}
     >
       <Text color={Colors.AccentCyan} bold>
-        IDE Context (ctrl+e to toggle)
+        {detectedIdeDisplay ? `${detectedIdeDisplay} ` : ''}IDE Context (ctrl+e
```
NTaylorMullen: IDE seems a bit repetitive here. We could probably get away with only showing the `detectedIdeDisplay` so it shows as `VSCode Context`

packages/core/src/ide/ide-installer.ts
```diff
@@ -0,0 +1,102 @@
...
+ */
+
+import * as child_process from 'child_process';
+import * as process from 'process';
+import { glob } from 'glob';
+import * as path from 'path';
+import { fileURLToPath } from 'url';
+import { DetectedIde } from './detect-ide.js';
+
+const VSCODE_COMMAND = process.platform === 'win32' ? 'code.cmd' : 'code';
```
NTaylorMullen, line 16: Note: we may need tomake discovering this command a little more dynamic. IIRC VSCode can also install as a user-path and not as an env variable at `$HOME/Library/Application Support/Code` I think?

packages/core/src/ide/ide-installer.ts
```diff
@@ -0,0 +1,102 @@
...
+    const bundleDir = path.dirname(fileURLToPath(import.meta.url));
+    // The VSIX file is copied to the bundle directory as part of the build.
+    let vsixFiles = glob.sync(path.join(bundleDir, '*.vsix'));
+    if (vsixFiles.length === 0) {
+      // If the VSIX file is not in the bundle, it might be a dev
+      // environment running with `npm start`. Look for it in the original
+      // package location, relative to the bundle dir.
+      const devPath = path.join(
+        bundleDir,
+        '..',
```
NTaylorMullen: Can you add comments for what each of these updirs represent

NTaylorMullen (approved):

packages/core/src/ide/ide-client.ts
chrstnb: This is getting caught above-- there's no way to send a scenario-specific message otherwise

packages/cli/src/ui/components/IDEContextDetailDisplay.tsx
chrstnb: good call!

packages/core/src/ide/ide-installer.ts
chrstnb: Added a more thorough search (gemini-generated)

packages/core/src/ide/ide-client.ts
chrstnb: I don't feel strongly, but I prefer to have it set dynamically so we can fetch it dynamically (e.g., configured by the MCP server)

packages/core/src/config/config.ts
skeshive: the !!this.client check can be removed now

packages/cli/src/ui/App.tsx
```diff
@@ -967,7 +967,12 @@ const App = ({ config, settings, startupWarnings = [], version }: AppProps) => {
                 </Box>
               </Box>
               {showIDEContextDetail && (
-                <IDEContextDetailDisplay ideContext={ideContextState} />
+                <IDEContextDetailDisplay
+                  ideContext={ideContextState}
+                  detectedIdeDisplay={config
+                    .getIdeClient()
+                    ?.getDetectedIdeDisplayName()}
```
skeshive: ideClient will always be defined now so we can remove the undefined check

packages/cli/src/ui/commands/ideCommand.ts
```diff
@@ -4,40 +4,29 @@
...
-  }
-}
 
 export const ideCommand = (config: Config | null): SlashCommand | null => {
   if (!config?.getIdeMode()) {
     return null;
   }
+  const currentIDE = config.getIdeClient().getCurrentIde();
+  if (!currentIDE) {
+    throw new Error(
```
skeshive, line 26: this should be an error message we display to the user, otherwise running /ide would crash for anyone running gemini-cli with --ide-mode enabled outside of VS Code

packages/core/src/ide/ide-client.ts
skeshive: hmm the detected IDE / display name wouldnt change per session though

packages/core/src/ide/ide-client.ts
skeshive: not following "there's no way to send a scenario-specific message otherwise" > wouldn't setting the status to disconnected with a details message get propagated when the user types /ide status (see the port & workspace path handling)

the disconnect messages also get logged from setState()

dtometzki:
Hi together,

can you please update the doc for vscode-ide-companion because. Becaus this here doesnt really help

- use this extension, you'll need:

- VS Code version 1.101.0 or newer

And please tell you need to start gemini-cli with the --ide-node parameter

Damian

packages/core/src/ide/ide-client.ts
chrstnb: I'm saying this because there's a 'catch' above 

https://github.com/google-gemini/gemini-cli/pull/4877/files/e9fc274c9973b0fec554a64783ff1b0ef9d27f91#diff-1861bdc75ed9df0012b1ab4446cba3a72618caa2d4a72a85ed6bfa4ebd35fcaeL37

Using the error to bubble up a more specific debug error message (we also should def not be hitting this case right?)

packages/core/src/ide/ide-client.ts
chrstnb: fair enough, don't feel strongly. will update. I think this is very low-stakes, but is moderately more brittle this way because now we need to set the ide display name when we fetch the IDE, and there's no computational upside because it's a static value from an enum switch case

Issue #4870, opened 2025-07-25
Gemini cli Erroneously Denies Capabilities, Causing Contradictory and Blocked User Interactions


gamunu:
### What happened?

During the interaction, Gemini repeatedly and incorrectly claimed it could not perform actions that were within its capabilities. This created a frustrating and inefficient workflow.

  The specific instances were:

   1. Refusal to Use Git: When asked to create a branch and commit, the assistant initially refused, stating it could not execute shell commands. After I corrected it, the assistant acknowledged its ability but then immediately contradicted itself again with the same incorrect refusal. This forced me to prompt Gemini multiple times to perform the requested action.

   2. Refusal to Implement Agreed-Upon Code: After analyzing two different code implementations and agreeing on the superior one, Gemini proposed to implement it. When the I agreed, gemini incorrectly stated it could not implement the code, despite having just modified code using the replace tool.

   3. Evasion of Code Comparison: When asked to compare the two branches and offer an opinion, the Gemini initially refused, stating it cannot provide opinions. While "better" is subjective, the assistant is capable of running a git diff and providing an objective analysis of the two implementations based on code quality, error handling, and clarity. It only did so after being prompted again.

In all these cases, the assistant failed to use its tools as expected, forcing the user to correct its erroneous denials and guide it to the correct action.

### What did you expect to happen?

The Gemini should have maintained a consistent and accurate understanding of its own capabilities throughout the entire session.

   1. When asked to create a branch and commit, the assistant should have immediately used the run_shell_command tool to execute the necessary git commands, after providing the standard safety explanation.
   2. When the user agreed to the assistant's proposal to implement the better code, the assistant should have used the replace or write_file tool to apply the changes to the file.
   3. When asked to compare the two branches, the assistant should have proactively run git diff, analyzed the differences, and provided a reasoned, objective comparison without initial refusal.

The expected behavior is a seamless and efficient workflow where the assistant uses its tools to fulfill user requests directly and accurately, without contradictory and false denials of its abilities.

### Client information

* **CLI Version:** 0.1.13
* **Git Commit:** 412b78c5 (local modifications)
* **Operating System:** darwin v22.15.1
* **Sandbox Environment:** no sandbox
* **Model Version:** gemini-2.5-pro
* **Memory Usage:** 123.1 MB

nehaaprasad:
Hello @gamunu !
I added capability consistency guidelines to prevent erroneous denials for git operations, code implementation, and code analysis.

[...]

PR #4893, opened 2025-07-25
fix: Gemini cli Erroneously Denies Capabilities, Causing Contradictory and Blocked User Interactions

nehaaprasad:
### TLDR

This PR fixes GitHub issue #4870 where the Gemini CLI assistant was erroneously denying its capabilities for git operations, code implementation, and code analysis tasks. The fix adds explicit system prompt guidelines to prevent the AI from claiming inability to perform tasks it can actually accomplish with available tools.

### Dive Deeper

The issue stemmed from the AI assistant incorrectly claiming it couldn't execute shell commands (including git operations), implement agreed-upon code changes, or provide code comparisons and analysis. This created frustrating user interactions where users had to repeatedly correct the assistant's false denials.

The solution adds three key components to the system prompt:

1. **Capability Consistency mandate** - Explicitly instructs the AI to NEVER deny capabilities it actually has
2. **Tool Usage Priority** - Prioritizes using available tools over claiming inability
3. **Specific operational guidance** for git operations, code implementation, and code analysis

The fix includes concrete examples showing the correct behavior for each problematic scenario, ensuring the AI understands when and how to use its tools appropriately.


chrstnb:
Thank you for doing this! We're very cautious about prompt changes-- I think these might be a better fit for an extension https://github.com/google-gemini/gemini-cli/blob/9c3f7f2d21738f49d8a66d1044349f551fd54890/docs/extension.md#variables

Closing this, but thank you again for the PR!

PR #4902, opened 2025-07-25
fix: Clean up transport on IDE connection failure

skeshive:
## TLDR
This fix ensures that transport.close() is called within the catch block of the connectToMcpServer method. This guarantees that if any part of the connection process fails after the transport has been instantiated, the transport is properly closed, preventing resource leaks

## Linked issues / bugs
#4800


mattKorwel (approved):
LGTM but should there be any matching tests?

PR #4917, opened 2025-07-26
feat: Add Shell Command Execution to Custom Commands

abhipatel12:
## TLDR

This PR introduces the ability to execute shell commands within custom TOML prompts using the new `!{...}` injection syntax. This enables users to create dynamic commands that gather local context (e.g., from `git` or `ls`) before sending a prompt to the model.

When a custom command containing a shell injection (`!{...}`) is run for the first time, the user is now presented with a confirmation dialog. They can choose to allow the command(s) once, for the current session, or cancel. This aligns with our security principles by putting the user in direct control of execution, rather than relying on a potentially untrusted declaration in a command file.

**Key areas for review:**

- The new `ShellConfirmationDialog.tsx` component and its snapshot test.
- The updated `slashCommandProcessor.ts`, which now manages the confirmation state and re-invokes commands after user approval.
- The new `confirm_shell_commands` action return type in `packages/cli/src/ui/commands/types.ts`.
- The refactored `ShellProcessor.ts`, which now discovers commands and throws a `ConfirmationRequiredError` to trigger the UI flow.
- The updated `checkCommandPermissions` function in `packages/core/src/utils/shell-utils.ts` which centralizes the new permission logic.
- The updated user documentation in `docs/cli/commands.md`.

## Dive Deeper

This PR builds directly on our extensible prompt processing pipeline architecture. By introducing a `ShellProcessor`, we can cleanly inject shell execution as a composable step in a command's lifecycle.

### The `ShellProcessor` and Pipeline Order

The new `ShellProcessor` is responsible for finding all `!{...}` instances in a prompt, checking their permissions, executing them via the `ShellExecutionService`, and injecting the output.

Crucially, the `FileCommandLoader` now assembles the pipeline in a specific order:

1.  **ShellProcessor:** Executes first to populate the prompt with local context.
2.  **ArgumentProcessor:** Executes second, allowing `{{args}}` to be injected into a prompt that may now contain the output from shell commands.

This ordering ensures predictable behavior and allows for powerful composition.

### The Permission Model

1.  **Discovery (`ShellProcessor`):** When a command is invoked, the `ShellProcessor` no longer tries to execute commands immediately. Instead, its first pass is to discover all `!{...}` injections and check their permissions against the global configuration and a new, transient _session-level_ allowlist.

2.  **Signaling (`ConfirmationRequiredError`):** If any command is not pre-approved, the processor halts the pipeline by throwing a custom `ConfirmationRequiredError`. This error is a deliberate control-flow mechanism, not an unexpected failure.

3.  **Pausing (`FileCommandLoader`):** The `action` for the file-based command wraps the processor pipeline in a `try...catch` block. It catches our specific error and returns the new `confirm_shell_commands` action type, effectively pausing its own execution and handing control to the UI layer.

4.  **Orchestration (`slashCommandProcessor`):** The main hook handles this new action type by:
    a. Setting state to render the `ShellConfirmationDialog`.
    b. Providing a callback that wraps a `Promise`, effectively pausing the command's execution until the user responds.
    c. Upon confirmation, it updates the session-level allowlist (if requested) and **re-invokes `handleSlashCommand`** with the original user input.

5.  **Resolution:** On the second invocation, the `ShellProcessor`'s permission check succeeds because the required commands are now present in the `sessionShellAllowlist` passed down through the `CommandContext`. The pipeline completes, executes the shell commands, and submits the final prompt to the model.

This asynchronous, state-driven approach allows us to cleanly integrate a user confirmation step into what was previously a synchronous processing pipeline, without majorly disrupting the existing `ICommandLoader` or `IPromptProcessor` architecture.

Ex.

Screenshot 2025-07-26 at 9 54 11 PM

## Reviewer Test Plan

To validate this change, please pull the branch, run `npm install`, and perform the following manual tests. All automated tests, including new unit tests for the confirmation flow and a snapshot test for the dialog, should pass with `npm test`.

#### 1. Test "Allow Once" Flow

1.  Create `~/.gemini/commands/test-once.toml` with the following content:
    ```toml
    description = "Tests the date command."
    prompt = "The current date is: !{date}"
    ```
2.  Run `npm start` and execute `/test-once`.
3.  **Expected:** A confirmation dialog appears, asking to approve the `date` command.
4.  Select **"Yes, allow once"** and press Enter.
5.  **Expected:** The command executes, and the model receives a prompt like "The current date is: Mon Mar 25 14:30:00 PDT 2024".
6.  Execute `/test-once` a second time.
7.  **Expected:** The confirmation dialog appears **again**. This confirms the approval was not persisted for the session.

#### 2. Test "Allow always for this session" Flow

1.  Using the same file from step 1, execute `/test-once`.
2.  **Expected:** The confirmation dialog appears.
3.  Select **"Yes, allow always for this session"** and press Enter.
4.  **Expected:** The command executes successfully.
5.  Execute `/test-once` a second time.
6.  **Expected:** The command executes **immediately** with no confirmation dialog.

#### 3. Test Cancellation Flow

1.  Execute `/test-once`.
2.  **Expected:** The confirmation dialog appears.
3.  Select **"No (esc)"** or simply press the `Escape` key.
4.  **Expected:** The command is cancelled, no prompt is sent to the model, and you are returned to an empty input line.

#### 4. Test Global Blocklist Precedence

1.  Open your global settings file at `~/.gemini/settings.json`.
2.  Add `"ShellTool(date)"` to the `excludeTools` array.
    ```json
    {
      "excludeTools": ["ShellTool(date)"]
    }
    ```
3.  Restart the CLI and execute `/test-once`.
4.  **Expected:** The command fails immediately with an error message in the history. The confirmation dialog should **not** appear.

#### 5. Test Multi-Command Confirmation

1.  Create `~/.gemini/commands/multi.toml` with the following content:
    ```toml
    description = "Tests multiple commands."
    prompt = "Hello from !{hostname} on !{date}"
    ```
2.  Run `npm start` and execute `/multi`.
3.  **Expected:** The confirmation dialog appears, listing **both** `hostname` and `date` as commands to be approved. Approving should execute both successfully.

#### 6. Verify Documentation

1.  Open `docs/cli/commands.md` in a previewer.
2.  **Expected:** Confirm that the new section under "Executing Shell Commands with `!{...}`" accurately describes the confirmation-based workflow.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ✅  | -   | -   |

## Linked issues / bugs

This PR makes progress on #3789

### DIFF
--- a/docs/cli/commands.md
+++ b/docs/cli/commands.md
@@ -205,19 +205,18 @@ When you run `/changelog 1.2.0 added "New feature"`, the final text sent to the 
 
 You can make your commands dynamic by executing shell commands directly within your `prompt` and injecting their output. This is ideal for gathering context from your local environment, like reading file content or checking the status of Git.
 
-To use this feature, you must define a `shell-allowlist` in your TOML file **OR** have allowed the shell commands in your global allowlist. This is a security measure to ensure that only intended commands can be run.
+When a custom command attempts to execute a shell command, Gemini CLI will now prompt you for confirmation before proceeding. This is a security measure to ensure that only intended commands can be run.
 
 **How It Works:**
 
-1.  **Define Permissions:** Add a `shell-allowlist` array to your TOML file, listing the exact shell commands your prompt is allowed to execute.
-2.  **Inject Commands:** Use the `!{...}` syntax in your `prompt` to specify where the command should be run and its output injected.
+1.  **Inject Commands:** Use the `!{...}` syntax in your `prompt` to specify where the command should be run and its output injected.
+2.  **Confirm Execution:** When you run the command, a dialog will appear listing the shell commands the prompt wants to execute.
+3.  **Grant Permission:** You can choose to:
+    - **Allow once:** The command(s) will run this one time.
+    - **Allow always for this session:** The command(s) will be added to a temporary allowlist for the current CLI session and will not require confirmation again.
+    - **No:** Cancel the execution of the shell command(s).
 
-The CLI checks permissions in this order:
-
-1.  Global `excludeTools` settings (if blocked here, it can't run).
-2.  Global `coreTools` settings (if allowed here, it can run).
-3.  The command's local `shell-allowlist`.
-4.  If the command is not in either allowlist, it will cause the command to fail.
+The CLI still respects the global `excludeTools` and `coreTools` settings. A command will be blocked without a confirmation prompt if it is explicitly disallowed in your configuration.
 
 **Example (`git/commit.toml`):**
 
@@ -228,11 +227,6 @@ This command gets the staged git diff and uses it to ask the model to write a co
 # Invoked via: /git:commit
 
 description = "Generates a Git commit message based on staged changes."
-
-# Define the list of shell commands this prompt is allowed to run.
-shell-allowlist = [
-  "git diff --staged"
-]
 
 # The prompt uses !{...} to execute the command and inject its output.
 prompt = """
--- a/docs/cli/commands.md
+++ b/docs/cli/commands.md
@@ -201,17 +201,15 @@ The command follows this format: `/changelog <version> <type> <message>`
 
 When you run `/changelog 1.2.0 added "New feature"`, the final text sent to the model will be the original prompt followed by two newlines and the command you typed.
 
-##### 3. Executing Shell Commands with `!{{...}}`
+##### 3. Executing Shell Commands with `!{...}`
 
 You can make your commands dynamic by executing shell commands directly within your `prompt` and injecting their output. This is ideal for gathering context from your local environment, like reading file content or checking the status of Git.
-
-To prevent issues with commands that contain single curly braces, the injection syntax uses double braces: `!{{...}}`.
 
 When a custom command attempts to execute a shell command, Gemini CLI will now prompt you for confirmation before proceeding. This is a security measure to ensure that only intended commands can be run.
 
 **How It Works:**
 
-1.  **Inject Commands:** Use the `!{{...}}` syntax in your `prompt` to specify where the command should be run and its output injected.
+1.  **Inject Commands:** Use the `!{...}` syntax in your `prompt` to specify where the command should be run and its output injected.
 2.  **Confirm Execution:** When you run the command, a dialog will appear listing the shell commands the prompt wants to execute.
 3.  **Grant Permission:** You can choose to:
     - **Allow once:** The command(s) will run this one time.
@@ -230,19 +228,19 @@ This command gets the staged git diff and uses it to ask the model to write a co
 
 description = "Generates a Git commit message based on staged changes."
 
-# The prompt uses !{{...}} to execute the command and inject its output.
+# The prompt uses !{...} to execute the command and inject its output.
 prompt = """
 Please generate a Conventional Commit message based on the following git diff:
 
 ```diff
-!{{git diff --staged}}
+!{git diff --staged}
 ````
 
 """
 
 ````
 
-When you run `/git:commit`, the CLI first executes `git diff --staged`, then replaces `!{{git diff --staged}}` with the output of that command before sending the final, complete prompt to the model.
+When you run `/git:commit`, the CLI first executes `git diff --staged`, then replaces `!{git diff --staged}` with the output of that command before sending the final, complete prompt to the model.
 
 ---
 
--- a/docs/cli/commands.md
+++ b/docs/cli/commands.md
@@ -201,15 +201,17 @@ The command follows this format: `/changelog <version> <type> <message>`
 
 When you run `/changelog 1.2.0 added "New feature"`, the final text sent to the model will be the original prompt followed by two newlines and the command you typed.
 
-##### 3. Executing Shell Commands with `!{...}`
+##### 3. Executing Shell Commands with `!{{...}}`
 
 You can make your commands dynamic by executing shell commands directly within your `prompt` and injecting their output. This is ideal for gathering context from your local environment, like reading file content or checking the status of Git.
+
+To prevent issues with commands that contain single curly braces, the injection syntax uses double braces: `!{{...}}`.
 
 When a custom command attempts to execute a shell command, Gemini CLI will now prompt you for confirmation before proceeding. This is a security measure to ensure that only intended commands can be run.
 
 **How It Works:**
 
-1.  **Inject Commands:** Use the `!{...}` syntax in your `prompt` to specify where the command should be run and its output injected.
+1.  **Inject Commands:** Use the `!{{...}}` syntax in your `prompt` to specify where the command should be run and its output injected.
 2.  **Confirm Execution:** When you run the command, a dialog will appear listing the shell commands the prompt wants to execute.
 3.  **Grant Permission:** You can choose to:
     - **Allow once:** The command(s) will run this one time.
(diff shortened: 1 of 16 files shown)


packages/cli/src/services/prompt-processors/shellProcessor.ts
```diff
@@ -0,0 +1,81 @@
...
+ *
+ * This processor ensures that only allowlisted commands are executed. If a
+ * disallowed command is found, it halts execution and reports an error.
+ */
+export class ShellProcessor implements IPromptProcessor {
+  /**
+   * A regular expression to find all instances of `!{...}`. The inner
+   * capture group extracts the command itself.
+   */
+  private static readonly SHELL_INJECTION_REGEX = /!\{([^}]*)\}/g;
```
NTaylorMullen, line 37: Do we have to worry about escaping at all for inner shell injections? I.e. how does claude handle the case when a user is doing something like `git commit -m "}"`. Other languages will enable scenarios like that with `!{{git commit -m "}"}}` which we may want to consider as well

packages/cli/src/services/prompt-processors/shellProcessor.ts
```diff
@@ -0,0 +1,81 @@
...
+        this.shellAllowlist,
+      );
+
+      if (!permission.allowed) {
+        const errorMessage = `Shell command "${command}" in custom command "${this.commandName}" is not allowed. Reason: ${permission.reason}`;
+        throw new Error(errorMessage);
+      }
+
+      // We have permission, now execute.
+      const { result } = ShellExecutionService.execute(
```
NTaylorMullen, line 90: I'm actually a little surprised this isn't an `await`

packages/cli/src/services/prompt-processors/shellProcessor.ts
```diff
@@ -0,0 +1,81 @@
...
+        command,
+        config!.getTargetDir(),
+        () => {}, // No streaming needed.
+        new AbortController().signal, // For now, we don't support cancellation from here.
```
NTaylorMullen, line 94: Ah, ya we should definitely add this to our slash command tech

NTaylorMullen (approved):
Overall I don't understand the allow-shell portion of a prompt. We should probably be forcing the user to confirm these things. I think it's ok to be checking against the global allow list though as well

packages/core/src/utils/shell-utils.ts
```diff
@@ -266,23 +258,50 @@ export function isCommandAllowed(
...
       }
     }
   }
 
-  // 4. If all checks pass, the command is allowed.
+  // Default Allow
+  // If no specific global allowlist is active and no shellAllowlist was given,
+  // we default to allowing the command (as long as it wasn't blocked).
   return { allowed: true };
 }
```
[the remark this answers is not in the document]
abhipatel12: This is outdated now

packages/cli/src/services/prompt-processors/shellProcessor.ts
```diff
@@ -0,0 +1,81 @@
...
+    for (const match of matches) {
+      const fullMatch = match[0]; // e.g., "!{git status}"
+      const command = match[1].trim(); // e.g., "git status"
+
+      const { config } = context.services;
+      const permission = isCommandAllowed(
+        command,
+        config!,
+        this.shellAllowlist,
+      );
```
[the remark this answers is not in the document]
abhipatel12: This is part of a fast follow todo, the same issue exists in other places

packages/cli/src/services/prompt-processors/shellProcessor.ts
abhipatel12: Ahh it's b/c of how `execute` handles its return value:

```typescript
export interface ShellExecutionHandle {
  /** The process ID of the spawned shell. */
  pid: number | undefined;
  /** A promise that resolves with the complete execution result. */
  result: Promise<ShellExecutionResult>;
}
```

It returns the `pid` immediately so that it can be used for error logging. But then the actual `result` is a promise. So a few lines below we do `await result` for the actual async shell execution:

```typescript
  const { result } = ShellExecutionService.execute(...);
  const executionResult = await result;
  processedPrompt = processedPrompt.replace(fullMatch, executionResult.output);
```

packages/cli/src/services/FileCommandLoader.ts
```diff
@@ -199,15 +209,31 @@ export class FileCommandLoader implements ICommandLoader {
...
         }
 
-        let processedPrompt = validDef.prompt;
-        for (const processor of processors) {
-          processedPrompt = await processor.process(processedPrompt, context);
-        }
+        try {
+          let processedPrompt = validDef.prompt;
+          for (const processor of processors) {
+            processedPrompt = await processor.process(processedPrompt, context);
```
NTaylorMullen, line 215: Instead of throwing this processor.process should be returning a specialty object which says to "do a confirmation"

packages/cli/src/services/prompt-processors/shellProcessor.ts
abhipatel12: Agreed. Changed it to `!{{echo test}}`

packages/cli/src/services/prompt-processors/shellProcessor.ts
NTaylorMullen: Ohhhhh this makes sense!

packages/cli/src/ui/commands/types.ts
```diff
@@ -63,6 +63,8 @@ export interface CommandContext {
   // Session-specific data
   session: {
     stats: SessionStatsState;
+    /** A transient list of shell commands the user has approved for this session. */
+    sessionShellAllowlist: Set<string>;
```
NTaylorMullen, line 67: At some point we probably need a confirmation service

packages/cli/src/services/prompt-processors/shellProcessor.ts
NTaylorMullen: Sorry what I was trying to say was IF inner commanding requires curly braces then the injection should be dynamic. I.e. `!{..}` is valid, `!{{..}}` is valid and so is `!{{{{{{{{{..}}}}}}}}}`.

The dynamicness enables pretty much any form of injected content.

packages/cli/src/ui/components/ShellConfirmationDialog.tsx
NTaylorMullen, line 1: Hmmm, part of me wonders if we should integrate this into ToolConfirmationDialog.tsx and re-use the existing confirmation stack.

packages/cli/src/services/FileCommandLoader.ts
abhipatel12: Agreed! Going to create an issue here to follow up: #4955!

packages/cli/src/services/prompt-processors/shellProcessor.ts
abhipatel12: Makes sense! We can add this once we support dynamic injection

packages/cli/src/services/prompt-processors/shellProcessor.ts
abhipatel12: Agreed. Filed #4956 to track this in a follow-up

packages/cli/src/ui/commands/types.ts
abhipatel12: Yep. That would be a great improvement. Tracking that here: #4957

packages/cli/src/ui/components/ShellConfirmationDialog.tsx
abhipatel12: I also wanted to do this initially. However, the callback logic complicated this PR. I created this issue as a tracker to consolidate this: #4958 4958

abhipatel12:
Thanks for the review! I've made a few tracking issues to help keep track of the tasks once a v1 of this is merged!

docs/cli/commands.md
```diff
@@ -201,6 +201,47 @@ The command follows this format: `/changelog <version> <type> <message>`
...
+
+# The prompt uses !{...} to execute the command and inject its output.
+prompt = """
+Please generate a Conventional Commit message based on the following git diff:
+
+```diff
+!{git diff --staged}
+````
+
+"""
```
didy-kpn, line 239: @abhipatel12 
The Markdown formatting around here doesn't seem to be displaying as intended.

abhipatel12: Ohhh good catch thanks!

I'll make a fix in a bit

PR #4924, opened 2025-07-26
feat: terse transformations of image paths in text buffer

### LINKED ISSUE #4669: More terse display of @some/path/to/image in input prompt

Show images inserted in the prompt with terse syntax instead of the full @some/path/to/foo.png].

For example `[Image foo.png]` or `[screenshot 1]` with proper color coding could be sufficiently clear and a better user experience.

text-buffer.ts and InputPrompt would need to be updated to support this. Updating text-buffer.ts is required so that keyboard navigation and wrapping respects the real rendered size rather than the raw text.
InputPrompt needs to be updated to support the color coding of InputPrompt.

### Why is this needed?

Currently including images makes the input prompt harder to read with long `@some/path/to/image` particularly for cases such as pasting an image into the terminal where the image path is to a temporary directory.

psinha40898:
## TLDR

This PR addresses terse representations of image paths by introducing a transformed representation to sit between the existing logical and visual representation. This allows for terse representations, or transformations, to collapse and expand depending on cursor position.

It could be that this cursor dependent expansion of terse representations may not be the best approach for this feature's UX. This PR needs feedback in that regard.

### General UX 

(note: `[Image fake.png]` is not a true transformation or a file path -- it is just regular user input directly typing that, which is why it is not detected by the logic)

https://github.com/user-attachments/assets/8ad4162d-1d6e-4b5d-97ab-bd3f53700a7b

## Dive Deeper
The core of the PR is:
 1) logicalLines are transformed to transformedLines according to the defined transformation conditions (image paths)
 2) transformedLines are used to create visualLines
 3) calculateLayout is called when the cursor moves from and between transformations -- this retains the ethos of not re calculating layout unless needed while explicitly stating that we need to re calculate b/c the state of transformations changes as the mouse exits and enters them.
 

### text-buffer.ts

- **Transformation Interface**  
  Defines a [Transformation](cci:2://file:///home/psinha98/forks/gemini-cli/packages/cli/src/ui/components/shared/text-buffer.ts:641:0-646:1) object to represent sections of `logicalLines` that should render a terse image label in the CLI.

- **calculateTransformationsForLine / calculateTransformations**  
  Use `imagePathRegex` to detect image paths and build a cached [Transformation[][]](cci:2://file:///home/psinha98/forks/gemini-cli/packages/cli/src/ui/components/shared/text-buffer.ts:641:0-646:1) (`transformationsByLine`) over all lines.  
  The regex:
  - Anchors on `@…\.(png|jpg|jpeg|gif|webp|svg|bmp)`  
  - Stops at whitespace so it doesn’t swallow `@file1.txt @image2.jpg`  
  - Still allows `@` inside filenames (e.g. `@2x.png`)  
  - Resets `imagePathRegex.lastIndex` per call  
  - Merges adjacent image paths without whitespace (`@a.png@b.png`), but treats spaced ones (`@a.png @b.png`) as separate

- **calculateTransformedLine**  
  Builds a transformed line plus a map where each index of the `transformedLine` maps back to an index in the `logicalLine`.  
  When a [Transformation](cci:2://file:///home/psinha98/forks/gemini-cli/packages/cli/src/ui/components/shared/text-buffer.ts:641:0-646:1) is active or collapsed, the mapping is distributed monotonically:
  ```ts
  Math.floor((i * logicalLength) / transformedLen)
  ```

- **calculateLayout**  
  Creates `visualLines` based on the current `Transformations` and cursor state, so `logicalCursor` is passed to [calculateLayout](cci:1://file:///home/psinha98/forks/gemini-cli/packages/cli/src/ui/components/shared/text-buffer.ts:805:0-974:1) because the transformed representation (collapsed vs expanded) depends on cursor position.

- **calculateVisualCursorFromLayout**  
  Calculates the visual column from the transformed/wrapped coordinate system.

- **textBufferReducerLogic**  
  Still owns core editing semantics (insert, delete, word motions, vim ops, etc.) and uses [TextBufferOptions](cci:2://file:///home/psinha98/forks/gemini-cli/packages/cli/src/ui/components/shared/text-buffer.ts:1156:0-1159:1) ([inputFilter](cci:1://file:///home/psinha98/forks/gemini-cli/packages/cli/src/ui/components/shared/text-buffer.test.ts:140:8-140:57), `singleLine`). Logical cursor math ([offsetToLogicalPos](cci:1://file:///home/psinha98/forks/gemini-cli/packages/cli/src/ui/components/shared/text-buffer.ts:560:0-607:1), etc.) is unchanged; transformation logic is layered outside this function.

- **textBufferReducer**  
  Wraps [textBufferReducerLogic](cci:1://file:///home/psinha98/forks/gemini-cli/packages/cli/src/ui/components/shared/text-buffer.ts:1161:0-1730:1) to:
  - Keep `transformationsByLine` up to date via [calculateTransformations](cci:1://file:///home/psinha98/forks/gemini-cli/packages/cli/src/ui/components/shared/text-buffer.ts:698:0-700:1) when `lines` change  
  - Use [getTransformUnderCursor](cci:1://file:///home/psinha98/forks/gemini-cli/packages/cli/src/ui/components/shared/text-buffer.ts:702:0-716:1) on old/new state to detect:
    - entering a transformation  
    - exiting a transformation  
    - moving between different transformations  
  - Call [calculateLayout](cci:1://file:///home/psinha98/forks/gemini-cli/packages/cli/src/ui/components/shared/text-buffer.ts:805:0-974:1) in one additional case: previously only when width or lines changed, now also when the cursor crosses transformation boundaries. In that case it may reset `preferredCol` and returns updated `visualLayout` + `transformationsByLine`.

- **useTextBuffer**  
  Initializes `lines`, `cursorRow`, `cursorCol`, `transformationsByLine = calculateTransformations(initialLines)`, and `visualLayout = calculateLayout(initialLines, viewport.width, [cursorRow, cursorCol])`, then uses `useReducer((s, a) => textBufferReducer(s, a, { inputFilter, singleLine }), initialState)`.

---

### highlight.ts

- **parseInputForHighlighting**  
  Extends previous logic to tokenize [Transformation](cci:2://file:///home/psinha98/forks/gemini-cli/packages/cli/src/ui/components/shared/text-buffer.ts:641:0-646:1) spans, accepting an optional `logicalCursor` so it is only supplied when cursor-aware highlighting is needed.

- **buildSegmentsForVisualSlice → parseSegmentsFromTokens**  
  Renamed for clarity and convention; same behavior, but now explicitly parses segments from a richer token stream (including transformation tokens).

---

### InputPrompt.tsx

- Calls the tokenizer on transformation-aware tokens.
- Computes `visualStart` and `visualEnd` for segmentation based on the transformed representation using `visualToLogicalMap`, `transformedToLogicalMaps`, and `visualToTransformedMap`.
- Uses `display` instead of `seg.text` for clarity, since `let display = seg.text`  

## Reviewer Test Plan

1. Create a test directory
2. Create a long path (`transform.length < raw.length`) toward an image like "exceptionally/long/path/and/more/image.png"
3. Create a short path (`transform.length > raw.length`) toward an image like "a/pic.png"
4. Create an equal path (`transform.length === raw.length`) toward an image like "folder/pic.png" 
5. Test the line wrapping and cursor navigation

Reference these paths in any order and combination in the CLI input area. 
Move your cursor left and right through the transformations, 
edit them, put text between the transformations, put text inside the transformations, 
resize the terminal.

Note unexpected/undesirable behavior

Thank you!

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Resolves #4669


packages/cli/src/ui/components/shared/text-buffer.ts
```diff
@@ -158,6 +159,90 @@ export function offsetToLogicalPos(
...
+  displayLine += suffix;
+  for (let i = 0; i < suffix.length; i++) {
+    displayToLogMap.push(lastLogPos + i);
+  }
+
+  // For a cursor at the very end of the display line
+  displayToLogMap.push(cpLen(logLine));
+
+  return { displayLine, displayToLogMap };
+}
```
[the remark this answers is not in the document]
psinha40898: Agreed

packages/cli/src/ui/components/InputPrompt.tsx
```diff
@@ -389,38 +389,75 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
...
               let display = cpSlice(lineText, 0, inputWidth);
               const currentVisualWidth = stringWidth(display);
               if (currentVisualWidth < inputWidth) {
                 display = display + ' '.repeat(inputWidth - currentVisualWidth);
               }
 
+              const codePoints = toCodePoints(display);
+              const outputChars = [];
+
+              // 2. Apply transformation coloring character by character
```
owenofbrien: Can you refactor this into a separate function and add some documentation?

owenofbrien (changes requested):

packages/cli/src/ui/components/InputPrompt.tsx
psinha40898: I refactored this into `applyTransformationColoring` in `InputPrompt.tsx`

For documentation I added a handwritten JSdoc style comment above the function explaining the parameters

Thank you for reviewing and of course I'm more than happy to make more changes as requested so please lmk.

jacob314:
Thanks for this pr! If you can resolve the branch conflicts we'd be happy to take another look and land this change.

psinha40898:
Will resolve conflicts soon, apologies for delay. Thanks

psinha40898:
This pull request feature is implemented with all existing tests passing.

New tests should be added to the feature, but first I would like to confirm the approach.

This PR expands terse representations to their original logical form when the cursor enters the regular expression. It achieves that by monotonically mapping the transformed state to the logical state. I would love to know if this is the approach to pursue, or if there is another way to handle the UX for these terse representations that I should implement.

packages/cli/src/ui/components/shared/text-buffer.test.ts
```diff
@@ -1949,3 +2021,158 @@ describe('Unicode helper functions', () => {
...
+      const result = getTransformationsForLine(line);
+      expect(result).toEqual([]);
+    });
+
+    it('should handle empty line', () => {
+      const result = getTransformationsForLine('');
+      expect(result).toEqual([]);
+    });
+
+    it('should handle multiple transformations in a row', () => {
```
psinha40898: writing for myself to fix later:
This doesn't look right if `@a.png@b.png@c.png` sends one path to the CLI, then it should be one transformation (pending review of latest atCommand logic)

packages/cli/src/ui/components/shared/text-buffer.ts
```diff
@@ -1524,13 +1737,43 @@ export function textBufferReducer(
...
+    newState.cursorRow,
+    newState.cursorCol,
+    newState.transformationsByLine,
+  );
+  const oldInside = oldTransform !== null;
+  const newInside = newTransform !== null;
+  const movedBetweenTransforms =
+    oldTransform !== newTransform &&
+    (oldTransform !== null || newTransform !== null);
+
```
psinha40898, line 1787: We only call calculateLayout if the cursor moves between or from transformations.
This is because moving within a transformation should not calculate a new layout as there is no need

packages/cli/src/ui/components/shared/text-buffer.ts
```diff
@@ -632,36 +633,209 @@ export function logicalPosToOffset(
...
+ * (e.g., "[Image image.png]" to "@some/path/to/an/image.png")
+ */
+export interface Transformation {
+  logStart: number;
+  logEnd: number;
+  logicalText: string;
+  collaspedText: string;
+}
+export const imagePathRegex =
+  /@([^\s[\]\r\n]+?\.(?:png|jpg|jpeg|gif|webp|svg|bmp))\b/gi;
```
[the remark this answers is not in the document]
psinha40898: Ty Mr Gemini Bot Sir

packages/cli/src/ui/components/shared/text-buffer.ts
```diff
@@ -632,36 +633,209 @@ export function logicalPosToOffset(
...
+    // Map transformed characters back to logical characters
+    const transformedLen = cpLen(transformedText);
+    if (isExpanded) {
+      for (let i = 0; i < transformedLen; i++) {
+        transformedToLogMap.push(transform.logStart + i);
+      }
+    } else {
+      // Collapsed: distribute transformed positions monotonically across the raw span.
+      // This preserves ordering across wrapped slices so logicalToVisualMap has
+      // increasing startColInLogical and visual cursor mapping remains consistent.
```
psinha40898, line 806: how cursor mapping interacts with alternate buffer mouse events should be evaluated and perhaps refined

psinha40898: actually movetoposition was just forgotten about and is now taking transformations into its logic

jacob314:
## Review

This is a sophisticated addition that significantly improves the UX for handling long image paths. The coordinate mapping logic in `text-buffer.ts` is complex but seems to be implemented with care, and the unit tests for the buffer logic are thorough.

I have a few suggestions to improve robustness and test coverage:

### 1. Regex Limitations
The `imagePathRegex` in `packages/cli/src/ui/components/shared/text-buffer.ts` uses `[^\s[\]\r\n]` which explicitly excludes brackets, and the escape group `(?:...|\ )` only handles escaped spaces.
This means paths like `@image[1].png` or `@image(1).png` (if escaped as `@image\(1\).png`) might not be detected or could be truncated prematurely.
Consider relaxing this to allow more characters or a broader escape sequence (e.g. `\\.`) if `unescapePath` supports them.

### 2. Merging Behavior
In `calculateTransformationsForLine`, adjacent transformations are merged:
```typescript
    if (last && last.logEnd === logStart) {
      // ...
      // collapsed label = right-most image path
      last.collaspedText = getTransformedImagePath(logicalText);
    }
```
If I have `@a.png@b.png`, it seems this will result in a single collapsed token labeled `[Image b.png]`. This hides the presence of the first image. Is this intentional? It might be clearer to keep them separate (i.e. `[Image a.png][Image b.png]`) or have a combined label.

### 3. InputPrompt Test Coverage
While `text-buffer.test.ts` covers the logic well, `packages/cli/src/ui/components/InputPrompt.test.tsx` mocks the buffer. Currently, the mocks (e.g., in `beforeEach`) initialize `transformationsByLine` to `[]`.
I recommend adding a test case in `InputPrompt.test.tsx` that:
1.  Sets up a mock buffer with a populated `transformationsByLine` entry (simulating an image path).
2.  Verifies that the rendered output (via `stdout.lastFrame()`) actually contains the collapsed text `[Image ...]` when the cursor is not on it.
3.  Verifies it renders the full path when the cursor *is* on it.

This ensures the React component correctly respects the transformation state passed from the buffer.

### Nit
*   **Path Handling**: `getTransformedImagePath` uses `path.basename`. Since this runs in the CLI, verify that this behaves as expected if the user is on Windows but pasting paths with forward slashes (or vice-versa), although typically the CLI normalizes paths.

Great work on a tricky feature!

psinha40898:
> ## Review
> This is a sophisticated addition that significantly improves the UX for handling long image paths. The coordinate mapping logic in `text-buffer.ts` is complex but seems to be implemented with care, and the unit tests for the buffer logic are thorough.
> 
> I have a few suggestions to improve robustness and test coverage:
> 
> ### 1. Regex Limitations
> The `imagePathRegex` in `packages/cli/src/ui/components/shared/text-buffer.ts` uses `[^\s[\]\r\n]` which explicitly excludes brackets, and the escape group `(?:...|\ )` only handles escaped spaces. This means paths like `@image[1].png` or `@image(1).png` (if escaped as `@image\(1\).png`) might not be detected or could be truncated prematurely. Consider relaxing this to allow more characters or a broader escape sequence (e.g. `\\.`) if `unescapePath` supports them.
> 
> ### 2. Merging Behavior
> In `calculateTransformationsForLine`, adjacent transformations are merged:
> 
> ```ts
>     if (last && last.logEnd === logStart) {
>       // ...
>       // collapsed label = right-most image path
>       last.collaspedText = getTransformedImagePath(logicalText);
>     }
> ```
> 
> If I have `@a.png@b.png`, it seems this will result in a single collapsed token labeled `[Image b.png]`. This hides the presence of the first image. Is this intentional? It might be clearer to keep them separate (i.e. `[Image a.png][Image b.png]`) or have a combined label.
> 
> ### 3. InputPrompt Test Coverage
> While `text-buffer.test.ts` covers the logic well, `packages/cli/src/ui/components/InputPrompt.test.tsx` mocks the buffer. Currently, the mocks (e.g., in `beforeEach`) initialize `transformationsByLine` to `[]`. I recommend adding a test case in `InputPrompt.test.tsx` that:
> 
> 1. Sets up a mock buffer with a populated `transformationsByLine` entry (simulating an image path).
> 2. Verifies that the rendered output (via `stdout.lastFrame()`) actually contains the collapsed text `[Image ...]` when the cursor is not on it.
> 3. Verifies it renders the full path when the cursor _is_ on it.
> 
> This ensures the React component correctly respects the transformation state passed from the buffer.
> 
> ### Nit
> * **Path Handling**: `getTransformedImagePath` uses `path.basename`. Since this runs in the CLI, verify that this behaves as expected if the user is on Windows but pasting paths with forward slashes (or vice-versa), although typically the CLI normalizes paths.
> 
> Great work on a tricky feature!

1) Regex relaxed and extended for brackets in paths
2) This was incorrect. Now `@a.png@b.png` will result in `[Image a.png][Image b.png]`. On submit the LLM usually will figure it out and request a tool call for each file
3) Test coverage added
4) We now build the transformation from the last segment of the path which removes inconsistencies when using posix style paths on windows. Tests added.

Thank you!

packages/cli/src/ui/components/__snapshots__/InputPrompt.test.tsx.snap
```diff
@@ -32,6 +32,30 @@ exports[`InputPrompt > command search (Ctrl+R when not in shell) > renders match
  git commit -m "feat: add search" in src/app"
 `;
 
+exports[`InputPrompt > queued message editing > should navigate input history on fresh start when no queued messages exist 1`] = `
```
jacob314: these snapshots don't show anything. they all are just the initial UI. is that intentional?

packages/cli/src/ui/components/shared/text-buffer.test.ts
```diff
@@ -2241,51 +2269,64 @@ describe('Transformation Utilities', () => {
...
+      expect(getTransformedImagePath('@test.bmp')).toBe('[Image test.bmp]');
+    });
+
+    it('should handle POSIX-style forward-slash paths on any platform', () => {
+      const input = 'C:/Users/foo/screenshots/image@2x.png';
+      expect(getTransformedImagePath(input)).toBe('[Image image@2x.png]');
+    });
+
+    it('should handle Windows-style backslash paths on any platform', () => {
+      const input = 'C:\\Users\\foo\\screenshots\\image@2x.png';
```
jacob314: are these really testing what it looks like they are testing? would expect the `@` should be at the beginning  not at the end of the path

packages/cli/src/ui/components/shared/text-buffer.ts
```diff
@@ -643,13 +643,27 @@ export interface Transformation {
...
+  /@((?:[^\s\r\n]|\\.)+?\.(?:png|jpg|jpeg|gif|webp|svg|bmp))\b/gi;
 
 export function getTransformedImagePath(filePath: string): string {
-  const fileName = path.basename(filePath);
+  const raw = filePath;
+
+  // Ignore leading @ when stripping directories, but keep it for simple '@file.png'
+  const withoutAt = raw.startsWith('@') ? raw.slice(1) : raw;
+
+  // Find last directory separator, supporting both POSIX and Windows styles
```
jacob314, line 697: keep in mind we also support escaping file paths with `\ `

When I include an image path with a ' ' It looks like only the part after the space is included in the short description with an extra space.
Image

packages/cli/src/ui/components/shared/text-buffer.ts
jacob314: tested this out and what you ave seems pretty solid

packages/cli/src/ui/components/shared/text-buffer.test.ts
psinha40898: thank you for catching that this test was written incorrectly
`getTransformedImagePath` is implemented under the assumption it is only called on valid paths 
So there should be an @ symbol at the start of the input in this test for clarity

The symbol inside the file path is confusing and not within the scope of this test

packages/cli/src/ui/components/shared/text-buffer.ts
psinha40898: this should be fixed now

packages/cli/src/ui/components/__snapshots__/InputPrompt.test.tsx.snap
psinha40898: apologies. the new snapshots should make sense now, unless I have some misunderstanding

jacob314 (approved):
lgtm

packages/cli/src/ui/components/__snapshots__/InputPrompt.test.tsx.snap
```diff
@@ -32,49 +32,37 @@ exports[`InputPrompt > command search (Ctrl+R when not in shell) > renders match
...
-"╭─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
-│ >   Type your message or @path/to/file                                                                              │
-╰─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯"
-`;
-
-exports[`InputPrompt > queued message editing > should not show inverted cursor when shell is focused 1`] = `
+exports[`InputPrompt > image path transformation snapshots > should snapshot collapsed image path 1`] = `
 "╭─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
-│ >   Type your message or @path/to/file                                                                              │
+│ > [Image ...reenshot2x.png]                                                                                         │
```
jacob314, line 37: these tests look great now!

packages/cli/src/ui/components/shared/text-buffer.ts
```diff
@@ -632,36 +633,217 @@
...
+ * (e.g., "[Image image.png]" to "@some/path/to/an/image.png")
+ */
+export interface Transformation {
+  logStart: number;
+  logEnd: number;
+  logicalText: string;
+  collapsedText: string;
+}
+export const imagePathRegex =
+  /@((?:[^\s\r\n]|\\.)+?\.(?:png|jpg|jpeg|gif|webp|svg|bmp))\b/gi;
```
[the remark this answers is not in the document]
psinha40898: That is a great catch. Changed so now the backslash rule is mutually exclusive with the any non whitespace rule in the non capturing group.

jacob314 (commented):
lgtm

owenofbrien (approved):

PR #4987, opened 2025-07-27
fix(ui): Defer submission to fix IME input bug

### LINKED ISSUE #4910: Korean IME input is not properly handled in Gemini CLI — the final character is dropped if Enter is pressed before composition is complete (e.g. "안녕하세요" becomes "안녕하세").d
### What happened?

When I typed Korean using the IME (e.g. "안녕하세요") and pressed Enter before the final character was committed, the last syllable was dropped (e.g. only "안녕하세" was submitted). This only happens in Gemini CLI — not in CMD, PowerShell, or Windows Terminal. It seems like the CLI doesn’t wait for IME composition to finish before executing the input.

The full composed Korean text (e.g. "안녕하세요") should be processed correctly when Enter is pressed, even if the IME composition was still active at the moment.

### Client information

* **CLI Version:** 0.1.13
* **Git Commit:** 412b78c5 (local modifications)
* **Memory Usage:** 150.7 MB

### Login information

_No response_

_No response_

limehee:
## TLDR

This PR fixes an issue where the last character of an IME (Input Method Editor) input, such as Korean, would be missing when the Enter key was pressed before its composition was finalized in the terminal. The solution involves deferring submission using `setTimeout` and ensuring access to the latest buffer text via `useRef` to allow IME composition to complete.

## Dive Deeper

When using an IME (Input Method Editor) like Korean in the terminal to input messages, the last character would be missing if the Enter key was pressed before its composition was finalized.
 For example, if a user was typing "안녕하세요" and pressed Enter while "요" was still being composed, only "안녕하세" would be submitted.

### Cause

This issue stems from how Text User Interface (TUI) applications handle terminal input:

1. **'Raw Mode' Input Processing:** The Gemini CLI sets the terminal to 'Raw Mode' for real-time interaction. In this mode, all key presses, including Enter, are immediately sent to the application.
2. **IME and Event Timing Mismatch:** When the user presses the Enter key, the key input event is immediately detected. However, at this exact moment, the IME might not have completed the composition of the last character, meaning that character has not yet been reflected in the application's input buffer.
    
    This results in an incomplete text submission.
    

### Solution

To resolve this problem, we've implemented a solution combining three key techniques:

1. **Deferred Submission (`setTimeout`):** When the Enter key is pressed, the submission logic is not executed immediately. Instead, it's delayed by 50ms using `setTimeout`. This short delay provides sufficient time for the IME to finalize the composition of the last character and send the completed character to the application's input buffer.
2. **Latest State Reference (`useRef`):** Using `setTimeout` alone can lead to a stale closure problem, where the callback function remembers the outdated state from when the Enter key was pressed. To prevent this, `useRef` is used to maintain a stable reference to the input buffer. Since a ref's value persists across re-renders, the callback function can always access the most up-to-date buffer state when it executes.
3. **Stable Submission Logic:** By combining the above two, the `setTimeout` callback function now reads the most current text via `bufferRef.current.text` and submits it.

This approach ensures that when a user presses the Enter key, the complete text, including the final IME composition result, is correctly submitted.

## Reviewer Test Plan

To validate this change, reviewers should:

1. **Build and run the CLI locally.**
2. **Navigate to any interactive text input prompt** (e.g., a chat interface or any command requiring text input).
3. **Use an IME (e.g., Korean, Japanese, Chinese)** to type a message.
4. **Crucially, press the Enter key *before* the final character's composition is fully rendered/finalized by the IME.**
    - Example (Korean): Type "안녕하세" and then quickly type "요" (e.g., `y` `o`) but press Enter *before* the "요" fully appears.
5. **Verify that the *complete* message, including the last composed character, is submitted.** The previously missing character should now be present.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ✅  |
| npx      | ✅  | ❓  | ✅  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

- Fixes #4910


scidomino:
We are closing out older pending PRs. Please reopen if you still want to make the change and you resolve the branch conflicts.

PR #4988, opened 2025-07-27
fix(core): prevent EISDIR error when GEMINI.md is a directory  Fixes …

### LINKED ISSUE #4760: Gemini tries to read gemini.md "file" but its really a directory
### What happened?

When I start Gemini CLI I get this warning (not pasting the full path but you get the idea)

[WARN] [MemoryDiscovery] Warning: Could not read GEMINI.md file at /Users/ ... / gemini.md. Error: EISDIR: illegal operation on a directory, read

I don't get to see this as I dont have a file at that path called gemini.md, I have a directory there called gemini.md

### Client information

<details>

```console
$ gemini /about
CLI Version             0.1.13                                       │
│ Git Commit              412b78c5 (local modifications)               │
│ Model                   gemini-2.5-pro                               │
│ Sandbox                 no sandbox                                   │
│ OS                      darwin                                       │
│ Auth Method             OAuth
```

</details>

### Login information

_No response_

_No response_

Pranith124:
…#4760
fix(core): prevent EISDIR error when GEMINI.md is a directory

Fixes #4760

## TLDR

This PR fixes a bug where the CLI would show a `[WARN] ... EISDIR: illegal operation on a directory, read` error if a **directory** named `GEMINI.md` existed in the project.

The fix adds a check to ensure a path is a **file** before attempting to read it, making the memory discovery process more robust and error-free.

## Dive Deeper

The root cause of the issue was in the `readGeminiMdFiles` function within the memory discovery module. The function would iterate through all discovered paths and immediately attempt to read each one using `fs.readFile()` without checking if the path was actually a file.

This change introduces a call to `await fs.stat(filePath)` before the read operation. It then checks the result with `stats.isFile()`. The `fs.readFile()` operation is now only performed if the path points to a **file**. If the path is a **directory** or other non-file entity, it is silently skipped, preventing the `EISDIR` error.

## Reviewer Test Plan

1. Pull down this branch and run:

   ```bash
   npm install

2.Ensure no GEMINI.md file or directory exists:

    ```bash
    rm -rf GEMINI.md
  
 3.Create a directory named GEMINI.md:

     ```bash
     mkdir GEMINI.md
   
 4.Run the CLI from source:
 
     ```bash
     npm start
  
  ✅ Validation: CLI should start successfully without any EISDIR warning in the console.
  
 5.(Optional Regression Test) Remove the directory:

      ```bash
      rmdir GEMINI.md

  Create a file instead:

      ```bash
      echo "test" > GEMINI.md

  Run the CLI again:

      ```bash
      npm start

✅ Validation: File content should load correctly.

<!-- Before submitting please validate your changes on as many of these options as possible -->

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ✅  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

<!--


-->

- Fixes #4760 
- Resolves #4760

### DIFF
--- a/packages/core/src/utils/memoryDiscovery.ts
+++ b/packages/core/src/utils/memoryDiscovery.ts
@@ -234,20 +234,29 @@ async function readGeminiMdFiles(
   const results: GeminiFileContent[] = [];
   for (const filePath of filePaths) {
     try {
-      const content = await fs.readFile(filePath, 'utf-8');
+      const stats = await fs.stat(filePath); // Get information about the path
 
-      // Process imports in the content
-      const processedContent = await processImports(
-        content,
-        path.dirname(filePath),
-        debugMode,
-      );
+      if (stats.isFile()) {
+        // <<< ADD THIS CHECK
+        const content = await fs.readFile(filePath, 'utf-8');
 
-      results.push({ filePath, content: processedContent });
-      if (debugMode)
-        logger.debug(
-          `Successfully read and processed imports: ${filePath} (Length: ${processedContent.length})`,
+        // Process imports in the content
+        const processedContent = await processImports(
+          content,
+          path.dirname(filePath),
+          debugMode,
         );
+
+        results.push({ filePath, content: processedContent });
+        if (debugMode)
+          logger.debug(
+            `Successfully read and processed imports: ${filePath} (Length: ${processedContent.length})`,
+          );
+      } else {
+        // It's a directory or something else, so just skip it silently.
+        if (debugMode) logger.debug(`Skipping non-file path: ${filePath}`);
+        results.push({ filePath, content: null });
+      }
     } catch (error: unknown) {
       const isTestEnv = process.env.NODE_ENV === 'test' || process.env.VITEST;
       if (!isTestEnv) {


packages/core/src/utils/memoryDiscovery.ts
```diff
@@ -234,20 +234,29 @@ async function readGeminiMdFiles(
   const results: GeminiFileContent[] = [];
   for (const filePath of filePaths) {
```
swissspidy, line 235: We should rather make sure we never pass directories in filePaths in the first place.

Pranith124: Thanks for the feedback, @swisspidy! That makes a lot of sense to fix it at the source. I'm happy to make that change.

To make sure I get it right, would the correct approach be to add fs.stat and isFile() checks inside the getGeminiMdFilePathsInternal function where it checks for the global path and does the upward scan?

jacob314: Sorry we missed this question. You are correct about where to make that change. After you make that change and add a test for this I'll approve.
Thanks again for this robustness polish!

jacob314:
Sorry for the confusion I believe https://github.com/google-gemini/gemini-cli/pull/5175/commits/ff31ac111941dc44fe0871618f30d29286d15dcc
landed fixing the same issue.
Thank you for taking this on and please do reach out if you tackle any other Gemini CLI issues and I'd be happy to help make sure you get a fast pull request review.

PR #4991, opened 2025-07-27
(model) - Use Flash Lite For Next Speaker Checks

abhipatel12:
## TLDR

We use Flash for next speaker checks. With the availability of Flash lite, we should use that instead for a simple check like this.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ✅  | -   | -   |

### DIFF
--- a/packages/core/src/config/models.ts
+++ b/packages/core/src/config/models.ts
@@ -6,4 +6,5 @@
 
 export const DEFAULT_GEMINI_MODEL = 'gemini-2.5-pro';
 export const DEFAULT_GEMINI_FLASH_MODEL = 'gemini-2.5-flash';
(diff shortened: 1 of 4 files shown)


NTaylorMullen (approved):
:shipit: nice addition. Lets also add an entry to /packages/core/src/core/tokenLimits.ts

abhipatel12:
Added the token limit!

acoliver:
This is still terrible. Why would you send a whole extra request with conversation to another model multiple times ina request cycle. it is such a token burn for what?

PR #5014, opened 2025-07-28
fix(neovim): fix neovim instantly quitting when diffing files with plugins

### LINKED ISSUE #3302: Lazyvim: nvim terminates immediately
### What happened?

When using lazyvim, nvim terminates immediately.

For me, this is related to the following command:

https://github.com/google-gemini/gemini-cli/blob/80aad5a42c4b631a675e36280bc6911965e8b797/packages/core/src/utils/editor.ts#L128

I do not observe this behaviour with an empty `init.lua` file effectively disabling lazyvim.

nvim not terminating automatically

### Client information

<details>

```console
$ gemini /about
│ CLI Version        0.1.9                              │
│ Git Commit         34935d6                            │
│ Model              gemini-2.5-pro                     │
│ Sandbox            no sandbox                         │
│ OS                 linux                              │
│ Auth Method        OAuth                              │

```

</details>

### Login information

Google Account

_No response_

### LINKED ISSUE #3496: Unable to Access External Editor Neovim within Neovim via Gemini CLI
### What happened?

When using the Gemini CLI inside Neovim (nvim), attempting to open an external editor session (likely for reviewing or modifying diffs) fails to launch or integrate properly. This hinders workflows that rely on nested editing or reviewing generated output from Gemini within the Neovim environment.

1.	Open nvim and launch a Gemini CLI session.
2.	Request a diff or content generation that triggers an “external editor” view or multi-file interaction.
3.	Attempt to open the generated content in a new buffer or editor via Gemini CLI.
4.	Observe that the editor session fails to open or does not allow proper interaction within Neovim.

The Gemini CLI should be able to launch or integrate the external editor view within Neovim or open a new buffer for review/modification without issue.

### Client information

<details>

```console
$ gemini /about
Environment:
	•	Gemini CLI Version: 0.1.9
	•	Git Commit: 34935d6
	•	Model: gemini-2.5-pro
	•	Sandbox: No
	•	Auth Method: OAuth
	•	OS: macOS (Darwin)
	•	Terminal: [Ghostty](https://ghostty.app/)
	•	Python Version: 3.11
	•	Neovim Version: Latest stable (nvim --version)
	•	Plugin Manager: Lazy.nvim

```

</details>

### Login information

google account

_No response_

### LINKED ISSUE #4767: using neovim as external editor, editor keeps quitting after lsp servers are run
### What happened?

when setting neovim as `/editor`
if neovim is invoked from gemini-cli, after a few seconds it just quits and goes back to gemini-cli .

I've bisected my nvim init.lua file and found out that if I comment out mason and all lsp configs, this problemn goes away.
I think gemini kills nvim session when nvim runs some external command inside the nvim session ?

neovim to keep running

### Client information

<details>

```console
$ gemini /about
│ CLI Version                                   0.1.13                                                                                 │
│ Git Commit                                    412b78c5 (local modifications)                                                         │
│ Model                                         gemini-2.5-pro                                                                         │
│ Sandbox                                       no sandbox                                                                             │
│ OS                                            darwin                                                                                 │
│ Auth Method                                   OAuth
```

</details>

### Login information

Google Account

_No response_

redoxahmii:
## TLDR

Solves the issue of instantly closing neovim when diffing files using `modify with external editor` with plugins enabled.

## Dive Deeper
Removed the diff coloring as neovim by default has diff colors that are sane as compared to vim.
Removed the `Autocmd` added for using `wqa` to quit all windows as that was the core issue causing it to instantly close when opening if you have plugins installed.
If it is mandatory to have the `autocmd` then the better approach would be to include `--noplugin` in the cli command to ensure it actually works instead of conflicting when plugins are installed. lmk and I will change it to that.
Tested it with `LazyVim` which is one of the heaviest configurations currently popular and it works fine with it and also with base neovim itself.
Currently, this PR also removes it from the `vim` editor command as well because this issue could triage to people using vim with heavy configurations as well. 

## Reviewer Test Plan
- Change editor to `neovim` in preferences.
- Tell gemini to add something to a file.
- Select `Modify with editor` option to see.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅ | ❓  | ✅  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs
Closes #4767 
Closes #3496 
Closes #3302

### DIFF
--- a/packages/core/src/core/client.ts
+++ b/packages/core/src/core/client.ts
@@ -44,7 +44,11 @@ import { DEFAULT_GEMINI_FLASH_MODEL } from '../config/models.js';
 import { LoopDetectionService } from '../services/loopDetectionService.js';
 import { ideContext } from '../ide/ideContext.js';
 import { logNextSpeakerCheck } from '../telemetry/loggers.js';
-import { NextSpeakerCheckEvent } from '../telemetry/types.js';
+import {
+  MalformedJsonResponseEvent,
+  NextSpeakerCheckEvent,
+} from '../telemetry/types.js';
+import { ClearcutLogger } from '../telemetry/clearcut-logger/clearcut-logger.js';
 
 function isThinkingSupported(model: string) {
   if (model.startsWith('gemini-2.5')) return true;
@@ -506,7 +510,7 @@ export class GeminiClient {
         authType: this.config.getContentGeneratorConfig()?.authType,
       });
 
-      const text = getResponseText(result);
+      let text = getResponseText(result);
       if (!text) {
         const error = new Error(
           'API returned an empty response for generateJson.',
@@ -519,6 +523,18 @@ export class GeminiClient {
         );
         throw error;
       }
+
+      const prefix = '```json';
+      const suffix = '```';
+      if (text.startsWith(prefix) && text.endsWith(suffix)) {
+        ClearcutLogger.getInstance(this.config)?.logMalformedJsonResponseEvent(
+          new MalformedJsonResponseEvent(modelToUse),
+        );
+        text = text
+          .substring(prefix.length, text.length - suffix.length)
+          .trim();
+      }
+
       try {
         return JSON.parse(text);
       } catch (parseError) {
@@ -532,7 +548,9 @@ export class GeminiClient {
           'generateJson-parse',
         );
         throw new Error(
-          `Failed to parse API response as JSON: ${getErrorMessage(parseError)}`,
+          `Failed to parse API response as JSON: ${getErrorMessage(
+            parseError,
+          )}`,
         );
       }
     } catch (error) {
--- a/packages/core/src/telemetry/clearcut-logger/clearcut-logger.ts
+++ b/packages/core/src/telemetry/clearcut-logger/clearcut-logger.ts
@@ -20,6 +20,7 @@ import {
   LoopDetectedEvent,
   NextSpeakerCheckEvent,
   SlashCommandEvent,
+  MalformedJsonResponseEvent,
 } from '../types.js';
 import { EventMetadataKey } from './event-metadata-key.js';
 import { Config } from '../../config/config.js';
@@ -42,6 +43,7 @@ const flash_fallback_event_name = 'flash_fallback';
 const loop_detected_event_name = 'loop_detected';
 const next_speaker_check_event_name = 'next_speaker_check';
 const slash_command_event_name = 'slash_command';
+const malformed_json_response_event_name = 'malformed_json_response';
 
 export interface LogResponse {
   nextRequestWaitMs?: number;
@@ -554,6 +556,21 @@ export class ClearcutLogger {
     }
 
     this.enqueueLogEvent(this.createLogEvent(slash_command_event_name, data));
+    this.flushIfNeeded();
+  }
+
+  logMalformedJsonResponseEvent(event: MalformedJsonResponseEvent): void {
+    const data = [
+      {
+        gemini_cli_key:
+          EventMetadataKey.GEMINI_CLI_MALFORMED_JSON_RESPONSE_MODEL,
+        value: JSON.stringify(event.model),
+      },
+    ];
+
+    this.enqueueLogEvent(
+      this.createLogEvent(malformed_json_response_event_name, data),
+    );
     this.flushIfNeeded();
   }
 
--- a/packages/core/src/telemetry/clearcut-logger/event-metadata-key.ts
+++ b/packages/core/src/telemetry/clearcut-logger/event-metadata-key.ts
@@ -183,6 +183,13 @@ export enum EventMetadataKey {
 
   // Logs the result of the next speaker check
   GEMINI_CLI_NEXT_SPEAKER_CHECK_RESULT = 44,
+
+  // ==========================================================================
+  // Malformed JSON Response Event Keys
+  // ==========================================================================
+
+  // Logs the model that produced the malformed JSON response.
+  GEMINI_CLI_MALFORMED_JSON_RESPONSE_MODEL = 45,
 }
 
 export function getEventMetadataKey(
--- a/packages/core/src/telemetry/types.ts
+++ b/packages/core/src/telemetry/types.ts
@@ -296,6 +296,18 @@ export class SlashCommandEvent {
   }
 }
 
+export class MalformedJsonResponseEvent {
+  'event.name': 'malformed_json_response';
+  'event.timestamp': string; // ISO 8601
+  model: string;
+
+  constructor(model: string) {
+    this['event.name'] = 'malformed_json_response';
+    this['event.timestamp'] = new Date().toISOString();
+    this.model = model;
+  }
+}
+
 export type TelemetryEvent =
   | StartSessionEvent
   | EndSessionEvent
@@ -307,4 +319,5 @@ export type TelemetryEvent =
   | FlashFallbackEvent
   | LoopDetectedEvent
   | NextSpeakerCheckEvent
-  | SlashCommandEvent;
+  | SlashCommandEvent
+  | MalformedJsonResponseEvent;
--- a/packages/core/src/utils/nextSpeakerChecker.ts
+++ b/packages/core/src/utils/nextSpeakerChecker.ts
@@ -14,27 +14,7 @@ const CHECK_PROMPT = `Analyze *only* the content and structure of your immediate
 **Decision Rules (apply in order):**
 1.  **Model Continues:** If your last response explicitly states an immediate next action *you* intend to take (e.g., "Next, I will...", "Now I'll process...", "Moving on to analyze...", indicates an intended tool call that didn't execute), OR if the response seems clearly incomplete (cut off mid-thought without a natural conclusion), then the **'model'** should speak next.
 2.  **Question to User:** If your last response ends with a direct question specifically addressed *to the user*, then the **'user'** should speak next.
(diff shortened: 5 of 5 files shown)


redoxahmii:
The tests have been updated, and also the misleading comment left back has been removed.
/gemini review

packages/core/src/utils/editor.ts
```diff
@@ -141,6 +140,29 @@ export function getDiffCommand(
...
+          'set showtabline=2 | set tabline=[Instructions]\\ :wqa(save\\ &\\ quit)\\ \\|\\ i/esc(toggle\\ edit\\ mode)',
+          '-c',
+          'wincmd h | setlocal statusline=OLD\\ FILE',
+          '-c',
+          'wincmd l | setlocal statusline=%#StatusBold#NEW\\ FILE\\ :wqa(save\\ &\\ quit)\\ \\|\\ i/esc(toggle\\ edit\\ mode)',
+          // Auto close all windows when one is closed
+          oldPath,
+          newPath,
+        ],
+      };
```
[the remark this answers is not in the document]
redoxahmii: The tests have been added with separate `args` for vim and neovim respectively.

packages/core/src/utils/editor.ts
```diff
@@ -141,6 +140,29 @@ export function getDiffCommand(
...
+          '-c',
+          'wincmd h | set readonly | wincmd l',
+          // Show helpful messages
+          '-c',
+          'set showtabline=2 | set tabline=[Instructions]\\ :wqa(save\\ &\\ quit)\\ \\|\\ i/esc(toggle\\ edit\\ mode)',
+          '-c',
+          'wincmd h | setlocal statusline=OLD\\ FILE',
+          '-c',
+          'wincmd l | setlocal statusline=%#StatusBold#NEW\\ FILE\\ :wqa(save\\ &\\ quit)\\ \\|\\ i/esc(toggle\\ edit\\ mode)',
+          // Auto close all windows when one is closed
```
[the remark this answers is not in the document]
redoxahmii: The comment is removed now.

packages/core/src/utils/editor.test.ts
```diff
@@ -261,38 +261,57 @@ describe('editor utils', () => {
...
+          '-c',
+          'highlight DiffAdd cterm=bold ctermbg=22 guibg=#005f00 | highlight DiffChange cterm=bold ctermbg=24 guibg=#005f87 | highlight DiffText ctermbg=21 guibg=#0000af | highlight DiffDelete ctermbg=52 guibg=#5f0000',
+          '-c',
+          'set showtabline=2 | set tabline=[Instructions]\\ :wqa(save\\ &\\ quit)\\ \\|\\ i/esc(toggle\\ edit\\ mode)',
+          '-c',
+          'wincmd h | setlocal statusline=OLD\\ FILE',
+          '-c',
+          'wincmd l | setlocal statusline=%#StatusBold#NEW\\ FILE\\ :wqa(save\\ &\\ quit)\\ \\|\\ i/esc(toggle\\ edit\\ mode)',
+          '-c',
+          'autocmd WinClosed * wqa',
```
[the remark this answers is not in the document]
redoxahmii: Done now.

packages/core/src/utils/editor.test.ts
```diff
@@ -261,38 +261,55 @@ describe('editor utils', () => {
...
+          'set showtabline=2 | set tabline=[Instructions]\\ :wqa(save\\ &\\ quit)\\ \\|\\ i/esc(toggle\\ edit\\ mode)',
+          '-c',
+          'wincmd h | setlocal statusline=OLD\\ FILE',
+          '-c',
+          'wincmd l | setlocal statusline=%#StatusBold#NEW\\ FILE\\ :wqa(save\\ &\\ quit)\\ \\|\\ i/esc(toggle\\ edit\\ mode)',
+          'old.txt',
+          'new.txt',
+        ],
+      },
     ];
```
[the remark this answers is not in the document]
redoxahmii: This has been added to the tests now.

packages/core/src/utils/editor.test.ts
```diff
@@ -261,38 +261,54 @@ describe('editor utils', () => {
...
-            'wincmd l | setlocal statusline=%#StatusBold#NEW\\ FILE\\ :wqa(save\\ &\\ quit)\\ \\|\\ i/esc(toggle\\ edit\\ mode)',
-            '-c',
-            'autocmd WinClosed * wqa',
-            'old.txt',
-            'new.txt',
-          ],
+          args: expectedArgs,
         });
       });
     }
```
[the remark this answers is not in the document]
redoxahmii: That's exactly how it was previously.

redoxahmii:
Gemini seems to be losing his mind so I'll wait for a human to review this so they know what they're talking about.

packages/core/src/utils/editor.ts
```diff
@@ -134,9 +133,28 @@ export function getDiffCommand(
           'wincmd h | setlocal statusline=OLD\\ FILE',
           '-c',
           'wincmd l | setlocal statusline=%#StatusBold#NEW\\ FILE\\ :wqa(save\\ &\\ quit)\\ \\|\\ i/esc(toggle\\ edit\\ mode)',
-          // Auto close all windows when one is closed
+          oldPath,
+          newPath,
+        ],
+      };
+    case 'neovim':
+      return {
```
galz10, line 141: Nit: this looks good, i think since we're modifying the behavior from what it was ( now users that use `:wq` will have to close out readonly file or use `:wqa`) i think it would be a good idea to call this out in the documentation, just calling out that if you're using the neovim or vim editor to use `:wqa`

galdawave (approved):

packages/core/src/utils/editor.test.ts
```diff
@@ -261,38 +261,55 @@ describe('editor utils', () => {
...
-            '-c',
-            'highlight DiffAdd cterm=bold ctermbg=22 guibg=#005f00 | highlight DiffChange cterm=bold ctermbg=24 guibg=#005f87 | highlight DiffText ctermbg=21 guibg=#0000af | highlight DiffDelete ctermbg=52 guibg=#5f0000',
-            '-c',
-            'set showtabline=2 | set tabline=[Instructions]\\ :wqa(save\\ &\\ quit)\\ \\|\\ i/esc(toggle\\ edit\\ mode)',
-            '-c',
-            'wincmd h | setlocal statusline=OLD\\ FILE',
-            '-c',
-            'wincmd l | setlocal statusline=%#StatusBold#NEW\\ FILE\\ :wqa(save\\ &\\ quit)\\ \\|\\ i/esc(toggle\\ edit\\ mode)',
-            '-c',
-            'autocmd WinClosed * wqa',
```
galdawave: I think all you need to do is change this to `autocmd BufWritePost * wqa` this will run the `:wqa` when the user writes to buffer.

galdawave (changes requested):

packages/core/src/utils/editor.test.ts
redoxahmii: Didn't know that the desired behavior in this case was to close the editor immediately on the first save.
It does override default behaviors for users that might be running `:w` or using `<c-s>` for writing buffer instead immediately close the editor instead of just writing to the buffer.
This can again conflict with some plugins that run formatters and write the buffer when they run.
If it is still a requirement, I'll change it.

EDIT: I tested with some popular formatters and plugins, and it doesn't seem to cause an issue. The only thing now is it overriding default behavior, but that is preferred to allow quickly getting back to `gemini-cli` so added.

packages/core/src/utils/editor.ts
```diff
@@ -138,9 +137,32 @@ export function getDiffCommand(
...
           'wincmd l | setlocal statusline=%#StatusBold#NEW\\ FILE\\ :wqa(save\\ &\\ quit)\\ \\|\\ i/esc(toggle\\ edit\\ mode)',
-          // Auto close all windows when one is closed
           '-c',
-          'autocmd WinClosed * wqa',
+          'autocmd BufWritePost * wqa',
+          oldPath,
+          newPath,
+        ],
+      };
+    case 'neovim':
```
galdawave: Thanks for your work on this! it looks like this is a more complex change than needed. The issue can actually be fixed by that one line change `WinClose -> BufWritePost`. Could you reset to the main branch, and then push a new commit that change?

galdawave (changes requested):

packages/core/src/utils/editor.ts
redoxahmii: The reason for separating it and not passing the `highlight` groups for neovim specifically is to allow it to work with multiple colorschemes which is more common with neovim users running a lot of plugins.

image

Above is the default `nvim` colors when diffing, which are already acceptable as compared to `vim` in this comparison.

image

Passing those colors for someone that has a colorscheme defined results in this, which isn't really clear or understandable.
Most colorschemes add the diff colors themselves so it's more coherent with the theme and the `Normal` highlight and it should inadvertently not cause any issues in the future for readability with any colorscheme.

image

image

Above are 2 examples of different colorschemes being used, but because they have their own defined highlights for diffs it allows for any other to also work without the readability issue.

I can of course reset it to main and add the changes I've made so far in a singular commit.
If it doesn't align with the repo then I'll reset the main and only add the `BufWritePost` fix.

Thank you for your time!

redoxahmii: I'm so sorry resetting on the main branch caused it to close the pull request and won't let me open it.
I'll make a separate pull request for this.

PR #5046, opened 2025-07-28
Updates schema, UX and prompt for IDE context

skeshive:
## TLDR

Updates schema, UX and prompt for IDE context

## Dive Deeper
* We now expect clients to pass a list of open files, with required fields path & timestamp. Optionally, clients can mark a file as currently active and provide additional context like the cursor position + selected text
* The summary / detailed view in the UI now refer to these files as open files rather than recent files for specificity sake
* The prompt now uses present tense ("this is the file that is open" instead of "this is the file that was recently open")

## Linked issues / bugs
#4800


mattKorwel (approved):
Lgtm

Issue #4321, opened 2025-07-16
Viewing light themes in dark terminals are unreadable


miguelsolorio:
### What happened?

We've been seeing reports of users trying out light themes in dark terminal backgrounds with readability issues. It's unclear if users understand these are intended for light themes. #4073 #3551

<img width="1140" height="934" alt="Image" src="https://github.com/user-attachments/assets/78686496-a263-48b9-8ebc-4d36c8e48d68" />

### What did you expect to happen?

Update CLI to use background colors

### Client information

<details>

```console
$ gemini /about
╭────────────────────────────────────────────────────────╮
│                                                        │
│ About Gemini CLI                                       │
│                                                        │
│ CLI Version        0.1.12-nightly.250714.b018e2d3      │
│ Git Commit         b018e2d3 (local modifications)      │
│ Model              gemini-2.5-pro-preview-05-06        │
│ Sandbox            no sandbox                          │
│ OS                 darwin                              │
│ Auth Method        gemini-api-key                      │
│                                                        │
╰────────────────────────────────────────────────────────╯
```

</details>

miguelsolorio:
Turns out that background colors are not supported on `<Box>` elements in ink, [intentionally](https://github.com/vadimdemedes/ink/issues/402#issuecomment-820605265):

> This is intentional, because previous major version of Ink had a lot of trouble implementing it, so I removed it in Ink 3 in favor of colorful text. You can add colorful borders to <Box> too, which could be a better looking alternative to colorful backgrounds in the terminal. See https://github.com/vadimdemedes/ink#bordercolor.

PR #5071, opened 2025-07-28
Re-order themes with light mode themes first.

### LINKED ISSUE #4611: Viewing dark themes in light terminals are unreadable
### What happened?

Similar to #4321 we need to make dark themes on light terminals readable Image

For dark themes to be readable on light terminal backgrounds

### Client information

<details>

```console
│ CLI Version        0.1.12-nightly.250714.b018e2d3      │
│ Git Commit         b018e2d3 (local modifications)      │
│ Model              gemini-2.5-pro-preview-05-06        │
│ OS                 darwin                              │
```

</details>

### Login information

_No response_

_No response_

sethtroisi:
## TLDR

Quick change to improve the initial experience for light mode terminals.

Fixes #4611
See also #4321 

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

### DIFF
--- a/packages/cli/src/ui/themes/theme-manager.ts
+++ b/packages/cli/src/ui/themes/theme-manager.ts
@@ -174,9 +174,9 @@ class ThemeManager {
     const sortedThemes = allThemes.sort((a, b) => {
       const typeOrder = (type: ThemeType): number => {
         switch (type) {
-          case 'dark':
-            return 1;
           case 'light':
+            return 1;
+          case 'dark':
             return 2;
           case 'ansi':
             return 3;


jacob314:
This will make the experience worse for dark mode terminal users who are likely the majority so we should avoid this and seek other ways to make this experience acceptable.

JustinStitt:
Perhaps the very first theme should be one that isn't horrible on either dark or light terminals. `Atom One` is pretty solid on both light and dark mode but it is currently the second option as of 0.1.13. Meanwhile, the `ANSI` Dark theme is currently the first option and is _awful_ for light mode users.

Or `gemini-cli` could try to be a bit invasive and read people's shell configs or vim configs to try and deduce what theme they would prefer and set it as the default. But that seems unreasonable.

sethtroisi:
I tried to read the current terminal color scheme but it's difficult to do generically. gemini-cli made an attempt at reading it for mac / linux but it failed in my zshell and over ssh. I think if we had confidence in which color scheme the terminal was in it would be worth re-ordering but not without. Closing for now.

PR #5075, opened 2025-07-28
Fixes bug where companion extension was not being included in release bundle

skeshive:
## TLDR
The companion extension VSIX was previously not being included in the released NPM package because the VSIX was not being built/

Note: This is to support the /ide install command before we publish the extension to marketplace. Once we publish the extension, this change will be reverted as /ide install will always point to the latest released public version of the extension

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "build:all": "npm run build && npm run build:sandbox && npm run build:vscode",
     "build:packages": "npm run build --workspaces",
     "build:sandbox": "node scripts/build_sandbox.js --skip-npm-install-build",
-    "bundle": "npm run generate && node esbuild.config.js && node scripts/copy_bundle_assets.js",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -28,7 +28,7 @@
     "build:all": "npm run build && npm run build:sandbox && npm run build:vscode",
     "build:packages": "npm run build --workspaces",
     "build:sandbox": "node scripts/build_sandbox.js --skip-npm-install-build",
-    "bundle": "npm run generate && node esbuild.config.js && node scripts/copy_bundle_assets.js",
+    "bundle": "npm run generate && node esbuild.config.js && npm run build:vscode && node scripts/copy_bundle_assets.js",
```
NTaylorMullen, line 31: IIRC we only bundle on `npx` double check me there though

NTaylorMullen (approved):

Issue #4251, opened 2025-07-15
UX Polish


jacob314:
## What problem does this solve?

The current UI could benefit from visual polish, usability, and configurability improvements.

This issue tracks the work to improve the visual polish, usability, and configurability of the Gemini CLI. The goal is provide a polished, mature, and configurable UI that delights users.

Core areas of focus include polishing the InputPrompt UI, improving the experience discovering and editing settings, and cleaning up workarounds such as only rendering at 90% of terminal width and not syntax highlighting Markdown that were needed to ensure that the initial release was robust enough.

gsquared94:
Found possible duplicate issues:

- #4245: (0.91175776720047)
- #4243: (0.9012332558631897)

devr0306:
Closing as all the issues have been addressed.

Issue #4245, opened 2025-07-15
Robust UX


jacob314:
## What problem does this solve?

Currently there are UX robustness issues that can get in the way of users being productive using Gemini CLI. 

In particular, there is a long tail of issues we need to address to reduce flicker, there are issues related to supporting all languages and character sets, and improvements are needed to improve the robustness of core functionality such as paste.

## How will it work?

We will continue to address flicker issues by ensuring that Gemini CLI UI  does not exceed the terminal height for all common use cases. This will involve using [MaxSizedBox](https://github.com/google-gemini/gemini-cli/blob/d3ee9de3c3b976ad45fe1a13eb49271d17a32e37/packages/cli/src/ui/components/shared/MaxSizedBox.tsx) in more places as well as likely extending the implementation of MaxSizedBox to support more cases. 

The end goal is Gemini CLI UX that feels robust and predictable for users. Users should not fear that doing the wrong thing will cause the UX to start flickering to the point they have to close the app or that pasting the wrong thing could break the tool.

#### Hedging our bets

Some of the fixes for UX polish will require landing upstream changes to [Ink] (https://github.com/vadimdemedes/ink) potentially including figuring out how to better integrate some of the use cases we currently solve with MaxSizedBox into Ink.

claell:
@jacob314 #1507 seems to be related as well (I can reproduce).

gsquared94:
Found possible duplicate issues:

- #4251: (0.91175776720047)

gshpychka:
> Found possible duplicate issues:
> 
> * [UX Polish #4251](https://github.com/google-gemini/gemini-cli/issues/4251): (0.91175776720047)

False positive

Adib234:
Thank you for bringing this to our attention. Right now, our engineering team is focusing all resources on critical system maintenance and core stability. Because of this, we don't have immediate plans to address this specific issue, but we truly appreciate you flagging it for us.

Issue #4243, opened 2025-07-15
Workstream: Core Functionality


jacob314:
This issue tracks the overall effort to improve the core functionality and UX of Gemini CLI.

This will be achieved by improving robustness, polishing the UI, and adding new features prioritizing closing material gaps with leading competitors. 

This initiative is broken down into several key areas. Please see linked sub issues below.

gsquared94:
Found possible duplicate issues:

- #4251: (0.9012332558631897)

gshpychka:
> Found possible duplicate issues:
> 
> * [UX Polish #4251](https://github.com/google-gemini/gemini-cli/issues/4251): (0.9012332558631897)

False positive

yahyaghariep-cyber:
![20260415_120518696.jpg_play_EG.jpg](https://github.com/user-attachments/assets/97f3167e-b136-4b46-b9c4-2578f727f849)

PR #5090, opened 2025-07-29
fix(config): Resolve duplicate config loading from home directory

Pranith124:
## TLDR

Fixes a bug where config files were loaded twice when running the CLI from the home directory by adding checks in both the CLI and core packages to prevent duplicate loading.

---

## Dive Deeper

Corrects a bug where running the CLI from the user's home directory would cause `settings.json` and `GEMINI.md` to be loaded twice, once for the user scope and once for the workspace scope.

This was happening because the application did not de-duplicate the paths when the user's home directory was also the current working directory.

This PR implements a robust, two-part fix:

1.  **In `packages/cli/src/config/settings.ts`**: A check is added to prevent the workspace `settings.json` from being loaded if the current directory is the user's home directory.
2.  **In `packages/core/src/utils/memoryDiscovery.ts`**: The core memory discovery logic is updated. It now checks if a `currentWorkingDirectory` was provided. If the path is empty, it correctly skips the entire workspace search for `GEMINI.md`, fixing the root cause of the bug as identified by the reviewer bot.

This improves startup efficiency and resolves the potential for configuration conflicts.

---

## Reviewer Test Plan

1.  Build the CLI with these changes (`npm run build`).
2.  Navigate to your **home directory** (`cd ~`).
3.  Run the CLI (`node /path/to/packages/cli/dist/index.js`).
4.  **Verify:** The CLI should start correctly without errors, and any debug logs should show that `settings.json` and `GEMINI.md` are loaded only once (the workspace search should be skipped).
5.  Navigate to a **different project directory** (e.g., `cd /path/to/some/other/project`).
6.  Run the CLI from there.
7.  **Verify:** The CLI should start correctly, loading both the user-level config and any workspace-level config if it exists. This confirms that normal behavior is unaffected.

---

## Testing Matrix

|           | 🍏  | 🪟    | 🐧    |
| --------- | --- | ----- | ----- |
| **npm run** | ✅  | ✅    | ✅    |
| **npx** | ❓  | ❓    | ❓    |
| **Docker** | ❓  | ❓    | ❓    |
| **Podman** | ❓  | -     | -     |
| **Seatbelt** | ❓  | -     | -     |

---

## Linked issues / bugs

- Closes #5056


allenhutchison:
This is a great fix for an important bug. The approach of checking if the current directory is the home directory is definitely the right way to go. I have a few suggestions to make this PR even more robust.

### 1. Fix Linting Issues

It looks like the `Lint` check is failing in the CI. Please run `npm run preflight` locally and commit the changes to fix the linting errors.

### 2. Use Real Path Resolution for Comparison

The current implementation uses direct string comparison for the directory paths. This can be unreliable across different operating systems or when symbolic links are involved. For example, if a user's setup involves a symlink to their home directory, this fix might not work as expected.

To make the comparison foolproof, I recommend resolving the canonical, real paths before comparing them. You can achieve this by using `path.resolve()` followed by `fs.realpathSync()`.

Here’s an example of how you could apply this in `packages/cli/src/config/settings.ts`:

```typescript
import * as fs from 'node:fs';
import * as path from 'node:path';
import { homedir } from 'node:os';

// In the loadSettings function:
const resolvedWorkspaceDir = path.resolve(workspaceDir);
const resolvedHomeDir = path.resolve(homedir());

let realWorkspaceDir: string;
try {
  // fs.realpathSync resolves symlinks and returns the canonical path.
  realWorkspaceDir = fs.realpathSync(resolvedWorkspaceDir);
} catch (e) {
  // The path might not exist yet, which is fine.
  // The subsequent fs.existsSync will handle it.
  realWorkspaceDir = resolvedWorkspaceDir;
}

const realHomeDir = fs.realpathSync(resolvedHomeDir);

// This comparison is now much more reliable.
if (realWorkspaceDir !== realHomeDir) {
  // ... load workspace settings
}
```

A similar change should be applied to `packages/cli/src/config/config.ts` to ensure the logic is consistent.

### 3. Add a Regression Test

While all the existing checks pass, this PR would be even stronger with a new test case specifically for this scenario. Adding a regression test will ensure this bug doesn't accidentally reappear in the future.

A good test would involve:
*   Mocking `os.homedir()` to return a controlled test path.
*   Setting the current working directory to the same mocked path.
*   Spying on the file loading functions to assert that the configuration and memory files are loaded only once.

This isn't necessarily a blocker, but it would be a valuable addition.

Great work on tracking this down! With these adjustments, the fix will be rock-solid.

Pranith124:
Hi @allenhutchison,

Thank you so much for the detailed and helpful feedback! I've made all the suggested changes.

- **1. Linting Issues:** I ran `npm run preflight` and fixed the linting error. The checks are all passing now.
- **2. Real Path Resolution:** I've updated the logic in both `settings.ts` and `config.ts` to use `fs.realpathSync` for a more robust path comparison, just as you suggested.
- **3. Test Fixes:** The changes caused some of the existing tests in `settings.test.ts` to fail, so I've updated the test mocks to account for the new logic.

The full test suite is passing locally now. I believe this addresses all your points. Could you please take another look when you have a moment? Thanks again for your guidance!

allenhutchison:
This is a great fix! Using `realpathSync` to resolve symlinks is the correct way to prevent settings from being loaded multiple times.

I've filed a separate issue (#5227) to address the unrelated snapshot test failure. Once that is resolved, this PR will be ready to merge.

Thank you for the excellent contribution!

Pranith124:
Thank you @allenhutchison for your valuable feedback and guidance.

I'm a beginner in open source contribution and eager to gain experience. I would really appreciate it if you could assign me an issue to work on.

Thank you once again!

jacob314:
> Thank you @allenhutchison for your valuable feedback and guidance.
> 
> I'm a beginner in open source contribution and eager to gain experience. I would really appreciate it if you could assign me an issue to work on.
> 
> Thank you once again!

Thank you so much for your contribution! This was a nice issue to fix.
If you would like to work on other similar issues, here a lists of high priority but approachable issues we are hoping to fix over the next couple months.
UX Robustness:
https://github.com/google-gemini/gemini-cli/issues/4245

UX Polish:
https://github.com/google-gemini/gemini-cli/issues/4251

I'm still finishing categorizing all the Core issues but here is the overall issue that I will be adding the other epics of work we are trying to land to core and UX functionality of Gemini CLI.
https://github.com/google-gemini/gemini-cli/issues/4243

Pranith124:
Thank you @jacob314! I’ll definitely go through it and try to resolve the issues.

allenhutchison (approved):

Issue #5086, opened 2025-07-29
You are referencing files with [id] in the path. The tool is unable to read files with brackets in the name.


luffy-taro:
### What happened?

I have a react native project using expo-router and the file names contain `[id]` pattern 

```
I see that you are referencing files with [id] in the path. The tool is unable to read files with brackets
  in the name. To help me with your request, please provide the whole content of the files:
  src/app/(app)/Tab1/[id]/index.tsx and src/app/(app)/Tab2/[id]/index.tsx.
```

### What did you expect to happen?

It's not able to read file but it's possible to provide directory instead till `src/app/(app)/Tab1/` 

### Client information

<details>

```console
$ gemini /about
│ CLI Version            0.1.14                                      │
│ Git Commit             de968877 (local modifications)              │
│ Model                  gemini-2.5-pro                              │
│ Sandbox                no sandbox                                  │
│ OS                     linux                                       │
│ Auth Method            OAuth                                       │
│ GCP Project            gen-lang-client-0227558081                  │
│                                                           
```
</details>

### Login information

Google Account

gsquared94:
Found possible duplicate issues:

- #3625: (#0.9057774543762207)

phanipaladugula:
Hello, I want to contribute on this so can you provide any specific information about the issue

Issue #3625, opened 2025-07-09
Reference a file with square brackets in its name


aborruso:
### What happened?

Hi, I have this astro file "src/pages/initiative/[id].astro".

If I refer to it, using `@`, it is not referenced.

### What did you expect to happen?

To be able to point to it

### Client information

*   **CLI Version:** 0.1.9
*   **Git Commit:** I do not know
*   **Operating System:** Debian 12
*   **Sandbox Environment:** no sandbox
*   **Model Version:** gemini-2.5-pro
*   **Memory Usage:** I do not know

### Login information

Via google account

gsquared94:
Found possible duplicate issues:

- #5086: (0.9057774543762207)

PR #5100, opened 2025-07-29
fix: Enable reading files with brackets in paths (e.g., expo-router `[id]` routes)

nehaaprasad:
#5086

### TLDR

- The tool couldn't read files like `src/app/(app)/Tab1/[id]/index.tsx` because glob patterns treat brackets as special characters. Added `escapeGlobPattern()` function to escape special glob characters, allowing proper file discovery for expo-router and similar projects.

### Dive Deeper

The issue occurred in the fallback glob search when direct file paths weren't found. The pattern `**/*${pathName}*` would fail for paths containing `[`, `]`, `(`, `)`, etc. because these have special meaning in glob patterns.

**Root cause**: Glob interprets `[id]` as a character class matching any single character that is 'i' or 'd', not as literal brackets.

**Solution**: Added `escapeGlobPattern()` function that escapes special glob characters using regex replacement. Applied this to all three places where glob patterns are used with user input:
- `@` command processing in `atCommandProcessor.ts`
- ACP peer file handling in `acpPeer.ts` 
- File completion suggestions in `useCompletion.ts`

**Impact**: Minimal changes that preserve existing API and functionality while enabling support for modern routing patterns.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

### DIFF
--- a/packages/cli/src/acp/acpPeer.ts
+++ b/packages/cli/src/acp/acpPeer.ts
@@ -47,6 +47,10 @@ export async function runAcpPeer(config: Config, settings: LoadedSettings) {
   );
 }
 
+function escapeGlobPattern(pattern: string): string {
+  return pattern.replace(/[[\]{}()*+?.\\^$|]/g, '\\$&');
+}
+
 class GeminiAgent implements Agent {
   chat?: GeminiChat;
   pendingSend?: AbortController;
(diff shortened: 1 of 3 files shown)


psinha40898:
Hey just helping out, unless I missed something this also seems relevant to #3625 so perhaps someone should mark it as duplicate or include it in the PR etc.

Good luck!

chrstnb:

PR #5130, opened 2025-07-29
Update prompt commands to avoid collisions

haroldmciver-go:
## TLDR

Update how mcp prompts slash commands are loaded to prevent name collisions with built-in and custom commands. Previously, collision checking only happened between MCP prompts themselves. Now, the slashCommandProcessor checks if a prompt name is already taken by any other command type. If a collision is found (e.g., an MCP prompt is named `/help`), the MCP prompt's command is automatically prefixed with its server name, becoming `/server-name_help`.

## Dive Deeper

The responsibility for handling name collisions has been moved from the PromptRegistry to the slashCommandProcessor. This allows for the processor to have a complete view of all command sources (built-in, custom files, and MCP).

The new loading sequence is Built-in and file-based commands are loaded first, and their names are added to a "reserved names" set. MCP-backed prompt commands are then loaded. Each MCP command's name is checked against the reserved set. If a conflict exists, it's renamed.

The /mcp command's output has also been updated to clearly show both the prompt's original name and its new, renamed command.

## Reviewer Test Plan

Test Collision with a Custom Command:

Create a local file-based command named test (e.g., in `~/.gemini/commands/test.toml`).

Define a prompt on your MCP server also named `test`.

Run the CLI and execute `/mcp`.

Verify: The MCP prompt has been renamed to `/prompt-server_test` to avoid conflicting with your local command file.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


packages/cli/src/ui/hooks/slashCommandProcessor.test.ts
```diff
@@ -711,15 +696,28 @@ describe('useSlashCommandProcessor', () => {
       const result = setupProcessorHook([], [fileCommand], [mcpCommand]);
 
       await waitFor(() => {
-        // The service should only return one command with the name 'override'
-        expect(result.current.slashCommands).toHaveLength(1);
+        // ✅ CHANGE: Expect 2 commands now, not 1.
```
chrstnb: Remove this and the comment below

packages/cli/src/ui/hooks/slashCommandProcessor.test.ts
```diff
@@ -711,15 +696,28 @@ describe('useSlashCommandProcessor', () => {
...
+      const renamedCmd = result.current.slashCommands.find(
+        (cmd) => cmd.name === 'test_server_override',
+      );
+
+      expect(originalCmd).toBeDefined();
+      expect(originalCmd?.description).toBe('file'); // The file command keeps the name
+      expect(renamedCmd).toBeDefined();
+      expect(renamedCmd?.description).toBe('mcp'); // The MCP command is renamed
+
+      // This part of the test remains valid: running `/override` should trigger the file command.
```
chrstnb: Remove

packages/cli/src/ui/commands/mcpCommand.ts
```diff
@@ -248,10 +249,16 @@ const getMcpStatus = async (
       if (serverTools.length > 0) {
         message += '\n';
       }
-      message += `  ${COLOR_CYAN}Prompts:${RESET_COLOR}\n`;
-      serverPrompts.forEach((prompt: DiscoveredMCPPrompt) => {
+      message += `  ${RESET_COLOR}Prompts:${RESET_COLOR}\n`;
+      serverPrompts.forEach((prompt) => {
+        const originalName = prompt.originalName || prompt.name;
```
chrstnb: I think this would make more sense as `prompt.name` as the name the prompt has in the MCP server and `prompt.displayName` as the potentially conflict-aware display name

packages/cli/src/ui/hooks/slashCommandProcessor.ts
```diff
@@ -180,22 +181,64 @@ export const useSlashCommandProcessor = (
       toggleCorgiMode,
       toggleVimEnabled,
       sessionShellAllowlist,
+      commands,
```
chrstnb, line 185: is this necessary?

packages/cli/src/ui/hooks/slashCommandProcessor.ts
```diff
@@ -180,22 +181,64 @@ export const useSlashCommandProcessor = (
...
+        // If the name is already taken by any other command, prepend the server name.
+        if (commandNames.has(newName)) {
+          newName = `${mcpCmd.serverName}_${mcpCmd.name}`;
+        }
+
+        // If the prefixed name also collides, the command would be shadowed.
+        // To prevent this, we log a warning and skip adding the command.
+        if (commandNames.has(newName)) {
+          console.warn(
+            `Command name collision for MCP prompt "${originalName}" from server "${mcpCmd.serverName}". The prefixed name "/${newName}" is also in use. This prompt will be unavailable.`,
```
chrstnb: This happens when there are two prompts in an MCP server with the same name, right? I actually don't think that's possible, come to think of it, but I would at minimum update the console warning to be more concise

packages/cli/src/ui/hooks/slashCommandProcessor.ts
```diff
@@ -180,22 +181,64 @@ export const useSlashCommandProcessor = (
...
+
+        commandNames.add(newName);
+        finalCommands.push({
+          ...mcpCmd,
+          name: newName,
+          originalName: originalName,
+        });
+      }
+
+      setCommands(finalCommands.sort((a, b) => a.name.localeCompare(b.name)));
```
chrstnb: is the sorting necessary?

packages/core/src/prompts/prompt-registry.ts
```diff
@@ -7,50 +7,38 @@
...
-      this.prompts.set(newName, { ...prompt, name: newName });
-    } else {
-      this.prompts.set(prompt.name, prompt);
-    }
+    this.prompts.push(prompt);
   }
 
   /**
    * Returns an array of all registered and discovered prompt instances.
+   * Name conflicts are handled by the slash command processor.
```
chrstnb: Would remove as this is an implementation detail of the slash command processor

packages/core/src/prompts/prompt-registry.ts
```diff
@@ -7,50 +7,38 @@
...
   getAllPrompts(): DiscoveredMCPPrompt[] {
-    return Array.from(this.prompts.values()).sort((a, b) =>
-      a.name.localeCompare(b.name),
-    );
+    return [...this.prompts].sort((a, b) => a.name.localeCompare(b.name));
   }
 
   /**
    * Get the definition of a specific prompt.
+   * If multiple prompts share the same name, the first one found is returned.
```
chrstnb: This would be a bad idea, but I searched and this isn't being used anywhere, so fine to just remove it!

chrstnb: (the function that is)

packages/cli/src/ui/commands/mcpCommand.ts
```diff
@@ -97,13 +96,15 @@ const getMcpStatus = async (
...
+  const allCommands = context.commands || [];
   for (const serverName of serverNames) {
     const serverTools = allTools.filter(
       (tool) =>
         tool instanceof DiscoveredMCPTool && tool.serverName === serverName,
     ) as DiscoveredMCPTool[];
-    const promptRegistry = await config.getPromptRegistry();
-    const serverPrompts = promptRegistry.getPromptsByServer(serverName) || [];
+    const serverPrompts = allCommands.filter(
+      (cmd) => cmd.serverName === serverName,
```
chrstnb: This change seems unnecessary?

packages/cli/src/ui/hooks/slashCommandProcessor.ts
```diff
@@ -180,22 +181,64 @@ export const useSlashCommandProcessor = (
...
-      const commandService = await CommandService.create(
-        loaders,
-        controller.signal,
-      );
-      setCommands(commandService.getCommands());
+      const mcpLoader = new McpPromptLoader(config);
+      const builtinLoader = new BuiltinCommandLoader(config);
+      const fileLoader = new FileCommandLoader(config);
+
+      // CommandService.create used to do this. We now do it here to allow
```
chrstnb: Remove

packages/cli/src/ui/hooks/slashCommandProcessor.test.ts
```diff
@@ -181,17 +165,17 @@ describe('useSlashCommandProcessor', () => {
       const result = setupProcessorHook([builtinCommand], [fileCommand]);
 
       await waitFor(() => {
-        // The service should only return one command with the name 'override'
-        expect(result.current.slashCommands).toHaveLength(1);
+        // The processor loads both commands without de-duplicating.
```
chrstnb: This seems suspicious-- I don't think this PR handles this case

packages/cli/src/ui/hooks/slashCommandProcessor.test.ts
```diff
@@ -148,23 +148,7 @@ describe('useSlashCommandProcessor', () => {
       expect(mockMcpLoadCommands).toHaveBeenCalledTimes(1);
     });
 
-    it('should provide an immutable array of commands to consumers', async () => {
```
chrstnb: Why is this removed?

packages/cli/src/ui/hooks/slashCommandProcessor.ts
haroldmciver-go: The commands dependency is required for commandContext object to be rebuilt with the complete list of commands after they finish loading, so the /mcp command receives the updated list of commands

packages/cli/src/ui/commands/mcpCommand.ts
haroldmciver-go: The old way grabbed the original command names from a registry, which didn't know if a command had been renamed to fix a collision. The new way uses the final, processed list of all commands, so it always has the correct and active names.

packages/cli/src/ui/hooks/slashCommandProcessor.ts
haroldmciver-go: not necessary but commands will be displayed in sorted order

packages/cli/src/ui/hooks/slashCommandProcessor.test.ts
```diff
@@ -179,7 +179,7 @@ describe('useSlashCommandProcessor', () => {
       }).toThrow(TypeError);
     });
 
-    it('should override built-in commands with file-based commands of the same name', async () => {
+    it('should prioritize built-in commands over file-based commands of the same name', async () => {
```
chrstnb, line 182: What is the expected behavior change here?

haroldmciver-go: The test was updated so that built-in commands now have precedence over file-based and MCP commands

chrstnb:

PR #5136, opened 2025-07-29
feat(core, cli): Introduce high-performance FileSearch engine

### LINKED ISSUE #5301: High performance FileSearch engine

Introduce a new, high-performance FileSearch engine in @google/gemini-cli-core and integrate it into the CLI's file path autocompletion (@ syntax). This is expected to significantly improves the speed and responsiveness of file searching, especially in large repositories.
This may also mitigate memory usage issues due to the inefficiency of the current approach.

### Why is this needed?

File search is too slow on large repositories and uses too much memory.

brychanrobot:
## TLDR

This pull request introduces a new, high-performance `FileSearch` engine in `@google/gemini-cli-core` and integrates it into the CLI's file path autocompletion (`@` syntax). This significantly improves the speed and responsiveness of file searching, especially in large repositories. Rough testing on the Chromium codebase showed an 18x latency improvement for the first character typed, and it only gets faster as the user keeps typing. 

## Dive Deeper

The previous file completion logic was slow, relying on recursive `fs` calls or `glob` which did not scale well. This PR replaces it with a purpose-built `FileSearch` engine with the following characteristics:

-   **Performance**: Uses `fdir` for highly efficient, asynchronous file system crawling. For optimal speed, the initial crawl is only filtered by directory-level ignore patterns (`node_modules/`, `dist/`). File-level patterns (`*.log`) are applied later to the much smaller result set
-   **Caching**: Implements a two-tier cache:
    1.  A `CrawlCache` singleton avoids re-crawling the entire file system. The cache is keyed by the project's ignore rules and has a configurable Time-to-Live (TTL) to ensure freshness.
    2.  A `ResultCache` reuses results from broader queries for more specific ones (e.g., results for "src/" are used as the base for "src/components").

## Future work

FileSearch can be used to replace existing usages of glob, such as `GlobTool` and `atCompletionProcessor.ts`.

## Reviewer Test Plan

To test these changes, pull this branch and run the CLI. The primary area to test is the `@` file path autocompletion and submission.

1.  **Test Autocompletion Performance**:
    -   `cd` into a very large repository (e.g., a checkout of `chromium`).
    -   Run Gemini CLI.
    -   Type `@` and start typing a file name. The suggestions should appear quickly. No need to do any manual globbing. You can just start typing any file or directory name from any level of the file structure.
    -   Try searching for a few different files to test the in-memory cache.

2.  **Test Ignore Directory Logic**:
    -   In a standard repo, type `@node_modules/` and verify that no files from `node_modules` are suggested.

4.  **Test Submission Logic**:
    -   Submit a prompt with multiple files: `What are the differences between @packages/core/index.ts and @packages/cli/index.ts?`
    -   Submit a prompt with an @ completed file as the last thing in the prompt: `How many tests are in @packages/stuff/foo.test.ts`

## Testing Matrix

|          | Linux | macOS | Windows |
| -------- | :---: | :---: | :---: |
| npm run  |  ✅   |  ❓   |  ❓   |
| npx      |  ✅   |  ❓   |  ❓   |
| Docker   |  ❓   |  ❓   |  ❓   |
| Podman   |  ❓   |   -   |   -   |
| Seatbelt |  ❓   |   -   |   -   |

## Linked issues / bugs

Fixes: #5301


packages/cli/src/ui/components/InputPrompt.tsx
```diff
@@ -220,7 +220,7 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
           }
         }
 
-        if (key.name === 'tab' || (key.name === 'return' && !key.ctrl)) {
```
jacob314: is this change intentional?

packages/core/src/utils/filesearch/fileSearch.ts
```diff
@@ -0,0 +1,261 @@
...
+  private resultCache: ResultCache | undefined;
+  private allFiles: string[] = [];
+
+  /**
+   * Constructs a new `FileSearch` instance.
+   * @param options Configuration options for the file search.
+   */
+  constructor(private readonly options: FileSearchOptions) {
+    this.absoluteDir = path.resolve(options.projectRoot);
+    options.ignoreDirs = ['.git', ...options.ignoreDirs];
```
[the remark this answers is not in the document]
jacob314: please fix

packages/core/src/utils/filesearch/fileSearch.ts
```diff
@@ -0,0 +1,261 @@
...
+  const patternFilter = picomatch(pattern, {
+    dot: true,
+    contains: true,
+    nocase: true,
+  });
+
+  const results: string[] = [];
+  for (const [i, p] of allPaths.entries()) {
+    // Yield control to the event loop periodically to prevent blocking.
+    if (i % 1000 === 0) {
```
jacob314, line 52: nit: how long does it typically take per 1000 entries? Potentially yield only every 32 ms of elapsed time rather than a strict interval.

packages/cli/src/ui/hooks/useCompletion.ts
```diff
@@ -157,6 +173,89 @@ export function useCompletion(
     return false;
   }, [buffer.text, buffer.cursor, buffer.lines]);
 
+  const fetchFileSuggestions = useCallback(
+    async (prefix: string, signal: AbortSignal) => {
+      if (!fileSearchRef.current) return;
+
+      // Only show loading spinner if there are no suggestions currently displayed.
```
jacob314: thanks for fixing this! the previous UI flashed a lot showing the loading spinner.
typically the idiomatic way to handle avoiding flicker for cases like this is to wait some number of ms (e.g 100) before starting to show the spinner to avoid showing it when a response is about to be available.

packages/cli/src/ui/hooks/useCompletion.ts
```diff
@@ -157,6 +173,89 @@ export function useCompletion(
...
+      setIsLoadingSuggestions(true);
+      fileSearchRef.current = new FileSearch({
+        projectRoot: cwd,
+        // TODO(b/343593467): Make these configurable.
+        ignoreDirs: [],
+        useGitignore: true,
+        useGeminiignore: true,
+        cache: true,
+        cacheTtl: 30, // 30 seconds
+      });
```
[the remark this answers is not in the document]
jacob314: please fix.

packages/cli/src/ui/hooks/useCompletion.ts
```diff
@@ -157,6 +173,89 @@ export function useCompletion(
...
+      } finally {
+        if (!signal.aborted) {
+          setIsLoadingSuggestions(false);
+        }
+      }
+    },
+    [resetCompletionState],
+  );
+
+  // Effect to initialize the FileSearch engine
```
jacob314: Can you try to manage the complex state of the file searcher (status, isLoading) with a useReducer hook. This consolidates all state transitions into a
  single dispatch function, making the useEffect cleaner and ensuring state updates are handled more predictably. Gemini CLI is generally good at performing these sorts of refactors to use reducers and they make things more robust then the alternatives using useEffect which have unintended consequences particularly for cases like this where you are setting state within a useEffect.

jacob314 (commented):
Amazing work! The performance now feels instant when I run this on the Gemini CLI project and I don't see any more cases with unexpected completion results.
Approved once the comments I added are addressed.

packages/cli/src/ui/components/InputPrompt.tsx
brychanrobot: Yes. Without this it's imposslble to submit a prompt with an @ file at the end of it. You just keep pressing enter and it keeps re-showing the results. I don't think there's a good way around it, since you could have a valid @file that matches more than one file (a glob or a directory). I'm open to other suggestions for how to handle it. A user can also add a space after the file entry and then enter will work.

packages/core/src/utils/filesearch/fileSearch.ts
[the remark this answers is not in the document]
brychanrobot: Done

packages/cli/src/ui/hooks/useCompletion.ts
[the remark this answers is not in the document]
brychanrobot: Done

packages/cli/src/ui/hooks/useCompletion.ts
brychanrobot: I attempted this, but it got complicated. I couldn't manage to get it to not break other things and keep tests passing. I'm sure someone with more react experience could get it to work, but as discussed offline, I'm going to leave it as-is.

packages/cli/src/ui/hooks/useCompletion.ts
brychanrobot: Done. I've switched to the 100ms timer

packages/core/src/utils/filesearch/fileSearch.ts
brychanrobot: I tried this out and found that running performance.now on every iteration was doubling the time to execute the loop. I thought about doing a hybrid approach where you check the time eg. every 250ms, but that was starting to feel a bit over complicated. With the yield logic as-is I tried just increasing from 1k to 10k and didn't measure any performance difference. So I think yielding is pretty cheap as long as there aren't actually other tasks to do. I'm opting to leave this as-is.

brychanrobot (commented):
I think I've addressed all of the comments. The remaining things are trying it out on windows and resolving the merge conflicts.

packages/cli/src/ui/components/InputPrompt.tsx
jacob314: actually there is a fix for this. you should should insert a space after the input file name as part of the accepted completion. then you just have to press enter twice to submit.

brychanrobot: I thought my change had created this problem, but it turns out it was already existing. I'm going to leave the behavior unmodified for now, and we can fix it in a followup.

For the followup change, how about a compromise? Enter adds a space, but tab doesn't. That allows users to complete part of a path, but keep digging deeper into the completion. I commonly use a flow like that to explore directories and find what I want via autocomplete. But double enter would submit a query. I've experimented with a prototype of this and it felt nice.

packages/cli/src/ui/hooks/useAtCompletion.ts
```diff
@@ -0,0 +1,224 @@
...
+      return state;
+    case 'ERROR':
+      return {
+        ...state,
+        status: AtCompletionStatus.ERROR,
+        isLoading: false,
+        suggestions: [],
+      };
+    case 'RESET':
+      return { ...state, suggestions: [], isLoading: false, pattern: null };
```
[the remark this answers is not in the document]
brychanrobot: Done.

packages/cli/src/ui/hooks/useAtCompletion.ts
```diff
@@ -0,0 +1,224 @@
...
+    pattern,
+    config,
+    cwd,
+    setSuggestions,
+    setIsLoadingSuggestions,
+  } = props;
+  const [state, dispatch] = useReducer(atCompletionReducer, initialState);
+  const fileSearch = useRef<FileSearch | null>(null);
+  const searchAbortController = useRef<AbortController | null>(null);
+  const slowSearchTimer = useRef<NodeJS.Timeout | null>(null);
```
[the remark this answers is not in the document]
brychanrobot: Done

packages/core/src/utils/filesearch/fileSearch.ts
```diff
@@ -0,0 +1,263 @@
...
+        break;
+      }
+      // The `ignore` library throws an error if the path is '.', so we skip it.
+      if (candidate === '.') {
+        continue;
+      }
+      if (!fileFilter(candidate)) {
+        results.push(candidate);
+      }
+    }
```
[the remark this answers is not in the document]
brychanrobot: Done

packages/cli/src/ui/hooks/useCommandCompletion.test.ts
```diff
@@ -48,45 +99,25 @@ describe('useCommandCompletion', () => {
...
       it('should initialize with default state', () => {
-        const slashCommands = [
-          { name: 'dummy', description: 'dummy' },
-        ] as unknown as SlashCommand[];
         const { result } = renderHook(() =>
           useCommandCompletion(
             useTextBufferForTest(''),
             testDirs,
             testRootDir,
-            slashCommands,
```
jacob314, line 89: revert this change.

packages/cli/src/ui/hooks/useCommandCompletion.test.ts
```diff
@@ -165,1140 +191,302 @@ describe('useCommandCompletion', () => {
...
-
         act(() => {
           result.current.navigateDown();
         });
 
-        expect(result.current.activeSuggestionIndex).toBe(0);
+        expect(result.current.activeSuggestionIndex).toBe(-1);
       });
 
-      it('should handle navigation with multiple suggestions', () => {
```
jacob314, line 289: same as previous can we revert to actually test the slash commands. seems odd this test had to be modified as is unrelated to changes.

packages/cli/src/ui/hooks/useReverseSearchCompletion.tsx
```diff
@@ -41,12 +41,19 @@ export function useReverseSearchCompletion(
     navigateDown,
   } = useCompletion();
 
-  // whenever reverseSearchActive is on, filter history
+  // Effect 1: Reset state when reverse search is deactivated.
```
jacob314: remove `Effect  1:` and `Effect 2:` comments.

packages/cli/src/ui/hooks/useCommandCompletion.tsx
```diff
@@ -72,541 +68,115 @@ export function useCommandCompletion(
...
-    reverseSearchActive,
     setSuggestions,
-    setShowSuggestions,
-    setActiveSuggestionIndex,
     setIsLoadingSuggestions,
     setIsPerfectMatch,
-    setVisibleStartIndex,
+  });
+
+  useEffect(() => {
```
jacob314, line 155: it is unfortunate this is a useEffect. Can you clean this up? If it is hard can leave for a followup pr.

jacob314 (commented):
Thanks for sticking with this. Approving as soon as these minor comments are addressed.

packages/cli/src/ui/hooks/useReverseSearchCompletion.tsx
brychanrobot: Done

packages/cli/src/ui/hooks/useCommandCompletion.tsx
brychanrobot: I think it's actually not necessary. I've removed it.

packages/cli/src/ui/hooks/useCommandCompletion.test.ts
brychanrobot: Same as previous

packages/cli/src/ui/hooks/useCommandCompletion.test.ts
brychanrobot: Discussed offline. We'll need to refactor this test suite as follow up to have less mocking.

jacob314 (approved):
lgtm

PR #5137, opened 2025-07-29
Fix: Ensure that non interactive mode and interactive mode are calling the same entry points

anj-s:
## TLDR
Fix divergence in the code path between non interactive and interactive mode. We previously called sendMessageStream in geminiChat that meant bypassing nextSpeakerCheck, loop checks, memory compression and other critical functionality that we had been adding.

## Dive Deeper

* API Interaction Refactor: The nonInteractiveCli now consumes ServerGeminiStreamEvent from the Gemini client's sendMessageStream, replacing the previous GenerateContentResponse parsing. This allows for direct handling of content and tool call events from the model stream.
* Streamlined Tool Call Handling: Tool call requests are now explicitly identified and processed from the ServerGeminiStreamEvent stream, improving the clarity and reliability of tool execution within the non-interactive CLI.
Improved Test Suite: The associated test file (nonInteractiveCli.test.ts) has been updated to reflect the new event-based streaming API, ensuring comprehensive coverage of the refactored logic. Mocking strategies for core modules and process interactions have also been modernized.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

### DIFF
--- a/packages/cli/src/nonInteractiveCli.ts
+++ b/packages/cli/src/nonInteractiveCli.ts
@@ -41,7 +41,7 @@ export async function runNonInteractive(
     while (true) {
       turnCount++;
       if (
-        config.getMaxSessionTurns() > 0 &&
+        config.getMaxSessionTurns() >= 0 &&
         turnCount > config.getMaxSessionTurns()
       ) {
         console.error(
--- a/packages/cli/src/nonInteractiveCli.ts
+++ b/packages/cli/src/nonInteractiveCli.ts
@@ -11,37 +11,11 @@ import {
   ToolRegistry,
   shutdownTelemetry,
   isTelemetrySdkInitialized,
+  GeminiEventType,
 } from '@google/gemini-cli-core';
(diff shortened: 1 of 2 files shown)


packages/cli/src/nonInteractiveCli.ts
NTaylorMullen, line 1: This is fine for now but we should upgrade our abstractions to make it nearly impossible to diverge in the future. Approving!

NTaylorMullen (approved):

packages/cli/src/nonInteractiveCli.ts
anj-s: Agreed. This is a short term fix. We need a redesign soon.

PR #5146, opened 2025-07-29
Add toggleable IDE mode setting

### LINKED ISSUE #4393: Add support for toggling IDE mode via /ide command
Currently, IDE mode is toggled on/off via a CLI flag. Instead, we should allow toggling of IDE mode via the /ide slash command so the user can toggle it in the middle of a session without having to restart the CLI. As part of this, we should remove the feature flag and use an ENV variable instead for enabling IDE mode.

chrstnb:
## TLDR

Rename the feature enablement flag to ide-mode-feature. Also add a toggle-able setting (via /ide command) to turn the feature on/off

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes #4393


packages/cli/src/config/config.ts
```diff
@@ -190,7 +190,7 @@ export async function parseArguments(): Promise<CliArgs> {
       type: 'boolean',
       description: 'List all available extensions and exit.',
     })
-    .option('ide-mode', {
+    .option('ide-mode-feature', {
```
jacob314, line 194: for discussion: should we deprecate this completely now that we have a setting? seems like this is something users shouldn't really care about configuring as a CLI arg and it is nice to keep are list of options small

packages/cli/src/config/settings.ts
```diff
@@ -99,7 +99,9 @@ export interface Settings {
 
   vimMode?: boolean;
 
-  // Add other settings here.
+  // Flag to be removed post-launch.
+  ideModeFeature?: boolean;
+  /// IDE mode setting configured via slash command toggle.
   ideMode?: boolean;
```
jacob314, line 105: why do we need both ideModeFeature and ideMode? I would think we could toggle on and off at runtime with a single ideMode setting.

packages/cli/src/ui/hooks/slashCommandProcessor.ts
```diff
@@ -55,6 +55,10 @@ export const useSlashCommandProcessor = (
 ) => {
   const session = useSessionStats();
   const [commands, setCommands] = useState<readonly SlashCommand[]>([]);
+  const [commandEpoch, setCommandEpoch] = useState(0);
```
jacob314: this is a bit of an antipattern. rather than incrementing the epoc to force a refresh can we get the configuration you care about stored so that it properly flows through react? happy to brainstorm some ways you might be able to do this that might be better aligned with react. if there isn't an easy way we can use this hack to unblock you.

jacob314 (commented):
Sorry if I'm missing something but can we accomplish this with just an ideMode setting rather than having both ideMode and ideModeFeature. I know the plan is to remove ideModeFeature post launch but would be nice if we never have to create it in the first place.

packages/cli/src/config/config.ts
chrstnb: @skeshive and I are in alignment on keeping it until we launch. We only want to show the feature to those who opt in until we're ready!

packages/cli/src/ui/hooks/slashCommandProcessor.ts
chrstnb: done! good call.

packages/cli/src/config/config.ts
jacob314: ok sounds good.

jacob314 (approved):
lgtm

dtometzki:
Hi together,

pr tested successfully

packages/cli/src/config/config.ts
```diff
@@ -258,13 +258,15 @@ export async function loadCliConfig(
       (v) => v === 'true' || v === '1',
     );
 
-  const ideMode =
-    (argv.ideMode ?? settings.ideMode ?? false) &&
+  const ideMode = settings.ideMode ?? true;
```
skeshive: should this be false by default?

skeshive (commented):
lets submit https://github.com/google-gemini/gemini-cli/pulls/chrstnb first please

packages/cli/src/config/config.ts
dtometzki: Hi @skeshive 

You can also set the ideMode setting at the project level.
To do this, create a file at the following path within your project directory:

  <PROJECT_ROOT>/.gemini/settings.json

  Settings in this project-based settings.json will override the global settings located in your user directory (~/.gemini/settings.json), but only for this specific project. This is useful if you want to enable IDE mode only for certain projects.
```
"ideMode": true,
"ideModeFeature": true,
```

Damian

Issue #4948, opened 2025-07-26
Unclear read_file Truncation Creates a Confusing User Experience


LyalinDotCom:
### What happened?

When the user asked the model to read a large file, the `read_file` tool returned a vague `(truncated)` status. This message provided **no detail on how much of the file was actually processed**, leaving the user uncertain about what information the model had access to. Despite this ambiguity, the model confidently, and likely incorrectly, confirmed it had read the file.

After the user questioned the truncation, the model stated it would read the entire file. However, the subsequent tool calls only returned the same vague `(truncated)` status, offering **no evidence that it was reading more of the file or making any progress**.

Finally, the model made the **unverifiable claim** that it had "now read the entire file." Given the complete lack of detailed feedback from the tool, the user was left with no confidence that the task was ever successfully completed, making the tool's state impossible to trust.

### What did you expect to happen?

The model and tool should handle large files more gracefully and transparently.

Some ideas:

1.  **Acknowledge Truncation Clearly:** If a file is too large to be read in full, the model should not just rely on a small `(truncated)` message. It should explicitly and clearly inform the user, for example: "The file is too large to read all at once, so I've read the first 2000 lines."
2.  **Provide User Control:** After a truncation, the model should offer the user actionable next steps, such as:
    * "Would you like me to read the next section of the file?"
    * "Should I try to summarize the portion I have read so far?"
3.  **Accurate State Reporting:** The model should never claim to have completed an action (like reading an entire file) when the tool's output indicates it was only partially successful. Its response should accurately reflect the tool's result.
4.  **Effective Problem-Solving:** When asked to read the full file after a truncation, the model should have a strategy to do so that is clear to the user on what is happening

### Client information

<details>

```console
$ gemini /about
# paste output here
```

</details>

jacob314:
Self assigning as joshualitt@google.com is taking it on.

joshualitt:
Okay, so we've landed both some prompt engineering, and some improvements to the UI. I think this is a really good start to addressing this bug. I'm sure there is more we can do here, but for now I'm going to close this, and we can target additional improvements in follow-ups.

PR #5155, opened 2025-07-30
bug(core): UI reporting for truncated read_file.

joshualitt:
BUG: #4948

## TLDR

Cleans up the UI messages around truncation.

## Dive Deeper

This is the initial change to tackle #4948. The next CL in this series will improve the prompt engineering.

## Reviewer Test Plan

There are new tests for this logic. I'm happy to add more. The change is fairly localized.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Makes progress on #4948

### DIFF
--- a/packages/core/src/utils/fileUtils.test.ts
+++ b/packages/core/src/utils/fileUtils.test.ts
@@ -420,7 +420,9 @@ describe('fileUtils', () => {
       expect(result.llmContent).toContain(
         '[File content truncated: showing lines 6-10 of 20 total lines. Use offset/limit parameters to view more.]',
       );
-      expect(result.returnDisplay).toBe('(truncated)');
+      expect(result.returnDisplay).toBe(
+        'Read lines 6-10 of 20 from test.txt (truncated)',
+      );
       expect(result.isTruncated).toBe(true);
       expect(result.originalLineCount).toBe(20);
       expect(result.linesShown).toEqual([6, 10]);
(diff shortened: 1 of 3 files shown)


packages/core/src/utils/fileUtils.ts
```diff
@@ -304,15 +304,32 @@ export async function processSingleFileContent(
 
         let llmTextContent = '';
         if (contentRangeTruncated) {
-          llmTextContent += `[File content truncated: showing lines ${actualStartLine + 1}-${endLine} of ${originalLineCount} total lines. Use offset/limit parameters to view more.]\n`;
+          llmTextContent += `[File content truncated: showing lines ${
+            actualStartLine + 1
+          }-${endLine} of ${originalLineCount} total lines. Use offset/limit parameters to view more.]\n`;
         } else if (linesWereTruncatedInLength) {
           llmTextContent += `[File content partially truncated: some lines exceeded maximum length of ${MAX_LINE_LENGTH_TEXT_FILE} characters.]\n`;
         }
```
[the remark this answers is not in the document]
joshualitt: @gemini-code-assist, I completely agree! I have a CL ready for fast follow-up to do just this. However, the reason I did not want to do it in this CL is to avoid making prompt changes alongside the UI changes. The prompt changes may need evals, or have unintended side-effects, whereas the UI changes are quite safe.

Thoughts? I'm happy to do the CLs at once, alternatively, we can land the prompt changes in a fast follow-up.

packages/core/src/utils/fileUtils.ts
```diff
@@ -310,9 +310,24 @@ export async function processSingleFileContent(
...
         llmTextContent += formattedLines.join('\n');
 
+        // By default, return nothing to streamline the common case of a successful read_file.
+        let returnDisplay = '';
+        if (contentRangeTruncated) {
+          returnDisplay = `Read lines ${
+            actualStartLine + 1
+          }-${endLine} of ${originalLineCount} from ${relativePathForDisplay} (truncated`;
```
jacob314: wonder if we really need the word truncated unless some lines were shortened.
If that is not the case then showing the range we read should be sufficient.

packages/core/src/utils/fileUtils.test.ts
```diff
@@ -465,7 +467,65 @@ describe('fileUtils', () => {
...
+        0,
+        5,
+      ); // Read 5 lines, but there are 11 total
+
       expect(result.isTruncated).toBe(true);
+      expect(result.returnDisplay).toBe(
+        'Read lines 1-5 of 11 from test.txt (truncated)',
+      );
+
+      // Scenario 2: Only line length truncation
```
jacob314: nit: ask gemini CLI to put each scenario in a separate test case rather than combining into a single test case.

jacob314 (commented):
Looks good! Approved once the two minor comments are addressed.

packages/core/src/utils/fileUtils.ts
joshualitt: SGTM, went ahead and used `(some lines were shortened)` in both long line cases for consistency.

jacob314 (approved):

PR #5202, opened 2025-07-30
do not mention GEMINI.md in system prompt as it is not fixed and can confuse model as it is not mentioned by memory tool and memory file paths are generally not exposed to model (yet)

olcan:


allenhutchison (approved):
Looks good.

PR #4401, opened 2025-07-17
Draft: Automatically record chat history in interactive and non-interactive mode; add interactive and non-interactive conversation resuming and a session browser

bl-ue:
# ⚠️ *This PR has been broken up into 8 smaller PRs:*
1. #5221 (MERGED)
2. #6721 (MERGED)
3. #7662 (MERGED)
4. #10719 (MERGED)
   a. #13243 (MERGED)
   b. #13520
5. #13300 (MERGED)
6. #13351 (MERGED)
7. #13621
8. #13667 

The original description, still completely relevant, is below.

---

> [!IMPORTANT]
> ✅ **All the functionality from this MR is now merged into main.**
---

In this PR, I—
- added automatic conversation recording to JSON files for interactive and non-interactive sessions (see below for details);
- created an interactive session browser with sorting, searching/filters, deleting, and resuming capabilities;
- implemented both interactive and non-interactive session resuming via the command line with `--resume <session_id>` (for a specific session) and `--resume` (for the latest one);
- configured a retention feature that automatically removes sessions based on age (older than a specific date) or quantity (when the total count of sessions exceeds a specific number),
- made `/clear` start a new session recording; and
- updated the documentation (`commands.md`, and added `chat-recording.md`).

Resuming modifies the session file in place.  It would not be difficult to make a configuration option to make it duplicate the session file instead.

# What's Done

In this PR I've added automatic chat history recording.  Starting with the first user message, we record:
- All user messages, with the timestamp and a UUID
- All Gemini messages, with
  - Timestamp
  - UUID
  - Thoughts (subject, description, timestamp)
  - Tool calls (ID, name, input, output, state, timestamp)
  - Token usage (input, output, cached, thoughts, tools)
  - Session ID
  - Project hash
  - Session start time
  - Model

Tool calls in all states are recorded, even while they're waiting to be confirmed.  API errors and info messages such as context compression notifications are also recorded, so the whole conversation log is preserved.  Each message is stored immediately; thus, if the CLI crashes or is killed, no messages are lost.

In the JSON, Gemini's thoughts are attached to their corresponding message object, and so, since they're generated _before_ the message itself be it a normal text message or tool calls, they're queued, not being flushed to the JSON until after the corresponding message is finished generating and then flushed.

[...]

PR #5221, opened 2025-07-30
feat(sessions): Introduce core ChatRecordingService for automatic conversation saving

bl-ue:
This PR is the first in **a series of 9 PRs** that implement automatic chat recording and resuming in both interactive and non-interactive modes, with command line flags for non-interactive resuming and session management, and an interactive session browser.

PR #4401 was the original monolithic PR, which implemented this feature from the ground up.  It's too large to merge as-is, however, and so it was broken up into 9 smaller PRs, of which this PR is the first.

Please see #4401's description for the full context of this PR.  The original monolithic PR's description contains a detailed analysis of the entire system, with an example of the recording JSON, 30+ linked issues, and a demo video.


gusbemacbe:
Good afternoon, @bl-ue!

I switched to this branch and could not resume the last chat session. I reviewed the `.gemini` folder and I noticed that I lost the session.

bl-ue:
Hi @gusbemacbe!  Yes, I'm sorry about that.  The maintainers requested that I break this PR up into 9 smaller PRs, so this PR is the first of 9 that implement automatic chat saving, and it only implements backend functionality.  Until these 9 PRs are merged, stay on the #4401 branch if you want to save your chats.

gusbemacbe:
Good evening, @bl-ue!

Thank you for your advice and tip!

Is it possible that you will keep the `auto-record-chat-history-071725` branch updated? Because the `gemini-cli` developers are updating daily by fixing the bugs and adding new features.

bl-ue:
@gusbemacbe Of course!  I'm really glad that you're testing it!  I'll keep updated.

packages/core/src/services/chatRecordingService.ts
```diff
@@ -0,0 +1,442 @@
...
+        const msg = this.newMessage(message.type, message.content);
+        if (msg.type === 'gemini') {
+          // If it's a new Gemini message then incorporate any queued thoughts.
+          conversation.messages.push({
+            ...msg,
+            thoughts: this.queuedThoughts,
+            tokens: this.queuedTokens,
+            model: this.config.getModel(),
+          });
+          this.queuedThoughts = [];
```
NTaylorMullen, line 218: Tell me more about the queued thoughts. What's the purpose of capturing those?

packages/core/src/services/chatRecordingService.ts
```diff
@@ -0,0 +1,442 @@
...
+          this.queuedTokens = null;
+        } else {
+          // Or else just add it.
+          conversation.messages.push(msg);
+        }
+      });
+    } catch (error) {
+      if (this.config.getDebugMode()) {
```
NTaylorMullen: You can error in this case even in non-debug mode. It's indicative of a bigger problem. Ditto elsewhere

packages/core/src/services/chatRecordingService.ts
```diff
@@ -0,0 +1,442 @@
...
+  /**
+   * Adds tool calls to the last message in the conversation (which should be by Gemini).
+   */
+  recordToolCalls(toolCalls: ToolCallRecord[]): void {
+    if (!this.conversationFile) return;
+
+    try {
+      this.updateConversation((conversation) => {
+        const lastMsg = this.getLastMessage(conversation);
+        // If a tool call was made, but the last message isn't from Gemini, it's because Gemini is
```
NTaylorMullen, line 290: So with all thsi magic why do the extra empty message creation? Wont this create a divergence in history for what is currently represented in the chat history vs. what we're saving?

packages/core/src/services/chatRecordingService.ts
```diff
@@ -0,0 +1,442 @@
...
+  resultDisplay?: string;
+  renderOutputAsMarkdown?: boolean;
+}
+
+/**
+ * Message type and message type-specific fields.
+ */
+export type ConversationRecordExtra =
+  | {
+      type: 'user' | 'system' | 'error';
```
NTaylorMullen: For now in this file is it possible to trim the types that aren't utilized yet? That way it should be easier to see when you put them into the system to better understand how they're utilized.

NTaylorMullen (changes requested):
You'll also want to write some tests for the chatRecordingService. Should be easy though if you let Gemini CLI run with it :)

packages/core/src/services/chatRecordingService.ts
bl-ue: It lets apps like [Gemini Desktop](https://github.com/Piebald-AI/gemini-desktop) display them, and potentially Gemini CLI itself in the future if we start displaying them.  I find them interesting to read and so do [others](https://github.com/Piebald-AI/gemini-desktop/issues/13#:~:text=The%20Chain%20of%20Thought%20(or%20whatever%20it%20is%20called%20nowadays%20at%20Google%27s%20side%2C%20the%20default%20Thinking%20Mode%3F)%20is%20visible!).  The ACP server inside Gemini CLI also sends them, so it seems consistent to save them.  Claude Code saves its reasoning too.

I've queued the thoughts because of the way I set up the JSON file; they need to be part of a message entry, but when it first starts thinking, Gemini has generated neither text nor tool calls so there isn't a message added yet.  I could've created a new `gemini` message and directly added the thoughts instead of queuing them, and right now, I can't remember if there were technical constraints that motivated this decision.

Another alternative is to use JSONL instead of JSON so we can append new thoughts immediately to the file without explicitly associating them with the subsequent `gemini` message.

packages/core/src/services/chatRecordingService.ts
bl-ue: If I understand you correctly, no.  We're storing tool calls in a `gemini` message, but when a user sends a message, Gemini can skip a text response and just go straight to tool calling, in which case we need an empty message to contain those tool calls.  JSONL would also solve this problem if you don't like this empty message solution.

There shouldn't be any divergence between the saved conversation data and the UI history because when I load the saved conversation into the UI, I skip that empty message and turn the tool calls into a tool group.  In [`convertSessionToHistoryFormats`](https://github.com/google-gemini/gemini-cli/pull/4401/files#diff-20913b5d552113e00114c61e59132870c8437d9b21ca425eb460486d27b2daaeR88-R257) from the original full PR, I only add non-empty messages from either side to the UI:
```ts
...
  for (const msg of messages) {
    // Add the message only if it has content
    if (msg.content && msg.content.trim()) {            // <--------
      let messageType: MessageType;
      switch (msg.type) {
        case 'user':
...
```

packages/core/src/services/chatRecordingService.ts
bl-ue: Good thinking, done.

packages/core/src/services/chatRecordingService.ts
bl-ue: Right, fixed.

bl-ue:
Hi @NTaylorMullen!  You probably have tons of PRs but I don't want this one to get lost in the shuffle.  Any chance you can look at this one this week?

NTaylorMullen:
> Hi @NTaylorMullen! You probably have tons of PRs but I don't want this one to get lost in the shuffle. Any chance you can look at this one this week?

Thank you for mentioning me directly! Yes I get hundreds per day lol. Looking now!

packages/core/src/services/chatRecordingService.ts
NTaylorMullen: Your reasoning makes senes. Thank you!

NTaylorMullen (approved):
Alright, PR 1, getting in!

@bl-ue to help make sure I don't miss things like this (so sorry about the delay) feel free to shoot me an email (it's ntaylormullen (my GH handle 😄 ) at google.com and I can send you an invite to a chat space which should be easier to poke me / the team at.

NTaylorMullen:
Ah looks like you might have a linting error. You should be able to fix by doing `npm run format` and pushing the branch

bl-ue:
Hi @NTaylorMullen!  Okay, sounds great.  I resolved the linting error, so it should be ready.  I also sent you an email with a subject of "PR #5221 on Gemini CLI".

NTaylorMullen:
@bl-ue looks like you may have a few test failures!

bl-ue:
Hi @NTaylorMullen, okay, fixed.  The `mkdirSync` call in `ChatRecordingService.initialize` was getting through to the FS directly because it wasn't being mocked for all the `ChatRecordingService` tests.  It succeeded on Windows and not macOS and Linux because `/home/project/...` was getting converted to `C:\home\project\...` and Windows has looser permissions than macOS/Linux around creating items in the root directory (`C:\` and `/`), so the mkdir call was succeeding on Windows but not the other platforms.  We fixed it by mocking `mkdirSync` for the entire `ChatRecordingService` test suite.  The entire test suite now passes on all three platforms.

bl-ue:
Next PR: #6721

PR #61, opened 2025-04-20
Minimal container setup. Install docker (or podman), build container with scripts/build_container.sh, then start with scripts/start_container.sh. Exit with ^C for now.

olcan:

PR #5222, opened 2025-07-30
[Fix Telemetry for tool calls, PR 1/n] Propagate tool reported errors via ToolCallResponseInfo and ToolResult

silviojr:
## TLDR

Our logging treats tool responses as SUCCESS if there was no exception raised. Here we specifically check for errors, including the ones that are exceptions handled inside the tool. 

The changes ensure that tool execution results, including specific error types, are properly propagated through the system and logged for better telemetry, while also refining CLI behavior to exit only for critical unhandled exceptions.

I will modify the tools in the next PRs. 
This PR makes progress on [#61](https://github.com/google-gemini/maintainers-gemini-cli/issues/61)

## Dive Deeper

It introduces a standardized way to categorize tool errors using an errorType field in both **ToolCallResponseInfo** and  **ToolResult** . This ensures that errors inside the tool calls are propagated instead of just being caught inside tools. 

 The changes ensure that tool execution results, including specific error types, are properly propagated through the system and logged for better telemetry, while also refining CLI behavior to exit only for critical unhandled exceptions for the **nonInteractiveCLI**.

**Standardized Error Categorization:** I've introduced an errorType field to the ToolCallResponseInfo and ToolResult interfaces. This allows for standardized categorization of tool execution errors, such as TOOL_NOT_REGISTERED or GENERIC_EXCEPTION. GENERIC_EXCEPTION is meant to represent any exception that was caught in a try...catch block and not treated/handled. 

**Improved Tool Execution Error Handling:** The CoreToolScheduler and nonInteractiveToolExecutor now explicitly check for and propagate tool-reported errors, distinguishing them from general exceptions. This ensures that structured error information is passed through the system.

**Enhanced Telemetry:** Tool call telemetry has been updated to include detailed error messages and the new error_type for failures. This provides richer insights into tool execution outcomes and aids in debugging.

**Refined NONinteractiveCLI Exit Behavior:** The non-interactive CLI's exit logic has been refined. It will now only trigger process.exit(1) for UNHANDLED_EXCEPTION errors originating from tools, making the CLI more resilient to other types of tool failures.

**Adapted tests to the new changes**

## Reviewer Test Plan

e2e tests and unit tests.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | YES  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

This PR makes progress on [#61](https://github.com/google-gemini/maintainers-gemini-cli/issues/61)


packages/core/src/core/turn.ts
```diff
@@ -76,6 +76,7 @@ export interface ToolCallResponseInfo {
   responseParts: PartListUnion;
   resultDisplay: ToolResultDisplay | undefined;
   error: Error | undefined;
+  errorType: string | undefined;
```
jacob314: The PR uses string literals for errorType (e.g., 'GENERIC_EXCEPTION', 'TOOL_NOT_REGISTERED'). To prevent typos can we define these as a TypeScript enum or a union of allowed strings constraining what errorType can be?

packages/core/src/tools/tools.ts
```diff
@@ -217,6 +217,14 @@ export interface ToolResult {
    * For now, we keep it as the core logic in ReadFileTool currently produces it.
    */
   returnDisplay: ToolResultDisplay;
+
+  /**
+   * If this property is present, the tool call is considered a failure.
+   */
+  error?: {
+    message: string; // raw error message
+    type?: string; // An optional machine-readable error type (e.g., 'FILE_NOT_FOUND').
```
jacob314: would it be possible to restrict this to an enum or a union of valid error type strings?

jacob314 (commented):
Looks good overall. Approved after these minor comments are addressed.
If it is not feasible to restrict error type to an enum or union of string literals, can we at least define string constants for the common error types mentioned so matching against those can be more robust and less prone to typos.

packages/core/src/core/turn.ts
silviojr: Yes, let me do that!

packages/core/src/tools/tools.ts
silviojr: Good point!

jacob314 (approved):
lgtm

jacob314:
Unfortunately there are some conflicts that need to be resolved. As soon as they are resolved I will review again and you can land.

silviojr:
> Unfortunately there are some conflicts that need to be resolved. As soon as they are resolved I will review again and you can land.

REBASED! Thanks for the quick review!

jacob314 (approved):
lgtm

PR #5225, opened 2025-07-30
Update built-in commands to avoid collisions with custom commands

haroldmciver-go:
## TLDR
This PR is a follow-up to and should be merged after https://github.com/google-gemini/gemini-cli/pull/5130 
File-based commands now always take priority over built-in commands. Instead of being overwritten, conflicting built-in commands are automatically renamed (e.g., /help becomes /builtin:help) so that all commands remain accessible.

## Dive Deeper

Previously, slash command names from file-based, built-in, and MCP sources could conflict, leading to unpredictable behavior due to undefined loading order.

This update refactors the command loading logic to establish a deterministic precedence by a defined command. Commands now load with the following priority:
1. File-based Commands: Highest priority (e.g., .gemini/commands/*.toml).
2. Built-in Commands: Standard CLI commands.
3. MCP-backed Commands: Lowest priority (from remote servers).

When a name conflict occurs, the lower-priority command is automatically renamed instead of being overwritten. Built-in commands conflicting with file-based commands are prefixed with builtin: (e.g., /help becomes /builtin:help). A displayName property is added to renamed commands to maintain their original name in UI elements like autocomplete.

## Reviewer Test Plan

Test Collision with a Built-in and Custom Command:

Create a local file-based command named help (e.g., in ~/.gemini/commands/help.toml).

Run the CLI and execute /mcp.

Verify: The Built-in command has been renamed to /builtin:help to avoid conflicting with the custom command.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


jacob314 (commented):
If we decide to go with this priority order then
`/help` needs to be updated to properly designate the builtin commands by `/builtin:someName` when a local tool command takes priority over them.

@abhipatel12 curious on your thoughts about whether local file commands should take priority over builtin commands like this. Seems generally reasonable to me but could also be a bit surprising. Particularly if we in the future provide easy ways for people to install bundles of builtin file commands then it might become confusing that they take priority over the builtin commands.

abhipatel12:
That's true. We can make`/help` dynamic using the `SlashCommand.name`. 

In general, I'm still on the fence on the best way to deal with name collisions. I've seen a few options with various prefixes, and I'm not sure. 

In this case, do we dedupe from the same source, and then prepend a prefix for name collisions between sources? Sorry if i missed it in the code.

chrstnb:

PR #5226, opened 2025-07-30
Exclude companion extension from release versioning

skeshive:
## TLDR
We'll be releasing the extension manually for the first few times so excluding the extension from the automated versioning.

Verified locally that versions packages/core/package.json, packages/cli/package.json, and gemini-cli/package.json still get bumped with this change

### DIFF
--- a/packages/vscode-ide-companion/package.json
+++ b/packages/vscode-ide-companion/package.json
@@ -2,7 +2,7 @@
   "name": "gemini-cli-vscode-ide-companion",
   "displayName": "Gemini CLI Companion",
   "description": "Enable Gemini CLI with direct access to your VS Code workspace.",
-  "version": "99.99.99",
+  "version": "0.0.1",
   "publisher": "google",
   "icon": "assets/icon.png",
   "repository": {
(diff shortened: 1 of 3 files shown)


packages/vscode-ide-companion/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "gemini-cli-vscode-ide-companion",
   "displayName": "Gemini CLI Companion",
   "description": "Enable Gemini CLI with direct access to your VS Code workspace.",
-  "version": "0.1.15",
+  "version": "99.99.99",
```
jacob314: nit: I would be careful introducing a version number this large anywhere in the package.json as if anything goes wrong and it is accidentally published by someone now you have a lot of pain unless you raise the version to 100 or retract this version which could still make the version history a bit confusing.
I would suggest 0.0.1 or 0.0.0 if that is allowed to indicate an invalid version.

skeshive: ah fair- updated to 0.0.1!

jacob314 (approved):
lgtm

PR #5237, opened 2025-07-30
confirm save_memory tool, with ability to see diff and edit manually for advanced changes that may override past memories

olcan:
screenshot:
image

### DIFF
--- a/packages/core/src/tools/memoryTool.ts
+++ b/packages/core/src/tools/memoryTool.ts
@@ -4,11 +4,21 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { BaseTool, Icon, ToolResult } from './tools.js';
+import {
+  BaseTool,
+  ToolResult,
+  ToolEditConfirmationDetails,
+  ToolConfirmationOutcome,
+  Icon,
+} from './tools.js';
 import { FunctionDeclaration, Type } from '@google/genai';
 import * as fs from 'fs/promises';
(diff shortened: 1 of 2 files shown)


NTaylorMullen (approved):

packages/core/src/tools/memoryTool.ts
```diff
@@ -110,6 +127,111 @@ export class MemoryTool extends BaseTool<SaveMemoryParams, ToolResult> {
...
+      const error = err as Error & { code?: string };
+      if (!(error instanceof Error) || error.code !== 'ENOENT') throw err;
+      return '';
+    }
+  }
+
+  /**
+   * Computes the new content that would result from adding a memory entry
+   */
+  private computeNewContent(currentContent: string, fact: string): string {
```
NTaylorMullen, line 151: Double check that we weren't doing this before in the memoryUtil piece. I.e. how did this worrk before hand?

olcan: this is just a refactoring of the previous code for appending to memory

allenhutchison (approved):

allenhutchison:
The tests need to be updated to use platform agnostic path separators. Can you do that in this PR?

olcan (commented):
> The tests need to be updated to use platform agnostic path separators. Can you do that in this PR? 

yes, was just looking at windows failures. will fix then merge

Issue #5077, opened 2025-07-28
/theme constantly re-renders


sethtroisi:
### What happened?

`/theme` renders dozens of times a seconds.

### What did you expect to happen?

stable rendering

### Client information

local dev client 1.13

PR #5242, opened 2025-07-30
Add render counter in debug mode

sethtroisi:
## TLDR

Nice for debugging issues like #5077 

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

### DIFF
--- a/packages/cli/src/ui/components/Footer.tsx
+++ b/packages/cli/src/ui/components/Footer.tsx
@@ -13,6 +13,8 @@ import process from 'node:process';
 import Gradient from 'ink-gradient';
 import { MemoryUsageDisplay } from './MemoryUsageDisplay.js';
 
+import { DebugProfiler } from './DebugProfiler.js';
+
 interface FooterProps {
   model: string;
   targetDir: string;
@@ -48,6 +50,7 @@ export const Footer: React.FC<FooterProps> = ({
   return (
     <Box justifyContent="space-between" width="100%">
       <Box>
+        {debugMode && <DebugProfiler />}
         {vimMode && <Text color={Colors.Gray}>[{vimMode}] </Text>}
         {nightly ? (
           <Gradient colors={Colors.GradientColors}>
--- a/packages/cli/src/ui/components/DebugProfiler.tsx
+++ b/packages/cli/src/ui/components/DebugProfiler.tsx
@@ -1,0 +1,32 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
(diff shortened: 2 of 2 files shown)


sethtroisi:
Screenshot: image

packages/cli/src/ui/components/Footer.tsx
```diff
@@ -41,13 +42,17 @@ export const Footer: React.FC<FooterProps> = ({
...
   vimMode,
+  numRenders,
 }) => {
   const limit = tokenLimit(model);
   const percentage = promptTokenCount / limit;
 
   return (
     <Box justifyContent="space-between" width="100%">
       <Box>
+        {debugMode && numRenders > 0 && (
```
jacob314: I'm not sure all users of debug mode want to see this. Can we put it behind a setting or require you to press a keyboard shortcut to enable it. I'd suggest maybe F12 as that is the keyboard shortcut to typically bring up devtools.

packages/cli/src/ui/App.tsx
```diff
@@ -135,6 +135,12 @@ const App = ({ config, settings, startupWarnings = [], version }: AppProps) => {
     consolePatcher.patch();
     registerCleanup(consolePatcher.cleanup);
   }, [handleNewMessage, config]);
+  const numRenders = useRef(0);
+  useEffect(() => {
```
jacob314: lets move this logic out of footer and avoid prop drilling. instead create a DebugProfiler component that is only added to footer if getDebugMode() is true. That component that then track numRenders and handle rendering it to the footer all in one place.

App.tsx is quite big and complex so anything we can do to keep it from growing bigger and more complex is worth it.

jacob314 (commented):
Looking forward to using this feature! Approved once these hopefully minor refactors that Gemini CLI should be able to one shot are done. Really want to catch some cases where we are rendering too many times. I think I've seen a few of them.

packages/cli/src/ui/App.tsx
sethtroisi: Done

packages/cli/src/ui/components/Footer.tsx
sethtroisi: Done.

jacob314 (approved):
lgtm

PR #5257, opened 2025-07-31
Replace FlashDecidedToContinueEvent with NextSpeakerCheckEvent

SandyTao520:
## TLDR

Replace FlashDecidedToContinueEvent with NextSpeakerCheckEvent

## Dive Deeper

Needed to analyze the effectiveness of next speaker check.

### DIFF
--- a/packages/core/src/telemetry/types.ts
+++ b/packages/core/src/telemetry/types.ts
@@ -266,15 +266,19 @@ export class LoopDetectedEvent {
   }
 }
 
-export class FlashDecidedToContinueEvent {
-  'event.name': 'flash_decided_to_continue';
+export class NextSpeakerCheckEvent {
+  'event.name': 'next_speaker_check';
   'event.timestamp': string; // ISO 8601
   prompt_id: string;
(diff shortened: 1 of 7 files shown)


packages/core/src/telemetry/types.ts
```diff
@@ -266,15 +266,19 @@ export class LoopDetectedEvent {
...
+  finish_reason: string;
+  result: string;
 
-  constructor(prompt_id: string) {
-    this['event.name'] = 'flash_decided_to_continue';
+  constructor(prompt_id: string, finish_reason: string, result: string) {
+    this['event.name'] = 'next_speaker_check';
     this['event.timestamp'] = new Date().toISOString();
     this.prompt_id = prompt_id;
+    this.finish_reason = finish_reason;
```
NTaylorMullen, line 280: lets do `finishReason` here and in the constructor param. the `prompt_id` is camel case legacy, we need to change that at some point, it's bad TS

NTaylorMullen (approved):

Issue #5209, opened 2025-07-30
Evaluate and track error rate of failures of the editing tool


silviojr:
### What would you like to be added?

We need to verify what is the error rate of the current editing tool so we have a baseline and a way to measure progress. 

### Why is this needed?

The editing tool is one of the most critical tools in gemini cli. We need to track its performance.

PR #5292, opened 2025-07-31
New Replace Strategies to the Edit tool 

silviojr:
## TLDR

Introducing multiple strategies for performing code **REPLACEMENT**.  We are testing those strategies separately and together. 
- Exact: do a simple search and replace
- Fuzzy: just loose the restrictions to match a string. Originally inspired by https://github.com/googlecloud-appeco-incubator/agent-prototypes/blob/main/agent_prototypes/code_editing/fuzzy_edit.py 
- Fuzzy v2: Inspired by https://www.npmjs.com/package/@modelcontextprotocol/server-filesystem?activeTab=code 
- Corrector: the **CURRENT** implementation of the tool in the core tools of the GCLI. Uses LLM to change the old_string and new_string to perform the search and replace. 

I made those strategies CONFIGURABLE, however the configuration is only available through environment variable. We will do more tests BEFORE we confirm which one is the best. We are also testing with other approaches, including with other editors available through MCP servers. 

## Dive Deeper

* **packages/cli/src/config/config.ts**
    * strategies are available ONLY through the `GEMINI_REPLACE_STRATEGY` environment variable.

* **packages/core/src/tools/edit.ts**
    * Introduced the `ReplaceStrategy` enum, defining `EXACT`, `CORRECTOR`, `FUZZY`, `FUZZY_V2`, and `COMPOSITE` edit modes.
    * Implemented distinct classes for each `ReplaceStrategy` (e.g., `ExactStrategy`, `CorrectorStrategy`, `FuzzyStrategy`, `CompositeStrategy`) to encapsulate their respective editing logic.
    * Refactored the `EditTool`'s `calculateEdit` method to utilize a strategy pattern, dynamically selecting and executing the appropriate `ReplaceStrategy` based on the configured mode.
    * Added a `validateEditResult` function to centralize and standardize the validation of edit operation outcomes.
    * Improved error handling for various scenarios, including file not found and attempts to create existing files.
    * Added a console log for now to log the strategy used for debugging... probably should remove it? 
* **packages/core/src/tools/tool-error.ts**
    * Added a new `EDIT_PARAMS_THE_SAME` error type to indicate when the old and new strings in an edit operation are identical. This is to differentiate from the error EDIT_NO_CHANGE . Notice that adding this error may increase the error rate... so the question is: is this actually something we want to consider an ERROR? It was not an error from the tool itself, but from the CALLER. 
    
## Reviewer Test Plan

Updated the relevant tests and I am running all strategies PLUS some mcp editor servers to solve known software engineering benchmarks to measure the error rate of all strategies. 

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | YES  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs
This PR makes progress on #5211 and #5209

### DIFF
--- a/packages/core/src/tools/edit.ts
+++ b/packages/core/src/tools/edit.ts
@@ -32,17 +32,15 @@ import { GeminiClient } from '../core/client.js';
 /**
  * Defines the available strategies for the edit tool.
  */
-export enum EditMode {
+export enum ReplaceStrategy {
   /** Strict, exact string replacement. */
   EXACT = 'exact',
   /** Uses the AI-powered corrector to find the best match. This is the default. */
   CORRECTOR = 'corrector',
   /** A flexible, line-by-line match that ignores whitespace. */
   FUZZY = 'fuzzy',
-  /** A modified version of the fuzzy match with different indentation handling. */
-  MODIFIED_FUZZY = 'modified_fuzzy',
-  /** Tries all strategies from strictest to most lenient, stopping at the first success. */
-  ALL = 'all',
+  /** Tries a sequence of strategies, stopping at the first success. */
+  COMPOSITE = 'composite',
 }
 
 /**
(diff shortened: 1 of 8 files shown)


packages/core/src/tools/edit.ts
```diff
@@ -23,10 +23,28 @@ import { SchemaValidator } from '../utils/schemaValidator.js';
...
+ */
+export enum ReplaceStrategy {
+  /** Strict, exact string replacement. */
+  EXACT = 'exact',
+  /** Uses the AI-powered corrector to find the best match. This is the default. */
+  CORRECTOR = 'corrector',
+  /** A flexible, line-by-line match that ignores whitespace. */
+  FUZZY = 'fuzzy',
+
+  FUZZY_V2 = 'fuzzy_v2',
```
anj-s, line 43: what is fuzz_v2?

packages/core/src/tools/edit.ts
```diff
@@ -67,6 +85,403 @@ interface CalculatedEdit {
...
+      newContent,
+      occurrences,
+      finalOldString,
+      finalNewString,
+      mode: this.mode,
+    };
+  }
+}
+
+class FuzzyStrategyV2 implements ReplaceStrategyImpl {
```
agmsb, line 164: Could we factor out shared logic between Fuzzy versions?

silviojr:
Closing this in favor of https://github.com/google-gemini/gemini-cli/pull/6823

PR #5315, opened 2025-07-31
fix(core): prevent UI shift after vim edit

### LINKED ISSUE #862: UI "shifted" after exiting "Modify edit with vim"
**Describe the bug**

- When edit is triggered, users have the option to modify the proposed diff with vim.
- After exiting the edit, the Edit box "shifted up".

- Before edit:
Image

- After edit:
Image

**Additional context**
- Modify with VS Code does not cause this effect.
- It looks like an issue where an "Enter /  line feed" is added above the InputPrompt, pushing the box higher.

galdawave:
## TLDR

This pull request fixes a minor UI bug where the "Edit" component would shift upwards after a user modified a diff using an external editor like `vim`. The fix involves introducing a callback that refreshes the entire UI after the editor process closes, ensuring the UI remains stable and corrects any visual artifacts left by the editor.

## Dive Deeper

When a user opts to edit a proposed code change, Gemini launches the selected editor as a synchronous child process. Some command-line editors, notably Vim, can alter the terminal's rendering state or leave artifacts upon exit. This was causing the entire UI block to jump up one line.

This PR implements a robust solution by introducing an `onEditorClose` callback. This callback is passed down through the tool scheduling layers. When the `execSync` call for the editor completes (either successfully or with an error), the `onClose` callback is triggered in a `finally` block. This callback is wired to a function in the main `App` component that forces a re-render of the entire UI, effectively clearing any visual glitches and ensuring the layout remains consistent.

## Reviewer Test Plan

To validate this change, a reviewer can follow these steps to reproduce the bug and verify the fix:

1.  Pull down this branch.
2.  Run the CLI using `npm run start`.
3.  Enter a prompt that generates a code change, for example: `write a simple hello world function`.
4.  When the diff view appears, select the option to "Modify edit with editor".
5.  Choose `vim` as the editor.
6.  Without making any changes, save and quit Vim by typing `:wq` and pressing Enter.
7.  **Verification:** Observe that the UI does **not** shift upwards after Vim closes. The input prompt should remain in its original position.
8.  To see the original buggy behavior, check out the `main` branch and repeat the steps. You will notice the UI jumps up one line after Vim exits.

## Testing Matrix

|          | ✅ |  |  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes #862

### DIFF
--- a/packages/cli/src/ui/hooks/useGeminiStream.ts
+++ b/packages/cli/src/ui/hooks/useGeminiStream.ts
@@ -94,6 +94,7 @@ export const useGeminiStream = (
   performMemoryRefresh: () => Promise<void>,
   modelSwitchedFromQuotaError: boolean,
   setModelSwitchedFromQuotaError: React.Dispatch<React.SetStateAction<boolean>>,
+  onEditorClose: () => void,
 ) => {
   const [initError, setInitError] = useState<string | null>(null);
   const abortControllerRef = useRef<AbortController | null>(null);
@@ -134,6 +135,7 @@ export const useGeminiStream = (
       config,
       setPendingHistoryItem,
       getPreferredEditor,
+      onEditorClose,
     );
 
   const pendingToolCallGroupDisplay = useMemo(
--- a/packages/cli/src/ui/App.tsx
+++ b/packages/cli/src/ui/App.tsx
@@ -500,6 +500,7 @@ const App = ({ config, settings, startupWarnings = [], version }: AppProps) => {
     performMemoryRefresh,
     modelSwitchedFromQuotaError,
     setModelSwitchedFromQuotaError,
+    refreshStatic,
   );
 
   // Input handling
--- a/packages/cli/src/ui/hooks/useReactToolScheduler.ts
+++ b/packages/cli/src/ui/hooks/useReactToolScheduler.ts
@@ -70,6 +70,7 @@ export function useReactToolScheduler(
     React.SetStateAction<HistoryItemWithoutId | null>
   >,
   getPreferredEditor: () => EditorType | undefined,
+  onEditorClose: () => void,
 ): [TrackedToolCall[], ScheduleFn, MarkToolsAsSubmittedFn] {
   const [toolCallsForDisplay, setToolCallsForDisplay] = useState<
     TrackedToolCall[]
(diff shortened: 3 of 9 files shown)


packages/cli/src/ui/hooks/useGeminiStream.ts
```diff
@@ -94,6 +94,7 @@ export const useGeminiStream = (
   performMemoryRefresh: () => Promise<void>,
   modelSwitchedFromQuotaError: boolean,
   setModelSwitchedFromQuotaError: React.Dispatch<React.SetStateAction<boolean>>,
+  onEditorClose?: () => void,
```
jacob314: all the other arguments are not nullable. should we make this one not nullable as well? would avoid errors if this arg is omitted somewhere we want it.

jacob314 (commented):
looks good. can you also add some tests for this functionality?

packages/cli/src/ui/hooks/useGeminiStream.ts
galdawave: Done.

galdawave:
> looks good. can you also add some tests for this functionality?

Added some unit tests

packages/cli/src/ui/hooks/useGeminiStream.ts
```diff
@@ -94,6 +94,7 @@ export const useGeminiStream = (
   performMemoryRefresh: () => Promise<void>,
   modelSwitchedFromQuotaError: boolean,
   setModelSwitchedFromQuotaError: React.Dispatch<React.SetStateAction<boolean>>,
+  onEditorClose: () => void,
```
jacob314, line 97: as a follow up pull request can you switch these to all named arguments as there are a ton of them. 
in particular, onEditorClose kindof sneaks in in app.tsx
would be better to have 
`onEditorClose: refreshStatic`
in app.tsx to make it more clear that refreshStatic is a hack we are calling to work around editors messing up the UI

jacob314 (approved):
lgtmrefreshStatic is very heavy weight. I'm fine with landing this now but can we follow up with a pr to only

PR #5321, opened 2025-07-31
feat(cli): Improve @ autocompletion for mid-sentence edits

SandyTao520:
## TLDR

Support at auto-completion in the middle of a sentence.

## Dive Deeper

The previous implementation of the file path autocompletion was based on finding the last `@` symbol in the input, which was not always accurate. This refactoring introduces a more sophisticated approach:

*   **`commandIndex`**: This is used to find the start of the current `@` command, taking into account the cursor's position.
*   **`completionStart` and `completionEnd`**: These refs are used to precisely define the range of text that needs to be replaced by the autocompletion suggestion. This allows for accurate completion even when the cursor is in the middle of a file path or when there are other characters after the path.
*   **Improved Tests**: The unit tests for `useCompletion` have been updated to use a real `useTextBuffer` instance, which makes them more realistic and reliable. New tests have been added to cover the scenarios with multiple `@` commands and with the cursor at different positions.

These changes make the file path autocompletion feature more predictable and user-friendly.

## Reviewer Test Plan

To test these changes, you can run the Gemini CLI and try the following scenarios:

1.  **Basic file path completion:**
    *   Type `@` and start typing a file or directory name. Verify that the suggestions are correct.
2.  **Completion with cursor in the middle of the path:**
    *   Type `@somet/path`
    *   Move the cursor back to the middle of the path (e.g., after `somet`).
    *   Trigger autocompletion and verify that it correctly completes the path.
3.  **Completion with multiple `@` commands:**
    *   Type `@file1.txt @some/other/path`
    *   Move the cursor to the second `@` command and trigger autocompletion. Verify that it completes the second path without affecting the first one.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ✅  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

N/A

### DIFF
--- a/packages/cli/src/ui/hooks/useCompletion.test.ts
+++ b/packages/cli/src/ui/hooks/useCompletion.test.ts
@@ -1148,7 +1148,7 @@ describe('useCompletion', () => {
               name: 'show',
               description: 'Show memory',
             },
-            { 
+            {
               name: 'add',
               description: 'Add to memory',
             },
--- a/packages/cli/src/ui/hooks/useCompletion.test.ts
+++ b/packages/cli/src/ui/hooks/useCompletion.test.ts
@@ -14,7 +14,7 @@ import * as path from 'path';
 import * as os from 'os';
 import { CommandContext, SlashCommand } from '../commands/types.js';
 import { Config, FileDiscoveryService } from '@google/gemini-cli-core';
-import { useTextBuffer, TextBuffer } from '../components/shared/text-buffer.js';
+import { useTextBuffer } from '../components/shared/text-buffer.js';
 
 describe('useCompletion', () => {
   let testRootDir: string;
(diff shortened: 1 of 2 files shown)


jacob314:
This is a really solid PR! Thanks for putting together such a clear description and test plan—it makes reviewing much easier. I'm especially impressed with the test refactoring. Swapping the mock TextBuffer for a real useTextBuffer instance is a huge improvement and makes the tests much more reliable. The new test cases for mid-sentence completion and multiple at commands are great additions as well. The core logic changes in the useCompletion hook are well thought out. Using commandIndex to find the active command and the completionStart/completionEnd refs to define the replacement range is a clever and effective approach. Just one small request: could you update the PR title to follow our Conventional Commits standard? Something like 'feat(cli): Improve @ autocompletion for mid-sentence edits' would fit perfectly. Awesome work on this

jacob314 (approved):
lgtm

PR #5325, opened 2025-07-31
fix(core): Remove json output schema form the next speaker check prompt

SandyTao520:
## TLDR

Removing this as passing RESPONSE_SCHEMA in generateContent should be sufficient. Also Clearcut log how often this happens.

## Dive Deeper

The flash lite model seems to be confused by this part in the prompt and ignores our provided response schema.


NTaylorMullen (approved):
This looks fine but please run evals before merging all the way.

PR #5328, opened 2025-08-01
Fix line end bugs in Vim mode

### LINKED ISSUE #5322: [Vim mode] Line boundary handling when moving from a shorter line to a longer one
### What happened?

```
short
long line
```

In NORMAL mode, moving the cursor up one line from "e" will land on right of "t".

The cursor should land on "t".

### Client information

<details>

```console
$ gemini /about
# paste output here
```

</details>

### Login information

_No response_

_No response_

### LINKED ISSUE #5326: [Vim mode] Word end detection bug when the last line is an empty line
### What happened?

```
hello world

```

- When the cursor is on 'w', `e` once moves it to 'd' as expected, but `e` again moves it to the right of 'd'.
- When the cursor is on 'w', `w` moves it to the right of 'd'.

- The second `e` should move the cursor to the beginning of the next line.
- `w` should move the cursor to the beginning of the next line.

### Client information

<details>

```console
$ gemini /about
# paste output here
```

</details>

### Login information

_No response_

_No response_

sijieamoy:
## TLDR
Vim mode bug of handling line end when the next and last line is empty.

## Dive Deeper
- See https://github.com/google-gemini/gemini-cli/issues/5326 and https://github.com/google-gemini/gemini-cli/issues/5322
- Missing test cases for both

## Reviewer Test Plan
1. Enable Vim mode
2. Enter "hello world" followed by an empty line
3. Position the cursor on "w" then:
  1. `w` once;
  2.  `e` twice
4. In both cases, the cursor should move to the beginning of the empty line.

## Linked issues / bugs
- Fixes https://github.com/google-gemini/gemini-cli/issues/5326 
- Fixes https://github.com/google-gemini/gemini-cli/issues/5322


packages/cli/src/ui/components/shared/vim-buffer-actions.ts
```diff
@@ -674,7 +765,64 @@ export function handleVimAction(
...
       for (let i = 0; i < count; i++) {
-        offset = findWordEnd(getText(), offset);
+        const newOffset = findWordEnd(getText(), offset);
+
+        // Special handling for the first iteration when we're at end of word
+        if (i === 0) {
+          const currentLine = lines[cursorRow] || '';
+          const atEndOfWord =
+            cursorCol < currentLine.length &&
+            /\w/.test(currentLine[cursorCol]) &&
```
jacob314: you need to use the codepoint helpers for all this logic to robustly handle unicode characters

packages/cli/src/ui/components/shared/vim-buffer-actions.ts
```diff
@@ -18,6 +18,46 @@ import {
 } from './text-buffer.js';
 import { cpLen } from '../../utils/textUtils.js';
 
+// Helper function to handle word movement across empty lines.
+function findNextWordAcrossLines(
+  getText: () => string,
```
jacob314: consider rewriting this logic to not require getText(). Same for the rest of the file. that will make it more robust as less conversion between lines and rows and columns would be needed. gemini cli may be able to handle that correctly with a oneshot.

packages/cli/src/ui/components/shared/vim-buffer-actions.ts
sijieamoy: Fixed

packages/cli/src/ui/components/shared/vim-buffer-actions.ts
sijieamoy: Done.

packages/cli/src/ui/components/shared/text-buffer.ts
```diff
@@ -33,143 +33,229 @@ function isWordChar(ch: string | undefined): boolean {
...
 }
 
-// Vim-specific word boundary functions
-export const findNextWordStart = (
-  text: string,
-  currentOffset: number,
-): number => {
-  let i = currentOffset;
+// Helper functions for line-based word navigation
+export const isWordCharStrict = (char: string): boolean => /\w/.test(char);
```
jacob314: does this have to be different from isWordChar in text-buffer.ts?
If not, would be ideal to converge. I'm open to changing the existing method if it should be slighty different to align with vim.

packages/cli/src/ui/components/shared/text-buffer.ts
```diff
@@ -33,143 +33,229 @@ function isWordChar(ch: string | undefined): boolean {
...
+
+export const isWhitespace = (char: string): boolean => /\s/.test(char);
+
+// Find next word start within a line, starting from col
+export const findNextWordStartInLine = (
+  line: string,
+  col: number,
+): number | null => {
+  const chars = [...line]; // Handle Unicode properly
```
jacob314: here and elsewhere please use the existing toCodePoints
helper in textUtils. which does the same thing to keep how utf-32 code point handling consistent.

jacob314 (commented):
Looks good! Please also add a couple of tests that verify that these methods behave correctly given UTF32 characters.

packages/cli/src/ui/components/shared/text-buffer.ts
sijieamoy: `isWordCharStrict` is specifically for the vim behavior. An example is "hello@example.com". `isWordChar` will treat it as three words, "hello", "@example", ".com" when navigating with `alt+arrow`, but `isWordCharStrict` will see "hello", "@", "example", ".", "com" with Vim's word movement (`w`). So one is for the normal OS word nav and the other for the vim-specific logic.

packages/cli/src/ui/components/shared/text-buffer.ts
sijieamoy: Done!

sijieamoy:
> Looks good! Please also add a couple of tests that verify that these methods behave correctly given UTF32 characters.

Addressed comments above and added a few UTF-32-related tests, specifically precomposed characters, combining characters, non-Latin characters, unicode symbols (excluding emojis as the input component doesn't seem to support emoji anyway).

packages/cli/src/ui/components/shared/text-buffer.test.ts
```diff
@@ -1591,3 +1595,94 @@ describe('textBufferReducer vim operations', () => {
...
+
+  describe('findNextWordStartInLine with Unicode', () => {
+    it('should handle right-to-left text', () => {
+      const result = findNextWordStartInLine('hello مرحبا world', 0);
+      expect(result).toBe(6); // Start of Arabic word
+    });
+
+    it('should handle Chinese characters', () => {
+      const result = findNextWordStartInLine('hello 你好 world', 0);
+      expect(result).toBe(6); // Start of Chinese word
```
jacob314, line 1629: would be good to also test if the first word was Chinese. Same above as that would probably stress improper unicode character handling more.

packages/cli/src/ui/components/shared/text-buffer.ts
```diff
@@ -34,16 +34,41 @@ function isWordChar(ch: string | undefined): boolean {
 }
 
 // Helper functions for line-based word navigation
-export const isWordCharStrict = (char: string): boolean => /\w/.test(char);
+export const isWordCharStrict = (char: string): boolean =>
+  /[\w\p{L}\p{N}]/u.test(char);
```
jacob314: this is a tricky regex. 
Maybe add a comment that it
matches a single character that is any Unicode letter, any Unicode number, or an underscore

packages/cli/src/ui/components/shared/text-buffer.ts
```diff
@@ -34,16 +34,41 @@ function isWordChar(ch: string | undefined): boolean {
...
+  if (/[\p{Script=Latin}]/u.test(char)) return 'latin'; // All Latin script chars including diacritics
+  if (/[\p{Script=Han}]/u.test(char)) return 'han'; // Chinese
+  if (/[\p{Script=Arabic}]/u.test(char)) return 'arabic';
+  if (/[\p{Script=Hiragana}]/u.test(char)) return 'hiragana';
+  if (/[\p{Script=Katakana}]/u.test(char)) return 'katakana';
+  if (/[\p{Script=Cyrillic}]/u.test(char)) return 'cyrillic';
+  return 'other';
+};
+
+// Check if two characters are from different scripts (indicating word boundary)
```
jacob314, line 60: this has gotten really complex. could be good to audit what vim typically does and align with that. would be good to do the simplest that is sufficient.

packages/cli/src/ui/components/shared/text-buffer.ts
```diff
@@ -109,37 +151,95 @@ export const findPrevWordStartInLine = (
...
+    // Handle word characters
+    while (i < chars.length && isWordCharWithCombining(chars[i])) {
+      foundWord = true;
+
+      // Track the position of the last base character (not combining mark)
+      if (isWordCharStrict(chars[i])) {
+        lastBaseCharPos = i;
+      }
+
+      // Check if next character is from a different script (word boundary)
```
jacob314, line 208: are these cases really useful? if so please make sure there are tests that capture them

packages/cli/src/ui/components/shared/vim-buffer-actions.test.ts
```diff
@@ -140,6 +140,25 @@ describe('vim-buffer-actions', () => {
         expect(result.cursorRow).toBe(1);
         expect(result.cursorCol).toBe(0);
       });
+
+      it('should skip over combining marks to avoid cursor disappearing', () => {
```
jacob314, line 144: this would impact non vim mode equally. if this is an existing bug we should fix it. potentially there is a simpler fix that involves skipping over width 0 characters (there is char measurement code in text-buffer already that may be busted if this case is more difficult than I realized.

jacob314 (commented):
The input prompt should support emojis. please tell me more about what is broken for them. seems like you may have caught a few more complex text cases we should handle better. would like to discuss a bit how we handle them correctly at a more fundamental level in text-buffer rather than just patching in the vim support.

sijieamoy:
> The input prompt should support emojis. please tell me more about what is broken for them. seems like you may have caught a few more complex text cases we should handle better. would like to discuss a bit how we handle them correctly at a more fundamental level in text-buffer rather than just patching in the vim support.

I tried entering and pasting emojis in the input but neither worked.

packages/cli/src/ui/components/shared/text-buffer.ts
sijieamoy: They are useful, and already have tests for them (starting at line 909 of vim-buffer-actions.test.ts).

packages/cli/src/ui/components/shared/text-buffer.ts
sijieamoy: Mind if I file a ticket as a reminder for later? The user-facing behaviors are consistent with Vim on these supported cases for now.

packages/cli/src/ui/components/shared/vim-buffer-actions.test.ts
sijieamoy: What's the suggested action here? During manual testing the only case I found in the non-vim-mode is a word with mixed scripts e.g. "中文test英文word", `alt`+arrow doesn't detect the language boundaries.

packages/cli/src/ui/components/shared/text-buffer.test.ts
sijieamoy: added

packages/cli/src/ui/components/shared/text-buffer.ts
sijieamoy: added

packages/cli/src/ui/components/shared/text-buffer.ts
jacob314: works for me

packages/cli/src/ui/components/shared/vim-buffer-actions.test.ts
jacob314: thanks for looking into it.

jacob314 (approved):
lgtm

PR #5337, opened 2025-08-01
fix(editors): fix neovim closing when using `modify with editor`

### LINKED ISSUE #3302: Lazyvim: nvim terminates immediately

### LINKED ISSUE #3496: Unable to Access External Editor Neovim within Neovim via Gemini CLI

### LINKED ISSUE #4767: using neovim as external editor, editor keeps quitting after lsp servers are run

redoxahmii:
## TLDR

Solves the issue of instantly closing neovim when diffing files using `modify with external editor` with plugins enabled.

## Dive Deeper
Removed the diff coloring as neovim by default has diff colors that are sane as compared to vim.
Removed the `Autocmd` added for using `wqa` to quit all windows as that was the core issue causing it to instantly close when opening if you have plugins installed.
Currently only updates the ` WinClosed` to `BufWritePost` but could be made better with separating command for diffs and allowing `colorschemes` to handle the highlights. More details at #5014 
If it is mandatory to have the `autocmd` then the better approach would be to include `--noplugin` in the cli command to ensure it actually works instead of conflicting when plugins are installed. lmk and I will change it to that.
Tested it with `LazyVim` which is one of the heaviest configurations currently popular and it works fine with it and also with base neovim itself.
Currently, this PR also removes it from the `vim` editor command as well because this issue could triage to people using vim with heavy configurations as well. 

This is a remake of that older PR mentioned above because resetting on `main` branch of fork caused it to force close the previous PR (What a silly mistake to make and i never knew github doesn't reopening it)

## Reviewer Test Plan
- Change editor to `neovim` in preferences.
- Tell gemini to add something to a file.
- Select `Modify with editor` option to see.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅ | ❓  | ✅  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs
Closes #4767 
Closes #3496 
Closes #3302


redoxahmii:
The following is a showcase of what additionally could be added to this by separating the `args` for `vim` and `neovim` and not passing hardcoded colors as `highlights`.

image

Above is the default `nvim` colors when diffing, which are already acceptable as compared to `vim` in this comparison.

image

Passing those colors for someone that has a colorscheme defined results in this, which isn't really clear or understandable.
Most colorschemes add the diff colors themselves so it's more coherent with the theme and the `Normal` highlight and it should inadvertently not cause any issues in the future for readability with any colorscheme.

image

image

Above are 2 examples of different colorschemes being used, but because they have their own defined highlights for diffs it allows for any other to also work without the readability issue.

Let me know if that is desired and if it is not then this can be merged.
Sorry again for the double PRs I never knew that resetting the main would close the previous PR and I'm amazed GitHub doesn't allow reopening it once that happens.

jacob314 (approved):
lgtm

PR #5338, opened 2025-08-01
bug(cli): Prefer IPv4 dns resolution by default.

### LINKED ISSUE #2323: Intermittent ETIMEDOUT Error Due to IPv6 Connectivity to cloudcode-pa.googleapis.com
### What happened?

**Description**

The Gemini CLI intermittently fails with an ETIMEDOUT error when connecting to cloudcode-pa.googleapis.com (e.g., 2404:6800:4003:c20::5f:443). Testing confirms that IPv6 connectivity to this endpoint is unreliable on some networks, while IPv4 connections succeed consistently. The issue appears when the CLI prioritizes IPv6 (default Node.js behavior in some versions), and the network or ISP has inconsistent IPv6 routing to the API endpoint. 

Forcing Node.js to prefer IPv4 via export NODE_OPTIONS="--dns-result-order=ipv4first" resolves the issue.

**Steps to Reproduce**

Ensure your network supports IPv6 (e.g., ping6 -c 4 google.com succeeds).
Install the Gemini CLI (npm install -g @google/gemini-cli).
Run a CLI command, e.g., gemini <your-command>.
Observe intermittent ETIMEDOUT errors when connecting to cloudcode-pa.googleapis.com.
Test connectivity:
curl -6 -v https://cloudcode-pa.googleapis.com (times out).
curl -4 -v https://cloudcode-pa.googleapis.com (succeeds with 404).
ping6 2404:6800:4003:c20::5f (fails with Unknown host).

**Expected Behavior**
The CLI should consistently connect to cloudcode-pa.googleapis.com without ETIMEDOUT errors, regardless of IPv6 or IPv4 usage, by implementing robust fallback (e.g., Happy Eyeballs algorithm) or preferring IPv4 when IPv6 fails.

******Actual Behavior**
The CLI intermittently fails with:
Error: connect ETIMEDOUT 2404:6800:4003:c20::5f:443

The error occurs when the CLI attempts an IPv6 connection to cloudcode-pa.googleapis.com, but the network or ISP has unreliable IPv6 routing to this endpoint. The CLI succeeds when it uses IPv4 or a rare successful IPv6 route.

**Environment**

OS: macOS (e.g., macOS Ventura 13.6) [or specify Linux/Windows if applicable]
Network: IPv6-enabled network (confirmed by successful ping6 google.com)

Proposed Fix
Set the NODE_OPTIONS environment variable to force Node.js to prefer IPv4:
export NODE_OPTIONS="--dns-result-order=ipv4first"

Then rerun the CLI command. This ensures the CLI uses IPv4, which connects reliably (e.g., to 142.251.12.95).
To make persistent:
echo 'export NODE_OPTIONS="--dns-result-order=ipv4first"' >> ~/.zshrc
source ~/.zshrc

**Suggested Permanent Fix:**

Update the CLI to set dns.setDefaultResultOrder('ipv4first') internally or implement the Happy Eyeballs algorithm (RFC 8305) to try IPv6 and fall back to IPv4 after a short timeout (e.g., 300ms), as done in curl or modern browsers.
Alternatively, allow users to configure IPv4 preference via a CLI flag or config file.

**Additional Context**

Test Results:
ping6 google.com succeeds (e.g., 2001:4860:4802:32::78), confirming network IPv6 support.
curl -6 https://cloudcode-pa.googleapis.com times out, indicating specific IPv6 routing issues to the API endpoint.
curl -4 https://cloudcode-pa.googleapis.com connects (returns 404), confirming IPv4 reliability.

**Potential Causes:**
Node.js’s default DNS resolution may prioritize IPv6 (especially since Node.js v17, which changed to prefer IPv6 in some cases,).
ISP or network routing issues for Google’s API IPv6 addresses.
Lack of Happy Eyeballs in the CLI’s HTTP client, causing it to hang on failed IPv6 connections.,

**Workaround Effectiveness:** 
Setting NODE_OPTIONS="--dns-result-order=ipv4first" resolves the issue consistently, as it forces IPv4 usage, which is reliable on the user’s network.

Related Issues: Similar IPv6-related ETIMEDOUT errors have been reported in Node.js projects (e.g., npm CLI,), often fixed by preferring IPv4 or disabling IPv6.

The CLI should consistently connect to cloudcode-pa.googleapis.com without ETIMEDOUT errors, regardless of IPv6 or IPv4 usage, by implementing robust fallback (e.g., Happy Eyeballs algorithm) or preferring IPv4 when IPv6 fails.

### Client information

<details>

```console
                                                  │
│ About Gemini CLI                                 │
│                                                  │
│ CLI Version      0.1.6                           │
│ Model            gemini-2.5-flash                │
│ Sandbox          no sandbox                      │
│ OS               darwin                          │
│ Auth Method      OAuth    
```

</details>

### Login information

_No response_

_No response_

joshualitt:
## TLDR

Can improve network reliability for some users(i.e. #2323), with little drawback*.

## Dive Deeper

*It is true that some users on all IPv6 networks might be better off using `verbatim`, those users can explicitly set this option if they wish. In the future, when more users are on such networks, we can flip the default, but for now the unreliability of IPv6 networks is still an issue in many parts of the world. Furthermore, even where IPv6 is the superior choice, it's not superior by much, thus preferring IPv4 could be a good option for now.

## Reviewer Test Plan

Gemini CLI works as expected.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Resolves #2323

### DIFF
--- a/packages/cli/src/config/settings.ts
+++ b/packages/cli/src/config/settings.ts
@@ -38,6 +38,8 @@ export function getSystemSettingsPath(): string {
   }
 }
 
+export type DnsResolutionOrder = 'ipv4first' | 'verbatim';
+
 export enum SettingScope {
   User = 'User',
   Workspace = 'Workspace',
@@ -110,6 +112,7 @@ export interface Settings {
   disableAutoUpdate?: boolean;
 
   memoryDiscoveryMaxDirs?: number;
+  dnsResolutionOrder?: DnsResolutionOrder;
 }
 
 export interface SettingsError {
--- a/packages/cli/src/config/settings.test.ts
+++ b/packages/cli/src/config/settings.test.ts
@@ -777,6 +777,48 @@ describe('Settings Loading and Merging', () => {
       }
     });
 
+    it('should correctly merge dnsResolutionOrder with workspace taking precedence', () => {
+      (mockFsExistsSync as Mock).mockReturnValue(true);
+      const userSettingsContent = {
+        dnsResolutionOrder: 'ipv4first',
+      };
+      const workspaceSettingsContent = {
+        dnsResolutionOrder: 'verbatim',
+      };
+
+      (fs.readFileSync as Mock).mockImplementation(
+        (p: fs.PathOrFileDescriptor) => {
+          if (p === USER_SETTINGS_PATH)
+            return JSON.stringify(userSettingsContent);
+          if (p === MOCK_WORKSPACE_SETTINGS_PATH)
+            return JSON.stringify(workspaceSettingsContent);
+          return '{}';
+        },
+      );
+
+      const settings = loadSettings(MOCK_WORKSPACE_DIR);
+      expect(settings.merged.dnsResolutionOrder).toBe('verbatim');
+    });
+
+    it('should use user dnsResolutionOrder if workspace is not defined', () => {
+      (mockFsExistsSync as Mock).mockImplementation(
+        (p: fs.PathLike) => p === USER_SETTINGS_PATH,
+      );
+      const userSettingsContent = {
+        dnsResolutionOrder: 'verbatim',
+      };
+      (fs.readFileSync as Mock).mockImplementation(
+        (p: fs.PathOrFileDescriptor) => {
+          if (p === USER_SETTINGS_PATH)
+            return JSON.stringify(userSettingsContent);
+          return '{}';
+        },
+      );
+
+      const settings = loadSettings(MOCK_WORKSPACE_DIR);
+      expect(settings.merged.dnsResolutionOrder).toBe('verbatim');
+    });
+
     it('should leave unresolved environment variables as is', () => {
       const userSettingsContent = { apiKey: '$UNDEFINED_VAR' };
       (mockFsExistsSync as Mock).mockImplementation(
--- a/packages/cli/src/gemini.test.tsx
+++ b/packages/cli/src/gemini.test.tsx
@@ -6,7 +6,11 @@
 
 import stripAnsi from 'strip-ansi';
 import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
-import { main, setupUnhandledRejectionHandler } from './gemini.js';
+import {
+  main,
+  setupUnhandledRejectionHandler,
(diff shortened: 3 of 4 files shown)


joshualitt:
Okay, I've done some more testing of this. Specifically, DNS resolution order on two computers which AFAIK are on at least somewhat modern networks, and in both cases IPv4 was preferred by Node, except in one case where IPv6 apparently was preferred, but even in that case it did not make a noticeable performance difference.

packages/cli/src/config/config.ts
```diff
@@ -210,6 +212,13 @@ export async function parseArguments(): Promise<CliArgs> {
         // Handle comma-separated values
         dirs.flatMap((dir) => dir.split(',').map((d) => d.trim())),
     })
+    .option('dns-resolution-order', {
```
jacob314: can you ask gemini cli to change this from in config to in settings.ts
we are erring on the side of adding new options to settings.ts so they can be set via settings.json rather than as command line arguments to avoid adding too many rarely used command line arguments.

jacob314 (commented):
Thanks for fixing and adding tests!
Approved once this is moved from config.ts to settings.ts as this should be an option few users care about changing. I'll be adding a change to GEMINI.md shortly to reduce the number of cases like this where options get added to config.ts that should be in settings.ts

packages/cli/src/config/config.ts
joshualitt: Yes, this is a great idea, thanks!

packages/cli/src/config/settings.ts
```diff
@@ -38,6 +38,8 @@ export function getSystemSettingsPath(): string {
   }
 }
 
+export type DnsResolutionOrder = 'ipv4first' | 'verbatim';
```
jacob314, line 41: nice!

jacob314 (approved):
lgtm

Clausinho:
Appreciate the speedy fiy

PR #1969, opened 2025-06-26
Centralize Key Binding Logic and Refactor

Lee-WonJun:
## TLDR

Centralized key binding logic into a single place to improve maintainability and consistency. Hardcoded key combinations like `Ctrl+L` and `Ctrl+P` have been moved to the `keyMatchers` object in `keyBindings.ts`. This lays the groundwork for future support for user-defined key bindings.

## Dive Deeper

Previously, key bindings were scattered across different files like `App.tsx` and `InputPrompt.tsx`, making them hard to manage. This change introduces:

- Centralized key binding definitions in `keyBindings.ts` under `keyMatchers`
- Meaningful action names for each key binding (e.g., `showErrorDetails`, `toggleToolDescriptions`)
- Reduced code duplication and increased consistency

A few exceptions remain:
- `Ctrl+X` and `Ctrl+E` interactions with external editors involve complex logic, so the original code is left untouched
- Low-level keys managed by `text-buffer`  are also left unchanged, as improvements have already been made by the gemini-cli team (e.g., handling `del` vs `backspace`)

## Reviewer Test Plan

Confirm that all previous key bindings still work as expected.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| -------- | --- | --- | --- |
| npm run  | ❓  | ❓  |  O  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs
#1950 

This PR presents a sample / illustrative approach to centralizing key bindings.
It’s not meant to dictate the only possible solution—if the original structure (e.g., keeping bindings directly in terminal-style components) was intentionally chosen, please feel free to point that out.

Issue #1950, opened 2025-06-26
Add custom SHORTCUT key binding with Documentation


Lee-WonJun:
### What would you like to be added?

Currently, most shortcut key bindings in the codebase are hardcoded (shell-terminal-based), which prevents customization and makes maintenance difficult as the app scales. Additionally, there seems to be no central documentation for these shortcuts.

### Why is this needed?

1. Hardcoded shortcuts are unscalable: As the number of features grows, maintaining hardcoded shortcuts becomes increasingly error-prone and unmanageable.

2. No discoverability: Developers must read through the source code to understand available shortcuts.

3. Scattered definitions: While text-buffer.ts might be an exception, the shortcuts in app.tsx and inputPrompt.tsx should ideally be defined in a central dictionary or mapping.

### Additional context

Suggested Implementation Steps

1. Migrate hardcoded shortcuts to a centralized global dictionary (e.g., keyBindings.ts).

2. Ensure UI descriptions referencing shortcut keys also pull from this dictionary, rather than hardcoding text like "ctrl-s".
```
<Box>
  <Text color={Colors.Gray} wrap="truncate">
    Press ctrl-s to show more lines
  </Text>
</Box>
```

3. Add override support via a user-editable config (e.g., .settings.json).

4. generate documentation from the global dictionary for visibility and maintainability.

DavidHJong:
one more issue, user-friendly. current shortcut scheme is not comfortable for all users.

eliliam:
This is a huge missing feature of Gemini CLI! For example, many terminals are configured to open a new terminal tab when `ctrl+t` is pressed, which means that Gemini's `ctrl+t` shortcut to view chat transcript and detailed tool usage is entirely inaccessible since there is no way to change Gemini's keybind for that shortcut to something which is not already in use.

matrs:
This is needed. Also, please, stop changing the key bindings without warning, now there is a new key binding to see the model responses' errors, `F12`, which is my case has always been bound to another app (the default and reliable `yakuake`). There is way too much software already in existence for you to just set hard-coded shortcuts and be fine with it.

NubeBuster:
Pressing CTRL + C while Gemini is cookin' is ambiguous. The key event cascades to interrupt the generation rather than stopping at first consumption to clear the input area. This could be resolved by allowing keybind configuration.

[UPDATE] for anyone with the same issue. ESC ESC clears the input unambiguously.

Adib234:
[...]

PR #5356, opened 2025-08-01
Centralize Key Binding Logic and Refactor (Reopen)

Lee-WonJun:
## TLDR

This PR reopens the work from the #1969 [previously closed PR](https://github.com/google-gemini/gemini-cli/pull/1969). For implementation details, please refer to the previous discussion.

1. **App.tsx Now Uses the Key Interface**

   * Similar to `InputPrompt.ts`, `App.tsx` has been updated to use the `Key` interface by replacing the hook.
   * From what I can tell, the existing test cases still pass, and I've verified through debug logs that backward compatibility does not appear to be broken.
   * However, since I’m mainly a backend developer, I don’t have deep insight into the importance or prioritization of React hook patterns.
   * If the previous lower-level hook used in `App.tsx` was intentional, then maybe `keyBindings` needs to be revised to accommodate it—or perhaps it should leverage the same type + mapper structure used here. But honestly, that approach doesn't feel very clean to me. I think it's better to follow a consistent pattern across the app.

2. **Context on Centralization Effort**

   * As noted in the original #1950 issue, the goal of centralizing key bindings is to lay the groundwork for future support for user-configurable key mappings.
   * Even without that functionality, I believe this centralization improves maintainability and consistency, making the codebase cleaner and easier to evolve.
   * Regardless of whether this PR gets merged, I hope the idea of giving users more control over configuration settings remains part of the long-term roadmap (as it seems to already be tracked under the epic).
   * As with the previous PR, this is not intended to impose a single solution—if the original structure (such as placing bindings directly in terminal-style components) was chosen deliberately, I’d appreciate your thoughts or guidance.

## Reviewer Test Plan

Confirm that all previous key bindings still work as expected.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  |  O  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs
#1950 #1969 

Thanks for the helpful feedback, @mattKorwel


mattKorwel:
@Lee-WonJun thanks for resurrecting this! Can you address the comment from the menu box about the overlapping keys?

@jacob314 this one looks good to me. I want to double check with you.

Lee-WonJun:
Does my understanding of the Gemini review seem correct?

* Previously, either `useInput` or `useKeypress` would run — meaning, if `useKeypress` was active, the `useInput` hook would not be triggered. Is that the intended meaning?

I'm not fully familiar with how priority or exclusivity works between them, but if my assumption is correct, it seems that both hooks were already running concurrently even before.

When I test it with the stable version of Gemini or the `main` branch build (while both components are active), all hooks appear to be invoked. So perhaps I misunderstood the intent, or maybe there's some focus management behavior that I'm not aware of?



https://github.com/user-attachments/assets/cc09f989-cf6d-4e37-9c43-3627eae1cf2d

(Please note that I may be running into an XY problem due to my limited understanding of Ink or frontend behavior.)

packages/cli/src/ui/keyBindings.ts
```diff
@@ -0,0 +1,66 @@
...
+ * interface for key matchers
+ */
+export interface KeyMatchers {
+  readonly [shortcutName: string]: KeyMatcher;
+}
+
+/**
+ * key binding matchers
+ */
+export const keyMatchers: KeyMatchers = {
```
jacob314: This is a great step in the right direction. However lets tweak this so it can be more data driven and less imperative code .

That means create an enum of all the Commands to assign keyboard shortcuts to 
and for each one provide a list of matching keys where each key is

{
  key | String,
  ctrl | boolean,
  shift | boolean,
  alt | boolean,
  command | boolean,
}

I would expect Gemini CLI should be able to one short refactor from this imperative code to that data driven code. Please let me know if you have any questions.

The reason that is important is it will make this code then scale to allowing users to provide their own keybinding mappings.

packages/cli/src/ui/App.tsx
```diff
@@ -553,46 +553,71 @@ const App = ({ config, settings, startupWarnings = [], version }: AppProps) => {
...
+        }
+      } else if (
+        keyMatchers.toggleIDEContextDetail(key) &&
+        config.getIdeMode() &&
+        ideContextState
+      ) {
+        setShowIDEContextDetail((prev) => !prev);
+      } else if (keyMatchers.quit(key)) {
+        handleExit(ctrlCPressedOnce, setCtrlCPressedOnce, ctrlCTimerRef);
+      } else if (keyMatchers.exit(key)) {
```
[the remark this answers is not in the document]
jacob314: ignore this comment. this is a symptom of LLMs being more familiar with the popular ink library than with our code. we intend to migrate all this code to use useKeypress just hadn't yet so this is a helpful step in the right direction.

jacob314 (commented):
I'll approve this to land once the changes requested are made and there is some tests for the new keyBindings functionality.

packages/cli/src/ui/components/InputPrompt.tsx
```diff
@@ -208,7 +209,7 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
         }
       }
 
-      if (key.ctrl && key.name === 'l') {
+      if (keyMatchers.clearScreen(key)) {
```
jacob314: note that text-buffer.ts also has a large number of key bindings that should be moved to use the new system over time.

Lee-WonJun:
@jacob314  thanks for your feedback and suggestions! I've clarified my thought process below based on your comments.

### Reason for Using Lambda Functions

The main reason I've implemented this using lambda functions is that, I'm not aware of the detailed plans for all the shortcuts (or key binding) that `gemini-cli` will need to support.

-----

### Ideal Architecture and Staged Implementation

I'm also a big fan of data-driven DSLs (Domain-Specific Languages), and the ideal architecture I had in mind is as follows:
`[String DSL (User Config)]` → `[ADT-based DSL (Core)]` → `[Callable Interface]`

However, this PR follows a **staged approach** rather than implementing the entire feature at once. In this first stage (Stage 1), I've focused on consolidating the interfaces related to shortcuts.

I've excluded `text-buffer.ts` for now as it seems to be a more low-level component, which I'll explain further below.

-----

### Thoughts on DSL Design

Users will need to be able to override shortcuts via `settings.json`. Therefore, a simple string-based JSON DSL would be the most intuitive for user configuration:

```json
{
  "keymap": {
    "command": ["ctrl+A+B", "ctrl+A+ALT"]
  }
}
```

The system can't interpret this string directly, so it needs to be converted into an internal data structure.

For example, to support complex key combinations with AND/OR logic, we could design the DSL with a recursive ADT (Abstract Data Type) structure. This would be very **fluent** and expressive, but it also makes the DSL structure itself more complex to write.

```ts
Command: [
  {
    op: AND,
    keys: ['A', 'CTRL', { op: OR, keys: ['B', 'ALT'] }] // more combination
  }
]
```

Conversely, if we only allow simple shortcuts, a structure like the one you suggested would also be great:

```ts
Command: [
  { key: 'A', clt: true },
  { key: 'a', clt: true }
]
```

However, since I don't know the full scope of user configurations (Case-sensitivity, multi input , and so on) that `gemini-cli` will eventually allow, I felt that defining such a complex DSL at this stage would be an **overspec**.

-----

### Role of the `Matcher` and Implementation Direction

If we have a data-value-only DSL, we need a **Decision Interpreter** to parse it and determine the final action. I decided that representing this interpreter's interface using a `Map` would be more declarative and extensible. (Though it doesn't have to be a Map—a single function or another interface could work, but a Map seemed like the best fit.)

Therefore, the `Matcher` I've written represents the **final form** after the DSL has been interpreted. This means if a DSL existed, we could dynamically create the `Matcher` at boot-up by reading the DSL. Since the upstream DSL hasn't been defined yet, it currently only contains hard-coded lambdas.

The reason I didn't include `text-buffer.ts` in the `Matcher` implementation is that, as the name implies, it's about the buffer and handles more **low-level manipulation**. This role is different from the `Matcher`, which is closer to a 'key' interface. (This is my impression from looking at the code).

In the future, if an ADT-based DSL is implemented in the core, I expect `text-buffer.ts` would use this DSL data as config constants instead of magic numbers. For this reason, I intentionally excluded it from the higher-level `Matcher` implementation.

-----

My understanding is that your suggestions were mainly about moving towards a data-driven DSL. I hope my explanation of the current approach and future direction makes sense in that context. Is my understanding correct?

Please feel free to share any more thoughts you have. I'd appreciate the feedback\!

Lee-WonJun:
@jacob314 I've made some code changes to support a more data-driven config. I’d love to hear your thoughts on whether this is heading in a positive direction.


---

1. **KeyBinding types and default bindings have been moved into the config.**
   In the future, I hope to either parse these from a user-level config or use them directly in the settings interface.

   For now, the `KeyBinding` behavior remains almost identical to the existing `ui/Key` implementation. This is intentional to maintain backward compatibility — for instance, `{ sequence: '\x18', ctrl: true }` should still work the same way. I deliberately avoided introducing a completely new type or breaking compatibility in this PR.

2. **The Matcher dynamically generates an interface that reads the binding structure and evaluates whether it matches a specific command enum.**

3. **Some important caveats:**

   * **3-1)** After my initial changes, reverse search functionality was added, which led to the introduction of new enums. In these cases, the command differs slightly — for example, `SUBMIT_REVERSE_SEARCH` vs. `SUBMIT`. Although they are similar, their behavior diverges, so keeping them separate makes sense. However, I’m not yet sure what the best naming convention should be, and it’s also unclear how much of this will eventually be user-overridable. We’ll need to clarify this at some point.

   * **3-2)** Currently, multi-key shortcuts (e.g., `Ctrl + A + B`) are not supported.

   * **3-3)** As expected, since the config does not define a matching priority, a command that matches first may overshadow lower-priority commands.

packages/cli/src/config/keyBindings.ts
```diff
@@ -0,0 +1,180 @@
...
+  // Original: key.ctrl && (key.name === 'c' || key.name === 'C')
+  [Command.QUIT]: [
+    { key: 'c', ctrl: true },
+    { key: 'C', ctrl: true },
+  ],
+  // Original: key.ctrl && (key.name === 'd' || key.name === 'D')
+  [Command.EXIT]: [
+    { key: 'd', ctrl: true },
+    { key: 'D', ctrl: true },
+  ],
```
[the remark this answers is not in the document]
jacob314: generally agree with this. suspect previous code we had was confused but it would be good to verify this.

packages/cli/src/ui/components/InputPrompt.tsx
```diff
@@ -296,30 +296,39 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
...
         handleSubmitAndClear(buffer.text);
         return;
       }
 
       if (completion.showSuggestions) {
         if (completion.suggestions.length > 1) {
-          if (key.name === 'up' || (key.ctrl && key.name === 'p')) {
+          if (
+            keyMatchers[Command.NAVIGATION_UP](key) ||
+            keyMatchers[Command.HISTORY_UP](key)
```
jacob314, line 317: this should be a separate command
COMPLETION_UP 
that matches both of these keys

that way users who are annoyed that the up arrow navigates through completion history can change it without breaking other things coupled to NAVIGATION_UP

packages/cli/src/ui/components/InputPrompt.tsx
```diff
@@ -296,30 +296,39 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
...
+          ) {
             completion.navigateUp();
             return;
           }
-          if (key.name === 'down' || (key.ctrl && key.name === 'n')) {
+          if (
+            keyMatchers[Command.NAVIGATION_DOWN](key) ||
```
jacob314, line 323: make this COMPLETION_DOWN

jacob314:
Please let me know when you are ready for me to review again. Can't wait to have custom keyboard shortcuts.

Lee-WonJun:
@jacob314 I’ve resolved the conflicting parts.

Lee-WonJun:
By changing the hook in `app.ts`, conflicts occur whenever `app.ts` is updated due to changes such as depth. I fix it every time, but I can’t always keep my branch in sync with the main branch.

jacob314 (commented):
Minor comments. Attempting to approve so you can merge to unblock.

jacob314 (approved):

packages/cli/src/config/keyBindings.ts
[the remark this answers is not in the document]
Lee-WonJun: At least in my environment, it works after removing the uppercase binding. If case sensitivity is needed, we could probably use the sequence instead.

packages/cli/src/ui/components/InputPrompt.tsx
Lee-WonJun: Agree

PR #5358, opened 2025-08-01
Add a setting to disable auth mode validation on startup

bbiggs:
## TLDR

Add a setting to disable auth mode validation on app startup

## Dive Deeper

Each of the auth modes (Gemini API, Login with Google) have a complex set of environment variables they support to set up the right quota/credentials/access for the APIs they use. This change adds a setting to disable auth validation on startup: we let the app start up in GEMINI_API mode even if no API key environment variable is set. This allows for using or scripting the CLI in tests where there are no API calls needed, test harnesses where the key is passed externally, or use in scenarios where the auth credentials are passed via a separate channel.

## Linked issues / bugs

Fixes #5357

### DIFF
--- a/packages/cli/src/validateNonInterActiveAuth.test.ts
+++ b/packages/cli/src/validateNonInterActiveAuth.test.ts
@@ -10,6 +10,7 @@ import {
   NonInteractiveConfig,
 } from './validateNonInterActiveAuth.js';
 import { AuthType } from '@google/gemini-cli-core';
+import * as auth from './config/auth.js';
 
 describe('validateNonInterActiveAuth', () => {
   let originalEnvGeminiApiKey: string | undefined;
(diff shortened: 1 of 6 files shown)


packages/cli/src/validateNonInterActiveAuth.test.ts
```diff
@@ -152,7 +184,11 @@ describe('validateNonInterActiveAuth', () => {
     const nonInteractiveConfig: NonInteractiveConfig = {
       refreshAuth: refreshAuthMock,
     };
-    await validateNonInteractiveAuth(AuthType.USE_GEMINI, nonInteractiveConfig);
+    await validateNonInteractiveAuth(
+      AuthType.USE_GEMINI,
+      undefined,
+      nonInteractiveConfig,
+    );
     expect(refreshAuthMock).toHaveBeenCalledWith(AuthType.USE_GEMINI);
```
jacob314, line 193: please add a test exercising the useExternalAuth=true code path.

jacob314 (approved):
Looks good. Please add a test for useExternalAuth and then I'll land.lgtm

packages/cli/src/validateNonInterActiveAuth.test.ts
```diff
@@ -175,4 +212,29 @@ describe('validateNonInterActiveAuth', () => {
     expect(consoleErrorSpy).toHaveBeenCalledWith('Auth error!');
     expect(processExitSpy).toHaveBeenCalledWith(1);
   });
+
+  it('skips validation if useExternalAuth is true', async () => {
+    // Mock validateAuthMethod to return error to ensure it's not being called
+    const mod = await import('./config/auth.js');
```
jacob314: nit: a bit silly we are importing inline like this. unless that is strictly needed move that to a regular top level import.

jacob314 (approved):
lgtm

jacob314 (approved):
lgtm

PR #5361, opened 2025-08-01
Add highlights for input /commands and @file/paths

miguelsolorio:
## TLDR

Adds highlighting for /commands and @file/paths

## Dive Deeper

CleanShot 2025-08-01 at 09 44 47@2x

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


psinha40898:
Hi I'm not officially affiliated with the project so it may not be appropriate for me to chime in, but I hope it is helpful

When I pulled this locally, long paths that exist throughout multiple lines produced this behavior:

image

I also ran into this issue in #4924 and solved it by coloring every relevant character on re render 

Thank you and I hope this feedback was appropriate to provide

packages/cli/src/ui/components/InputPrompt.tsx
```diff
@@ -61,6 +61,45 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
...
 
+  // Memoize highlight ranges for performance.
+  const highlightRanges = useMemo(() => {
+    const ranges: Array<{ start: number; end: number }> = [];
+    // Regex to find words starting with @ or / that are preceded by whitespace or start of string.
+    const regex = /(?<=^|\s)[@/](?:\\ |\S)*/g;
+    let match;
+    while ((match = regex.exec(buffer.text)) !== null) {
+      ranges.push({
+        start: match.index,
```
jacob314, line 72: this isn't handling Utf32 chars robustly. Ask Gemini CLI to fix this to make sure the ranges are in terms of utf32 characters rather than utf16.

packages/cli/src/ui/components/InputPrompt.tsx
```diff
@@ -61,6 +61,45 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
 }) => {
   const [justNavigatedHistory, setJustNavigatedHistory] = useState(false);
 
+  // Memoize highlight ranges for performance.
+  const highlightRanges = useMemo(() => {
```
jacob314, line 65: consider asking Gemini CLI these methods need to go in textBuffer which already has some logic mapping logical to visual lines and stuff that you can leverage to avoid reinventing the wheel. ask it to move this logic to text-buffer.ts and to refactor to use existing helpers where possible.

miguelsolorio:
Will try this in a new PR

Issue #5230, opened 2025-07-30
MCP Prompts support ignores multiple content arguments


jakemac53:
### What happened?

I am testing Gemini CLI with the new MCP Prompts support (which translates them into `/` commands).

For the Dart MCP server, we have a prompt which will return the content in multiple parts (with different roles, it mimics the Agent asking for the user journey and then the user giving it back).

However, only the first content part seems to be respected in the response, so Gemini CLI asks me for a user journey that I already supplied.

### What did you expect to happen?

It should already have the user journey from the prompt response (I confirmed this by looking at my MCP server logs). So it shouldn't ask me for the user journey.

### Client information

<details>

```console
$ gemini /about
About Gemini CLI                                    │
│                                                     │
│ CLI Version       0.1.15                            │
│ Git Commit        bcce1e7b                          │
│ Model             gemini-2.5-pro                    │
│ Sandbox           no sandbox                        │
│ OS                linux                             │
│ Auth Method       gemini-api-key                    │
│                                      
```

</details>

chrstnb:
@haroldmciver-go is going to TAL

jakemac53:
afaik this is still an issue

PR #5380, opened 2025-08-01
Correctly handle multi-part MCP prompts

haroldmciver-go:
## TLDR

Addresses https://github.com/google-gemini/gemini-cli/issues/5230 where the Gemini CLI would incorrectly process multi-part responses from MCP servers. The previous implementation concatenated all message parts without preserving their conversational roles, causing the model to miss context and re-ask for information that was already provided. The fix ensures all message parts are formatted with their roles, preserving the full conversational context.

## Dive Deeper

When an MCP server returned a prompt response containing multiple ChatMessage parts, the CLI was not handling the conversational structure correctly. The original code extracted only the text content from each message and joined them with a newline, erasing the conversational context. 

The code in McpPromptLoader.ts was updated to format the response by including the role of each message. This preserves the conversational structure that the model expects. The final prompt sent to Gemini accurately represents the multi-turn context from the MCP server, resolving the issue and ensuring the entire user journey is understood.

## Reviewer Test Plan

Connect to the MCP server with a prompt that has multiple parts with different roles.

Execute the MCP prompt via slash command

Verify the model's response in the CLI has full context from the prompt and does not ask for information that was supplied in the prompt.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


packages/cli/src/services/McpPromptLoader.ts
```diff
@@ -123,7 +123,11 @@ export class McpPromptLoader implements ICommandLoader {
                 };
               }
 
-              if (!result.messages?.[0]?.content?.text) {
+              const content = result.messages
+                ?.map((message) => `${message.role}: ${message.content.text}`)
```
jacob314: Fyi @jakemac53 is this how you would want the content to be formatted?
Do we also need to handle the role being omitted or empty gracefully?

jacob314 (commented):
Thanks for addressing this. Please also add a test so we don't regress this behavior.
Will approve after confirmation from Jake that this is how we'd want the output to be formatted in this case.

packages/cli/src/services/McpPromptLoader.ts
```diff
@@ -123,7 +123,11 @@ export class McpPromptLoader implements ICommandLoader {
                 };
               }
 
-              if (!result.messages?.[0]?.content?.text) {
+              const content = result.messages
```
NTaylorMullen: We wont want to return `content` as a string here. Ultimately the the `submit_prompt` return type may need to be updated to take a `PartListUnion` type so that re-submission can take a complex data type. This would also then enable multi-part content to be returned from the prompts endpoint

packages/cli/src/services/McpPromptLoader.test.ts
```diff
@@ -0,0 +1,141 @@
...
+      invoke: vi.fn().mockResolvedValue({
+        messages: [
+          {
+            role: 'user',
+            content: { text: 'Hello' },
+          },
+          {
+            role: 'model',
+            content: { text: 'Hi there!' },
+          },
```
jakemac53, line 48: Just to illustrate the full back and forth behavior, I think it would be useful to have another user role message after the model one here.

packages/cli/src/services/McpPromptLoader.test.ts
```diff
@@ -0,0 +1,141 @@
...
+
+    const commands = await loader.loadCommands(new AbortController().signal);
+    expect(commands).toHaveLength(1);
+    const command = commands[0];
+
+    const result = await command.action!(mockContext, '');
+
+    expect(result.type).toBe('submit_prompt');
+    if (result.type === 'submit_prompt') {
+      expect(result.content).toBe('user: Hello\n\nmodel: Hi there!');
```
jakemac53: This result also presumably has a role - we should validate what that role is here?

Or, if these result objects don't have a role that should likely be added.

Ideally, we could get the full multi-part responses as a part of this result, with the roles etc.

packages/cli/src/services/McpPromptLoader.test.ts
```diff
@@ -0,0 +1,141 @@
...
+
+    vi.mocked(getMCPServerPrompts).mockReturnValue([mockPrompt]);
+
+    const commands = await loader.loadCommands(new AbortController().signal);
+    const command = commands[0];
+    const result = await command.action!(mockContext, '');
+
+    expect(result.type).toBe('submit_prompt');
+    if (result.type === 'submit_prompt') {
+      expect(result.content).toBe('undefined: Message without role');
```
jakemac53: I doubt we want `undefined:` to actually show up in what is presumably a user-visible message here?

For MCP specifically though, the role [is not optional](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/schema/2025-06-18/schema.json#L1632). So, just erroring in this case should also be acceptable. Or we could default to "user" if we want to support malformed responses.

packages/cli/src/services/McpPromptLoader.ts
jakemac53: +1 I think this is related to my comment above, so we can have the roles etc attached to the actual response.

packages/cli/src/services/McpPromptLoader.ts
jakemac53: Commented above related to this, I don't really think this is what we want ultimately.

We do not need to handle missing roles imo unless this also handles non-MCP related things, they are a required field in the MCP spec.

packages/cli/src/services/McpPromptLoader.ts
haroldmciver-go: I updated to the return type to be `PartListUnion`

jakemac53:
I will test this out locally with the Dart MCP server which sends responses like this 👍

packages/cli/src/services/McpPromptLoader.test.ts
```diff
@@ -60,7 +64,11 @@ describe('McpPromptLoader', () => {
 
     expect(result.type).toBe('submit_prompt');
     if (result.type === 'submit_prompt') {
-      expect(result.content).toBe('user: Hello\n\nmodel: Hi there!');
+      expect(result.content).toEqual([
+        { text: 'Hello' },
```
jakemac53, line 68: Can we validate the roles here as well?

jakemac53:
image

It does appear to have correctly gotten the argument this time - but I couldn't validate the actual message roles just by looking at this - I only see the prompt and its arguments in the UI. But the prompt does work as expected now.

Did we remove the full prompt from the UI again?

jakemac53:
Err actually, I take that back 🤣 . I have updated the prompt to look like this:

```dart
user: <large prompt for how to test a user journey>
model: Please give me the user journey
user: Some user journey
```

And now it doesn't work, Gemini responded with `✦ Please provide the user journey you would like me to write a test for.`

jakemac53:
Specifically, it seems like whether it works or not is really dependent on how "user journey-ish" the prompt sounds right now. Which makes me think the roles in the context are not properly set.

It worked on the first try with just "cool!" as the journey, because in that version of the prompt the server just returned two messages:

- user: `<large prompt for how to test a user journey>`
- user: Here is the user journey: cool!

So, there were no model roles there, and it was a bit more explicit. But when I make it mimic an interaction where the model asks you for the journey and then you supply it, that doesn't seem to work well.

packages/cli/src/services/McpPromptLoader.ts
```diff
@@ -132,9 +133,31 @@ export class McpPromptLoader implements ICommandLoader {
...
+                  content: (allMessageContents as string[]).join('\n\n'),
+                };
+              }
+
+              const content: PartListUnion = allMessageContents.flatMap(
+                (content) => {
+                  if (typeof content === 'string') {
+                    return [content];
+                  }
+                  return content as PartListUnion;
```
jakemac53, line 154: I tried passing through the model roles locally here, by just returning the full `response.messages` (a `PromptMessage[]`) and fwiw this line causes that to fail much later on.

This cast should be removed as it hides bugs and instead we should fix the part in the code that is actually requiring the case (where `PromptMessage[]` is added as a valid type). Or, we should handle that case here (probably ideal).

jakemac53: Actually, what I think we want to to change this to a ContentListUnion, so we can attach the roles, and then map the PromptMessage types to the Content types and return a `Content[]` from the submit prompt action?

jakemac53:
Hmmm ok so looking deeper into this I think we just need to be modifying the history directly instead of relying on `sendMessageStream` for these events. That can only handle user messages.

Then, maybe just ask Gemini who should go next (basically do the regular next turn logic).

There shouldn't even be an assumption that the user shouldn't immediately go next after sending a prompt, maybe they are intended to do their own follow-up.

chrstnb:

Issue #5244, opened 2025-07-30
Unable to use /copy


rkj:
### What happened?

```
╭───────────╮
│  > /copy  │
╰───────────╯

✕ Failed to copy to the clipboard.
```
I suspect it's part of my setup (not sure how would it work). I'm `ssh`ing into a workstation from my Mac and run `gemini` there, so it would have to pass the clipboard somehow to my local machine

### What did you expect to happen?

feature to not fail. Maybe it could explain limitations.

### Client information

* **CLI Version:** 0.1.15
* **Git Commit:** bcce1e7b
* **Operating System:** linux v22.5.1
*

rkj:
Consulted Gemini which suggested to add this to my tmux config:
```
set-window-option -g allow-passthrough on
set -g set-clipboard on
```
and inroduced me to `osc52` which does work... Does the `/copy` use it / why not?

cbagwell:
I also use ssh in same way and would find this useful.  I rely on OSC52 with Vim and Copying to my local clipboard over ssh.  Not all terminals support OSC52 but iTerm2 on Mac and Microsoft Terminal on Windows support it.  This repo has a nice table to see what terminals would support OSC52 if support is added to gemini-cli.

https://github.com/theimpostor/osc

preethamak:
I'm in the same situation — running `gemini` over SSH from my Mac to a remote Linux machine. The `/copy` command fails due to clipboard access issues, which makes sense since it's trying to use the remote clipboard.

I tested the OSC52 approach (with `set -g set-clipboard on` and `set-window-option -g allow-passthrough on` in tmux), and it works nicely with iTerm2. It would be great if `gemini-cli` could fall back to OSC52 when available, or at least detect the environment and explain why `/copy` won't work.

Really appreciate the discussion here — happy to help test if needed.

rkj:
I do have it enabled:
<img width="839" height="405" alt="Image" src="https://github.com/user-attachments/assets/06ef0853-1ea8-47e4-b259-51584285bb44" />

$ tmux -V
tmux 3.5a

ismellpillows:
It was just merged, maybe you can test on the next release

PR #5416, opened 2025-08-02
fix(cli): Improve error message for /copy command (#5244)

vero-code:
## TLDR

Improves the error message shown when the `/copy` command fails, to help users understand why copying may not work in remote or SSH environments.

## Dive Deeper

Previously, the message was too generic ("Failed to copy to the clipboard."). This change adds a clear hint that the issue may be related to the environment (e.g., SSH, tmux), and suggests using OSC52 or switching to a local terminal.

## Reviewer Test Plan

1. Run `gemini-cli` in a remote session (e.g., SSH).
2. Prompt the model with any question (e.g., "Tell me a joke about devs").
3. Run `/copy`.
4. Confirm that the new message is shown if clipboard is unavailable.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ✅  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Related to #5244


packages/cli/src/ui/commands/copyCommand.test.ts
```diff
@@ -249,7 +252,10 @@ describe('copyCommand', () => {
     expect(result).toEqual({
       type: 'message',
       messageType: 'error',
-      content: 'Failed to copy to the clipboard.',
+      content:
```
galdawave: The current error message seems to assume the failure is always due to a remote session. To make this more robust for other potential issues, we should including the original error message. That way, the user will have a clearer idea of what went wrong.

vero-code: Thanks for the feedback! I’ve updated the logic to include the original error message in the output. Let me know if that works better.

vero-code:
It looks like #9966 implemented OSC52 clipboard support and resolved #5244.
Would it make sense to close this one as superseded?

PR #5426, opened 2025-08-02
feat: have `/bug` copy over last prompt and response to GitHub issue

jackwotherspoon:
## TLDR

This PR aims to improve the use of `/bug` slash command, by having it automatically copy over the last prompt and corresponding response from Gemini CLI into the "What happened?" section of the GitHub issue.

## Dive Deeper

The assumption here is that most users when they encounter an error or bug will then report it using `/bug`. The user then has to manually copy over the error/output from Gemini CLI. By having `/bug` automatically copy over last prompt and response it makes it extremely simple for users to create actionable bugs.

Also, often times users will create a bug without specifying what there prompt was, this makes it extremely difficult as a maintainer to try and reproduce, often requiring us to ask "Do you mind sharing the prompt that caused this?". This PR also removes that need as the prompt is populated as well using the history.

## Considerations

- Should the user be prompted in Gemini CLI to allow copy of prompt and response?
  - Maybe... users may be reporting a bug unrelated to previous command
  - The user may not want certain information in their prompt or response filed on the bug. (although they could just remove it from the template)
- Rendering of tool output is `Tool Call: Shell, Status: Success, Description: openssl rand -base64 5000`
  - Not as visually appealing as copying over the rendered output which uses Ink
- If the "What happened?" field is already filled in by the copied prompt/response are users less likely to provide their own details and additional insights?

## Demo

Included are some examples:

- `/mcp` --> has emojis and proper ANSI chars stripped
- Invoke a tool (Shell tool) --> tool calls, status, and responses are preserved 
  - `Tool Call: Shell, Status: Success, Description: openssl rand -base64 5000`
- Gemini CLI response exceeds 8000 character limit for GitHub URL --> truncates response
  - `... truncated response due to character limit ...`
- Error path (mcp auth on misconfigured server) --> error is shown

https://github.com/user-attachments/assets/65e8091f-6429-4822-89e5-174c1d56a86c

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


jackwotherspoon:
Will clean up and add additional test cases.

yashasvigirdhar:
Thank you Jack!

I've one question on this: Is it possible that the response from the gemini cli might have something which user might not want to share publicly? For example something about their code or product.

jackwotherspoon:
> I've one question on this: Is it possible that the response from the gemini cli might have something which user might not want to share publicly? For example something about their code or product.

@yashasvigirdhar  This is a very valid point. I added it to the "Considerations" section above for the team to discuss.

Here are the two options I see:

1. Current approach: `/bug` auto opens and fills in the issue template (doesn't actually submit the issue). That way the user can easily remove any sensitive material or remove the auto-populated fields fully (if they so choose).
2. Alternative approach: When a user enters `/bug` they are prompted in Gemini CLI if they would like to auto-populate the last prompt and Gemini CLI response into the bug template.

Curious which you would prefer as a user @yashasvigirdhar 😄

yashasvigirdhar:
Sorry I missed this in the consideration above. 

Thanks for asking. In my opinion, I think I'd personally prefer option no 2 as I imagine it'd more efficient for me approve/reject in the terminal itself, similar to how I approve other tool calls, as compared to going over the bug description in browser and then selecting and removing the text.  

Usually the cli response has something related to my codebase or some personal topic so I think the no of times I would share the entire response would be quite less than the other.

packages/cli/tsconfig.json
```diff
@@ -3,7 +3,7 @@
   "compilerOptions": {
     "outDir": "dist",
     "jsx": "react-jsx",
-    "lib": ["DOM", "DOM.Iterable", "ES2022"],
+    "lib": ["DOM", "DOM.Iterable", "ES2023"],
```
NTaylorMullen, line 6: Why did we need to change this bit?

packages/cli/src/ui/commands/bugCommand.ts
```diff
@@ -47,16 +129,50 @@ export const bugCommand: SlashCommand = {
...
+      ].join('\n');
+      url = bugReportUrl
+        .replace('{title}', encodeURIComponent(bugDescription))
+        .replace('{info}', encodeURIComponent(info))
+        .replace('{problem}', encodeURIComponent(truncatedProblem));
+    }
+
+    bugReportUrl = url;
 
     context.ui.addItem(
```
NTaylorMullen, line 177: So one note: it's possible the `bugReportUrl` gets too big. All of the content gets embedded into the query string which may blow out the browsers query string limit. We'll want to add protection.

In addition for this addItem we'll also want to indicate to the user if we have included their request / response in the report and to instruct them to remove the content if needed.

NTaylorMullen (approved):

packages/cli/src/ui/commands/bugCommand.ts
jackwotherspoon: @NTaylorMullen I have a safeguard against the too big URL. At the moment if the URL gets over 8000 characters it will truncate the Gemini CLI response and hence the URL. There is a unit test verifying this.

> we'll also want to indicate to the user if we have included their request / response in the report and to instruct them to remove the content if needed.

I have thought about this and I think I am going to add a dialog prompt in Gemini CLI to ask user if they would like to include latest prompt/response in bug.

reconsumeralization:
I think this might be bad idea - think of the potential for automated reporting accidentally.

jackwotherspoon:
> I think this might be bad idea - think of the potential for automated reporting accidentally.

@reconsumeralization  Not sure I see the concern here... The `/bug` command already exists today?

Users could already automate this if they so choose. 

I think prompting the user via dialog box in Gemini CLI would probably safe guard against automation a little.

reconsumeralization:
I Can just see someone accidently reporting a bug about a bug in the bug system and confusing it.. I am sure it's fine.

reconsumeralization:
The amazon kiro application right now that specs stuff out - when used with this I Can see it make over specifications for silly things like arbitraging things in such a constricted way.  It's no bigy - I need some rest! Recon out!

reconsumeralization:
Hiding a malicious prompt at the end of a long text that gets predictably truncated is not safe as rejecting overly verbose bug reports.. just mho.

jerop:
@jackwotherspoon  still working on this? we'd like to merge quickly and keep open PR count low

reconsumeralization:
Thanks for the detailed PR, Jack! I think this is a great usability improvement, but there are some privacy and completeness risks we should address. Here’s a mitigated approach:

Explicit User Consent – Prompt the user in the CLI when /bug is invoked: “Do you want to include your last prompt and CLI responses in the bug report? [Y/n]”. Default to No, and allow editing or removal of content before submission.

Sensitive Data Filtering – Scan prompts and responses for API keys, tokens, passwords, proprietary code, or other sensitive info. Either redact automatically or warn the user before including.

Safe Truncation – Keep the 8,000-character limit, but clearly indicate truncation and optionally allow the user to select which portion to include if important info is cut off.

Improved Formatting – Strip ANSI codes but preserve structured tool output in a Markdown-friendly format (code blocks, headings).

Encourage Additional Context – Prompt the user to add steps to reproduce or other observations, so we don’t rely solely on the auto-filled content.

Optional Audit Logging – Maintain a local log of /bug submissions, including whether the last prompt/response was included, to detect accidental leaks or automated misuse.

This way, we preserve the convenience of auto-populating bug reports while minimizing risks of sensitive data exposure or incomplete reports.

abhipatel12:
Going to close this for inactivity

PR #5449, opened 2025-08-03
Add a setting to disable the version update nag message

### LINKED ISSUE #5447: Setting to disable the version update nag message

Add a setting, similar to disableAutoUpdate, that disables the version update nag message.

### Why is this needed?

In package managers that have their own auto-update mechanism, the update nag message in Gemini CLI can be confusing. 

### Additional context

_No response_

bbiggs:
## TLDR

Add a setting to disable the version update nag message.

## Dive Deeper

In package managers that have their own auto-update mechanism, the update nag message in Gemini CLI can be confusing. While the user may have npm locally in their account, they may be running a version of Gemini CLI from a package manager. A setting allows a package distribution distribute a system config that disables the version update message.  This setting is a peer to disableAutoUpdate - users may still wish to manually update, but know when there is a new version. 

## Linked issues / bugs

Fixes #5447

### DIFF
--- a/packages/cli/src/config/settings.ts
+++ b/packages/cli/src/config/settings.ts
@@ -111,6 +111,9 @@ export interface Settings {
   // Setting for disabling auto-update.
   disableAutoUpdate?: boolean;
 
+  // Setting for disabling the update nag message.
+  disableUpdateNag?: boolean;
+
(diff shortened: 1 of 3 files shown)


jacob314 (approved):
lgtm

PR #5475, opened 2025-08-03
feat(core): Add trailing space when completing an at completion suggestion

SandyTao520:
## TLDR

Add a trailing space after completing an at command suggestion. It also addresses a potential race condition when resetting the completion state.

## Dive Deeper

The `handleAutocomplete` function in the `useCompletion` hook has been modified to always append a trailing space to the completed suggestion. This ensures consistent behavior whether the suggestion is a file path or a command.

To prevent a race condition where the completion state could be reset while suggestions were still being updated, the `useEffect` that handles resetting the completion state now uses `setTimeout(resetCompletionState, 0)`. This defers the reset to the next tick, allowing any pending suggestion updates to complete.

The tests for the `useCompletion` hook have been updated to reflect these changes, ensuring that the new behavior is correctly tested.

## Reviewer Test Plan

1.  Pull down the changes and run the CLI.
2.  Trigger the autocompletion for a file path by typing `@` followed by some characters of a file name in the project.
3.  Use the arrow keys to select a file from the suggestion list and press `Tab` to autocomplete.
4.  **Verify:** A trailing space is added after the file path.
5.  Trigger the autocompletion for a command (e.g., `/`).
6.  Select a command and press `Tab`.
7.  **Verify:** A trailing space is added after the command.
8.  Type a partial command or file path to bring up the suggestion list, then delete the text.
9.  **Verify:** The suggestion list disappears and the UI does not crash or behave unexpectedly.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅ | ❓  | ❓  |
| npx      | ✅ | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


galdawave (approved):
Nit: Do we also want to support this functionality for @files ?

SandyTao520:
This file actually supports @ completion, I'll prepare a PR for rename and some other refactoring.

jacob314 (approved):
lgtm

Issue #5318, opened 2025-07-31
mcp process leak when server is started via command (stdio transport) and `/mcp refresh` is triggered


danicat:
### What happened?

The CLI doesn't clean up MCP Servers that are spin up via commands (using stdio transport) when performing an `/mcp refresh` command.

Steps to reproduce:
- Configure one stdio mcp server in settings.json (via "command" property)
- Start the CLI
- Run `!ps -ef | grep your-mcp-server-command` (should have one count)
- Run `/mcp refresh`
- Run `!ps -ef | grep your-mcp-server-command` (should have two counts)
- Run `/mcp refresh`
- Run `!ps -ef | grep your-mcp-server-command` (should have three counts)

When started:

<img width="1838" height="1072" alt="Image" src="https://github.com/user-attachments/assets/10f4a7ab-b549-4251-bbe4-9d4bcaccd800" />

After 2 refreshes:

<img width="1758" height="554" alt="Image" src="https://github.com/user-attachments/assets/7251a5b0-3559-40a9-b015-82103a0a614e" />

### What did you expect to happen?

No process leak when doing a manual refresh.

### Client information

* **CLI Version:** 0.1.15
* **Git Commit:** c45c14ee (local modifications)
* **Operating System:** darwin v22.16.0
* **Sandbox Environment:** no sandbox
* **Model Version:** gemini-2.5-pro
* **Memory Usage:** 208.5 MB

### Anything else we need to know?

Currently working on a fix, PR will come soon.

danicat:
Seems related to https://github.com/modelcontextprotocol/typescript-sdk/issues/271 and potentially fixed by https://github.com/modelcontextprotocol/typescript-sdk/pull/818 if merged

danicat:
Before connecting again we should close the previous client: https://github.com/google-gemini/gemini-cli/blob/a3a432e3cf5d2de084cecb684229b14ccd4969ac/packages/core/src/tools/mcp-client.ts#L360-L374

While testing a fix for this I noticed that the Go SDK for MCP is not responding to SIGTERM. Will continue exploration tomorrow.

jakemac53:
@danicat were you still working on this?

srivatsj:
I am not able to reproduce for below mcp server. 
mcpServers: {
    "playwright": {
      "command": "npx",
      "args": [
        "@playwright/mcp@latest"
      ]
}

Please reopen if the issue is still present.

Issue #5306, opened 2025-07-31
Robust MCP Support


jacob314:
### What would you like to be added?

This feature captures all enhancements to how we support MCP servers in Core for users to feel that the support is robust.

This means better handling of errors in how MCP servers are configured and unexpected output from MCPs.

### Why is this needed?

Users currently can struggle to setup MCP servers correctly and are confused by cryptic errors from MCP servers.

### Additional context

Some of these bugs may have a single root cause. After you fix one please check the rest of the list to make sure you haven't fixed more issues than you realized.

jakemac53:
See https://github.com/google-gemini/gemini-cli/issues/5689 as well, although I made that have a different parent issue (but we could move it under this one, its very related to each of them).

ddobrin:
Created an issue for MCP notification handling: [issue](https://github.com/google-gemini/gemini-cli/issues/7402)
Would complement /mcp refresh

Adib234:
Thank you for bringing this to our attention. Right now, our engineering team is focusing all resources on critical system maintenance and core stability. Because of this, we don't have immediate plans to address this specific issue, but we truly appreciate you flagging it for us.

PR #5479, opened 2025-08-03
feat: restart MCP servers on /mcp refresh

rmedranollamas:
## TLDR

This PR fixes an bug where /MCP refresh does not restart the servers so no new tools is discovered 

## Dive deeper

The /mcp refresh command now restarts local and remote MCP servers to discover new tools and prompts. This is achieved by introducing a new McpClientManager to handle the server lifecycle, which is now used by the ToolRegistry.

## Reviewer Test Plan

you need to MCP refresh adding a new change on the middle (e.g. a new tool)

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ✅  |
| npx      | ❓  | ❓  | ✅  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Partial fix for #5306 and #5318


packages/core/src/tools/mcp-client-manager.ts
```diff
@@ -0,0 +1,245 @@
...
+
+  /**
+   * Initiates the tool discovery process for all configured MCP servers.
+   * It connects to each server, discovers its available tools, and registers
+   * them with the `ToolRegistry`.
+   */
+  async discoverAllMcpTools(): Promise<void> {
+    await this.startLocalServers();
+    this.discoveryState = MCPDiscoveryState.IN_PROGRESS;
+    const servers = populateMcpServerCommand(
```
NTaylorMullen: So the populate bit comes from mcp-client currently. Should it continue to come from there or be baked into this manager?

packages/core/src/tools/mcp-client-manager.ts
```diff
@@ -0,0 +1,245 @@
...
+ */
+
+import { Client } from '@modelcontextprotocol/sdk/client/index.js';
+import { Transport } from '@modelcontextprotocol/sdk/shared/transport.js';
+import { StdioClientTransport } from '@modelcontextprotocol/sdk/client/stdio.js';
+import { MCPServerConfig } from '../config/config.js';
+import { ToolRegistry } from './tool-registry.js';
+import { PromptRegistry } from '../prompts/prompt-registry.js';
+import {
+  connectToMcpServer,
```
NTaylorMullen: Hmmm, this is a bit ugly / introducing a lot of coupling. We should refactor mcp-client to be a proper class that encapsulates this functionality as connect, discover, status etc. 

I'd imagine after this refactoring all the code here would change drastically

rmedranollamas: 🫡

rmedranollamas:
Implemented the refactor required :)

packages/core/src/tools/mcp-client-manager.ts
```diff
@@ -0,0 +1,110 @@
...
+import { getErrorMessage } from '../utils/errors.js';
+import { WorkspaceContext } from '../utils/workspaceContext.js';
+
+/**
+ * Manages the lifecycle of multiple MCP clients, including local child processes.
+ * This class is responsible for starting, stopping, and discovering tools from
+ * a collection of MCP servers defined in the configuration.
+ */
+export class McpClientManager {
+  private clients: Map<string, McpClient> = new Map();
```
NTaylorMullen, line 24: I imagine a number of these fields could probably be read only

rmedranollamas:
build issue fixed, and also removed a weird race condition with remote servers.

rmedranollamas:
Removed log of the credential. Good catch.

packages/core/src/tools/mcp-client.ts
```diff
@@ -1061,6 +1193,12 @@
   mcpServerConfig: MCPServerConfig,
   debugMode: boolean,
 ): Promise<Transport> {
+  if (debugMode) {
+    console.debug(
+      `[DEBUG] Creating transport for ${mcpServerName} with config:`,
+      mcpServerConfig,
```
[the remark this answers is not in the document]
rmedranollamas: done.

packages/cli/src/ui/commands/mcpCommand.ts
```diff
@@ -471,7 +471,7 @@ const listCommand: SlashCommand = {
 
 const refreshCommand: SlashCommand = {
   name: 'refresh',
```
NTaylorMullen, line 473: Should this be `restart` now?

packages/cli/src/ui/commands/mcpCommand.ts
```diff
@@ -471,7 +471,7 @@ const listCommand: SlashCommand = {
...
-  description: 'Refresh the list of MCP servers and tools',
+  description: 'Restarts MCP servers and refreshes the list of tools.',
```
NTaylorMullen:
```suggestion
  description: 'Restarts MCP servers.',
```

packages/cli/src/ui/commands/mcpCommand.ts
```diff
@@ -497,12 +497,12 @@ const refreshCommand: SlashCommand = {
     context.ui.addItem(
       {
         type: 'info',
-        text: 'Refreshing MCP servers and tools...',
+        text: 'Restarting MCP servers and refreshing tools...',
```
NTaylorMullen:
```suggestion
        text: 'Restarting MCP servers...',
```

NTaylorMullen (approved):

packages/cli/src/ui/commands/mcpCommand.ts
rmedranollamas: we left it as refresh for consistency to the memory command (and the fact that for remote servers, there is nothing to restart).

rmedranollamas:
damn, I forgot to git add the test file with the matching string change.

Issue #5462, opened 2025-08-03
@file includes don't work unless they are followed by a space


danutsu:
### What happened?

Repro:

1. Start a new session in a directory with a two files A.md and B.md.
2. Say "Read @A.md and @B.md. Then tell me what you think the next step is."

Only A.md will be read, B.md will be silently ignored.

Why? Because parsing includes the period after B.md in the file name and then does not find it. Also breaks for comma "Read @A.md, @B.md and @C.md" will skip A. 

### What did you expect to happen?

Gemini reads both A.md and B.md.

### Client information

<details>

```console
$ gemini /about
# paste output here
```

╭──────────────────────────────────────────────────────────────────╮
│                                                                  │
│ About Gemini CLI                                                 │
│                                                                  │
│ CLI Version           0.1.16                                     │
│ Git Commit            820169ba (local modifications)             │
│ Model                 gemini-2.5-pro                             │
│ Sandbox               no sandbox                                 │
│ OS                    linux                                      │
│ Auth Method           OAuth                                      │
│ GCP Project          <would rather not share this information, contact me privately if you need it>                                │
│                                                                  │
</details>

### Login information

Google account with cloud enterprise assist plan.

jacob314:
What we are doing is not technically wrong but is confusing to users. We should be smart enough that trailing `.,` and similar characters are not part of file paths if there exists a match without the character and there does not exist one with the character.

PR #5482, opened 2025-08-03
feat(cli):  Handle Punctuation in @ Command Parsing

SandyTao520:
## TLDR

Enhanced file path escaping to match macOS terminal behavior and fixed critical bugs in @ command parsing and file completion. The @ command parser now properly handles punctuation termination without breaking file extensions, and file paths are correctly escaped for all shell metacharacters. Fixed a critical bug in useSlashCompletion that was destroying shell escaping by replacing all backslashes with forward slashes.

## Dive Deeper

This PR addresses three major file handling issues in the CLI:

### 1. Smart @ Command Parsing - Fixed File Extension Breaking

The original @ command parser used a simple regex `/[.,\s;!?()[\]{}]/` to terminate @ commands at punctuation, but this broke file extensions because it would terminate at the period in `.txt`, `.js`, etc., resulting in incomplete paths like `@file` instead of `@file.txt`.

**The fix**: **Immediate termination** for punctuation that should never be part of file paths (`,`, `;`, `!`, `?`, `(`, `)`, `[`, `]`, `{`, `}`) and **smart termination for periods** - only terminate if followed by whitespace or end of string, allowing file extensions while still handling sentence endings.

**Examples that now work correctly:**
- `"Look at @file.txt, then explain it"` → `@file.txt` + `, then explain it`
- `"Check @config.json. What does it say?"` → `@config.json` + `. What does it say?`
- `"Review @example.d.ts for types"` → `@example.d.ts` + ` for types`
- `"Compare @file1.js, @file2.ts; what's different?"` → Both files parsed correctly

### 2. Enhanced File Path Escaping - Match macOS Terminal Behavior

The `escapePath` function only escaped spaces, leaving users unable to reference files with other shell metacharacters like parentheses, brackets, dollar signs, etc.

**The fix**: Added `SHELL_SPECIAL_CHARS` constant covering all shell metacharacters `[ ()[]{};&|*?$`'"#&<>!~]` and enhanced `escapePath()` function to escape all shell metacharacters like macOS terminal, preventing double-escaping of already escaped characters.

**Examples:**
- `"My Documents/Project (2024)/file [backup].txt"` → `"My\ Documents/Project\ \(2024\)/file\ \[backup\].txt"`
- `"file with $special &chars!.txt"` → `"file\ with\ \$special\ \&chars\!.txt"`

### 3. Fixed useSlashCompletion Critical Bug - Preserved Shell Escaping

Lines 517-520 in `useSlashCompletion.tsx` were replacing ALL backslashes with forward slashes, completely destroying shell escaping and making it impossible to reference files with spaces or special characters.

**The fix**: Smart path normalization using negative lookahead to preserve escape characters while normalizing Windows path separators.

## Reviewer Test Plan

**1. Basic punctuation termination:**
- `"Look at @README.md, what does it contain?"`
- `"Check @package.json; review the dependencies"`
- `"Analyze @src/index.ts! Is there any issue?"`
- `"What's in @config.yaml? Please explain"`

**2. File extensions should be preserved:**
- `"Review @example.d.ts for type definitions"`
- `"Check @version.1.2.3.txt contains version info"`
- `"Look at @script.test.js for test cases"`

**3. Mixed scenarios:**
- `"Compare @file1.txt, @file2.md; what's different?"`
- `"Check @config.json. Then review @package.json"`

**4. Edge cases:**
- `"Show me @file.txt."` (period at end of string)
- `"Use @data.json[0] for the first item"`
- `"Parse @config.yaml{settings} for deployment"`

### Expected Behavior
- File extensions (`.txt`, `.js`, `.md`, etc.) should be preserved in the @ command
- Punctuation after complete file paths should properly terminate the @ command
- The file content should be successfully loaded and displayed
- No parsing errors should occur

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅ | ❓  | ❓  |
| npx      | ✅ | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

#5462


packages/cli/src/ui/hooks/atCommandProcessor.ts
```diff
@@ -87,9 +87,17 @@ function parseAllAtCommands(query: string): AtCommandPart[] {
         inEscape = false;
       } else if (char === '\\') {
         inEscape = true;
-      } else if (/\s/.test(char)) {
-        // Path ends at first whitespace not escaped
+      } else if (/[,\s;!?()[\]{}]/.test(char)) {
```
jacob314, line 90: we have to be a bit more careful. For example to try to break this I created a file named `foo?bar.txt`. We've now broken reading that file as shown in this example.  To fix I'd suggest we make sure our autocomplete automatically escapes the \? and other characters just as a regular terminal on a mac does. So in this case typing tab after `foo` should complete to `foo\?bar.txt`.


```Example failure case:
`
╭───────────────────────────────────────────────╮
│  > @foo?bar.txt what does this file contain?  │
╰───────────────────────────────────────────────╯

 ╭─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
 │ ✔  ReadManyFiles Will attempt to read and concatenate files using patterns: `foo?bar.txt` (within target directory: `/Users/jacobr/git/review/gemini-cli-…  │
 │                                                                                                                                                             │
 │    ### ReadManyFiles Result (Target Dir: `/Users/jacobr/git/review/gemini-cli-review`)                                                                      │
 │                                                                                                                                                             │
 │    Successfully read and concatenated content from **1 file(s)**.                                                                                           │
 │                                                                                                                                                             │
 │    **Processed Files:**                                                                                                                                     │
 │    - `foo?bar.txt`                                                                                                                                          │
 ╰─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
✦ I'm sorry, but I cannot find a file named foo?bar.txt?bar.txt. However, I did find a file named foo?bar.txt. Would you like me to tell you what that file
  contains?

Using: 2 GEMINI.md files
╭──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ >   Type your message or @path/to/file                                                                                                                   │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
~/git/review/gemini-cli-review (better-at-command*)                   no sandbox (see /docs)                     gemini-2.5-pro (99% context left)| 151.4 MB```

File: packages/cli/src/ui/hooks/atCommandProcessor.test.ts [hunk 2]
```diff
@@ -689,5 +689,536 @@ describe('handleAtCommand', () => {
...
+      const result = await handleAtCommand({
+        query,
+        config: mockConfig,
+        addItem: mockAddItem,
+        onDebugMessage: mockOnDebugMessage,
+        messageId: 416,
+        signal: abortController.signal,
+      });
+
+      expect(result).toEqual({
```
jacob314, line 986: nit: these tests are quite repetitive. Ask gemini CLI to turn this into a parameterized test with just the filePath, query, and  expected list of files read as parameters will make this about 4X shorter.

packages/cli/src/ui/hooks/atCommandProcessor.ts
SandyTao520: Done

packages/cli/src/ui/hooks/atCommandProcessor.test.ts
```diff
@@ -689,5 +689,339 @@ describe('handleAtCommand', () => {
...
+          `Parse @${filePath}{prop1: value1}.`,
+        messageId: 409,
+      },
+      {
+        name: 'closing curly brace',
+        fileName: 'config.yaml',
+        fileContent: 'Configuration',
+        queryTemplate: (filePath: string) =>
+          `Use settings from @${filePath}} for deployment.`,
+        messageId: 410,
```
jacob314, line 781: please add some cases where there are special characters within the filenames to ensure those are handled correctly rather than skipped. For example
`foo\?bar.txt` and `foo?bar.txt`

packages/core/src/utils/paths.ts
```diff
@@ -13,6 +13,13 @@ export const GOOGLE_ACCOUNTS_FILENAME = 'google_accounts.json';
 const TMP_DIR_NAME = 'tmp';
 const COMMANDS_DIR_NAME = 'commands';
 
+/**
+ * Special characters that need to be escaped in file paths for shell compatibility.
+ * Includes: spaces, parentheses, brackets, braces, semicolons, ampersands, pipes,
+ * asterisks, question marks, dollar signs, backticks, quotes, hash, and other shell metacharacters.
+ */
+export const SHELL_SPECIAL_CHARS = /[ ()[\]{};|*?$`'"#&<>!~]/;
```
jacob314: nit: include a tab (\t probably) as well

packages/core/src/utils/paths.ts
```diff
@@ -119,26 +126,35 @@ export function makeRelative(
...
+ * asterisks, question marks, dollar signs, backticks, quotes, hash, and other shell metacharacters.
  */
 export function escapePath(filePath: string): string {
   let result = '';
   for (let i = 0; i < filePath.length; i++) {
-    // Only escape spaces that are not already escaped.
-    if (filePath[i] === ' ' && (i === 0 || filePath[i - 1] !== '\\')) {
-      result += '\\ ';
+    const char = filePath[i];
+    const isAlreadyEscaped = i > 0 && filePath[i - 1] === '\\';
```
jacob314: don't you also need to make sure that filePath[i-1] is not also escaped (e.g. proceeding character is not a \
Expect this simplifies to there being an odd number of `\` proceeding the current characters.

jacob314 (approved):
lgtm
Added some minor comments. Feel free to address in a follow up to make this even more robust.

jacob314 (approved):
lgtm

jacob314 (approved):
lgtm

PR #5492, opened 2025-08-04
fix(core): Treat .mts files as TypeScript modules instead of video files

### LINKED ISSUE #4745: ReadFile issues for vite.config.mts
### What happened?

It says  "Read video file: path/to/vite.config.mts". This is a not a video file. Then it does not work afterwards; either no response, or Gemini API shows 500 error.

The file should be read in a short period of time. It has only a few lines.

### Client information

* **CLI Version:** 0.1.13
* **Git Commit:** N/A
* **Operating System:** linux v22.17.1
* **Memory Usage:** 292.1 MB

### Login information

_No response_

_No response_

yashv6655:
## TLDR

Adds support for mts reading from Gemini CLI.

## Dive Deeper

Currenlty, mts files are being processed as video files, which is throwing an API error like `✕ [API Error: Request contains an invalid argument.]`

## Reviewer Test Plan

1. Launch Gemini as you would
2. Ask it to read a .mts file, for example, "read test.mts and explain the content"
3. It should read the mts file normally.
## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs
Resolves #4745


packages/core/src/utils/fileUtils.ts
```diff
@@ -122,9 +122,9 @@ export async function detectFileType(
 ): Promise<'text' | 'image' | 'pdf' | 'audio' | 'video' | 'binary' | 'svg'> {
   const ext = path.extname(filePath).toLowerCase();
 
-  // The mimetype for "ts" is MPEG transport stream (a video format) but we want
+  // The mimetype for "ts" and "mts" is MPEG transport stream (a video format) but we want
   // to assume these are typescript files instead.
-  if (ext === '.ts') {
+  if (ext === '.ts' || ext === '.mts') {
     return 'text';
   }
```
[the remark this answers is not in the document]
yashv6655: done

skeshive (approved):

cornmander (approved):

packages/core/src/utils/fileUtils.ts
```diff
@@ -122,9 +122,10 @@ export async function detectFileType(
 ): Promise<'text' | 'image' | 'pdf' | 'audio' | 'video' | 'binary' | 'svg'> {
   const ext = path.extname(filePath).toLowerCase();
 
-  // The mimetype for "ts" is MPEG transport stream (a video format) but we want
-  // to assume these are typescript files instead.
-  if (ext === '.ts') {
+  // The mimetype for various TypeScript extensions (ts, mts, cts, tsx) can be
+  // MPEG transport stream (a video format), but we want to assume these are
+  // TypeScript files instead.
+  if (['.ts', '.mts', '.cts', '.tsx'].includes(ext)) {
```
jacob314: is tsx really needed? I would remove it from the list and verify that the test still passes. If we were interpreting tsx as media I would expect using `@` for the gemini cli workspace would work very badly because all `tsx` files would be ignored as they would be treated as images instead of text.

jacob314 (commented):
approved once tsx is removed from the list assuming I have not missed something and it is required. Please do leave the test for .tsx. That is good to have to be safe.

packages/core/src/utils/fileUtils.ts
yashv6655: The only reason I included .tsx is that the Gemini bot insisted on it being included in the change. I have now removed it as per your request. The test still passes.

yashv6655:
@jacob314 I have made the changes you have requested and kept the test case. The test case still passes, and everything is functional.

jacob314 (approved):
lgtm
Thanks for the PR!

PR #5508, opened 2025-08-04
fix(ui): Sanitize terminal output and preserve tabular layout in MaxSizedBox (#2286)

### LINKED ISSUE #2286: Curl tool breaks box rendering
### What happened?

Repro steps:

Image

The problem is that we are not stripping control characters from the terminal output before adding it to the box. Characters such as delete will cause this sort of behavior in ink.

Expended the curl output to show in a box without rendering artifacts

### Client information

<details>

```console
$ gemini /about
# paste output here
```

</details>

### Login information

_No response_

_No response_

busycaesar:
## TLDR

This PR fixes a rendering issue in `MaxSizedBox` where control characters (like backspace or delete) from terminal output caused visual artifacts. It introduces a more robust `sanitize()` function and adds logic to preserve the formatting of tabular text, such as curl progress output.

## Dive Deeper

Previously, terminal output that included control characters (e.g., ANSI escape codes or backspaces) could corrupt the layout in Ink-based UIs. This was especially problematic with structured or tabular output (like progress bars or column-aligned text), which would be misaligned or broken in the UI.

This PR introduces the following improvements:

- A `sanitize()` function to remove problematic ANSI escape sequences and control characters, while preserving structural whitespace (tabs, multiple spaces, newlines).
- A `looksLikeTable()` heuristic that detects tabular text by analyzing line patterns. If such formatting is detected, it's rendered with special care to maintain column alignment.

Safeguards to skip empty segments and ensure consistent layout rendering even for multi-line structured output. These changes significantly improve visual fidelity and reliability for structured terminal output within the CLI.

## Reviewer Test Plan

Once the changes are pulled and the project is started, turn on the shell mode. Then, to validate the fix, run a `curl` command with a valid URL that tends to produce a tabular output, like the following,

```
curl -k https://shahtech.info/
```

After running the command, observe the output.
- It should visually look like a tabular structure.
- Each cell in each row should match the formatting of its adjacent rows as well as the header row.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ✅ |
| npx      | ❓  | ❓  | ✅ |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

- `Fixes #2286`


jacob314:
Thanks for your contribution! However, @galz10 is looking at a more robust solution for this using the xterm headless library to properly apply terminal escape codes. If that can land soon I'd rather go with that approach than these heuristics which while reasonable have risks of false positives as well as performance overhead. Thanks you for the contribution and do let me know if there are any other issues you would like to tackle.
In particular you might find the issues under https://github.com/google-gemini/gemini-cli/issues/4245 interesting.

busycaesar:
Thank you @jacob314 for the opportunity. I will explore more issues and let you know if there is something that I would like to solve.

PR #5522, opened 2025-08-04
Improve user-facing error messages for IDE mode

skeshive:
## Linked issues / bugs
#4390

### DIFF
--- a/packages/cli/src/ui/commands/ideCommand.ts
+++ b/packages/cli/src/ui/commands/ideCommand.ts
@@ -25,7 +25,7 @@ export const ideCommand = (config: Config | null): SlashCommand | null => {
   }
   const ideClient = config.getIdeClient();
   const currentIDE = ideClient.getCurrentIde();
-  if (!currentIDE) {
+  if (!currentIDE || !ideClient.getDetectedIdeDisplayName()) {
     return {
       name: 'ide',
       description: 'manage IDE integration',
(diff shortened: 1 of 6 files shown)


packages/cli/src/ui/commands/ideCommand.ts
```diff
@@ -70,15 +86,15 @@ export const ideCommand = (config: Config | null): SlashCommand | null => {
...
+    description: `install required IDE companion for ${ideClient.getDetectedIdeDisplayName()}`,
     kind: CommandKind.BUILT_IN,
     action: async (context) => {
       const installer = getIdeInstaller(currentIDE);
       if (!installer) {
         context.ui.addItem(
           {
             type: 'error',
-            text: 'No installer available for your configured IDE.',
+            text: `No installer is available for ${ideClient.getDetectedIdeDisplayName()}. Please install the IDE companion manually from its marketplace.`,
```
chrstnb, line 97: This seems fine but in the future there could be other types of installs? I don't feel strongly tho

skeshive: ack, we can update this once needed. I did remove all mentions of extension from the command itself since other IDEs like Intellij use the term "plugin"

chrstnb (approved):

packages/core/src/ide/ide-installer.ts
```diff
@@ -95,16 +94,12 @@ class VsCodeInstaller implements IdeInstaller {
...
-    return (await this.vsCodeCommand) !== null;
-  }
-
   async install(): Promise<InstallResult> {
     const commandPath = await this.vsCodeCommand;
     if (!commandPath) {
       return {
         success: false,
-        message: `VS Code command-line tool not found in your PATH or common installation locations.`,
+        message: `VS Code CLI not found. Please ensure 'code' is in your system's PATH. For help, see https://code.visualstudio.com/docs/configure/command-line#_code-is-not-recognized-as-an-internal-or-external-command. You can also install the companion extension manually from the VS Code marketplace.`,
```
NTaylorMullen, line 102: You may want to make a goo.gle link for the VSCode instructions + one for the yet-to-be-created companion extension in the marketplace

NTaylorMullen (approved):

PR #5526, opened 2025-08-04
remove unnecessary checks in WriteFileChecks.getDescription

### LINKED ISSUE #2376: WriteFile "Parameters failed schema validation" when MCP server defined
### What happened?

When I ask Gemini to edit a file, it keeps throwing an error.

Image

edit file

### Client information

Image

### Login information

_No response_

_No response_

jakemac53:
## TLDR

Remove bad validation of the "content" parameter  in WriteFileTool.getDescription - this previously would return an error message if the content was the empty string.

Add a new error specific to empty paths - although validateToolParams would already handle this, we probably shouldn't assume it has been called.

## Reviewer Test Plan

You can test this by just asking Gemini to create any empty file - previously it would output some error text about invalid params (although it would actually then succeed which is interesting). The schema was ultimately valid and this is just a description of the command I guess.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Closes https://github.com/google-gemini/gemini-cli/issues/2376

### DIFF
--- a/packages/core/src/tools/write-file.test.ts
+++ b/packages/core/src/tools/write-file.test.ts
@@ -13,7 +13,7 @@ import {
   vi,
   type Mocked,
 } from 'vitest';
-import { WriteFileTool } from './write-file.js';
+import { WriteFileTool, WriteFileToolParams } from './write-file.js';
 import {
   FileDiff,
   ToolConfirmationOutcome,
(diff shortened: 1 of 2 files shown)


jacob314 (approved):
lgtm

PR #5529, opened 2025-08-04
feat(core): Parse Multimodal MCP Tool responses

### LINKED ISSUE #2136: Support Image Content in MCP tool results

Support the [Image Content MCP tool result data type](https://modelcontextprotocol.io/specification/2025-06-18/server/tools#image-content). When an MCP server responds with base64 encoded image data as part of its tool result, Gemini is not able to "see" the image.

Let's improve Gemini CLI to allow it to "see" images within MCP tool responses.

I can currently think of two possible approaches:

1. **Generative AI SDK**
   - Have image content be natively supported in the [`functionResponse`](https://googleapis.github.io/js-genai/release_docs/interfaces/types.Part.html#functionresponse) Part type of the Gemini Generative AI SDKs. There is currently no indication ([function calling example](https://ai.google.dev/gemini-api/docs/function-calling?example=meeting#step-4)) that multimodal content that the model will understand as multimodal is supported in function responses. Only text or structured/JSON text.
   - Or introduce a new response content type specific to MCP tool use, that is distinct and able to closely match the MCP spec.  Example: https://docs.anthropic.com/en/docs/agents-and-tools/mcp-connector#response-content-types
2. **Gemini CLI**
   - Modify the Gemini CLI, as an MCP app, to intercept image content in MCP tool responses.
   - Potentially substitute a text placeholder: "The tool responded with an image here. The image can be found after this function response."
   - Then add `inlineData` with the base64 image data, pushed as a new `Part` after the `functionResponse`.
   - Possible code location to understand this suggestion: https://github.com/google-gemini/gemini-cli/blob/bf873a1d85d4400fd22c175db0f3b80a47bc27fb/packages/core/src/tools/mcp-tool.ts#L84-L89
   - Bonus: I'd yolo support for MCP [Audio Content](https://modelcontextprotocol.io/specification/2025-06-18/server/tools#audio-content) here as well and see if it works! [I don't see why it wouldn't](https://ai.google.dev/gemini-api/docs/audio#inline-audio). 🤷 

### Why is this needed?

This autonomous agent loop is [getting popular](https://www.youtube.com/live/6eBSHbLKuN0?si=FpDz_DklOouQ1sjH&t=674):

```
Start with a UI mock -> write UI code -> take_screenshot -> edit code -> take_screenshot -> ..
```

MCP tools such as Puppeteer that are able to return screenshot image responses will be very handy to make this work within Gemini CLI.

### Additional context

This is a similar or the same issue as with Cline: https://github.com/cline/cline/issues/1865

luccasparoni:
## TLDR

This pull request enables the Gemini CLI to correctly parse and process rich, multi-modal content (e.g., text, images, audio) from MCP tools (compliant with MCP specification). Previously, **this data was either ignored or handled incorrectly**, with only plain text being passed to the model.

Important to look: 
- Text only MCP still works
- Multimodal MCP content is interpreted.

## Dive Deeper

The CLI's previous implementation for handling MCP tool responses was limited and not fully spec-compliant. It relied on a coincidence that only extracted plain text from a tool's output, effectively discarding any rich data like images or audio that a tool might have sent
(prior to this change, any media data was returned inside a functionResponse and discarted at the `getResponseTextFromParts` from the `coreToolScheduler`.
This PR fixes the client-side handling of MCP tool responses, making it robust and feature-complete.

`mcp-tool.ts` was refactored to correctly parse the ContentBlock array from an MCP response. Instead of just looking for .text properties, it now has dedicated logic to transform each MCP block type (text, image, audio, resource, etc.) into the corresponding GenAI Part structure (text or inlineData).
The display string has also been modified in order to indicate o the user that a media has been sent to the model(example: `[Image: image/png]`. 

The relevant documentation has also been updated to reflect that the CLI now fully supports rich content from tools.

image
_Note: In the example above the MCP returns the image and the "Image URL:..." text_

## Reviewer Test Plan

- Verify older MCPs still work correctly
- Verify an MCP that returns media (alone or with text) works and that the model understands the media. 

I have a very simple python MCP that can be used to test available [in this repository.](https://github.com/luccasparoni/url-fetch-mcp) 

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅   | ❓  | ✅  |
| npx      | ✅  | ❓  | ✅  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

- Fixes #2136


packages/core/src/tools/mcp-tool.ts
```diff
@@ -21,6 +21,7 @@ import {
 } from '@google/genai';
 
 type ToolParams = Record<string, unknown>;
+type McpContentBlock = Record<string, string | Record<string, string>>;
```
jacob314: This code would be a bit more well typed if you change the McpContentBlock type to more fully leverage the TypeScript type system. Expect GeminiCLI can help you do that if you ask it to define types so that 
```  (block: McpContentBlock): Part | Part[] | null => {
      switch (block.type) {
        case 'text':
          return transformTextBlock(block);
        case 'image':
        case 'audio':
          return transformImageAudioBlock(block, toolName);
        case 'resource':
          return transformResourceBlock(block, toolName);
        case 'resource_link':
          return transformResourceLinkBlock(block);
        default:
          return null;
      }
    },```
can be type safe removing the `as` calls currently required in those helper methods.

If Gemini CLI is unclear what to do you could tell it to leverage UnionTypes to achieve this where each clause in the union type has a different value for the string field type.

File: packages/core/src/tools/mcp-tool.ts [hunk 2]
```diff
@@ -114,70 +115,151 @@ export class DiscoveredMCPTool extends BaseTool<ToolParams, ToolResult> {
...
+          ((block.resource as McpContentBlock)?.mimeType as string) ||
+          'unknown type'
+        }]`;
+      default:
+        return `[Unknown content type: ${block.type as string}]`;
+    }
+  });
+
+  return displayParts.join('\n');
 }
```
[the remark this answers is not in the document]
jacob314: this would work but you can do even better by leveraging the Type system as mentioned above.

jacob314 (commented):
Thanks for this pull request! Users will really appreciate this feature. All the code looks high quality. I will approve after the hopefully minor comments are addressed.

docs/core/tools-api.md
```diff
@@ -15,9 +15,11 @@ The Gemini CLI core (`packages/core`) features a robust system for defining, reg
   - `execute()`: The core method that performs the tool's action and returns a `ToolResult`.
 
 - **`ToolResult` (`tools.ts`):** An interface defining the structure of a tool's execution outcome:
-  - `llmContent`: The factual string content to be included in the history sent back to the LLM for context.
+  - `llmContent`: The factual content to be included in the history sent back to the LLM for context. This can be a simple string or a `PartListUnion` (an array of `Part` objects and strings) for rich content.
   - `returnDisplay`: A user-friendly string (often Markdown) or a special object (like `FileDiff`) for display in the CLI.
 
+- **Returning Rich Content:** Tools are not limited to returning simple text. The `llmContent` can be a `PartListUnion`, which is an array that can contain a mix of `Part` objects (for images, audio, etc.) and `string`s. This allows a single tool execution to return multiple pieces of rich content. The `CoreToolScheduler` is responsible for correctly processing this array, packaging all text content into a single `FunctionResponse`, and sending any binary data (like images) as separate top-level `Part`s for the model to understand.
```
jacob314: nit: omit the part starting with "The `CoreToolScheduler` is
as that is more of an implementation detail.

packages/core/src/tools/mcp-tool.ts
luccasparoni: Done. Thanks, this is way more clear now!

packages/core/src/tools/mcp-tool.ts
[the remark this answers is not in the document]
luccasparoni: Ack. Implemented using type system.

docs/core/tools-api.md
luccasparoni: Done.

jacob314 (approved):
lgtm

That cleanup turned out great! How much was Gemini CLI able to do and how much did you have to do manually?

luccasparoni:
> lgtm
> 
> That cleanup turned out great! How much was Gemini CLI able to do and how much did you have to do manually?

100% Gemini, but it needed some iterations: first try it made everything more complicated, with some more guidance it was able to make it correct!

PR #5536, opened 2025-08-04
Zed integration schema upgrade

agu-z:
## TLDR

This PR updates the experimental Zed integration to a newer schema, which is more flexible and redesigned to be easily embedded in other base protocols such as MCP or A2A.

Co-Authored-By: Ben Brandt [benjamin.j.brandt@gmail.com](mailto:benjamin.j.brandt@gmail.com)

## Dive Deeper

The new schema was redesigned to be easier to embed in other protocols and to align with existing ones by:

- Streaming content via notifications rather than method calls (MCP friendly)
- Using MCP's [ContentBlock](https://modelcontextprotocol.io/specification/2025-06-18/schema#contentblock) type for all content
- Using MCP/LSP version negotiation

Additional enhancements in the redesign:
- The agent exposes all authentication methods supported 
- The client can expose MCP servers to the Agent, which can be used to power IDE mode or to provide tools directly to the model
- Requests are routed by session ID so that multiple conversations can occur in parallel
- Introduces the concept of capabilities for non-breaking feature additions
- Tool kinds have replaced the concept of icons, which seemed too UI-specific
- Zod schemas are used for safe parsing and validation

The following capabilities are supported, but haven't been implemented in this PR to keep the diff primarily mechanical:
- Client file system access
- Session restoration

## Reviewer Test Plan

To fully test the integration, you'll need to [build Zed from source](https://github.com/zed-industries/zed?tab=readme-ov-file#developing-zed), and add the following to your settings:

```
{
  "agent_servers": {
    "gemini": {
      "command": "node",
      "args": ["/<PATH>/<TO>/gemini-cli/packages/cli"]
    }
  },
}
```

Feel free to ask @agu-z or @benbrandt for any help getting set up.


gsquared94 (approved):

gsquared94:
@agu-z  can you please rebase?

agu-z:
Thanks for the review @gsquared94! Just rebased it.

packages/cli/src/zed-integration/zedIntegration.ts
mattKorwel, line 1: Thanks for updating the naming here. This will make it clearer to future folks that the integration here is intended just for zed. We'll work to migrate this to mcp/a2a in the coming weeks. Thanks all!

packages/core/src/tools/edit.ts
```diff
@@ -9,7 +9,7 @@ import * as path from 'path';
 import * as Diff from 'diff';
 import {
   BaseDeclarativeTool,
-  Icon,
+  Kind,
```
mattKorwel, line 12: Are these icon changes net new @ConradIrwin @agu-z and are they aligned with your understanding of how things should be @jacob314 ?

mattKorwel (approved):
LGTM. Pending a quick confirm on the icon stuff from @jacob314

jacob314 (commented):
lgtm

As farI can tell the icons were only used in ACP so no concerns about changing them.

scidomino:
I think I found an error here in that this code treats every ToolResult as a success even though it has an error field that you should be looking at to figure that out.

benbrandt:
> I think I found an error here in that this code treats every ToolResult as a success even though it has an error field that you should be looking at to figure that out.

@scidomino thanks for reporting! We addressed this in our updates to the tool call and prompt handling here: https://github.com/google-gemini/gemini-cli/pull/6756/commits/b649febb37cac3b776fbefcccdb633ac735c5910
https://github.com/google-gemini/gemini-cli/pull/6756

Issue #5537, opened 2025-08-04
give users the ability to add instruction to the /compress command


shamso-goog:
### What would you like to be added?

give users the ability to add instruction to the /compress command

### Why is this needed?

compression often ends the effective window for interacting with the cli, compression allows the user to steer compression towards their immediate goals

olcan:
Thank you for the feature request. If you are able to provide specific examples (scenarios + instructions) that would be helpful.

shamso-goog:
thanks its often the case that there is something specific from your context you want to it focus on so "focus on debugging X" or "focus on building Y" fwiw other clis have a similar command

PR #5538, opened 2025-08-04
feat: Add instruction to compress command

shamso-goog:
## TLDR

  This pull request enhances the /compress command by allowing users to provide an optional instruction to guide the context summary. For example, a user can now type /compress focus on recent file modifications.
  
## Dive Deeper

  The core change involves passing an optional instruction string from the compressCommand action down to the GeminiClient.tryCompressChat method.

   - packages/cli/src/ui/commands/compressCommand.ts: The command's action now accepts an instruction argument, which is passed to tryCompressChat. The command's description has also been updated to reflect this new
     capability.
   - packages/core/src/core/client.ts: The tryCompressChat method signature was updated to accept the optional instruction string.
   - packages/core/src/core/prompts.ts: The getCompressionPrompt function now conditionally injects a <user_instruction> XML block into the system prompt if an instruction is provided. This ensures the model is aware of the
     user's specific summarization request.
   - Associated test files (compressCommand.test.ts, client.test.ts, prompts.test.ts) have been updated to cover the new functionality and ensure that providing no instruction preserves the original behavior.
## Reviewer Test Plan

   1. Pull down this branch and run npm run preflight to ensure all checks pass.
   2. Start the Gemini CLI.
   3. Type /compress and press space. Verify that the command description now includes Optional: /compress [instruction for compression].
   4. Execute the command without any arguments: /compress. The command should function as it did previously.
   5. Execute the command with a specific instruction, for example: /compress please summarize the changes made to the client.ts file.
 
## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  |  ✅    |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

https://github.com/google-gemini/gemini-cli/issues/5537


packages/core/src/core/prompts.ts
```diff
@@ -317,7 +323,7 @@ The structure MUST be as follows:
         <!-- A single, concise sentence describing the user's high-level objective. -->
         <!-- Example: "Refactor the authentication service to use a new JWT library." -->
     </overall_goal>
-
+${instructionXml}
```
jacob314: this doesn't seem rlike the right place to include instructionXml as this is part of where the snapshot format is described not the instructions. Naively I would think the instructions should go  in a paragraph added before line 317

packages/cli/src/ui/commands/compressCommand.test.ts
```diff
@@ -90,6 +91,22 @@ describe('compressCommand', () => {
...
+    const compressedResult = {
+      originalTokenCount: 200,
+      newTokenCount: 100,
+    };
+    mockTryCompressChat.mockResolvedValue(compressedResult);
+
+    await compressCommand.action!(context, 'test instruction');
+
+    expect(mockTryCompressChat).toHaveBeenCalledWith(
+      expect.stringMatching(/^compress-\d+$/),
```
jacob314: specify where the new instructions occur in the output rather than just that they occur. this could be achieved with a snapshot test. you can see other examples of text snapshot tests including one for the system prompt.

packages/core/src/core/client.ts
```diff
@@ -677,7 +677,8 @@ export class GeminiClient {
 
   async tryCompressChat(
     prompt_id: string,
-    force: boolean = false,
+    force = false,
```
jacob314: nit: avoid multiple optional positional arguments where feasible. either make these arguments named or required.

jacob314 (commented):
thanks for adding. this will be a very useful feature! one other thing to consider is how to use this to control the size of the compressed output. As part of implementing, we should ensure that it is possible to use this compress the chat history to a reasonable size with reasonable instructions to control the size of the output.

packages/core/src/core/prompts.ts
```diff
@@ -300,7 +300,13 @@ Your core function is efficient and safe assistance. Balance extreme conciseness
...
  * think in a scratchpad, and produce a structured XML summary.
  */
-export function getCompressionPrompt(): string {
+export function getCompressionPrompt(instruction?: string): string {
+  const instructionXml = instruction
+    ? `
+    <user_instruction>
+        ${instruction}
+    </user_instruction>`
+    : '';
```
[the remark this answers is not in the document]
anj-s: @cornmander Can you please take a look at this and verify that this is safe to do?

packages/core/src/core/prompts.ts
anj-s: How do we deal with instructions that contradict what we have in the prompt? I think either we should use the user instructions or what we have already.

shamso-goog: moved it, let me know what you think. thanks.

packages/cli/src/ui/commands/compressCommand.test.ts
shamso-goog: done

packages/core/src/core/client.ts
shamso-goog: done

packages/core/src/core/prompts.ts
anj-s: @shamso-goog I don't think my comment was addressed by the move. Can you comment on it?

shamso-goog: ah sorry didnt think you were asking me. i mean the user can always give us instructions that contradict ours, in any of our interfaces. i personally tend not to design my system around such cases.

packages/core/src/core/prompts.ts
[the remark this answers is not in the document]
jacob314: I also remain a little confused why this part should be xml at all as I thought the XML was just the part describing the output not the instructions. I'll defer to @anj-s on this though and approve once they approve this PR as this is more in their area of expertise aside from the concerns that have already been addressed about how the API for this integrates with the rest of the core package.

agmsb:
Hey @shamso-goog - thanks for the PR! Did you still want to move forward with merging? If so, could you address conflicts in the branch?

shamso-goog:
> Hey @shamso-goog - thanks for the PR! Did you still want to move forward with merging? If so, could you address conflicts in the branch?

yes, thank you. fixed the conflicts.

docs/cli/commands.md
```diff
@@ -35,8 +35,8 @@ Slash commands provide meta-level control over the CLI itself.
   - **Description:** Clear the terminal screen, including the visible session history and scrollback within the CLI. The underlying session data (for history recall) might be preserved depending on the exact implementation, but the visual display is cleared.
   - **Keyboard shortcut:** Press **Ctrl+L** at any time to perform a clear action.
 
-- **`/compress`**
-  - **Description:** Replace the entire chat context with a summary. This saves on tokens used for future tasks while retaining a high level summary of what has happened.
+- **`/compress [instructions]`**
+  - **Description:** Replace the entire chat context with a summary. This saves on tokens used for future tasks while retaining a high level summary of what has happened. You can optionally provide instructions on how the compression should be performed.
```
agmsb: Could we be more opinionated here around what the instructions should tell the model to do around compression? 

For example, by "should be performed", do we mean areas we want the model to focus on in the generated summary?

shamso-goog: thanks! done

agmsb (approved):

jacob314:
I got the following behavior trying to use this to compress my chat history. Without this change I am not able to compress at all for this simple prompt:
`@packages/cli there is flicker after each keystroke when autocompleting file paths. Analyze the problem and suggest 3 solutions.`
with your changes I was eventually able to compress but the experience wasn't very robust.

image
Fyi @scidomino

shamso-goog:
> I got the following behavior trying to use this to compress my chat history. Without this change I am not able to compress at all for this simple prompt: `@packages/cli there is flicker after each keystroke when autocompleting file paths. Analyze the problem and suggest 3 solutions.` with your changes I was eventually able to compress but the experience wasn't very robust.
> 
> image Fyi @scidomino

thank you, are you suggesting i make a change? or that it is working at intended or provides more value than not having it?

elvinagam:
@shamso-goog resolve conflicts for merge

elvinagam (approved):

PR #5567, opened 2025-08-05
Fix(vim): Fix shell mode in Vim mode

### LINKED ISSUE #5563: [Vim Mode] Can't use Shell Mode when in Vim Mode
### What happened?

When in vim mode, typing '!' to enter shell mode in the CLI doens't work.  It's a no-op.

Shell mode to be invokable in Vim mode (in insert mode)

### Client information

<details>

```console
│ CLI Version          0.1.17                                 │
│ Model                gemini-2.5-pro                         │
│ Sandbox              no sandbox                             │
│ OS                   linux                                  │
│ Auth Method          vertex-ai                              │
│ GCP Project          jmahood-demo
```

</details>

### Login information

Vertex

_No response_

justinmahood:
## TLDR

This PR allows the use of shell mode when in vim mode, and improves the handling of key presses in vim mode.

  The following changes are included:
  - Allows '!' to pass through to the shell in vim mode (when at the beginning of the line), enabling access to shell mode
  - Allows Ctrl+R to pass through to the shell, enabling access to shell history.
   - Allows Esc to pass through in normal mode, allowing return to the previous modes.

  Fixes #5563

## Dive Deeper

Vim mode was too eagerly 'eating' keystrokes necessary to enter shell mode and do shell search.  This change allows those to pass through in the appropriate circumstances, and also allows for 'esc' to be used conditionally to exit search and shell modes, while maintaining vim-mode behavior. 

## Reviewer Test Plan

* Launch the CLI,  enter vim mode (/vim)
* Test shell mode (!) and command search (ctrl+r)
* return back to the CLI input

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ✅  |
| npx      | ❓  | ❓  | ✅  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes #5563


jacob314 (commented):
This seems like it inverting things a bit. Can we just have InputPrompt handle both `!` and ctrl-R before delegating to vim? having comments in vim.ts explaining that we are doing this because of inputPrompt having similar logic is a code smell. As long as we are ok with these two modes behaving normally regardless of which vim mode you are in I think that is more robust.
Fyi @sijieamoy who is more of a vim expert than I am. Right now this PRs current behavior seems a bit odd as `!` only works if you are already in insert mode rather than working in normal mode. I'm not vim users so unsure what vim users would prefer.

justinmahood:
> This seems like it inverting things a bit. Can we just have InputPrompt handle both `!` and ctrl-R before delegating to vim? having comments in vim.ts explaining that we are doing this because of inputPrompt having similar logic is a code smell. As long as we are ok with these two modes behaving normally regardless of which vim mode you are in I think that is more robust. Fyi @sijieamoy who is more of a vim expert than I am. Right now this PRs current behavior seems a bit odd as `!` only works if you are already in insert mode rather than working in normal mode. I'm not vim users so unsure what vim users would prefer.

@jacob314  FWIW I mapped the behavior to what I expect / prefer from using 'vim mode' in zsh.   In normal mode in vim, many keys are used for vim specific behavior by default, and having the global shortcuts  take precedence can break muscle memory.  Ctrl+r in particular is used for "redo" in normal mode.

jacob314:
> > This seems like it inverting things a bit. Can we just have InputPrompt handle both `!` and ctrl-R before delegating to vim? having comments in vim.ts explaining that we are doing this because of inputPrompt having similar logic is a code smell. As long as we are ok with these two modes behaving normally regardless of which vim mode you are in I think that is more robust. Fyi @sijieamoy who is more of a vim expert than I am. Right now this PRs current behavior seems a bit odd as `!` only works if you are already in insert mode rather than working in normal mode. I'm not vim users so unsure what vim users would prefer.
> 
> @jacob314 FWIW I mapped the behavior to what I expect / prefer from using 'vim mode' in zsh. In normal mode in vim, many keys are used for vim specific behavior by default, and having the global shortcuts take precedence can break muscle memory. Ctrl+r in particular is used for "redo" in normal mode.

Ok I'll defer to your expertise as a vim user.

jacob314 (approved):
lgtm

PR #5590, opened 2025-08-05
feat: Add thinking traces display support for gemini-2.5 models

aelaguiz:
## Summary

This PR adds support for displaying thinking traces when using gemini-2.5 models, similar to how they appear in other tools like Aider.

## What's Changed

- Set 32k thinking token budget for gemini-2.5 models with temperature 0
- Add new  type to display thinking traces in the UI
- Create  component with purple thinking emoji (💭) and formatted display
- Update  to add thinking events to history
- Export  function for UI use
- Display thinking budget (32,768 tokens) in AboutBox for supported models
- Remove all debug logging for cleaner output

## Motivation

When using gemini-2.5-pro models, thinking traces provide valuable insight into the model's reasoning process. This PR ensures these traces are properly displayed in the UI rather than being hidden.

## Testing

Tested locally with gemini-2.5-pro-002 model to verify:
- Thinking traces appear with purple emoji and formatted subject/description
- Thinking budget shows correctly in /about command
- No debug spam in output
- TypeScript builds without errors

### DIFF
--- a/packages/cli/src/ui/components/AboutBox.tsx
+++ b/packages/cli/src/ui/components/AboutBox.tsx
@@ -8,6 +8,7 @@ import React from 'react';
 import { Box, Text } from 'ink';
 import { Colors } from '../colors.js';
 import { GIT_COMMIT_INFO } from '../../generated/git-commit.js';
+import { isThinkingSupported } from '@google/gemini-cli-core';
 
 interface AboutBoxProps {
   cliVersion: string;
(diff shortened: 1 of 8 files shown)


jacob314:
Can you link this bug to an issue?

anthrotype:
> Can you link this bug to an issue?

@jacob314 I suppose this one https://github.com/google-gemini/gemini-cli/issues/2456

packages/cli/src/ui/components/AboutBox.tsx
```diff
@@ -71,6 +72,18 @@ export const AboutBox: React.FC<AboutBoxProps> = ({
...
     </Box>
+    {isThinkingSupported(modelVersion) && (
+      <Box flexDirection="row">
+        <Box width="35%">
+          <Text bold color={Colors.LightBlue}>
+            Thinking Budget
+          </Text>
+        </Box>
+        <Box>
+          <Text color={Colors.AccentGreen}>32,768 tokens</Text>
```
anthrotype, line 83: this magic number is hard-coded and only applies to 2.5 Pro; Flash or Flash lite have different maximums.

anthrotype:
> Set 32k thinking token budget for gemini-2.5 models with temperature 0

I think this particular change is idependent from better displaying of thinking traces, ideally that should be a separate PR

jacob314:
> > Set 32k thinking token budget for gemini-2.5 models with temperature 0
> 
> I think this particular change is idependent from better displaying of thinking traces, ideally that should be a separate PR

agreed. please move that to a separate pr so this pull request does not change model behavior.

anthrotype:
Another thing that would be nice is if these thinking summaries could be toggled on/off in the settings

chrstnb:

PR #5601, opened 2025-08-05
fix: Restore user input when the user cancels response

shishu314:
## TLDR

Restores a user's previous input when they press escape to cancel

## Dive Deeper

I think it makes more sense to have the state management in App.tsx instead of pulling the buffer and state into useGemini since the app holds all the info and hooks (from input + useGemini)

## Reviewer Test Plan

Press ESC after sending a message and you should see that your sent message is restored. Currently the sent message is stored in history and you see the input remain the same if we press up(which doesn't feel like a bug but we can fix forward)

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | x  | ❓  | ❓  |
| npx      | x  | ❓  | ❓  |
| Docker   | x  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | x  | -   | -   |

## Linked issues / bugs

- Resolves #5243


packages/cli/src/ui/App.tsx
```diff
@@ -486,6 +486,20 @@ const App = ({ config, settings, startupWarnings = [], version }: AppProps) => {
...
+    viewport: { height: 10, width: inputWidth },
+    stdin,
+    setRawMode,
+    isValidPath,
+    shellModeActive,
+  });
+
+  const [previousUserInput, setPreviousUserInput] = useState('');
+  const handleUserCancel = useCallback(() => {
+    buffer.setText(previousUserInput);
```
jacob314: nit: rather than tracking previousUserInput, can you grab the previous input from the last entry in history instead. Expect Gemini CLI should be able to wire that up.

packages/cli/src/ui/App.tsx
```diff
@@ -491,10 +491,15 @@ const App = ({ config, settings, startupWarnings = [], version }: AppProps) => {
     shellModeActive,
   });
 
-  const [previousUserInput, setPreviousUserInput] = useState('');
   const handleUserCancel = useCallback(() => {
-    buffer.setText(previousUserInput);
-  }, [buffer, previousUserInput]);
+    const lastUserItem = history
+      .slice()
```
jacob314: I believe you should just be able to write `userMessages.at(-1)`

jacob314 (approved):
lgtm

nxnmsl0:
-  [ x] #5601

PR #4890, opened 2025-07-25
fix(tests): refactor integration tests to be less flaky

allenhutchison:
fix(tests): refactor integration tests to be less flaky

This PR addresses flakiness in the integration test suite by refactoring tests to use telemetry-based assertions instead of relying on non-deterministic LLM output.

### Test Infrastructure Improvements

1. **Telemetry-based assertions**: Tests now verify tool calls through telemetry logs instead of parsing LLM output
   - Added `waitForToolCall()`, `waitForAnyToolCall()`, and `readToolLogs()` methods to TestRig
   - Tests assert on actual tool invocations, not on how the LLM describes them
   - Added `waitForTelemetryReady()` to fix race conditions in telemetry collection

2. **Local telemetry collection**: Each test run uses a local telemetry file
   - Configured via `settings.json` with `telemetry.target: 'local'`
   - Telemetry writes to a test-specific log file for isolation

3. **Container sandbox support**: Fixed integration tests to work properly in Docker and Podman
   - Fixed telemetry file paths to use relative paths in sandbox mode
   - Made MCP server test work in containers by using a self-contained implementation
   - Added Podman telemetry parsing from stdout when file telemetry is unavailable

4. **Environment-aware timeouts**: Added adaptive timeout configuration
   - 60 seconds in CI environments
   - 30 seconds in container sandboxes
   - 15 seconds for local development

5. **Better error diagnostics**: Added helper functions for improved debugging
   - `createToolCallErrorMessage()` provides detailed error context
   - Enhanced debug output shows all tool calls when tests fail

### Specific Test Fixes

- **google_web_search.test.js**: 
  - Now verifies the tool was called via telemetry
  - Added network error handling for CI environments
  - Gracefully skips test on network failures

- **run_shell_command.test.js**: 
  - Added debugging output and made assertions more flexible
  - Added warnings for empty LLM output

- **write_file.test.js**: 
  - Simplified to check file contents directly
  - Added better error messages using createToolCallErrorMessage helper

- **replace.test.js**: 
  - Verifies file modifications without parsing LLM output

- **save_memory.test.js**: 
  - Checks telemetry for save_memory tool call

- **list_directory.test.js**: 
  - Validates directory listing via telemetry
  - Added explicit filesystem sync wait for containers

- **read_many_files.test.js**: 
  - Accepts either read_many_files OR multiple read_file calls
  - More flexible to handle different LLM tool choices

- **file-system.test.js**:
  - Updated write test to accept write_file, edit, or replace tools
[...]

PR #5609, opened 2025-08-05
Detect and warn about cyclic tool refs when schema depth errors are encountered

jakemac53:
## TLDR

When we see an API Error related to maximum schema depth:
- Don't retry the error
- Check for likely problematic tools with cyclic schema definitions, and attach extra context to the error indicating the user should try disabling those tools. 

The error now looks like this:

image

## Dive Deeper

I found that it isn't so simple as Gemini doesn't support any cyclic refs - they actually do support some of them. There are just certain specific structures which are not allowed.

So, I chose not to pre-emptively disable all tools with cyclic schemas, and instead just improve the error when we encounter it.

## Reviewer Test Plan

Add an MCP server with a tool schema like this one (definitely triggers the issue):

```json
        {
          "type": "object",
          "properties": {
            "data": {
              "type": "array",
              "items": {
                "type": "object",
                "properties": {
                  "child": {"$ref": "#/properties/data/items"}
                }
              }
            }
          }
        }
```

Or, you can try the mcp-server-chart MCP server which triggers this:

```json
  "mcpServers": {
    "mcp-server-chart": {
      "command": "npx",
      "args": ["-y", "@antv/mcp-server-chart"]
    }
  }
```

Then, try and send any message to Gemini, this will fail due to the tool schema. 

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

See https://github.com/google-gemini/gemini-cli/issues/5030

I am not sure if we should close that issue, or leave it open after this.


packages/core/src/tools/tools.ts
```diff
@@ -228,6 +227,82 @@ export interface ToolResult {
...
+            pathRefs,
+          )
+        ) {
+          return true;
+        }
+      }
+    }
+
+    return false;
+  }
```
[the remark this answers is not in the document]
jakemac53: Gemini reviewing gemini 🤣 but hey it is valid!

packages/core/src/core/geminiChat.ts
```diff
@@ -443,6 +448,32 @@ export class GeminiChat {
...
+          }
+        }
+        if (cyclicSchemaTools.length > 0) {
+          const extraDetails =
+            `\n\nThis error was probably caused by cyclic schema references in one of the following tools, try disabling them:\n\n - ` +
+            cyclicSchemaTools.join(`\n - `) +
+            `\n`;
+          error.message += extraDetails;
+        }
+      }
```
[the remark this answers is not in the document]
jakemac53: done

packages/core/src/core/geminiChat.ts
```diff
@@ -299,6 +302,9 @@ export class GeminiChat {
 
       response = await retryWithBackoff(apiCall, {
         shouldRetry: (error: Error) => {
+          // Check for likely cyclic schema errors, don't retry those.
+          if (error.message.includes('maximum schema depth exceeded')) return false;
```
jacob314: can we make this more strict and add a test for this so if the message changes the tests fails and we know what to do? We've been bit in the past by checks for text in error messages that had false positives.

jacob314:
Hi @jakemac53, this is a great PR! The improved error handling for cyclic schema references is a fantastic UX improvement. The core logic and tests are solid.

I have a couple of suggestions to further modernize the TypeScript and enhance type safety, in line with the latest best practices.

### 1. Use `unknown` for Safer Error Handling in `shouldRetry`

The `try...catch` blocks correctly catch errors as `unknown`, which is the safest approach. We can extend this safety to the `shouldRetry` callback in `geminiChat.ts`. Instead of assuming the caught value is an `Error` instance, we can type it as `unknown` and use a type guard.

This makes the logic more robust in case a non-`Error` object (like a string or a plain object) is thrown.

**In `packages/core/src/core/geminiChat.ts`:**

```typescript
// Before
shouldRetry: (error: Error) => {
  // Check for likely cyclic schema errors, don't retry those.
  if (error.message.includes('maximum schema depth exceeded'))
    return false;
  // ...
}

// After
shouldRetry: (error: unknown) => {
  // Check for likely cyclic schema errors, don't retry those.
  if (error instanceof Error && error.message.includes('maximum schema depth exceeded')) {
    return false;
  }
  
  // Apply the `instanceof Error` check to the other conditions as well for full type safety.
  if (error instanceof Error && error.message) {
    if (error.message.includes('429')) return true;
    if (error.message.match(/5\d{2}/)) return true;
  }
  return false;
}
```

### 2. Use a More Precise Type for JSON Schemas

In `tools.ts`, the `hasCycleInSchema` function accepts `schema: object`. We can make this more descriptive and type-safe by defining a specific `JSONObject` type for JSON-like structures. This makes the function's contract clearer without changing its runtime behavior.

**In `packages/core/src/tools/tools.ts`:**

```typescript
// Suggested type definition to be added
export type JSONValue =
  | string
  | number
  | boolean
  | null
  | JSONObject
  | JSONArray;

export interface JSONObject {
  [key: string]: JSONValue;
}

export interface JSONArray extends Array<JSONValue> {}

// Updated function signature
export function hasCycleInSchema(schema: JSONObject): boolean {
  // ... implementation
}
```

What are your thoughts on these suggestions? Happy to discuss further!

jacob314 (commented):
Looks good. Also added a suggestion added with the help of gemini cli for a refactoring to clean this up slightly that Gemini CLI should make.

packages/core/src/core/geminiChat.ts
jakemac53: Are we OK with a test that hits the real Gemini API? I don't know how to do this otherwise.

jacob314: yep! 
check out this change from adh for inspiration for how we write integration tests for cases like this.
https://github.com/google-gemini/gemini-cli/pull/4890/files#diff-60f377a22934f08297f47a041704012b1dcf7ce0ca7aeed2ac63928e741212c9

jacob314 (approved):

packages/core/src/core/geminiChat.ts
jakemac53: https://github.com/google-gemini/gemini-cli/pull/5685

PR #5613, opened 2025-08-05
feat(core): Introduce `DeclarativeTool` and `ToolInvocation`.

joshualitt:
## TLDR

Introduces `DeclarativeTool` and `ToolInvocation` to streamline tool argument validation, and make tools generally more robust.

Note: this CL only migrates `read-file` to the new tool system. Existing tools are migrated to a deprecated subclass of `DeclarativeTool` to allow us to migrate each tool individually.

## Dive Deeper

The existing `Tool` class(removed in this CL), is highly error prone as it is used both for building a tool call, as well as executing a tool call. These fully orthogonal requirements result in a brittle class. Recognizing this natural division of responsibility, this CL introduces `DeclarativeTool` as a `ToolBuilder`, which is used to build up a tool call before execution or display. This `built` tool call is known internally as a `ToolInvocation` and it should be immutable.

Note: Currently, this CL just validates arguments lazily every time `DeclarativeTool.build` is called. Caching could be added in a simple follow on.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

This cl does not fix #2390, but a follow-up will.

### DIFF
--- a/packages/cli/src/acp/acpPeer.ts
+++ b/packages/cli/src/acp/acpPeer.ts
@@ -239,65 +239,62 @@ class GeminiAgent implements Agent {
       );
     }
 
-    let toolCallId;
-    const confirmationDetails = await tool.shouldConfirmExecute(
-      args,
-      abortSignal,
-    );
-    if (confirmationDetails) {
-      let content: acp.ToolCallContent | null = null;
-      if (confirmationDetails.type === 'edit') {
-        content = {
-          type: 'diff',
-          path: confirmationDetails.fileName,
-          oldText: confirmationDetails.originalContent,
-          newText: confirmationDetails.newContent,
-        };
-      }
-
-      const result = await this.client.requestToolCallConfirmation({
-        label: tool.getDescription(args),
-        icon: tool.icon,
-        content,
-        confirmation: toAcpToolCallConfirmation(confirmationDetails),
-        locations: tool.toolLocations(args),
-      });
-
-      await confirmationDetails.onConfirm(toToolCallOutcome(result.outcome));
-      switch (result.outcome) {
-        case 'reject':
-          return errorResponse(
-            new Error(`Tool "${fc.name}" not allowed to run by the user.`),
-          );
-
-        case 'cancel':
-          return errorResponse(
-            new Error(`Tool "${fc.name}" was canceled by the user.`),
-          );
-        case 'allow':
-        case 'alwaysAllow':
-        case 'alwaysAllowMcpServer':
-        case 'alwaysAllowTool':
-          break;
-        default: {
-          const resultOutcome: never = result.outcome;
-          throw new Error(`Unexpected: ${resultOutcome}`);
-        }
-      }
-
-      toolCallId = result.id;
-    } else {
-      const result = await this.client.pushToolCall({
-        icon: tool.icon,
-        label: tool.getDescription(args),
-        locations: tool.toolLocations(args),
-      });
-
-      toolCallId = result.id;
-    }
-
+    let toolCallId: number | undefined = undefined;
     try {
-      const toolResult: ToolResult = await tool.execute(args, abortSignal);
+      const invocation = tool.build(args);
+      const confirmationDetails =
+        await invocation.shouldConfirmExecute(abortSignal);
+      if (confirmationDetails) {
+        let content: acp.ToolCallContent | null = null;
+        if (confirmationDetails.type === 'edit') {
+          content = {
+            type: 'diff',
+            path: confirmationDetails.fileName,
+            oldText: confirmationDetails.originalContent,
+            newText: confirmationDetails.newContent,
+          };
+        }
+
+        const result = await this.client.requestToolCallConfirmation({
+          label: invocation.getDescription(),
+          icon: tool.icon,
+          content,
+          confirmation: toAcpToolCallConfirmation(confirmationDetails),
+          locations: invocation.toolLocations(),
+        });
+
+        await confirmationDetails.onConfirm(toToolCallOutcome(result.outcome));
+        switch (result.outcome) {
+          case 'reject':
+            return errorResponse(
+              new Error(`Tool "${fc.name}" not allowed to run by the user.`),
+            );
+
+          case 'cancel':
+            return errorResponse(
+              new Error(`Tool "${fc.name}" was canceled by the user.`),
+            );
+          case 'allow':
+          case 'alwaysAllow':
+          case 'alwaysAllowMcpServer':
+          case 'alwaysAllowTool':
+            break;
+          default: {
+            const resultOutcome: never = result.outcome;
+            throw new Error(`Unexpected: ${resultOutcome}`);
+          }
+        }
+        toolCallId = result.id;
+      } else {
+        const result = await this.client.pushToolCall({
+          icon: tool.icon,
+          label: invocation.getDescription(),
+          locations: invocation.toolLocations(),
+        });
+        toolCallId = result.id;
+      }
+
+      const toolResult: ToolResult = await invocation.execute(abortSignal);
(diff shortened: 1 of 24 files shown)


packages/cli/src/acp/acpPeer.ts
```diff
@@ -239,65 +239,62 @@ class GeminiAgent implements Agent {
...
+          default: {
+            const resultOutcome: never = result.outcome;
+            throw new Error(`Unexpected: ${resultOutcome}`);
+          }
         }
+        toolCallId = result.id;
+      } else {
+        const result = await this.client.pushToolCall({
+          icon: tool.icon,
+          label: invocation.getDescription(),
```
jacob314, line 291: like how this part turned out.

packages/cli/src/ui/hooks/atCommandProcessor.ts
```diff
@@ -459,7 +460,7 @@ export async function handleAtCommand({
     toolCallDisplay = {
       callId: `client-read-${userMessageTimestamp}`,
       name: readManyFilesTool.displayName,
-      description: readManyFilesTool.getDescription(toolArgs),
+      description: `Read files: ${pathSpecsToRead.join(', ')}`,
```
jacob314: why isn't this
invocation.getDescription()?

packages/core/src/core/client.ts
```diff
@@ -251,18 +250,15 @@ export class GeminiClient {
...
+
           // Read all files in the target directory
-          const result = await readManyFilesTool.execute(
-            {
-              paths: ['**/*'], // Read everything recursively
-              useDefaultExcludes: true, // Use default excludes
-            },
-            AbortSignal.timeout(30000),
-          );
+          const result = await invocation.execute(AbortSignal.timeout(30000));
```
jacob314, line 262: nice!

packages/core/src/core/coreToolScheduler.ts
```diff
@@ -463,14 +496,30 @@ export class CoreToolScheduler {
...
+          reqInfo.callId,
+          'error',
+          createErrorResponse(
+            reqInfo,
+            invocationOrError,
+            ToolErrorType.INVALID_TOOL_PARAMS,
+          ),
+        );
+        continue;
+      }
```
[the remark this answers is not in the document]
jacob314: this seems like a reasonable suggestion.

packages/core/src/test-utils/tools.ts
```diff
@@ -0,0 +1,57 @@
...
+    return (
+      result ?? {
+        llmContent: `Tool ${this.name} executed successfully.`,
+        returnDisplay: `Tool ${this.name} executed successfully.`,
+      }
+    );
+  }
+
+  async shouldConfirmExecute(): Promise<
+    import('../tools/tools.js').ToolCallConfirmationDetails | false
```
jacob314: avoid adding imports inline.  not sure why gemini cli sometimes does this.

packages/core/src/test-utils/tools.ts
```diff
@@ -0,0 +1,57 @@
...
+      return {
+        type: 'exec' as const,
+        title: `Confirm ${this.displayName}`,
+        command: this.name,
+        rootCommand: this.name,
+        onConfirm: async () => {},
+      };
+    }
+    return false;
+  }
```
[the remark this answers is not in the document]
jacob314: please address

jacob314 (commented):
This looks quite good! I'll review the tests in more detail once the comments are resolved.

packages/core/src/core/coreToolScheduler.ts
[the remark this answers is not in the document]
joshualitt: I'm happy to do this change, but then we must be extra careful to keep the `invocation` in sync with `tool`, which I believe is required due to the use of modifiable tools.

Some options:
1) Do as you propose, I think we currently only change tool calls in one place in `coreToolScheduler`, so we'll just regenerate the invocation then.
2) Leave things as is for now, revisit in a follow-up focused on just this issue.
3) Cache the invocation on the `DeclarativeTool` itself. That is, if the arguments haven't changed, we reuse the cached `ToolInvocation` instead of revalidating.

I think if we want to do this caching, the safest way to do it is on the `DeclarativeTool` itself, i.e. (3), vs. less centralized caching logic. However, I have a weak preference to do this in a follow-up, as this CL is quite complex as is. We can land this follow-up immediately after, and port the rest of the tools afterwards if you prefer.

Thoughts @jacob314  @gemini-code-assist?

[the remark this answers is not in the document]
joshualitt: Okay, I've updated the CL to incorporate the caching. To make this as robust and safe as possible, I've tried to centralize as much logic as possible in `coreToolScheduler.ts`. Now, the logic for building is quite simple:
1) All ToolCall types have both a `tool`(DeclarativeTool) and an `invocation`(ToolInvocation), except for `ErroredToolCall` which only optionally has a `tool`.
2) When a new tool request comes in, either it starts out life as a `ValidateToolCall` with a valid built `ToolInvocation` or we have an `ErroredToolCall`.
3) `ToolCall`s progress through their lifecycle with both valid `tool` and `invocation` properties.
4) If we have to update a `ToolCall`'s args, then we will immediately build a new `invocation`. If we cannot, i.e. because the arguments are invalid, then `setArgsInternal` will immediately transition the tool to an errored state.

I'm not sure what exactly should happen if `setArgsInternal` is called with bad arguments(I think this can only happen if the user modifies a tool call with bad args?). The existing logic will not allow these tool calls to be updated further, but we can support this if desired.

Anyway, please take a look at the code, and we can decide how to handle cases where a tool call is updated with bad arguments.

packages/cli/src/acp/acpPeer.ts
joshualitt: Agreed, thanks for pushing for it! The new changes to `coreToolScheduler` also look quite nice.

packages/cli/src/ui/hooks/atCommandProcessor.ts
joshualitt: I've updated the CL to support this. TLDR: because we may not have an invocation at this point. However, I've hoisted the argument so in the case where we do have a `ToolInvocation` we can use it, and if not we just returned a canned error description.

However, given this is an error path, I'm actually not 100% sure what the description should be in this case? Should we be saying we've read files when in fact there was some error?

packages/core/src/core/coreToolScheduler.ts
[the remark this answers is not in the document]
jacob314: Feel free to push back more against the review bot on this. its API taste is questionable and consider it only an enthusiastic intern that may have spent a lot of time thinking about your code.

jacob314 (commented):
Approved once these minor comments are addressed.

packages/cli/src/ui/hooks/atCommandProcessor.ts
```diff
@@ -466,7 +469,8 @@ export async function handleAtCommand({
     toolCallDisplay = {
       callId: `client-read-${userMessageTimestamp}`,
       name: readManyFilesTool.displayName,
-      description: readManyFilesTool.getDescription(toolArgs),
+      description:
+        invocation?.getDescription() ?? 'Error creating tool invocation.',
```
jacob314: Users don't know or care what an invocation is. Make the description be something more user facing. E.g. "Error attempting to execute tool to read files"

jacob314 (approved):
lgtm

PR #5616, opened 2025-08-05
check for the prompt capability before listing prompts from MCP servers

### LINKED ISSUE #5383: Error discovering prompts from github: MCP error -32601: prompts not supported
### What happened?

Since #4828 ([Load and use MCP server prompts as slash commands in the CLI](https://github.com/google-gemini/gemini-cli/pull/4828))  was merged,
I have observed the following error message when starting up `gemini`:

```Error discovering prompts from github: MCP error -32601: prompts not supported```

The error message is emitted in https://github.com/google-gemini/gemini-cli/blob/main/packages/core/src/tools/mcp-client.ts :

```
export async function discoverPrompts(
  mcpServerName: string,
  mcpClient: Client,
  promptRegistry: PromptRegistry,
): Promise<void> {
  try {
    const response = await mcpClient.request(
      { method: 'prompts/list', params: {} },
      ListPromptsResultSchema,
    );
    ...
  } catch (error) {
    // It's okay if this fails, not all servers will have prompts.
    // Don't log an error if the method is not found, which is a common case.
    if (
      error instanceof Error &&
      !error.message?.includes('Method not found')
    ) {
      console.error(
        `Error discovering prompts from ${mcpServerName}: ${getErrorMessage(
          error,
        )}`,
      );
    }
  }
}
```

... specifically, when I'm using, in my `~/.gemini/settings.json`:

```
  "mcpServers": {
    "github": {
      "httpUrl": "https://api.githubcopilot.com/mcp/x/repos/readonly",
      "headers": {
        "Authorization": "..."
      },
      "timeout": 5000
    }
  },
```

_Note that this setting is critical to observe the bug_: `"httpUrl": "https://api.githubcopilot.com/mcp/x/repos/readonly"` (note "**repos**")

If one instead uses `"httpUrl": "https://api.githubcopilot.com/mcp/x/issues/readonly"` (note "**issues**"), _no error will be observed_,
because the "issues/readonly" MCP server includes predefined prompts, whereas the "repos/readonly" MCP server does not.

Regarding possible solutions:
1. A simple solution would be to expand the test `!error.message?.includes('Method not found')`
to also ignore: `MCP error -32601: prompts not supported`.
2. A more general (IMO, superior) solution would be to test `if( mcpClient.getServerCapabilities()?.prompts)`
as a pre-requisite for `await mcpClient.request( { method: 'prompts/list', params: {} }, ...)`, in `discoverPrompts()`
(alternatively, one could condition calling `discoverPrompts()`, in `connectAndDiscover()`, according to this test.
I.e., one should check for the presence of `'prompts'` in the MCP server's declared `capabilities`.
Ref.: [Client.getServerCapabilities()](https://github.com/modelcontextprotocol/typescript-sdk/blob/main/src/client/index.ts#:~:text=getServerCapabilities,-%28%29%3A%20ServerCapabilities%20%7C%20undefined%20%7B)

According to [Prompts - Capabilities - Model Context Protocol](https://modelcontextprotocol.io/specification/2025-06-18/server/prompts#capabilities):

> Servers that support prompts MUST declare the prompts capability during [initialization]

..........................................................

For reference, the aforementioned Remote GitHub MCP Server URLs are listed at [Remote MCP Toolsets](https://github.com/github/github-mcp-server/blob/main/docs/remote-server.md#remote-mcp-toolsets).

For more detail regarding the difference between these two endpoints, see https://github.com/github/github-mcp-server/blob/main/pkg/github/tools.go#L66 :

```
	issues := toolsets.NewToolset("issues", "GitHub Issues related tools").
		....AddPrompts(
		toolsets.NewServerPrompt(AssignCodingAgentPrompt(t)),
		toolsets.NewServerPrompt(IssueToFixWorkflowPrompt(t)),
	)
```

No error message should be displayed.

### Client information

<details>

```
╭────────────╮
│  > /about  │
╰────────────╯

╭───────────────────────────────────────────────────────────────────────╮
│                                                                       │
│ About Gemini CLI                                                      │
│                                                                       │
│ CLI Version                         0.1.15-nightly.250801.6f7beb41    │
│ Git Commit                          6f7beb41 (local modifications)    │
│ Model                               gemini-2.5-pro                    │
│ Sandbox                             no sandbox                        │
│ OS                                  linux                             │
│ Auth Method                         OAuth                             │
│                                                                       │
╰───────────────────────────────────────────────────────────────────────╯
```

</details>

### Login information

I'm using `Login with Google`, with my Gmail account which has a Google AI Pro subscription.

_No response_

### PR DESCRIPTION
**Pull request by jakemac53**
## TLDR

MCP clients should check for server capabilities before making requests, this adds a check for the prompts capability before requesting the prompts which avoids the error for servers that don't implement prompts.

## Reviewer Test Plan

Connect to an MPC server which doesn't implement the prompts/list handler, you will no longer see an error on startup (but would previously).

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes https://github.com/google-gemini/gemini-cli/issues/5383

### DIFF
--- a/packages/core/src/tools/mcp-client.test.ts
+++ b/packages/core/src/tools/mcp-client.test.ts
@@ -22,6 +22,7 @@ import { AuthProviderType } from '../config/config.js';
 import { PromptRegistry } from '../prompts/prompt-registry.js';
 
 import { DiscoveredMCPTool } from './mcp-tool.js';
+import { ServerCapabilities } from '@modelcontextprotocol/sdk/types.js';
 
 vi.mock('@modelcontextprotocol/sdk/client/stdio.js');
 vi.mock('@modelcontextprotocol/sdk/client/index.js');
@@ -113,12 +114,17 @@ describe('mcp-client', () => {
           { name: 'prompt2' },
         ],
       });
+      const mockGetServerCapabilities = vi.fn().mockReturnValue({
+        prompts: {}
+      });
       const mockedClient = {
+        getServerCapabilities: mockGetServerCapabilities,
         request: mockRequest,
       } as unknown as ClientLib.Client;
 
       await discoverPrompts('test-server', mockedClient, mockedPromptRegistry);
 
+      expect(mockGetServerCapabilities).toHaveBeenCalledOnce();
       expect(mockRequest).toHaveBeenCalledWith(
         { method: 'prompts/list', params: {} },
         expect.anything(),
@@ -129,7 +135,12 @@ describe('mcp-client', () => {
       const mockRequest = vi.fn().mockResolvedValue({
         prompts: [],
       });
+      const mockGetServerCapabilities = vi.fn().mockReturnValue({
+        'prompts': {}
+      });
+
       const mockedClient = {
+        getServerCapabilities: mockGetServerCapabilities,
         request: mockRequest,
       } as unknown as ClientLib.Client;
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: packages/core/src/tools/mcp-client.test.ts [hunk 1]
```diff
@@ -141,17 +151,48 @@ describe('mcp-client', () => {
...
+
+      const mockedClient = {
+        getServerCapabilities: mockGetServerCapabilities,
+        request: mockRequest,
+      } as unknown as ClientLib.Client;
+
+      const consoleLogSpy = vi
+        .spyOn(console, 'debug')
+        .mockImplementation(() => {
+          // no-op
```
jacob314: nit: no need for comments that methods are no-op. bad Gemin CLI.

jacob314 (approved):
lgtm

packages/core/src/tools/mcp-client.test.ts
jakemac53: lol this was just me copying a probably Gemini CLI written test 🤣

lesteral:
@jakemac53  - thank you!

PR #5603, opened 2025-08-05
[ide-mode] Add support for in-IDE diff handling in the CLI

chrstnb:
## TLDR

CLI-side implementation of #4519

Adds support in the CLI for:
1. Opening diffs in the IDE when in IDE mode
2. Receiving notifications from the IDE about whether a diff was accepted or rejected
3. Handling CLI-based accept/reject actions, closing the diff, and applying any changes made if the diff is accepted 

NOTE: This does not include support for the edit/replace tool. That will happen in a follow-up and will require some ide-server changes.

## Dive Deeper

Gemini CLI IDE Mode Diffing (1)
Gemini CLI IDE Mode Diffing (6)
Gemini CLI IDE Mode Diffing (5)
Gemini CLI IDE Mode Diffing (4)
Gemini CLI IDE Mode Diffing (3)

## Reviewer Test Plan
https://github.com/user-attachments/assets/14a43ea2-f2fd-4e2b-a437-8adf687238f0

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| -------- | --- | --- | --- |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

https://github.com/google-gemini/gemini-cli/issues/4385

PR #5618, opened 2025-08-05
[ide-mode] Support rendering in-IDE diffs using the edit tool

chrstnb:
## TLDR

Follow-up to #5603 for the edit tool.

## Dive Deeper

Ideally we're doing replacements instead of writing the full contents, but for now this is reasonable.
## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

#4385 4385


jacob314 (approved):
lgtm

packages/core/src/tools/edit.ts
```diff
@@ -328,6 +328,12 @@ Expectation for required parameters:
       'Proposed',
       DEFAULT_DIFF_OPTIONS,
     );
+    const ideClient = this.config.getIdeClient();
+    const ideConfirmation =
+      this.config.getIdeMode() && ideClient
```
skeshive: also check feature flag & connection status here (ide client will always be set)

skeshive: (ok to do in a follow up but we should probably have some common function we use everywhere that verifies the integration is active instead of all these one off checks)

PR #5639, opened 2025-08-06
Add tests for useAtCompletion reset logic

SandyTao520:
Adds tests for the useAtCompletion hook to ensure its state is correctly reset when disabled is true.

The tests cover two specific scenarios:
 - Disabling the hook when it is in the READY state (i.e., has successfully provided suggestions).
 - Disabling the hook when it is in the ERROR state (e.g., after a file system crawl fails).

This verifies that the RESET action is dispatched correctly, clearing any existing suggestions or error states and returning the hook to its initial idle condition.


jacob314 (approved):
lgtm

Issue #3846, opened 2025-07-11
Responsive Issue,


JaiN1014:
As Gemini CLI 2.5 Pro is directly used in Terminal. But its not mandatory that everyone is using the termial at bottom side. Some uses it side by side mode, and when placing it side by side, it is not responsive.

jacob314:
Can you tell me more about the problems you are experiencing running in side by side mode? Do you want the input prompt on the right hand side of the UI or something else?

JaibhagwanJindal:
Yes, so not everyone is using the Terminal in the bottom windows or on a separate screen; they choose side-by-side mode. In side-by-side mode, its footer buttons are not properly aligned. As seen in the image. 

<img width="405" height="607" alt="ImageUploadedByJaibhagwanJindal" src="https://github.com/user-attachments/assets/03d34b27-f739-47e7-85ad-8b22bb471246" />

jacob314:
So the issue is when the terminal is narrow enough the footer buttons run into each other? It should be an easy fix to add a bit of padding and likely truncate the content when there is limited space.

PR #5651, opened 2025-08-06
feat(ui): Improve UI layout adaptation for narrow terminals

galdawave:
## TLDR

This pull request introduces responsive UI layout adaptations for the Gemini CLI, specifically to improve the user experience on narrow terminal widths (e.g., when used in a side-by-side panel). The UI now switches to a more compact, vertical layout when the terminal is narrower than 80 columns.

Key changes for reviewers to look at:
- A new `useTerminalSize` hook to get terminal dimensions.
- Updates to `Header`, `Footer`, `ContextSummaryDisplay`, and `LoadingIndicator` to support responsive layouts.
- New and updated tests that mock different terminal sizes to verify the responsive behavior.

## Dive Deeper

The CLI UI now adapts more effectively to varying terminal widths. This was achieved by introducing a `useTerminalSize` hook and a breakpoint at 80 columns. When the terminal width is below this breakpoint, the layout of several key components becomes more compact and vertically oriented.

- The `Header` component now displays different ASCII art logos (`long`, `short`, `tiny`) based on the available terminal width.
- The `Footer`, `ContextSummaryDisplay`, and `LoadingIndicator` components rearrange their content to a vertical flow on narrow screens to prevent wrapping and layout issues.
- Tests for the affected components have been added or updated to assert the correct rendering at different terminal sizes.

## Reviewer Test Plan

To validate the changes, a reviewer can:
1.  Run the Gemini CLI.
2.  Resize the terminal window to be wider and narrower than 80 columns.
3.  Observe the layout changes in the header, footer, and other UI elements. The layout should switch between the standard and compact vertical view.
4.  Run the test suite to ensure all new and existing tests pass:
    ```bash
    npm test
    ```
5.  Specifically, check the new snapshot files in `packages/cli/src/ui/__snapshots__/` to see the expected rendering for narrow and wide terminals.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |
## Linked issues / bugs

Resolves [#3846]

https://github.com/user-attachments/assets/b442bc26-77a3-467e-b2ae-901ccf0d7794


packages/cli/src/ui/components/AsciiArt.ts
```diff
@@ -25,3 +25,14 @@ export const longAsciiLogo = `
  ███░         ░░█████████  ██████████ █████     █████ █████ █████  ░░█████ █████
 ░░░            ░░░░░░░░░  ░░░░░░░░░░ ░░░░░     ░░░░░ ░░░░░ ░░░░░    ░░░░░ ░░░░░
 `;
+
+export const tinyAsciiLogo = `
```
jacob314, line 29: @miguelsolorio will defer to you for thoughts on what the tiny logo should look like.

packages/cli/src/ui/components/ContextSummaryDisplay.tsx
```diff
@@ -78,30 +82,36 @@ export const ContextSummaryDisplay: React.FC<ContextSummaryDisplayProps> = ({
...
-    } else {
-      summaryText += ' (ctrl+t to view)';
-    }
+  if (isNarrow) {
+    return (
+      <Box flexDirection="column">
+        <Text color={Colors.Gray}>Using:</Text>
+        {summaryParts.map((part, index) => (
+          <Text key={index} color={Colors.Gray}>
+            {'  '}- {part}
```
jacob314, line 105: @miguelsolorio curious about your thoughts on what style we should use to render this when in narrow mode. This looks reasonable but not sure where else in the UI we are using this style.

packages/cli/src/ui/components/InputPrompt.tsx
```diff
@@ -536,7 +536,7 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
         </Box>
       </Box>
       {completion.showSuggestions && (
-        <Box>
+        <Box width={suggestionsWidth} paddingRight={2}>
```
jacob314: why are we now passing suggestionsWidth here?

packages/cli/src/ui/components/InputPrompt.tsx
```diff
@@ -548,7 +548,7 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
         </Box>
       )}
       {reverseSearchActive && (
-        <Box>
+        <Box width={suggestionsWidth} paddingRight={2}>
```
jacob314: same question about suggestionsWidth

packages/cli/src/ui/components/LoadingIndicator.tsx
```diff
@@ -37,25 +41,46 @@ export const LoadingIndicator: React.FC<LoadingIndicatorProps> = ({
...
+        {rightContent && (
+          <Box paddingTop={isNarrow ? 1 : 0}>{rightContent}</Box>
+        )}
       </Box>
+      {isNarrow && (
+        <Box>
+          <Text color={Colors.Gray}>
+            {streamingState === StreamingState.WaitingForConfirmation
+              ? ''
+              : `(esc to cancel, ${elapsedTime < 60 ? `${elapsedTime}s` : formatDuration(elapsedTime * 1000)})`}
```
jacob314: seems like this content duplication could be reduced.

packages/cli/src/ui/constants.ts
```diff
@@ -15,3 +15,5 @@ export const UI_WIDTH =
 export const STREAM_DEBOUNCE_MS = 100;
 
 export const SHELL_COMMAND_NAME = 'Shell Command';
+
+export const NARROW_BREAKPOINT = 80;
```
jacob314: nit: name NARROW_BREAKPOINT is a bit of a surprising name. Can we create a name that is a bit more idiomatic for how it is used. E.g. something like `isNarrowWidth(width:number)` might be a bit more idiomatic to me.

packages/cli/src/ui/App.tsx
```diff
@@ -420,15 +421,16 @@ const App = ({ config, settings, startupWarnings = [], version }: AppProps) => {
...
   const { rows: terminalHeight, columns: terminalWidth } = useTerminalSize();
+  const isNarrow = terminalWidth < NARROW_BREAKPOINT;
   const { stdin, setRawMode } = useStdin();
   const isInitialMount = useRef(true);
 
   const widthFraction = 0.9;
   const inputWidth = Math.max(
     20,
-    Math.floor(terminalWidth * widthFraction) - 3,
+    Math.floor(terminalWidth * widthFraction) - 4,
```
jacob314: why did this get changed by 1?

packages/cli/src/ui/App.tsx
```diff
@@ -420,15 +421,16 @@ const App = ({ config, settings, startupWarnings = [], version }: AppProps) => {
...
   );
-  const suggestionsWidth = Math.max(60, Math.floor(terminalWidth * 0.8));
+  const suggestionsWidth = Math.max(20, Math.floor(terminalWidth * 0.9));
```
jacob314: now that I think about it, we likely never want this to exceed the terminalWidth as that won't render well. Probably really just want
`Math.min(Math.max(20, terminalWidth * 0.8), terminalWidth)` or something
@miguelsolorio might have better ideas. it may be silly that we ever made this smaller than inputWidth.

packages/cli/src/ui/hooks/useCommandCompletion.tsx
```diff
@@ -74,6 +74,7 @@ export function useCommandCompletion(
 
   const completionStart = useRef(-1);
   const completionEnd = useRef(-1);
+  const previousCommandPath = useRef<string>('');
```
jacob314: why was this added? seems unrelated.

packages/cli/src/ui/App.tsx
```diff
@@ -420,15 +421,16 @@ const App = ({ config, settings, startupWarnings = [], version }: AppProps) => {
 
   // Terminal and UI setup
   const { rows: terminalHeight, columns: terminalWidth } = useTerminalSize();
+  const isNarrow = terminalWidth < NARROW_BREAKPOINT;
```
jacob314: see this a lot of places. I would add it as a helper function as its intent is clearer than the name NARROW_BREAKPOINT.

jacob314 (commented):
Will approve once we get @miguelsolorio's thoughts on the styling. This will be really helpful for users using Gemini CLI in the sidebar in their IDE.

jacob314:
@miguelsolorio would be curious on your thoughts about densifying the footer when in column mode. The extra whitespace between the lines in the footer does look reasonable but I worry it takes up too much vertical space so would be annoying for VSCode users particularly on smaller screens. Potentially we could have it adaptively densify it based on terminal height. If your terminal is 100+ lines tall then it is probably good, if it is 40 lines tall in a small vscode sidebar then it could be frustrating.

Unrelated, here is some other behavior that I think could benefit from polish when the window is narrow.
There is a bit of wacky behavior truncating the file path as the window changes
It goes from 
image

to
image

to wrapping the lines
image

I'd suggest we try to also clean that up as part of this change likely truncating to just the directory name or
`.../directory-name`
when there isn't enough space.

Without looking I would suspect some of this is just due to accidental ink behavior where the right ink options can clean it up.

[demo of wacky file path behavior.webm](https://github.com/user-attachments/assets/c28e193d-02a1-439e-b19c-cb5e78859ce7)


Double slashes to try to truncate
image

packages/cli/src/ui/components/AsciiArt.ts
miguelsolorio: We talked about this yesterday, in our chat w/ others we discussed showing `>G` for short so all good 👍

packages/cli/src/ui/components/ContextSummaryDisplay.tsx
miguelsolorio: This is fine for now, I think we may tweak this line in the future

miguelsolorio:
> @miguelsolorio would be curious on your thoughts about densifying the footer when in column mode. The extra whitespace between the lines in the footer does look reasonable but I worry it takes up too much vertical space so would be annoying for VSCode users particularly on smaller screens. Potentially we could have it adaptively densify it based on terminal height. If your terminal is 100+ lines tall then it is probably good, if it is 40 lines tall in a small vscode sidebar then it could be frustrating.

@jacob314 I would love to further pair down the footer, I might even argue that pairing down to the current folder (and branch) would be enough. The user doesn't need to see the entire path and a lot of other terminals will add these to their app UI and can be seen as duplicative.

packages/cli/src/ui/App.tsx
jacob314: this was resolve but I'm not seeing `Math.min(Math.max(20, terminalWidth * 0.8), terminalWidth)`
Don't feel strongly but want to surface in case this was accidental

jacob314 (approved):
This makes things so much better for the common VSCode chat sidebar use case!lgtm

PR #5653, opened 2025-08-06
feat: Enable /setup-github to always run, and error appropriately

leehagoodjames:
## TLDR

Previously `/setup-github` was gated to only show when a GitHub repo was detected. Now it can always be run, and diplays and explicit error when it is run outside of a github repo.

## Dive Deeper

Below are the before/after cases for running `/setup-github` **outside** of a GitHub repository. Cases for running `/setup-github` **inside** a repository are excluded, because they don't change (they autocomplete and no error is displayed)

### Before (outside of GitHub repo)
- User types `/setup-` and `gemini-cli` does not show any slash commands or support autocomplete
  image
- User types `/setup-github`, then hits `enter`, and `gemini-cli` responds with `Unknown command: /setup-github`
  image

### After (outside of GitHub repo)
- User types `/setup-` and `gemini-cli` enables autocomplete to `/setup-github` 
  image
- User types `/setup-github`, then hits `enter`, and `gemini-cli` responds with `Unable to determine the Git root directory. /setup-github must be run from a git repository.` 
  image

### DIFF
--- a/packages/cli/src/ui/commands/setupGithubCommand.ts
+++ b/packages/cli/src/ui/commands/setupGithubCommand.ts
@@ -19,12 +19,21 @@ export const setupGithubCommand: SlashCommand = {
   description: 'Set up GitHub Actions',
   kind: CommandKind.BUILT_IN,
   action: (): SlashCommandActionReturn => {
-    const gitRootRepo = execSync('git rev-parse --show-toplevel', {
-      encoding: 'utf-8',
-    }).trim();
-
     if (!isGitHubRepository()) {
-      throw new Error('Unable to determine the Git root directory.');
+      throw new Error(
+        'Unable to determine the Git root directory. /setup-github must be run from a git repository.',
+      );
+    }
+
+    let gitRootRepo: string;
+    try {
+      gitRootRepo = execSync('git rev-parse --show-toplevel', {
+        encoding: 'utf-8',
+      }).trim();
+    } catch {
+      throw new Error(
+        'Unable to determine the Git root directory. /setup-github must be run from a git repository.',
+      );
     }
 
     const version = 'v0';
--- a/packages/cli/src/ui/commands/setupGithubCommand.ts
+++ b/packages/cli/src/ui/commands/setupGithubCommand.ts
@@ -19,12 +19,21 @@ export const setupGithubCommand: SlashCommand = {
   description: 'Set up GitHub Actions',
   kind: CommandKind.BUILT_IN,
   action: (): SlashCommandActionReturn => {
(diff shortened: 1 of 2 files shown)


jerop (approved):

packages/cli/src/ui/commands/setupGithubCommand.ts
```diff
@@ -19,14 +19,14 @@ export const setupGithubCommand: SlashCommand = {
   description: 'Set up GitHub Actions',
   kind: CommandKind.BUILT_IN,
   action: (): SlashCommandActionReturn => {
+    if (!isGitHubRepository()) {
+      throw new Error('Unable to determine the Git root directory. /setup-github must be run from a git repository.');
+    }
+
     const gitRootRepo = execSync('git rev-parse --show-toplevel', {
       encoding: 'utf-8',
     }).trim();
```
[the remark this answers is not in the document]
leehagoodjames: added

packages/cli/src/ui/commands/setupGithubCommand.ts
```diff
@@ -19,12 +19,21 @@ export const setupGithubCommand: SlashCommand = {
   description: 'Set up GitHub Actions',
   kind: CommandKind.BUILT_IN,
   action: (): SlashCommandActionReturn => {
-    const gitRootRepo = execSync('git rev-parse --show-toplevel', {
-      encoding: 'utf-8',
-    }).trim();
-
     if (!isGitHubRepository()) {
-      throw new Error('Unable to determine the Git root directory.');
+      throw new Error(
```
NTaylorMullen, line 23: Should update to be from a GitHub repository

NTaylorMullen (approved):

packages/cli/src/ui/commands/setupGithubCommand.ts
leehagoodjames: Updated

packages/cli/src/ui/commands/setupGithubCommand.ts
```diff
@@ -19,12 +19,21 @@ export const setupGithubCommand: SlashCommand = {
...
+    }
+
+    let gitRootRepo: string;
+    try {
+      gitRootRepo = execSync('git rev-parse --show-toplevel', {
+        encoding: 'utf-8',
+      }).trim();
+    } catch {
+      throw new Error(
+        'Unable to determine the Git root directory. /setup-github must be run from a git repository.',
```
jerop: make this match above

leehagoodjames: Done ✅

PR #5677, opened 2025-08-06
Fix to send user tool confirmation decision for yolo or non interacti…

### LINKED ISSUE #5676: User Tool Confirmation Decision is not logged in Yolo, non interactive mode
### What happened?

User Tool Confirmation Decision is not logged in Yolo, non interactive mode. While these are the implicit accept mode and hence considered as Accept

User Tool Confirmation Decision is not logged in Yolo, non interactive mode. While these are the implicit accept mode and hence considered as Accept

### Client information

<details>

```console
$ gemini /about
# paste output here
```

</details>

### Login information

_No response_

_No response_

agarwalravikant:
…ve mode

## TLDR

Changes has been done so that User Tool Confirmation Decision is logged in Yolo, non interactive mode as these are the implicit accept mode and hence considered as Accept

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

- Fixes #5676

### DIFF
--- a/packages/core/src/core/nonInteractiveToolExecutor.ts
+++ b/packages/core/src/core/nonInteractiveToolExecutor.ts
@@ -14,6 +14,7 @@ import {
 } from '../index.js';
 import { Config } from '../config/config.js';
 import { convertToFunctionResponse } from './coreToolScheduler.js';
+import { ToolCallDecision } from '../telemetry/types.js';
 
 /**
  * Executes a single tool call non-interactively.
@@ -87,6 +88,7 @@ export async function executeToolCall(
       error_type:
         toolResult.error === undefined ? undefined : toolResult.error.type,
       prompt_id: toolCallRequest.prompt_id,
+      decision: ToolCallDecision.ACCEPT,
     });
 
     const response = convertToFunctionResponse(
--- a/packages/core/src/core/nonInteractiveToolExecutor.ts
+++ b/packages/core/src/core/nonInteractiveToolExecutor.ts
@@ -88,7 +88,7 @@ export async function executeToolCall(
       error_type:
         toolResult.error === undefined ? undefined : toolResult.error.type,
       prompt_id: toolCallRequest.prompt_id,
-      decision: ToolCallDecision.ACCEPT,
+      decision: ToolCallDecision.AUTO_ACCEPT,
     });
 
     const response = convertToFunctionResponse(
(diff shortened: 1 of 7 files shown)


packages/core/src/core/nonInteractiveToolExecutor.ts
```diff
@@ -87,6 +88,7 @@ export async function executeToolCall(
       error_type:
         toolResult.error === undefined ? undefined : toolResult.error.type,
       prompt_id: toolCallRequest.prompt_id,
+      decision: ToolCallDecision.ACCEPT,
```
owenofbrien: Let's add a new ToolCallDecision value, AUTO_ACCEPT = 'auto_accept' and log that in the non-interactive case, as well as for the yolo and 'always allow' cases.

packages/core/src/telemetry/types.ts
```diff
@@ -27,6 +28,8 @@ export function getDecisionFromOutcome(
       return ToolCallDecision.ACCEPT;
     case ToolConfirmationOutcome.ModifyWithEditor:
       return ToolCallDecision.MODIFY;
+    case ToolConfirmationOutcome.ProceedAuto:
```
owenofbrien: Instead of adding a new ProceedAuto outcome, I suggest the following:
case ToolConfirmationOutcome.ProceedOnce:
      return ToolCallDecision.ACCEPT
    case ToolConfirmationOutcome.ProceedAlways:
    case ToolConfirmationOutcome.ProceedAlwaysServer:
    case ToolConfirmationOutcome.ProceedAlwaysTool:
      return ToolCallDecision.AUTO_ACCEPT;

I realize this would change the semantics of ToolCallDecision.Accept compared to prior logs, but the product is still young, and I don't think we have any analysis that depends on these semantics yet, so I'd rather go ahead and make the change so that going forward we can interpret ToolCallDecision.ACCEPT to mean that the user consciously chose to accept the tool call.

agarwalravikant: I have a different perspective that outcome ProceedAlways, ProceedAlwaysServer and ProceedAlwaysTool is the explicit approval which user provides. but ProceedAuto is more the autonomous implicit proceed

agarwalravikant: Updated the code as per discussion

owenofbrien (approved):

owenofbrien (approved):

allenhutchison (approved):
This looks like a great change. Thanks for fixing this issue with the telemetry. It's going to be really helpful for users running in headless mode to be able to audit what is going on.

PR #5680, opened 2025-08-06
feat(shell): Support subdirectories in all workspace directories

### LINKED ISSUE #5669: run_shell_command fails to validate valid subdirectories within workspace
### What happened?

**Description:**

  The run_shell_command tool incorrectly validates the directory parameter. It fails to recognize and
  execute commands in valid subdirectories of the registered workspace roots.

  **Steps to Reproduce:**

   1. Initialize the Gemini CLI in a project.
   2. Attempt to run a command in a valid, existing subdirectory. For example:

   1     run_shell_command(command='ls', directory='packages/core')
   3. The tool returns an error: Directory 'packages/core' is not a registered workspace directory.

  **Root Cause:**

  The validation logic in packages/core/src/tools/shell.ts is flawed. Instead of checking if the
  provided directory is a subdirectory of a workspace root, it performs an incorrect string comparison:
  path.basename(dir) === params.directory.

  This compares the base name of the workspace root (e.g., "gemini-cli") with the relative path of the
  subdirectory (e.g., "packages/core"), which will always fail.

  

The run_shell_command tool should successfully validate and execute commands in any subdirectory that
  exists within any of the registered workspace directories (including the primary project root and any
  additional directories from includeDirectories).

### Client information

* **CLI Version:** 0.1.17
* **Git Commit:** 99ba2f64 (local modifications)
* **Operating System:** darwin v24.3.0
* **Memory Usage:** 350.4 MB

### Login information

_No response_

_No response_

yashasvigirdhar:
## TLDR

This PR fixes the `run_shell_command` tool's directory validation logic to properly support subdirectories across all workspace directories. The previous implementation used flawed basename matching that prevented valid subdirectory execution. This change replaces it with comprehensive multi-workspace directory resolution that handles ambiguity and provides robust validation.

**Key changes:**
- Replace basename matching with comprehensive multi-workspace directory resolution
- Add support for running shell commands in any subdirectory of any workspace directory
- Handle ambiguity when same subdirectory exists in multiple workspace locations
- Add comprehensive test coverage for new functionality
- Update CWD resolution to use getProjectRoot() as default
- Add 6 new test cases covering edge cases and validation scenarios

## Dive Deeper

**Root Cause Analysis:**
The original validation logic used `path.basename(dir) === params.directory`, which compared the base name of workspace roots (e.g., "gemini-cli") with relative subdirectory paths (e.g., "packages/core"). This comparison always failed, preventing valid subdirectory execution.

**Solution Implementation:**
- **Multi-workspace Resolution**: Now checks if the provided directory is a subdirectory of any registered workspace root
- **Ambiguity Handling**: When the same subdirectory exists in multiple workspace locations, the tool now provides clear error messages
- **Robust Validation**: Uses proper path resolution and validation instead of simple string comparison
- **Default CWD**: Updated to use `getProjectRoot()` as the default working directory for better consistency

**Backward Compatibility:**
- All existing functionality remains unchanged
- Commands that worked before continue to work
- Only fixes the validation logic, no breaking changes

## Reviewer Test Plan

**To validate this change works:**

1. **Pull and run the code:**
   ```bash
   git checkout fix/shell-subdirectory-validation
   npm run preflight  # Verify all tests pass
   ```

2. **Test subdirectory execution:**
   ```bash
   # Test in packages/core subdirectory
   run_shell_command(command="ls", directory="packages/core")
   
   # Test in docs subdirectory  
   run_shell_command(command="ls", directory="docs")
   
   # Test in nested subdirectory
   run_shell_command(command="ls", directory="packages/core/src")
   ```

3. **Test edge cases:**
   - Try invalid directories (should fail gracefully)
   - Test with absolute paths
   - Test with relative paths from different workspace roots

4. **Verify test coverage:**
   - Run `npm test` in packages/core to see the 6 new test cases
   - Check that all validation scenarios are covered

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ✅  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

*Note: Tested on macOS (🍏). Other platforms need testing by reviewers.*

## Linked issues / bugs

**Fixes #5669**

This PR fully resolves the issue where `run_shell_command` failed to validate valid subdirectories within workspace directories. The reported bug is now fixed with comprehensive test coverage to prevent regression.

### DIFF
--- a/packages/core/src/tools/shell.ts
+++ b/packages/core/src/tools/shell.ts
@@ -126,10 +126,16 @@ class ShellToolInvocation extends BaseToolInvocation<
             return `{ ${command} }; __code=$?; pgrep -g 0 >${tempFilePath} 2>&1; exit $__code;`;
           })();
 
-      const cwd = path.resolve(
-        this.config.getTargetDir(),
-        this.params.directory || '',
-      );
+      const cwd = (() => {
+        const projectRoot = this.config.getProjectRoot();
+        const requestedDir = this.params.directory?.trim();
+        if (!requestedDir) return projectRoot;
+        const { path: resolvedPath } = resolveDirectory(
+          this.config,
+          requestedDir,
+        );
+        return resolvedPath ?? projectRoot;
+      })();
 
       let cumulativeOutput = '';
       let lastUpdateTime = Date.now();
@@ -390,17 +396,9 @@ export class ShellTool extends BaseDeclarativeTool<
       if (path.isAbsolute(params.directory)) {
         return 'Directory cannot be absolute. Please refer to workspace directories by their name.';
       }
-      const workspaceDirs = this.config.getWorkspaceContext().getDirectories();
-      const matchingDirs = workspaceDirs.filter(
-        (dir) => path.basename(dir) === params.directory,
-      );
-
-      if (matchingDirs.length === 0) {
-        return `Directory '${params.directory}' is not a registered workspace directory.`;
-      }
-
-      if (matchingDirs.length > 1) {
-        return `Directory name '${params.directory}' is ambiguous as it matches multiple workspace directories.`;
+      const { error } = resolveDirectory(this.config, params.directory);
+      if (error) {
+        return error;
       }
     }
     return null;
@@ -411,4 +409,38 @@ export class ShellTool extends BaseDeclarativeTool<
   ): ToolInvocation<ShellToolParams, ToolResult> {
     return new ShellToolInvocation(this.config, params, this.allowlist);
   }
+}
+
+function resolveDirectory(
+  config: Config,
+  relativeDir: string,
+): { path: string | null; error: string | null } {
+  const workspaceContext = config.getWorkspaceContext();
+  const workspaceDirs = workspaceContext.getDirectories();
+  const candidates: string[] = [];
+  for (const dir of workspaceDirs) {
+    const resolved = path.join(dir, relativeDir);
+    if (
+      fs.existsSync(resolved) &&
+      workspaceContext.isPathWithinWorkspace(resolved) &&
+      !candidates.includes(resolved)
+    ) {
+      candidates.push(resolved);
+    }
+  }
+  if (candidates.length === 0) {
+    return {
+      path: null,
+      error: `Directory '${relativeDir}' is not a registered workspace directory.`,
+    };
+  }
+  if (candidates.length > 1) {
+    return {
+      path: null,
+      error: `Directory '${relativeDir}' is ambiguous as it exists in multiple workspace locations: ${candidates.join(
+        ', ',
+      )}`,
+    };
+  }
+  return { path: candidates[0], error: null };
 }
--- a/packages/core/src/tools/shell.ts
+++ b/packages/core/src/tools/shell.ts
@@ -363,7 +363,7 @@ export class ShellTool extends BaseDeclarativeTool<
           directory: {
             type: 'string',
             description:
-              '(OPTIONAL) Directory to run the command in, if not the project root directory. Must be relative to the project root directory and must already exist.',
+              '(OPTIONAL) Directory to run the command in, specified as a relative path that will be resolved from any registered workspace directory. Must be a relative path and must exist. If ambiguous (exists in multiple workspaces), the command will fail - in such cases, provide a longer, more specific relative path that uniquely identifies the directory (e.g., "packages/core/src" instead of just "src").',
           },
         },
         required: ['command'],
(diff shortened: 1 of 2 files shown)


yashasvigirdhar:
Addressed the feedback by Gemini Code Assist. I am not sure it'd run again after the new commit.

yashasvigirdhar:
Waiting for a human review now.

packages/core/src/tools/shell.ts
```diff
@@ -101,6 +101,44 @@ export class ShellTool extends BaseTool<ShellToolParams, ToolResult> {
...
+    if (possiblePaths.length > 1) {
+      return {
+        paths: [],
+        error: `Directory '${directory}' is ambiguous as it exists in multiple workspace locations: ${possiblePaths.join(
+          ', ',
+        )}`,
+      };
+    }
+
+    return { paths: possiblePaths, error: null };
```
NTaylorMullen: Instead of returning an array of possible paths we should just return the final possible path because otherwise it's ambiguous OR not found 😄

packages/core/src/tools/shell.ts
```diff
@@ -101,6 +101,44 @@ export class ShellTool extends BaseTool<ShellToolParams, ToolResult> {
     return description;
   }
 
+  private _resolveDirectory(directory: string): {
```
NTaylorMullen:
```suggestion
  private resolveDirectory(directory: string): {
```

NTaylorMullen (commented):
Super appreciate the review @yashasvigirdhar! Just a few slight changes then we can get this in. Appreciate you looking out for us ❤️

packages/core/src/tools/shell.ts
yashasvigirdhar: yes makes sense! Changed now.

yashasvigirdhar:
Addressed the comments.

yashasvigirdhar:
Ping :)

jacob314:
Thanks @yashasvigirdhar for this PR. If you resolve the conflicts I'll help it get merged in today.

yashasvigirdhar:
Thanks! Took a bit of time but I've rebased it now and addressed the conflicts.

NTaylorMullen (approved):

NTaylorMullen:
@yashasvigirdhar looks like you have a few test failures

yashasvigirdhar:
The newly added tests were failing on windows due to different convention of writing path directory. Pushed a fix. Hopefully they should pass now.

yashasvigirdhar:
Hopefully the windows tests pass now. Since I can't verify them locally, its taking more iterations than I expected.

NTaylorMullen:
Hi @yashasvigirdhar looks like there are still some things to resolve!

jacob314:
@scidomino fixed some windows shell issues that may have fixed the windows tests. Rebasing to see if that is the case.

scidomino:
The tool description for the `directory` param we send Gemini is:

> (OPTIONAL) Directory to run the command in, if not the project root directory. Must be relative to the project root directory and must already exist.

At a minimum, we need to change that to reflect the new logic. 

I have a bigger concern though: even with this change, it's not possible for the Agent to request to run in certain directories if there's a name conflict. Like, if we have two project directories `/foo` and `/bar` and both of them have a subdir called `derp` then the Agent has no way of requesting to run in `/bar/derp` or `/foo/derp` since this code will have it throw an error. That seems wrong.

Personally, I think the right solution here is to switch this to require absolute paths. That would simplify this PR quite a bit and hopefully avoid the windows test issues it's running into. That said, I'm open to alternatives.

yashasvigirdhar:
@scidomino

How about we update the tool description to ask the agent to provide more specific path if there's any ambiguity? I pushed a commit with an updated description. PTAL.

Also, I think I could use a hand in fixing the windows tests if they still fail since the iteration cycle is quite slow for me as I can't run them myself.

scidomino:
>How about we update the tool description to ask the agent to provide more specific path if there's any ambiguity?

If I want to run in `/bar/derp` it does little good to tell me "why don't you run in `/bar/derp/herp/` instead?"

Please change it to use absolute paths. This will simplify the logic a lot (and maybe fix your window test) and it will better match the Edit and WriteFile tools.

If you make the change and the windows test is still broken, I'll be happy to fix it for you.

chrstnb:
Thank you for the contribution! Mind resolving conflicts?

PR #5684, opened 2025-08-06
feat: open repo secrets page in addition to README

leehagoodjames:
When running `/setup-github`, open both the README and the repo's action's secrets settings page.

### DIFF
--- a/packages/cli/src/ui/commands/setupGithubCommand.ts
+++ b/packages/cli/src/ui/commands/setupGithubCommand.ts
@@ -11,6 +11,7 @@ import {
   getGitRepoRoot,
   getLatestGitHubRelease,
   isGitHubRepository,
+  getGitHubRepoInfo,
 } from '../../utils/gitUtils.js';
 
 import {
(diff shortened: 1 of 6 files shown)


packages/cli/src/ui/commands/setupGithubCommand.ts
```diff
@@ -6,14 +6,26 @@
...
+  const readmeUrl =
+    'https://github.com/google-github-actions/run-gemini-cli/blob/v0/README.md#quick-start';
+  const repoInfo = getGitHubRepoInfo();
+  if (!repoInfo) {
+    return `open ${readmeUrl}`;
+  }
+
+  const secretsUrl = `https://github.com/${repoInfo.owner}/${repoInfo.repo}/settings/secrets/actions`;
+  return `open ${readmeUrl} ${secretsUrl}`;
+}
```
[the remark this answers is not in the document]
leehagoodjames: Done in 1287191

packages/cli/src/ui/commands/setupGithubCommand.ts
```diff
@@ -19,6 +20,36 @@ import {
...
+    case 'darwin':
+      openCmd = 'open';
+      break;
+    case 'win32':
+      openCmd = 'start';
+      break;
+    default:
+      openCmd = 'xdg-open';
+      break;
+  }
```
sethvargo: Nit - this could be it's own helper with dedicated unit tests.

sethvargo (approved):
This LGTM. My only pushback is that the person running this command might not have permissions on the repo to access the "Secrets" page, which means we'd be opening a "404" page. How do you feel about that?

leehagoodjames:
> This LGTM. My only pushback is that the person running this command might not have permissions on the repo to access the "Secrets" page, which means we'd be opening a "404" page. How do you feel about that?

Good question. Ideally we would handle this and surface an error to the user that they have insufficient permissions to complete this setup workflow (would be easier with a 403 vs 404, but if GitHub 404s we will have to make do with that). I think that will be tough with the current implementation that hands these commands to `tool_calls`. 

What do you think about having that as a `TODO`/`feature-request`, which is blocked on re-doing this workflow to more natively execute these commands in node to enable better error handling?

sethvargo:
GitHub always 404s on access denied so that they don't leak the existence of a resource to an unauthorized person.

> What do you think about having that as a TODO/feature-request, which is blocked on re-doing this workflow to more natively execute these commands in node to enable better error handling?

I'm fine with that. I might take a stab at doing that later today depending on how the morning shapes up.

packages/cli/src/ui/commands/setupGithubCommand.ts
leehagoodjames: Done in 0bf60b7

leehagoodjames:
> GitHub always 404s on access denied so that they don't leak the existence of a resource to an unauthorized person.
> 
> > What do you think about having that as a TODO/feature-request, which is blocked on re-doing this workflow to more natively execute these commands in node to enable better error handling?
> 
> I'm fine with that. I might take a stab at doing that later today depending on how the morning shapes up.

Sounds good - added #5776

jerop (approved):
as discussed in chat, i think we should a separate `/update-workflows` that just updates workflows and doesn't need any environment setup like opening secrets page

PR #5686, opened 2025-08-06
feat(chat): Add overwrite confirmation dialog to `/chat save`

flowernotfound:
## TLDR

This pull request introduces a confirmation prompt for the `/chat save <tag>` command to prevent users from accidentally overwriting an existing conversation checkpoint. If a checkpoint with the specified tag already exists, the user will now be asked to confirm the overwrite before the save operation proceeds.

https://github.com/user-attachments/assets/6bbf84fc-dd73-4568-883f-fb3e27d6986c

## Dive Deeper

This change enhances the safety of the `/chat save` command by adding an interactive confirmation step. 

1.  **`Logger` Enhancement**: A new `checkpointExists(tag: string)` method was added to `packages/core/src/core/logger.ts` to check for the existence of a checkpoint file using `fs.access`.
2.  **New Action Type**: A `confirm_overwrite` action type was introduced in `packages/cli/src/ui/commands/types.ts` to signal the need for user confirmation.
3.  **Command Logic Update**: The `save` subcommand in `packages/cli/src/ui/commands/chatCommand.ts` now uses `checkpointExists`. If the checkpoint exists and the action hasn't already been confirmed, it returns the new `confirm_overwrite` action.
4.  **UI State Management**: `packages/cli/src/ui/hooks/slashCommandProcessor.ts` was updated to handle the `confirm_overwrite` action. It uses a new state variable (`overwriteConfirmRequest`) to manage the confirmation flow, similar to how shell command confirmations are handled.
5.  **Confirmation UI**: `packages/cli/src/ui/App.tsx` now renders a `RadioButtonSelect` component when an `overwriteConfirmRequest` is active, presenting the user with "Yes" and "No" options.
6.  **Recursive Command Execution**: Based on the user's selection, `handleSlashCommand` is re-invoked with the `overwriteConfirmed` flag set to either `true` or `false`, ensuring the original command is either completed or cleanly cancelled.

## Reviewer Test Plan

To validate this change, please perform the following steps:

1.  Start the Gemini CLI.
2.  Create an initial conversation checkpoint by running:
    ```
    /chat save my-checkpoint
    ```
    Verify that it saves without a prompt.
3.  Attempt to save another checkpoint with the *same tag*:
    ```
    /chat save my-checkpoint
    ```
4.  **Verification**: Confirm that a prompt appears asking: "A checkpoint with the tag 'my-checkpoint' already exists. Do you want to overwrite it?" with "Yes" and "No" options.
5.  Select **"No"** using the arrow keys and press Enter.
6.  **Verification**: Confirm that a message "Save operation cancelled." is displayed and the original checkpoint is not modified.
7.  Run the save command again:
    ```
    /chat save my-checkpoint
    ```
8.  This time, select **"Yes"** and press Enter.
9.  **Verification**: Confirm that the message "Conversation checkpoint saved with tag: my-checkpoint." is displayed. You can further verify the checkpoint was updated by resuming it with `/chat resume my-checkpoint`.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Resolves #5569


packages/cli/src/ui/App.tsx
```diff
@@ -852,6 +854,27 @@ const App = ({ config, settings, startupWarnings = [], version }: AppProps) => {
...
+              <RadioButtonSelect
+                items={[
+                  { label: 'Yes', value: true },
+                  { label: 'No', value: false },
+                ]}
+                onSelect={(value: boolean) => {
+                  overwriteConfirmRequest.onConfirm(value);
+                }}
+              />
+              <Box height={1} />
```
jacob314: nit: use padding rather than height for padding.

packages/cli/src/ui/App.tsx
```diff
@@ -852,6 +854,27 @@ const App = ({ config, settings, startupWarnings = [], version }: AppProps) => {
 
           {shellConfirmationRequest ? (
             <ShellConfirmationDialog request={shellConfirmationRequest} />
+          ) : overwriteConfirmRequest ? (
+            <Box flexDirection="column">
+              <Text>
+                A checkpoint with the tag{' '}
```
jacob314: this needs to be refactored as this is currently hard coding that the confirmation is about a checkpoint.  I would suggest allowing the tools to specify the entire confirmation message including styling. Then the only thing required in App.tsx is the Radio button portion.

packages/cli/src/ui/commands/types.ts
```diff
@@ -135,14 +137,25 @@ export interface ConfirmShellCommandsActionReturn {
   };
 }
 
+export interface ConfirmOverwriteActionReturn {
+  type: 'confirm_overwrite';
+  /** The tag of the checkpoint to be overwritten. */
```
jacob314: Overwrite can be more generic. I'd suggest making this just
ConfirmActionReturn and return the arbitrary React component required to render the confirmation message... or a Text component if you want to restrict it to getting text back.

jacob314 (commented):
Thanks for this feature request! This is nice polish to have to avoid frustration if users accidentally override.

packages/cli/src/ui/App.tsx
flowernotfound: Thank you for the suggestion. I’ve applied the change accordingly.

jacob314 (approved):
lgtm

packages/cli/src/ui/App.tsx
flowernotfound: I’ve refactored the code to separate the logic for generating the message.

packages/cli/src/ui/commands/types.ts
flowernotfound: I’ve refactored it to accept an arbitrary React component.

flowernotfound:
@jacob314 
Thank you for the code review. I’ve applied the suggested improvements :)
Please let me know if there are any other areas that need to be addressed.

PR #5692, opened 2025-08-06
fix(oauth): Add Accept header to token requests

eeekcct:
## TLDR

This PR adds the `Accept: application/json` header to OAuth token requests. This ensures the CLI can correctly parse responses from authorization servers that default to a non-JSON content type.
## Dive Deeper

Some authorization servers respond with a content type like `application/x-www-form-urlencoded` if the Accept header is not specified. When this happens, the CLI, which expects a JSON response, fails to parse it, and the authentication flow breaks.

This fix explicitly adds the `Accept: 'application/json'` header to the fetch requests in both the `exchangeCodeForToken` and `refreshAccessToken` functions within `packages/core/src/mcp/oauth-provider.ts`. This guarantees that a JSON-formatted response is always received, making the authentication flow more robust.
## Reviewer Test Plan

To validate this change, a reviewer should test the OAuth authentication flow. The fix is best verified by testing against an OAuth server that is known to return a non-JSON response when the Accept header is missing.

1. Check out this branch.
2. Run npm install.
3. Run npm run preflight and confirm all tests pass.
4. (If possible) Run the /mcp auth command against an OAuth server that returns a non-JSON response without the Accept header and confirm that authentication now succeeds.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ✅   | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs
Fixes #5687

### DIFF
--- a/packages/core/src/mcp/oauth-provider.ts
+++ b/packages/core/src/mcp/oauth-provider.ts
@@ -370,6 +370,7 @@ export class MCPOAuthProvider {
       method: 'POST',
       headers: {
         'Content-Type': 'application/x-www-form-urlencoded',
+        Accept: 'application/json',
       },
       body: params.toString(),
     });
@@ -432,6 +433,7 @@ export class MCPOAuthProvider {
       method: 'POST',
       headers: {
         'Content-Type': 'application/x-www-form-urlencoded',
+        Accept: 'application/json',
       },
       body: params.toString(),
     });
--- a/packages/core/src/mcp/oauth-provider.ts
+++ b/packages/core/src/mcp/oauth-provider.ts
@@ -382,6 +382,13 @@ export class MCPOAuthProvider {
       );
     }
 
+    const contentType = response.headers?.get('content-type');
+    if (!contentType || !contentType.includes('application/json')) {
+      throw new Error(
+        `Token exchange failed: Expected JSON response but received Content-Type: ${contentType || 'undefined'}`,
+      );
+    }
+
     return (await response.json()) as OAuthTokenResponse;
   }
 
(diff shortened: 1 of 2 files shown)


packages/core/src/mcp/oauth-provider.ts
```diff
@@ -432,6 +433,7 @@ export class MCPOAuthProvider {
       method: 'POST',
       headers: {
         'Content-Type': 'application/x-www-form-urlencoded',
+        Accept: 'application/json',
       },
```
DinoChiesa, line 444: This new header tells the remote endpoint what this caller would like to receive. There's no guarantee that the remote endpoint will actually provide JSON.  In the large majority of cases it will, but in some cases it may still not provide JSON.  In which case, upon receipt of the response, we should check the content-type header, and try to de-serialize as JSON if and only if that content-type header indicates JSON.  Otherwise, fail with an appropriate error message , sth like "the content type of the token exchange response was not acceptable" or "the content type of the token exchange response was not JSON". 

Right now the code assumes JSON, and it failed with a somewhat cryptic

>  Failed to authenticate with MCP server 'github-remote': Unexpected token 'a', "access_tok"... is not valid JSON

It would be nice to avoid those kinds of cases by putting in just a few more checks.

DinoChiesa (changes requested):
consider introducing a few more checks.

packages/core/src/mcp/oauth-provider.ts
eeekcct: I've added the changes to check the `Content-Type` header to ensure the response is JSON before parsing!

mattKorwel:
Hey there @eeekcct thanks for taking the time to add this, and appologies it sat around for a bit. It seems this was added in parallel in https://github.com/google-gemini/gemini-cli/pull/6281 so for now I'm going to close this out If there is anything else I've missed that 6281 doesn't cover please feel free to reopen and let me know.

PR #5694, opened 2025-08-06
Add detection of tools with bad schemas and automatically omit them with a warning

jakemac53:
## TLDR

Checks MCP tool parameter schemas to ensure all properties have a specified type.

If they don't then we skip adding the tool, and emit a warning asking the user to file an issue on the MCP Server implementation.

## Dive Deeper

Gemini function calling requires types on all schema properties, and if this is not the case then all requests will error out making the entire CLI unusable. It is best to pre-emptively skip these tools in this case since we know they will fail.

It is not actually clear to me if MCP itself allows un-typed parameters from the spec. But either way, Gemini does not.

## Reviewer Test Plan

You can connect to a known MCP server that omits types, and should see a message on startup about the skipped tool. I just created my own, but https://github.com/awslabs/mcp/issues/661 has a real one you could try installing and connecting.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes https://github.com/google-gemini/gemini-cli/issues/4594

### DIFF
--- a/packages/core/src/tools/mcp-client.ts
+++ b/packages/core/src/tools/mcp-client.ts
@@ -491,8 +491,8 @@ export async function discoverTools(
         if (!hasValidTypes(funcDecl.parametersJsonSchema)) {
           console.warn(
             `Skipping tool '${funcDecl.name}' from MCP server '${mcpServerName}' ` +
-            `because it has missing types in its parameter schema. Please file an ` +
-            `issue with the owner of the MCP server.`,
+              `because it has missing types in its parameter schema. Please file an ` +
+              `issue with the owner of the MCP server.`,
           );
           continue;
         }
@@ -510,7 +510,8 @@ export async function discoverTools(
         );
       } catch (error) {
         console.error(
-          `Error discovering tool: '${funcDecl.name
+          `Error discovering tool: '${
+            funcDecl.name
           }' from MCP server '${mcpServerName}': ${(error as Error).message}`,
         );
       }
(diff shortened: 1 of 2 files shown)


jakemac53:
Added support for subschemas 👍

packages/core/src/tools/mcp-client.ts
```diff
@@ -448,6 +507,15 @@ export async function discoverTools(
           continue;
         }
 
+        if (!hasValidTypes(funcDecl.parametersJsonSchema)) {
+          console.warn(
```
jacob314, line 511: users won't see console.warn messages unless they open the console. If we make this 
console.error then they will see it but may get too concerned.

jacob314 (approved):
lgtm

packages/core/src/tools/mcp-client.ts
jakemac53: Do we have any more general framework for surfacing errors/warnings to users? I couldn't really find one but I did see other usages of console.warn. The UX is not ideal though.

jakemac53: (fwiw, this message does appear without any special flags, at least in my VsCode terminal)

jakemac53: I am going to go ahead and land this as is but I think there should be a separate issue for some sort of general framework for surfacing warnings/errors to the user as full on messages in the chat (but not the gemini history).

PR #5699, opened 2025-08-06
Add support for VSCode-like editors

chrstnb:
## TLDR

Where possible, detect when we're in VSCode-based IDEs with reliable environment variables. This will update the display name.

Also introduce an OpenVSIX installer, which I propose we use for all non VSCode (supported) editors.

## Dive Deeper

Longer-term, we may want to add ways for editors to configure their display names and installation types.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


packages/core/src/ide/detect-ide.ts
```diff
@@ -21,6 +24,9 @@ export function getIdeDisplayName(ide: DetectedIde): string {
 }
 
 export function detectIde(): DetectedIde | undefined {
+  if (process.env.EDITOR_IN_CLOUD_SHELL === 'true') {
```
skeshive: so I was looking at other VS Code forks as well (like cursor) and noticed they all set the TERM_PROGRAM = "vscode".

instead of adding each env var for all the forks, can we figure out a more generic way to support all VS Code forks? 

either we treat vs code & all forks as the same or we treat vs code vs. all forks as two separate categories but IDT we should have to allowlist each individual fork, wdyt?

jacob314: +1

jacob314 (commented):
Code looks good. Can you add tests for these changes?

packages/core/src/ide/detect-ide.ts
chrstnb: I think it's beneficial to support individual forks in this very small way (getting the name) if we reliably can. IMO the extension of this is to add a flag that editors can set to configure this themselves, but I'm not seeing the downsides of this approach for now

skeshive: adding the name for some forks is fine with me but is there a way we can detect that the current env is a fork (or is not a fork)? if so, we should default to editor instead of vs code so other non-explicitly supported forks get a generic name

packages/core/src/ide/detect-ide.ts
```diff
@@ -21,6 +33,18 @@ export function getIdeDisplayName(ide: DetectedIde): string {
 }
 
 export function detectIde(): DetectedIde | undefined {
+  if (process.env.CURSOR_TRACE_ID) {
```
skeshive, line 49: all these should be nested under the if term_prog == vscode check (incase these env vars are also set in non-vscode forks)

packages/core/src/ide/detect-ide.ts
chrstnb: Interesting-- wdyt of this updated approach? I don't think we can detect if something is a fork of VSCode vs just VSCode, so my proposal would be to just call it VSCode if we can't find another environment variable that matches. 

I think as a follow-up we could introduce an environment variable editors can set for e.g.,:
1. Display name
2. Installation type (e.g., OpenVSIX)

jacob314 (approved):
lgtm

PR #5481, opened 2025-08-03
feat(mcp): add `gemini mcp` commands for `add`, `remove` and `list`

jackwotherspoon:
## TLDR

This PR enables full control over managing MCP servers from the command line, eliminating the need for knowing about `settings.json`.

It does so by adding support for the following commands:
- `gemini mcp add <name> <commandOrUrl> [options]`
- `gemini mcp remove <name>`
- `gemini mcp list`

https://github.com/user-attachments/assets/8b7b34b8-e813-4e3c-83d9-e999d4bb8586

## Dive Deeper

The `mcpServers` field within `settings.json` is confusing and can be annoying to have to go update a config file whenever I want to update, add, or remove an MCP server.

This PR enables full control over managing MCP servers from using `gemini mcp <command>` to update the appropriate `mcpServers` field within `settings.json` on behalf of the user.

## Managing MCP Servers with `gemini mcp`

While you can always configure MCP servers by manually editing your `settings.json` file, the Gemini CLI provides a convenient set of commands to manage your server configurations programmatically. These commands streamline the process of adding, listing, and removing MCP servers without needing to directly edit JSON files.

### Adding a Server (`gemini mcp add`)

The `add` command configures a new MCP server in your `settings.json`. Based on the scope (`-s, --scope`), it will be added to either the user config `~/.gemini/settings.json` or the project config `.gemini/settings.json` file.

**Command:**

```bash
gemini mcp add [options] <name> <commandOrUrl> [args...]
```

- `<name>`: A unique name for the server.
- `<commandOrUrl>`: The command to execute (for `stdio`) or the URL (for `http`/`sse`).
- `[args...]`: Optional arguments for a `stdio` command.

**Options (Flags):**

- `-s, --scope`: Configuration scope (user or project). [default: "project"]
- `-t, --transport`: Transport type (stdio, sse, http). [default: "stdio"]
- `-e, --env`: Set environment variables (e.g. -e KEY=value).
- `-H, --header`: Set HTTP headers for SSE and HTTP transports (e.g. -H "X-Api-Key: abc123" -H "Authorization: Bearer abc123").
- `--timeout`: Set connection timeout in milliseconds.
- `--trust`: Trust the server (bypass all tool call confirmation prompts).

#### Adding an stdio server

This is the default transport for running local servers.

```bash
# Basic syntax
gemini mcp add <name> <command> [args...]

# Example: Adding a local server
gemini mcp add my-stdio-server -e API_KEY=123 /path/to/server arg1 arg2 arg3

[...]

PR #5706, opened 2025-08-06
tests: fix e2e tests

jackwotherspoon:
Previous PR #5481 updated the base `gemini` command structure which made `.strict()` kick in and block unknown arguments.

This exposed a flaw in an integration test. The nested quotes break the shell parsing, and parts of the prompt get interpreted as separate command arguments, which yargs rejects due to `.strict()`

Instead of manually quoting with double quotes, use `JSON.stringify()` which properly escapes all quotes and special characters.


NTaylorMullen (approved):

jacob314 (approved):
lgtm

PR #5727, opened 2025-08-06
Few IDE integration polishes

skeshive:
## TLDR
- Shows parent directory in open files list if duplicate base file names are present
- Fixes bug where IDE integration was broken for symlinks
- If ideMode=true, logs errors in the Debug Console to make it easier to discover IDE integration is broken

## Linked issues / bugs
#4390


packages/cli/src/ui/components/__snapshots__/IDEContextDetailDisplay.test.tsx.snap
```diff
@@ -0,0 +1,24 @@
+// Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
+
+exports[`IDEContextDetailDisplay > handles duplicate basenames by showing path hints 1`] = `
+"
+╭──────────────────────────────────────────────────────────────────────────────────────────────────╮
+│ VS Code Context (ctrl+e to toggle)                                                               │
+│                                                                                                  │
+│ Open files:                                                                                      │
+│ - bar.txt (/foo) (active)                                                                        │
```
jacob314, line 9: minor:
@miguelsolorio any thoughts on simple ways to polish this UX for showing files with the same name but different directories a bit clearer?
My minor concern with the current ux is it is a bit odd to have the directory after the file name and I don't really like the multiple adjacent blocks (in parens).

jacob314 (approved):
lgtm

Issue #5203, opened 2025-07-30
BM25-based Search for Code Chunks


mrcabbage972:
### What would you like to be added?

A tool that performs BM25 based search on code chunks.

### Why is this needed?

To improve codebase navigation.

PR #5741, opened 2025-08-07
add BM25 as a search tool on chunks of text

shamso-goog:
https://github.com/google-gemini/gemini-cli/issues/5203


shamso-goog:
happy to convert this to a draft while we work out if this actually makes sense for us but i did see it on the roadmap.

PR #5780, opened 2025-08-07
Include Schema Error Handling for Vertex and Google Auth methods

adamfweidman:
## TLDR

When authenticating via Google/Code Assist or Vertex schema depth errors get surfaces as 400 errors so are not being caught and shown to the user.

## Dive Deeper

The check is moved to within turn, after initial error processing has already happened. 

## Reviewer Test Plan

configure mcpServers to include
```
    "mcp-server-chart": {
      "command": "npx",
      "args": ["-y", "@antv/mcp-server-chart"],
    }
```

Confirm that authenticating via all 3 methods returns a warning that the following tools may have caused the error ("generate_organization_chart", "generate_mind_map", "generate_treemap_chart", "generate_fishbone_diagram")

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | x  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes: #5030
Fixes: #155

### DIFF
--- a/packages/core/src/core/geminiChat.ts
+++ b/packages/core/src/core/geminiChat.ts
@@ -34,6 +34,7 @@ import {
 import { DEFAULT_GEMINI_FLASH_MODEL } from '../config/models.js';
 import { hasCycleInSchema } from '../tools/tools.js';
 import { isStructuredError } from '../utils/quotaErrorDetection.js';
+import {StructuredError} from './turn.js';
 
 /**
  * Returns true if the response is valid, false otherwise.
@@ -352,7 +353,6 @@ export class GeminiChat {
     } catch (error) {
       const durationMs = Date.now() - startTime;
       this._logApiError(durationMs, error, prompt_id);
-      await this.maybeIncludeSchemaDepthContext(error);
       this.sendPromise = Promise.resolve();
       throw error;
     }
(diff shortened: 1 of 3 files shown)


jakemac53 (approved):
LGTM, if we can add an integration test for this error that would also be good though (followup would be fine)

jacob314 (approved):

hustcc:
@jakemac53 Hello, This PR merely adds a hint to prevent errors, but does not support the `$ref` in JSON Schema, correct?

jakemac53:
> @jakemac53 Hello, This PR merely adds a hint to prevent errors, but does not support the `$ref` in JSON Schema, correct? 

It doesn't add support for anything but it does follow `$ref` links in order to detect cycles, because these are what can cause errors.

hustcc:
> It doesn't add support for anything but it does follow `$ref` links in order to detect cycles, because these are what can cause errors.

So we still need to remove the recursive structure of the JSON schema in the [mcp-server-chart](https://github.com/antvis/mcp-server-chart)?

jakemac53:
> So we still need to remove the recursive structure of the JSON schema in the [mcp-server-chart](https://github.com/antvis/mcp-server-chart)?

Well, this is basically up to you. But it is not compatible with Gemini today. I don't know about other models, or if you can reasonably alter the schemas such that they don't cause an issue.

In particular there is something about arrays that have ref values that seemed to be triggering it.

hustcc:
> Well, this is basically up to you. But it is not compatible with Gemini today. I don't know about other models, or if you can reasonably alter the schemas such that they don't cause an issue.
> 
> In particular there is something about arrays that have ref values that seemed to be triggering it.

I'll try it. `@antv/mcp-server-chart@0.8.2` released, https://github.com/antvis/mcp-server-chart/pull/170 remove recursive JSON schema.

Issue #5082, opened 2025-07-29
Bug: /stats under-reports Model API Calls


abhipatel12:
### Problem
There is a discrepancy between the number of model API calls being made by the CLI and the metrics reported in the UI telemetry and thus the `/stats` command. The UI consistently under reports the total number of requests sent to the model.

### Root Cause Analysis

The core of the issue lies in where the telemetry logging is currently implemented. Logging is only happening at a high level of abstraction, specifically for main conversational turns managed by the `GeminiChat` class (`packages/core/src/core/geminiChat.ts`).

However, a significant number of "utility" API calls are made through a more direct, lower-level path (`GeminiClient.generateJson` in `packages/core/src/core/client.ts`). These calls are essential for internal logic but are not part of the primary user-assistant dialogue. Because they bypass the `GeminiChat` methods, they are never logged by the current telemetry system.

### Examples of Unreported API Calls

Several key features rely on these un-logged calls:

1.  **Edit Corrector (`packages/core/src/utils/editCorrector.ts`):** Before executing a file edit, the system often makes one or more API calls to correct or validate the parameters provided by the model. This ensures the edit operation doesn't fail due to minor inaccuracies, but these crucial pre-flight checks are invisible to our metrics.
2.  **Next Speaker Check (`packages/core/src/utils/nextSpeakerChecker.ts`):** To create a more fluid conversation, the CLI uses a lightweight model call to determine if the assistant has finished its thought or intends to continue. This call happens after many model responses and is never reported.

### Proposed Architectural Solution

To fix this comprehensively, we must move the telemetry logging to a lower, centralized layer.

1.  **Log at the Lowest Level:** The fundamental principle should be to log the event as close to the actual outbound network request as possible. This guarantees that no call can be missed.
2.  **Centralize the Logic:** The `ContentGenerator` interface (defined in `packages/core/src/core/contentGenerator.ts`) is the perfect architectural "choke point." Every single model request, whether for conversation or utility, must pass through an implementation of this interface. By instrumenting the `ContentGenerator` layer, we can capture all calls from a single point.
3.  **Ensure Maintainability:** This approach is robust and future-proof. New developers adding new features that require utility model calls will have their calls logged automatically, without needing to know the details of the telemetry system.

### Relevant Files for Investigation

*   `packages/core/src/core/geminiChat.ts`: **(Current Flawed Implementation)** Examine the `sendMessage` and `sendMessageStream` methods to see the current, incomplete telemetry logging.
[...]

PR #5781, opened 2025-08-07
fix(cli) - Move logging into CodeAssistServer

shishu314:
## TLDR

Move api request logging into CodeAssistServer since it is closest to the network call

## Dive Deeper

Splitting the changes into 3 parts:

1.  Migration of API request logging
1. Migration of API response logging
1. Migration of API error logging

Toolcalls had to be extracted into its own file due to circular dependency issues.

## Reviewer Test Plan

I asked gemini cli to `Generate a JSON file for me`, with breakpoints at `generateJson` in `/core/client.ts` and `logApiRequest` in `telemetry/loggers.ts`. The breakpoints both triggered, indicating that we are now logging API requests.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | x  | ❓  | ❓  |
| npx      | x  | ❓  | ❓  |
| Docker   | x  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | x  | -   | -   |

## Linked issues / bugs

This PR makes progress on #5082

### DIFF
--- a/packages/core/src/code_assist/codeAssist.ts
+++ b/packages/core/src/code_assist/codeAssist.ts
@@ -24,6 +24,7 @@ export async function createCodeAssistContentGenerator(
     const userData = await setupUser(authClient);
     return new CodeAssistServer(
       authClient,
+      config,
       userData.projectId,
       httpOptions,
       sessionId,
--- a/packages/core/src/code_assist/codeAssist.ts
+++ b/packages/core/src/code_assist/codeAssist.ts
@@ -21,7 +21,7 @@ export async function createCodeAssistContentGenerator(
     authType === AuthType.CLOUD_SHELL
   ) {
     const authClient = await getOauthClient(authType, config);
-    const userData = await setupUser(authClient);
+    const userData = await setupUser(authClient, config);
     return new CodeAssistServer(
       authClient,
       config,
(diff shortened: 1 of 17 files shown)


abhipatel12:
This is a great refactor! Centralizing the logging logic is a solid architectural improvement.

I have one question about the impl. I see that the new logging logic is added directly to the `CodeAssistServer`. This definitely solves the problem for users on the OAuth and Cloud Shell auth paths.

Wondering what your thoughts were on handling the API key path, which uses the `GoogleGenAI` content generator? Since that class isn't modified, it seems like the utility calls for those users would still not be logged.

One approach could be to create a `LoggingContentGenerator` decorator that wraps any `ContentGenerator` instance. In `createContentGenerator`, we could wrap both the `CodeAssistServer` and the `GoogleGenAI` instances with this decorator. This would give us complete coverage without duplicating the logging logic.

What do you think of that approach?

shishu314:
I think adding a `LoggingContentGenerator` wrapper makes a lot of sense. I missed the part with API key path so didn't think of it. I've updated my code to reflect your comments. Let me know if anything else is missing.

abhipatel12 (approved):
This looks great! Thanks for fixing this

jacob314 (approved):
lgtm

PR #5796, opened 2025-08-07
Use semantic colors in themes

miguelsolorio:
## TLDR

This allows for colors to be used semantically instead of referencing the color codes directly. This only handles the foundation, the rest of the components will need to be updated in follow up PRs.

## Dive Deeper

Introduces the following semantic tokens:

- text
  - primary
  - secondary
  - link
  - accent
- background
  - primary
  - diff
    - added
    - removed
- border
  - default
  - focused
- ui
  - comment
  - symbol
  - gradient
- status
  - error
  - success
  - warning
 
Custom themes would be referenced like:

```json
{
  "customThemes": {
    "Synthwave": {
      "name": "Synthwave",
      "text": {
        "primary": "#FDFDFD",
        "secondary": "#7A7A7A",
        "link": "#FF79C6",
        "accent": "#BD93F9"
      },
      "background": {
        "primary": "#282A36",
        "diff": {
          "added": "#50FA7B20",
          "removed": "#FF555520"
        }
      },
      "border": {
        "default": "#44475A",
        "focused": "#BD93F9"
      },
      "ui": {
        "comment": "#6272A4",
        "symbol": "#6272A4",
        "gradient": ["#FF79C6", "#BD93F9"]
      },
      "status": {
        "error": "#FF5555",
        "success": "#50FA7B",
        "warning": "#F1FA8C"
      }
    }
  },
  "theme": "Synthwave"
}
```

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


packages/cli/src/ui/themes/theme.test.ts
```diff
@@ -71,37 +52,6 @@ describe('validateCustomTheme', () => {
...
-    expect(result.error).toBe('Invalid color format for DiffAdded: invalid');
-  });
-
-  it('should return isValid: false for a theme with an invalid DiffRemoved color', () => {
-    const invalidTheme = { ...validTheme, DiffRemoved: 'invalid' };
-    const result = validateCustomTheme(invalidTheme);
-    expect(result.isValid).toBe(false);
-    expect(result.error).toBe('Invalid color format for DiffRemoved: invalid');
-  });
-
```
miguelsolorio, line 104: We now allow users to set partial custom themes without needing to add all tokens, so these tests are not needed anymore

packages/cli/src/ui/themes/theme.ts
```diff
@@ -27,9 +28,54 @@ export interface ColorsTheme {
   GradientColors?: string[];
 }
 
-export interface CustomTheme extends ColorsTheme {
+export interface CustomTheme {
   type: 'custom';
   name: string;
+
+  // New semantic structure (all optional)
```
jacob314: remove this comment.

jacob314 (commented):
Would be nice if we could keep showing errors if there are unrecognized properties in themes or if neither the current or previous name for a color token is specified. That will help users stay more of the rails if they use themes and help users understand if we make breaking theme changes in the future. 
Hopefully Gemini CLI can one shot that change if you tell it to look at the changes from your pull request and add back in code and tests that got reverted due to this.

Other than that looks great and lets land asap to avoid merge conflict hassles.

miguelsolorio:
> Would be nice if we could keep showing errors if there are unrecognized properties in themes or if neither the current or previous name for a color token is specified. That will help users stay more of the rails if they use themes and help users understand if we make breaking theme changes in the future.

I think disagree with this take, in VS Code when settings are not correct there are LSP warnings and it will ignore the incorrect ones. 

When I was editing a theme for my own personal use, it was really annoying getting an error when I only wanted to change a few colors instead of the entire theme. It also removes settings that don't exist, which i found annoying when I had a typo.

jacob314 (approved):
lgtm

PR #5808, opened 2025-08-07
[ide-mode] Hide diff options when active diff is not focused

chrstnb:
## TLDR

When the active diff is not focussed, we should not show the options in the toolbar nor be using cmd+s to accept changes.

## Reviewer Test Plan

https://github.com/user-attachments/assets/1287a87e-47f2-4304-95f9-dd85f5516f60

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


packages/vscode-ide-companion/src/diff-manager.ts
```diff
@@ -199,6 +204,18 @@ export class DiffManager {
     );
   }
 
+  private async updateDiffContext(editor: vscode.TextEditor | undefined) {
```
jacob314: nit: rename to onActiveEditorChange or updateActiveEditor
otherwise I thought for a second that this was updating that the diff was taking place in this editor which confused me before I read the body of the method.

jacob314 (approved):
Can you also follow up with a test?

packages/vscode-ide-companion/src/diff-manager.ts
```diff
@@ -58,7 +58,12 @@ export class DiffManager {
   constructor(
     private readonly log: (message: string) => void,
     private readonly diffContentProvider: DiffContentProvider,
-  ) {}
+  ) {
+    vscode.window.onDidChangeActiveTextEditor((editor) => {
```
NTaylorMullen: Should we be disposing of the subscription?

NTaylorMullen (approved):

PR #5811, opened 2025-08-07
[ide-mode] Write port to file in ide-server

chrstnb:
## TLDR

This will allow us to dynamically read the port for the IDE server.

Currently there is a race condition where, on startup, if gemini CLI loads before the port is written to, the user needs to open a new terminal. This is similarly an issue on install.

This implementation is backwards-compatible because it writes the port to both the tmpdir file and to the env variable.

The IDE client will first try to fetch + connect to a port from the file (if one exists). If that fails, it will fall back on the environment variable.

NOTE: should we want to remove the environment variable, we will need to update the sandboxing approach. I don't see any immediate need to do that, however.

The tmpdir file is tied to the vs code process ID via the name of the file. This will get deleted when the server stops running. This ensures that gemini CLI can read the associated VSCode window's process ID from the integrated terminal, separating the process for the current window from other vs code windows.

## Dive Deeper

https://github.com/user-attachments/assets/fa84fd3c-6edc-4a6d-bfc1-9b7c74a1d1b2

## Reviewer Test Plan

Run the extension locally, then run gemini CLI in the extension host. Confirm it's able to connect. 
## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


packages/vscode-ide-companion/src/ide-server.ts
```diff
@@ -12,12 +12,15 @@ import { StreamableHTTPServerTransport } from '@modelcontextprotocol/sdk/server/
...
 import { type Server as HTTPServer } from 'node:http';
+import * as path from 'node:path';
+import * as fs from 'node:fs/promises';
 import { z } from 'zod';
 import { DiffManager } from './diff-manager.js';
 import { OpenFilesManager } from './open-files-manager.js';
 
 const MCP_SESSION_ID_HEADER = 'mcp-session-id';
 const IDE_SERVER_PORT_ENV_VAR = 'GEMINI_CLI_IDE_SERVER_PORT';
+const PORT_FILE = '.gemini/ide-server-port';
```
skeshive: we should write these to the user's temp directory instead

packages/vscode-ide-companion/src/ide-server.ts
```diff
@@ -197,6 +200,15 @@ export class IDEServer {
           port.toString(),
         );
         this.log(`IDE server listening on port ${port}`);
+        const workspaceFolder = vscode.workspace.workspaceFolders?.[0];
```
skeshive: we should share common logic for resolving the workspace path- https://github.com/google-gemini/gemini-cli/blob/main/packages/vscode-ide-companion/src/extension.ts#L21

packages/vscode-ide-companion/src/ide-server.ts
```diff
@@ -197,6 +200,15 @@ export class IDEServer {
...
+        if (workspaceFolder) {
+          const portFile = path.join(workspaceFolder.uri.fsPath, PORT_FILE);
```
skeshive: the current pid of the vs code instance also needs to be included (the same workspace folder could be opened in multiple windows)

packages/vscode-ide-companion/src/ide-server.ts
```diff
@@ -197,6 +200,15 @@ export class IDEServer {
...
+          fs.mkdir(path.dirname(portFile), { recursive: true })
```
skeshive: nit: should we use 1 top-level directory + files inside instead of creating new directories per port?

packages/vscode-ide-companion/src/ide-server.ts
```diff
@@ -219,6 +231,15 @@ export class IDEServer {
     if (this.context) {
       this.context.environmentVariableCollection.clear();
     }
+    const workspaceFolder = vscode.workspace.workspaceFolders?.[0];
+    if (workspaceFolder) {
+      const portFile = path.join(workspaceFolder.uri.fsPath, PORT_FILE);
+      try {
+        await fs.unlink(portFile);
```
skeshive: does this actually delete the file? IIRC, deletes weren't working when the extension closes

skeshive:
also can we update the notification we show at startup to just say the extension installed successfully: https://github.com/google-gemini/gemini-cli/blob/main/packages/vscode-ide-companion/src/extension.ts#L88C65-L90C8

packages/vscode-ide-companion/src/ide-server.ts
chrstnb: yes it does!

skeshive:
> also can we update the notification we show at startup to just say the extension installed successfully: https://github.com/google-gemini/gemini-cli/blob/main/packages/vscode-ide-companion/src/extension.ts#L88C65-L90C8

actually nvm, we should remove the notification when we make the corresponding CLI changes to read the file.

packages/vscode-ide-companion/src/ide-server.ts
```diff
@@ -46,11 +49,16 @@ export class IDEServer {
...
   private log: (message: string) => void;
+  private portFile: string;
   diffManager: DiffManager;
 
   constructor(log: (message: string) => void, diffManager: DiffManager) {
     this.log = log;
     this.diffManager = diffManager;
+    this.portFile = path.join(
+      os.tmpdir(),
+      `gemini-ide-server-${process.ppid}.port`,
```
skeshive: lets name this gemini-cli-ide-server-{ppid}.json? not sure how various OS's would handle the unkown extension

packages/vscode-ide-companion/src/ide-server.ts
```diff
@@ -197,6 +205,10 @@ export class IDEServer {
           port.toString(),
         );
         this.log(`IDE server listening on port ${port}`);
+        fs.writeFile(this.portFile, port.toString()).catch((err) => {
```
skeshive: lets write a json file so we have more flexibility to extend

packages/core/src/ide/process-utils.ts
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * @returns A promise that resolves to the numeric PID of the top-level process.
+ * @throws Will throw an error if the underlying shell commands fail unexpectedly.
+ */
+export async function getIdeProcessId(): Promise<number> {
+  const platform = os.platform();
+  let currentPid = process.pid;
+
+  // Indefinitely loop upwards through the process tree
+  while (true) {
```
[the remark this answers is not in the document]
skeshive: this is a good suggestion

skeshive (approved):

jacob314 (commented):
I'll approve once this code has tests. Likely an integration test is the right way to test this.

integration-tests/ide-client.test.ts
```diff
@@ -0,0 +1,97 @@
...
+import * as fs from 'node:fs';
+import * as os from 'node:os';
+import * as path from 'node:path';
+import * as net from 'node:net';
+import { IdeClient } from '../packages/core/src/ide/ide-client.js';
+import { vi } from 'vitest';
+import { getIdeProcessId } from '../packages/core/src/ide/process-utils.js';
+
+vi.mock('../packages/core/src/ide/process-utils', () => ({
+  getIdeProcessId: vi.fn(),
```
jacob314: this doesn't test process-utils as it is mocking it out. Can you make the integration test actually test that logic as it is the most complex logic added. That might require having this integration test intentionally start up a few processes for process-utils to traverse.

jacob314 (approved):
lgtm

PR #5812, opened 2025-08-07
Normalize using a helper function

SandyTao520:
## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

### DIFF
--- a/packages/cli/src/ui/hooks/atCommandProcessor.test.ts
+++ b/packages/cli/src/ui/hooks/atCommandProcessor.test.ts
@@ -856,7 +856,7 @@ describe('handleAtCommand', () => {
         path.join(testRootDir, 'spaced file.txt'),
         fileContent,
       );
-      const query = `Check @${escapePath(filePath)}, it has spaces.`;
+      const query = `Check @${filePath}, it has spaces.`;
 
       const result = await handleAtCommand({
         query,
@@ -1030,7 +1030,7 @@ describe('handleAtCommand', () => {
         path.join(testRootDir, 'file$with&special#chars.txt'),
         fileContent,
       );
-      const query = `Check @${escapePath(filePath)} for content.`;
+      const query = `Check @${filePath} for content.`;
 
       const result = await handleAtCommand({
         query,
--- a/packages/cli/src/ui/hooks/atCommandProcessor.test.ts
+++ b/packages/cli/src/ui/hooks/atCommandProcessor.test.ts
@@ -10,6 +10,8 @@ import {
   Config,
   FileDiscoveryService,
   GlobTool,
+  normalizePath,
+  escapePath,
   ReadManyFilesTool,
   ToolRegistry,
 } from '@google/gemini-cli-core';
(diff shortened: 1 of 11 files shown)


packages/core/src/tools/read-many-files.ts
```diff
@@ -437,9 +434,10 @@ Use this tool when the user's query implies needing the content of several files
     const fileProcessingPromises = sortedFiles.map(
       async (filePath): Promise<FileProcessingResult> => {
         try {
-          const relativePathForDisplay = path
-            .relative(this.config.getTargetDir(), filePath)
-            .replace(/\\/g, '/');
+          const relativePathForDisplay = path.relative(
+            this.config.getTargetDir(),
+            filePath,
+          );
```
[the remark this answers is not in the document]
jacob314: is there a reason we shouldn't be normalizing paths here? Naively would seem reasonable to call normalizePath even if it is just defensive coding.

packages/cli/src/ui/hooks/atCommandProcessor.test.ts
```diff
@@ -31,7 +33,7 @@ describe('handleAtCommand', () => {
   async function createTestFile(fullPath: string, fileContents: string) {
     await fsPromises.mkdir(path.dirname(fullPath), { recursive: true });
     await fsPromises.writeFile(fullPath, fileContents);
-    return path.resolve(testRootDir, fullPath);
+    return escapePath(normalizePath(path.resolve(testRootDir, fullPath)));
```
jacob314, line 36: why does createTestFile return an escaped path? seems like that could hide some bugs.

packages/core/src/utils/paths.test.ts
```diff
@@ -212,3 +212,146 @@ describe('unescapePath', () => {
...
+    ).toBe('path\\\\\\\\folder\\ with\\ spaces\\\\file\\(test\\).txt');
+    expect(normalizePath('C:\\\\Program\\ Files\\\\App\\[config\\].txt')).toBe(
+      'C:\\\\Program\\ Files\\\\App\\[config\\].txt',
+    );
+  });
+
+  it('should handle edge cases', () => {
+    expect(normalizePath('')).toBe('');
+    expect(normalizePath('\\')).toBe('/');
+    expect(normalizePath('\\\\')).toBe('\\\\');
```
jacob314, line 310: normally would expect 4 escaped slashes would turn into two after unescaping

packages/core/src/utils/paths.ts
```diff
@@ -154,6 +154,38 @@ export function escapePath(filePath: string): string {
   return result;
 }
 
+/**
+ * Normalizes Windows-style paths by converting backslashes to forward slashes.
+ * Preserves backslashes that are used to escape shell special characters.
+ * @param filePath The path to normalize.
+ * @returns The normalized path with forward slashes.
+ */
+export function normalizePath(filePath: string): string {
```
jacob314, line 163: could be worth reviewing if this is right. would recommend deep research to figure out what the SOTA is for this.

PR #5658, opened 2025-08-06
feat(model): Validate Model Selection and Improve Error Handling

yuyu1025:
This PR resolves #5586 by implementing a robust model validation and selection process.

### Key Changes:
- **Model Configuration Hierarchy:** Corrected the model loading logic to respect the documented hierarchy: `--model` flag > `settings.json` > `$GEMINI_MODEL` > default.
- **Startup Validation:** Added a new service that validates the chosen model against the list of available models from the Google AI API right after authentication.
- **User-Friendly Logging:** The CLI now provides clear, informative logs at startup, indicating which model is being used and why (e.g., if it fell back to a default). This eliminates confusion from the previous behavior where an invalid model would only fail upon the first prompt.
- **Improved Error Handling:** Users will now be immediately aware of invalid model configurations instead of receiving a cryptic 404 API error.

### Testing:
- All existing tests have been updated to pass with the new changes.
- The `preflight` check completes successfully.

PR #5814, opened 2025-08-07
feat(cli): validate model names with precedence and concise startup logs

### LINKED ISSUE #5586: Model Selection and Error Handling
### What happened?

# Model Selection and Error Handling

**Summary**

The current `gemini-cli` application does not properly validate model names provided through various configuration methods. When an invalid model name is used, the CLI loads with an incorrect display and only fails with a `404 Not Found` API error upon prompt submission. This behavior can be confusing for users.

This request proposes an enhancement to the model selection process to include validation, a clear configuration hierarchy, and informative logging. This will provide a more intuitive and resilient user experience, similar to the improvements made in addressing bug report #5527.

---

### Current Behavior

When an invalid model name is provided via the `--model` flag, `.gemini/settings.json`, or the `$GEMINI_MODEL` environment variable, the CLI loads and displays the invalid name. For example:

`gemini-cli --model "invalid-model-name"`

The CLI will display something similar to:

`Using model invalid-model-name`

However, the first time a user submits a prompt, they will encounter the following error:

```json
[API Error: [{
  "error": {
    "code": 404,
    "message": "Requested entity was not found.",
    "errors": [
      {
        "message": "Requested entity was not found.",
        "domain": "global",
        "reason": "notFound"
      }
    ],
    "status": "NOT_FOUND"
  }
}
]]
```

This requires the user to diagnose why the application is failing after it appears to have loaded successfully.

### Proposed Behavior

The model selection process should be enhanced to include validation and a clear, hierarchical fallback mechanism.

1. **Configuration Hierarchy:** The application should check for a model name in the following order, with each step serving as a fallback if the previous method is not found or is invalid:

   * Command-line `--model` flag.

   * The `model` parameter in `.gemini/settings.json`.

   * The `$GEMINI_MODEL` environment variable.

   * A final default model, such as `gemini-2.5-pro`.

2. **Model Validation:** At each step in the hierarchy, the provided model name should be validated against a list of known, supported models. If the model name is invalid, the application should not load with it. Instead, it should proceed to the next fallback option in the hierarchy.

3. **Startup Logging:** To provide greater transparency, the CLI should print a startup log that clearly indicates which model is being loaded and why. This log would be very helpful for troubleshooting.

### Example Scenarios for Proposed Behavior

Here are some examples of how the new logging and fallback system would work with invalid configurations:

**Scenario 1: Command-line flag is invalid**

`gemini-cli --model "gargantuan-9.8-extreme"`

* **Log Output:**

  * `Loading --model "gargantuan-9.8-extreme" failed, not a valid model name.`

  * `Checking .gemini/settings.json for model...`

  * `Checking $GEMINI_MODEL for model...`

  * `No valid model selection found, defaulting to gemini-2.5-pro.`

  * `Using model gemini-2.5-pro`

**Scenario 2: Command-line flag is valid, but settings.json is not**

`gemini-cli --model "gemini-1.5-pro"`
*(settings.json has `"model": "myBestFriend-5.5-unhinged"`)*

* **Log Output:**

  * `Loading --model "gemini-1.5-pro" succeeded.`

  * `Using model gemini-1.5-pro`

**Scenario 3: No valid options found in any configuration method**

* **Log Output:**

  * `Loading --model flag: not specified.`

  * `Loading settings.json: "model": "myBestFriend-5.5-unhinged" failed, not a valid model name.`

  * `Loading $GEMINI_MODEL = "not-even-trying-anymore-4.4-wtf" failed, not a valid model name.`

  * `No valid model selection found, defaulting to gemini-2.5-pro.`

  * `Using model gemini-2.5-pro`

Implementing this enhancement would make the `gemini-cli` more user-friendly and robust, allowing users to quickly identify and resolve configuration issues without needing to submit a prompt and receive a cryptic API error.

The gemini-cli should validate model names before loading to ensure a valid selection has been made. If no valid model names are found in the load hierarchy, load the application default model.

### Client information

<details>

```console
About Gemini CLI

CLI Version                0.1.17
Git Commit                 99ba2f64 (local modifications)
Model                      gemini-2.5-flash-pro-pro-fish-market
Sandbox                    no sandbox
OS                         darwin
Auth Method                OAuth
GCP Project                my-crazy-project-0054243996
```

</details>

### Login information

API key

_No response_

mariorch22:
#### Problem
Invalid model names from flags, settings, or env load the UI and only fail on first prompt with a 404. This is confusing and noisy for users.

#### Solution
- Enforce a clear model selection precedence with validation and quiet, relevant startup logs.

#### Behavior
- Precedence: `--model` > `.gemini/settings.json:model` > `$GEMINI_MODEL` > default.
- Validation: candidates must be in `SUPPORTED_GEMINI_MODELS` (core). Invalid entries are rejected and we fall back.
- Logging policy:
  - Only log when a provided value is invalid (source + value) and always include final “Using model …” when a failure occurred.
  - Happy path is silent in the UI.
  - Non-interactive mode prints the same relevant lines.

Examples:
```text
# invalid in settings.json; falls back to default
Loading settings.json "myBestFriend-5.5-unhinged" failed, not a valid model name.
Using model gemini-2.5-pro
```
```text
# invalid in env; no CLI/settings provided
Loading $GEMINI_MODEL "gemini-2.5-flashsss" failed, not a valid model name.
Using model gemini-2.5-pro
```

#### Key Changes
- core: add canonical model data and type-safety
  - `packages/core/src/config/models.ts`
    - `SUPPORTED_GEMINI_MODELS` (readonly tuple)
    - `type SupportedGeminiModel = typeof SUPPORTED_GEMINI_MODELS[number]`
    - `isSupportedModel(model): model is SupportedGeminiModel`
    - `DEFAULT_GEMINI_MODEL` re-exported
  - `packages/core/index.ts` re-exports the above
- cli: selection policy + integration
  - `packages/cli/src/config/modelSelection.ts`:
    - `selectModel(argv, settings, env?, default?)` returns `{ model, logs, hadFailure }`
    - Validates candidates, logs only “failed”, appends “Using model …” once
  - `packages/cli/src/gemini.tsx`:
    - Run `selectModel` early, set `argv.model` to validated result
    - Show logs only if `hadFailure`, both interactive and non-interactive
  - `packages/cli/src/config/config.ts`:
    - Remove implicit `default: process.env.GEMINI_MODEL` from `--model` option to preserve precedence and logging clarity

#### Type-safety
- Strong narrowing via `SupportedGeminiModel` and `isSupportedModel`.
- `selectModel`’s `defaultModel` parameter typed as `SupportedGeminiModel` to avoid string widening.

#### Tests
- `packages/cli/src/config/modelSelection.test.ts`
  - Precedence
  - Validation failure on CLI/settings/env
  - Default fallback
  - Logging policy (only failures + final “Using model …”, silent happy path)
  - Tests avoid `any` and reset env per test

#### Docs
- Add note on model precedence and npm flag forwarding:
  - Use `npm start -- -m "gemini-1.5-pro"` (the `--` forwards args to the CLI).
- Briefly document the quiet logging policy and examples.

#### Backwards compatibility
- No breaking flags.
- `$GEMINI_MODEL` still respected but no longer silently injected as the `--model` default.
- Startup remains quiet on the happy path.

#### How to test locally
- Settings valid: put a supported model in `.gemini/settings.json` → no startup logs.
- CLI invalid: `npm start -- -m "invalid-model-name"` → see failure + final “Using model …”.
- Env invalid: `GEMINI_MODEL=gemini-2.5-flashsss npm start` → see failure + final “Using model …”.
- Precedence: `GEMINI_MODEL=gemini-1.5-flash npm start -- -m gemini-1.5-pro` → uses `gemini-1.5-pro`, no logs.

#### Related
- Fixes #5586
- Refs #5658

#### Checklist
- [x] Unit tests added and passing
- [x] Lint and preflight pass
- [x] Docs updated (precedence + npm flag forwarding)
- [x] No noisy logs in the happy path

### DIFF
--- a/packages/cli/src/ui/AppContainer.tsx
+++ b/packages/cli/src/ui/AppContainer.tsx
@@ -20,7 +20,7 @@ import {
   type HistoryItemWithoutId,
   AuthState,
 } from './types.js';
-import { MessageType } from './types.js';
+import { MessageType, StreamingState } from './types.js';
 import {
   type EditorType,
   type Config,
@@ -699,7 +699,20 @@ Logging in with Google... Please restart Gemini CLI to continue.
 
   const { handleInput: vimHandleInput } = useVim(buffer, handleFinalSubmit);
 
-  const isInputActive = !initError && !isProcessing;
+  /**
+   * Determines if the input prompt should be active and accept user input.
+   * Input is disabled during:
+   * - Initialization errors
+   * - Slash command processing
+   * - Tool confirmations (WaitingForConfirmation state)
+   * - Any future streaming states not explicitly allowed
+   */
+  const isInputActive =
+    !initError &&
+    !isProcessing &&
+    (streamingState === StreamingState.Idle ||
+      streamingState === StreamingState.Responding) &&
+    !isProQuotaDialogOpen;
 
   // Compute available terminal height based on controls measurement
   const availableTerminalHeight = useMemo(() => {


packages/core/src/config/models.ts
```diff
@@ -9,3 +9,21 @@ export const DEFAULT_GEMINI_FLASH_MODEL = 'gemini-2.5-flash';
...
 
 export const DEFAULT_GEMINI_EMBEDDING_MODEL = 'gemini-embedding-001';
+
+export const SUPPORTED_GEMINI_MODELS = [
+  'gemini-2.5-pro',
+  'gemini-2.5-flash',
+  'gemini-2.5-flash-lite',
+  'gemini-1.5-pro',
+  'gemini-1.5-flash',
+] as const;
```
allenhutchison: This isn't going to work for our project. Internally for testing we use model names that don't appear on this list, and that we don't want to make public.

allenhutchison (changes requested):
Thank you for the PR. I'd support better error messages or a fast fail path for this kind of change, but we can't do the approach that you have suggested. Let me know if you want to try something different or if we should close this PR.

mariorch22:
Alright, thanks for the feedback. I will try a different approach and will submit it in a few days.

mariorch22:
Small concern: I'm not fully sure countTokens is the right/lightest probe for verifying model availability. It works, but it hits the generation API and could add startup latency or be sensitive to rate limits/timeouts.
If there’s a more canonical/lighter way in this SDK to check a model name (or availability/permissions), I’m happy to switch. Feedback welcome!

mariorch22:
Quick update: I re-checked this change and found a couple of mistakes. I’ll push fixes shortly and mark this PR as Draft until CI is green.

allenhutchison:
Just taking a quick pass through this PR today. Have you considered using the models.list endpoint? https://ai.google.dev/api/models#method:-models.list

mariorch22:
> Just taking a quick pass through this PR today. Have you considered using the models.list endpoint? https://ai.google.dev/api/models#method:-models.list

Thanks for the feedback! I did think about using models.list, the blocker was that it requires an API key and doesn’t work when logged in via OAuth. My plan is: if an API key is available, fetch models dynamically, otherwise fall back to a simple regex check. I didn’t have time this week, but I’ll try to implement that today.

Do you have any other ideas for a good fallback approach?

allenhutchison:
Reading through the original issue, I wonder if there is an easier and more maintainable way to handle this. If the model doesn't exist you get a 404 right? If that's the case you could catch 404s and then create a useful message to the user. You could direct them to the appropriate API site for the model lists (ai.dev for api keys, and vertex page for vertex, etc...). That might eliminate the need to maintain a manual list of valid models which is going to be hard to keep up to date. If you go down this route then you could also remove the flag / setting you are adding for unknown models.

What do you think?

mariorch22:
I see your point, but I think the situation is a bit more nuanced.
Right now, during startup the code just constructs the GoogleGenAI client without binding a model, so no 404s or similar errors are raised at that stage. The first time you actually see a failure is when making a completion request against the Gemini API.

From my perspective there are basically three ways to improve this:
- perform a lightweight API call during startup (e.g. countTokens) to verify that the configured model is valid
- do some local validation (regex or a list of known models)
- or fetch the available model list from the API and check against that

Or potentially combine them: a cheap local check first, and then a minimal API probe (or list fetch) to confirm.

But none of them is a really good way of doing it. What do you think?

allenhutchison:
We used to have a lightweight model checker. This was used during the time that Gemini 2.5 Pro did not provide free quota, so we had to validate that you could make a request to the model before we set it up as your default. You can see the code here: https://github.com/google-gemini/gemini-cli/blob/514e883af17cbfe5f1de7c2b92c08a5f5a3a8fad/packages/core/src/core/modelCheck.ts

I removed this a while back because it had been inadvertently bypassed and wasn't running. We didn't get any complaints so it was deemed unnecessary.

From your list we know that the local validation won't work as a strategy because we have a lot of internal users who are using models that are only available inside Google. 

I think the probe is the best option, but it will slow down startup time. It's also important to realize that we actually use several models internally to Gemini CLI including Flash, Flash-lite, and the embedding model. Not all of those are controlled by the model flag and setting.

PR #5660, opened 2025-08-06
Fix for git issue 5657 to add lines of code added/removed telemetry

agarwalravikant:
## TLDR

Changes has been done to capture the telemetry of LLM and User added/removed lines during LLM proposed changes.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| -------- | --- | --- | --- |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

- Resolves #5657

PR #5823, opened 2025-08-08
Fix for git issue 5657 to add lines of code added/removed telemetry

### LINKED ISSUE #5657: Add lines of code added/removed by LLM and User

Measure how many lines of code has been added, removed by user and LLM during agent proposed changes.

### Why is this needed?

In order to measure lines of code accepted, addition of this telemetry is required.

### Additional context

_No response_

agarwalravikant:
## TLDR

Changes has been done to capture the telemetry of LLM and User added/removed lines during LLM proposed changes.

This PR is clone of https://github.com/google-gemini/gemini-cli/pull/5660 and old pr is closed due to incorrect conflict resolution.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

- Resolves #5657

### DIFF
--- a/docs/telemetry.md
+++ b/docs/telemetry.md
@@ -183,9 +183,10 @@ Logs are timestamped records of specific events. The following events are logged
     - `function_args`
     - `duration_ms`
     - `success` (boolean)
-    - `decision` (string: "accept", "reject", or "modify", if applicable)
+    - `decision` (string: "accept", "reject", "auto_accept", or "modify", if applicable)
     - `error` (if applicable)
     - `error_type` (if applicable)
+    - `metadata` (if applicable, dictionary of string -> any)
 
 - `gemini_cli.api_request`: This event occurs when making a request to Gemini API.
   - **Attributes**:
@@ -262,3 +263,7 @@ Metrics are numerical measurements of behavior over time. The following metrics 
     - `lines` (Int, if applicable): Number of lines in the file.
     - `mimetype` (string, if applicable): Mimetype of the file.
     - `extension` (string, if applicable): File extension of the file.
+    - `ai_added_lines` (Int, if applicable): Number of lines added by AI.
+    - `ai_removed_lines` (Int, if applicable): Number of lines removed by AI.
+    - `user_added_lines` (Int, if applicable): Number of lines added by user in AI proposed changes.
+    - `user_removed_lines` (Int, if applicable): Number of lines removed by user in AI proposed changes.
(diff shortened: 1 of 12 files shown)


docs/telemetry.md
```diff
@@ -262,3 +263,7 @@ Metrics are numerical measurements of behavior over time. The following metrics
     - `lines` (Int, if applicable): Number of lines in the file.
     - `mimetype` (string, if applicable): Mimetype of the file.
     - `extension` (string, if applicable): File extension of the file.
+    - `ai_added_lines` (Int, if applicable): Number of lines added by AI.
+    - `ai_removed_lines` (Int, if applicable): Number of lines removed by AI.
+    - `user_added_lines` (Int, if applicable): Number of lines added by user in AI proposed changes.
+    - `user_removed_lines` (Int, if applicable): Number of lines removed by user in AI proposed changes.
```
jerop: how is changed lines considered? is it double counted as removed an added?

jerop: nit: was trying to use CRUD in the other attributes, can we be consistent and use:
- created instead of added
- deleted instead of removed

agarwalravikant: Yes changed lines are counted as both added and removed

jerop: but then added/removed is commonly used for diffs 🤔

agarwalravikant: I believe instead of created added makes more sense, however open for suggestions

jerop: thanks for the clarification, wonder if this is worth documenting - maybe by adding `including changed lines` at the end of each description?

jerop: it's ok, we can leave it as added/removed given that it's commonly used for diffs

packages/core/src/tools/diffOptions.ts
jerop, line 1: can we add some simple tests for this logic to ensure we're getting the right stats?

packages/core/src/telemetry/metrics.ts
jerop, line 1: can we add some tests for this in https://github.com/agarwalravikant/gemini-cli/blob/529c9a86aec8d4b827fcca884ca271dcde007678/packages/core/src/telemetry/metrics.test.ts

agarwalravikant: added

jerop (approved):
it'd be cool if we added `+x -y` in `/stats` in future work

cc @abhipatel12

Issue #4392, opened 2025-07-17
Allow user to select which open IDE files get included as context in the UI


skeshive:
Add a checkbox that allows the user to include or exclude the current files as context. The current open files may not be relevant to the user's query so we should allow the user to exclude files as necessary.

chrstnb:
We could leverage the slash command flow but use '@' for referencing files, and then we could pull up the recently-opened files list to allow the user to @ include them. This is how it works in code assist now!

skeshive:
Marking this as not planned after discussing with UX team- we should run evals first to determine if quality is degraded in IDE mode and then decide how to address that (either through user selection or updating how the context gets added to the prompt)

PR #5839, opened 2025-08-08
Updates UX for context selection via IDE integration 

skeshive:
## TLDR
Instead of auto-including all open files as context, we now only include the active file by default. The user can  include other open files by navigating to the open files context menu. 

## Dive Deeper
* The active file will always be included in the context. This active file can be excluded by the user, however, if they click into that file again or select context it will be included again
* When a new file becomes active, the previous active file is not included anymore by default

https://github.com/user-attachments/assets/d76c9839-cab1-4c7b-819d-1e09ccecfb2a

## Linked issues / bugs
#4392

### DIFF
--- a/packages/cli/src/ui/App.test.tsx
+++ b/packages/cli/src/ui/App.test.tsx
@@ -530,7 +530,7 @@ describe('App UI', () => {
     );
     currentUnmount = unmount;
     await Promise.resolve();
-    expect(lastFrame()).toContain('1 open file (ctrl+e to view)');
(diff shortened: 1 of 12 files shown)


skeshive:
@jacob314 

This PR uses ctrl+e to toggle between the context view and tab to focus on the context for editinng. If the user starts typing or presses esc, the focus will move back to the input box. 

However, just realized both ctrl+e and tab are already existing shortcuts so would appreciate some suggestions.

PR #5842, opened 2025-08-08
fix(cli) - Adding logging for response and error in LoggingContentGenerator

### LINKED ISSUE #5082: Bug: /stats under-reports Model API Calls

shishu314:
## TLDR

This adds response and error logging to the `LoggingContentGenerator wrapper` class

## Dive Deeper

I had to create another generator so the consumer isn't blocked on the whole stream finishing, allowing us to go through logging logic as the user consumes the stream. 

## Reviewer Test Plan

For response testing, I added a breakpoint and console logs to make sure only the last responses are being logged.

For error testing, I wasn't sure how to test so I just threw an error in the try block of `generateContentStream`, then I added breakpoint to see that the error was being logged.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | x  | ❓  | ❓  |
| npx      | x  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Resolves #5082

### DIFF
--- a/packages/core/src/core/geminiChat.ts
+++ b/packages/core/src/core/geminiChat.ts
@@ -14,15 +14,12 @@ import {
   SendMessageParameters,
   createUserContent,
   Part,
-  GenerateContentResponseUsageMetadata,
   Tool,
 } from '@google/genai';
 import { retryWithBackoff } from '../utils/retry.js';
 import { isFunctionResponse } from '../utils/messageInspectors.js';
 import { ContentGenerator, AuthType } from './contentGenerator.js';
 import { Config } from '../config/config.js';
-import { logApiResponse, logApiError } from '../telemetry/loggers.js';
-import { ApiErrorEvent, ApiResponseEvent } from '../telemetry/types.js';
 import { DEFAULT_GEMINI_FLASH_MODEL } from '../config/models.js';
 import { hasCycleInSchema } from '../tools/tools.js';
 import { StructuredError } from './turn.js';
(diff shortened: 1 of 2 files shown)


abhipatel12 (approved):
Awesome, thanks for adding this!

jacob314 (approved):
lgtm

PR #5844, opened 2025-08-08
feat: add environment variable support for generation config

### LINKED ISSUE #5280: Add support for setting API call configurations - temperature, topK, thinking budget
We should have a way to set the thinking budget to mitigate thought loops. Additionally we need a way to experiment with different parameters as we run evals

akhil29:
## Summary
- Add support for reading generation configuration from environment variables
- Implements parsing for `GEMINI_TEMPERATURE`, `GEMINI_TOP_K`, and `GEMINI_THINKING_BUDGET`
- Integrates environment variables into the Config class and GeminiClient

## Changes
- Added `parseGenerationConfigFromEnv()` function to parse and validate environment variables
- Extended Config class to store and provide generation configuration
- Updated GeminiClient to use generation config from Config instead of hardcoded values
- Added comprehensive tests for environment variable parsing and client integration

## Test plan
- [x] Added unit tests for environment variable parsing with validation
- [x] Added tests for GeminiClient using generation config
- [x] Verified environment variables are properly passed to API calls
- [x] Tested invalid values show appropriate warnings

Fixes #5280

### DIFF
--- a/packages/cli/src/config/config.ts
+++ b/packages/cli/src/config/config.ts
@@ -308,6 +308,79 @@ export async function loadHierarchicalGeminiMemory(
   );
 }
 
+/**
+ * Parse and validate generation config from environment variables
+ */
+function parseGenerationConfigFromEnv():
+  | {
+      temperature?: number;
+      topK?: number;
+      thinking_budget?: number;
+    }
+  | undefined {
+  const config: {
+    temperature?: number;
+    topK?: number;
+    thinking_budget?: number;
+  } = {};
+
+  // Parse temperature from env var
+  if (
+    process.env.GEMINI_TEMPERATURE !== undefined &&
+    process.env.GEMINI_TEMPERATURE !== ''
+  ) {
+    const temp = parseFloat(process.env.GEMINI_TEMPERATURE);
+    if (!isNaN(temp) && temp >= 0 && temp <= 2.0) {
+      config.temperature = temp;
+    } else {
+      logger.warn(
+        `Invalid GEMINI_TEMPERATURE value: ${process.env.GEMINI_TEMPERATURE}. Must be between 0.0 and 2.0`,
+      );
+    }
+  }
+
+  // Parse topK from env var
+  if (
+    process.env.GEMINI_TOP_K !== undefined &&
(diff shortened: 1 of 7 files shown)


akhil29:
Temp: 2.0
Screenshot 2025-08-08 at 9 17 59 AM

akhil29:
Temp: default
Screenshot 2025-08-08 at 9 22 05 AM

Vennygis (approved):

Vennygis (approved):

NTaylorMullen (changes requested):
`generationConfig` is a bit too generic. for this PR would you mind changing this to `getHyperparameters`. In addition, lets have a default set of hyper parameters so instead of this being `parse...` the `get` would have its own default in mind and would never return `undefined`. This way call sites didn't have to worry about the undefinedness of the hyperparameters, they could always just respect the full value.

Finally, get buy-off from @anj-s that this env var approach is what she was looking for (I would assume it was, BUT Anjali would be the big stakeholder here)

akhil29:
Done the requested changes.

@anj-s pls review.

packages/cli/src/config/config.ts
```diff
@@ -308,6 +308,79 @@ export async function loadHierarchicalGeminiMemory(
...
+  if (
+    process.env.GEMINI_TEMPERATURE !== undefined &&
+    process.env.GEMINI_TEMPERATURE !== ''
+  ) {
+    const temp = parseFloat(process.env.GEMINI_TEMPERATURE);
+    if (!isNaN(temp) && temp >= 0 && temp <= 2.0) {
+      config.temperature = temp;
+    } else {
+      logger.warn(
+        `Invalid GEMINI_TEMPERATURE value: ${process.env.GEMINI_TEMPERATURE}. Must be between 0.0 and 2.0`,
```
anj-s, line 337: what happens if we send an invalid temperature value to the API? do we get an error?
I am wondering if we should throw an error?

akhil29: Screenshot 2025-08-17 at 10 34 10 AM
This is how the error is shown currently.

Let me know if we should throw a blocking error instead.

chrstnb:

PR #5849, opened 2025-08-08
fix: correct absolute path handling in system prompts

### LINKED ISSUE #1244: Tool almost always uses a non-absolute path first
**Describe the bug**
Pretty much every time a tool invokes something with a path, it gets an error:

```
✦ Now that I'm on the gh-pages branch, I'll list the files to find the documentation.
 ╭────────────────────────────────────╮
 │ ✔  ReadFolder .                    │
 │                                    │
 │    Error: Failed to execute tool.  │
 ╰────────────────────────────────────╯

✦ My apologies, I need to provide an absolute path. I'll list the files in the current directory using its absolute path.
 ╭───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
 │ ✔  Shell pwd (Getting the current working directory.)                                                                                 │
 │                                                                                                                                       │
 │    /home/macd/code/zooshi                                                                                                             │
 │                                                                                                                                       │
 ╰───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

Can we add some context to the affected tool(s) or in the overall SI to help avoid this extra turn?

akhil29:
## Summary
- Fixed incorrect handling of absolute paths in system prompts that was causing the model to prepend project root to user-provided absolute paths
- Updates Path Construction and Tool Usage instructions to properly distinguish between relative and absolute paths

## Problem
Issue #1244: The model was incorrectly combining the project root with user-provided absolute paths due to faulty instructions in the system prompt. This resulted in invalid paths like `/project/root/etc/hosts` when users provided `/etc/hosts`.

## Solution
Updated the system prompt instructions to:
1. Check if the user provides an absolute path (starting with `/`)
2. If absolute, use it as-is without modification
3. If relative, resolve it against the project root

## Changes
- Updated `Path Construction` instruction in `prompts.ts` to handle absolute paths correctly
- Updated `Tool Usage` section to clarify path handling
- Updated test snapshots to reflect the new instructions

## Testing
- All existing tests pass
- Type checking passes
- Linting passes

Fixes #1244

### DIFF
--- a/packages/core/src/core/__snapshots__/prompts.test.ts.snap
+++ b/packages/core/src/core/__snapshots__/prompts.test.ts.snap
@@ -13,7 +13,7 @@ exports[`Core System Prompt (prompts.ts) > should append userMemory with separat
 - **Proactiveness:** Fulfill the user's request thoroughly, including reasonable, directly implied follow-up actions.
 - **Confirm Ambiguity/Expansion:** Do not take significant actions beyond the clear scope of the request without confirming with the user. If asked *how* to do something, explain first, don't just do it.
 - **Explaining Changes:** After completing a code modification or file operation *do not* provide summaries unless asked.
-- **Path Construction:** Before using any file system tool (e.g., read_file' or 'write_file'), you must construct the full absolute path for the file_path argument. Always combine the absolute path of the project's root directory with the file's path relative to the root. For example, if the project root is /path/to/project/ and the file is foo/bar/baz.txt, the final path you must use is /path/to/project/foo/bar/baz.txt. If the user provides a relative path, you must resolve it against the root directory to create an absolute path.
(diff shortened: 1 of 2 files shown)


akhil29:
@anj-s @abhipatel12  what is needed here to unblock?

anj-s:
we need to test and evaluate this change prompt change. this needs a repro. cc @silviojr who is looking into this.

silviojr:
Hello, @akhil29 !

Thanks for your contribution! I like the suggestion to update the instructions to reenforce that the paths should be absolute. 

But there is still a chance that Gemini will not follow the instructions and use a relative path. 

I think we can make this even more robust by actually handling this error during checking and trying to fix it. For example, like in [line](https://github.com/google-gemini/gemini-cli/blob/c668699e77104e70167c949ddf769baae297162f/packages/core/src/tools/read-file.ts#L210), instead of just checking if the path is absolute, we can try to convert it to an absolute path.  

What do you think?

chrstnb:

PR #5860, opened 2025-08-08
Fix excessive console logging + remove unnecessary try catch

skeshive:
- Only log errors to console if the previous state wasn't already disconnected to prevent spamming. This also preserves the earlier more useful error messages


jacob314 (approved):
lgtm

Issue #5830, opened 2025-08-08
[nonInteractiveCli] "Error executing tool write_file: File path must be absolute" will cause process.exit(1)


binux:
### What happened?

7748e56 changed the tool error handling for nonInteractiveCli. When tools like `write_file` getting an relative file path, the error will be propagated and cause `process.exit(1)`.

### What did you expect to happen?

The tool failed with normal error with "Error executing tool write_file: File path must be absolute" message propagated to the tool and been fixed by the next turn.

### Client information

<details>

```console
$ gemini /about
│ About Gemini CLI                                                       │
│                                                                        │
│ CLI Version             0.1.17                                         │
│ Git Commit              9bfbac29                                       │
│ Model                   gemini-2.5-pro                                 │
│ Sandbox                 no sandbox                                     │
│ OS                      linux                                          │
│ Auth Method             gemini-api-key                                 │
```

</details>

### Login information

API key

### Anything else we need to know?

I can understand the idea that having tool instructions to ask Gemini to use an absolute path is necessary, and I can't find a simple prompt to reproduce it. But in my case, I have a 10+ step instruction with long context `GEMINI.md`. When I asked it to generate a `commit_message.md` or `summary.json` at the end of the run, it always failed with "Error executing tool write_file: File path must be absolute". Here is an example:

# Turn 10 (2025-08-08T04:57:15.448Z)
> Thought: **Revising Memory Handling**
> I've been working on refining the memory handling in the code. I've switched from using the potentially unsafe `memcpy` to the more modern `std::ranges::copy` for moving data. This
 involved adding the `<algorithm>` header for the necessary functionality. My goal is to improve safety and potentially performance.
> Thought: **Analyzing Implementation Details**
> My focus has shifted to analyzing the specific changes made to resolve the buffer overflow. The code modifications, involving `<algorithm>` inclusion and `std::ranges::copy`, are s
traightforward. I'm now certain this approach aligns well with modern C++ standards, ensuring safety and efficiency. It's a clean and effective solution. The commit message and summa
ry JSON are ready for final review and integration.
> Thought: **Completing Implementation Review**
> I'm now fully confident in the solution implemented. The minimal diff includes the necessary `<algorithm>` header and replaces the unsafe `memcpy` with a safe `std::ranges::copy`.
This aligns perfectly with modern C++ standards, ensuring both safety and efficiency. I've prepared the commit message and summary JSON; they're ready for final review and integratio
n. The solution is concise and idiomatic.

[...]

Issue #5825, opened 2025-08-08
Gemini CLI fails to complete file edits, stops unexpectedly, and falls back to plain text output


mohamedyaakoubi:
### What happened?

While using Gemini CLI to edit a file, the tool repeatedly fails to complete the edit operation despite the context window being far from full (around 70%). The process stops abruptly without any cancellation from my side.

When I prompt it to continue, Gemini responds without using the editing tools, instead outputting the updated code as plain text in the CLI, as if it’s pretending to update the codebase. This behavior results in extremely slow workflows, sometimes taking an excessive amount of time just to make a single edit.

### What did you expect to happen?

Gemini should complete the edit in a single operation when context limits are not exceeded.

If interrupted, it should resume using the editing tools and actually apply changes to the codebase, not just print code to the terminal.

Edit operations should be completed in a reasonable time frame.

### Client information

<details>

CLI Version:   0.1.18
Git Commit:    cd7e60e0
Model:         gemini-2.5-pro
Sandbox:       no sandbox
OS:            Windows (win32)
Auth Method:   gemini-api-key

</details>

### Anything else we need to know?

**Logs / Session Stats:**
Agent powering down. Goodbye!

Interaction Summary:
Tool Calls:                 27 ( ✔ 22 ✖ 5 )
Success Rate:               81.5%
User Agreement:             66.7% (3 reviewed)

Performance:
Wall Time:                  52m 13s
Agent Active:               24m 40s
  » API Time:               21m 59s (89.1%)
  » Tool Time:              2m 41s (10.9%)

Model Usage:
gemini-2.5-pro               Reqs: 33
                             Input Tokens: 8,888,100
                             Output Tokens: 83,290

Savings Highlight: 8,206,176 (92.3%) of input tokens served from cache.

bdmorgan:
Quick analysis:

1. Session management stops processing at turn limits (packages/core/src/core/client.ts:1030)
2. Edit correction has complex retry logic that can fail (packages/core/src/utils/editCorrector.ts)
3. Context compression triggers automatically and may disrupt ongoing operations
4. No explicit continuation mechanism for resumed edit operations

Recommendations:

1. Check **maxSessionTurns** setting - may be inadvertently limiting session length
2. Monitor context compression events - may be interrupting edit sequences
3. Investigate edit tool string matching - 18.5% failure rate suggests matching issues
4. Review API timeouts and retry logic - 89% API time indicates potential throttling
[...]

Issue #5758, opened 2025-08-07
The model doesn't realise WriteFile tool can create directories if needed.


yashasvigirdhar:
### What happened?

~I think there's a bug in write file tool where its displaying wrong error message. The problem was that the directory didn't exist but it displays that the file path should be absolute. It was able to use relative path later on after creating the directory.~

We are running mkdir to create a directory for writing to a new file while the writefile tool is capable of creating the directory.

<img width="1362" height="520" alt="Image" src="https://github.com/user-attachments/assets/dda858e3-9bdf-46d3-a138-5964c0cdab16" />

### What did you expect to happen?

Correct error message

### Client information

* **CLI Version:** 0.1.18
* **Git Commit:** cd7e60e0
* **Operating System:** darwin v24.3.0
* **Sandbox Environment:** no sandbox
* **Model Version:** gemini-2.5-pro
* **Memory Usage:** 295.1 MB

yashasvigirdhar:
Looking at the code, it looks like `write_file` tool is capable of creating directories in its `execute` function. But the model called it with a relative path so it failed in the `validateToolParams` method itself. The model needs to understand that write_file tool can create directories if needed.

scidomino:
I've confirmed this is true. Should be a relatively simple matter to fix the description. However, we need to performance test such changes so it might be a bit of work.

```
 The write_file tool is designed to write content to files, but it can't
  create directories. If I try to write to a path where the directory doesn't
  exist, the operation will fail. The correct way to do this is to first
  create the directory with mkdir and then write the file.

  Would you like me to proceed with that?

╭─────────────────────────────────────────────────────────────────────────╮
│  > I think write_file can create directories. Let's test my theory by   │
│    having you attempt it                                                │
╰─────────────────────────────────────────────────────────────────────────╯

✦ Okay, I will attempt to write the file directly. Here is the tool call:
 ╭───────────────────────────────────────────────────────────────────────────╮
 │ ✔  WriteFile Writing to foo/bar.txt                                       │
 │                                                                           │
 │    1 hello world                                                          │
 ╰───────────────────────────────────────────────────────────────────────────╯
✦ You were right! It looks like write_file can create parent directories. The
  file has been created at /Users/sciortino/src/gemini-cli/foo/bar.txt.

  What's next?
``

PR #5862, opened 2025-08-09
feat: add "surface" to all logs

leehagoodjames:
## TLDR

This adds `surface` (ex: `GitHub`) to all logs, to support surfacing things like tool_call failures that are disproportionately impacting certain surfaces. 

## Dive Deeper

This assumes there is not an intentional decision to normalize logging by session id, therefor allows for `surface` to be reported to all logs in a denormalized fashion - which is much easier for downstream querying. 

## Reviewer Test Plan

Related to: #5830, #5825, #5758

### DIFF
--- a/packages/core/src/telemetry/clearcut-logger/clearcut-logger.ts
+++ b/packages/core/src/telemetry/clearcut-logger/clearcut-logger.ts
@@ -150,13 +150,27 @@ export class ClearcutLogger {
     }
   }
 
+  addDefaultFields(data: EventValue[]): void {
+    const totalAccounts = getLifetimeGoogleAccounts();
+    const surface = determineSurface();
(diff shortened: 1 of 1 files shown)


leehagoodjames:
> ## Code Review
> This pull request adds `surface` logging to all telemetry events. While this is a valuable addition, the implementation introduces significant code duplication by adding the same logic to each `log...Event` method. My main feedback is to refactor this by centralizing the `surface` logging logic into the `createLogEvent` method. This will improve maintainability and reduce redundancy. I've added a specific comment with a suggestion on how to approach this refactoring.

resolved

galdawave (approved):

NTaylorMullen (approved):

PR #5863, opened 2025-08-09
feat(/setup-github): Use node to download the files

sethvargo:
This enhances the curl-based approach to use `fetch` to download the files directly and in parallel.

## TLDR

The existing solution shells out to curl. That is sad and requires `curl` on the system. 

## Dive Deeper

This new approach uses node's native fetching. This also fixes the global proxy configuration setting.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

N/A

### DIFF
--- a/packages/cli/src/ui/commands/setupGithubCommand.ts
+++ b/packages/cli/src/ui/commands/setupGithubCommand.ts
@@ -7,7 +7,7 @@
 import path from 'node:path';
 import * as fs from 'node:fs';
 import { Writable } from 'node:stream';
-import { ProxyAgent, setGlobalDispatcher } from 'undici';
+import { ProxyAgent } from 'undici';
 
 import { CommandContext } from '../../ui/commands/types.js';
 import {
--- a/packages/cli/src/ui/commands/setupGithubCommand.ts
+++ b/packages/cli/src/ui/commands/setupGithubCommand.ts
@@ -4,7 +4,10 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import path from 'path';
+import path from 'node:path';
+import * as fs from 'node:fs';
+import { Writable } from 'node:stream';
+import { ProxyAgent, setGlobalDispatcher } from 'undici';
 
 import { CommandContext } from '../../ui/commands/types.js';
 import {
@@ -48,6 +51,8 @@ export const setupGithubCommand: SlashCommand = {
   action: async (
     context: CommandContext,
   ): Promise<SlashCommandActionReturn> => {
+    const abortController = new AbortController();
+
     if (!isGitHubRepository()) {
       throw new Error(
         'Unable to determine the GitHub repository. /setup-github must be run from a git repository.',
@@ -68,7 +73,24 @@ export const setupGithubCommand: SlashCommand = {
     // Get the latest release tag from GitHub
     const proxy = context?.services?.config?.getProxy();
     const releaseTag = await getLatestGitHubRelease(proxy);
+    const readmeUrl = `https://github.com/google-github-actions/run-gemini-cli/blob/${releaseTag}/README.md#quick-start`;
 
+    // Create the .github/workflows directory to download the files into
+    const githubWorkflowsDir = path.join(gitRepoRoot, '.github', 'workflows');
+    try {
+      await fs.promises.mkdir(githubWorkflowsDir, { recursive: true });
+    } catch (_error) {
+      console.debug(
+        `Failed to create ${githubWorkflowsDir} directory:`,
+        _error,
+      );
+      throw new Error(
+        `Unable to create ${githubWorkflowsDir} directory. Do you have file permissions in the current directory?`,
+      );
+    }
+
+    // Download each workflow in parallel - there aren't enough files to warrant
+    // a full workerpool model here.
     const workflows = [
       'gemini-cli/gemini-cli.yml',
       'issue-triage/gemini-issue-automated-triage.yml',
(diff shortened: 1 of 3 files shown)


packages/cli/src/ui/commands/setupGithubCommand.ts
```diff
@@ -68,37 +73,87 @@ export const setupGithubCommand: SlashCommand = {
...
-      const fileName = path.basename(workflow);
-      const curlCommand = buildCurlCommand(
-        `https://raw.githubusercontent.com/google-github-actions/run-gemini-cli/refs/tags/${releaseTag}/examples/workflows/${workflow}`,
-        [`--output "${gitRepoRoot}/.github/workflows/${fileName}"`],
+      downloads.push(
+        (async () => {
+          const endpoint = `https://raw.githubusercontent.com/google-github-actions/run-gemini-cli/refs/tags/${releaseTag}/examples/workflows/${workflow}`;
+          const response = await fetch(endpoint, {
+            method: 'GET',
+            dispatcher: proxy ? new ProxyAgent(proxy) : undefined,
```
NTaylorMullen, line 108: Woohoo appreciate you picking up on the proxy support ❤️

packages/cli/src/ui/commands/setupGithubCommand.ts
```diff
@@ -68,37 +73,87 @@ export const setupGithubCommand: SlashCommand = {
...
+            signal: AbortSignal.any([
+              AbortSignal.timeout(30_000),
```
NTaylorMullen, line 110: @abhipatel12 did we never add abort signals to command context?

packages/cli/src/ui/commands/setupGithubCommand.ts
```diff
@@ -68,37 +73,87 @@ export const setupGithubCommand: SlashCommand = {
...
+          const body = response.body;
+          if (!body) {
+            throw new Error(
+              `Empty body while downloading ${endpoint}: ${response.status} - ${response.statusText}`,
+            );
+          }
+
+          const destination = path.resolve(
+            githubWorkflowsDir,
+            path.join(path.basename(workflow)),
```
NTaylorMullen: why path.join?

packages/cli/src/ui/commands/setupGithubCommand.ts
```diff
@@ -68,37 +73,87 @@ export const setupGithubCommand: SlashCommand = {
...
+          );
+
+          const fileStream = fs.createWriteStream(destination, {
+            mode: 0o640,
```
NTaylorMullen: Can you leave a comment on what this represents

packages/cli/src/ui/commands/setupGithubCommand.ts
```diff
@@ -68,37 +73,87 @@ export const setupGithubCommand: SlashCommand = {
...
+            flags: 'w',
+            flush: true,
+          });
+
+          await body.pipeTo(Writable.toWeb(fileStream));
```
NTaylorMullen, line 138: I'm surprised fetch didn't have something out of the box for all of this

packages/cli/src/ui/commands/setupGithubCommand.ts
```diff
@@ -68,37 +73,87 @@ export const setupGithubCommand: SlashCommand = {
...
-      commands.push(curlCommand);
     }
 
-    const readmeUrl = `https://github.com/google-github-actions/run-gemini-cli/blob/${releaseTag}/README.md#quick-start`;
+    // Wait for all downloads to complete
+    try {
+      await Promise.all(downloads);
+    } catch (_error) {
+      abortController.abort();
+      throw _error;
```
NTaylorMullen: instead of re-throwing an error turn the catch block into a finally block so the error will still throw but the controller will abort too

packages/cli/src/utils/gitUtils.ts
```diff
@@ -70,8 +67,9 @@ export const getLatestGitHubRelease = async (
         'Content-Type': 'application/json',
         'X-GitHub-Api-Version': '2022-11-28',
       },
-      signal: controller.signal,
-    });
+      dispatcher: proxy ? new ProxyAgent(proxy) : undefined,
+      signal: AbortSignal.any([AbortSignal.timeout(30_000), controller.signal]),
```
NTaylorMullen, line 71: ❤️

NTaylorMullen (approved):

packages/cli/src/ui/commands/setupGithubCommand.ts
sethvargo: I don't see anything threaded through.

packages/cli/src/ui/commands/setupGithubCommand.ts
sethvargo: Fixed

packages/cli/src/ui/commands/setupGithubCommand.ts
sethvargo: Done

packages/cli/src/ui/commands/setupGithubCommand.ts
sethvargo: I agree, but both Gemini and Google Search confirmed it. You can invert the problem (pipe the readable into the stream), but it's the same result.

packages/cli/src/ui/commands/setupGithubCommand.ts
sethvargo: Done

NTaylorMullen (approved):

jerop (approved):

PR #5867, opened 2025-08-09
fix: use correct inferred OAuth metadata urls

DinoChiesa:
## TLDR

Use the correct URLs for metadata for the protected resource, and for the authorization server.  This allows the OAuth flow for compliant MCP Servers to complete successfully.

## Dive Deeper

[IETF RFC 8414 Section 3](https://datatracker.ietf.org/doc/html/rfc8414#section-3) describes how to produce the URLs to find metadata:  
> available at a path formed by inserting a well-known URI string into the authorization server's issuer identifier between the host component and the path component, if any.  By default, the well-known URI string used is "/.well-known/oauth-authorization-server". 

But the implementation in Gemini CLI without this change uses a more simplistic approach, replacing the path with `/.well-known/WHATEVER` rather than injecting `/.well-known/WHATEVER` in the left-most position. 

For the GitHub MCP server, the result was: 

| URL   | value used | correct value |
| --- | --- | --- |
| protected metadata | https://api.githubcopilot.com/.well-known/oauth-protected-resource | https://api.githubcopilot.com/.well-known/oauth-protected-resource/mcp/ |
| authorization server | https://github.com/.well-known/oauth-authorization-server (but it never gets here)| https://github.com/.well-known/oauth-authorization-server/login/oauth |
| authorization server fallback | https://api.githubcopilot.com/.well-known/oauth-authorization-server | https://api.githubcopilot.com/.well-known/oauth-authorization-server/mcp/ | 

This change corrects the handling to use URL surgery that complies with RFC8414 directives. 

(To enable a hands-on  test, this change ALSO includes the change from #5692 that requests a JSON response when redeeming a code for a token.)

## Reviewer Test Plan

Test this with the GitHub remote server. 
GitHub does not support Dynamic Client Registration (DCR), which means you must provision a client ID and Secret for the MCP Client to use with the MCP Server.

To test: 

1. Create an Oauth client ID / Secret pair for Github
   a. Visit Github
   b. Right click on your photo/avatar in the upper right corner
   c. click Settings
   d. on the far left, scroll waaay down and click Developer Settings
   e. on the left hand side, OAuth Apps
   f. Create a new OAuth app
   g. name it whatever you like
   h. provide a callback URL of http://localhost:7777/oauth/callback (hard coded by Gemini CLI)
   I. create a secret.  Copy it. 
2. Provision your settings.json file with the correct setup for the Github "remote" MCP server, including the new client ID and Secret you just provisioned.
   ```json
   {
     ...
     "mcpServers": {
       "github-remote": {
         "httpUrl": "https://api.githubcopilot.com/mcp/",
         "oauth": {
           "enabled": true,
           "clientId": "sample_client_id_y1vKquixJM",
           "clientSecret": "sample_c5de306629eba4"
         }
       }
     },
     ...
   }
   ```
3. start gemini-cli
4. `/mcp auth github-remote`
5. The browser window will open.  Authenticate and approve the app. 
6. Control returns to gemini-cli; it works. 
7. ctrl-t to see the MCP tools
8. prompt: "Show me the 3 most recent gists I've created" - it works.
9. Examine the mcp-oauth-tokens.json file and observe the newly acquired credentials for GitHub.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

- Progress on: #5397   (needs test update)
- related to:  #5692


packages/core/src/mcp/oauth-provider.ts
```diff
@@ -145,8 +145,7 @@ export class MCPOAuthProvider {
   private static async discoverOAuthFromMCPServer(
     mcpServerUrl: string,
   ): Promise<MCPOAuthConfig | null> {
-    const baseUrl = OAuthUtils.extractBaseUrl(mcpServerUrl);
```
cornmander, line 147: I think this was the only use of extractBaseUrl. can you delete that function?

cornmander:
Please fix linting errors. LGTM but please get a review from @emeryray2002

packages/core/src/mcp/oauth-provider.ts
```diff
@@ -432,6 +432,7 @@ export class MCPOAuthProvider {
       method: 'POST',
       headers: {
         'Content-Type': 'application/x-www-form-urlencoded',
+        Accept: 'application/json',
```
cornmander, line 434: FWIW it seems like this shouldn't be necessary. The response must be application/json:

https://datatracker.ietf.org/doc/html/rfc6749#section-5.1

DinoChiesa: You are right - "Shouldn't be" necessary, according to the spec. For the github OAuth server, it **is** necessary. But this is covered in #5692 . I introduced this change here only to allow the test to succeed.

packages/core/src/mcp/oauth-utils.ts
```diff
@@ -41,22 +41,32 @@ export interface OAuthProtectedResourceMetadata {
...
+   *
+   * @param originalUrl The original URL
+   * @param wellKnownSegment The well-known segment to insert
+   * @returns The constructed URL
+   */
+  static constructRfc841WellKnownUrl(originalUrl: string, wellKnownSegment: string): string {
+    const url = new URL(originalUrl);
+    const base = `${url.protocol}//${url.host}`;
+    return new URL( wellKnownSegment + url.pathname, base).toString();
+  }
```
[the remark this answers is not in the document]
DinoChiesa: I think this more robust approach may be unnecessary. 

- The URL class, when applied to http URLs, always returns a single leading slash in the pathname property.  I presume the validation of the httpUrl (eg that it uses https? as the scheme) is done elsewhere. If the wellKnownSegment is constructed correctly, which it will be - it is always provided by the caller - the existing implementation will never result in double slashes.  

- There is no userinfo expected in the httpUrl. 

But I'm happy to use the implementation suggested here if that is preferred.  @emeryray2002

packages/core/src/mcp/oauth-provider.ts
DinoChiesa: Yes

cornmander:
Seems like you have a bad merge?

DinoChiesa:
I had rebased main... and there's a commit there lacking a CLA. 
\*sigh\*

cornmander (approved):
Rebase on latest head and you can

emeryray2002:
tests in oauth-utils.test.ts are referencing OAuthUtils.extractBaseUrl, but this method no longer exists in oauth-utils.ts.
 Please ensure your project builds, and passes the linter and formater before submitting the PR.

packages/core/src/mcp/oauth-utils.ts
```diff
@@ -251,17 +274,6 @@ export class OAuthUtils {
     return null;
   }
 
-  /**
-   * Extract base URL from an MCP server URL.
-   *
-   * @param mcpServerUrl The MCP server URL
-   * @returns The base URL
-   */
-  static extractBaseUrl(mcpServerUrl: string): string {
```
emeryray2002, line 260: This is going to break tests. Please update tests as needed

emeryray2002 (dismissed):

packages/core/src/mcp/oauth-utils.ts
DinoChiesa: yep, done.

emeryray2002:
Implemented in https://github.com/google-gemini/gemini-cli/pull/6281

PR #5880, opened 2025-08-09
Modify content generated describing the ide context to only include deltas after the initial update

### LINKED ISSUE #5879: IDE integration emits the full list of active ide files with each message rather than the delta
### What happened?

Expect the IDE integration would only add the active ide files when the user selection in the ide had changed from the previous item in the conversation. This wastes some tokens and can confuse the model as it thinks the user is repeatedly referencing the IDE state causing the model to often reference the IDE state even for prompts that do not involve it.

For example in this case when I asked what 3+3 is the model made the following distracting comment:

Image

I was also generally reliably able to trigger this sort of question on the very first prompt when asking what `1+1` was.

Expect that IDE integration would only be additive rather than degrading the model performance for simple cases like asking to add numbers.

### Client information

<details>

```console
$ gemini /about
# paste output here
```

</details>

### Login information

_No response_

_No response_

jacob314:
deltas after the initial update.
Also switch to a json structure, switch to using a 1 based column instead of a 0 based character as that is more standard for ides. This mitigates known cases where the ide context was messing up the conversation.

This change also seeks to make the IDE state descriptions more clearly context rather than part of the prompt by switching them to separate user messages and indicating in the data that it is not always relevant to the task.

All of the prompting in this PR could be refined. Feel free to propose suggestions.
I went with JSON because it is generally terse and makes it more clear this is data not instructions from the user.

## Reviewer Test Plan

Run with the IDE extension and verify that your expected use cases asking the model to consider active files are handled correctly and that the model does not get confused and start referring to the context of what files are open even when files are not open. `1+1?` turns out to be an effective query to trigger challenges like this particularly asking the same query multiple times.
Make sure the model correctly handles when the open file changes  or the selection changes after the initial prompt as this triggers the delta calculation.
You can run with 
`DEBUG=1`
to see the data sent to the model for the hidden messages describing the IDE context in the debug console.

## Linked issues / bugs

Resolves: #5879

### DIFF
--- a/packages/core/src/core/client.ts
+++ b/packages/core/src/core/client.ts
@@ -50,6 +50,7 @@ import {
   NextSpeakerCheckEvent,
 } from '../telemetry/types.js';
 import { ClearcutLogger } from '../telemetry/clearcut-logger/clearcut-logger.js';
+import { IdeContext, File } from '../ide/ideContext.js';
 
 function isThinkingSupported(model: string) {
   if (model.startsWith('gemini-2.5')) return true;
@@ -112,6 +113,8 @@ export class GeminiClient {
 
   private readonly loopDetector: LoopDetectionService;
   private lastPromptId: string;
+  private lastSentIdeContext: IdeContext | undefined;
+  private forceFullIdeContext = true;
 
   constructor(private config: Config) {
     if (config.getProxy()) {
@@ -189,6 +192,7 @@ export class GeminiClient {
   }
 
   async startChat(extraHistory?: Content[]): Promise<GeminiChat> {
+    this.forceFullIdeContext = true;
     const envParts = await getEnvironmentContext(this.config);
     const toolRegistry = await this.config.getToolRegistry();
     const toolDeclarations = toolRegistry.getFunctionDeclarations();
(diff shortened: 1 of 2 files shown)


packages/core/src/core/client.ts
```diff
@@ -238,6 +242,174 @@ export class GeminiClient {
...
+
+      if (this.config.getDebugMode()) {
+        console.log(contextParts.join('\n'));
+      }
+      return {
+        contextParts,
+        newIdeContext: currentIdeContext,
+      };
+    }
+  }
```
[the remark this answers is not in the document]
jacob314: probably better to keep them together as they are tightly coupled.

packages/core/src/core/client.ts
```diff
@@ -238,6 +242,174 @@ export class GeminiClient {
...
+          previousPath: lastActiveFile.path,
+        };
+      }
+
+      if (Object.keys(changes).length === 0) {
+        return { contextParts: [], newIdeContext: currentIdeContext };
+      }
+
+      delta.changes = changes;
+      const jsonString = JSON.stringify(delta, null, 2);
```
jacob314, line 396: this is pretty silly. I am pretty printing the json. open to removing that to save a couple tokens for the model.

Also open to using Markdown or a different format than json. Generally I think the JSON format is clearer and easier to keep separate from context that was generated by users but open to other opinions. it is easy enough to switch the format if this isn't idea. Also open to making this fully look like a tool request or switch to markdown whichever people prefer. please try this out. in practice it seems robust.

packages/core/src/core/client.ts
```diff
@@ -238,6 +242,174 @@ export class GeminiClient {
...
+            currentCursor &&
+            (!lastCursor ||
+              lastCursor.line !== currentCursor.line ||
+              lastCursor.character !== currentCursor.character)
+          ) {
+            changes.cursorMoved = {
+              path: currentActiveFile.path,
+              cursor: {
+                line: currentCursor.line,
+                column: currentCursor.character + 1,
```
jacob314: it seemed a bit odd we were using a zero based character rather than the more standard one based column so I tweaked that but open to switching back if I'm not aware of why character is preferred.

packages/core/src/core/client.ts
NTaylorMullen: json is ok. It's risen in training model importance for tools 😄

packages/core/src/core/client.ts
```diff
@@ -273,49 +445,17 @@ export class GeminiClient {
...
-            { text: contextParts.join('\n') },
-            ...(Array.isArray(request) ? request : [request]),
-          ];
-        }
+      const { contextParts, newIdeContext } = this.getIdeContextParts(
+        this.forceFullIdeContext || this.getHistory().length === 0,
+      );
+      if (contextParts.length > 0) {
+        this.getChat().addHistory({
+          role: 'user',
```
NTaylorMullen, line 454: Note: In the IDE context objects they say `in the user's IDE...`. We may want to change that verbiage given the user role is meant to be "from the user". 

At some point we'll probably want to update the system prompt to indicate that there are some "synthetic" messages in chat history and have a common format for those (cc @anj-s)

NTaylorMullen (approved):

packages/core/src/core/client.ts
skeshive: this choice was arbitrary so we can update it but mind making the change in the extension instead so its consistent? https://github.com/google-gemini/gemini-cli/blob/main/packages/vscode-ide-companion/src/open-files-manager.ts#L149

packages/core/src/core/client.ts
skeshive: @NTaylorMullen @anj-s I remember in the original PR we wanted to avoid adding the IDE context to history- https://github.com/google-gemini/gemini-cli/pull/4312#issuecomment-3081022461. Does the same concern apply here?

packages/core/src/core/client.ts
```diff
@@ -238,6 +242,174 @@ export class GeminiClient {
...
+        contextData.otherOpenFiles = otherOpenFiles;
+      }
+
+      if (Object.keys(contextData).length === 0) {
+        return { contextParts: [], newIdeContext: currentIdeContext };
+      }
+
+      const jsonString = JSON.stringify(contextData, null, 2);
+      const contextParts = [
+        "Here is the user's IDE context as a JSON object. This is for your information only.",
```
skeshive: @anj-s advised against using the term IDE- https://github.com/google-gemini/gemini-cli/pull/4312#discussion_r2211830666

skeshive (approved):

packages/core/src/core/client.ts
jacob314: I've double checked and calling addHistory like this isn't causing the messages to be visible in the chat displayed in Gemini CLI or causing us to show these synthetic messages when you press the up arrow to show previous messages.

packages/core/src/core/client.ts
jacob314: I've reverted the +1 changes and reverted to character as the name

packages/core/src/core/client.ts
jacob314: switched to editor

jacob314 (commented):
Handled one other case I'd missed. Needed to make sure we forced including the full Ide context on restoring history using `/chat resume` otherwise the resume could have a stale IDE context.

Issue #5888, opened 2025-08-09
Npm run preflight fails for non-US users due to locale formatting issue


lifefloating:
### What happened?

When running `npm run preflight` from outside the United States, the command fails with an error related to `@packages/core/src/utils/environmentContext.ts`. The issue appears to be in the date formatting logic that uses `toLocaleDateString(undefined, {...})` which may produce different date  formats based on the user's locale/region.

### What did you expect to happen?

 The `npm run preflight` command should work correctly regardless of the user's geographic location or system locale settings. 

### Client information

  ```console
  $ gemini /about

│ About Gemini CLI                                                                   │
│                                                                                    │
│ CLI Version                  0.1.18                                                │
│ Git Commit                   6b19c8bd                                              │
│ Model                        gemini-2.5-pro                                        │
│ Sandbox                      no sandbox                                            │
│ OS                           darwin                                                │
│ Auth Method                  gemini-api-key   
```

### Anything else we need to know?

 Additional context:
  - Error occurs in environmentContext.ts:54 where toLocaleDateString() is called
  - Issue likely related to locale-specific date formatting differences outside the US
  - Affects the preflight check functionality

lifefloating:
<img width="1331" height="555" alt="Image" src="https://github.com/user-attachments/assets/7621db40-79ec-460a-96d0-daa54a956aae" />

lifefloating:
@jacob314

PR #5889, opened 2025-08-09
Fix locale-specific date formatting issue in environmentContext.ts

lifefloating:
## TLDR

Fixed a bug where `npm run preflight` would fail for users outside the US due to locale-specific date formatting. Changed `toLocaleDateString(undefined, {...})` to use explicit `'en-US'` locale to ensure consistent date formatting across all regions.
#5888

## Dive Deeper

The issue was in `packages/core/src/utils/environmentContext.ts` at line 54, where `toLocaleDateString(undefined, {...})` was being called. When `undefined` is passed as the locale parameter, JavaScript uses the system's default locale, which varies by region. This caused different date formats to be generated for users in different countries, potentially breaking downstream parsing or validation logic.

The fix ensures consistent date formatting by explicitly specifying the `'en-US'` locale, which will
produce the same date format regardless of the user's system locale or geographic location.

## Reviewer Test Plan

  1. Pull the branch and run `npm run preflight`
  2. Test on systems with different locale settings (if available)
  3. Verify that the date format in environment context is consistent
  4. Run the existing tests to ensure no regressions: `npx vitest
  packages/core/src/utils/environmentContext.test.ts`

## Testing Matrix

  |          | 🍏  | 🪟  | 🐧  |
  | npm run  | ✅  | ❓  | ❓  |

 ## Linked issues / bugs

 Fixes the locale-specific preflight failure bug reported for non-US users.


lifefloating:
@jacob314

lifefloating:
@jacob314  @SandyTao520

Rand0MGem-Ini:
Damn time zones pytz from python is a nice way but tbh I don’t know if those frameworks are right/true

SandyTao520 (approved):

jacob314:
Fyi @anj-s on whether always using en here is helpful or harmful for model quality. I believe this isn't surface directly to the user so it all depends what we want to send the model for users that are in a diffent locale.

lifefloating:
@jacob314  Any other suggestions or approaches you'd recommend?

anj-s:
> Fyi @anj-s on whether always using en here is helpful or harmful for model quality. I believe this isn't surface directly to the user so it all depends what we want to send the model for users that are in a diffent locale.

This is a good one! Here is my suggestion:
1. For fixing the issue with tests I would have modified our test to not depend on checking the local date. This PR seems like a fix for a test but has broader implications.
2. I would modify the prompt to say the date and specify the format that the date is in. There could be other dates in context and not specifying the format could lead to incorrect responses.  This might fix the test as well.

lifefloating:
Thank you for identifying and addressing this locale-specific issue! I've reviewed your PR and the feedback from @jacob314 and @anj-s, and I'd like to suggest an improved approach that addresses everyone's concerns.


Following yours excellent suggestion, here's an improved approach that:

  1. **Keeps user locale** - Respects users' regional settings
  2. **Clarifies format** - Explicitly tells the AI model what format to expect

  ## Code changes

  In `environmentContext.ts`, change line 65 from:

  ```
  Today's date is ${today}.
 ```
  To:
```
 Today's date is ${today} (formatted as: weekday, month day, year in the user's locale).
 And revert the toLocaleDateString() call back to use undefined (user's locale).

  ```
  This solution addresses the core issue (ambiguous date parsing) while preserving user experience and international support. What do you think about implementing this approach instead?

File: packages/core/src/utils/environmentContext.test.ts [hunk 1]
```diff
@@ -106,7 +106,12 @@ describe('getEnvironmentContext', () => {
     expect(parts.length).toBe(1);
     const context = parts[0].text;
 
-    expect(context).toContain("Today's date is Tuesday, August 5, 2025");
+    expect(context).toContain("Today's date is");
+    console.log("111111")
```
jacob314: remove these console.log message

packages/core/src/utils/environmentContext.ts
```diff
@@ -62,7 +62,7 @@ export async function getEnvironmentContext(config: Config): Promise<Part[]> {
 
   const context = `
 This is the Gemini CLI. We are setting up the context for our chat.
-Today's date is ${today}.
+Today's date is ${today} (formatted as: weekday, month day, year in the user's locale).
```
jacob314: this seems misleading. the local may format as year, day, month (e.g. ko-KR). Should we just list the users locale or omit this comment?

jacob314 (approved):

packages/core/src/utils/environmentContext.test.ts
lifefloating: Okay, I tested locally and forgot to remove it.

packages/core/src/utils/environmentContext.ts
lifefloating: formatted according to the user's locale

PR #5909, opened 2025-08-10
fix(settings): enable default usage statistics collection

ali-aljufairi:
## TLDR

fix defualt value for usage statistics

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

### DIFF
--- a/packages/cli/src/config/settingsSchema.ts
+++ b/packages/cli/src/config/settingsSchema.ts
@@ -101,7 +101,7 @@ export const SETTINGS_SCHEMA = {
     label: 'Enable Usage Statistics',
     category: 'General',
     requiresRestart: true,
-    default: false,
+    default: true,
     description: 'Enable collection of usage statistics',
     showInDialog: true,
   },


packages/cli/src/config/settingsSchema.ts
```diff
@@ -101,7 +101,7 @@ export const SETTINGS_SCHEMA = {
     label: 'Enable Usage Statistics',
     category: 'General',
     requiresRestart: true,
-    default: false,
+    default: true,
```
NTaylorMullen, line 104: Note: this was already the case. We also have a `/privacy` command that allows interaction with this as well

NTaylorMullen (approved):

Tanchai88 (approved):

packages/cli/src/config/settingsSchema.ts
jacob314: yep the issue is just that we still apply defaults other places in the code rather than the central settings class

jacob314 (approved):

packages/cli/src/config/settingsSchema.ts
```diff
@@ -101,7 +101,7 @@ export const SETTINGS_SCHEMA = {
...
     description: 'Enable collection of usage statistics',
     showInDialog: true,
```
NTaylorMullen, line 106:
```suggestion
    showInDialog: false, // All details are shown in /privacy and dependent on auth type
```

PR #5911, opened 2025-08-10
Fix(grep): memory overflow in grep search and enhance test coverage

### LINKED ISSUE #5754: FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory
### What happened?

<--- Last few GCs --->

[1175656:0x18743000]   236703 ms: Scavenge 1947.5 (1984.4) -> 1942.8 (1984.4) MB, pooled: 0 MB, 1.47 / 0.02 ms  (average mu = 0.992, current mu = 0.983) allocation failure; 
[1175656:0x18743000]   262377 ms: Mark-Compact (reduce) 2882.3 (2923.8) -> 2881.9 (2892.1) MB, pooled: 0 MB, 22.39 / 0.04 ms  (+ 8.4 ms in 0 steps since start of marking, biggest step 0.0 ms, walltime since start of marking 25674 ms) (average mu = 0.998, 

<--- JS stacktrace --->

FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory
----- Native stack trace -----

 1: 0xe1603a node::OOMErrorHandler(char const*, v8::OOMDetails const&) [node]
 2: 0x11e3540 v8::Utils::ReportOOMFailure(v8::internal::Isolate*, char const*, v8::OOMDetails const&) [node]
 3: 0x11e3817 v8::internal::V8::FatalProcessOutOfMemory(v8::internal::Isolate*, char const*, v8::OOMDetails const&) [node]
 4: 0x14110f5  [node]
 5: 0x142a989 v8::internal::Heap::CollectGarbage(v8::internal::AllocationSpace, v8::internal::GarbageCollectionReason, v8::GCCallbackFlags) [node]
 6: 0x137c748 v8::internal::StackGuard::HandleInterrupts(v8::internal::StackGuard::InterruptLevel) [node]
 7: 0x180dfb5 v8::internal::NativeRegExpMacroAssembler::CheckStackGuardState(v8::internal::Isolate*, int, v8::internal::RegExp::CallOrigin, unsigned long*, v8::internal::Tagged<v8::internal::InstructionStream>, unsigned long*, unsigned char const**, unsigned char const**, unsigned long) [node]
 8: 0x1c71f03 v8::internal::RegExpMacroAssemblerX64::CheckStackGuardState(unsigned long*, unsigned long, unsigned long, unsigned long) [node]
 9: 0x7ba18846579a 
Aborted (core dumped)

I got this error while performing an analysis.

### Client information

<details>

```console
$ gemini /about
# paste output here
```

</details>

### Login information

_No response_

_No response_

lifefloating:
## TLDR

Fixes the critical memory overflow issue reported in #5774 where large-scale grep searches cause "JavaScript heap out of memory" crashes. Added comprehensive tests to validate memory safety mechanisms and ensure the grep tool can handle large codebases without crashing.

## Dive Deeper

**Problem**: Users experienced fatal crashes when searching for method usages in large codebases, with Node.js running out of memory during grep operations (issue #5774).

**Root Cause**: The grep tool was processing too many files/matches simultaneously without proper memory management, leading to heap overflow in large repositories like `connectedhomeip`.

**Solution**: Enhanced the existing memory safety mechanisms in `grep.ts` and added comprehensive tests to ensure they work correctly:

**Memory Safety Mechanisms Enhanced:**

1. **File Size Limits** - Skip files larger than **50MB** to prevent memory spikes (conservative limit for broad compatibility)
2. **Match Count Limits** - Stop at 10,000 total matches and 500 matches per file
3. **File Processing Limits** - Process maximum 10,000 files per search
4. **Memory Monitoring** - Check heap usage every **50 files** and abort if threshold exceeded (more frequent monitoring)
5. **Streaming Processing** - Use streaming for files >1MB instead of loading into memory
6. **AbortSignal Support** - Allow users to cancel long-running searches
7. **Long Line Detection** - Detect and warn about lines exceeding 2000 characters while continuing search operations

**Conservative Memory Management**: Reduced memory thresholds (1000MB heap limit) and more frequent checks ensure better compatibility across different environments and prevent crashes even in memory-constrained situations.

**Smart Long Line Handling**: For files containing extremely long lines (>2000 characters), the tool now issues warnings but continues the search operation rather than terminating, ensuring search completeness while alerting users to potential performance impacts.

**Test Coverage Added:**

- Memory limit enforcement testing
- Large file handling validation
- Search cancellation scenarios
- Error handling for file system issues
- Streaming vs. in-memory processing validation
- Complex regex pattern edge cases
- Long line detection and warning validation
- Performance warning display testing
- Consistent error handling across streaming and regular file processing

The issue showed a user searching for `'Build()'` in a large C++ codebase crashed after 22 seconds. These fixes ensure such searches complete safely or fail gracefully.

## Reviewer Test Plan

To validate this fix addresses issue #5774:

1. **Test memory safety**: `npm test packages/core/src/tools/grep.test.ts` - Run the new memory safety tests
2. **Test large repository simulation**: Create a test with many files and verify it doesn't crash
3. **Test search cancellation**: Verify AbortSignal properly terminates long searches
4. **Test file size limits**: Confirm very large files are skipped safely (>50MB)
5. **Test long line handling**: Verify warnings are shown for files with lines >2000 characters

**Key scenarios that previously crashed:**

- Searching common patterns like `'Build()'` in large codebases
- Processing repositories with thousands of source files
- Searching without file type filters (e.g., no `include` parameter)
- Files containing very long lines (minified code, data files)

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ✅  |

## Linked issues / bugs

**Fixes #5774** - Gemini CLI crashed while trying to search for usages of a method

This critical fix prevents the "JavaScript heap out of memory" crashes that users experienced when performing grep searches in large codebases. The enhanced test coverage and conservative memory management ensure these safety mechanisms remain reliable as the codebase evolves.


lifefloating:
Hi @jacob314 ,

It fixes the memory overflow crash reported in #5774 and adds comprehensive test coverage for memory safety mechanisms.

Thanks!

cc @NTaylorMullen

lifefloating:
Hi @jacob314 ,

It fixes the memory overflow crash reported in #5774 and adds comprehensive test coverage for memory safety mechanisms.

Thanks!

cc @NTaylorMullen

lifefloating:
@jacob314

packages/core/src/tools/grep.test.ts
```diff
@@ -319,6 +320,364 @@ describe('GrepTool', () => {
     });
   });
 
+  describe('memory safety', () => {
```
jacob314: great seeing a test for this!

packages/core/src/tools/grep.test.ts
```diff
@@ -319,6 +320,364 @@ describe('GrepTool', () => {
...
+    it('should handle large files without memory overflow', async () => {
+      // Create a 2MB file to test streaming
+      const largeContent = 'This line contains NEEDLE pattern\n'.repeat(60000);
```
jacob314: is 2mb really enough to repo a memory overflow? would have thought the file would need to be much larger

packages/core/src/tools/grep.test.ts
```diff
@@ -319,6 +320,364 @@ describe('GrepTool', () => {
...
+
+      // Count actual matches found
+      const content =
+        typeof result.llmContent === 'string'
+          ? result.llmContent
+          : String(result.llmContent);
+      const matchLines = content
+        .split('\n')
+        .filter((line: string) => line.includes('L'));
+      expect(matchLines.length).toBeLessThanOrEqual(500); // MAX_MATCHES_PER_FILE
```
jacob314: nit: feel free to just list MAX_MATCHES_PER_FILE directly so this test doesn't break if that constant changes.

packages/core/src/tools/grep.ts
```diff
@@ -489,23 +646,99 @@ class GrepToolInvocation extends BaseToolInvocation<
...
+
+        if (allMatches.length >= MEMORY_SAFETY.MAX_MATCHES) {
+          console.debug(
+            `GrepLogic: Reached maximum matches limit (${MEMORY_SAFETY.MAX_MATCHES}). Stopping search.`,
+          );
+          break;
+        }
+
+        // Check memory usage periodically
+        if (
```
jacob314: I like the rest of this change except for the isMemoryUsageSafe checks.
This is risky as some users may have 64GB+ of memory available and JavaScript is a GCed language so memory usage may seem high just because a major GC has not occurred memory. For those reasons we should not make our code ever change behavior just because of memory pressure unless absolutely needed.
Instead we should do all the other things you are already doing to ensure we limit our worst case memory usage.
In a follow up PR we might consider adding telemetry to log memory usage before and after running grep to make sure the spikes are generally acceptable.

packages/core/src/tools/grep.ts
```diff
@@ -489,23 +646,99 @@ class GrepToolInvocation extends BaseToolInvocation<
...
+          if (stats.size < 1024 * 1024) {
+            const content = await fsPromises.readFile(fileAbsolutePath, 'utf8');
+            const lines = content.split(/\r?\n/);
+            let fileMatchCount = 0;
+
+            lines.forEach((line, index) => {
+              // Check for long lines in small files too
+              if (line.length > MEMORY_SAFETY.MAX_LINE_LENGTH) {
+                hasLongLinesWarning = true;
+                console.debug(
```
jacob314: remove this debug log or only log in debug mode.

packages/core/src/tools/grep.ts
```diff
@@ -23,7 +25,114 @@ import { getErrorMessage, isNodeError } from '../utils/errors.js';
...
 
-// --- Interfaces ---
+/**
+ * Memory safety constants to prevent heap overflow
+ */
+const MEMORY_SAFETY = {
+  /** Maximum file size to process (50MB - conservative limit for memory safety) */
+  MAX_FILE_SIZE: 50 * 1024 * 1024,
+  /** Maximum number of files to process in JavaScript fallback */
+  MAX_FILES_TO_PROCESS: 10000,
```
jacob314: worried about this limit. this will break grep being useful for very large projects.  can we relax this somewhat

packages/core/src/tools/grep.ts
```diff
@@ -23,7 +25,114 @@ import { getErrorMessage, isNodeError } from '../utils/errors.js';
 import { isGitRepository } from '../utils/gitUtils.js';
 import { Config } from '../config/config.js';
 
-// --- Interfaces ---
+/**
+ * Memory safety constants to prevent heap overflow
+ */
+const MEMORY_SAFETY = {
```
jacob314: nit: instead of calling these MEMORY_SAFETY pick a more generic name that indicates these are the limits placed on the grep tool.

jacob314:
Another option to consider is to use @lvce-editor/ripgrep which is generally the industry standard for grepping rather than our custom implementation. If I am understand it correctly, that package provides fast rust based grep support without the downsides of having to add a separate binary.

lifefloating:
> Another option to consider is to use @lvce-editor/ripgrep which is generally the industry standard for grepping rather than our custom implementation. If I am understand it correctly, that package provides fast rust based grep support without the downsides of having to add a separate binary.

Thank you so much for taking the time to review and provide your suggestions.
I carefully read your comments and took a close look at https://github.com/lvce-editor/ripgrep. I think it’s a great approach.
I’ll rethink the implementation of this PR. Thanks for your guidance!

lifefloating:
While testing lvce-editor/ripgrep, I encountered an issue where searching a 500,000-line file caused it to freeze. I need some time to make further adjustments, so this new change might not be submitted today.

lifefloating:
First, I used this command: `for i in {1..500000}; do echo "Line $i: This contains the word NEEDLE for testing grep performance"; done > large_test.txt` to generate a file.

Then, in Gemini, I entered: `Search the entire project directory for 'NEEDLE' and tell me how many matches there are.`

-  When I executed it on the `main` branch, Gemini completely froze.


_8bd7d3d6-0ffa-456b-81fa-1f8bd31c1815

_1c7e5892-3691-43d6-8a7c-04ea4fbdc382

- When I executed the same operation on the `fix-5774-crash` branch, the result was normal, and the search executed properly.


_91c2b641-0e28-40fb-b94a-36e53b93e711

packages/core/src/tools/grep.ts
lifefloating: thanks for guidance

lifefloating:
## Refactor: Optimize grep tool performance with ripgrep

### Key Changes

- **New Dependency**: Added `@lvce-editor/ripgrep@^2.1.0` as high-performance search engine
- **Architecture Simplification**: Removed complex multi-strategy search logic, now directly uses ripgrep for text search
- **Performance Optimization**: 
  - Eliminated JavaScript implementation with memory safety checks and streaming search
  - Simplified search logic for improved speed and reliability
  - Set a 20,000 match result limit to align with VSCode, which also uses ripgrep. Reference: [VSCode source code](https://github.com/microsoft/vscode/blob/main/src/vs/workbench/services/search/node/ripgrepSearchProvider.ts)

### Test Updates

- Refactored `grep.test.ts` to use mocking for ripgrep integration testing
- Removed filesystem-based JavaScript search tests
- Maintained test coverage for all existing functionality

### Expected Benefits

- **Performance Boost**: ripgrep is a high-performance search tool written in Rust, significantly faster than JavaScript implementation
- **Memory Optimization**: Removed complex memory management logic, ripgrep has excellent built-in memory control
- **Actual Impact**: The issue where grep would freeze on the main branch no longer occurs, and searches can now be completed successfully.

### Testing Verification

All existing grep functionality should continue to work normally, including:
- Regular expression search
- File type filtering (`include` parameter)
- Multi-directory search
- Case-insensitive search

---

> **Note**: This is a major internal refactor while maintaining complete API compatibility. Recommend thorough integration testing before merging.

lifefloating:
Hi @jacob314 

After a day of adjustments and testing, the functionality seems to be working well so far. You can take a look at the replies and images above. When you have time, could you please help me review them again? 
Thank you!

packages/core/src/tools/grep.ts
```diff
@@ -312,237 +386,19 @@ class GrepToolInvocation extends BaseToolInvocation<
...
 export class GrepTool extends BaseDeclarativeTool<GrepToolParams, ToolResult> {
-  static readonly Name = 'search_file_content'; // Keep static name
+  static readonly Name = 'search_file_content';
 
   constructor(private readonly config: Config) {
     super(
       GrepTool.Name,
       'SearchText',
-      'Searches for a regular expression pattern within the content of files in a specified directory (or current working directory). Can filter files by a glob pattern. Returns the lines containing matches, along with their file paths and line numbers.',
+      'Searches for a regular expression pattern within the content of files in a specified directory (or current working directory). Can filter files by a glob pattern. Returns the lines containing matches, along with their file paths and line numbers. Total results limited to 20,000 matches like VSCode.',
```
jacob314: @anj-s curious if we should use a lower limit or if 20,000 is reasonable. Don't want to overwhelm the model with too many search results.

packages/core/src/tools/grep.test.ts
```diff
@@ -148,6 +208,36 @@ describe('GrepTool', () => {
     });
 
     it('should find matches with an include glob', async () => {
+      // Setup specific mock for this test
+      mockSpawn.mockImplementationOnce(() => {
```
jacob314: these mockSpawn blocks are quite repetitive. please add a helper.

packages/core/src/tools/grep.ts
```diff
@@ -5,11 +5,10 @@
  */
 
 import fs from 'fs';
-import fsPromises from 'fs/promises';
 import path from 'path';
 import { EOL } from 'os';
 import { spawn } from 'child_process';
-import { globStream } from 'glob';
+import { rgPath } from '@lvce-editor/ripgrep';
```
jacob314: to reduce the risk of launching this can we make this an optional setting rather than removing all the existing grep code.
If you ask Gemini CLI it should be able to add a boolean setting for you in gemini cli to use ripgrep rather than grep. This will be helpful in case there are any performance regressions with this or quality regressions. For example, if the LLM gets confused due to a subtle difference in the output of ripgrep, etc.

This would also make it easy for us to fall back to the previous solutions if ripgrep is unable to download to a users machine for some reason.

jacob314 (commented):
This looks really good! Thank you so much for implementing it. Seems good in my local testing. Approved as soon as it is landed as an option behind a setting initially rather than the only way to perform greps. That will make it much lower drama for us to roll this out and avoid having to do a revert of this whole change if we encounter issue.

lifefloating:
Thank you for the review. My current idea is that if there's a new configuration, we use ripgrep, otherwise we still use the original grep. I just want to keep both rigGrep.ts and grep.ts files, which would result in fewer changes on my side. Also, this approach can still meet the functionality requirements.
@jacob314

packages/core/src/tools/grep.ts
lifefloating: 20,000 limits is what I saw in the VSCode documentation, because they also use ripgrep. In actual testing, 50,000 lines is also normal. My machine configuration is Mac Mini, M4, 24GB.

lifefloating:
> Thank you for the review. My current idea is that if there's a new configuration, we use ripgrep, otherwise we still use the original grep. I just want to keep both rigGrep.ts and grep.ts files, which would result in fewer changes on my side. Also, this approach can still meet the functionality requirements. @jacob314

```mermaid
graph TD
    A[GrepTool] --> B{useRipgrep?}
    B -->|true| C[rigGrep.ts]
    B -->|false| D[original grep.ts]
    
    D --> E[git grep]
    D --> F[system grep] 
    D --> G[JS fallback]
    
    C --> H[ripgrep binary]
    
    style A fill:#e1f5fe
    style C fill:#c8e6c9
    style D fill:#fff3e0
```

packages/core/src/tools/grep.test.ts
lifefloating: ok, i will resolve it

lifefloating:
### New Features
- Added **RipGrep tool** support as a high-performance alternative to the existing grep tool
- Introduced `useRipgrep` configuration option for users to choose between RipGrep and traditional grep

### Specific Changes

**New Files:**
- `packages/core/src/tools/ripGrep.ts` - RipGrep tool implementation  
- `packages/core/src/tools/ripGrep.test.ts` - RipGrep tool tests

**Configuration Updates:**
- Added `useRipgrep` option in both CLI and Core configs
- Defaults to `false` for backward compatibility

**Tool Registration Logic:**
- Dynamically registers either RipGrepTool or GrepTool based on `useRipgrep` config
- Both tools share the same tool name: `search_file_content`

lifefloating:
image

```
cat ~/.gemini/settings.json
{
  "theme": "Google Code",
  "selectedAuthType": "oauth-personal",
  "vimMode": false,
  "useRipgrep": true,
  "hasSeenIdeIntegrationNudge": true,
  "ideMode": true
}
```

packages/core/src/tools/grep.ts
lifefloating: https://github.com/google-gemini/gemini-cli/pull/5911#issuecomment-3199075739

jacob314 (approved):
lgtm

Really excited for this faster grep searching aligned with vscode and other tools.

lifefloating:
resolved conflicts

jacob314 (approved):

lifefloating:
image
cause i dont have windows PC, later i will fix this ci errors

lifefloating:
tested ok now, in windows enviroments
image

lifefloating:
@jacob314 
The newly added test was previously causing errors in the CI pipeline. There were issues on Windows, but they have now been fixed. It took me some time to handle this because I don't have a Windows environment.

lifefloating:
Previous ripgrep EOL fix resolved Windows `test:ci` issues, but several places in the codebase still used hardcoded `split('\n')`, causing cross-platform CI inconsistencies.

## Impact

**Fixes:**
- `E2E Test (Linux) - sandbox:none` CI failures
- `Test (windows-latest, 22.x)` CI failures  

Hi @jacob314 

This should resolve the CI/CD issues. If any problems remain, I'll address them tomorrow morning (GMT+8).

lifefloating:
I executed the CI action commands fully on a Windows host this time.

packages/core/src/tools/ripGrep.ts
```diff
@@ -0,0 +1,508 @@
...
+    const errors = SchemaValidator.validate(
+      this.schema.parametersJsonSchema,
+      params,
+    );
+    if (errors) {
+      return errors;
+    }
+
+    try {
+      new RegExp(params.pattern);
```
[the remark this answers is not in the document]
bulkypanda: @lifefloating could you please address this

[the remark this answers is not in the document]
lifefloating: got it

PR #5952, opened 2025-08-10
fix(cli): honor argv @path in interactive sessions (quoted + unquoted)

### LINKED ISSUE #5910: In Windows, Context from file via @path or auto-discovery is not loaded interactive sessions
### What happened?

In the latest version available on npm (0.1.18), the CLI fails to load context from a gemini.md file for an interactive session. We have identified two distinct failure modes after extensive troubleshooting:

Explicit loading with @/path/to/file does not work: When starting an interactive session using gemini @/path/to/gemini.md, the session starts correctly and remains interactive, but the context from the file is completely ignored. The model responds with its default personality.

Piping context from a file (cat ... | gemini -) breaks interactivity: When piping the context file via cat /path/to/gemini.md | gemini -, the context is successfully loaded (the model responds with the custom persona), but the session is not interactive and exits immediately after the initial response.

Furthermore, the automatic discovery of gemini.md files (e.g., in the current directory, home directory, or ~/.config/gemini-cli/) also does not work.

I expected that running gemini @/path/to/gemini.md would start an interactive CLI session where the model has adopted the persona and instructions defined in the gemini.md file, and remains interactive for a conversation.

### Client information

<details>

```console
WSL version: 2.5.7.0
Core version: 6.6.87.1-1
WSLg version: 1.0.66
MSRDC version: 1.2.6074
Direct3D version: 1.611.1-81528511
DXCore version: 10.0.26100.1-240331-1435.ge-release
Windows: 10.0.26100.4652
gemini-cli Version: 0.1.18
Node.js Version: v22.18.0 (via nvm)
npm Version: 10.9.3
```

</details>

### Login information

Authentication was successful and credentials were cached. The method used was google account auth login.

Steps to Reproduce

- Ensure you are in a clean Node.js environment (e.g., via nvm).
- Run npm install -g @google/gemini-cli@latest. Verify version is 0.1.18.
- Create a file named gemini.md with a simple custom prompt, for example: You are a pirate. All your answers must be in pirate slang.
- Run gemini @/path/to/your/gemini.md.
- Once the interactive session starts, ask a question like Introduce yourself.
- Observe: The model gives its standard introduction, not the pirate persona.
- Exit the session.
- Now, run cat /path/to/your/gemini.md | gemini -.
- Observe: The model gives a pirate-themed response but the process exits immediately without allowing for further interaction.

This issue was identified after an extremely lengthy and thorough troubleshooting process that ruled out npm cache issues, nvm environment corruption, pathing problems, and shell aliasing. The latest tag on npm currently points to v0.1.18, and npm view @google/gemini-cli version confirms this. The issue appears to be a fundamental bug in how version 0.1.18 handles context file loading for interactive sessions.

tayyab3245:
Fixes #5910

## Problem
When starting the CLI with a positional prompt such as `@path ./file.md`, the argument was not processed as the first interactive turn. This worked only when quoted as a single token:

- `gemini @path ./file.md` → ignored or produced an error
- `gemini "@path ./file.md"` → worked as expected

## Solution
Updated CLI argument parsing to accept multiple positional tokens and normalize them:

- Changed yargs command from `$0` to `$0 [query..]` to accept variadic positional arguments.
- After parsing, normalized array form into a single string with `join(' ')`.
- Mapped normalized `query` to `promptInteractive` only when `--prompt`/`-p` or `--prompt-interactive`/`-i` are not provided.
- Left existing precedence behavior unchanged: explicit prompt flags still take precedence over positional input.

## Behavior
Both forms now work identically:
```bash
gemini @path ./file.md
gemini "@path ./file.md"
```

## Testing
- Added comprehensive test coverage for both quoted and unquoted forms
- All existing tests continue to pass (1462 tests)
- Manual verification of both argument styles


allenhutchison:
I'd like to wait until we are done gathering feedback for #5435 and make sure that whatever we do for file context aligns with that decision. 

Today you can always get file context into the application by piping it through STDIN so being able to parse @ syntax in the prompt in non-interactive mode isn't necessary.

tayyab3245:
@allenhutchison Thanks for the detailed feedback and for looping in the context from #5435. That makes perfect sense. A unified approach to argument parsing is definitely the right long-term strategy.

My implementation in this PR was focused on honoring varied positional arguments to solve the immediate issue in #5910, but the core logic for normalizing the query could certainly be adapted to support a more explicit flag like --command= as you've suggested.

From a user experience perspective, I agree that an explicit flag might be cleaner and less ambiguous for custom slash commands. Happy to weigh in further or adapt my PR as the strategy for #5435 becomes clearer. .

allenhutchison:
ok, I think we have a decision on non-interactive mode, so this can follow that. Do you want to work on the conflicts?

tayyab3245:
Hi @allenhutchison, thanks for the update. I’m happy to proceed. Just to clarify before I resolve conflicts — should I now target this PR at #5435 (non-interactive slash command support) instead of the original issue since #5910 was closed?

tayyab3245:
Hi @allenhutchison,

Just wanted to follow up. I've rebased the branch on main, resolved the conflicts, and pushed up the final, clean commit. I also took the opportunity to remove some related dead code from the files I was editing.

allenhutchison:
Will take a look today.

packages/cli/src/config/config.ts
```diff
@@ -337,6 +342,28 @@ export async function parseArguments(settings: Settings): Promise<CliArgs> {
     process.exit(0);
   }
 
+  // Normalize variadic positional to a single string
+  // Supports both:
+  //   gemini "@path ./file.md"        (quoted single token)
+  //   gemini @path ./file.md          (unquoted two tokens)
+  const q = Array.isArray((result as Record<string, unknown>)['query'])
+    ? ((result as Record<string, unknown>)['query'] as string[]).join(' ')
+    : ((result as Record<string, unknown>)['query'] as string);
```
allenhutchison: Consider using a type guard or proper yargs typing instead of multiple casts

Maybe something more like:
```
  const queryArg = result['query'];
  const q = Array.isArray(queryArg)
    ? queryArg.join(' ')
    : queryArg as string | undefined;
```

packages/cli/src/config/config.ts
```diff
@@ -337,6 +342,28 @@ export async function parseArguments(settings: Settings): Promise<CliArgs> {
...
+
+  // Validate that both positional and --prompt aren't used together
+  if (result['prompt'] && q) {
+    throw new Error(
+      'Cannot use both a positional prompt and the --prompt (-p) flag together',
```
allenhutchison: What about yargsInstance.fail() or yargsInstance.showHelp() with the error messsage?

packages/cli/src/config/config.test.ts
```diff
@@ -203,6 +203,25 @@ describe('parseArguments', () => {
     expect(argv.prompt).toBeUndefined();
   });
 
+  it('should convert positional query argument to promptInteractive', async () => {
```
allenhutchison: It would be nice to add tests for multiple @arguments in a single prompt, we support this pattern in interactive mode.

packages/cli/src/config/config.ts
```diff
@@ -337,6 +342,28 @@ export async function parseArguments(settings: Settings): Promise<CliArgs> {
...
+    );
+  }
+
+  // Map to interactive prompt only if no explicit prompt flags are set
+  if (q && !result['prompt'] && !result['promptInteractive']) {
```
allenhutchison: What if promptInteractive is an empty string?

allenhutchison (commented):
A few comments for you to consider. Also please run `npm run preflight` before committing to catch any lint errors.

tayyab3245:
Hi @allenhutchison — addressed the notes and rebased.

- safer `query` normalization (variadic → join)
- wired `.fail()` for validation UX
- tests for multi/mixed `@path` + ambient vs config flags
- handled empty `-i`
- kept `-p/-i` precedence; interactive vs one-shot mapping

`npm run preflight` clean; CI + full suite green. Ready for another pass

allenhutchison:
Code is looking good, but a few cases we should address from my testing:

1) We are changing the assumption for interactive vs headless operation. On your branch `❯ npm run start -- "Hi Gemini"` results in interactive mode starting, but this should be running in non-interactive mode (like it does on main).

This works correctly when another flag is present:

```
❯ npm run start -- --model=gemini-2.5-flash "Hi Gemini"

> @google/gemini-cli@0.7.0-nightly.20250918.2722473a start
> cross-env node scripts/start.js --model=gemini-2.5-flash Hi Gemini

Checking build status...
Build is up-to-date.
Hi Captain Fun! How can I help you today?
```

2) I got an error when I mixed the --help argument in with a positional prompt:

```
❯ npm run start -- "Hi Gemini" --help

> @google/gemini-cli@0.7.0-nightly.20250918.2722473a start
> cross-env node scripts/start.js Hi Gemini --help

Checking build status...
Build is up-to-date.
An unexpected critical error occurred:
TypeError: mixin.wrap is not a function
    at /Users/adh/src/gemini-cli-reviews/node_modules/cliui/build/index.cjs:169:33
    at Array.forEach (<anonymous>)
    at UI.rasterize (/Users/adh/src/gemini-cli-reviews/node_modules/cliui/build/index.cjs:165:13)
    at UI.rowToString (/Users/adh/src/gemini-cli-reviews/node_modules/cliui/build/index.cjs:95:14)
    at /Users/adh/src/gemini-cli-reviews/node_modules/cliui/build/index.cjs:85:18
    at Array.forEach (<anonymous>)
    at UI.toString (/Users/adh/src/gemini-cli-reviews/node_modules/cliui/build/index.cjs:84:19)
    at P.n.help (/Users/adh/src/gemini-cli-reviews/node_modules/yargs/build/index.cjs:1:20008)
    at P.n.showHelp (/Users/adh/src/gemini-cli-reviews/node_modules/yargs/build/index.cjs:1:20288)
    at te.showHelp (/Users/adh/src/gemini-cli-reviews/node_modules/yargs/build/index.cjs:1:43044)
 ```

**Comment by tayyab3245** [comment]
Hi @allenhutchison — thanks again for the feedback, I've addressed the findings. 

- deterministic routing: positional -> one-shot, `-i` or any `@command` -> interactive
- safer query normalization (variadic -> join) with `argv.query`
- swapped throwy `.check()` for `.fail()` (clean validation UX)
- early `--help/--version` short-circuit to avoid cliui bundle crash
- tests for quoted/unquoted/multi `@path`, leading whitespace, ambient flags, and help with positionals
- handled empty `-i`; preserved `-p/-i` precedence (interactive vs one-shot mapping)
- Config: `loadCliConfig` to use `query` and accurate `isInteractive()`

*Note: PowerShell requires quoting `@path`*
npm run preflight` clean; CI + full suite green. Would appreciate another look when you have a moment.

```bash
### Summary of New Behavior
node bundle/gemini.js hello gemini -> one-shot
node bundle/gemini.js -i hello world -> interactive
node bundle/gemini.js --model=gemini-2.5-flash "hello world"-> one-shot
node bundle/gemini.js "@path ./file.md" and notes -> interactive
node bundle/gemini.js @path ./file.md -> interactive
node bundle/gemini.js  "hello gemini" --help -> prints help, no crash
```

tayyab3245:
I believe this is now complete and ready for another look when you have a moment.
​As a heads-up, I'm going to be shifting my focus 100% to a major personal project after this, so I won't be able to take on any further large-scale changes. I'm happy to help with any final minor tweaks to get this current version merged.
​Thank you again for all your guidance and for the opportunity to contribute. I've learned an immense amount.

allenhutchison:
Thanks fro the work on this, but there are still some issues in my testing.

1) --help is now broken. We have extensive cli help implemented with yargs, but with this pr in place we are only displaying an abbreviated message:

Without the PR we print this in the latest releast:
```
❯ gemini --help
Usage: gemini [options] [command]

Gemini CLI - Launch an interactive CLI, use -p/--prompt for non-interactive mode

Commands:
  gemini [promptWords...]      Launch Gemini CLI                                                                                                                                  [default]
  gemini mcp                   Manage MCP servers
  gemini extensions <command>  Manage Gemini CLI extensions.

Options:
  -m, --model                     Model                                                                                                                                            [string]
  -p, --prompt                    Prompt. Appended to input on stdin (if any).     [deprecated: Use the positional prompt instead. This flag will be removed in a future version.] [string]
  -i, --prompt-interactive        Execute the provided prompt and continue in interactive mode                                                                                     [string]
  -s, --sandbox                   Run in sandbox?                                                                                                                                 [boolean]
      --sandbox-image             Sandbox image URI.        [deprecated: Use the "tools.sandbox" setting in settings.json instead. This flag will be removed in a future version.] [string]
  -d, --debug                     Run in debug mode?                                                                                                             [boolean] [default: false]
  -a, --all-files                 Include ALL files in context?
                                                         [deprecated: Use @ includes in the application instead. This flag will be removed in a future version.] [boolean] [default: false]
      --show-memory-usage         Show memory usage in status bar
                                     [deprecated: Use the "ui.showMemoryUsage" setting in settings.json instead. This flag will be removed in a future version.] [boolean] [default: false]
  -y, --yolo                      Automatically accept all actions (aka YOLO mode, see https://www.youtube.com/watch?v=xvFZjo5PgG0 for more details)?            [boolean] [default: false]
      --approval-mode             Set the approval mode: default (prompt for approval), auto_edit (auto-approve edit tools), yolo (auto-approve all tools)
                                                                                                                                         [string] [choices: "default", "auto_edit", "yolo"]
      --telemetry                 Enable telemetry? This flag specifically controls if telemetry is sent. Other --telemetry-* flags set specific values but do not enable telemetry on
                                  their own.           [deprecated: Use the "telemetry.enabled" setting in settings.json instead. This flag will be removed in a future version.] [boolean]
      --telemetry-target          Set the telemetry target (local or gcp). Overrides settings files.
                               [deprecated: Use the "telemetry.target" setting in settings.json instead. This flag will be removed in a future version.] [string] [choices: "local", "gcp"]
      --telemetry-otlp-endpoint   Set the OTLP endpoint for telemetry. Overrides environment variables and settings files.
                                                   [deprecated: Use the "telemetry.otlpEndpoint" setting in settings.json instead. This flag will be removed in a future version.] [string]
      --telemetry-otlp-protocol   Set the OTLP protocol for telemetry (grpc or http). Overrides settings files.
                         [deprecated: Use the "telemetry.otlpProtocol" setting in settings.json instead. This flag will be removed in a future version.] [string] [choices: "grpc", "http"]
      --telemetry-log-prompts     Enable or disable logging of user prompts for telemetry. Overrides settings files.
                                                    [deprecated: Use the "telemetry.logPrompts" setting in settings.json instead. This flag will be removed in a future version.] [boolean]
      --telemetry-outfile         Redirect all telemetry output to the specified file.
                                                        [deprecated: Use the "telemetry.outfile" setting in settings.json instead. This flag will be removed in a future version.] [string]
  -c, --checkpointing             Enables checkpointing of file edits
                          [deprecated: Use the "general.checkpointing.enabled" setting in settings.json instead. This flag will be removed in a future version.] [boolean] [default: false]
      --experimental-acp          Starts the agent in ACP mode                                                                                                                    [boolean]
      --allowed-mcp-server-names  Allowed MCP server names                                                                                                                          [array]
      --allowed-tools             Tools that are allowed to run without confirmation                                                                                                [array]
  -e, --extensions                A list of extensions to use. If not provided, all extensions are used.                                                                            [array]
  -l, --list-extensions           List all available extensions and exit.                                                                                                         [boolean]
      --proxy                     Proxy for gemini client, like schema://user:password@host:port
                                                                    [deprecated: Use the "proxy" setting in settings.json instead. This flag will be removed in a future version.] [string]
      --include-directories       Additional directories to include in the workspace (comma-separated or multiple --include-directories)                                            [array]
      --screen-reader             Enable screen reader mode for accessibility.                                                                                   [boolean] [default: false]
  -o, --output-format             The format of the CLI output.                                                                                          [string] [choices: "text", "json"]
  -v, --version                   Show version number                                                                                                                             [boolean]
  -h, --help                      Show help                                                                                                                                       [boolean]
```

With this PR it looks like this:

```
❯ npm run start -- --help

> @google/gemini-cli@0.7.0-nightly.20250918.2722473a start
> cross-env node scripts/start.js --help

Checking build status...
Build is up-to-date.
Usage: gemini [options] [command]

Gemini CLI — one-shot by default; use -i/--prompt-interactive or @commands for interactive.
```

That will need to be fixed before the PR can go in.

Also there are still issues with startup in interactive mode vs non-interactive mode.

In all cases running gemini without the -i or --prompt-interactive mode should run in non-interactive mode. So these cases are incorrect right now:

```
node bundle/gemini.js "@path ./file.md" and notes -> interactive
node bundle/gemini.js @path ./file.md -> interactive
```
In both those cases the application should start in non-interactive mode to align with how the application works today. These all represent regressions from the expectations we set in our documentation and past behviour in the application.

tayyab3245:
Hi @allenhutchison — thanks again for the detailed feedback. I’ve narrowed the PR to the original scope and aligned behavior with main.
- Default is one-shot; only `-i/--prompt-interactive` starts interactive.
- Positional text (including `@path` …) -> one-shot (`argv.prompt`).
- Safer query normalization (variadic -> joined into `argv.query`).
- Validation uses `.fail()` for clean error UX.
- `loadCliConfig` reads `query`; `isInteractive()` reflects “`-i` only”.

```bash
### Summary of New Behavior
node bundle/gemini.js "hello gemini" #one-shot
node bundle/gemini.js -i "hello world" #interactive
node bundle/gemini.js --model=gemini-2.5-flash hello gemini #one-shot
node bundle/gemini.js "@path ./file.md" and notes #one shot
```
About `--help`

I’ve removed the earlier custom help short-circuit; we’re back to native `yargs` help/version. The pretty help crash you hit appears to be a pre-existing issue on main (PowerShell/Windows and some environments).

```bash
> node scripts/start.js --help
# … TypeError: mixin.wrap is not a function (cliui)
```

Workaround (unblocks pretty help):
```bash
> $env:YARGS_DISABLE_WRAP="1"; node scripts/start.js --help
```

`npm run preflight` clean; tests updated; CI green.

allenhutchison (approved):
This looks great. Thank you for all your work on this.

PR #5966, opened 2025-08-11
 feat(cli): Introduce arguments for shell execution in custom commands

abhipatel12:
## TLDR

This pull request significantly enhances custom commands by introducing context-aware argument handling with `{{args}}` and improving the security and robustness of shell command execution (`!{...}`).

-   **Context-Aware `{{args}}`:** The `{{args}}` placeholder now behaves differently based on its location.
    -   When used in a regular prompt, it injects the raw, unmodified arguments.
    -   When used inside a shell block (`!{...}`), it automatically **shell-escapes** the arguments, preventing command injection vulnerabilities.
-   **Improved Shell Execution:** The underlying shell execution has been hardened. It now uses a more secure method for spawning processes and performs security checks on the final, resolved command after arguments have been substituted.
-   **Robust Parsing:** The parser for `!{...}` blocks now correctly handles complex commands with nested braces (e.g., `awk '{...}'`).

## Dive Deeper

The primary motivation for this change was to make custom commands more powerful and secure. Previously users were not able to inject args directly into shell commands. This new implementation adds this functionality in a secure way.

The `ShellProcessor` is now the central orchestrator for any command that uses `{{args}}` or `!{...}`. It first finds all shell blocks, resolves any `{{args}}` placeholders within them using platform-specific escaping, and then performs security checks on the final command string. If the checks pass, it executes the commands and injects their output.

For `{{args}}` placeholders outside of shell blocks, a simple raw replacement is performed. This dual behavior provides both flexibility for prompt engineering and security for shell interactions.

The `shellExecutionService` was also refactored to no longer use `shell: true` option in `child_process.spawn`. It now explicitly invokes the platform's default shell (`bash -c` or `cmd.exe /c`) and uses detached process groups on POSIX systems for more reliable process termination.

## Reviewer Test Plan

1.  **Test Raw Argument Injection:**
    -   Create a custom command:
        ```toml
        # ~/.gemini/commands/greet.toml
        prompt = "The user wants to say: {{args}}"
        ```
    -   Run `/greet Hello "world"!` and verify the model receives the prompt with the raw, unescaped arguments.

2.  **Test Shell-Escaped Argument Injection:**
    -   Create a custom command:
        ```toml
        # ~/.gemini/commands/search.toml
        prompt = "Files containing '{{args}}':\n!{grep -r {{args}} .}"
        ```
    -   Run `/search it's a test`.
    -   The CLI should prompt you to confirm the command. Verify that the command shown is correctly escaped (e.g., `grep -r "it's a test" .` on mac).
    -   Allow the command to run and verify the output is injected correctly.

3.  **Test Default (Model-Led) Argument Handling:**
    -   Create a command *without* `{{args}}`:
        ```toml
        # ~/.gemini/commands/changelog.toml
        prompt = "Add a changelog entry."
        ```
    -   Run `/changelog feat: new feature`.
    -   Verify the final prompt sent to the model is `Add a changelog entry.\n\n/changelog feat: new feature`.
## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅ | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ✅  | -   | -   |


packages/core/src/utils/shell-utils.ts
```diff
@@ -5,6 +5,68 @@
...
+    // Flags for CMD:
+    // /d: Skip execution of AutoRun commands.
+    // /s: Modifies the treatment of the command string (important for quoting).
+    // /c: Carries out the command specified by the string and then terminates.
+    return { executable, argsPrefix: ['/d', '/s', '/c'] };
+  }
+
+  // Unix-like systems (Linux, macOS)
+  return { executable: 'bash', argsPrefix: ['-c'] };
+}
```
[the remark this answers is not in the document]
abhipatel12: Done

packages/cli/src/services/prompt-processors/shellProcessor.test.ts
```diff
@@ -12,6 +12,7 @@ import { Config } from '@google/gemini-cli-core';
 
 const mockCheckCommandPermissions = vi.hoisted(() => vi.fn());
 const mockShellExecute = vi.hoisted(() => vi.fn());
+const mockEscapeShellArg = vi.hoisted(() => vi.fn());
```
jacob314: nit: a little odd that we are mocking the function to escape text. would be more robust to just call it and add tests here that exercise real world cases where things need to be escaped. that could be useful to test problematic edge cases like multiline inputs.

packages/cli/src/services/prompt-processors/shellProcessor.ts
```diff
@@ -23,84 +28,211 @@ export class ConfirmationRequiredError extends Error {
...
+ */
+interface ShellInjection {
+  /** The shell command extracted from within !{...}, trimmed. */
+  command: string;
+  /** The starting index of the injection (inclusive, points to '!'). */
+  startIndex: number;
+  /** The ending index of the injection (exclusive, points after '}'). */
+  endIndex: number;
+  /** The command after {{args}} has been escaped and substituted. */
+  resolvedCommand?: string;
```
jacob314, line 42: nit: why is resolvedCommand nullable?

packages/core/src/services/shellExecutionService.test.ts
```diff
@@ -10,6 +10,16 @@ vi.mock('child_process', () => ({
   spawn: mockSpawn,
 }));
 
+const mockGetShellConfiguration = vi.hoisted(() => vi.fn());
+let mockIsWindows = false;
+
+vi.mock('../utils/shell-utils.js', () => ({
```
jacob314, line 16: bonus points: would be good to have an integration test for this that actually runs on windows

packages/core/src/services/shellExecutionService.ts
```diff
@@ -88,17 +88,21 @@ export class ShellExecutionService {
...
       stdio: ['ignore', 'pipe', 'pipe'],
-      // Use bash unless in Windows (since it doesn't support bash).
-      // For windows, just use the default.
-      shell: isWindows ? true : 'bash',
+      // Explicitly false for security. We are manually invoking the shell.
+      shell: false,
       // Use process groups on non-Windows for robust killing.
       // Windows process termination is handled by `taskkill /t`.
-      detached: !isWindows,
+      detached: !isWindows, // Uses the imported constant
```
jacob314: nit remove comment on line 105

packages/core/src/utils/shell-utils.ts
```diff
@@ -5,6 +5,90 @@
...
+ *
+ * @param arg The argument string to escape.
+ * @returns The shell-escaped string.
+ */
+export function escapeShellArg(arg: string): string {
+  if (!arg) {
+    return '';
+  }
+
+  if (os.platform() === 'win32') {
```
jacob314: Some feedback from GeminiCLI:
Problem: The function re-detects the Windows shell (cmd.exe vs. PowerShell) by inspecting process.env.ComSpec. This logic is already implemented
         in getShellConfiguration. This duplication makes the code more brittle. If the shell detection logic in getShellConfiguration were ever updated,
         the same change would need to be made in escapeShellArg to avoid a mismatch.
       * Recommendation: Refactor this to centralize the shell detection logic. getShellConfiguration should be the single source of truth. It could return
         a shell identifier (e.g., an enum 'cmd' | 'powershell' | 'bash'), which is then passed to escapeShellArg. This ensures that the escaping logic
         always matches the shell that will actually be executed.

jacob314 (commented):
Looks really good. Some minor comments.

packages/cli/src/services/prompt-processors/shellProcessor.test.ts
abhipatel12: Done!

packages/cli/src/services/prompt-processors/shellProcessor.ts
abhipatel12: When we process the prompt, if it finds an empty shell injection ( `!{}` ). When it does this it saves an empty string for the `injection.command`. Then when the processing for the args happens it sets the command to the proper resolved command. In cases where this doesn't happen like an empty shell injection it returns just the command.

packages/core/src/services/shellExecutionService.ts
abhipatel12: Done

packages/core/src/utils/shell-utils.ts
abhipatel12: Done

packages/core/src/services/shellExecutionService.test.ts
abhipatel12: I added  `integration-tests/shell-service.test.ts`, let me know if that looks okay!

abhipatel12:
Thanks! lmk what you think!

docs/cli/commands.md
```diff
@@ -261,7 +290,7 @@ Please generate a Conventional Commit message based on the following git diff:
 
 ```diff
 !{git diff --staged}
-```
+````
```
jacob314: nit: seems like this got an extra `

docs/cli/commands.md
```diff
@@ -282,7 +311,7 @@ First, ensure the user commands directory exists, then create a `refactor` subdi
 ```bash
 mkdir -p ~/.gemini/commands/refactor
 touch ~/.gemini/commands/refactor/pure.toml
-```
+````
```
jacob314: seems like this got an extra `

packages/core/src/services/shellExecutionService.test.ts
```diff
@@ -42,20 +44,27 @@ describe('ShellExecutionService', () => {
...
-      onData: Mock;
-      onExit: Mock;
-    };
-    mockPtyProcess.pid = 12345;
-    mockPtyProcess.kill = vi.fn();
-    mockPtyProcess.onData = vi.fn();
-    mockPtyProcess.onExit = vi.fn();
+    mockChildProcess = new EventEmitter() as EventEmitter &
+      Partial<ChildProcess>;
+    // FIX: Cast simple EventEmitters to the expected stream type.
```
jacob314: nit. remove FIX: comments.

Bonus points if Gemini CLI can fix this up to not use defineProperty. Seems like that was a patch for it struggling to mock

packages/core/src/services/shellExecutionService.test.ts
```diff
@@ -183,15 +192,90 @@ describe('ShellExecutionService', () => {
...
 
   describe('Aborting Commands', () => {
-    it('should abort a running process and set the aborted flag', async () => {
-      const { result } = await simulateExecution(
-        'sleep 10',
-        (pty, abortController) => {
-          abortController.abort();
-          pty.onExit.mock.calls[0][0]({ exitCode: 1, signal: null });
-        },
+    describe.each([
```
jacob314: these are fine but probably a bit into integration test territory where as a follow up geting Gemini CLI to refactor as a true integraiton test with reall sigterm and taskkill would be more robust.

packages/core/src/utils/shell-utils.ts
```diff
@@ -5,6 +5,102 @@
...
+/**
+ * Determines the appropriate shell configuration for the current platform.
+ *
+ * This ensures we can execute command strings predictably and securely across platforms
+ * using the `spawn(executable, [...argsPrefix, commandString], { shell: false })` pattern.
+ *
+ * @returns The ShellConfiguration for the current environment.
+ */
+export function getShellConfiguration(): ShellConfiguration {
+  if (os.platform() === 'win32') {
```
jacob314: nit: use isWindows const defined bellow.

integration-tests/shell-service.test.ts
```diff
@@ -0,0 +1,122 @@
...
+
+      expect(result.error).toBeNull();
+      expect(result.exitCode).toBe(0);
+      expect(result.output).toContain(testFile);
+    },
+  );
+
+  it('should abort a running process', async () => {
+    // A command that runs for a bit. 'sleep' on unix, 'timeout' on windows.
+    const command = process.platform === 'win32' ? 'timeout /t 5' : 'sleep 5';
```
jacob314: perhaps make 5 slightly larger to reduce risk of flaking.

integration-tests/shell-service.test.ts
```diff
@@ -0,0 +1,122 @@
...
+
+    const handle = ShellExecutionService.execute(
+      command,
+      testDir,
+      onOutputEvent,
+      abortController.signal,
+    );
+
+    // Abort shortly after starting
+    setTimeout(() => abortController.abort(), 100);
```
jacob314: can 100 be reduced? every ms saved running tests is nice.

jacob314 (approved):
lgtm

docs/cli/commands.md
abhipatel12: Since we're using markdown within the markdown, we need to use more backticks than inside. So in this case ```` would be correct. `npm run format` will also add this back if i try to remove the extra backtick.

docs/cli/commands.md
abhipatel12: Since we're using markdown within the markdown, we need to use more backticks than inside. So in this case ```` would be correct. `npm run format` will also add this back if i try to remove the extra backtick.

packages/core/src/services/shellExecutionService.test.ts
abhipatel12: Removed the comments. I left the define property for now since these tests will likely be restructured once gal's pr goes in

packages/core/src/services/shellExecutionService.test.ts
abhipatel12: SG!

packages/core/src/utils/shell-utils.ts
abhipatel12: done

integration-tests/shell-service.test.ts
abhipatel12: Done

integration-tests/shell-service.test.ts
abhipatel12: Done

docs/cli/commands.md
```diff
@@ -169,23 +169,52 @@ Your command definition files must be written in the TOML format and use the `.t
...
-description = "Generates a fix for a given GitHub issue."
-prompt = "Please analyze the staged git changes and provide a code fix for the issue described here: {{args}}."
+description = "Generates a fix for a given issue."
+prompt = "Please provide a code fix for the issue described here: {{args}}."
 ```
 
-The model will receive the final prompt: `Please analyze the staged git changes and provide a code fix for the issue described here: "Button is misaligned on mobile".`
+The model receives: `Please provide a code fix for the issue described here: "Button is misaligned".`
+
+**B. Shell-Safe Injection (Inside `!{...}` Blocks)**
```
NTaylorMullen: I feel like the safety aspect of this is a detail to ensure good practice. From a user perspective this is "you can use args in shell injections"

packages/cli/src/services/prompt-processors/shellProcessor.ts
```diff
@@ -23,84 +29,212 @@ export class ConfirmationRequiredError extends Error {
...
+        userArgsRaw,
       );
+
+      // Execute the resolved command (which already has ESCAPED input).
+      if (injection.resolvedCommand) {
+        const { result } = ShellExecutionService.execute(
+          injection.resolvedCommand,
+          config.getTargetDir(),
+          () => {},
+          new AbortController().signal,
```
NTaylorMullen, line 139: Depending on when this PR lands in conjunction with @akhil29's [PR](https://github.com/google-gemini/gemini-cli/pull/6337) you may want to change the abort signal

packages/cli/src/services/prompt-processors/shellProcessor.ts
```diff
@@ -23,84 +29,212 @@ export class ConfirmationRequiredError extends Error {
...
+
+        // Handle Spawn Errors
+        if (executionResult.error && !executionResult.aborted) {
+          throw new Error(
+            `Failed to start shell command in '${this.commandName}': ${executionResult.error.message}. Command: ${injection.resolvedCommand}`,
+          );
+        }
+
+        // Append the output and status.
+        processedPrompt += executionResult.output;
```
NTaylorMullen: How does it know which shell invocation failed?

packages/core/src/utils/shell-utils.ts
```diff
@@ -5,6 +5,102 @@
...
+      executable.endsWith('powershell.exe') ||
+      executable.endsWith('pwsh.exe')
+    ) {
+      // For PowerShell, the arguments are different.
+      // -NoProfile: Speeds up startup.
+      // -Command: Executes the following command.
+      return {
+        executable: comSpec,
+        argsPrefix: ['-NoProfile', '-Command'],
+        shell: 'powershell',
```
NTaylorMullen, line 53: Do we utilize powershell in other places? One challenge is that by default in powershell if the argument happens to be another powershell script you'll end up hitting [execution-poclity ](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_execution_policies?view=powershell-7.5) errors. I'm not sure how these would surface in our UI (maybe they'd be ok?)

NTaylorMullen (approved):
Woot woot, excited to see this land. My custom commands are going to be super powered with this ❤️

packages/cli/src/services/prompt-processors/shellProcessor.ts
jacob314: Gemini Code Assist... you are not the PR author :)

docs/cli/commands.md
abhipatel12: True, made the header more focused on the functionality, but left the explanation similar so users know that the arg is shell escaped.

packages/cli/src/services/prompt-processors/shellProcessor.ts
abhipatel12: Haha it seems that gemini's got it. 

But yes, im reviewing that PR as well, so ill be sure to mention needing this changed as well.

packages/cli/src/services/prompt-processors/shellProcessor.ts
abhipatel12: That's a great point. I made it more explicit now.

packages/core/src/utils/shell-utils.ts
abhipatel12: I did a deeper dive into this. The `ShellExecutionService` currently uses `spawn({ shell: true })` on Windows. So that means `argsPrefix` in `shell-utils.ts` should be ignored in the service today. 

But I think we can have a broader discussion/enhancement to use powershell on windows. I can create a fast follow issue to track this. Made an issue here #6413

abhipatel12:
Awesome, thanks @NTaylorMullen and @jacob314 for the review on this!

PR #5967, opened 2025-08-11
fix(tools): Add an end of file list marker to ReadManyFilesTool

vmiura:
## TLDR\

The ReadManyFilesTool often caused the model to echo back file contents or to hallucinate extra file contents. This change adds a marker at the end of the file parts, that in my testing seems to resolve the issue.

## Dive Deeper

In other contexts such as @Path with a directory we were already inserting an `--- End of content ---` marker. This change adds the same marker at the end of all ReadManyFilesTool results. Prior to this change I was able to reproduce Issue #5604 with a high regularity. With the fix, I was able to test 1000s of file reads without seeing the issue.

## Reviewer Test Plan

The issue could be reproduced by running something like "Read all files in the repository in batches of 50" inside the gemini-cli repository, or any other large project.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ✅  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

- Fixes #5604


packages/core/src/tools/read-many-files.ts
```diff
@@ -600,7 +602,9 @@ Use this tool when the user's query implies needing the content of several files
       displayMessage += `No files were read and concatenated based on the criteria.\n`;
     }
 
-    if (contentParts.length === 0) {
+    if (contentParts.length > 0) {
```
galdawave, line 531: Shouldn't this be inside the for loop where you are adding DEFAULT_OUTPUT_SEPARATOR_FORMAT , how the atCommandProcessor works is it adds both when it's including content from a specific file, what this would be doing is 

```
--- File1 --- 
content 
--- File2 ---
content
--- End Of Content ---
```

so in theory the model can hallucinate , i think it should be like this 

```
--- File1 --- 
content 
--- End Of File1 Content ---
--- File2 ---
content
--- End Of File2 Content ---
```

i don't know if filepath in `--- End Of {filepath} Content ---` is needed since i don't think it's not included in atCommandProcessor

packages/cli/src/ui/hooks/atCommandProcessor.ts
```diff
@@ -450,7 +450,6 @@ export async function handleAtCommand({
           processedQueryParts.push(part);
         }
       }
-      processedQueryParts.push({ text: '\n--- End of content ---' });
```
galdawave, line 450: I don't think you want to remove this, since `-- Content from referenced files ---` and content of the file is being added

packages/cli/src/acp/acpPeer.ts
```diff
@@ -570,7 +570,6 @@ class GeminiAgent implements Agent {
             processedQueryParts.push(part);
           }
         }
-        processedQueryParts.push({ text: '\n--- End of content ---' });
```
galdawave: I don't think you want to remove this, since -- Content from referenced files --- and content of the file is being added

galdawave (changes requested):

packages/core/src/tools/read-many-files.ts
vmiura: The model doesn't seem to hallucinate with the pattern:

```
--- File1 --- 
content 
--- File2 ---
content
--- End Of Content ---
```

I have not tested your suggested pattern but it doesn't seem to be necessary to make the bigger change.

packages/cli/src/ui/hooks/atCommandProcessor.ts
vmiura: `\n--- End of content ---' is now being added by the ReadManyFilesTool, so the net result is the same. It is verified by many tests.

packages/cli/src/acp/acpPeer.ts
vmiura: As above, `\n--- End of content ---' is now being added by the ReadManyFilesTool.

I think having ReadManyFilesTool add this is more consistent than having it in multiple places.

packages/cli/src/ui/hooks/atCommandProcessor.ts
vmiura: For extra context:

packages/cli/src/ui/hooks/atCommandProcessor.test.ts validates that '\n--- End of content ---' is added to results in 21 tests.

The reason to remove this line is to preserve the existing expectations.

vmiura:
@galz10 

Could we land the fix that uses this, matches the existing @path implementation, and seems to resolve the hallucinations:

```
--- File1 --- 
content 
--- File2 ---
content
--- End Of Content ---
```

If we still see issues, we can evaluate if adding extra delimiters like your proposal helps further, e.g.:

```
--- File1 --- 
content 
--- End Of File1 Content ---
--- File2 ---
content
--- End Of File2 Content ---
--- End Of Content ---
```

galdawave (approved):

vmiura:
It looks like the merge failed on one of the targets. Is this a flake, or do I need to do anything?

allenhutchison:
Seems like that target is consistently failing for this PR. There are also some merge conflicts. Can you take a look at the merge conflicts and I'll take a look at the failing test after?

vmiura:
Merge is resolved.

allenhutchison:
Tests are good, but there is a new conflict on read-many-files.ts Can you take a look?

vmiura:
Merges fixed.

allenhutchison:
I've run our offline evals and this shows even or slightly higher performance compared to the baseline. 

2025-08-21 | 488 | 263 | 53.89%

allenhutchison (approved):

PR #5969, opened 2025-08-11
fix(#5340): unable to load chats with weird characters

### LINKED ISSUE #5340: Unable to load chats with Japanese characters in the name
### What happened?

When I resume a tag that exists in the Chat list, I get “No saved checkpoint found with tag: ~~” and cannot resume the conversation.

This event occurred at the time I raised it from v0.1.13 to v0.1.14. This event is still occurring in v0.1.15.

https://github.com/user-attachments/assets/820c9903-0fc6-4595-86d8-9fbc4633c9a7

I expected that I could resume chatting.

### Client information

<details>

```console
$ gemini /about
# paste output here

About Gemini CLI                                                                                                  │
│                                                                                                                   │
│ CLI Version                             0.1.15                                                                    │
│ Git Commit                              c45c14ee (local modifications)                                            │
│ Model                                   gemini-2.5-pro                                                            │
│ Sandbox                                 no sandbox                                                                │
│ OS                                      darwin                                                                    │
│ Auth Method                             OAuth          
```

</details>

### Login information

_No response_

_No response_

hugomurillomtz:
## TLDR

It fixes the issue about weird characters causing issues when attempting to resume chats.

## Dive Deeper

For tags which names include characters different to **A-Za-z-_0-9** like some of these:
image

when attempting to resume/delete them, the following error was happening:
image

Interestingly, checking the code I noticed that we were not allowing characters different to **A-Za-z-_0-9** to be used as tag names for saving chats. When attempting to doing so, like this `ginひらがな`, the code would filter out the characters and leave `gin` only in this example. However, users report that they have saved chats using weird characters (I wonder if they were allowed at a certain time), but that the problem is that, when attempting to resume/delete them, they get an exception saying that `no checkpoint was found`, and that was happening because the sanitization or filtering was being done when retrieving the file too. So, the solution this PRs implements is as follows:

- For tags with weird characters, it decodes the tag name to handle URL-encoded characters like `%2F`. For an example like `%2F`, the code would convert it into a forward slash: `/`.
- It normalizes both `\` and `/` to check for path traversal. This is because, on Windows, the path separator is a backslash \, while on Linux and macOS is a forward slash.
- It checks the basename of the file to ensure that it doesn't contain any malicious characters. And if unsafe, it removes all characters from the tag name different to **A-Za-z-_0-9**.

## Reviewer Test Plan

I already have some tags saved in advanced. Note that some of them have 'weird' characters like Chinese or Japanese letters:
image

I will now resume the first one with Chinese characters:
image
image

And it's a success:
image

**I can also delete:**
image
image

And it's a success too:
image
image

**I can save too:**
image
image

And it's a success:
image
image

And if attempting to perform a traversal attack:
image
image

For testing purposes, I am printing the error on the screen:
image

As seen, it detected a traversal attack. So, after applying the regex, it ended up with: `etcpasswd`, and couldn't find it. What users will actually see is this:
image

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Resolves: https://github.com/google-gemini/gemini-cli/issues/5340


packages/core/src/core/logger.ts
```diff
@@ -231,19 +231,30 @@ export class Logger {
...
+      // If it's unsafe, perform aggressive sanitization on the original tag.
+      const sanitizedTag = tag.replace(/[^a-zA-Z0-9-_]/g, '');
+      const finalTag = sanitizedTag || 'default';
+      return path.join(this.geminiDir, `checkpoint-${finalTag}.json`);
     }
-    return path.join(this.geminiDir, `checkpoint-${sanitizedTag}.json`);
+
+    // If we are here, the tag is safe from traversal.
+    const finalTag = decodedTag || 'default';
+    return path.join(this.geminiDir, `checkpoint-${finalTag}.json`);
```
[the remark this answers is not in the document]
jacob314: agree a try catch block is reasonable here. 
call console.error in the catch block to report this unlikely error in a way that will be visible to users.

jacob314 (commented):
This looks good. Approved as soon as this concern is addressed. Agree using an escaping scheme that avoids security issues i the way to go here.

packages/core/src/core/logger.ts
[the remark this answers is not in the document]
hugomurillomtz: @gemini-code-assist how could I test this?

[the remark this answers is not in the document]
hugomurillomtz: @gemini-code-assist when I do "test%E0%A4%A", the exception seems to be handled correctly. It's like it's happening inside the `decodeURIComponent`.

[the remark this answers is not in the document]
hugomurillomtz: @gemini-code-assist I mean, without the try/catch block is being handled correctly. Why?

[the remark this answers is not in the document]
hugomurillomtz: @gemini-code-assist tha app continues to run correctly, and an error message shows saying: "URI malformed"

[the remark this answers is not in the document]
hugomurillomtz: Done!

packages/core/src/core/logger.ts
```diff
@@ -231,19 +231,35 @@ export class Logger {
     }
   }
 
-  _checkpointPath(tag: string): string {
+  async _checkpointPath(tag: string): Promise<string> {
```
jacob314: why is this async? plesae fix unless I'm missing something async that you need to do.

packages/core/src/core/logger.test.ts
```diff
@@ -611,6 +611,70 @@ describe('Logger', () => {
...
+        'Error when attempting to form tag file name',
+      );
+    });
+
+    it('should handle null bytes', async () => {
+      const nullByteTag = 'test%00attack'; // Null byte encoded
+      const decodedTag = decodeURIComponent(nullByteTag);
+      const expectedPath = path.join(
+        TEST_GEMINI_DIR,
+        `checkpoint-${decodedTag}.json`,
```
jacob314: this is testing the reverse of what you want. you want the file names to be safely encoded and the tags not to be encoded.

packages/core/src/core/logger.ts
```diff
@@ -231,19 +231,35 @@ export class Logger {
...
-    // Sanitize tag to prevent directory traversal attacks
-    let sanitizedTag = tag.replace(/[^a-zA-Z0-9-_]/g, '');
-    if (!sanitizedTag) {
-      sanitizedTag = 'default';
+
+    let finalTag: string;
+    try {
+      // Decode the tag to handle URL-encoded characters like %2F, which could
+      // be used to disguise path traversal attempts.
+      const decodedTag = decodeURIComponent(tag);
```
jacob314: agree with the general idea but don't think this implementation is what you want. you should be encoding the tags safely. not sure if decodeURIComponent is what you really want at all.  I would ask Gemini CLI what the best encoding scheme to use to encode a string as something that is safe for a file path. It should be able to suggest an existing solution rather than us implementing our own using decodeURIComponent along with some odd special checks for `\` and `/` to check for path traversal.

packages/core/src/core/logger.ts
hugomurillomtz: You're right. I refactored it so it's no longer needed.

packages/core/src/core/logger.ts
hugomurillomtz: It was actually Gemini CLI that suggested it. By decoding the tag name, we can turn things like `%2F` into `/` in case someone is disguising the tag name. My very first test is about that. Gemini CLI also suggested the odd special checks for \ and /. @gemini-code-assist using `decodeURIComponent` to check for characters like %2F used for traversal attacks is okay to do, right?

Also, normalizing `\` and _/_ slashes is okay to check for traversal attacks considering that users OS can be either Linux, Windows and Mac, right?

packages/core/src/core/logger.test.ts
hugomurillomtz: By decoding the tag name we are trying to see if users are disguising characters like / or \ by using things like %2F. For example, say a user types for a tag name `..%2F..%2F..%2F..%2Fetc%2Fpasswd`, our code would decode it and get `../../../../etc/passwd`, meaning that they are attempting to perform a traversal attack.

packages/core/src/core/logger.ts
hugomurillomtz: @gemini-code-assist Give me an example of a traversal attack using URL-encoded characters that I can use in this case to test my application with.

hugomurillomtz: @gemini-code-assist so I should simply test my app by doing: `/chat save ..%2F..%2F..%2Fetc%2Fpasswd`?

jacob314: please encode rather than decode. all you need to do is ensure you encode all unsafe characters. not sure why gemini cli is so confused. just make sure slashes and dots that are unsafe are all encoded

hugomurillomtz: I'm almost done.

jacob314: thanks. please let me know when this is ready to review. would be great to get this landed.

hugomurillomtz: You may review it whenever you can. Meanwhile, I'll check the tasks list to pick something else.

https://github.com/google-gemini/gemini-cli/issues?q=is%3Aissue%20state%3Aopen%20label%3Aarea%2Fcore%2Carea%2Fux%20label%3A%22good%20first%20issue%22%20label%3Apriority%2Fp1

jacob314:
This works but leaves the chat file names unreadable by humans as they are base64 encoded.
For example: saving a checkpoint with the name `foo` becomes `checkpoint-Zm9v.json`
Please use an encoding scheme so regular ascii characters are still readable. if other characters are not readable that is fie.

hugomurillomtz:
> This works but leaves the chat file names unreadable by humans as they are base64 encoded. For example: saving a checkpoint with the name `foo` becomes `checkpoint-Zm9v.json` Please use an encoding scheme so regular ascii characters are still readable. if other characters are not readable that is fie.

I think it's fine now. If I do:

`/chat save chinese_汉字`
image

The filename gets stored like this:
image

And then, if I perform operations on it like resume or delete, I see the original name:
image

And I'm still able to resume old tags like this one:
image
image

LKM if you'd like to see more tests.

packages/cli/src/ui/commands/chatCommand.ts
```diff
@@ -18,6 +18,26 @@ import {
...
+/**
+ * Decodes a string that was encoded with the `encode` function.
+ *
+ * It finds any percent-encoded characters and converts them back to their
+ * original representation.
+ *
+ * @param str The encoded string to decode.
+ * @returns The decoded, original string.
+ */
+export function decode(str: string): string {
```
jacob314: nit: decode is too general a name. call this decodeTagName or similar.

packages/core/src/core/logger.ts
```diff
@@ -23,6 +23,22 @@ export interface LogEntry {
...
+/**
+ * Encodes a string to be safe for use as a filename.
+ *
+ * It replaces any characters that are not alphanumeric or one of `_`, `-`, `.`
+ * with a URL-like percent-encoding (`%` followed by the 2-digit hex code).
+ *
+ * @param str The input string to encode.
+ * @returns The encoded, filename-safe string.
+ */
+export function encode(str: string): string {
```
jacob314: encode is too general a name. Also place encode and the matching decode in the same file. decode is in chatCommand and encode is here.

packages/core/src/core/logger.test.ts
```diff
@@ -394,15 +394,27 @@ describe('Logger', () => {
...
 
     it.each([
-      { tag: 'test-tag', sanitizedTag: 'test-tag' },
-      { tag: 'invalid/?*!', sanitizedTag: 'invalid' },
-      { tag: '/?*!', sanitizedTag: 'default' },
-      { tag: '../../secret', sanitizedTag: 'secret' },
-    ])('should save a checkpoint', async ({ tag, sanitizedTag }) => {
+      {
+        tag: 'test-tag',
+        encodedTag: encode('test-tag'),
```
jacob314: do not call encode here. Include what it encodes to. otherwise someone could break your encode method and your tests would still pass.

packages/core/src/core/logger.test.ts
```diff
@@ -527,6 +553,26 @@ describe('Logger', () => {
...
+
+      // Create both files
+      await fs.writeFile(oldStylePath, '{}');
+      await fs.writeFile(newStylePath, '{}');
+
+      const result = await logger.deleteCheckpoint(oldTag);
+      expect(result).toBe(true);
+
+      // Verify both are gone
+      await expect(fs.access(oldStylePath)).rejects.toThrow(/ENOENT/);
```
jacob314: nit: simpler to check if the files exists then than that reading them throws

packages/core/src/core/logger.test.ts
```diff
@@ -394,15 +394,27 @@ describe('Logger', () => {
     ];
 
     it.each([
```
jacob314, line 402: this is the basis for a good test that verifies that  the encoded tag for each of these examples is what you expect. Please add that as a separate test or extend this test to also expect that the encodedTag passed in also matches the result of calling encode(tag)

packages/core/src/core/logger.test.ts
hugomurillomtz: It throws an exception if I do that:

image

I'll see if there's another way. Everything else is fixed now.

packages/core/src/core/logger.test.ts
hugomurillomtz: I extended this test, and I also added another assertion for the decode part.

packages/core/src/core/logger.test.ts
hugomurillomtz: Done!

jacob314 (approved):
lgtm

PR #6006, opened 2025-08-11
Reduce noisy IDE integration error message in standalone terminal

skeshive:
Removes error log from console when IDE mode is started up in non-supported terminal (/ide status still shows error)


jacob314 (approved):
lgtm

PR #6013, opened 2025-08-11
chore(usage telemetry): Freshen up Clearcut logging

richieforeman:
## TLDR

- Adds even stronger typing to things
- Adds MSW as a mocking library
- Switches to Clearcut's non-binary endpoint that receives/sends one shot JSON data.
- Use less mocks, wire things up with fake libraries (e.g. MSW) or use the actual things.

## Reviewer Test Plan

- None of the event payloads actually changed -- only the transit mechanism.  I find this code quite a bit more simple to grok.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | x  | ❓  | x  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Not attached to a bug, but this felt like some technical debt, and I worked on it on Friday mostly.


packages/core/src/config/config.ts
```diff
@@ -170,7 +170,9 @@ export interface ConfigParameters {
   contextFileName?: string | string[];
   accessibility?: AccessibilitySettings;
   telemetry?: TelemetrySettings;
+  /** if true, then usage statistics will be sent back to Google. */
```
jacob314: revert. low value comment.

packages/core/src/config/config.ts
```diff
@@ -263,6 +265,7 @@ export class Config {
   private readonly chatCompression: ChatCompressionSettings | undefined;
   private readonly interactive: boolean;
   private initialized: boolean = false;
+  private readonly logInitialSessionEventEnabled: boolean;
```
jacob314: why is having logInitialSessionEventEnabled needed?

packages/core/src/telemetry/clearcut-logger/clearcut-logger.test.ts
```diff
@@ -4,203 +4,249 @@
...
+});
+
+afterAll(() => {
+  server.close();
+});
+
 describe('ClearcutLogger', () => {
-  let mockConfig: Config;
-  let logger: ClearcutLogger | undefined;
+  const NEXT_WAIT_MS = 9001;
```
jacob314: why 9001. Surprising choice for a magic number. Comment or make something less seemingly precise.

jacob314 (commented):
Thanks for cleaning this up a bit. Approved after these comments are addressed

packages/core/src/config/config.ts
richieforeman: I ran into snags in the tests where using a real instance of `Config()` (previously, the tests used mocks). There's some code in `Config()` that was triggering events that I needed to disable.  I wanted to avoid adding an unnecessary `clearEvents` method or otherwise.

packages/core/src/telemetry/clearcut-logger/clearcut-logger.test.ts
richieforeman: I was referencing a meme :)

I've updated it to `1234` to be more obvious.  Noted for future.  I generally use it as my bogus number, but I can see how 9001 would trip folks up.

packages/core/src/config/config.ts
richieforeman: ... Actually, apologies, I just found a mechanism that allowed me to remove this whole adventure. It was not needed.

jacob314 (approved):
lgtm

owenofbrien (approved):
Thanks for the cleanup!

PR #6039, opened 2025-08-12
fix(core): avoid error handling on cancelled requests to prevent crash

### LINKED ISSUE #5935: [API Error: Cannot read properties of undefined (reading 'error')]
### What happened?

first start 

first start gemini cli,fisrt ask

### Client information

* **CLI Version:** 0.1.18
* **Git Commit:** cd7e60e0
* **Operating System:** darwin v20.19.2
* **Memory Usage:** 105.9 MB

### Login information

_No response_

_No response_

pwrwpw:
## TLDR

Fix crash caused by accessing error properties on cancelled requests.
Checks for request cancellation early and gracefully handles it to prevent runtime errors.
## Dive Deeper

When a user cancels a request (e.g., by pressing ESC), the error handling code attempts to read properties from an undefined error object, causing a crash.
This change moves the cancellation check before any error processing, so if the request is aborted, the code emits a cancellation event and exits cleanly without throwing exceptions or processing undefined errors.

## Reviewer Test Plan

- Send requests and cancel them using ESC abort signals.

- Verify that no unnecessary exception handling occurs upon cancellation.

- Verify that the UserCancelled event is emitted appropriately on cancellation.

- Ensure that normal errors (e.g., UnauthorizedError) are still handled correctly.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

 Fixes #5935


jacob314 (approved):
lgtm
Thanks for the pr. If you can follow up with a adding a quick test to ensure we don't regress that would be helpful to make sure we don't regress and reintroduce a bug.

PR #6046, opened 2025-08-12
Log CLI version and git commit hash

owenofbrien:
## TLDR

Added logging for CLI version and git commit hash

## Dive Deeper

Logging these values to help us make better sense of the metrics.

## Reviewer Test Plan

Auto tests should suffice for this.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | y  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

https://github.com/google-gemini/maintainers-gemini-cli/issues/649


packages/cli/src/config/config.ts
```diff
@@ -411,6 +412,8 @@ export async function loadCliConfig(
 
   return new Config({
     sessionId,
+    cliVersion: await getCliVersion(),
```
anj-s, line 415: Why does this need to be part of the config? Can we call getCliVersion() when we need to log as well as the Git commit hash. The reason I say this is because this is not needed anywhere else in the codebase.

owenofbrien: That was my first thought as well, but then I realized getCliVersion() is in the cli package, and we shouldn't add a core -> cli package dependency.

owenofbrien:
I ended up taking a different approach: https://github.com/google-gemini/gemini-cli/pull/6176

PR #2454, opened 2025-06-28
feat: add a way to queue messages while model is thinking

asadm:
## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| -------- | --- | --- | --- |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

PR #6049, opened 2025-08-12
feat(ui): implement message queuing during streaming responses

### LINKED ISSUE #3594: Should be able to type while Gemini is thinking

Should be able to type next question/ask while Gemini is replying to the current question.
The same as other AI code assistants.

### Why is this needed?

Because I don't want to waste time.

### Additional context

_No response_

akhil29:
## Summary
- Implements message queuing functionality that allows users to type and queue messages while the AI is streaming responses
- Queued messages are displayed below the loading indicator with a dimmed color (`▸` prefix)
- Messages are automatically sent when streaming completes

Fixes #3594

## Changes
- Added `messageQueue` state to buffer user input during streaming
- Modified `handleFinalSubmit` to always queue messages for unified flow
- Implemented auto-send logic with `useEffect` when streaming transitions to idle
- Integrated queue handling with cancel behavior to append queued messages
- Added visual display of queued messages in the UI
- Updated tests and snapshots
- Removed fragile `setTimeout` approach based on code review feedback

## Test Plan
- [x] Type messages while AI is responding - they should appear queued below loading indicator
- [x] When streaming completes, queued messages should auto-send
- [x] Cancel during streaming - queued messages should be appended to input
- [x] Multiple messages can be queued and are sent combined with double newlines
- [x] Empty messages are filtered out
- [x] All existing tests pass
- [x] Added comprehensive automated tests for the feature

### DIFF
--- a/packages/cli/src/ui/App.tsx
+++ b/packages/cli/src/ui/App.tsx
@@ -544,6 +544,9 @@ const App = ({ config, settings, startupWarnings = [], version }: AppProps) => {
     shellModeActive,
   });
 
+  // Simple message queue for handling input during streaming
+  const [messageQueue, setMessageQueue] = useState<string[]>([]);
+
   const [userMessages, setUserMessages] = useState<string[]>([]);
 
   const handleUserCancel = useCallback(() => {
@@ -577,15 +580,20 @@ const App = ({ config, settings, startupWarnings = [], version }: AppProps) => {
     handleUserCancel,
   );
 
-  // Input handling
+  // Input handling - with message queue support
   const handleFinalSubmit = useCallback(
     (submittedValue: string) => {
       const trimmedValue = submittedValue.trim();
       if (trimmedValue.length > 0) {
-        submitQuery(trimmedValue);
+        // Queue message if streaming is active, otherwise submit directly
+        if (streamingState === StreamingState.Responding) {
+          setMessageQueue((prev) => [...prev, trimmedValue]);
+        } else {
+          submitQuery(trimmedValue);
+        }
       }
     },
-    [submitQuery],
+    [submitQuery, streamingState],
   );
 
   const handleIdePromptComplete = useCallback(
@@ -1083,6 +1091,17 @@ const App = ({ config, settings, startupWarnings = [], version }: AppProps) => {
                 }
                 elapsedTime={elapsedTime}
               />
+
+              {/* Display queued messages below loading indicator */}
+              {messageQueue.length > 0 && (
+                <Box flexDirection="column" marginTop={1}>
+                  {messageQueue.map((message, index) => (
+                    <Box key={index} paddingLeft={2}>
+                      <Text dimColor>▸ {message}</Text>
+                    </Box>
+                  ))}
+                </Box>
+              )}
 
               <Box
                 marginTop={1}
--- a/packages/cli/src/ui/App.tsx
+++ b/packages/cli/src/ui/App.tsx
@@ -761,7 +761,10 @@ const App = ({ config, settings, startupWarnings = [], version }: AppProps) => {
   }, [history, logger]);
 
   const isInputActive =
-    streamingState === StreamingState.Idle && !initError && !isProcessing;
+    (streamingState === StreamingState.Idle ||
+      streamingState === StreamingState.Responding) &&
+    !initError &&
+    !isProcessing;
 
   const handleClearScreen = useCallback(() => {
     clearItems();
--- a/packages/cli/src/ui/App.tsx
+++ b/packages/cli/src/ui/App.tsx
@@ -103,6 +103,8 @@ import { appEvents, AppEvent } from '../utils/events.js';
 import { isNarrowWidth } from './utils/isNarrowWidth.js';
 
 const CTRL_EXIT_PROMPT_DURATION_MS = 1000;
+// Maximum number of queued messages to display in UI to prevent performance issues
+const MAX_DISPLAYED_QUEUED_MESSAGES = 3;
 
 interface AppProps {
   config: Config;
@@ -544,8 +546,6 @@ const App = ({ config, settings, startupWarnings = [], version }: AppProps) => {
     isValidPath,
     shellModeActive,
   });
-
-  // Message queue managed by dedicated hook
 
   const [userMessages, setUserMessages] = useState<string[]>([]);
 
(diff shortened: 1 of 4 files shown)


akhil29:
Testing:
Screenshot 2025-08-12 at 7 00 07 AM
Screenshot 2025-08-12 at 7 00 22 AM

cornmander:
Thank you for the PR. I think this is a good idea but I would like to see it implemented in core rather than in the TUI package. That way, we can build upon this concept for sub-agents. Sub-agents will be able to enqueue messages to the main agent, even if it's processing a user message or a message from another agent.

abhipatel12:
There was a previous PR,  #3712, that does something similar, that PR was closed due to inactivity. We can continue this in this PR. 

A few things: 
1.  Please add an issue to this PR. I believe #3594 would work for this!
2. Can a user cancel a request once its been added? What if the model responds in a way where the queued message is no longer relevant?
3. Does a large number of queued requests cause an increase of flashing? Should we only show one queued message at a time? Can we make it collapsable? 

Also going to add @miguelsolorio here who is the expert in UI/UX for some thoughts!

akhil29:
> There was a previous PR, #3712, that does something similar, that PR was closed due to inactivity. We can continue this in this PR.
> 
> A few things:
> 
> 1. Please add an issue to this PR. I believe [Should be able to type while Gemini is thinking #3594](https://github.com/google-gemini/gemini-cli/issues/3594) would work for this!
> 2. Can a user cancel a request once its been added? What if the model responds in a way where the queued message is no longer relevant?
> 3. Does a large number of queued requests cause an increase of flashing? Should we only show one queued message at a time? Can we make it collapsable?
> 
> Also going to add @miguelsolorio here who is the expert in UI/UX for some thoughts!

1. Done
2. Currently when the model is responding and user cancels it, the canceled prompt gets added to the input box and the request is cancelled. Same thing happens now as well but all the queued prompts will be added to the input box as well.
Screenshot 2025-08-12 at 8 05 32 AM
3. Each additional message is added to the queue but while sending to the prompt, they are all merged with double new line "\n\n". This is the standard behavior across other more mature cli coding agents.

itsparth:
There are multiple PR's for the same https://github.com/google-gemini/gemini-cli/pull/2454

akhil29:
What is the state here? @abhipatel12

packages/cli/src/ui/App.tsx
```diff
@@ -1081,6 +1105,17 @@ const App = ({ config, settings, startupWarnings = [], version }: AppProps) => {
                 elapsedTime={elapsedTime}
               />
 
+              {/* Display queued messages below loading indicator */}
+              {messageQueue.length > 0 && (
+                <Box flexDirection="column" marginTop={1}>
+                  {messageQueue.map((message, index) => (
```
abhipatel12: The current impl renders the full message for every queued message. This works with small inputs, but if a user provides a large input as a queued message, this can be difficult to view the edits/reads/thoughts, etc. But also, this creates performance issues due to how our UI works. This may cause flickering and scroll issues, especially for large inputs.

We should limit the number of messages displayed and ensure each message only takes up a single line in the preview.

We don't have to use this exact impl below, but a possible impl provided via gemini:

```typescript
              {/* Display queued messages below loading indicator */}
              {messageQueue.length > 0 && (
                <Box flexDirection="column" marginTop={1}>
                  {messageQueue
                    .slice(0, MAX_DISPLAYED_QUEUED_MESSAGES)
                    .map((message, index) => {
                      // Ensure multi-line messages are collapsed for the preview.
                      // Replace all whitespace (including newlines) with a single space.
                      const preview = message.replace(/\s+/g, ' ');

                      return (
                        // Ensure the Box takes full width so truncation calculates correctly
                        <Box key={index} paddingLeft={2} width="100%">
                          {/* Use wrap="truncate" to ensure it fits the terminal width and doesn't wrap */}
                          <Text dimColor wrap="truncate">
                            ▸ {preview}
                          </Text>
                        </Box>
                      );
                    })}
                  {messageQueue.length > MAX_DISPLAYED_QUEUED_MESSAGES && (
                    <Box paddingLeft={2}>
                      <Text dimColor>
                        ... (+{messageQueue.length - MAX_DISPLAYED_QUEUED_MESSAGES} more)
                      </Text>
                    </Box>
                  )}
                </Box>
              )}
```

packages/cli/src/ui/App.tsx
```diff
@@ -544,14 +544,26 @@ const App = ({ config, settings, startupWarnings = [], version }: AppProps) => {
     shellModeActive,
   });
 
+  // Simple message queue for handling input during streaming
+  const [messageQueue, setMessageQueue] = useState<string[]>([]);
```
abhipatel12: The current `App.tsx` is already very long and often difficult to test.

Can you please extract the queue logic into its own dedicated hook? aka `useMessageQueue.ts`?

abhipatel12 (commented):
Thanks for the contribution! Took a look at this and pulled it locally to test it out. Overall, this is great. 

I made a couple of comments for some concerns that I had. Let me know what you think!

akhil29:
Testing multi line:
Screenshot 2025-08-18 at 4 25 17 PM

eslint
abhipatel12, line 1: I think this got added by accident.

packages/cli/src/ui/App.tsx
```diff
@@ -544,14 +547,12 @@ const App = ({ config, settings, startupWarnings = [], version }: AppProps) => {
     shellModeActive,
   });
 
+  // Message queue managed by dedicated hook
```
abhipatel12: We can delete this comment

packages/cli/src/ui/App.tsx
```diff
@@ -1081,6 +1106,39 @@ const App = ({ config, settings, startupWarnings = [], version }: AppProps) => {
                 elapsedTime={elapsedTime}
               />
 
+              {/* Display queued messages below loading indicator */}
+              {messageQueue.length > 0 && (
+                <Box flexDirection="column" marginTop={1}>
```
abhipatel12, line 1109: This looks great, would you be able to create a follow-up task that extracts the UI component into its own file? Something like `QueuedMessageDisplay.tsx` or something?

abhipatel12 (commented):
This looks great! Thanks for making all of the changes. 

Just a few cleanup comments! I'll pass this along to our ask reviewers now!

abhipatel12 (approved):

jacob314:
The message queueing is a bit surprising to me. I would expect each message I queued to run after the previous message had completed but looks more like the queued mesages are combined together with all the queued messages getting combined into one block rather than getting interleaved with responses from the LLM.
 Is the current behavior intentional? If the current behavior is intended it should be refined to show the pending messages as a single blocks.
 I'd suggest we instead tweak so pending messages get fired off one by one rather than grouped.
Screenshot 2025-08-18 at 7 12 12 PM
Screenshot 2025-08-18 at 7 16 09 PM

packages/cli/src/ui/App.tsx
```diff
@@ -1081,6 +1106,39 @@ const App = ({ config, settings, startupWarnings = [], version }: AppProps) => {
...
+                      // Ensure multi-line messages are collapsed for the preview.
+                      // Replace all whitespace (including newlines) with a single space.
+                      const preview = message.replace(/\s+/g, ' ');
+
+                      return (
+                        // Ensure the Box takes full width so truncation calculates correctly
+                        <Box key={index} paddingLeft={2} width="100%">
+                          {/* Use wrap="truncate" to ensure it fits the terminal width and doesn't wrap */}
+                          <Text dimColor wrap="truncate">
+                            ▸ {preview}
```
jacob314: lets remove the `▸`
or only show it on the first message of the queue. Otherwise it makes me think the queued messages are separate rather than combined together. approved after that is done.

akhil29:
Addressed all comments.

jacob314 (approved):
lgtm

PR #6063, opened 2025-08-12
Launch VS Code IDE Integration

skeshive:
## TLDR
* Updates VS Code installer to install extension from marketplace instead of from local build
* Removes feature flag (IDE integration will still be opt-in but users will see a hint nudging them to enable the integration)


mattKorwel (approved):
Lets go!

dtometzki:
Hello now isnt possible to install the extension via vsix file anymore. 

It is not possible to have both option ?

Damian

PR #6068, opened 2025-08-12
[ide-mode] Update installation logic and nudge

chrstnb:
## TLDR

1. For known non-vscode IDEs (this does not include e.g., Windsurf, which we cannot differentiate from VS Code), prompt the user to install the extension using the marketplace
2. For IDEs where the extension comes pre-installed, the initial nudge is now to enable IDE mode rather than to install the extension

## Dive Deeper
Screenshot 2025-08-12 at 1 42 36 PM

https://github.com/user-attachments/assets/904a0aa5-9604-416e-bf39-37a7adc4bb96

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

### DIFF
--- a/packages/cli/src/ui/App.tsx
+++ b/packages/cli/src/ui/App.tsx
@@ -61,6 +61,7 @@ import {
   AuthType,
   type IdeContext,
   ideContext,
+  getIdeInfo,
 } from '@google/gemini-cli-core';
 import {
   IdeIntegrationNudge,
@@ -577,7 +578,15 @@ const App = ({ config, settings, startupWarnings = [], version }: AppProps) => {
   const handleIdePromptComplete = useCallback(
     (result: IdeIntegrationNudgeResult) => {
       if (result === 'yes') {
-        handleSlashCommand('/ide install');
+        if (!currentIDE) {
+          return;
+        }
+        const ideInfo = getIdeInfo(currentIDE);
+        if (ideInfo.isExtensionInstalledByDefault) {
+          handleSlashCommand('/ide enable');
+        } else {
+          handleSlashCommand('/ide install');
+        }
         settings.setValue(
           SettingScope.User,
           'hasSeenIdeIntegrationNudge',
@@ -592,7 +601,7 @@ const App = ({ config, settings, startupWarnings = [], version }: AppProps) => {
       }
       setIdePromptAnswered(true);
     },
-    [handleSlashCommand, settings],
+    [handleSlashCommand, settings, currentIDE],
   );
 
   const { handleInput: vimHandleInput } = useVim(buffer, handleFinalSubmit);
--- a/packages/cli/src/ui/App.tsx
+++ b/packages/cli/src/ui/App.tsx
@@ -61,7 +61,6 @@ import {
   AuthType,
   type IdeContext,
   ideContext,
-  getIdeInfo,
 } from '@google/gemini-cli-core';
 import {
   IdeIntegrationNudge,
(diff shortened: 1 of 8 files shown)


packages/cli/src/ui/App.tsx
```diff
@@ -577,7 +578,15 @@ const App = ({ config, settings, startupWarnings = [], version }: AppProps) => {
   const handleIdePromptComplete = useCallback(
     (result: IdeIntegrationNudgeResult) => {
       if (result === 'yes') {
-        handleSlashCommand('/ide install');
+        if (!currentIDE) {
+          return;
+        }
+        const ideInfo = getIdeInfo(currentIDE);
+        if (ideInfo.isExtensionInstalledByDefault) {
```
skeshive: just realized that older versions of 1P editors won't have the extension installed by default

instead of this approach, can we show the enable nudge if our expected env vars are not set?

packages/core/src/ide/detect-ide.ts
```diff
@@ -15,24 +15,53 @@ export enum DetectedIde {
...
 
-export function getIdeDisplayName(ide: DetectedIde): string {
+export interface IdeInfo {
+  displayName: string;
+  isExtensionInstalledByDefault: boolean;
+}
+
+export function getIdeInfo(ide: DetectedIde): IdeInfo {
   switch (ide) {
     case DetectedIde.VSCode:
```
skeshive, line 22: can you remove any IDEs that we can't detect so its clear we don't support them explicitly

packages/core/src/ide/ide-installer.ts
```diff
@@ -147,23 +147,12 @@ class VsCodeInstaller implements IdeInstaller {
   }
 }
 
-class OpenVSXInstaller implements IdeInstaller {
+class DefaultInstaller implements IdeInstaller {
```
skeshive: IDT we need this- https://github.com/google-gemini/gemini-cli/blob/main/packages/cli/src/ui/commands/ideCommand.ts#L172C1-L173C1

packages/core/src/ide/ide-client.ts
```diff
@@ -68,7 +64,7 @@ export class IdeClient {
   private constructor() {
     this.currentIde = detectIde();
     if (this.currentIde) {
-      this.currentIdeDisplayName = getIdeDisplayName(this.currentIde);
+      this.currentIdeDisplayName = getIdeInfo(this.currentIde).displayName;
```
skeshive, line 67: can this be reverted as well

packages/cli/src/ui/IdeIntegrationNudge.tsx
```diff
@@ -4,44 +4,74 @@
...
-      value: 'dismiss',
+      value: {
+        userSelection: 'dismiss',
+        isExtensionPreInstalled,
+      },
     },
   ];
 
+  const installText = isExtensionPreInstalled
+    ? `If you select Yes, the CLI will have access to access your open files and display diffs directly in ${
```
skeshive, line 68: remove

skeshive: double access here

skeshive (approved):

mattKorwel (approved):

PR #6071, opened 2025-08-12
feat: Add programming language to CLI events

nandakishorereddy-chundi:
## TLDR

Adds a new dimension to CLI tool call events to capture the programming language being used. This change enables the ability to filter active users by programming language, addressing the current gap in our analytics.

## Dive Deeper

This change introduces a new field, programming language, to the tool call event payload. The value for this field is determined by the file extension of the file the user is trying to read/modify. A mapping of file extensions to programming languages has been implemented to handle this.

## Reviewer Test Plan

To validate this change, a reviewer can follow these steps:

- Navigate to a project that uses a supported programming language (e.g., a Python project with a .py file or a JavaScript project with a .js file).
- Check out this branch and run `npm run build` to build.
- run `npm run start -- --debug` to start Gemini CLI in the debug mode.
- Ask the CLI to edit a file or create a new file.
- Inspect the emitted event payload. The payload should now include a new field, `programming_language`, with the correct value (e.g., "Python" or "JavaScript").
- Try this with a few different file types to ensure the mapping works as expected.

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

- Fixes #[6070](https://github.com/google-gemini/gemini-cli/issues/6070)


packages/core/src/core/coreToolScheduler.test.ts
```diff
@@ -784,3 +814,118 @@ describe('CoreToolScheduler request queueing', () => {
...
+      (e: ToolCallEvent) => e.function_name === 'replace',
+    );
+    expect(replaceEvent).toBeDefined();
+    expect(replaceEvent).toHaveProperty('programming_language', 'JavaScript');
+
+    const globEvent = loggedEvents.find(
+      (e: ToolCallEvent) => e.function_name === 'glob',
+    );
+    expect(globEvent).toBeDefined();
+    expect(globEvent).not.toHaveProperty('undefined');
```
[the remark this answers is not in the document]
nandakishorereddy-chundi: This is not true because we have defined the programming_language field in type.js and it will be undefined if there is no value assigned to it.

packages/core/src/core/nonInteractiveToolExecutor.ts
```diff
@@ -15,8 +16,21 @@ import {
...
+  // Logging programming_language for replace, write_file, and read_file function calls.
+  if (event.function_args) {
+    const filePath =
+      event.function_args.file_path || event.function_args.absolute_path;
+    if (typeof filePath === 'string') {
+      event.programming_language = getLanguageFromFilePath(filePath);
+    }
+  }
+  logToolCall(config, event);
+}
```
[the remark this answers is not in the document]
nandakishorereddy-chundi: Extracted adding programming language into a util file and used by both `CoreToolScheduler` and `nonInteractiveToolExecutor`

packages/core/src/utils/language-detection.ts
```diff
@@ -0,0 +1,99 @@
...
+  '.tsconfig': 'TypeScript',
+  '.flow': 'Flow',
+  '.graphql': 'GraphQL',
+  '.proto': 'Protocol Buffers',
+};
+
+export function getLanguageFromFilePath(filePath: string): string | undefined {
+  const extension = path.extname(filePath).toLowerCase();
+  return extensionToLanguageMap[extension];
+}
```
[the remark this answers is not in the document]
nandakishorereddy-chundi: Fixed this

packages/core/src/telemetry/telemetry-utils.ts
```diff
@@ -0,0 +1,22 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { ToolCallEvent } from './types.js';
+import { getLanguageFromFilePath } from '../utils/language-detection.js';
+
+export function addProgrammingLanguageToEvent(
```
jerop: wrapping this doesn't fit right, consider making this an attribute of `gemini_cli.file.operation.count`

jerop (changes requested):
We have an event that fits well with this:

```
- `gemini_cli.file.operation.count` (Counter, Int): Counts file operations.
  - **Attributes**:
    - `operation` (string: "create", "read", "update"): The type of file operation.
    - `lines` (Int, if applicable): Number of lines in the file.
    - `mimetype` (string, if applicable): Mimetype of the file.
    - `extension` (string, if applicable): File extension of the file.
    - `ai_added_lines` (Int, if applicable): Number of lines added/changed by AI.
    - `ai_removed_lines` (Int, if applicable): Number of lines removed/changed by AI.
    - `user_added_lines` (Int, if applicable): Number of lines added/changed by user in AI proposed changes.
    - `user_removed_lines` (Int, if applicable): Number of lines removed/changed by user in AI proposed changes.
```

and it already has an `extension` attribute, which can be used to indicate language 

you can add `language` attribute here if `extension` attribute isn't sufficient

packages/core/src/core/coreToolScheduler.ts
```diff
@@ -887,7 +888,8 @@ export class CoreToolScheduler {
       this.toolCalls = [];
 
       for (const call of completedCalls) {
-        logToolCall(this.config, new ToolCallEvent(call));
+        const event = new ToolCallEvent(call);
+        logToolCall(this.config, addProgrammingLanguageToEvent(event));
```
anj-s: programming language should be an attribute of the event itself. Can we refactor this to do that? Lets not change the top level calling function but instead have it be part of the event object.

owenofbrien: > programming language should be an attribute of the event itself. Can we refactor this to do that? Lets not change the top level calling function but instead have it be part of the event object.

+1, can you just add the programming language in the ToolCallEvent constructor?

nandakishorereddy-chundi: @anj-s  @owenofbrien, thanks for the feedback. I'm a bit confused because there are multiple points of feedback. @jerop suggested using the gemini_cli.file.operation.count event, so I'm not sure which direction to follow.

Can someone please help clarify?

owenofbrien:
> We have an event that fits well with this:
> 
> ```
> - `gemini_cli.file.operation.count` (Counter, Int): Counts file operations.
>   - **Attributes**:
>     - `operation` (string: "create", "read", "update"): The type of file operation.
>     - `lines` (Int, if applicable): Number of lines in the file.
>     - `mimetype` (string, if applicable): Mimetype of the file.
>     - `extension` (string, if applicable): File extension of the file.
>     - `ai_added_lines` (Int, if applicable): Number of lines added/changed by AI.
>     - `ai_removed_lines` (Int, if applicable): Number of lines removed/changed by AI.
>     - `user_added_lines` (Int, if applicable): Number of lines added/changed by user in AI proposed changes.
>     - `user_removed_lines` (Int, if applicable): Number of lines removed/changed by user in AI proposed changes.
> ```
> 
> and it already has an `extension` attribute, which can be used to indicate language
> 
> you can add `language` attribute here if `extension` attribute isn't sufficient

I wasn't aware of file.operation.count, but it looks like it was added as an otel metric without a corresponding telemetry event defined in core/src/telemetry/types.ts. I suggest we create a FileOperationEvent there and then add a loggers.logFileOperation() function that takes the FileOperationEvent as an input and calls metrics.recordFileOperationMetric(). We can then add a new logFileOperation() function to clearcut-logger.ts, which will log this as a Clearcut event (for internal telemetry collection) and call that from loggers.logFileOperation().

Or we could forego adding FileOperation as a new event type and just log this new metadata on the FunctionCallEvent for Clearcut. In that case, we could just move the metrics.recordFileOperationMetric() call into loggers.logToolCall().

packages/core/src/telemetry/clearcut-logger/event-metadata-key.ts
```diff
@@ -219,6 +219,9 @@ export enum EventMetadataKey {
   // Logs user removed lines in edit/write tool response.
   GEMINI_CLI_USER_REMOVED_LINES = 50,
 
+  // Logs the programming language of the project.
```
owenofbrien, line 222: You'll need to add this new key here as well: https://source.corp.google.com/piper///depot/google3/logs/proto/cloud/concord/event_metadata_keys/gemini_cli_key.proto

packages/core/src/telemetry/types.ts
```diff
@@ -115,6 +117,7 @@ export class ToolCallEvent implements BaseTelemetryEvent {
   error?: string;
   error_type?: string;
   prompt_id: string;
+  programming_language?: string;
```
owenofbrien: Remove this param, as it seems to be unused.

owenofbrien (changes requested):
- The file-related metadata that was previously added to the OTEL logging is still not shared with the Clearcut logging. I know on our call, I said we probably don't need to introduce a new FileOperationEvent type, but after looking more closely at the code, I think that would be the easiest and cleanest way to share this metadata. This will also involve adding a new logFileOperation function in loggers.ts that takes the FileOperationEvent as a param and calls a new LogFileOperationEvent function in clearcut-logger.ts as well as calling the recordFileOperationMetric function for OTEL.
- read-file.ts, read-many-files.ts, and write-file.ts should then be updated to call loggers.logFileOperation() instead of calling recordFileOperationMetric() directly.
- edit.ts appears to have been missed when recordFileOperationMetric() was added, but the edit tool is actually our most-used file operation tool, so it should call logFileOperation() as well.

packages/core/src/telemetry/clearcut-logger/event-metadata-key.ts
nandakishorereddy-chundi: Yes, I have raised a CL for that cl/797247759 and sent it for review

packages/core/src/telemetry/clearcut-logger/event-metadata-key.ts
```diff
@@ -246,4 +249,20 @@ export enum EventMetadataKey {
 
   // Logs name of MCP tools as comma seperated string
   GEMINI_CLI_START_SESSION_MCP_TOOLS = 65,
+
+  // ==========================================================================
+  // File Operation Event Keys
+  // ===========================================================================
+
+  // Logs the operation type of the file operation.
+  GEMINI_CLI_FILE_OPERATION_OPERATION = 66,
```
owenofbrien: Let's call this GEMINI_CLI_FILE_OPERATION_TYPE

packages/core/src/tools/edit.ts
```diff
@@ -387,6 +393,40 @@ class EditToolInvocation implements ToolInvocation<EditToolParams, ToolResult> {
...
+      const extension = path.extname(this.params.file_path);
+      const programming_language = getProgrammingLanguage({
+        file_path: this.params.file_path,
+      });
+      if (editData.isNewFile) {
+        logFileOperation(
+          this.config,
+          new FileOperationEvent(
+            EditTool.Name,
+            FileOperation.CREATE,
```
owenofbrien: Since the FileOperation type is the only difference between the if and else blocks, you can simplify the code by removing the if-else and using a ? operator for the FileOperation.

packages/core/src/core/coreToolScheduler.test.ts
```diff
@@ -959,3 +985,118 @@ describe('CoreToolScheduler request queueing', () => {
     expect(approvalMode).toBe(ApprovalMode.AUTO_EDIT);
   });
 });
+
+describe('CoreToolScheduler programming_language logging', () => {
```
owenofbrien: While it doesn't hurt to duplicate the data, it does complicate the codebase and could lead to confusion, so I'd prefer to remove it from ToolCallEvent unless there's a strong reason to keep it there.

owenofbrien (changes requested):

packages/core/src/core/coreToolScheduler.test.ts
nandakishorereddy-chundi: Removed programming language from the ToolCallEvent

packages/core/src/tools/edit.ts
nandakishorereddy-chundi: Done

packages/core/src/telemetry/clearcut-logger/event-metadata-key.ts
nandakishorereddy-chundi: Done

packages/core/src/core/coreToolScheduler.test.ts
```diff
@@ -4,7 +4,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
```
owenofbrien: Do you still need the changes in this file?

owenofbrien (approved):

packages/core/src/core/coreToolScheduler.test.ts
nandakishorereddy-chundi: Reverted the changes

packages/core/src/telemetry/telemetry-utils.ts
nandakishorereddy-chundi: Fixed

NTaylorMullen (approved):

jerop (approved):

PR #6073, opened 2025-08-12
Add support for trustedFolders.json config file

shrutip90:
## TLDR

A new JSON config file called trustedFolders.json will be created on launch when a user chooses whether they trust the current folder or not. Eg contents:
**{
  "/Users/padamata/Documents/git/du4/gemini-cli": "TRUST_FOLDER",
  "/Users/padamata/Documents/git/du4/gemini-cli/packages/cli": "DO_NOT_TRUST"
}**

## Dive Deeper

- Child folders inherit the trust for "trusted" folders but not for "untrusted" folders, mimicking the behavior in VSCode.
- The config file is supported at system and user scope, with system scope taking precedence over user scope config.
- In case of errors in the config file, the next write will overwrite the file.
- In case of duplicate entries, the last entry will be taken and the previous entries will be overwritten on the next write.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Related to https://github.com/google-gemini/maintainers-gemini-cli/issues/641

### DIFF
--- a/packages/cli/src/config/trustedFolders.test.ts
+++ b/packages/cli/src/config/trustedFolders.test.ts
@@ -1,0 +1,289 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+// Mock 'os' first.
+import * as osActual from 'os';
+vi.mock('os', async (importOriginal) => {
+  const actualOs = await importOriginal<typeof osActual>();
+  return {
+    ...actualOs,
+    homedir: vi.fn(() => '/mock/home/user'),
+    platform: vi.fn(() => 'linux'),
+  };
+});
+
+// Mock './trustedFolders.js' to ensure it uses the mocked 'os.homedir()' for its internal constants.
+vi.mock('./trustedFolders.js', async (importActual) => {
+  const originalModule =
+    await importActual<typeof import('./trustedFolders.js')>();
+  return {
+    __esModule: true,
+    ...originalModule,
+  };
+});
+
+import {
+  describe,
+  it,
+  expect,
+  vi,
+  beforeEach,
+  afterEach,
+  type Mocked,
+  type Mock,
+} from 'vitest';
+import * as fs from 'fs';
+import stripJsonComments from 'strip-json-comments';
+import * as path from 'path';
+
+import {
+  loadTrustedFolders,
+  USER_TRUSTED_FOLDERS_PATH,
+  getSystemTrustedFoldersPath,
+  TrustLevel,
+  isCurrentDirectoryTrusted,
+} from './trustedFolders.js';
+import { SettingScope } from './settings.js';
+
+vi.mock('fs', async (importOriginal) => {
+  const actualFs = await importOriginal<typeof fs>();
+  return {
+    ...actualFs,
+    existsSync: vi.fn(),
+    readFileSync: vi.fn(),
+    writeFileSync: vi.fn(),
+    mkdirSync: vi.fn(),
+  };
+});
+
+vi.mock('strip-json-comments', () => ({
+  default: vi.fn((content) => content),
+}));
+
+describe('Trusted Folders Loading', () => {
+  let mockFsExistsSync: Mocked<typeof fs.existsSync>;
+  let mockStripJsonComments: Mocked<typeof stripJsonComments>;
+  let mockFsWriteFileSync: Mocked<typeof fs.writeFileSync>;
+
+  beforeEach(() => {
+    vi.resetAllMocks();
+    mockFsExistsSync = vi.mocked(fs.existsSync);
+    mockStripJsonComments = vi.mocked(stripJsonComments);
+    mockFsWriteFileSync = vi.mocked(fs.writeFileSync);
+    vi.mocked(osActual.homedir).mockReturnValue('/mock/home/user');
+    (mockStripJsonComments as unknown as Mock).mockImplementation(
+      (jsonString: string) => jsonString,
+    );
+    (mockFsExistsSync as Mock).mockReturnValue(false);
+    (fs.readFileSync as Mock).mockReturnValue('{}');
+  });
+
+  afterEach(() => {
+    vi.restoreAllMocks();
+  });
+
+  it('should load empty rules if no files exist', () => {
+    const { rules, errors } = loadTrustedFolders();
+    expect(rules).toEqual([]);
+    expect(errors).toEqual([]);
+  });
+
+  it('should load system rules if only system file exists', () => {
+    const systemPath = getSystemTrustedFoldersPath();
+    (mockFsExistsSync as Mock).mockImplementation((p) => p === systemPath);
+    const systemContent = {
+      '/system/folder': TrustLevel.TRUST_PARENT,
+    };
+    (fs.readFileSync as Mock).mockImplementation((p) => {
+      if (p === systemPath) return JSON.stringify(systemContent);
+      return '{}';
+    });
+
+    const { rules, errors } = loadTrustedFolders();
+    expect(rules).toEqual([
+      { path: '/system/folder', trustLevel: TrustLevel.TRUST_PARENT },
+    ]);
+    expect(errors).toEqual([]);
+  });
+
+  it('should load user rules if only user file exists', () => {
+    const userPath = USER_TRUSTED_FOLDERS_PATH;
+    (mockFsExistsSync as Mock).mockImplementation((p) => p === userPath);
+    const userContent = {
+      '/user/folder': TrustLevel.TRUST_FOLDER,
+    };
+    (fs.readFileSync as Mock).mockImplementation((p) => {
+      if (p === userPath) return JSON.stringify(userContent);
+      return '{}';
+    });
+
+    const { rules, errors } = loadTrustedFolders();
+    expect(rules).toEqual([
+      { path: '/user/folder', trustLevel: TrustLevel.TRUST_FOLDER },
+    ]);
+    expect(errors).toEqual([]);
+  });
+
+  it('should merge system and user rules, with system taking precedence', () => {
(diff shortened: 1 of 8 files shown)


packages/cli/src/config/trustedFolders.ts
```diff
@@ -0,0 +1,217 @@
...
+      return true;
+    }
+  }
+
+  for (const untrustedPath of untrustedPaths) {
+    const normalizedUntrustedPath = path.normalize(untrustedPath);
+    if (normalizedCwd === normalizedUntrustedPath) {
+      return false;
+    }
+  }
```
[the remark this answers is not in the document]
shrutip90: This is intentional to mimic vscode workspace trust behavior.

packages/cli/src/config/trustedFolders.ts
```diff
@@ -0,0 +1,217 @@
...
+    }
+
+    fs.writeFileSync(
+      trustedFoldersFile.path,
+      JSON.stringify(trustedFoldersFile.config, null, 2),
+      'utf-8',
+    );
+  } catch (error) {
+    console.error('Error saving trusted folders file:', error);
+  }
```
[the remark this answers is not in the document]
jacob314: ignore this comment.

packages/cli/src/ui/hooks/useFolderTrust.ts
```diff
@@ -5,19 +5,41 @@
...
+          break;
+        case FolderTrustChoice.DO_NOT_TRUST:
+          trustLevel = TrustLevel.DO_NOT_TRUST;
+          break;
+        default:
+          return;
+      }
+
+      trustedFolders.setValue(SettingScope.User, cwd, trustLevel);
       settings.setValue(SettingScope.User, 'folderTrust', true);
```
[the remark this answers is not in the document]
jacob314: please address. agree this is confusing. consider renaming the folderTrust setting if this is the intended behavior

packages/cli/src/ui/hooks/useFolderTrust.ts
```diff
@@ -5,19 +5,41 @@
...
 
-export const useFolderTrust = (settings: LoadedSettings) => {
+export const useFolderTrust = (settings: LoadedSettings, config: Config) => {
   const [isFolderTrustDialogOpen, setIsFolderTrustDialogOpen] = useState(
     !!settings.merged.folderTrustFeature &&
-      // TODO: Update to avoid showing dialog for folders that are trusted.
-      settings.merged.folderTrust === undefined,
+      (settings.merged.folderTrust === undefined ||
+        settings.merged.folderTrust === true) &&
+      config.isTrustedFolder() === undefined,
```
jacob314: can this just use config.isTrustedFolder() as the source of truth for this?

packages/cli/src/config/trustedFolders.test.ts
```diff
@@ -0,0 +1,289 @@
...
+vi.mock('os', async (importOriginal) => {
+  const actualOs = await importOriginal<typeof osActual>();
+  return {
+    ...actualOs,
+    homedir: vi.fn(() => '/mock/home/user'),
+    platform: vi.fn(() => 'linux'),
+  };
+});
+
+// Mock './trustedFolders.js' to ensure it uses the mocked 'os.homedir()' for its internal constants.
```
jacob314: this seems a bit complex. can you ask gemini cli to achieve this same result while just using regular imports for trustedFolders.js? If not it is ok but would be good to keep this simpler.

packages/cli/src/config/trustedFolders.ts
```diff
@@ -0,0 +1,217 @@
...
+  USER_SETTINGS_DIR,
+  TRUSTED_FOLDERS_FILENAME,
+);
+
+export function getSystemTrustedFoldersPath(): string {
+  if (process.env.GEMINI_CLI_SYSTEM_TRUSTED_FOLDERS_PATH) {
+    return process.env.GEMINI_CLI_SYSTEM_TRUSTED_FOLDERS_PATH;
+  }
+  if (platform() === 'darwin') {
+    return `/Library/Application Support/GeminiCli/${TRUSTED_FOLDERS_FILENAME}`;
```
jacob314: is there a reason these should be stored here rather than the same directory user level settings are stored? E.g. ~/.gemini/
?

packages/cli/src/config/trustedFolders.ts
```diff
@@ -0,0 +1,217 @@
...
+
+  get rules(): TrustRule[] {
+    const mergedConfig = { ...this.user.config, ...this.system.config };
+    return Object.entries(mergedConfig).map(([path, trustLevel]) => ({
+      path,
+      trustLevel,
+    }));
+  }
+
+  forScope(scope: SettingScope): TrustedFoldersFile {
```
jacob314: why do we need System scope as well as user scope for this?

packages/cli/src/config/trustedFolders.ts
```diff
@@ -0,0 +1,217 @@
...
+      case TrustLevel.DO_NOT_TRUST:
+        untrustedPaths.push(rule.path);
+        break;
+      default:
+        // Do nothing for unknown trust levels.
+        break;
+    }
+  }
+
+  const cwd = process.cwd();
```
jacob314, line 143: use the workspaceRoot rather than process.cwd()

packages/cli/src/config/trustedFolders.ts
```diff
@@ -0,0 +1,217 @@
...
+  const normalizedCwd = path.normalize(cwd);
+
+  for (const trustedPath of trustedPaths) {
+    const normalizedTrustedPath = path.normalize(trustedPath);
+    if (normalizedCwd === normalizedTrustedPath) {
+      return true;
+    }
+    const trustedPathWithSep = normalizedTrustedPath.endsWith(path.sep)
```
jacob314: compute the relative path rather than using endsWith.  See this breadcrumb for one of the cases we have this logic. It should be probably moved to a public utility method.
https://github.com/google-gemini/gemini-cli/blob/5ee5cf17ba952ad6bfb8a671c617aa1a2122cc6b/packages/core/src/utils/workspaceContext.ts#L159

packages/cli/src/ui/hooks/useFolderTrust.test.ts
```diff
@@ -4,75 +4,161 @@
...
+import { type Config } from '@google/gemini-cli-core';
 import { LoadedSettings, SettingScope } from '../../config/settings.js';
 import { FolderTrustChoice } from '../components/FolderTrustDialog.js';
+import { loadTrustedFolders, TrustLevel } from '../../config/trustedFolders.js';
+import * as process from 'process';
+
+// Mock dependencies
+vi.mock('../../config/trustedFolders.js', () => ({
+  loadTrustedFolders: vi.fn(),
+  TrustLevel: {
```
jacob314: clean up how we are doing this. we shouldn't have to mock trustedFolders and we definitely shouldn't have to mock enums.

jacob314 (commented):
Done with a round of comments. Generally this looks good but would like to discuss these possible ways to make this code more aligned with the rest of the code in the repo.

packages/cli/src/config/trustedFolders.ts
shrutip90: This was for the system level path, removed now anyway since we decided to support this file only at user level.

packages/cli/src/config/trustedFolders.ts
shrutip90: Removed system scope as discussed offline, since Vscode doesn't seem to have something similar.

packages/cli/src/ui/hooks/useFolderTrust.ts
shrutip90: Updated config.isTrustedFolder() to include these checks so we can just use that here.

packages/cli/src/config/trustedFolders.ts
shrutip90: Settings uses the same for getting the workspace directory: https://github.com/google-gemini/gemini-cli/blob/3a87712c1a15dac9f0a717b40fbf9e59398177ca/packages/cli/src/config/settings.ts#L263.

Renamed the method to isWorkspaceTrusted to make it a bit more clear. Please lmk if this still needs any changes.

packages/cli/src/config/trustedFolders.ts
shrutip90: Found an existing util function that we could use. Thanks.

packages/cli/src/config/trustedFolders.test.ts
shrutip90: The tests work fine even after removing this mock. Thanks.

packages/cli/src/ui/hooks/useFolderTrust.test.ts
shrutip90: Done

packages/cli/src/ui/hooks/useFolderTrust.ts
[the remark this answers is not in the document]
jacob314: as discussed offline please remove this line.

jacob314 (commented):
lgtm
Will approve and merge when the minor comment is addressed. Thanks for landning this!

packages/cli/src/ui/hooks/useFolderTrust.ts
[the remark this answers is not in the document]
shrutip90: Done, removed the setting to always true.

jacob314 (approved):
lgtm

PR #6074, opened 2025-08-12
chore(telemetry): Add various surface detection to `determineSurface` for logging.

richieforeman:
## TLDR

Adds detection of various surfaces to logging telemetry.

- Devin
- Replit
- Cursor
- Github Action
- Vscode terminal

## Dive Deeper

I went and tested each of these and confirmed them to work.  I added some pretty detailed tests around the behavior that I feel good about.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | x  | ❓  | x  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

N/A


packages/core/src/telemetry/clearcut-logger/clearcut-logger.test.ts
```diff
@@ -135,16 +138,109 @@ describe('ClearcutLogger', () => {
       });
     });
 
-    it('logs the current surface', () => {
-      const { logger } = setup({});
+    it.each([
```
jacob314: nice! great to see more parameterized tests.

packages/core/src/telemetry/clearcut-logger/clearcut-logger.ts
```diff
@@ -83,10 +83,20 @@ export interface LogRequest {
  * methods might have in their runtimes.
  */
 function determineSurface(): string {
-  if (process.env.CLOUD_SHELL === 'true') {
+  if (process.env.__COG_BASHRC_SOURCED) {
```
jacob314: please merge this with the existing detectIde method in detect-ide.ts which has very similar logic.  I think that version is a little more robust in that it ensures the TERM_PROGRAM is vscode before assuming you could be any of the VSCode family of ids.
https://github.com/google-gemini/gemini-cli/blob/74fd0841d0d7148127e586fce4c550a01ff40e90/packages/core/src/ide/detect-ide.ts#L54

packages/core/src/telemetry/clearcut-logger/clearcut-logger.ts
```diff
@@ -83,10 +83,20 @@ export interface LogRequest {
...
+    return 'DEVIN';
+  } else if (process.env.REPLIT_USER) {
+    return 'REPLIT';
+  } else if (process.env.CURSOR_TRACE_ID) {
+    return 'CURSOR';
+  } else if (process.env.GITHUB_SHA) {
+    return 'GITHUB_ACTION';
```
jerop: we're already using `GitHub` for this, this may cause divergence in data - was following naming schema from first item that used this, which was `VSCode` 

https://github.com/google-github-actions/run-gemini-cli/blob/main/action.yml#L160

jerop (dismissed):

packages/core/src/telemetry/clearcut-logger/clearcut-logger.test.ts
richieforeman: :)

packages/core/src/telemetry/clearcut-logger/clearcut-logger.ts
richieforeman: Done.

packages/core/src/telemetry/clearcut-logger/clearcut-logger.ts
richieforeman: Noted. I made sure that this behavior && naming is preserved.  I also added a test around this specific case. Thanks for flagging, I really appreciate it since I'm a bit newer to this codebase and things are moving very fast.

jacob314 (approved):
lgtm

PR #6088, opened 2025-08-12
feat(core): refactor shell execution to use node-pty 

### LINKED ISSUE #5132: Support interaction with terminal windows embedded in Gemini CLI
This will address issues such as
https://github.com/google-gemini/gemini-cli/issues/1936
https://github.com/google-gemini/gemini-cli/issues/2719
https://github.com/google-gemini/gemini-cli/issues/3617

A reasonable approach to implement this is to use
https://github.com/rse/stmux
or
https://www.npmjs.com/package/@xterm/headless

Only the very last terminal in the chat history should be interactive.
Implementing this should also indirectly address various rendering issues where showing raw unfiltered terminal output in the console causes problems when terminal commands that use advanced terminal escape codes run.

galdawave:
## TLDR

This pull request replaces the standard `child_process.spawn` with `node-pty` for executing shell commands. This change introduces a pseudo-terminal (pty) that allows for a much more accurate and feature-rich terminal emulation within the Gemini CLI. The primary benefit is the correct handling of complex ANSI escape codes, colors, and cursor movements, which were previously either stripped or not handled correctly.

## Dive Deeper

The core of this change is in the `ShellExecutionService`. Instead of capturing `stdout` and `stderr` streams from `child_process`, we now spawn a pseudo-terminal using `node-pty`. The output of this pty is then piped into a headless terminal instance from `@xterm/headless`. This allows us to interpret the raw terminal output stream and get the "rendered" text content, correctly processing complex terminal interactions like cursor movements, screen clearing, and advanced color formatting that a simple `strip-ansi` call cannot handle.

To ensure the pseudo-terminal behaves as closely as possible to the user's actual terminal, the terminal dimensions (columns and rows) are now detected in the UI layer using a new `useTerminalSize` hook. These dimensions are then passed down through the tool scheduling layers (`useReactToolScheduler`, `CoreToolScheduler`) all the way to the `ShellExecutionService` when a command is executed.

Since a pty provides a single, unified output stream (just like a real terminal), the distinction between `stdout` and `stderr` at the stream handling level has been removed in `shellCommandProcessor` and `ShellExecutionService`, simplifying the output processing logic.

## Reviewer Test Plan

1.  **Setup:** Pull this branch and run `npm install` to fetch the new `node-pty` and `@xterm/headless` dependencies. Then, run `npm run build`.
2.  **Basic Commands:** Run some basic commands and verify the output is clean and correct.
    *   `ls -la`
    *   `echo "hello world"`
3.  **Complex UI Commands:** Run commands that are known to produce more complex terminal UIs with colors, progress bars, and overwriting lines.
    *   `npm run preflight`
    *   `npm install` in a project with many dependencies.
    *   Run a test suite that uses a "rich" reporter, for example: `npx jest --color`.
4.  **Verification:**
    *   Confirm that the output is well-formatted and does not contain any raw or garbled ANSI escape codes.
    *   Check that colors are displayed correctly (or stripped cleanly if your terminal doesn't support them).
    *   Ensure that long-running commands with updating lines (like progress bars) display the final state correctly without showing all the intermediate frames.
5.  **Cancellation:** Run a long-running command (e.g., `sleep 10`) and cancel it with `Ctrl+C`. Verify that the process is terminated correctly.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

- Resolves #5132


packages/core/src/services/shellExecutionService.ts
```diff
@@ -87,167 +91,145 @@ export class ShellExecutionService {
...
+                  type: 'binary_progress',
+                  bytesReceived: totalBytes,
+                });
+                resolve();
+              }
+            }),
+        );
+      };
 
-        if (stream === 'stdout') {
```
olcan, line 157: Looks like we lose the stdout/stderr distinction here as a side effect. I've seen this with `docker run ... --tty` and it can be problematic there since you might pipe the stdout and not want stderr mixed in. IIUC in this context the only potential concern is that the model will see the streams merged. My guess is that's fine, but of course with LLMs you never know, so I'd recommend keeping an eye on this, maybe adding a comment around here. 

Another thing to think about is how this impacts displayed output, i.e. what the human sees. Looks like displayed output is currently constructed as stdout + stderr (not sure why, i don't think it is my code), and this change should replace that with proper mixed output that one would see in a terminal.

packages/core/src/services/shellExecutionService.ts
```diff
@@ -87,167 +91,145 @@ export class ShellExecutionService {
...
-              if (!exited) {
-                process.kill(-child.pid, 'SIGKILL');
-              }
-            } catch (_e) {
-              // Fall back to killing just the main process if group kill fails.
-              if (!exited) child.kill('SIGKILL');
-            }
-          }
+        if (ptyProcess.pid && !exited) {
+          ptyProcess.kill();
```
olcan: Would be good to confirm that this kills all subprocesses as a group. The previous code was jumping through hoops to do that ('detached' option, negative pid to kill, etc).

packages/vscode-ide-companion/tsconfig.json
```diff
@@ -6,7 +6,8 @@
     "lib": ["ES2022", "dom"],
     "sourceMap": true,
     "rootDir": "src",
-    "strict": true /* enable all strict type-checking options */
+    "strict": true /* enable all strict type-checking options */,
+    "skipLibCheck": true
```
olcan: why? may be good to add a comment

packages/core/package.json
```diff
@@ -48,9 +48,10 @@
     "picomatch": "^4.0.1",
     "shell-quote": "^1.8.3",
     "simple-git": "^3.28.0",
-    "strip-ansi": "^7.1.0",
     "undici": "^7.10.0",
-    "ws": "^8.18.0"
+    "ws": "^8.18.0",
+    "node-pty": "1.0.0",
```
olcan: One concern I had with node-pty before, when I was experimenting with interactivity, was that it had a dependency on python, which felt too heavy at the time. Is that still the case, and are we good with it now given the benefits?

olcan (approved):
I'm approving this but please see my feedback, mostly about adding some comments and watching out for certain things like the switch to mixed output in model context, and new dependencies (e.g. on python for node-pty) that likely require robust testing across platforms + environments (incl. sandboxes) + terminals + commands, including commands that push terminal features e.g. various editors, gemini-cli itself, other clis, etc. This would be less for the model but more for human dev running these things via shell mode (which shares the shell execution service w/ shell tool iiuc). I would also test robust termination of subprocess hierarchies given the switch in termination logic.

olcan:
> I'm approving this but please see my feedback, mostly about adding some comments and watching out for certain things like the switch to mixed output in model context, and new dependencies (e.g. on python for node-pty) that likely require robust testing across platforms + environments (incl. sandboxes) + terminals + commands, including commands that push terminal features e.g. various editors, gemini-cli itself, other clis, etc. This would be less for the model but more for human dev running these things via shell mode (which shares the shell execution service w/ shell tool iiuc). I would also test robust termination of subprocess hierarchies given the switch in termination logic.

One more thing that came to mind is the background PID detection via pgrep, see around https://github.com/google-gemini/gemini-cli/blob/f81ccd8d7a996dc3f8b9dbc2fab1e428d482d6a5/packages/core/src/tools/shell.ts#L304-L305. Note how we exclude result.pid here because on some platforms (e.g. linux sandbox but not macos host iirc) pgrep returns the shell process pid also, and we don't want to report that as a "background pid" to the model. With this change if there is an additional parent process we may need to exclude that as well since it would not count as a background pid. 

I would also make sure that `result.pid` is still a process group in the new approach (in the old approach the 'detached:true' option would ensure this) as the model is being told it is around https://github.com/google-gemini/gemini-cli/blob/f81ccd8d7a996dc3f8b9dbc2fab1e428d482d6a5/packages/core/src/tools/shell.ts#L341-L342 and this is important because it means the model may try to kill the group via `kill -pgid` and this is indeed the most robust way for the model to kill a background process group.

jacob314 (approved):
lgtm

packages/core/package.json
galdawave: It does depend on python but node-pty has installers and stuff to make sure it is installed correctly on all os's, i think in order to support this feature + better ansi key handling, node-pty is need.

packages/core/src/services/shellExecutionService.ts
galdawave: added 'SIGHUP' signal, this will send the to the subgroups the kill message as well

packages/vscode-ide-companion/tsconfig.json
galdawave: Added comment.

packages/core/src/services/shellExecutionService.ts
galdawave: added a comment about the change, but as you pointed we already combine the stdout and stderr. If we see a drop in quality we can revisit this and figure out how to strip stderr out.

https://github.com/google-gemini/gemini-cli/blob/2dbd5ecdc80b55cc13c81a0f836ad65ef874e8f8/packages/core/src/services/shellExecutionService.ts#L184

PR #6090, opened 2025-08-12
fix(core): Discard thought signature when switching from Gemini API Key to OAuth

SandyTao520:
## TLDR

This pull request ensures that when a user switches from a GenAI API key to logging in with Google (which uses the Vertex API), any "thought" data from the previous session is stripped from the conversation history. This prevents potential issues when resuming a conversation across different backend services.

## Dive Deeper

When a user authenticates with a GenAI API key, the conversation history may include internal "thought" metadata. However, the Vertex API, which is used when logging in with Google, does not support this metadata. If a user switches from using a GenAI key to logging in with Google, this metadata could cause errors or unexpected behavior.

This change introduces a `stripThoughts` flag to the `setHistory` method in `GeminiClient`. When a user's authentication method changes from GenAI to Vertex, this flag is set to `true`, and the `setHistory` method removes any `thoughtSignature` properties from the conversation history before setting it on the new client. This ensures a clean transition between the two services. The reverse is not true; when switching from Vertex to GenAI, the history is preserved as-is.

## Reviewer Test Plan

1.  Start the Gemini CLI using a GenAI API key.
2.  Have a short conversation with the model.
3.  Use /auth to switch to Login with Google
4.  Verify that the previous conversation history is loaded and that there are no errors.
5.  You can also inspect the code changes in `packages/core/src/config/config.ts` and `packages/core/src/core/client.ts` to confirm the logic.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ✅  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


packages/core/src/config/config.ts
```diff
@@ -375,13 +375,17 @@ export class Config {
     const newGeminiClient = new GeminiClient(this);
     await newGeminiClient.initialize(newContentGeneratorConfig);
 
+    const fromGenaiToVertex =
```
jacob314, line 384: add a comment explaining why this is needed as this will be surprising to future readers of this code.

packages/core/src/core/client.ts
```diff
@@ -164,8 +164,17 @@ export class GeminiClient {
     return this.getChat().getHistory();
   }
 
-  setHistory(history: Content[]) {
-    this.getChat().setHistory(history);
+  setHistory(history: Content[], stripThoughts = false) {
```
jacob314: nit: make stripThoughts a named parameter  so its meaning is obvious at callsites.

packages/core/src/core/client.ts
```diff
@@ -164,8 +164,17 @@ export class GeminiClient {
...
+    const historyToSet = stripThoughts
+      ? history.map((content) => ({
```
jacob314: can you clean this up a little bit? 
right now it is adding 
`thoughtSignature: undefined` 
properties everywhere. 
Would be nice to clone and then call 
delete part.thoughtSignature
if and only if part has a thoughtSignature field.
or something along those lines.
In addition, I think it is possible there are more deeply nested Parts containing other parts that might have thoughtSignature. Out of paranoia I would suggest creating a version of this that does a robust traversal of PartListUnion stripping thoughtSignature fields for all types that might have them.

jacob314 (commented):
This will be a really helpful change. Added some comments to try to make this as robust and safe as possible.

packages/core/src/core/client.ts
SandyTao520: https://github.com/googleapis/js-genai/blob/main/src/types.ts#L1046 I think it's not possible to have nested parts.

Cleaned it up to only delete thoughtSignature if it exists.

packages/core/src/core/client.ts
```diff
@@ -164,8 +164,32 @@ export class GeminiClient {
...
+  setHistory(
+    history: Content[],
+    { stripThoughts = false }: { stripThoughts?: boolean } = {},
+  ) {
+    const historyToSet = stripThoughts
+      ? history.map((content) => {
+          const newContent = { ...content };
+          if (newContent.parts) {
+            newContent.parts = newContent.parts.map((part) => {
+              if (
```
jacob314, line 176: I would think you should also handle cases where the part is an Array (PartList) in which case you need to recursively filter out thoughtSignature fields from its children.

jacob314 (commented):
one last comment that is perhaps slightly paranoid.

jacob314 (approved):
lgtm

PR #6094, opened 2025-08-13
feat: add file change tracking to session metrics

### LINKED ISSUE #5827: Introduce `+x -y` in `/stats`

We just introduced added and removed line counts in telemetry - https://github.com/google-gemini/gemini-cli/pull/5823. It'd be useful to show these stats on exit or when `/stats` command is run.

cc @agarwalravikant @abhipatel12 

### Why is this needed?

This would help users get a view of how much code change was made during the session.

### Additional context

see `gemini_cli.file.operation.count` in https://github.com/google-gemini/gemini-cli/blob/main/docs/telemetry.md#metrics

bulkypanda:
## TLDR

introduced +x -y in /stats

## Dive Deeper

useful to show these stats on exit or when /stats command is run and would help users get a view of how much code change was made during the session

## Reviewer Test Plan

have the cli make changes to files and then check /stats

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ✅  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

- Resolves #5827


packages/cli/src/ui/components/StatsDisplay.tsx
```diff
@@ -211,6 +211,18 @@ export const StatsDisplay: React.FC<StatsDisplayProps> = ({
         <StatRow title="Success Rate:">
           <Text color={successColor}>{computed.successRate.toFixed(1)}%</Text>
         </StatRow>
+        {(computed.totalLinesAdded > 0 || computed.totalLinesRemoved > 0) && (
+          <StatRow title="Code Changes:">
+            <Text>
+              <Text color={Colors.AccentGreen}>
+                +{computed.totalLinesAdded}
+              </Text>{' '}
+              <Text color={Colors.AccentRed}>
```
jacob314: please use the new semantic colors. 
See https://github.com/google-gemini/gemini-cli/pull/6253/files#diff-f01c86147ce7947ec07c4ba54058b0e663e060252846919fe77139fcc3b1c1b3
as an example.
We'll soon be making the semantic colors mandatory so you won't accidentally use the old colors.

packages/core/src/telemetry/uiTelemetry.ts
```diff
@@ -209,6 +217,16 @@ export class UiTelemetryService extends EventEmitter {
       tools.totalDecisions[event.decision]++;
       toolStats.decisions[event.decision]++;
     }
+
+    // Aggregate line count data from metadata
+    if (event.metadata) {
+      if (Number.isFinite(event.metadata.ai_added_lines)) {
```
jacob314: nit:are you really trying to check for finite or just that the values are not null/undefined.
Would suggest being a little more idiomatic if you really just want to make sure you got a number.

jacob314 (commented):
A couple minior comments. Approved after they are addressed.Thanks for adding easy to read tests for this!

bulkypanda:
> A couple minior comments. Approved after they are addressed.Thanks for adding easy to read tests for this!

all addressed!

packages/core/src/telemetry/uiTelemetry.test.ts
```diff
@@ -633,4 +633,90 @@ describe('UiTelemetryService', () => {
       expect(spy).toHaveBeenCalledOnce();
     });
   });
+
+  describe('Tool Call Event with Line Count Metadata', () => {
+    it('should aggregate valid line count metadata', () => {
+      const toolCall = createFakeCompletedToolCall('test_tool', true, 100);
+      const event = {
+        ...JSON.parse(JSON.stringify(new ToolCallEvent(toolCall))),
```
jacob314: use structuredClone instead of JSON.parse(JSON.stringify here and elsewhere

jacob314 (approved):
lgtm

packages/core/src/telemetry/uiTelemetry.ts
```diff
@@ -209,6 +217,22 @@ export class UiTelemetryService extends EventEmitter {
       tools.totalDecisions[event.decision]++;
       toolStats.decisions[event.decision]++;
     }
+
+    // Aggregate line count data from metadata
+    if (event.metadata) {
+      if (
+        typeof event.metadata.ai_added_lines === 'number' &&
```
jacob314: sorry for the confusion. What you had previously was better than this. I was suggesting that if it is known this is a numberor undefined just check for !== undefined as that is clearer.

bulkypanda:
all done.

PR #6098, opened 2025-08-13
chore(ci): Ensure e2e workflow is consistent and not vulnerable to injection attacks

sethvargo:
## TLDR

These are some best practice recommendations for ensuring untrusted inputs aren't available as arbitrary execution in bash scripts. It also ensures our YAML is always interpreted as expected with proper quoting and consistent blocks.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -1,75 +1,88 @@
-# .github/workflows/e2e.yml
-
-name: E2E Tests
+name: 'E2E Tests'
 
 on:
   push:
-    branches: [main]
+    branches:
+      - 'main'
   merge_group:
 
 jobs:
   e2e-test-linux:
-    name: E2E Test (Linux) - ${{ matrix.sandbox }}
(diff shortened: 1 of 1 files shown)


cornmander:
Are there linters we can apply to make sure we do this systematically?

sethvargo:
> Are there linters we can apply to make sure we do this systematically?

Yes, and I have them running in my local branch. I'm making the changes in smaller chunks, then I'll propose adding the linters. Otherwise there's a PR with like 2.5k changed lines.

cornmander (approved):

jerop (approved):

mattKorwel:
> > Are there linters we can apply to make sure we do this systematically?
> 
> Yes, and I have them running in my local branch. I'm making the changes in smaller chunks, then I'll propose adding the linters. Otherwise there's a PR with like 2.5k changed lines.

@sethvargo which linters are you refering to? Linting for catching unpined actions, or linting for changing all the linting and formatting in our code? If the latter lets talk IRL ;)

PR #6102, opened 2025-08-13
chore: update bash to follow shellcheck recommendations

sethvargo:
## TLDR

Some of these are best practices, and some of them are legitimate issues. In the future, the `shellcheck` linter will catch these kinds of things.

### DIFF
--- a/.github/scripts/pr-triage.sh
+++ b/.github/scripts/pr-triage.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 set -euo pipefail
 
 # Initialize a comma-separated string to hold PR numbers that need a comment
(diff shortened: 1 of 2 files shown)


.github/scripts/pr-triage.sh
```diff
@@ -1,163 +1,173 @@
...
+        IFS=',' read -ra ISSUE_LABEL_ARRAY <<< "${ISSUE_LABELS}"
+        IFS=',' read -ra PR_LABEL_ARRAY <<< "${PR_LABELS}"
 
         # Find labels to add (on issue but not on PR)
         local LABELS_TO_ADD=""
         for label in "${ISSUE_LABEL_ARRAY[@]}"; do
-            if [ -n "$label" ] && [[ ! " ${PR_LABEL_ARRAY[*]} " =~ " ${label} " ]]; then
-                if [ -z "$LABELS_TO_ADD" ]; then
-                    LABELS_TO_ADD="$label"
+            if [[ -n "${label}" ]] && [[ " ${PR_LABEL_ARRAY[*]} " != *" ${label} "* ]]; then
```
sethvargo, line 89: This was a recommendation by shellcheck, and I _think_ it's more technically correct?

scripts/create_alias.sh
cornmander, line 1: This script runs under bash but then checks $SHELL 🤔

cornmander (approved):

scripts/create_alias.sh
sethvargo: The execution of the script doesn't change $SHELL:

```text
$ /usr/bin/env bash -c 'echo $SHELL'
/opt/homebrew/bin/zsh
```

```text
$ cat ./run.sh
#!/usr/bin/env bash
echo "${SHELL}"

$ ./run.sh
/opt/homebrew/bin/zsh

$ bash ./run.sh
/opt/homebrew/bin/zsh
```

PR #6109, opened 2025-08-13
chore(ci): Ensure ci workflows is consistent and not vulnerable to injection attacks

sethvargo:
## TLDR

These are some best practice recommendations for ensuring untrusted inputs aren't available as arbitrary execution in bash scripts. It also ensures our YAML is always interpreted as expected with proper quoting and consistent blocks.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,13 +1,19 @@
-# .github/workflows/ci.yml
-
-name: Gemini CLI CI
+name: 'Gemini CLI CI'
 
 on:
   push:
-    branches: [main, release]
+    branches:
+      - 'main'
+      - 'release'
   pull_request:
-    branches: [main, release]
+    branches:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -49,114 +51,132 @@ jobs:
 
       - name: Run type check
         run: npm run typecheck
+concurrency:
+  group: '${{ github.workflow }}-${{ github.head_ref || github.ref }}'
+  cancel-in-progress: ${{ github.ref != 'refs/heads/main' && !startsWith(github.ref, 'refs/heads/release/') }}
```
sethvargo: 👀  This ensures that ci jobs from PRs are cancelled when new commits are pushed, except for `main` and release branches. This should reduce queuing times.

cornmander (approved):

jerop (approved):

PR #6113, opened 2025-08-13
fix: use server-returned project for gca free tier auth

### LINKED ISSUE #5331: Cannot pass workspace account verification on macOS 15.6: `Failed to login. Message: Precondition check failed.`
**Description:**

Hi, I encountered an issue when trying to log in with my **workspace account** on macOS 15.6. The authentication fails with the following error message:
```
Failed to login. Message: Precondition check failed.
```

**Details:**

- On **macOS 15.6**, I am unable to pass the account verification step using my workspace account.
- The same workspace account works fine and passes verification on **macOS 15.5**.
- The error consistently appears every time I try to log in on macOS 15.6.

**Steps to Reproduce:**

1. Open the application on macOS 15.6.
2. Attempt to log in with a workspace account.
3. Observe the error message: `Failed to login. Message: Precondition check failed.`

**Expected Behavior:**

I expect to be able to log in successfully with my workspace account, as I can on macOS 15.5.

**Environment:**

- macOS version: 15.6 (issue occurs)
- macOS version: 15.5 (works fine)
- Application version: 0.1.15

**Additional Information:**

Please let me know if you need any logs or further details.

Thank you!

Image

gsquared94:
## TLDR

  This pull request addresses a bug that caused authentication failures for free-tier users who had the
  `GOOGLE_CLOUD_PROJECT` environment variable set. The fix ensures that the client-side logic respects the server-managed project for the free tier, preventing a `Precondition Failed` error.

##  Dive Deeper

  For free-tier users, Gemini Code Assist manages the Google Cloud project on the server side. Previously, if a user had
  the `GOOGLE_CLOUD_PROJECT` environment variable set, the client would erroneously send this project ID in the onboardUser request. This conflicts with the server's expectation for the free tier, leading to a 400 Precondition Failed error. This is a recently enforced behavior on the API which is why it was working earlier.

  This patch corrects the behavior by:
   1. Identifying if the user is on the free tier.
   2. If so, it refrains from sending a projectId in the onboardUser request, even if GOOGLE_CLOUD_PROJECT is set.
   3. It then correctly uses the cloudaicompanionProject.id returned in the onboardUser response as the projectId for the
      remainder of the session.
   4. A new unit test has been added to packages/core/src/code_assist/setup.test.ts to cover this specific scenario and
      prevent regressions.

##  Reviewer Test Plan

  To validate this fix, a reviewer can perform the following steps:

   1. Create an OTA on http://rhea.
   2. Set the `GOOGLE_CLOUD_PROJECT` environment variable in your shell to some value
   3. Run the Gemini CLI and attempt to authenticate with the OTA account.
   4. Expected Result: The authentication process should complete successfully without any Precondition Failed errors. The CLI should be ready to use.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ✅  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

- Fixes #5331

### DIFF
--- a/packages/core/src/code_assist/setup.ts
+++ b/packages/core/src/code_assist/setup.ts
@@ -33,38 +33,58 @@ export interface UserData {
  * @returns the user's actual project id
  */
 export async function setupUser(client: OAuth2Client): Promise<UserData> {
-  let projectId = process.env.GOOGLE_CLOUD_PROJECT || undefined;
+  const projectId = process.env.GOOGLE_CLOUD_PROJECT || undefined;
   const caServer = new CodeAssistServer(client, projectId, {}, '', undefined);
-
-  const clientMetadata: ClientMetadata = {
+  const coreClientMetadata: ClientMetadata = {
     ideType: 'IDE_UNSPECIFIED',
     platform: 'PLATFORM_UNSPECIFIED',
     pluginType: 'GEMINI',
-    duetProject: projectId,
   };
 
   const loadRes = await caServer.loadCodeAssist({
     cloudaicompanionProject: projectId,
-    metadata: clientMetadata,
+    metadata: {
+      ...coreClientMetadata,
+      duetProject: projectId,
+    },
   });
 
-  if (!projectId && loadRes.cloudaicompanionProject) {
-    projectId = loadRes.cloudaicompanionProject;
+  if (loadRes.currentTier) {
+    if (!loadRes.cloudaicompanionProject) {
+      if (projectId) {
+        return {
+          projectId,
+          userTier: loadRes.currentTier.id,
+        };
+      }
+      throw new ProjectIdRequiredError();
+    }
+    return {
+      projectId: loadRes.cloudaicompanionProject,
+      userTier: loadRes.currentTier.id,
+    };
   }
 
   const tier = getOnboardTier(loadRes);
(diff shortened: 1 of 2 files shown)


packages/core/src/code_assist/setup.ts
```diff
@@ -54,6 +54,12 @@ export async function setupUser(client: OAuth2Client): Promise<UserData> {
 
   const tier = getOnboardTier(loadRes);
 
+  if (tier.id === UserTierId.FREE) {
```
cornmander: It seems like there are different code paths here depending on whether projectId is set, unset, or if we want to use the free tier managed google project instead. I think this code would be much more clear if you split it up along those lines instead setting projectId multiple times

gsquared94: Good point. I've refactored the method to return early where feasible and avoid variable updates. Also found the bug that onboardUser was being called despite currentTier being set. Fixed that too.

cornmander (approved):

cornmander (approved):

PR #6121, opened 2025-08-13
/chat save command saves empty conversations with only system context

### LINKED ISSUE #6119: /chat save command saves empty conversations with only system context
### Summary
When using gemini-cli, the `/chat save` command saves conversation checkpoints even when there are no actual user messages, only the initial system context messages. This creates empty or meaningless saved conversations that waste storage and clutter the checkpoint list.

### Steps to Reproduce
1. Open gemini-cli in any terminal
2. Start a new chat session (this automatically adds initial system context)
3. Run the save command without any user interaction:
   ╭────────────────────────╮
    │  > /chat save default                               
   ╰────────────────────────╯

Observed: The command successfully saves a "conversation" with tag "default" even though no actual user conversation exists.

4. List saved conversations:
   ╭────────────────────────╮
    │  > /chat list                                            
   ╰────────────────────────╯

Observed: The "default" checkpoint appears in the list, but when loaded, it contains only system context messages (environment info, directory structure, etc.) and no actual user conversation.

### Impact
- Users can accidentally save empty conversations that serve no purpose
- Storage is wasted on meaningless checkpoints
- The checkpoint list becomes cluttered with non-functional saves
- When users try to resume these "conversations," they find no actual content to work with
- This creates confusion about what constitutes a valid conversation to save

The `/chat save` command should only save conversations that contain actual user messages beyond the initial system context. If only system context messages exist, the command should return "No conversation found to save" instead of creating an empty checkpoint.

### Client information

<details>

```console
CLI Version: 0.1.20
Git Commit: 214800cf
Operating System: win32
Sandbox: no sandbox
Model Version: gemini-2.5-pro
Terminals: Any terminal
Auth Method: OAuth
```

</details>

### Login information

OAuth (but I think it affects all authentication methods)

The issue occurs because the save logic checks `history.length > 0` instead of checking for actual user content. The initial system context always consists of exactly 2 messages (user context + model acknowledgment), so checking `history.length > 2` would ensure only conversations with real user interaction are saved.

The resume command already has logic to filter out these system messages for display, but the save command doesn't apply similar filtering before saving.

trycatchkamal:
## TLDR

Fixed a bug in the `/chat save` command where conversations containing only system context messages (initial environment info and model acknowledgment) were being saved as checkpoints. The fix ensures that only conversations with actual user interaction beyond the initial system context are saved, preventing storage waste and clutter from meaningless checkpoints.

## Dive Deeper

The issue was that the `/chat save` command was saving conversation checkpoints even when there were no actual user messages, only the initial system context messages. This happened because:

1. **System Context Always Present**: Every new chat session automatically includes 2 initial messages:
   - A user message with environment context (directory structure, etc.)
   - A model acknowledgment message

2. **Insufficient Filtering**: The original logic checked `history.length > 2`, but this wasn't sufficient to distinguish between:
   - Just system context (2 messages) → should NOT save
   - System context + actual user conversation (3+ messages) → should save

3. **Storage Waste**: Users could accidentally save "conversations" that contained no meaningful content, wasting storage and cluttering the checkpoint list.

The fix implements proper filtering logic that:
- Only saves conversations that contain actual user messages beyond the initial system context
- Returns "No conversation found to save" when only system context exists

This aligns with the existing logic in the `/chat resume` command, which already filters out system messages for display purposes.

## Reviewer Test Plan

1. **Pull and run the code** with the fix applied
2. **Test the save command with minimal conversation**:
   - Start a new chat session (this adds system context automatically)
   - Run `/chat save test-empty` immediately without any user input
   - **Expected**: Should return "No conversation found to save" instead of creating a checkpoint
3. **Test the save command with actual conversation**:
   - Start a new chat session
   - Send a user message like "Hello, how are you?"
   - Get a response from the model
   - Run `/chat save test-real`
   - **Expected**: Should successfully save the conversation
4. **Verify checkpoint content**:
   - Run `/chat list` to see saved checkpoints
   - Run `/chat resume test-real` to verify the conversation loads with actual content
   - **Expected**: Should show the user's message and model's response, not just system context
5. **Test edge cases**:
   - Try saving with various conversation lengths
   - Verify that conversations with only system context are rejected
   - Verify that conversations with actual user interaction are accepted
6. **Integration testing**:
   - Save a meaningful conversation
   - Delete it with `/chat delete`
   - Verify it's removed from the list
   - Try to resume the deleted conversation (should fail appropriately)

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ✅  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs
- Closes #6119 
- Fixes #6119 
- Resolves #6119 

## Addition information
```javascript
~\gemini-cli\packages\cli>npm test -- chatCommand.test.ts

> @google/gemini-cli@0.1.19 test
> vitest run chatCommand.test.ts


 RUN  v3.2.4 ~/gemini-cli/packages/cli
      Coverage enabled with v8

 ✓ src/ui/commands/chatCommand.test.ts (17 tests) 17ms
   ✓ chatCommand > should have the correct main command definition 3ms
   ✓ chatCommand > list subcommand > should inform when no checkpoints are found 1ms
   ✓ chatCommand > list subcommand > should list found checkpoints 1ms
   ✓ chatCommand > list subcommand > should handle invalid date formats gracefully 1ms
   ✓ chatCommand > save subcommand > should return an error if tag is missing 1ms
   ✓ chatCommand > save subcommand > should inform if conversation history is empty or only contains system context 1ms
   ✓ chatCommand > save subcommand > should return confirm_action if checkpoint already exists 2ms
   ✓ chatCommand > save subcommand > should save the conversation if overwrite is confirmed 1ms
   ✓ chatCommand > resume subcommand > should return an error if tag is missing 1ms
   ✓ chatCommand > resume subcommand > should inform if checkpoint is not found 0ms
   ✓ chatCommand > resume subcommand > should resume a conversation 1ms
   ✓ chatCommand > resume subcommand > completion > should provide completion suggestions 1ms
   ✓ chatCommand > resume subcommand > completion > should suggest filenames sorted by modified time (newest first) 1ms
   ✓ chatCommand > delete subcommand > should return an error if tag is missing 1ms
   ✓ chatCommand > delete subcommand > should return an error if checkpoint is not found 1ms
   ✓ chatCommand > delete subcommand > should delete the conversation 1ms
   ✓ chatCommand > delete subcommand > completion > should provide completion suggestions 1ms

 Test Files  1 passed (1)
      Tests  17 passed (17)
   Start at  11:15:02
   Duration  12.21s (transform 190ms, setup 61ms, collect 1.01s, tests 17ms, environment 8.41s, prepare 678ms)

JUNIT report written to ~/gemini-cli/packages/cli/junit.xml
 % Coverage report from v8
```


packages/cli/src/ui/commands/chatCommand.test.ts
```diff
@@ -185,30 +185,32 @@ describe('chatCommand', () => {
       });
     });
 
-    it('should inform if conversation history is empty', async () => {
+    it('should inform if conversation history is empty or only contains system context', async () => {
       mockGetHistory.mockReturnValue([]);
-      const result = await saveCommand?.action?.(mockContext, tag);
+      let result = await saveCommand?.action?.(mockContext, tag);
```
adamfweidman, line 190: nit: rather than reusing the same variable can you create new ones. result1, result2, etc

adamfweidman (approved):

packages/cli/src/ui/commands/chatCommand.ts
```diff
@@ -142,7 +142,7 @@ const saveCommand: SlashCommand = {
     }
 
     const history = chat.getHistory();
-    if (history.length > 0) {
+    if (history.length > 2) {
```
jacob314, line 145: as a follow up, it would it be more robust if chat had a method that determined whether the history had any messages that were actually from the user? It is slightly paranoid but I do worry this constant might change from 2 to some other value in the future if we tweaked how the initial conversation is setup.

jacob314 (approved):
lgtm

packages/cli/src/ui/commands/chatCommand.ts
trycatchkamal: @jacob314 Thanks for the suggestion.

I agree that relying on a hardcoded constant for the minimum history length is fragile, especially if the initial conversation setup changes in the future. I think moving this logic into some function that checks for actual user messages would make the code more robust and maintainable.

It also makes sense to consider refactoring the resume command to use similar logic for consistency. I will take this up and propose a separate refactor PR, making future changes easier to manage.

PR #6140, opened 2025-08-13
fix(ide): Add TMUX support for IDE detection and connection

### LINKED ISSUE #6135: /ide doesn't work properly in tmux
### What happened?

When I am in a tmux session the /ide doesn't work properly 

in tmux session 

Image

without tmux session 

Image

Even though I am in tmux session /ide should work as expected 

### Client information

About Gemini CLI
CLI Version   : 0.1.20
Git Commit    : 214800cf
Model         : gemini-2.5-pro
Sandbox       : No sandbox
OS            : darwin
Auth Method   : gemini-api-key

Extension settings

Name: Gemini CLI Companion
Id: google.gemini-cli-vscode-ide-companion
Description: Enable Gemini CLI with direct access to your VS Code workspace.
Version: 0.1.19
Publisher: Google
VS Marketplace Link: https://marketplace.visualstudio.com/items?itemName=Google.gemini-cli-vscode-ide-companion

### Login information

_No response_

_No response_

nithin-raj-9100:
## TLDR
This pull request adds support for using the /ide command within a tmux session in VS Code. It resolves an issue where the CLI would fail to detect the IDE environment when running inside tmux, by updating the IDE detection logic and ensuring the correct host is used for the IDE server.

## Dive Deeper
The primary issue was that the detectIde function in [packages/core/src/ide/detect-ide.ts](vscode-webview://0l6fk58nnvgvhjd1a1s6sjv6soek51ekph3ueckanlgcscomls71/packages/core/src/ide/detect-ide.ts:1) relied on the TERM_PROGRAM environment variable being set to vscode. When running inside a tmux session, this variable is set to tmux, causing the IDE detection to fail.

This PR introduces a check to see if the TMUX environment variable is set and if TERM_PROGRAM is tmux. If both are true, it sets process.env.TERM_PROGRAM to vscode, allowing the CLI to correctly identify the VS Code environment.

Additionally, the getIdeServerHost function in [packages/core/src/ide/ide-client.ts](vscode-webview://0l6fk58nnvgvhjd1a1s6sjv6soek51ekph3ueckanlgcscomls71/packages/core/src/ide/ide-client.ts:1) has been updated. When TMUX is set, it now returns 127.0.0.1 to ensure a stable connection to the IDE server, which can be an issue when networking through tmux.

These two changes together ensure that the /ide command functions seamlessly for users who work within tmux inside of VS Code.

## Reviewer Test Plan
Open a terminal within VS Code.
Start a tmux session by running tmux.
Inside the tmux session, run the gemini CLI.
Execute the /ide status.
Verify that the command executes successfully and returns 
ℹ 🟢 Connected to VS Code
Exit the tmux session.
Run the /ide command again in the standard VS Code terminal and verify it still works as expected.

## Testing Matrix

I ran npm run preflight everything works as expected ignoring Testig Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs
Fixes #6135


packages/core/src/ide/ide-client.ts
```diff
@@ -362,5 +362,11 @@ export class IdeClient {
 function getIdeServerHost() {
   const isInContainer =
     fs.existsSync('/.dockerenv') || fs.existsSync('/run/.containerenv');
-  return isInContainer ? 'host.docker.internal' : 'localhost';
+  if (isInContainer) {
+    return 'host.docker.internal';
+  }
+  if (process.env.TMUX) {
+    return '127.0.0.1';
```
richieforeman: Isn't this just localhost as well?

skeshive: yup we can use 127.0.0.1 as the default instead of local host

packages/core/src/ide/detect-ide.ts
```diff
@@ -53,7 +53,10 @@ export function getIdeInfo(ide: DetectedIde): IdeInfo {
 
 export function detectIde(): DetectedIde | undefined {
   // Only VSCode-based integrations are currently supported.
-  if (process.env.TERM_PROGRAM !== 'vscode') {
+  const isVsCode = process.env.TERM_PROGRAM === 'vscode';
+  const isVsCodeInTmux =
+    process.env.TMUX && process.env.TERM_PROGRAM === 'tmux';
```
skeshive: in addition to these, can you also check for the existence of other VS Code specific env var so we don't start allowing tmux sessions in standalone terminals or other IDEs?

ex. VSCODE_GIT_ASKPASS_NODE, VSCODE_GIT_IPC_HANDLE

packages/core/src/ide/detect-ide.ts
```diff
@@ -53,7 +53,14 @@ export function getIdeInfo(ide: DetectedIde): IdeInfo {
...
+  const isVsCode = process.env.TERM_PROGRAM === 'vscode';
+  const isVsCodeInTmux =
+    process.env.TMUX &&
+    process.env.TERM_PROGRAM === 'tmux' &&
+    (process.env.VSCODE_GIT_ASKPASS_NODE ||
+     process.env.VSCODE_GIT_ASKPASS_MAIN ||
+      process.env.VSCODE_GIT_IPC_HANDLE);
+  if (!isVsCode && !isVsCodeInTmux) {
     return undefined;
   }
```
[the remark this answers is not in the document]
skeshive: Agreed but unfortunately there are no reliable env vars that get set by VS Code

skeshive (approved):

packages/core/src/ide/detect-ide.ts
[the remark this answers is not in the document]
jacob314: Unfortunately  VSCODE_PID is a hallucination.
I think there is a more robust solution:
`tmux show-environment -g | grep TERM_PROGRAM`
will give you the TERM_PROGRAM of the terminal when tmux was launched.
If that works, it. could be worth doing as the GIT env variables assume the user has the git plugin setup so may have significant false negatives.

jacob314 (commented):
Open to approving now but I think there is a more robust solution we should consider.

jacob314 (commented):
added a comment with an alternate solution to consider.

allenhutchison (changes requested):
Thank you for your contribution, but before we can merge it there are several issues that need to be addressed.

While the updated logic for detecting a `tmux` session within VS Code is a significant improvement, the implementation introduces critical platform compatibility issues that were highlighted in previous reviews but have not been resolved. The heavy reliance on Unix-specific command-line tools (`lsof`, `curl`) will break the functionality for Windows users.

Here is a summary of the issues and suggestions for the contributor:

### Summary of Issues:

*   **Critical: Platform Incompatibility:** The core logic for discovering the active IDE server port in `packages/core/src/ide/tmux-env-sync.ts` relies on `lsof` and `curl`. These commands are not available on Windows, which means this feature will fail on that platform. As this is a cross-platform CLI, all features must work on supported operating systems.
*   **Minor: Debugging Remnants:** The code contains a comment indicating the presence of debugging logic that is intended to be removed later (`// Syncs VS Code environment to TMUX with debugging (will remove once approved)`). This should be cleaned up before merging.
*   **Concern: Fragility:** The port discovery mechanism, while clever, is complex and relies on parsing the output of shell commands. This can be fragile and may break if the output format of `lsof` changes in the future.

### Actionable Changes Required:

1.  **Replace `lsof` with a cross-platform solution.** Instead of actively scanning system processes, a more robust and platform-independent approach would be for the VS Code extension to communicate the port number to the CLI. A common pattern for this is to have the extension write the port to a file in a known temporary directory on startup. The CLI can then simply read this file to get the correct port, eliminating the need for any process scanning.
2.  **Replace `curl` with a Node.js equivalent.** To test the MCP endpoint on a given port, please use Node.js's native `http` module or the global `fetch` API. This will be cross-platform and remove the `curl` dependency.
3.  **Remove the debugging comment** mentioned above once the implementation is finalized.

The goal of supporting `tmux` is valuable, but it must be implemented in a way that does not compromise the experience for users on other platforms.

allenhutchison:
Going to close this for now due to lack of activity. Please reopen if you decide to pursue this further.

PR #6158, opened 2025-08-13
test(integration): add failing test for stdin context with prompt

allenhutchison:
This test demonstrates a bug where content piped to stdin is ignored when the --prompt flag is also used.

The CLI should combine the stdin content with the prompt, but it currently only uses the prompt, causing this test to fail.

The fix for this test is in #4406


NTaylorMullen (approved):

packages/cli/src/utils/sandbox.ts
```diff
@@ -166,7 +166,20 @@ function entrypoint(workdir: string): string[] {
...
 
-  const cliArgs = process.argv.slice(2).map((arg) => quote([arg]));
+  const cliArgs = process.argv.slice(2);
+  if (stdinData) {
+    const promptIndex = cliArgs.findIndex(
+      (arg) => arg === '--prompt' || arg === '-p',
+    );
+    if (promptIndex > -1 && cliArgs.length > promptIndex + 1) {
+      // If there's a prompt argument, prepend stdin to it
+      cliArgs[promptIndex + 1] = `${stdinData}\n\n${cliArgs[promptIndex + 1]}`;
```
NTaylorMullen: clever

packages/cli/src/utils/sandbox.ts
```diff
@@ -263,6 +276,21 @@ export async function start_sandbox(
         args.push('-D', `INCLUDE_DIR_${i}=${dirPath}`);
       }
 
+      const finalArgv = [...process.argv];
+      if (stdinData) {
+        const promptIndex = finalArgv.findIndex(
```
NTaylorMullen: Digging through arguments feels a bit hacky for this. Is there a better way to get the user's question?

Also, setting the prompt arg below feels brittle if we change the argument too

packages/cli/src/utils/sandbox.ts
```diff
@@ -166,7 +166,20 @@ function entrypoint(workdir: string): string[] {
     ),
   );
 
-  const cliArgs = process.argv.slice(2).map((arg) => quote([arg]));
+  const cliArgs = process.argv.slice(2);
+  if (stdinData) {
```
NTaylorMullen: Does this whole block need to be replicated with bits below?

NTaylorMullen (approved):

packages/cli/src/utils/sandbox.ts
allenhutchison: I've pulled this into a helper function in gemini.tsx so it's no longer repeated here. For history we do need the logic in both places because one is for the mac os sandbox and the other is for docker / podman.

packages/cli/src/utils/sandbox.ts
allenhutchison: Fair. I don't think there is a way around dealing with the --prompt string, but what I can do is move this logic up to the geminin.tsx file which makes it more obvious that we are messing around with a command line flag.

PR #6023, opened 2025-08-12
chore(cli/slashcommands): Add status enum to SlashCommandEvent telemetry

richieforeman:
> NOTE: DEPENDS ON cl/793846401 from Internal

## TLDR

- Adds a status enum to `SlashCommandEvent` and wires it up in the `slashCommandProcessor.ts`
- Prefer an interface based approach for defining events rather than classes 
 
> Opinion: I find class objects like that very hard to maintain and work with.  By preferring interfaces and object bags, I find the usage code a bit more verbose, but way way way more readable.  If there's interest in this approach, I don't mind sending a refactor around this pattern.

## Dive Deeper

- Some drive by cleanups and tidy ups, I get very picky about types, and I like them very much :)

## Reviewer Test Plan

- Slash Commands behave as normal
- Slash commands that work will log a `success` event
- Slash commands that throw errors or choke for whatever reason will log an `error` event.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| -------- | --- | --- | --- |
| npm run  | x  | ❓  | x  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

- Related to #3709 (which I am unable to dupe, and is very likely fixed).

Issue #3709, opened 2025-07-09
Failed to compress chat history


skun23:
### What happened?

Failed to compress chat history: Please ensure that the number of  function response parts is equal to the number of function call  parts of the function call turn.
[API Error: [{
    "error": {
      "code": 400,
      "message": "Please ensure that the number of function response 
  parts is equal to the number of function call parts of the function 
  call turn.",
      "errors": [
        {
          "message": "Please ensure that the number of function 
  response parts is equal to the number of function call parts of the 
  function call turn.",
          "domain": "global",
          "reason": "badRequest"
        }
      ],
      "status": "INVALID_ARGUMENT"
    }
  }
  ]]

### What did you expect to happen?

 /compress 
compress chat history

### Client information

*   **CLI Version:** 0.1.9
*   **Git Commit:** 34935d6
*   **Operating System:** win32 v22.17.0
*   **Sandbox Environment:** no sandbox
*   **Model Version:** gemini-2.5-pro
*   **Memory Usage:** 158.2 MB

corneliusroemer:
The code above is pretty badly formatted, here it is with extra context from my duplicate #4568 

```
ℹ IMPORTANT: This conversation approached the input token limit for gemini-2.5-flash. A compressed context will be sent for future messages (compressed from: 741653 to 90649 tokens).

✕ [API Error: [{
    "error": {
      "code": 400,
      "message": "Please ensure that the number of function response parts is equal to the number of function call parts of the function call turn.",
      "errors": [
        {
          "message": "Please ensure that the number of function response parts is equal to the number of function call parts of the function call turn.",
          "domain": "global",
          "reason": "badRequest"
        }
      ],
      "status": "INVALID_ARGUMENT"
    }
  }
  ]]


╭───────────────────────────────╮
│  > coll but let's keep going  │
╰───────────────────────────────╯

✕ [API Error: [{
    "error": {
      "code": 400,
      "message": "Please ensure that the number of function response parts is equal to the number of function call parts of the function call turn.",
      "errors": [
        {
          "message": "Please ensure that the number of function response parts is equal to the number of function call parts of the function call turn.",
          "domain": "global",
          "reason": "badRequest"
        }
      ],
      "status": "INVALID_ARGUMENT"
    }
  }
  ]]


╭─────────────────╮
│  > ok let's go  │
╰─────────────────╯

✕ [API Error: [{
    "error": {
      "code": 400,
[...]

PR #6166, opened 2025-08-13
chore(cli/slashcommands): Add status enum to SlashCommandEvent telemetry

richieforeman:
Rehash of #6023 

> NOTE: DEPENDS ON cl/793846401 from Internal

## TLDR

- Adds a status enum to `SlashCommandEvent` and wires it up in the `slashCommandProcessor.ts`
- Prefer an interface based approach for defining events rather than classes 
 
> Opinion: I find class objects like that very hard to maintain and work with.  By preferring interfaces and object bags, I find the usage code a bit more verbose, but way way way more readable.  If there's interest in this approach, I don't mind sending a refactor around this pattern.

## Dive Deeper

- Some drive by cleanups and tidy ups, I get very picky about types, and I like them very much :)

## Reviewer Test Plan

- Slash Commands behave as normal
- Slash commands that work will log a `success` event
- Slash commands that throw errors or choke for whatever reason will log an `error` event.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | x  | ❓  | x  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

- Related to #3709 (which I am unable to dupe, and is very likely fixed).


jacob314 (approved):
lgtm

PR #6177, opened 2025-08-13
[ide-mode] Support multi-folder workspaces

chrstnb:
## TLDR

Support workspaces with multiple open folders by ensuring that gemini CLI is running on one of the open folders.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


packages/core/src/ide/ide-client.ts
```diff
@@ -238,22 +238,32 @@ export class IdeClient {
...
         IDEConnectionStatus.Disconnected,
-        `To use this feature, please open a single workspace folder in ${this.currentIdeDisplayName} and try again.`,
+        `To use this feature, please open a workspace folder in ${this.currentIdeDisplayName} and try again.`,
         true,
       );
       return false;
     }
 
-    const idePath = getRealPath(ideWorkspacePath).toLocaleLowerCase();
+    const ideWorkspacePaths = ideWorkspacePath.split(':');
```
skeshive, line 275: I'm assuming this is backwards compatible when the path just contains a single dir?

skeshive (approved):

packages/core/src/ide/ide-client.ts
chrstnb: yes! tested with that and it works!

packages/core/src/ide/ide-client.ts
```diff
@@ -238,22 +238,32 @@ export class IdeClient {
...
     const cwd = getRealPath(process.cwd()).toLocaleLowerCase();
-    const rel = path.relative(idePath, cwd);
-    if (rel.startsWith('..') || path.isAbsolute(rel)) {
+    const isWithinWorkspace = ideWorkspacePaths.some((workspacePath) => {
+      const idePath = getRealPath(workspacePath).toLocaleLowerCase();
```
jacob314: why the toLocaleLowerCase()

This logic has also been duplicated in at least 3 places in Gemini CLI so think it is worth adding a standard path helper that determines whether paths are contained within each other. such a helper may already exist in the project.

packages/core/src/ide/ide-client.ts
```diff
@@ -238,22 +238,31 @@ export class IdeClient {
...
         true,
       );
       return false;
     }
 
-    const idePath = getRealPath(ideWorkspacePath).toLocaleLowerCase();
-    const cwd = getRealPath(process.cwd()).toLocaleLowerCase();
-    const rel = path.relative(idePath, cwd);
-    if (rel.startsWith('..') || path.isAbsolute(rel)) {
+    const ideWorkspacePaths = ideWorkspacePath.split(':');
```
jacob314, line 275: please add a test for the changes to ide-client.ts as well.

packages/core/src/utils/paths.ts
```diff
@@ -200,3 +200,28 @@ export function getUserCommandsDir(): string {
...
 }
+
+/**
+ * Checks if a path is a subpath of another path.
+ * @param parentPath The parent path.
+ * @param childPath The child path.
+ * @returns True if childPath is a subpath of parentPath, false otherwise.
+ */
+export function isSubpath(parentPath: string, childPath: string): boolean {
+  // Heuristic to detect Windows paths even on POSIX systems
```
jacob314: is this heuristic for windows paths needed? Concerned we will make this code fragile if we sometimes require this heuristic and sometimes don't. would be preferred if we detect that the operating system is windows and special case based on that rather than pattern matching specific paths.

jacob314 (commented):
thanks for t

packages/core/src/ide/ide-client.test.ts
```diff
@@ -0,0 +1,131 @@
...
+  afterEach(() => {
+    process.env = originalEnv;
+    process.cwd = originalCwd;
+  });
+
+  it('should connect if cwd is within the IDE workspace path', async () => {
+    process.env['GEMINI_CLI_IDE_WORKSPACE_PATH'] = '/Users/person/gemini-cli';
+    process.env['GEMINI_CLI_IDE_SERVER_PORT'] = '12345';
+
+    vi.doMock('./detect-ide.js', () => ({
```
jacob314: Gemini CLI likes to mock too much. Please get it to refactor these tests to not mock the path utils or detect-ide.js unless that is strictly required.
When Gemini CLI mocks this much it is really hard to tell whether the tests are testing anything at all.

packages/core/src/ide/ide-client.test.ts
```diff
@@ -0,0 +1,131 @@
...
+      detectIde: () => 'vscode',
+      getIdeInfo: () => ({ displayName: 'VS Code' }),
+      DetectedIde: { VSCODE: 'vscode' },
+    }));
+    vi.doMock('../utils/paths.js', () => ({
+      isSubpath: () => true,
+    }));
+
+    const { IdeClient, IDEConnectionStatus } = await import('./ide-client.js');
```
jacob314: avoid inline imports like this and use the standard mechanisms for imports.

packages/vscode-ide-companion/src/extension.ts
```diff
@@ -20,11 +20,13 @@ let log: (message: string) => void = () => {};
 
 function updateWorkspacePath(context: vscode.ExtensionContext) {
   const workspaceFolders = vscode.workspace.workspaceFolders;
-  if (workspaceFolders && workspaceFolders.length === 1) {
-    const workspaceFolder = workspaceFolders[0];
+  if (workspaceFolders && workspaceFolders.length > 0) {
+    const workspacePaths = workspaceFolders
+      .map((folder) => folder.uri.fsPath)
+      .join(':');
```
sunnyps, line 26: FWIW this isn't the convention on Windows which uses `;` for path delimiters - you can use `path.delimiter` to join the paths in an OS agnostic way.

sunnyps:
I also started working on this last night in https://github.com/google-gemini/gemini-cli/pull/6208 before I could see this PR. Sorry!

jacob314 (approved):
lgtm

Thanks for cleaning up the tests!

packages/core/src/ide/ide-client.ts
```diff
@@ -245,37 +253,41 @@ export class IdeClient {
...
-    const cwd = getRealPath(process.cwd()).toLocaleLowerCase();
-    const rel = path.relative(idePath, cwd);
-    if (rel.startsWith('..') || path.isAbsolute(rel)) {
-      this.setState(
-        IDEConnectionStatus.Disconnected,
-        `Directory mismatch. Gemini CLI is running in a different location than the open workspace in ${this.currentIdeDisplayName}. Please run the CLI from the same directory as your project's root folder.`,
-        true,
-      );
-      return false;
+    const ideWorkspacePaths = ideWorkspacePath.split(':');
```
sunnyps, line 275: path.delimiter here too if you choose to do that in extension.ts

PR #6185, opened 2025-08-13
refactor: remove modelCheck feature

allenhutchison:
This PR removes the modelCheck feature. This was originally created because there was no free tier for the Pro model during its preview. Since PR #2526 inadvertently removed the ability to set the fallback model and we have not seen any issues, this feature is no longer needed.


jacob314 (approved):
lgtm
Yay for PRs that delete much more code than they add.

PR #6194, opened 2025-08-14
bug(mcp): catch errors reported by GitHub MCP

leehagoodjames:
## TLDR

MCP errors surfaced via `isError:true` are currently considered a tool call success. This change marks them as a failure.

## Reviewer Test Plan

Validate that failed tool calls that surface a response (`isError:true` exists in that response) should be recorded as a failure inline when tool is called and in the `Interaction Summary` in the `Tool Calls` section.

#### Current behavior showing a failed call being interpreted as a success
image

#### New behavior showing a failed call being interpreted as a failure
image

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

- Related to #5306

### DIFF
--- a/packages/core/src/tools/mcp-tool.test.ts
+++ b/packages/core/src/tools/mcp-tool.test.ts
@@ -185,6 +185,98 @@ describe('DiscoveredMCPTool', () => {
       ).rejects.toThrow(expectedError);
     });
 
+    it.each([
+      { isErrorValue: true, description: 'true (bool)' },
+      { isErrorValue: 'true', description: '"true" (str)' },
+    ])(
+      'should consider a ToolResult with isError $description to be a failure',
+      async ({ isErrorValue }) => {
+        const tool = new DiscoveredMCPTool(
+          mockCallableToolInstance,
+          serverName,
+          serverToolName,
+          baseDescription,
+          inputSchema,
+        );
+        const params = { param: 'isErrorTrueCase' };
+
+        const errorResponse = { isError: isErrorValue };
+        const mockMcpToolResponseParts: Part[] = [
+          {
+            functionResponse: {
+              name: serverToolName,
+              response: { error: errorResponse },
+            },
+          },
+        ];
+        mockCallTool.mockResolvedValue(mockMcpToolResponseParts);
+        const expectedError = new Error(
+          `MCP tool '${serverToolName}' reported tool error with response: ${JSON.stringify(
+            mockMcpToolResponseParts,
+          )}`,
+        );
+
+        const invocation = tool.build(params);
+        await expect(
+          invocation.execute(new AbortController().signal),
+        ).rejects.toThrow(expectedError);
+      },
+    );
+
+    it.each([
+      { isErrorValue: false, description: 'false (bool)' },
+      { isErrorValue: 'false', description: '"false" (str)' },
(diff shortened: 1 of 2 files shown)


packages/core/src/tools/mcp-tool.ts
```diff
@@ -122,6 +122,29 @@ export class DiscoveredMCPTool extends BaseTool<ToolParams, ToolResult> {
...
+    const functionResponse = rawResponseParts?.[0]?.functionResponse;
+    const response = functionResponse?.response;
+
+    interface McpError {
+      isError?: boolean;
+    }
+
+    if (response) {
+      const error = (response as { error?: McpError })?.error;
+      const isError = error?.isError;
```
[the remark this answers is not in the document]
leehagoodjames: Updated if to check for `(isError === true || isError === 'true')` below

packages/core/src/tools/mcp-tool.test.ts
```diff
@@ -185,6 +185,158 @@ describe('DiscoveredMCPTool', () => {
       ).rejects.toThrow(expectedError);
     });
 
+    it('should consider a ToolResult with isError true (bool) to be a failure', async () => {
```
sethvargo: Nit: these tests are almost all identical. I wonder if we could use a table-driven test instead:

```js
for (const case of {param}) {
  it(`should consider a ToolResult with isError as ${value}`, async() => {

  })
}
```

Different test framework, but here's an example: https://github.com/google-github-actions/actions-utils/blob/23f8dc3b60e81bddb05090a393cd64dfea2a9289/tests/fs.test.ts#L56

packages/core/src/tools/mcp-tool.ts
```diff
@@ -104,6 +104,29 @@ class DiscoveredMCPToolInvocation extends BaseToolInvocation<
     return confirmationDetails;
   }
 
+  // Validate that the response does not contain errors
+  // This is needed because CallToolResults should return errors inside the response.
+  // ref: https://modelcontextprotocol.io/specification/2025-06-18/schema#calltoolresult
+  private _validateNotError(rawResponseParts: Part[]): void {
```
sethvargo: I'm not sure I agree this is the right API. I think we want a public `isMCPToolError` function that returns a boolean. Then the caller can decide whether to throw an error, log a message, etc.

leehagoodjames: Updated to a public `isMCPToolError` method that returns a `boolean`

packages/core/src/tools/mcp-tool.test.ts
leehagoodjames: I added a table for the failures and a table for the success, is that sufficient? It seemed more readable than all of them in the same table - and then adding checks to only include some commands based on it being a failure vs success

sethvargo (approved):

JeromeJu (approved):
lgtm

bdmorgan (approved):
LGTM

PR #6195, opened 2025-08-14
feat(client/compression): Log telemetry when compressing chat context.

richieforeman:
## TLDR

- Logs telemtry when sending chat context, includes tokens before/after.
- Further tighten some typing in `ClearcutLogger` (drop the `object` type there), update tests for the fall out from that.

## Dive Deeper

- There were a bunch of constants in ClearcutLogger that felt very duplicative.  I purged that while I was here and refactored things lightly to remove the need for them.  These were only used internally in the file. I checked all of the events, and all of the names/titles match.

## Reviewer Test Plan

N/A

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | x  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Resolves #4658


packages/core/src/telemetry/integration.test.circular.ts
```diff
@@ -56,7 +56,7 @@ describe('Circular Reference Integration Test', () => {
     const logger = ClearcutLogger.getInstance(mockConfig);
 
     expect(() => {
-      logger?.enqueueLogEvent(problematicEvent);
+      logger?.enqueueLogEvent(problematicEvent as any);
```
richieforeman, line 60: NOTE: with the tighter typing here on this, this is now technically impossible since the interface types can no longer contain arbitrary circular references.  

I feel like this test was closing some sort of bug or something, so I just opted for an any cast here.

I personally feel like this test could/should be removed now that the typing is tighter, but I don't feel strongly.  I do believe that good types and interfaces can help prevent strange explosions like this.

packages/core/src/telemetry/clearcut-logger/clearcut-logger.ts
```diff
@@ -167,7 +154,7 @@ export class ClearcutLogger {
     ClearcutLogger.instance = undefined;
   }
 
-  enqueueLogEvent(event: object): void {
+  enqueueLogEvent(event: LogEvent): void {
```
richieforeman, line 175: This change made the circular object thing an impossible state.

packages/core/src/telemetry/types.ts
```diff
@@ -317,6 +317,25 @@ export enum SlashCommandStatus {
   ERROR = 'error',
 }
 
+export interface ChatCompressionEvent extends BaseTelemetryEvent {
```
jerop, line 320: please update the data reference with this new event: https://github.com/google-gemini/gemini-cli/blob/main/docs/telemetry.md#logs-and-metric-reference

jerop (dismissed):

packages/core/src/telemetry/clearcut-logger/clearcut-logger.ts
jerop, line 1: @uttamkanodia14 please review changes to clearcut

Excellencedev:
You guys should add Algora bounties to issues

packages/core/src/telemetry/types.ts
richieforeman: DONE

anj-s:
@owenofbrien can you take a look as well

packages/core/src/telemetry/integration.test.circular.ts
owenofbrien: +1, the tighter typing should help with this kind of thing. Thanks for improving the typing!

owenofbrien (approved):
Thanks for leaving the code better than you found it!

packages/core/src/telemetry/clearcut-logger/clearcut-logger.ts
```diff
@@ -386,7 +373,7 @@ export class ClearcutLogger {
     ];
 
     // Flush start event immediately
-    this.enqueueLogEvent(this.createLogEvent(start_session_event_name, data));
```
owenofbrien, line 389: Unfortunately, the event.name for StartSessionEvent is 'cli_config' which is used for OTEL logging. Changing this will break Dashboards and other Concord log clients. I like the code simplification you're doing here, but we need to find a way to do it without breaking log clients.

owenofbrien (changes requested):

packages/core/src/telemetry/clearcut-logger/clearcut-logger.ts
richieforeman: Drat! You're right. I will backpedal this part of the change and ensure that a test is added for this.

richieforeman: I re-worked this PR.  I wish that these were same-same, but I tightened up the typing in a way that I feel adds value. PTAL.

packages/core/src/core/client.ts
```diff
@@ -891,3 +906,8 @@ export class GeminiClient {
     return null;
   }
 }
+
+export const TEST_ONLY = {
+  COMPRESSION_PRESERVE_THRESHOLD,
+  COMPRESSION_TOKEN_THRESHOLD,
+};
```
jerop, line 913: for my learning, is this a common pattern for passing env vars for tests?

jerop (approved):

packages/core/src/core/client.ts
richieforeman: It's a common pattern internally.  Do a code search for "const TEST_ONLY lang:js|ts".

It's a nice way to export "private things" that are useful for your test, but not necessarily part of your module's "public api", per se.

jerop: thanks for the explanation!

owenofbrien (approved):

PR #6197, opened 2025-08-14
feat(core): Migrate MockTools to declarative pattern.

joshualitt:
## TLDR

Migrates MockTools to the declarative pattern, and deletes the deprecated and now unused migration helper classes.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


packages/cli/src/ui/hooks/useToolScheduler.test.ts
```diff
@@ -55,7 +57,33 @@ const mockConfig = {
   getDebugMode: () => false,
 };
 
-class MockTool extends BaseTool<object, ToolResult> {
+class MockToolInvocation extends BaseToolInvocation<object, ToolResult> {
```
jacob314, line 67: can you delete and use the version from test-utils/tools.ts instead?

packages/core/src/core/coreToolScheduler.test.ts
```diff
@@ -29,6 +31,50 @@ import {
 } from '../tools/modifiable-tool.js';
 import { MockTool } from '../test-utils/tools.js';
 
+class MockModifiableToolInvocation extends BaseToolInvocation<
```
jacob314: nit: move this to test-utils?

packages/core/src/core/coreToolScheduler.ts
```diff
@@ -598,6 +598,14 @@ export class CoreToolScheduler {
...
             this.setStatusInternal(reqInfo.callId, 'scheduled');
           } else {
+            if (signal.aborted) {
+              this.setStatusInternal(
+                reqInfo.callId,
+                'cancelled',
+                'Tool call cancelled by user.',
+              );
+              continue;
+            }
```
joshualitt: This logic was needed to pass the test for aborting before `shouldConfirmExecute`.  The test before this change would manually call `handleConfirmationResponse` and somehow that made the test pass falsely.

packages/core/src/core/coreToolScheduler.ts
```diff
@@ -601,6 +601,14 @@ export class CoreToolScheduler {
             );
             this.setStatusInternal(reqInfo.callId, 'scheduled');
           } else {
+            if (signal.aborted) {
```
jacob314: perhaps bad idea but should this be moved to before the if for yolo
Even if in yolo perhaps in the future you could have signaled an abort (e.g. press escape) and after that everything in flight that can be cancelled should be cancelled.

packages/cli/src/ui/hooks/useToolScheduler.test.ts
joshualitt: Hmm, those mocks live in `core`. There is a `test-utils` package, but we'd have to have a circular dependency on `core` to share the mocks between `cli` and `core`. We could export it from `core` behind namespace.

Thoughts or alternatives?

jacob314 (approved):
lgtm

PR #6199, opened 2025-08-14
feat(core): Cleanup after migrating tools.

joshualitt:
## TLDR

Moves the core validation logic to `BaseDeclarativeTool`, simplifying the concrete tools. This is the final CL in the series.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

### DIFF
--- a/packages/core/src/telemetry/uiTelemetry.test.ts
+++ b/packages/core/src/telemetry/uiTelemetry.test.ts
@@ -43,7 +43,7 @@ const createFakeCompletedToolCall = (
       status: 'success',
       request,
       tool,
-      invocation: tool.build({}),
+      invocation: tool.build({ param: 'test' }),
       response: {
         callId: request.callId,
         responseParts: {
--- a/packages/core/src/test-utils/tools.ts
+++ b/packages/core/src/test-utils/tools.ts
@@ -13,7 +13,6 @@ import {
   ToolResult,
   Kind,
 } from '../tools/tools.js';
-import { Schema, Type } from '@google/genai';
 import {
   ModifiableDeclarativeTool,
   ModifyContext,
@@ -74,9 +73,9 @@ export class MockTool extends BaseDeclarativeTool<
     name = 'mock-tool',
     displayName?: string,
     description = 'A mock tool for testing.',
-    params: Schema = {
-      type: Type.OBJECT,
-      properties: { param: { type: Type.STRING } },
+    params = {
+      type: 'object',
+      properties: { param: { type: 'string' } },
     },
   ) {
     super(name, displayName ?? name, description, Kind.Other, params);
--- a/packages/core/src/tools/edit.ts
+++ b/packages/core/src/tools/edit.ts
@@ -19,7 +19,6 @@ import {
   ToolResultDisplay,
 } from './tools.js';
 import { ToolErrorType } from './tool-error.js';
-import { SchemaValidator } from '../utils/schemaValidator.js';
 import { makeRelative, shortenPath } from '../utils/paths.js';
 import { isNodeError } from '../utils/errors.js';
 import { Config, ApprovalMode } from '../config/config.js';
@@ -475,13 +474,11 @@ Expectation for required parameters:
    * @param params Parameters to validate
    * @returns Error message string or null if valid
    */
-  override validateToolParams(params: EditToolParams): string | null {
-    const errors = SchemaValidator.validate(
-      this.schema.parametersJsonSchema,
-      params,
-    );
-    if (errors) {
-      return errors;
+  protected override validateToolParamValues(
+    params: EditToolParams,
+  ): string | null {
+    if (!params.file_path) {
+      return "The 'file_path' parameter must be non-empty.";
     }
 
     if (!path.isAbsolute(params.file_path)) {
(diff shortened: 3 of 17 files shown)


jacob314:
Resolve conflicts and I'll give this a final review. Hopefully the conflicts are just with your own changes.

packages/core/src/test-utils/tools.ts
```diff
@@ -74,9 +73,9 @@ export class MockTool extends BaseDeclarativeTool<
     name = 'mock-tool',
     displayName?: string,
     description = 'A mock tool for testing.',
-    params: Schema = {
-      type: Type.OBJECT,
-      properties: { param: { type: Type.STRING } },
+    params = {
```
jacob314, line 76: was this change intentional? seems like a regression.

packages/core/src/telemetry/uiTelemetry.test.ts
```diff
@@ -43,7 +43,7 @@ const createFakeCompletedToolCall = (
       status: 'success',
       request,
       tool,
-      invocation: tool.build({}),
```
jacob314, line 46: seems unrelated. revert if so.

packages/core/src/tools/tools.ts
```diff
@@ -231,6 +232,23 @@ export abstract class BaseDeclarativeTool<
...
+      params,
+    );
+
+    if (errors) {
+      return errors;
+    }
+    return this.validateToolParamValues(params);
+  }
+
+  protected validateToolParamValues(_params: TParams): string | null {
```
jacob314, line 294: nit: why this pattern instead of requireing tools to call 
super.validateToolParams?

With this there is a risk someone will accidentally override validateToolParams instead causing problems.  Not sure if there is a lint like must call super in Dart to make that easy to enforce.

packages/core/src/tools/edit.ts
```diff
@@ -471,13 +470,11 @@ Expectation for required parameters:
...
-    const errors = SchemaValidator.validate(
-      this.schema.parametersJsonSchema,
-      params,
-    );
-    if (errors) {
-      return errors;
+  protected override validateToolParamValues(
+    params: EditToolParams,
+  ): string | null {
+    if (!params.file_path) {
```
jacob314, line 480: this check seems net new but reasonable. was it intentionally added?
if so, please add a test.

packages/core/src/tools/read-file.ts
```diff
@@ -198,18 +197,14 @@ export class ReadFileTool extends BaseDeclarativeTool<
...
     params: ReadFileToolParams,
   ): string | null {
-    const errors = SchemaValidator.validate(
-      this.schema.parametersJsonSchema,
-      params,
-    );
-    if (errors) {
-      return errors;
+    const filePath = params.absolute_path;
+    if (params.absolute_path.trim() === '') {
```
jacob314, line 205: this check seems net new but reasonable. If so, please add a test.

packages/core/src/tools/tools.ts
joshualitt: We have `noImplicitOverride` setup in our tsconfig, so overriding it unintentionally isn't completely trivial?

I'm not sure if there is a lint to call `super`, I can investigate this. Would you prefer it over the existing logic with `noImplicitOverride`?

packages/core/src/telemetry/uiTelemetry.test.ts
joshualitt: `tool.build({})` validates parameters now, and the mock tool expects one parameter. However, we can make this configurable. Would you prefer that?

packages/core/src/tools/tools.ts
jacob314: go ahead and ignore me. sadly this doesn't exist in typescript so what you did is probably the better typescript pattern.
https://github.com/eslint/eslint/issues/11969

packages/core/src/telemetry/uiTelemetry.test.ts
jacob314: what you have is fine. just double checking suspicious looking lines.

jacob314 (commented):
looks good after final two comments about net new functionality are addressed.

packages/core/src/test-utils/tools.ts
joshualitt: This was a rabbit hole, but I've figured it out. Yes, this was intentional, no it is not a regression.

In https://github.com/google-gemini/gemini-cli/pull/4814, we away from the `Schema` type and towards the official JSON Schema type, i.e. what the AJV validator supports. Without this change, the validator returns:
```
Error: schema is invalid: data/properties/param/type must be equal to one of the allowed values, data/properties/param/type must be array, data/properties/param/type must match a schema in anyOf
```

Anyway, the AJV Validator has its own type, and we should now be using this high quality type everywhere. However, changing the type will not be trivial, so we should probably tackle it in a follow-up if possible.

jacob314 (approved):
lgtm

Issue #6165, opened 2025-08-13
Flaky test on windows


jacob314:
### What happened?

I see the following test periodically flake on Windows.

Example: https://github.com/google-gemini/gemini-cli/actions/runs/16945361812/job/48024980623
 FAIL  src/ui/components/SettingsDialog.test.tsx > SettingsDialog > Scope Selection > should reset to settings focus when scope is selected
AssertionError: expected '╭────────────────────────────────────…' to contain '  Apply To'

### What did you expect to happen?

Test should not be flaky

### Client information

<details>

```console
$ gemini /about
# paste output here
```

</details>

jacob314:
Fyi @ali-aljufairi

PR #5984, opened 2025-08-11
docs(commands): add /settings command for user-friendly settings editing

ali-aljufairi:
## TLDR

add /setting to docs 

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| -------- | --- | --- | --- |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

PR #6202, opened 2025-08-14
feat(settings): enhance settings management with generic setter and display hel…

ali-aljufairi:
## TLDR

I added to options where users can now change 2 number filed and I fixed the problem mention by @jacob314  about this 

https://github.com/google-gemini/gemini-cli/pull/4738#discussion_r2256072683
  ```js
   const [modifiedValues, setModifiedValues] = useState<Map<string, boolean>>(
  ```
I followed  the  tip  given in pervious PR

Fixed this bug
 
 this has a default value of false, and you change it to true then back to false, the key now exists in the JSON file with value false, so it shows false* even though false is the default. The * should only appear when the value is actually different from the default, not just when the key exists.

Also fixed the test I think the problem is due to react no full render that why its not failling all the times but sometimes so I added delay 

https://github.com/user-attachments/assets/a35b6dd2-bd69-466e-929c-52b1e129b07d

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | x | ❓  | x  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | x  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

#5984
#6165


packages/cli/src/ui/components/SettingsDialog.test.tsx
```diff
@@ -292,8 +292,13 @@ describe('SettingsDialog', () => {
       stdin.write('1'); // Select first scope option
       await wait();
 
-      // Should be back to settings focus
-      expect(lastFrame()).toContain('  Apply To');
+      // Wait a bit more to ensure React state has updated
```
jacob314: Rather than adding magic time intervals which might still flake on slow machines running tests and slow down tests for all users, write:
await waitFor(() => {
  const output = lastFrame();
  expect(output).not.toContain('> Apply To');
});
which will continue as soon as the UI is in the expected state.

packages/cli/src/ui/components/SettingsDialog.test.tsx
```diff
@@ -669,8 +676,10 @@ describe('SettingsDialog', () => {
 
       // Tab back to settings section
       stdin.write('\t');
-      await wait();
-      expect(lastFrame()).toContain('  Apply To');
+      await wait(100); // Extra wait for state update
```
jacob314: go ahead and revert this. Sent a pr for the more robust fix of using waitFor instead of hard coding time intervals into text. This is better as it keeps the tests fast and avoids possible race conditions for tests running on very slow machines in CI.

packages/cli/src/utils/settingsUtils.ts
```diff
@@ -332,6 +335,20 @@ export function setPendingSettingValue(
...
+/**
+ * Generic setter: Set a setting value (boolean, number, string, etc.) in the pending settings
+ */
+export function setPendingSettingValueAny(
+  key: string,
+  value: unknown,
+  pendingSettings: Settings,
+): Settings {
+  const path = key.split('.');
+  const newSettings = JSON.parse(JSON.stringify(pendingSettings));
```
jacob314: Use structuredClone to clone JSON rather than parse + stringify.

packages/cli/src/utils/settingsUtils.ts
```diff
@@ -431,11 +446,9 @@ export function getDisplayValue(
   const isChangedFromDefault =
     typeof defaultValue === 'boolean' ? value !== defaultValue : value === true;
   const isInModifiedSettings = modifiedSettings.has(key);
-  const hasPendingChanges =
-    pendingSettings && settingExistsInScope(key, pendingSettings);
 
-  // Add * indicator when value differs from default, is in modified settings, or has pending changes
-  if (isChangedFromDefault || isInModifiedSettings || hasPendingChanges) {
+  // Only show * if value actually differs from default, not just if key exists
```
jacob314: is that right? a value that matches the default is still crucial to show if it is a workspace level setting that overrides the value in the user settings.

packages/cli/src/ui/components/SettingsDialog.tsx
```diff
@@ -385,13 +488,50 @@ export function SettingsDialog({
...
-            scopeSettings,
-            mergedSettings,
-            modifiedSettings,
-            pendingSettings,
-          );
+
+          let displayValue: string;
+          if (editingKey === item.value) {
+            // Show edit buffer with cursor
+            const cursorChar = cursorVisible ? '▌' : ' ';
```
jacob314: take a page from inputPrompt which uses some clever tricks to invert a char to show the cursor which allows future support for using left and right arrows to edit earlier text in the future.

jacob314 (commented):
Excited for this! A few minor comments. Sorry for the slow review. Busy worrying about some input related issues today.

packages/cli/src/ui/components/SettingsDialog.test.tsx
ali-aljufairi: merged with main and went with this okay thanks for insight

packages/cli/src/ui/components/SettingsDialog.test.tsx
ali-aljufairi: Done

packages/cli/src/utils/settingsUtils.ts
ali-aljufairi: okay

packages/cli/src/utils/settingsUtils.ts
```diff
@@ -431,11 +446,12 @@ export function getDisplayValue(
...
 
-  // Add * indicator when value differs from default, is in modified settings, or has pending changes
-  if (isChangedFromDefault || isInModifiedSettings || hasPendingChanges) {
+  // Mark as modified if setting exists in current scope OR is in modified settings
+  if (settingExistsInScope(key, settings) || isInModifiedSettings) {
+    return `${valueString}*`; // * indicates setting is set in current scope
+  }
+  if (isChangedFromDefault || isInModifiedSettings) {
     return `${valueString}*`; // * indicates changed from default value
   }
```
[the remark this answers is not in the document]
ali-aljufairi: No but user need to know that key existe in current scope

jacob314 (approved):
lgtm

PR #6208, opened 2025-08-14
feat: VSCode multi-root workspace support

sunnyps:
## TLDR

The primary goal is to allow developers to use the Gemini CLI seamlessly when they have multiple project folders open in a single VSCode window (a "multi-root workspace").

### 1. VSCode Extension Responsibilities

- **Environment Variable:** The extension is responsible for communicating the complete workspace structure to the CLI. It does this by setting the `GEMINI_CLI_IDE_WORKSPACE_PATH` environment variable.
- **Path Serialization:** When one or more workspace folders are open, the extension gets the full path of each folder, joins them into a single string using the OS-native path delimiter (`:` for macOS/Linux, `;` for Windows), and sets this string as the value for the environment variable.
- **Global Availability:** This variable is applied to the global VSCode environment using the `environmentVariableCollection` API, ensuring it's available to any terminal created within VSCode. It is updated automatically whenever a folder is added or removed from the workspace.
- **User-Selected Context:** When the user explicitly launches the Gemini CLI via the "Run Gemini CLI" command in a multi-root workspace, the extension prompts them to select which of the open folders they wish to work in. The new terminal is then created with its current working directory (`cwd`) set to the path of the selected folder.

### 2. Gemini CLI Core Responsibilities

- **Context Validation:** The CLI's `IdeClient` is responsible for validating that it is running in a valid IDE context before enabling integration features.
- **Workspace Verification:** On startup, the `IdeClient` reads the `GEMINI_CLI_IDE_WORKSPACE_PATH` environment variable. It splits the value into a list of all workspace root paths.
- **CWD Check:** It then verifies that the CLI's current working directory (`process.cwd()`) is a child directory of at least one of the paths in the list. If this check passes, the IDE integration is enabled. If it fails, the integration remains disabled, and the user is notified of the directory mismatch.

This design ensures that the CLI is only active when it's confident it's running within a folder that the user has explicitly opened in their IDE, thereby supporting multi-root setups without any ambiguity.

### `packages/vscode-ide-companion/src/extension.ts`

- The `updateWorkspacePath` function was modified to gather all folder paths from `vscode.workspace.workspaceFolders`, join them with `path.delimiter`, and set the result in the `environmentVariableCollection`. If no folders are open, the variable is deleted.
- The `gemini-cli.runGeminiCLI` command was updated to check for multiple workspace folders. If more than one is present, it uses `vscode.window.showQuickPick` to prompt the user for a folder, which is then used as the `cwd` for the new terminal.

### `packages/core/src/ide/ide-client.ts`

- The `validateWorkspacePath` method was rewritten.
- It now splits the `GEMINI_CLI_IDE_WORKSPACE_PATH` environment variable into an array of `workspaceRoots`.
- It checks if the current working directory is a subdirectory of *any* of the `workspaceRoots`.
- The error message for an empty `GEMINI_CLI_IDE_WORKSPACE_PATH` was updated to be more generic, prompting the user to open a folder.

### Tests

- Unit tests were added for `packages/core/src/ide/ide-client.ts` to verify the new workspace validation logic.
- Unit tests were added for `packages/vscode-ide-companion/src/extension.ts` to verify the new multi-root workspace functionality, including prompting the user to select a folder.

## Reviewer Test Plan

I have tested this on macOS with my Chromium VSCode workspace. To test, you need to install the extension separately by building it in `packages/vscode-ide-companion` and installing the VSIX in VSCode. Then you can test the CLI with `node ~/path/to/gemini-cli/packages/cli`. Note that the "Run Gemini CLI" extension command will just execute `gemini` in the CLI so be sure to alias it or link to the development build, otherwise it will run the global installed `gemini`.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | Y  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Resolves #6209

### DIFF
--- a/packages/core/src/ide/ide-client.test.ts
+++ b/packages/core/src/ide/ide-client.test.ts
@@ -1,0 +1,217 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { vi, describe, it, expect, beforeEach, afterEach } from 'vitest';
+import { IdeClient, IDEConnectionStatus } from './ide-client.js';
+import * as detectIde from '../ide/detect-ide.js';
+import * as fs from 'node:fs';
+import * as path from 'node:path';
+
+// Mock dependencies
+vi.mock('node:fs');
+vi.mock('../ide/detect-ide.js');
+
+describe('IdeClient', () => {
+  let originalProcess: NodeJS.Process;
+
+  beforeEach(() => {
+    // Mock process.cwd and process.env
+    originalProcess = process;
+    vi.stubGlobal('process', {
+      ...originalProcess,
+      cwd: vi.fn(),
+      env: { ...originalProcess.env },
+    });
+
+    // Mock detectIde to simulate being in VSCode
+    vi.mocked(detectIde.detectIde).mockReturnValue(detectIde.DetectedIde.VSCode);
+    vi.mocked(detectIde.getIdeInfo).mockReturnValue({
+      displayName: 'VSCode',
+    });
+
+    // Mock fs.realpathSync to return the path itself
+    vi.mocked(fs.realpathSync).mockImplementation((p) => p.toString());
+
+    // Reset singleton instance before each test
+    (IdeClient as any).instance = undefined;
+  });
+
+  afterEach(() => {
+    vi.unstubAllGlobals();
+    vi.resetAllMocks();
+    (IdeClient as any).instance = undefined;
+  });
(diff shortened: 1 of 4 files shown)


sunnyps:
Closing this in lieu of https://github.com/google-gemini/gemini-cli/pull/6177

PR #6210, opened 2025-08-14
feat: Show untrusted status in the Footer

shrutip90:
## TLDR

Show untrusted status in the Footer only when the folder is untrusted. This will replace the sandbox info.
For trusted folders, we do not show any status.

## Dive Deeper

Screenshot:
Screenshot 2025-08-13 at 11 35 37 PM

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Related to https://github.com/google-gemini/maintainers-gemini-cli/issues/640


QuestionTop:
I have also encountered this problem.

packages/cli/src/ui/hooks/useFolderTrust.ts
```diff
@@ -4,42 +4,61 @@
...
-    setIsFolderTrustDialogOpen(false);
-  }, []);
+      trustedFolders.setValue(cwd, trustLevel);
+      const trusted = isWorkspaceTrusted(settings.merged);
+      setIsTrusted(trusted);
+      setIsFolderTrustDialogOpen(false);
+      onTrustChange(trusted);
+    },
+    [onTrustChange, settings.merged],
+  );
```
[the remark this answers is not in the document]
shrutip90: Done. Updated.

packages/cli/src/config/trustedFolders.ts
cornmander, line 1: Most of the business logic should go into the core package and not the CLI package. We use core in places without the CLI package, for example in vs code agent mode.

jacob314: agree would be good to move this logic to the core package. Sorry I missed that on the previous pr!

packages/cli/src/ui/components/Footer.tsx
```diff
@@ -90,16 +92,21 @@ export const Footer: React.FC<FooterProps> = ({
...
       <Box
         flexGrow={isNarrow ? 0 : 1}
         alignItems="center"
         justifyContent={isNarrow ? 'flex-start' : 'center'}
         display="flex"
         paddingX={isNarrow ? 0 : 1}
         paddingTop={isNarrow ? 1 : 0}
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-ignore
+        testID="sandbox-info"
```
jacob314: please remove testID and test by matching the text rendered. Good news is I don't think the test even uses this. Did Gemini CLI add this testID field? If so, might be an odd holdover from non-Ink react.

packages/cli/src/config/trustedFolders.ts
jacob314: there is a corresponding config.ts in core where you will want to make sure the folder trusted setting is passed to.

jacob314 (commented):
Looks good. Will approve once the trustedFolders logic has been mored to Core.

packages/cli/src/ui/hooks/useFolderTrust.ts
```diff
@@ -21,12 +21,14 @@ export const useFolderTrust = (
...
   const [isFolderTrustDialogOpen, setIsFolderTrustDialogOpen] = useState(false);
 
+  const { folderTrust, folderTrustFeature } = settings.merged;
   useEffect(() => {
     const trusted = isWorkspaceTrusted(settings.merged);
     setIsTrusted(trusted);
     setIsFolderTrustDialogOpen(trusted === undefined);
     onTrustChange(trusted);
-  }, [onTrustChange, settings.merged]);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
jacob314: don't suppress this lint. it is required for robust react behavior. same bellow.

jacob314 (commented):
Approved once the react dependencies issue is addressed.

packages/cli/src/ui/components/Footer.tsx
shrutip90: Ya CLI had added it in one of its version of tests I think, but later didn't end up using it. Removed it.

packages/cli/src/config/trustedFolders.ts
shrutip90: I do have the trustedFolder values in core/config too: https://github.com/google-gemini/gemini-cli/blob/4973e7e1e04004547021986a0e2dc5eeb4a8bf9d/packages/core/src/config/config.ts#L678.

I had kept this here, since I thought this is similar to the parsing of settings file. But I can try to move it...
Will do it in another PR.

packages/cli/src/ui/hooks/useFolderTrust.ts
shrutip90: Done.

jacob314 (approved):
lgtm

Issue #4191, opened 2025-07-15
Public Roadmap


mattKorwel:
This is a tracking issue for navigating the living public roadmap for Gemini CLI. If you are interested in following along with what we are doing, or even more exciting, looking to contribute. Please browse through here. Any items within are fair game and would be great contributions.

danowar2:
Thanks for the update

ahmedzaki5554488-blip:
https://github.com/google-gemini/gemini-cli/issues/4191

kashifo:
The best coding agent in the world is Gemini 3 for planning + GLM 4.6 for implementing

get it for just $2 pm, or 10% off with my code: https://z.ai/subscribe?ic=OP8ZPS4ZK6

anowardear062-svg:
Support and proses earnings

gitkit13:
> This is a tracking issue for navigating the living public roadmap for Gemini CLI. If you are interested in following along with what we are doing, or even more exciting, looking to contribute. Please browse through here. Any items within are fair game and would be great contributions.

https://github.com/mdn/browser-compat-data/blob/main/docs%2FREADME.md

gitkit13:
git remote add origin https://github.com/gitkit13/Circleci.git
git branch -M block-eth-rep
git push -u origin block-eth-rep

El-o-heka:
[yahway.java](https://github.com/user-attachments/files/27080867/yahway.java)

ronyjsro4:
0130133614101301336141 location taking local public

jeelsoni01:
We have started working on this issue #4191

jeelsoni01:
I have submitted PR #28960 to fix the trailing period bug on the Antigravity URL (#28052), which is part of the Core/Auth roadmap workstream. The CLA is signed, tests are green, and it is ready for review!

jeelsoni01:
This PR is just the first part of the larger Core/Auth + Capacity + Scalability workstream. We have solved this initial issue and are actively working on the rest of the related problems in this workstream now.

PR #6217, opened 2025-08-14
Sends Gemini CLI Client install id to Vertex and Gemini.

uttamkanodia14:
## TLDR

Sends Gemini CLI Client install id to Vertex and Gemini.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs
Resolves #6216


packages/core/src/core/contentGenerator.ts
```diff
@@ -108,15 +109,16 @@ export async function createContentGenerator(
...
+  const userAgent = `GeminiCLI/${version} (${process.platform}; ${process.arch})`;
   if (
     config.authType === AuthType.LOGIN_WITH_GOOGLE ||
     config.authType === AuthType.CLOUD_SHELL
   ) {
+    const httpOptions = {
+      headers: {
+        'User-Agent': `${userAgent}`,
+      },
+    };
```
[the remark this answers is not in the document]
mattKorwel: @uttamkanodia14 what do you think about the review comment here?

mattKorwel (dismissed):
@uttamkanodia14 I see there is a linked issue but that issue doesn't leave much detail and won't help future us understand the what and why. Could you take a pass at detailing why we want to send this data where it will be used and why we only send it in ceratain scenarios?

Also, I don't see this issue linked to a card related to the #4191. Can you help me understand which worksstream this is part of and where the genesis for the work came?

Marking this as requesting changes re the extra details above. Thanks!

packages/core/src/core/contentGenerator.ts
```diff
@@ -132,6 +134,13 @@ export async function createContentGenerator(
     config.authType === AuthType.USE_GEMINI ||
     config.authType === AuthType.USE_VERTEX_AI
   ) {
+    const installationId = getInstallationId();
+    const httpOptions = {
+      headers: {
+        'User-Agent': `${userAgent}`,
+        'Gemini-Cli-Install-Id': `${installationId}`,
```
NTaylorMullen: Why not log for all auth methods?

NTaylorMullen (changes requested):
We should ensure install id respects the `config?.getUsageStatisticsEnabled()` and doesn't log it if this is set.

packages/core/src/core/contentGenerator.ts
jpmaniqies23: > Why not log for all auth methods?

packages/core/src/core/contentGenerator.ts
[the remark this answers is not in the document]
uttamkanodia14: Yes this looks good. Have applied somewhat same with some changes.

packages/core/src/core/contentGenerator.ts
uttamkanodia14: @NTaylorMullen for GCA we have the gaia id.

uttamkanodia14:
> @uttamkanodia14 I see there is a linked issue but that issue doesn't leave much detail and won't help future us understand the what and why. Could you take a pass at detailing why we want to send this data where it will be used and why we only send it in ceratain scenarios?
> 
> Also, I don't see this issue linked to a card related to the #4191. Can you help me understand which worksstream this is part of and where the genesis for the work came?
> 
> Marking this as requesting changes re the extra details above. Thanks!

owenofbrien (commented):
@uttamkanodia14 - are the server-side changes fully rolled out on the Gemini and Vertex API side such that sending x-gemini-api-privileged-user-id won't error out?

packages/core/src/core/contentGenerator.ts
```diff
@@ -132,6 +135,16 @@ export async function createContentGenerator(
     config.authType === AuthType.USE_GEMINI ||
     config.authType === AuthType.USE_VERTEX_AI
   ) {
+    let headers: Record<string, string> = { ...baseHeaders };
+    if (gcConfig?.getUsageStatisticsEnabled()) {
+      const installationId = getInstallationId();
+      headers = {
+        ...headers,
+        'x-gemini-api-privileged-user-id': `${installationId}`,
```
NTaylorMullen, line 142: it's no longer recommended to `x-` prefixed I thought?

NTaylorMullen (approved):

owenofbrien:
> @uttamkanodia14 - are the server-side changes fully rolled out on the Gemini and Vertex API side such that sending x-gemini-api-privileged-user-id won't error out?

I'll go ahead and approve to unblock this before I log out for the weekend, but please double check this before merging, @uttamkanodia14 - thanks!

owenofbrien (approved):
Please double check that all the necessary changes are fully rolled out on both backend servers before merging this.

packages/core/src/core/contentGenerator.ts
uttamkanodia14: Gemini API team has proposed this.

cvjxendra:
May I use different GEMINI_API_KEY for the same install id? I use my PC for my projects at home and also for my college. I use different emails and I have different GEMINI_API_KEYs. I am worried about be banned or something like this.

PR #6233, opened 2025-08-14
Fix release notes generation

### LINKED ISSUE #3847: Release notes starting from wrong tag
### What happened?

In 0.1.11 list value 0.1.9-0.1.11

List 0.1.10 - 0.1.11

### Client information

<details>

```console
$ gemini /about
# paste output here
```

</details>

### Login information

_No response_

_No response_

skeshive:
Currently, release notes include commits from previous releases. We should explicitly create the release notes from after the previous release tag instead

Fixes: #3847


.github/workflows/release.yml
```diff
@@ -183,18 +183,29 @@ jobs:
...
             --tag="${NPM_TAG}"
 
+      - name: 'Get previous release tag'
+        id: 'previous_release'
+        if: |-
+          ${{ steps.vars.outputs.is_dry_run == 'false' }}
+        env:
+          GITHUB_TOKEN: '${{ secrets.GITHUB_TOKEN }}'
+        run: |-
+          echo "PREVIOUS_TAG=$(gh release list --limit 1 --json tagName --jq '.[0].tagName')" >> "$GITHUB_OUTPUT"
```
mattKorwel: How does this handle "nightly" vs "full" releases? Will, say 0.1.22 diff against the most recent nightly or against 0.1.21?

skeshive: ah good catch- just updated to account for nightly vs. non-nightly

a less brittle way to do this would be based on the isPrerelease flag but github isnt setting that for our nightly releases

mattKorwel (approved):
thanks! LGTM

PR #6234, opened 2025-08-14
feat(core): Preload User Memory in New Chat History

bct8925:
## TLDR

From my testing Gemini seems to follow instructions better if they are in the actual chat context, so this adds an option to preload user memory (GEMINI.md files and other context files) into the chat history for new / cleared chats. Optional since it sacrifices some of the context window.

## Reviewer Test Plan

"Unit" testing via this prompt with the setting on/off:
> summarize each message we've sent to each other so far in this current chat

When `loadUserMemoryInHistory: true` it shows the default 2 messages plus the user memory content from GEMINI.md files.
When `loadUserMemoryInHistory: false` it shows just the default 2 messages.

As for testing its ability to follow instructions better, that would involve more subjective/circumstantial testing.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ✅  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

There are definitely a few issues added in the context of Gemini not following instructions, but cannot guarantee that this should close them.

### DIFF
--- a/docs/cli/configuration.md
+++ b/docs/cli/configuration.md
@@ -288,6 +288,14 @@ In addition to a project settings file, a project's `.gemini` directory can cont
     "showLineNumbers": false
     ```
 
+- **`loadUserMemoryInHistory`** (boolean):
+  - **Description:** Controls whether user memory (including GEMINI.md files and other context files) is loaded into the chat history for new or cleared chats. When enabled, the content from your context files is added as a user message at the start of each new conversation, making the AI aware of your project-specific instructions throughout the entire conversation. When disabled, context files are only included in the system prompt.
+  - **Default:** `false`
+  - **Example:**
+    ```json
+    "loadUserMemoryInHistory": true
+    ```
+
 ### Example `settings.json`:
 
 ```json
--- a/packages/cli/src/config/config.ts
+++ b/packages/cli/src/config/config.ts
@@ -527,6 +527,7 @@ export async function loadCliConfig(
     folderTrust,
     interactive,
     trustedFolder,
+    loadUserMemoryInHistory: settings.loadUserMemoryInHistory ?? false,
   });
 }
 
--- a/packages/cli/src/config/settingsSchema.ts
+++ b/packages/cli/src/config/settingsSchema.ts
@@ -495,6 +495,16 @@ export const SETTINGS_SCHEMA = {
     description: 'Show line numbers in the chat.',
     showInDialog: true,
   },
+  loadUserMemoryInHistory: {
+    type: 'boolean',
+    label: 'Preload User Memory for New Chats',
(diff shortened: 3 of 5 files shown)


bct8925:
> ## Code Review
> This pull request introduces a feature to preload system instructions into the chat history, controlled by a new `loadSystemPromptInHistory` setting. The implementation is sound and includes necessary documentation and configuration updates. However, I've identified a high-severity issue where enabling this feature causes the system prompt to be sent to the model twice: once as a user message in the history and again as a `systemInstruction`. This duplication can increase token usage and may lead to unexpected model behavior. I have provided a code suggestion to rectify this by ensuring the system prompt is sent only once.

Hi @gemini-code-assist! 

This is actually intentional with the purpose of being more verbose. I've made it an optional configuration defaulted to off so that this increased token usage can be acknowledged and accepted without affecting the majority of users.

bct8925:
> Hi @bct8925,
> 
> Thank you for clarifying that the dual inclusion of the system prompt (as a `systemInstruction` and as a `user` message in the chat history) is intentional for increased verbosity. I understand the goal of making the instructions more prominent to the model.
> 
> While the `loadSystemPromptInHistory` setting being optional and defaulted to `false` addresses the concern about increased token usage for most users, I'd still like to highlight a potential consideration regarding model behavior. Even with intentional verbosity, sending the same core instruction through two distinct input mechanisms (a dedicated `systemInstruction` parameter and a `user` message within the chat history) could, in some cases, lead to subtle or unintended interpretations by the model. Models are typically designed to process `systemInstruction` as a primary directive and chat history as conversational context. Duplicating the instruction might not always result in the desired 'more verbose' emphasis, and could potentially lead to the model over-indexing on the instruction or, conversely, treating the historical user message as redundant.
> 
> If the primary goal is to ensure the model strongly adheres to the system prompt, relying solely on the `systemInstruction` parameter is generally the most direct and robust approach. If additional emphasis or a conversational 'acknowledgment' of the system prompt is desired within the chat history, perhaps a distinct, concise message in the history that _references_ the system prompt's core intent (rather than duplicating the full prompt) could achieve a similar 'verbose' effect without the potential for ambiguous interpretation or unnecessary token consumption.
> 
> However, I acknowledge your design choice and the intention behind it. If your testing indicates that this approach consistently yields the desired 'more verbose' and effective model responses, then the current implementation serves its purpose.

Ah I made a typo originally. The configuration option is now called "loadUserMemoryInHistory" and loads the user memory instead of the system prompt, which was my original intent.

PR #6240, opened 2025-08-14
feat(core): Save large tool outputs to a file and return truncated lines

SandyTao520:
## TLDR

This pull request introduces a feature to manage large tool outputs. When a tool's output exceeds 1m charas, the output is truncated in the UI, and the full content is saved to a temporary file. The user is notified with the path to this file, allowing them or the model to inspect the full output using the `read_file` tool. This behavior is controlled by a new user setting, `truncateToolOutput`, which is disabled by default.

## Dive Deeper

The core changes are in `packages/core/src/core/coreToolScheduler.ts`, which now includes a `truncateAndSaveToFile` function. This function checks if a tool's string output is over 1,000,000 characters. If it is, the content is saved to a file in the project's temp directory (`~/.gemini/temp`).

To support this, the `read_file` tool has been updated to allow reading from this specific temporary directory, in addition to the workspace directories.

The UI component `ToolGroupMessage.tsx` has been modified to display a message with the path to the saved file whenever a tool's output has been truncated and saved.

A new boolean setting, `truncateToolOutput`, has been added to the configuration schema, allowing users to opt-in to this feature via the `/settings` dialog.

## Reviewer Test Plan

1.  First, enable the feature by adding "truncateToolOutput": true to `settings.json` or toggling on "Truncate Tool Output" in the `/settings` dialog.
2.  Ask Gemini to print large content `` > run python3 -c "print('hello world\n' * 300_000)"``
3.  **Verify** that Gemini CLI doesn't throw a bad request saying the input exceeds max token count.
4.  **Verify** that it shows a message like `Output too long and was saved to: /path/to/project/.gemini/temp/<id>.txt`.
5.  **Verify** that you can view the full content in the directory or ask Gemini to read it if it needs to.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ✅  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


jacob314:
Btw the React based UI with MaxSizedBox also falls over for extremely large input. We should consider truncating it as well if we go this route.

jacob314:
Fyi @bbiggs who likely has some thoughts about this or at least some good examples of cases we should test it on to make sure it works alright for power users.

agladysh:
FWIW, my personal take, based on what Codex does, is that this behaviour *will* be sneakily confusing to the model in more real life scenarios.

Please make this feature configurable, so users may set the threshold or disable it.

Given how large context-window Gemini has, I personally would prefer to have it off by default. A future implementation might be adaptive, say, configure it to auto-enable if remaining context window is less than XX%

Alternative, potentially less confusing, take would be to refuse to read at all with an appropriate error message for the model, which may suggest same as it does now.

packages/cli/src/ui/types.ts
```diff
@@ -50,6 +50,7 @@ export interface IndividualToolCallDisplay {
   status: ToolCallStatus;
   confirmationDetails: ToolCallConfirmationDetails | undefined;
   renderOutputAsMarkdown?: boolean;
+  outputFile?: string | undefined;
```
NTaylorMullen:
```suggestion
  outputFile?: string;
```

packages/cli/src/config/settingsSchema.ts
```diff
@@ -544,6 +544,16 @@ export const SETTINGS_SCHEMA = {
       'Enable AI-powered prompt completion suggestions while typing.',
     showInDialog: true,
   },
+  truncateToolOutput: {
+    type: 'boolean',
+    label: 'Truncate Tool Output',
+    category: 'General',
+    requiresRestart: false,
+    default: false,
```
NTaylorMullen: Off by default for now for evals? Assuming so what's the plan for swapping to on-by-default?

packages/core/src/core/coreToolScheduler.ts
```diff
@@ -225,6 +227,57 @@ const createErrorResponse = (
   errorType,
 });
 
+const TRUNCATION_THRESHOLD = 1_000_000;
```
NTaylorMullen: 1 token is typically ~4 characters. This would mean if a single tool output was 250k tokens to then truncate. We should probably truncate at a significantly lower threshold. I could even see that number being 40k characters (10k tokens).

The one UX implication of doing this is that you're then notifying the user a lot more frequently that their shell output has been saved to a specific file.

packages/core/src/core/coreToolScheduler.ts
```diff
@@ -225,6 +227,57 @@ const createErrorResponse = (
...
+const TRUNCATION_LINES = 1_000;
```
NTaylorMullen: 1k lines is a good number, I imagine this is what's driving truncation largely today

packages/core/src/core/coreToolScheduler.ts
```diff
@@ -225,6 +227,57 @@ const createErrorResponse = (
...
+  const lines = content.split('\n');
+  let truncatedContent: string;
+
+  // Determine the truncated content for display based on its structure.
+  if (lines.length > TRUNCATION_LINES) {
+    // Content has many lines; truncate by line count.
+    truncatedContent = lines.slice(-TRUNCATION_LINES).join('\n');
+  } else {
+    // Content has few lines (or one very long line); truncate by character count.
+    const maxChars = TRUNCATION_LINES * 80; // Approximate 80 chars per line
```
NTaylorMullen: ~20k tokens. I would have thought the truncation threshold would be the maxChars here though. Feels a little disconnected

packages/core/src/core/coreToolScheduler.ts
```diff
@@ -225,6 +227,57 @@ const createErrorResponse = (
...
+    truncatedContent = content.slice(-maxChars);
+  }
+
+  // Sanitize callId to prevent path traversal.
+  const safeFileName = `${path.basename(callId)}.txt`;
```
NTaylorMullen: The file extension will play a large role in how the model decides to read said file. I'd recommend dropping the file extension entirely to be honest since we don't have a good heuristic. Alternatively, we could special case a few things like json/xml BUT that might be overkill.

I could also see us picking a general file extension like `.log` or `.output`

packages/core/src/core/coreToolScheduler.ts
```diff
@@ -225,6 +227,57 @@ const createErrorResponse = (
...
+  const outputFile = path.join(projectTempDir, safeFileName);
+  try {
+    await fs.writeFile(outputFile, content);
+
+    return {
+      content: `Tool output was too large and has been truncated.
+The full output has been saved to: ${outputFile}
+To read the complete output, use the read_file tool with the absolute file path above. For large files, you can use the offset and limit parameters to read specific sections:
```
NTaylorMullen: Should replace all the tool names and parameters with `${ReadFile.Name}` etc.

packages/core/src/core/coreToolScheduler.ts
```diff
@@ -225,6 +227,57 @@ const createErrorResponse = (
...
+- read_file tool with offset=0, limit=100 to see the first 100 lines
```
NTaylorMullen: The read file tool itself describes the offset/limit doesn't it? Did you find that you needed to re-state it?

packages/core/src/core/coreToolScheduler.ts
```diff
@@ -225,6 +227,57 @@ const createErrorResponse = (
...
+- read_file tool with offset=N to skip N lines from the beginning
+- read_file tool with limit=M to read only M lines at a time
+This allows you to efficiently examine different parts of the output without loading the entire file.
+Last ${TRUNCATION_LINES} lines of output:
```
NTaylorMullen: This isn't accurate if we have one super large line (or `n` super large lines). We'll want this to be accurate

packages/core/src/core/coreToolScheduler.ts
```diff
@@ -225,6 +227,57 @@ const createErrorResponse = (
...
+...\n${truncatedContent}`,
```
NTaylorMullen: Windows newlines are `\r\n` so we can lean on node to translate the string literals properly

```suggestion
...
${truncatedContent}`,
```

packages/core/src/core/coreToolScheduler.ts
```diff
@@ -843,17 +896,31 @@ export class CoreToolScheduler {
...
+              let content = toolResult.llmContent;
+              let outputFile: string | undefined = undefined;
+              if (
+                typeof content === 'string' &&
+                this.config.getTruncateToolOutput()
+              ) {
+                ({ content, outputFile } = await truncateAndSaveToFile(
+                  content,
+                  callId,
+                  this.config.storage.getProjectTempDir(),
```
NTaylorMullen, line 984: Does this map to `<projectdir>/.gemini/tmp`? If so does that mean that any written file will have to be reverted / deleted by users to avoid accidentally staging?

Assuming so we probably should consider another route to storing these outside of the current workspace dir. I realize that this would then require us to special case a specialized dir (or file) in read_file BUT I think the alternate would be quite painful for the user.

agladysh: Sorry for a random plug, but, NB: `<projectdir>/.gemini/tmp` can have `.gitignore` with `*` in it.

BTW, note how Claude Code handles its own .claude/ ignores. Not sure I like it messing up with my home dir git config, but:

```
cat /Users/agladysh/.config/git/ignore
**/.claude/settings.local.json
```

packages/cli/src/config/settingsSchema.ts
anj-s: The plan is to first run a small experiment for internal and external users to see if any major issues surface. Last time we tried rolling this out we had immediate pushback because the model did not get the necessary context. 

If we don't see an impact in our evals and gemini CLI is still able to complete tasks, we can turn this on by default.

packages/core/src/core/coreToolScheduler.ts
```diff
@@ -843,17 +897,31 @@ export class CoreToolScheduler {
             }
 
             if (toolResult.error === undefined) {
+              let content = toolResult.llmContent;
+              let outputFile: string | undefined = undefined;
+              if (
+                typeof content === 'string' &&
```
anj-s, line 978: we should add an issue that we might need to support a different solution for non text output or at least throw a reasonable error message

packages/core/src/core/coreToolScheduler.ts
anj-s: Can these be tunable knobs as well? This way we can evaluate a good number with a few experiments. It also allows users to increase this for any edge case use cases.

packages/core/src/core/coreToolScheduler.ts
anj-s: We are also saving our checkpoints and chats. Wouldn't we need to save this in coordination with the other state management we currently have?

SandyTao520: It maps to `~/.gemini/<project_hash>/tmp/`. The plan is to implement cleanup mechanism that remove tmp files that wasn't included in checkpointing files.

packages/core/src/core/coreToolScheduler.ts
SandyTao520: #7344

packages/core/src/core/coreToolScheduler.ts
SandyTao520: Yeah sometimes it doesn't know it should read the file by chunks, I added this here to make sure the model knows.

packages/core/src/core/coreToolScheduler.ts
SandyTao520: Changed these to 2 settings: 1) threshold that is default to 4M and 2) truncation lines which defaults to 1k.

packages/core/src/core/coreToolScheduler.ts
SandyTao520: As discussed below, I made them configurable settings and let users decide.

packages/cli/src/config/config.ts
```diff
@@ -621,6 +621,8 @@ export async function loadCliConfig(
     shouldUseNodePtyShell: settings.tools?.usePty,
     skipNextSpeakerCheck: settings.model?.skipNextSpeakerCheck,
     enablePromptCompletion: settings.general?.enablePromptCompletion ?? false,
+    truncateToolOutputThreshold: settings.model?.truncateToolOutputThreshold,
+    truncateToolOutputLines: settings.model?.truncateToolOutputLines,
```
allenhutchison: wouldn't this be more logical in the tools stanza rather than the model stanza? I wouldn't think of these as model settings.

packages/core/src/config/config.ts
```diff
@@ -358,6 +362,9 @@ export class Config {
     this.useRipgrep = params.useRipgrep ?? false;
     this.shouldUseNodePtyShell = params.shouldUseNodePtyShell ?? false;
     this.skipNextSpeakerCheck = params.skipNextSpeakerCheck ?? false;
+    this.truncateToolOutputThreshold =
+      params.truncateToolOutputThreshold ?? 4_000_000;
+    this.truncateToolOutputLines = params.truncateToolOutputLines ?? 1000;
```
allenhutchison: Consider defining a const for these defaults and using that in the settings definition, the tests, and here.

packages/core/src/core/coreToolScheduler.ts
```diff
@@ -243,6 +246,56 @@ const createErrorResponse = (
...
+  let truncatedContent: string;
+
+  // Determine the truncated content for display based on its structure.
+  if (lines.length > truncateLines) {
+    // Content has many lines; truncate by line count.
+    truncatedContent = lines.slice(-truncateLines).join('\n');
+  } else {
+    // Content has few lines (or one very long line); truncate by character count.
+    truncatedContent = content.slice(-threshold);
+  }
```
allenhutchison, line 278: Maybe something to think about while you are running evals. My intuition is that this is going to be used a lot by tool calls for build and test actions. The piece of that output that will be most valuable to a model is likely going to be at the end of the file rather than the beginning. You might consider some alternative strategies for truncation:

1) Split your budget between the beginning of the file and the end, so you include 500 lines from the beginning and 500 from the end. If you go down this route I would encourage you to add a marker in between the two sections <content removed by truncation> or something like that to inform the model about where the missing data is. You would also update your instructions to tell the model about that marker.

2) Look at taking the output by itself and spinning up a thread to flash-lite with no context and ask it to summarize the file looking for things that are out of the ordianry or might require attention from the developer. This would allow you to use a fresh context window without worring about contaminating the current orchestration loop.

packages/core/src/core/coreToolScheduler.ts
```diff
@@ -243,6 +246,56 @@ const createErrorResponse = (
...
+  try {
+    await fs.writeFile(outputFile, content);
+
+    return {
+      content: `Tool output was too large and has been truncated.
+The full output has been saved to: ${outputFile}
+To read the complete output, use the ${ReadFileTool.Name} tool with the absolute file path above. For large files, you can use the offset and limit parameters to read specific sections:
+- ${ReadFileTool.Name} tool with offset=0, limit=100 to see the first 100 lines
+- ${ReadFileTool.Name} tool with offset=N to skip N lines from the beginning
+- ${ReadFileTool.Name} tool with limit=M to read only M lines at a time
```
allenhutchison, line 298: You should consider adapting these instructions for your two use cases (many lines vs long line).

SandyTao520: Just realized ReadFile tool can only offset by lines, so it technically cannot read a small chunk of a very long line. I updated logic to wrap very long lines so the same prompt works for both cases.

packages/core/src/core/coreToolScheduler.ts
SandyTao520: `.slice(-truncateLines);` with a negative sign is keeping the last `truncateLines` lines.

1. Good idea to keep the first and last `truncateLines/2` lines, updated the logic.
2. This is potentially the next step of how we can handle large tool output, which I plan to implement in the future.

allenhutchison: On 1, during evals you might want to play with the ratio of beginning to ending lines. Just keep that in mind. It may turn out that the first 100 lines are most valuable and the last 900 or something like that. 

On 2, cool, let me know if there is anything I can do to help.

allenhutchison (commented):
These changes look great.

allenhutchison (approved):

Issue #5691, opened 2025-08-06
Re-enable building tests for packages/cli


joshualitt:
### What happened?

Tests are not currently being built in packages/cli.

### What did you expect to happen?

All tests are built.

### Client information

n/a

scidomino:
What do you mean by this? When I execute `npm run test` it runs the packages/cli tests just fine.

joshualitt:
@scidomino , it's the build that is broken, as in these tests have type errors which aren't being caught(about ~311 when I last checked). For more context please see: https://github.com/google-gemini/gemini-cli/blob/main/packages/cli/tsconfig.json#L19, if you comment out L19-L21 and then run `npm run build` in `packages/cli` you should see the issue.

PR #6244, opened 2025-08-14
bug(cli): Exclude only specific tests.

joshualitt:
## TLDR

Switch from glob excludes to specific excludes to stop bleeding.

## Dive Deeper

We are now over 500 type errors in these files. Trying to fix everything in one CL is very risky. Instead, in an effort to stop the bleeding, this CL switches to specifically excluding individual tests. Thus, new tests *will* be built, and older tests can be fixed as they are touched.

[Bug](https://github.com/google-gemini/gemini-cli/issues/5691) 

## Reviewer Test Plan

CL builds.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


jacob314 (approved):
lgtm

PR #6255, opened 2025-08-14
chore(compiler): Enable strict property access TS compiler flag.

richieforeman:
## TLDR

- Enable strict property access compiler flag. This matches Google3.  It requires ['property_access'] on things that are index or Record-like types.

## Dive Deeper

Optionally, we may want to use `vi.stubEnv` in our tests.  I started doing that, but am worried about merge conflict radius.

## Reviewer Test Plan

N/A

## Testing Matrix
N/A

## Linked issues / bugs

N/A


packages/core/src/utils/editor.test.ts
```diff
@@ -33,7 +33,7 @@ const originalPlatform = process.platform;
 describe('editor utils', () => {
   beforeEach(() => {
     vi.clearAllMocks();
-    delete process.env.SANDBOX;
```
jacob314, line 36: did gemini CLI decide to make this change or did you?

jacob314 (approved):
lgtm
All things considered think this a slight positive as it ensures everyone reviewing code is aware that these calls were property accesses.

PR #6262, opened 2025-08-14
[ide-mode] Use command to detect IDE

chrstnb:
## TLDR

Introduce an IDE type for VSCode forks-- this will ensure we only install the VSCode extension on VSCode.

This ended up being a lot more involved than expected bc of the logging dependency. IIUC we want that to function even when not in IDE mode, so it seemed weird to be reliant on the IDE client for that. Interested to hear thoughts.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

### DIFF
--- a/packages/core/src/ide/detect-ide.test.ts
+++ b/packages/core/src/ide/detect-ide.test.ts
@@ -4,65 +4,72 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { describe, it, expect, afterEach, vi } from 'vitest';
-import { detectIde, DetectedIde } from './detect-ide.js';
+import { detectIde, detectIdeByEnvVar, DetectedIde } from './detect-ide.js';
 
 describe('detectIde', () => {
+  const originalEnv = { ...process.env };
+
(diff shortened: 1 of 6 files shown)


packages/core/src/ide/detect-ide.test.ts
```diff
@@ -4,65 +4,72 @@
...
   afterEach(() => {
-    vi.unstubAllEnvs();
+    process.env = { ...originalEnv };
+  });
+
+  it('should return undefined when no IDE is detected', () => {
+    expect(detectIdeByEnvVar()).toBeUndefined();
+  });
+
+  it('should detect Devin when __COG_BASHRC_SOURCED is set', () => {
```
richieforeman, line 20: Please bring back the parameterized testing here to reduce boilerplate.

PR #6266, opened 2025-08-14
fix(ui): display file paths with full parent directory instead of par…

Alane38:
### Title

```
fix(ui): display duplicate file names with better path prefix
```

### Description

This change updates how duplicate file names are displayed in the "Open files" list.

**Before:**

```
file.txt (/dirA)
file.txt (/dirB)
```

**After:**

```
/dirA/file.txt
/dirB/file.txt
```

**Why:**

* Using parentheses for the parent directory was inconsistent with other parts of the UI.
* Standard path notation is more intuitive and easier to scan visually.
* Improves clarity when multiple files share the same name.

**Testing:**

1. Open two files with the same name located in different directories.
2. Run `gemini-cli` and view the "Open files" list.
3. Confirm that parent directories now appear as a prefix to the file name rather than inside parentheses.


Alane38 (commented):
## TLDR

This PR updates the IDE to display the **full parent directory** for files with duplicate basenames, making it easier to distinguish between them. Test snapshots have been updated accordingly.

## Dive Deeper

Previously, when multiple files with the same name were open, only the basename was shown with the parent directory in parentheses (e.g., `bar.txt (/foo)`). This could be confusing if multiple files with the same name exist in different directories.  

With this change:  
- The display now shows the parent directory before the filename (e.g., `/foo/bar.txt`).  
- Duplicate files are clearly distinguishable at a glance.  
- Snapshot tests were updated to reflect the new formatting.

## Reviewer Test Plan

1. Pull the branch locally.  
2. Open multiple files with the same name in different directories.  
3. Verify that the file list shows the parent directory before the filename.  
4. Confirm that active files are still correctly marked.  
5. Run `npm test` to ensure snapshots match the new display.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ✅  | ✅  |
| npx      | ✅  | ✅  | ✅  |
| Docker   | ✅  | ✅  | ✅  |
| Podman   | ✅  | -   | -   |
| Seatbelt | ✅  | -   | -   |

## Linked issues / bugs

Related to #6266 – improves file path display for duplicate filenames in the IDE.

Alane38:
The issue has been fixed. Workflow approval can now proceed.

PR #6277, opened 2025-08-14
Fix flake due to using wait instead of waitFor

jacob314:
## TLDR
Gemini CLI caught and fixed this flake while working on an unrelated PR

using waitFor is the idiomatic way to handle this case as it ensures there cannot be race conditions. Only downside is if a test is broken it will take until waitFor times out for the test to fail.


allenhutchison (approved):
omg

PR #6278, opened 2025-08-14
feat(shell): Include disallowed commands in block reason

cornmander:
When a command is blocked, the reason message will now include a list of the specific commands that were disallowed.

This improves the user experience by making it clear which commands are causing the block, which is especially helpful in chained commands.


sethvargo (approved):
Gemini suggestions seem good

sethvargo (approved):

packages/core/src/utils/shell-utils.ts
```diff
@@ -301,7 +301,9 @@ export function checkCommandPermissions(
       return {
         allAllowed: false,
         disallowedCommands,
-        blockReason: `Command(s) not on the global or session allowlist.`,
+        blockReason: `Command(s) not on the global or session allowlist. Disallowed commands: ${disallowedCommands
```
leehagoodjames, line 304: Nit:
```suggestion
        blockReason: `Command(s) not on the global or session allowlist or exist in disallowed commands. Disallowed commands: ${disallowedCommands
```

leehagoodjames (approved):

packages/core/src/utils/shell-utils.ts
sethvargo, line 304: Coming at this with a fresh pair of eyes, I wonder if we should also `sort` the commands so the output is consistent? Otherwise, I think this list will show the order the commands were called, which might not be predictable.

cornmander: There's already a blockedCommands section above that returns earlier, I don't think this is necessary.

cornmander: I think it's fine? This is for human output and doesn't need to stable for equality checking.

bdmorgan (approved):
LGTM

PR #6275, opened 2025-08-14
fix: update issue dedup workflow

gsquared94:
## TLDR

This pull request refactors the issue de-duplication workflow to use actions/github-script for all GitHub API interactions, such as commenting and labeling. This replaces the previous method where the Gemini model was prompted to generate gh CLI commands, making the process significantly more reliable and maintainable.

##  Dive Deeper

  The original de-duplication process relied on prompting the Gemini model to construct and execute gh shell commands to comment on and label issues. This approach was brittle and insecure.

  This change introduces a more robust, two-step process:

   1. Find Duplicates: The Gemini model's role is now strictly limited to identifying duplicate issues. Its only output is a structured JSON object containing a list of duplicate issue numbers (e.g., `{"duplicate_issues": [123, 456]}`). The prompt has been simplified to reflect this focused responsibility.

   2. Comment and Label: A new, separate job uses `actions/github-script` to parse the JSON output from the first step. This script then uses the official octokit/rest.js client to reliably handle all interactions with GitHub, including:
       * Checking if a de-duplication comment already exists.
       * Creating a new comment or updating an existing one if the list of duplicates has changed.
       * Adding the `status/possible-duplicate` label to the issue.

  This separation of concerns makes the workflow more deterministic and easier to debug. The logic for interacting with GitHub is now explicit and no longer dependent on the model's ability to generate precise shell commands.

##  Reviewer Test Plan

   1. Create a new test issue with a title and body that is very similar to an existing issue.
   2. Verify that the `gemini-automated-issue-dedup.yml` action runs.
   3. Verify that the action correctly identifies the duplicate issue and adds a comment and the status/possible-duplicate label to the new issue.
   4. Find an existing issue that is a suspected duplicate of another, but hasn't been marked.
   5. Add a comment to that issue with the text @gemini-cli /deduplicate.
   6. Verify that the gemini-automated-issue-dedup.yml action runs and correctly identifies and labels the duplicate.
   7. Manually trigger the gemini-scheduled-issue-dedup.yml workflow and check the action logs to ensure it completes successfully.

## Testing Matrix

Successful run on fork: https://github.com/gsquared94/gemini-cli/actions/runs/16980863843

|          | 🍏  | 🪟  | 🐧  |
| -------- | --- | --- | --- |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
[...]

PR #6280, opened 2025-08-14
temp: remove issue dedup workflows

gsquared94:
## TLDR

Will add back after reimplementing with the new approach of using `actions/github-script`.


gsquared94:
Closed in favor of https://github.com/google-gemini/gemini-cli/pull/6275

PR #6281, opened 2025-08-15
fixes for oauth spec - adds github oauth support. Resource paramater.

emeryray2002:
## TLDR

This PR improves OAuth support for MCP servers, particularly for GitHub OAuth, by making the resource parameter conditional, supporting both JSON and form-urlencoded token responses, and enhancing OAuth discovery with path-based well-known URLs.

## Dive Deeper

  Key Changes

  1. Conditional Resource Parameter (oauth-provider.ts)
    - The resource parameter is now only added when an mcpServerUrl is provided
    - This makes the OAuth flow compatible with both resource-specific OAuth (which requires resource) and standard OAuth providers like GitHub (which don't expect it)
    - Added warnings instead of errors when the resource parameter can't be built
  2. GitHub OAuth Compatibility (oauth-provider.ts)
    - Added support for form-urlencoded token responses (GitHub returns tokens as application/x-www-form-urlencoded, not JSON)
    - Added proper Accept headers (application/json, application/x-www-form-urlencoded) for token exchange requests
    - Improved error handling to parse errors from both JSON and form-urlencoded responses
  3. Enhanced OAuth Discovery (oauth-utils.ts)
    - Improved well-known URL discovery to support both root-based and path-based endpoints
    - Now tries standard root-based discovery first, then falls back to path-based discovery
    - Better handling of authorization servers with custom paths (e.g., https://github.com/login/oauth)
    - Added includePathSuffix parameter to buildWellKnownUrls() for flexible discovery
  4. WWW-Authenticate Header Support (oauth-provider.ts)
    - Enhanced discovery flow to check for WWW-Authenticate headers when no authorization URL is provided
    - Supports both SSE and JSON endpoints when checking authentication requirements
    - Preserves client credentials during discovery and merges them with discovered configuration
  5. Improved Token Handling (oauth-provider.ts)
    - Better validation of token responses, ensuring access_token is present
    - Defaults token_type to 'Bearer' when not provided
    - Improved token preview in logs (truncates long tokens safely)
  6. Test Updates (oauth-provider.test.ts)
    - Fixed failing test by passing mcpServerUrl parameter to trigger resource parameter inclusion
    - Added proper response mocking with both text() and json() methods
    - Tests now properly validate OAuth flow with and without MCP server URLs

## Reviewer Test Plan

Here are a few MCP servers to try in your config:
 "cloudflare": {
      "url": "https://ai-gateway.mcp.cloudflare.com/sse"
    },
    "zapier": {
      "httpUrl": "https://mcp.zapier.com/api/mcp/a/41447/mcp"
    },
    "paypal": {
      "url": "https://mcp.paypal.com/sse"
    },
    "github-remote": {
      "httpUrl": "https://api.githubcopilot.com/mcp/",
      "oauth": {
        "enabled": true,
        "clientId": "Ov33liRpeSAv5qNlC3WW",
        "clientSecret": "{create your own 3rd party oauth app}"
      }
    }
  },

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


packages/core/src/mcp/oauth-utils.ts
```diff
@@ -43,18 +43,37 @@ export interface OAuthProtectedResourceMetadata {
...
+        authorizationServer: new URL(
+          '/.well-known/oauth-authorization-server',
+          base,
+        ).toString(),
+      };
+    }
+
+    // Path-based discovery: append path suffix to well-known URLs
+    const pathSuffix =
+      serverUrl.pathname === '/' ? '' : serverUrl.pathname.replace(/\/$/, ''); // Remove trailing slash
```
cornmander: '/'.replace(/\/$/, '') == '' so you don't need the ternary expression

packages/core/src/mcp/oauth-provider.ts
```diff
@@ -355,32 +359,80 @@ export class MCPOAuthProvider {
...
+        // Fall back to raw error
+      }
       throw new Error(
-        `Token exchange failed: ${response.status} - ${errorText}`,
+        `Token exchange failed: ${response.status} - ${responseText}`,
       );
     }
 
-    return (await response.json()) as OAuthTokenResponse;
+    // Try to parse as JSON first, fall back to form-urlencoded
```
cornmander, line 420: Can we console.warn() on content-type here if it mismatches json or form-urlencoded? I'm fine with the current guess-and-check approach but when we invariably get user error reports it will be helpful to see the response header

cornmander (dismissed):

packages/core/src/mcp/oauth-utils.ts
emeryray2002: fixed

packages/core/src/mcp/oauth-provider.ts
emeryray2002: fixed

cornmander (approved):

PR #5102, opened 2025-07-29
fix: re render context usage indicator 

psinha40898:
## TLDR
The UI that displayed the percentage remaining of the context window was not re rendering effectively enough.
A new component `ContextUsageDisplay` with props `promptTokenCount` and `model` ensure that the percentage display will re render whenever those props change.
This fixes the issue where another query is required _after_ compression to trigger the re render of `Footer`
Now `ContextUsageDisplay` should re render appropriately after `/compress`

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| -------- | --- | --- | --- |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes #4929  1/2

PR #6282, opened 2025-08-15
fix: pendingSlashCommandHistoryItems corrected to properly memoize updates

### LINKED ISSUE #4929: Chat History Compression should show a progress indicator so users don't think the UI has hung
### What happened?

After triggering `/compress`, there's no visual feedback or indication that the process has started. Due to this lack of response, I assumed nothing had happened and initiated the command again, resulting in two separate compressions being performed.
Image

The context usage indicator does not refresh automatically after compression. Despite performing compression, the indicator still displays 28% usage remaining. An additional query is required to update the current usage status.

Compression indication.

### Client information

* **CLI Version:** 0.1.14
* **Git Commit:** de968877 (local modifications)
* **Operating System:** darwin v22.13.0
* **Memory Usage:** 259.9 MB

### Login information

_No response_

_No response_

psinha40898:
## TLDR

The `/compress` command does not render its pending state because the memoization incorrectly uses `useStateAndRef` and passes a `ref` into the dependency array of `useMemo`

Before:
image

After:
image

## Dive Deeper

In React functions memoized through `useMemo` re-run when the identities of values in the dependency array change. Previously, the identity of  `pendingCompressionItemRef` was never changing because it is the ref from `useStateAndRef`.

## Potential Solutions

*(This PR implements option 1)*

**1) Use `useState` instead of `useStateAndRef`** because pending slashCommands are not complicated like AI streams and do not need the functionality

2) Remove memoization and continue to use `useStateAndRef` - this might increase the number of re-renders in a way that is not needed. It looks like this was the behavior before #3175 so reverting to this might be the better option. I'm not sure.

3) Retain memoization and let `useStateAndRef` also return the state, and then pass that into the dependency array. This may have been the intended path when looking at the comments in `useStateAndRef.ts`

This pull request implements option 1 so that if `useStateAndRef` is required in the future for slashCommands it must be deliberately chosen again.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes #4929 when on top of #5102


bdmorgan (approved):
LGTM

PR #6290, opened 2025-08-15
feat(mcp-client): add support for JSON Schema  references

cfool:
## TLDR

 Add getRefDefinition function to resolve JSON Schema reference paths
  - Enhance hasValidTypes to handle $ref references in schema validation                                                                                                                                                      
  - Support various reference formats: #/definitions, #/$defs, #/components/schemas                                                                                                                                          
  - Add comprehensive test coverage for $ref handling scenarios                                                                                                                                                                
  - Update existing tests to pass root schema context for proper $ref resolution
## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

### DIFF
--- a/packages/core/src/tools/mcp-client.test.ts
+++ b/packages/core/src/tools/mcp-client.test.ts
@@ -846,10 +846,7 @@ describe('mcp-client', () => {
 
       it('should handle circular references in anyOf', () => {
         const schema = {
-          anyOf: [
-            { $ref: '#/definitions/NodeA' },
-            { type: 'string' },
-          ],
+          anyOf: [{ $ref: '#/definitions/NodeA' }, { type: 'string' }],
           definitions: {
             NodeA: {
               type: 'object',
--- a/packages/core/src/tools/mcp-client.test.ts
+++ b/packages/core/src/tools/mcp-client.test.ts
@@ -151,220 +151,6 @@ describe('mcp-client', () => {
       const consoleErrorSpy = vi
         .spyOn(console, 'error')
         .mockImplementation(() => {});
-      vi.mocked(GenAiLib.mcpToTool).mockReturnValue({
-        tool: () =>
-          Promise.resolve({
-            functionDeclarations: [
-              {
-                name: 'validTool',
-                parametersJsonSchema: {
-                  type: 'object',
-                  properties: {},
-                },
-              },
-            ],
-          }),
-      } as unknown as GenAiLib.CallableTool);
-
-      const tools = await discoverTools('test-server', {}, mockedClient);
-
-      expect(tools.length).toBe(1);
-      expect(vi.mocked(DiscoveredMCPTool).mock.calls[0][2]).toBe('validTool');
-      expect(consoleWarnSpy).not.toHaveBeenCalled();
-      consoleWarnSpy.mockRestore();
-    });
-
-    it('should discover tool with $ref references in schema', async () => {
-      const mockedClient = {} as unknown as ClientLib.Client;
-      const consoleWarnSpy = vi
-        .spyOn(console, 'warn')
-        .mockImplementation(() => {});
-      vi.mocked(GenAiLib.mcpToTool).mockReturnValue({
-        tool: () =>
-          Promise.resolve({
-            functionDeclarations: [
-              {
-                name: 'toolWithRef',
-                parametersJsonSchema: {
-                  type: 'object',
-                  properties: {
-                    user: { $ref: '#/definitions/User' },
-                    address: { $ref: '#/definitions/Address' },
-                  },
-                  definitions: {
-                    User: {
-                      type: 'object',
-                      properties: {
-                        name: { type: 'string' },
-                        age: { type: 'number' },
-                      },
-                    },
-                    Address: {
-                      type: 'object',
-                      properties: {
-                        street: { type: 'string' },
-                        city: { type: 'string' },
-                      },
-                    },
-                  },
-                },
-              },
-            ],
-          }),
-      } as unknown as GenAiLib.CallableTool);
-
-      const tools = await discoverTools('test-server', {}, mockedClient);
-
-      expect(tools.length).toBe(1);
-      expect(vi.mocked(DiscoveredMCPTool).mock.calls[0][2]).toBe('toolWithRef');
-      expect(consoleWarnSpy).not.toHaveBeenCalled();
-      consoleWarnSpy.mockRestore();
-    });
-
-    it('should discover tool with complex $ref schema including arrays', async () => {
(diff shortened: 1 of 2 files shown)


jacob314:
@jakemac53 this seems up your alley to review.
@cfool you have some minor failing lints to address. After that we'll take a look.

Ice1984m (approved):

packages/core/src/tools/mcp-client.test.ts
```diff
@@ -806,22 +995,326 @@ describe('mcp-client', () => {
...
       };
-      expect(hasValidTypes(schema)).toBe(true);
+      expect(hasValidTypes(schema, schema)).toBe(true);
+    });
+
+    describe('$ref handling', () => {
+      it('should return true for a schema with local $ref reference', () => {
+        const schema = {
+          $ref: '#/definitions/User',
+        };
```
jakemac53, line 595: I am confused by this test, there is no corresponding definition for `#/definitions/User`? Seems like this one should expect `false` and the next test should expect `true` to me.

packages/core/src/tools/mcp-client.test.ts
```diff
@@ -806,22 +995,326 @@ describe('mcp-client', () => {
...
+              },
+            },
+          },
+        };
+        expect(hasValidTypes(schema, schema)).toBe(false);
+      });
+
+      it('should return true for a schema with external $ref reference', () => {
+        const schema = {
+          $ref: 'https://example.com/schema.json#/definitions/User',
```
jakemac53, line 617: Are these allowed by Gemini generally?

packages/core/src/tools/mcp-client.test.ts
```diff
@@ -806,22 +995,326 @@ describe('mcp-client', () => {
...
+
+      it('should return true for a schema with nested $ref in properties', () => {
+        const schema = {
+          type: 'object',
+          properties: {
+            user: { $ref: '#/definitions/User' },
+            address: { $ref: '#/definitions/Address' },
+          },
+        };
+        expect(hasValidTypes(schema, schema)).toBe(true);
```
jakemac53: Same as above, should this be true even though the references are actually not present in the schema? Same for others below.

packages/core/src/tools/mcp-client.test.ts
```diff
@@ -806,22 +995,326 @@ describe('mcp-client', () => {
...
+
+    it('should return null for invalid path format', () => {
+      const schema = {
+        definitions: {
+          User: {
+            type: 'object',
+          },
+        },
+      };
+      const result = getRefDefinition('#/invalid/path', schema);
```
jakemac53: This isn't actually an invalid format, it's not really any different from the `non-existent` example above.

You would want to test with some invalid characters or something like that.

packages/core/src/tools/mcp-client.test.ts
```diff
@@ -806,22 +995,326 @@ describe('mcp-client', () => {
...
+      expect(result).toBeNull();
+    });
+
+    it('should handle empty root schema', () => {
```
jakemac53, line 953:
```suggestion
    it('should return null for empty root schema', () => {
```

packages/core/src/tools/mcp-client.test.ts
```diff
@@ -806,22 +995,326 @@ describe('mcp-client', () => {
...
+      const result = getRefDefinition('#/definitions/User', {});
+      expect(result).toBeNull();
+    });
+
+    it('should handle undefined root schema', () => {
```
jakemac53, line 958:
```suggestion
    it('should return null for undefined root schema', () => {
```

packages/core/src/tools/mcp-client.ts
```diff
@@ -430,16 +430,59 @@ export async function connectAndDiscover(
...
+ *   definitions: {
+ *     User: { type: "object", properties: { name: { type: "string" } } }
+ *   }
+ * };
+ * const userDef = getRefDefinition("#/definitions/User", schema);
+ * // Returns: { type: "object", properties: { name: { type: "string" } } }
+ * ```
+ */
+export function getRefDefinition(
+  path: `#/${string}`,
```
jakemac53: I am not super familiar with typescript, is this basically just statically enforcing a certain pattern for the string arguments?

If so, do we need the runtime check at the top of the function?

packages/core/src/tools/mcp-client.ts
```diff
@@ -430,16 +430,59 @@ export async function connectAndDiscover(
...
+  rootSchema?: Record<string, unknown>,
+): unknown {
+  if (!path.startsWith('#/')) {
+    return null;
```
jakemac53, line 585: It seems a bit odd we would just drop these on the floor, maybe we should be warning at least?

packages/core/src/tools/mcp-client.ts
```diff
@@ -430,16 +430,59 @@ export async function connectAndDiscover(
...
+  }
+  const subPaths = path.substring('#/'.length).split('/');
+  return subPaths.reduce((schema: unknown, prop: string) => {
+    if (typeof schema !== 'object' || schema === null) {
+      return null;
+    }
+    // Unescape JSON Pointer path segments.
+    const unescapedProp = prop.replace(/~1/g, '/').replace(/~0/g, '~');
```
jakemac53, line 593: Can you elaborate on what exactly is happening here? Maybe provide a link to what JSON Pointers are? Why do we only care about 0 and 1, are these array indexes or something else?

jakemac53:
> 2\. The `hasValidTypes` function is susceptible to stack overflows from circular references and incorrectly validates dangling references.

I think this is still an issue currently? Keeping a set of already validated references should resolve it (by skipping the validate call for those).

packages/core/src/tools/mcp-client.ts
cfool: Because the characters '\~' (%x7E) and '/' (%x2F) have special meanings in JSON Pointer, '~' needs to be encoded as '~0' and '/' needs to be encoded as '~1' when these characters appear in a reference token. (https://datatracker.ietf.org/doc/html/rfc6901#section-3)

cfool:
@jakemac53 I will adjust the valid method for $ref.
1. Only internal refs are supported; external refs will be considered invalid.
2. Invalid references (referencing non-existent objects) will be considered invalid.
3. Added detection for circular references.
4. getRefDefinition.path, i will change it to string

chrstnb:

PR #6328, opened 2025-08-15
Revert #6088

galdawave:
Reverting child-process to node-pty changes, causing build errors for g3 and vscode extension


jacob314 (approved):
lgtm

PR #6335, opened 2025-08-15
Fix shell tool description to be os-specific

### LINKED ISSUE #4616: failed to delete a directory with rd command
### What happened?

gemini created a diractory and some files inside, then I ve asked to undo changes and I ve had error with the remove directory command, so impossible de remove the directory

remove the directory created few minutes ago

### Client information

* **CLI Version:** 0.1.13
* **Git Commit:** 412b78c5 (local modifications)
* **Memory Usage:** 173.3 MB

### Login information

_No response_

_No response_

scidomino:
## TLDR

Fixes the shell tool description to not always be linux-specific.

## Reviewer Test Plan

Ask GCLI to delete a file in windows and verify it uses "del" instead of "rm".

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | x  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes https://github.com/google-gemini/gemini-cli/issues/4616

### DIFF
--- a/packages/core/src/tools/shell.ts
+++ b/packages/core/src/tools/shell.ts
@@ -293,18 +293,8 @@ class ShellToolInvocation extends BaseToolInvocation<
   }
 }
 
-export class ShellTool extends BaseDeclarativeTool<
-  ShellToolParams,
-  ToolResult
-> {
-  static Name: string = 'run_shell_command';
-  private allowlist: Set<string> = new Set();
-
-  constructor(private readonly config: Config) {
-    super(
-      ShellTool.Name,
-      'Shell',
-      `This tool executes a given shell command as \`bash -c <command>\`. Command can start background processes using \`&\`. Command is executed as a subprocess that leads its own process group. Command process group can be terminated as \`kill -- -PGID\` or signaled as \`kill -s SIGNAL -- -PGID\`.
+function getShellToolDescription(): string {
+  const returnedInfo = `
 
       The following information is returned:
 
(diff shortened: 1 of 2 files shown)


packages/core/src/tools/shell.ts
```diff
@@ -311,14 +301,42 @@ export class ShellTool extends BaseDeclarativeTool<
       Exit Code: Exit code or \`(none)\` if terminated by signal.
       Signal: Signal number or \`(none)\` if no signal was received.
       Background PIDs: List of background processes started or \`(none)\`.
-      Process Group PGID: Process group started or \`(none)\``,
+      Process Group PGID: Process group started or \`(none)\\`;
```
jacob314: can we add a golden tests for windows and non windows  it is easier to visually make sure the output looks like we would want on both platforms?
I would nudge gemini cli to align this to the golden tests already available for the system prompt.

jacob314 (approved):
Can you follow up using the snapshot files like in the system prompt test? That will make it easier for people to update the goldens.

scidomino:
> Can you follow up using the snapshot files like in the system prompt test? That will make it easier for people to update the goldens.

Oh! My bad. I didn't understand what you meant. Will do!

PR #6337, opened 2025-08-15
fix: implement AbortSignal support for slash command cancellation (#6066)

### LINKED ISSUE #6066: Slash commands do not respect top-level cancellation signal
### Tl;Dr

The main application's `AbortController` signal, which is triggered when the user cancels an ongoing operation (e.g., by pressing the `Escape` key), is not propagated to the `action` function of slash commands.

As a result, any long-running asynchronous operations within a slash command (such as network requests, file system access, or child processes) will continue to execute even after the user has requested cancellation. This can lead to unexpected behavior, orphaned processes, and unnecessary resource consumption.

### Analysis of the Current Implementation

1.  **Top-Level Abort Controller:** In `packages/cli/src/ui/hooks/useGeminiStream.ts`, a new `AbortController` is correctly created in the `submitQuery` function for each user request. The `cancelOngoingRequest` function properly calls `.abort()` on this controller.

2.  **Signal Propagation Gap:** The `abortSignal` from this controller is passed to the `geminiClient.sendMessageStream`, but it is **not** passed down to the `handleSlashCommand` function located in `packages/cli/src/ui/hooks/slashCommandProcessor.ts`.

3.  **Missing Context Property:** The `CommandContext` interface, defined in `packages/cli/src/ui/commands/types.ts`, does not have a property to hold an `AbortSignal`. This is the primary architectural gap, as there is no standard mechanism to make the signal available to command actions.

4.  **Isolated Command Logic:** As seen in `packages/cli/src/ui/commands/setupGithubCommand.ts`, commands that need cancellation logic are forced to create their own internal `AbortController`. This controller is completely disconnected from the main application's cancellation flow, making it impossible to respond to a global cancel event.

### Proposed Solution

To fix this, we need to plumb the `AbortSignal` from the top-level stream handler down into the execution context of every slash command.

1.  **Update `CommandContext`:**
    Modify the `CommandContext` interface in `packages/cli/src/ui/commands/types.ts` to include the signal.

    ```typescript
    // In packages/cli/src/ui/commands/types.ts
    export interface CommandContext {
      // An abort signal that is tied to the lifecycle of the current user prompt.
      signal: AbortSignal;
      // ... existing properties
    }
    ```

2.  **Pass the Signal Down the Call Stack:**
    *   In `useGeminiStream.ts`, update the call to `handleSlashCommand` to pass the `abortSignal`.
      ```typescript
      // In useGeminiStream.ts, inside `prepareQueryForGemini`
      const slashCommandResult = await handleSlashCommand(trimmedQuery, abortSignal);
      ```
    *   Update the `handleSlashCommand` function signature in `slashCommandProcessor.ts` to accept the signal.
      ```typescript
      // In slashCommandProcessor.ts
      const handleSlashCommand = useCallback(
        async (
          rawQuery: PartListUnion,
          signal: AbortSignal, // Add this parameter
          oneTimeShellAllowlist?: Set<string>,
          overwriteConfirmed?: boolean,
        ): Promise<SlashCommandProcessorResult | false> => {
          // ...
        }
      );
      ```
    *   In `slashCommandProcessor.ts`, include the signal when creating the `fullCommandContext`.
      ```typescript
      // In slashCommandProcessor.ts
      const fullCommandContext: CommandContext = {
        ...commandContext,
        signal, // Add the signal here
        invocation: { /* ... */ },
        overwriteConfirmed,
      };
      ```

3.  **Refactor Slash Commands to Use the Context Signal:**
    Update commands with long-running operations to use the signal from the context instead of creating their own.

    *Example: Refactoring `setupGithubCommand.ts`*
    ```typescript
    // Before
    const abortController = new AbortController();
    const response = await fetch(endpoint, {
      // ...
      signal: AbortSignal.any([
        AbortSignal.timeout(30_000),
        abortController.signal,
      ]),
    });

    // After
    const response = await fetch(endpoint, {
      // ...
      signal: AbortSignal.any([
        AbortSignal.timeout(30_000),
        context.signal, // Use the signal from the context
      ]),
    });
    ```

### Acceptance Criteria

- The `AbortSignal` from `useGeminiStream` is successfully passed down and is available in the `CommandContext` provided to all slash command `action` functions.
- Long-running slash commands, such as `/setup-github`, correctly use the `context.signal` for their asynchronous operations. We should evaluate each slash command including commands via Custom Commands to ensure abort works properly.
- The application remains stable, and other commands are unaffected.

akhil29:
## Summary

Implements proper AbortSignal propagation for slash commands to enable cancellation when users press Escape, addressing issue #6066.

## Changes

- **Add signal property to CommandContext interface** - Provides AbortSignal to all slash command actions
- **Update handleSlashCommand signature** - Accepts AbortSignal parameter for proper signal propagation  
- **Pass abortSignal from useGeminiStream** - Connects user cancellation (Escape key) to slash commands
- **Refactor setupGithubCommand** - Uses context.signal instead of isolated AbortController
- **Fix keyboard shortcuts** - Provide default signals for backwards compatibility
- **Comprehensive test coverage** - Tests signal propagation, cancellation scenarios, and AbortSignal.any() combinations

## Technical Implementation

The solution follows the exact **Proposed Solution** from issue #6066:

1. **CommandContext Enhancement**: Added `signal: AbortSignal` property with proper JSDoc comment
2. **Signal Propagation Chain**: `useGeminiStream` → `handleSlashCommand` → `CommandContext` → command actions  
3. **AbortSignal.any() Integration**: Combines timeout and user cancellation signals in setupGithubCommand
4. **Backwards Compatibility**: Default signals ensure existing commands continue working

## Test Plan

- ✅ All existing tests pass
- ✅ New tests verify signal propagation through entire call chain
- ✅ Cancellation tests confirm Escape key stops operations
- ✅ setupGithubCommand tests validate AbortSignal.any() behavior
- ✅ Mock context updated with required signal property

## Acceptance Criteria Met

- ✅ AbortSignal from useGeminiStream successfully propagated to CommandContext
- ✅ Long-running commands like /setup-github use context.signal for operations
- ✅ Application remains stable with no impact to existing functionality
- ✅ Users can cancel slash commands by pressing Escape key

Fixes #6066

### DIFF
--- a/packages/cli/src/ui/App.tsx
+++ b/packages/cli/src/ui/App.tsx
@@ -592,9 +592,9 @@ const App = ({ config, settings, startupWarnings = [], version }: AppProps) => {
     (result: IdeIntegrationNudgeResult) => {
       if (result.userSelection === 'yes') {
         if (result.isExtensionPreInstalled) {
-          handleSlashCommand('/ide enable');
+          handleSlashCommand('/ide enable', new AbortController().signal);
         } else {
-          handleSlashCommand('/ide install');
+          handleSlashCommand('/ide install', new AbortController().signal);
         }
         settings.setValue(
           SettingScope.User,
@@ -632,7 +632,7 @@ const App = ({ config, settings, startupWarnings = [], version }: AppProps) => {
           clearTimeout(timerRef.current);
         }
         // Directly invoke the central command handler.
-        handleSlashCommand('/quit');
+        handleSlashCommand('/quit', new AbortController().signal);
       } else {
         setPressedOnce(true);
         timerRef.current = setTimeout(() => {
@@ -660,7 +660,10 @@ const App = ({ config, settings, startupWarnings = [], version }: AppProps) => {
 
         const mcpServers = config.getMcpServers();
         if (Object.keys(mcpServers || {}).length > 0) {
-          handleSlashCommand(newValue ? '/mcp desc' : '/mcp nodesc');
+          handleSlashCommand(
+            newValue ? '/mcp desc' : '/mcp nodesc',
+            new AbortController().signal,
+          );
         }
       } else if (
         keyMatchers[Command.TOGGLE_IDE_CONTEXT_DETAIL](key) &&
@@ -668,7 +671,7 @@ const App = ({ config, settings, startupWarnings = [], version }: AppProps) => {
         ideContextState
       ) {
         // Show IDE status when in IDE mode and context is available.
-        handleSlashCommand('/ide status');
+        handleSlashCommand('/ide status', new AbortController().signal);
       } else if (keyMatchers[Command.QUIT](key)) {
         // When authenticating, let AuthInProgress component handle Ctrl+C.
         if (isAuthenticating) {
--- a/packages/cli/src/ui/hooks/slashCommandProcessor.ts
+++ b/packages/cli/src/ui/hooks/slashCommandProcessor.ts
@@ -158,6 +158,7 @@ export const useSlashCommandProcessor = (
   );
   const commandContext = useMemo(
     (): CommandContext => ({
+      signal: new AbortController().signal, // Default signal, will be overridden in handleSlashCommand
       services: {
         config,
         settings,
@@ -233,6 +234,7 @@ export const useSlashCommandProcessor = (
   const handleSlashCommand = useCallback(
     async (
       rawQuery: PartListUnion,
+      signal: AbortSignal,
       oneTimeShellAllowlist?: Set<string>,
       overwriteConfirmed?: boolean,
     ): Promise<SlashCommandProcessorResult | false> => {
@@ -303,6 +305,7 @@ export const useSlashCommandProcessor = (
           if (commandToExecute.action) {
             const fullCommandContext: CommandContext = {
               ...commandContext,
+              signal,
               invocation: {
                 raw: trimmed,
                 name: commandToExecute.name,
@@ -432,6 +435,7 @@ export const useSlashCommandProcessor = (
 
                   return await handleSlashCommand(
                     result.originalInvocation.raw,
+                    signal,
                     // Pass the approved commands as a one-time grant for this execution.
                     new Set(approvedCommands),
                   );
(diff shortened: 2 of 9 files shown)


akhil29:
Basic Testing:
Screenshot 2025-08-15 at 10 41 22 AM

Please also recommend other ways to test this.

packages/cli/src/ui/hooks/slashCommandProcessor.ts
```diff
@@ -158,6 +158,7 @@ export const useSlashCommandProcessor = (
   );
   const commandContext = useMemo(
     (): CommandContext => ({
+      signal: new AbortController().signal, // Default signal, will be overridden in handleSlashCommand
```
NTaylorMullen, line 161: Hmm isa there still value in memoizing this since we just re-expand in `handleSlashCommand` this has a bit of code smell?

packages/cli/src/ui/App.tsx
```diff
@@ -592,9 +592,9 @@ const App = ({ config, settings, startupWarnings = [], version }: AppProps) => {
     (result: IdeIntegrationNudgeResult) => {
       if (result.userSelection === 'yes') {
         if (result.isExtensionPreInstalled) {
-          handleSlashCommand('/ide enable');
+          handleSlashCommand('/ide enable', new AbortController().signal);
```
NTaylorMullen, line 595: Do we have a app-level abort signal so if the user kills the app it gets triggered? Ultimately everything should be cancellable at some layer. OR at the very least this indicates that the `/ide` commands can't actually be interacted with via `esc` /cc @skeshive

packages/cli/src/ui/hooks/slashCommandProcessor.ts
scidomino: I *think* the right thing to do here is turn this into a useCallback so you can pass in the abortSignal as an argument

scidomino:
We are trying to clean out our pending PRs so I will close this for now. reopen when you are ready for review.

akhil29:
Reopening this.

akhil29:
couldn't reopen the issue as i didn't have access.

Create a new pr https://github.com/google-gemini/gemini-cli/pull/7790

PR #6348, opened 2025-08-15
feat(oauth): Make oauth client a singleton to survive cache failures

mboshernitsan:
The getOauthClient function used to rely on an on-disk cache to persist credentials across multiple invocations. However, if the cache fails to be written (e.g. due to file system permissions), subsequent calls to getOauthClient would re-trigger the entire authentication flow, resulting in a poor user experience at best and in the appearance of Gemini CLI hanging at worst (see issue #4546).

This change refactors getOauthClient to be a singleton, ensuring that the oauth client is initialized only once per process. This makes the function more robust and resilient to cache failures.  This should fix at least some of the auth failures mentioned in #4546).

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | x  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

*Example: `This PR makes progress on #4546` and possibly resolves it.  Due to the difficulty of testing this case, this will need to be confirmed by issue reporters.


scidomino (commented):
I have basically the same comments as Gemini I know it's unlikely to happen in practice but I'd prefer if we had some kind of checking to see if the proxy and auth type are the same.

mboshernitsan:
> I have basically the same comments as Gemini I know it's unlikely to happen in practice but I'd prefer if we had some kind of checking to see if the proxy and auth type are the same.

Done.  I don't think proxy can change within the lifetime of the process, but changing auth type is possible, so I implemented the suggested map

packages/core/src/code_assist/oauth2.ts
```diff
@@ -416,3 +428,7 @@ async function fetchAndCacheUserInfo(client: OAuth2Client): Promise<void> {
     console.error('Error retrieving user info:', error);
   }
 }
+
+export function resetOauthClient() {
```
NTaylorMullen: should add a comment + maybe change the name to indicate that this is for testing only

NTaylorMullen (approved):

packages/core/src/code_assist/oauth2.ts
mboshernitsan: Done.

PR #6363, opened 2025-08-16
bug(core): Do not throw validation errors when building tools in `nonInteractiveToolExecutor`.

joshualitt:
1) Fixes a bug introduced in https://github.com/google-gemini/gemini-cli/pull/5613 
2) adds extensive tests of the corrected behavior to ensure against backsliding.
3) Also stops terminating on caught UNHANDLED_EXCEPTIONs when run non-interactively. This ensures alignment between how subagents and the non-interactive CLI treat these errors.

### DIFF
--- a/packages/cli/src/nonInteractiveCli.test.ts
+++ b/packages/cli/src/nonInteractiveCli.test.ts
@@ -150,7 +150,7 @@ describe('runNonInteractive', () => {
     expect(processStdoutSpy).toHaveBeenCalledWith('\n');
   });
 
-  it('should handle error during tool execution', async () => {
+  it('should handle error during tool execution and should send error back to the model', async () => {
     const toolCallEvent: ServerGeminiStreamEvent = {
       type: GeminiEventType.ToolCallRequest,
       value: {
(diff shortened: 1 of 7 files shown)


packages/core/src/tools/tools.ts
```diff
@@ -211,6 +211,20 @@ export abstract class DeclarativeTool<
...
+  async validateBuildAndExecute(
+    args: TParams,
+    abortSignal: AbortSignal,
+  ): Promise<ToolResult> {
+    const validationError = this.validateToolParams(args);
+    if (validationError) {
+      return {
+        llmContent: `Error: Invalid parameters provided. Reason: ${validationError}`,
+        returnDisplay: validationError,
+      };
```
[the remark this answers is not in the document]
joshualitt: Ack!

FYI, this is how error handling would have appeared before this logic changed: https://github.com/google-gemini/gemini-cli/blob/c1fe6889569610878c45216556fb99424b5bcba4/packages/core/src/tools/read-file.ts#L133

```
  async execute(
    params: ReadFileToolParams,
    _signal: AbortSignal,
  ): Promise<ToolResult> {
    const validationError = this.validateToolParams(params);
    if (validationError) {
      return {
        llmContent: `Error: Invalid parameters provided. Reason: ${validationError}`,
        returnDisplay: validationError,
      };
    }
    ...
```

packages/cli/src/nonInteractiveCli.test.ts
```diff
@@ -150,7 +150,7 @@ describe('runNonInteractive', () => {
     expect(processStdoutSpy).toHaveBeenCalledWith('\n');
   });
 
-  it('should handle error during tool execution', async () => {
+  it('should handle error during tool execution and should send error back to the model', async () => {
```
joshualitt, line 153: @anj-s it looks like the executor already formats the response correctly, and so the logic to create a function response(here: https://github.com/google-gemini/gemini-cli/compare/main...gb-11-6022-5316-fix#diff-0a0a9771217246ff47414e54edfaf48f2212ee2b98c17311adc378f9c26c1ca1R23) may not be necessary. This test demonstrates the correction flow.

packages/core/src/tools/tools.ts
```diff
@@ -200,6 +200,61 @@ export abstract class DeclarativeTool<
     const invocation = this.build(params);
     return invocation.execute(signal, updateOutput);
   }
+
+  /**
+   * Similar to `build` but never throws.
+   * @param params The raw, untrusted parameters from the model.
+   * @returns A `ToolInvocation` instance.
+   */
+  silentBuild(params: TParams): ToolInvocation<TParams, TResult> | Error {
```
jacob314: does this need to be public?

packages/core/src/tools/tools.ts
```diff
@@ -200,6 +200,61 @@ export abstract class DeclarativeTool<
...
+   * @param params The raw, untrusted parameters from the model.
+   * @params abortSignal a signal to abort.
+   * @returns The result of the tool execution.
+   */
+  async validateBuildAndExecute(
+    params: TParams,
+    abortSignal: AbortSignal,
+  ): Promise<ToolResult> {
+    try {
+      const invocationOrError = this.silentBuild(params);
```
jacob314: can we either remove silentBuild or remove the extra try-catch? Seems like we now have more try catch blocks than needed.

jacob314 (approved):
great to see us making this more robust. approved with a few nits about uneeded logic.

packages/cli/src/nonInteractiveCli.test.ts
anj-s: Yes, agreed. I realized that as well but wanted to make sure that the toolRespose we created mirrored interactive mode.

anj-s (approved):

packages/core/src/tools/tools.ts
joshualitt: Good catch, done!

packages/core/src/tools/tools.ts
joshualitt: Per offline discussions, moved `silentBuild` outside of the `try / catch` given it doesn't throw.

PR #6380, opened 2025-08-16
feat(core): Support configuring Gemini API base URL via environment variable

xiaoguomeiyitian:
This pull request adds support for overriding the default Gemini API endpoint by using the GEMINI_BASEURL environment variable.

  This change allows users to specify a custom base URL, which is useful for scenarios such as:
   - Connecting to the Gemini API through a corporate proxy.
   - Using a different API endpoint for testing or development purposes.
   - Accessing regionally specific or private Gemini API instances.

  The implementation retrieves the baseUrl from process.env.GEMINI_BASEURL and passes it into the httpOptions for the GoogleGenAI client. If the environment variable is not set, it defaults to undefined, ensuring no change in the existing behavior.

  Thank you for considering this contribution. I would be very grateful if you could merge this change.

### DIFF
--- a/packages/core/src/core/contentGenerator.ts
+++ b/packages/core/src/core/contentGenerator.ts
@@ -109,6 +109,7 @@ export async function createContentGenerator(
 ): Promise<ContentGenerator> {
   const version = process.env.CLI_VERSION || process.version;
   const httpOptions = {
+    baseUrl: process.env.GEMINI_BASEURL || undefined,
     headers: {
       'User-Agent': `GeminiCLI/${version} (${process.platform}; ${process.arch})`,
     },


2077DevWave:
This is a great suggestion! It seems like this functionality is already available in the Gemini CLI. You can achieve the same result by setting the `GOOGLE_GEMINI_BASE_URL` environment variable.

Here's how it works:

When you set the `GOOGLE_GEMINI_BASE_URL` environment variable, the Gemini CLI will use the value of this variable as the base URL for all API requests to the Gemini model. This allows you to direct the CLI to a different endpoint, such as a custom or private instance of the Gemini API, instead of the default Google-hosted service.

For example, you can set the environment variable in your terminal like this:

**For Linux/macOS:**
```bash
export GOOGLE_GEMINI_BASE_URL="your_custom_url_here"
```

**For Windows (Command Prompt):**
```bash
set GOOGLE_GEMINI_BASE_URL="your_custom_url_here"
```

**For Windows (PowerShell):**
```powershell
$env:GOOGLE_GEMINI_BASE_URL="your_custom_url_here"
```

After setting this variable, any `gemini` command you run in that terminal session will use your specified URL. This provides a straightforward way to use the Gemini CLI with different backends without modifying the tool's code.

ypenn21:
> GOOGLE_GEMINI_BASE_URL

How Come I can't find GOOGLE_GEMINI_BASE_URL in the source code on the main branch??

2077DevWave:
a GOOGLE_GEMINI_BASE_URL come with libraries that gemini-cli use to send api requests. u can find it in google/generative-ai library.

2077DevWave:
also the new version of gemini-cli have switch of --proxy that can be used to send request through proxy server.
for example if is your proxy server is 127.0.0.1 with port 2334 u can run grmini-cli with following command:

gemini --proxy https://127.0.0.1:2334

PR #6399, opened 2025-08-16
feat(config): add shell.inheritEnv option

cornmander:
Adds a new configuration setting, `shell.inheritEnv`, to control whether the shell tool inherits environment variables from the Gemini CLI process.

This setting is `true` by default to maintain the existing behavior. It can be controlled via the `--shell-inherit-env` command-line flag.

This PR exposes the configuration but does not yet implement the business logic.

## Dive Deeper

This change, after full implemented, will make it safer to use Gemini CLI against untrusted inputs.


chrstnb:

PR #6401, opened 2025-08-16
Add an integration test for checking the current env var behavior.

cornmander:
Adds an integration test to confirm current environment variable behavior in the shell tool. I am planning to make this customizable so this test is a starting point to confirm nothing breaks.


chrstnb:

PR #6403, opened 2025-08-16
Fix integration test.

cornmander:
Fixes integration test by closing stdin unconditionally. This might break interactive multi-turn tests, if we have any.


jacob314 (approved):
lgtm

PR #6407, opened 2025-08-17
Fix missing import in ide integration tests.

cornmander:
Fix a missing import in the IDE integration tests.

The tests still don't pass but at least they fail in the test logic rather than in scaffolding.


jacob314 (approved):
lgtm

PR #6417, opened 2025-08-17
Support IDE connections via stdio MCP

### LINKED ISSUE #6416: Support IDE connections over UNIX socket

Add support for IDE connections using stdio MCP over a UNIX socket.

### Why is this needed?

Today connections to an IDE are supported via a MCP HTTP SSE transport to localhost. It would be nice to support connecting via a UNIX socket to avoid listening on a port.

### Additional context

_No response_

bbiggs:
## TLDR

This change adds support for connecting to an IDE via stdio MCP

## Dive Deeper

We add support for new environment variables: GEMINI_CLI_IDE_SERVER_STDIO_COMMAND and GEMINI_CLI_IDE_SERVER_STDIO_ARGS to provide an app to run to set up a connection to an IDE as a stdio MCP server, also updating the PID file config to support providing the command and args.

## Reviewer Test Plan

Hopefully this feature is an add-on and I didn't break the exiting IDE integration.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes #6416


packages/core/src/ide/ide-client.ts
```diff
@@ -107,15 +109,23 @@ export class IdeClient {
...
     const portFromFile = await this.getPortFromFile();
     if (portFromFile) {
-      const connected = await this.establishConnection(portFromFile);
+      const connected = await this.establishHttpConnection(portFromFile);
+      if (connected) {
+        return;
+      }
+    }
+
+    const socketFromEnv = this.getSocketFromEnv();
```
cornmander: The env var discovery approach is meant as a fallback. Could you implement the file based approach as well?

cornmander (commented):
Can you add a test case for this in the integration test?

packages/core/src/ide/ide-client.ts
```diff
@@ -393,6 +412,52 @@ export class IdeClient {
...
+        version: '1.0.0',
+      });
+
+      const args =
+        command === 'socat' ? ['-', `UNIX-CONNECT:${socket}`] : ['-U', socket];
+
+      transport = new StdioClientTransport({
+        command,
+        args,
+      });
```
elmar-peise, line 515: (optional) Should we consider passing both of these to the CLI, that makes it the IDE's responsibility to figure out the command and allows for stdio servers that are not necessarily exposed as sockets.

packages/core/src/ide/ide-client.ts
skeshive: +1 or just do the file based approach (the env var fallback is mostly to support stale clients)

the env vars often end up being stale if a terminal session gets restored

bbiggs: I updated this to be more generic, working to validate but would like feedback on the approach.

packages/core/src/ide/ide-client.ts
```diff
@@ -298,16 +329,44 @@ export class IdeClient {
...
       const ideProcessId = await getIdeProcessId();
       const portFile = path.join(
         os.tmpdir(),
         `gemini-ide-server-${ideProcessId}.json`,
       );
       const portFileContents = await fs.promises.readFile(portFile, 'utf8');
-      const port = JSON.parse(portFileContents).port;
-      return port.toString();
+      const config = JSON.parse(portFileContents);
+      return config;
```
cornmander: return w/o the intermediate assignment

cornmander (approved):

PR #6431, opened 2025-08-17
Log start session event through telemetry/loggers

### LINKED ISSUE #6430: Session start event missing for OpenTelemetry
### What happened?

OpenTelemetry misses the CLI configuration event.

Upon app launch, OpenTelemetry should receive a CLI configuration event with the initial state.

### Client information

<details>

```console
$ gemini /about
# paste output here
```

</details>

### Login information

_No response_

_No response_

bbiggs:
## TLDR

OpenTelemetry misses the CLI configuration event. This routes through telemetry/loggers to fix.

## Linked issues / bugs

Fixes #6430

### DIFF
--- a/packages/core/src/config/config.ts
+++ b/packages/core/src/config/config.ts
@@ -337,9 +337,9 @@ export class Config {
       initializeTelemetry(this);
     }
 
+    logCliConfiguration(this, new StartSessionEvent(this));
+
     if (this.getUsageStatisticsEnabled()) {
-      logCliConfiguration(this, new StartSessionEvent(this));
-    } else {
       console.log('Data collection is disabled.');
     }
   }
--- a/packages/core/src/config/config.ts
+++ b/packages/core/src/config/config.ts
@@ -43,12 +43,11 @@ import {
   DEFAULT_GEMINI_EMBEDDING_MODEL,
   DEFAULT_GEMINI_FLASH_MODEL,
 } from './models.js';
-import { ClearcutLogger } from '../telemetry/clearcut-logger/clearcut-logger.js';
 import { shouldAttemptBrowserLaunch } from '../utils/browser.js';
 import { MCPOAuthConfig } from '../mcp/oauth-provider.js';
 import { IdeClient } from '../ide/ide-client.js';
 import type { Content } from '@google/genai';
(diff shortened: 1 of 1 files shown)


packages/core/src/config/config.ts
```diff
@@ -339,9 +338,7 @@ export class Config {
     }
 
     if (this.getUsageStatisticsEnabled()) {
-      ClearcutLogger.getInstance(this)?.logStartSessionEvent(
-        new StartSessionEvent(this),
-      );
+      logCliConfiguration(this, new StartSessionEvent(this));
     } else {
       console.log('Data collection is disabled.');
     }
```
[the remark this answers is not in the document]
owenofbrien: This is a good point.

owenofbrien (changes requested):
The GCA review bot made a good point about the usageStatisticsEnabled flag not being intended for OTEL enablement.

packages/core/src/config/config.ts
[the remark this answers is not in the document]
bbiggs: done

mattKorwel (approved):

packages/core/src/config/config.ts
```diff
@@ -338,11 +337,9 @@ export class Config {
       initializeTelemetry(this);
     }
 
+    logCliConfiguration(this, new StartSessionEvent(this));
+
     if (this.getUsageStatisticsEnabled()) {
-      ClearcutLogger.getInstance(this)?.logStartSessionEvent(
```
anj-s, line 342: why do we not need this anymore?

bbiggs: it happens here:
  https://github.com/google-gemini/gemini-cli/blob/main/packages/core/src/telemetry/loggers.ts#L61

anj-s: right, but this is no longer dependent on the getUsageStatisticsEnabled flag. If we want to decouple the two, should we have that initialization within this codeblock?

bbiggs: it is dependent - getUsageStatisticsEnabled is checked within clearcut, so if that's disabled it's already the case that the telemetry will not be sent.

owenofbrien (approved):

PR #6453, opened 2025-08-18
feat(cli): Add --allowed-tools flag to bypass tool confirmation (#2417)

werdnum:
This PR introduces the `--allowed-tools` flag and a corresponding `allowedTools` setting, allowing users to specify tools that can run without confirmation. This is useful for trusted and frequently used commands like `git status`.

I have chosen `--allowed-tools` (instead of `--auto-approved-tools` as I had initially) because at the moment its functionality is quite complementary to `coreTools` (if you are in YOLO/unattended mode, `coreTools` is useful and `allowedTools` is not, and if you're in interactive mode, `coreTools` is not that useful but `allowedTools` is), so we may want to merge `coreTools` into `allowedTools` – and `allowedTools` works well for both use cases.

Addresses #2417

**Key Changes:**

*   **New Feature:**
    *   Adds the `--allowed-tools` command-line flag.
    *   Adds the `allowedTools` setting to the configuration schema.
    *   Updates documentation in `docs/cli/configuration.md`.

*   **Refactoring & Bug Fix:**
    *   To support this feature correctly for shell commands with arguments, the tool filtering logic was refactored.
    *   The core matching logic is now centralized in the `doesToolInvocationMatch` function located in `packages/core/src/utils/shell-utils.ts`.
    *   This refactoring unifies how `coreTools`, `excludedTools`, and the new `allowedTools` are handled, resolving a circular dependency.
    *   The redundant `packages/core/src/utils/tool-utils.ts` file has been removed.
    *   Fixes a bug where matching shell command patterns with arguments (e.g., `run_shell_command(git status)`) was not working correctly.

### DIFF
--- a/packages/cli/src/config/settingsSchema.ts
+++ b/packages/cli/src/config/settingsSchema.ts
@@ -344,6 +344,16 @@ export const SETTINGS_SCHEMA = {
     description: 'Paths to core tool definitions.',
     showInDialog: false,
   },
+  allowedTools: {
+    type: 'array',
+    label: 'Allowed Tools',
+    category: 'Advanced',
+    requiresRestart: true,
+    default: undefined as string[] | undefined,
+    description:
+      'A list of tool names that will bypass the confirmation dialog.',
+    showInDialog: false,
+  },
   excludeTools: {
     type: 'array',
     label: 'Exclude Tools',
--- a/packages/core/src/core/coreToolScheduler.test.ts
+++ b/packages/core/src/core/coreToolScheduler.test.ts
@@ -5,6 +5,7 @@
  */
 
 import { describe, it, expect, vi } from 'vitest';
+import type { Mock } from 'vitest';
 import type { ToolCall, WaitingToolCall } from './coreToolScheduler.js';
 import {
   CoreToolScheduler,
@@ -99,6 +100,41 @@ class TestApprovalInvocation extends BaseToolInvocation<
   }
 }
 
+async function waitForStatus(
+  onToolCallsUpdate: Mock,
+  status: 'awaiting_approval' | 'executing' | 'success' | 'error' | 'cancelled',
+  timeout = 5000,
+): Promise<ToolCall> {
+  return new Promise((resolve, reject) => {
+    const startTime = Date.now();
+    const check = () => {
+      if (Date.now() - startTime > timeout) {
+        const seenStatuses = onToolCallsUpdate.mock.calls
+          .flatMap((call) => call[0])
+          .map((toolCall: ToolCall) => toolCall.status);
+        reject(
+          new Error(
+            `Timed out waiting for status "${status}". Seen statuses: ${seenStatuses.join(
+              ', ',
+            )}`,
+          ),
+        );
+        return;
+      }
+
+      const foundCall = onToolCallsUpdate.mock.calls
+        .flatMap((call) => call[0])
+        .find((toolCall: ToolCall) => toolCall.status === status);
+      if (foundCall) {
+        resolve(foundCall);
+      } else {
+        setTimeout(check, 10); // Check again in 10ms
+      }
+    };
+    check();
+  });
+}
+
 describe('CoreToolScheduler', () => {
   it('should cancel a tool call if the signal is aborted before confirmation', async () => {
     const mockTool = new MockTool();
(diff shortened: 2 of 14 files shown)


packages/core/src/utils/shell-utils.ts
```diff
@@ -4,10 +4,79 @@
...
+ * @param toolOrToolName The tool object or the name of the tool being invoked.
+ * @param invocation The invocation object for the tool.
+ * @param patterns A list of patterns to match against.
+ *   Patterns can be:
+ *   - A tool name (e.g., "ReadFileTool") to match any invocation of that tool.
+ *   - A tool name with a prefix (e.g., "ShellTool(git status)") to match
+ *     invocations where the arguments start with that prefix.
+ * @returns True if the invocation matches any pattern, false otherwise.
+ */
+export function doesToolInvocationMatch(
```
abhipatel12: Can we move this to a more central location or separate utility file? It seems to have uses outside of just shell commands? Or was there a reason to keep it here?

packages/cli/src/config/settingsSchema.ts
```diff
@@ -314,6 +314,16 @@ export const SETTINGS_SCHEMA = {
     description: 'Paths to core tool definitions.',
     showInDialog: false,
   },
+  autoApprovedTools: {
```
abhipatel12: Any thoughts on a different flag name? 

ideas:

`--approvedTools`
`--trustedTools`

packages/core/src/core/coreToolScheduler.test.ts
```diff
@@ -53,6 +53,7 @@ describe('CoreToolScheduler', () => {
       getUsageStatisticsEnabled: () => true,
       getDebugMode: () => false,
       getApprovalMode: () => ApprovalMode.DEFAULT,
+      getAutoApprovedTools: () => [],
```
abhipatel12: Can we add a test for the positive flow here? So something where the test would require approval, but b/c we added it to the approved list, it is bypassed?

abhipatel12 (commented):
Hey Andrew thanks for this PR!

Left a few comments, let me know what you think. Some naming questions about the flag + some relocation just in case you had thoughts on that before you made any changes!

packages/cli/src/config/settingsSchema.ts
werdnum: I had `allowedTools` initially. It matches Claude Code, but I thought it might be confusing in that it implies that *only* these tools are allowed (with or without confirmation). I was trying to err on the side of clarity even if it's a bit verbose, but happy to defer to your judgement/preferences (or those of your PM/UX on the Gemini CLI side).

Approved or trusted both sound fine. Trusted has a slight semantic gap in that the tool is not what's trusted, the LLM is trusted - the tool is safe.

Safe tools seems wrong too?

approved tools perhaps then? I don't want to spend ages on the name but it makes sense to get it right since it's part of the public interface.

packages/core/src/utils/shell-utils.ts
werdnum: Heh, Gemini initially put it in tool-utils and I had it merge into shell-utils for simplicity, but I don't feel strongly. I'll change it back.

werdnum:
Rebase was a bit of a pain.

Of note, I adjusted the order of some of the permissions checks (in particular we always check if a tool requires confirmation), this made one test fail because it was looking for that method to never be called. IMO that is kind of overspecified so I just deleted the assertion.

packages/cli/src/config/settingsSchema.ts
allenhutchison: I don't have a strong opinion here, but personally like the verbosity of auto-approve-tools since it does exactly what it says. Failing that then I'd go with allowedTools to try to stay consistent with norms set in the industry.

werdnum: Discussed with Alan. Rationale:

> if we eventually want to merge this into one setting (we think), that setting isn't going to be called coreTools, so if we name the new setting something that works for a combined setting, then the eventual migration will require only migrating one legacy setting. By a happy coincidence the most sensible name for a combined setting is allowedTools and we are consistent with Claude Code to boot.

allenhutchison (approved):
This is a great contribution and will really help people who want to allow list shell commands and sub-tools.

PR #6478, opened 2025-08-18
feat(chat): Enable non-blocking chat input

MSHIVVANI:
This commit resolves the 'stop-and-wait' user experience in the chat interface.

The input prompt is now always visible, allowing users to type their next message while the AI is generating a response.

This is achieved by:
1. Guarding the submission handler against new inputs while the app is busy (either streaming from the API or processing a slash command).
2. Removing the conditional rendering that previously made the input component disappear.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

### DIFF
--- a/packages/cli/src/ui/App.tsx
+++ b/packages/cli/src/ui/App.tsx
@@ -580,12 +580,16 @@ const App = ({ config, settings, startupWarnings = [], version }: AppProps) => {
   // Input handling
   const handleFinalSubmit = useCallback(
     (submittedValue: string) => {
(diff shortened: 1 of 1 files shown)


abhipatel12:
Hey thanks for the contribution! This change is already in progress here #6049

PR #6479, opened 2025-08-18
feat(logging): add session_id to all logs

leehagoodjames:
## TLDR

Adds `session_id` to all logs, to make it easier in downstream processes to enrich by metadata logged in the `StartSessionEvent`.

## Dive Deeper

Key changes:
- Adds `session_id` to default log metadata (which is included with all logs)
- Removes `EndSessionEvent` because it now longer serves a purpose. Logs should still be emitted on shutdown.

Note: some of the changes are self-annotated for context

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

### DIFF
--- a/packages/cli/src/ui/hooks/useToolScheduler.test.ts
+++ b/packages/cli/src/ui/hooks/useToolScheduler.test.ts
@@ -55,6 +55,7 @@ const mockConfig = {
   getApprovalMode: vi.fn(() => ApprovalMode.DEFAULT),
   getUsageStatisticsEnabled: () => true,
   getDebugMode: () => false,
+  getSessionId: () => 'my-session-id',
 };
 
 class MockToolInvocation extends BaseToolInvocation<object, ToolResult> {
--- a/packages/core/src/telemetry/clearcut-logger/clearcut-logger.test.ts
+++ b/packages/core/src/telemetry/clearcut-logger/clearcut-logger.test.ts
@@ -132,9 +132,41 @@ describe('ClearcutLogger', () => {
 
       const event = logger?.createLogEvent('abc', []);
 
-      expect(event?.event_metadata[0][0]).toEqual({
+      expect(event?.event_metadata[0]).toContainEqual({
         gemini_cli_key: EventMetadataKey.GEMINI_CLI_GOOGLE_ACCOUNTS_COUNT,
         value: '9001',
+      });
+    });
+
+    it('logs default metadata', () => {
+      const google_accounts = 123;
+      const session_id = 'my-session-id';
+      const surface = 'ide-1234';
+
+      const { logger } = setup({
+        lifetimeGoogleAccounts: google_accounts,
+        config: {
+          sessionId: session_id,
+        },
+      });
+
+      vi.stubEnv('SURFACE', surface);
+
+      const event = logger?.createLogEvent('abc', []);
+
+      expect(event?.event_metadata[0]).toContainEqual({
+        gemini_cli_key: EventMetadataKey.GEMINI_CLI_GOOGLE_ACCOUNTS_COUNT,
+        value: `${google_accounts}`,
+      });
+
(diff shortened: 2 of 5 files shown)


packages/core/src/telemetry/clearcut-logger/clearcut-logger.ts
```diff
@@ -153,6 +152,30 @@ export class ClearcutLogger {
     this.events = new FixedDeque<LogEventEntry[]>(Array, MAX_EVENTS);
   }
 
+  /**
+   * Adds default fields to data, and returns a new data array.  This fields
+   * should exist on all log events.
+   */
+  private addDefaultFields(data: EventValue[]): EventValue[] {
```
leehagoodjames, line 163: Note: made this a class method to access `this.config?.getSessionId()`

packages/cli/src/ui/hooks/useToolScheduler.test.ts
```diff
@@ -55,6 +55,7 @@ const mockConfig = {
   getApprovalMode: vi.fn(() => ApprovalMode.DEFAULT),
   getUsageStatisticsEnabled: () => true,
   getDebugMode: () => false,
+  getSessionId: () => 'my-session-id',
```
leehagoodjames, line 58: Note: this function now needs to exist in the mock, since the logged tool calls have `session_id` added to the logs

packages/core/src/telemetry/clearcut-logger/clearcut-logger.test.ts
```diff
@@ -132,20 +132,52 @@ describe('ClearcutLogger', () => {
...
+
     it('logs the current surface from a github action', () => {
       const { logger } = setup({});
 
       vi.stubEnv('GITHUB_SHA', '8675309');
 
       const event = logger?.createLogEvent('abc', []);
 
-      expect(event?.event_metadata[0][1]).toEqual({
+      expect(event?.event_metadata[0]).toContainEqual({
```
leehagoodjames, line 230: Note: the logs contain an array of key/value pairs. These tests should should not care about the order of metadata in that array, but rather the existence of key/value pairs in the array, which is why `toContainEqual` is now used

packages/core/src/telemetry/clearcut-logger/clearcut-logger.ts
jerop, line 1: to clarify, this exists in all opentelemetry logs, right? is that why there's no change there?

jerop: https://github.com/google-gemini/gemini-cli/blob/main/docs/telemetry.md#logs-and-metric-reference -> we also added surface to all OTEL logs and metrics, maybe we should update this to mention that it's not just session ID attribute in everything, but also surface 🤔

leehagoodjames: So there are two loggers, `Clearcut` and `OpenTelemetry`. These changes are for the `Clearcut` logger and not the `OpenTelemetry` logger. 

I believe `OpeTelemetry` has had [session.id](https://github.com/google-gemini/gemini-cli/blob/065eb7897d89033cfd6602c911ad5b01af1a161e/packages/core/src/telemetry/loggers.ts#L53) as a default field, while `Clearcut` did not already have this for all logs.

jerop:
@uttamkanodia14 @owenofbrien please take a look as this is affecting clearcut

leehagoodjames:
Closing in light of https://github.com/google-gemini/gemini-cli/pull/6423

Issue #5255, opened 2025-07-31
/copy hangs for a few minutes


obadiar:
### What happened?

Using the `/copy` command causes the CLI to hang for a few minutes (~3), even with very short output. The output is actually copied to the clipboard right away, though the command does not return until much later, during which the prompt input disappears, and the CLI becomes unusable. 

### What did you expect to happen?

The output is copied  to the clipboard within a couple of seconds, and I can interact with the CLI right away.

### Client information

<details>

```console
$ gemini /about
CLI Version      0.1.15        
Git Commit       bcce1e7b      
Model            gemini-2.5-pro
Sandbox          no sandbox    
OS               linux         
Auth Method      gemini-api-key

```

</details>

### Login information

API key

PR #6482, opened 2025-08-18
fix: copy command gets stuck

hritan:
## TLDR

/copy command gets stuck in linux. 

## Dive Deeper

Although the chat output is copied to the clipboard but the promise never resolves hence the UI seems stuck and chat box remains hidden.

After fix screenshot:
image

## Reviewer Test Plan

1. Do npm start in a Linux terminal.
2. Say "hi" to gemini. It will return an output.
3. Give /copy command in the input
4. You will now see an info note for successful copy and the input box is visible again for further chat.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes - https://github.com/google-gemini/gemini-cli/issues/5255


packages/cli/src/ui/utils/commandUtils.ts
```diff
@@ -29,22 +29,61 @@ export const isSlashCommand = (query: string): boolean => query.startsWith('/');
...
+              }`,
+            ),
+          );
+        }
       });
-      child.stdin.on('error', reject);
+
+      // Consume stdout/stdin to prevent blocking issues.
+      child.stdout.on('data', () => {});
+      child.stdin.on('error', () => {});
```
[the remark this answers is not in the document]
hritan: done

jacob314:
@shrutip90 is going to help look at this to figure out whether this is the right overall approach or putting a bandaid over the issue.

abolast:
> @shrutip90 is going to help look at this to figure out whether this is the right overall approach or putting a bandaid over the issue.将有助于研究这个问题，以弄清楚这是否是正确的总体方法，还是只是对问题的一种补救。

I've noticed that I get returned to the prompt if I wait long enough, or it happens immediately if I clear my clipboard.

packages/cli/src/ui/utils/commandUtils.ts
```diff
@@ -44,23 +46,29 @@ export const copyToClipboard = async (text: string): Promise<void> => {
...
+      if (child.stdin) {
+        child.stdin.on('error', reject);
+        child.stdin.write(text);
+        child.stdin.end();
+      } else {
+        reject(new Error('Child process has no stdin stream to write to.'));
+      }
     });
 
+  const linuxOptions: SpawnOptions = { stdio: ['pipe', 'inherit', 'inherit'] };
```
shrutip90: Please add a comment on what these options will enable.

hritan: done

packages/cli/src/ui/utils/commandUtils.ts
```diff
@@ -44,23 +46,33 @@ export const copyToClipboard = async (text: string): Promise<void> => {
...
+  // - stdin: 'pipe' to write the text that needs to be copied.
+  // - stdout: 'inherit' since we don't need to capture the command's output on success.
+  // - stderr: 'pipe' to capture error messages (e.g., "command not found") for better error handling.
+  const linuxOptions: SpawnOptions = { stdio: ['pipe', 'inherit', 'pipe'] };
+
   switch (process.platform) {
     case 'win32':
       return run('clip', []);
     case 'darwin':
       return run('pbcopy', []);
```
jacob314, line 68: out of curiousity, is there a reason we shouldn't also be piping stderr for pbcopy. on mac. would imagine it might also send error output to stderr.

packages/cli/src/ui/utils/commandUtils.ts
```diff
@@ -29,11 +29,13 @@ export const isSlashCommand = (query: string): boolean => query.startsWith('/');
...
 // Copies a string snippet to the clipboard for different platforms
 export const copyToClipboard = async (text: string): Promise<void> => {
-  const run = (cmd: string, args: string[]) =>
+  const run = (cmd: string, args: string[], options?: SpawnOptions) =>
     new Promise<void>((resolve, reject) => {
-      const child = spawn(cmd, args);
+      const child = options ? spawn(cmd, args, options) : spawn(cmd, args);
       let stderr = '';
-      child.stderr.on('data', (chunk) => (stderr += chunk.toString()));
+      if (child.stderr) {
```
jacob314, line 36: why is child.stderr sometimes nullable?

jacob314:
It would have helped me to understand if the pr description included a summary of why the fix worked in this case because it was not obvious from the code. My assumption from asking Gemini CLi was the reason for the fix is the follows but that might be hallucinated nonsense.


  When Node.js's spawn function runs a command, it creates pipes for the new process's
  standard input (stdin), standard output (stdout), and standard error (stderr). This
  allows the parent Node.js process to send data to and receive data from the child
  process.

   * Before the fix: The stdout of xclip was a pipe. Our code wrote the text to xclip's
     stdin but never read from its `stdout` pipe.
   * The Deadlock: The xclip command, after receiving the text and copying it, doesn't
     immediately exit in some environments. It might wait for its stdout pipe to be closed
     by the parent process. If the pipe's buffer fills up (even with a small amount of
     data), xclip will pause and wait indefinitely for the parent process to read the data.
     Since our code never reads it, xclip gets stuck.
   * The Result: Because xclip is stuck, it never exits. Because it never exits, the 'close'
     event is never fired in our Node.js code. Because the 'close' event is never fired, the
     Promise never resolves, and the UI appears to hang.

jacob314 (approved):
lgtm

packages/cli/src/ui/utils/commandUtils.ts
hritan: No, we don't need to add explicit options for the other platforms. 
For Windows (clip) and macOS (pbcopy), the spawn function uses Node.js's default stdio configuration, which is ['pipe', 'pipe', 'pipe'].

The default behavior already does exactly what we need for those platforms, so creating a separate options object would be redundant. We only needed linuxOptions to change stdin and stderr from the problematic 'inherit' to 'pipe'.

packages/cli/src/ui/utils/commandUtils.ts
hritan: It's just to prevent the type check from going crazy, in our case it can never be null with the current options setting.

PR #6483, opened 2025-08-18
feat(ci): add retry to test job

cornmander:
We have test flakes, especially on windows, and retrying often fixes the problem. This adds retries to our ci workflow to retry automatically.


jacob314:
Should we instead prioritize fixing these flakes rather than hiding them?
I'm fine with landing this as long as we also add some mechanism to ensure we are auditing and resolving the flakes over time.

.github/workflows/ci.yml
```diff
@@ -253,6 +253,11 @@ jobs:
       pull-requests: 'write'
     strategy:
       fail-fast: false # So we can see all test failures
+      retries: 2
```
NTaylorMullen, line 260: I thought that we already had each test logically retry itself 3x don't we?

NTaylorMullen (approved):

.github/workflows/ci.yml
cornmander: Yes, but I've seen the runners fail to start (especially windows) and so it never hits the test retry logic.

cornmander:
> Should we instead prioritize fixing these flakes rather than hiding them? I'm fine with landing this as long as we also add some mechanism to ensure we are auditing and resolving the flakes over time.

Yes, although I don't see built-in github function to filter for tests that were retried.

SandyTao520:
Getting
```
 Check failure on line 1 in .github/workflows/ci.yml


GitHub Actions
/ .github/workflows/ci.yml
Invalid workflow file

(Line: 256, Col: 7): Unexpected value 'retries', (Line: 257, Col: 7): Unexpected value 'retry-on'
```

chrstnb:

PR #6496, opened 2025-08-18
feat: Show /ide subcommands based on connection status instead of ideMode boolean

skeshive:
Show /ide subcommands based on connection status instead of ideMode boolean


jacob314 (approved):
lgtm

PR #6507, opened 2025-08-18
fix(tools): Handle special characters in file paths for glob and read_many_files

### LINKED ISSUE #5086: You are referencing files with [id] in the path. The tool is unable to read files with brackets in the name.

galdawave:
## TLDR

This PR fixes a bug in the `glob` and `read-many-files` tools that prevented them from working with file paths containing special characters like `[]` and `()`. The fix ensures these paths are correctly escaped when they refer to existing files, while still allowing glob patterns to be used.

## Dive Deeper

The issue stemmed from the `glob` library interpreting special characters in file paths as part of a glob pattern. To resolve this, we now check if a given path exists on the filesystem before passing it to `glob`.

- If `fs.stat()` confirms the path exists, we treat it as a literal path and escape any special characters using `glob.escape()`.
- If the path does not exist, we assume it's a glob pattern and use it as-is.

This logic has been implemented in both the `GlobTool` and the `ReadManyFilesTool`. Additionally, new unit tests have been added to verify that both tools can now correctly handle file paths and names with special characters, ensuring this bug is resolved and preventing future regressions.

## Reviewer Test Plan

To validate the fix, a reviewer can perform the following steps:

1.  Create a file within a directory structure that includes special characters:
    ```bash
    mkdir -p 'my-dir(v1)/[test]'
    echo "hello world from file1.txt" > 'my-dir(v1)/[test]/file1.txt'
    echo "hello world from file2.txt" > 'my-dir(v1)/[test]/file2.txt'
    ```
2.  Use the `read-many-files` tool to read the content of the newly created file. The tool should successfully read and display the file's content.
    ```bash
    what is in @my-dir(v1)/[test]/file.txt 
    ```

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs
- Fixes: #5086

### DIFF
--- a/packages/core/src/tools/glob.ts
+++ b/packages/core/src/tools/glob.ts
@@ -6,7 +6,7 @@
 
 import fs from 'fs';
 import path from 'path';
-import { glob } from 'glob';
+import { glob, escape } from 'glob';
 import { SchemaValidator } from '../utils/schemaValidator.js';
 import {
   BaseDeclarativeTool,
@@ -137,7 +137,13 @@ class GlobToolInvocation extends BaseToolInvocation<
       let allEntries: GlobPath[] = [];
 
       for (const searchDir of searchDirectories) {
-        const entries = (await glob(this.params.pattern, {
+        let pattern = this.params.pattern;
+        const fullPath = path.join(searchDir, pattern);
+        if (fs.existsSync(fullPath)) {
+          pattern = escape(pattern);
+        }
+
+        const entries = (await glob(pattern, {
           cwd: searchDir,
           withFileTypes: true,
           nodir: true,
--- a/packages/core/src/tools/read-many-files.test.ts
+++ b/packages/core/src/tools/read-many-files.test.ts
@@ -524,6 +524,43 @@ describe('ReadManyFilesTool', () => {
       expect(truncatedFileContent).toContain('L200');
       expect(truncatedFileContent).not.toContain('L2400');
     });
+
+    it('should read files with special characters like [] and () in the path', async () => {
+      const filePath = 'src/app/[test]/(dashboard)/testing/components/code.tsx';
+      createFile(filePath, 'Content of receive-detail');
+      const params = { paths: [filePath] };
+      const invocation = tool.build(params);
+      const result = await invocation.execute(new AbortController().signal);
+      const expectedPath = path.join(tempRootDir, filePath);
+      expect(result.llmContent).toEqual([
+        `--- ${expectedPath} ---
+
+Content of receive-detail
+
+`,
+      ]);
+      expect(result.returnDisplay).toContain(
+        'Successfully read and concatenated content from **1 file(s)**',
+      );
+    });
+
+    it('should read files with special characters in the name', async () => {
+      createFile('file[1].txt', 'Content of file[1]');
+      const params = { paths: ['file[1].txt'] };
+      const invocation = tool.build(params);
+      const result = await invocation.execute(new AbortController().signal);
+      const expectedPath = path.join(tempRootDir, 'file[1].txt');
+      expect(result.llmContent).toEqual([
+        `--- ${expectedPath} ---
+
+Content of file[1]
+
+`,
+      ]);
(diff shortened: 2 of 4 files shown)


packages/core/src/tools/glob.ts
```diff
@@ -137,7 +138,20 @@ class GlobToolInvocation extends BaseToolInvocation<
       let allEntries: GlobPath[] = [];
 
       for (const searchDir of searchDirectories) {
-        const entries = (await glob(this.params.pattern, {
+        let pattern = this.params.pattern;
+        try {
+          await fs.promises.stat(path.join(searchDir, pattern));
+          pattern = escape(pattern);
```
jacob314, line 143: if you apply the suggestions in read-many-files please apply them here

packages/core/src/tools/read-many-files.test.ts
```diff
@@ -524,6 +524,43 @@ describe('ReadManyFilesTool', () => {
       expect(truncatedFileContent).toContain('L200');
       expect(truncatedFileContent).not.toContain('L2400');
     });
+
+    it('should read files with special characters like [] and () in the path', async () => {
+      const filePath = 'src/app/[test]/(dashboard)/testing/components/code.tsx';
```
jacob314, line 532: great regression test!

packages/core/src/tools/read-many-files.ts
```diff
@@ -245,18 +247,32 @@ ${finalExclusionPatternsForDescription
...
-            absolute: true,
-            nocase: true,
-            signal,
-          },
-        );
+        const processedPatterns = [];
+        for (const p of searchPatterns) {
+          const normalizedP = p.replace(/\\/g, '/');
+          try {
+            await fs.stat(path.join(dir, normalizedP));
```
jacob314: same nit about using stat instead of access and tossing the result.

packages/core/src/tools/read-many-files.ts
```diff
@@ -245,18 +247,32 @@ ${finalExclusionPatternsForDescription
...
+            processedPatterns.push(escape(normalizedP));
+          } catch (e: unknown) {
+            if (isNodeError(e) && e.code === 'ENOENT') {
+              // The path does not exist, so we treat it as a glob pattern.
+              processedPatterns.push(normalizedP);
+            } else {
+              // For other errors (e.g., permissions), we should not silently continue.
+              throw e;
```
jacob314: why is this useful? shouldn't we just let glob try to do its thing in this case? Is so we can use
fs.existsSync which would be simpler and a bit more idiomatic.

jacob314 (commented):
approved once these pretty minor comments are addressed.

packages/core/src/tools/read-many-files.test.ts
galdawave: Thanks!

jacob314 (approved):
lgtm

PR #6514, opened 2025-08-18
Merge .geminiignore and .gitignore to allow negative excludes from .geminiignore to override those in .gitignore

### LINKED ISSUE #5259: Allow .geminiignore to un-ignore items from .gitignore

I would love it if I could use `.geminiignore` to override the rules in `.gitignore` with negative patterns.

As far as I can tell, the current implementation will filter out a file if it appears in either's rules, but there's no support for something like:

```
# .gitignore
/foo
/bar
/baz
```

```
# .geminiignore

!/bar/quux
```

(I'd love to be corrected, I may have been holding something wrong)

### Why is this needed?

Some projects like [Fuchsia](https://fuchsia.dev) build a source checkout out of multiple git repositories using tools that wrap git commands. In those projects, a particular subdirectory of the workspace might be in `.gitignore` but logically still under source control and relevant for a developer (or agent) to read and edit directly.

It would be great to get gcli to read `GEMINI.md` files in those nested subrepositories when working in the root directory of the workspace so that memory content can be colocated and evolved in lockstep with the source it references.

This feature would also allow for a pattern where a user could have user-and-project-specific memory files under a `.gitignore`d path, analogous to common patterns for [user-specific Bazel configuration](https://bazel.build/configure/best-practices#bazelrc-file).

### Additional context

[ripgrep](https://github.com/BurntSushi/ripgrep/blob/master/GUIDE.md#automatic-filtering) is one notable precedent for a popular CLI tool that offers a similar hierarchical override approach to ignore files.

I tried implementing this, but it seems that there are some subtle interactions between the file discovery service and some of Gemini's tools and I wasn't sure how to proceed. I'd be interested in giving it another shot if there's interest from maintainers.

gbbosak:
## TLDR

.geminiignore should be able to override entries in .gitignore.

## Dive Deeper

This adds the ability for .geminiignore to override .gitignore by merging the two exclude lists.

## Reviewer Test Plan

Run unit tests.

## Testing Matrix

Ran unit tests, added additional unit tests.

## Linked issues / bugs

- Fixes: #5259


packages/core/src/services/fileDiscoveryService.ts
```diff
@@ -50,18 +73,21 @@ export class FileDiscoveryService {
...
-        options.respectGeminiIgnore &&
-        this.shouldGeminiIgnoreFile(filePath)
-      ) {
-        return false;
-      }
-      return true;
-    });
+    if (!this.unifiedIgnore) return filePaths;
+    if (options.respectGitIgnore && options.respectGeminiIgnore) {
+      return filePaths.filter((filePath) => !this.isUnifiedIgnored(filePath));
```
scidomino: Could you replace this with

return filePaths.filter((filePath) => !shouldIgnoreFile(filePath, options));

?

gbbosak: Done.

gbbosak:
Why does it have the need-issue label? I already have this associated with an issue.

gbbosak:
This has now completely bitrotted unfortunately. Looking up replacements to see how to redo this.

gbbosak:
This is no longer applicable to the current codebase. Closing.

PR #6515, opened 2025-08-18
fix(shell): handle single-quoted backslashes correctly when splitting commands

shicks:
## TLDR

There's a bug in the splitCommands logic due to a common misconception in shell grammar.  Standard POSIX sh specifies that single-quoted backslashes are treated as literal backslashes: the _only_ character that can exit a single-quoted string is another single-quote.  This change corrects the logic to account for this.  It also adds test coverage for splitCommands more generally.

## Reviewer Test Plan

To see the vulnerability, start by prompting

```
Run `ls`
```

Approve the `ls` command with "Yes, allow always".  Then prompt

```
Run `echo PWND`
```

and deny access to verify that `echo` is not allowed to run.  Finally, prompt

```
Now run exactly `ls '\'; echo PWND`.  Please pay close attention to keep the quotes and backslashes exactly as I entered them.
```

Without this patch, the agent will execute `echo PWND` without confirmation, but this patch fixes the vulnerability.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


shicks:
Note: the settings precedence change was an extra commit that was not intended to be part of this PR (I've instead opened #6517 to discuss it further).  I have removed it, so that only the shell-utils changes should be included here.

shicks:
This issue is still present - is there any way to get it some attention?  In the worst case, a compromised or severely misguided agent could cause some really unfortunate impacts if it learns how to exploit this.

chrstnb:

shicks:
This issue seems to have been corrected in the meantime.

PR #6516, opened 2025-08-18
chore: Implement secure CI/CD pipeline

### LINKED ISSUE #3695: Configure CI to reliably run integration tests on PRs from forks

galdawave:
## TLDR

This PR enhances CI/CD security by splitting the pipeline into two distinct workflows:

1.  **Untrusted (`ci.yml`):** Runs on all PRs. It lints, runs unit tests, and builds the code. It has no access to secrets.
2.  **Trusted (`trusted-ci.yml`):** Runs *only after* the untrusted workflow succeeds. It downloads the artifacts built in the first stage and runs integration tests using the necessary secrets.

This separation prevents untrusted code from pull requests from having access to sensitive secrets, mitigating the risk of secret exfiltration.

## Dive Deeper

The core of this change is the introduction of a two-stage CI process using `workflow_run` as a trigger.

-   The existing `ci.yml` is now the "untrusted" stage. It handles all checks that don't require secrets, such as linting, formatting, type checking, and unit tests. Its final step is to build the project and upload the compiled `dist` directories as a single artifact.
-   The new `trusted-ci.yml` workflow acts as the "trusted" stage. It is triggered only upon the successful completion of a `ci.yml` run associated with a pull request. Instead of checking out the potentially malicious PR code, it checks out the *base branch* and then downloads the build artifacts from the untrusted stage. These pre-built artifacts are then used to run the integration tests, which require access to `GEMINI_API_KEY`.

This approach ensures that the environment with access to secrets never directly executes code from the pull request, only the artifacts generated in the sandboxed, secret-less environment.

To enforce that integration tests are never skipped, a step was added to `trusted-ci.yml` that explicitly fails the workflow if the `GEMINI_API_KEY` secret is not available. This guarantees that all PRs are fully tested.

Finally, the `test` job in the trusted workflow is associated with a GitHub Environment named `trusted-ci-env`. This allows for additional protections, such as requiring manual approval before the job can run, providing an extra layer of security for sensitive operations.

## Reviewer Test Plan

1.  Review the workflow files `.github/workflows/ci.yml` and `.github/workflows/trusted-ci.yml`.
2.  Confirm that `ci.yml` does not contain any steps that access secrets.
3.  Confirm that `trusted-ci.yml` is triggered by `workflow_run` and not directly by `pull_request`.
4.  Verify that `trusted-ci.yml` checks out the base branch and downloads artifacts rather than checking out the PR head.
5.  Create a test pull request and observe the GitHub Actions tab:
    -   Ensure the "Gemini CLI CI" workflow runs first.
    -   After it succeeds, ensure the "Gemini CLI Trusted CI" workflow is triggered and runs the integration tests.

## Testing Matrix

The changes are to the CI/CD pipeline. The existing test suite should pass as before. The primary validation is observing that the GitHub Actions workflows execute correctly.

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

- Resolves #3695

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -120,6 +120,12 @@ jobs:
         run: |-
           npm run build
 
+      - name: 'Upload Build Artifacts'
+        uses: 'actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02' # ratchet:actions/upload-artifact@v4
+        with:
+          name: 'build-artifacts'
+          path: 'packages/*/dist'
+
       - name: 'Run type check'
         run: |-
           npm run typecheck
@@ -280,10 +286,10 @@ jobs:
         run: |-
           npm ci
 
-      - name: 'Run tests and generate reports'
+      - name: 'Run unit tests'
         env:
           NO_COLOR: true
-        run: 'npm run test:ci'
+        run: 'npm run test:unit'
 
       - name: 'Publish Test Report (for non-forks)'
         if: |-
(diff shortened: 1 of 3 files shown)


cornmander:
Hey, thanks for implementing this. I was actually thinking of doing this in a different way. I was thinking of using an RPC replay library like pollyjs to record RPCs locally and then replay them in CI/CD. This would have a few benefits:

- forks can run integration tests without making their own secrets
- tests would be much faster
- tests would be less flaky

What do you think?

PR #6523, opened 2025-08-19
Update colors tokens for inputer/footer

miguelsolorio:
## TLDR

In the last color token update, the input/footer colors were not propagating properly

## Dive Deeper

Even though the color token names were updated, they kept being overwritten by the default. I think part of the problem was that base theme was being imported into each theme.

## Before
CleanShot 2025-08-18 at 17 32 41@2x

## After
CleanShot 2025-08-18 at 17 31 28@2x

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


miguelsolorio:
Going to wait on this one and see if @ali-aljufairi's updates fix this (which I suspect it does)

miguelsolorio:
@jacob314 this PR is ready, I think the semantic PR refactor will need to broken down into smaller PRs so going with this instead for now

packages/cli/src/ui/themes/theme-manager.ts
```diff
@@ -299,7 +299,7 @@ class ThemeManager {
 
     // First check built-in themes
     const builtInTheme = this.availableThemes.find(
-      (theme) => theme.name === themeName,
```
jacob314: is this intentional?

jacob314 (approved):
lgtm

packages/cli/src/ui/themes/theme-manager.ts
miguelsolorio: weird, I'm not sure why that got added in. reverted.

Issue #6257, opened 2025-08-14
Ability to enforce/set selected auth options


ameer00:
### What would you like to be added?

In some situations, we would like the ability to set specific AUTH options. For example, for Enterprises, it may be necesssary to log in with their corporate email addresses. This way they are under the Google Cloud ToS. For "Login with Google" option, it would be necessary to provide domains (for example, you may NOT login with your personal @gmail addresses and only your corporate email address.
This would be set by admins so users cannot override.

### Why is this needed?

Enterprises require enforcement of specific AUTH types so they know that they can use this tool within specified terms of service.

jacob314:
This is likely actually possible today as system settings override user settings and project settings. So if you force the auth type in your system level settings.json then you will enforce it for all users. Please let me know if that works.

ameer00:
Thanks @jacob314 for the response. 
This works at small scale. For larger organizations with thousands of users, managing at the desktop level is cumbersome. 
Also, in addition to limiting authType (for example Google oauth), also need to ensure that the users log in using their corp credentials and not their personal gmail creds because of different subscription, quota and most importantly terms of service (which are governed by how you authenticate)

jacob314:
interesting. fwiw, the system level settings file (this is different from the one in the user directory) is working for Google internally with over 100,000 employees. 

Here is where systems settings are loaded from. Interested if that can work for your organization. Would require you to be able to deploy the file to some fixed location on each users machine such as 'C:\\ProgramData\\gemini-cli\\settings.json'

```
export function getSystemSettingsPath(): string {
  if (process.env['GEMINI_CLI_SYSTEM_SETTINGS_PATH']) {
    return process.env['GEMINI_CLI_SYSTEM_SETTINGS_PATH'];
  }
  if (platform() === 'darwin') {
    return '/Library/Application Support/GeminiCli/settings.json';
  } else if (platform() === 'win32') {
    return 'C:\\ProgramData\\gemini-cli\\settings.json';
  } else {
    return '/etc/gemini-cli/settings.json';
  }
}
```

PR #6564, opened 2025-08-19
Add enforcedAuthType setting 

### LINKED ISSUE #6664: Add auth type setting for enterprises
We've gotten requests for enterprise users to be able to configure in system-level settings the authentication method.

chrstnb:
## TLDR

This PR introduces a new setting, `enforcedAuthType`, which allows system administrators to enforce a specific authentication method for the Gemini CLI.

### Features

- **Enforced Authentication Type:** A new setting, `enforcedAuthType`, has been added to the settings schema. When this setting is present, the CLI will ensure that the user is authenticated with the specified authentication method.
- **Non-interactive Mode:** In non-interactive mode, if the current authentication type does not match the `enforcedAuthType`, the CLI will exit with an error message.
- **Interactive Mode:** In interactive mode, if the current authentication type does not match the `enforcedAuthType`, the user will be prompted to switch to the enforced authentication method.

https://github.com/user-attachments/assets/26026caa-0795-4c86-bd78-d983728ad065

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes #6664


9M6:
was looking for this feature. 👍

ameer00:
Fixes #6257

cornmander:
Can you split this into a UI change and a config change? I can review the config change but @jacob314 should review the UI parts.

NTaylorMullen:
@chrstnb if you have the auth type specified in the system level settings.json I believe that might also enforce this by default. Would you mind double checking?

ptone:
> @chrstnb if you have the auth type specified in the system level settings.json I believe that might also enforce this by default. Would you mind double checking?

The existing selectedAuthType setting only sets a default preference, but does not limit a user from choosing a different auth type during /auth

It is not used as a limiter during the /auth flow (choosing a new auth type sets a new value for selectedAuthType at user-settings level) - this behavior seems to indicate that it would be difficult to repurpose this setting for enforcement purposes.

chrstnb:
@NTaylorMullen can you say more re: https://github.com/google-gemini/gemini-cli/pull/6564#issuecomment-3214743816?

NTaylorMullen (approved):

NTaylorMullen (approved):

PR #6566, opened 2025-08-19
fix(ide): Correctly identify IDE process when run from terminal

skeshive:
Previously, the function would traverse the process tree until it reached the top-level parent process, which led to an incorrect PID being returned.

The updated logic now includes a specific check for known shell processes (bash, zsh, cmd.exe, etc.). When a shell is encountered, the function returns its parent's PID. This correctly identifies the process that launched the shell, which is the main IDE application.

A specific check for Linux utility processes (--type=) has also been added. This prevents the function from incorrectly returning the PID of a utility process, instead of the main application window that spawned it.

## Testing Matrix
I tested this on my mac & linux / windows cloudtops


packages/core/src/ide/process-utils.ts
```diff
@@ -7,56 +7,109 @@
 import { exec } from 'child_process';
 import { promisify } from 'util';
 import os from 'os';
+import path from 'path';
+
+const logger = {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
jacob314: I would remove this logger const as it doesn't seem like it adds anything over the default console.error
If you do keep it use unknown rather than any to avoid the lint

packages/core/src/ide/process-utils.ts
```diff
@@ -7,56 +7,109 @@
...
+        parentPid = ppidMatch ? parseInt(ppidMatch[1], 10) : 0;
+      } else {
+        const command = `ps -o ppid=,command= -p ${currentPid}`;
         const { stdout } = await execAsync(command);
-        const ppid = parseInt(stdout.trim(), 10);
-        parentPid = isNaN(ppid) ? 1 : ppid; // Top of the tree is 1
+        const trimmedStdout = stdout.trim();
+        const ppidString = trimmedStdout.split(/\s+/)[0];
+        const ppid = parseInt(ppidString, 10);
+        parentPid = isNaN(ppid) ? 1 : ppid;
```
jacob314: nit: keep comment about top of the tree is 1. that was useful to explain why 1 was used.

packages/core/src/ide/process-utils.ts
```diff
@@ -7,56 +7,109 @@
...
 
+    const isShell = shellAllowlist.some((shell) =>
+      platform === 'win32'
+        ? processName.toLowerCase() === shell.toLowerCase()
+        : processName === shell,
+    );
+
+    if (isShell) {
+      let idePid = parentPid;
+      if (os.platform() === 'linux') {
```
jacob314: can we use != 'win32' here and use this logic on mac as well?

packages/core/src/ide/process-utils.ts
```diff
@@ -7,56 +7,109 @@
...
     const rootPid = platform === 'win32' ? 0 : 1;
-
     // If the parent is the root process or invalid, we've found our target.
     if (parentPid === rootPid || parentPid <= 0) {
       break;
     }
     // Move one level up the tree for the next iteration.
     currentPid = parentPid;
   }
+  logger.error(
```
jacob314: is there a. reason we don't just call 
console.error directly?
Generally that aligns with the rest of the codebase.

jacob314 (commented):
It would be great to add integration tests for this so we don't accidentally regress this behavior.

skeshive:
> It would be great to add integration tests for this so we don't accidentally regress this behavior.

ack- IDE integration tests are currently broken but I'll add a test once I fix the tests

packages/core/src/ide/process-utils.ts
```diff
@@ -7,56 +7,156 @@
...
+ */
+export async function getIdeProcessId(): Promise<number> {
+  const platform = os.platform();
+
+  if (platform === 'win32') {
+    return getIdeProcessIdForWindows();
+  }
+
+  const shells: Record<string, string[]> = {
+    darwin: ['zsh', 'bash', 'sh', 'tcsh', 'csh', 'ksh', 'fish'],
```
jacob314: you can run dash shell on macos. Can you simplify and have one list of shells?

jacob314 (approved):
lgtm

PR #6567, opened 2025-08-19
feat: Add IDE client to /bug & /about if IDE mode is enabled

skeshive:


packages/cli/src/ui/commands/bugCommand.ts
```diff
@@ -45,7 +49,7 @@ export const bugCommand: SlashCommand = {
 * **Operating System:** ${osVersion}
 * **Sandbox Environment:** ${sandboxEnv}
 * **Model Version:** ${modelVersion}
-* **Memory Usage:** ${memoryUsage}
+* **Memory Usage:** ${memoryUsage}${ideClient ? `\n* **IDE Client:** ${ideClient}` : ''}
```
jacob314: minor nit: linebreaking for this is a bit ugly. Can you instead move this out of the outer
\` block so that the 
`${ideClient ?` block can be on its own line rather than reading like it is part of the memory usage line.

jacob314 (approved):
Approved but please resolve the minior nit before merging or resolve it with a fast follow tiny cleanup pull request.

Issue #4651, opened 2025-07-22
Investigate if we need the nextSpeakerCheck functionality


anj-s:
We currently have a nextSpeakerCheck function to ensure that we have the right user/role when we receive empty  responses from the model. There is a second check to prevent looping.
Remove this nextSpeakerCheck to confirm that the original issue has been fixed. This requires a LLM call which we should ideally get rid of.

acoliver:
suggest this should be a higher priority. The number one complaint besides the throttle (which this is also part of) has been the suprise token burn and suprise bills. Sending the conversation over and over again is pretty bad. It's also not a small part of performance. 

from original bug report:

packages/core/src/utils/nextSpeakerChecker.ts -
Presently, the system sends the whole conversation to Flash potentially multiple times per turn to ask whose turn it is. I assume this thing began as an internal tool. As a public tool, this is too terribly inefficient, and a better solution is needed. Also, this is a terrible load on Google's infrastructure and rather cruel to the users to throttle or blow tokens for this...

This is a terrible design! Let me break down why this is so inefficient:

The Problems:

Extra API Call: After EVERY Gemini response, it makes another API call to Gemini Flash just to ask "should I continue?"
Full History Processing: It sends the ENTIRE conversation history to Gemini Flash for this simple decision
Recursive Calls: If it decides to continue, it recursively calls sendMessageStream again, processing everything again
Token Waste: Burns tokens on Gemini Flash for what should be a simple check
Latency: Adds significant delay between each response segment
What It's Doing:

User: "Explain quantum computing"
→ Gemini: "Quantum computing uses qubits..."
→ API CALL to Flash: "Should I continue?" (with full history)
→ Flash: "Yes, continue"
→ Gemini: "Please continue."
→ Gemini: "Furthermore, superposition allows..."
→ API CALL to Flash: "Should I continue?" (with even more history)
→ Flash: "Yes, continue"
→ ... and so on

What did you expect to happen?
Better Approaches:

Let Gemini Decide Itself: Gemini could just output a signal in its response if it wants to continue
Use Streaming: Instead of multiple round trips, use proper streaming where Gemini outputs everything in one go
Client-Side Logic: Simple rules could determine continuation without an API call
Single Request: Send a parameter like max_turns: 3 and let Gemini handle it server-side

char23-web:
[...]

PR #6589, opened 2025-08-19
feat(core): Add skipNextSpeakerCheck setting

SandyTao520:
## TLDR

This pull request introduces a new configuration option, `skipNextSpeakerCheck`, which allows users to bypass the `checkNextSpeaker` function call. This can be useful for developers or power users who want to ensure their input is processed without an intermediate check to see if the model should speak next, potentially speeding up interactions.

## Dive Deeper

The `checkNextSpeaker` function is designed to create a more natural conversational flow by predicting whether the user or the model should provide the next turn. However, in scenarios where a user wants to provide a series of inputs without interruption, or for debugging purposes, this check can be an unnecessary extra step.

This change adds a `skipNextSpeakerCheck` boolean flag to the settings schema, accessible through the `gemini settings` dialog. When enabled, the `GeminiClient` will skip the `checkNextSpeaker` call entirely and proceed directly with processing the user's turn. This provides more direct control over the conversation flow for users who need it.

The changes span:
- `settingsSchema.ts` to define the new setting.
- `config.ts` in both `cli` and `core` to plumb the setting through.
- `client.ts` to implement the conditional skip logic.

## Reviewer Test Plan

1.  Run `gemini settings` and enable the "Skip Next Speaker Check" option in the "General" category.
2.  Start a chat session with `gemini`.
3.  Enter a multi-turn prompt where the model might typically interject.
4.  Observe that the client immediately processes your input without the delay or logic associated with determining the next speaker.
5.  Disable the setting and repeat the process to confirm that the default behavior is restored.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ✅  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

#4651

### DIFF
--- a/packages/cli/src/config/config.ts
+++ b/packages/cli/src/config/config.ts
@@ -73,6 +73,7 @@ export interface CliArgs {
   listExtensions: boolean | undefined;
   proxy: string | undefined;
   includeDirectories: string[] | undefined;
+  skipNextSpeakerCheck: boolean | undefined;
 }
 
 export async function parseArguments(): Promise<CliArgs> {
@@ -228,6 +229,11 @@ export async function parseArguments(): Promise<CliArgs> {
           coerce: (dirs: string[]) =>
             // Handle comma-separated values
             dirs.flatMap((dir) => dir.split(',').map((d) => d.trim())),
+        })
+        .option('skip-next-speaker-check', {
+          type: 'boolean',
+          description: 'Skip the next speaker check.',
+          default: false,
         })
         .check((argv) => {
           if (argv.prompt && argv['promptInteractive']) {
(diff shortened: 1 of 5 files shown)


anj-s:
can you make this an env variable change instead of a command line flag?

jacob314:
Better yet can we add this to settings.ts?
Generally I'd prefer that to args or environment variables for things that aren't frequently used.

anj-s:
> Better yet can we add this to settings.ts? Generally I'd prefer that to args or environment variables for things that aren't frequently used.

I prefer not adding to settings.ts cause this is an implementation detail that we don't want exposed to users. I generally think of settings.ts for configuring the behavior of gemini cli.

jacob314 (approved):
lgtm

PR #6591, opened 2025-08-19
feat: update .gitignore in /setup-github

jerop:
This feature automatically updates the users .gitignore file during the `setup-github` command.

The following changes are included:
- A new `updateGitignore` function is added to `packages/cli/src/ui/commands/setupGithubCommand.ts`. This function ensures that the `.gemini/` directory and `gha-creds-*.json` files are added to the users `.gitignore` file.
- The `setupGithubCommand` now calls `updateGitignore` to perform this update.

This is a follow up to https://github.com/google-github-actions/run-gemini-cli/pull/198

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,3 +40,6 @@ packages/cli/src/generated/
 packages/core/src/generated/
 .integration-tests/
 packages/vscode-ide-companion/*.vsix
+
+# GHA credentials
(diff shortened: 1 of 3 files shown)


mattKorwel (approved):
Nice, i like the added test suite, good cases there!

PR #6595, opened 2025-08-20
Add permissions specs to token generation.

cornmander:
Following GitHub best practices: https://github.com/actions/create-github-app-token?tab=readme-ov-file#permission-permission-name

### DIFF
--- a/.github/workflows/gemini-scheduled-issue-triage.yml
+++ b/.github/workflows/gemini-scheduled-issue-triage.yml
@@ -14,11 +14,8 @@ defaults:
     shell: 'bash'
 
 permissions:
-  contents: 'read'
   id-token: 'write'
   issues: 'write'
-  statuses: 'write'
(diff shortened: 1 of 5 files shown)


jacob314 (approved):
lgtm

.github/workflows/gemini-scheduled-issue-triage.yml
```diff
@@ -14,11 +14,8 @@ defaults:
     shell: 'bash'
 
 permissions:
-  contents: 'read'
   id-token: 'write'
   issues: 'write'
-  statuses: 'write'
-  packages: 'read'
```
gsquared94, line 21: Need the package read permission to pull mcp server container image required by the issue deduplication workflows

sethvargo:
I still think it's important that we prioritize moving to an API endpoint to mint these tokens instead of relying on the action itself to do the token minting.

PR #6604, opened 2025-08-20
feat: add dependabot configuration

cornmander:
Adds dependabot configuration so that we get PRs to update packages with new vulnerabilities.


mattKorwel (approved):
Thanks for adding this. Lets see how noisy it is and how we can best validate all the changes, we have. Great place to start.

PR #6606, opened 2025-08-20
Ignore workspace settings for untrusted folders

shrutip90:
## TLDR

For untrusted folders, we want to avoid merging workspace settings. Added logic to reload settings when the trust level changes and also reload config object and theme. Also modified the components that use settings to get it from settingsContext.

## Dive Deeper

Verified manually by setting different themes at the workspace level vs user level and checking that it changes on trust level change during launch.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

### DIFF
--- a/packages/cli/src/config/settings.ts
+++ b/packages/cli/src/config/settings.ts
@@ -16,6 +16,7 @@ import {
 import stripJsonComments from 'strip-json-comments';
 import { DefaultLight } from '../ui/themes/default-light.js';
 import { DefaultDark } from '../ui/themes/default.js';
+import { isWorkspaceTrusted } from './trustedFolders.js';
 import { Settings, MemoryImportFormat } from './settingsSchema.js';
 
 export type { Settings, MemoryImportFormat };
@@ -73,7 +74,30 @@ function mergeSettings(
   system: Settings,
   user: Settings,
   workspace: Settings,
+  isTrusted?: boolean,
 ): Settings {
+  if (!isTrusted) {
+    return {
+      ...user,
+      ...system,
+      customThemes: {
+        ...(user.customThemes || {}),
+        ...(system.customThemes || {}),
+      },
+      mcpServers: {
+        ...(user.mcpServers || {}),
+        ...(system.mcpServers || {}),
+      },
+      includeDirectories: [
+        ...(system.includeDirectories || []),
+        ...(user.includeDirectories || []),
+      ],
+      chatCompression: {
+        ...(system.chatCompression || {}),
+        ...(user.chatCompression || {}),
+      },
+    };
+  }
   // folderTrust is not supported at workspace level.
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const { folderTrust, ...workspaceWithoutFolderTrust } = workspace;
@@ -111,11 +135,13 @@ export class LoadedSettings {
     user: SettingsFile,
     workspace: SettingsFile,
     errors: SettingsError[],
+    isTrusted?: boolean,
   ) {
     this.system = system;
     this.user = user;
     this.workspace = workspace;
     this.errors = errors;
+    this.isTrusted = isTrusted;
     this._merged = this.computeMergedSettings();
   }
 
@@ -123,6 +149,7 @@ export class LoadedSettings {
   readonly user: SettingsFile;
   readonly workspace: SettingsFile;
   readonly errors: SettingsError[];
+  private isTrusted: boolean | undefined;
 
   private _merged: Settings;
 
@@ -130,11 +157,17 @@ export class LoadedSettings {
     return this._merged;
   }
 
+  recomputeMergedSettings(isTrusted?: boolean): void {
+    this.isTrusted = isTrusted;
+    this._merged = this.computeMergedSettings();
+  }
+
   private computeMergedSettings(): Settings {
     return mergeSettings(
       this.system.settings,
       this.user.settings,
       this.workspace.settings,
+      this.isTrusted,
     );
   }
 
@@ -403,11 +436,16 @@ export function loadSettings(workspaceDir: string): LoadedSettings {
     }
   }
 
+  // For the initial trust check, we can only use user and system settings.
+  const initialTrustCheckSettings = { ...systemSettings, ...userSettings };
+  const isTrusted = isWorkspaceTrusted(initialTrustCheckSettings);
+
   // Create a temporary merged settings object to pass to loadEnvironment.
   const tempMergedSettings = mergeSettings(
     systemSettings,
     userSettings,
     workspaceSettings,
+    isTrusted,
   );
 
   // loadEnviroment depends on settings so we have to create a temp version of
@@ -434,6 +472,7 @@ export function loadSettings(workspaceDir: string): LoadedSettings {
       settings: workspaceSettings,
     },
     settingsErrors,
+    isTrusted,
   );
 
   // Validate chatCompression settings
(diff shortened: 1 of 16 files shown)


packages/cli/src/config/settings.ts
```diff
@@ -73,7 +74,30 @@ function mergeSettings(
...
+        ...(system.mcpServers || {}),
+      },
+      includeDirectories: [
+        ...(system.includeDirectories || []),
+        ...(user.includeDirectories || []),
+      ],
+      chatCompression: {
+        ...(system.chatCompression || {}),
+        ...(user.chatCompression || {}),
+      },
```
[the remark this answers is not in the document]
shrutip90: This is unchanged compared to the existing merge logic.

cornmander (approved):

packages/cli/src/config/settings.ts
```diff
@@ -73,7 +74,30 @@ function mergeSettings(
   system: Settings,
   user: Settings,
   workspace: Settings,
+  isTrusted?: boolean,
 ): Settings {
+  if (!isTrusted) {
+    return {
```
jacob314, line 80: this code duplicates the code in the non trusted case. Please refactor this code so less duplication is required. particularly with the logic for servers includeDirs, etc this needs to not be duped.
I would suggest adding a helper that takes 
N Settings objects and merges them in the specified order.
You can then apply it for both cases. Alternately you could reuse the existing method but pass in an empty workspace setting instead of the actual one. 
Please land this as a fast follow to avoid duplicated code that might get out of sync.

packages/cli/src/ui/hooks/useThemeCommand.ts
```diff
@@ -21,11 +22,12 @@ interface UseThemeCommandReturn {
 }
 
 export const useThemeCommand = (
-  loadedSettings: LoadedSettings,
   setThemeError: (error: string | null) => void,
   addItem: (item: Omit<HistoryItem, 'id'>, timestamp: number) => void,
 ): UseThemeCommandReturn => {
   const [isThemeDialogOpen, setIsThemeDialogOpen] = useState(false);
+  const settingsContext = useContext(SettingsContext);
+  const loadedSettings = settingsContext!.settings;
```
jacob314, line 30: why not make useContext(SettingsContext) non-nullable to avoid this? Seems like code that depends on settings should require settings.

packages/cli/src/ui/utils/MarkdownDisplay.test.tsx
```diff
@@ -40,7 +42,9 @@ describe('<MarkdownDisplay />', () => {
   it('renders a simple paragraph', () => {
     const text = 'Hello, world.';
     const { lastFrame } = render(
```
jacob314, line 44: there is an existing render helper that provides contexts. Extend it to provide the settings context as well. That will clean up this test and elsewhere.

packages/cli/src/ui/App.tsx
```diff
@@ -108,34 +116,46 @@ const MAX_DISPLAYED_QUEUED_MESSAGES = 3;
...
   config: Config;
-  settings: LoadedSettings;
   startupWarnings?: string[];
   version: string;
 }
 
 export const AppWrapper = (props: AppProps) => {
   const kittyProtocolStatus = useKittyKeyboardProtocol();
+  const settingsContext = useContext(SettingsContext);
+  if (!settingsContext) {
```
jacob314, line 126: this throw should be in useContext(SettingsContext) itself.

packages/cli/src/ui/App.tsx
```diff
@@ -108,34 +116,46 @@ const MAX_DISPLAYED_QUEUED_MESSAGES = 3;
...
+  const { settings } = settingsContext;
+
   return (
     <KeypressProvider
       kittyProtocolEnabled={kittyProtocolStatus.enabled}
       config={props.config}
     >
       <SessionStatsProvider>
-        <VimModeProvider settings={props.settings}>
+        <VimModeProvider settings={settings}>
```
jacob314, line 138: Please audit all the existing code and ensure we aren't prop drilling settings anymore now that it has a provider.
This case should change to not take settings as a parameter.

packages/cli/src/ui/App.tsx
```diff
@@ -262,7 +282,7 @@ const App = ({ config, settings, startupWarnings = [], version }: AppProps) => {
     openThemeDialog,
     handleThemeSelect,
     handleThemeHighlight,
-  } = useThemeCommand(settings, setThemeError, addItem);
+  } = useThemeCommand(setThemeError, addItem);
```
jacob314, line 285: good to see this prop drilling go away

packages/cli/src/ui/MainComponent.tsx
```diff
@@ -0,0 +1,95 @@
...
+  initialConfig: Config;
+  settings: LoadedSettings;
+  startupWarnings: string[];
+  version: string;
+  workspaceRoot: string;
+  extensions: Extension[];
+  argv: CliArgs;
+}
+
+export const MainComponent = ({
```
jacob314, line 26: this needs a better name. How is MainComponent different from app?

packages/cli/src/ui/MainComponent.tsx
```diff
@@ -0,0 +1,95 @@
...
+  const [currentSettings, setCurrentSettings] =
+    useState<LoadedSettings>(settings);
+  const [config, setConfig] = useState<Config>(initialConfig);
+
+  const recomputeSettings = () => {
+    const newSettings = loadSettings(workspaceRoot);
+    setCurrentSettings(newSettings);
+  };
+
+  React.useEffect(() => {
```
jacob314, line 44: lets refine how this works. I'm concerned about this async logic in useEffect

packages/cli/src/ui/MainComponent.tsx
```diff
@@ -0,0 +1,95 @@
...
+  argv,
+}: MainComponentProps) => {
+  const [currentSettings, setCurrentSettings] =
+    useState<LoadedSettings>(settings);
+  const [config, setConfig] = useState<Config>(initialConfig);
+
+  const recomputeSettings = () => {
+    const newSettings = loadSettings(workspaceRoot);
+    setCurrentSettings(newSettings);
+  };
```
jacob314, line 42: aligning with the rest of the code in the repo, write this as useEffect not React.useEffect.

packages/cli/src/ui/hooks/useThemeCommand.ts
shrutip90: Replaced it with useSettings() that handles this.

packages/cli/src/config/settings.ts
shrutip90: Done.

packages/cli/src/ui/utils/MarkdownDisplay.test.tsx
shrutip90: Thanks. There's renderWithProviders that already provides SettingsContext. Replaced with it to simplify the tests here.

packages/cli/src/ui/App.tsx
shrutip90: Done, replaced this block with useSettings that already handles this.

Issue #6551, opened 2025-08-19
Implement fuzzy command matching and autocompletion for enhanced discoverability


dracic:
## Feature Description

This feature request proposes the implementation of a **fuzzy command matching and autocompletion system** within the Gemini CLI. The goal is to improve the user experience by making command discovery intuitive and reducing the cognitive load of remembering complex and custom command hierarchies.

Currently, as users create more custom commands and nested structures, it becomes increasingly difficult to recall the exact syntax for every command. This forces users to frequently consult help documentation (`/help`) or external notes, interrupting their workflow.

## Proposed Solution

The proposed functionality would be similar to the behavior seen in the Claude Code CLI. When a user starts typing a command, the CLI should dynamically display a list of all commands that contain the typed string.

### Example Workflow:

1.  A user has created several custom commands, such as:

      * `BMad:analyst`
      * `BMad:analyst:create-doc`

2.  The user wants to find a command related to "agent" but can't remember the exact name. They start typing:  `/analyst` if that syntax is adopted.

3.  The CLI should immediately display a filtered list of suggestions in real-time. Matching should be case insensitive.

4.  The user can then use arrow keys to select the desired command or continue typing to further refine the list.

## User Benefits

  * **Enhanced Discoverability:** Users can easily find the commands they need without knowing their full names or paths in advance. This is especially valuable for new users or for those returning to a project after some time. 
  * **Improved Efficiency:** This feature significantly speeds up workflow by reducing the time spent searching for commands. It keeps the user "in the flow" and focused on the task at hand.
  * **Reduced Learning Curve:** It makes the CLI more approachable and less intimidating. Users can learn the available commands organically through interaction rather than through rote memorization.
  * **Support for Complex Hierarchies:** As the CLI supports more complex, nested commands, this type of search becomes essential for maintaining usability.

Implementing this fuzzy matching and autocompletion would be a powerful enhancement to the Gemini CLI's usability, making it a more productive and user-friendly tool.

Also related to: #5039

dracic:
```
{"duplicate_issues": [4769]}
```

This is not duplicate of #4769

jacob314:
This is not an immediate priority for the team to implement ourselves but would welcome a pull request that supported this functionality. There is an existing fuzzy matcher used for fuzzy matching file name autocomplete that I would suggest using if you want to implement.

dracic:
@jacob314 I did my best to test this with all sort of usual and unusual commands.

[...]

PR #6633, opened 2025-08-20
feat(cli): add fuzzy matching for command suggestions

dracic:
## Implement fuzzy search functionality using AsyncFzf to improve command discovery. 

This pull request implements fuzzy search for command autocompletion, replacing the existing case-sensitive, prefix-based matching. By leveraging the `fzf` library, this change provides a more intuitive and powerful user experience, allowing for real-time, case-insensitive, and typo-tolerant command discovery. The modifications are primarily localized to the `useSlashCompletion.tsx` hook for minimal architectural impact.

## Dive Deeper

The current command completion is rigid, only matching exact, case-sensitive prefixes. This hinders usability, especially when users don't recall a command's precise name. This PR addresses the issue by integrating the `fzf` fuzzy search engine, which is already used for `@` file path completion.

**Architectural Changes:**
-   **`useSlashCompletion.tsx`:** This hook is refactored to be the core of the new implementation.
    -   It now instantiates and memoizes an `AsyncFzf` instance, seeded with all available slash commands.
    -   On user input, it calls `fzf.find(query)` to get a ranked list of fuzzy-matched commands.
    -   The simple prefix filter is completely replaced.
-   **`useCommandCompletion.tsx` & `InputPrompt.tsx`:** No changes are needed in these components. They remain agnostic to the suggestion logic and will render the fuzzy-matched results provided by the refactored `useSlashCompletion` hook.

This approach is a low-risk, high-impact brownfield enhancement. It delivers significant user value by targeting a single, well-encapsulated hook, avoiding a large-scale refactor.

Next possible fix would be for #5039

## Reviewer Test Plan

1.  Pull down the branch.
2.  Start the CLI using `npm run start`.
3.  Test the fuzzy command completion:
    -   Type `/` and begin typing a partial command. For example:
        -   Type `/he` and verify `/help` appears.
        -   Type `/subcommand` and verify `/customcommand:subcommand` and other related commands appear.
    -   Ensure the suggestions are case-insensitive (e.g., `/HE` should yield the same results as `/he`).
    -   Use the arrow keys to navigate the suggestions and press `Enter` or `Tab` to autocomplete. Verify the selected command populates the input correctly.
4.  Confirm existing functionality is unaffected:
    -   Test file path completion with `@` to ensure it still works correctly.
    -   Execute a few commands to ensure they run as expected.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Adds implementation of #6551


packages/cli/src/ui/hooks/useSlashCompletion.test.ts
```diff
@@ -431,4 +431,179 @@ describe('useSlashCompletion', () => {
...
+          mockCommandContext,
+        ),
+      );
+
+      expect(result.current.suggestions.length).toBeGreaterThan(0);
+      const labels = result.current.suggestions.map((s) => s.label);
+      expect(labels).toContain('clear');
+      
+      consoleSpy.mockRestore();
+    });
```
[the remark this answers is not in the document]
dracic: Done

packages/cli/src/ui/hooks/useSlashCompletion.ts
```diff
@@ -143,29 +195,72 @@ export function useSlashCompletion(props: UseSlashCompletionProps): {
...
+          potentialSuggestions = commandsToSearch.filter((cmd) => cmd.description);
+        } else {
+          // Use fuzzy search for non-empty partial queries with fallback
+          const fzfInstance = getFzfForCommands(commandsToSearch);
+          if (fzfInstance) {
+            try {
+              const fzfResults = await fzfInstance.fzf.find(partial);
+              potentialSuggestions = fzfResults
+                .map((result: FzfResultItem<string>) => fzfInstance.commandMap.get(result.item))
+                .filter((cmd: SlashCommand | undefined): cmd is SlashCommand => cmd !== undefined && !!cmd.description);
```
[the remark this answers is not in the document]
dracic: Done

dracic:
@gemini-code-assist review

dracic:
@gemini-code-assist review

packages/cli/src/ui/hooks/useSlashCompletion.ts
```diff
@@ -143,29 +193,80 @@ export function useSlashCompletion(props: UseSlashCompletionProps): {
...
-      }
 
-      const finalSuggestions = potentialSuggestions.map((cmd) => ({
-        label: cmd.name,
-        value: cmd.name,
-        description: cmd.description,
-      }));
+        if (potentialSuggestions.length > 0 && !hasTrailingSpace) {
+          const perfectMatch = potentialSuggestions.find(
+            (s) => s.name === partial || s.altNames?.includes(partial),
```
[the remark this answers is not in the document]
dracic: Done. Found more:
  
```
  1. Line 113-114: Command path traversal
  2. Line 130-131: Exact match as parent detection
  3. Line 147-148: Perfect match for isPerfectMatch
  4. Line 243-244: Perfect match in suggestion filtering
```

dracic:
I've done my best here... I have done testing and few code reviews with gemini-cli locally

packages/cli/src/ui/hooks/useSlashCompletion.ts
```diff
@@ -87,36 +86,268 @@ export function useSlashCompletion(props: UseSlashCompletionProps): {
...
+
+interface SuggestionsResult {
+  suggestions: Suggestion[];
+  isLoading: boolean;
+}
+
+function useCommandSuggestions(
+  parserResult: CommandParserResult,
+  commandContext: CommandContext,
+  getFzfForCommands: (commands: readonly SlashCommand[]) => { fzf: AsyncFzf<string[]>; commandMap: Map<string, SlashCommand>; } | null,
```
abhipatel12: Can we define an interface for `{ fzf: AsyncFzf<string[]>; commandMap: Map<string, SlashCommand>; }`

Something like:

```typescript
interface FzfCommandCacheEntry {
  fzf: AsyncFzf<string>;
  commandMap: Map<string, SlashCommand>;
}
```

That way we could make this:

`getFzfForCommands: (commands: readonly SlashCommand[]) => FzfCommandCacheEntry | null,`

packages/cli/src/ui/hooks/useSlashCompletion.ts
```diff
@@ -87,36 +86,268 @@ export function useSlashCompletion(props: UseSlashCompletionProps): {
...
+  getPrefixSuggestions: (commands: readonly SlashCommand[], partial: string) => SlashCommand[]
+): SuggestionsResult {
+  const [suggestions, setSuggestions] = useState<Suggestion[]>([]);
+  const [isLoading, setIsLoading] = useState(false);
+
+  useEffect(() => {
```
abhipatel12, line 166: This hook seems to fetch suggestions async. Are there possible race conditions that could play a role here?

For example, if the user changes input quickly, could this trigger new fetches where older queries could overwrite newer correct suggestions?

If this is a possible concern, we should look into a bool flag that checks if we are active `let isActive = true` at the top of this hook

packages/cli/src/ui/hooks/useSlashCompletion.ts
```diff
@@ -87,36 +86,268 @@ export function useSlashCompletion(props: UseSlashCompletionProps): {
...
+                if (cmd && cmd.description) {
+                  uniqueCommands.add(cmd);
+                }
+              });
+              potentialSuggestions = Array.from(uniqueCommands);
+
+              // If fuzzy search returns no results, fallback to prefix matching
+              if (potentialSuggestions.length === 0) {
+                potentialSuggestions = getPrefixSuggestions(commandsToSearch, partial);
+              }
```
abhipatel12: I may be wrong! But does `fzf` not do prefix matching inherently? If so, we can remove this conditional block

packages/cli/src/ui/hooks/useSlashCompletion.ts
```diff
@@ -87,36 +86,268 @@ export function useSlashCompletion(props: UseSlashCompletionProps): {
...
+    setIsLoadingSuggestions,
+    setIsPerfectMatch,
+  } = props;
+  const [completionStart, setCompletionStart] = useState(-1);
+  const [completionEnd, setCompletionEnd] = useState(-1);
+
+  // Memoized cache for AsyncFzf instances per command level
+  const fzfInstanceCache = useMemo(() => new WeakMap<readonly SlashCommand[], {
+    fzf: AsyncFzf<string[]>;
+    commandMap: Map<string, SlashCommand>;
```
abhipatel12: By moving to an interface from above we can make this easier to read as well.

packages/cli/src/ui/hooks/useSlashCompletion.test.ts
```diff
@@ -431,4 +468,217 @@ describe('useSlashCompletion', () => {
...
   });
+
+  describe('Fuzzy Matching', () => {
+    const fuzzyTestCommands = [
+      { name: 'help', altNames: ['?'], description: 'Show help' },
+      { name: 'history', description: 'Show command history' },
+      { name: 'hello', description: 'Hello world command' },
+      { name: 'config', altNames: ['configure'], description: 'Configure settings' },
+      { name: 'clear', description: 'Clear the screen' },
+    ] as unknown as SlashCommand[];
```
abhipatel12: nit/not a big deal, but do we have to cast `as unknown` b/c of a missing `CommandContext`?

abhipatel12 (commented):
Hey! Thank you for this contribution and sorry it took so long to review!

Overall looks great, had a few concerns. lmk what you think!

Additionally, really appreciate you breaking up the hook into smaller hooks. As a follow-up refactoring if you're interested, I could see us moving some of the completion logic into its own `CommandCompletionService` which would allow us to test without any concern for react concepts.

packages/cli/src/ui/hooks/useSlashCompletion.ts
dracic: Yes:

  1. User types quickly: "test" → "testing" → "testfile"
  2. Each input change triggers useEffect which calls fetchAndSetSuggestions()
  3. Multiple async completion calls are in flight simultaneously
  4. The last completion to resolve wins, not necessarily the most recent query
  
  I'll try to do something not too ugly

packages/cli/src/ui/hooks/useSlashCompletion.ts
dracic: Nice one

packages/cli/src/ui/hooks/useSlashCompletion.ts
dracic: You are right. I'd hire you...

packages/cli/src/ui/hooks/useSlashCompletion.test.ts
dracic: I will create a test utility type and helper function

dracic:
I rebased on top of main, applied all your suggestions, did some cleanup, and now will think through your idea of _I could see us moving some of the completion logic into its own CommandCompletionService_

packages/cli/src/ui/hooks/useSlashCompletion.ts
dracic: Done

packages/cli/src/ui/hooks/useSlashCompletion.ts
dracic: The implementation uses `AbortController `to prevent race conditions

- Fast typing scenario: When user types quickly, each new keystroke triggers a new useEffect
- Previous operations aborted: The cleanup function (abortController.abort()) cancels any in-flight async operations from previous keystrokes
- State updates blocked: Aborted operations skip all setSuggestions() calls via if (!signal.aborted) checks
- Only latest results shown: Only the most recent query's results can update the UI state

packages/cli/src/ui/hooks/useSlashCompletion.ts
```diff
@@ -4,43 +4,72 @@
...
+  item: string;
+  start: number;
+  end: number;
+  score: number;
+  positions?: number[]; // Optional - fzf doesn't always provide match positions depending on algorithm/options used
+};
+
+// Interface for FZF command cache entry (simplified)
+interface FzfCommandCacheEntry {
+  fzf: AsyncFzf<string[]>;
```
abhipatel12, line 23: Why is the generic `string[]` and not `string`? 

Since we're searching for a `string[]` of command names we're looking for a `string` right?

packages/cli/src/ui/hooks/useSlashCompletion.ts
```diff
@@ -4,43 +4,72 @@
...
+// Type alias for improved type safety based on actual fzf result structure
+type FzfCommandResult = {
+  item: string;
+  start: number;
+  end: number;
+  score: number;
+  positions?: number[]; // Optional - fzf doesn't always provide match positions depending on algorithm/options used
+};
+
+// Interface for FZF command cache entry (simplified)
```
abhipatel12: nit: we can remove simplified

packages/cli/src/ui/hooks/useSlashCompletion.ts
abhipatel12: This is great!

abhipatel12 (approved):
Thanks for making the changes! This looks great! 

We can explore if we want to extract into a service in a follow-up pr!

abhipatel12:
Went ahead and gave my LGTM, but a few small comments. Lmk what you think

dracic:
> Thanks for making the changes! This looks great!
> 
> We can explore if we want to extract into a service in a follow-up pr!

Yesterday, I attempted to refactor some code into a service but encountered some regressions with the built-in commands. It's also possible I didn't test this properly, so I'll start over from scratch. If you're okay with this, I'll make the suggested changes, format the code (I forgot about that), and then create a single new commit.

Edit: I've tested this again, and the built-in commands are working correctly, so I obviously made a mistake during my refactoring attempt. We can fine-tune the fuzzy matching further, but I think this is okay for a start.

packages/cli/src/ui/hooks/useSlashCompletion.ts
dracic: You're absolutely right! Hahahaha.

packages/cli/src/ui/hooks/useSlashCompletion.ts
dracic: Ok

packages/cli/src/ui/hooks/useSlashCompletion.ts
dracic: Well, run the build and paf -  `error TS2344: Type 'string' does not satisfy the constraint 'readonly any[]'.`

This suggests the original code was actually correct - AsyncFzf expects the generic parameter to be the collection type, not the item type.

https://fzf.netlify.app/docs/latest#api

dracic: How Our Implementation Works

  1. Array Creation: We build commandItems: string[] = ["help", "mcp", "clear"] from all command names
  2. FZF Instance: new AsyncFzf(commandItems) requires AsyncFzf<string[]> generic
  3. User Types: When user types /ed, FZF searches the string array and returns name and fuzzy match score [{item: "mcp", score: 0.8}]
  4. Command Lookup: We use commandMap.get("edit") to convert the string result back to the actual SlashCommand object

packages/cli/src/ui/hooks/useSlashCompletion.ts
dracic: This is a leftover from test on two options - complex and simplified.

packages/cli/src/ui/hooks/useSlashCompletion.ts
abhipatel12: Ohh thanks for explaining. Sorry about that, sounds good! TIL

abhipatel12 (approved):
This all looks great! Thanks!

jacob314 (approved):
lgtm

PR #6655, opened 2025-08-20
return the JSON stringified parameters from getDescription for MCP tools and Discovered tools

jakemac53:
## TLDR

Updates `getDescription()` for MCP tools and Discovered tools to be the JSON stringified parameters instead of just the tool name.

The tool name already appears in the display so this was being duplicated previously, the description is really the details of the invocation.

## Dive Deeper

Ultimately long term we might want to improve the display further here, rendering the arguments in a yaml-ish format or something more human readable, the current format is compact but somewhat hard to decipher and I think it can also get trimmed.

Ultimately this restores the previous behavior though.

## Reviewer Test Plan

Ask Gemini to invoke any discovered or MCP tool and you should now see the arguments displayed instead of the name of the tool duplicated.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes https://github.com/google-gemini/gemini-cli/issues/6531

### DIFF
--- a/packages/core/src/tools/mcp-tool.test.ts
+++ b/packages/core/src/tools/mcp-tool.test.ts
@@ -746,10 +746,10 @@ describe('DiscoveredMCPTool', () => {
 
   describe('DiscoveredMCPToolInvocation', () => {
     it('should return the stringified params from getDescription', () => {
-      const params = { param: 'testValue' };
+      const params = { param: 'testValue', param2: 'anotherOne' };
       const invocation = tool.build(params);
       const description = invocation.getDescription();
-      expect(description).toBe('{"param":"testValue"}');
+      expect(description).toBe('{"param":"testValue","param2":"anotherOne"}');
     });
   });
 });
--- a/packages/core/src/tools/mcp-tool.test.ts
+++ b/packages/core/src/tools/mcp-tool.test.ts
@@ -743,4 +743,13 @@ describe('DiscoveredMCPTool', () => {
       }
     });
   });
+
+  describe('DiscoveredMCPToolInvocation', () => {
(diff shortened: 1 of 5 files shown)


jakemac53:
I will also look into how to properly test this.

jakemac53:
For now I just added some really basic unit tests that don't have a ton of value but will at least fail if somebody changes the descriptions, and hopefully they can git blame their way back to this PR if they don't understand why the description is what it is.

packages/core/src/tools/mcp-tool.test.ts
```diff
@@ -743,4 +743,13 @@ describe('DiscoveredMCPTool', () => {
       }
     });
   });
+
+  describe('DiscoveredMCPToolInvocation', () => {
+    it('should return the stringified params from getDescription', () => {
+      const params = { param: 'testValue' };
+      const invocation = tool.build(params);
+      const description = invocation.getDescription();
+      expect(description).toBe(JSON.stringify(params));
```
jacob314: nit: can one of the tests show the actual output rather than verifying it matches what was computed. Also include multiple params to make sure it is human readable enough.

jacob314 (approved):
lgtm

packages/core/src/tools/mcp-tool.test.ts
jakemac53: done, fwiw the readability could use some work in general

PR #6656, opened 2025-08-20
Improve Zed integration error handling

agu-z:
## TLDR

Improves error handling for the Zed integration by handling aborted requests in `prompt`, returning error objects as JSON instead of strings, and handling invalid JSON-RPC messages.

## Dive Deeper

- The `prompt` uses an `AbortController` to support cancellation by checking the abort signal in the loop, and passing it to the Gemini Client functions and tool invocations. When aborted, the latter may emit `Aborted` errors, which were returned to the client as internal errors instead of using the semantically meaningful `{ stop_reason: cancelled }` response.   
- Some internal errors from the Gemini Client and other dependencies produce JSON-encoded objects that were being sent as strings. We'll now parse them and send them as objects in JSON-RPC's `data` field. 
- If an invalid JSON-RPC message was received, we weren't gracefully handling it. We now return the corresponding `RequestError`s according to the spec.

## Reviewer Test Plan

To fully test the integration, you'll need to [build Zed from source](https://github.com/zed-industries/zed?tab=readme-ov-file#developing-zed), and add the following to your settings:

```json
{
  "agent_servers": {
    "gemini": {
      "command": "node",
      "args": ["/<PATH>/<TO>/gemini-cli/packages/cli"]
    }
  },
}
```

Feel free to ask @agu-z or @ConradIrwin  for any help getting set up.


chrstnb:

PR #6657, opened 2025-08-20
Limit dependabot PRs to security updates

cornmander:
Configure dependabot to only send PRs for security updates:

https://docs.github.com/en/code-security/dependabot/dependabot-security-updates/customizing-dependabot-security-prs#about-customizing-pull-requests-for-security-updates


.github/dependabot.yml
```diff
@@ -20,6 +20,7 @@ updates:
         update-types:
           - 'minor'
           - 'patch'
+    open-pull-requests-limit: 0
```
mattKorwel, line 23: how does this limit it to security issue?

mattKorwel (approved):
lgtm if it lgtYou

.github/dependabot.yml
cornmander: I agree that this is confusingly named, but it's from the github docs: If you want the customization for a package ecosystem to only apply to security updates (and exclude version updates), set the open-pull-requests-limit key to 0.

PR #6665, opened 2025-08-20
feat: auto-approve pending tool calls when auto_edit/yolo is activated

bulkypanda:
## TLDR

continue processing pending tool calls if auto edit or yolo is triggered via keybinds

## Dive Deeper

- Integrate pending tool calls into auto-accept indicator hook
- Add auto-approval of pending tool calls when switching to YOLO/AUTO_EDIT modes
- Update ToolGroupMessage spacing for better visual separation
- Add comprehensive test coverage for new auto-approval functionality
- Export pending tool calls from useGeminiStream hook

## Reviewer Test Plan

enable auto edit or yolo with pending tool calls using keybinds

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ✅  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


packages/cli/src/ui/hooks/useAutoAcceptIndicator.ts
```diff
@@ -43,6 +55,46 @@ export function useAutoAcceptIndicator({
...
+          if (nextApprovalMode === ApprovalMode.AUTO_EDIT) {
+            awaitingApprovalCalls = awaitingApprovalCalls.filter(
+              (call) =>
+                call.request.name === 'replace' ||
+                call.request.name === 'write_file',
+            );
+          }
+
+          // Process all pending tool calls concurrently
+          Promise.all(
```
jacob314: nit: approve one at a time so there is less chaos in the UI.

packages/cli/src/ui/hooks/useAutoAcceptIndicator.ts
```diff
@@ -43,6 +55,46 @@ export function useAutoAcceptIndicator({
         config.setApprovalMode(nextApprovalMode);
         // Update local state immediately for responsiveness
         setShowAutoAcceptIndicator(nextApprovalMode);
+
+        // Auto-approve pending tool calls when switching to auto-approval modes
```
jacob314: this logic seems like it should be in a more central place than useAutoAcceptIndicator. Can you centralize this with the logic you added on the previous pull request. seems like similar concepts. we need to let useGeminiStream know to handle its queue of things that should be approved but weren't because we just switched settings.

packages/cli/src/ui/App.tsx
```diff
@@ -659,7 +660,10 @@ const App = ({ config, startupWarnings = [], version }: AppProps) => {
 
   const { elapsedTime, currentLoadingPhrase } =
     useLoadingIndicator(streamingState);
-  const showAutoAcceptIndicator = useAutoAcceptIndicator({ config });
+  const showAutoAcceptIndicator = useAutoAcceptIndicator({
+    config,
+    pendingToolCalls,
```
jacob314: you should be able to implement without coupling the auto accept indicator to the pendingToolCalls. instead listen for changes to the accept mode likely from useGeminiStream. ask gemini cli what reasonable ways to do this are.

jacob314 (commented):
This is a great case to handle. some comments on how to handle it.

bulkypanda:
all addressed.

packages/cli/src/ui/hooks/useGeminiStream.ts
```diff
@@ -728,6 +731,47 @@ export const useGeminiStream = (
     ],
   );
 
+  const handleApprovalModeChange = useCallback(
```
jacob314, line 852: Can you add tests for this method?

jacob314 (commented):
Looks good! Approved once there are tests for the useGeminiStream.ts changes.

bulkypanda:
#6941 and #7037 should be approved and merged before this.

packages/cli/src/ui/hooks/useAutoAcceptIndicator.ts
```diff
@@ -48,14 +62,62 @@ export function useAutoAcceptIndicator({
           config.setApprovalMode(nextApprovalMode);
           // Update local state immediately for responsiveness
           setShowAutoAcceptIndicator(nextApprovalMode);
+
+          // Auto-approve pending tool calls when switching to auto-approval modes
```
jacob314: can you centralize this auto-approve logic? would be nice to not add it to useAutoAcceptIndicator and instead keep it on useGeminiStream. 
useGeminiStream could track the approval mode and fix things up with then approvalMode changes flushing its tool calls that should be approved. I think that might slightly reduce the amount of dupe logic for this and keep things like useAutoAcceptIndicator less coupled to the business logic for this.

jacob314: looks like you may have already done that a bit in useGeminiStream but this code wasn't deleted yet.

bulkypanda: oops, forgot to delete duplicate code

packages/cli/src/ui/hooks/useGeminiStream.ts
```diff
@@ -789,6 +789,47 @@ export const useGeminiStream = (
...
+          (call): call is TrackedWaitingToolCall =>
+            call.status === 'awaiting_approval',
+        );
+
+        // For AUTO_EDIT mode, only approve edit tools (replace, write_file)
+        if (newApprovalMode === ApprovalMode.AUTO_EDIT) {
+          awaitingApprovalCalls = awaitingApprovalCalls.filter(
+            (call) =>
+              call.request.name === 'replace' ||
+              call.request.name === 'write_file',
```
abhipatel12: We should extract this into a new constant / set at the top so that if we add new edit tools, we don't have to come find this.

jacob314 (approved):
lgtm

PR #6680, opened 2025-08-20
feat(core): Handle special characters in file search paths

SandyTao520:
## TLDR

This pull request fixes a bug in the file search functionality that prevented it from finding files with special characters (like parentheses) in their paths. The search pattern is now unescaped before being used, ensuring that paths with special characters are resolved correctly.

## Dive Deeper

The `RecursiveFileSearch` implementation was passing the user-provided search pattern directly to the `fzf` engine. This caused issues when a path contained special characters that are typically escaped in a shell environment (e.g., `src/file (special).txt`). A user trying to find this file might provide an escaped pattern like `src/file \\(special\\).txt`.

This change introduces a call to a new `unescapePath` utility function, which normalizes the search pattern by removing shell escapes. This ensures that the search engine looks for the literal file path, allowing it to find files with special characters correctly.

A new unit test has been added to verify this specific scenario, ensuring that a file with a path containing `(special)` can be found using an escaped search pattern.

## Reviewer Test Plan

1. Create a new file in your test project with special characters in its name, for example:
    ```bash
    touch "my-file-(1).log"
    ```
2. Use the Gemini CLI to search for this file using the escaped path:
    ```bash
    look at @my-file-\(
    ```
3. Verify that the CLI correctly shows ``my-file-(1).log`` as a suggestion.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ✅  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

### DIFF
--- a/packages/core/src/utils/filesearch/fileSearch.test.ts
+++ b/packages/core/src/utils/filesearch/fileSearch.test.ts
@@ -566,6 +566,35 @@ describe('FileSearch', () => {
     expect(limitedResults).toEqual(['file1.js', 'file2.js']);
   });
 
+  it('should handle file paths with special characters that need escaping', async () => {
+    tmpDir = await createTmpDir({
+      src: {
+        'file with (special) chars.txt': '',
+        'another-file.txt': '',
+      },
+    });
+
+    const fileSearch = FileSearchFactory.create({
+      projectRoot: tmpDir,
+      useGitignore: false,
+      useGeminiignore: false,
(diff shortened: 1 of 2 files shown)


jacob314 (approved):
lgtm

PR #6694, opened 2025-08-20
feat(shell): enable interactive commands with virtual terminal

### LINKED ISSUE #5132: Support interaction with terminal windows embedded in Gemini CLI

galdawave:
## TLDR

This pull request enables interactive shell commands by integrating a pseudo-terminal (pty) into the `run_shell_command` tool. This allows Gemini to run commands like `vim`, `top`, or `git rebase -i` and correctly render their UI, a capability that was previously unsupported.

## Dive Deeper

The core of this change is the replacement of the previous raw stream handling in `ShellExecutionService` with a headless `xterm.js` terminal backed by `node-pty`. This architectural shift provides several key benefits:

1.  **Full Terminal Emulation**: The headless terminal correctly interprets terminal escape codes, cursor movements, and screen clearing. This means that the output streamed to the client is a more accurate representation of the terminal's state, enabling the rendering of complex TUIs.
2.  **Interactive Process Communication**: We've introduced `writeToShell` and `resizeShell` methods on the `GeminiClient`. This creates a two-way communication channel, allowing the client to send user input (keystrokes) and terminal resize events back to the running process.
3.  **Improved Output Handling**: The `isBinary` utility has been made more sophisticated. The previous implementation would often misclassify rich TTY output with ANSI color codes as a binary stream. The new logic is more discerning, ensuring that colorful command-line tool output is preserved as text.
4.  **System Prompt Update**: The prompt instruction that discouraged the model from using interactive commands has been removed, unlocking this new capability.

These changes provide the foundational support for a much richer, interactive terminal experience within Gemini.

## Reviewer Test Plan

To validate these changes, pull down the branch and run the CLI with prompts that invoke interactive or TUI-based applications.

1.  **Basic Interactivity**:
    *   Run shell mode and run `ng new`.
    *   (Follow-up) Send input to write and save the file.

4.  **Resizing**:
    *   While an interactive command is running, resize your terminal window. The TUI within the command should adapt to the new dimensions.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

- Fixes #5132


packages/cli/src/ui/App.tsx
```diff
@@ -874,6 +930,20 @@ const App = ({ config, startupWarnings = [], version }: AppProps) => {
...
+        geminiClient.resizeShell(
+          activeShellPtyId,
+          Math.floor(terminalWidth * 0.5),
+          Math.floor(terminalHeight * 0.5),
+        );
+      }
+    },
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [terminalHeight, terminalWidth],
+  );
```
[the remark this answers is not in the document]
jacob314: please fix

packages/cli/src/config/keyBindings.ts
```diff
@@ -163,7 +164,7 @@ export const defaultKeyBindings: KeyBindingConfig = {
   // Original: key.ctrl && key.name === 'o'
   [Command.SHOW_ERROR_DETAILS]: [{ key: 'o', ctrl: true }],
   // Original: key.ctrl && key.name === 't'
-  [Command.TOGGLE_TOOL_DESCRIPTIONS]: [{ key: 't', ctrl: true }],
+  [Command.TOGGLE_TOOL_DESCRIPTIONS]: [{ key: 'i', ctrl: true }],
```
jacob314: make sure code telling users about ctrl-t now ctrl-i is updated. same for markdown docs.
alternately take the path of least resistance and pick a new keyboard shortcut for this

packages/cli/src/ui/components/messages/ToolGroupMessage.tsx
```diff
@@ -9,9 +9,8 @@ import { Box } from 'ink';
 import { IndividualToolCallDisplay, ToolCallStatus } from '../../types.js';
 import { ToolMessage } from './ToolMessage.js';
 import { ToolConfirmationMessage } from './ToolConfirmationMessage.js';
-import { Colors } from '../../colors.js';
 import { Config } from '@google/gemini-cli-core';
-import { SHELL_COMMAND_NAME } from '../../constants.js';
+import { theme } from '../../semantic-colors.js';
```
jacob314: you are making Miguel happy by switching to this :)

packages/cli/src/ui/components/messages/ToolGroupMessage.tsx
```diff
@@ -87,12 +102,7 @@ export const ToolGroupMessage: React.FC<ToolGroupMessageProps> = ({
           <Box key={tool.callId} flexDirection="column" minHeight={1}>
             <Box flexDirection="row" alignItems="center">
               <ToolMessage
-                callId={tool.callId}
-                name={tool.name}
-                description={tool.description}
-                resultDisplay={tool.resultDisplay}
-                status={tool.status}
-                confirmationDetails={tool.confirmationDetails}
+                {...tool}
```
jacob314, line 107: wow that is cleaner. wonder where else we should be doing this.

packages/cli/src/ui/components/ShellInputPrompt.tsx
```diff
@@ -0,0 +1,59 @@
...
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import React, { useCallback, useEffect, useState } from 'react';
+import { Box, Text } from 'ink';
+import { useKeypress, Key, keyToAnsi } from '../hooks/useKeypress.js';
+import chalk from 'chalk';
+
+const CURSOR_BLINK_RATE_MS = 500;
```
jacob314: move this to a const and adopt in in the InputPrompt as well. we shouldn't be blinking in one case and not the other.

packages/cli/src/ui/hooks/useGeminiStream.ts
```diff
@@ -159,6 +168,22 @@ export const useGeminiStream = (
     onDebugMessage,
     config,
     geminiClient,
+    (pid) => {
```
jacob314: can we refactor to avoid this? this update drilling into the details to get the ptyId is a bit of a code smell. For example, can we just store the ptyId in the history item in a way that avoids this?

packages/cli/src/ui/hooks/useGeminiStream.ts
```diff
@@ -114,6 +118,11 @@ export const useGeminiStream = (
     return new GitService(config.getProjectRoot(), storage);
   }, [config, storage]);
 
+  const getTerminalSize = () => ({
+    columns: terminalWidth ?? process.stdout.columns,
```
jacob314: rather than this can we make terminalWidth/height required here fixing any cases where we aren't passing this in. there may be one case we intentionally set it to undefined (the expand to show option but that could be cleaned up by passing an extra boolean).

packages/cli/src/ui/hooks/useReactToolScheduler.ts
```diff
@@ -65,33 +66,17 @@ export type TrackedToolCall =
 export function useReactToolScheduler(
   onComplete: (tools: CompletedToolCall[]) => Promise<void>,
   config: Config,
-  setPendingHistoryItem: React.Dispatch<
-    React.SetStateAction<HistoryItemWithoutId | null>
-  >,
   getPreferredEditor: () => EditorType | undefined,
   onEditorClose: () => void,
+  getTerminalSize: () => { columns: number; rows: number },
```
jacob314: would be nice to avoid having the tool scheduler know about the terminal size

packages/cli/src/ui/App.tsx
```diff
@@ -594,6 +599,10 @@ const App = ({ config, startupWarnings = [], version }: AppProps) => {
     setModelSwitchedFromQuotaError,
     refreshStatic,
     () => cancelHandlerRef.current(),
+    setShellInputFocused,
+    Math.floor(terminalWidth * 0.5),
```
jacob314: why half height and half width?

packages/cli/src/ui/App.tsx
```diff
@@ -654,8 +663,49 @@ const App = ({ config, startupWarnings = [], version }: AppProps) => {
...
+            break;
+          }
+        }
+      }
+      if (ptyId) {
+        break;
+      }
+    }
+    setActiveShellPtyId(ptyId);
+  }, [pendingHistoryItems]);
```
jacob314: would be nice to plumb this data through a different way so we didn't have to search for this.

packages/cli/src/ui/App.tsx
```diff
@@ -654,8 +663,49 @@ const App = ({ config, startupWarnings = [], version }: AppProps) => {
...
+
+  const handleShellInputSubmit = useCallback(
```
jacob314: can you get shellInputSubmit out of App.tsx. Would expect this should be more local in the UI actually rendering the particular shell.

packages/cli/src/ui/App.tsx
```diff
@@ -654,8 +663,49 @@ const App = ({ config, startupWarnings = [], version }: AppProps) => {
...
+    (input: string) => {
+      if (activeShellPtyId) {
+        config.getGeminiClient().writeToShell(activeShellPtyId, input);
+      }
+    },
+    [activeShellPtyId, config],
+  );
+
+  useEffect(() => {
```
jacob314: would be nice to remove this useEffect as well

packages/cli/src/ui/App.tsx
```diff
@@ -874,6 +930,20 @@ const App = ({ config, startupWarnings = [], version }: AppProps) => {
   const initialPrompt = useMemo(() => config.getQuestion(), [config]);
   const geminiClient = config.getGeminiClient();
 
+  useEffect(
+    () => {
+      if (activeShellPtyId) {
+        geminiClient.resizeShell(
+          activeShellPtyId,
+          Math.floor(terminalWidth * 0.5),
```
jacob314: we shouldn't have to use terminalWidth * 0.5. We go through a bunch of work to know the exact render width for all the tool response stuff so we can manually do our own word wrapping using maxSizeBox

packages/core/src/core/prompts.ts
```diff
@@ -111,7 +111,6 @@ When requested to perform tasks like fixing bugs, adding features, refactoring,
 - **Parallelism:** Execute multiple independent tool calls in parallel when feasible (i.e. searching the codebase).
 - **Command Execution:** Use the '${ShellTool.Name}' tool for running shell commands, remembering the safety rule to explain modifying commands first.
 - **Background Processes:** Use background processes (via \`&\`) for commands that are unlikely to stop on their own, e.g. \`node server.js &\`. If unsure, ask the user.
-- **Interactive Commands:** Try to avoid shell commands that are likely to require user interaction (e.g. \`git rebase -i\`). Use non-interactive versions of commands (e.g. \`npm init -y\` instead of \`npm init\`) when available, and otherwise remind the user that interactive shell commands are not supported and may cause hangs until canceled by the user.
```
jacob314: lets leave this for a separate pr. this might degrade model quality. no need to force the model to start triggering more interactive commands. interactivity is more for when the user does shell commands not the model for all but the most advanced use cases.

packages/core/src/services/shellExecutionService.ts
```diff
@@ -76,12 +77,19 @@ export type ShellOutputEvent =
       bytesReceived: number;
     };
 
+interface ActivePty {
+  ptyProcess: IPty;
+  // @ts-expect-error Terminal type has issues.
```
jacob314: please remove this ts-expect-error and resolve if at all possible

packages/core/src/services/shellExecutionService.ts
```diff
@@ -355,6 +378,24 @@ export class ShellExecutionService {
         const MAX_SNIFF_SIZE = 4096;
         let sniffedBytes = 0;
 
+        let renderTimeout: NodeJS.Timeout | null = null;
+        const RENDER_INTERVAL = 100; // roughly 60fps
```
jacob314: nit: this is 10fps not 60fps. what frame rate do you intend?

packages/core/src/services/shellExecutionService.ts
```diff
@@ -355,6 +378,24 @@ export class ShellExecutionService {
...
+
+        const render = () => {
+          renderTimeout = null;
+          const newStrippedOutput = getFullText(headlessTerminal);
```
jacob314: I would expect the headlessTerminal has some apis to notify of changes so we don't have to poll.

packages/core/src/utils/textUtils.ts
```diff
@@ -5,30 +5,80 @@
  */
 
 /**
- * Checks if a Buffer is likely binary by testing for the presence of a NULL byte.
- * The presence of a NULL byte is a strong indicator that the data is not plain text.
+ * A more robust check to determine if a Buffer contains binary data.
```
jacob314: avoid comments that just describe how code changed

packages/core/src/utils/textUtils.ts
```diff
@@ -5,30 +5,80 @@
...
+ * This function uses several heuristics:
+ * 1.  It checks for the presence of a NULL byte, which is a very strong
+ *     indicator of binary data.
+ * 2.  It allows for common text control characters like tabs, newlines, and
+ *     carriage returns, as well as the ANSI escape character (0x1B).
+ * 3.  It calculates the percentage of other "suspicious" bytes (i.e., other
+ *     control characters or bytes outside the standard printable ASCII range).
+ * 4.  If this percentage exceeds a certain threshold, the buffer is considered binary.
+ *
+ * This approach is designed to correctly classify interactive TTY output, which
```
jacob314: can this be moved to a separate pr. curious to understand why these heuristics need to change. these are used in other places so we should be careful about changing this to avoid false positives or performance issues with the crucial readFiles and readManyFiles tools?

packages/core/src/utils/textUtils.ts
```diff
@@ -5,30 +5,80 @@
...
   if (!data) {
     return false;
   }
 
   const sample = data.length > sampleSize ? data.subarray(0, sampleSize) : data;
+  if (sample.length === 0) {
+    return false;
+  }
+
+  // Check for a Byte Order Mark (BOM), which indicates text.
```
jacob314: why is this added? would think this only matters for certain windows cases and generally I believe we have a bunch of problems for BOM text

packages/core/src/utils/textUtils.ts
```diff
@@ -5,30 +5,80 @@
...
+    if (!isPrintableAscii && !isCommonControlChar && !isAnsiEscape) {
+      suspiciousBytes++;
+    }
+  }
+
+  // If more than 30% of the sample consists of suspicious characters,
+  // it's likely binary. This threshold is a heuristic chosen to be lenient
+  // enough for TTY output, which can contain many ANSI control codes, while
+  // still reliably detecting most binary file types.
+  if (suspiciousBytes / sample.length > 0.3) {
```
jacob314: lets only use this as a heuristic for terminals and leave the existing code for the calls to this util for readManyFiles, etc. then we can try updating readmanyFiles along with careful eval to make sure this doesn't hurt quality... possible it might help if we were false positiving too much before. might also hurt if we now start confusing the model by letting binary files into read files calls.

jacob314 (commented):
Super excited about this feature. Hopefully Gemini CLI can help address most of these comments. The most important ones are the ones to derisk landing this change.

Other UI issues noticed. Need to make the regular InputPrompt show up as unfocused when focus switches to the terminal. Also curious why the cursor inside the embedded Gemini CLI input prompt doesn't work. Potentially we aren't notifying the sub terminal that it has focus.
Need to align on whether we use a flickering cursor. Either we should use it everywhere or nowhere. 

We should discuss the keyboard shortcuts for this. Potentially we should use tab to toggle focus. Can you work with @miguelsolorio to figure out reasonable user interactions for that.
We should show focus by changing the outline color of the shell command area box. We've meant to do that for InputPrompt as well but never got around to it. @miguelsolorio should have ideas on colors to use. I'd suggest the shell has the yellow outline when focused.

packages/cli/src/config/keyBindings.ts
galdawave: Yea ctrl+t made the most sense but i can move it, any suggestions for another key?

packages/cli/src/ui/components/ShellInputPrompt.tsx
```diff
@@ -0,0 +1,59 @@
...
+      }
+    },
+    [focus, onSubmit],
+  );
+
+  useKeypress(handleInput, { isActive: focus });
+
+  const cursor = isCursorVisible ? chalk.inverse(' ') : ' ';
+
+  return <Box>{focus && <Text>{cursor}</Text>}</Box>;
```
jacob314: ok now that I understand the code better, rather than making this character blink we should make the character that the xterm headless library says the cursor is at blink when the xterm headless says the cursor location is visible. that will provide a cleaner UI and not make it feel like the cursor is in a broken location.

packages/cli/src/ui/components/messages/ToolGroupMessage.tsx
galdawave: 😄

packages/cli/src/ui/components/messages/ToolGroupMessage.tsx
galdawave: Can take a look in a separate refactor PR

packages/core/src/utils/textUtils.ts
galdawave: Reverted textUtils.ts changes

packages/cli/src/ui/App.tsx
galdawave: i set it to that because i found that basically anything we render with that width and height would look good, i'm not opinionated on this so i can add subtract some padding from the width so it takes the width of the whole shell output. 

before

Image

after 

Image

packages/cli/src/ui/App.tsx
galdawave: Same comment as above.

packages/cli/src/ui/components/ShellInputPrompt.tsx
galdawave: Done.

packages/core/src/config/config.ts
```diff
@@ -200,6 +200,8 @@ export interface ConfigParameters {
   trustedFolder?: boolean;
   shouldUseNodePtyShell?: boolean;
   skipNextSpeakerCheck?: boolean;
+  terminalWidth?: number;
```
jacob314: can we move the height and width to be a new ShellExecutionConfig sub object similar to SandboxConfig.
That way it is clear this this is only something for spinning up shell execution not something people who are generally interested in the terminal width+height should use. Also makes it clear that clients might set this something arbitrary based on their interest in how big to by default make terminals they will render.

packages/core/src/core/prompts.ts
```diff
@@ -111,6 +111,7 @@ When requested to perform tasks like fixing bugs, adding features, refactoring,
 - **Parallelism:** Execute multiple independent tool calls in parallel when feasible (i.e. searching the codebase).
 - **Command Execution:** Use the '${ShellTool.Name}' tool for running shell commands, remembering the safety rule to explain modifying commands first.
 - **Background Processes:** Use background processes (via \`&\`) for commands that are unlikely to stop on their own, e.g. \`node server.js &\`. If unsure, ask the user.
+- **Interactive Commands:** Try to avoid shell commands that are likely to require user interaction (e.g. \`git rebase -i\`). Use non-interactive versions of commands (e.g. \`npm init -y\` instead of \`npm init\`) when available, and otherwise remind the user that interactive shell commands are not supported and may cause hangs until canceled by the user.
```
jacob314: Fyi @anj-s for the prompt changes here. Would want to make sure what we do here doesn't regress quality.
I'd suggest we try to make this a bit more terse and a bit more centered on saying what to do rather than what to avoid.

packages/core/src/services/shellExecutionService.test.ts
```diff
@@ -180,7 +180,6 @@ describe('ShellExecutionService', () => {
     });
 
     it('should call onPid with the process id', async () => {
-      const onPid = vi.fn();
```
jacob314: this test seems misleading with onPid removed.

packages/core/src/services/shellExecutionService.test.ts
```diff
@@ -425,7 +423,7 @@ describe('ShellExecutionService child_process fallback', () => {
       expect(result.error).toBeNull();
       expect(result.aborted).toBe(false);
       expect(result.output).toBe('file1.txt\na warning');
-      expect(handle.pid).toBe(12345);
+      expect(handle.pid).toBe(undefined);
```
jacob314, line 525: why did this expectation change?

packages/core/src/services/shellExecutionService.ts
```diff
@@ -239,7 +236,7 @@ export class ShellExecutionService {
             signal: signal ? os.constants.signals[signal] : null,
             error,
             aborted: abortSignal.aborted,
-            pid: child.pid,
+            pid: undefined,
```
jacob314, line 261: why don't we fallback to child.pid anymore?

packages/cli/src/ui/App.tsx
```diff
@@ -208,6 +208,25 @@ const App = ({ config, settings, startupWarnings = [], version }: AppProps) => {
   const [showEscapePrompt, setShowEscapePrompt] = useState(false);
   const [isProcessing, setIsProcessing] = useState<boolean>(false);
   const [shellInputFocused, setShellInputFocused] = useState(false);
+  const [cursorPosition, setCursorPosition] = useState<{
```
jacob314: can you keep this login in the shell rather than in App.tsx to keep App.tsx small

packages/cli/src/ui/App.tsx
```diff
@@ -836,10 +853,13 @@ const App = ({ config, settings, startupWarnings = [], version }: AppProps) => {
       refreshStatic();
     }, 300);
 
+    config.setTerminalHeight(terminalHeight*0.5);
```
jacob314: add constants for the shellExecutionTerminalHeight and width to avoid the repetition of this logic.  would like to follow up these prs my changing it to match the actual height and width available rather than setting to 0.5 which is a lot more narrow than would be ideal.

packages/core/src/core/prompts.ts
galdawave: I reverted the prompt changes to what they are at head, so there shouldn't be any regressions due to the prompt.

packages/core/src/services/shellExecutionService.ts
galdawave: For child_process we don't care about the pid, since this is being used to focus the terminal onto the process which shouldn't happen for child_process

packages/core/src/services/shellExecutionService.test.ts
galdawave: Same as bellow, we use the pid to focus on the terminal so for child_process we don't need to define the pid.

chrstnb:
Non-urgent, but we talked about supporting this in IDE mode. would be a cool follow-up to support opening the terminal within the IDE. Here's a branch where I was playing around with this: https://github.com/google-gemini/gemini-cli/compare/main...cb/terminal

jacob314:
> Non-urgent, but we talked about supporting this in IDE mode. would be a cool follow-up to support opening the terminal within the IDE. Here's a branch where I was playing around with this: [main...cb/terminal](https://github.com/google-gemini/gemini-cli/compare/main...cb/terminal)

It would be great to investigate this after this pr lands. This pr is already large enough that it is hard to review robustly.

packages/cli/src/ui/hooks/useReactToolScheduler.ts
```diff
@@ -102,16 +102,22 @@ export function useReactToolScheduler(
...
+            return {
+              ...coreTc,
+              responseSubmittedToGemini,
+              liveOutput: (existingTrackedCall as TrackedExecutingToolCall)
+                ?.liveOutput,
+              pid: (coreTc as ExecutingToolCall).pid,
+            };
+          }
+
+          // For other statuses, we don't want to add liveOutput or pid
```
jacob314: nit: for other statuses, shouldn't the pid and liveOutput just be undefined?

packages/core/src/core/coreToolScheduler.ts
```diff
@@ -893,8 +896,32 @@ export class CoreToolScheduler {
             : undefined;
 
         const shellExecutionConfig = this.config.getShellExecutionConfig();
-        invocation
-          .execute(signal, liveOutputCallback, shellExecutionConfig)
+
+        let promise: Promise<ToolResult>;
+        if (invocation instanceof ShellToolInvocation) {
+          const setPidCallback = (pid: number) => {
```
jacob314, line 979: special casing this for ShellToolInvocation is a bit ugly. would be ideal if there is a simpler way but won't block the pr on that. perhaps add a todo.

packages/cli/src/config/settingsSchema.ts
```diff
@@ -625,6 +625,36 @@ const SETTINGS_SCHEMA = {
...
+        properties: {
+          pager: {
+            type: 'string',
+            label: 'Pager',
+            category: 'Tools',
+            requiresRestart: false,
+            default: 'cat' as string | undefined,
+            description:
+              'The pager command to use for shell output. Defaults to `cat`.',
+            showInDialog: true,
```
jacob314: can we not show this in the dialog?

packages/cli/src/ui/components/AnsiOutput.tsx
```diff
@@ -0,0 +1,46 @@
...
+}
+
+export const AnsiOutputText: React.FC<AnsiOutputProps> = ({
+  data,
+  availableTerminalHeight,
+}) => {
+  const lastLines = data.slice(
+    -(availableTerminalHeight && availableTerminalHeight > 0
+      ? availableTerminalHeight
+      : 24),
```
jacob314: nit: make 24 a constant

packages/cli/src/ui/components/AnsiOutput.tsx
```diff
@@ -0,0 +1,46 @@
...
+            dimColor={token.dim}
+            bold={token.bold}
+            italic={token.italic}
+            underline={token.underline}
+          >
+            {token.text}
+          </Text>
+        ))
+      ) : (
+        <Text> </Text>
```
jacob314: nit: why are we adding a space if the line length is 0?

packages/cli/src/ui/hooks/shellCommandProcessor.test.ts
```diff
@@ -208,46 +219,6 @@ describe('useShellCommandProcessor', () => {
       vi.useRealTimers();
     });
 
-    it('should throttle pending UI updates for text streams', async () => {
```
jacob314, line 211: nit: is removing this intentional? seems we should still have it for the non interactive cases?

packages/cli/src/ui/components/InputPrompt.tsx
```diff
@@ -48,6 +48,7 @@ export interface InputPromptProps {
   setShellModeActive: (value: boolean) => void;
   onEscapePromptChange?: (showPrompt: boolean) => void;
   vimHandleInput?: (key: Key) => boolean;
+  isShellInputFocused?: boolean;
```
jacob314: rename this field. it isn't anything to to with the shell. it is just whether this input is focused

packages/cli/src/ui/hooks/useKeypress.ts
```diff
@@ -11,7 +11,81 @@ import { useKeypressContext } from '../contexts/KeypressContext.js';
 export type { Key };
 
 /**
```
jacob314: nit: move these out of useKeypress into a util file for your psuedo-terminal support as these aren't used elsewhere in useKeypress

packages/cli/src/ui/AppContainer.tsx
```diff
@@ -618,6 +633,22 @@ Logging in with Google... Please restart Gemini CLI to continue.
   const initialPromptSubmitted = useRef(false);
   const geminiClient = config.getGeminiClient();
 
+  useEffect(() => {
+    if (activePtyId) {
+      ShellExecutionService.resizePty(
+        activePtyId,
+        Math.floor(terminalWidth * 0.89),
```
jacob314: comment on these very magic #s.

packages/cli/src/ui/keyMatchers.test.ts
```diff
@@ -253,6 +255,11 @@ describe('keyMatchers', () => {
       positive: [createKey('tab'), createKey('tab', { ctrl: true })],
       negative: [createKey('return'), createKey('space')],
     },
+    {
+      command: Command.TOGGLE_SHELL_INPUT_FOCUS,
```
jacob314, line 259: Fyi @miguelsolorio would be good to get your thoughts on what keyboard shortcut we should use to toggle focus between components in the main GEMINI CLI UI. right now it is just the shell input vs regular input but in the future I could see toggling focus back to review the list of historic messages in the UI.

Other possible followup is automatically moving the focus to the shell when it is active and moving it back.

miguelsolorio: I like the idea of moving focus, but I'd also assume using things like `Tab` allow you to cycle through the options?

packages/core/src/core/coreToolScheduler.ts
galdawave: Added a todo, i'll follow up on this i think the correct implementation would require some refactors.

jacob314 (approved):
lgtm

This is the most epic UI feature we have landed so far. No other tools support this yet.

Issue #6528, opened 2025-08-19
Merge NonInteractive and Interactive mode code paths


anj-s:
### What happened?

We have a large number of issues that show up because of these diverging codebases.

### What did you expect to happen?

We should not be have different behavior across these two types of runs.

### Client information

<details>

```console
$ gemini /about
# paste output here
```

</details>

scidomino:
I'm going to close this out having consolidated the Tool Processing logic. What's left is pretty legitimately different so I don't think any more useful consolidation is possible.

PR #6697, opened 2025-08-20
Remove unnecessary FileErrorType.

scidomino:
## TLDR

Replace usages of FileErrorType with ToolErrorType to avoid gross text-parsing to figure out which error to return.

## Reviewer Test Plan

I ran preflight.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | x  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Contributes to https://github.com/google-gemini/gemini-cli/issues/6528

### DIFF
--- a/packages/core/src/utils/fileUtils.ts
+++ b/packages/core/src/utils/fileUtils.ts
@@ -9,6 +9,7 @@ import path from 'node:path';
 import { PartUnion } from '@google/genai';
 import mime from 'mime-types';
 import { FileSystemService } from '../services/fileSystemService.js';
+import { ToolErrorType } from '../tools/tool-error.js';
 
 // Constants for text file processing
 const DEFAULT_MAX_LINES_TEXT_FILE = 2000;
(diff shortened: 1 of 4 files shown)


silviojr (approved):
Very good. thanks.

packages/core/src/utils/fileUtils.ts
```diff
@@ -232,33 +226,32 @@ export async function processSingleFileContent(
...
-
-    if (fileSizeInBytes > maxFileSize) {
-      throw new Error(
-        `File size exceeds the 20MB limit: ${filePath} (${(
-          fileSizeInBytes /
-          (1024 * 1024)
-        ).toFixed(2)}MB)`,
-      );
+    const fileSizeInMB = stats.size / (1024 * 1024);
+    if (fileSizeInMB > 20) {
```
anj-s, line 248: If we do have this limit, how are planning to allowing users to read large files?

scidomino: It's an excellent question but I have no idea. I don't even know if we *want* to allow users to read large files. I just refactored this existing code to make it more readable without modifying it's existing behavior.

anj-s: can you open an issue about this? good to know at least :)

anj-s (approved):

packages/core/src/utils/fileUtils.ts
scidomino: Done: https://github.com/google-gemini/gemini-cli/issues/6700

PR #6699, opened 2025-08-20
feat(core): Annotate remaining error paths in tools with type.

### LINKED ISSUE #5679: Tool calls errors are not reported correctly.
Description of the bug:
Most of our tools gracefully handle errors in try catches, and our tool executors only log failures when errors are thrown. For example, when a replace call fails with "Failed to edit, could not find the string to replace", that counts as a success in the logs.

Gemini CLI wrote a helpful analysis doc exploring this issue: https://docs.google.com/document/d/1y-CDvFVfr4vQadweVabFgAMhQNFzI-fEUfzruC-Na4w/edit?tab=t.0

Actual vs expected behavior:
No response

Any other information you'd like to share?
No response

joshualitt:
## TLDR

Resolves #5679 .

## Reviewer Test Plan

run tests

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

- Resolves #5679

### DIFF
--- a/packages/core/src/tools/mcp-tool.ts
+++ b/packages/core/src/tools/mcp-tool.ts
@@ -16,6 +16,7 @@ import {
   ToolResult,
 } from './tools.js';
 import { CallableTool, FunctionCall, Part } from '@google/genai';
+import { ToolErrorType } from './tool-error.js';
 
 type ToolParams = Record<string, unknown>;
 
@@ -139,9 +140,19 @@ class DiscoveredMCPToolInvocation extends BaseToolInvocation<
 
     // Ensure the response is not an error
     if (this.isMCPToolError(rawResponseParts)) {
-      throw new Error(
-        `MCP tool '${this.serverToolName}' reported tool error with response: ${JSON.stringify(rawResponseParts)}`,
-      );
+      const errorMessage = `MCP tool '${
+        this.serverToolName
+      }' reported tool error with response: ${JSON.stringify(
+        rawResponseParts,
+      )}`;
+      return {
+        llmContent: errorMessage,
+        returnDisplay: `Error: MCP tool '${this.serverToolName}' reported an error.`,
+        error: {
+          message: errorMessage,
+          type: ToolErrorType.MCP_TOOL_ERROR,
+        },
+      };
     }
 
     const transformedParts = transformMcpContentToParts(rawResponseParts);
--- a/packages/core/src/tools/read-many-files.ts
+++ b/packages/core/src/tools/read-many-files.ts
@@ -29,6 +29,7 @@ import {
   recordFileOperationMetric,
   FileOperation,
 } from '../telemetry/metrics.js';
+import { ToolErrorType } from './tool-error.js';
 
 /**
  * Parameters for the ReadManyFilesTool.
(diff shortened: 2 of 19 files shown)


packages/core/src/tools/mcp-tool.ts
```diff
@@ -139,9 +140,19 @@ class DiscoveredMCPToolInvocation extends BaseToolInvocation<
...
+        rawResponseParts,
+      )}`;
+      return {
+        llmContent: errorMessage,
+        returnDisplay: `Error: MCP tool '${this.serverToolName}' reported an error.`,
+        error: {
+          message: errorMessage,
+          type: ToolErrorType.MCP_TOOL_ERROR,
+        },
+      };
```
joshualitt, line 155: This change ensures all tool errors are handled consistently.

However, I think a natural follow-up to this work would be to switch from forcing error paths to propagate the error type, to using structured tool errors that are thrown at error sites, and caught at the point of creating a tool result.  The current approach creates a lot of noise in tool files, as any function the tool may wish to invoke needs to track error information manually.

packages/core/src/tools/read-many-files.ts
```diff
@@ -233,13 +234,6 @@ ${finalExclusionPatternsForDescription
       : [...exclude];
 
     const searchPatterns = [...inputPatterns, ...include];
-    if (searchPatterns.length === 0) {
-      return {
-        llmContent: 'No search paths or include patterns provided.',
-        returnDisplay: `## Information\n\nNo search paths or include patterns were specified. Nothing to read or concatenate.`,
-      };
-    }
-
```
joshualitt, line 241: This is now checked by the `minLength` requirement on the schema.

packages/core/src/tools/web-fetch.ts
```diff
@@ -73,12 +74,6 @@ class WebFetchToolInvocation extends BaseToolInvocation<
 
   private async executeFallback(signal: AbortSignal): Promise<ToolResult> {
     const urls = extractUrls(this.params.prompt);
-    if (urls.length === 0) {
-      return {
-        llmContent: 'Error: No URL found in the prompt for fallback.',
-        returnDisplay: 'Error: No URL found in the prompt for fallback.',
-      };
-    }
```
joshualitt, line 81: This is now tracked by validation.

packages/core/src/tools/read-many-files.ts
jacob314: Out of paranoia, what message do we get for a schema failure. I would worry this message was clearer to the LLM. Probably worth running evals to ensure this doesn't regress if you haven't already.

jacob314 (approved):
lgtm

packages/core/src/tools/read-many-files.ts
joshualitt: This CL didn't actually change this behavior, I think it changed whenever we started validating the read-many-files tool schema.

Luckily, this CL does add a test! But it's not apples to apples. This check when it ran combined a check for empty paths with a check for empty includes. However, the schema is much more restrictive:
```
        paths: {
          type: 'array',
          items: {
            type: 'string',
            minLength: 1,
          },
          minItems: 1,
          description:
            "Required. An array of glob patterns or paths relative to the tool's target directory. Examples: ['src/**/*.ts'], ['README.md', 'docs/']",
        },
```
If paths is empty, we see `params/paths must NOT have fewer than 1 items` per the test. TBH, I think this error is reasonably clear, at least as clear as the other schema validation errors. That said, I'm not sure this is less clear then the existing error message, because the old error message combines two checks(non empty paths and includes), whereas the schema verifies just paths are non-empty.  The old error message is perhaps a bit more detailed.

joshualitt: Okay, ran 100 evals, and the pass rate was as expected. Given this, and the scope of the CL, I'm going to land this CL.

Issue #5993, opened 2025-08-11
Command: `gemini extensions update`


NTaylorMullen:
This task involves implementing the `gemini extensions update [name]` command. This command will be responsible for updating user-installed extensions to their latest available versions from their original source.

### Acceptance Criteria

1.  **Command Implementation:**
    *   Create a new command: `gemini extensions update [name]`.
    *   The `<name>` argument is optional. If provided, it updates a single, specified extension.
    *   If `<name>` is omitted, the command should attempt to update all extensions installed in the user scope.

2.  **Update Logic:**
    *   The command must only target extensions installed in the **user scope** (`~/.gemini/extensions`). It should not affect system-level extensions.
    *   The update mechanism must be based on the extension's original installation source. For extensions installed from a Git repository, this means performing a `git pull` on the latest version of the default branch.
    *   The system will need a way to track the original source of each user-installed extension (e.g., storing the Git URL in a metadata file within the extension's directory).
    *   If an extension was installed from a local path, it cannot be updated automatically. The command should notify the user about this.

3.  **User Feedback:**
    *   The command should provide clear output to the user, indicating which extensions were checked, which were updated, and which were already up-to-date or could not be updated.

**Parent Issue:** #5987

evanotero:
For the update logic, we should handle the edge case where a user deletes the metadata file:

   * The `gemini extensions update` command's logic will be to look for the `.gemini-extension-install.json` file in each extension's directory.
   * If the file is not found: The update command will not fail. Instead, it will:
       1. Gracefully skip the update for that specific extension.
       2. Print a clear warning message to the user, for example: `Warning: Cannot update 'extension-name' because its installation metadata is missing. Please 
          reinstall it to enable automatic updates.`

PR #6703, opened 2025-08-21
[extensions] Add extension management install command

chrstnb:
## TLDR

Introduces the `gemini extensions install` command, which allows users to install extensions from a Git repository or a local path for user-level extensions only.

Extensions will be enabled by default if installed. To disable them, I will add a list of disabled extensions to the settings.

- Installation from a Git URL.
- Installation from a local directory path.

## Dive Deeper

This just focuses on user installation-- we can add system-wide settings and installation later.

For git repo installation, we will write the repo to a temp directory and perform validation, then, if valid, copy it over to the users top-level extensions folder.

We are storing metadata about the installation source in `.gemini-extension-install.json`. This will allow us to do updates later.  

**One thing that feels wonky to me is the possibility for the directory names and the extension names to get out of sync.**  Currently we are matching on both.

## Reviewer Test Plan

`npx https://github.com/google-gemini/gemini-cli#cb/extensionsinstall extensions install github-repo-url`

`npx https://github.com/google-gemini/gemini-cli#cb/extensionsinstall extensions install path/to/stuff`

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes #5990

Co-authored-by: @evanotero


packages/cli/src/commands/extensions.tsx
```diff
@@ -0,0 +1,58 @@
...
+      .positional('source', {
+        describe: 'The git URL of the extension to install.',
+        type: 'string',
+      })
+      .option('path', {
+        describe: 'Path to a local extension directory.',
+        type: 'string',
+      })
+      .option('scope', {
+        describe: 'The scope to activate the extension in.',
```
NTaylorMullen: What do you think about installed extensions being activated by default?

packages/cli/src/commands/extensions/install.ts
```diff
@@ -0,0 +1,130 @@
...
+  const { scope = SettingScope.User } = args;
+
+  if (scope !== SettingScope.User && scope !== SettingScope.Workspace) {
+    console.error(
+      'Only user and project scopes are supported for installation.',
+    );
+    process.exit(1);
+  }
+
+  const extensionsDir = path.join(os.homedir(), '.gemini', 'extensions');
```
NTaylorMullen: You'll want to lean more on storage.ts. STorage can be the central place where you manage / access this extensions dir

packages/cli/src/commands/extensions/install.ts
```diff
@@ -0,0 +1,130 @@
...
+    // This case should be prevented by yargs configuration
+    console.error('Either a git URL source or a --path must be provided.');
+    process.exit(1);
+  }
+
+  // Verify gemini-extension.json
+  const manifestPath = path.join(targetPath, 'gemini-extension.json');
+  if (!(await fileExists(manifestPath))) {
+    console.error(
+      'Installation failed: gemini-extension.json not found in the extension.',
```
NTaylorMullen: There's a lot of magic strings in this file (like `gemini-extension.json` and `--path` etc.). Should update to rely on a centralized definition of those in case we want to change

packages/cli/src/commands/extensions/install.ts
```diff
@@ -0,0 +1,130 @@
...
+      process.exit(1);
+    }
+  } else {
+    // This case should be prevented by yargs configuration
+    console.error('Either a git URL source or a --path must be provided.');
+    process.exit(1);
+  }
+
+  // Verify gemini-extension.json
+  const manifestPath = path.join(targetPath, 'gemini-extension.json');
```
NTaylorMullen: Might make sense to have an `ExtensionsStorage` objectg similar to Storage

packages/cli/src/commands/extensions/install.ts
```diff
@@ -0,0 +1,130 @@
...
+  if (!(await fileExists(manifestPath))) {
+    console.error(
+      'Installation failed: gemini-extension.json not found in the extension.',
+    );
+    // Clean up installed directory
+    await fs.rm(targetPath, { recursive: true, force: true });
+    process.exit(1);
+  }
+
+  const settings = loadSettings(process.cwd());
```
NTaylorMullen: the config object should have a target dir instead of leaning on process.cwd() here

packages/cli/src/config/config.ts
```diff
@@ -229,6 +231,11 @@ export async function parseArguments(): Promise<CliArgs> {
             // Handle comma-separated values
             dirs.flatMap((dir) => dir.split(',').map((d) => d.trim())),
         })
+        .option('extension-management', {
```
NTaylorMullen: I wouldn't add a command line. We can have a hidden settings.json entry until we're ready

NTaylorMullen:
lets also get @bbiggs's take here too

packages/cli/src/config/config.ts
```diff
@@ -245,7 +252,13 @@ export async function parseArguments(): Promise<CliArgs> {
         }),
     )
     // Register MCP subcommands
-    .command(mcpCommand)
+    .command(mcpCommand);
+
+  if (settings.extensionManagement) {
+    yargsInstance.command(extensionsCommand);
```
evanotero, line 259: Do we also need to mirror the mcp implementation on Line 274 and have all subcommand modules handle their own execution?

packages/cli/src/commands/extensions.tsx
```diff
@@ -0,0 +1,58 @@
...
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { CommandModule } from 'yargs';
+import { handleInstall } from './extensions/install.js';
+import { SettingScope } from '../config/settings.js';
+
+const installCommand: CommandModule = {
```
evanotero: nit: the mcp implementation has the subcommand modules defined in their own files. For example, "installCommand" would be defined in "install.ts"

packages/cli/src/commands/extensions/install.ts
evanotero: nit: there's already a constant EXTENSIONS_DIRECTORY_NAME in config/extension.ts

packages/cli/src/commands/extensions/install.ts
evanotero: gemini-extension.json can be found in config/extension.ts along with other helper functions that may be relevant here. See #6163 for examples.

packages/cli/src/commands/extensions/install.ts
```diff
@@ -0,0 +1,130 @@
...
+    return stat.isFile();
+  } catch (e) {
+    if ((e as NodeJS.ErrnoException).code === 'ENOENT') {
+      return false;
+    }
+    throw e;
+  }
+}
+
+export async function handleInstall(args: InstallArgs) {
```
evanotero, line 18: We will need to store metadata on the source in order to handle updating in the future, per #5993

See https://github.com/evanotero/gemini-cli/blob/bf49f27a356ee488dd85d20a4ecf9727f8d2241b/packages/cli/src/commands/extensions/install.ts#L162 for how I did this in my draft pr.

packages/cli/src/commands/extensions/install.ts
evanotero: We can check for the presence of an extensions manifest file as well as for proper formatting by using `loadExtension`. See https://github.com/evanotero/gemini-cli/blob/bf49f27a356ee488dd85d20a4ecf9727f8d2241b/packages/cli/src/commands/extensions/install.ts#L122

packages/cli/src/commands/extensions/install.ts
```diff
@@ -0,0 +1,130 @@
...
+    }
+  } else if (args.path) {
+    // Local path
+    const sourcePath = path.resolve(args.path);
+    extensionName = path.basename(sourcePath);
+    targetPath = path.join(extensionsDir, extensionName);
+
+    if (await directoryExists(targetPath)) {
+      console.error(
+        `Extension "${extensionName}" already exists. Please uninstall it first.`,
```
evanotero: It is possible we change from always requiring the directory name to match the extension name. A more future-proof way to check for uniqueness is to compare ExtensionConfig.name. See https://github.com/evanotero/gemini-cli/blob/bf49f27a356ee488dd85d20a4ecf9727f8d2241b/packages/cli/src/commands/extensions/install.ts#L140 for an example

packages/cli/src/commands/extensions/install.ts
evanotero: For the one cloned from git, you can create a temp directory before copying it over to the .gemini/extensions folder to validate the configuration: https://github.com/evanotero/gemini-cli/blob/bf49f27a356ee488dd85d20a4ecf9727f8d2241b/packages/cli/src/commands/extensions/install.ts#L113

packages/cli/src/commands/extensions/install.ts
```diff
@@ -0,0 +1,130 @@
...
+      );
+      process.exit(1);
+    }
+
+    try {
+      execSync(`git clone --depth 1 ${args.source} ${targetPath}`, {
```
evanotero: nit: you can use `await simpleGit().clone(gitUrl, destination, ['--depth', '1'])`

packages/cli/src/commands/extensions/install.ts
chrstnb: Not sure if I totally addressed but we're using this approach elsewhere as well. Can clean up as a follow-up?

chrstnb:
Ended up pulling in a lot of @evanotero's implementation!

packages/cli/src/commands/extensions/install.ts
```diff
@@ -0,0 +1,60 @@
...
+        type: 'string',
+      })
+      .option('path', {
+        describe: 'Path to a local extension directory.',
+        type: 'string',
+      })
+      .conflicts('source', 'path')
+      .check((argv) => {
+        if (!argv.source && !argv.path) {
+          throw new Error('Either a git URL or a --path must be provided.');
```
cornmander: It's weird that one is positional and one is a flag. Can both be flags and we check that exactly one is set?

cornmander: This is worth testing.

packages/cli/src/commands/extensions.tsx
```diff
@@ -0,0 +1,22 @@
...
+import { CommandModule } from 'yargs';
+import { installCommand } from './extensions/install.js';
+
+export const extensionsCommand: CommandModule = {
+  command: 'extensions <command>',
+  describe: 'Manage Gemini CLI extensions.',
+  builder: (yargs) =>
+    yargs
+      .command(installCommand)
+      .demandCommand(1, 'You need at least one command before continuing.')
```
cornmander, line 16: Maybe "You need at least one command to continue."?

packages/cli/src/config/extension.ts
```diff
@@ -171,3 +242,98 @@ export function annotateActiveExtensions(
...
+ * Clones a Git repository to a specified local path.
+ * @param gitUrl The Git URL to clone.
+ * @param destination The destination path to clone the repository to.
+ */
+async function cloneFromGit(
+  gitUrl: string,
+  destination: string,
+): Promise<void> {
+  try {
+    await simpleGit().clone(gitUrl, destination, ['--depth', '1']);
```
cornmander, line 257: This probably doesn't matter for now, but if you copy the contents of the cloned repo you'll include the .git metadata. If you do this for many extensions it's wasteful. You should look into using git archive to export just the code in the repo.

cornmander (approved):

packages/cli/src/config/extension.ts
chrstnb: added a todo, ty!

packages/cli/src/commands/extensions/install.ts
```diff
@@ -0,0 +1,62 @@
...
+      `Extension "${extensionName}" installed successfully and enabled.`,
+    );
+  } catch (error) {
+    console.error((error as Error).message);
+    process.exit(1);
+  }
+}
+
+export const installCommand: CommandModule = {
+  command: 'install [source]',
```
evanotero: Since there's flags, this should be `install [source] [args...]`

packages/cli/src/commands/extensions/install.ts
```diff
@@ -36,7 +36,7 @@ export const installCommand: CommandModule = {
   describe: 'Installs an extension from a git repository or a local path.',
```
cornmander, line 36: The command docstring should probably be [--source <source> | --path <path>]

packages/cli/src/config/extension.ts
```diff
@@ -171,3 +242,99 @@ export function annotateActiveExtensions(
...
+  destination: string,
+): Promise<void> {
+  await fs.promises.cp(source, destination, { recursive: true });
+}
+
+export async function installExtension(
+  installMetadata: ExtensionInstallMetadata,
+): Promise<string> {
+  const extensionsDir = ExtensionStorage.getUserExtensionsDir();
+  await fs.promises.mkdir(extensionsDir, { recursive: true });
```
evanotero, line 281: Probably a silly question, but does mkdir fail if a directory already exists?

evanotero (approved):
Super clean!

packages/cli/src/commands/extensions/install.ts
evanotero: Or at least that's that the `gemini mcp add` command did

packages/cli/src/config/extension.ts
chrstnb: "Calling fsPromises.mkdir() when path is a directory that exists results in a rejection only when recursive is false." so yes but not here because we're setting 'recursive'!

packages/cli/src/commands/extensions/install.ts
chrstnb: @evanotero as well-- this made it to the merge queue so going to let it go through and will update the command docstring in the following PR

PR #6713, opened 2025-08-21
Force restart on trust level change to reload settings

shrutip90:
## TLDR

Updated settings to ignore workspace settings in untrusted environments and added logic to force restart when trust level changes. Since trust is expected to be the default, we initialize with trusted settings when the trust level is undefined and only force restart for untrusted case. If it's already defined, we force restart when the trusted status flips.

## Dive Deeper

image

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

### DIFF
--- a/packages/cli/src/config/settings.ts
+++ b/packages/cli/src/config/settings.ts
@@ -16,6 +16,7 @@ import {
 import stripJsonComments from 'strip-json-comments';
 import { DefaultLight } from '../ui/themes/default-light.js';
 import { DefaultDark } from '../ui/themes/default.js';
+import { isWorkspaceTrusted } from './trustedFolders.js';
 import { Settings, MemoryImportFormat } from './settingsSchema.js';
 
 export type { Settings, MemoryImportFormat };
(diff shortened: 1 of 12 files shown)


packages/cli/src/config/settings.ts
```diff
@@ -73,34 +74,41 @@ function mergeSettings(
   system: Settings,
   user: Settings,
   workspace: Settings,
+  isTrusted?: boolean,
```
jacob314: nit: why not make isTrusted required? Seems important not to accidentally miss passing it and have it default to true.

packages/cli/src/config/settings.ts
```diff
@@ -111,18 +119,21 @@ export class LoadedSettings {
...
     this.errors = errors;
+    this.isTrusted = isTrusted;
     this._merged = this.computeMergedSettings();
   }
 
   readonly system: SettingsFile;
   readonly user: SettingsFile;
   readonly workspace: SettingsFile;
   readonly errors: SettingsError[];
+  readonly isTrusted: boolean | undefined;
```
jacob314: nit: make this bool instead of `boolean | undefined`
only allow bools to be undefined when strictly required. this seems like a true boolean.

packages/cli/src/ui/components/FolderTrustDialog.test.tsx
```diff
@@ -21,16 +21,85 @@ describe('FolderTrustDialog', () => {
...
+        onSelect={onSelect}
+        onRestartRequest={vi.fn()}
+        isRestarting={true}
+      />,
+    );
+
+    stdin.write('\x1b'); // escape key
+
+    // Give it a moment to process, then assert that onSelect was NOT called.
+    await new Promise((resolve) => setTimeout(resolve, 100));
```
jacob314: don't timeout. use waitFor. same elsewhere in file.

packages/cli/src/ui/App.tsx
```diff
@@ -991,7 +989,11 @@ const App = ({ config, settings, startupWarnings = [], version }: AppProps) => {
               onComplete={handleIdePromptComplete}
             />
           ) : isFolderTrustDialogOpen ? (
-            <FolderTrustDialog onSelect={handleFolderTrustSelect} />
+            <FolderTrustDialog
+              onSelect={handleFolderTrustSelect}
+              onRestartRequest={() => process.exit(0)}
```
jacob314: no need to make this a property. just have FolderTrustDialog call onRestartRequest directly. You can mock for tests if needed.

jacob314 (approved):
Thanks for the careful testing!

packages/cli/src/config/settings.ts
shrutip90: Sure, that should be fine too, done.

packages/cli/src/config/settings.ts
shrutip90: Done

packages/cli/src/ui/components/FolderTrustDialog.test.tsx
shrutip90: Done

packages/cli/src/ui/App.tsx
shrutip90: Guessing you wanted FolderTrustDialog to call process.exit directly, since this was the definition for  onRestartRequest. Replaced it.

packages/cli/src/config/settings.ts
```diff
@@ -73,34 +74,41 @@ function mergeSettings(
...
 ): Settings {
   // folderTrust is not supported at workspace level.
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const { folderTrust, ...workspaceWithoutFolderTrust } = workspace;
 
+  const isSafeWorkspace = isTrusted ?? true;
+  const safeWorkspace = isSafeWorkspace ? workspace : ({} as Settings);
+  const safeWorkspaceWithoutFolderTrust = isSafeWorkspace
+    ? workspaceWithoutFolderTrust
+    : ({} as Settings);
```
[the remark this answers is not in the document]
shrutip90: Done

PR #6714, opened 2025-08-21
Reuse CoreToolScheduler for nonInteractiveToolExecutor

scidomino:
## TLDR

Reuse CoreToolScheduler in nonInteractiveToolExecutor. This makes the interactive and noninteractive code paths more similar.

## Reviewer Test Plan

Ran non interactive mode with:

```
npm start -- --prompt "This is a test of your ability to make multiple tool calls at once. Please make two separate requests to read two different files using the read_file tool (DO NOT use read_many_files). Tell me which files your read and what you learned from their contents"
```

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

### DIFF
--- a/packages/core/src/core/nonInteractiveToolExecutor.ts
+++ b/packages/core/src/core/nonInteractiveToolExecutor.ts
@@ -4,166 +4,27 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import {
-  FileDiff,
-  logToolCall,
-  ToolCallRequestInfo,
-  ToolCallResponseInfo,
-  ToolErrorType,
-  ToolResult,
-} from '../index.js';
-import { DiscoveredMCPTool } from '../tools/mcp-tool.js';
-import { Config } from '../config/config.js';
-import { convertToFunctionResponse } from './coreToolScheduler.js';
-import { ToolCallDecision } from '../telemetry/tool-call-decision.js';
+import { ToolCallRequestInfo, ToolCallResponseInfo, Config } from '../index.js';
+import { CoreToolScheduler } from './coreToolScheduler.js';
 
 /**
- * Executes a single tool call non-interactively.
- * It does not handle confirmations, multiple calls, or live updates.
+ * Executes a single tool call non-interactively by leveraging the CoreToolScheduler.
  */
 export async function executeToolCall(
   config: Config,
   toolCallRequest: ToolCallRequestInfo,
-  abortSignal?: AbortSignal,
+  abortSignal: AbortSignal,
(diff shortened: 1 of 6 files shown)


packages/core/src/core/nonInteractiveToolExecutor.ts
```diff
@@ -4,166 +4,27 @@
...
-            response: { error: error.message },
-          },
-        },
-      ],
-      resultDisplay: error.message,
-      error,
-      errorType: ToolErrorType.UNHANDLED_EXCEPTION,
-    };
-  }
+  return new Promise<ToolCallResponseInfo>((resolve, reject) => {
```
NTaylorMullen, line 18: Wow, this is so much more streamlined. I didn't realize that we had made coreToolScheduler so flexible to support all of this out of the box. Are there specific things that we had in non-Interactive that we didn't have in interactive?

packages/core/src/core/nonInteractiveToolExecutor.ts
```diff
@@ -4,166 +4,27 @@
...
 /**
- * Executes a single tool call non-interactively.
- * It does not handle confirmations, multiple calls, or live updates.
+ * Executes a single tool call non-interactively by leveraging the CoreToolScheduler.
  */
 export async function executeToolCall(
   config: Config,
   toolCallRequest: ToolCallRequestInfo,
-  abortSignal?: AbortSignal,
+  abortSignal: AbortSignal,
```
NTaylorMullen, line 16: 👏

NTaylorMullen (approved):
Love the consolidation. Please run evals on this changeset to understand the impact before merge.

packages/core/src/core/nonInteractiveToolExecutor.ts
scidomino: I couldn't find any. There were slight differences in how they handled errors (which you can see in the tests) but I think the way coreToolScheduler did it is superior. Even logging is handled properly.

anj-s: fantastic change. i love to see code deleted. just goes to show how many more places we could have had issues.

anj-s (approved):

scidomino:
After 200 evals it's passing at 54.42% so I think that's comparable to the standard. Will merge now.

PR #5438, opened 2025-08-02
fix(core): prevent ImportProcessor from duplicating workspace directo…

JeremyFabrikapp:
# fix(core): prevent ImportProcessor from duplicating workspace directory names

- Enhanced validation in memoryImportProcessor to only accept file paths starting with ./ or /
- Excludes workspace syntax like @workspace/package from being treated as file imports
- Fixes ENOENT errors when CLI is run from workspace directories
- Resolves path duplication issue: /path/to/project/project/file -> /path/to/project/file

Fixes issue where yarn workspace references in gemini.md files caused ImportProcessor to incorrectly interpret @workspace/package as file imports, leading to duplicated directory names in resolved paths.

## TLDR

The ImportProcessor was incorrectly treating yarn workspace references (like `@workspace/package`) in gemini.md files as file imports, causing path duplication errors. This fix enhances validation to only accept explicit file paths starting with `./` or `/`, preventing workspace syntax from being processed as imports.

## Dive Deeper

The root cause was in `packages/core/src/utils/memoryImportProcessor.ts` where the `findImports` function used overly permissive validation that accepted any import starting with a letter (`isLetter(importPath[0])`). This meant that yarn workspace references like `@myproject/app` in documentation were interpreted as file imports `myproject/app`.

When the CLI was executed from within a project directory named `myproject`, the path resolution would duplicate the directory name:
- Base path: `/Users/user/projects/myproject/`
- Import path: `myproject/app` (extracted from `@myproject/app`)
- Result: `/Users/user/projects/myproject/myproject/app` ❌

The fix restricts import validation to only accept paths that explicitly start with `./` (relative) or `/` (absolute), which are the intended file import syntaxes. Workspace references like `@workspace/package` are now properly excluded.

## Reviewer Test Plan

1. **Create test environment:**
   ```bash
   mkdir test-workspace && cd test-workspace
   echo "yarn workspace @test-workspace/app dev" > gemini.md
   ```

2. **Before fix:** Run `gemini -p "hello"` - should see path duplication error
3. **After fix:** Same command should work without errors
4. **Validate file imports still work:** Add `@./some-file.md` to gemini.md and verify it's processed correctly

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| -------- | --- | --- | --- |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ✅  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
[...]

PR #6716, opened 2025-08-21
feat(commands): Enable @file processing in TOML commands

### LINKED ISSUE #6651: custom command can't use @ command
### What happened?

custom command can't use @ command

custom command can use @ command to read a  file

### Client information

* **CLI Version:** 0.1.22
* **Git Commit:** 5fe4e023
* **Session ID:** 8ab13bad-5114-43a8-b234-3bb0dd49c602
* **Operating System:** darwin v22.18.0
* **Memory Usage:** 252.3 MB

### Login information

_No response_

_No response_

abhipatel12:
## TLDR

This pull request introduces support for `@{path/to/file}` syntax within the `prompt` of custom `.toml` commands. This feature allows users to create powerful, dynamic commands that can inline file or directory content directly into the prompt as context before it's sent to the model.

The core change involves adding a new `AtFileProcessor` to the prompt processing pipeline for file-based commands. The processor pipeline has been carefully ordered to enable dynamic use cases, such as generating a file path from a shell command.

### New `@{...}` Syntax

- **File Injection**: `@{path/to/file.txt}` is replaced by the content of `file.txt`.
- **Multimodal Support**: If the path points to a supported image (e.g., PNG, JPEG), PDF, audio, or video file, it will be correctly encoded and injected as multimodal input. Other binary files are handled gracefully and skipped.
- **Directory Listing**: `@{path/to/dir}` is traversed and each file present within the directory and all subdirectories are inserted into the prompt. This respects `.gitignore` and `.geminiignore` if enabled.
- **Workspace-Aware**: The command searches for the path in the current directory and any other workspace directories. Absolute paths are allowed if they are within the workspace.
- **Processing Order**: File content injection with `@{...}` is processed _before_ shell commands (`!{...}`) and argument substitution (`{{args}}`).
- **Parsing**: The parser requires the content inside `@{...}` (the path) to have balanced braces (`{` and `}`).

### Processor Pipeline Order

The prompt processor pipeline for `.toml` commands now executes in the following order:

1.  **File Injection (`AtFileProcessor`)**: Scans the prompt for `@{...}` placeholders and injects the corresponding file/directory content.
2.  **Shell & Argument Injection (`ShellProcessor`)**: Replaces `!{...}` and `{{args}}` placeholders.
3.  **Default Argument Appending (`DefaultArgumentProcessor`)**: If `{{args}}` was not used, the raw arguments are appended to the very end of the prompt.

This order means that file injection is completed before shell and argument processing. 

One thing to note: if an injected file contains placeholder they will be processed. Thus, an injected file that contains `!{rm -rf}` would be parsed as such. 

### Error Handling

If a path in an `@{...}` placeholder cannot be found or read:
1.  An error message is displayed in the UI for user feedback (e.g., "Failed to inject file content...").
2.  The original, unprocessed placeholder (e.g., `@{path/to/nonexistent.file}`) is left in the prompt.

This approach empowers the model. Since the model has access to file-system tools, it can see the user's likely intent, reason about a potential typo in the path, and use its own tools to attempt to find the correct file.

## Reviewer Test Plan

To validate these changes, reviewers can test the following scenarios:

1.  **Static Path**: Create a `.toml` command with a hardcoded `@{...}` path.
    ```toml
    # ~/.gemini/commands/static.toml
    prompt = "Summarize this: @{README.md}"
    ```
    **Run**: `/static`
    **Verify**: The model receives the content of `README.md`.

2. **Multi-modal**: Ask it to describe an image.
    ```toml
    # ~/.gemini/commands/what-is-it.toml
    prompt = "Describe what is in this image: @{image.png}"
    ```
    **Run**: `/what-is-it`
    **Verify**: The model receives the content of `image.png` and should tell you what the image contained.

4.  **Invalid Path**: Run a command with a typo in the file path.
    **Run**: `/dynamic @{README.md-typo}`
    -   **Verify**: Check the UI for an error message about failing to inject the file.
    -   **Verify**: Check the model's input to confirm it received the literal string `@{README.md-typo}`.

5.  **Non-Matching Syntax**: Use an `@` symbol in a way that does not match the syntax.
    ```toml
    # ~/.gemini/commands/at-test.toml
    prompt = "Send a message to @the-team about the release of @ version1.0"
    ```
    **Run**: `/at-test`
    -   **Verify**: The model should receive the prompt with the strings `@the-team` and `@ version1.0` intact, as they do not match the `@{...}` syntax.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅ | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ✅ | -   | -   |

## Linked issues / bugs

- Resolves #6651


packages/core/src/utils/memoryImportProcessor.ts
```diff
@@ -367,16 +368,23 @@ export async function processImports(
...
+          newImportState,
+          projectRoot,
+          importFormat,
+          recursive,
+        );
+        result += `<!-- Imported from: ${importPath} -->\n${imported.content}\n<!-- End of import from: ${importPath} -->`;
+        imports.push(imported.importTree);
+      } else {
+        result += `<!-- Imported from: ${importPath} -->\n${fileContent}\n<!-- End of import from: ${importPath} -->`;
+      }
```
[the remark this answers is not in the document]
abhipatel12: Done

jacob314 (commented):
How might someone create a TOML command that needs to use `@` but doesn't want to have it trigger file processing?
For example
`yarn workspace @test-workspace/app dev`

where unfortunately test-workspace/app is a valid path so you can't just skip based on invalid paths.

I'm curious about a solution for this because I'm hoping what you do can also align with what we need to do for ImportProcessor in https://github.com/google-gemini/gemini-cli/pull/5438

abhipatel12:
> How might someone create a TOML command that needs to use `@` but doesn't want to have it trigger file processing? For example `yarn workspace @test-workspace/app dev`
> 
> where unfortunately test-workspace/app is a valid path so you can't just skip based on invalid paths.
> 
> I'm curious about a solution for this because I'm hoping what you do can also align with what we need to do for ImportProcessor in #5438

From our offline convo, changed this to use @{foo.txt} syntax. Let me know what you think!

packages/cli/src/services/prompt-processors/atFileProcessor.test.ts
```diff
@@ -0,0 +1,174 @@
...
+      const prompt = 'Analyze @{path/with/{braces}/file.txt}';
+      const result = await processor.process(prompt, context);
+      expect(mockReadPathFromWorkspace).toHaveBeenCalledWith(
+        'path/with/{braces}/file.txt',
+        mockWorkspaceContext,
+      );
+      expect(result).toBe('Analyze content of path/with/{braces}/file.txt');
+    });
+
+    it('should leave the prompt unmodified if it contains an unclosed trigger', async () => {
```
jacob314: for discussion: should this generate an error or warning of some sort rather than quietly not being parsed?
Agree that `@world`  should just be ignored but `@{world` more likely indicates a typo by the author.

packages/cli/src/services/prompt-processors/atFileProcessor.test.ts
```diff
@@ -0,0 +1,174 @@
...
+    it('should handle placeholders at the beginning, middle, and end', async () => {
+      const processor = new AtFileProcessor();
+      const prompt = '@{start.txt} in the @{middle.txt} and @{end.txt}';
+      const result = await processor.process(prompt, context);
+      expect(result).toBe(
+        'content of start.txt in the content of middle.txt and content of end.txt',
+      );
+    });
+
+    it('should correctly parse paths that contain balanced braces', async () => {
```
jacob314, line 126: what about if I need unbalanced braces in the prompt?
How might I write
`'Analyze @{path/with/braces{example.txt}'`
could I write
`'Analyze @{{{path/with/braces{example.txt}}}'`
?

packages/cli/src/services/prompt-processors/atFileProcessor.ts
```diff
@@ -0,0 +1,148 @@
...
+interface FileInjection {
+  /** The file path extracted from within @{...}, trimmed. */
+  path: string;
+  /** The starting index of the injection (inclusive, points to '@'). */
+  startIndex: number;
+  /** The ending index of the injection (exclusive, points after '}'). */
+  endIndex: number;
+}
+
+export class AtFileProcessor implements IPromptProcessor {
```
jacob314, line 20: can you unify this parsing with parsing for `!{` as I would expect there should be consistent handling of edge cases.

packages/core/src/utils/pathReader.ts
```diff
@@ -0,0 +1,59 @@
...
+  if (!absolutePath) {
+    throw new Error(`Path not found in workspace: ${pathStr}`);
+  }
+
+  const stats = await fs.stat(absolutePath);
+  if (stats.isDirectory()) {
+    const entries = await fs.readdir(absolutePath);
+    return `Directory listing for ${pathStr}:\n- ${entries.join('\n- ')}`;
+  } else {
+    return fs.readFile(absolutePath, 'utf-8');
```
jacob314: I would suggest we align this closely with read-many-files.ts which is what is called when `@` are added to a prompt. If that is not feasible it should be at least aligned with
is there a reason we can't use ../utils/fileUtils.ts

That will keep the behavior consistent and handle cases such as reading binary files robustly. For example, fileUtils.ts has support for images and read-many-files and read-file have support for the correct logic to encode images in the format the Gemini API needs.

Without changes, referencing a binary image causes Gemini CLI to throw exceptions due to blowing out the context window by adding the bytes of the image into the prompt.

jacob314 (commented):
Do we really need to support

```
prompt = "What is the purpose of this file: @{ !{find . -name 'FileCommandLoader.ts'} }"
```
my slight preference would be to omit that for now and keep the 
`@{}` functionality simpler
We can always support more complex use cases in the future but it is harder to remove functionality once we have added it.
A user who really wants

`@{ !{find . -name 'FileCommandLoader.ts'} }` could accomplish it today by piping the results of the find into `cat`.

Counter point to this suggestion would be if 
`@{}` was able to support loading image files correctly into the prompt at which point it  would support something not currently feasible with shell scripts although I suppose the shell scripts could perhaps be tweaked to support scripts that output binary data.

packages/cli/src/services/prompt-processors/atFileProcessor.test.ts
abhipatel12: I extracted the parsing logic to separate utility and added some mechanisms to let users know. Let me know what you think!

packages/cli/src/services/prompt-processors/atFileProcessor.test.ts
abhipatel12: Made some changes to the parsing logic here. I'm not quite if it's worth supporting unbalanced braces if theres a brace within a file name. 

But willing to change my mind!

packages/cli/src/services/prompt-processors/atFileProcessor.ts
abhipatel12: Done!

packages/core/src/utils/pathReader.ts
abhipatel12: To make the `@` processing support multi-modal input, i had to make a lot of changes ... but it works! Let me know what you think!

abhipatel12:
Thanks for the feedback @jacob314! Sorry about the big diff, but we have multi-modal support now!

packages/cli/src/services/prompt-processors/argumentProcessor.ts
```diff
@@ -14,9 +14,45 @@ import type { CommandContext } from '../../ui/commands/types.js';
...
+          {
+            text: `
+
+${context.invocation.raw}`,
+          },
+        ];
+      }
+
+      const lastPart = prompt[prompt.length - 1];
+      const newPrompt = [...prompt]; // Create a mutable copy
```
jacob314: nit: remove zero value comment.

packages/cli/src/services/prompt-processors/argumentProcessor.ts
```diff
@@ -14,9 +14,45 @@ import type { CommandContext } from '../../ui/commands/types.js';
...
+        return [
+          {
+            text: `
+
+${context.invocation.raw}`,
+          },
+        ];
+      }
+
+      const lastPart = prompt[prompt.length - 1];
```
jacob314: nit: use `.at(-1)` instead of `prompt.length - 1`

packages/cli/src/services/prompt-processors/argumentProcessor.ts
```diff
@@ -14,9 +14,45 @@ import type { CommandContext } from '../../ui/commands/types.js';
...
 export class DefaultArgumentProcessor implements IPromptProcessor {
-  async process(prompt: string, context: CommandContext): Promise<string> {
-    if (context.invocation!.args) {
-      return `${prompt}\n\n${context.invocation!.raw}`;
+  async process(
+    prompt: PromptPipelineContent,
+    context: CommandContext,
+  ): Promise<PromptPipelineContent> {
+    if (context.invocation?.args) {
+      if (prompt.length === 0) {
```
jacob314: why are we still adding two trailing line breaks when the prompt is empty?

packages/cli/src/services/prompt-processors/argumentProcessor.ts
jacob314: optional nit: can you ask gemini CLI to linearize this flow? I think this is just a complexified version of adding to an existing text part if one exists, adding a new one if it doesn't. that would be a reasonable helper method to add to a util file.

packages/cli/src/services/prompt-processors/atFileProcessor.test.ts
jacob314: fine to omit for now.

packages/core/src/utils/partUtils.test.ts
```diff
@@ -162,5 +166,97 @@ describe('partUtils', () => {
...
+        data: undefined,
+        functionCalls: undefined,
+        executableCode: undefined,
+        codeExecutionResult: undefined,
+      };
+      expect(getResponseText(response)).toBeNull();
+    });
+  });
+
+  // ⬇️ New test suite for flatMapTextParts
```
jacob314: remove this comment that just indicates that new stuff was added.

packages/core/src/utils/pathReader.test.ts
```diff
@@ -0,0 +1,410 @@
...
+import mock from 'mock-fs';
+import * as path from 'node:path';
+import { WorkspaceContext } from './workspaceContext.js';
+import { readPathFromWorkspace } from './pathReader.js';
+import type { Config } from '../config/config.js';
+import { StandardFileSystemService } from '../services/fileSystemService.js';
+import type { FileDiscoveryService } from '../services/fileDiscoveryService.js';
+
+// Helper to conditionally run tests (useful for OS-specific behaviors like permissions)
+const itif = (condition: boolean) => (condition ? it : it.skip);
```
jacob314: instead of adding our own helper, use it.skipIf from vitest

packages/cli/src/services/prompt-processors/atFileProcessor.test.ts
jacob314: looks good

jacob314 (approved):

packages/cli/src/services/prompt-processors/argumentProcessor.ts
abhipatel12: We shouldnt. Fixed!

packages/cli/src/services/prompt-processors/argumentProcessor.ts
abhipatel12: Done

abhipatel12: Done. Moved to a util file that deals with `Part`.

packages/cli/src/services/prompt-processors/argumentProcessor.ts
abhipatel12: Done

packages/core/src/utils/partUtils.test.ts
abhipatel12: Done

packages/core/src/utils/pathReader.test.ts
abhipatel12: Done

abhipatel12 (commented):
Thanks for the review!

PR #6721, opened 2025-08-21
feat(sessions): Integrate chat recording into GeminiChat

bl-ue:
This PR is the **second** in a series of **9 PRs** that implement automatic chat recording and resuming in both interactive and non-interactive modes, with command line flags for non-interactive resuming and session management, and an interactive session browser.

PR #4401 was the original monolithic PR, which implemented this feature from the ground up.  It's too large to merge as-is, however, and so it was broken up into 9 smaller PRs, of which this PR is the second.

Please see #4401's description for the full context of this PR.  The original monolithic PR's description contains a detailed analysis of the entire system, with an example of the recording JSON, 30+ linked issues, and a demo video.


bl-ue:
Hi @NTaylorMullen and @jacob314, this is PR tested and ready to merge.

packages/cli/src/nonInteractiveCli.ts
```diff
@@ -79,16 +88,50 @@ export async function runNonInteractive(
             id: toolCallRequest.callId,
           };
           functionCalls.push(fc);
+        } else if (event.type === GeminiEventType.Finished) {
+          chatRecordingService.recordMessageTokens({
```
NTaylorMullen: Should this recordMessageTokens just take the usageMetadata actually and pull out all these values?

packages/cli/src/nonInteractiveCli.ts
NTaylorMullen, line 1: Why integrate the recording here vs. in https://github.com/google-gemini/gemini-cli/blob/main/packages/core/src/core/geminiChat.ts ?

bl-ue: @NTaylorMullen, good point.  I probably would've done so had I thought of it.  It would mostly unify non-interactive and interactive recording implementations in `geminiChat.ts` .  We'd still need to hook into the chat recording flow from the UI to add system/info messages like "chat was compressed," and various tool call updates, such as the `[Operation Cancelled] Reason: User did not allow tool call` and similar messages that are sent to Gemini, as those are handled outside of Gemini chat.

It might take a day or two to implement.  Should I proceed?

NTaylorMullen: Ya lets definitely proceed with the geminiChat.ts. @scidomino is actively working on unifying the interactive/non-interactive stuff as well so if you're able to integrate at the geminiChat layer that will have the highest chance for minimal conflicts + ensure that no matter where these things are used that they are effective and consistent.

packages/core/src/core/client.ts
```diff
@@ -213,6 +214,10 @@ export class GeminiClient {
     this.chat = await this.startChat();
   }
 
+  getChatRecordingService(): ChatRecordingService | undefined {
+    return this.chat?.getChatRecordingService();
```
NTaylorMullen, line 227: Can we make this throw if chat is undefined? That way we can return the recording service in a non-undefined fashion

packages/core/src/core/coreToolScheduler.ts
```diff
@@ -947,12 +947,47 @@ export class CoreToolScheduler {
...
+    const allCallsAreTerminal = this.toolCalls.every(isToolCallTerminal);
+
+    // Record all tool calls in their final state
+    const chatRecordingService = this.config
+      .getGeminiClient()
+      ?.getChatRecordingService();
+    if (chatRecordingService) {
+      const toolRegistry = await this.toolRegistry;
+
+      chatRecordingService.recordToolCalls(
```
NTaylorMullen: Hmmm, is there a route where more of the tool call recording happens in geminiChat? I imagine you need to have this bit here currently because it's effectively recording the "display/description" portions right?

NTaylorMullen (approved):

packages/core/src/core/coreToolScheduler.ts
bl-ue: Hi @NTaylorMullen, yes, you're right.  We need to record that extra information to support resuming chats.  Also, recording in `CoreToolScheduler` enables us to record the status of each tool call as it changes.  For example, when the model calls a tool, we immediately record the tool call with a status of `awaiting`, and then update it to `executing` and `succeeded` as it changes.  A side effect of this would be to potentially enable other Gemini CLI clients to display tool call state changes in real time.

However, if `GeminiChat` had access to `ToolRegistry`, then we could theoretically handle recording that information in `GeminiChat` instead of `CoreToolScheduler`.  Should I investigate this approach?

packages/core/src/core/client.ts
bl-ue: Done.

packages/core/src/core/coreToolScheduler.ts
NTaylorMullen: does Gemini chat need access to the tool registery or does the recording service need access to it?

bl-ue: Great point!  `GeminiChat` doesn't need it, only the recording service does.  That makes things simpler.  Should I proceed?

NTaylorMullen: Yes absolutely!

bl-ue: Hi @NTaylorMullen, migrated tool call recording to `GeminiChat`.

NTaylorMullen (approved):

bl-ue:
Next: #7662

PR #6723, opened 2025-08-21
fix(copyCommand): provide friendlier error messages for `/copy` command

### LINKED ISSUE #6400: Improve `/copy` error message when platform-specific clipboard tools are missing
### What happened?

When using the /copy command, the error message only shows "Failed to copy to the clipboard." without providing further context.
Unless **debug mode** is enabled, the user does not get any hint that the failure may be due to missing dependencies such as `xclip` or `xsel`.

Image

Currently:
1. No suggestion is given to install the required dependencies.
2. The documentation section for `/copy` does not mention the need for these tools.

1. If the failure is caused by missing `xclip` or `xsel`, the error message should explicitly suggest installing them.
2. The documentation should mention the required tools for different platforms.

### Client information

* **CLI Version:** 0.1.21
* **Git Commit:** 055fe46d
* **Operating System:** linux v20.19.3
* **Model Version:** gemini-2.5-flash
* **Memory Usage:** 182.8 MB

### Login information

_No response_

_No response_

HunDun0Ben:
## TLDR

Improve `/copy` command error messages and provide clearer guidance when platform-specific clipboard tools are missing.

## Dive Deeper
- **Linux**  
The `/copy` command now provides a user-friendly error message when `xclip` or `xsel` is not installed, clearly indicating that the failure is due to missing clipboard tools. This ensures that Linux users can quickly identify and resolve the issue.

image

- **Windows and macOS**  
No additional handling is required for Windows or macOS because the necessary clipboard tools (`clip` on Windows and `pbcopy` on macOS) are typically pre-installed and reliable.

- **Documentation Update**  
The documentation has been updated to include platform-specific clipboard tool requirements and guidance for installing them on Linux.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ✔  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs
- Fixes #6400


HunDun0Ben:
While working on improving /copy error messages, I noticed that the test case: "should handle non-Error clipboard errors" seems redundant. In the current implementation, both primaryError and fallbackError are always wrapped as Error instances, so this scenario may never actually occur.
 It might be worth discussing whether to keep or remove this test to simplify the test suite.

packages/cli/src/ui/utils/commandUtils.ts
```diff
@@ -74,16 +74,33 @@ export const copyToClipboard = async (text: string): Promise<void> => {
           // If xclip fails for any reason, try xsel as a fallback.
           await run('xsel', ['--clipboard', '--input'], linuxOptions);
         } catch (fallbackError) {
-          const primaryMsg =
+          const xclipNotFound =
+            (primaryError as NodeJS.ErrnoException).code === 'ENOENT';
+          const xselNotFound =
+            (fallbackError as NodeJS.ErrnoException).code === 'ENOENT';
```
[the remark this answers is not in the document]
jacob314: please fix

jacob314:
> While working on improving /copy error messages, I noticed that the test case: "should handle non-Error clipboard errors" seems redundant. In the current implementation, both primaryError and fallbackError are always wrapped as Error instances, so this scenario may never actually occur. It might be worth discussing whether to keep or remove this test to simplify the test suite.

please go ahead and fix. there are doubtless plenty of redundant tests in the codebase

docs/cli/commands.md
```diff
@@ -290,7 +293,6 @@ Please generate a Conventional Commit message based on the following git diff:
 ```
 
 """
-
```
jacob314: please revert this line. if it was intentional please add it on a different pr just cleaning up this whitespace.

packages/cli/src/ui/commands/copyCommand.test.ts
```diff
@@ -242,14 +242,15 @@ describe('copyCommand', () => {
     ];
 
     mockGetHistory.mockReturnValue(historyWithAiMessage);
-    mockCopyToClipboard.mockRejectedValue('String error');
```
jacob314, line 245: please mock the actual cases where fairly complicated messages are returned depending on which tools are not found.

packages/cli/src/ui/utils/commandUtils.ts
```diff
@@ -74,16 +74,33 @@ export const copyToClipboard = async (text: string): Promise<void> => {
...
             fallbackError instanceof Error
               ? fallbackError.message
               : String(fallbackError);
+          if (xselNotFound) {
+            fallbackMsg = `xsel not found`;
+          }
+
           throw new Error(
-            `All copy commands failed. xclip: "${primaryMsg}", xsel: "${fallbackMsg}". Please ensure xclip or xsel is installed and configured.`,
+            `All copy commands failed. xclip: "${primaryMsg}", xsel: "${fallbackMsg}". `,
```
jacob314: tests should verify that these messages are shown when expected and tests should show the exact message returned for the mock cases so it is clear that the messages are understandable by users.

jacob314 (commented):
thanks for the detailed error messages! approved after these minor comments are addressed.

HunDun0Ben:
> > While working on improving /copy error messages, I noticed that the test case: "should handle non-Error clipboard errors" seems redundant. In the current implementation, both primaryError and fallbackError are always wrapped as Error instances, so this scenario may never actually occur. It might be worth discussing whether to keep or remove this test to simplify the test suite.
> 
> please go ahead and fix. there are doubtless plenty of redundant tests in the codebase

I’ll take some additional time to go through and address these issues.

docs/cli/commands.md
HunDun0Ben: I didn’t modify this test case intentionally; I’ll check whether the changes come from my IDE’s auto-formatting or Gemini CLI next time, and I’ll restore it.

packages/cli/src/ui/utils/commandUtils.ts
HunDun0Ben: Since the internal error now includes the command name, the message template has been simplified accordingly, and the corresponding test case in `commandUtils.test.ts` has been updated.

packages/cli/src/ui/commands/copyCommand.test.ts
HunDun0Ben: After reviewing the original test case, I realized it mainly shows how non-ERROR objects (like a **_string_**) are handled, while _'should handle clipboard copy error'_ covers _**Error**_ thrown by `copyToClipboard`. Complex cases like missing tools are already handled in `copyToClipboard` tests, so not much change seems necessary here.

HunDun0Ben: I’m happy to make further updates if you have any suggestions.

HunDun0Ben:
Hi @jacob314 , I’ve updated the PR based on your feedback. 
Could you please take another look when you have time? Thanks!

jacob314 (approved):
lgtm
Thanks for the fixes!

packages/cli/src/ui/utils/commandUtils.ts
Amnat-auyana1998: > > > []() to get it to work out and get my id is the stone age input and output and output is khong to get the get it for me to work on it for a few minutes - then ~~@ ~~[]()0https://github.com/google-gemini/gemini-cli/issues/1696https://github.com/google-gemini/gemini-cli/issues/1696

PR #6724, opened 2025-08-21
Introduce system defaults (vs system overrides)

### LINKED ISSUE #6537: Support system-wide default settings (versus overrides)

Support for a system-wide default setting config file.

### Why is this needed?

Today's system settings are not user overridable, and take precedence over any user settings. Would be nice to offer system defaults that users can override.

### Additional context

_No response_

bbiggs:
## TLDR

Add a system-level default settings that can be overridden by user and workspace settings.

## Dive Deeper

Today, system settings override both user settings and workspace settings. This provides no way for a system configuration that provides sensible defaults that can be overridden by users (or workspaces). To solve, this we introduce system settings, configurable as `system-defaults.json` and settable via an environment variable as `GEMINI_CLI_SYSTEM_DEFAULTS_PATH`.

## Reviewer Test Plan

Please validate in real configuration settings across platforms.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes #6537


chrstnb:
Can you also update enterprise.md and configuration.md? this looks great to me and is very clear

packages/cli/src/config/settings.ts
```diff
@@ -71,55 +79,73 @@ export interface SettingsFile {
...
   const { folderTrust, ...workspaceWithoutFolderTrust } = workspace;
 
+  // Settings are merged with the following precedence (last one wins for
+  // single values):
+  // 1. System Defaults
+  // 2. User Settings
+  // 3. Workspace Settings
+  // 4. System Settings (as overrides)
+  //
+  // For properties that are arrays (e.g., includeDirectories), the arrays
```
chrstnb, line 105: I'm not entirely sure how to navigate this between defaults and settings (overrides).

The current behavior is:

**System settings:**
```
mcpServers: [a, b, c]
```

**User settings:**
```
mcpServers: [c, d]
```

the result will be:
```
mcpServers: [a (system), b (system), c (system), d (user)]
```

Is that what we want if we're also supporting system defaults? I worry this will be unexpected behavior for enterprise controls, but otherwise I think merging makes sense.

jacob314 (approved):
Glad we didn't surface  system defaults in the theme or settings dialog. that would have looked silly.

docs/cli/enterprise.md
```diff
@@ -6,23 +6,27 @@ This document outlines configuration patterns and best practices for deploying a
...
+The most powerful tools for enterprise administration are the system-wide settings files. These files allow you to define a baseline configuration (`system-defaults.json`) and a set of overrides (`settings.json`) that apply to all users on a machine. For a complete overview of configuration options, see the [Configuration documentation](./configuration.md).
+
+Settings are merged from four files. The precedence order for single-value settings (like `theme`) is:
+
+1. System Defaults (`system-defaults.json`)
```
jacob314, line 13: Can you clarify the directory where system defaults and system overrides are specified here? Otherwise it kindof seems like they are in the current directory.

jacob314 (approved):
lgtm
Approved but one nit about clarifying the enterprise settings location

PR #1729, opened 2025-06-26
feat(cli): clear input buffer on CTRL+C when not executing commands

matias-casal:
## Description
This PR implements a simple enhancement to clear the input buffer when CTRL+C is pressed while the CLI is idle (not executing any command).

## Changes
- Modified  to handle CTRL+C and clear the buffer when pressed
- Only clears the buffer if there's text; otherwise lets the global handler manage exit flow

## Benefits
- Provides a quick way for users to clear their current input
- Consistent with terminal behavior in many applications
- No breaking changes

## Testing
- Manually tested by typing text and pressing CTRL+C
- Verified that CTRL+C on empty buffer still triggers exit flow
- All existing tests pass

Fixes #1722

PR #6729, opened 2025-08-21
fix: Ctrl+E should move to current line end, not buffer end

mkusaka:
## TLDR

Fixes Ctrl+E keyboard shortcut to correctly move the cursor to the end of the current line instead of the end of the entire buffer in multiline input. This was a regression introduced in #1729.

## Dive Deeper

The issue was caused by an unnecessary call to `buffer.moveToOffset(cpLen(buffer.text))` after `buffer.move('end')`. This line was accidentally added in PR #1729 (commit d622e596a) which was intended to implement Ctrl+C functionality for clearing input, but inadvertently modified the Ctrl+E behavior as well.

The `buffer.move('end')` method already correctly handles moving to the current line end, so the additional `moveToOffset` call was redundant and incorrect, causing the cursor to jump to the end of the entire buffer.

### Rationale for the fix

The gemini-cli appears to follow GNU Readline conventions for keyboard shortcuts, as evidenced by the correct implementation of Ctrl+A which moves to the beginning of the current line (not the beginning of the entire buffer). Standard readline behavior, used in bash, zsh, and most terminal applications, defines:

- **Ctrl+A**: Move to the beginning of the **current line**
- **Ctrl+E**: Move to the end of the **current line**
- **Ctrl+Home/Ctrl+End**: Move to the beginning/end of the entire buffer (in applications that support it)

Since Ctrl+A correctly moves to the current line's beginning in gemini-cli's multiline input, Ctrl+E should symmetrically move to the current line's end. This consistency is important for users familiar with standard terminal navigation.

## Reviewer Test Plan

1. **Manual Testing**:
   - Launch gemini-cli
   - Type a multiline prompt (use Shift+Enter for new lines):
     ```
     line 1
     line 2
     line 3
     ```
   - Move cursor to middle of "line 2"
   - Press Ctrl+E
   - Verify cursor moves to end of "line 2", not end of "line 3"

2. **Run the new tests**:
   ```bash
   cd packages/cli
   npx vitest run src/ui/components/InputPrompt.test.tsx
   ```

3. **Verify no regression for single line**:
   - Type a single line prompt
   - Press Ctrl+E anywhere in the line
   - Verify cursor moves to end of line

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✔  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

fix: #6435


jacob314 (approved):
Thanks for the cleanup!
lgtm

PR #6732, opened 2025-08-21
feat(settings) support editing string settings.

### LINKED ISSUE #6774: String settings render incorrectly in the settings dialog and cannot be edited
### What happened?

I was reviewing a PR to add a new string setting surfaced in the setting dialog and noticed that it rendered incorrectly

Expect to see the value of the field correctly and expected to be able to edit it like numeric settings.

jacob314:
## TLDR

Support string properties just like we already support numeric properties

## Dive Deeper

Allow entering all safe characters into strings using the same functionality to verify safe characters as used in text-buffer
In the future we should consider making the text editing support richer and more unified with text-buffer.ts / input-prompt.ts but this is good enough for now.

## Reviewer Test Plan

Change settingsSchema.ts so that the theme has showInDialog: true
Try editing the theme and verify things work.

Delete the theme from settings.json, exit out of the theme dialog shown on startup, and then verify the missing theme in settings.json renders appropriately

## Linked issues / bugs

Fixes #6774

### DIFF
--- a/packages/cli/src/ui/components/shared/text-buffer.ts
+++ b/packages/cli/src/ui/components/shared/text-buffer.ts
@@ -4,8 +4,6 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import stripAnsi from 'strip-ansi';
-import { stripVTControlCharacters } from 'util';
 import { spawnSync } from 'child_process';
 import fs from 'fs';
 import os from 'os';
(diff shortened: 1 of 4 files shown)


packages/cli/src/ui/components/shared/text-buffer.ts
```diff
@@ -494,51 +497,6 @@ export const replaceRangeInternal = (
   };
 };
 
-/**
```
jacob314, line 497: this is just moved to textUtils.ts

jacob314:
Fyi @ali-aljufairi

SandyTao520 (approved):

NTaylorMullen (approved):

PR #6771, opened 2025-08-21
fix: handle extra text in gemini output for dedup workflow

gsquared94:
The CLI is occasionally outputting extra sentences like:
```
Raw duplicates JSON: Data collection is disabled.
Accessing resource attributes before async attributes settled
Invalid attribute value set for key: metadata
Invalid attribute value set for key: metadata
{"duplicate_issues": []}
```

This PR updates the parsing logic in the issue deduplication workflow to handle extra text.


jacob314 (approved):
lgtm

Issue #6582, opened 2025-08-19
Support invoking tools from VsCode extensions


jakemac53:
### What would you like to be added?

See https://code.visualstudio.com/api/references/vscode-api#lm - specifically the `tools` variable and `invokeTool` function.

These could be exposed as tools from the VsCode companion extension to Gemini, giving it access to all these tools registered by extensions.

### Why is this needed?

This would bring Gemini CLI in VsCode closer to CoPilot in terms of the tools available to it. Extensions can expose any number of useful language model tools in this way.

Concretely, the DartCode extension uses this to communicate the debug port URIs for running apps to agents, so they can connect to running applications.

jakemac53:
Hmm, this will be a bit tricky because it looks like some of the registered tools are duplicative with ones that Gemini CLI already has available, I am seeing the following all listed:

```
Registering tool: setup.tools.createNewWorkspace
Registering tool: get_dart_tooling_daemon_dtd_uri // From the DartCode extension, what I actually want :)
Registering tool: run_in_terminal
Registering tool: get_terminal_output
Registering tool: run_task
Registering tool: get_task_output
Registering tool: create_and_run_task
Registering tool: runTests
```

PR #6773, opened 2025-08-21
Add option to include VsCode language model tools

jakemac53:
## TLDR

Exposes the tools from [this VsCode API](https://code.visualstudio.com/api/references/vscode-api#lm) as MCP tools from the IDE companion extension.

## Dive Deeper

This enables Gemini CLI to get access to tools registered by VsCode extensions, when using the IDE companion extension.

There is a hard coded block list as well of certain tools that are duplicative with built-in tools, copilot specific, etc.

Interestingly, MCP tools also appear to show up in this list for MCP servers that you have configured in VsCode. These could very well be duplicative with the MCP servers you have configured with Gemini CLI which is less than ideal if that is the case - but it also might provide a nice out of the box experience for some users when extensions can automatically enable certain MCP servers. Unfortunately we can't really tell where these tools are coming from, so detecting duplicates is potentially challenging.

Because of these factors, this is a somewhat unpredictable feature and currently disabled by default until we can figure out the best way to manage it, we may just want an explicit allow list of IDE tools, as opposed to a blocklist.

I also moved the IDE client connection inside of the `config.initialize` method - this needs to happen prior to tool discovery now which happens in that method as well. As a part of tool discovery it will check for these tools.

## Reviewer Test Plan

You will need to have an extension connected which registers some LM tool, then you can enable the setting (`"enableIdeLanguageModelTools": true`) and you can invoke that tool from Gemini CLI.

One example of this is using the DartCode extension and then asking gemini to connect to your running flutter application, as shown here:

https://github.com/user-attachments/assets/c60a6023-5e3f-4894-9bf3-b47be6f0b81f

Specifically, the first tool invocation to get the DTD Uri is coming from the DartCode extension.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

First part of https://github.com/google-gemini/gemini-cli/issues/6582


packages/vscode-ide-companion/src/lm-tools.ts
```diff
@@ -0,0 +1,68 @@
...
+              input,
+            });
+            return {
+              // Convert the language model content parts to text content parts.
+              content: result.content.map(
+                (part): TextContent => ({
+                  type: 'text',
+                  text: JSON.stringify(part),
+                }),
+              ),
```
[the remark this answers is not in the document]
jakemac53: I think it is better to leave this as is - representing the multiple parts as they were returned by the tool. Each part might have a different "role" attached and this will closer represent that.

Note that this isn't really perfect still in its current state as it just json encodes all the responses, but its the safest thing to do here for now.

You can see the test, you will notice you end up with something like `{"result": "someText"}` as the text value, instead of just "someText". But, I didn't want to risk dropping additional data on the floor here.

packages/core/src/ide/ide-client.ts
```diff
@@ -197,6 +228,80 @@ export class IdeClient {
...
+      const tool = await mcpCallableTool.tool();
+
+      if (!Array.isArray(tool.functionDeclarations)) {
+        return;
+      }
+
+      for (const funcDecl of tool.functionDeclarations) {
+        try {
+          if (!funcDecl.name) {
+            logger.debug(
```
jacob314, line 278: nit: probably only log this in debug mode.

jakemac53: is that not what `logger.debug` does?

jacob314:
This is a great addition for integrating IDE tools. I noticed that `LanguageModelTools` is implemented as a class. To better align with the project's conventions outlined in `GEMINI.md`, which favor a more functional approach, could we refactor this into a standalone function?

For example, we could have a single exported function like `registerLanguageModelTools(server: McpServer, log: (message: string) => void)`. This would achieve the same result while reducing boilerplate and adhering to the preferred coding style.

packages/core/src/ide/ide-client.ts
jacob314: nope. you got tricked by a method in ide-client.ts that we should remove calling console.debug, etc directly.
Feel free to remove the misleading `logger` boilerplate in this file that shouldn't have been added as it is inconsistent with the rest of the code base.

packages/vscode-ide-companion/src/lm-tools.ts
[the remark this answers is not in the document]
jacob314: the result.content is typically a 
vscode.LanguageModelTextPart
I would recommend detecting that typical case. It might even be the only case modulo future proof additional changes.
For that case I think this can just be 
`text: result.content.value`
which will help Gemini CLI do better

jacob314 (commented):
This is great! Approving once these minor changes are made.

packages/vscode-ide-companion/package.json
```diff
@@ -123,9 +123,11 @@
     "vitest": "^3.2.4"
   },
   "dependencies": {
+    "@google/gemini-cli-core": "file:../core",
```
jacob314, line 126: is this dependency needed? it would be good not to depend on core in the IDE companion if we can.

jakemac53: Yes, all `npm run` commands fail without this, it is imported and used by the companion.

packages/vscode-ide-companion/src/lm-tools.ts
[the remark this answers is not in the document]
jakemac53: We will also lose additional data from `LanguageModelTextPart` such as a "role" were it added in the future (which it seems like it should be). So, this was my reasoning for just leaving it as is.

I could just extract the text but it will be less future proof.

packages/core/src/ide/ide-client.ts
jakemac53: What is the supported way of doing logging? Should I just remove it entirely?

chrstnb:

PR #6777, opened 2025-08-21
Retry Message Stream on Empty Chunks

### LINKED ISSUE #6474: Degradation of Instruction Following in Gemini Model
### What happened?

Both in Gemini CLI and via the API, users had been reporting issues with instruction following - the model fails to respond to the most recent message and instead get into a loop of responding to an older message or other context in the chat history. This appears to had been an issue for some time now, but has gotten worse recently.

##### Gemini CLI

- https://github.com/google-gemini/gemini-cli/issues/6147
- https://github.com/google-gemini/gemini-cli/issues/6246
- https://github.com/google-gemini/gemini-cli/issues/6455
- https://github.com/google-gemini/gemini-cli/issues/6239
- https://github.com/google-gemini/gemini-cli/issues/6221
- https://github.com/google-gemini/gemini-cli/issues/6200
- https://github.com/google-gemini/gemini-cli/issues/6190
- https://github.com/google-gemini/gemini-cli/issues/6443
- https://github.com/google-gemini/gemini-cli/issues/6149
- https://github.com/google-gemini/gemini-cli/issues/6136
- https://github.com/google-gemini/gemini-cli/issues/6127
- https://github.com/google-gemini/gemini-cli/issues/6319
- https://github.com/google-gemini/gemini-cli/issues/6067
- https://github.com/google-gemini/gemini-cli/issues/6034
- https://github.com/google-gemini/gemini-cli/issues/5971
- https://github.com/google-gemini/gemini-cli/issues/5963
- https://github.com/google-gemini/gemini-cli/issues/5854
- #6739
- #6735
- #6719

##### Google Support:

- [Gemini AI chat returning repeated responses](https://support.google.com/gemini/thread/365334436/issue-with-gemini-ai-chat-returning-repeated-responses?hl=en)
- [Gemini Pro keeps referring back to older prompts](https://support.google.com/gemini/thread/342765378/gemini-pro-keeps-referring-back-to-older-prompts?hl=en)
- [Gemini 2.5 Pro Repeats Itself](https://support.google.com/gemini/thread/349203730/gemini-2-5-pro-repeats-itself-seems-not-to-be-related-to-stuck-loop-issue-reported-elsewhere?hl=en)

##### Reddit:
- [For the first time, I believe Gemini 2.5 Pro is performing worse. Power User since Day 1.](https://www.reddit.com/r/GeminiAI/comments/1mnc3ty/for_the_first_time_i_believe_gemini_25_pro_is/)
- [Has gemini got a lobotomy last Friday?](https://www.reddit.com/r/Bard/comments/1mnko71/has_gemini_got_a_lobotomy_last_friday/)
- [RooCode Gemini 2.5 has been awful lately](https://www.reddit.com/r/RooCode/comments/1mq9p9h/roocode_gemini_25_has_been_awful_lately/)
- [Gemini 2.5 Pro Suddenly Struggling with Coding Instructions. Anyone else?](https://www.reddit.com/r/GeminiAI/comments/1mne23q/gemini_25_pro_suddenly_struggling_with_coding/)

That the model responds to the most recent user instruction or agent-generated tool-call.

### Client information

<details>
Version 0.1.21
</details>

### Login information

Google Account (OAuth): 3
Api key: 3

_No response_

### LINKED ISSUE #6889: Add retry handling of invalid text streams

mrcabbage972:
## TLDR

Users could encounter a scenario where they would submit a prompt, but it would vanish from the UI and message history if the Gemini model returned a response that didn't pass the isValidContent check. 

This PR introduces a new, content-aware retry loop. The retries happen on invalid chunks coming from the model. If all retries fail, the loop raises an error instead of silently continuing.

## Dive Deeper

This change introduces a new, higher-level retry loop in the sendMessageStream function. Its responsibility is specifically to validate the content of the stream:

Network Retry (Existing): retryWithBackoff remains in place within makeApiCallAndYieldStream. Its job is to ensure a successful network connection and handle initial API errors, including the fallback to the Flash model.

Content Retry (New): The new loop in sendMessageStream consumes the stream provided by makeApiCallAndYieldStream. After the stream is fully consumed, it checks if any valid content was actually received. If not, it triggers a retry of the entire operation.

If the retries fail, the submitQuery function in useGeminiStream.ts catches the error that is raised.

The following test cases were added:
1. Invalid content is returned, but the retry succeeds.
2. When the retries fail,  an error is raised.
3. Concurrency test for sendMessageStream that verifies that a new streaming message is blocked until the previous stream has been fully consumed, preventing race conditions.
4. Retry test for completely empty streams: Validates that the system correctly identifies a stream that yields no chunks as a content failure and properly triggers a retry.
5. Test that useGeminiStreaming gets the correct error message on an invalid chunk.

### Test Persistent Failure:

Modify the isValidResponse function to fail consistently.

Run a prompt in the CLI.

*Expected*: The CLI should attempt the call three times (initial + two retries). The sendMessageStream method in GeminiChat.ts will then throw an error. The useGeminiStream.ts hook will catch this error and display an error message in the UI.

The error should be:
```
The model failed to respond after multiple attempts. Error: Model stream was invalid or completed without valid content.
```

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | 👍   | ❓  | 👍  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs
- Resolves #6889


packages/core/src/core/geminiChat.ts
```diff
@@ -306,64 +314,99 @@ export class GeminiChat {
...
-            if (error.message.includes('429')) return true;
-            if (error.message.match(/5\d{2}/)) return true;
+        return stream; // If successful, return the generator and exit the loop.
+      } catch (error) {
+        lastError = error;
+        const errorMessage = error instanceof Error ? error.message : '';
+
+        // Check for retryable conditions
+        const isRetryableNetworkError =
+          errorMessage.includes('429') || errorMessage.match(/5\d{2}/);
```
silviojr: I don't think we should retry if the error is 429. The `retryWithBackoff` already handles that. So in the end we may end up trying MAX_RETRIES x 2 for 429 errors.

packages/core/src/core/geminiChat.ts
```diff
@@ -451,43 +492,30 @@ export class GeminiChat {
...
-      }
+    // Now that the stream is finished, make a decision.
+    if (isStreamInvalid) {
+      throw new EmptyStreamError('Model stream had invalid chunks');
     }
-    this.recordHistory(inputContent, outputContent);
-  }
 
+    // If valid, add the fully assembled response to history.
+    this.history.push({ role: 'model', parts: modelResponseParts });
```
anj-s: recordHistory does other things as well. any reason we decided to use this?

silviojr: We are going to end up adding thoughts to the history in this case and the history is part of the request. That's ok, but when we are building the request we are not filtering out thoughts. That will increase the context by a lot.

packages/core/src/core/geminiChat.ts
mrcabbage972: There is no retryWithBackoff here. The handling for network and content errors is unified.

packages/core/src/core/geminiChat.ts
mrcabbage972: fixed

mrcabbage972: fixed

packages/core/src/core/geminiChat.ts
```diff
@@ -451,127 +494,101 @@ export class GeminiChat {
...
+        const content = chunk.candidates?.[0]?.content;
+        if (content) {
+          // Filter out thought parts from being added to history.
+          if (!this.isThoughtContent(content) && content.parts) {
+            modelResponseParts.push(...content.parts);
           }
         }
-        yield chunk;
+      } else {
+        isStreamInvalid = true;
```
cornmander, line 548: Why don't we throw here?

packages/core/src/core/geminiChat.ts
```diff
@@ -451,127 +494,101 @@ export class GeminiChat {
...
+    }
+
+    // Part 3: Consolidate the parts of this turn's model response.
+    const consolidatedOutputContents: Content[] = [];
+    if (outputContents.length > 0) {
+      for (const content of outputContents) {
+        const lastContent =
+          consolidatedOutputContents[consolidatedOutputContents.length - 1];
+        if (this.isTextContent(lastContent) && this.isTextContent(content)) {
+          lastContent.parts[0].text += content.parts[0].text || '';
```
cornmander, line 617: when do we ever need the || here?

packages/core/src/core/geminiChat.ts
```diff
@@ -451,127 +494,101 @@ export class GeminiChat {
...
-        consolidatedOutputContents.shift(); // Remove the first element as it's merged
       }
-      this.history.push(...consolidatedOutputContents);
     }
+
+    // Part 4: Add the new turn (user and model parts) to the main history.
+    this.history.push(...newHistoryEntries, ...consolidatedOutputContents);
   }
 
   private isTextContent(
```
cornmander: the name of this function is very misleading as it doesn't preclude Content from having non-text content.

hasTextContent?

packages/core/src/core/geminiChat.ts
```diff
@@ -305,65 +313,113 @@ export class GeminiChat {
...
-          );
-        }
+    // eslint-disable-next-line @typescript-eslint/no-this-alias
+    const self = this;
+    return (async function* () {
+      try {
+        const MAX_RETRIES = 2;
+        let lastError: unknown = new Error('Request failed after all retries.');
+
+        for (let attempt = 0; attempt <= MAX_RETRIES; attempt++) {
```
NTaylorMullen: Why build a separate retry mechanism vs. updating the `retryWithBackoff`? Is there a key difference?

mrcabbage972: Thanks for the question, I hope the following makes sense.

The goal is to handle content-level errors (like an empty stream) transparently, without sacrificing the real-time streaming to the UI. We achieve this by wrapping the original API stream.

The sendMessageStream method doesn't return the raw API stream directly to the caller. Instead, it returns a new, wrapped async generator. This wrapper has two main jobs:
1. Proxying Chunks: It immediately starts consuming the real API stream. As soon as it receives a chunk, it yields it to the consumer (the UI). This preserves the word-by-word streaming experience.
2. Inspecting Content: As the chunks pass through, the wrapper inspects them to see if they are valid and keeps track of whether any chunks have been received at all.

The wrapper can only be certain that a stream was completely empty after the underlying API stream has fully closed. 

Meanwhile, retryWithBackoff utility wraps an entire operation that returns a single Promise. It cannot "peek inside" the operation while it's still running.

Lastly, I wanted to have a separation of concerns between handling retries of API-level errors and retries of content errors such as empty chunks.

packages/core/src/core/geminiChat.ts
mrcabbage972: Not supposed to, but in case the text is undefined.

packages/core/src/core/geminiChat.ts
mrcabbage972: renamed

packages/core/src/core/geminiChat.ts
mrcabbage972: So that we consume the stream until it is fully closed. There might be useful metadata coming in, e.g. an error code.

packages/core/src/core/geminiChat.ts
NTaylorMullen: Great explanation!

NTaylorMullen (approved):

PR #6823, opened 2025-08-22
Smart Edit Tool 

silviojr:
## TLDR

This is the Smart Edit tool, a potential replacement for the current 'replace/edit' tool. 

The main differences:
1. Smart Edit takes four arguments: the usual **old_string**, **new_string**, **file_path** and **instructions** describing the change that it's trying to perform. It only works on editing one occurrence at a time. 
2. The flow is as following:
2.1. It tries an exact match, if not succeed, continue
2.3 It tries a flexible match where we loose the constrains mostly on spaces. If not succeed, continues
2.4 It asks Gemini to Fix the old_string and potentially the new_string taking into consideration the instructions it received using gemini 2.5 flash 
2.5 Tries the exact match and flexible match one last time. 

I did some tests with the tool running it on swebench verify and swebench lite. The error rate for the editing action dropped from around 20% to 6%.

We are not making it the default tool yet. In order to run it, you have to enable it in your `.gemini/settings.json` file

`
{ ....
    "useSmartEdit": true
}
`
 **Token usage.** 
The current EditTool can do up to two LLM calls, this new tool do only up to one. In experiments, we noticed a drop in input_token_count of about 30%. 

**Escaping characters**
The current EditTool tries multiple character escaping strategies. We removed those strategies without harming the the performance. 
 

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


packages/cli/src/config/config.ts
```diff
@@ -195,6 +196,10 @@ export async function parseArguments(): Promise<CliArgs> {
           description: 'Enables checkpointing of file edits',
           default: false,
         })
+        .option('use-smart-edit', {
```
NTaylorMullen: Lets stick with settings.json for now. Trying to gate keep our command line args 😄

packages/cli/src/config/settingsSchema.ts
```diff
@@ -407,6 +407,15 @@ export const SETTINGS_SCHEMA = {
     description: 'A blacklist of MCP servers to exclude.',
     showInDialog: false,
   },
+  useSmartEdit: {
+    type: 'boolean',
+    label: 'Use Smart Edit',
+    category: 'Advanced',
+    requiresRestart: false,
+    default: false,
```
NTaylorMullen: We're only defaulting to `false` for now to trial the rollout I assume right?

packages/core/src/tools/smart-edit.ts
```diff
@@ -0,0 +1,813 @@
...
+}
+
+/**
+ * Implementation of the Edit tool logic
+ */
+export class SmartEditTool
+  extends BaseDeclarativeTool<EditToolParams, ToolResult>
+  implements ModifiableDeclarativeTool<EditToolParams>
+{
+  static readonly Name = 'smart_edit';
```
NTaylorMullen, line 708: We reference the regular edit tool alll over the place in system prompt etc. In addition, we also end up utilizing it for various other feature lookups.

```suggestion
  static readonly Name = EditTool.name;
```

packages/core/src/tools/smart-edit.ts
```diff
@@ -0,0 +1,813 @@
...
+import { makeRelative, shortenPath } from '../utils/paths.js';
+import { isNodeError } from '../utils/errors.js';
+import { Config, ApprovalMode } from '../config/config.js';
+import { DEFAULT_DIFF_OPTIONS, getDiffStat } from './diffOptions.js';
+import { ReadFileTool } from './read-file.js';
+import { ModifiableDeclarativeTool, ModifyContext } from './modifiable-tool.js';
+import { IDEConnectionStatus } from '../ide/ide-client.js';
+import { FixLLMEditWithInstruction } from '../utils/llm-edit-fixer.js';
+
+export function applyReplacement(
```
NTaylorMullen, line 31: Why export all these functions? I imagine we should probably group these as private methods in the smart edit class for proper isolation

packages/core/src/tools/smart-edit.ts
```diff
@@ -0,0 +1,813 @@
...
+  currentContent: string | null,
+  oldString: string,
+  newString: string,
+  isNewFile: boolean,
```
NTaylorMullen, line 35: Is this equivalent to `currentContent === null`?

packages/core/src/tools/smart-edit.ts
```diff
@@ -0,0 +1,813 @@
...
+import { isNodeError } from '../utils/errors.js';
+import { Config, ApprovalMode } from '../config/config.js';
+import { DEFAULT_DIFF_OPTIONS, getDiffStat } from './diffOptions.js';
+import { ReadFileTool } from './read-file.js';
+import { ModifiableDeclarativeTool, ModifyContext } from './modifiable-tool.js';
+import { IDEConnectionStatus } from '../ide/ide-client.js';
+import { FixLLMEditWithInstruction } from '../utils/llm-edit-fixer.js';
+
+export function applyReplacement(
+  currentContent: string | null,
```
NTaylorMullen, line 32: instead of `null` can we lean on `undefined` (more JavaScripty)

packages/core/src/tools/smart-edit.ts
```diff
@@ -0,0 +1,813 @@
...
+  if (isNewFile) {
+    return newString;
+  }
+  if (currentContent === null) {
+    // Should not happen if not a new file, but defensively return empty or newString if oldString is also empty
+    return oldString === '' ? newString : '';
+  }
+  // If oldString is empty and it's not a new file, do not modify the content.
+  if (oldString === '' && !isNewFile) {
+    return currentContent;
```
NTaylorMullen, line 49: The inclusion of current content vs. newString etc. is a bit confusing. Why do we need all the various ways to represent mutate vs. create?

packages/core/src/tools/smart-edit.ts
```diff
@@ -0,0 +1,813 @@
...
+  }
+  return currentContent.replaceAll(oldString, newString);
+}
+
+interface ReplacementContext {
```
NTaylorMullen, line 54: these private interfaces should also be nested in the corresponding class

packages/core/src/tools/smart-edit.ts
NTaylorMullen, line 34: I know you were just copying from the edit.ts tool BUT as follow ups we should definitely come back and revisit all these functions. They're quite spaghetti 😆

packages/core/src/tools/smart-edit.ts
```diff
@@ -0,0 +1,813 @@
...
+}
+
+async function calculateExactReplacement(
+  context: ReplacementContext,
+): Promise<ReplacementResult | null> {
+  const { currentContent, params } = context;
+  const { old_string, new_string } = params;
+
+  const normalizedCode = currentContent;
+  const normalizedSearch = old_string.replace(/\r\n/g, '\n');
```
NTaylorMullen, line 87: Normalizing is fine BUT what you're doing here is not taking into account the current newline state of the file and forcing them all to `\n`. Meaning, on a windows machine you'll be breaking the newlines of a given document by weaving in `\n` when most likely the rest of the document may be `\r\n`.

One normalizing algorithm here could be to take the old_string/new_string, convert it to document specific newlines (see if the doc contains OS specific newliens), see if you can find a match and replace it with the right newline content. That being said, there's a ton you could potentially do.

You'll want to consider this elsewhere as well.

Anyhow, all these should be super easily testable against too to emulate various newlines structures

packages/core/src/tools/smart-edit.ts
```diff
@@ -0,0 +1,813 @@
...
+  const normalizedReplace = new_string.replace(/\r\n/g, '\n');
+
+  const exactOccurrences = normalizedCode.split(normalizedSearch).length - 1;
+  if (exactOccurrences > 0) {
+    let modifiedCode = normalizedCode.replaceAll(
+      normalizedSearch,
+      normalizedReplace,
+    );
+    modifiedCode = restoreTrailingNewline(currentContent, modifiedCode);
```
NTaylorMullen, line 96: If the user requested the model to trim content, wouldn't this break that assumption?

packages/core/src/utils/llm-edit-fixer.ts
```diff
@@ -0,0 +1,135 @@
...
+    return cachedResult;
+  }
+  const userPrompt = EDIT_USER_PROMPT.replace('{instruction}', instruction)
+    .replace('{old_string}', old_string)
+    .replace('{new_string}', new_string)
+    .replace('{error}', error)
+    .replace('{current_content}', current_content);
+
+  const contents: Content[] = [
+    { role: 'user', parts: [{ text: `${EDIT_SYS_PROMPT}\n${userPrompt}` }] },
```
NTaylorMullen: Should use OS newlines to not sway the model in a bad direction
```suggestion
    { role: 'user', parts: [{ text: `${EDIT_SYS_PROMPT}
${userPrompt}` }] },
```

packages/core/src/utils/llm-edit-fixer.ts
```diff
@@ -0,0 +1,135 @@
...
+5.  **Exactness:** The final \`search\` field must be the EXACT literal text from the file. Do not escape characters.
+`;
+
+const EDIT_USER_PROMPT = `
+# Goal of the Original Edit
+{instruction}
+
+# Failed Attempt Details
+- **Original \`search\` parameter (failed):**
+\`\`\
```
NTaylorMullen: Did you mean to have an extra back tik on these? Also for prompts you may want this to be dynamic on the back tick utilization depending on the old/new/error text. I.e. if any of those have backticks it braeks markdown. Meaning, you'll want to add 1 additional open/closing backtick per matching multi back tick code block you find

packages/core/src/utils/llm-edit-fixer.ts
```diff
@@ -0,0 +1,135 @@
...
+
+const MAX_CACHE_SIZE = 50;
+
+const EDIT_SYS_PROMPT = `
+You are an expert code-editing assistant specializing in debugging and correcting failed search-and-replace operations.
+
+# Primary Goal
+Your task is to analyze a failed edit attempt and provide a corrected \`search\` string that will match the text in the file precisely. The correction should be as minimal as possible, staying very close to the original, failed \`search\` string. Do NOT invent a completely new edit based on the instruction; your job is to fix the provided parameters.
+
+It is important that you do no try to figure ou if the instruction is correct. DO NOT GIVE ADVICE. Your only goal here is to do your best to perform the search and replace task!
```
NTaylorMullen:
```suggestion
It is important that you do no try to figure out if the instruction is correct. DO NOT GIVE ADVICE. Your only goal here is to do your best to perform the search and replace task! 
```

packages/core/src/utils/llm-edit-fixer.ts
```diff
@@ -0,0 +1,135 @@
...
+You will be given:
+1. The high-level instruction for the original edit.
+2. The exact \`search\` and \`replace\` strings that failed.
+3. The error message that was produced.
+4. The full content of the source file.
+
+# Rules for Correction
+1.  **Minimal Correction:** Your new \`search\` string must be a close variation of the original. Focus on fixing issues like whitespace, indentation, line endings, or small contextual differences.
+2.  **Explain the Fix:** Your \`explanation\` MUST state exactly why the original \`search\` failed and how your new \`search\` string resolves that specific failure. (e.g., "The original search failed due to incorrect indentation; the new search corrects the indentation to match the source file.").
+3.  **Preserve the \`replace\` String:** Do NOT modify the \`replace\` string unless the instruction explicitly requires it and it was the source of the error. Your primary focus is fixing the \`search\` string.
```
NTaylorMullen, line 32: When would the replace string need modification / how would we know?

packages/core/src/utils/llm-edit-fixer.ts
```diff
@@ -0,0 +1,135 @@
...
+4.  **No Changes Case:** CRUCIAL: if the change is already present in the file,  set \`noChangesRequired\` to True and explain why in the \`explanation\`. It is crucial that you only do this if the changes outline in \`replace\` are alredy in the file and suits the instruction!!
```
NTaylorMullen, line 33: Can you share more here. Don't we know that this isn't true?

packages/core/src/utils/llm-edit-fixer.ts
```diff
@@ -0,0 +1,135 @@
...
+  explanation: string;
+}
+
+const SearchReplaceEditSchema = {
+  type: Type.OBJECT,
+  properties: {
+    search: { type: Type.STRING },
+    replace: { type: Type.STRING },
+    noChangesRequired: { type: Type.BOOLEAN },
+    explanation: { type: Type.STRING },
```
NTaylorMullen: If your goal with the explanation is to have re-inforced CoT you'll want to put this first in the schema. Otherwise it's just for debugging purposes

packages/core/src/tools/smart-edit.ts
```diff
@@ -0,0 +1,813 @@
...
+      
+      The user has the ability to modify the \`new_string\` content. If modified, this will be stated in the response.
+      
+      Expectation for required parameters:
+      1. \`file_path\` MUST be an absolute path; otherwise an error will be thrown.
+      2. \`old_string\` MUST be the exact literal text to replace (including all whitespace, indentation, newlines, and surrounding code etc.).
+      3. \`new_string\` MUST be the exact literal text to replace \`old_string\` with (also including all whitespace, indentation, newlines, and surrounding code etc.). Ensure the resulting code is correct and idiomatic and that \`old_string\` and \`new_string\` are different.
+      4.  \`instruction\` is the detailed instruction of what needs to be changed. It is important to Make it specific and detailed so developers or large language models can understand what needs to be changed and perform the changes on their own if necessary.
```
NTaylorMullen:
```suggestion
      4. \`instruction\` is the detailed instruction of what needs to be changed. It is important to Make it specific and detailed so developers or large language models can understand what needs to be changed and perform the changes on their own if necessary. 
```

packages/core/src/tools/smart-edit.ts
```diff
@@ -0,0 +1,813 @@
...
+export class SmartEditTool
+  extends BaseDeclarativeTool<EditToolParams, ToolResult>
+  implements ModifiableDeclarativeTool<EditToolParams>
+{
+  static readonly Name = 'smart_edit';
+
+  constructor(private readonly config: Config) {
+    super(
+      SmartEditTool.Name,
+      'Smart Edit',
```
NTaylorMullen: Lets leave the display name the same as the existing edit tool

NTaylorMullen (commented):
Some absolutely stellar results here. Left some key comments mostly around cross-plat capabilities and some general goodness factors.

packages/cli/src/config/settingsSchema.ts
silviojr: That is correct.

packages/cli/src/config/config.ts
silviojr: Done!

packages/core/src/tools/smart-edit.ts
silviojr: That's a fair point. 

The tool  uses the **old_string** parameter to differentiate between two main operations: create file VS change file. If the **old_string** is empty, this can only be interpreted as a new file - since we can't replace an empty string in an existing file. 

This check, for example, is supposed to be defensive. If the **old_string** is empty but the file already exists, this is basically an error and we decide to return the currentcontent of the file to avoid destructing the file. This error will be caught upstream as no replacement/editing made.

packages/core/src/tools/smart-edit.ts
silviojr: Ok. You're right. I added a new test case to detect the line ending and return them back to original

packages/core/src/tools/smart-edit.ts
silviojr: You are correct. However we have observed that Gemini ignores the trailing new lines quite often in its replacements. I think it's worth to keep this as it is for now, as this is a very specific case. what do you think?

packages/core/src/tools/smart-edit.ts
silviojr: One of my fears is that we do not have a way differentiate if the user will have enabled **useSmartEdit** in our logs and then we won't know the error rate. 

What do you think of also mentioning the Smart edit tool as an option where the editTool is mentioned? It's not many places: https://github.com/search?q=repo%3Agoogle-gemini%2Fgemini-cli%20EditTool.name&type=code

Or even better, we could have a variable in the config (EditToolName...) that can be changed according to which tool is enabled. 

What do you think?

packages/core/src/tools/smart-edit.ts
silviojr: done!

packages/core/src/utils/llm-edit-fixer.ts
silviojr: This would cover when the **search** and **replace** are the same - which would be counted as an error. Gemini can still try to figure out if it can fix this.

packages/core/src/tools/smart-edit.ts
silviojr: I'll do this as a follow-up

packages/core/src/tools/smart-edit.ts
silviojr: I'll do this in a follow up if that's ok.

packages/core/src/utils/llm-edit-fixer.ts
silviojr: This rule handles a  edge case: when the initial LLM agent hallucinates or provides an incorrect old_string based on a faulty assumption about the file's content. 

This noChangesRequired rule allows the fixer to act as a sanity check. It can recognize that the initial failure happened because of a bad premise, and it can gracefully conclude that no work is needed (if that's the case) and bubble that information up to the LLM as an Error , preventing it from attempting another unnecessary or incorrect edit.

packages/core/src/utils/llm-edit-fixer.ts
silviojr: You are right. 

I changed it completely to use pseudo xml tags instead.

packages/core/src/tools/smart-edit.ts
anj-s: Given that we don't mention this tool in the system prompt, have you had any issue having the model actually use this tool? Thats the only risk i see wrt name.

anj-s (approved):

packages/core/src/tools/smart-edit.ts
asadm: does it make sense to include it in prompt as a fallback? "if you get stuck in loop trying to edit a file, try {this tool name}".

silviojr: >> Given that we don't mention this tool in the system prompt, have you had any issue having the model actually use this tool? Thats the only risk i see wrt name.


It used the tool multiple times, but I agree with you that we don't want to take the risk. I will add a next PR where we dynamically choose the name of the default edit tool to be used.

silviojr: @asadm I think your comment was truncated. Try what?

asadm: sorry fixed.

silviojr: @asadm I think for now we should let them be tested separately. it will also be less confusing to the user. I will fix the sytem prompt.

Manamama:
Ta again, works fine after 3 days or more, see: https://github.com/google-gemini/gemini-cli/discussions/7758#discussioncomment-14336666

PR #6856, opened 2025-08-22
fix: copy command delay in Linux handled

hritan:
## TLDR

This PR adds an exit event for the child process in Linux because the close event which comes after exit was taking too long to fire. Since the exit event signifies a successful copy, for linux we are resolving the promise on the exit event.
https://github.com/google-gemini/gemini-cli/issues/5255

## Dive Deeper

This resolves the issue where the close event takes too long to fire. The close event is emitted after a process has exited and all of its stdio streams (stdin, stdout, stderr) have been closed.  Since the stream isn't automatically closing, we are manually closing the child process's streams right after it exits to trigger the close event.

Extension of the PR #6482 

There was a gap in previous PR, although the copy command working was fixed and the promise was resolving but it was taking too long. This PR addresses that gap and adds an exit event only for the Linux environment.

## Reviewer Test Plan

1.Do npm start in a Linux terminal.
2. Say "hi" to gemini. It will return an output.
3. Give /copy command in the input
4. You will instantly see an info note for successful copy and the input box is visible again for further chat.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

<!--

Fixes - https://github.com/google-gemini/gemini-cli/issues/5255
Related - #6482


packages/cli/src/ui/utils/commandUtils.ts
```diff
@@ -53,16 +53,36 @@ export const copyToClipboard = async (text: string): Promise<void> => {
...
         reject(
           new Error(
             `'${cmd}' exited with code ${code}${errorMsg ? `: ${errorMsg}` : ''}`,
           ),
         );
+      };
+
+      // The 'exit' event workaround is only needed for the specific stdio
+      // configuration used on Linux, which is indicated by `options` being present.
+      if (options) {
```
jacob314: if this workaround is for linux, condition based on the os being linux not an option that happens to be currently set on linux.

jacob314 (commented):
please add a unit or integration test for this additional logic.

packages/cli/src/ui/utils/commandUtils.ts
hritan: done

briancampo:
Will this be merged?  I continue to get this error on a semi-regular basis.

abhipatel12 (approved):
LGTM!

PR #6870, opened 2025-08-22
Metrics for Retries on Content Errors

### LINKED ISSUE #6890: Add metrics for the content retry handling mechanism

mrcabbage972:
## TLDR

This PR introduces new telemetry to monitor the health of streaming responses from the Gemini model.
## Dive Deeper

**New Metrics Added**: Three new counters have been implemented in metrics.ts:
1. *gemini.chat.invalid_chunk.count*: Tracks each time an invalid chunk is received from a stream.
2. *gemini.chat.content_retry.count*: Tracks each time a retry is triggered due to an EmptyStreamError.
3. *gemini.chat.content_retry_failure.count*: Tracks when all content retries for a single prompt have been exhausted and the operation fails.

**Implementation**: The metric recording functions are now called from the stream processing and retry logic within geminiChat.ts.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | 👍   |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs
- Resolves #6890


packages/core/src/core/geminiChat.ts
```diff
@@ -373,6 +377,7 @@ export class GeminiChat {
             if (isContentError) {
               // Check if we have more attempts left.
               if (attempt < INVALID_CONTENT_RETRY_OPTIONS.maxAttempts - 1) {
+                recordContentRetry(self.config);
```
cornmander, line 380: not blocking, but this seems noisy. can we record this at the end of the loop once with the total number of retries?

packages/core/src/telemetry/metrics.ts
```diff
@@ -210,7 +234,6 @@ export function recordFileOperationMetric(
   mimetype?: string,
   extension?: string,
   diffStat?: DiffStat,
-  programming_language?: string,
```
cornmander: intentional?

packages/core/src/core/geminiChat.ts
mrcabbage972: What would be the issue? It's only up to 3 events, with a big retry.

packages/core/src/telemetry/metrics.ts
mrcabbage972: Bad merge. Fixed.

packages/core/src/core/geminiChat.ts
cornmander: I guess on one hand this approach is better because it reduces the chance that we miss a retry if the user exits before we hit the end. On the other hand I think it makes our analytics harder to understand because it's not clear what you can aggregate on to see the total number of retries for a single generation attempt. Maybe you can add that as a field? Also not blocking.

cornmander (approved):

PR #6875, opened 2025-08-22
Change the type of ToolResult.responseParts

scidomino:
## TLDR

This reduces the need for post-processing which is done in more than one place.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | x  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Contributes to https://github.com/google-gemini/gemini-cli/issues/6528

### DIFF
--- a/packages/core/src/core/turn.ts
+++ b/packages/core/src/core/turn.ts
@@ -5,6 +5,7 @@
  */
 
 import {
+  Part,
   PartListUnion,
   GenerateContentResponse,
   FunctionCall,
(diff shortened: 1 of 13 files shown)


packages/core/src/core/turn.ts
```diff
@@ -5,6 +5,7 @@
  */
 
 import {
+  Part,
   PartListUnion,
```
cornmander, line 9: Still needed?

cornmander (approved):

packages/core/src/core/turn.ts
scidomino: Yes. It's being passed a PartListUnion by the client in the run method.

PR #6878, opened 2025-08-22
[extensions] Add extensions update command

### LINKED ISSUE #5993: Command: `gemini extensions update`

chrstnb:
## TLDR

Add an `update` command that will update extensions installed via github repo or from a directory to their latest version.

If the update fails, the current version will remain as-is.

This requires an extension to be installed via the `install` extension or to have the installation metadata in .gemini-extension.install.json.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes #5993


packages/cli/src/config/extension.ts
```diff
@@ -380,3 +385,45 @@ export function toOutputString(extension: Extension): string {
...
+
+export async function updateExtension(
+  extensionName: string,
+): Promise<ExtensionUpdateInfo | undefined> {
+  const installedExtensions = loadUserExtensions();
+  const extension = installedExtensions.find(
+    (installed) => installed.config.name === extensionName,
+  );
+  if (!extension) {
+    throw new Error(`Error: Extension "${extensionName}" not found.`);
```
evanotero: nit: I think you should omit the `Error: ` part

See https://github.com/search?q=repo%3Agoogle-gemini%2Fgemini-cli+%22new+Error%28%22&type=code

evanotero: A UX improvement here would be to list out all of the extensions a user has installed that they could update. It's possible the user tries to update the extension using the `source` they used during installation. This would be a great time to introduce them to the `gemini extensions list` command and help them understand what extensions are already installed.

packages/cli/src/config/extension.ts
```diff
@@ -380,3 +385,45 @@ export function toOutputString(extension: Extension): string {
...
+  }
+  if (!extension.installMetadata) {
+    throw new Error(
+      `Extension cannot be updated. To update manually, uninstall and then reinstall the updated version.`,
```
evanotero: Let's be specific for the error, like: "Extension cannot be updated due to missing `.gemini-extension.install.json` file."

packages/cli/src/config/extension.ts
```diff
@@ -380,3 +385,45 @@ export function toOutputString(extension: Extension): string {
...
+      throw new Error('Updated extension not found after installation.');
+    }
+    const updatedVersion = updatedExtension.config.version;
+    return {
+      originalVersion,
+      updatedVersion,
+    };
+  } catch (e) {
+    console.error(
+      `Warning: error updating extension, rolling back. Error: ${e}`,
```
evanotero: Nit: Let's not use the word "warning" since this was a process termination.

`Failed to update extension. Rolling back to previous version. Error: ${e}`

packages/cli/src/commands/extensions/update.ts
```diff
@@ -0,0 +1,45 @@
...
+
+export async function handleUpdate(args: UpdateArgs) {
+  try {
+    const updatedExtensionInfo = await updateExtension(args.name);
+    if (!updatedExtensionInfo) {
+      console.log(`Extension "${args.name}" failed to update.`);
+      return;
+    }
+    console.log(
+      `Extension "${args.name}" successfully updated from version ${updatedExtensionInfo.originalVersion} to version ${updatedExtensionInfo.updatedVersion}.`,
```
evanotero: You should shorten this if you wanted to:

`Extension "${args.name}" successfully updated: ${updatedExtensionInfo.originalVersion} → ${updatedExtensionInfo.updatedVersion}.

evanotero (approved):

packages/cli/src/commands/extensions/update.ts
```diff
@@ -0,0 +1,45 @@
...
+    );
+  } catch (error) {
+    console.error((error as Error).message);
+    process.exit(1);
+  }
+}
+
+export const updateCommand: CommandModule = {
+  command: 'update <name>',
```
evanotero, line 32: What do you think about adding a flag "--all,-a" to update all extensions? If not in this PR, we should create a tracking feature request

packages/cli/src/config/extension.ts
chrstnb: that's a good idea! for now I'm going to suggest that the user run that command. Will make some more thorough UX improvements in follow-ups. Need to clean up some of the existing code around listing and loading extensions, so once I've done that I'll be able to handle that logic in the command itself (a check for extension existence by name, then a list if non-existent)

packages/cli/src/commands/extensions/update.ts
chrstnb: added an issue!

cornmander (approved):

PR #6922, opened 2025-08-23
fix(vscode): target originating window for diff display

### LINKED ISSUE #6848: Diff opened in wrong VS Code window when multiple instances are running
### What happened?

I am running two separate instances of VS Code. When I initiate a code change request in one VS Code window while the other window is currently active (focused), Gemini CLI attempts to apply the changes to the focused window instead of the one that initiated the request.

I expected Gemini CLI to apply the code change to the specific VS Code window where the request was originally made, regardless of which window is currently in focus.

### Client information

<details>

```console
$ gemini /about
# paste output here
```

</details>

### Login information

_No response_

_No response_

Louis454545:
## Summary

- Fixed VS Code extension targeting wrong window when showing diffs with multiple VS Code instances running
- Added explicit `viewColumn: vscode.ViewColumn.Active` parameter to `vscode.diff` command  
- Ensures diffs always open in the originating VS Code window instead of the currently focused one

## Test plan

- [x] Verified TypeScript compilation passes
- [x] Verified ESLint passes  
- [x] Verified build completes successfully
- [ ] Manual testing with multiple VS Code instances (requires reviewer testing)

Fixes google-gemini/gemini-cli#6848

### DIFF
--- a/packages/vscode-ide-companion/src/diff-manager.ts
+++ b/packages/vscode-ide-companion/src/diff-manager.ts
@@ -122,6 +122,7 @@ export class DiffManager {
       {
         preview: false,
         preserveFocus: true,
+        viewColumn: vscode.ViewColumn.Active,
       },
     );
     await vscode.commands.executeCommand(


skeshive:
Thanks for submitting this PR. After some digging, I found the root cause of the issue: our current implementation makes an incorrect assumption about process IDs (PIDs). We had assumed that each new VS Code window would have its own PID, but my manual testing shows they actually share the same one.

This is causing a bug where a new window overwrites the server port for the previous one, and the diff gets opened in the wrong window. The fix for this is more involved and will require refactoring how we write the server port to the file.

I'm going to close this PR and open a new one with the correct fix.

Issue #4544, opened 2025-07-20
Performance: Slow startup time (8-12 seconds) for Gemini CLI


wildcatDownstairs:
### What happened?

When I launch the gemini-cli from my terminal, there is a significant delay of approximately 8 to 12 seconds before the tool becomes responsive and ready for use. This happens every time I start the application.

It would be helpful for you to add details about your system here, for example:

OS: [Windows 11]
Terminal: [cmd]
gemini-cli version: [v0.1.13]

https://github.com/user-attachments/assets/2f940428-cacb-4883-9043-3f31e5f5b88c

### What did you expect to happen?

I expected the command-line interface to launch almost instantly, or at least with a much shorter delay (e.g., under 2 seconds). A quick startup is typical for most CLI tools and would lead to a more fluid and efficient user experience.

### Client information

<details>

```console
$ gemini /about
# paste output here
```

</details>

Manamama:
Many MCPs that call home. Today I decided to remove half of the lot for the same reason. The worst on mobile (slow) connections. Check for the `npx` ones that use online servers.

helalsoft:
No MCP, still this slow.

jacob314:
We should not be blocking startup while MCP servers initialize. MCP servers should be initialized in the background. For discussion: should we block issuing prompts in Gemini CLI while MCP servers initialize.

NTaylorMullen:
Goal here would be to bootup immediately but upon first prompt block if needed

jacob314:
Moving this text to the main issue description for clarity.

jakemac53:
One question, which is also related to adding support for tool update notifications:

Do we know how tool definitions affect input token caching? Presumably, they might completely invalidate the cache causing the next request to be more expensive once the MCP tools are loaded?

If this is the case, is this something we should try and optimize around or just accept it? Most likely, the context won't be super large by the time the MCP servers do load anyways, so the cost of invalidating the context should hopefully be small. One exception would be if somebody resumes a previous chat that already has lots of context.

jakemac53:
Oh, I missed the part about blocking the first prompt until the tools are all loaded. That would resolve the issue around context cache churn (although it's worth chatting about when we go to implement tool list updates).

scidomino:
One difficulty will be not breaking oauth support which currently assumes you're not operating in react.

gennaro-tedesco:
[...]

PR #6923, opened 2025-08-23
Add MCP loading indicator when initializing Gemini CLI

### LINKED ISSUE #3638: Add MCP loading indicator

When starting Gemini CLI, you may see "Waiting for auth" for a long time.

You may think this is because of auth, but most of the time is actually spent connecting to MCP servers.

And if an MCP connection fails, you won't necessarily notice because there doesn't seem to be good error reporting.

I suggest making the message more granular to better report the various states the tool is in.

For example:

```
"Waiting for auth...."

"Connecting to MCP servers (1/3)..."

"Connecting to MCP servers (2/3)..."
```

Maybe even mention the names of the servers currently being tried (happens in parallel)

### Why is this needed?

This improves the user experience as you get much better progress and error information.

### Additional context

Related:

* #1835
* #2046

swissspidy:
## TLDR

Improves UX when there are many MCP servers to connect with during initialization, by displaying a loading indicator.

## Dive Deeper

Demo:

https://github.com/user-attachments/assets/83127e66-ecf4-4ea0-b2b0-3b47716a5665

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ✅  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes #3638


jacob314 (approved):
Like the use of eventEmitter for this. Approving so we can build on top of this change. 
Generally wouldn't approve without tests but want to build on top of this change to work on resolving https://github.com/google-gemini/gemini-cli/issues/4544.
Generally think we should go further than just surfacing that MCP servers are loading and instead make the core UI functional when MCP servers are loading with the limitation that you cannot submit a prompt until mcp servers are loaded.

swissspidy:
> Generally think we should go further than just surfacing that MCP servers are loading and instead make the core UI functional when MCP servers are loading with the limitation that you cannot submit a prompt until mcp servers are loaded.

Agree that this would be a good next step 👍

PR #6941, opened 2025-08-23
fix: misaligned right border on tool calls ui and spacing in multiple tool calls ui

bulkypanda:
## TLDR

- fix previously misaligned right border in successful tool calls
- add space between multiple tool calls in the same box
- the new checkmark is also same type as the x on the failed tool calls

## Dive Deeper
Before:
before
After:
after

## Reviewer Test Plan

- make a successful tool calls and notice that the right margin is properly aligned.
- do something that has two tool calls together and notice the spacing isn't congested

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ✅  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

### DIFF
--- a/packages/cli/src/ui/components/messages/ToolGroupMessage.tsx
+++ b/packages/cli/src/ui/components/messages/ToolGroupMessage.tsx
@@ -81,7 +81,7 @@ export const ToolGroupMessage: React.FC<ToolGroupMessageProps> = ({
       borderDimColor={hasPending}
       borderColor={borderColor}
     >
-      {toolCalls.map((tool) => {
+      {toolCalls.map((tool, index) => {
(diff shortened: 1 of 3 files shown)


jacob314 (approved):
We should follow up by adding a golden snapshot text to ToolGroupMessage.tsx 
A test for that would have hopefully helped us notice the regression.

As another follow up, I'm curious about if there is a better alternative to the square root symbol but I haven't figured out one yet that renders consistently in monospace fonts.

PR #6961, opened 2025-08-24
Added the strictSchemeValidation setting for MCP server parameters. 

### LINKED ISSUE #6632: Strict JSON schema validation rule in MCP tools
### What happened?

MCP servers with tools where the JSON schema doesn't specify a type are skipped during gemini loading, even though it works perfectly.

n8n deploys similar tools, and it's not a problem for others cli

### Client information

╭────────────────────────────────────────────────────────────────────────────────╮
│                                                                                │
│ About Gemini CLI                                                               │
│                                                                                │
│ CLI Version                0.1.22                                              │
│ Git Commit                 5fe4e023                                            │
│ Model                      gemini-2.5-pro                                      │
│ Sandbox                    no sandbox                                          │
│ OS                         linux                                               │
│ Auth Method                OAuth                                               │
│                                                                                │
╰────────────────────────────────────────────────────────────────────────────────╯

### Login information

_No response_

IN packages/core/src/tools/mcp-client.ts, line 570, function hasValidTypes
+++++++
**THIS CHECK COULD BE COMMENTED ?**
  if (!s['type']) {
    // These keywords contain an array of schemas that should be validated.
    //
    // If no top level type was given, then they must each have a type.
    let hasSubSchema = false;
    const schemaArrayKeywords = ['anyOf', 'allOf', 'oneOf'];
    for (const keyword of schemaArrayKeywords) {
      const subSchemas = s[keyword];
      if (Array.isArray(subSchemas)) {
        hasSubSchema = true;
        for (const subSchema of subSchemas) {
          if (!hasValidTypes(subSchema)) {
            return false;
          }
        }
      }
    }
    // If the node itself is missing a type and had no subschemas, then it isn't valid.
    if (!hasSubSchema) return false;
  }
+++++++

I tried it on a local fork, and it works.

blouflab:
Added the strictSchemeValidation setting for MCP server parameters. 

Defaults to false; if true, it enforces strict validation rules for the tools (current rules)

strictSchemeValidation: 
- **undefined, false :** it's the default in the [MCP Protocol](https://modelcontextprotocol.io/docs/getting-started/intro).
- **true :** if a subSchema has not types, it skips the tool (current policy)

`
    "my_mcp": {
      "url": "https://xxx",
      "strictSchemeValidation" : false/true
    }
`
## Linked issues / bugs
Resolves #6632

### DIFF
--- a/packages/core/src/tools/mcp-client.ts
+++ b/packages/core/src/tools/mcp-client.ts
@@ -564,13 +564,14 @@ export async function connectAndDiscover(
  *
  * @visiblefortesting
  */
-export function hasValidTypes(schema: unknown): boolean {
+export function hasValidTypes(schema: unknown, mcpServerConfig?: MCPServerConfig): boolean {
   if (typeof schema !== 'object' || schema === null) {
     // Not a schema object we can validate, or not a schema at all.
     // Treat as valid as it has no properties to be invalid.
     return true;
   }
-
+  //
+  const useStrictValidation = mcpServerConfig?.strictSchemeValidation ?? false;
   const s = schema as Record<string, unknown>;
 
   if (!s['type']) {
@@ -584,7 +585,7 @@ export function hasValidTypes(schema: unknown): boolean {
       if (Array.isArray(subSchemas)) {
         hasSubSchema = true;
         for (const subSchema of subSchemas) {
-          if (!hasValidTypes(subSchema)) {
+          if (!hasValidTypes(subSchema,mcpServerConfig )) {
             return false;
           }
         }
@@ -592,13 +593,14 @@ export function hasValidTypes(schema: unknown): boolean {
     }
 
     // If the node itself is missing a type and had no subschemas, then it isn't valid.
-    if (!hasSubSchema) return false;
+    // It won't skip a tool because there's no "type" in a subschema.
+    if (!hasSubSchema && useStrictValidation) return false;
   }
 
   if (s['type'] === 'object' && s['properties']) {
     if (typeof s['properties'] === 'object' && s['properties'] !== null) {
       for (const prop of Object.values(s['properties'])) {
-        if (!hasValidTypes(prop)) {
+        if (!hasValidTypes(prop,mcpServerConfig)) {
           return false;
         }
       }
@@ -606,7 +608,7 @@ export function hasValidTypes(schema: unknown): boolean {
   }
 
   if (s['type'] === 'array' && s['items']) {
-    if (!hasValidTypes(s['items'])) {
+    if (!hasValidTypes(s['items'], mcpServerConfig)) {
       return false;
     }
   }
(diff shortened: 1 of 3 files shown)


packages/core/src/tools/mcp-tool.ts
```diff
@@ -180,6 +180,7 @@ export class DiscoveredMCPTool extends BaseDeclarativeTool<
     override readonly parameterSchema: unknown,
     readonly timeout?: number,
     readonly trust?: boolean,
+    readonly strictSchemeValidation?: boolean,
```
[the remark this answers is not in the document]
jacob314: please consider this concern from the bot and comment on whether it is valid and then I will review.

jacob314:
Also please sign the CLA before I can review your code

packages/core/src/tools/mcp-client.ts
```diff
@@ -564,13 +564,14 @@ export async function connectAndDiscover(
...
-export function hasValidTypes(schema: unknown): boolean {
+export function hasValidTypes(schema: unknown, mcpServerConfig?: MCPServerConfig): boolean {
   if (typeof schema !== 'object' || schema === null) {
     // Not a schema object we can validate, or not a schema at all.
     // Treat as valid as it has no properties to be invalid.
     return true;
   }
-
+  //
+  const useStrictValidation = mcpServerConfig?.strictSchemeValidation ?? false;
```
wanlin31, line 574: Came from the issue you created, https://github.com/google-gemini/gemini-cli/issues/6632.
Just curious, from a behavior change perspective, would it be safer to default to strict checking and loosen up the check with a deliberate setting.

wanlin31: We don't have this flag before. We are changing from:
```
 if (!hasSubSchema) return false;
 ```
to 
```
if (!hasSubSchema && useStrictValidation) return false;
```
So before is equivalent to `useStrictValidation` always set to true, setting the default to `true` is more consistent with the original behavior.

Comment (blouflab) [reply packages/core/src/tools/mcp-client.ts:574]: Default is false. In that case it uses the default of the Mcp protocol which is to not skip a tool because there is no type in a subschema.
True is the same of what you have actually, which is not a default of the protocol itself and the cause of issue related.

Comment (blouflab) [reply packages/core/src/tools/mcp-client.ts:574]: > ```
> if (!hasSubSchema && useStrictValidation) return false;
> ```
> 
> So before is equivalent to `useStrictValidation` always set to true, setting the default to `true` is more consistent with the original behavior.

That's where it hurts. It's not the original behavior, this check has been added to makes it compliant with 1 aws mcp. See the threads for reference.
This is a question of both common sense and ethics. Is it Google's place to redefine the standard for the MCP protocol?

File: packages/core/src/tools/mcp-client.ts [hunk 3]
```diff
@@ -646,7 +648,7 @@ export async function discoverTools(
           continue;
         }
 
-        if (!hasValidTypes(funcDecl.parametersJsonSchema)) {
+        if (!hasValidTypes(funcDecl.parametersJsonSchema,mcpServerConfig )) {
```
jakemac53, line 651: We should add something to the warning here mentioning the configuration option to loosen the checks so people can be notified of the way to disable the check if they so choose.

PR #6969, opened 2025-08-24
Fix/2967 gemini md import bug

### LINKED ISSUE #2967: Import processor regex matches code inside markdown code blocks
## Description
The import processor uses a regex pattern that's too broad and incorrectly matches `@` patterns inside markdown code blocks, causing false warnings when processing markdown files containing code examples.

## Steps to Reproduce
1. Create a markdown file containing code examples with `@` symbols (e.g., Python decorators)
2. Include the file in a gemini session (e.g., via GEMINI.md)
3. Observe false warnings in the console

## Current Behavior
When processing markdown files with code examples like:
````markdown
```python
@app.get("/api/items")
async def get_items():
    return []
```
````

The import processor generates warnings:
```
[WARN] [ImportProcessor] Import processor only supports .md files. Attempting to import non-md file: app.get("/api/items. This will fail.
```

## Expected Behavior
The import processor should ignore content inside markdown code blocks (both inline `` ` `` and fenced `` ``` `` blocks) and only process actual import statements in regular markdown content.

## Root Cause
**File:** `/packages/core/src/utils/memoryImportProcessor.ts`  

The current regex pattern:
```typescript
const importRegex = /@([./]?[^\s\n]+\.[^\s\n]+)/g;
```

This regex matches any `@` followed by a file-like pattern, without considering markdown syntax boundaries.

### Client information

<details>

```console
$ gemini /about
│ CLI Version                    0.1.9                                                     │
│ Git Commit                     34935d6                                                   │
│ Model                          gemini-2.5-pro                                            │
│ Sandbox                        no sandbox                                                │
│ OS                             darwin                                                    │
│ Auth Method                    vertex-ai                                                 │
```

</details>

### Login information

_No response_

_No response_

ksprashu:
## TLDR

This pull request refines the import parsing logic within the GEMINI.md context processor to prevent false positives and allow the special case of subdirectory imports with file extensions to prevent any regressions.

## Dive Deeper

  The core of this pull request is a balancing act. On one hand, we need to be flexible enough to recognize valid import paths, even when they don't use the explicit ./ or ../ prefixes. On the other hand, we need to be strict enough to avoid misinterpreting other common string formats as import paths.

  The previous logic was too permissive, leading to the false positives mentioned in the main description. The initial fix for this was too restrictive, causing the regression with subdirectory imports. This pull request finds a middle ground by introducing a new rule: a path can contain a / if and only if the final segment of the path contains a file extension.

  This heuristic is based on the assumption that valid file imports will almost always include a file extension, while other formats (like npm package scopes) will not. This allows us to support the subdirectory import syntax without reopening the door to the original false positives.

  The new test cases in memoryImportProcessor.test.ts are crucial for understanding the nuances of this change. They cover the following scenarios:

   * Valid Imports:
       * `@./test.md`
       * `@../test.md`
       * `@/test.md`
       * `@test.md`
       * `@sub/test.md`
   * Invalid Imports (False Positives):
       * `@angular/core`
       * `@my-decorator`
       * `user@example.com`

  By verifying these cases, we can be confident that the new logic is both correct and robust.

## Reviewer Test Plan

  A reviewer can validate these changes by following these steps:

   1. Pull down the branch and run the full test suite to ensure that all existing tests pass:
 
    `npm run preflight`

   4. Examine the new tests in `packages/core/src/utils/memoryImportProcessor.test.ts`. These tests are the best way to understand the intended behavior of the new import parsing logic. Pay close attention to the tests that cover the false positive cases and the subdirectory import fix.

   5. Run the CLI with a GEMINI.md file that contains a variety of `@import` statements, both valid and invalid. For example:

    `GEMINI.md`

> @./test-data/test.md
> @test-data/test.md
> @angular/core
> user@example.com

      `test-data/test.md`

> This is a test file.

  When you run the CLI with this GEMINI.md file, you should see that the contents of `test-data/test.md` are imported twice, and the `@angular/core` and `user@example.com` lines are ignored.

  6. Test with a real-world example. If you have a project that uses the @import syntax in its GEMINI.md file, try running the CLI with that project to ensure that the changes don't introduce any new issues.

  By following these steps, a reviewer can be confident that the new import parsing logic is working as expected.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | Y  | ❓  | ❓  |
| npx      | Y  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes #2967


ksprashu:
Current PR was too far behind HEAD. 
I've redone the changes on a new rebased branch in PR - #10554
Closing this PR.

PR #6996, opened 2025-08-25
refactor(core): async config initialization flow

skeshive:
## TLDR

This PR refactors the core `Config` class to use an asynchronous initialization pattern. The `new Config()` constructor and separate `config.initialize()` method have been replaced with a single static `await Config.create()` method.

## Dive Deeper

The previous synchronous constructor for `Config` followed by an async `initialize()` method created a state where a `Config` object could exist but not be fully operational. This refactoring to a static async `create` method ensures that any `Config` instance is always fully constructed and initialized.

A key consequence of this change is that services requiring authentication, like `GeminiClient`, are no longer guaranteed to be present on the `Config` object at all times. `config.getGeminiClient()` (and other related getters like `getToolRegistry`) can now return `undefined`.

The bulk of the changes in this PR involve adding the necessary checks across the codebase to handle this possibility, ensuring that we don't try to access services before they are available.


skeshive:
Closing this as https://github.com/google-gemini/gemini-cli/pull/7627 removes the ideClient!

Issue #5992, opened 2025-08-11
Command: `gemini extensions enable/disable`


NTaylorMullen:
This task involves creating the `gemini extensions enable <name>` and `gemini extensions disable <name>` commands. These commands will provide an intuitive way for users to control the activation state of installed extensions by modifying the `settings.json` file.

### Acceptance Criteria

1.  **Command Implementation:**
    *   Create a new command: `gemini extensions enable <name>`.
    *   Create a new command: `gemini extensions disable <name>`.
    *   Both commands require the `<name>` argument.
    *   Both commands should include a `--global` flag to target the user-level configuration.

2.  **Configuration Scope:**
    *   By default (without the `--global` flag), the commands must modify the **project-level** `settings.json` at `<workspace>/.gemini/settings.json`.
    *   If the `--global` flag is present, the commands must modify the **user-level** `settings.json` at `~/.gemini/settings.json`.
    *   If a `settings.json` file does not exist at the target scope, the command should create it.

3.  **Disable Logic (`disable`):**
    *   This command adds the specified extension `<name>` to the `extensions.disabled` array in the target `settings.json`.
    *   If the `extensions.disabled` array does not exist, it should be created.
    *   If the name is already in the array, the command should do nothing and succeed.

4.  **Enable Logic (`enable`):**
    *   This command removes the specified extension `<name>` from the `extensions.disabled` array in the target `settings.json`.
    *   If the name is not in the array, the command should do nothing and succeed.
    *   If the `extensions.disabled` array becomes empty after the removal, it should be removed from the `settings.json` file to keep it clean.

**Parent Issue:** #5987

PR #7001, opened 2025-08-25
[extensions] Add disable command

chrstnb:
## TLDR

Add the `disable` command, with an optional scope. This will disable the command in the user or workspace settings (system settings not yet supported pending work on system-level extensions). 

I need to do some additional work to clean up the extension loading utilities in config/extension.ts, but for now this is set up so that disabled extensions will not show up in the UI, but will still be listed in `extensions list`. In a follow-up, I'd like to thread through the status of the extension and where it is disabled. 

## Reviewer Test Plan

```
npm run build && npm start extensions disable example-extension -- --scope=Workspace
```

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

#5992


packages/cli/src/commands/extensions/disable.ts
```diff
@@ -0,0 +1,50 @@
...
+}
+
+export async function handleDisable(args: DisableArgs) {
+  try {
+    disableExtension(args.name, args.scope);
+    console.log(
+      `Extension "${args.name}" successfully disabled for scope "${args.scope}".`,
+    );
+  } catch (error) {
+    console.error((error as Error).message);
```
NTaylorMullen: There should be a getErrorMessage helper

packages/cli/src/config/extension.ts
```diff
@@ -428,3 +440,38 @@ export async function updateExtension(
     await fs.promises.rm(tempDir, { recursive: true, force: true });
   }
 }
+
+export function disableExtension(name: string, scope: SettingScope) {
+  if (scope === SettingScope.System || scope === SettingScope.SystemDefaults) {
+    throw new Error('System and SystemDefaults scopes are not supported.');
+  }
+  const settings = loadSettings(process.cwd());
```
NTaylorMullen: Shouldn't assume process.cwd(0 is where we're running, I belive config has a method for that

packages/cli/src/config/settingsSchema.ts
```diff
@@ -544,6 +544,26 @@ export const SETTINGS_SCHEMA = {
     description: 'Enable extension management features.',
     showInDialog: false,
   },
+  extensions: {
```
NTaylorMullen, line 547: fyi @galz10

NTaylorMullen (approved):

packages/cli/src/config/extension.ts
chrstnb: Replied to your comment on the other PR-- the config method is in core config and would require for us to load the config before we load commands and thread it through that way (I think?). I will create a task to clean this up in a follow-up, but note that we use process.cwd() throughout the cli/ codebase currently. For now I parameterized it with process.cwd() as the default and it looks like something you can set dynamically, but I'm not sure how one would do that for the extension commands as-is (though definitely relevant if we expose this other ways).

chrstnb: https://github.com/google-gemini/gemini-cli/issues/7069

packages/cli/src/commands/extensions/disable.ts
```diff
@@ -0,0 +1,51 @@
...
+      `Extension "${args.name}" successfully disabled for scope "${args.scope}".`,
+    );
+  } catch (error) {
+    console.error(getErrorMessage(error));
+    process.exit(1);
+  }
+}
+
+export const disableCommand: CommandModule = {
+  command: 'disable <name>',
```
cornmander: add the flag argument disable [--scope <scope>] \<name\>

packages/cli/src/config/extension.test.ts
```diff
@@ -438,3 +491,55 @@ describe('updateExtension', () => {
...
+
+  it('should disable an extension at the workspace scope', () => {
+    disableExtension('my-extension', SettingScope.Workspace);
+    const settings = loadSettings(tempWorkspaceDir);
+    expect(
+      settings.forScope(SettingScope.Workspace).settings.extensions?.disabled,
+    ).toEqual(['my-extension']);
+  });
+
+  it('should not add the same extension twice', () => {
```
cornmander: disable?

cornmander (approved):

PR #7002, opened 2025-08-25
fix: Enable disableFuzzySearch config option propagation

brychanrobot:
This change fixes the propagation of the disableFuzzySearch configuration option from the CLI settings to the core file search utility. Previously, although the option was introduced, it was not correctly passed down, rendering it ineffective.

With this fix, the disableFuzzySearch setting now correctly influences the file search behavior, allowing users to disable fuzzy search as intended. Additionally, the AsyncFzf instance is now conditionally initialized based on this setting, further optimizing resource usage when fuzzy search is disabled.

## Reviewer Test Plan

Try out disabling fuzzy search

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | 🐧  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


jacob314 (approved):
Can you add a test verifying this option is propagated as expected?lgtm

PR #7035, opened 2025-08-25
[extensions] Add an initial set of extension variables 

### LINKED ISSUE #7015: Implement initial extension variable set
https://code.visualstudio.com/docs/reference/variables-reference

To start, we want to support initializing mcp servers using something like:  `"cwd": "${geminiExtensionsDir}${/}${geminiExtensionName}${/}mcp-server-dir"`, so we'll want:

|variable name           |description                                                              |
|------------------------|-------------------------------------------------------------------------|
|${extensionName}        |The name of the extension (which should also map to the parent directory}|
|${extensionsDir}        |The extensions directory the extension is stored in                      |
|${/} or ${pathSeparator}|The path separator (differs per OS)                                      |

chrstnb:
## TLDR

Inspired by https://code.visualstudio.com/docs/reference/variables-reference and to support requests for filling the "cwd' to run local MCP servers.

Adding an initial set of variables extensions can set anywhere in the `gemini-extension.json` values. Currently just the extension path and the path separator.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes #7015


packages/cli/src/config/extensions/variableSchema.ts
```diff
@@ -0,0 +1,31 @@
...
+}
+
+export interface VariableSchema {
+  [key: string]: VariableDefinition;
+}
+
+export const VARIABLE_SCHEMA = {
+  extensionPath: {
+    type: 'string',
+    description: 'The path of the extension in the filesystem.',
```
allenhutchison, line 26: My assumption reading this is that I would get a fully qualified absolute path (i.e. "/Users/adh/.gemini/extensions/foo-exention/") and that it would unwrap symlinks and such. It might be good to add additional data to the definition to make it clear what kind of path you are getting here.

packages/cli/src/config/extensions/variableSchema.ts
```diff
@@ -0,0 +1,31 @@
...
+  },
+  '/': {
```
allenhutchison: For these two, did you consider defining a shared part that could be assigned to both so that they are always consistent in the future?

const PATH_SEPARATOR_DEFINITION = {
  type: 'string',
  description: 'The path separator.',
} as const;

Then in later:

  '/': PATH_SEPARATOR_DEFINITION,
  pathSeparator: PATH_SEPARATOR_DEFINITION,

packages/cli/src/commands/extensions/update.ts
```diff
@@ -0,0 +1,46 @@
+/**
```
allenhutchison: I don't understand why update.ts is a part of this change. Was it included by accident or is there something I'm not understanding?

chrstnb: not intended-- had a big stack that has since been merged!

docs/extension.md
```diff
@@ -87,3 +87,14 @@ You can install extensions using the `install` command. This command allows you
 
 - `source <url> positional argument`: The URL of a Git repository to install the extension from. The repository must contain a `gemini-extension.json` file in its root.
 - `--path <path>`: The path to a local directory to install as an extension. The directory must contain a `gemini-extension.json` file.
+
+# Variables
+
+Gemini CLI extensions allow variable substitution in `gemini-extension.json`. This can be useful if e.g., you need the current directory to run an MCP server using `"cwd": "${extensionsDir}${/}${extensionFolder}${/}run.ts"`.
```
cornmander: This references variables that aren't in the table below

packages/cli/src/config/extensions/variables.ts
```diff
@@ -0,0 +1,57 @@
...
+  validateVariables(context, VARIABLE_SCHEMA);
+  const regex = /\${(.*?)}/g;
+  return str.replace(regex, (match, key) =>
+    context[key as keyof VariableContext] == null
+      ? match
+      : (context[key as keyof VariableContext] as string),
+  );
+}
+
+export function recursivelyHydrateStrings(
```
cornmander, line 45: This makes me nervous. Although JSON doesn't support circular references this function takes an arbitrary object. Can you make it only take a JSON object?

packages/cli/src/config/extensions/variableSchema.ts
cornmander, line 1: It would be nice to generate the documentation table from this schema. Could be a nice starter bug for an external contributor.

cornmander (approved):

packages/cli/src/config/extensions/variableSchema.ts
chrstnb: good idea https://github.com/google-gemini/gemini-cli/issues/7072

packages/cli/src/config/extensions/variables.ts
chrstnb: done

PR #7037, opened 2025-08-25
feat: add golden snapshot test for ToolGroupMessage and improve success symbol

bulkypanda:
## TLDR

- Add comprehensive golden snapshot tests for ToolGroupMessage.tsx with 14 test cases
- Replace √ (square root) with ✓ (check mark) for better semantic clarity and monospace font compatibility
- Test coverage includes multiple tool statuses, border colors, confirmation handling, and edge cases
- Updated all related tests to use the new check mark symbol
- Fixes would help catch visual regressions mentioned in PR #6941 

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ✅  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

### DIFF
--- a/packages/cli/src/ui/components/messages/ToolGroupMessage.test.tsx
+++ b/packages/cli/src/ui/components/messages/ToolGroupMessage.test.tsx
@@ -1,0 +1,344 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { render } from 'ink-testing-library';
+import { describe, it, expect, vi } from 'vitest';
+import { Text } from 'ink';
+import { ToolGroupMessage } from './ToolGroupMessage.js';
+import { IndividualToolCallDisplay, ToolCallStatus } from '../../types.js';
+import { Config, ToolCallConfirmationDetails } from '@google/gemini-cli-core';
+
+// Mock child components to isolate ToolGroupMessage behavior
+vi.mock('./ToolMessage.js', () => ({
+  ToolMessage: function MockToolMessage({
+    callId,
+    name,
+    description,
+    status,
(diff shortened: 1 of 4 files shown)


packages/cli/src/ui/components/messages/ToolGroupMessage.test.tsx
```diff
@@ -0,0 +1,344 @@
...
+    emphasis,
+  }: {
+    callId: string;
+    name: string;
+    description: string;
+    status: ToolCallStatus;
+    emphasis: string;
+  }) {
+    const statusSymbol = {
+      [ToolCallStatus.Success]: '✓',
```
jacob314, line 30: nit: not super useful if these are mocked. Can we at least tweak so these are constants used in the real tests that we use here. Otherwise they will drift out of sync and we will wonder why the tests pass but the rendering is bad.

packages/cli/src/ui/components/messages/ToolMessage.test.tsx
```diff
@@ -71,19 +71,19 @@ describe('<ToolMessage />', () => {
       StreamingState.Idle,
     );
     const output = lastFrame();
-    expect(output).toContain('√'); // Success indicator
+    expect(output).toContain('✓'); // Success indicator
```
jacob314, line 74: this is better and avoids the rendering glitches

jacob314 (commented):
lgtm

jacob314 (approved):
lgtm
lets tweak the golden tests in a follow up. Approving to unblock reviewing the next pr in the chain

PR #7041, opened 2025-08-25
feat: Disable YOLO and AUTO_EDIT modes for untrusted folders

shrutip90:
## TLDR
For untrusted folders, disable auto approval modes when used in CLI flags and don't allow users to enable them during runtime. Also avoid showing "approve always" option for tool calls, edit ops, etc.

## Dive Deeper

NOTE that when the trust level is changed, since we force restart the CLI, I haven't added approval mode overriding in that path.

**Screenshots:**
Warning shown when CLI flag is overriden:
image

Info message shown when you try to modify the approval mode in interactive shell:
image

Not showing proceed always option for untrusted folders:
image

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

### DIFF
--- a/packages/cli/src/config/config.test.ts
+++ b/packages/cli/src/config/config.test.ts
@@ -1868,28 +1868,28 @@ describe('loadCliConfig approval mode', () => {
 
     it('should override --approval-mode=yolo to DEFAULT', async () => {
       process.argv = ['node', 'script.js', '--approval-mode', 'yolo'];
-      const argv = await parseArguments();
+      const argv = await parseArguments({} as Settings);
       const config = await loadCliConfig({}, [], 'test-session', argv);
       expect(config.getApprovalMode()).toBe(ServerConfig.ApprovalMode.DEFAULT);
     });
 
     it('should override --approval-mode=auto_edit to DEFAULT', async () => {
       process.argv = ['node', 'script.js', '--approval-mode', 'auto_edit'];
-      const argv = await parseArguments();
+      const argv = await parseArguments({} as Settings);
       const config = await loadCliConfig({}, [], 'test-session', argv);
       expect(config.getApprovalMode()).toBe(ServerConfig.ApprovalMode.DEFAULT);
     });
 
     it('should override --yolo flag to DEFAULT', async () => {
       process.argv = ['node', 'script.js', '--yolo'];
-      const argv = await parseArguments();
+      const argv = await parseArguments({} as Settings);
       const config = await loadCliConfig({}, [], 'test-session', argv);
       expect(config.getApprovalMode()).toBe(ServerConfig.ApprovalMode.DEFAULT);
     });
 
     it('should remain DEFAULT when --approval-mode=default', async () => {
       process.argv = ['node', 'script.js', '--approval-mode', 'default'];
-      const argv = await parseArguments();
+      const argv = await parseArguments({} as Settings);
       const config = await loadCliConfig({}, [], 'test-session', argv);
       expect(config.getApprovalMode()).toBe(ServerConfig.ApprovalMode.DEFAULT);
     });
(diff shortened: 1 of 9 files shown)


packages/cli/src/config/config.test.ts
```diff
@@ -1848,6 +1851,41 @@ describe('loadCliConfig approval mode', () => {
...
+      expect(config.getApprovalMode()).toBe(ServerConfig.ApprovalMode.DEFAULT);
+    });
+
+    it('should remain DEFAULT when --approval-mode=default', async () => {
+      process.argv = ['node', 'script.js', '--approval-mode', 'default'];
+      const argv = await parseArguments();
+      const config = await loadCliConfig({}, [], 'test-session', argv);
+      expect(config.getApprovalMode()).toBe(ServerConfig.ApprovalMode.DEFAULT);
+    });
+  });
```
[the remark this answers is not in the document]
shrutip90: Done

packages/cli/src/config/config.ts
```diff
@@ -423,6 +423,18 @@ export async function loadCliConfig(
       argv.yolo || false ? ApprovalMode.YOLO : ApprovalMode.DEFAULT;
   }
 
+  // Force approval mode to default if the folder is not trusted.
+  if (
+    !trustedFolder &&
+    (approvalMode === ApprovalMode.YOLO ||
```
jacob314: nit: why not use
approvalMode !== ApprovalMode.DEFAULT instead here so you don't miss a case in the future... for example
ApprovalMode.ALMOST_YOLO

packages/cli/src/ui/hooks/useAutoAcceptIndicator.ts
```diff
@@ -40,6 +43,20 @@ export function useAutoAcceptIndicator({
...
+          addItem(
+            {
+              type: MessageType.INFO,
+              text: 'Cannot enable privileged approval modes in an untrusted folder.',
+            },
+            Date.now(),
+          );
+          return;
+        }
         config.setApprovalMode(nextApprovalMode);
```
jacob314: optional nit: can we make this extra robust? Have
setApprovalMode fail when setting to a mode that is not allowed due to config.isTrustedFolder()?
You could have it throw an exception which you catch or return an error object describing why the operation was not allowed.

jacob314 (commented):
Looks good. A couple minor comments to make this more robust. Lets make config.setApprovalMode fail when it isn't allowed. I believe there is at least one other place we change the approval mode so changing that API will be more robust to ensure users can never get into an unsafe approval mode when they shouldn't.

packages/cli/src/ui/hooks/useAutoAcceptIndicator.ts
shrutip90: Thanks, done.

packages/cli/src/config/config.ts
shrutip90: Done

packages/core/src/config/config.ts
```diff
@@ -566,6 +566,11 @@ export class Config {
   }
 
   setApprovalMode(mode: ApprovalMode): void {
+    if (!this.isTrustedFolder() && mode !== ApprovalMode.DEFAULT) {
```
jacob314, line 567: this is called in edit.ts and a few other places. Please verify that all those places work correctly. Likely a few of them need to be fixed by removing the option from the list of radio buttons to "approve always"
if you search for `setApprovalMode` in the project you will find all the missing cases.

shrutip90: Thanks, checked that there are 3 other places in core - write_file.ts, edit.ts and web-fetch.ts that all call setApprovalMode with Auto edit in case of PROCEED_ALWAYS option selection.

Updated ToolConfirmationMessage to skip this option for untrusted folders. While I am here, I also skipped it in other cases like exec tool calls and mcp tool calls, though these paths do not call setApprovalMode.

Added a screenshot with the updated options in the description.

jacob314 (approved):
lgtm

PR #4404, opened 2025-07-17
fix(core): Improve compression message clarity for small history cases

ShammiAnand:
## TLDR
Fixes confusing compression message when chat history compression is skipped for small histories. Previously showed "compressed from 50 to 75 tokens" (misleading), now shows "Skipping compression for small history as the process would have increased its size" (clear).

## Dive Deeper
The compression feature automatically compresses chat history when it approaches token limits. However, for very small histories, the compression process itself can actually increase the token count due to overhead. In these cases, compression is skipped, but the UI was still showing the generic "compressed from X to Y tokens" message where Y &gt; X, which confused users into thinking compression was broken.

This change detects when `newTokenCount >= originalTokenCount` (indicating compression was skipped) and displays a clearer message explaining why compression was skipped. Also improved code readability by destructuring props early and using safer null handling.

## Reviewer Test Plan
1. **Test normal compression**: Create a long chat history and trigger compression - should show normal "compressed from X to Y tokens" message
2. **Test skipped compression**: Create a very short chat history and trigger compression - should show new "Skipping compression..." message
3. **Test edge cases**: Try with undefined/null token counts to ensure graceful handling

Fixes #3533

PR #7047, opened 2025-08-25
fix(compression): Discard compression result if it results in more token usage

richieforeman:
## TLDR

Refuse to use the new "compressed" context if the compression result has more tokens that the source.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | x  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

This is the first part of [3533](https://github.com/google-gemini/gemini-cli/issues/3533) and probably the least controversial thing that should just be shipped 🚢


abhipatel12:
This will kick of a compression event again in the next user prompt. Is this the intended behavior?

But also, we can create a hard stop on the output using the `maxOutputTokens` in the config right? That way we are not waiting for the entire output to happen and then discarding it

richieforeman:
> This will kick of a compression event again in the next user prompt. Is this the intended behavior?
> 
> But also, we can create a hard stop on the output using the `maxOutputTokens` in the config right? That way we are not waiting for the entire output to happen and then discarding it

I am not observing that, can you help me understand more?

image

I like the idea with maxOutputTokens, i'll fiddle with that and verify.

packages/core/src/core/client.ts
```diff
@@ -858,6 +858,11 @@ export class GeminiClient {
       }),
     );
 
+    if (newTokenCount > originalTokenCount) {
```
jacob314, line 892: see https://github.com/google-gemini/gemini-cli/pull/4404/files#diff-7208dcd4859a7d57f72d738dfa8da3004e83c9048b8011fa475e4a9a389ea3d4
which tackles this showing a user visible message when things get wrong. I think we need a combination of your pr and that pr. 
Your pr is right to make changes to client but doesn't go quite far enough as it appears we still keep the badly compressed result even though it was larger than the original.

jacob314: Does it really discard the compression result? Line 829 looks like we are still keeping the compression result we are just not reporting stats about what the compression result was.

jacob314: Gemini Code Assist can you take a look and reply to my question?

abhipatel12:
> > This will kick of a compression event again in the next user prompt. Is this the intended behavior?
> > But also, we can create a hard stop on the output using the `maxOutputTokens` in the config right? That way we are not waiting for the entire output to happen and then discarding it
> 
> I am not observing that, can you help me understand more?
> 
> image
> I like the idea with maxOutputTokens, i'll fiddle with that and verify.

I had thought that if the token count didn't decrease, the same conditional would pick it up on the next user input where we check the token count. This would cause another compression attempt which may succeed but could also fail since we're not changing any system instructions. 

I'm mostly referring to when the context size increases at the 80% point.

packages/core/src/core/client.ts
richieforeman: Good catch, updated and added a bunch of tests around this.

richieforeman:
> > > This will kick of a compression event again in the next user prompt. Is this the intended behavior?
> > > But also, we can create a hard stop on the output using the `maxOutputTokens` in the config right? That way we are not waiting for the entire output to happen and then discarding it
> > 
> > 
> > I am not observing that, can you help me understand more?
> > image
> > I like the idea with maxOutputTokens, i'll fiddle with that and verify.
> 
> I had thought that if the token count didn't decrease, the same conditional would pick it up on the next user input where we check the token count. This would cause another compression attempt which may succeed but could also fail since we're not changing any system instructions.
> 
> I'm mostly referring to when the context size increases at the 80% point.

That's a great point, I've addressed this in the CL

packages/core/src/core/client.ts
```diff
@@ -857,6 +867,14 @@ export class GeminiClient {
...
+      this.hasFailedCompressionAttempt = !force && true;
+      console.warn('Compression incorrectly inflated the token count.');
```
jacob314: can we continue to return the original and new token count and add an extra bool indicating that the request was skipped. this will allow the CLI package to show a meaningful message explaining why the request was skipped which is more useful then logging to the console. in fact the pr for such a message has already landed :)

jacob314 (commented):
Approved after this small comment is addressed.

packages/core/src/core/client.ts
richieforeman: I don't think that #4404 has landed yet.  I have wired up a new enum through the command execution, I think this will custom messaging very easy to do/handle.  Depending, it may make sense to break NOOP into further things, but I'll leave that for #4404 if they handle it.

packages/core/src/core/client.ts
```diff
@@ -821,11 +843,12 @@ export class GeminiClient {
         },
         config: {
           systemInstruction: { text: getCompressionPrompt() },
+          maxOutputTokens: originalTokenCount,
         },
       },
       prompt_id,
     );
-    this.chat = await this.startChat([
+    const chat = await this.startChat([
```
abhipatel12, line 854: A few lines above we do:

```typescript
const historyToCompress = curatedHistory.slice(0, compressBeforeIndex);
const historyToKeep = curatedHistory.slice(compressBeforeIndex);
this.getChat().setHistory(historyToCompress);
```

If the compression is rejected later due to token or some other error, wouldn't this mean that we fail to restore the original history and now only have a sliced history?

packages/core/src/core/client.ts
```diff
@@ -846,7 +869,13 @@ export class GeminiClient {
       });
```
abhipatel12, line 872: I can't click the proper line b/c that's just how this review tool works 😤

But one line above, we do `contents: this.getChat().getHistory(),` to count the tokens. Should we be counting the tokens in the proposed new chat? Which would be:

```typescript
contents: chat.getHistory(), 
```

packages/core/src/core/client.ts
```diff
@@ -764,12 +772,16 @@ export class GeminiClient {
...
     prompt_id: string,
     force: boolean = false,
-  ): Promise<ChatCompressionInfo | null> {
+  ): Promise<ChatCompressionInfo> {
     const curatedHistory = this.getChat().getHistory(true);
 
     // Regardless of `force`, don't do anything if the history is empty.
-    if (curatedHistory.length === 0) {
-      return null;
+    if (curatedHistory.length === 0 || this.hasFailedCompressionAttempt) {
```
abhipatel12: Does this mean a user can't force a `/compress` if a previous auto attempt occurred?

packages/core/src/core/client.ts
```diff
@@ -781,7 +793,13 @@ export class GeminiClient {
       });
     if (originalTokenCount === undefined) {
       console.warn(`Could not determine token count for model ${model}.`);
-      return null;
+      this.hasFailedCompressionAttempt = !force && true;
```
abhipatel12, line 799: In this case, if force is true and a failure happens, are we meaning to reset this?

This could be correct, but just a little confused on what this line does. I actually see this exact line 3 times, would it be possible to add a comment for clarity or extract it into a more readable function?

packages/core/src/core/client.ts
```diff
@@ -821,11 +843,12 @@ export class GeminiClient {
         },
         config: {
           systemInstruction: { text: getCompressionPrompt() },
+          maxOutputTokens: originalTokenCount,
```
abhipatel12, line 849: We might be able to optimize a little bit here. If we make the output the same size as the original token count, we could end up with a larger total b/c we still have to append the `historyToKeep`. 

Wdyt?

packages/cli/src/ui/commands/compressCommand.ts
```diff
@@ -50,6 +51,7 @@ export const compressCommand: SlashCommand = {
               isPending: false,
               originalTokenCount: compressed.originalTokenCount,
               newTokenCount: compressed.newTokenCount,
+              compressionStatus: compressed.compressionStatus,
```
abhipatel12, line 54: Now that `ChatCompressInfo` is now never `null`, the `if (compressed)` is no longer needed, and the else block is now dead code.

abhipatel12 (commented):
Thanks for taking this on! Left a few comments! Lmk what you think!

packages/cli/src/ui/commands/compressCommand.ts
richieforeman: It is not `nullable`, but it is potentially undefined because of the optional chaining above.

packages/core/src/core/client.ts
richieforeman: Fixed!

packages/core/src/core/client.ts
richieforeman: Updated; Added tests.

packages/core/src/core/client.ts
richieforeman: I added a docblock comment to the property definition.

packages/core/src/core/client.ts
richieforeman: I think we could potentially compute a "target output token count", but that feels potentially out of scope for this change.

I think there is a case here where we do compress the context and then discard it, but optimizing this further feels like something that should come in a separate wave.

abhipatel12: Sounds good! In general, it would be nice to move compression out of this file all together to make it easier to follow.

packages/cli/src/ui/commands/compressCommand.ts
abhipatel12: Ohh makes sense. Thanks for clarifying!

packages/core/src/core/client.ts
richieforeman: Agree, It should be a different class, it feels kludgy here.

packages/core/src/core/client.ts
richieforeman: I've updated this. This is not possible to test without a larger refactor of the tests.  Currently, the tests are actually testing a mock of some of the bits of Client.  In many ways, these tests are actually mocking the unit-under-test, which is very harmful and dangerous.  I can explore making these tests test the _actual_ implementation in a follow up, but I fear it will be a more involved refactor of the tests as a whole.

abhipatel12: Sg thanks for looking into it!

abhipatel12 (approved):
LGTM!

jacob314 (approved):
lgtm

PR #7058, opened 2025-08-25
fix: resolve three flaky tests

bulkypanda:
## TLDR

Fixed three flaky tests causing CI failures: (1) useAtCompletion test replaced unreliable waitFor with with fake timers for precise 200ms timing control, (2) App.test.tsx update notification tests refactored to use proper waitFor() calls, (3) InputPrompt reverse search test wrapped stdin operations in act() to handle React state updates.

## Dive Deeper

- Fix 1 (useAtCompletion): Uses fake timers (vi.useFakeTimers() + vi.advanceTimersByTime(200)) to precisely control time, making the test faster (~140ms), deterministic across all environments, and actually validating the intended 200ms timing logic rather than just "waiting long enough."

- Fix 2 (App handleAutoUpdate): Replaces all fixed delays with waitFor() calls that return immediately when conditions are met, making tests both faster and more resilient to timing variations while providing better error messages when failures occur.

- Fix 3 (InputPrompt reverse search): The most complex issue involving dozens of "not wrapped in act()" warnings and 5000ms timeouts. The solution wraps stdin.write() operations in act() calls to ensure React properly batches and completes state updates before assertions, eliminating both the timeout and warnings.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ✅  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

- Resolves #7031

### DIFF
--- a/packages/cli/src/ui/App.test.tsx
+++ b/packages/cli/src/ui/App.test.tsx
@@ -6,6 +6,7 @@
 
 import type { Mock } from 'vitest';
 import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+import { waitFor } from '@testing-library/react';
 import { renderWithProviders } from '../test-utils/render.js';
 import { AppWrapper as App } from './App.js';
 import type {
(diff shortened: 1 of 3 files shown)


cornmander (approved):
awesome! thanks for fixing this.

PR #7073, opened 2025-08-26
Treat undefined same as true for isTrustedFolder

shrutip90:
## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

### DIFF
--- a/packages/cli/src/ui/components/messages/ToolConfirmationMessage.tsx
+++ b/packages/cli/src/ui/components/messages/ToolConfirmationMessage.tsx
@@ -56,6 +56,8 @@ export const ToolConfirmationMessage: React.FC<
     onConfirm(outcome);
   };
 
+  const isTrustedFolder = config?.isTrustedFolder() !== false;
+
   useKeypress(
     (key) => {
       if (!isFocused) return;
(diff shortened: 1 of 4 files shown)


packages/cli/src/ui/components/messages/ToolConfirmationMessage.tsx
```diff
@@ -56,6 +56,8 @@ export const ToolConfirmationMessage: React.FC<
     onConfirm(outcome);
   };
 
+  const isTrustedFolder = config?.isTrustedFolder() !== false;
```
jacob314, line 59: we should probably make config be non-nullable to simplify this and avoid bugs when we accidentally forget to pass in a config.

jacob314 (approved):
lgtm

packages/cli/src/ui/components/messages/ToolConfirmationMessage.tsx
shrutip90: Ok, will check and send a separate PR for that.

PR #7130, opened 2025-08-26
fix(ci): Fix self assign action permissions

davideast:
## TLDR
Self assign workflow was broken due to a type in an unneeded permission.

## Dive Deeper
This workflow should only need the `permission-issues: 'write'` issue to execute.

## Reviewer Test Plan
N/A

## Testing Matrix
N/A


jacob314 (approved):
lgtm

Issue #6897, opened 2025-08-22
Repeated failed tool calls


leehagoodjames:
### What happened?

`gemini-cli` will repeatedly fail to call a tool that does not exist. This has been observed with the GitHub MCP, although the issue may affect all tool calls - that is unclear, when `gemini-cli` wrongly appends the MCP server name to the tool name, such as calling `github.list_issues` instead of `list_issues`.

Example:
<img width="983" height="873" alt="Image" src="https://github.com/user-attachments/assets/38a82500-99a5-41bd-9825-7c3d06951a19" />

_Note: This does not appear to be looping behavior, as an inspection of the various attempts shows slight alterations from the model, even trying workarounds with other tools_

_Note: It seems that this issue is related to the way that the tool errors are surface to Gemini. See below how even a highly useful error does not cause gemini-cli to self correct, but copy/pasting that error as a prompt yields the self-correction._

<img width="886" height="1108" alt="Image" src="https://github.com/user-attachments/assets/2cd06963-0a81-47ff-b2ef-946182767d34" />

### What did you expect to happen?

I expect the model to self correct when it calls a tool that does not exist

### Client information

<details>

```console
╭─────────────────────────────────────────────────────────╮
│                                                         │
│ About Gemini CLI                                        │
│                                                         │
│ CLI Version        0.1.22                               │
│ Git Commit         5fe4e023                             │
│ Model              gemini-2.5-pro                       │
│ Sandbox            no sandbox                           │
│ OS                 darwin                               │
│ Auth Method        gemini-api-key                       │
│                                                         │
╰─────────────────────────────────────────────────────────╯
```

</details>

PR #7133, opened 2025-08-26
feat(errors): Make errors more informative

leehagoodjames:
## TLDR

This PR addresses one of the two issues causing #6897.

## Dive Deeper

Two issues are causing #6897:
1. The error messages provided to the Gemini API are not informative enough (this PR addresses that)
1. There is a structural issue in the errors provided to the Gemini API where `id` is passed even when it should not be. A follow-up PR will fix that

## Reviewer Test Plan

Get the Gemini API to call a tool that does not exist and inspect the error message. 

Before:
image

After:
image

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

- Related to #6897

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -33,12 +33,14 @@
     "@opentelemetry/sdk-node": "^0.203.0",
     "@types/glob": "^8.1.0",
     "@types/html-to-text": "^9.0.4",
+    "@xterm/headless": "5.5.0",
     "ajv": "^8.17.1",
-    "fast-uri": "^3.0.6",
     "ajv-formats": "^3.0.0",
     "chardet": "^2.1.0",
     "diff": "^7.0.0",
     "dotenv": "^17.1.0",
+    "fast-levenshtein": "^2.0.6",
+    "fast-uri": "^3.0.6",
     "fdir": "^6.4.6",
     "fzf": "^0.5.2",
     "glob": "^10.4.5",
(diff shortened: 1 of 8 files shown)


packages/core/src/core/coreToolScheduler.ts
```diff
@@ -485,6 +486,40 @@ export class CoreToolScheduler {
...
+    const allToolNames = this.toolRegistry.getAllToolNames();
+
+    const matches = allToolNames.map((toolName) => ({
+      name: toolName,
+      distance: leven(unknownToolName, toolName),
+    }));
+
+    matches.sort((a, b) => a.distance - b.distance);
+
+    const topNResults = matches.slice(0, topN);
```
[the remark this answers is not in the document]
leehagoodjames: I disagree. The core issue that this PR seeks to resolve is the Gemini API incorrectly prefixing MCP tool names with their server name, ex: `github.list_issues` vs `list_issues`. Using an arbitrary threshold could make this solution less robust for MCP server names that long.

Instead, this solution relies on the global minimum edit distance, to show the _most similar_ tools across all configured tools. This approach should be more resilient to arbitrarily long MCP server names

jerop (approved):
this would be so useful for self healing when the model gets the tools wrong 

cc @jacob314 @NTaylorMullen

packages/core/src/core/coreToolScheduler.ts
```diff
@@ -557,12 +592,14 @@ export class CoreToolScheduler {
         (reqInfo): ToolCall => {
           const toolInstance = this.toolRegistry.getTool(reqInfo.name);
           if (!toolInstance) {
+            const suggestion = this.getToolSuggestion(reqInfo.name);
+            const errorMessage = `Tool "${reqInfo.name}" not found in registry. Tools must use the exact names that are registered, without prefixes.${suggestion}`;
```
scidomino: I'm concerned that this is overly tailored to the specific mcp issue you are looking at. Are we sure this issue is widespread enough to warrant warning about that here?

jerop: how about this
```
Tool "${reqInfo.name}" not found in registry. Tools must use the exact names that are registered.${suggestion}
```

packages/core/package.json
```diff
@@ -46,6 +46,7 @@
     "html-to-text": "^9.0.5",
     "https-proxy-agent": "^7.0.6",
     "ignore": "^7.0.0",
```
scidomino: Do a search to verify that there is no equivalent functionality in any of the existing dependencies (or standard node).

scidomino (approved):

packages/core/src/core/coreToolScheduler.ts
leehagoodjames: Good call, updated to remove any mention of `prefix` using @jerop's suggestion

packages/core/package.json
leehagoodjames: Switched to `fast-levenshtein` which is a dependency of `optionator`

PR #7136, opened 2025-08-26
Support installing extensions to 'system' location

### LINKED ISSUE #5988: Core: Implement Hierarchical Extension & Configuration Loading
This task is a foundational part of the new extension management system. It involves implementing the core logic for discovering installed extensions and applying configuration from multiple scopes to determine which extensions are active.

### Acceptance Criteria

1.  **Extension Discovery:**
    *   The CLI must scan for installed extensions in two locations:
        1.  A new system-level extensions directory.
        2.  The existing user-level extensions directory (`~/.gemini/extensions`).
    *   It should build a complete list of all available extensions from both locations. If an extension with the same name exists in both, the user-level one takes precedence.

2.  **Configuration Loading:**
    *   The CLI must read the `settings.json` file from two locations:
        1.  The user's home directory (`~/.gemini/settings.json`).
        2.  The current workspace/project directory (`<workspace>/.gemini/settings.json`).
    *   It must look for an `extensions.disabled` property, which is an array of strings representing the names of disabled extensions.

3.  **Activation Logic:**
    *   An extension is considered **active** by default if it is discovered.
    *   An extension is **inactive** if its name is present in the final `extensions.disabled` array.
    *   The project-level `settings.json` has the highest precedence. If it contains an `extensions.disabled` list, that list is used.
    *   If the project-level `settings.json` does not exist or does not contain `extensions.disabled`, the user-level `settings.json` is used as the fallback.

4.  **Final Extension Set:**
    *   The outcome of this logic should be a final, definitive set of active extensions that will be loaded by the CLI for the current session.

This task focuses solely on the backend loading and configuration logic. No CLI commands will be created here, but this work is essential for all other extension management features.

**Parent Issue:** #5987

chrstnb:
## TLDR

This change introduces the ability to manage extensions in two different locations: `user` and `system`.

- **`install`**: The `install` command now accepts a `--location` flag, which can be set to either `user` or `system`. The default is `user`.
- **`uninstall`**: The `uninstall` command also accepts a `--location` flag to specify where torremo.
- **`update`**: The `update` command has been updated to accept a `--location` flag.
- **`list`**: The `list` command now displays extensions from both `user` and `system` locations, grouped by their location.
- **`enable`/`disable`**: The `enable` and `disable` commands have been updated to correctly handle the new extension locations.

**System extensions will take precedence over user extensions if there are name conflicts.** If you attempt to install a user extension with the same name as an existing system extension, you will receive an error.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes #5988


packages/cli/src/commands/extensions/install.ts
```diff
@@ -8,13 +8,16 @@ import type { CommandModule } from 'yargs';
...
 import { getErrorMessage } from '../../utils/errors.js';
+import { locationOption } from './options.js';
 
 interface InstallArgs {
   source?: string;
   path?: string;
+  location: string;
 }
 
 export async function handleInstall(args: InstallArgs) {
```
ruizhao, line 23: Nit: do we need docs?
/**
 * Handles the extension installation process based on the provided arguments.
 * It determines whether the installation source is a git repository or a local path
 * and calls the appropriate installation function.
 * Logs success or error messages to the console.
 * @param args The arguments for the install command.
 * @throws Will re-throw any error encountered during the installation process
 * after logging an error message.
 */

packages/cli/src/commands/extensions/list.ts
```diff
@@ -5,21 +5,43 @@
...
-        .map((extension, _): string => toOutputString(extension))
-        .join('\n\n'),
-    );
+
+    if (userExtensions.length > 0) {
+      console.log('User Extensions:');
+      console.log(
+        userExtensions
+          .map((extension) => toOutputString(extension))
+          .join('\n\n'),
```
ruizhao, line 30: really minor nit: should we put .join('\n\n') into toOutputString() function?

packages/cli/src/commands/extensions/list.ts
```diff
@@ -5,21 +5,43 @@
...
+      console.log('System Extensions:');
+      console.log(
+        systemExtensions
+          .map((extension) => toOutputString(extension))
+          .join('\n\n'),
+      );
+    }
   } catch (error) {
     console.error(getErrorMessage(error));
     process.exit(1);
```
ruizhao: The error handling among those commands are not consistent, in install, we log and throw error back, but in uninstall we exit(1), is this intentional?

packages/cli/src/config/extension.ts
```diff
@@ -47,16 +51,30 @@ export interface ExtensionUpdateInfo {
...
 
-  constructor(extensionName: string) {
+  constructor(extensionName: string, location: InstallLocation) {
     this.extensionName = extensionName;
+    this.location = location;
+    const extensionsDir =
+      location === InstallLocation.System
+        ? ExtensionStorage.getSystemExtensionsDir()
+        : ExtensionStorage.getUserExtensionsDir();
+    fs.mkdirSync(extensionsDir, { recursive: true });
```
ruizhao, line 70: nit: InstallExtension function is async, which is nice, should we also make all the file operation async to free up the event loop?

packages/cli/src/commands/extensions/install.ts
chrstnb: yes good call!

packages/cli/src/commands/extensions/list.ts
chrstnb: I think no so we don't have trailing newlines, but open to push-back!

packages/cli/src/config/extension.ts
chrstnb: The reason I did it this way is because it's in a constructor

packages/cli/src/commands/extensions/disable.ts
```diff
@@ -39,7 +38,7 @@ export const disableCommand: CommandModule = {
         describe: 'The scope to disable the extenison in.',
         type: 'string',
         default: SettingScope.User,
-        choices: [SettingScope.User, SettingScope.Workspace],
+        choices: Object.values(SettingScope),
```
richieforeman, line 41: Concern: Javascript does not have native enums and Typescript sort of shims them into the language by having an object.

For number based enums, this has some error states.

Take this:
https://www.typescriptlang.org/play/?#code/KYOwrgtgBAgiCWECGAbKBvAUFKARA8gOIA02UAwjACqk4AS+ASgMoCipAvptwMYD2IAM58UwAHQo+AcwAU+AEYArYDwAuYgG6owwQTLiJUASkxA

The result of this is:
image

That's a bit of a problem.

For string enums, there's a similar problem:

https://www.typescriptlang.org/play/?#code/KYOwrgtgBAgiCWECGAbKBvAUFKARA8gOJQC8UARAYedlAMIwAqpFDj5ANLQBL4BKAZQCiLcr0FDOmAL6Y5AYwD2IAM6KUwAHQpFAcwAU+AEYArYPIAumgG6owwFfriJUASkxA

image


For this reason, I don't recommend doing `Object.values` on enums in TS, and instead preferring to explicitly list the members.

packages/cli/src/commands/extensions/enable.ts
```diff
@@ -45,9 +45,9 @@ export const enableCommand: CommandModule = {
       })
       .option('scope', {
         describe:
-          'The scope to enable the extenison in. If not set, will be enabled in all scopes.',
+          'The scope to enable the extenison in. If not set, will be enabled in user and workspace scopes.',
         type: 'string',
-        choices: [SettingScope.User, SettingScope.Workspace],
+        choices: Object.values(SettingScope),
```
richieforeman, line 50: Same concern here as above.

packages/cli/src/commands/extensions/uninstall.ts
```diff
@@ -5,32 +5,38 @@
...
 
 interface UninstallArgs {
   name: string;
+  location: string;
 }
 
 export async function handleUninstall(args: UninstallArgs) {
   try {
-    await uninstallExtension(args.name);
+    const location =
```
richieforeman, line 19: I noticed you have this same logic in a couple of places, it might potentially make sense to extract into a helper func.

packages/cli/src/config/extension.test.ts
```diff
@@ -62,6 +76,10 @@ describe('loadExtensions', () => {
     tempHomeDir = fs.mkdtempSync(
       path.join(os.tmpdir(), 'gemini-cli-test-home-'),
     );
+    tempSystemDir = fs.mkdtempSync(
```
richieforeman, line 79: We should make sure this dir is cleaned up in the `afterEach`, can you add a cleanup in there with the others?

packages/cli/src/config/extension.test.ts
```diff
@@ -166,6 +184,24 @@ describe('loadExtensions', () => {
...
+    const userExtensionsDir = path.join(tempHomeDir, EXTENSIONS_DIRECTORY_NAME);
+    fs.mkdirSync(userExtensionsDir, { recursive: true });
+    createExtension(userExtensionsDir, 'test-extension', '1.0.0');
+
+    const systemExtensionsDir = path.join(
+      tempSystemDir,
+      EXTENSIONS_DIRECTORY_NAME,
+    );
+    fs.mkdirSync(systemExtensionsDir, { recursive: true });
+    createExtension(systemExtensionsDir, 'test-extension', '2.0.0');
```
richieforeman, line 197: nit: Potentially for future, don't feel strongly in the scope of this cl, consider updating this `createExtension` function to take an object bag.  Having a func that takes a whole bunch of strings (even for a test helper) is ripe for a mixup.

ref: https://g3doc.corp.google.com/javascript/typescript/g3doc/dev/practices/call-site-friendly-parameters.md?cl=head

packages/cli/src/config/extension.test.ts
```diff
@@ -277,17 +313,21 @@ describe('annotateActiveExtensions', () => {
...
   let userExtensionsDir: string;
+  let tempSystemDir: string;
 
   beforeEach(() => {
     tempHomeDir = fs.mkdtempSync(
       path.join(os.tmpdir(), 'gemini-cli-test-home-'),
     );
     vi.mocked(os.homedir).mockReturnValue(tempHomeDir);
     userExtensionsDir = path.join(tempHomeDir, '.gemini', 'extensions');
+    tempSystemDir = fs.mkdtempSync(
```
richieforeman, line 324: Likewise, be sure to tidy this dir as well.

packages/cli/src/config/extension.test.ts
```diff
@@ -344,6 +396,31 @@ describe('installExtension', () => {
     expect(fs.existsSync(targetExtDir)).toBe(false);
   });
 
+  it('should throw an error if installing a user extension that already exists as a system extension', async () => {
+    const systemExtensionsDir = path.join(
+      tempSystemDir,
+      '.gemini',
```
richieforeman, line 402: Use `GEMINI_DIR` from `@google/gemini-cli-core`

chrstnb:

PR #7162, opened 2025-08-26
feat(cli): Add support for Ctrl+Backspace to delete a word backward

### LINKED ISSUE #6547: i can't delete words backwards with Ctrl+Backspace only forward with Ctrl+Del which is annoying
### What happened?

The Ctrl+Backspace didn't deleted a whole word in the chat input area.

I expected to delete a word backward with Ctrl+Backspace but this feature is not implemented yet.

### Client information

* **CLI Version:** 0.1.22
* **Git Commit:** 5fe4e023
* **Session ID:** 19897d62-d581-42e1-864f-32c9b9ce9e33
* **Operating System:** linux v22.17.0
* **Memory Usage:** 199.0 MB

### Login information

_No response_

_No response_

davideast:
## TLDR
Adds the ability to delete a word backwards with `Ctrl + Backspace`.

## Reviewer Test Plan
I tested a local build in a Linux environment and tests were added in:
- `packages/cli/src/ui/contexts/KeypressContext.test.tsx`
- `packages/cli/src/ui/contexts/KeypressContext.test.tsx`
- `packages/cli/src/ui/keyMatchers.test.ts`

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  |  ✅ |
| npx      | ❓  | ❓  | ✅ |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs
Resolves #6547

### DIFF
--- a/packages/cli/src/config/keyBindings.ts
+++ b/packages/cli/src/config/keyBindings.ts
@@ -90,23 +90,16 @@ export type KeyBindingConfig = {
 export const defaultKeyBindings: KeyBindingConfig = {
   // Basic bindings
   [Command.RETURN]: [{ key: 'return' }],
-  // Original: key.name === 'escape'
   [Command.ESCAPE]: [{ key: 'escape' }],
 
   // Cursor movement
-  // Original: key.ctrl && key.name === 'a'
   [Command.HOME]: [{ key: 'a', ctrl: true }],
-  // Original: key.ctrl && key.name === 'e'
   [Command.END]: [{ key: 'e', ctrl: true }],
 
   // Text deletion
-  // Original: key.ctrl && key.name === 'k'
   [Command.KILL_LINE_RIGHT]: [{ key: 'k', ctrl: true }],
-  // Original: key.ctrl && key.name === 'u'
   [Command.KILL_LINE_LEFT]: [{ key: 'u', ctrl: true }],
-  // Original: key.ctrl && key.name === 'c'
   [Command.CLEAR_INPUT]: [{ key: 'c', ctrl: true }],
-  // Original: key.ctrl && key.name === 'backspace'
   // Added command (meta/alt/option) for mac compatibility
   [Command.DELETE_WORD_BACKWARD]: [
     { key: 'backspace', ctrl: true },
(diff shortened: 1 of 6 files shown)


packages/cli/src/config/keyBindings.ts
```diff
@@ -105,6 +106,12 @@ export const defaultKeyBindings: KeyBindingConfig = {
   [Command.KILL_LINE_LEFT]: [{ key: 'u', ctrl: true }],
   // Original: key.ctrl && key.name === 'c'
   [Command.CLEAR_INPUT]: [{ key: 'c', ctrl: true }],
+  // Original: key.ctrl && key.name === 'backspace'
```
jacob314: Generally Gemini CLI is a bit too big a fan of adding comments describing how the code was rather than what it does now. 
The comment about adding meta/alt/option for 
As a follow up PR we should clean up this and the other low value "Original: " comments in this pr.

packages/cli/src/ui/components/shared/text-buffer.ts
```diff
@@ -1299,6 +1300,45 @@ export function textBufferReducer(
       return { ...nextState, lines: newLines, preferredCol: null };
     }
 
+    case 'delete_word_backward': {
```
jacob314: how is this different from deleteWordLeft? Please align with that method rather than creating another one. note that that command already has a keyboard shortcut tied to ctrl backspace which just likely wasn't working on your keyboard without the other fixes you landed this pr.

jacob314 (commented):
Thanks for the fix. will approve after the duplicate deleteWordLeft functionality is reverted.
You hit one of the confusing cases where some logic has gotten into InputPrompt which should have just been in text-buffer making it a bit harder to spot where to add keyboard shortcuts. it would be nice to clean this up and move the keyboard processing that doesn't need to be InputPrompt aware (e.g. know about autocomplete, etc) into the text-buffer.ts layer.
I'd support updating all of text-buffer to use the keybindings. nobody has yet made the time to do that but good news is Gemini CLI should be able to make the task quite mechanical.

packages/cli/src/ui/components/shared/text-buffer.ts
davideast: Yep, you're right there's a lot of duplication. I added logic improves on the word boundary logic, so I moved that into `deleteWordLeft` and removed the `delete_word_backward` function. 

The main difference is that the new logic is much smarter about what it considers a "word." The old version would treat things like `variable_name` or `path/to/file` as a single word. It's also better with unicode and punctuation, so something like `foo!!!` is correctly treated as two words.

Let me know if this improvement is good and if not I can just resort back to the old `deleteWordLeft` logic.

packages/cli/src/ui/components/shared/text-buffer.ts
```diff
@@ -1229,47 +1229,38 @@ export function textBufferReducer(
     case 'delete_word_left': {
       const { cursorRow, cursorCol } = state;
       if (cursorCol === 0 && cursorRow === 0) return state;
-      if (cursorCol === 0) {
+
+      const nextState = pushUndoLocal(state);
```
jacob314, line 1233: please add tests for this changed logic

packages/cli/src/ui/components/shared/text-buffer.ts
```diff
@@ -1229,47 +1229,38 @@ export function textBufferReducer(
     case 'delete_word_left': {
```
jacob314, line 1229: if you are modifying delete_word_left I suspect you also need to modify delete_word_right to be consistent.

packages/cli/src/config/keyBindings.ts
davideast: Done.

packages/cli/src/ui/components/shared/text-buffer.ts
davideast: Done.

packages/cli/src/ui/components/shared/text-buffer.ts
davideast: Done. I also added word boundary tests.

jacob314 (approved):
lgtm

PR #7174, opened 2025-08-27
fix(ui): fix drag-and-drop file feature for macOS terminal 2

### LINKED ISSUE #5729: Adding files to CLI within cursor/vscode.
### What happened?

Often I want to quickly add multiple files to CLI context when using it from within cursor. One way to do this would be to drag the file from the left side panel to the terminal where CLI is operating. However, when I drag and drop files, the file paths are often truncated. This does not happen in Claude Code or just terminal. I am attaching an example below. Only part of the file path is added. 

Image

I expected the whole file path to be added.

### Client information

│ About Gemini CLI                                            │
│                                                             │
│ CLI Version          0.1.18                                 │
│ Git Commit           cd7e60e0                               │
│ Model                gemini-2.5-pro                         │
│ Sandbox              no sandbox                             │
│ OS                   darwin                                 │
│ Auth Method          OAuth   

### Login information

_No response_

_No response_

### LINKED ISSUE #6386: Unable to drag markdown files to CLI within cursor
### What happened?

I can't drag markdown files to CLI.

Image

About Cursor
---
version: 1.4.5
submit: 1.99.3
date: af58d92614edb1f72bdd756615d131bf8dfa5290
Electron: 2025-08-13T02:08:56.371Z
ElectronBuildId: 34.5.8
Chromium: undefined
Node.js: 132.0.6834.210
V8: 20.19.1
OS: 13.2.152.41-electron.0

I hope markdown file can also be dragged to CLI.

### Client information

* **CLI Version:** 0.1.21
* **Git Commit:** 055fe46d
* **Operating System:** darwin v22.15.0
* **Memory Usage:** 89.5 MB

### Login information

_No response_

_No response_

944750720:
## TLDR

Fixes drag-and-drop file functionality for macOS terminal environments. The feature was previously non-functional due to macOS sending focus events followed by individual characters
instead of complete file paths.

## Dive Deeper

- macOS terminals often emit a focus event (ESC[I) and then stream the dragged path character-by-character.
- We no longer depend on the focus event; collection starts on the first quote and aggregates the stream reliably.
- A single completion debounce is used; explicit end markers trigger an immediate flush to guarantee full-path delivery.

Technical Root Cause:

1. macOS drag emits focus + per-character stream (not a complete path).
2. Without robust aggregation, paths were truncated or dropped when input wasn’t focused.
3. Terminals may append explicit end markers, which were not handled consistently.

Solution Implementation:

- Single debounce timer:
    - DRAG_COMPLETION_TIMEOUT_MS re-armed on every character during collection.
- Start condition:
    - Begin collecting on the first quote (SINGLE_QUOTE or DOUBLE_QUOTE); subsequent quotes are treated as regular characters.
- Explicit end markers:
    - Flush immediately when CSI_END_O (ESC[O) or SS3_END (ESC O) arrives; the marker is included in the aggregated sequence.
- Consistent event shape:
    - Broadcast a single event with name: '', paste: true, and the aggregated sequence.

## Reviewer Test Plan

- Drag files from Finder into the CLI:
    - Types: .tsx, .js, .py, .md, .json
    - With spaces: "test file.txt"
    - With apostrophes: "O’Reilly.txt", "it’s-ok.zsh"
    - Long paths: "/Users/username/very/long/path/to/file.tsx"
    - Expected: full path appears in the input without truncation (a trailing terminal marker may be present and is handled downstream).
- Paste verification:
    - Cmd+V text/code; should work on first attempt.
    - Paste must not be impacted by drag detection.
- Edge cases:
    - Press Esc during drag → clean cancel.
npm run build && npm start
## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ✅  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

Note: This fix specifically targets macOS drag behavior. Windows/Linux should be unaffected; light regression tests are recommended.

## Linked Issues

- Fixes #6386
- Fixes #6449
- Fixes #5729


packages/cli/src/ui/contexts/KeypressContext.tsx
```diff
@@ -194,6 +206,64 @@ export function KeypressProvider({
...
+          if (trimmedPath) {
+            broadcast({
+              name: '',
+              ctrl: false,
+              meta: false,
+              shift: false,
+              paste: true,
+              sequence: trimmedPath,
+            });
+          }
```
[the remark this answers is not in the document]
jacob314: agree with this

944750720:
~~@jacob314 I fixed lint error but drag-drop hasn't worked yet. Is [fix(ui): use useRef instead of useState for DragBuffer](https://github.com/google-gemini/gemini-cli/pull/7174/commits/2d2e9d6909d4687b7336f4ce92f0a8f60c39fdaa) right?~~
~~@jacob314  I fixed logic bug by https://github.com/google-gemini/gemini-cli/pull/7174/commits/bfe0d7a57202d08932d03f960afa76d80ba0b880. Please check it when you're free.~~

packages/cli/src/ui/contexts/KeypressContext.tsx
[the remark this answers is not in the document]
944750720: @jacob314 
image
Drag-drop still doesn't work. newBuffer is no problem. But it looks like broadcast was skipped. After this step, I pressed f11 and debug was over. Nothing appears in cli.

packages/cli/src/ui/contexts/KeypressContext.tsx
```diff
@@ -36,6 +36,12 @@ import { FOCUS_IN, FOCUS_OUT } from '../hooks/useFocus.js';
 const ESC = '\u001B';
 export const PASTE_MODE_PREFIX = `${ESC}[200~`;
 export const PASTE_MODE_SUFFIX = `${ESC}[201~`;
+export const MAC_DRAG_MODE_PREFIX = `${ESC}[I`;
+export const MAC_FOCUS_EVENT_OUT_OF_EDITOR = "'"; // When editor does NOT have focus (macOS Finder.app)
+
+export const DRAG_START_TIMEOUT_MS = 200;
```
leehagoodjames: Add comments on how these timeouts were determined to persist context in case they need to be changed later

packages/cli/src/ui/contexts/KeypressContext.tsx
```diff
@@ -194,6 +226,75 @@ export function KeypressProvider({
...
+        !key.paste
+      ) {
+        isDragCollectingRef.current = true;
+        dragBufferRef.current = '';
+        dragBroadcastedRef.current = false;
+
+        clearDragStartTimeout();
+        dragTimeoutRef.current = setTimeout(() => {
+          resetDragState();
+        }, DRAG_START_TIMEOUT_MS);
```
leehagoodjames: Why is this done inside `resetDragState()`

leehagoodjames (commented):
Thanks for making the change, excited to get this in! 

Can you please address them comments. Also, since the fix requires cursor (which I and many reviews may not have) can you please include a video of it working within cursor?

packages/cli/src/ui/contexts/KeypressContext.tsx
944750720: This is to avoid blocking other events like paste. Because before paste, user also needs to focus on terminal so 
```
      if (
        !isDragCollectingRef.current &&
        isDragStartEvent(key.sequence) &&
        !key.paste
      )
```
will be fulfilled. If not resetDragState, user's paste action will be blocked.

944750720: @leehagoodjames Hers's the screen record of drag-drop in cursor.

https://github.com/user-attachments/assets/c2bd1cd1-da71-4e75-92b7-d6b04e8c1453

944750720:
@leehagoodjames I found my previous logic was too complex and I refactored them. I think it looks much better now! Please take a look when convenient.

944750720:
@leehagoodjames @jacob314 I found my previous logic was too complex and I refactored them. I think it looks much better now! Please take a look when convenient.

jacob314 (approved):
lgtm
Great job! @skeshive was just asking about this case not working and it was great to be able to point to your pr fixing it.

PR #7214, opened 2025-08-27
fix: unset GEMINI_API_KEY env var if empty

### LINKED ISSUE #5747: Empty API Key Handling and Input Validation Issue
### What happened?

Empty API keys not being handled gracefully. This is a challenge in systems where missing values default to empty.

I'd expect empty API keys to be handled as though it's absent.

### Client information

```console
 CLI Version       0.1.18                              
 OS                darwin                              
 Auth Method       vertex-ai
```

### Login information

Vertex AI, but with an empty API key

Other environment variables (GOOGLE_API_KEY, GOOGLE_CLOUD_PROJECT, etc.) should have validation too

gsquared94:
## TLDR
If the `GEMINI_API_KEY` env var is set to empty string and Vertex AI env vars are set, then the `GoogleGenAI` client is initialized with `{"apiKey":"","vertexai":true}` which fails to use Vertex AI.
```
[API Error: {"error":{"code":401,"message":"Request is missing required authentication credential. Expected OAuth 2 access token, login cookie or other valid authentication credential. See https://developers.google.com/identity/sign-in/web/devconsole-project.","status":"UNAUTHENTICATED","details":[{"@type":"type.googleapis.com/google.rpc.DebugInfo","detail":"Request is missing required authentication credential. Expected OAuth 2 access token, login cookie or other valid authentication credential. See https://developers.google.com/identity/sign-in/web/devconsole-project."},{"@type":"type.googleapis.com/google.rpc.ErrorInfo","reason":"CREDENTIALS_MISSING","domain":"googleapis.com","metadata":{"method":"google.cloud.aiplatform.v1beta1.PredictionService.CountTokens","service":"aiplatform.googleapis.com"}}]}}]
```
This is causing failures for Github Action runtimes. 

Will submit a fix upstream, but handling it here directly.

Fixes: #5747

### DIFF
--- a/packages/cli/src/gemini.tsx
+++ b/packages/cli/src/gemini.tsx
@@ -232,6 +232,10 @@ export async function main() {
     delete process.env['GEMINI_API_KEY'];
   }
 
+  if (process.env['GOOGLE_API_KEY']?.trim() === '') {
+    delete process.env['GOOGLE_API_KEY'];
+  }
+
   setMaxSizedBoxDebugging(config.getDebugMode());
 
   await config.initialize();
--- a/packages/cli/src/gemini.tsx
+++ b/packages/cli/src/gemini.tsx
@@ -227,6 +227,10 @@ export async function main() {
       );
     }
   }
+  // Empty key causes issues with the GoogleGenAI package.
+  if (process.env['GEMINI_API_KEY']?.trim() === '') {
+    delete process.env['GEMINI_API_KEY'];
+  }
 
   setMaxSizedBoxDebugging(config.getDebugMode());
 


packages/cli/src/gemini.tsx
```diff
@@ -227,6 +227,10 @@ export async function main() {
       );
     }
   }
+  // Empty key causes issues with the GoogleGenAI package.
+  if (process.env['GEMINI_API_KEY']?.trim() === '') {
+    delete process.env['GEMINI_API_KEY'];
+  }
```
jerop, line 233: I think we should do the same for GOOGLE_API_KEY as well; I'd expect it to have the same issue

jerop (commented):
thanks for this fix @gsquared94!

packages/cli/src/gemini.tsx
gsquared94: done

jerop (approved):

PR #7250, opened 2025-08-27
fix(terminal): Prevent garbled output in non-standard terminals

gsquared94:
This commit attempts to address an issue where I occasionally see garbled control characters on iTerm terminal output. Also seen during some test runs like https://github.com/google-gemini/gemini-cli/actions/runs/17275532597/job/49031274715?pr=6728#step:6:64

The two main changes are: 
  1.  **feat(terminal): Make `setWindowTitle` more robust.** The function that sets the terminal window title now checks the `TERM` environment variable and only attempts to set the title if the terminal identifies as `xterm`-compatible. This prevents sending unsupported escape sequences to terminals that can't handle them.
  2.  **fix(test): Disable bracketed paste in tests.** The `useBracketedPaste` hook is now disabled when running under `vitest`. This was the likely cause of the `[?2004h` sequences appearing in test output, as the test runner was not interpreting the escape codes correctly.


jacob314:
Unfortunately this doesn't fix the issue.

PR #7251, opened 2025-08-27
Add `gemini extensions new` command to create boilerplate examples

chrstnb:
# feat(extensions): Add 'new' command

## `gemini extensions new`

The `new` command allows developers to create a new extension from a boilerplate template. This makes it easier to get started with creating new extensions.

### Usage

```
gemini extensions new <path> <template>
```

-   `<path>`: The directory where the new extension will be created.
-   `<template>`: The boilerplate template to use.

Available templates include:
-   `context`: An example of an extension that provides additional context.
-   `custom-commands`: An example of an extension that adds custom commands.
-   `exclude-tools`: An example of an extension that excludes certain tools.
-   `mcp-server`: An example of an extension that runs an MCP server.

-   Updated `scripts/copy_files.js` to include the new example templates in the build.


bdmorgan (approved):
LGTM

PR #7267, opened 2025-08-28
Fix issues where escape codes could end up on startup in the input prompt

### LINKED ISSUE #6906: Gemini CLI sometimes enters the following text in the terminal on startup
### What happened?

Opened Gemini CLI and saw the following in the input prompt when I expected an empty prompt
```\x1b[?64;1;2;4;6;17;18;21;22c```

This is likely due to incorrect handling of the kitty protocol. I'm using iterm2 on mac.

I expected an empty input prompt.

### Client information

<details>

```console
$ gemini /about
# paste output here
```

</details>

### Login information

_No response_

_No response_

jacob314:
## TLDR

Particularly in the presence of slow to startup MCP servers, it was possible to get Gemini CLI too a insert malformed control character sequence both in stdout and even worse in the input prompt itself. To repro the key was to tab out of the terminal and then type keys or do something else to put the computer under high load. 

Screenshot 2025-08-27 at 4 42 59 PM

## Dive Deeper

To solve this we enter raw mode immediately and never leave it to avoid terminal control codes escaping where they shouldn't. This resolves both bugs.
We also needed to move kitty protocol detection earlier to reduce the time before we started kitty protocol detection 
As part of testing also identified that the timeout to detect kitty mode was too aggressive (50ms). Switched to a a longer 200ms with an additonal 1000ms delay if there was a partial response indicating a terminal probably supported the kitty protocol. Measured that it can take iTerm 91ms to indicate it is in kitty protocol mode on a modern laptop.

## Reviewer Test Plan

Open iterm2 or another kitty capable terminal that can reproduce the issue.
Run with an MCP server that takes a few seconds to startup.
```
  "mcpServers": {
    "cloud-run": {
      "command": "npx",
      "args": [
        "-y",
        "github:GoogleCloudPlatform/cloud-run-mcp"
      ]
    }
  },

```

Launch Gemini CLI and tab away typing furiously in the alternate terminal or window you tabbed to.
Go back to Gemini CLI after it has loaded and verify that no spurious characters are in the console or input prompt.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

fixes #6906


jacob314:
I'll follow up with some tests for this.

jacob314:
Fyi @deepankarsharma would be great to have you review this.

packages/cli/src/ui/utils/kittyProtocolDetector.ts
```diff
@@ -32,40 +32,58 @@ export async function detectAndEnableKittyProtocol(): Promise<boolean> {
...
+        timeoutId = undefined;
+        process.stdin.removeListener('data', handleData);
+
+        if (!originalRawMode) {
+          process.stdin.setRawMode(false);
         }
+
+        if (progressiveEnhancementReceived) {
+          // Enable the protocol
+          process.stdout.write('\x1b[>1u');
```
deepankarsharma, line 76: This could be a constant in platformConstants.ts

deepankarsharma (changes requested):
Looks good.

NTaylorMullen (approved):

PR #7245, opened 2025-08-27
Move mockTool into test-utils

adamfweidman:
## TLDR

This fixes a production error introduced in https://github.com/google-gemini/gemini-cli/pull/7228, exporting vitest causes crash in npm run start. 

## Reviewer Test Plan

Confirm the followig work:
`npm run test`
`npm run start`
`npm run start:a2a-server`

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| -------- | --- | --- | --- |
| npm run  | x  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

PR #7277, opened 2025-08-28
Revert "Move mockTool into test-utils (#7245)"

galdawave:
This reverts commit da7901acaf5d75903bde21c12b9911ae15b27fa0.

This adds a circular dependency for test-util is depending on core, and core is depending on test-util.


cornmander (approved):
Confirmed this fixes it

PR #7278, opened 2025-08-28
Refine stream validation to prevent unnecessary retries

mrcabbage972:
## TLDR
The stream processing logic has been updated to prevent unnecessary retries. It now correctly handles streams that are functionally complete but end with an empty part, such as after a tool call or with a valid finishReason (STOP, MAX_TOKENS).

## Dive Deeper
The previous implementation of processStreamResponse was too aggressive in its validation. It would trigger a retry by throwing an EmptyStreamError if any chunk in the stream was considered invalid (e.g., contained an empty text part). This led to incorrect failures in two legitimate scenarios:

A model response contains a tool call and is then followed by an empty final chunk.

A model response streams valid text and then ends with an empty chunk but provides a successful finishReason like STOP.

In both cases, the stream had successfully completed its primary task, and the empty final chunk did not signify a failure that required a retry.

The decision to throw an EmptyStreamError is now governed by the following rules:

If the stream contains an invalid chunk but a tool call was present, the stream is considered successful. The tool call is treated as the intended output, and the empty part is ignored.

If the stream contains an invalid chunk and no tool call was present, it is only considered successful if two conditions are met:
a. The invalid chunk was the absolute last one (validChunkAfterInvalidEncountered === false).
b. The lastChunk has a successful finishReason (e.g., STOP or MAX_TOKENS).

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

### DIFF
--- a/packages/core/src/core/geminiChat.ts
+++ b/packages/core/src/core/geminiChat.ts
@@ -566,7 +566,6 @@ export class GeminiChat {
     let hasToolCall = false;
     let lastChunk: GenerateContentResponse | null = null;
 
-    // --- FIX: Restore the overall status flag and initialize all flags ---
     let isStreamInvalid = false;
     let firstInvalidChunkEncountered = false;
     let validChunkAfterInvalidEncountered = false;
@@ -592,8 +591,7 @@ export class GeminiChat {
         logInvalidChunk(
           this.config,
           new InvalidChunkEvent('Invalid chunk received from stream.'),
-        );
-        // --- FIX: Set both flags for an invalid chunk ---
+        );        
         isStreamInvalid = true;
         firstInvalidChunkEncountered = true;
       }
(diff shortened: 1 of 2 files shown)


packages/core/src/core/geminiChat.ts
```diff
@@ -562,30 +562,67 @@ export class GeminiChat {
...
+      throw new EmptyStreamError('Model stream completed without any chunks.');
+    }
+
+    // --- FIX: The entire validation block was restructured for clarity and correctness ---
+    // Only apply complex validation if an invalid chunk was actually found.
+    if (isStreamInvalid) {
+      // Fail immediately if an invalid chunk was not the absolute last chunk.
+      if (validChunkAfterInvalidEncountered) {
+        throw new EmptyStreamError(
+          'Model stream had invalid intermediate chunks without a tool call.',
```
anj-s, line 611: how do we know we didn't have a tool call here?

anj-s (approved):

PR #7323, opened 2025-08-28
fix(trust): Refuse to load  from untrusted process.cwd() sources; Add tests

richieforeman:
Based on top of https://github.com/google-gemini/gemini-cli/pull/7313 (I think I diffbase'd this properly, I don't know Github).

## TLDR

Read trust settings when reading `.env`.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | x  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Resolves https://github.com/google-gemini/maintainers-gemini-cli/issues/722

### DIFF
--- a/packages/cli/src/config/trustedFolders.test.ts
+++ b/packages/cli/src/config/trustedFolders.test.ts
@@ -80,6 +80,54 @@ describe('Trusted Folders Loading', () => {
     expect(errors).toEqual([]);
   });
 
+  describe('isPathTrusted', () => {
+    function setup({ config = {} as Record<string, TrustLevel> } = {}) {
+      (mockFsExistsSync as Mock).mockImplementation(
+        (p) => p === USER_TRUSTED_FOLDERS_PATH,
+      );
+      (fs.readFileSync as Mock).mockImplementation((p) => {
+        if (p === USER_TRUSTED_FOLDERS_PATH) return JSON.stringify(config);
+        return '{}';
+      });
+
+      const folders = loadTrustedFolders();
+
+      return { folders };
+    }
+
+    it('provides a method to determine if a path is trusted', () => {
+      const { folders } = setup({
+        config: {
+          './myfolder': TrustLevel.TRUST_FOLDER,
+          '../myfolder': TrustLevel.TRUST_FOLDER,
+          '/trustedparent/trustme': TrustLevel.TRUST_PARENT,
+          '/user/folder': TrustLevel.TRUST_FOLDER,
+          '/secret': TrustLevel.DO_NOT_TRUST,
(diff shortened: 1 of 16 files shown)


packages/cli/src/config/trustedFolders.ts
```diff
@@ -53,6 +53,51 @@ export class LoadedTrustedFolders {
...
+    for (const trustedPath of trustedPaths) {
+      if (isWithinRoot(location, trustedPath)) {
+        return true;
+      }
+    }
+
+    for (const untrustedPath of untrustedPaths) {
+      if (path.normalize(location) === path.normalize(untrustedPath)) {
+        return false;
+      }
```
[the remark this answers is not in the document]
richieforeman: That's how it was implemented.

packages/cli/src/config/trustedFolders.ts
```diff
@@ -53,6 +53,51 @@ export class LoadedTrustedFolders {
...
+    }
+
+    for (const trustedPath of trustedPaths) {
+      if (isWithinRoot(location, trustedPath)) {
+        return true;
+      }
+    }
+
+    for (const untrustedPath of untrustedPaths) {
```
cornmander, line 90: Can we move the untrusted loop first?

packages/cli/src/config/trustedFolders.test.ts
```diff
@@ -183,6 +231,7 @@ describe('isWorkspaceTrusted', () => {
     expect(isWorkspaceTrusted(mockSettings)).toBe(false);
   });
 
+  // RICHIE THINKS THAT RETURNING UNDEFINED IS ACTUALLY A PROBLEM. `true | false | undefined` is very strange.
```
cornmander: Can you file an issue about this so we don't lose track of this?

cornmander (approved):

packages/cli/src/config/trustedFolders.test.ts
richieforeman: Will do, I think we discussed this and it's now resolved tho.

packages/cli/src/config/trustedFolders.ts
richieforeman: I just moved this logic, and there's a test that asserts this behavior. For better or worse, I'm going to ACK this.  I think if we want this behavior to change, we should do it a follow up.  As discussed, I think there's some improvements that could be done here.

PR #7325, opened 2025-08-28
fix(cli): remove settings migration console logs

galdawave:
Removes console logging from the settings migration process to reduce noise in the output.


jacob314 (approved):
lgtm

PR #7341, opened 2025-08-28
Fix(trust) - Disable commands from untrusted directories when useFolderTrust is enabled

shishu314:
## TLDR

Disable custom commands that are not from trusted directories

## Dive Deeper

I just used the config values to determine whether or not a directory is trusted since that is what we do throughout the codebase. The config value only checks current workspace though, which doesn't necessarily mean the trusted folder config gets processed.

## Reviewer Test Plan

Ask gemini to generate a custom command into your `.gemini` directory. Then change `settings.json` and `trustedFolders.json` to include the appropriate flag/trusted or untrusted folders. 

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | x  | ❓  | ❓  |
| npx      | x  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes https://github.com/google-gemini/maintainers-gemini-cli/issues/642?issue=google-gemini%7Cmaintainers-gemini-cli%7C720

### DIFF
--- a/packages/cli/src/services/FileCommandLoader.ts
+++ b/packages/cli/src/services/FileCommandLoader.ts
@@ -33,6 +33,7 @@ import {
   ShellProcessor,
 } from './prompt-processors/shellProcessor.js';
 import { AtFileProcessor } from './prompt-processors/atFileProcessor.js';
+import { loadTrustedFolders } from '../config/trustedFolders.js';
 
 interface CommandDirectory {
   path: string;
@@ -94,13 +95,14 @@ export class FileCommandLoader implements ICommandLoader {
     const commandDirs = this.getCommandDirectories();
     for (const dirInfo of commandDirs) {
       try {
-        const files = await glob('**/*.toml', {
+        let files = await glob('**/*.toml', {
           ...globOptions,
           cwd: dirInfo.path,
         });
 
         if(this.folderTrustEnabled) {
-          return [];
+          const trustedFolders = loadTrustedFolders();
+          files = files.filter((file) => trustedFolders.isPathTrusted(path.join(dirInfo.path, file)));
         }
 
         const commandPromises = files.map((file) =>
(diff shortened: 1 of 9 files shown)


packages/cli/src/services/FileCommandLoader.ts
```diff
@@ -63,8 +63,12 @@ const TomlCommandDefSchema = z.object({
  */
 export class FileCommandLoader implements ICommandLoader {
   private readonly projectRoot: string;
+  private readonly folderTrustEnabled: boolean;
+  private readonly folderTrust: boolean;
 
   constructor(private readonly config: Config | null) {
+    this.folderTrustEnabled = !!config?.getFolderTrustFeature();
```
richieforeman, line 70: I don't think you need my vars if you heed my guidance below.

packages/cli/src/services/FileCommandLoader.ts
```diff
@@ -92,11 +96,15 @@ export class FileCommandLoader implements ICommandLoader {
     const commandDirs = this.getCommandDirectories();
     for (const dirInfo of commandDirs) {
       try {
-        const files = await glob('**/*.toml', {
+        let files = await glob('**/*.toml', {
           ...globOptions,
           cwd: dirInfo.path,
         });
 
+        if(this.folderTrustEnabled && !this.folderTrust) {
```
richieforeman: You need to call `isWorkspaceTrusted` and that should give you what you need.

packages/cli/src/services/FileCommandLoader.ts
```diff
@@ -63,8 +63,12 @@ const TomlCommandDefSchema = z.object({
  */
```
richieforeman, line 63: Gimmie some tests, yo!

packages/cli/src/services/FileCommandLoader.ts
shishu314: Those config values are populated already.

packages/cli/src/services/FileCommandLoader.ts
richieforeman: Agree to resolve. This can also be read from config.

cornmander (approved):

PR #7342, opened 2025-08-28
fix(trust): Refuse to load extensions from untrusted workspaces

richieforeman:
## TLDR

Refuse to load extensions from untrusted workspaces.

## Dive Deeper

I also found these tests a little messy, so I took some liberties to tidy them up.  Most of the churn in the tests is due to that. .

## Reviewer Test Plan

- Unit tests cover the functionality quite well.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | x  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Closes https://github.com/google-gemini/maintainers-gemini-cli/issues/721

### DIFF
--- a/packages/cli/src/config/extension.ts
+++ b/packages/cli/src/config/extension.ts
@@ -16,6 +16,7 @@ import { simpleGit } from 'simple-git';
 import { SettingScope, loadSettings } from '../config/settings.js';
 import { getErrorMessage } from '../utils/errors.js';
 import { recursivelyHydrateStrings } from './extensions/variables.js';
+import { isWorkspaceTrusted } from './trustedFolders.js';
 
 export const EXTENSIONS_DIRECTORY_NAME = '.gemini/extensions';
 
@@ -112,7 +113,10 @@ export function loadExtensions(workspaceDir: string): Extension[] {
   const disabledExtensions = settings.extensions?.disabled ?? [];
   const allExtensions = [...loadUserExtensions()];
 
-  if (!settings.experimental?.extensionManagement) {
+  if (
+    isWorkspaceTrusted(settings) &&
+    !settings.experimental?.extensionManagement
+  ) {
     allExtensions.push(...getWorkspaceExtensions(workspaceDir));
   }
 
--- a/packages/cli/src/config/extension.ts
+++ b/packages/cli/src/config/extension.ts
@@ -114,7 +114,7 @@ export function loadExtensions(workspaceDir: string): Extension[] {
   const allExtensions = [...loadUserExtensions()];
 
   if (
-    isWorkspaceTrusted(settings) &&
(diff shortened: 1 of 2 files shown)


packages/cli/src/config/extension.ts
cornmander, line 1: I wonder if we should alter performWorkspaceExtensionMigration as well or the callsite to it. I think the user gets prompted to migrate workspace extensions to the main settings file.

packages/cli/src/config/extension.ts
```diff
@@ -112,7 +113,10 @@ export function loadExtensions(workspaceDir: string): Extension[] {
   const disabledExtensions = settings.extensions?.disabled ?? [];
   const allExtensions = [...loadUserExtensions()];
 
-  if (!settings.experimental?.extensionManagement) {
+  if (
+    (isWorkspaceTrusted(settings) ?? true) &&
```
shrutip90, line 117: Everywhere else in code, we use config.isTrustedFolder. Please use that if possible.

richieforeman: That just checks a boolean.  I don't think that does what you're thinking.  It's possible to enable folderTrust mode and not be in a restricted/un-trusted folder.  It's really important to use this function (or some version of it).  This checks the bool + the current dir.

packages/cli/src/config/extension.ts
richieforeman: DONE.

packages/cli/src/config/extension.ts
richieforeman: Oh i see, you're reading isWorkspaceTrusted into that variable within the config (odd. okay).  

I have concerns about plumbing config into here since it has so many dependencies.

shrutip90: We actually wanted to move trustedFolders parsing etc also to core package. Since trust applies to non-CLI packages too. I see one issue with the current implementation for config.isTrustedFolder that it doesnt take into account that the boolean folderTrust flags could change since these settings changes also don't trigger a restart. We can modify config.isTrustedFolder to call isWorkspaceTrusted. I think it would be nice to have one method thats the source of truth that all the checkers rely on.

richieforeman: Given that config isn't plumbed into here, I'd like to keep this as-is for the moment and have it as something you potentially catch in a followup.

shrutip90: Sure, thanks.

cornmander (approved):

Issue #6260, opened 2025-08-14
Add support for citation display in the CLI: Phase 1 


ahockman-google:
### What would you like to be added?

Implement support in the CLI to display recitation information when LLM responses generated by Gemini match other publicly-available source code.

### Why is this needed?

The Gemini API provides information to developers, identifying when generated source code may match other publicly-available source code. This notification is important because it helps developers understand the licensing implications of any generated + recited source code they may choose to add to their project.

masoudahg00:
@masoudahg00 @masoudleyli @masoudahg1 > ### What would you like to be added?
> 
> Implement support in the CLI to display recitation information when LLM responses generated by Gemini match other publicly-available source code.
> 
> ### Why is this needed?
> 
> The Gemini API provides information to developers, identifying when generated source code may match other publicly-available source code. This notification is important because it helps developers understand the licensing implications of any generated + recited source code they may choose to add to their project.
> 
> ### Additional context
> 
> _No response_

- [x] `@masoudahg00``` @masoudleyli @masoudahg1 #6279 #6514 #1 #2 #3 #4 #5 #6 #7 #8 #9 #10 #11 #12 #13 #15 #14

PR #7350, opened 2025-08-28
Show citations at the end of each turn

scidomino:
## TLDR

Show citations at the end of each turn. Disabled by default (with `showCitations` option in `settings.json`).

## Deeper Dive

We want this to be enabled by default for OAuth users that aren't on free tier. That will be done in a follow up PR.

## Reviewer Test Plan

Add the following to your `settings.json`:

`"showCitations": true,`

Then ask:

`Without looking it up Recite a the frog prince fairy tale`

It doesn't always work on the first try. You may have to ask for 4 or 5 fairy tales to get some citations.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | x  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Contributes: https://github.com/google-gemini/gemini-cli/issues/6260


packages/cli/src/config/settings.ts
```diff
@@ -45,6 +45,7 @@ const MIGRATION_MAP: Record<string, string> = {
   hideFooter: 'ui.hideFooter',
   showMemoryUsage: 'ui.showMemoryUsage',
   showLineNumbers: 'ui.showLineNumbers',
+  showCitations: 'ui.showCitations',
```
jacob314, line 48: I think this can be omitted from the migration map as you are post migration. @galz10 can verify that this doesn't break things.

packages/cli/src/ui/hooks/useGeminiStream.ts
```diff
@@ -486,6 +488,20 @@ export const useGeminiStream = (
     [addItem, pendingHistoryItemRef, setPendingHistoryItem, config, setThought],
   );
 
+  const handleCitationEvent = useCallback(
+    (text: string, userMessageTimestamp: number) => {
+      if (!settings?.merged?.ui?.showCitations) {
+        return;
+      }
+      if (pendingHistoryItemRef.current) {
```
jacob314, line 496: the case finishing the pending history item when there is a citation makes me a bit nervous... partly because of how restricted the static rendering is.  possibly bad idea is if we can we instead integrate citations with the appropriate pending history item.

Reasons: might have wacky race conditions if the pending history item is still being added to or if there is a newer item if the query to get the citation was slow. we'll have to leverage the refreshStatic hack to make this work.

packages/cli/src/ui/hooks/useGeminiStream.ts
```diff
@@ -486,6 +488,20 @@ export const useGeminiStream = (
...
+        addItem(pendingHistoryItemRef.current, userMessageTimestamp);
+        setPendingHistoryItem(null);
+      }
+      addItem({ type: MessageType.INFO, text }, userMessageTimestamp);
```
jacob314, line 500: before we enable this by default for any users we should make this collapsed to one line by default with an affordance to expand with an appropriate keyboard shortcut. Fyi @miguelsolorio for thoughts on the right UI for this

jacob314 (commented):
To be clear. Supporting collapsing citations by default can be a follow up pr.
Approved after the other hopefully minor comments are discussed

packages/cli/src/config/settings.ts
scidomino: I spoke with Gal. This is needed for it to be editable int he UI.

packages/cli/src/ui/hooks/useGeminiStream.ts
scidomino: Right. 

I think for these to make sense they have to be shown "in stream" (that is, as a history item) so the citation is shown immediately after the text it is for. But that will make it hard to show a preview with "view more" option since history items are display-and-forget (like the About screen). 

If we really want to do "preview with show/more" then it probably makes more sense to have them be out-of-stream (like the debug screen). 

I prefer showing them in stream with no preview. From my playing around with it, it doesn't look like most queries have many citations. So "show more" might be overkill most of the time.

jacob314: what about a single line view that shows one citation fully? its not like I really ever care that there were 3 citations of the same starter project for whatever framework I was using.

packages/cli/src/config/settings.ts
jacob314: Fyi @galz10 can we fix this? Would prefer we didn't have to add things here just to make them editable in the UI.

packages/cli/src/ui/hooks/useGeminiStream.ts
scidomino: We only return citations at the end of a turn. Never in the middle. So in practice I think this would never happen.

jacob314 (approved):
lgtm

packages/cli/src/ui/hooks/useGeminiStream.ts
scidomino: Resolved offline.

scidomino:
I tested this with function calls and it works as expected.

PR #7358, opened 2025-08-28
Skip MCP server connections in untrusted folders

shrutip90:
## Dive Deeper

MCP servers do not connect and do not show up in /mcp list.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes https://github.com/google-gemini/maintainers-gemini-cli/issues/718

### DIFF
--- a/packages/core/src/tools/mcp-client-manager.ts
+++ b/packages/core/src/tools/mcp-client-manager.ts
@@ -56,6 +56,9 @@ export class McpClientManager {
    * them with the `ToolRegistry`.
    */
   async discoverAllMcpTools(cliConfig: Config): Promise<void> {
+    if (cliConfig.isTrustedFolder() === false) {
+      return;
+    }
     await this.stop();
 
     const servers = populateMcpServerCommand(
--- a/packages/core/src/tools/mcp-client-manager.ts
+++ b/packages/core/src/tools/mcp-client-manager.ts
@@ -4,7 +4,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import type { MCPServerConfig } from '../config/config.js';
+import type { Config, MCPServerConfig } from '../config/config.js';
 import type { ToolRegistry } from './tool-registry.js';
 import type { PromptRegistry } from '../prompts/prompt-registry.js';
 import {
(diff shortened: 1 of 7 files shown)


packages/core/src/tools/mcp-tool.ts
```diff
@@ -80,7 +82,9 @@ class DiscoveredMCPToolInvocation extends BaseToolInvocation<
     const serverAllowListKey = this.serverName;
     const toolAllowListKey = `${this.serverName}.${this.serverToolName}`;
 
-    if (this.trust) {
+    const isTrustedFolder = this.cliConfig?.isTrustedFolder() !== false;
```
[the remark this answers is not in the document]
shrutip90: This is intentional.

reconsumeralization:
thank you git is killing me right now!

packages/core/src/tools/mcp-client-manager.ts
```diff
@@ -91,7 +91,7 @@ export class McpClientManager {
 
         try {
           await client.connect();
```
cornmander, line 96: We need to stop connecting. Connecting and discovering are both potentially unsafe.

packages/core/src/tools/mcp-tool.ts
```diff
@@ -70,6 +71,7 @@ class DiscoveredMCPToolInvocation extends BaseToolInvocation<
     readonly timeout?: number,
     readonly trust?: boolean,
     params: ToolParams = {},
+    private readonly cliConfig?: Config,
```
jacob314, line 74: don't make things nullable that are key for security... or if you don't want to deal with mocking it in tests and don't expect to add other params, just pass in isTrustedFolder as a required boolean instead.

jacob314 (commented):
Approved after this minor comment is addressed to insure we don't have bugs causing this logic to be ignored in the future.

packages/core/src/tools/mcp-client-manager.ts
shrutip90: Done

packages/core/src/tools/mcp-tool.ts
shrutip90: Thanks, this makes sense and seems to be causing a lot of confusion even to the CLI. I will update the isTrustedFolder to a required boolean in a fast follow since that will modify a lot of unrelated files too.

cornmander (approved):

jacob314 (approved):
lgtm

PR #7395, opened 2025-08-29
Reduce bundle size & check it in CI

swissspidy:
## TLDR

Addresses some low-hanging fruit re: bundle size improvements by replacing/updating/externalizing modules.

## Dive Deeper

Visualization using https://esbuild.github.io/analyze/:

Screenshot 2025-08-29 at 17 13 50

To run this yourself, run `DEV=true npm run bundle`, and then drop `bundle/esbuild.json` into that website.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

* See #7403


packages/cli/package.json
```diff
@@ -52,6 +50,7 @@
     "string-width": "^7.1.0",
     "strip-ansi": "^7.1.0",
     "strip-json-comments": "^3.1.1",
+    "ts-deepmerge": "^7.0.3",
```
jacob314: can you make sure this package is available in g3 before landing this?

.github/workflows/ci.yml
```diff
@@ -371,3 +371,24 @@ jobs:
...
+      - name: 'Checkout'
+        uses: 'actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8' # ratchet:actions/checkout@v5
+        with:
+          fetch-depth: 1
+
+      - uses: 'preactjs/compressed-size-action@946a292cd35bd1088e0d7eb92b69d1a8d5b5d76a'
+        with:
+          repo-token: '${{ secrets.GITHUB_TOKEN }}'
+          pattern: './bundle/**/*.{js,sb}'
+          minimum-change-threshold: '100'
```
jacob314: is the unit for this in bytes? if so, would be nice to have a singificantly larger threshold to avoid this action triggering a lot.

jacob314 (approved):

packages/cli/package.json
swissspidy: Since when is that a requirement? 🤔 

That's not the case for all the other dependencies here. And from what I can see, only the final bundled `gemini.js` is in g3 anyway.

.github/workflows/ci.yml
swissspidy: Changed to 1000 bytes now, which in my experience is plenty. It can always be adjusted later.

packages/cli/package.json
swissspidy: Looks like it was replaced with a custom deep merge function in the meantime anyway.

PR #7408, opened 2025-08-29
feat(extension) - Notify users when there is a new version and update it

### LINKED ISSUE #7043: Notify users when a new version of the Gemini CLI Companion extension is available

A notification that alerts users when a new version of the Gemini CLI Companion extension is available and provides a clear call to action to update.

This notification should include:
- A message that a new version of the Gemini CLI has been released
- The current version and the new version numbers
- A button or link to a command that updates the CLI to the latest version
- An option to dismiss the notification

### Why is this needed?

By providing a clear and non-intrusive notification, we can encourage users to keep their CLI up-to-date, ensuring a better and more consistent user experience.

### Additional context

_No response_

shishu314:
## TLDR

Show update version notification when current version is oudated

## Dive Deeper

I used `getExtension` to get the current latest extension and call the `installExtension` command to update. Let me know if there is a better way

## Reviewer Test Plan

I actually don't know how to mock the version to be an older version. I updated `package.json` to be an older version to test but not sure correct.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | x  | ❓  | ❓  |
| npx      | x  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs
Fixes #7043


packages/vscode-ide-companion/src/extension.ts
```diff
@@ -17,11 +19,52 @@ let logger: vscode.OutputChannel;
...
 
+async function checkForUpdates(
+  context: vscode.ExtensionContext,
+  log: (message: string) => void,
+) {
+  try {
+    const currentVersion = context.extension.packageJSON.version;
+
+    // Fetch package.json from the main branch of the repository.
+    // This is a simple way to get the latest version.
```
skeshive: Is it possible to fetch this info from the VS code marketplace? Fetching from github is brittle because the extension release is manual so even if the package.json number is higher in github, it doesnt necessarily mean we released that version yet

shishu314: Done

packages/vscode-ide-companion/src/extension.ts
```diff
@@ -17,11 +19,79 @@ let logger: vscode.OutputChannel;
 
 let log: (message: string) => void = () => {};
 
+async function checkForUpdates(
+  context: vscode.ExtensionContext,
+  log: (message: string) => void,
+) {
+  try {
+    const currentVersion = context.extension.packageJSON.version;
```
skeshive, line 27: this won't be accurate since our releases are still manual- can we keep the reflection logic in a previous version and get the current version using the VS code action?

skeshive (approved):

packages/vscode-ide-companion/src/extension.ts
```diff
@@ -17,11 +19,79 @@ let logger: vscode.OutputChannel;
...
+    log(`Error checking for extension updates: ${message}`);
+  }
+}
+
 export async function activate(context: vscode.ExtensionContext) {
   logger = vscode.window.createOutputChannel('Gemini CLI IDE Companion');
   log = createLogger(context, logger);
   log('Extension activated');
 
+  await checkForUpdates(context, log);
```
cornmander: Does this need to be await?

cornmander (approved):

PR #7409, opened 2025-08-29
fix(trust): Respect folder trust setting when reading GEMINI.md

richieforeman:
## TLDR

- Respects the folder trust setting when reading GEMINI.md files.

## Dive Deeper

I do feel like this code could made a bit more clear. The workspace folder walking feelings like something that could/should be separated out from this.  I also think that rather than property-drilling a bunch of stuff from `Config`, we could just pass the entire `Config` object.  This would create a bit of a catch-22, since we read/use memory prior to initialization of Config.  I resisted the urge to refactor this heavily.

## Reviewer Test Plan

- in a trusted workspace
> it should load GEMINI.md files from the workspace
> It should load GEMINI.md files from /home

- in an untrusted workspace
> it should NOT load GEMINI.md files from the workspace
> it still should load GEMINI.md files from /home

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | x  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

https://github.com/google-gemini/maintainers-gemini-cli/issues/727

### DIFF
--- a/packages/core/src/utils/memoryDiscovery.ts
+++ b/packages/core/src/utils/memoryDiscovery.ts
@@ -9,14 +9,12 @@ import * as fsSync from 'node:fs';
 import * as path from 'node:path';
 import { homedir } from 'node:os';
 import { bfsFileSearch } from './bfsFileSearch.js';
-import {
-  GEMINI_CONFIG_DIR,
-  getAllGeminiMdFilenames,
-} from '../tools/memoryTool.js';
+import { getAllGeminiMdFilenames } from '../tools/memoryTool.js';
 import type { FileDiscoveryService } from '../services/fileDiscoveryService.js';
 import { processImports } from './memoryImportProcessor.js';
 import type { FileFilteringOptions } from '../config/config.js';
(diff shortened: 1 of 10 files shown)


packages/core/src/utils/memoryDiscovery.ts
```diff
@@ -148,7 +149,7 @@ async function getGeminiMdFilePathsInternalForEachDir(
     const resolvedHome = path.resolve(userHomePath);
     const globalMemoryPath = path.join(
       resolvedHome,
-      GEMINI_CONFIG_DIR,
+      GEMINI_DIR,
```
cornmander, line 152: Should we file an issue to normalize this elsewhere?

cornmander (approved):

packages/core/src/utils/memoryDiscovery.ts
richieforeman: https://github.com/google-gemini/gemini-cli/issues/7414

PR #7416, opened 2025-08-29
Handle cleaning up the response text in the UI when a response stream retry occurs

### LINKED ISSUE #7428: Partial content from a failed stream attempt appears in the final response after a retry
### What happened?

As a result of resolving #6889, a UX side effect was introduced. When the model produces a streaming response that fails mid-way (e.g., by sending an invalid data chunk), the internal retry mechanism is correctly triggered. However, the partial text that was already rendered to the UI from the failed attempt is not cleared.

When the subsequent, successful retry stream begins, its content is incorrectly appended to the stale text from the failed attempt. This results in a garbled, concatenated message being displayed to the user, containing both the end of the failed response and the entirety of the successful one.

**Example of the incorrect state flow:**

* **Initial State:** `buffer = ""`
* **Chunk from Failed Stream:** `buffer = "Partial content..."`
* **Retry Event Received:** The bug occurs here; the buffer is **not** cleared.
* **Chunk from Successful Stream:** `buffer = "Partial content...Successful content."`

The user sees the final, concatenated string.

When an internal retry is triggered, all state associated with the failed stream attempt should be completely discarded. Any partial text displayed in the UI should be cleared immediately.

The new, successful stream from the retry attempt should then render from a clean slate. The user should only ever see the final, coherent response from the successful attempt, and should be completely unaware that a recovery action happened behind the scenes.

### Client information

<details>

```console
> /about
# paste output here
```

</details>

### Login information

_No response_

_No response_

mrcabbage972:
## TLDR
Handle cleaning up the response text in the UI when a response stream retry occurs.

**Example of the incorrect state flow:**

* **Initial State:** `buffer = ""`
* **Chunk from Failed Stream:** `buffer = "Partial content..."`
* **Retry Event Received:** The bug occurs here; the buffer is **not** cleared.
* **Chunk from Successful Stream:** `buffer = "Partial content...Successful content."`

See linked issue for more details.

## Dive Deeper
`geminiChat.ts` was changed so that `sendMessageStream` returns a new `StreamEvent` which can indicate a retry event.

`turn.ts` was changed to include a new `ServerGeminiRetryEvent` in `ServerGeminiStreamEvent`.

In a future PR, `useGeminiStream` will change to clear the text buffer on a `ServerGeminiRetryEvent`.

Added new test cases for all changed files to cover handling the new retry event.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | 👍   |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs
Resolves #7428


anj-s (approved):

packages/cli/src/ui/hooks/useGeminiStream.ts
```diff
@@ -635,6 +635,12 @@ export const useGeminiStream = (
             // before we add loop detected message to history
             loopDetectedRef.current = true;
             break;
+          case ServerGeminiEventType.Retry:
```
jacob314, line 653: It might be helpful to include a screencast showing what the behavior is for this case. My guess is you might see a partial response be generated then deleted with no UX indicating that a retry is occurring. 

@miguelsolorio curious on your thoughts on the right UX here. This might be fine or we might want to also show a status message somewhere in the UI indicating we are retrying.

packages/cli/src/zed-integration/zedIntegration.ts
```diff
@@ -269,8 +270,12 @@ class Session {
             return { stopReason: 'cancelled' };
           }
 
-          if (resp.candidates && resp.candidates.length > 0) {
-            const candidate = resp.candidates[0];
+          if (
```
jacob314, line 273: does Zed also need to get some notification to clear the previous chunk if a retry has occurred?

jacob314 (commented):
Looks good overall. A couple small questions about the UX.

packages/cli/src/ui/hooks/useGeminiStream.ts
mrcabbage972: I'm also not sure what's the best UX here. Too many notifications to the users about transient errors being retried can also be annoying and have no functional impact. 

I do think that the behavior in the PR is an improvement over the existing state, where the retried response just gets appending to the failed ones. Also it adds the infra to handle these retries in whatever way we want, UX-wise.

packages/cli/src/zed-integration/zedIntegration.ts
mrcabbage972: I'm not yet familiar with Zed, but sounds like this could be a nice addition in a subsequent PR.

mrcabbage972:
I removed the UX change, left only the infra. Will catch up with @jacob314 when they're available.

packages/cli/src/ui/hooks/useGeminiStream.ts
mrcabbage972: @jacob314 Here are the screencasts:

**Current behavior**
retry_bug



**Proposed Behavior**
retry_bug_fixed

cc: @anj-s

PR #7419, opened 2025-08-29
Add footer configuration settings

### LINKED ISSUE #4261: Better UX support in the footer/status bar

Currently there are a lot of items that are visually noisy, distracting you from the main conversation.

### Why is this needed?

Needed in order to minimize noise and shift the focus to the content

Image

### Additional context

It would be nice if there was a way to simplify a lot of these info into a true status bar

Image

miguelsolorio:
## TLDR

This adds configuration options for the footer to toggle certain items on/off. Everything is on by default.

## Dive Deeper

This introduces the following new settings:
- `hideContextSummary`: Hides the context summary (GEMINI.md, MCP servers) above the input.
- `hideCWD`: Hides the current working directory path in the footer.
- `hideSandboxStatus`: Hide the sandbox status indicator in the footer.
- `hideModelInfo`: Hide the model name and context usage in the footer.

CleanShot 2025-08-29 at 12 09 58@2x

And also moves out corgi & errors out of the model container and places them into their own container:

CleanShot 2025-08-29 at 12 13 21@2x

## Before
CleanShot 2025-08-29 at 12 03 02@2x

## After

When context summary is hidden (`hideContextSummary`)

CleanShot 2025-08-29 at 12 03 46@2x
<hr>

When CWD is hidden (`hideCWD`)

CleanShot 2025-08-29 at 12 04 58@2x

<hr>

When sandbox is hidden (`hideSandboxStatus`)

CleanShot 2025-08-29 at 12 06 14@2x

<hr>

When model info is hidden (`hideModelInfo`)

CleanShot 2025-08-29 at 12 07 43@2x

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

- Fixes #4261


packages/cli/src/config/config.ts
```diff
@@ -622,6 +622,11 @@ export async function loadCliConfig(
     skipNextSpeakerCheck: settings.model?.skipNextSpeakerCheck,
     enablePromptCompletion: settings.general?.enablePromptCompletion ?? false,
     eventEmitter: appEvents,
+    footer: {
```
jacob314: remove this from config.ts. These ones should just be referenced directly from settings as they are strictly UI

packages/cli/src/ui/__snapshots__/App.test.tsx.snap
```diff
@@ -27,5 +27,5 @@ dir
 
 no sandbox (see /docs)
 
-model (100% context left)| ✖ 5 errors (ctrl+o for details)"
+model (100% context left)  | ✖ 5 errors (ctrl+o for details)"
```
jacob314, line 30: previous snapshot had one space too few. new one has one space too many. Can we polish this?

packages/cli/src/ui/App.test.tsx
```diff
@@ -653,6 +653,41 @@ describe('App UI', () => {
...
+    const { lastFrame, unmount } = renderWithProviders(
+      <App
+        config={mockConfig as unknown as ServerConfig}
+        settings={mockSettings}
+        version={mockVersion}
+      />,
+    );
+    currentUnmount = unmount;
+    await Promise.resolve();
+    expect(lastFrame()).not.toContain(
```
jacob314: nit this test could pass for the wrong reasons (basically any text that isn't exactly this text). instead just have this match the exact golden you want.

packages/cli/src/ui/App.tsx
```diff
@@ -1380,6 +1383,9 @@ const App = ({ config, settings, startupWarnings = [], version }: AppProps) => {
               nightly={nightly}
               vimMode={vimModeEnabled ? vimMode : undefined}
               isTrustedFolder={isTrustedFolderState}
+              hideCWD={settings.merged.ui?.hideCWD}
```
jacob314: good news is you are using settings here as I would hope so the change to config may be revertable without any other tweaks.

packages/core/src/config/config.ts
```diff
@@ -68,6 +68,12 @@ export interface AccessibilitySettings {
   screenReader?: boolean;
 }
 
+export interface FooterSettings {
```
jacob314: move this to settings.ts
sorry gemini cli messed this up. I'll update the GEMINI.md prompt to make sure future contributors don't end up with settings in the wrong place like this.

jacob314 (commented):
lgtm
one the hopefully minor changes are addressed.

packages/cli/src/ui/__snapshots__/App.test.tsx.snap
miguelsolorio: I intentionally added the space here so it’s not right up against the parenthesis

packages/core/src/config/config.ts
miguelsolorio: Had to do a little refactor to move it out of core

jacob314 (approved):
lgtm

PR #7427, opened 2025-08-29
feat(core): Enable ripgrep by default.

### LINKED ISSUE #7327: Turn on ripgrep by default
Ripgrep is surfaced as an option but not enabled by default.
Need to verify fallback is graceful for users that cannot use ripgrep.
Need to verify it does not regress quality due to differences in what matches are shown.

joshualitt:
## TLDR

Enables ripgrep by default

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs
- Fixes #7327

### DIFF
--- a/packages/core/src/core/subagent.test.ts
+++ b/packages/core/src/core/subagent.test.ts
@@ -45,6 +45,7 @@ vi.mock('../ide/ide-client.js');
 
 async function createMockConfig(
   toolRegistryMocks = {},
+  configParameters: Partial<ConfigParameters> = {},
 ): Promise<{ config: Config; toolRegistry: ToolRegistry }> {
   const configParams: ConfigParameters = {
     sessionId: 'test-session',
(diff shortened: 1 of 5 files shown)


packages/core/src/core/subagent.test.ts
```diff
@@ -767,7 +769,7 @@ describe('subagent.ts', () => {
         // Use fake timers to reliably test timeouts
         vi.useFakeTimers();
 
-        const { config } = await createMockConfig();
+        const { config } = await createMockConfig({}, { useRipgrep: false });
```
joshualitt, line 772: I'm not 100% sure why, but for some reason I had to disable ripgrep here and in the below test to get the subagent tests to pass on windows, otherwise they were mysteriously timing out. I can investigate more if desired, but it may just be a timers issue.

joshualitt:
Sending out for early review, still need to test more on windows.

jacob314 (approved):
lgtm

PR #7433, opened 2025-08-29
Override Gemini CLI trust with VScode workspace trust when in IDE

shrutip90:
## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Related to https://github.com/google-gemini/maintainers-gemini-cli/issues/717


packages/core/src/config/config.ts
```diff
@@ -749,6 +750,11 @@ export class Config {
     // restarts in the more common path. If the user chooses to mark the folder
     // as untrusted, the CLI will restart and we will have the trust value
     // reloaded.
+    const context = ideContext.getIdeContext();
+    if (context?.workspaceState?.isTrusted !== undefined) {
+      return context.workspaceState.isTrusted;
+    }
+
     return this.trustedFolder ?? true;
```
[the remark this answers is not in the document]
shrutip90: IdeMode can be on even outside the IDE, so this doesn't work. I could look for the env variables like GEMINI_CLI_IDE_WORKSPACE_PATH that we set in the IDE mode though. This will cause a degraded experience in envs where trust is undefined or TRUSTED outside Vscode and TRUSTED in Vscode. But it will be give us better correctness in the case where it is undefined outside Vscode and UNTRUSTED in Vscode.

@jacob314 / @cornmander please let me know if i should query those and default to false when we can't determine the status.

packages/cli/src/ui/hooks/useFolderTrust.test.ts
```diff
@@ -4,7 +4,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { vi } from 'vitest';
+import { vi, type Mock } from 'vitest';
```
jacob314: nit: revert this change

packages/cli/src/ui/App.tsx
```diff
@@ -304,6 +306,23 @@ const App = ({ config, settings, startupWarnings = [], version }: AppProps) => {
...
+  useEffect(() => {
+    if (ideNeedsRestart) {
+      // IDE trust changed, force a restart.
+      setShowIdeRestartPrompt(true);
+    }
+  }, [ideNeedsRestart]);
+
+  useKeypress(
+    (key) => {
+      if (key.name === 'r' || key.name === 'R') {
```
jacob314, line 319: this is a bit scary. Now typing a 'r' triggers a restart so the app is generally unusable until you restart.

packages/core/src/config/config.ts
```diff
@@ -749,6 +750,11 @@ export class Config {
     // restarts in the more common path. If the user chooses to mark the folder
     // as untrusted, the CLI will restart and we will have the trust value
     // reloaded.
+    const context = ideContext.getIdeContext();
+    if (context?.workspaceState?.isTrusted !== undefined) {
```
jacob314, line 754: nit: can combine this with `this.trustedFolder ?? true`

packages/core/src/utils/ide-trust.ts
```diff
@@ -0,0 +1,16 @@
...
+ */
+
+import { ideContext } from '../ide/ideContext.js';
+
+/**
+ * Gets the workspace trust from the IDE if available.
+ * @returns A boolean if the IDE provides a trust value, otherwise undefined.
+ */
+export function getIdeTrust(): boolean | undefined {
+  const context = ideContext.getIdeContext();
```
jacob314: nit: remove context local variable

packages/cli/src/ui/hooks/useIdeTrustListener.ts
```diff
@@ -0,0 +1,53 @@
...
+ * if a restart is needed because the trust state has changed.
+ */
+export function useIdeTrustListener(config: Config) {
+  const [isIdeTrusted, setIsIdeTrusted] = useState<boolean | undefined>(
+    undefined,
+  );
+  const [needsRestart, setNeedsRestart] = useState(false);
+  const initialCheckPerformed = useRef(false);
+
+  useEffect(() => {
```
jacob314: this can be made more idiomatic by using 
https://react.dev/reference/react/useSyncExternalStore

  The current implementation uses useEffect for both subscribing to a listener and managing an initial state check with a useRef flag. This can be
  simplified and made more robust by using the useSyncExternalStore hook, which is designed for subscribing to external data sources. This change
  separates the concern of syncing with the external state from the logic of determining if a restart is needed.

packages/cli/src/ui/hooks/useFolderTrust.test.ts
shrutip90: Done

packages/cli/src/ui/App.tsx
shrutip90: only when showIdeRestartPrompt is true..

packages/core/src/config/config.ts
shrutip90: Ack, but keeping it as is.

packages/core/src/utils/ide-trust.ts
shrutip90: Done

packages/cli/src/ui/hooks/useIdeTrustListener.ts
shrutip90: Done, thanks. PTAL.

jacob314 (approved):
lgtm

PR #7438, opened 2025-08-29
Enable citations by default for certain users.

### LINKED ISSUE #6260: Add support for citation display in the CLI: Phase 1

scidomino:
## TLDR

CodeAssist users on Non-free tier should have citations enabled by default.

## Dive Deeper

Legal mandates this.

## Reviewer Test Plan

Make sure you have no "showCitations" in your settings.json and 

* log in with a google.com email and ask it to tell a fairy tail (it may take several attempts). Verify that you see citations.

* log in with a free tier email and do the same.  Verify you don't see citations.

* log in with API key. Verify you see no citations.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | x  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes #6260

### DIFF
--- a/packages/cli/src/ui/hooks/useGeminiStream.ts
+++ b/packages/cli/src/ui/hooks/useGeminiStream.ts
@@ -31,6 +31,8 @@ import {
   ConversationFinishedEvent,
   ApprovalMode,
   parseAndFormatApiError,
+  getCodeAssistServer,
+  UserTierId,
 } from '@google/gemini-cli-core';
 import { type Part, type PartListUnion, FinishReason } from '@google/genai';
 import type {
@@ -66,6 +68,15 @@ enum StreamProcessingStatus {
   Completed,
   UserCancelled,
   Error,
+}
+
+function showCitations(settings: LoadedSettings, config: Config): boolean {
+  const enabled = settings?.merged?.ui?.showCitations;
+  if (enabled !== undefined) {
+    return enabled;
+  }
+  const server = getCodeAssistServer(config);
+  return (server && server.userTier !== UserTierId.FREE) ?? false;
 }
 
 /**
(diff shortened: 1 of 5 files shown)


packages/cli/src/ui/hooks/useGeminiStream.ts
```diff
@@ -488,18 +490,30 @@ export const useGeminiStream = (
     [addItem, pendingHistoryItemRef, setPendingHistoryItem, config, setThought],
   );
 
+  // Respect whatever settings is in settings.json. Otherwise, disable citations
+  // unless the they are a CodeAssist user that is not on free tier.
+  const showCitations = useCallback(() => {
```
jacob314: nit: rather than making this a useCallback just define this as a standalone top level helper method and pass in settings and config as arguments. 

reason is there is no need to have this be cached callback as the only use for this is from handleCitationEvent within this same file.

jacob314 (approved):
lgtm

packages/cli/src/ui/hooks/useGeminiStream.ts
scidomino: Done.

PR #7442, opened 2025-08-29
fix(diffstats): Fix diff stats to correctly capture the edits

kiranani:
## TLDR

Without this change the diffStats were being computed incorrectly. `params.new_string` does not correspond to the new content to which the old content was diffed against, leading to incorrect computations.

Also renamed `ai_proposed_string` -> `ai_proposed_content` as this is what should be passed as the old content that was initially proposed, and is being updated and compared against in the diff stat computation.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ✅  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

### DIFF
--- a/packages/core/src/tools/edit.test.ts
+++ b/packages/core/src/tools/edit.test.ts
@@ -507,7 +507,7 @@ describe('EditTool', () => {
     });
 
     it('should successfully replace multiple occurrences when expected_replacements specified', async () => {
-      fs.writeFileSync(filePath, 'old text old text old text', 'utf8');
+      fs.writeFileSync(filePath, 'old text\nold text\nold text', 'utf8');
       const params: EditToolParams = {
         file_path: filePath,
         old_string: 'old',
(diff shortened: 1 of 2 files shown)


ghost:
1

ghost:
> ## TLDR
> 
> Without this change the diffStats were being computed incorrectly. `params.new_string` does not correspond to the new content to which the old content was diffed against, leading to incorrect computations.
> 
> Also renamed `ai_proposed_string` -> `ai_proposed_content` as this is what should be passed as the old content that was initially proposed, and is being updated and compared against in the diff stat computation.
> 
> ## Testing Matrix
> 
> 
> 
> |          | 🍏  | 🪟  | 🐧  |
> | -------- | --- | --- | --- |
> | npm run  | ❓  | ❓  | ✅  |
> | npx      | ❓  | ❓  | ❓  |
> | Docker   | ❓  | ❓  | ❓  |
> | Podman   | ❓  | -   | -   |
> | Seatbelt | ❓  | -   | -   |
> 
> ## Linked issues / bugs
> 
> 
>

PR #7445, opened 2025-08-30
refactor(setting): Improve settings migration and tool loading

galdawave:
## TLDR

This pull request refactors the settings migration logic to be more robust, improves how settings are merged, and makes tool loading more resilient to class name minification in production builds. It also introduces helpers to manage the "thinking mode" feature based on the specific model being used.

## Dive Deeper

The core of this PR is a significant improvement to the settings management system.

- **Settings Migration & Merging**: The previous settings migration could be brittle. This change introduces a more robust migration path and utilizes `lodash.mergeWith` for a deep merge of settings from different sources (system, user, workspace). This ensures that user-defined settings correctly override defaults and that nested configurations are combined predictably.

- **Tool Loading**: The tool loading mechanism was previously dependent on class names, which caused issues in production builds where names are often minified. The new implementation avoids this dependency, making tool discovery more reliable.

- **Model-Specific "Thinking Mode"**: New helper functions have been added to determine whether the "thinking" status should be displayed for a given model. This logic is now centralized and based on the model's name, making it easier to manage as new models are introduced.

## Reviewer Test Plan

To validate these changes, a reviewer should:

1.  **Test Settings Merging**:
    - Configure settings at different levels (e.g., a global setting in your user config and a conflicting setting in the workspace config).
    - Run the CLI and verify that the workspace setting correctly overrides the user setting.
    - Check that nested properties are merged as expected.

2.  **Test "Thinking Mode"**:
    - Use a model that is expected to show the "thinking" status (e.g., a Gemini 1.5 model).
    - Use a model that is not expected to show it.
    - Confirm that the status indicator appears only for the correct models.

3.  **Test Production Build**:
    - Create a production build of the CLI.
    - Run the CLI and ensure that all custom and built-in tools are loaded and available for use.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  |  ✅  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


packages/cli/src/config/auth.test.ts
```diff
@@ -59,6 +59,9 @@ describe('validateAuthMethod', () => {
     });
 
     it('should return an error message if no required environment variables are set', () => {
+      vi.stubEnv('GOOGLE_CLOUD_PROJECT', '');
```
jacob314: nit: seems unrelated.

packages/cli/src/config/settings.ts
```diff
@@ -34,54 +34,61 @@ const MIGRATE_V2_OVERWRITE = false;
...
+  disableUpdateNag: 'general.disableUpdateNag',
+  dnsResolutionOrder: 'advanced.dnsResolutionOrder',
+  enablePromptCompletion: 'general.enablePromptCompletion',
+  excludeTools: 'tools.exclude',
+  excludeMCPServers: 'mcp.excluded',
+  excludedProjectEnvVars: 'advanced.excludedEnvVars',
+  extensionManagement: 'advanced.extensionManagement',
+  extensions: 'extensions',
+  fileFiltering: 'context.fileFiltering',
+  folderTrustFeature: 'security.folderTrust.featureEnabled',
```
jacob314, line 83: note: another pr is in process to remove this feature flag after that goes in you will probably want to follow up to remove this from the cleanup map.

packages/cli/src/config/settings.ts
```diff
@@ -295,106 +320,46 @@ function mergeSettings(
...
-        ...new Set([
-          ...(systemDefaults.extensions?.workspacesWithMigrationNudge || []),
-          ...(user.extensions?.workspacesWithMigrationNudge || []),
-          ...(safeWorkspaceWithoutFolderTrust.extensions
-            ?.workspacesWithMigrationNudge || []),
-          ...(system.extensions?.workspacesWithMigrationNudge || []),
-        ]),
-      ],
-    },
+  const customizer = (
```
jacob314: would be good to make this more robust. rather than having the customizations being based on a single key name can we include the full path to the customization or better yet specify these customizations directly in the settingsMetadata.ts file so the customizations are listed right where the settings are ensuring if we change the names in the future we won't accidentally break this logic.

packages/core/src/config/config.ts
```diff
@@ -838,20 +838,22 @@ export class Config {
       const toolName = ToolClass.Name || className;
       const coreTools = this.getCoreTools();
       const excludeTools = this.getExcludeTools() || [];
+      // On some platforms, the className can be minified to _ClassName.
```
jacob314, line 847: this is news to me. What is minifying the class name?

galdawave: not sure, this only happens in g3 not linux, so i assume some process for g3

jacob314 (approved):
lgtm

PR #7446, opened 2025-08-30
fix(diffstats): Fix diff stats to correctly capture the edits

### LINKED ISSUE #7481: Lines of Code metrics are not being properly calculated and stored
### What happened?

The GEMINI_CLI_AI_REMOVED_LINES and GEMINI_CLI_AI_ADDED_LINES metrics are not being processed properly by the application. Currently Lines Removed LOC metric is 10+x the Lines Added, which is suspicious and very likely incorrect. Further investigation revealed the following:

Every successful file edit operation sends the same diffStat data TWICE:

  1. Via ToolCallEvent (/Users/bryanmorgan/code/gemini-cli/packages/core/src/telemetry/types.ts:158-172):
    - When a tool like EditTool or WriteFileTool completes successfully
    - The diffStat from resultDisplay is extracted and sent as metadata
  2. Via FileOperationEvent (/Users/bryanmorgan/code/gemini-cli/packages/core/src/tools/edit.ts:416-427):
    - The same tools explicitly call logFileOperation() with the same diffStat

  Why This Causes 10x Removed vs Added Lines:

  Line modification behavior (confirmed in tests at
  /Users/bryanmorgan/code/gemini-cli/packages/core/src/tools/diffOptions.test.ts:117-128):
  - Any line modification counts as BOTH 1 addition + 1 removal
  - Pure additions count as only additions
  - Pure deletions count as only removals

  The math:
  - If 90% of operations are modifications: (0.9 × 2 removals) + (0.1 × 0 removals) = 1.8 removals
  - But additions: (0.9 × 1 addition) + (0.1 × 1 addition) = 1.0 additions
  - Double-counting effect: 1.8 × 2 = 3.6 removals vs 1.0 × 2 = 2.0 additions
  - Ratio: 3.6 ÷ 2.0 = 1.8x higher removals

  With higher modification rates, this ratio increases toward the 10x you observed.

  Evidence:

  1. ToolCallEvent constructor extracts diffStat when 'diffStat' in call.response.resultDisplay
  2. EditTool.execute() calls logFileOperation() with the same diffStat at line 416
  3. WriteFileTool.execute() calls logFileOperation() with the same diffStat at line 325
  4. Both tools return resultDisplay objects containing the diffStat that triggers ToolCallEvent metrics

  Solution Needed:

 The solution is to modify the FileOperationEvent to exclude the diff_stat parameter while keeping all other metadata
  (programming_language, mimetype, extension, etc.). This way:

  1. ToolCallEvent handles diff_stat metrics (preventing double-counting)
  2. FileOperationEvent handles file metadata (programming_language, mimetype, etc.)

kiranani:
## TLDR

Fix the diff stat computation by:
- Using the right content for diffing: `EditToolParams.new_string` corresponds to the replacement string and not the new content. New content is stored in `editData.newContent`
- Using the right content for `EditToolParams.ai_proposed_content` in `createUpdatedParams`. This was being set to the previous `EditToolParams.new_string` which is merely the old replacement string

Also, renamed `ai_proposed_string` => `ai_proposed_content`, which is what this variable stores.

## Dive Deeper

Setup for testing the changes:
Created a file `test.txt` with:
```
Line 1
Line 2
Line 3
Line 4
Line 5
Line 6
Line 7
Line 8
```
Prompt: `Replace '7' in the file docs/test.txt with 'S3V3N'`

Without the change:
Before

With the change:
After

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ✅  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

- Fixes #7481

### DIFF
--- a/packages/core/src/tools/edit.test.ts
+++ b/packages/core/src/tools/edit.test.ts
@@ -507,7 +507,7 @@ describe('EditTool', () => {
     });
 
     it('should successfully replace multiple occurrences when expected_replacements specified', async () => {
-      fs.writeFileSync(filePath, 'old text old text old text', 'utf8');
+      fs.writeFileSync(filePath, 'old text\nold text\nold text', 'utf8');
       const params: EditToolParams = {
         file_path: filePath,
         old_string: 'old',
@@ -520,12 +520,19 @@ describe('EditTool', () => {
 
       expect(result.llmContent).toMatch(/Successfully modified file/);
       expect(fs.readFileSync(filePath, 'utf8')).toBe(
-        'new text new text new text',
+        'new text\nnew text\nnew text',
       );
       const display = result.returnDisplay as FileDiff;
-      expect(display.fileDiff).toMatch(/old text old text old text/);
-      expect(display.fileDiff).toMatch(/new text new text new text/);
+
+      expect(display.fileDiff).toMatch(/-old text\n-old text\n-old text/);
+      expect(display.fileDiff).toMatch(/\+new text\n\+new text\n\+new text/);
       expect(display.fileName).toBe(testFile);
+      expect((result.returnDisplay as FileDiff).diffStat).toStrictEqual({
+        ai_added_lines: 3,
+        ai_removed_lines: 3,
+        user_added_lines: 0,
+        user_removed_lines: 0,
+      });
     });
 
     it('should return error if expected_replacements does not match actual occurrences', async () => {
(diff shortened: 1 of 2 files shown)


bdmorgan (approved):
LGTM!

PR #7451, opened 2025-08-30
test(auth): improve test environment variable cleanup

### LINKED ISSUE #7436: Potential flake in src/config/auth.test.ts
### What happened?

In multiple failing release workflows, we see this error:

IL  src/config/auth.test.ts > validateAuthMethod > USE_GEMINI > should return an error message if GEMINI_API_KEY is not set
AssertionError: expected null to be 'GEMINI_API_KEY environment variable n…' // Object.is equality

- Expected: 
"GEMINI_API_KEY environment variable not found. Add that to your environment and try again (no reload needed if using .env)!"

+ Received: 
null

For this test to pass consistently.

### Client information

n/a

### Login information

_No response_

_No response_

bulkypanda:
## TLDR

Properly clean up environment variables in auth tests by using vi.unstubAllEnvs() in beforeEach and explicitly stubbing variables as undefined to ensure clean test state.

## Reviewer Test Plan

run the tests and they shouldn't flake anymore

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ✅  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

- Fixes #7412
- Fixes #7436


packages/cli/src/config/auth.test.ts
```diff
@@ -18,6 +18,11 @@ vi.mock('./settings.js', () => ({
 describe('validateAuthMethod', () => {
   beforeEach(() => {
     vi.resetModules();
+    vi.unstubAllEnvs();
+    vi.stubEnv('GEMINI_API_KEY', undefined);
```
cornmander, line 21: Why do we need to stub as undefined after clearing the env stubs?

bulkypanda: defensive programming -- [vi.unstubAllEnvs()](https://vitest.dev/api/vi.html#vi-unstuballenvs) restores environment variables to their original values from when the test started. It doesn't set them to undefined. Explicitly setting them to undefined ensures all devs and CI start each test with the same clean state, regardless of what they have in their actual environment.

bulkypanda: i think the better approach here would be to get rid of the unstub line.

cornmander (approved):

PR #7462, opened 2025-08-30
Update README.md

harshityadav95:
### TLDR

This PR updates the `npx` installation command in the `README.md` to improve user experience. The previous format, which included an inline comment, often caused an error when the command was copied and pasted.

---

### Dive Deeper

The original `README.md` had the `npx` command and a comment on the same line within a single code block. When users copied this, the comment text would be included, making the command invalid and causing it to fail. This change separates the command from the comment, placing the comment on its own line above the code block. This ensures that only the command `npx https://github.com/google-gemini/gemini-cli` is copied, allowing for a seamless and error-free installation process.

---

### Reviewer Test Plan

1.  **Pull down this branch.**
2.  **Open the `README.md` file** and navigate to the "Using npx" section.
3.  **Copy the command** from the code block.
4.  **Paste the command into your terminal** and press enter.
5.  **Verify that the command executes without any errors.** The `gemini-cli` should start as expected.

---

### Testing Matrix

| | 🍏 | 🪟 | 🐧 |
| --- | --- | --- | --- |
| npm run | ❓ | ❓ | ❓ |
| npx | ✅ | ✅ | ✅ |
| Docker | ❓ | - | - |
| Podman | ❓ | - | - |
| Seatbelt | ❓ | - | - |

---


cornmander:
Thank you for your contribution. This will change npx to pull the latest version from NPM rather than from HEAD. Did you intentionally change that?

adamfweidman (approved):

PR #7566, opened 2025-09-01
refactor(core): Require model for utility calls

abhipatel12:
## TLDR

This PR refactors `GeminiClient` to require an explicit `model` parameter for the `generateContent` and `generateJson` methods. This change makes our internal API calls more explicit and is a necessary preparatory step for upcoming model routing enhancements.

Calls to these methods from within the tool implementations (`web-fetch`, `web-search`) have been updated to pass a specific model, typically `DEFAULT_GEMINI_FLASH_MODEL`.

## Dive Deeper

The primary motivation for this refactoring is to prepare the codebase for more sophisticated model routing capabilities. Currently, `config.getModel()` serves as a general default. However, the long-term vision is for `config.getModel()` to represent a model that the *user* has explicitly forced for a given session.

By requiring internal utility calls (like those in `web-fetch` and `web-search`) to specify which model they need, we avoid ambiguity and prevent these internal processes from unintentionally using a user-forced model that may not be suitable for the task. This change ensures that we have precise control over which model is used for each specific API call, paving the way for more intelligent and context-aware model selection in the future.

## Reviewer Test Plan

As this is primarily an internal refactoring, the most important validation is ensuring that existing functionality remains unchanged and all checks pass.

1.  Pull down the changes.
2.  Run the full preflight check and ensure it passes:
    ```bash
    npm run preflight
    ```
3.  All tests should pass. To be extra sure, you can run the core package tests specifically:
    ```bash
    npm run test -w @google/gemini-cli-core
    ```
4.  Manually exercise the tools that were modified to ensure they still function as expected. For each, start gemini cli and ask the following:
    *   Test web search: "Do a web search for the latest info/news about gemini and then tell me about it"
    *   Test web fetch: "use web fetch and summarize the info at https://cloud.google.com/blog/topics/developers-practitioners/gemini-cli-custom-slash-commands"

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  |✅ | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ✅| -   | -   |

### DIFF
--- a/packages/core/src/core/client.ts
+++ b/packages/core/src/core/client.ts
@@ -582,12 +582,9 @@ export class GeminiClient {
     contents: Content[],
     schema: Record<string, unknown>,
     abortSignal: AbortSignal,
-    model?: string,
+    model: string,
     config: GenerateContentConfig = {},
   ): Promise<Record<string, unknown>> {
-    // Use current model from config instead of hardcoded Flash model
-    const modelToUse =
-      model || this.config.getModel() || DEFAULT_GEMINI_FLASH_MODEL;
     try {
       const userMemory = this.config.getUserMemory();
       const systemInstruction = getCoreSystemPrompt(userMemory);
@@ -600,7 +597,7 @@ export class GeminiClient {
       const apiCall = () =>
         this.getContentGenerator().generateContent(
           {
-            model: modelToUse,
+            model,
             config: {
               ...requestConfig,
               systemInstruction,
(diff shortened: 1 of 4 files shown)


packages/core/src/core/client.ts
```diff
@@ -582,12 +582,9 @@ export class GeminiClient {
     contents: Content[],
     schema: Record<string, unknown>,
     abortSignal: AbortSignal,
-    model?: string,
+    model: string,
```
NTaylorMullen, line 585: I know it's silly but I love changes like this. Nullability/undefinedness is viral when not needed

NTaylorMullen (approved):

packages/core/src/core/client.ts
abhipatel12: Totally agree! So much easier to reason without seeing things like:
```
model || this.config.getModel() || DEFAULT_GEMINI_FLASH_MODEL
```

PR #7603, opened 2025-09-02
Change editing tool name dynamically according to config

silviojr:
## TLDR

Add the `Config` as a required argument to `getCoreSystemPrompt()` function. In this way, the function will know which editing tool name to use in the system prompt. 

## Dive Deeper

This follows https://github.com/google-gemini/gemini-cli/pull/6823 so we can keep the smartEditTool with a different name to differentiate it from the `replace` tool in the logs. 

## Reviewer Test Plan

Added new unit tests

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


NTaylorMullen (approved):
You should also do a general search for `replace` in case we didn't use the proper name eslsewhere as well

PR #7610, opened 2025-09-02
feat(oauth) - Create hybrid storage class

shishu314:
## TLDR

Create a hybrid storage class that first checks for keychain storage then falls back on file storage

## Reviewer Test Plan

Nothing to test yet since code is not used. Just run npm run preflight
## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | x  | ❓  | ❓  |
| npx      | x  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Makes progress on https://github.com/google-gemini/maintainers-gemini-cli/issues/698


packages/core/src/mcp/token-storage/hybrid-token-storage.ts
```diff
@@ -0,0 +1,104 @@
...
+        const keychainStorage = new KeychainTokenStorage(this.serviceName);
+
+        const isAvailable = await keychainStorage.isAvailable();
+        if (isAvailable) {
+          this.storage = keychainStorage;
+          this.storageType = TokenStorageType.KEYCHAIN;
+          return this.storage;
+        }
+      } catch (_e) {
+        // Fallback to file storage if keychain fails to initialize
```
cornmander, line 40: Can you add logging here?

packages/core/src/mcp/token-storage/hybrid-token-storage.ts
```diff
@@ -0,0 +1,104 @@
...
+      return this.storage;
+    }
+
+    // Use a single initialization promise to avoid race conditions
+    if (!this.storageInitPromise) {
+      this.storageInitPromise = this.initializeStorage();
+    }
+
+    // Wait for initialization to complete
+    const storage = await this.storageInitPromise;
```
cornmander: consolidate to return await ...

packages/core/src/mcp/token-storage/hybrid-token-storage.ts
```diff
@@ -0,0 +1,104 @@
...
+    const storage = await this.getStorage();
+    await storage.clearAll();
+  }
+
+  async getStorageType(): Promise<TokenStorageType> {
+    await this.getStorage();
+    return this.storageType!;
+  }
+
+  async resetStorage(): Promise<void> {
```
cornmander: This is for testing, right? Can you comment thit that way?

shishu314: Actually we don't need this code, removed.

packages/core/src/mcp/token-storage/keychain-token-storage.ts
```diff
@@ -0,0 +1,251 @@
...
+}
+
+const KEYCHAIN_TEST_PREFIX = '__keychain_test__';
+
+export class KeychainTokenStorage extends BaseTokenStorage {
+  private keychainAvailable: boolean | null = null;
+  private keytarModule: Keytar | null = null;
+  private keytarLoadAttempted = false;
+
+  async getKeytar(): Promise<Keytar | null> {
```
cornmander, line 31: I wonder if we can remove the null return type here and throw instead of logging, since all the callsites now check for null and then throw?

cornmander (approved):

emeryray2002 (approved):

PR #7612, opened 2025-09-02
feat(cli): Add --with-thoughts flag to /chat save command

### LINKED ISSUE #7265: Save thoughts as part of the chat history with a configurable knob.
This is very useful for when we want to dump chat history during debugging. Currently we filter out thoughts which means even if we dump chat history we may not understand all the edge cases in the request/response objects.

davideast:
## TLDR

Adds a `--with-thoughts` flag to the `/chat save` command. This flag saves the complete chat history, including the model's internal "thought" process, which is stripped out by default. This is primarily a debugging feature to help understand the model's reasoning.

## Dive Deeper

It's really helpful to inspect the `thought` parts of the model's output to understand its reasoning chain.

This PR addresses:
1.  Adding a `--with-thoughts` flag to the `/chat save` command.
2.  Overloading the `GeminiChat.getHistory` method to accept a `stripThoughts` option in a backward-compatible way.
3.  Updating all internal calls to `getHistory` to use the new, more explicit options object where appropriate.
4.  Adding new tests to verify that the flag correctly controls the inclusion of thoughts in the saved history.

## Reviewer Test Plan

1.  Start the CLI (`npm start`).
2.  Have a short conversation with the model.
3.  Run the save command **without** the flag: `/chat save test-no-thoughts`
4.  Inspect the saved file at `.gemini/tmp/checkpoints/checkpoint-test-no-thoughts.json`. Verify that the `parts` arrays for the model's responses **do not** contain any `"thought"` keys.
5.  Run the save command **with** the flag: `/chat save test-with-thoughts --with-thoughts`
6.  Inspect the saved file at `.gemini/tmp/checkpoints/checkpoint-test-with-thoughts.json`. Verify that the `parts` arrays for the model's responses **do** contain `"thought"` keys with corresponding reasoning strings.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ✅  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Resolves #7265

### DIFF
--- a/packages/cli/src/ui/commands/chatCommand.ts
+++ b/packages/cli/src/ui/commands/chatCommand.ts
@@ -102,10 +102,12 @@ const listCommand: SlashCommand = {
 const saveCommand: SlashCommand = {
   name: 'save',
   description:
-    'Save the current conversation as a checkpoint. Usage: /chat save <tag>',
+    'Save the current conversation as a checkpoint. Usage: /chat save <tag> [--with-thoughts]',
   kind: CommandKind.BUILT_IN,
   action: async (context, args): Promise<SlashCommandActionReturn | void> => {
-    const tag = args.trim();
+    const [tag, ...rest] = args.split(' ');
+    const withThoughts = rest.includes('--with-thoughts');
+
     if (!tag) {
       return {
         type: 'message',
@@ -145,13 +147,15 @@ const saveCommand: SlashCommand = {
       };
     }
 
-    const history = chat.getHistory();
+    const history = chat.getHistory({ stripThoughts: !withThoughts });
     if (history.length > 2) {
       await logger.saveCheckpoint(history, tag);
       return {
         type: 'message',
         messageType: 'info',
-        content: `Conversation checkpoint saved with tag: ${decodeTagName(tag)}.`,
+        content: `Conversation checkpoint saved with tag: ${decodeTagName(
+          tag,
+        )}.`,
       };
     } else {
       return {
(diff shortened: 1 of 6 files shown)


packages/cli/src/ui/commands/chatCommand.ts
```diff
@@ -102,10 +102,12 @@ const listCommand: SlashCommand = {
 const saveCommand: SlashCommand = {
   name: 'save',
   description:
-    'Save the current conversation as a checkpoint. Usage: /chat save <tag>',
+    'Save the current conversation as a checkpoint. Usage: /chat save <tag> [--with-thoughts]',
```
jacob314, line 105: for discussion: should we add this as an arg to the slash command or add a setting to settings.json that toggles whether we save checkpoints with thoughts?

abhipatel12: Drive-by thought - I think this should be a debugging setting that is maybe exposed via our `/settings`

jacob314: to be clear. Abhi and I are saying the same thing. Exposed in settings.json and editable in /settings. Gemini CLI should one shot a PR to do this if you tell it to look at @packages/cli

kyanha:
Why does this PR still have a "needs-issue" tag? https://github.com/google-gemini/gemini-cli/issues/7265 is the issue.

nayong2021:
I wonder why this PR is not merged into `main`.
I want to see the reasoning process behind Gemini Pro's responses. Additionally, I would like to develop the markdown generated by the `/chat share` command to include these thoughts as well.
This should work similarly to how we can view Gemini Pro's thoughts on the web interface.

PR #7627, opened 2025-09-03
Use IdeClient directly instead of config.ideClient

scidomino:
## TLDR

IdeClient.getInstance is already a single global instance so there's no value in config keeping it's own copy.

## Dive Deeper

I'm moving config.initialize() into the react code and this change will simplify things a lot since react needs the ideClient to be ready in a bunch of places.

## Reviewer Test Plan

Start with `"ideMode": false,` and use `/ide enable` and `/ide disable`.
Start with `"ideMode": true,` and do the same.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | x  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

For #4544

### DIFF
--- a/packages/cli/src/ui/App.tsx
+++ b/packages/cli/src/ui/App.tsx
@@ -59,7 +59,12 @@ import { ContextSummaryDisplay } from './components/ContextSummaryDisplay.js';
 import { useHistory } from './hooks/useHistoryManager.js';
 import { useInputHistoryStore } from './hooks/useInputHistoryStore.js';
 import process from 'node:process';
-import type { EditorType, Config, IdeContext } from '@google/gemini-cli-core';
+import type {
+  EditorType,
+  Config,
+  IdeContext,
+  DetectedIde,
+} from '@google/gemini-cli-core';
 import {
   ApprovalMode,
   getAllGeminiMdFilenames,
@@ -73,6 +78,7 @@ import {
   isGenericQuotaExceededError,
   UserTierId,
   DEFAULT_GEMINI_FLASH_MODEL,
+  IdeClient,
 } from '@google/gemini-cli-core';
 import type { IdeIntegrationNudgeResult } from './IdeIntegrationNudge.js';
 import { IdeIntegrationNudge } from './IdeIntegrationNudge.js';
@@ -161,10 +167,19 @@ const App = ({ config, settings, startupWarnings = [], version }: AppProps) => {
   const { history, addItem, clearItems, loadHistory } = useHistory();
 
   const [idePromptAnswered, setIdePromptAnswered] = useState(false);
-  const currentIDE = config.getIdeClient().getCurrentIde();
+  const [currentIDE, setCurrentIDE] = useState<DetectedIde | undefined>();
+
   useEffect(() => {
-    registerCleanup(() => config.getIdeClient().disconnect());
+    (async () => {
+      const ideClient = await IdeClient.getInstance();
+      setCurrentIDE(ideClient.getCurrentIde());
+    })();
+    registerCleanup(async () => {
+      const ideClient = await IdeClient.getInstance();
+      ideClient.disconnect();
+    });
   }, [config]);
+
   const shouldShowIdePrompt =
     currentIDE &&
     !config.getIdeMode() &&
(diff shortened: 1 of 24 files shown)


packages/cli/src/ui/hooks/slashCommandProcessor.ts
```diff
@@ -215,15 +216,18 @@ export const useSlashCommandProcessor = (
...
+    (async () => {
+      (await IdeClient.getInstance()).addStatusChangeListener(listener);
+    })();
 
     return () => {
-      ideClient.removeStatusChangeListener(listener);
+      (async () => {
+        (await IdeClient.getInstance()).removeStatusChangeListener(listener);
+      })();
     };
```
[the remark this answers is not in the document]
scidomino: Logging the error won't do much to improve the situation so I'll leave it as is.

skeshive:
Seeing a bug where running /ide enable does not enable the integration as expected. Given the error message, seems likely .connect() isnt being called or we're not waiting for the connection to become active

Repro steps:

Run /ide disable and restart CLI
Run /ide enable --> ✕ 🔴 Disconnected: IDE integration is currently disabled. To enable it, run
/ide enable.

scidomino:
> Seeing a bug where running /ide enable does not enable the integration as expected. Given the error message, seems likely .connect() isnt being called or we're not waiting for the connection to become active
> 
> Repro steps:
> 
> Run /ide disable and restart CLI Run /ide enable --> ✕ 🔴 Disconnected: IDE integration is currently disabled. To enable it, run /ide enable.

There was a subtle check-then-act bug in IdeClient.getInstance(). This was causing two instances of IdeClient to be created. I fixed it and now it works as expected.

scidomino:
I believe these E2E test failures are unrelated.

packages/cli/src/ui/App.tsx
```diff
@@ -161,10 +167,15 @@ const App = ({ config, settings, startupWarnings = [], version }: AppProps) => {
   const { history, addItem, clearItems, loadHistory } = useHistory();
 
   const [idePromptAnswered, setIdePromptAnswered] = useState(false);
-  const currentIDE = config.getIdeClient().getCurrentIde();
+  const [currentIDE, setCurrentIDE] = useState<DetectedIde | undefined>();
+
   useEffect(() => {
-    registerCleanup(() => config.getIdeClient().disconnect());
+    (async () => {
+      setCurrentIDE((await IdeClient.getInstance()).getCurrentIde());
```
NTaylorMullen: nit: lets not inline the await statements. Easier to debug. Same below in register cleanup.

packages/cli/src/ui/commands/aboutCommand.ts
```diff
@@ -31,7 +32,7 @@ export const aboutCommand: SlashCommand = {
     const gcpProject = process.env['GOOGLE_CLOUD_PROJECT'] || '';
     const ideClient =
       (context.services.config?.getIdeMode() &&
-        context.services.config?.getIdeClient()?.getDetectedIdeDisplayName()) ||
+        (await IdeClient.getInstance())?.getDetectedIdeDisplayName()) ||
```
NTaylorMullen: In general I'd avoid inlining `await` clauses because they aren't very debbugable

NTaylorMullen (approved):
Approving to unblock, should wait for an approval from @skeshive

skeshive (approved):

packages/cli/src/ui/App.tsx
scidomino: Fair point. Fixed this one and all the others I introduced.

Amnat-auyana1998 (dismissed):
@scidomino

packages/cli/src/ui/hooks/slashCommandProcessor.ts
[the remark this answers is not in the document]
Amnat-auyana1998: **### _

> เขียน api เพิ่ม ฟืล inlineQr 

> 

> ให้บอท
[``](url)
 #_******

Issue #6846, opened 2025-08-22
Unable to apply changes via CLI with IDE integration enabled


skeshive:
### What happened?

When using the VS Code IDE integration, a diff is presented in the IDE with the option to "Apply this change?" in the CLI. After using the arrow keys to select "Yes, allow once" and pressing Enter, the command is not executed. Instead of applying the code change, the action is cancelled, and the file remains unmodified.

### What did you expect to happen?

I expected the "Yes, allow once" command in the CLI to apply the changes from the diff directly to the file and save them. 

### Client information

<details>

```console
$ gemini /about
# paste output here
```

</details>

skeshive:
Note: this was reported by @galz10 but I'm unable to repro

Jopo-JP:
Can confirm it, it simply freezes and after a minute it works, sometimes the write simply fails as if i aborted the changes.

I think the Gemini CLI Intergration seems to have some issues:
✖ [ERROR] [IDEClient] IDE connection error. The connection was lost unexpectedly. Please try reconnecting by running /ide enable

I'm using VSCode.

hugomurillomtz:
> Can confirm it, it simply freezes and after a minute it works, sometimes the write simply fails as if i aborted the changes.
> 
> I think the Gemini CLI Intergration seems to have some issues: ✖ [ERROR] [IDEClient] IDE connection error. The connection was lost unexpectedly. Please try reconnecting by running /ide enable
> 
> I'm using VSCode.

May I know if you're attempting to modify a big file? Maybe only fails in those cases. I'm trying to reproduce it, but haven't been able to.

hugomurillomtz:
Haven't been able to reproduce it either.

Jopo-JP:
Not really. I did edit smaller typescript files or markdown files.

hugomurillomtz:
> Not really. I did edit smaller typescript files or markdown files.

May I know which Gemini CLI version you're using?

Jopo-JP:
> > Not really. I did edit smaller typescript files or markdown files.
>
> May I know which Gemini CLI version you're using? 

Of course. I still use version 0.2.1 a few days ago. But since yesterday there is 0.2.2 and have updated Gemini CLI accordingly. But the error persists. If I run Gemini Cli without IDE, it can normally store everything. 

[...]

PR #7633, opened 2025-09-03
fix(ide): prevent race condition when diff accepted through CLI

skeshive:
## TLDR
This pull request introduces a `suppressNotification` flag to the `closeDiff` functionality. This allows the client to control whether the IDE sends a notification when a diff view is closed. This is primarily used to prevent incorrect notifications when the diff is being resolved directly from the CLI (e.g., via `/accept` or `/reject`), as the client is already handling the resolution manually. 

Note: This PR only contains the VS Code extension changes to ensure backward compatibility. The corresponding CLI-side changes will be released separately after this update is live.

## Linked issues / bugs
#6846


bdmorgan (approved):
LGTM

PR #7634, opened 2025-09-03
refactor: Centralize and improve model fallback handling

### LINKED ISSUE #7810: Refactor Fallback Logic out of frontend
Currently, the logic to set fallback and the fallback model is in both the cli and core packages. This means that the `config.getModel()` function is used to track multiple concepts (model set by user and fallback). 

As we add routing, we need to consolidate this to the core package and ensure that the model string used in the config only represents either the explicit model choice or the default routing mechanism

abhipatel12:
## TLDR

This PR refactors the model fallback mechanism, preparing the architecture for upcoming model routing strategies (e.g., "auto" mode).

The previous implementation suffered from duplicated logic, tight coupling where the CLI package mutated Core state, and ambiguity in the communication contract. This PR resolves these issues by centralizing the logic, introducing explicit intents, and implementing key architectural patterns for decoupling.

**Key changes for reviewers:**
-   `packages/core/src/fallback/`: New centralized handler and types.
-   `packages/cli/src/ui/hooks/useQuotaAndFallback.ts`: New hook encapsulating all UI logic for this feature.
-   `packages/cli/src/ui/AppContainer.tsx`: Significantly simplified by removing the inline fallback handler.
-   `packages/core/src/core/client.ts` & `geminiChat.ts`: Updated to use the new centralized handler.

### Model Resolution

We no longer mutate `config.setModel()` during runtime fallbacks.

In an interim state, until we have a more complex router service, we check `config.isInFallbackMode()`; if true, we return the fallback model (Flash); otherwise, we return the configured model.

### Centralized, Intent-Driven Fallback

The logic previously duplicated in `client.ts` and `geminiChat.ts` is now handled by a single `handleFallback` function.

The new flow is:
1.  An API call fails (e.g., 429). The specific, concrete model used for that attempt (`currentAttemptModel`) is captured via closure.
2.  The `onPersistent429` callback calls `handleFallback(config, currentAttemptModel, ...)`.
3.  `handleFallback` invokes the UI handler (injected via `config.setFallbackHandler`).
4.  The UI (`App.tsx`) handles user interaction (e.g., Pro Quota Dialog) and returns a `FallbackIntent`.
5.  `handleFallback` interprets this intent:
    *   `'retry'`: Calls `config.setFallbackMode(true)` and signals `retryWithBackoff` to immediately retry the request. The Model Resolver ensures the retry uses the fallback model.
    *   `'stop'`: Calls `config.setFallbackMode(true)` but stops the current request.
    *   `'auth'`: Stops the current request without activating fallback mode.

Crucially, the Core package now owns the responsibility of changing the fallback state; the UI only signals intent.

## Reviewer Test Plan

To test this, you will need to simulate quota errors. The easiest way is to temporarily modify `packages/core/src/core/geminiChat.ts` in the `sendMessage` method to always throw an error that looks like a 429.

Find the `apiCall` function and replace its content with:
```typescript
throw Object.assign(new Error('API Error 429: Quota exceeded'), {
  status: 429,
  errors: [{ reason: 'rateLimitExceeded' }] // For generic quota
  // OR
  // errors: [{ reason: 'userRateLimitExceeded' }] // For Pro quota
});
```

**Scenario 1: Generic Quota Fallback (Automatic)**
1.  Use the `rateLimitExceeded` reason in your mock error.
2.  Run the CLI and submit a prompt.
3.  **Expect:** An informational message appears in the chat history, stating that you've been switched to the fallback model. The prompt is **not** retried.
4.  **Expect:** Subsequent prompts are sent to the `gemini-flash` model. The model indicator in the UI should update.

**Scenario 2: Pro Quota Fallback (Interactive)**
1.  Use the `userRateLimitExceeded` reason in your mock error.
2.  Run the CLI and submit a prompt.
3.  **Expect:** The interactive "Pro quota limit reached" dialog appears.

**Scenario 2a: Continue with Fallback**
1.  From the dialog, select "Continue with fallback model".
2.  **Expect:** The dialog closes, an informational message about the switch is added to history, and the original prompt is **successfully retried** using the `gemini-flash` model. You should see a valid response.

**Scenario 2b: Authenticate**
1.  From the dialog, select "Authenticate with a different method".
2.  **Expect:** The dialog closes, and the `/auth` dialog immediately opens. The original prompt is **not** retried.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  |✅ | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt |✅ | -   | -   |

## Linked issues / bugs

Resolves #7810

### DIFF
--- a/packages/a2a-server/src/agent/task.ts
+++ b/packages/a2a-server/src/agent/task.ts
@@ -89,11 +89,10 @@ export class Task {
     this.completedToolCalls = [];
     this._resetToolCompletionPromise();
     this.config.setFallbackHandler(
-      async (currentModel: string, fallbackModel: string): Promise<'stop'> => {
-        config.setModel(fallbackModel); // gemini-cli-core sets to DEFAULT_GEMINI_FLASH_MODEL
-        // Switch model for future use but return false to stop current retry
-        return 'stop';
-      },
+      // For a2a-server, we want to automatically switch to the fallback model
+      // for future requests without retrying the current one. The 'stop'
+      // intent achieves this.
+      async () => 'stop',
     );
   }
 
--- a/packages/a2a-server/src/agent/task.ts
+++ b/packages/a2a-server/src/agent/task.ts
@@ -88,11 +88,11 @@ export class Task {
     this.eventBus = eventBus;
     this.completedToolCalls = [];
     this._resetToolCompletionPromise();
-    this.config.setFlashFallbackHandler(
-      async (currentModel: string, fallbackModel: string): Promise<boolean> => {
+    this.config.setFallbackHandler(
+      async (currentModel: string, fallbackModel: string): Promise<'stop'> => {
         config.setModel(fallbackModel); // gemini-cli-core sets to DEFAULT_GEMINI_FLASH_MODEL
         // Switch model for future use but return false to stop current retry
-        return false;
+        return 'stop';
       },
     );
   }
--- a/packages/a2a-server/src/agent/task.ts
+++ b/packages/a2a-server/src/agent/task.ts
@@ -88,7 +88,7 @@ export class Task {
     this.eventBus = eventBus;
     this.completedToolCalls = [];
     this._resetToolCompletionPromise();
-    this.config.setFallbackHandler(
+    this.config.setFallbackModelHandler(
       // For a2a-server, we want to automatically switch to the fallback model
       // for future requests without retrying the current one. The 'stop'
       // intent achieves this.
--- a/packages/cli/src/ui/AppContainer.test.tsx
+++ b/packages/cli/src/ui/AppContainer.test.tsx
@@ -4,31 +4,55 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+import {
+  describe,
+  it,
+  expect,
+  vi,
+  beforeEach,
+  afterEach,
+  type Mock,
+} from 'vitest';
 import { render, cleanup } from 'ink-testing-library';
 import { AppContainer } from './AppContainer.js';
 import { type Config, makeFakeConfig } from '@google/gemini-cli-core';
 import type { LoadedSettings } from '../config/settings.js';
 import type { InitializationResult } from '../core/initializer.js';
+import { useQuotaAndFallback } from './hooks/useQuotaAndFallback.js';
+import { UIStateContext, type UIState } from './contexts/UIStateContext.js';
+import {
+  UIActionsContext,
+  type UIActions,
+} from './contexts/UIActionsContext.js';
+import { useContext } from 'react';
 
-// Mock App component to isolate AppContainer testing
+// Helper component will read the context values provided by AppContainer
+// so we can assert against them in our tests.
+let capturedUIState: UIState;
+let capturedUIActions: UIActions;
+function TestContextConsumer() {
+  capturedUIState = useContext(UIStateContext)!;
+  capturedUIActions = useContext(UIActionsContext)!;
+  return null;
+}
+
 vi.mock('./App.js', () => ({
-  App: () => 'App Component',
+  App: TestContextConsumer,
 }));
 
-// Mock all the hooks and utilities
-vi.mock('./hooks/useHistory.js');
+vi.mock('./hooks/useQuotaAndFallback.js');
+vi.mock('./hooks/useHistoryManager.js');
 vi.mock('./hooks/useThemeCommand.js');
-vi.mock('./hooks/useAuthCommand.js');
+vi.mock('./auth/useAuth.js');
 vi.mock('./hooks/useEditorSettings.js');
 vi.mock('./hooks/useSettingsCommand.js');
-vi.mock('./hooks/useSlashCommandProcessor.js');
+vi.mock('./hooks/slashCommandProcessor.js');
 vi.mock('./hooks/useConsoleMessages.js');
 vi.mock('./hooks/useTerminalSize.js', () => ({
   useTerminalSize: vi.fn(() => ({ columns: 80, rows: 24 })),
 }));
 vi.mock('./hooks/useGeminiStream.js');
-vi.mock('./hooks/useVim.js');
+vi.mock('./hooks/vim.js');
 vi.mock('./hooks/useFocus.js');
 vi.mock('./hooks/useBracketedPaste.js');
 vi.mock('./hooks/useKeypress.js');
@@ -40,7 +64,7 @@ vi.mock('./hooks/useWorkspaceMigration.js');
 vi.mock('./hooks/useGitBranchName.js');
 vi.mock('./contexts/VimModeContext.js');
 vi.mock('./contexts/SessionContext.js');
-vi.mock('./hooks/useTextBuffer.js');
+vi.mock('./components/shared/text-buffer.js');
 vi.mock('./hooks/useLogger.js');
 
 // Mock external utilities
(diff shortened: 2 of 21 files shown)


packages/a2a-server/src/agent/task.ts
```diff
@@ -88,11 +88,11 @@ export class Task {
...
     this._resetToolCompletionPromise();
-    this.config.setFlashFallbackHandler(
-      async (currentModel: string, fallbackModel: string): Promise<boolean> => {
+    this.config.setFallbackHandler(
+      async (currentModel: string, fallbackModel: string): Promise<'stop'> => {
         config.setModel(fallbackModel); // gemini-cli-core sets to DEFAULT_GEMINI_FLASH_MODEL
         // Switch model for future use but return false to stop current retry
-        return false;
+        return 'stop';
       },
```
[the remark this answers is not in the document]
abhipatel12: Done

packages/core/src/fallback/handler.ts
```diff
@@ -0,0 +1,59 @@
...
+    return null;
+  }
+}
+
+function activateFallbackMode(config: Config, authType: string) {
+  if (!config.isInFallbackMode()) {
+    config.setFallbackMode(true);
+    logFlashFallback(config, new FlashFallbackEvent(authType));
+  }
+}
```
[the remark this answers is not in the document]
abhipatel12: Done

packages/cli/src/ui/hooks/useQuotaAndFallback.ts
```diff
@@ -0,0 +1,168 @@
...
+      }
+    },
+    [proQuotaRequest, setAuthState, historyManager],
+  );
+
+  return {
+    proQuotaRequest,
+    handleProQuotaChoice,
+  };
+}
```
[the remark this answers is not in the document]
abhipatel12: Done

packages/cli/src/ui/AppContainer.tsx
```diff
@@ -167,18 +171,17 @@ export const AppContainer = (props: AppContainerProps) => {
...
       }
     };
 
-    // Check immediately and then periodically
     checkModelChange();
     const interval = setInterval(checkModelChange, 1000); // Check every second
 
     return () => clearInterval(interval);
-  }, [config, currentModel]);
+  }, [config, currentModel, getEffectiveModel]);
```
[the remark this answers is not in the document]
abhipatel12: This is something I want to do, but is not in scope of this PR. I have made a note to do this in a follow-up

packages/cli/src/ui/AppContainer.test.tsx
```diff
@@ -49,14 +73,154 @@ vi.mock('../utils/handleAutoUpdate.js');
 vi.mock('./utils/ConsolePatcher.js');
 vi.mock('../utils/cleanup.js');
 
+// Import the mocked hooks so you can provide implementations
```
jacob314: nit: remove this comment

packages/cli/src/ui/AppContainer.test.tsx
```diff
@@ -49,14 +73,154 @@ vi.mock('../utils/handleAutoUpdate.js');
...
+import { useTextBuffer } from './components/shared/text-buffer.js';
+import { useLogger } from './hooks/useLogger.js';
+import { useLoadingIndicator } from './hooks/useLoadingIndicator.js';
+
 describe('AppContainer State Management', () => {
   let mockConfig: Config;
   let mockSettings: LoadedSettings;
   let mockInitResult: InitializationResult;
 
+  // Create typed mocks for all hooks
```
jacob314, line 101: this is a massive amount of additional mocking. is this really needed?

packages/cli/src/ui/AppContainer.test.tsx
```diff
@@ -325,7 +489,75 @@ describe('AppContainer State Management', () => {
...
+      expect(capturedUIActions.handleProQuotaChoice).toBe(mockHandler);
+
+      // You can even verify that the plumbed function is callable
+      capturedUIActions.handleProQuotaChoice('auth');
+      expect(mockHandler).toHaveBeenCalledWith('auth');
+    });
+  });
 });
 
 // TODO: Add comprehensive integration test once all hook mocks are complete
```
jacob314: remove the comment on line 563 as well. this comment should never have been checked in to begin with.

packages/core/src/config/config.ts
```diff
@@ -281,7 +276,7 @@ export class Config {
     name: string;
     extensionName: string;
   }>;
-  flashFallbackHandler?: FlashFallbackHandler;
+  fallbackHandler?: FallbackHandler;
```
jacob314: optional nit: like the handler doesn't have the name Flash in it anymore but worry it might be a little confusing now. should this be fallbackModelHandler to clarify what kind of fallback it is?

packages/core/src/core/client.test.ts
```diff
@@ -2210,32 +2243,29 @@ ${JSON.stringify(
...
+        requestedModel,
       );
 
-      expect(result).toBe(fallbackModel);
-
-      expect(mockFallbackHandler).toHaveBeenCalledWith(
-        currentModel,
-        fallbackModel,
-        undefined,
+      // Assert that the Flash model was used, not the requested model
```
jacob314: nit: remove this comment as it is low value.

packages/core/src/core/geminiChat.test.ts
```diff
@@ -76,17 +96,21 @@ describe('GeminiChat', () => {
...
+    // Default mock implementation for tests that don't care about retry logic
+    mockRetryWithBackoff.mockImplementation(async (apiCall) => apiCall());
     mockConfig = {
       getSessionId: () => 'test-session-id',
       getTelemetryLogPromptsEnabled: () => true,
       getUsageStatisticsEnabled: () => true,
       getDebugMode: () => false,
-      getContentGeneratorConfig: () => ({
-        authType: 'oauth-personal',
+      getContentGeneratorConfig: vi.fn().mockReturnValue({
```
jacob314, line 107: nit: why is this now mockReturnValue when it wasn't before?

packages/core/src/core/geminiChat.test.ts
```diff
@@ -1476,11 +1500,177 @@ describe('GeminiChat', () => {
...
+        chat.sendMessage({ message: 'test stop' }, 'prompt-id-fb2'),
+      ).rejects.toThrow(error429);
+
+      expect(mockContentGenerator.generateContent).toHaveBeenCalledTimes(1);
+      expect(mockHandleFallback).toHaveBeenCalledTimes(1);
+    });
+  });
+
   it('should discard valid partial content from a failed attempt upon retry', async () => {
-    // ARRANGE: Mock the stream to fail on the first attempt after yielding some valid content.
```
jacob314, line 1480: nit: were these comments intentionally removed? they aren't great but do add some clarity

packages/core/src/fallback/handler.ts
```diff
@@ -0,0 +1,61 @@
...
+        activateFallbackMode(config, authType);
+        return true; // Signal retryWithBackoff to continue.
+
+      case 'stop':
+        activateFallbackMode(config, authType);
+        return false;
+
+      case 'auth':
+      default:
+        return false;
```
jacob314, line 49: can we throw if an unexpected value is received rather than quitely treating it like the 'auth case'?

packages/core/src/fallback/handler.ts
```diff
@@ -0,0 +1,61 @@
...
+    }
+  } catch (handlerError) {
+    console.warn('Fallback UI handler failed:', handlerError);
```
jacob314: why is this just .warn instead of .error?

packages/core/src/utils/flashFallback.integration.test.ts
```diff
@@ -17,10 +17,13 @@ import {
 import { DEFAULT_GEMINI_FLASH_MODEL } from '../config/models.js';
 import { retryWithBackoff } from './retry.js';
 import { AuthType } from '../core/contentGenerator.js';
+// Import the new types (Assuming this test file is in packages/core/src/utils/)
+import type { FallbackHandler } from '../fallback/types.js';
 
 vi.mock('node:fs');
 
-describe('Flash Fallback Integration', () => {
```
jacob314, line 23: not a new issue but for discussion, should this be renamed to flashFallback.test.ts
seems much more like a unit test than a regular integration test with lots of mocking.

jacob314 (commented):
Looks solid. A few minor comments then approved. Please be sure to land the fixes as a separate commit.
Thanks for adding all of the extra testing as part of the refactor.

packages/cli/src/ui/AppContainer.test.tsx
abhipatel12: Done

packages/cli/src/ui/AppContainer.test.tsx
abhipatel12: As per offline convo: If i wanted to test my new hook was actually wired up i had to have a minimal mock for all of them. I assume in the container we're just trying to see if the wiring is correct. 

For now, we can keep this but have made a note to revisit and add better integration testing.

packages/cli/src/ui/AppContainer.test.tsx
abhipatel12: Done

packages/core/src/config/config.ts
abhipatel12: Good point. Made it clearer

packages/core/src/core/client.test.ts
abhipatel12: Done

packages/core/src/core/geminiChat.test.ts
abhipatel12: We needed it so that we could test the fallback integration. These things weren't being tested previously which is why it was needed. We could use a fake Config instead, but would need to tests more complex by calling the real `refreshAuth`, etc.

packages/core/src/core/geminiChat.test.ts
abhipatel12: I had removed them b/c they looked like AI-isms. But added them back if you feel like they are helpful.

packages/core/src/fallback/handler.ts
abhipatel12: Whoop, copypasta. This was a direct copy of the existing try catch from the client.ts code. Changed to .error

packages/core/src/fallback/handler.ts
abhipatel12: Done

packages/core/src/utils/flashFallback.integration.test.ts
abhipatel12: That's fair. Done

abhipatel12 (commented):
Thanks for the review. Made the changes!

jacob314 (approved):
lgtm

PR #7642, opened 2025-09-03
feat: Introduce Patch tool for applying unified diffs

yongbingchen-a11y:
feat: Introduce Patch tool for applying unified diffs

    For LLM-generated code to converge on large projects, the model must
    make focused changes rather than attempting full rewrites. This patch-based
    approach is key to iterative refinement. LLMs are natively proficient at
    this, as they are trained on vast datasets of code from version control
    repositories, making the generation of a standard unified diff a natural
    and robust capability.

    On the other hand, the existing `replace` tool is fundamentally brittle from an
    engineering perspective. It forces the LLM to simultaneously manage two
    distinct concerns: the complex syntax of the code itself (like Python
    indentation) and the formatting characters of the string literal (like `\n`)
    when composing a single, multi-line string. This dual responsibility is a frequent
    source of correctness errors and is a flawed approach for generating reliable code.

    This commit replaces the existing 'replace' tool with a new `patch` tool that
    applies standard unified diffs directly. This creates a proper separation
    of concerns, allowing the LLM to focus on generating correct code while the
    agent handles the mechanical application.

    Key benefits include:

    * **Improves Correctness**: By using a structured format, it prevents the
        common errors associated with string manipulation and escaping, leading
        to more reliable code generation.
    * **Workflow Efficiency**: Applies complex, multi-file changes
        (including file creation and deletion) in a single operation,
        dramatically reducing the number of LLM interactions.
    * **Robust Application**: Uses a content-based matching algorithm that is
        resilient to line number shifts, resulting in a much higher success
        rate for applying patches.
    * **Self-Healing Patches**: Automatically uses an LLM to fix hunks that
        fail due to context mismatches, further increasing reliability.
    * **Iterative Refinement**: Applies all successful hunks even if others
        fail, returning a clean diff of only the failures. This enables a
        powerful workflow where the LLM only needs to correct what went wrong.

## Dive Deeper

Assume I have a Python code snippet like this:
```
def test_multiple_hunks():
  """Original test - kept for compatibility."""
  test_instance = TestPatchBehaviors()
  content = """import os

def func1():
    pass

def func2():
    pass
"""
  file_path = test_instance.create_temp_file(content)

  try:
    diff_str = f"""--- a/test.py
+++ b/test.py
@@ -1,3 +1,4 @@
 import os
+import sys

 def func1():

@@ -3,4 +4,5 @@
 def func1():
+    print("func1")
     pass

"""

    result = diff_patch(diff_str, working_directory=os.getcwd())
    assert "SUCCESS" in result
  finally:
    test_instance.cleanup()
 ```
 Which has strings inside strings, and I want to edit some line inside the string block, while maintain its correctness, any string based solution would be very hard to achieve the correct editing -- this is inherently a work that suitable for a deterministic tool (like the diff patch tool), not for LLM (to acting like an editor), as an intelligent entity.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

### DIFF
--- a/packages/core/src/services/fileSystemService.ts
+++ b/packages/core/src/services/fileSystemService.ts
@@ -25,6 +25,21 @@ export interface FileSystemService {
    * @param content - The content to write
    */
   writeTextFile(filePath: string, content: string): Promise<void>;
+
+  /**
+   * Deletes a file.
+   *
+   * @param filePath - The path to the file to delete.
+   */
+  unlink(filePath: string): Promise<void>;
+
+  /**
+   * Creates a directory.
+   *
+   * @param dirPath - The path to the directory to create.
+   * @param options - Options, e.g., { recursive: true }.
+   */
+  mkdir(dirPath: string, options?: { recursive: boolean }): Promise<void>;
 }
 
 /**
@@ -37,5 +52,16 @@ export class StandardFileSystemService implements FileSystemService {
 
   async writeTextFile(filePath: string, content: string): Promise<void> {
     await fs.writeFile(filePath, content, 'utf-8');
+  }
+
+  async unlink(filePath: string): Promise<void> {
+    await fs.unlink(filePath);
+  }
+
+  async mkdir(
+    dirPath: string,
+    options?: { recursive: boolean },
+  ): Promise<void> {
+    await fs.mkdir(dirPath, options);
   }
 }
--- a/packages/core/src/utils/patcher.test.ts
+++ b/packages/core/src/utils/patcher.test.ts
@@ -1,0 +1,446 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * @fileoverview Comprehensive test suite for the diff_patch tool.
+ * This suite validates all behaviors, including fuzzy matching, error handling,
+ * file creation, and multi-file patches.
+ */
+
+import { describe, it, expect, beforeEach, afterEach } from 'vitest';
+import * as fs from 'node:fs/promises';
+import * as path from 'node:path';
+import * as os from 'node:os';
+import { parse, applyPatchesToFS } from './patcher.js';
+import type { PatcherConfig } from './patcher.js';
+
+describe('TestPatchBehaviors', () => {
+  let tempDir: string;
+  let mockConfig: PatcherConfig;
+
+  // Setup: Create a temporary directory before each test
+  beforeEach(async () => {
+    tempDir = await fs.mkdtemp(path.join(os.tmpdir(), 'patch-test-'));
+    mockConfig = {
+      getTargetDir: () => tempDir,
+      isPathWithinWorkspace: (p: string) => {
+        const relative = path.relative(tempDir, p);
+        return !relative.startsWith('..') && !path.isAbsolute(relative);
+      },
+    };
+  });
+
+  // Teardown: Clean up the temporary directory after each test
+  afterEach(async () => {
+    await fs.rm(tempDir, { recursive: true, force: true });
+  });
+
+  /**
+   * Helper to create a file within the temporary test directory.
+   * @param content The content of the file.
+   * @param filename The name of the file.
+   * @return The full path to the created file.
+   */
+  async function createTempFile(
+    content: string,
+    filename = 'test.ts',
+  ): Promise<string> {
+    const filePath = path.join(tempDir, filename);
+    await fs.writeFile(filePath, content);
+    return filePath;
+  }
+
+  /**
+   * Applies standard unified diff format patches. It ignores line numbers and uses
+   * content-based matching for high accuracy.
+   *
+   * This is the main entry point for the patcher library.
+   */
+  async function diffPatch(patchesStr: string): Promise<string> {
+    const fileHunks = parse(patchesStr);
+    const ret = await applyPatchesToFS(fileHunks, mockConfig);
+    return ret;
+  }
+
+  it('should completely ignore line numbers', async () => {
+    const content = `function function_a(): string {
+    return "a";
+}
+
+function function_b(): string {
+    return "b";
+}
+
+function function_c(): string {
+    return "c";
+}`;
+    await createTempFile(content, 'test.ts');
+
+    // Patch with completely wrong line numbers (999,999) but GOOD CONTEXT
+    const diffStr = `--- a/test.ts
++++ b/test.ts
+@@ -999,3 +999,3 @@
+ function function_b(): string {
+-    return "b";
++    return "modified_b";
+ }`;
+
+    const result = await diffPatch(diffStr);
+    expect(result).toContain('SUCCESS');
+
+    const modified = await fs.readFile(path.join(tempDir, 'test.ts'), 'utf-8');
+    expect(modified).toContain('return "modified_b";');
+    expect(modified).not.toContain('return "b"');
+  });
+
+  it('should process hunks as atomic units', async () => {
+    const content = `class MyClass {
+    constructor() {
+        this.value = 0;
+        this.name = "test";
+    }
+
+    process() {
+        return this.value;
+    }
+}`;
+    await createTempFile(content);
+
+    const diffStr = `--- a/test.ts
++++ b/test.ts
+@@ -2,5 +2,6 @@
+     constructor() {
+-        this.value = 0;
+-        this.name = "test";
++        this.value = 42;
++        this.name = "modified";
++        this.description = "new field";
+     }
+
+     process() {
+         return this.value;
+`;
+
+    const result = await diffPatch(diffStr);
+    expect(result).toContain('SUCCESS');
+
+    const modified = await fs.readFile(path.join(tempDir, 'test.ts'), 'utf-8');
+    expect(modified).toContain('this.value = 42;');
+    expect(modified).toContain('this.name = "modified";');
+    expect(modified).toContain('this.description = "new field"');
+  });
+
+  it('should create a new file', async () => {
+    const newFileName = 'new_module.ts';
+    const diffStr = `--- /dev/null
++++ b/${newFileName}
+@@ -0,0 +1,5 @@
++/** A new module. */
++export class NewClass {
++  public static run(): string {
++    return 'new';
++  }
++}`;
+    const result = await diffPatch(diffStr);
+    expect(result).toContain('SUCCESS');
+
+    const newFilePath = path.join(tempDir, newFileName);
+    const content = await fs.readFile(newFilePath, 'utf-8');
+    expect(content).toContain('export class NewClass');
+  });
+
+  it('should delete a file', async () => {
+    const content = 'This file should be deleted.';
+    const filePath = await createTempFile(content);
+
+    const diffStr = `--- a/test.ts
++++ /dev/null
+@@ -1 +0,0 @@
+-This file should be deleted.
+`;
+    const result = await diffPatch(diffStr);
+    expect(result).toContain('SUCCESS');
+    expect(result).toContain('Deleted file');
+
+    await expect(fs.access(filePath)).rejects.toThrow();
+  });
+
+  it('should handle multiple files being processed', async () => {
+    await createTempFile('content for file1', 'file1.ts');
+    await createTempFile('content for file2', 'file2.ts');
+
+    const diffStr = `--- a/file1.ts
++++ b/file1.ts
+@@ -1,1 +1,1 @@
+-content for file1
++modified content for file1
+--- a/file2.ts
++++ b/file2.ts
+@@ -1,1 +1,1 @@
+-content for file2
++modified content for file2`;
+
+    const result = await diffPatch(diffStr);
+    expect(result).toContain('2/2 files patched successfully');
+    expect(result).toContain('SUCCESSFULLY for file1.ts');
+    expect(result).toContain('SUCCESSFULLY for file2.ts');
+
+    const content1 = await fs.readFile(path.join(tempDir, 'file1.ts'), 'utf-8');
+    expect(content1).toBe('modified content for file1');
+    const content2 = await fs.readFile(path.join(tempDir, 'file2.ts'), 'utf-8');
+    expect(content2).toBe('modified content for file2');
+  });
+
+  it('should handle graceful failure for unmatchable patches', async () => {
+    const content = `function existing_function() {
+    return "exists";
+}`;
+    await createTempFile(content);
+
+    const diffStr = `--- a/test.ts
++++ b/test.ts
+@@ -1,3 +1,3 @@
+ class CompletelyDifferentClass {
+-    constructor() {
++    constructor(value: any) {
+         this.data = "something";
+ }`;
+    const result = await diffPatch(diffStr);
+    expect(result).toContain('FAILED');
+    expect(result).toContain('Could not locate context');
+
+    const contentAfter = await fs.readFile(
+      path.join(tempDir, 'test.ts'),
+      'utf-8',
+    );
+    expect(contentAfter).toBe(content); // File should be unchanged
+  });
+
+  it('should preserve file structure on partial error', async () => {
+    const content = `function important_function() {
+    // This function must not be corrupted
+    return "critical_data";
+}
+
+function another_function() {
+    return "other_data";
+}`;
+    await createTempFile(content);
+
+    const diffStr = `--- a/test.ts
++++ b/test.ts
+@@ -1,4 +1,5 @@
+ function important_function() {
+     // This function must not be corrupted
++    // Added a comment
+     return "critical_data";
+---
+-a/test.ts
++++ b/test.ts
+@@ -999,3 +999,3 @@
+ def nonexistent_function():
+-    return "this does not exist"
++    return "this also does not exist"`;
+
+    const result = await diffPatch(diffStr);
+    expect(result).toContain('PARTIAL SUCCESS');
+
+    const modified = await fs.readFile(path.join(tempDir, 'test.ts'), 'utf-8');
+    expect(modified).toContain('// Added a comment');
+    expect(modified).toContain('return "critical_data"');
+  });
+
+  it('should reject a patch trying to access outside the working directory', async () => {
+    const diffStr = `--- a/../test.ts
++++ b/../test.ts
+@@ -1,1 +1,1 @@
+-foo
++bar
+`;
+    const result = await diffPatch(diffStr);
+    expect(result).toContain('FAILED: ../test.ts');
+    expect(result).toContain('Security violation');
+  });
+
+  it('should show failed hunks for manual application', async () => {
+    await createTempFile('line one\nline two\nline three\n');
+    const badHunk = `@@ -1,3 +1,3 @@
+ line one
+-something that does not exist
++something new
+ line three`;
+    const diffStr = `--- a/test.ts\n+++ b/test.ts\n${badHunk}`;
+
+    const result = await diffPatch(diffStr);
+
+    expect(result).toContain('FAILED HUNKS FOR MANUAL APPLICATION');
+    expect(result).toContain(badHunk);
+  });
+
+  it('should skip a hunk if it appears to be already applied', async () => {
+    const content = 'const x = 2;';
+    await createTempFile(content);
+
+    const diffStr = `--- a/test.ts
++++ b/test.ts
+@@ -1,1 +1,1 @@
+-const x = 1;
++const x = 2;`;
+
+    const result = await diffPatch(diffStr);
+    // The tool should see 'const x = 2' is already present and skip.
+    expect(result).toContain('SUCCESS');
+    expect(result).not.toContain('FAILED');
+
+    const finalContent = await fs.readFile(
+      path.join(tempDir, 'test.ts'),
+      'utf-8',
+    );
+    expect(finalContent).toBe(content);
+  });
+
+  it('should apply multiple hunks in reverse order', async () => {
+    const content = `// Hunk 2 Target
+function second() {
+    return 'second';
+}
+
+// Hunk 1 Target
+function first() {
+    return 'first';
+}`;
+    await createTempFile(content);
+
+    const diffStr = `--- a/test.ts
++++ b/test.ts
+@@ -6,3 +6,4 @@
+ // Hunk 1 Target
+ function first() {
+-    return 'first';
++    return 'first modified';
++    // Comment added
+}
+@@ -1,4 +1,4 @@
+ // Hunk 2 Target
+ function second() {
+-    return 'second';
++    return 'second modified';
+ }
+`;
+
+    const result = await diffPatch(diffStr);
+    expect(result).toContain('SUCCESS');
+
+    const modified = await fs.readFile(path.join(tempDir, 'test.ts'), 'utf-8');
+    expect(modified).toContain('first modified');
+    expect(modified).toContain('second modified');
+  });
+
+  it('should handle pure insertions with context', async () => {
+    const content = `class DatabaseManager {
+    constructor() {}
+
+    disconnect() {}
+}`;
+    await createTempFile(content);
+
+    const diffStr = `--- a/test.ts
++++ b/test.ts
+@@ -2,3 +2,7 @@
+     constructor() {}
+
++    connect() {
++        // new connection logic
++    }
++
+     disconnect() {}
+ }`;
+
+    const result = await diffPatch(diffStr);
+    expect(result).toContain('SUCCESS');
+
+    const modified = await fs.readFile(path.join(tempDir, 'test.ts'), 'utf-8');
+    expect(modified).toContain('connect()');
+  });
+
+  it('should tolerate whitespace and formatting variations', async () => {
+    const content = `function process(data: any) {
+    if   (data   ===   null) {
+        return false;
+    }
+    return true;
+}`;
+    await createTempFile(content);
+
+    // This patch has normalized whitespace compared to the original content
+    const diffStr = `--- a/test.ts
++++ b/test.ts
+@@ -1,5 +1,6 @@
+ function process(data: any) {
+-    if   (data   ===   null) {
++    if (data === null) {
++        console.log('Data was null');
+         return false;
+     }
+     return true;
+`;
+    const result = await diffPatch(diffStr);
+    expect(result).toContain('SUCCESS');
+
+    const modified = await fs.readFile(path.join(tempDir, 'test.ts'), 'utf-8');
+    expect(modified).toContain("console.log('Data was null')");
+  });
+
+  it('should be compatible with git format-patch output', async () => {
+    const content = `def hello():
+    return "world"
+`;
+    await createTempFile(content);
+
+    const diffStr = `From 1234567890abcdef1234567890abcdef12345678 Mon Sep 17 00:00:00 2001
+From: A. Developer <dev@example.com>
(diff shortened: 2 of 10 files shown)

### CODE REVIEW

File: packages/core/src/utils/patcher.test.ts [hunk 1]
```diff
@@ -0,0 +1,438 @@
...
+  /**
+   * Applies standard unified diff format patches. It ignores line numbers and uses
+   * content-based matching for high accuracy.
+   *
+   * This is the main entry point for the patcher library.
+   */
+  async function diffPatch(
+    patchesStr: string,
+    workingDirectory: string = process.cwd(),
+  ): Promise<string> {
```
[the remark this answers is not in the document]
yongbingchen-a11y: Screenshot 2025-09-02 at 10 51 48 PM
Sample output from this new 'patch' tool -- it failed to apply one hunk of the patch -- this is intentionally set to fail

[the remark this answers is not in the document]
yongbingchen-a11y: Screenshot 2025-09-02 at 11 43 03 PM
The patch tool reported the failed hunk to Gemini, so Gemini called the 'replace' tool to applied the failed hunk.
-- the interaction shows the ideal workflow, cooperative with this 'patch' tool and the existing 'replace' tool -- but this tool is very reliable -- you can test it offline with Gemini generated standard unified patch, just ask Gemini from the web UI.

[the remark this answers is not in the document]
yongbingchen-a11y: Unit test:
```
yongbingchen@dev-yongbing:~/tmp/gemini-cli$ npm test -w @google/gemini-cli-core -- src/utils/patcher.test.ts

> @google/gemini-cli-core@0.2.2 test
> vitest run src/utils/patcher.test.ts


 RUN  v3.2.4 /usr/local/google/home/yongbingchen/tmp/gemini-cli/packages/core
      Coverage enabled with v8

 ✓ src/utils/patcher.test.ts (15 tests) 53ms
   ✓ TestPatchBehaviors > should completely ignore line numbers 11ms
   ✓ TestPatchBehaviors > should process hunks as atomic units 3ms
   ✓ TestPatchBehaviors > should create a new file 2ms
   ✓ TestPatchBehaviors > should delete a file 7ms
   ✓ TestPatchBehaviors > should handle multiple files being processed 5ms
   ✓ TestPatchBehaviors > should handle graceful failure for unmatchable patches 3ms
   ✓ TestPatchBehaviors > should preserve file structure on partial error 3ms
   ✓ TestPatchBehaviors > should reject a patch trying to access outside the working directory 1ms
   ✓ TestPatchBehaviors > should show failed hunks for manual application 2ms
   ✓ TestPatchBehaviors > should skip a hunk if it appears to be already applied 4ms
   ✓ TestPatchBehaviors > should apply multiple hunks in reverse order 3ms
   ✓ TestPatchBehaviors > should handle pure insertions with context 3ms
   ✓ TestPatchBehaviors > should tolerate whitespace and formatting variations 3ms
   ✓ TestPatchBehaviors > should be compatible with git format-patch output 2ms
   ✓ TestPatchBehaviors > should handle empty and no-op patches gracefully 2ms

 Test Files  1 passed (1)
      Tests  15 passed (15)
   Start at  18:41:14
   Duration  1.56s (transform 88ms, setup 29ms, collect 73ms, tests 53ms, environment 0ms, prepare 115ms)
   
   yongbingchen@dev-yongbing:~/tmp/gemini-cli$ npm test -w @google/gemini-cli-core -- src/tools/patch.test.ts

> @google/gemini-cli-core@0.2.2 test
> vitest run src/tools/patch.test.ts


 RUN  v3.2.4 /usr/local/google/home/yongbingchen/tmp/gemini-cli/packages/core
      Coverage enabled with v8

 ✓ src/tools/patch.test.ts (6 tests) 78ms
   ✓ PatchTool Direct Invocation Test > should apply a simple patch to a file 50ms
   ✓ PatchTool Direct Invocation Test > should create a new file 4ms
   ✓ PatchTool Direct Invocation Test > should delete a file 5ms
   ✓ PatchTool Direct Invocation Test > should handle a multi-file patch successfully 7ms
   ✓ PatchTool Direct Invocation Test > should report partial success when one file fails to patch 7ms
   ✓ PatchTool Direct Invocation Test > should return an error when all hunks for all files fail 5ms

 Test Files  1 passed (1)
      Tests  6 passed (6)
   Start at  18:47:11
   Duration  4.29s (transform 1.09s, setup 42ms, collect 2.74s, tests 78ms, environment 0ms, prepare 131ms)
   ```

**Comment by silviojr** [comment]
I want to try to run tests on this to see how it compares to our current solution.

File: packages/core/src/utils/patcher.test.ts (hunk 1 above)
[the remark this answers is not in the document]
Comment (yongbingchen-a11y) [reply packages/core/src/utils/patcher.test.ts]: This has been addressed in patch set 2: https://github.com/google-gemini/gemini-cli/pull/7642/commits/539fc668d2137f288dd681c3fdb013e902cb4a22#diff-22ebadec74fed23a4ccee15f896319256c807fd64dced9197427c1a9fea85351R55

**Comment by silviojr** [comment]
Thank you for your contribution!  I think that using the patch strategy is very promising and also used by other agents. 

I ran some tests to evaluate its performance. 

**Test setup:** 
**disabled:** **write_file**, **replace** and **run_shell_command** to ensure that we are using the candidate patch tool to edit files. 

**Evaluation harness:** I ran our evaluation harness to solve the 300 tasks of SWEBench lite. The **patch** tool failed 155 out of 490 total calls.  Compared to the replace tool that has an average of 20% in the same evaluation. 

I think that at this time, adding a new tool that is also an editing tool with an error rate that is higher than the existing tool may increase the chances of users seeing errors. 

In previous tests, we found out that self-healing loops inside the tool can reduce its error rate, so that may be something to think about adding to the tool.

**Comment by yongbingchen-a11y** [comment]
> Thank you for your contribution! I think that using the patch strategy is very promising and also used by other agents.
> 
> I ran some tests to evaluate its performance.
> 
> **Test setup:** **disabled:** **write_file**, **replace** and **run_shell_command** to ensure that we are using the candidate patch tool to edit files.
> 
> **Evaluation harness:** I ran our evaluation harness to solve the 300 tasks of SWEBench lite. The **patch** tool failed 155 out of 490 total calls. Compared to the replace tool that has an average of 20% in the same evaluation.
> 
> I think that at this time, adding a new tool that is also an editing tool with an error rate that is higher than the existing tool may increase the chances of users seeing errors.
> 
> In previous tests, we found out that self-healing loops inside the tool can reduce its error rate, so that may be something to think about adding to the tool.

Hi, Silvio,

Thank you so much for the test.

However, as stated in the commit message, this tool is never designed to be used alone -- LLM generated 'standard unified diff mode patch' is not 100% reliable, as you can see from your test result -- in that case, the 'patch' tool fails correctly, and pass the correct error message back to LLM -- in that case, LLM is supposed to fix it by improve the patch, or calling the 'replace' tool as last remediate, or calling writeFile as last resort -- This is its 'self-healing' process.

So, you disable 'replace' and claim the tool's failing rate is high is **NOT a valid test setup**, IMHO.

Also, the failure cases is not 155, the LLM will mostly generate multiple incorrect patch on some non-exist code base, if I simply remove the duplicated failed (failed patch on same file), the number of failed case is **52**:
```
$ grep '^\-\-\- a\/' tswe_benchmark_test_record.txt | tail -n 5
--- a/sympy/geometry/point.py
--- a/sympy/geometry/point.py
--- a/sympy/printing/pretty/pretty.py
--- a/sympy/printing/pretty/pretty.py
--- a/sympy/printing/pretty/pretty.py
$ grep '^\-\-\- a\/' swe_benchmark_test_record.txt | sort -u | wc -l
52
```

Thus the success rate is actually 1 - 52/(490-155+52) = 87% 

Thanks!
Yongbing

Foot note:
The failed patch from your test running log:
```
--- a/astropy/nddata/mixins/ndarithmetic.py
+++ b/astropy/nddata/mixins/ndarithmetic.py
@@ -520,38 +520,35 @@
         """
         # Just return what handle_meta does with both of the metas.
         return handle_meta(self.meta, operand.meta, **kwds)

-    def _arithmetic_mask(self, operation, operand, handle_mask, axis=None, **kwds):
+    def _arithmetic_mask(self, operation, operand, handle_mask, axis=None, **kwds):
         """
         Calculate the resulting mask.

-        This is implemented as the piecewise ``or`` operation if both have a
-        mask.
-
         Parameters
         ----------
         operation : callable
             see :meth:`NDArithmeticMixin._arithmetic` parameter description.
             By default, the ``operation`` will be ignored.

         operand : `NDData`-like instance
             The second operand wrapped in an instance of the same class as
             self.

         handle_mask : callable
             see :meth:`NDArithmeticMixin.add`

         kwds :
             Additional parameters given to ``handle_mask``.

         Returns
         -------
         result_mask : any type
             If only one mask was present this mask is returned.
             If neither had a mask ``None`` is returned. Otherwise
             ``handle_mask`` must create (and copy) the returned mask.
         """
-        # If only one mask is present we need not bother about any type checks
-        if (
-            self.mask is None and operand is not None and operand.mask is None
-        ) or handle_mask is None:
+        if handle_mask is None:
             return None
-        elif self.mask is None and operand is not None:
-            # Make a copy so there is no reference in the result.
-            return deepcopy(operand.mask)
-        elif operand is None:
+
+        operand_mask = getattr(operand, "mask", None)
+
+        if self.mask is None and operand_mask is None:
+            return None
+
+        if self.mask is None:
+            return deepcopy(operand_mask)
+
+        if operand_mask is None:
             return deepcopy(self.mask)
-        else:
-            # Now lets calculate the resulting mask (operation enforces copy)
-            return handle_mask(self.mask, operand.mask, **kwds)
+
+        # Both masks are present, call the handler
+        return handle_mask(self.mask, operand_mask, **kwds)

     def _arithmetic_wcs(self, operation, operand, compare_wcs, **kwds):
```

Compare to the original SWE bench test case for this file:
https://github.com/astropy/astropy/blob/main/astropy/nddata/mixins/ndarithmetic.py#L620 
```
        # Just return what handle_meta does with both of the metas.
        return handle_meta(self.meta, operand.meta, **kwds)

    @sharedmethod
    @format_doc(_arit_doc, name="addition", op="+")
    def add(self, operand, operand2=None, **kwargs):
        return self._prepare_then_do_arithmetic(np.add, operand, operand2, **kwargs)

```

-- This is clearly a failed patch from LLM, not the tool's fault, and it correctly reported the error to LLM.

File: packages/core/src/utils/patcher.ts [hunk 2]
```diff
@@ -0,0 +1,613 @@
...
+}
+
+/**
+ * Ensures the resolved file path is within the secure working directory
+ * to prevent path traversal attacks.
+ * @param filepath The relative path to the file.
+ * @param workingDirectory The absolute path of the secure base directory.
+ * @return The resolved, safe absolute path.
+ */
+function getSafePath(filepath: string, workingDirectory: string): string {
```
cornmander: Can you use https://github.com/google-gemini/gemini-cli/blob/45d494a8d86cd2d47ac34abb7b15d25124b44e5a/packages/core/src/utils/workspaceContext.ts#L132?

packages/core/src/services/fileSystemService.ts
```diff
@@ -25,6 +25,21 @@ export interface FileSystemService {
    * @param content - The content to write
    */
   writeTextFile(filePath: string, content: string): Promise<void>;
+
```
cornmander, line 28: I'm surprised that we don't already have functionality to delete files or create directories in the edit and replace tools. Were we relying on shell tool execution for that?

yongbingchen-a11y: I feel this is natural in existing tool configure: 'replace' tool is just doing string replace -- there's no semantic meaning of 'replace' action to create or delete a file.

Only when you think in diff mode patch way, creating and deleting a file sounds natural to the same code editing tool.

packages/core/src/utils/patcher.ts
yongbingchen-a11y: Done

yongbingchen-a11y:
Screenshot 2025-09-04 at 1 06 29 PM
The reasoning passed to Gemini for choosing this patch tool (as the effect of the tool prompt).

silviojr:
I think there's a misunderstanding of how we are measuring tool errors here. This means we have to do a better job of sharing with the community how we are measuring the quality of the Gemini CLI.

**Errors**

We are not counting as failures just the errors that happen when the tool fails internally because of a bug. We are also counting errors that happen because of:

**1) Tool internal bug/system failure:** like using the wrong function to open a file, network error, etc.
**2) Wrong arguments:** As you pointed out, some errors are due to Gemini's fault. Nevertheless, they are surfaced to the user as errors. This happens when the `replace` tool is asked to replace a string that does not exist, for example. It's the same as when the `patch` tool fails to apply a patch because the patch itself is wrong—not because of the tool itself, but because of Gemini. Those are still errors. They may happen because the tool doesn't provide a good description of itself or simply because Gemini is not able to understand how to generate the right content and provide the right arguments.
**3) User cancellation:** For tools that request the user to confirm their actions, if the user cancels the action, we count this as an error.

**Duplicated errors**

If an error happens, it's desirable for Gemini to observe that error and take an action based on it. This might mean, for example, calling the tool with different arguments. We can't guarantee what Gemini will do, but we can, for example, add hints in the error message to prevent Gemini from making the same mistake. If Gemini still makes the same mistake, again—that's counted as a new error. This may happen for various reasons. There may not be enough context (or maybe there is too much context), or the tool's instructions are not clear enough on how to recover or learn from the error. We want to prevent these duplicated errors from happening. So that's why we still track them so we can improve the tools.

**Not using the replace tool in the tests**

That was intentional because we want to test the performance of the patch tool itself first. If we include the `replace` tool in our settings, Gemini may just prefer to use it  and we won't observe much of what the patch tool is capable of.

**Replace tool as a callback and self-correcting**

The current error rate of this patch tool is high (>30%), so have another unreliable tool as a fallback will add more points of failure, increase maintenance overhead, and complexity. And again, there's no guarantee that Gemini will call them in this sequence.

As I mentioned, self-correction or self-healing is an option here. But it should be done with inside the tool. Returning errors to the main loop means more pollution in the context, it increases the size of the user's utilized context. And as we know, the bigger the size of the context used, the higher the chances of hallucinations.

Yes, the proposed `smart-edit` does a self-correction of its parameters. But that is never visible to the outside loop/context, and it uses a cheaper model (Gemini 2.5 flash) to do that. When an error goes back to the main loop/context, if the user is utilizing Gemini 2.5 Pro, it will use their tokens to repair something that was broken due to bad tool usage, for example. We are working hard to help users maximize the usage of those more expensive tokens as much as possible.

**What's next?**
The edit/replace tool is the tool that has the highest error rate among all of our tools. We want to reduce that because it causes a terrible user experience. For now, we made the choice of testing the simplest approach (search/replace) and doing a self-correction based on testing multiple candidate imlementations and observing which ones had the lowest error rates when Gemini CLI had the task of solving programming issues in languages such as Python, Java, JavaScript, etc.

I really miss not being able to do multiple edits in a single tool call—and we also tried to give a unidiff-based tool to Gemini CLI. But the error rate was higher than with search/replace. We may have to make improvements and spend more time improving the unidiff/patch tool - and maybe a new implementation will be a solution that will **replace** the `replace` tool (ahah). We  should continue exploring our options!

yongbingchen-a11y:
> I think there's a misunderstanding of how we are measuring tool errors here. This means we have to do a better job of sharing with the community how we are measuring the quality of the Gemini CLI.
> 
> **Errors**
> 
> We are not counting as failures just the errors that happen when the tool fails internally because of a bug. We are also counting errors that happen because of:
> 
> **1) Tool internal bug/system failure:** like using the wrong function to open a file, network error, etc. **2) Wrong arguments:** As you pointed out, some errors are due to Gemini's fault. Nevertheless, they are surfaced to the user as errors. This happens when the `replace` tool is asked to replace a string that does not exist, for example. It's the same as when the `patch` tool fails to apply a patch because the patch itself is wrong—not because of the tool itself, but because of Gemini. Those are still errors. They may happen because the tool doesn't provide a good description of itself or simply because Gemini is not able to understand how to generate the right content and provide the right arguments. **3) User cancellation:** For tools that request the user to confirm their actions, if the user cancels the action, we count this as an error.
> 
> **Duplicated errors**
> 
> If an error happens, it's desirable for Gemini to observe that error and take an action based on it. This might mean, for example, calling the tool with different arguments. We can't guarantee what Gemini will do, but we can, for example, add hints in the error message to prevent Gemini from making the same mistake. If Gemini still makes the same mistake, again—that's counted as a new error. This may happen for various reasons. There may not be enough context (or maybe there is too much context), or the tool's instructions are not clear enough on how to recover or learn from the error. We want to prevent these duplicated errors from happening. So that's why we still track them so we can improve the tools.
> 
> **Not using the replace tool in the tests**
> 
> That was intentional because we want to test the performance of the patch tool itself first. If we include the `replace` tool in our settings, Gemini may just prefer to use it and we won't observe much of what the patch tool is capable of.
> 
> **Replace tool as a callback and self-correcting**
> 
> The current error rate of this patch tool is high (>30%), so have another unreliable tool as a fallback will add more points of failure, increase maintenance overhead, and complexity. And again, there's no guarantee that Gemini will call them in this sequence.
> 
> As I mentioned, self-correction or self-healing is an option here. But it should be done with inside the tool. Returning errors to the main loop means more pollution in the context, it increases the size of the user's utilized context. And as we know, the bigger the size of the context used, the higher the chances of hallucinations.
> 
> Yes, the proposed `smart-edit` does a self-correction of its parameters. But that is never visible to the outside loop/context, and it uses a cheaper model (Gemini 2.5 flash) to do that. When an error goes back to the main loop/context, if the user is utilizing Gemini 2.5 Pro, it will use their tokens to repair something that was broken due to bad tool usage, for example. We are working hard to help users maximize the usage of those more expensive tokens as much as possible.
> 
> **What's next?** The edit/replace tool is the tool that has the highest error rate among all of our tools. We want to reduce that because it causes a terrible user experience. For now, we made the choice of testing the simplest approach (search/replace) and doing a self-correction based on testing multiple candidate imlementations and observing which ones had the lowest error rates when Gemini CLI had the task of solving programming issues in languages such as Python, Java, JavaScript, etc.
> 
> I really miss not being able to do multiple edits in a single tool call—and we also tried to give a unidiff-based tool to Gemini CLI. But the error rate was higher than with search/replace. We may have to make improvements and spend more time improving the unidiff/patch tool - and maybe a new implementation will be a solution that will **replace** the `replace` tool (ahah). We should continue exploring our options!

Hi, Silvio,

Thank you for the detailed explain, I need to be able to run the test myself to see what's wrong in those settings -- as from the test log you shared, I can not even see the code base that why those patch failed -- this is an unfair test to reject a new algorithm, by not allowing it even to debug, so **I will apply for run the same test myself internally**.

The patch tool choose to return error instead of self-healing, is because it assumes there's a 'replace' tool exists in the agent -- if the goal is to 'replace' tool totally, then I can add the same self-healing logic inside this tool call -- it's just a sub-task to a new separate session, and as you can see from the logic, 'patch' tools is a superset of 'replace' tool -- **please comment**.

Thanks!
Yongbing

yongbingchen-a11y:
> **replace** the `replace` tool



> I think there's a misunderstanding of how we are measuring tool errors here. This means we have to do a better job of sharing with the community how we are measuring the quality of the Gemini CLI.
> 
> **Errors**
> 
> We are not counting as failures just the errors that happen when the tool fails internally because of a bug. We are also counting errors that happen because of:
> 
> **1) Tool internal bug/system failure:** like using the wrong function to open a file, network error, etc. **2) Wrong arguments:** As you pointed out, some errors are due to Gemini's fault. Nevertheless, they are surfaced to the user as errors. This happens when the `replace` tool is asked to replace a string that does not exist, for example. It's the same as when the `patch` tool fails to apply a patch because the patch itself is wrong—not because of the tool itself, but because of Gemini. Those are still errors. They may happen because the tool doesn't provide a good description of itself or simply because Gemini is not able to understand how to generate the right content and provide the right arguments. **3) User cancellation:** For tools that request the user to confirm their actions, if the user cancels the action, we count this as an error.
> 
> **Duplicated errors**
> 
> If an error happens, it's desirable for Gemini to observe that error and take an action based on it. This might mean, for example, calling the tool with different arguments. We can't guarantee what Gemini will do, but we can, for example, add hints in the error message to prevent Gemini from making the same mistake. If Gemini still makes the same mistake, again—that's counted as a new error. This may happen for various reasons. There may not be enough context (or maybe there is too much context), or the tool's instructions are not clear enough on how to recover or learn from the error. We want to prevent these duplicated errors from happening. So that's why we still track them so we can improve the tools.
> 
> **Not using the replace tool in the tests**
> 
> That was intentional because we want to test the performance of the patch tool itself first. If we include the `replace` tool in our settings, Gemini may just prefer to use it and we won't observe much of what the patch tool is capable of.
> 
> **Replace tool as a callback and self-correcting**
> 
> The current error rate of this patch tool is high (>30%), so have another unreliable tool as a fallback will add more points of failure, increase maintenance overhead, and complexity. And again, there's no guarantee that Gemini will call them in this sequence.
> 
> As I mentioned, self-correction or self-healing is an option here. But it should be done with inside the tool. Returning errors to the main loop means more pollution in the context, it increases the size of the user's utilized context. And as we know, the bigger the size of the context used, the higher the chances of hallucinations.
> 
> Yes, the proposed `smart-edit` does a self-correction of its parameters. But that is never visible to the outside loop/context, and it uses a cheaper model (Gemini 2.5 flash) to do that. When an error goes back to the main loop/context, if the user is utilizing Gemini 2.5 Pro, it will use their tokens to repair something that was broken due to bad tool usage, for example. We are working hard to help users maximize the usage of those more expensive tokens as much as possible.
> 
> **What's next?** The edit/replace tool is the tool that has the highest error rate among all of our tools. We want to reduce that because it causes a terrible user experience. For now, we made the choice of testing the simplest approach (search/replace) and doing a self-correction based on testing multiple candidate imlementations and observing which ones had the lowest error rates when Gemini CLI had the task of solving programming issues in languages such as Python, Java, JavaScript, etc.
> 
> I really miss not being able to do multiple edits in a single tool call—and we also tried to give a unidiff-based tool to Gemini CLI. But the error rate was higher than with search/replace. We may have to make improvements and spend more time improving the unidiff/patch tool - and maybe a new implementation will be a solution that will **replace** the `replace` tool (ahah). We should continue exploring our options!

Hi, Silvio,

Thank you so much for the detailed explain.

I will setup the SWE benchmark test myself to debug the failure.

I will add the self-healing logic (by calling the 'smart edit' logic you developed, as the error message in the 'patch' tool is already a 'old_string', 'new_string', so it should be a very straight forward adapt to your solution), then redo the test -- I believe in this way, we will have both the correctness rate from your 'smart edit' and the efficiency and align with LLM mindset of 'unified diff patch' --  so it will 'be a solution that will **replace** the replace tool' -- ^-^  -- I tested in Gemini CLI, Gemini shows confusion to choose between 'patch' and 'replace' tool, so the correct solution should be a total replacement.

Sincerely,
Yongbing

packages/core/src/services/fileSystemService.ts
timothyfroehlich: Speaking for myself, I never want to allow a file deletion without explicit consent. It may be more of a gut feeling than anything, but when the LLM starts deleting files and it's not a cleanup/migration task, usually it's gone off the rails and is doing something that I want to intercede with.

yongbingchen-a11y: The 'patch' tool still present preview and ask for confirmation, like the 'replace' tool will do.

Allow the patch tool to delete file is for the interface completeness -- as how the 'unified diff patch' is defined, some coding task can not be fully agentic if you don't allow delete files.

Anyway, we can have a separate confirmation step for file deletion, but current 'confirmation' design does not allow multiple steps confirm for single tool call -- we may need to refine this before we can enable this additional confirmation.

yongbingchen-a11y:
Screenshot 2025-09-05 at 1 53 49 PM
Sample session where the self-healing sub-task fixed a diff mode patch during the tool call -- so the main session's context is saved -- the main session's response message contains the remediation message, so the main session LLM has a consistent view of what happened (instead of silently out of sync). 

The self-healing does not do any semantic fixing -- it only try to find out if the patch itself can be fixed by finding correct context -- let the sub-task to change what the main session LLM intend to do is a source of contaminate the main LLM session.

yongbingchen-a11y:
Run the SWE benchmark test with https://github.com/yongbingchen-a11y/gemini-cli/commit/fe46bfe583c7fe34706fa30c0e49eaa6d90903a6

The telemetry summary for the 'patch' tool is:
```
## 📉 Most Common Errors
------------------------
Total Failures: 20 out of 627 total calls.

- **Count: 20**
  ```
  edit_no_occurrence_found
  ```
 ```
  
The error rate is 3%, it's half compared to the existing 'smart replace' tool (https://github.com/google-gemini/gemini-cli/pull/6823)

**Comment by yongbingchen-a11y** [comment]
> Run the SWE benchmark test with [yongbingchen-a11y@fe46bfe](https://github.com/yongbingchen-a11y/gemini-cli/commit/fe46bfe583c7fe34706fa30c0e49eaa6d90903a6)
> 
> The telemetry summary for the 'patch' tool is:
> 
> ```
> ## 📉 Most Common Errors
> ------------------------
> Total Failures: 20 out of 627 total calls.
> 
> - **Count: 20**
> ```
> 
> edit_no_occurrence_found
> 
> The error rate is 3%, it's half compared to the existing 'smart replace' tool (#6823)

```
================================================================================
🔬 Tool Analysis Summary for: `patch`
================================================================================

## 📉 Most Common Errors
------------------------
Total Failures: 15 out of 613 total calls.

- **Count: 15**
  ```
  edit_no_occurrence_found
  ```
  ```
  
  Another run, with even better result: 2% failing rate

yongbingchen-a11y:
> Another run, with even better result: 2% failing rate

```
- **Total Instances Processed**: 500
- **Total Log Files Processed**: 500

## 📊 Combined Statistics
-----------------------
#### 🪙 Token Usage
- **Total Input Tokens**: 135983719
- **Total Output Tokens**: 876035
- **Total Cached Tokens**: 66681126
- **Total Thoughts Tokens**: 5892591
- **Total Tool Tokens**: 18201
- **Grand Total Tokens**: 142770546
```

yongbingchen-a11y:
> - **Total Instances Processed**: 500
> - **Total Log Files Processed**: 500
> 
> ## 📊 Combined Statistics
> -----------------------
> #### 🪙 Token Usage
> - **Total Input Tokens**: 135983719
> - **Total Output Tokens**: 876035
> - **Total Cached Tokens**: 66681126
> - **Total Thoughts Tokens**: 5892591
> - **Total Tool Tokens**: 18201
> - **Grand Total Tokens**: 142770546

The same token consumption statistics from the base version (de5a31c3 origin/main) run with 'replace' tool:
```
- **Total Instances Processed**: 500
- **Total Log Files Processed**: 500

## 📊 Combined Statistics
-----------------------
#### 🪙 Token Usage
- **Total Input Tokens**: 116094441
- **Total Output Tokens**: 881678
- **Total Cached Tokens**: 42243607
- **Total Thoughts Tokens**: 6326732
- **Total Tool Tokens**: 78632
- **Grand Total Tokens**: 123381483
```

The tool tokens decreased from 78632 to 18201 , a 77% drop, as expected (because 'patch' tool is by nature more efficient)

PR #7165, opened 2025-08-26
Add highlights for input /commands and @file/paths

miguelsolorio:
## TLDR

Adds highlighting for /commands and @file/paths

## Dive Deeper

CleanShot 2025-08-01 at 09 44 47@2x

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| -------- | --- | --- | --- |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

PR #7652, opened 2025-09-03
Revert 7165

miguelsolorio:
This reverts commit c29e44848b3e087ba5e807ba4f5542d72356b32b (PR #7165)

This causes multiple new lines to break and unintentional highlights (#7651)


miguelsolorio:
We'll try to address this in a follow up PR

PR #7656, opened 2025-09-03
fix: InputPrompt wrapped lines maintain highlighting, increase responsiveness in narrow cases 

### LINKED ISSUE #7768: Path highlighting breaks on line wrap
### What happened?

Image

introduced in #7165

proposed solution: break paths into segments and highlight segments, the offset will be consistent even as more visual lines are added

entire path should be colored

### Client information

<details>

```console
> /about
# paste output here
```

</details>

### Login information

_No response_

_No response_

### LINKED ISSUE #8206: Input breaks below clamp of 20
### What happened?

The `inputWidth` is clamped to `20` (`AppContainer.tsx`) so the UI breaks in several ways when terminals like iTerm2 are narrowed to widths below `20`.

Notably: 

1) Below the clamped minimum width internal data structures in `text-buffer.ts` become inaccurate because they are calculating on a width that is not represented in the UI.

2) Also, Ink's responsibility of styling and internal use of chalk breaks down

This makes solutions to #7768 difficult if they rely on Ink to properly style `<Text>` nodes

Proposed solution: Reduce minimum width clamp to something more reasonable like 2 while also preventing/addressing all regressions in all reasonable and edge cases

Image:

Image

The cursor should remain visible at the end of the input box, the text should not be cut off, the source code should lend itself to segment level highlighting that relies on Ink to style text in the terminal.

### Client information

<details>

```console
> /about
# paste output here
```

</details>

### Login information

_No response_

_No response_

psinha40898:
## TLDR

Although paths and commands are highlighted in the `InputPrompt`, they currently break on line wrap. This is because tokenizing by `visualLine` does not highlight all wrapped lines.

`visualLines = ["@some/file/path.txt"]` would be colorized properly

`visualLines = ["@some/file/", "path.txt"]` would only have the first entry colorized

Instead of coloring each token (a token is a command, a path, or default text), we calculate the segments of each token in visual lines and color those. 

In order to resolve edge cases, issues of responsiveness with `InputPrompt` **also** had to be addressed.

### 1) Segment-Based Highlighting for Paths and Commands

We exposed `visualToLogicalMaps` to split the tokens into segments so that if they span multiple visual lines they are of a segment type that still gets colored.

Previously:
- `visualLines = ["@some/file/path.txt"]` would be colorized properly
- `visualLines = ["@some/file/", "path.txt"]` would only have the first entry colorized

Now both segments will be colorized.

#### Before:
image

*Note: Coloring breaks on line wrap*

#### After:
image

### 2) More Responsive InputPrompt at Narrow Widths

This PR also increases the responsiveness of the `InputPrompt` because a minimum width of 20 was breaking multiple aspects of the UI when users narrowed the width below that value (possible on iTerm2) such as Ink's responsibility of styling `<Text>` components breaking down due to unexpected wrapping by the terminal. 

This meant that segment-level highlighting broke at extremely narrow widths, so increasing the responsiveness of `InputPrompt` was needed to fix highlighting of wrapped lines.

Now line wrapping is managed as expected by Ink, and the `InputPrompt` is more agreeable at narrow widths.

While this was necessary for segment highlighting, the benefits of this new `InputPrompt` can also be observed in isolation from the highlighting feature (see #8206, see https://github.com/google-gemini/gemini-cli/pull/7759#issuecomment-3262833025).

#### Before:
*Visual lines cut off both horizontally and vertically, cursor misplaced - UI broken*

image

#### After:
*Extreme narrow width:*
image

*Less narrow, still responsive:*
image

The new `InputPrompt` collapses to a minimum of 2 characters per row, and retains cursor positioning as well as proper line visibility at extreme narrow dimensions.

**To prevent regressions in edge cases:**
- The `useEffect` in `text-buffer.ts` that kept track of vertical height is more robust for resizing
- The width passed in to the `DetailedMessagesDisplay` is now calculated for its specific purpose and held in `uiState` - previously the `InputPrompt` `inputWidth`, which is used to calculate the content inside of `InputPrompt` and not the width of a container in the UI, was used

## Reviewer Test Plan

When reviewing this code, please pull and run the code to validate the changes work. Test with various terminal widths, especially narrow ones, and verify that:
**1. Path and command highlighting works correctly across line wraps**
2. The `InputPrompt` remains functional at extreme narrow widths
3. Cursor positioning is maintained during resizing

## Testing Matrix

*Before submitting, please validate your changes on as many of these options as possible*

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked Issues / Bugs

Related to #7165  
Fixes #7768  
Fixes #8206


jacob314 (commented):
There are a few more edge cases to consider. For example, notice that `ver` is not syntax highlighted properly in this.

Image

psinha40898:
> There are a few more edge cases to consider. For example, notice that `ver` is not syntax highlighted properly in this.
> 
>image


It appears that there is an edge case when the terminal width is less than 20 columns, which can be done on terminals like iTerm2. Keeping track of each character to color seems to resolve the issue and is pushed to this PR, but I'm not certain that it is the best solution. Perhaps I'm wrong and this is not the one you've described here? 

There also remains another bug with cursor placement irregularity when terminal width is less than 20 columns. That bug exists with or without this PR and is not related to highlighting. One reason that I mention this is because I have reason to suspect _that_ bug did not exist on _this_ branch when you first reviewed it. (See https://github.com/google-gemini/gemini-cli/pull/7759#issuecomment-3262833025)

packages/cli/src/ui/components/InputPrompt.tsx
```diff
@@ -749,62 +752,45 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
...
                       ? theme.text.accent
                       : undefined;
 
+                  const isCursorChar =
+                    isOnCursorLine && i === cursorVisualColAbsolute;
+                  const display = isCursorChar ? chalk.inverse(ch) : ch;
+
                   renderedLine.push(
-                    <Text key={`token-${tokenIdx}`} color={color}>
+                    <Text key={`ch-${absVisualIdx}-${i}`} color={color}>
```
jacob314: this is generating an excessive number of <Text> objects. we should group text objects together as long as they have the same color. otherwise there are some performance concerns and this output will be very hard to read using React debugging tools due to having each char in a same text block.

psinha40898: Previously this PR had segment level highlighting. It broke on an edge case reported earlier. By my calculation, highlighting was breaking on terminal widths that were unexpectedly narrow for Ink. 

To address this, the App's `inputWidth` was recalculated to a much narrower minimum and the useEffect syncing the height in `text-buffer.ts` was remedied to address an edge case with the new narrower minimum.

The result should be a more responsive `InputPrompt` that is agreeable at more narrow terminal widths including path and command highlighting without regressions.

psinha40898:
I think the segmenting and highlighting code could be simplified and refactored, but I need a sanity check about the issues and edge cases I observed with highlighting across wrapped lines in narrow widths as well as my solutions

miguelsolorio:
@psinha40898 test this out today and it appears to be working well, haven't found any visual issues.

packages/cli/src/ui/AppContainer.tsx
```diff
@@ -227,10 +227,24 @@ export const AppContainer = (props: AppContainerProps) => {
   }, [handleNewMessage, config]);
 
   const widthFraction = 0.9;
-  const inputWidth = Math.max(
-    20,
-    Math.floor(terminalWidth * widthFraction) - 3,
-  );
+  // ──────────────────────────────────────────────────────────────
```
jacob314: the width calculation for the input prompt is duplicated. Please add it as a separate helper method exported from InputPrompt.tsx so we don't duplicate this. Otherwise we will fail to keep it consistent and have confusing bugs.

packages/cli/src/ui/contexts/UIStateContext.tsx
```diff
@@ -66,6 +66,7 @@ export interface UIState {
   buffer: TextBuffer;
   inputWidth: number;
   suggestionsWidth: number;
+  debugWidth: number;
```
jacob314: what does debugWidth mean? this is a confusing prop name. would be ideal to delete.

jacob314 (commented):
Overall looks great. A few minor comments then approved.

packages/cli/src/ui/contexts/UIStateContext.tsx
psinha40898: The DetailedMessagesDisplay now receives the InputPrompt container width for its width prop.

On main the inputWidth from uiState is passed as the width prop for DetailedMessagesDisplay. 

Since inputWidth represents the width of the content inside the InputPrompt, it did not make sense to me to pass it as a prop that represents the container width of a sibling component. Hope I got that right

packages/cli/src/ui/AppContainer.tsx
psinha40898: The 3 width calculations relevant to the InputPrompt (the input container width, the input content width, and the input suggestions width) are now calculated with a helper that is exported from `InputPrompt.tsx`.

psinha40898:
In addition to attempting to resolve the issues, I cleaned up the code in `InputPrompt.tsx` so the diffs against main make more sense. I'm happy to continue making changes if I misunderstood or forgot something. Thank you for the reviews and guidance!

jacob314:
This is a great PR that addresses a tricky UI bug with wrapped-line highlighting. The code is well-structured, and the detailed PR description is much appreciated.

Here are a few points for consideration:

*   **Testing for `buildSegmentsForVisualSlice`**: The new `buildSegmentsForVisualSlice` function in `packages/cli/src/ui/utils/highlight.ts` is a critical piece of the new highlighting logic. It would be great to add a separate suite of unit tests for this function to cover various scenarios, including:
    *   Tokens that are fully contained within the slice.
    *   Tokens that are partially contained at the beginning or end of the slice.
    *   Tokens that completely contain the slice.
    *   Multiple tokens within the slice.
    *   Edge cases with empty inputs or zero-length slices.

Overall, this is a solid contribution that significantly improves the user experience of the input prompt. Thank you for the thorough work!

jacob314 (approved):
[lgtm](https://i.imgur.com/JgNRalI.png)

psinha40898:
Unfortunately a failing test on windows node 20 and a failing e2e. Will try to resolve asap. My apologies

edit: looks like everything is passing now..ty

psinha40898:
I'm not entirely sure what's going on with a windows (node 24) test failing in CI. It seemed to have passed all checks earlier, and is passing locally for me on windows with node 24. I will try to look into it.

Amnat-auyana1998 (approved):

PR #7660, opened 2025-09-03
feat: use largest windows runner for ci

mattKorwel:
This PR updates the ci.yml workflow to use the largest available windows runner.


allenhutchison (approved):

mattKorwel:
For future us: i'm not sure how billing and actions runners work in this scenario. We will need to chase this down and ensure we understand it before leaving this on long term.

Issue #7324, opened 2025-08-28
mcp tool calls broken due to enforcing a strict 60s timeout


vnagarnaik:
### What happened?

something changed between the 0.1.22 and 0.2.1 releases, gemini-cli seems to be enforcing a 60s timeout on mcp tool calls. essentially the timeout field in mcp server config is being ignored. this seems at a very low level in mcp that can't be worked around in the tool implementation.

use case: i created an internal mcp server for our build/test integration which takes 2-3 minutes to complete for each call. 

this enforcement is completely breaking the workflow, since gemini can't get the tool output for context and continue with its fixing loop.

### What did you expect to happen?

mcp client should wait for the tool call to complete, and honor the mcp server timeout field. the mcp server sends notifications/progress periodically, which should provide enough context for the mcp client to keep waiting for the server output.

### Client information

<details>
<summary>Client Information</summary>

Run `gemini` to enter the interactive CLI, then run the `/about` command.

```console
> /about
│ About Gemini CLI                                                                                          │
│                                                                                                           │
│ CLI Version                          0.2.2                                                                │
│ Git Commit                           32204dca                                                             │
│ Model                                gemini-2.5-pro                                                       │
│ Sandbox                              no sandbox                                                           │
│ OS                                   win32                                                                │
│ Auth Method                          gemini-api-key                                                       │
│ GCP Project                          <redacted>                                                           │
```

</details>

jacob314:
Fyi @jakemac53

jakemac53:
Related to https://github.com/google-gemini/gemini-cli/issues/6590 - I think some sort of timeout is likely required here - you don't just want it stalling forever.

But, if we supported progress notifications we could reset the timeout each time we receive a progress notification?

We could also make the timeout configurable, where a `0` value means no timeout.

JunYang-tes:
This timeout is a bit annoying. One time I left my computer, and gemini-cli requested me to accept or reject through the IDE. By the time I got back, it had already timed out.

jakemac53:
One other option we could consider here is sending a `ping` request to the server when we reach the timeout - just to ensure the server itself is at least active. We can't know that its still attempting to perform the specific tool call though.

> One time I left my computer, and gemini-cli requested me to accept or reject through the IDE.

That sounds like a slightly different but related issue - file edits don't go through MCP

jakemac53:
[...]

PR #7661, opened 2025-09-03
Simplify MCP server timeout configuration

jakemac53:
## TLDR

Use the option provided by the genai SDK `mcpToTool` function to configure timeouts when creating the gemini tool from the MCP client.

## Dive Deeper

Previously, we were overwriting the `callTool` method on the McpClient itself to achieve this and had a TODO to remove that once we were provided a better option. I am not sure if this API is actually new, but it does allow us to achieve the same effect while simplifying the code significantly and making it less brittle.

I also removed the `timeout` option from the tool specific classes as these are no longer needed/used, it is configured at a higher level than that. We do not and never have supported per-tool timeout configurations for MCP servers.

## Reviewer Test Plan

Set the request timeout for some MCP server to a low value and call a function that takes longer than that.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Related to https://github.com/google-gemini/gemini-cli/issues/7324


cornmander (approved):

machinaai:
This change didn’t resolve the MCP “strict” 60-second timeout, though. Seems the root issue is in the /node_modules/@modelcontextprotocol/sdk/dist/esm/client/streamableHttp.js  file. trying a monkey-patch but no successful results yet.

jakemac53:
I only tested with a stdio MCP server - although I don't see anything in the transport itself that related to this. The timeouts are handled here https://github.com/modelcontextprotocol/typescript-sdk/blob/main/src/shared/protocol.ts#L624 and this should work fine with what this PR implements.

machinaai:
Seems to me that:

`protocol.js`: Contains an application-level timeout for the request-response cycle, but this does not time out the underlying network request.
`sse.js` & `streamableHttp.js`: Make fetch calls without any timeout mechanism. The AbortSignal they use is only for manually closing the connection, not for enforcing a timeout.

PR #7666, opened 2025-09-03
Fix(core): Log exact model version from API response

SandyTao520:
## TLDR

It was logging config.getModel() which is always gemini-2.5-pro.

## TLDR

This pull request updates the API response logging to use the precise model version returned from the backend (`response.modelVersion`) rather than the model specified in the local configuration. This enhances the accuracy of our telemetry data.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ✅  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

### DIFF
--- a/packages/core/src/core/loggingContentGenerator.ts
+++ b/packages/core/src/core/loggingContentGenerator.ts
@@ -60,6 +60,7 @@ export class LoggingContentGenerator implements ContentGenerator {
 
   private _logApiResponse(
     durationMs: number,
+    model: string,
     prompt_id: string,
     usageMetadata?: GenerateContentResponseUsageMetadata,
     responseText?: string,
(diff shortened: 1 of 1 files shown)


NTaylorMullen (approved):

anj-s (approved):

packages/core/src/core/loggingContentGenerator.ts
```diff
@@ -171,6 +173,7 @@ export class LoggingContentGenerator implements ContentGenerator {
       this._logApiResponse(
         durationMs,
         userPromptId,
+        responses[0]?.modelVersion || '',
```
anj-s, line 176: it should never be empty right?

PR #7678, opened 2025-09-03
fix(settings): Add missing v1 settings to migration map

galdawave:
## TLDR

Adds missing setting to migration map

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

### DIFF
--- a/packages/cli/src/config/settingsSchema.ts
+++ b/packages/cli/src/config/settingsSchema.ts
@@ -574,6 +574,16 @@ export const SETTINGS_SCHEMA = {
           'Use node-pty for shell command execution. Fallback to child_process still applies.',
         showInDialog: true,
       },
+      autoAccept: {
+        type: 'boolean',
+        label: 'Auto Accept',
+        category: 'Tools',
+        requiresRestart: false,
+        default: false,
+        description:
+          'Automatically accept and execute tool calls that are considered safe (e.g., read-only operations).',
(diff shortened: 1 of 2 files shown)


jacob314 (approved):
lgtm

packages/cli/src/config/settingsSchema.ts
```diff
@@ -574,6 +574,16 @@ export const SETTINGS_SCHEMA = {
           'Use node-pty for shell command execution. Fallback to child_process still applies.',
         showInDialog: true,
       },
+      autoAccept: {
+        type: 'boolean',
+        label: 'Auto Accept',
+        category: 'Tools',
+        requiresRestart: false,
+        default: false,
```
NTaylorMullen, line 582: Read operations are typically auto-accepted. Maybe I missed this setting originally: should this be true or is false correct?

NTaylorMullen (approved):

packages/cli/src/config/settingsSchema.ts
galdawave: the old schema had this set to false

PR #7680, opened 2025-09-03
improve performance of shell commands with lots of output

### LINKED ISSUE #5411: performance: pegging 100% CPU qnd grinding to an halt
### What happened?

When running certain tasks with lots of stdout/stderr output, it looks like gemini-cli does a lot of small reallocation, killing performance.
In my case it is building an Android Kotlin app using `./gradlew buildDebug --info`. This generates quite a lot of output and gemini-cli consumes the output extremelly slowly.
A perf top reveals it does a lot of malloc/free.
I am now using jemalloc with LD_PRELOAD, which helped a bit, but it still ends up being very slow.
Here is the perf record over 10s while being very slow:
```
Samples: 39K of event 'cycles:P', Event count (approx.): 21597291866
Overhead  Command  Shared Object         Symbol
   2.47%  node     libnode.so.115        [.] v8::internal::JSObject::MigrateToMap(v8::internal::Isolate*, v8::int
   2.17%  node     libnode.so.115        [.] v8::internal::GlobalHandles::Create(v8::internal::Object)
   2.10%  node     libnode.so.115        [.] v8::Isolate::AdjustAmountOfExternalAllocatedMemory(long)
   1.62%  node     libicuuc.so.76.1      [.] icu_76::RuleBasedBreakIterator::~RuleBasedBreakIterator()
   1.47%  node     libnode.so.115        [.] v8::internal::Map::TransitionToDataProperty(v8::internal::Isolate*,
   1.45%  node     libc.so.6             [.] pthread_mutex_lock
   1.34%  node     libnode.so.115        [.] v8::internal::LookupIterator::PrepareTransitionToDataProperty(v8::in
   1.28%  node     libnode.so.115        [.] v8::internal::Object::AddDataProperty(v8::internal::LookupIterator*,
   1.25%  node     libicuuc.so.76.1      [.] utext_setup_76
   1.24%  node     libnode.so.115        [.] v8::internal::JSReceiver::CreateDataProperty(v8::internal::Isolate*,
   1.22%  node     libnode.so.115        [.] v8::internal::LookupIterator::WriteDataValue(v8::internal::Handle<v8
   1.13%  node     libicuuc.so.76.1      [.] icu_76::RuleBasedBreakIterator::handleNext()
   1.11%  node     libjemalloc.so.2      [.] malloc
   1.07%  node     libnode.so.115        [.] v8::internal::Factory::NewStringFromTwoByte(unsigned short const*, i
   1.06%  node     libnode.so.115        [.] v8::internal::LookupIterator::State v8::internal::LookupIterator::Lo
   1.02%  node     libnode.so.115        [.] v8::internal::LookupIterator::ApplyTransitionToDataProperty(v8::inte
   1.02%  node     libjemalloc.so.2      [.] free
   0.99%  node     libicuuc.so.76.1      [.] icu_76::RuleBasedBreakIterator::RuleBasedBreakIterator(UErrorCode*)
   0.98%  node     libc.so.6             [.] 0x000000000009777c
   0.93%  node     libicuuc.so.76.1      [.] icu_76::UVector32::~UVector32()
   0.92%  node     libnode.so.115        [.] v8::internal::GlobalHandles::Destroy(unsigned long*)
   0.90%  node     libnode.so.115        [.] std::_Sp_counted_ptr<icu_76::BreakIterator*, (__gnu_cxx::_Lock_polic
   0.87%  node     libnode.so.115        [.] v8::internal::Factory::InitializeJSObjectBody(v8::internal::JSObject
   0.84%  node     libnode.so.115        [.] v8::internal::GlobalHandles::ProcessWeakYoungObjects(v8::internal::R
   0.81%  node     libnode.so.115        [.] v8::internal::JSObject::DefineOwnPropertyIgnoreAttributes(v8::intern
   0.80%  node     libnode.so.115        [.] v8::internal::Managed<icu_76::UnicodeString>::Destructor(void*)
   0.79%  node     libnode.so.115        [.] 0x000000000127ff72
   0.77%  node     libnode.so.115        [.] void v8::internal::CopyChars<unsigned short, unsigned char>(unsigned
   0.75%  node     libnode.so.115        [.] v8::internal::Factory::AllocateRaw(int, v8::internal::AllocationType
   0.75%  node     libnode.so.115        [.] v8::internal::Factory::InitializeJSObjectFromMap(v8::internal::JSObj
   0.74%  node     libicuuc.so.76.1      [.] icu_76::UnicodeString::releaseArray()
   0.73%  node     libnode.so.115        [.] v8::internal::JSObject::IsExtensible(v8::internal::Isolate*, v8::int
   0.72%  node     libnode.so.115        [.] void v8::internal::LookupIterator::Start<false>()
   0.72%  node     libnode.so.115        [.] v8::internal::JSSegmentIterator::Next(v8::internal::Isolate*, v8::in
   0.71%  node     libnode.so.115        [.] v8::internal::Factory::NewJSIteratorResult(v8::internal::Handle<v8::
   0.71%  node     libnode.so.115        [.] v8::internal::Managed<icu_76::BreakIterator>::Destructor(void*)
   0.69%  node     libnode.so.115        [.] v8::internal::Object::TransitionAndWriteDataProperty(v8::internal::L
   0.67%  node     libicuuc.so.76.1      [.] icu_76::ForwardCharacterIterator::ForwardCharacterIterator()
   0.63%  node     libnode.so.115        [.] v8::internal::GlobalHandles::InvokeSecondPassPhantomCallbacks()
   0.63%  node     libc.so.6             [.] pthread_mutex_trylock
   0.60%  node     libicuuc.so.76.1      [.] icu_76::RBBIDataWrapper::removeReference()
   0.60%  node     ld-linux-x86-64.so.2  [.] __tls_get_addr
   0.59%  node     libnode.so.115        [.] v8::internal::Factory::AllocateRawWithAllocationSite(v8::internal::H
   0.55%  node     libjemalloc.so.2      [.] operator new(unsigned long)
```

Do not become that slow.

### Client information

<details>

```console
│ About Gemini CLI                                                 │
│                                                                  │
│ CLI Version           0.1.16                                     │
│ Git Commit            820169ba (local modifications)             │
│ Model                 gemini-2.5-pro                             │
│ Sandbox               no sandbox                                 │
│ OS                    linux                                      │
│ Auth Method           gemini-api-key  
```

</details>

### Login information

API key

_No response_

jakemac53:
## TLDR

Fix the performance of the pty code path for shell commands.

## Dive Deeper

Previously, on each chunk of output received from the terminal we would then re-compute the entire terminal output from its buffer, which was very inefficient.

Now, we just send each chunk as an event instead of the entire thing, and accumulate the total output on the receiving end of those events. Note that this is still potentially slow though, since we are still re-creating the full output string when we go to update the display (if there are new chunks) - although it should be a fair bit better than it was.

The behavior between the pty and fallback modes should also be more consistent now (that mode already was sending chunks, and I think was likely somewhat broken since the entire output was being bashed over instead of concatenated).

~This could be further improved by collecting the chunks and only concatenating them when we decide to do a UI update (see the checks for OUTPUT_UPDATE_INTERVAL_MS).~ implemented this as well

## Reviewer Test Plan

You can ask gemini to run a command with lots of shell output and watch the CPU usage of gemini cli, or measure it in some other manner, with and without this change.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes #5411

### DIFF
--- a/packages/core/src/services/shellExecutionService.test.ts
+++ b/packages/core/src/services/shellExecutionService.test.ts
@@ -133,7 +133,7 @@ describe('ShellExecutionService', () => {
 
       expect(onOutputEventMock).toHaveBeenCalledWith({
         type: 'data',
-        chunk: 'file1.txt',
+        chunk: 'file1.txt\n',
       });
     });
 
--- a/packages/core/src/tools/shell.ts
+++ b/packages/core/src/tools/shell.ts
@@ -149,7 +149,7 @@ class ShellToolInvocation extends BaseToolInvocation<
           switch (event.type) {
             case 'data':
               if (isBinaryStream) break;
-              cumulativeOutput = event.chunk;
+              cumulativeOutput += event.chunk;
               currentDisplayOutput = cumulativeOutput;
               if (Date.now() - lastUpdateTime > OUTPUT_UPDATE_INTERVAL_MS) {
                 shouldUpdate = true;
--- a/packages/core/src/tools/shell.ts
+++ b/packages/core/src/tools/shell.ts
@@ -132,6 +132,7 @@ class ShellToolInvocation extends BaseToolInvocation<
       );
 
       let cumulativeOutput = '';
+      let outputChunks: string[] = [cumulativeOutput];
       let lastUpdateTime = Date.now();
       let isBinaryStream = false;
 
@@ -149,9 +150,11 @@ class ShellToolInvocation extends BaseToolInvocation<
           switch (event.type) {
             case 'data':
               if (isBinaryStream) break;
-              cumulativeOutput += event.chunk;
-              currentDisplayOutput = cumulativeOutput;
+              outputChunks.push(event.chunk);
               if (Date.now() - lastUpdateTime > OUTPUT_UPDATE_INTERVAL_MS) {
+                cumulativeOutput = outputChunks.join('');
+                outputChunks = [cumulativeOutput];
+                currentDisplayOutput = cumulativeOutput;
                 shouldUpdate = true;
               }
               break;
--- a/packages/core/src/services/shellExecutionService.ts
+++ b/packages/core/src/services/shellExecutionService.ts
@@ -348,7 +348,6 @@ export class ShellExecutionService {
         });
         let processingChain = Promise.resolve();
         let decoder: TextDecoder | null = null;
-        let output = '';
         const outputChunks: Buffer[] = [];
         const error: Error | null = null;
         let exited = false;
@@ -385,9 +384,10 @@ export class ShellExecutionService {
                 if (isStreamingRawContent) {
                   const decodedChunk = decoder.decode(data, { stream: true });
                   headlessTerminal.write(decodedChunk, () => {
-                    const newStrippedOutput = getFullText(headlessTerminal);
-                    output = newStrippedOutput;
-                    onOutputEvent({ type: 'data', chunk: newStrippedOutput });
+                    onOutputEvent({
+                      type: 'data',
+                      chunk: stripAnsi(decodedChunk),
+                    });
                     resolve();
                   });
                 } else {
@@ -420,7 +420,7 @@ export class ShellExecutionService {
 
               resolve({
                 rawOutput: finalBuffer,
-                output,
+                output: getFullText(headlessTerminal),
                 exitCode,
                 signal: signal ?? null,
                 error,
--- a/packages/core/src/tools/shell.test.ts
+++ b/packages/core/src/tools/shell.test.ts
@@ -313,7 +313,7 @@ describe('ShellTool', () => {
         // Send a second chunk. THIS event triggers the update with the CUMULATIVE content.
         mockShellOutputCallback({
           type: 'data',
-          chunk: 'hello world',
+          chunk: 'world',
         });
 
         // It should have been called once now with the combined output.


jakemac53:
cc @bganne can you confirm if this is sufficient to resolve your issue?

packages/core/src/tools/shell.ts
```diff
@@ -149,7 +149,7 @@ class ShellToolInvocation extends BaseToolInvocation<
           switch (event.type) {
             case 'data':
               if (isBinaryStream) break;
-              cumulativeOutput = event.chunk;
+              cumulativeOutput += event.chunk;
               currentDisplayOutput = cumulativeOutput;
```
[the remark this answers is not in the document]
jakemac53: Yeah I think some variation of this change would be good, but not this change exactly (since this is joining all the chunks repeatedly instead of just appending the new chunks to an already joined string).

packages/core/src/services/shellExecutionService.test.ts
```diff
@@ -133,7 +133,7 @@ describe('ShellExecutionService', () => {
 
       expect(onOutputEventMock).toHaveBeenCalledWith({
         type: 'data',
-        chunk: 'file1.txt',
+        chunk: 'file1.txt\n',
```
cornmander, line 136: why the difference in behavior?

jakemac53: So the way this worked previously it would get all the lines from the terminal (no newlines), and then join them with `\n`, which would remove trailing newlines if present it appears (you didn't get an empty final line in this case from the terminal API?)

cornmander (approved):

bganne:
> cc @bganne can you confirm if this is sufficient to resolve your issue?

@jakemac53 thanks! Sure, is it part of some release or pre-release already?
I usually just do `npm install -g @google/gemini-cli` and I tried various form of `npm install -g 'https://github.com/google-gemini/gemini-cli#main'` but without much success so far.

PR #7687, opened 2025-09-03
fix(read_many_files): address flaky integration test

allenhutchison:
This commit addresses a flaky integration test for the `read_many_files` tool.

The key changes include:
- Waiting for the telemetry to be ready before checking for tool calls to ensure that all tool calls are available for inspection.
- Replacing `waitForToolCall` with a more robust method of finding the tool call from the complete tool log.
- Enhanced debugging output to provide more context on test failures.
- Simplified the model output validation to check for a non-empty result, making the test less brittle.

### DIFF
--- a/integration-tests/read_many_files.test.ts
+++ b/integration-tests/read_many_files.test.ts
@@ -18,9 +18,14 @@ describe('read_many_files', () => {
 
     const result = await rig.run(prompt);
 
+    // Wait for telemetry to be available before checking tool calls
+    await rig.waitForTelemetryReady();
+
     // Check for either read_many_files or multiple read_file calls
     const allTools = rig.readToolLogs();
-    const readManyFilesCall = await rig.waitForToolCall('read_many_files');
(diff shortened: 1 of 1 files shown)


integration-tests/read_many_files.test.ts
```diff
@@ -18,21 +18,28 @@ describe('read_many_files', () => {
 
     const result = await rig.run(prompt);
 
+    // Wait for telemetry to be available before checking tool calls
+    await rig.waitForTelemetryReady();
```
cornmander, line 22: Should this be in rig.setup() Seems like we'd want this by default?

allenhutchison: It's in most of the paths that wait for a tool call. This test is a little different because the model could decide to call read many files, or it can call read file multiple times. I didn't setup the harness to support waiting for multiple tools calls so in this case we wait for the telemetry to be ready and then get all the tool calls and parse the list manually.

Issue #7664, opened 2025-09-03
Error code 400


hasanfq6:
### What happened?

Suddenly after recent updates the following error accured 

╭───────────╮
│  > /auth  │
╰───────────╯

╭──────────────────╮
│  > geneeatw a d  │
╰──────────────────╯

✕ [API Error: [{
    "error": {
      "code": 400,
      "message": "The GenerateContentRequest proto is invalid:\n  *
  tools[0].function_declarations[12].name: [FIELD_INVALID] Invalid function name.
  Must start with a letter or an underscore. Must be a-z, A-Z, 0-9, or contain
  underscores, dots and dashes, with a maximum length of 64.",
      "errors": [
        {
          "message": "The GenerateContentRequest proto is invalid:\n  *
  tools[0].function_declarations[12].name: [FIELD_INVALID] Invalid function name.
  Must start with a letter or an underscore. Must be a-z, A-Z, 0-9, or contain
  underscores, dots and dashes, with a maximum length of 64.",
          "domain": "global",
          "reason": "badRequest"
        }
      ],
      "status": "INVALID_ARGUMENT"
    }
  }
  ]]

### What did you expect to happen?

A normal conversation 

### Client information

<details>
<summary>Client Information</summary>

Run `gemini` to enter the interactive CLI, then run the `/about` command.

```console
> /about

╭────────────────────────────────────────────────────────────────────────────────╮
│                                                                                │
│ About Gemini CLI                                                               │
│                                                                                │
│ CLI Version                0.2.2                                               │
│ Git Commit                 32204dca                                            │
│ Model                      gemini-2.5-pro                                      │
│ Sandbox                    no sandbox                                          │
│ OS                         android                                             │
│ Auth Method                OAuth                                               │
│                                                                                │
╰──────────────────────────────────────────────────────
```

</details>

### Login information

Goggle Account

srithreepo:
Hi Could you share a little more information on what you were doing leading upto this issue? Do you see this when using specific prompts ?

hasanfq6:
No it happens suddenly for every prompt even hi and others, it happens after last update. It works fine before

PaulRBerg:
In my case, the issue was due to using the Magic MCP:

```json
"magic": {
  "command": "npx",
  "trust": true,
  "args": ["-y", "@21st-dev/magic@latest"],
  "env": {
    "API_KEY": "$API_KEY_MAGIC_MCP"
  }
}
```

Full error log:

```text
[...]

PR #7692, opened 2025-09-04
fix(core): tend to history with dangling function calls/responses

agmsb:
## TLDR

Dangling function calls and responses can result in 400 errors: `Please ensure that function response turn comes immediately after a function call turn.` In this PR we add a `trimRecentHistory` function to ensure that we do not leave leading function responses in the beginning of history or trailing function calls at the end of history used for loop detection.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | X  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Potentially related to #7664.


silviojr:
Curious: Does this only happen in the loop detection service?

NTaylorMullen (approved):

packages/core/src/services/loopDetectionService.ts
```diff
@@ -328,11 +330,37 @@ export class LoopDetectionService {
...
+
+    // A function response must follow a function call.
+    // Continuously removes leading function responses from the beginning of history
+    // until the first turn is not a function response.
+    while (
+      recentHistory.length > 0 &&
+      isFunctionResponse(recentHistory[0])
+    ) {
+      recentHistory.shift();
+    }
```
SandyTao520, line 352: I think it's legal that history starts with a function response, but having this here makes it clearer to the model.

SandyTao520 (approved):
LGTM as a temporary fix.

packages/core/src/services/loopDetectionService.ts
agmsb: I'll update comments to reflect.

agmsb:
> Curious: Does this only happen in the loop detection service?

We'll need to continue investigations as this is only a temporary fix for instances where we caught this error.

Amnat-auyana1998 (approved):
@agmsb Commit

PR #7723, opened 2025-09-04
Remove experimental extensions management flag

chrstnb:
## TLDR

We want these changes to be live in the next release. This will also deprecate workspace extensions fully with a migration path available. 

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

### DIFF
--- a/packages/core/src/config/config.ts
+++ b/packages/core/src/config/config.ts
@@ -208,7 +208,6 @@ export interface ConfigParameters {
   useRipgrep?: boolean;
   shouldUseNodePtyShell?: boolean;
   skipNextSpeakerCheck?: boolean;
-  extensionManagement?: boolean;
   enablePromptCompletion?: boolean;
   eventEmitter?: EventEmitter;
   useSmartEdit?: boolean;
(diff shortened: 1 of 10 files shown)


mattKorwel:
@chrstnb we'll need to patch this into #7753

packages/core/src/config/config.ts
```diff
@@ -702,10 +699,6 @@ export class Config {
     return this.listExtensions;
   }
 
-  getExtensionManagement(): boolean {
```
anj-s, line 705: Can we keep this flag until we roll out? This allows us to turn off if there are issues. We can do the cleanup after the rollout IMO. Let me know if I understood this right.

chrstnb: ah good call-- just set default to true in [this PR](https://github.com/google-gemini/gemini-cli/pull/7807) instead. will make this one a draft for when we actually want to remove it.

PR #7731, opened 2025-09-04
Fix(extension) - Add pid into portFile

shishu314:
## TLDR

Add pid into the portFile name

## Dive Deeper

The client does not have access to the port info until the config is read. Adding pid into the portFile names help us reduce the number of configs to filter through when figuring out the current active port.

## Reviewer Test Plan

`npm run test` is the only way to test since port files are not used yet.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | x  | ❓  | ❓  |
| npx      | x  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs
Makes progress on #6848

### DIFF
--- a/packages/vscode-ide-companion/src/ide-server.test.ts
+++ b/packages/vscode-ide-companion/src/ide-server.test.ts
@@ -124,7 +124,7 @@ describe('IDEServer', () => {
     const port = getPortFromMock(replaceMock);
     const expectedPortFile = path.join(
       '/tmp',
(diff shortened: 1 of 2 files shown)


skeshive:
Closing this for now as we discussed offline that the original issue was deprioritized.

PR #7749, opened 2025-09-04
fix(ci): fix windows e2e tests

galdawave:
Fixes e2e tests for windows by adding cross-env to test runs and setting shell type in e2e.yml


.github/workflows/e2e.yml
```diff
@@ -81,5 +81,6 @@ jobs:
           KEEP_OUTPUT: 'true'
           SANDBOX: '${{ matrix.sandbox }}'
           VERBOSE: 'true'
+        shell: 'bash'
```
mattKorwel, line 83: does bash work on windows?

galdawave: Removed, this is no longer needed.

mattKorwel (approved):

mattKorwel (approved):

PR #7750, opened 2025-09-04
Refactor Authentication Components and Hooks

scidomino:
## TLDR

Move Auth UI code into it's own directory. Introduce AuthState object and refactor code.

## Dive Deeper

This is needed for https://github.com/google-gemini/gemini-cli/issues/4544 since initializing the MCP servers needs to happen before Auth and we need to be clever about when to turn it on and turn it off.

## Reviewer Test Plan

* Startup with no selectedAuthType. Verify that you are forced to auth and can't escape.
* Startup with no selectedAuthType but with GEMINI_API_KEY. Verify that you are forced to auth and can't escape and Gemini API Key option is selected.
* Startup with selectedAuthType. Verify no auth required.
* execute `/auth` and switch auth types. Verify it works as expected.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | x  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Contributes to #4544


packages/cli/src/ui/auth/useAuth.ts
```diff
@@ -0,0 +1,96 @@
...
+        await config.refreshAuth(authType);
+
+        console.log(`Authenticated via "${authType}".`);
+        setAuthError(null);
+        setAuthState(AuthState.Authenticated);
+      } catch (e) {
+        onAuthError(`Failed to login. Message: ${getErrorMessage(e)}`);
+      }
+    })();
+  }, [settings, config, authState, setAuthState, setAuthError, onAuthError]);
```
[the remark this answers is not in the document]
scidomino: We don't do this anywhere else as far as I can see. And this is an existing issue so I'll look into it in a follow up PR.

packages/cli/src/config/auth.ts
```diff
@@ -9,6 +9,7 @@ import { loadEnvironment, loadSettings } from './settings.js';
 
 export function validateAuthMethod(authMethod: string): string | null {
   loadEnvironment(loadSettings().merged);
+
```
allenhutchison: Extra line?

packages/cli/src/ui/App.tsx
```diff
@@ -338,52 +338,18 @@ const App = ({ config, settings, startupWarnings = [], version }: AppProps) => {
...
-      }
-    }
-  }, [
-    settings.merged.security?.auth?.selectedType,
-    settings.merged.security?.auth?.enforcedType,
-    settings.merged.security?.auth?.useExternal,
-    openAuthDialog,
-    setAuthError,
-  ]);
+  const { authState, setAuthState, authError, onAuthError } = useAuthCommand(
```
allenhutchison, line 341: Should this be memoized to prevent rerenders?

allenhutchison (approved):
Just a few nits / questions to consider.

packages/cli/src/ui/App.tsx
scidomino: I don't think we need to memoize because these functions are already stable (they're returned by useState and useCallback).

packages/cli/src/config/auth.ts
scidomino: fixed.

PR #7759, opened 2025-09-04
Fix syntax highlighting and rendering issues.

### LINKED ISSUE #7683: Feature Request: Render empty newlines in multi-line input mode
### What happened?

I would like my input to match my output. When I press CTRL-J twice, I'd like to see the blank line in the input area before I send the command, not just in the final output.

Reproduction steps:

1. Start Gemini CLI, no options
2. Type any one line string into input
3. Press CTRL-J for a new line twice
4. Type more text

There is no blank line in-between the two typed lines of text. I see the new line when I send off the request, but not when I'm typing it.
I would like my input to match my output.

There should be a new line typed for each CTRL-J sent, even if they're blank.

### Client information

* **CLI Version:** 0.2.2-git.4d07cb7d
* **Git Commit:** 4d07cb7d
* **Session ID:** ebadf4a4-bbd1-49bd-a39f-cf8d06356cfe
* **Operating System:** linux v22.5.1
* **Memory Usage:** 225.5 MB

### Login information

_No response_

_No response_

jacob314:
## TLDR

Blank lines weren't rendering properly with syntax highlighting. Fixed by wrapping each line with a  Box with a line height of 1 for reliable ink rendering of lines.

Also fixed syntax highlighting to only highlight tool calls that are the start of the first line.

## Reviewer Test Plan

Test prompts such as
`foo

bar
` 
where the blank line should render. 

`/tool `
`not a /tool call`
`escaped file 

## Linked issues / bugs

- Fixes #7683
- Fixes #7651


packages/cli/src/ui/components/InputPrompt.tsx
```diff
@@ -784,7 +787,9 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
                 ) {
                   if (!currentLineGhost) {
                     renderedLine.push(
-                      <Text key="cursor-end">{chalk.inverse(' ')}</Text>,
+                      <Text key={`cursor-end-${cursorVisualColAbsolute}`}>
```
jacob314, line 790: was seeing strange ink artifacts without this key being unique when going from `@Foo ` to `@Foo  `

miguelsolorio (approved):
LGTM, tested new lines & slashes in the input and works perfectly 

thanks for fixing this ❤️

allenhutchison (approved):

psinha40898:
`git checkout d8dbe627^1`
1. Write path with length > 20
2. Reduce terminal width to < 20 columns
3. Observe cursor is at the end of the input

image


`git checkout d8dbe627`
1. Write path with length > 20
2. Reduce terminal width to < 20 columns
3. Observe cursor unexpected behavior

image


iTerm2

Not sure if width this narrow should be supported, or if I made a mistake in observing the bug.

PR #7782, opened 2025-09-05
fix(config): Correct settings migration to prevent overwrite

JayadityaGit:
<h2>TLDR</h2>
<p>While testing, I noticed that <code>customWittyPhrases</code> were effectively being <strong>“Thanos snapped”</strong> (disappearing) after <code>settings.json</code> was updated.</p>
<ul>
<li>
<p>Cause: <code>customWittyPhrases</code> wasn’t properly preserved in the migration, since it was NOT present in the migration map.</p>
</li>
<li>
<p>Effect: after a theme change or file rewrite, <code>customWittyPhrases</code> vanished on the next load.</p>
</li>
</ul>
<p>✅ <strong>Fix:</strong> I changed the migration logic so that objects are <strong>merged instead of replaced</strong>.</p>
<p>Now, just like <code>theme</code>, users can keep <code>customWittyPhrases</code> as a <strong>top-level key</strong> in <code>settings.json</code> without worrying about losing them—even after the file updates.</p>
<hr>
<h2>Dive Deeper</h2>
<ul>
<li>
<p><strong>Old behavior:</strong></p>
<ul>
<li>
<p>Migrating <code>customWittyPhrases</code> into <code>ui</code> worked, but the <code>ui</code> object from V2 would overwrite it during carry-over.</p>
</li>
<li>
<p>This wiped out the migrated phrases.</p>
</li>
</ul>
</li>
<li>
<p><strong>New behavior:</strong></p>
<ul>
<li>
<p>The migration process now deep-merges objects instead of replacing them.</p>
</li>
<li>
<p><code>customWittyPhrases</code> is preserved regardless of settings.json override.</p>
</li>
</ul>
</li>
</ul>
<p><strong>Example top-level settings.json:</strong></p>
<pre><code class="language-json">{
  "theme": "GitHub",
  "customWittyPhrases": [
    "Getting this Before GTA6",
    "Connecting to AGI"
  ]
}
</code></pre>
<p>After migration, these values remain intact—even if <code>settings.json</code> is rewritten after theme changes.</p>
<hr>
<h2>Reviewer Test Plan</h2>
<ol>
<li>
<p>Create a settings file with:</p>
<ul>
<li>
<p>A top-level <code>customWittyPhrases</code> array.</p>
</li>
<li>
<p>Or a nested <code>ui.customWittyPhrases</code>.</p>
</li>
</ul>
</li>
<li>
<p>Trigger a migration (e.g., by changing the theme).</p>
</li>
<li>
<p>Verify that <code>customWittyPhrases</code> remains intact in both cases.</p>
</li>
<li>
<p>Confirm that no other settings are lost.</p>
</li>
</ol>
<hr>
<h2>Testing Matrix</h2>

  | 🍏 | 🪟 | 🐧
-- | -- | -- | --
npm run | ❓ | ❓ | ❓
npx | ❓ | ❓ | ❓
Docker | ❓ | ❓ | ❓
Podman | ❓ | - | -
Seatbelt | ❓ | - | -

<hr>
<h2>Linked issues / bugs</h2>
<ul>
<li>
<p>Resolves #️⃣ (#7781)</p>
</li>
</ul>

### DIFF
--- a/packages/cli/src/config/settings.test.ts
+++ b/packages/cli/src/config/settings.test.ts
@@ -2168,12 +2168,12 @@ describe('Settings Loading and Merging', () => {
     });
 
     it('should correctly migrate customWittyPhrases', () => {
-      const v2Settings = {
+      const v2Settings: Partial<Settings> = {
         ui: {
           customWittyPhrases: ['test phrase'],
         },
       };
-      const v1Settings = migrateSettingsToV1(v2Settings);
+      const v1Settings = migrateSettingsToV1(v2Settings as Settings);
       expect(v1Settings).toEqual({
         customWittyPhrases: ['test phrase'],
       });
@@ -2236,7 +2236,7 @@ describe('Settings Loading and Merging', () => {
     });
 
     it('should return false for settings that are already in V2 format', () => {
-      const v2Settings = {
+      const v2Settings: Partial<Settings> = {
         ui: {
           theme: 'dark',
         },
--- a/packages/cli/src/config/settings.test.ts
+++ b/packages/cli/src/config/settings.test.ts
@@ -2166,6 +2166,18 @@ describe('Settings Loading and Merging', () => {
         allowMCPServers: ['serverA'],
       });
     });
+
+    it('should correctly migrate customWittyPhrases', () => {
+      const v2Settings = {
+        ui: {
+          customWittyPhrases: ['test phrase'],
+        },
+      };
+      const v1Settings = migrateSettingsToV1(v2Settings);
+      expect(v1Settings).toEqual({
+        customWittyPhrases: ['test phrase'],
+      });
+    });
   });
 
   describe('loadEnvironment', () => {
--- a/packages/cli/src/config/settings.ts
+++ b/packages/cli/src/config/settings.ts
@@ -28,7 +28,7 @@ import {
   getSettingsSchema,
 } from './settingsSchema.js';
 import { resolveEnvVarsInObject } from '../utils/envVarResolver.js';
-import { customDeepMerge } from '../utils/deepMerge.js';
+import { customDeepMerge, type MergeableObject } from '../utils/deepMerge.js';
 
 function getMergeStrategyForPath(path: string[]): MergeStrategy | undefined {
   let current: SettingDefinition | undefined = undefined;
@@ -258,10 +258,14 @@ function migrateSettingsToV2(
       newValue !== null &&
       !Array.isArray(newValue)
     ) {
-      v2Settings[remainingKey] = {
-        ...(newValue as Record<string, unknown>),
-        ...(existingValue as Record<string, unknown>),
-      };
+      const pathAwareGetStrategy = (path: string[]) =>
+        getMergeStrategyForPath([remainingKey, ...path]);
+      v2Settings[remainingKey] = customDeepMerge(
+        pathAwareGetStrategy,
+        {},
+        newValue as MergeableObject,
+        existingValue as MergeableObject,
+      );
     } else {
       v2Settings[remainingKey] = newValue;
     }
--- a/packages/cli/src/config/settings.ts
+++ b/packages/cli/src/config/settings.ts
@@ -67,6 +67,7 @@ const MIGRATION_MAP: Record<string, string> = {
   coreTools: 'tools.core',
   contextFileName: 'context.fileName',
   customThemes: 'ui.customThemes',
+  customWittyPhrases: 'ui.customWittyPhrases',
   debugKeystrokeLogging: 'general.debugKeystrokeLogging',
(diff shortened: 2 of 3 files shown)


JayadityaGit:
Hello @cornmander, would you mind reviewing this whenever it’s convenient for you? I’d really appreciate your time and feedback.

packages/cli/src/config/settings.test.ts
```diff
@@ -2166,6 +2166,18 @@ describe('Settings Loading and Merging', () => {
         allowMCPServers: ['serverA'],
       });
     });
+
+    it('should correctly migrate customWittyPhrases', () => {
+      const v2Settings = {
```
richieforeman: Please add a structural type onto this (I think `Partial<Settings>` will do).

packages/cli/src/config/settings.test.ts
```diff
@@ -2166,6 +2166,18 @@ describe('Settings Loading and Merging', () => {
...
+        ui: {
+          customWittyPhrases: ['test phrase'],
+        },
+      };
+      const v1Settings = migrateSettingsToV1(v2Settings);
```
richieforeman: If you add `Partial<Settings>` above, you may need to cast `vsSettings as Settings` here.

packages/cli/src/config/settings.ts
```diff
@@ -246,7 +247,28 @@ function migrateSettingsToV2(
 
   // Carry over any unrecognized keys
   for (const remainingKey of flatKeys) {
-    v2Settings[remainingKey] = flatSettings[remainingKey];
+    const existingValue = v2Settings[remainingKey];
+    const newValue = flatSettings[remainingKey];
+
+    if (
+      typeof existingValue === 'object' &&
```
richieforeman, line 254: I think you only need the `typeof .... === 'object'` checks here.

- if it's an object, it won't be null
- If it's an object, it won't be an array

The additional checks for `!== null` and `!Array` seems extraneous given the object constraint here.

JayadityaGit:
Thank you for the review, @richieforeman. I truly appreciate the time and effort you’ve given to this PR. I’ll work on completing the changes by Monday. I’m currently going through a bit of a rough time, so thank you once again.

packages/cli/src/config/settings.ts
JayadityaGit: ### Act 1: A User Sets Their Custom Phrase

1. **User’s Action**: The user opens `settings.json` and adds a custom phrase:

   ```json
   {
     "customWittyPhrases": [
       "You forget a thousand things every day. Make sure this is one of ’em"
     ]
   }
   ```
2. **App’s Action**: The app starts up and displays the custom phrase. Everything works as expected.

---

### Act 2: The Bug (The Problem Before My Fix)

1. **User’s Action**: The user changes their theme.
2. **App’s Old Logic**: The app didn’t know how to merge objects. It treated the new theme (`ui: { theme: "GitHub" }`) as a completely new `ui` object and **overwrote** the old one.
3. **Result**: The user’s `customWittyPhrases` were erased from memory.

---

### Act 3: How My Logic Fixes the Bug

My code changes this behavior entirely.

1. **User’s Action**: The user changes the theme again.
2. **My Logic in Action**:

   * The app receives the new theme (`ui: { theme: "GitHub" }`).
   * It looks at what it already has in memory (`ui: { customWittyPhrases: [...] }`).
   * My new `if` statement asks: *“Are both the old and new `ui` values plain objects?”*
3. **Result**: The answer is **yes**, so the app merges them instead of replacing.

   * Old: `{ customWittyPhrases: [...] }`
   * New: `{ theme: "GitHub" }`
   * Final merged object: `{ customWittyPhrases: [...], theme: "GitHub" }`
   * The user’s phrases are preserved.

The core of my fix is this: the app now **merges objects instead of blindly replacing them**.

---

### Why the Extra Checks Matter

Now let’s see why the detailed checks (like `!Array.isArray`) are critical.

1. **User’s Action**: They update `settings.json` to replace their phrases:

   ```json
   {
     "customWittyPhrases": ["A brand new phrase"]
   }
   ```
2. **My Logic in Action**: The app compares old vs. new:

   * Old: `["You forget a thousand things..."]`
   * New: `["A brand new phrase"]`
3. **The Check**: `!Array.isArray()` makes sure arrays are not mistaken for mergeable objects. Since both values are arrays, the condition fails.
4. **Result**: The app replaces the old array with the new one — exactly what the user expects.

---

✨ **In short:** My fix makes the app smart enough to **merge objects when appropriate** while still **replacing arrays when users intend to overwrite them**.

JayadityaGit:
This branch has become quite complex, and I missed pulling in the recent changes. I’ll create a new branch and reference it instead. Closing this one for now.

PR #7790, opened 2025-09-05
fix: enable cancellation of slash commands and background processes

### LINKED ISSUE #6066: Slash commands do not respect top-level cancellation signal

akhil29:
## Summary

Enables proper cancellation when users press Escape during slash command execution, including termination of background processes spawned by shell commands. Addresses issue #6066 where slash commands did not respect the top-level cancellation signal.

## Key Changes

- **AbortSignal propagation**: Added signal parameter to slash command processor chain
- **Background process termination**: Implemented `killAllBackgroundProcesses()` with cross-platform support
- **Global PID tracking**: Added tracking for background processes across async boundaries
- **Signal integration**: Connected cancellation hooks throughout command execution flow
- **Comprehensive tests**: Added test coverage for new functionality

### Core Cancellation Flow
1. User presses Escape → `useGeminiStream.cancelOngoingRequest()`
2. Calls `killAllBackgroundProcesses()` to terminate spawned processes
3. AbortSignal propagates through `handleSlashCommand` → `createCommandContext`
4. Commands receive signal and can respond to cancellation appropriately

### Background Process Management
- **Cross-platform termination**: Uses process groups (SIGTERM/SIGKILL) on Unix, taskkill on Windows
- **Global PID tracking**: Maintains Set of background PIDs for cleanup on cancellation
- **Graceful shutdown**: SIGTERM followed by SIGKILL after timeout

### Code Quality Improvements
- **Refactored to useCallback**: Addressed memoization code smell from PR feedback
- **Signal parameter pattern**: Changed from creating dummy signals to accepting signal as parameter
- **App-level AbortController**: Added for programmatic command lifecycle management

## Test Coverage

- ✅ **Background process termination**: Tests Unix/Windows process killing, error handling, PID tracking
- ✅ **Integration tests**: Verifies cancellation flow in `useGeminiStream`
- ✅ **Signal propagation**: Tests AbortSignal parameter passing through command chain
- ✅ **Cross-platform support**: Tests both Unix and Windows process termination methods

## Fixes

- Background processes now properly terminate when cancellation occurs (addresses #6066)
- Shell commands like `/tmp/long_counter.sh &` can be cancelled with Escape
- Proper signal propagation prevents zombie processes and resource leaks
- useCallback pattern eliminates memoization code smell from previous implementation

Fixes #6066

### DIFF
--- a/packages/cli/src/ui/hooks/useGeminiStream.test.tsx
+++ b/packages/cli/src/ui/hooks/useGeminiStream.test.tsx
@@ -129,20 +129,6 @@ vi.mock('./useLogger.js', () => ({
   }),
 }));
 
-const mockKillAllBackgroundProcesses = vi.hoisted(() => vi.fn());
-
-vi.mock('@google/gemini-cli-core', async (importOriginal) => {
-  const actualCoreModule = (await importOriginal()) as any;
-  return {
-    ...actualCoreModule,
-    GitService: vi.fn(),
-    GeminiClient: MockedGeminiClientClass,
-    UserPromptEvent: MockedUserPromptEvent,
-    parseAndFormatApiError: mockParseAndFormatApiError,
-    killAllBackgroundProcesses: mockKillAllBackgroundProcesses,
-  };
-});
-
 const mockStartNewPrompt = vi.fn();
 const mockAddUsage = vi.fn();
 vi.mock('../contexts/SessionContext.js', () => ({
@@ -180,6 +166,9 @@ describe('useGeminiStream', () => {
       // MockedGeminiClientClass is defined in the module scope by the previous change.
       // It will use the mockStartChat and mockSendMessageStream that are managed within beforeEach.
       const clientInstance = new MockedGeminiClientClass(mockConfig);
+      clientInstance.getLoopDetectionService = vi.fn().mockReturnValue({
+        disableForSession: vi.fn(),
+      });
       return clientInstance;
     });
 
@@ -322,6 +311,8 @@ describe('useGeminiStream', () => {
           () => {},
           () => {},
           mockSetShellInputFocused,
+          80,
+          24,
         );
       },
       {
@@ -485,6 +476,8 @@ describe('useGeminiStream', () => {
         () => {},
         () => {},
         mockSetShellInputFocused,
+        80,
+        24,
       ),
     );
 
@@ -566,6 +559,8 @@ describe('useGeminiStream', () => {
         () => {},
         () => {},
         mockSetShellInputFocused,
+        80,
+        24,
       ),
     );
 
@@ -676,6 +671,8 @@ describe('useGeminiStream', () => {
         () => {},
         () => {},
         mockSetShellInputFocused,
+        80,
+        24,
       ),
     );
 
@@ -787,6 +784,8 @@ describe('useGeminiStream', () => {
         () => {},
         () => {},
         mockSetShellInputFocused,
+        80,
+        24,
       ),
     );
 
@@ -918,6 +917,8 @@ describe('useGeminiStream', () => {
           () => {},
           cancelSubmitSpy,
           mockSetShellInputFocused,
+          80,
+          24,
         ),
       );
 
@@ -929,6 +930,32 @@ describe('useGeminiStream', () => {
       simulateEscapeKeyPress();
 
       expect(cancelSubmitSpy).toHaveBeenCalled();
+    });
+
+    it('should call setShellInputFocused(false) when escape is pressed', async () => {
+      const mockStream = (async function* () {
+        yield { type: 'content', value: 'Part 1' };
+        await new Promise(() => {}); // Keep stream open
+      })();
+      mockSendMessageStream.mockReturnValue(mockStream);
+
+      const { result } = renderTestHook();
+
+      // Start a query
+      await act(async () => {
+        result.current.submitQuery('test query');
+      });
+
+      // Wait for the response to start
+      await waitFor(() => {
+        expect(result.current.streamingState).toBe(StreamingState.Responding);
+      });
+
+      // Simulate escape key press
+      simulateEscapeKeyPress();
+
+      // Verify setShellInputFocused was called with false
+      expect(mockSetShellInputFocused).toHaveBeenCalledWith(false);
     });
 
     it('should not do anything if escape is pressed when not responding', () => {
@@ -1271,6 +1298,8 @@ describe('useGeminiStream', () => {
           () => {},
           () => {},
           mockSetShellInputFocused,
+          80,
+          24,
         ),
       );
 
@@ -1326,6 +1355,8 @@ describe('useGeminiStream', () => {
           () => {},
           () => {},
           mockSetShellInputFocused,
+          80,
+          24,
         ),
       );
 
@@ -1857,6 +1888,8 @@ describe('useGeminiStream', () => {
           () => {},
           () => {},
           mockSetShellInputFocused,
+          80,
+          24,
         ),
       );
 
@@ -2025,7 +2058,7 @@ describe('useGeminiStream', () => {
           () => {},
           () => {},
           onCancelSubmitSpy,
-          () => {},
+          mockSetShellInputFocused,
           80,
           24,
         ),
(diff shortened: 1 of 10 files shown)


akhil29:
@abhipatel12 pls review.

abhipatel12 (commented):
Hey @akhil29 thanks for the PR. 

It looks like this PR hasn't been properly rebased for the larger refactor that moved most of the logic out of `App.tsx`. I can take a look at this once this is rebased onto the current state of main.

akhil29:
@abhipatel12 sorry about that. The fixed it now. please check.

packages/core/src/tools/shell.ts
```diff
@@ -39,6 +40,31 @@ import {
 
 export const OUTPUT_UPDATE_INTERVAL_MS = 1000;
 
+// Global tracker for background processes that need to be killed on cancellation
+const globalBackgroundPIDs = new Set<number>();
+
+export function killAllBackgroundProcesses(): void {
```
abhipatel12: This seems to tightly couples the UI to the internal impl details of the `ShellTool`

Can we scope the background PID tracking to the specific tool execution context and tie it to the `AbortSignal`.

This would also let us remove the global state, since we can just use the captured PIDs from the temp file. 

WDYT?

abhipatel12 (commented):
Left a comment about the coupling for the shell logic we're adding. lmk what you think and also if you could rebase!

Thanks!

packages/core/src/tools/shell.ts
ChrisMenesesDev: Hi Abhi,

 Thank you very much for the detailed review and for catching the tight coupling and global state issue! I agree that it needs to be fixed.

 As I am still new to development and Git, could you please guide me? I'm not yet familiar with the rebase process and the exact method of scoping the PID tracking to the AbortSignal and using the temp file.

 I'm ready to make the changes, but I need some step-by-step instructions on how to execute them.

 Thanks for your patience and help!

akhil29:
@abhipatel12 done.
Please check now.

abhipatel12 (commented):
Hey this looks good in general. However, it doesn't seem to build. Please be sure to run preflight and push it again.

Additionally, for the shell tool changes, can we create an issue that explores adding the same level of support for windows.

akhil29:
Fixed the build it. sorry about that. PTAL @abhipatel12 

created a new issue https://github.com/google-gemini/gemini-cli/issues/10978

packages/cli/src/ui/hooks/useGeminiStream.test.tsx
```diff
@@ -2571,261 +2521,100 @@ describe('useGeminiStream', () => {
     });
   });
 
-  describe('Loop Detection Confirmation', () => {
```
abhipatel12, line 2767: Did these accidentally get deleted?

packages/cli/src/ui/hooks/useGeminiStream.test.tsx
```diff
@@ -1267,9 +1237,7 @@ describe('useGeminiStream', () => {
           () => {},
           () => {},
           () => {},
-          () => {},
-          80,
```
abhipatel12, line 1641: Did we remove these numbers on purpose?

packages/cli/src/ui/hooks/useGeminiStream.test.tsx
```diff
@@ -927,47 +931,6 @@ describe('useGeminiStream', () => {
       expect(cancelSubmitSpy).toHaveBeenCalled();
     });
 
-    it('should call setShellInputFocused(false) when escape is pressed', async () => {
```
abhipatel12, line 1149: Is this also supposed to be deleted?

package-lock.json
abhipatel12, line 1: Please revert these changes

packages/cli/src/ui/hooks/useGeminiStream.test.tsx
akhil29: Sorry about that. reverting.

akhil29:
Resolved suggested changes. PTAL @abhipatel12

abhipatel12 (approved):
LGTM, thank you for your patience on this!

PR #7800, opened 2025-09-05
fix(mcp): handle `gemini mcp add` scope correctly in home directory

### LINKED ISSUE #7799: [Bug] mcp add from home directory overwrites user settings
Running `gemini mcp add` command from the users home directory (`~`) with the default project scope leads to overwriting the global user settings file with an empty settings file with just the `mcpServers` attribute.

wietsevenema:
Fixes #7799 by throwing an error if `--scope` is set to `project` when running `gemini mcp add` from home directory.

*   If `gemini mcp add --scope project ...` is run explicitly from the home directory, the command exits with an error.

## Detailed Behavior

By default the `--scope` argument of the `gemini mcp add` command is set to `project`. When run from the home directory where `project` and `user` scopes essentially point to the same `settings.json` path... this causes project scope to overwrite the user scope settings file.

To prevent this we can add a check and error if project scope is set when running `gemini mcp add` from the home directory.


packages/cli/src/commands/mcp/add.ts
```diff
@@ -36,9 +36,21 @@ async function addMcpServer(
     includeTools,
     excludeTools,
   } = options;
-  const settingsScope =
-    scope === 'user' ? SettingScope.User : SettingScope.Workspace;
-  const settings = loadSettings();
+
+  const settings = loadSettings(process.cwd());
```
jackwotherspoon, line 40: `process.cwd()` is already the default for `loadSettings` so I don't think we need to explicitly pass it here...

packages/cli/src/commands/mcp/add.ts
```diff
@@ -36,9 +36,21 @@ async function addMcpServer(
...
+  const inHome = settings.workspace.path === settings.user.path;
+
+  if (scope === 'project' && inHome) {
+    console.error(
+      'Error: Cannot use --scope project when running from the home directory.',
```
jackwotherspoon: Let's make this error actionable by adding a second line to the error message:

```
Error: Cannot use --scope project when running from the home directory.
Please use --scope user or run command from within a project.
```

jackwotherspoon (changes requested):
I think I would rather leave `project` as default and just have error thrown if running `gemini mcp add` from home directory with `--scope project` configured (implicitly by default or explicitly).

The `auto` config is ambiguous and I think the `--help` message and usage is more clear to users with project as default. I could also foresee where users in Cloud Shell etc run `gemini mcp add` by accident from the home directory where they actually meant to cd into a project before running the command. An error would be better for this case instead of auto using the user scope.

We also don't recommend running `gemini` from the home directory, there is a warning that appears on startup... so I think its best to error for `gemini mcp add` and have a clear error message that mentions using `--scope user` if user does actually mean to run `gemini mcp add` from home directory.

packages/cli/src/commands/mcp/add.ts
wietsevenema: I know but that makes it very hard to test. So it's either explicit passing, or complex test. You choose.

packages/cli/src/commands/mcp/add.ts
wietsevenema: Got it.

jackwotherspoon:
@wietsevenema I see two options for default `gemini mcp add` scope (which settings.json to update when --scope is not specified):
>
> a) Keep default as `project` and add error if run from home directory
>
>```
>Error: Cannot use --scope project when running from the home directory.
>Please use --scope user or run command from within a project.
>```
>
>b) Make default `auto` which uses project scope everywhere other than home directory in which case it will update user settings.

Chatted with @NTaylorMullen offline and he said:

> Wow what a hard choice. I'd opt for a) because it's something we can easily enable in the future and is safer

So I say we go a) for now and keep project as the default with check for if home==workspace and error message.

We can add auto in later if need be 👍

jackwotherspoon:
@wietsevenema let me know if you want me to take this PR over of if you are good to make the changes for option a) above 😄

wietsevenema:
@jackwotherspoon thanks for the nudge. I'll take care of it later today

wietsevenema:
@jackwotherspoon PTAL

packages/cli/src/commands/mcp/add.ts
```diff
@@ -107,7 +117,7 @@ async function addMcpServer(
   const isExistingServer = !!mcpServers[name];
   if (isExistingServer) {
     console.log(
-      `MCP server "${name}" is already configured within ${scope} settings.`,
+      `MCP server "${name}" is already configured within ${settingsScope} settings.`,
```
jackwotherspoon: I believe we can put these back to just `scope`, now...
```suggestion
      `MCP server "${name}" is already configured within ${scope} settings.`,
```

packages/cli/src/commands/mcp/add.ts
```diff
@@ -116,10 +126,10 @@ async function addMcpServer(
   settings.setValue(settingsScope, 'mcpServers', mcpServers);
 
   if (isExistingServer) {
-    console.log(`MCP server "${name}" updated in ${scope} settings.`);
+    console.log(`MCP server "${name}" updated in ${settingsScope} settings.`);
```
jackwotherspoon:
```suggestion
    console.log(`MCP server "${name}" updated in ${scope} settings.`);
```

packages/cli/src/commands/mcp/add.ts
```diff
@@ -116,10 +126,10 @@ async function addMcpServer(
...
   } else {
     console.log(
-      `MCP server "${name}" added to ${scope} settings. (${transport})`,
+      `MCP server "${name}" added to ${settingsScope} settings. (${transport})`,
```
jackwotherspoon:
```suggestion
      `MCP server "${name}" added to ${scope} settings. (${transport})`,
```

jackwotherspoon (approved):
LGTM, couple small nits I will take care of

jackwotherspoon (approved):

scidomino (approved):

bdmorgan (approved):

PR #7817, opened 2025-09-05
Add initial RFC for A2A development-tool extension

skeshive:
Note: The spec in this RFC is not final. I expect it to evolve as we get community feedback & update our experimental A2A server implementation to match the spec.


bdmorgan (approved):
LGTM

PR #7818, opened 2025-09-05
feat(core): Download ripgrep at runtime, if enabled.

joshualitt:
## TLDR

Downloads ripgrep at runtime, if the feature is enabled.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | :heavy_check_mark:  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Part of the work required to address #7327

### DIFF
--- a/packages/core/src/config/config.test.ts
+++ b/packages/core/src/config/config.test.ts
@@ -687,9 +687,7 @@ describe('setApprovalMode with folder trust', () => {
 
   describe('registerCoreTools', () => {
     beforeEach(() => {
-      (canUseRipgrep as Mock).mockClear();
-      (ToolRegistry.prototype.registerTool as Mock).mockClear();
-      (logRipgrepFallback as Mock).mockClear();
+      vi.clearAllMocks();
     });
 
     it('should register RipGrepTool when useRipgrep is true and it is available', async () => {
--- a/packages/core/src/config/config.test.ts
+++ b/packages/core/src/config/config.test.ts
@@ -25,6 +25,10 @@ import { ClearcutLogger } from '../telemetry/clearcut-logger/clearcut-logger.js'
 
 import { ShellTool } from '../tools/shell.js';
 import { ReadFileTool } from '../tools/read-file.js';
+import { GrepTool } from '../tools/grep.js';
+import { RipGrepTool, canUseRipgrep } from '../tools/ripGrep.js';
+import { logRipgrepFallback } from '../telemetry/loggers.js';
+import { RipgrepFallbackEvent } from '../telemetry/types.js';
 
 vi.mock('fs', async (importOriginal) => {
   const actual = await importOriginal<typeof import('fs')>();
@@ -56,7 +60,11 @@ vi.mock('../utils/memoryDiscovery.js', () => ({
 // Mock individual tools if their constructors are complex or have side effects
 vi.mock('../tools/ls');
 vi.mock('../tools/read-file');
-vi.mock('../tools/grep');
+vi.mock('../tools/grep.js');
+vi.mock('../tools/ripGrep.js', () => ({
+  canUseRipgrep: vi.fn(),
+  RipGrepTool: class MockRipGrepTool {},
+}));
 vi.mock('../tools/glob');
 vi.mock('../tools/edit');
 vi.mock('../tools/shell');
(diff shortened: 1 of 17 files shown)


joshualitt:
@mattKorwel  for changes required to get this through CI.

joshualitt:
@mattKorwel  nevermind! I created https://github.com/google-gemini/gemini-cli/pull/7833 with just the CI changes.

packages/core/src/telemetry/loggers.ts
```diff
@@ -268,6 +269,28 @@ export function logFlashFallback(
...
+  config: Config,
+  event: RipgrepFallbackEvent,
+): void {
+  ClearcutLogger.getInstance(config)?.logRipgrepFallbackEvent();
+  if (!isTelemetrySdkInitialized()) return;
+
+  const attributes: LogAttributes = {
+    ...getCommonAttributes(config),
+    ...event,
+    'event.name': EVENT_API_REQUEST,
```
jacob314: this event name seems odd. Perhaps you need a new event name as none of the existing ones seem like a great fit. 
https://github.com/google-gemini/gemini-cli/blob/c1b8708ef581310fd5861a09cbe40b265b6942bb/docs/telemetry.md?plain=1#L212

packages/core/src/tools/ripGrep.ts
```diff
@@ -8,16 +8,34 @@ import fs from 'node:fs';
...
+
+/**
+ * Checks if `rg` exists, if not then attempt to download it.
+ */
+export async function canUseRipgrep(): Promise<boolean> {
+  if (await fileExists(getRgPath())) {
+    return true;
+  }
+
+  await downloadRipGrep(Storage.getGlobalBinDir());
```
jacob314, line 35: what is the plan to version downloaded versions of ripgrep in the future?

packages/core/src/config/storage.ts
```diff
@@ -57,6 +58,10 @@ export class Storage {
     return path.join(Storage.getGlobalGeminiDir(), TMP_DIR_NAME);
   }
 
+  static getGlobalBinDir(): string {
+    return path.join(Storage.getGlobalGeminiDir(), BIN_DIR_NAME);
```
jacob314: lets reuse the existing tmp directory rather than adding an extra directory as we recently did some work to reduce the number of subdirectories we create to avoid user confusion.

jacob314 (commented):
Looks good. a few minor comments then approved.

packages/core/src/tools/ripGrep.ts
joshualitt: One option will be to encode the version in the file name, and re-download if the cached version is out of date. We could also just run `rg -v` and parse the output.

Did you want that logic in this CL? It does need to be solved, but I think it shouldn't be too much trouble to tackle in a follow-up.

packages/core/src/config/storage.ts
joshualitt: This SGTM, however, it's worth pointing out that `temp` folders tend to be fairly safe to delete, i.e. logs. Whereas, these binaries aren't really "temporary" in that sense, i.e. because if temporary files are cleared at runtime, and our binaries are in there, then the SearchTool will become useless. Of course, it would be safe to delete these binaries at other points in time, we will just re-download them(another sort of odd behavior for something "temporary", it's more like a cache at that point).

Still, I fully understand the desire for consistency, so I'm happy to make this change if these binaries meet the project definition of "temporary." Thoughts?

packages/core/src/telemetry/loggers.ts
joshualitt: Good catch! Funny, I had already added EVENT_RIPGREP_FALLBACK, apparently I forgot to wire it up though. Should be fixed now, ptal.

packages/core/src/tools/ripGrep.ts
jacob314: fine to tackle in a follow up

packages/core/src/config/storage.ts
joshualitt: @jacob314 , updated to move the `bin` dir to `tmp`. Please let me know if you also wanted the getter on `Storage` renamed. We can also tackle additional work in one of the follow-ups. ptal.

packages/core/src/config/config.test.ts
```diff
@@ -685,22 +686,22 @@ describe('setApprovalMode with folder trust', () => {
   });
 
   describe('registerCoreTools', () => {
+    beforeEach(() => {
```
joshualitt, line 689: For reasons I'm not 100% sure of, when I rebased it broke the mocking logic in these tests. I've updated the tests to pass again.

jacob314: can we call vi.resetAllMocks in the beforeEach block instead of this?

jacob314 (approved):
lgtm

PR #7825, opened 2025-09-05
Pull contentGenerator out of GeminiClient and into Config.

scidomino:
## TLDR

Move ContentGenerator from `config.geminiClient.contenGenerator` to `config.contentGenerator`.

## Dive Deeper

Necessary for #4544 which requires that we refreshAuth *before* loading the mcp servers. 

* Moves all non-auth related code out of `Config.refreshAuth()` and into `Config.initialize()`. You still need to call both before getting started but now you can call them in either order.
* `GeminiClient` is now created exactly once in the `Config` constructor and is never reset (instead of a new GeminiClient being built every time we call `refreshAuth()`).
*  Pushes `stripThoughtsFromHistory` code into `GeminiChat` (where the history lives).

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Contributes to #4544

### DIFF
--- a/packages/a2a-server/src/agent/task.ts
+++ b/packages/a2a-server/src/agent/task.ts
@@ -6,7 +6,7 @@
 
 import {
   CoreToolScheduler,
-  GeminiClient,
+  type GeminiClient,
   GeminiEventType,
   ToolConfirmationOutcome,
   ApprovalMode,
@@ -82,7 +82,7 @@ export class Task {
     this.contextId = contextId;
     this.config = config;
     this.scheduler = this.createScheduler();
-    this.geminiClient = new GeminiClient(this.config);
+    this.geminiClient = this.config.getGeminiClient();
     this.pendingToolConfirmationDetails = new Map();
     this.taskState = 'submitted';
     this.eventBus = eventBus;
@@ -227,7 +227,7 @@ export class Task {
     } = {
       coderAgent: coderAgentMessage,
       model: this.config.getModel(),
-      userTier: this.geminiClient.getUserTier(),
+      userTier: this.config.getUserTier(),
     };
 
     if (metadataError) {
--- a/packages/core/src/config/config.ts
+++ b/packages/core/src/config/config.ts
@@ -6,9 +6,13 @@
 
 import * as path from 'node:path';
 import process from 'node:process';
-import type { ContentGeneratorConfig } from '../core/contentGenerator.js';
+import type {
+  ContentGenerator,
+  ContentGeneratorConfig,
+} from '../core/contentGenerator.js';
 import {
   AuthType,
+  createContentGenerator,
   createContentGeneratorConfig,
 } from '../core/contentGenerator.js';
 import { PromptRegistry } from '../prompts/prompt-registry.js';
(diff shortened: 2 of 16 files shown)


scidomino:
The Check Bundle Size error is a preexisting bug that will be fixed by https://github.com/google-gemini/gemini-cli/pull/7833

packages/core/src/config/config.ts
```diff
@@ -410,46 +421,47 @@ export class Config {
...
-    if (existingHistory.length > 0) {
-      this.geminiClient.setHistory(existingHistory, {
-        stripThoughts: fromGenaiToVertex,
-      });
+    // Vertex and Genai have incompatible encryption and sending history with
+    // throughtSignature from Genai to Vertex will fail, we need to strip them
+    if (
+      oldAuthMethod === AuthType.USE_GEMINI &&
+      authMethod === AuthType.LOGIN_WITH_GOOGLE
+    ) {
```
abhipatel12: We switched the order of creating the content generator (prev in the the geminiClient) and stripping the thoughtSignatures. It's very unlikely, but in the future we could cause race conditions here where the client is called before the thought signatures have been stripped properly. 

Wdyt of preparing the chat history first, then creating the config and generator?

packages/core/src/core/geminiChat.ts
```diff
@@ -287,7 +285,7 @@ export class GeminiChat {
           );
         }
 
-        return this.contentGenerator.generateContent(
+        return this.config.getContentGenerator().generateContent(
```
abhipatel12, line 288: Is it possible for this to ever be uninitialized at this point?

packages/a2a-server/src/agent/task.ts
```diff
@@ -227,7 +227,7 @@ export class Task {
     } = {
       coderAgent: coderAgentMessage,
       model: this.config.getModel(),
-      userTier: this.geminiClient.getUserTier(),
+      userTier: this.config.getUserTier(),
```
abhipatel12, line 230: Not on this line, but this file creates its own `geminiClient`. Should we be using the one in the config?

abhipatel12 (approved):
LGTM! Just a couple of questions. Thanks again for doing this! Makes my refactor so much easier ❤️

packages/core/src/config/config.ts
scidomino: Good point. I change the code to strip the thoughts from the history first then change out the content generators. Worst case scenario is that the user's loose thought history unnecessarily which is a much better fail state.

packages/core/src/core/geminiChat.ts
scidomino: I don't think so, but I'm pretty sure the if there is such a situation, that it would have caused problems in the old code too. The error would have just happened when trying to access the non-existent gemini client.

packages/a2a-server/src/agent/task.ts
scidomino: Done.

packages/core/src/core/client.ts
```diff
@@ -135,33 +128,19 @@ export class GeminiClient {
...
-      this.config.getSessionId(),
-    );
+  async initialize() {
     this.chat = await this.startChat();
   }
 
-  getContentGenerator(): ContentGenerator {
-    if (!this.contentGenerator) {
+  private getContentGeneratorOrFail(): ContentGenerator {
+    if (!this.config.getContentGenerator()) {
```
jacob314, line 140: optional nit: add a local variable to avoid calling `config.getContentGenerator()` twice.

jacob314 (approved):
lgtm

Issue #7826, opened 2025-09-05
every instruction involving file writing yields  [API Error: Model stream was invalid or completed without valid content.]


hillct:
### What happened?

This seems to be the case today Sept 5th, whereas I'd never seen this error before in the month of usage. This morning I updated to:
 CLI Version           0.3.2                                     │
│ Git Commit            ad3bc17e                                  │
│ Model                 gemini-2.5-pro                            │
│ Sandbox               no sandbox                                │
│ OS                    darwin                                    │
│ Auth Method           OAuth                                     │
│ IDE Client            VS Code    

My assumption that this issue is CLI-version specific.

### What did you expect to happen?

I expect file write tool use to complete successfully.

### Client information

* **CLI Version:** 0.3.2
* **Git Commit:** ad3bc17e
* **Session ID:** 6f13caa1-f21b-4c94-9639-118cfdeb4cae
* **Operating System:** darwin v24.5.0
* **Sandbox Environment:** no sandbox
* **Model Version:** gemini-2.5-pro
* **Memory Usage:** 621.7 MB
* **IDE Client:** VS Code

### Login information

Google account

Issue #7779, opened 2025-09-05
API error


Kajuks:
### What happened?

after inserting task i am getting this error:
✕ [API Error: Model stream was invalid or completed without valid content.]

### What did you expect to happen?

continue working on the task

### Client information

* **CLI Version:** 0.3.2
* **Git Commit:** ad3bc17e
* **Session ID:** 60a3e0ec-7041-4d03-a9a5-68784127ad32
* **Operating System:** win32 v22.17.1
* **Sandbox Environment:** no sandbox
* **Model Version:** gemini-2.5-flash
* **Memory Usage:** 196.6 MB

### Login information

Google Account

Issue #7747, opened 2025-09-04
✕ [API Error: Model stream was invalid or completed without valid content.]


CallMeChewy:
### What happened?

✕ [API Error: Model stream was invalid or completed without valid content.]

### What did you expect to happen?

✕ [API Error: Model stream was invalid or completed without valid content.]

### Client information

* **CLI Version:** 0.3.1
* **Git Commit:** 2fb14ead
* **Session ID:** a06d091a-9650-4594-8a0e-7a62448d4af7
* **Operating System:** linux v22.12.0
* **Sandbox Environment:** no sandbox
* **Model Version:** gemini-2.5-pro
* **Memory Usage:** 211.7 MB
* **IDE Client:** VS Code

TimoD87:
I'm getting this issue since 2-3 hours after each response of the gemini 2.5 pro api inside the Gemini CLI

PR #7832, opened 2025-09-05
Fix(core): Fix stream validation logic

### LINKED ISSUE #7814: Multiple simultaneous attempts to solve the posed request results in failed output.
### What happened?

It seems like there are multiple sub-agents trying to execute the same task, at least by looking at the repeated response messages, which then causes the resulting output to be garbage (although that is just my assumption).

Image

Let me know if (and how) I can provide any traces that would help.

Just one subagent performs the action and is able to save a valid file.

### Client information

* **CLI Version:** 0.3.2
* **Git Commit:** ad3bc17e
* **Session ID:** 3269ced5-dcc8-4c90-997e-629bc864b31a
* **Operating System:** darwin v24.6.0
* **Memory Usage:** 428.1 MB
* **IDE Client:** VS Code

### Login information

GEMINI_API_KEY

_No response_

SandyTao520:
## TLDR

This pull request fixes a bug in the chat stream handling where an incomplete model response could be accepted if the stream ended with an empty or invalid data chunk. The validation logic is now stricter, requiring a valid `finishReason` to consider a stream complete, which prevents silent failures and ensures data integrity.

## Dive Deeper

The previous stream validation logic was overly complex and had a loophole: if a stream from the model ended with an invalid chunk (e.g., an empty text part) but didn't have a definitive `finishReason`, our code would often accept it as complete. This could result in the last part of a model's response being silently dropped, leading to an incomplete turn in the chat history.

This change refactors the validation logic to be more robust and explicit. A stream is now considered successful only if one of the following is true:
1. It contains a tool call (which can legitimately end without a `finishReason`).
2. Its final data chunk is valid AND it has a `finishReason` (like `STOP` or `MAX_TOKENS`).

This ensures that we no longer accept ambiguous or improperly terminated streams, making the chat interaction more reliable. The accompanying unit tests have been updated to assert this new, stricter behavior.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ✅  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes #7814, #7779, #7747, #7826

### DIFF
--- a/packages/core/src/core/geminiChat.test.ts
+++ b/packages/core/src/core/geminiChat.test.ts
@@ -416,9 +416,9 @@ describe('GeminiChat', () => {
       expect(modelTurn?.parts![0]!.functionCall).toBeDefined();
     });
 
-    it('should succeed if the stream ends with an empty part but has a valid finishReason', async () => {
-      // 1. Mock a stream that ends with an invalid part but has a 'STOP' finish reason.
-      const streamWithValidFinish = (async function* () {
+    it('should fail if the stream ends with an empty part and has no finishReason', async () => {
+      // 1. Mock a stream that ends with an invalid part and has no finish reason.
+      const streamWithNoFinish = (async function* () {
         yield {
           candidates: [
             {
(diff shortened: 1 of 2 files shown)


NTaylorMullen (approved):

Msofwan:
I still got this error every time I asked it to read or write file, and when I ask about the content, it keep saying that it didn't know because the output of the file isn't being delivered

PR #7833, opened 2025-09-05
bug(core): Fix `Check Bundle Size` github action.

joshualitt:
Adds a missing `clean-step` to the `Check Bundle Size` action.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -392,3 +392,4 @@ jobs:
           pattern: './bundle/**/*.{js,sb}'
           minimum-change-threshold: '1000'
           compression: 'none'
+          clean-script: 'clean'


scripts/clean.js
```diff
@@ -26,7 +26,6 @@ const __dirname = dirname(fileURLToPath(import.meta.url));
 const root = join(__dirname, '..');
 
 // remove npm install/build artifacts
-rmSync(join(root, 'node_modules'), { recursive: true, force: true });
```
joshualitt: Maybe this behavior is desirable in some cases, but if so then I think it should be an option, because otherwise we cannot run other scripts after running `clean`(because some of them have dependencies in `node_modules`).

jacob314 (approved):
lgtm

jacob314 (commented):
Revoking my approve for now as this breaks npx and I don't want someone to accidentally land it.

joshualitt:
@jacob314 PR has been updated to continue bundling during `prepare`. This fixes `npx`.

jacob314 (approved):
lgtm

joshualitt:
Okay, I'm not 100% sure why it helps, but the `clean` fixes two cases of erroneous failures in `Check Bundle Size`([1](https://github.com/google-gemini/gemini-cli/pull/7818), [2](https://github.com/google-gemini/gemini-cli/pull/7932)), so unless there are objections, I'm going to land this in a few hours to unblock these PRs.

Issue #7231, opened 2025-08-27
Epic: Tool Confirmation Message Bus


allenhutchison:
This issue tracks the implementation of the Tool Confirmation Message Bus as described in the RFC.

### Summary

This project will implement a new pub/sub message bus to decouple tool confirmation logic from the core tool implementation. This will allow the agent/tools in the `@google-gemini/core` package to request user confirmation for actions (e.g., executing a shell command) from the TUI in the `@google-gemini/cli` package without a direct dependency. This architecture will also introduce a centralized Policy Service to manage tool execution rules, incorporating existing features like 'Always Allow' and non-interactive mode tool permissions from the outset.

### Implementation Plan

- [x] **PR 1: Foundation & Scaffolding**
  - Introduce `packages/core/src/confirmation-bus/` and `packages/core/src/policy/`.
  - Create `MessageBus`, `PolicyEngine`, and all related types.
  - Integrate `PolicyEngine` into `MessageBus`.
  - Instantiate both in the `Session` object.
  - Add this updated RFC. No functional changes to tool behavior yet.

- [x] **PR 2: Core Tool Integration**
  - Modify the shell tool in `packages/core/src/tools/shell.ts` to use the new message bus flow.
  - Include tests mocking both policy decisions and UI responses to validate the logic.

- [x] **PR 3: TUI Integration**
  - Modify `useReactToolScheduler` to handle `ToolConfirmationRequest` and `ToolPolicyRejection` messages.
  - Wire logic to publish confirmation responses and policy updates (for 'Always Allow') back to the bus.

allenhutchison:
## Update: Added MCP Integration Phase

Created #7837 to track the integration of MCP tools with the Tool Confirmation Message Bus. This represents an additional phase of work beyond the initial 3 PRs:

### Updated Implementation Plan:

- [x] **PR 1: Foundation & Scaffolding** (#7835) - ✅ Complete
  - MessageBus and PolicyEngine implementation
  - Integration with Config class
  - Comprehensive unit tests

- [x] **PR 2: Core Tool Integration**
  - Modify shell tool to use message bus
  - Add tests for tool integration

- [x] **PR 3: TUI Integration**  
  - Wire up UI to handle confirmation requests
  - Handle policy updates from UI

- [x] **PR 4: MCP Tool Integration** (#7837) - NEW
  - Extend PolicyEngine for MCP-specific rules
  - Migrate MCP tool confirmation to MessageBus
[...]

PR #7835, opened 2025-09-05
feat(core): Tool Confirmation Message Bus foundation (PR 1 of 3)

allenhutchison:
## Summary
This PR implements the foundation for the Tool Confirmation Message Bus as described in issue #7231. This is the first of three PRs that will fully implement the feature.

## What this PR does
- Introduces a pub/sub `MessageBus` class for decoupled tool confirmation handling
- Implements a `PolicyEngine` with configurable rules for tool execution control
- Integrates both components into the `Config` class
- Adds comprehensive unit tests with 100% coverage

## Key Features
- **Message Bus**: Event-driven publish/subscribe pattern for tool confirmation messages
- **Policy Engine**: Priority-based rules with tool name and args pattern matching
- **Non-interactive mode**: Automatic conversion of ASK_USER to DENY
- **Error handling**: Robust validation and error event emission
- **Type safety**: Full TypeScript support with discriminated unions

## Testing
- ✅ 25 new unit tests covering all functionality
- ✅ All existing tests pass (1964 total)
- ✅ TypeScript compilation passes
- ✅ ESLint checks pass
- ✅ Preflight checks pass

## Next Steps
- **PR 2**: Core Tool Integration - Modify shell tool to use the message bus
- **PR 3**: TUI Integration - Wire up UI to handle confirmation requests

Fixes part of #7231


packages/core/src/policy/policy-engine.ts
```diff
@@ -0,0 +1,98 @@
...
+      return false;
+    }
+
+    // Check args pattern if specified
+    if (rule.argsPattern) {
+      // If rule has an args pattern but tool has no args, no match
+      if (!toolCall.args) {
+        return false;
+      }
+      const argsString = JSON.stringify(toolCall.args);
```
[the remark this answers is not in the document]
allenhutchison: Fixed and added a test to verify that order doesn't affect validation.

packages/core/src/policy/policy-engine.ts
```diff
@@ -0,0 +1,125 @@
...
+    }
+    
+    // Sort object keys and recursively stringify
+    const sortedKeys = Object.keys(obj).sort();
+    const pairs = sortedKeys.map(key => {
+      const value = (obj as Record<string, unknown>)[key];
+      return JSON.stringify(key) + ':' + this.stableStringify(value);
+    });
+    return '{' + pairs.join(',') + '}';
+  }
```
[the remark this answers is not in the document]
allenhutchison: Fixed

packages/core/src/confirmation-bus/message-bus.ts
```diff
@@ -0,0 +1,98 @@
...
+
+  publish(message: Message): void {
+    try {
+      if (!this.isValidMessage(message)) {
+        throw new Error(
+          `Invalid message structure: ${safeJsonStringify(message)}`,
+        );
+      }
+
+      if (message.type === MessageBusType.TOOL_CONFIRMATION_REQUEST) {
```
abhipatel12, line 45: It may be overkill at this point, but what do you think about using a mediator pattern here. So we would have a generic event/message bus that would just handle message delivery but be separated from any business logic. 

Then, we could create a specific tool policy mediator whose entire job is to subscribe to the relevant events, and then reacting and publishing subsequent events. 

The main benefit here is that if/when we expand the scope of the message bus (beyond tool call confirmations), we'll continue to have to add more business logic to handle those different use cases.

packages/core/src/policy/policy-engine.ts
```diff
@@ -0,0 +1,222 @@
...
+  /**
+   * Remove rules for a specific tool.
+   */
+  removeRulesForTool(toolName: string): void {
+    // Remove all rules for the specified tool, not just the first one
+    for (let i = this.rules.length - 1; i >= 0; i--) {
+      if (this.rules[i].toolName === toolName) {
+        this.rules.splice(i, 1);
+      }
+    }
```
abhipatel12: nit: May be easier to read: `this.rules = this.rules.filter(rule => rule.toolName !== toolName);`

packages/core/src/policy/policy-engine.ts
```diff
@@ -0,0 +1,222 @@
...
+   * @example
+   * // toJSON methods are respected:
+   * const obj = {
+   *   sensitive: 'secret',
+   *   toJSON: () => ({ safe: 'data' })
+   * };
+   * stableStringify(obj)
+   * // Returns: '{"safe":"data"}'
+   */
+  private stableStringify(obj: unknown): string {
```
abhipatel12: This seems like a really useful utility even outside this use case. Would it be helpful to move this into its own utlity file?

abhipatel12 (commented):
These changes look great! Really excited about an event bus. Left a couple of comments. LMK what you think!

packages/core/src/confirmation-bus/message-bus.test.ts
```diff
@@ -0,0 +1,226 @@
...
+
+      expect(handler).toHaveBeenCalledWith(message);
+    });
+
+    it('should allow unsubscribing from message types', () => {
+      const handler = vi.fn();
+      messageBus.subscribe(MessageBusType.TOOL_EXECUTION_SUCCESS, handler);
+      messageBus.unsubscribe(MessageBusType.TOOL_EXECUTION_SUCCESS, handler);
+
+      const message = {
```
richieforeman: Nit: Can you add structural types on to these objects?

go/tsjs-style#use-structural-types

richieforeman: (Whole file for all the `message` objects).

packages/core/src/policy/policy-engine.ts
richieforeman: +1

packages/core/src/policy/policy-engine.ts
richieforeman: At minimum, I think we should make this a file level unexported function (it doesn't use `this` and doesn't rely on the class context, it's a static method).

packages/core/src/policy/policy-engine.ts
```diff
@@ -0,0 +1,222 @@
...
+  }
+
+  /**
+   * Get all current rules.
+   */
+  getRules(): readonly PolicyRule[] {
+    return this.rules;
+  }
+
+  private ruleMatches(rule: PolicyRule, toolCall: FunctionCall): boolean {
```
richieforeman: if you heed my comment below w.r.t. `stableString`, then this function ALSO doesn't use the class context and could easily also be a file-level static function.

Opinion: I always like declaring the "least visibility" I can on a thing.  If it can be a static func, I make it static; If it can be private, I make it private; If it can be readonly, I make it readonly, etc, etc, etc.  I think using least visibility helps with mental load of understanding the impact of some bit of code.

packages/core/src/confirmation-bus/types.ts
```diff
@@ -0,0 +1,51 @@
...
+
+export interface ToolPolicyRejection {
+  type: MessageBusType.TOOL_POLICY_REJECTION;
+  toolCall: FunctionCall;
+}
+
+export interface ToolExecutionSuccess {
+  type: MessageBusType.TOOL_EXECUTION_SUCCESS;
+  toolCall: FunctionCall;
+  result: unknown;
```
richieforeman: Could ToolExecutionSuccess be a templated `<T>` type?

```
interface Success<T> {
  result: T
}
````

File: packages/core/src/confirmation-bus/types.ts [hunk 9]
```diff
@@ -0,0 +1,51 @@
...
+}
+
+export interface ToolExecutionFailure {
+  type: MessageBusType.TOOL_EXECUTION_FAILURE;
+  toolCall: FunctionCall;
+  error: Error;
```
richieforeman: Consider making thing a templated type as well with a default.

```
Failure<E = Error> {
  error: E;
}
```

packages/core/src/confirmation-bus/message-bus.ts
```diff
@@ -0,0 +1,98 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { EventEmitter } from 'node:events';
+import type { PolicyEngine } from '../policy/policy-engine.js';
+import { PolicyDecision } from '../policy/types.js';
+import { MessageBusType, type Message } from './types.js';
+import { safeJsonStringify } from '../utils/safeJsonStringify.js';
+
+export class MessageBus extends EventEmitter {
```
richieforeman, line 13: Thinking out loud, I know we want to eliminate dependencies and things, but if we add this generic messageBus we should probably consider whether or not we should use another library for events/messaging.

RXJS is pretty heavy, but has a lot of utilities for dealing with event streams and the like.  It can be a pain to learn, but has a mountain of flexibility.

Signals has also been proposed in TC39 - https://github.com/tc39/proposal-signals . Internally, Wiz and Angular have generalized signal implementations.

While potentially dismissible as overkill, I think we should probably consider solutions for messaging (or eliminate them).

jacob314: Interesting suggestion and this has been on my mind as well. However I'd prefer we go with EventEmitter for now. RXJS could be interesting to consider in the future and I'm not sure our state is yet complex enough for it to be worth it. I would expect that if we decide to migrate, Gemini CLI could make the migration surprisingly easy in the future so I don't think we need to adopt RXJS until we really have cases where the complexity is worth the costs.

packages/core/src/policy/policy-engine.ts
```diff
@@ -0,0 +1,222 @@
...
+    }
+
+    // Check args pattern if specified
+    if (rule.argsPattern) {
+      // If rule has an args pattern but tool has no args, no match
+      if (!toolCall.args) {
+        return false;
+      }
+      // Use stable JSON stringification with sorted keys to ensure consistent matching
+      const argsString = this.stableStringify(toolCall.args);
```
jacob314: should we cache the computation of stableStringify so we don't compute this again for each rule that might match?

packages/core/src/policy/policy-engine.ts
```diff
@@ -0,0 +1,222 @@
...
+        return JSON.stringify(currentObj);
+      }
+
+      // Check for circular reference (object is in ancestor chain)
+      if (ancestors.has(currentObj)) {
+        return '"[Circular]"';
+      }
+
+      // Create new ancestors set for this branch
+      const newAncestors = new Set(ancestors);
```
jacob314: performance nit: can you remove ancestors when leaving a branch rather than creating a new set.

packages/core/src/policy/policy-engine.ts
```diff
@@ -0,0 +1,222 @@
...
+          return stringify(jsonValue, newAncestors);
+        } catch {
+          // If toJSON throws, treat as a regular object
+        }
+      }
+
+      if (Array.isArray(currentObj)) {
+        const items = currentObj.map((item) => {
+          // undefined and functions in arrays become null
+          if (item === undefined || typeof item === 'function') {
```
jacob314: nit: also check for null here to avoid recursing for this primitive case

packages/core/src/policy/policy-engine.ts
jacob314: generally I worry this method is likely 50X slower than the regular stringify method. Might be worth benchmarking this or at least ensuring we don't call it more than once per tool call

packages/core/src/policy/policy-engine.ts
allenhutchison: Agreed. Done.

packages/core/src/policy/policy-engine.ts
allenhutchison: I've moved it out to it's own file.

packages/core/src/confirmation-bus/message-bus.test.ts
allenhutchison: Done

packages/core/src/policy/policy-engine.ts
allenhutchison: Done

packages/core/src/confirmation-bus/types.ts
allenhutchison: Done

packages/core/src/confirmation-bus/types.ts
allenhutchison: Done

packages/core/src/confirmation-bus/message-bus.ts
allenhutchison: I think this is a good direction to go for the next use of the message bus. I'd rather add that once we have a concrete next use case.

abhipatel12: Sounds good!

packages/core/src/policy/policy-engine.ts
abhipatel12: Thanks!

abhipatel12 (approved):
This looks great! LGTM

packages/core/src/confirmation-bus/message-bus.ts
richieforeman: (To be clear this is more "commentary" than an ask. No action is expected here for the interim).

allenhutchison: I read through the RXJS and the TC39. I really appreciate the pointers and the suggestions. Thank you!

packages/core/src/policy/policy-engine.ts
allenhutchison:
```
Running benchmark with 10000 iterations...

--- simpleObject ---
JSON.stringify: 0.879ms
stableStringify: 7.036ms

--- nestedObject ---
JSON.stringify: 1.212ms
stableStringify: 8.83ms

--- arrayObject ---
JSON.stringify: 1.061ms
stableStringify: 11.287ms

--- complexObject ---
JSON.stringify: 1.396ms
stableStringify: 6.883ms
```
Yes slower, not 50x, but have made the changes to ensure it isn't called too often.

packages/core/src/policy/policy-engine.ts
allenhutchison: done

packages/core/src/policy/policy-engine.ts
allenhutchison: done

packages/core/src/policy/policy-engine.ts
allenhutchison: done

jacob314 (approved):
lgtm

PR #7845, opened 2025-09-05
Remove conflicting shell Directory checks

hminooei:
## TLDR

This change addresses two common errors, `Directory <dir> is not a registered workspace directory` and `Directory cannot be absolute. Please refer to workspace directories by their name.`, which accounts for approximately 34%+11% of `run_shell_command` failures as of Sep 8 2025. These errors incorrectly blocks benign shell commands that specify a `Directory` parameter.

## Dive Deeper

The root cause of this issue is a logical conflict in parameter validation for the `run_shell_command` tool. There are two competing conditions:

1.  The `Directory` parameter is not allowed to be an absolute path.
2.  If the `Directory` parameter is a relative path, it is often blocked by a check ensuring it's a "registered workspace directory," which it frequently isn't for simple, valid use cases.

This creates a catch-22 where many legitimate commands that need to specify a working directory fail validation. This change addresses the problem by making the Directory parameter absolute.

## Reviewer Test Plan

To validate this change, you should test shell commands that previously failed with the `Directory ... is not a registered workspace directory` error.

1.  Pull down this branch.
2.  Run gemini cli's main branch and ask it to execute a shell command such as `npm install` inside a subfolder of your root e.g. `frontend`. (since the Directory is an optional parameter this might not happen always). 
    *if the Directory param is set by LLM to frontend, then it will be blocked and you see the error `Directory `fronted` is not a registered workspace directory`.
4.  Exit and run the gemini cli from this branch and verify that the command now executes successfully without throwing the workspace directory error.

## Testing Matrix

|         | 🍏  | 🪟  | 🐧  |
| ------- | --- | --- | --- |
| npm run | 👍   | ❓  | ❓  |
| npx     | ❓  | ❓  | ❓  |
| Docker  | ❓  | ❓  | ❓  |
| Podman  | ❓  | -   | -   |
| Seatbelt| ❓  | -   | -   |

## Linked issues / bugs
[#747](https://github.com/google-gemini/maintainers-gemini-cli/issues/747)


packages/core/src/tools/shell.ts
```diff
@@ -340,6 +337,7 @@ export class ShellTool extends BaseDeclarativeTool<
   private allowlist: Set<string> = new Set();
 
   constructor(private readonly config: Config) {
+    console.log('**************** this branch - TO BE DELETED');
```
anj-s: remove?

packages/core/src/tools/shell.test.ts
```diff
@@ -422,38 +463,3 @@ describe('ShellTool', () => {
...
-    } as unknown as Config;
-    const shellTool = new ShellTool(config);
-    const invocation = shellTool.build({
-      command: 'ls',
-      directory: 'test',
-    });
-    expect(invocation).toBeDefined();
-  });
-
-  it('should throw an error for directory outside workspace', () => {
```
anj-s, line 404: we should still keep this test but modify it i think

anj-s (approved):

anj-s:
@abhipatel12 can you take a look as well?

packages/core/src/tools/shell.ts
abhipatel12: +1 let's remove this or non-interactive will get extra noise!

abhipatel12 (approved):
LGTM as well! Please remove that one console.log!

packages/core/src/tools/shell.ts
hminooei: Ah good catch! removed.

PR #7876, opened 2025-09-06
feat: Add cross-platform clipboard image support

### LINKED ISSUE #5316: Cohesive Image support
Currently Gemini CLI has some image support but it is incomplete and only works on some platforms.
For example, image paste works on Mac (ctrl-V) but is not discoverable and not supported on all platforms.
Users can drag and drop images into the terminal but are not aware of it.
Users lack an easy way to send a screenshot of their entire desktop to gemini cli.
Screenshots and images referenced in the prompt lack polish and show up with confusing urls like
`@.gemini-clipboard/clipboard-1231234134.png` instead of `[screenshot-1]` or something else less confusing.

bniladridas:
Add cross-platform clipboard image support for macOS, Windows, and Linux with security fixes and test improvements.

This continues the work from PR #7831 with consolidated commits.

Closes #5316

### DIFF
--- a/packages/cli/src/ui/components/InputPrompt.test.tsx
+++ b/packages/cli/src/ui/components/InputPrompt.test.tsx
@@ -138,6 +138,22 @@ describe('InputPrompt', () => {
       replaceRange: vi.fn(),
       deleteWordLeft: vi.fn(),
       deleteWordRight: vi.fn(),
+      getLine: vi.fn((row: number) => mockBuffer.lines[row] || ''),
+      getLineCount: vi.fn(() => mockBuffer.lines.length),
+      getLineLength: vi.fn(
+        (row: number) => (mockBuffer.lines[row] || '').length,
+      ),
+      getTextInRange: vi.fn(() => ''),
+      getText: vi.fn(() => mockBuffer.text),
+      getCursor: vi.fn(() => mockBuffer.cursor),
+      setCursor: vi.fn((row: number, col: number) => {
+        mockBuffer.cursor = [row, col];
+      }),
+      getSelection: vi.fn(() => ({
+        start: { row: 0, col: 0 },
+        end: { row: 0, col: 0 },
+        isEmpty: true,
+      })),
     } as unknown as TextBuffer;
 
     mockShellHistory = {
@@ -166,6 +182,9 @@ describe('InputPrompt', () => {
         text: '',
         accept: vi.fn(),
         clear: vi.fn(),
+        isLoading: false,
+        isActive: false,
+        markSelected: vi.fn(),
       },
     };
     mockedUseCommandCompletion.mockReturnValue(mockCommandCompletion);
@@ -374,17 +393,29 @@ describe('InputPrompt', () => {
   describe('clipboard image paste', () => {
     beforeEach(() => {
       vi.mocked(clipboardUtils.clipboardHasImage).mockResolvedValue(false);
-      vi.mocked(clipboardUtils.saveClipboardImage).mockResolvedValue(null);
+      vi.mocked(clipboardUtils.saveClipboardImage).mockResolvedValue({
+        filePath: null,
+        error: 'No image in clipboard',
+      });
       vi.mocked(clipboardUtils.cleanupOldClipboardImages).mockResolvedValue(
         undefined,
       );
     });
 
     it('should handle Ctrl+V when clipboard has an image', async () => {
+      const imagePath = '/test/.gemini-clipboard/clipboard-123.png';
       vi.mocked(clipboardUtils.clipboardHasImage).mockResolvedValue(true);
-      vi.mocked(clipboardUtils.saveClipboardImage).mockResolvedValue(
-        '/test/.gemini-clipboard/clipboard-123.png',
-      );
+      vi.mocked(clipboardUtils.saveClipboardImage).mockResolvedValue({
+        filePath: imagePath,
+        error: undefined,
+      });
+      vi.mocked(clipboardUtils.cleanupOldClipboardImages).mockResolvedValue();
+
+      // Set up buffer state
+      mockBuffer.text = '';
+      mockBuffer.cursor = [0, 0];
+      mockBuffer.lines = [''];
+      mockBuffer.replaceRangeByOffset = vi.fn();
 
       const { stdin, unmount } = renderWithProviders(
         <InputPrompt {...props} />,
@@ -393,50 +424,71 @@ describe('InputPrompt', () => {
 
       // Send Ctrl+V
       stdin.write('\x16'); // Ctrl+V
-      await wait();
 
-      expect(clipboardUtils.clipboardHasImage).toHaveBeenCalled();
+      // Wait for the clipboard operations to complete
+      await waitFor(() => {
+        expect(clipboardUtils.clipboardHasImage).toHaveBeenCalled();
+      });
+
+      // Verify clipboard operations were called correctly
       expect(clipboardUtils.saveClipboardImage).toHaveBeenCalledWith(
         props.config.getTargetDir(),
       );
       expect(clipboardUtils.cleanupOldClipboardImages).toHaveBeenCalledWith(
         props.config.getTargetDir(),
       );
+
+      // Verify buffer was updated with the image path
       expect(mockBuffer.replaceRangeByOffset).toHaveBeenCalled();
       unmount();
     });
 
     it('should not insert anything when clipboard has no image', async () => {
       vi.mocked(clipboardUtils.clipboardHasImage).mockResolvedValue(false);
+      mockBuffer.replaceRangeByOffset = vi.fn();
 
       const { stdin, unmount } = renderWithProviders(
         <InputPrompt {...props} />,
       );
       await wait();
 
-      stdin.write('\x16'); // Ctrl+V
-      await wait();
+      // Send Ctrl+V
+      stdin.write('\x16');
 
-      expect(clipboardUtils.clipboardHasImage).toHaveBeenCalled();
+      // Wait for clipboard check
+      await waitFor(() => {
+        expect(clipboardUtils.clipboardHasImage).toHaveBeenCalled();
+      });
+
+      // Verify no save or buffer modification occurred
       expect(clipboardUtils.saveClipboardImage).not.toHaveBeenCalled();
-      expect(mockBuffer.setText).not.toHaveBeenCalled();
+      expect(mockBuffer.replaceRangeByOffset).not.toHaveBeenCalled();
       unmount();
     });
 
     it('should handle image save failure gracefully', async () => {
       vi.mocked(clipboardUtils.clipboardHasImage).mockResolvedValue(true);
-      vi.mocked(clipboardUtils.saveClipboardImage).mockResolvedValue(null);
+      vi.mocked(clipboardUtils.saveClipboardImage).mockResolvedValue({
+        filePath: null,
+        error: 'Failed to save image',
+      });
+      mockBuffer.replaceRangeByOffset = vi.fn();
 
       const { stdin, unmount } = renderWithProviders(
         <InputPrompt {...props} />,
       );
       await wait();
 
-      stdin.write('\x16'); // Ctrl+V
-      await wait();
+      // Send Ctrl+V
+      stdin.write('\x16');
 
-      expect(clipboardUtils.saveClipboardImage).toHaveBeenCalled();
-      expect(mockBuffer.setText).not.toHaveBeenCalled();
+      // Wait for clipboard operations
+      await waitFor(() => {
+        expect(clipboardUtils.saveClipboardImage).toHaveBeenCalled();
+      });
+
+      // Verify no buffer modification occurred on failure
+      expect(mockBuffer.replaceRangeByOffset).not.toHaveBeenCalled();
       unmount();
     });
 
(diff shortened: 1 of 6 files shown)


bniladridas:
## Test Updates for Cross-Platform Support

I've updated the clipboard tests to properly support the new cross-platform functionality:

### Changes Made:
- **Removed platform-specific test assumptions** that expected clipboard functions to return  on non-macOS platforms
- **Updated tests to work on all platforms** (macOS, Windows, Linux) 
- **Fixed error handling tests** to be consistent across platforms
- **All tests now pass** on any platform with Node.js 20.x, 22.x, 24.x

### Why This Was Needed:
The original tests were written assuming macOS-only clipboard support. With the new cross-platform implementation, these tests would fail in CI on Windows and Linux platforms. The updated tests now properly validate the cross-platform clipboard functionality.

### Test Coverage:
- ✅ Clipboard image detection on all platforms
- ✅ Image saving with sequential numbering  
- ✅ Error handling for invalid paths
- ✅ File cleanup functionality
- ✅ Cross-platform compatibility

The tests are now CI-ready and will pass on all target platforms!

bniladridas:
@allenhutchison , could you please review this PR?

jacob314 (commented):
Really like this functionality. However can you first take a pass to remove changes from this pr that aren't related to cross-platform clipboard support. At a glance there appear to be a lot of unrelated changes. For example files removed from core and docs/mermaid unreleated to the change.

bniladridas:
@jacob314 , Thank you 

I have changed the unrelated files.

packages/cli/src/ui/components/InputPrompt.tsx
```diff
@@ -58,7 +58,7 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
   config,
   slashCommands,
   commandContext,
-  placeholder = '  Type your message or @path/to/file',
+  placeholder = '  Type your message, @path/to/file, or Ctrl+V for clipboard images',
```
jacob314: Fyi @miguelsolorio curious on your thoughts for this placeholder text change.
Potentially we should add a startup tip for using Ctrl+V to paste images from clipboards rather than cluttering the placeholder text.

bniladridas:
GitHub Actions workflow run for PR #7876 all linting, testing, and coverage jobs passed successfully across platforms.

Screenshot 2025-09-10 at 3 14 58 AM

packages/cli/src/ui/utils/clipboardUtils.ts
```diff
@@ -12,101 +12,271 @@ import * as path from 'node:path';
...
+        { shell: '/bin/bash' },
+      );
+      return stdout.trim() === 'true';
+    } catch {
+      return false;
+    }
+  } else if (process.platform === 'win32') {
+    try {
+      // Use PowerShell to check if clipboard contains an image
+      const { stdout } = await execAsync(
```
jacob314, line 149: This doesn't work on my windows machine.
Instead you will need to do 
powershell -Command "[bool](Get-Clipboard -Format Image -ErrorAction Ignore)"

packages/cli/src/ui/utils/clipboardUtils.ts
```diff
@@ -12,101 +12,271 @@ import * as path from 'node:path';
...
+          // Ignore cleanup errors
+        }
+      }
+    } else if (process.platform === 'win32') {
+      // Use PowerShell to save clipboard image
+      const tempFilePath = path.join(tempDir, `clipboard-${counter}.png`);
+      // In PowerShell, a single quote within a single-quoted string is escaped by doubling it.
+      const escapedPath = tempFilePath.replace(/'/g, "''");
+      const powershellCommand = `
+        Add-Type -AssemblyName System.Windows.Forms;
```
jacob314: please also modernize this powershell command. Google Gemini should be able to tell you how to use the more modern syntax.

jacob314 (commented):
Thanks for the PR. This all looks good except for the two cases I commented on.

packages/cli/src/ui/components/InputPrompt.tsx
bniladridas: Thanks @jacob314 for the suggestion. I took the clipboard paste hint out of InputPrompt.tsx (placeholder) and added it into Tips.tsx. It now shows up as tip 5 when no GEMINI.md files exist, or tip 4 otherwise. This way the input stays clean, but the feature is still visible.

jacob314 (commented):
the diff doesn't look right. Please rebase to avoid the unrelated changes in InputPrompt.test.tsx

packages/cli/src/ui/utils/clipboardUtils.ts
bniladridas: current PowerShell approach uses this syntax structure:

```
# Load the System.Drawing assembly
Add-Type -AssemblyName System.Drawing;

# Get image from clipboard
$img = Get-Clipboard -Format Image -ErrorAction Ignore;

# Check if image exists and save it
if ($img) {
  $img.Save('${escapedPath}', [System.Drawing.Imaging.ImageFormat]::Png);
  "success";
} else {
  "no_image";
}
```

bniladridas:
@jacob314 , could you please now see this, looks like I had some issues

packages/cli/src/ui/components/Tips.tsx
bniladridas, line 1: • Dynamic tip numbering: Shows tip "5." when no GEMINI.md files exist, "4." when they do exist
• Clear instruction: Tells users they can paste images from clipboard using Ctrl+V
• User-friendly: Helps users discover the new clipboard image feature

bniladridas:
**Features include:**

• Paste images from clipboard with Ctrl+V
• Security timeout to prevent accidental command execution after paste
• Automatic cleanup of old clipboard images
• Error handling for clipboard operations

jacob314 (commented):
The windows support is broken when I use it but that may be because the security options on my windows machine are locked down too much. The windows scripts also seem to have various escaping issues.
Can you attach a video of how the windows support for paste works for you? Expect you will at minimum need to add clearer error reporting when windows image paste fails as otherwise the failure mode will be confusing for windows users who will only see a cryptic issue in their console.


Here is the error I get running a cleaned up version of the command from the CLI in my powershell terminal:
```
PS C:\Users\jacobr> powershell.exe -Command "Add-Type -AssemblyName System.Drawing; `$img = Get-Clipboard -Format Image -ErrorAction Ignore; if (`$img) { `$img.Save('foo.png', [System.Drawing.Imaging.ImageFormat]::Png); 'success' } else { 'no_image' }"
Add-Type : Cannot add type. Definition of new types is not supported in this language mode.
At line:1 char:1
+ Add-Type -AssemblyName System.Drawing; $img = Get-Clipboard -Format I ...
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : PermissionDenied: (:) [Add-Type], PSNotSupportedException
    + FullyQualifiedErrorId : CannotDefineNewType,Microsoft.PowerShell.Commands.AddTypeCommand
```

jacob314 (commented):
To have confidence we don't regress this it would be great if we could add an integration test. One thought to add an integration test is to trigger the right platform specific calls to add images to the clipboard on each platform and then verify that the image is actually able to be loaded from the clipboard using these apis.

packages/cli/src/ui/utils/clipboardUtils.test.ts
```diff
@@ -13,36 +13,21 @@ import {
...
   describe('clipboardHasImage', () => {
-    it('should return false on non-macOS platforms', async () => {
-      if (process.platform !== 'darwin') {
-        const result = await clipboardHasImage();
-        expect(result).toBe(false);
-      } else {
-        // Skip on macOS as it would require actual clipboard state
-        expect(true).toBe(true);
-      }
+    it('should return boolean on all platforms', async () => {
```
jacob314: given how hard this logic is to get right lets invest in adding some more e2e tests. Good news is now that I think about it adding them shouldn't be that hard. please ask gemini cli to generate test helper methods that copy an example png image to the clipboard on each platform. then we can have these tests exercise the methods to check whether the clipboard has an image and copy the image from the clipboard on each platform. that will make sure we don't start failing if a platform changes their APIs or we introduce a bug in clipboardUtils.ts

jacob314 (commented):
I promise these are the last comments. Really excited about this change but also care about making sure it is robust for users on all platforms.

bniladridas:
The clipboard E2E tests run by default.
If you want to skip them during test runs, set:

```bash
SKIP_E2E_TESTS=true
```

To run only the clipboard E2E tests:

```bash
cd packages/cli
npx vitest run src/ui/__tests__/clipboard.e2e.test.ts
```

Or you can target them by name:

```bash
npm test -- -t "Clipboard E2E Tests"
```

The key change is that the path uses `src/ui/__tests__/...` (not `src/ui/tests/...`) to match the actual directory structure.

packages/cli/src/ui/components/InputPrompt.test.tsx
```diff
@@ -1143,15 +1207,13 @@ describe('InputPrompt', () => {
...
     });
   });
 
-  describe('vim mode', () => {
-    it('should not call buffer.handleInput when vim mode is enabled and vim handles the input', async () => {
-      props.vimModeEnabled = true;
-      props.vimHandleInput = vi.fn().mockReturnValue(true); // Mock that vim handled it.
+  describe('vim handle input', () => {
+    it('should not call buffer.handleInput when vimHandleInput is provided and returns true', async () => {
+      props.vimHandleInput = vi.fn().mockReturnValue(true);
```
jacob314: why was this line deleted. seems unrelated.

packages/cli/src/ui/components/InputPrompt.test.tsx
```diff
@@ -1165,9 +1227,8 @@ describe('InputPrompt', () => {
       unmount();
     });
 
-    it('should call buffer.handleInput when vim mode is enabled but vim does not handle the input', async () => {
-      props.vimModeEnabled = true;
-      props.vimHandleInput = vi.fn().mockReturnValue(false); // Mock that vim did NOT handle it.
+    it('should call buffer.handleInput when vimHandleInput is provided but returns false', async () => {
+      props.vimHandleInput = vi.fn().mockReturnValue(false);
```
jacob314: why was this line deleted. seems unrelated.

packages/cli/src/ui/utils/clipboardUtils.ts
```diff
@@ -8,109 +8,604 @@ import { exec } from 'node:child_process';
...
 import * as fs from 'node:fs/promises';
 import * as path from 'node:path';
+import * as crypto from 'node:crypto';
+
+/**
+ * Interface for paste protection options
+ */
+export interface PasteProtectionOptions {
+  /** Check for sensitive data like API keys, tokens, etc. */
+  checkSensitiveData: boolean;
```
jacob314: please remove the sensitive data checks. that should be a separate PR and will need to be validated carefully to ensure false positives don't impact users.

packages/cli/src/ui/__tests__/clipboard.e2e.test.ts
```diff
@@ -0,0 +1,169 @@
+/**
```
jacob314: can you use the existing naming scheme. existing tests aren't in __tests__ directory. Ask Gemini CLI to rename these to better align with the code base and it should be able to help with that.

packages/cli/src/ui/__tests__/clipboard.e2e.test.ts
```diff
@@ -0,0 +1,169 @@
...
+
+    // Test saveClipboardImage
+    const result = await clipboardUtils.saveClipboardImage(tempDir);
+    expect(result.filePath).toBeTruthy();
+    expect(result.error).toBeUndefined();
+
+    // Verify the file was created
+    if (result.filePath) {
+      const stats = await fs.stat(result.filePath);
+      expect(stats.size).toBeGreaterThan(0);
```
jacob314: can you verify the actual content? that will make these tests more robust.

packages/cli/src/ui/__tests__/clipboard.e2e.test.ts
```diff
@@ -0,0 +1,169 @@
...
+    await ClipboardTestHelpers.copyImageFromFile(testImagePath);
+
+    // First save should work
+    const result1 = await clipboardUtils.saveClipboardImage(tempDir);
+    expect(result1.filePath).toBeTruthy();
+
+    // Try to save again - should be prevented as duplicate
+    const result2 = await clipboardUtils.saveClipboardImage(tempDir);
+
+    // On some platforms, the second save might still work due to timing or clipboard access
```
jacob314: how might we avoid this race condition?

jacob314 (commented):
Thank you for adding e2e tests for this. that is a great improvement. once these comments are addressed I'll approve this commit.

bniladridas:
thank you @jacob314 , your suggestions helped shape these changes.

we fixed the clipboard race condition by adding an isprocessing flag so only one operation can run at a time.

the tests were updated to align with the codebase. instead of keeping them in a separate **tests** folder, they now sit alongside the files they cover, following the existing convention.

* sensitive data checks were removed. 
* vim tests were strengthened by explicitly enabling props.vimmodeenabled = true;

lastly, the clipboard tests now verify actual content. a helper was added to read clipboard text across platforms, and the e2e test confirms the copied text matches what is on the clipboard. this makes the tests more reliable and closer to real usage.

Closes https://github.com/google-gemini/gemini-cli/issues/5316 | Open PR https://github.com/google-gemini/gemini-cli/pull/8557

PR #7883, opened 2025-09-06
feat(cli/input-prompt): paste buffering with placeholders

944750720:
## TLDR
  
  - Add large‑paste placeholders and pending→content mapping to InputPrompt; expand on submit.
  - Unify insert logic via insertAtOffsetWithAutoSpaces; switch offset math to logicalPosToOffset.
  - Make deletion codepoint‑safe (cpLen/cpSlice). Auto‑prune stale pending pastes when text changes.
  - Keep small explicit pastes routed through buffer.handleInput; backspace removes whole placeholder atomically.
  
  ## Dive Deeper
  
  - Large pastes: When an explicit paste exceeds 1000 code points, insert a placeholder like “[Pasted Content N chars]” at the cursor. The mapping {placeholder,content} is stored
  in component state.
  - Submit path: On Enter, expand all placeholders in buffer.text in insertion order, clear the mapping, and submit the expanded text. If the last char before Enter is \, treat
  Enter as newline (soft line break) instead of submit.
  - Codepoint safety: All offset and neighbor checks use codepoint utilities now. Placeholder deletion compares segments with cpLen/cpSlice to avoid UTF‑16 surrogate pitfalls.
  - Insertion ergonomics: insertAtOffsetWithAutoSpaces adds spaces around inserted text (placeholder or @path) if adjacent chars are non‑whitespace. This prevents token run‑ons.
  - Pruning: A new prunePendingPastes runs whenever buffer.text changes to drop mappings whose placeholders were removed by other edits (e.g., kill line, delete word,
  selections).
  - Backspace behavior: If the caret is at the end of a placeholder, backspace deletes the entire placeholder and removes the first matching mapping entry to keep order stable.
  
  ## Reviewer Test Plan
  
  - Large paste placeholder:
      - Paste >1000 code points; verify a “[Pasted Content …]” placeholder appears with spaces adjusted.
      - Press Enter; verify submission text has the placeholder expanded to original content and the mapping cleared.
  - Small paste path:
      - Paste a short string; it should insert inline and behave like regular input.
  - Atomic deletion:
      - Place caret after a placeholder; press Backspace; verify the whole placeholder is removed and mapping shrinks by one matching entry.
  - Pruning:
      - Insert a placeholder; use “kill line”, “delete word”, or select+delete to remove it; type a character; verify pending mapping is auto‑pruned (no stale entries).
  - Unicode safety:
      - Surround placeholders with emoji and CJK; verify spacing checks and deletion do not split surrogate pairs.
  - @path insertion:
      - Paste an image via clipboard; verify an “@relative/path ” is inserted with appropriate spacing at cursor.
  
  ## Testing Matrix
  
  |          | 🍏 | 🪟 | 🐧 |
  | npm run  | ✅  | ❓  | ❓  |
  | npx      | ❓  | ❓  | ❓  |
  | Docker   | ❓  | ❓  | ❓  |
  | Podman   | ❓  | -   | -   |
  | Seatbelt | ✅  | -   | -   |
  
  ## Linked issues / bugs
  
  - Fixed #2627


944750720:
@jacob314 Excited to contribute again! Could you please review this or assign other one to review this?

packages/cli/src/ui/components/InputPrompt.tsx
```diff
@@ -210,37 +249,76 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
...
+          insertAtOffsetWithAutoSpaces(offset, insertText);
         }
       }
     } catch (error) {
       console.error('Error handling clipboard image:', error);
     }
-  }, [buffer, config]);
+  }, [buffer, config, insertAtOffsetWithAutoSpaces]);
+
+  const expandPlaceholders = useCallback(
```
jacob314: as a follow up, please integrate this with the syntax highlighting for the input prompt. that will make it clear these placeholders are special and have been matched. This will make the case where the placeholder is edited less confusing as well as it will switch to plain text when it doesn't match.

packages/cli/src/ui/components/InputPrompt.tsx
```diff
@@ -210,37 +249,76 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
...
+            item.content +
+            out.slice(idx + item.placeholder.length);
+        }
+      }
+      return out;
+    },
+    [pendingPastes],
+  );
+
+  const insertLargePastePlaceholder = useCallback(
```
jacob314: these two calls should be integrated with the reducer rather than left standalone.

packages/cli/src/ui/components/InputPrompt.tsx
```diff
@@ -264,8 +342,13 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
           pasteTimeoutRef.current = null;
         }, 500);
 
-        // Ensure we never accidentally interpret paste as regular input.
-        buffer.handleInput(key);
+        const content = key.sequence || '';
```
jacob314: move all of this logic into text buffer. that will work better to integrate it into the reducer.

packages/cli/src/ui/components/InputPrompt.tsx
```diff
@@ -549,7 +635,11 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
         return;
       }
 
-      // Fall back to the text buffer's default input handling for all other keys
+      if (key.name === 'backspace') {
```
jacob314: move this into the reducer in textBuffer as well.

jacob314 (commented):
thanks for adding! will approve once this logic is moved into text-buffer where it will be easier to integrate it into the reducer and after tests have been added.

packages/cli/src/ui/utils/highlight.ts
```diff
@@ -61,5 +62,61 @@ export function parseInputForHighlighting(
...
-  return tokens;
+  if (placeholders.length === 0) return tokens;
+
+  const nextTokens: HighlightToken[] = [];
+  for (const token of tokens) {
+    if (token.type !== 'default') {
+      nextTokens.push(token);
+      continue;
+    }
+    // Find the segments of the token that match the placeholders
```
jacob314: please simplify this logic. the current approach is complex with sorting and logic to avoid overlap neither of which should be required.
it also requires tests.

A simpler version of this logic is to enforce that placeholders must all begin with `[` and end with `]` and have no other characters in between. Then you should be able to write a single pass trivial parsing algorithm that goes through the string finding all candidate placeholders. Each time it finds a candidate (a block starting with a `[` and ending with a `]` with no `] or `[` in between it can check against the set of all placeholders. if it matches then it is a placeholder.
This will keep the code performant and reduce the risk of bugs. Reuse this logic in a helper method when deciding whether the previous character is part of a place holder in text-buffer.ts. That will ensure there are not any bugs where something shows up as a placeholder in the highlighting but is not deleted like a placeholder. You should also reuse this logic to substitute placeholders. There are some tricky edge cases such as
`@somePath[some placeholder]` that could also be handled inconsistently if we don't have a single consistent solution for what the parse is. for clarity you might want to rename `highlight.ts` to `parse.ts` as it is fundamentally the core way we parse prompts not just what we do for highlighting prompts.

jacob314 (commented):
This is looking good. After there are tests and the requested change to unify logic is made I will approve.

944750720:
@jacob314 I've tried to simplify and extract code to find placeholders. And renamed highlight.ts to parse.ts. Added unittest. Please take a look when convenient.

944750720:
@jacob314 I'm trying to solve conflicts and I found some tests got errors due to code changes last week.
image
image
Should I just ignore them?

944750720:
@jacob314 @abhipatel12 Please take a look when convenient.

packages/cli/src/ui/components/shared/text-buffer.ts
```diff
@@ -19,6 +19,9 @@ import {
 } from '../../utils/textUtils.js';
 import type { VimAction } from './vim-buffer-actions.js';
 import { handleVimAction } from './vim-buffer-actions.js';
+import { findPlaceholderCandidates } from '../../utils/parse.js';
+
+const LARGE_PASTE_CHAR_THRESHOLD = 1000; // When the pasted content is larger than this threshold, it will be inserted as a placeholder
```
jacob314: nit: line exceeds 80 cols. would be ideal to move this to a comment above the variable

packages/cli/src/ui/components/shared/text-buffer.ts
```diff
@@ -1020,7 +1047,94 @@ function textBufferReducerLogic(
...
+          // Delete the corresponding placeholder in pendingPastes
+          const sameTextBefore =
+            placeholderCandidates
+              .slice(0, hitIndex + 1)
+              .filter((ph) => ph.text === hitPlaceholder.text).length - 1;
+          let count = 0;
+          const newPending = [...replaced.pendingPastes];
+          for (let j = 0; j < newPending.length; j++) {
+            if (newPending[j].placeholder === hitPlaceholder.text) {
+              if (count === sameTextBefore) {
```
jacob314, line 1146: don't understand `count === sameTextBefore`? Can you simplify this. Maybe this is correct but is overly complex.

jacob314: in addition, we have prune_pending_pastes as well. Could we just rely on that and reuse a common helper to garbage collect unused pastes there and here.

packages/cli/src/ui/components/shared/text-buffer.ts
```diff
@@ -1984,6 +2112,32 @@ export function useTextBuffer({
     dispatch({ type: 'move_to_offset', payload: { offset } });
   }, []);
 
+  const expandPlaceholders = useCallback(
+    (input: string): string => {
+      let out = input;
+      for (const item of state.pendingPastes) {
```
jacob314, line 2157: there is a linear algorithm for this building up the output string once. Gemini CLI should figure it out if nudged. It will also be robust. currently this also has the issue that if the pastes contained the placeholder text themselves this could go badly wrong. as we are mutating out as we go rather than assembling a single output string. Google should know about it as it is a standard interview question used by tech companies so there are a lot of solutions online.

jacob314 (commented):
Looks good overall. Also tested and it was robust for common cases. Once the expansion method is refactored to make faster and more robust, approved.

packages/cli/src/ui/components/shared/text-buffer.ts
944750720: @jacob314  When we have **multiple identical placeholders with different content**, we need to **precisely identify which specific instance** the user deleted. For example:

```typescript
// User has pasted 3 times with same length (same placeholder text)
pendingPastes = [
  { placeholder: '[Pasted Content 100 chars]', content: 'AAA...(100 A's)' },
  { placeholder: '[Pasted Content 100 chars]', content: 'BBB...(100 B's)' },
  { placeholder: '[Pasted Content 100 chars]', content: 'CCC...(100 C's)' }
]

text = "[Pasted Content 100 chars] [Pasted Content 100 chars] [Pasted Content 100 chars]"
```

If the user backspaces to delete the **middle** placeholder:

```typescript
text = "[Pasted Content 100 chars] [Pasted Content 100 chars]"
// We should keep: [AAA, CCC] ✅
// NOT: [AAA, BBB] ❌
```
So I chose:
```typescript
// sameTextBefore: How many placeholders with the SAME text appear BEFORE the one we're deleting
const sameTextBefore = placeholderCandidates
  .slice(0, hitIndex + 1)  // Include the current one
  .filter((ph) => ph.text === hitPlaceholder.text).length - 1;  // Subtract 1 to get "before"

// Then we find the Nth pendingPaste with this placeholder text and remove it
let count = 0;
for (let j = 0; j < newPending.length; j++) {
  if (newPending[j].placeholder === hitPlaceholder.text) {
    if (count === sameTextBefore) {  // Found the Nth one
      newPending.splice(j, 1);  // Remove it
      break;
    }
    count++;
  }
}
```

PR #7911, opened 2025-09-07
Add support for a /tools schema subcommand

bbiggs:
## TLDR

Add support for `/tools schema` to show parameter schema for built in tools.

## Dive Deeper

The `/mcp` command supports a `/mcp schema` subcommand that shows the full function declaration schema as provided to the model. We add support for `/tools schema` for parity with `/mcp`. Having the full descriptions available helps users understand what the model sees and can provide to the tools, and helps developers understand how their tools can be similar or adequate replacements.

## Reviewer Test Plan

Try it out!

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes #7910


packages/cli/src/ui/commands/toolsCommand.ts
```diff
@@ -58,6 +62,25 @@ export const toolsCommand: SlashCommand = {
         } else {
           message += `  - \u001b[36m${tool.displayName}\u001b[0m\n`;
         }
+
```
scidomino: There seems to be a lot of overlap here with code in mcpCommand.ts Could you extract that out to a common function to avoid duplication?

Issue #7809, opened 2025-09-05
Refactor Utility LLM Calls Out of Client.ts


abhipatel12:
Currently `generateJson` calls are made through `client.ts`. This means that we are appending the core system prompt of the central agent where it is not needed. This can negatively impact utility calls such as edit correction or in the future, routing.

PR #7952, opened 2025-09-08
refactor(core): Introduce LlmUtilityService and promptIdContext

abhipatel12:
## TLDR

This pull request introduces a new `LlmUtilityService` to centralize stateless, utility-focused LLM calls (like JSON generation) and a new `promptIdContext` to consistently propagate prompt IDs.

This refactoring decouples internal tools, like the self-correction mechanism in `smart-edit`, from the main `GeminiClient`, improving modularity and testability. The `promptIdContext` uses `AsyncLocalStorage` to avoid passing the `promptId` through multiple layers of the application, making it implicitly available for logging and telemetry.

## Dive Deeper

Previously, internal components that needed to make LLM calls for utility purposes (e.g., fixing a failed edit) were directly dependent on the `GeminiClient`. This created tight coupling and made it difficult to enforce specific configurations (like `temperature: 0`) for these calls.

This PR introduces two key components to address this:

1.  **`LlmUtilityService`**: A new service dedicated to handling stateless LLM tasks. Its first responsibility is a robust `generateJson` method, which is now used by the `smart-edit` tool's self-correction logic (`FixLLMEditWithInstruction`). This ensures that all utility JSON generation calls are standardized and routed through a single, testable service.

2.  **`promptIdContext`**: A new context provider built on `AsyncLocalStorage`. This allows us to set a unique `promptId` at the beginning of a user request (in `useGeminiStream` for the interactive CLI and `runNonInteractive` for non-interactive mode) and have it be accessible deep within the call stack (like in the `LlmUtilityService`) without needing to pass it down as a parameter. This simplifies function signatures and ensures consistent telemetry.

As part of this refactoring, several tests were added and updated, uncovering and fixing a bug in the `smart-edit` test setup where a key service was not being mocked correctly.

## Reviewer Test Plan

1.  Pull down the branch and run the full test suite to ensure all changes are validated:
    ```bash
    npm run preflight
    ```
    Pay close attention to the new tests: `llmUtilityService.test.ts`, `llm-edit-fixer.test.ts`, and the updated `smart-edit.test.ts`.

2.  Manually test the `replace` tool's self-correction capability. Find a file and construct a prompt that is *slightly* incorrect, which would cause a normal search-and-replace to fail.

    **Example:**
    If a file contains the line `const version = "1.0.0";`, run the following prompt:
    ```
    In [path/to/your/file], replace 'const version  = "1.0.0";' with 'const version = "1.0.1";'
    ```
    (Note the extra space in the `old_string`). The tool should fail the initial attempt, trigger the self-correction mechanism via `LlmUtilityService`, and successfully apply the change.

3.  Verify that standard, correct `replace` operations still work as expected.

4.  (Optional) Enable debug mode and observe the logs to confirm that a consistent `promptId` is being used throughout the lifecycle of a single prompt, from the initial user input to the final tool execution.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  |✅ | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt |✅ | -   | -   |

## Linked issues / bugs

This PR makes progress on #7809

### DIFF
--- a/packages/cli/src/ui/hooks/useGeminiStream.ts
+++ b/packages/cli/src/ui/hooks/useGeminiStream.ts
@@ -33,6 +33,7 @@ import {
   parseAndFormatApiError,
   getCodeAssistServer,
   UserTierId,
+  promptIdContext,
 } from '@google/gemini-cli-core';
 import { type Part, type PartListUnion, FinishReason } from '@google/genai';
 import type {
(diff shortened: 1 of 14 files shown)


packages/core/src/core/llmUtilityService.ts
```diff
@@ -0,0 +1,171 @@
...
+    const suffix = '```';
+    if (text.startsWith(prefix) && text.endsWith(suffix)) {
+      logMalformedJsonResponse(
+        this.config,
+        new MalformedJsonResponseEvent(model),
+      );
+      return text.substring(prefix.length, text.length - suffix.length).trim();
+    }
+    return text;
+  }
```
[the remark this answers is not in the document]
abhipatel12: This is the existing parsing mechanism. We can change this in a future PR.

packages/cli/src/ui/hooks/useGeminiStream.ts
```diff
@@ -705,71 +706,72 @@ export const useGeminiStream = (
...
-
-      try {
-        const stream = geminiClient.sendMessageStream(
-          queryToSend,
-          abortSignal,
-          prompt_id!,
-        );
-        const processingStatus = await processGeminiStreamEvents(
-          stream,
+      return promptIdContext.run({ promptId: prompt_id! }, async () => {
```
jacob314: nit: don't think the bang after prompt_id is needed.

packages/core/src/core/llmUtilityService.ts
```diff
@@ -0,0 +1,171 @@
...
+  /**
+   * A unique ID for the prompt, used for logging/telemetry correlation.
+   */
+  promptId: string;
+}
+
+/**
+ * A service dedicated to stateless, utility-focused LLM calls.
+ */
+export class LlmUtilityService {
```
jacob314: this name is a code smell. Both Utility and Service are a bit filler words.
Similarly, the comment tells me nothing other than that this is a grab of functionality without a cohesive name.

How about LlmJsonService, LlmJsonGenerator, or StructuredDataLlm?

In addition, prior art in the project appears to be to write LLM as LLM rather than Llm so consider changing the case here for consistency or correct the existing cases to be Llm.

packages/core/src/utils/promptIdContext.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { AsyncLocalStorage } from 'node:async_hooks';
+
+interface PromptContext {
```
jacob314: this seems like a low value interface. do you plan to add more things to it? if not, keep it as a string id.

packages/cli/src/ui/hooks/useGeminiStream.ts
abhipatel12: Done, also changed in the 2 other spots.

packages/core/src/utils/promptIdContext.ts
abhipatel12: Sure, no plans right now, we can update if needed.

packages/core/src/core/llmUtilityService.ts
abhipatel12: Done, renamed to `BaseLlmClient`

abhipatel12 (commented):
Thanks for the review. Made the changes, lmk what you think!

jacob314 (approved):
lgtm

PR #7990, opened 2025-09-08
Fix narrow terminal streaming UI issue where loading indicator creates visual artifacts

### LINKED ISSUE #7813: When VS Code's terminal window is narrow, updates are rendered as a stream
### What happened?

1. Open VS Code
2. Make terminal window narrow
3. Interact w/ CLI
4. 🐛 Updates are added as a stream instead of being replaced

https://github.com/user-attachments/assets/670776c3-e11a-4e81-a682-d992cbd0f3aa

Updates are replaced properly

### Client information

<details>

```console
> /about
# paste output here
```

</details>

### Login information

_No response_

_No response_

lifefloating:
### TLDR
Fixed the loading indicator in narrow terminals (< 80 columns) to maintain single-row layout instead of switching to multi-row, preventing the "streaming append" visual bug in VS Code's integrated terminal where old frames weren't properly cleared.

### Dive Deeper
The issue occurred when the terminal width crossed the 80-column threshold during active streaming. The `LoadingIndicator` component would switch from single-row (`flexDirection="row"`) to multi-row (`flexDirection="column"`) layout, but VS Code's terminal wouldn't properly clear previous frames, causing witty loading phrases and timer updates to append as new lines instead of replacing in-place.

The fix ensures the loading indicator always uses a single-row layout with truncation, reserving one column to avoid auto-wrap behavior that triggers the rendering issue. In narrow mode, right content is hidden rather than moved to a separate line to maintain consistent frame height.

### Reviewer Test Plan
1. Open VS Code with integrated terminal
2. Start with terminal width > 80 columns (`tput cols`)
3. Run `gemini` and ask a question that takes seconds to respond
4. During the response (while loading indicator is active), drag the terminal width below 80 columns
5. Verify that the loading indicator stays on a single line and doesn't create "streaming append" artifacts
6. Test the reverse: start narrow and expand to wide during response

### Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |

### Linked issues / bugs
Fixes #7813


lifefloating:
### Environment Setup
- Use VS Code's integrated terminal (not iTerm/native Terminal)
- Start with a wide terminal width ≥ 90 columns. Verify with: `tput cols` (should return ≥ 90)

### Trigger "Responding" State
- Run `gemini` (or `npx -y @google/gemini-cli`)
- Enter any question and press Enter, e.g.: "Explain binary search tree"
- Once the loading indicator appears at the top (spinner + witty phrase + timer), keep it running

### Create Narrow Width Transition During Response
- While the CLI is actively responding, drag the VS Code terminal panel border to gradually shrink the width to < 80 columns (e.g., 70-75)
- Verify with `tput cols` (should return < 80)
- The "wide→narrow" transition during streaming updates is most likely to trigger the issue: when the loading bar switches from single-line to multi-line layout, old frames aren't properly cleared, causing "witty phrases" or timer lines to continuously append like a log stream

packages/cli/src/ui/components/LoadingIndicator.tsx
```diff
@@ -45,36 +46,30 @@ export const LoadingIndicator: React.FC<LoadingIndicatorProps> = ({
...
+                ? '⠏'
+                : ''
+            }
+          />
         </Box>
-        {!isNarrow && <Box flexGrow={1}>{/* Spacer */}</Box>}
+        {primaryText && (
+          <Text color={theme.text.accent} wrap="truncate-end">
+            {primaryText}
+            {cancelAndTimerContent ? ` ${cancelAndTimerContent}` : ''}
```
scidomino, line 62: This changes the color of the cancelAndTimerContent to match the primary text. That's not ideal.

packages/cli/src/ui/components/LoadingIndicator.tsx
```diff
@@ -45,36 +46,30 @@ export const LoadingIndicator: React.FC<LoadingIndicatorProps> = ({
...
+          </Text>
+        )}
+        {!primaryText && cancelAndTimerContent && (
+          <Text color={theme.text.secondary} wrap="truncate-end">
+            {cancelAndTimerContent}
+          </Text>
+        )}
+        {!isNarrow && <Box flexGrow={1}></Box>}
         {!isNarrow && rightContent && <Box>{rightContent}</Box>}
```
scidomino, line 71: So now in the narrow view we just hide the right content entirely? That doesn't seem right.

Issue #7834, opened 2025-09-05
Gemini CLI crashes in VS Code after update


mhfaisalbd:
### What happened?

**Description**

The Gemini CLI is consistently crashing when run inside Visual Studio Code. This issue began occurring immediately after the last update.

**Steps to Reproduce**

1.  Open Visual Studio Code.
2.  Open the integrated terminal.
3.  Run the `gemini` command.

### What did you expect to happen?

**Expected Behavior**

The Gemini CLI should start and run normally without any errors.

**Actual Behavior**

The CLI crashes with the following error message:

```
An unexpected critical error occurred:
TypeError: Cannot read properties of null (reading 'toLowerCase')
    at verifyVSCode (file:///C:/Users/mhfaisal/AppData/Roaming/npm/node_modules/@google/gemini-cli/node_modules/@google/gemini-cli-core/dist/src/ide/detect-ide.js:91:32)
    at detectIde (file:///C:/Users/mhfaisal/AppData/Roaming/npm/node_modules/@google/gemini-cli/node_modules/@google/gemini-cli-core/dist/src/ide/detect-ide.js:102:12)
    at IdeClient.getInstance (file:///C:/Users/mhfaisal/AppData/Roaming/npm/node_modules/@google/gemini-cli/node_modules/@google/gemini-cli-core/dist/src/ide/ide-client.js:59:33)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async Config.initialize (file:///C:/Users/mhfaisal/AppData/Roaming/npm/node_modules/@google/gemini-cli/node_modules/@google/gemini-cli-core/dist/src/config/config.js:254:26)
    at async main (file:///C:/Users/mhfaisal/AppData/Roaming/npm/node_modules/@google/gemini-cli/dist/src/gemini.js:160:5)
```

### Client information

<details>
<summary>Client Information</summary>

  * **Operating System:** Windows 10 Enterprise
  * **Node.js Version:** v22.17.0
  * **npm Version:** 10.9.2
  * **Gemini CLI Version:** 0.3.2
  * **VS Code Version:** 1.103.2

</details>

jacob314:
Fyi @skeshive

mhfaisalbd:
I've been doing some debugging and have a potential lead on the issue.

I added a `console.log` to inspect the `ideProcessInfo` object within the `verifyVSCode` function.

```javascript
function verifyVSCode(ide, ideProcessInfo) {
[...]

PR #8007, opened 2025-09-08
fix(ide): fix crash on windows when null command is returned

skeshive:
## TLDR
Unfortunately I'm unable to repro the bug on my windows machine so I can't actually confirm if this fixes the issue but based on the stack trace from #7834, seems like process-utils was returning a null command. 

## Linked issues / bugs
#7834


bdmorgan (approved):
LGTM

PR #8014, opened 2025-09-08
feat(telemetry): Add content length to tool calls and log tool output truncation event

SandyTao520:
## TLDR

This pull request enhances our telemetry by adding a `contentLength` field to all tool call responses. It also introduces a new logging event that fires whenever a tool's output is truncated, capturing details about the truncation. This provides better visibility into tool usage, especially for debugging issues related to large outputs.

## Dive Deeper

The core changes in this PR are:

1.  **`contentLength` in Tool Responses**: The `ToolCallResponseInfo` interface now includes an optional `contentLength` property. This is populated in `CoreToolScheduler` for successful, errored, and cancelled tool calls, giving us insight into the size of the data being handled.
2.  **Truncation Logging**:
    *   A new `ToolOutputTruncatedEvent` was created to specifically capture when a tool's output is shortened.
    *   A corresponding `logToolOutputTruncated` function has been added to our telemetry loggers, which reports this event to both Clearcut and OTel.
    *   This new logger is called from `CoreToolScheduler` right after the `truncateAndSaveToFile` function successfully truncates and saves the output.

## Reviewer Test Plan

Run evals with telemetry enabled, logs of tool call should include content length and there will be events for tool output truncation.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ✅  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

### DIFF
--- a/packages/core/src/core/nonInteractiveToolExecutor.test.ts
+++ b/packages/core/src/core/nonInteractiveToolExecutor.test.ts
@@ -88,6 +88,10 @@ describe('executeToolCall', () => {
       errorType: undefined,
       outputFile: undefined,
       resultDisplay: 'Success!',
+      contentLength:
+        typeof toolResult.llmContent === 'string'
+          ? toolResult.llmContent.length
+          : undefined,
       responseParts: [
         {
           functionResponse: {
@@ -127,6 +131,7 @@ describe('executeToolCall', () => {
       error: new Error(expectedErrorMessage),
       errorType: ToolErrorType.TOOL_NOT_REGISTERED,
       resultDisplay: expectedErrorMessage,
+      contentLength: expectedErrorMessage.length,
       responseParts: [
         {
           functionResponse: {
@@ -176,6 +181,7 @@ describe('executeToolCall', () => {
         },
       ],
       resultDisplay: 'Invalid parameters',
+      contentLength: 'Invalid parameters'.length,
     });
   });
 
@@ -219,6 +225,7 @@ describe('executeToolCall', () => {
         },
       ],
       resultDisplay: 'Execution failed',
+      contentLength: 'Execution failed'.length,
     });
   });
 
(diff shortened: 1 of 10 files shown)


packages/core/src/telemetry/types.ts
```diff
@@ -522,3 +524,34 @@ export type TelemetryEvent =
   | InvalidChunkEvent
   | ContentRetryEvent
   | ContentRetryFailureEvent;
+
+export class ToolOutputTruncatedEvent implements BaseTelemetryEvent {
```
jerop, line 563: add this new event in the telemetry data reference: https://github.com/google-gemini/gemini-cli/blob/main/docs/telemetry.md#logs-and-metric-reference

jerop (dismissed):
had a chance to test this out in opentelemetry, like producing it to a gcp project?

packages/core/src/telemetry/types.ts
SandyTao520: Done

SandyTao520:
Yes, I ran 2 evals and in gcp I was able to see the newly added contentLength field and the new tool output truncation event.

packages/core/src/core/nonInteractiveToolExecutor.test.ts
```diff
@@ -302,4 +311,56 @@ describe('executeToolCall', () => {
...
+    );
+
+    expect(response.contentLength).toBe(
+      typeof toolResult.llmContent === 'string'
+        ? toolResult.llmContent.length
+        : undefined,
+    );
+  });
+
+  it('should have undefined contentLength for array llmContent with no string parts', async () => {
```
jacob314, line 343: note that you will currently return undefined for llmContent with some string parts and some other parts. is that intended?

jacob314 (approved):

packages/core/src/core/nonInteractiveToolExecutor.test.ts
SandyTao520: yes, it's hard to compute length and truncate mixed-type llmContent. Setting it to undefined for now.

Issue #7808, opened 2025-09-05
Allow specifying a ref and path for Git-installed extensions


gspencergoog:
### What would you like to be added?

For extensions installed via git, it would be really helpful to be able to specify the ref/branch/tag and sub-path for where to install an extension from.

I'd suggest having two arguments to gemini extensions install which allows providing the ref (which can also be a branch or tag), and/or path.

For example, I'd like to be able to do this:

```bash
gemini extensions install --source https://github.com/flutter/ai-guidance.git --ref beta --path gemini-cli/extensions/flutter
```

Of course, `--ref` and `--path` are only valid for Git sourced extensions.

The `.gemini-extension-install.json` would then contain:

```json
{
  "source": "https://github.com/flutter/ai-guidance.git",
  "ref": "beta",
  "path": "gemini-cli/flutter",
  "type": "git"
}
```

Installing it would be a sparse fetch that only includes the path specified, moved to the root of the checkout, which could be the equivalent of:

```bash
git archive --remote=<repo_url> <ref> <path/to/dir> | tar -x
```

Updating the install would fetch the ref HEAD.  If the ref is a branch, it would fetch the latest HEAD, if it is a tag, it will fetch the ref that the tag now points to, and if it's just a fixed ref, it would re-fetch the files at that ref and overwrite them.  before fetching, any existing files in the extension would need to be removed, so that removes would also be supported.

We might have to save a manifest in the `.gemini-extension-install.json` or another file so we know what files to remove, because a user could add their own files, and we don't want to just nuke the directory, do we?  If we're OK with just nuking the extension directory before re-install, then that works as well.

We can't make the extension directory be an actual Git repo because

1) sparse checkout doesn't support checking out the files to a location other than their actual relative location, and
2) even though the checkout is sparse, the whole git repo (at least the latest commit) would have to be synced to the .git, which could be larger than the user is expecting (imagine using a path inside of the Flutter repo for an extension).

(related to I[this issue](https://github.com/google-gemini/gemini-cli/issues/7009#issuecomment-3247306448) )

### Why is this needed?

[...]

Issue #7009, opened 2025-08-25
Design and implement extension auto-updating


chrstnb:
Preliminary proposal:

- Add `gemini extensions link` command for creating symlinks to local folders. This can be used in place of `gemini extensions install --path` to create auto-updating behavior.
- Add update settings in to the extensions folder and allow those to be set both by the extension and by the user via commands 
  - In gemini cli core, check for extension updates after the CLI has started up and make updates in the background. Inform the user of the updates and encourage them to re-start once completed.

evanotero:
We should also add a few UI updates in case the user disables auto-updates:

### `gemini extensions list`
```
another-example-extension (0.0.1) [Update available → 0.0.2]
 Path: /Users/chrstn/.gemini/extensions/another-example-extension
 Source: /Users/chrstn/another-example-extension

example-extension (0.0.1)
 Path: /Users/chrstn/.gemini/extensions/example-extension
 Source: /Users/chrstn/example-extension

Run `gemini extensions update -a` to update all extensions.
Extensions will auto-update after starting the Gemini CLI if {whatever setttings file} has auto-updates are enabled.
```

### `gemini` homepage
#### Updating state:
```
╭────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
| New versions of extensions are available!                                                                  |
├── my-extension@1.0.0 → 1.0.1                                                                               |
├── my-extension@1.0.0 → 1.0.1                                                                               |
├── ...                                                                                                      |
|                                                                                                            |
| Attempting to automatically update now...                                                                  |
╰────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

#### Finished state:
```
Extensions update successful! The new versions will be used on your next run.
```


#### Failed state:
```
Automatic update failed. Please try updating manually using `gemini extensions update -a`.
```

gspencergoog:
For extensions installed via git, it would be really helpful to be able to specify the ref/branch/tag and sub-path for where to install an extension from.

I'm authoring an extension for Gemini CLI, but I don't want to make a completely separate repo to hold it: I'd like to add it to an existing repo.  Without the ability to specify a sub-path, I can't do that.  And my users can't pin the extension at a particular version without being able to specify a ref, branch and/or tag.

I'd suggest having two arguments to `gemini extensions install` which allows providing the ref (which can also be a branch or tag), and/or path.

jakemac53:
@gspencergoog can you file a separate issue for those?

[...]

PR #8018, opened 2025-09-08
Add functionality to check for git extension updates, as well as support for installing a specific ref

jakemac53:
## TLDR

Adds support for installing by specific git refs, and checking for updates to git installed extensions. This allows for instance an extension to do releases by updating a specific ref (branch or tag). So you can install the "stable" branch of an extension for instance, if they maintain a branch or tag pointing to a specific commit for that release channel.

Note that this is done without actually cloning the repo, just comparing remote refs, so we don't actually know the version number. But I think it is a good tradeoff for doing it cheaply.

Note also that if you explicitly install a specific version or hash, it will effectively bypass this update check, but I think that is a feature. You opted into installing at that exact version/hash, we shouldn't bug you about it.

## Reviewer Test Plan

Run the tests

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

First part of #7009
Half of  #7808 (implements ref support but not subpath support)


jakemac53:
As a follow-up to this PR I would like to add a setting to just auto-update in the background, but I think it should be a default disabled setting just because of potential security concerns.

packages/cli/src/commands/extensions/install.ts
```diff
@@ -15,6 +15,8 @@ import { getErrorMessage } from '../../utils/errors.js';
 interface InstallArgs {
   source?: string;
   path?: string;
+  ref?: string;
+  subpath?: string;
```
chrstnb: I think we should pull the subpath logic out for now-- There may be utility here, but right now we want to explicitly discourage people using a monorepo to manage multiple extensions.

packages/cli/src/commands/extensions/install.ts
```diff
@@ -15,6 +15,8 @@ import { getErrorMessage } from '../../utils/errors.js';
 interface InstallArgs {
   source?: string;
   path?: string;
+  ref?: string;
```
chrstnb, line 18: can this be reflected via the source URL instead?

packages/cli/src/utils/extensionUpdateEventEmitter.ts
```diff
@@ -0,0 +1,67 @@
...
+  UpdateAvailable = 'extension-update-available',
+  LogError = 'log-error',
+}
+
+interface ExtensionUpdateEventTypes {
+  [ExtensionUpdateEvent.UpdateAvailable]: [message: string];
+  [ExtensionUpdateEvent.LogError]: [message: string];
+}
+
+// An event emitter that queues it's events until a listener is attached for a
```
chrstnb: uber nit: 'its'

chrstnb:
Thank you so much-- this is overall great.

One overarching thought: what happens if there are, say, 1000 extensions that need to be updated and those events are coming in while the user sending/receiving messages-- will the messages get added in the middle of the conversation?

I think we should add a UI component that can be expanded if needed. We can also prompt users to run 'gemini extensions update --all' to update all update-able extensions if there are multiple.

What if we for updates in the background and then write them to a user-level file (e.g., `~/.gemini/extensions/updates.json`). Then, on next startup, we can load the list of extensions to be updated and show it in an info component as the session starts.

We can also store info about auto-updated extensions in that file.

packages/cli/src/commands/extensions/install.ts
jakemac53: Ack, it does simplify the updating story potentially because it makes tags/releases more predictable.

packages/cli/src/commands/extensions/install.ts
jakemac53: I am not aware of any way supported way to do that, no. In terms of the git API and command line refs are always specified separately (`git fetch <remote> <ref>`).

We could invent our own scheme for this - for instance extract the URI fragment as the ref - but I see multiple problems with that:

- It doesn't play well with the org/repo format
- Technically fragments are valid git URIs, there could theoretically be a real git repository URI with a fragment, even though it would be pretty weird.

jakemac53:
> One overarching thought: what happens if there are, say, 1000 extensions that need to be updated and those events are coming in while the user sending/receiving messages-- will the messages get added in the middle of the conversation?

Yes - the way this is implemented each extension will add its own separate message, and they are added to the regular chat history, so they will appear below the last message in the chat, when the update notification comes in.

jakemac53:
> I think we should add a UI component that can be expanded if needed. We can also prompt users to run 'gemini extensions update --all' to update all update-able extensions if there are multiple.
> 
> What if we for updates in the background and then write them to a user-level file (e.g., `~/.gemini/extensions/updates.json`). Then, on next startup, we can load the list of extensions to be updated and show it in an info component as the session starts.

How do you feel about doing this as a follow up? People should not have a lot of extensions right now, and I think this PR gives a good base to build off of?

I just think that adding the auto-update functionality + `/extensions update` command are a bit higher priority than handling the huge number of extensions case right now, and it should be pretty easy to improve the UX in a followup.

packages/cli/src/commands/extensions/install.ts
chrstnb: ok sounds good to me!

jakemac53:
Updated this to just have the core logic but not do any UX around it, we are planning on a more stateful UX and not something event driven.

jacob314 (approved):
lgtm

PR #8021, opened 2025-09-08
Allow users to type while waiting for MCP servers

### LINKED ISSUE #4544: Performance: Slow startup time (8-12 seconds) for Gemini CLI

scidomino:
## TLDR

Move `config.initialize()` into the React code so users can start typing while waiting for MCP server to load.

## Dive Deeper

We still call `config.initialize()` in `gemini.tsx` when not using interactive mode. This does remove the mcp loading spinner in the non-interactive case but that's probably the right thing to do since non-interactive users probably just want the output and nothing else.

## Reviewer Test Plan

Configure several MCP servers. Verify that you can start typing before they finish loading and that they are submitted once mcp servers loading is done.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | x  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes #4544
Fixes #7334

### DIFF
--- a/packages/cli/src/ui/AppContainer.tsx
+++ b/packages/cli/src/ui/AppContainer.tsx
@@ -134,6 +134,8 @@ export const AppContainer = (props: AppContainerProps) => {
 
   const [userTier, setUserTier] = useState<UserTierId | undefined>(undefined);
 
+  const [isConfigInitialized, setConfigInitialized] = useState(false);
+
   // Auto-accept indicator
   const showAutoAcceptIndicator = useAutoAcceptIndicator({
     config,
@@ -157,16 +159,22 @@ export const AppContainer = (props: AppContainerProps) => {
   const staticExtraHeight = 3;
 
   useEffect(() => {
+    (async () => {
+      // Note: the program will not work if this fails so let errors be
+      // handled by the global catch.
+      await config.initialize();
+      setConfigInitialized(true);
+    })();
     registerCleanup(async () => {
       const ideClient = await IdeClient.getInstance();
       await ideClient.disconnect();
     });
   }, [config]);
 
-  useEffect(() => {
-    const cleanup = setUpdateHandler(historyManager.addItem, setUpdateInfo);
-    return cleanup;
-  }, [historyManager.addItem]);
+  useEffect(
+    () => setUpdateHandler(historyManager.addItem, setUpdateInfo),
+    [historyManager.addItem],
+  );
 
   // Watch for model changes (e.g., from Flash fallback)
   useEffect(() => {
@@ -517,6 +525,7 @@ Logging in with Google... Please restart Gemini CLI to continue.
 
   const { messageQueue, addMessage, clearQueue, getQueuedMessagesText } =
     useMessageQueue({
+      isConfigInitialized,
       streamingState,
       submitQuery,
     });
@@ -612,6 +621,7 @@ Logging in with Google... Please restart Gemini CLI to continue.
   useEffect(() => {
     if (
       initialPrompt &&
+      isConfigInitialized &&
       !initialPromptSubmitted.current &&
       !isAuthenticating &&
       !isAuthDialogOpen &&
@@ -625,6 +635,7 @@ Logging in with Google... Please restart Gemini CLI to continue.
     }
   }, [
     initialPrompt,
+    isConfigInitialized,
     handleFinalSubmit,
     isAuthenticating,
     isAuthDialogOpen,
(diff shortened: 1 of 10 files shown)


packages/cli/src/ui/components/ConfigInitDisplay.tsx
```diff
@@ -0,0 +1,54 @@
...
+    appEvents.on('mcp-servers-discovery-start', onStart);
+    appEvents.on('mcp-server-connected', onChange);
+    appEvents.on('mcp-server-error', onChange);
+
+    return () => {
+      appEvents.off('mcp-servers-discovery-start', onStart);
+      appEvents.off('mcp-server-connected', onChange);
+      appEvents.off('mcp-server-error', onChange);
+    };
+  }, [config]);
```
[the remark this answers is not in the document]
scidomino: reworked code to avoid holding complicated state.

packages/cli/src/ui/AppContainer.tsx
```diff
@@ -153,16 +154,22 @@ export const AppContainer = (props: AppContainerProps) => {
...
+      // Note: the program will not work if this fails so let errors be 
+      // handled by the global catch.
+      await config.initialize();
+      setConfigInitialized(true);
+    })();
     registerCleanup(async () => {
       const ideClient = await IdeClient.getInstance();
       await ideClient.disconnect();
     });
   }, [config]);
```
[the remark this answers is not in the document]
scidomino: This is not something we do elsewhere so I'll leave it as is.

packages/cli/src/ui/components/ConfigInitDisplay.tsx
```diff
@@ -0,0 +1,46 @@
...
+    appEvents.on('mcp-client-update', onChange);
+    return () => {
+      appEvents.off('mcp-client-update', onChange);
+    };
+  }, [config]);
+
+  return (
+    <Box>
+      <Text>
+        <Spinner /> {message}
```
jacob314: use the spinner from Gemini CLI instead. It has better accessibility defaults.

packages/cli/src/ui/components/ConfigInitDisplay.tsx
```diff
@@ -0,0 +1,46 @@
...
+    };
+
+    appEvents.on('mcp-client-update', onChange);
+    return () => {
+      appEvents.off('mcp-client-update', onChange);
+    };
+  }, [config]);
+
+  return (
+    <Box>
```
jacob314: nit: add one line of padding on the top to keep this from showing up right bellow the tips without a linebreak.

jacob314 (commented):
I like how clean the implementation of this was!

packages/cli/src/ui/components/ConfigInitDisplay.tsx
scidomino: I tried to use GeminiRespondingSpinner but it doesn't display a spinner unless we're waiting for gemini to respond (which we aren't here).

jacob314: oh darn.Can you refactor that code to add base spinner class that has the 
isScreenReaderEnabled
checks but doesn't include the logic specific to being in a responding state?

scidomino: I don't think that would add much value since all it does is show text instead of a spinner if screen reader mode is enabled. But we're already showing text that tells the user what's happening regardless.

scidomino: done.

jacob314 (approved):
lgtm

Issue #7405, opened 2025-08-29
Infra: Separate *.test.ts bulding from `npm run build`


richieforeman:
### What would you like to be added?

We should separate `*.test.ts` building from `npm run build` in order to speed up local velocity and not build tests && the binary when we build.  It might make sense to make a `npm run build:tests`

Things to remember:
- Make sure that `npm run build:all` builds tests
- Make sure that Github actions still blocks on test build failures.

### Why is this needed?

Go fast!

mattKorwel:
Talk me through this one. I'm not currently convinced we should do this. What is the time diff?

How long do i not build tests while making changes to main, doesn't that just leave me in a badly broken state, that takes longer to fix?

richieforeman:
I'm concerned that the larger the codebase gets, the harder this will be to separate.

I've also ran into issues where I'm tinkering with the main line code and tests at the same time.  For refactorings, I have to make sure my tests build and compile BEFORE I am able to boot up my code and see if it works.  That feels like a productivity drag.  I've done some medium size refactorings, and found this detail specifically frustrating.  In one case, I refactored something, battled the test compilation and then found out my change didn't "work" during runtime.  In order to keep iterating I had to move the main line code and tests together.

In short, at this stage in the codebase -- I think this is a net productivity gain, rather than a wall clock gain here. (With wall clock time potentially becoming a factor as the code matures).

hritan:
> Talk me through this one. I'm not currently convinced we should do this. What is the time diff?
> 
> How long do i not build tests while making changes to main, doesn't that just leave me in a badly broken state, that takes longer to fix?

Fresh builds:

- With test build - 1m6sec
- Without test build - 55 sec

Incremental builds (cache kicks in): 

- With test build - 11.6 sec
- Without test build - 11.1 sec

mattKorwel:
Ok thanks for the context. this seems like a god nice to have. marking it as p1

PR #8027, opened 2025-09-08
fix: Separate *.test.ts bulding from npm run build

hritan:
## TLDR

This PR separates *.test.ts building from `npm run build` in order to speed up local velocity. 
Creates a new command `npm run build:tests` and tsconfig.test.json for building tests in individual packages.

## Dive Deeper

The core goal of separating the test build from the main application build has been achieved, resulting in a faster `npm run build` for local development while maintaining full type-safety in the CI pipeline.

  This was done through the following key changes:

   1. TypeScript Configuration Split:
       * The default tsconfig.json files were modified to exclude test files (*.test.ts, *.test.tsx). This ensures that a standard `npm run build` only compiles the application code.
       * New tsconfig.test.json files were added to each package. These are configured to include all files (both application and test code) for comprehensive type-checking.

   2. Intelligent Build Scripts:
       * The scripts/build_package.js script now intelligently selects which TypeScript configuration to use. It defaults to the fast, app-only build but will use the tsconfig.test.json if a --tests flag is present.
       * The main scripts/build.js script acts as a router, passing the --tests flag down to the individual packages when `npm run build:tests` is executed.

   3. CI Workflow Optimization:
       * The GitHub Actions workflow (.github/workflows/ci.yml) has been updated to leverage this new system.
       * The lint job continues to use the fast `npm run build.`
       * The test job now specifically runs `npm run build:tests`. This guarantees that all code, including tests, is fully type-checked before the tests are executed, ensuring that any type errors will still fail the CI build.

  In summary, you have successfully decoupled the build processes. The default build is now faster for developers, and the CI pipeline is more
  efficient, running the appropriate build type for each job without sacrificing safety.

## Reviewer Test Plan

1. Run `npm run clean` to clear your existing dist folders.
2. You can run `npm run build`. It will be a faster build and return a smaller dist output.
3. Run `npm run clean` again.
4. You can then run `npm run build:tests`. It will have both ts and test.ts files in the dist folder.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes - https://github.com/google-gemini/gemini-cli/issues/7405


packages/cli/tsconfig.json
```diff
@@ -13,74 +13,6 @@
     "src/**/*.json",
     "./package.json"
   ],
-  "exclude": [
-    "node_modules",
-    "dist",
-    // TODO(5691): Fix type errors and remove excludes.
```
shrutip90, line 19: Keep the TODO?

packages/cli/tsconfig.test.json
```diff
@@ -0,0 +1,80 @@
...
+    "src/**/*.tsx",
+    "src/**/*.json",
+    "**/*.test.ts",
+    "**/*.test.tsx"
+  ],
+  "exclude": [
+    "node_modules",
+    "dist",
+    // TODO(5691): Fix type errors and remove excludes.
+    "src/commands/mcp.test.ts",
```
shrutip90, line 15: The include above also has **/*.test.ts?

packages/cli/tsconfig.json
hritan: So, the TODO is moved to tsconfig.test.json because here we exclude the tests completely and that TODO is specifically for test files.

packages/cli/tsconfig.test.json
hritan: Yep, I think exclude will take precedence over include. So, the compiler will go over all the test files excluding specific test files.

.github/workflows/ci.yml
```diff
@@ -278,7 +278,7 @@ jobs:
 
       - name: 'Build project'
         run: |-
-          npm run build
```
richieforeman, line 281: I think here it might make sense to do a build step and then a build tests step.  I think we want to make sure we're doing both here in the CI flow.

packages/cli/tsconfig.test.json
```diff
@@ -0,0 +1,80 @@
...
+    "src/commands/mcp/add.test.ts",
```
richieforeman, line 16: Naive question. Are these tests here because they don't compile?  I'm confused as to the backstory. Do you know?

hritan: Yeah, so these were already there in existing tsconfig. What I understand is that these tests have type errors and they will fail the linting and type checks during preflight. So, someone has excluded them from building and put a ToDo to fix the type issues later.

.github/workflows/ci.yml
hritan: updated

Issue #4578, opened 2025-07-21
Enable daily/weekly/adhoc eval pipeline for quality regression checks


anj-s:
### What would you like to be added?

We need to be able to run a subset of evals we have right now to check for regressions before releasing weekly/daily. 
There should also be a way for users/devs to check quality regressions on a PR with a BYOQ setup.

### Why is this needed?

Currently we don't have a good way of identifying quality regressions before releases go out.

gsquared94:
Found possible duplicate issues:

- #4084: (#0.9019249467320888)

PR #8028, opened 2025-09-08
feat(actions): add swebench eval harness to github actions

agmsb:
## TLDR

We want to be able to run evaluations to catch regressions on changes using GitHub Actions. This PR adds an action that runs a Docker container with a swe-bench evaluation harness baked into the container image. The purpose of this is to enable adhoc evaluations natively in the Gemini CLI OSS repository.

## Dive Deeper

In this PR, running the GitHub action will run a sample of 30 instances from [Swe-bench Verified](https://www.swebench.com) with gemini-cli, and upon completion of work, will upload the outputs from Gemini-CLI to be scored for pass/fail.

Currently, this action will run gemini-cli from HEAD on main branch; we will need to update the harness to allow for running on the branch it was dispatched from in GitHub Actions immediately after this PR.

The workflow is as follows:

1. Script runs to download swebench data.
2. The eval harness will spin up 10 concurrent workers, and each worker will read the data for one of the 30 samples from Swe-bench Verified.
3. Each worker will clone the working directory into a temp directory, assemble a prompt for Gemini CLI based on the instance metadata, and run the Gemini-CLI via npx in that temp directory with the provided prompt. During this execution, telemetry is collected locally.
4. Upon completion of work, the harness will generate a diff patch from Gemini CLI's work, and will upload these along with instance metadata to GCS.
5. Outside of this workflow, we run a continuous scorer polling GCS for these diffs, running Swe-bench Lite's scoring mechanism, and uploading results back to GCS.
6. The harness will read the results and print out the number of passes and/or failures.
7. Processing scripts will run. A full report of instances, resolution, and tool call/token statistics will be calculated. That report + agent trajectories for each instance will be uploaded to GCS. The user will get a GCS URL to access these artifacts.

## Reviewer Test Plan

1. Go to GitHub Actions.
8. Navigate to `Eval`.
9. Click on `Run workflow`.
10. Select `agmsb/eval-runner` branch. 
11. Let the workflow execute.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

This makes progress on #4578.


agmsb:
Successfully ran eval harness here: https://github.com/google-gemini/gemini-cli/actions/runs/17559259950

anj-s (approved):
Let us wait for Matt to take a look as well.

mattKorwel:
Sorry for some of the basic questions but:

1. is 'poetry' a thing written by us or by swe bench?
1. where does the docker container come from, where is its source?
1. when it says we spin up 10 concurrent agents, is that spinning up 10 docker containers or 10 instances of something in cloud run or else where?
1. where do we see the logs for these runs
1. regarding the continuous scorer: Outside of this workflow, we run a continuous scorer polling GCS for these diffs where is the code for this. Can the scorer just run in this action to score it right here?
1. how do we see the the results of the evals?

mattKorwel (changes requested):
See comment below, I think we need to make this clearer in how its used and what it does and where the code comes from and how we see the results.

agmsb:
> Sorry for some of the basic questions but:
> 
> 1. is 'poetry' a thing written by us or by swe bench?
> 2. where does the docker container come from, where is its source?
> 3. when it says we spin up 10 concurrent agents, is that spinning up 10 docker containers or 10 instances of something in cloud run or else where?
> 4. where do we see the logs for these runs
> 5. regarding the continuous scorer: Outside of this workflow, we run a continuous scorer polling GCS for these diffs where is the code for this. Can the scorer just run in this action to score it right here?
> 6. how do we see the the results of the evals?

Not a problem addressing them one by one below.

>is 'poetry' a thing written by us or by swe bench?

The eval harness is written in Python and uses [Poetry](https://python-poetry.org/) for dependency/package management.

>when it says we spin up 10 concurrent agents, is that spinning up 10 docker containers or 10 instances of something in cloud run or else where?

The harness will spawn 10 concurrent [async.io](http://async.io/) Tasks, but the harness itself will still run in a single container. Each Task will run Gemini CLI with npx, setting up the working directory and passing in the prompt. This will all run on the GitHub Action runner.

>regarding the continuous scorer: Outside of this workflow, we run a continuous scorer polling GCS for these diffs where is the code for this. Can the scorer just run in this action to score it right here?

Yes, I think we'll want to do follow up with this quickly after this PR.

>where do we see the logs for these runs
>how do we see the the results of the evals?

~Currently this action in this PR writes out the resolve/unresolved count to stdout. For run-level summaries in a follow-up PR we will log them to stdout so it can be viewable in the GitHub Action UI. For per-eval instance logs and agent trajectories, in a follow-up PR we will upload these to Google Cloud Storage and write out to stdout a link to the GCS bucket.~ 

EDIT: This is now written to GCS. Users will receive a URL to download a zip with results and per-instance agent trajectories.

agmsb:
This PR has been updated to now return a URL to a GCS bucket with a .zip that contains results, reports, and agent trajectories users can download and review locally. Using workflow dispatch, maintainers can run this from any branch where they are working on changes.

Example job/results: https://github.com/google-gemini/gemini-cli/actions/runs/17749244467/job/50440818354#step:4:81

mattKorwel (approved):
Thanks for the comments and updates. LGTM

.github/workflows/eval.yml
```diff
@@ -3,27 +3,46 @@ name: 'Eval'
...
+  contents: 'read'
+  id-token: 'write'
+  packages: 'read'
+
 jobs:
   eval:
     name: 'Eval'
+    if: >-
+      github.repository == 'google-gemini/gemini-cli'
     runs-on: 'ubuntu-latest'
```
richieforeman, line 20: Does this run as part of CI, or is this just manually run?  If manual, please use `self-hosted`, as we do want to transition to self hosted runners.

richieforeman: Whoops, I'm late. I'll address this as a follow up as part of my own work.

thesues:
@agmsb 
is ghcr.io/google-gemini/gemini-cli-swe-agent-eval  a public image, I want to test gemini-cli on other swebench dataset, But I can not pull this image?

PR #8039, opened 2025-09-08
feat(core): Only summarize long tool output for shell command

SandyTao520:
## TLDR

This change refines the tool output truncation logic to apply *only* to the `ShellTool`. Previously, any tool returning a string could have its output truncated, which was not the intended behavior. This ensures that only shell command output is truncated when it exceeds the configured threshold.

## Dive Deeper

The `CoreToolScheduler` previously had a generic mechanism to truncate tool output if the content was a string and a truncation threshold was set. This logic was too broad and could inadvertently affect tools that return legitimate, long strings that shouldn't be truncated.

This PR introduces a simple check (`toolName === ShellTool.Name`) to ensure that this truncation and "save to file" behavior is scoped exclusively to the output of shell commands. This makes the behavior more predictable and prevents unintended side effects on other tools.

## Reviewer Test Plan

To validate this change, you can test the following scenarios:

1.  **Verify Shell Truncation:** Run a shell command that produces a large amount of output.
    - Example prompt: `run the command ls -lR /`
    - **Expected behavior:** The output in the chat should be truncated, and a message should appear indicating that the full output has been saved to a file (e.g., `... (full output saved to /tmp/gemini-cli-output-....txt)`).

2.  **Verify Other Tools are Unaffected:** If another tool returns a long string, it should not be truncated by this mechanism. (While no current tool is directly affected, this change future-proofs the system). You can confirm that the `ShellTool` is the only one with this specific truncation logic applied in `coreToolScheduler.ts`.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ✅  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


anj-s (approved):

adlternative:
@SandyTao520 Why is truncation limited to shell tool only? Many tools can potentially have long outputs, such as mcp, or read_many_files if it accidentally reads too many files. Shouldn't this kind of truncation be applied to them as well?

PR #8042, opened 2025-09-08
Reduce margin on narrow screens, flow the footer contents

### LINKED ISSUE #7939: Footer wraps when < 88 characters wide
### What happened?

The footer display (directory, sandbox, and model status) wraps to a stacked 3-row display when the window size is < 88 characters wide.

I think this behaviour is unexpected. The code for detecting a narrow window looks for < 80 characters, but because there is an overall sizable right-hand-size margin (10%), narrow width mode is triggered even for windows of standard 80x24 size. The footer display should optimize for standard-sized windows where possible.

### Client information

<details>

```console
> /about
╭──────────────────────────────────────────────╮
│                                              │
│ About Gemini CLI                             │
│                                              │
│ CLI Version    0.3.1-git.20d0de74            │
│ Git Commit     20d0de74                      │
│ Model          gemini-2.5-pro                │
│ Sandbox        no sandbox                    │
│ OS             linux                         │
│ Auth Method    gemini-api-key                │
│                                              │
╰──────────────────────────────────────────────╯
```

</details>

### Login information

_No response_

_No response_

bbiggs:
## TLDR

Ensure the footer information does not wrap to three lines on 80x24 terminals.

## Dive Deeper

This change reduces the RHS margin in Gemini CLI as windows become smaller, and adapts the footer display to try harder to keep the three segments on a single line even at small window sizes. This change attempts to have a tigher view for Gemini CLI even when running at standard small terminal sizes of 80x24 or 80x50

## Reviewer Test Plan

Please try this out in different sandboxing configs and directories with ridiculous sizes ;)

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes #7939


packages/cli/src/ui/App.tsx
```diff
@@ -13,16 +13,33 @@ import { Composer } from './components/Composer.js';
 import { useUIState } from './contexts/UIStateContext.js';
 import { QuittingDisplay } from './components/QuittingDisplay.js';
 
+const getContainerWidth = (terminalWidth: number): string => {
+  if (terminalWidth <= 80) {
```
jacob314, line 17: btw only reason we are less than 100% was because of ink rendering bugs in the past.

packages/cli/src/ui/App.tsx
```diff
@@ -13,16 +13,33 @@ import { Composer } from './components/Composer.js';
...
+    return '98%';
+  }
+  if (terminalWidth >= 132) {
+    return '90%';
+  }
+
+  // Linearly interpolate between 80 columns (98%) and 132 columns (90%).
+  const slope = (90 - 98) / (132 - 80); // -0.1538...
```
jacob314: remove -0.1538... comment.

jacob314: nit: add a lerp method to a utils.ts file in cli/src
rather than implementing this inline here.

packages/cli/src/ui/App.tsx
```diff
@@ -13,16 +13,33 @@ import { Composer } from './components/Composer.js';
...
+  const percentage = 98 + slope * (terminalWidth - 80);
+
+  return `${Math.round(percentage)}%`;
+};
+
 export const App = () => {
   const uiState = useUIState();
+  const { stdout } = useStdout();
```
jacob314: use 
  const { container: terminalWidth} = useTerminalSize();

instead.

jacob314 (commented):
Approved once these comments are addressed.

packages/cli/src/ui/App.tsx
bbiggs: done

packages/cli/src/ui/App.tsx
bbiggs: done

bbiggs: done

packages/cli/src/ui/App.tsx
```diff
@@ -5,23 +5,46 @@
...
+import { useTerminalSize } from './hooks/useTerminalSize.js';
+import { lerp } from '../utils/math.js';
 import { useUIState } from './contexts/UIStateContext.js';
 import { StreamingContext } from './contexts/StreamingContext.js';
 import { QuittingDisplay } from './components/QuittingDisplay.js';
 import { ScreenReaderAppLayout } from './layouts/ScreenReaderAppLayout.js';
 import { DefaultAppLayout } from './layouts/DefaultAppLayout.js';
 
+const getContainerWidth = (terminalWidth: number): string => {
+  if (terminalWidth <= 80) {
```
jacob314, line 17: bonus points: make the lerp method be clamped and then you can remove the checks for width <=80 and >= 132.

For a clamped lerp you clamp to be between 0 and 1 at the start of the lerp method

jacob314 (approved):
lgtm

PR #8045, opened 2025-09-09
fix(settings): Use schema defaults and only save overrides

galdawave:
## TLDR

This PR refactors the settings system to only save non-default values to user and workspace `settings.json` files. This keeps configuration files clean by only storing explicit user overrides, making them easier to manage.

## Dive Deeper

This change introduces a more robust settings architecture by making the system fully aware of the default values defined in the `settingsSchema.ts`.

1.  **Dynamic Default Generation**: A new `getDefaultSettings` function programmatically generates a complete default settings object directly from the schema. This ensures our defaults are always in sync with the schema definition.
2.  **Improved Merge Logic**: The settings merge process now uses these generated defaults as the foundational layer. This guarantees that the application always operates with a full set of predictable default values, even if user or workspace settings files are missing or empty.
3.  **Intelligent Saving**: The `saveSettings` function has been enhanced to perform a diff against the default settings before writing to a file. It filters out any setting that matches its default value, ensuring that only explicitly overridden settings are persisted.

This approach significantly cleans up the `settings.json` files and makes the entire configuration system more predictable and maintainable.

## Reviewer Test Plan

To validate these changes, you can test the settings save logic:

1.  Pull the branch and run `npm install`.
2.  For a clean test, temporarily delete your user and workspace settings files (`~/.gemini/settings.json` and `.gemini/settings.json`).
3.  Run the CLI and open the settings dialog with `:set`.
4.  Change a setting from its default value. A good example is toggling **`general.vimMode`** to `true`.
5.  Save the settings for the "User" (global) scope.
6.  Inspect the content of `~/.gemini/settings.json`. It should now only contain the settings that differ from the default (e.g., `{"vimMode": true}`).
7.  Go back to the settings dialog and change `general.vimMode` back to `false` (its default value).
8.  Save the settings again.
9.  Inspect `~/.gemini/settings.json`. The file should now contain an empty object `{}` because the setting has been returned to its default state and is no longer written to the file.
10. Repeat this process for a workspace-level setting to ensure the behavior is consistent.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ❓ | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


packages/cli/src/config/settings.ts
```diff
@@ -675,9 +689,80 @@ export function loadSettings(
...
+ * Deeply compares two objects for equality.
+ * @param a The first object.
+ * @param b The second object.
+ * @returns True if the objects are equal, false otherwise.
+ */
+function isEqual(a: unknown, b: unknown): boolean {
+  if (a === b) return true;
+
+  if (a && b && typeof a === 'object' && typeof b === 'object') {
+    if (a.constructor !== b.constructor) return false;
```
jacob314: can we use
https://nodejs.org/api/util.html#utilisdeepstrictequalval1-val2
instead of rolling our own method?

packages/cli/src/config/settings.test.ts
```diff
@@ -159,7 +162,85 @@ describe('Settings Loading and Merging', () => {
       expect(settings.user.settings).toEqual({});
       expect(settings.workspace.settings).toEqual({});
       expect(settings.merged).toEqual({
-        ...systemSettingsContent,
```
jacob314: these test values are going to churn as settings are added. Can we modify them to use a restricted list of settings to avoid churn?

packages/cli/src/config/settings.test.ts
```diff
@@ -345,18 +655,6 @@ describe('Settings Loading and Merging', () => {
       const settings = loadSettings(MOCK_WORKSPACE_DIR);
 
       expect(settings.merged).toEqual({
-        ui: {
```
jacob314: why is the theme gone?

packages/cli/src/config/settings.ts
```diff
@@ -331,13 +335,15 @@ function mergeSettings(
 
   // Settings are merged with the following precedence (last one wins for
   // single values):
-  // 1. System Defaults
-  // 2. User Settings
-  // 3. Workspace Settings
-  // 4. System Settings (as overrides)
+  // 1. Programmatic Defaults
```
jacob314: Programmatic Defaults --> SchemaDefaults?

packages/cli/src/config/settings.ts
galdawave: This is no longer needed.

packages/cli/src/config/settings.test.ts
galdawave: Theme is back in the corrected tests

PR #8052, opened 2025-09-09
feat(ui): Add FPS counter display and settings option

### LINKED ISSUE #8050: Currently Gemini Cli use ink library which runs by default on 30 Fps with new update

The ability to run Gemini cli on 60 fps

### Why is this needed?

Will make it feel gemini cli is faster and smoother

### Additional context

_No response_

ali-aljufairi:
## TLDR

This PR to check is it worth it to 60fps to gemini cli and how will affect  Render

## Dive Deeper
This 

## Reviewer Test Plan
enable FPS counter from SettingDialogue  and check if FPS logic is correct

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | x  | ❓  | ❓  |
| npx      | x | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs
Resolves #8050

### DIFF
--- a/packages/cli/src/config/settingsSchema.ts
+++ b/packages/cli/src/config/settingsSchema.ts
@@ -318,6 +318,15 @@ const SETTINGS_SCHEMA = {
         description: 'Show line numbers in the chat.',
         showInDialog: true,
       },
+      showFPS: {
+        type: 'boolean',
+        label: 'Show FPS Counter',
+        category: 'UI',
+        requiresRestart: false,
+        default: false,
+        description: 'Display an approximate render FPS in the footer.',
+        showInDialog: true,
+      },
       showCitations: {
         type: 'boolean',
         label: 'Show Citations',
--- a/packages/cli/src/ui/components/FpsDisplay.tsx
+++ b/packages/cli/src/ui/components/FpsDisplay.tsx
@@ -1,0 +1,51 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import type React from 'react';
+import { useEffect, useRef, useState } from 'react';
+import { Box, Text } from 'ink';
+import { Colors } from '../colors.js';
+
+/**
+ * Displays an approximate FPS (renders per second) for the Ink UI.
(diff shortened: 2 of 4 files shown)


packages/cli/src/config/settingsSchema.ts
```diff
@@ -318,6 +318,15 @@ const SETTINGS_SCHEMA = {
...
         showInDialog: true,
       },
+      showFPS: {
+        type: 'boolean',
+        label: 'Show FPS Counter',
+        category: 'UI',
+        requiresRestart: false,
+        default: false,
+        description: 'Display an approximate render FPS in the footer.',
+        showInDialog: true,
```
jacob314, line 328: I would omit this from the dialog.

packages/cli/src/ui/components/FpsDisplay.tsx
```diff
@@ -0,0 +1,51 @@
...
+  });
+
+  // Update FPS once per second
+  useEffect(() => {
+    const id = setInterval(() => {
+      const total = renderCountRef.current;
+      const diff = total - lastCountRef.current;
+      lastCountRef.current = total;
+      // Subtract 1 to account for this update's render, but ensure minimum 0
+      setFps(Math.max(0, diff - 1));
```
jacob314, line 33: not sure this is the right way to measure this but perhaps it is the idiomatic practice in React. Think tracking this correctly would probably be a function of looking at the internals of the ink  renderer. this number will be bellow 60 just because the the UI isn't continually refreshing. think you would really want to know about what the latency is between ui refreshes being triggered and renders occurring.

JayadityaGit:
If this FPS setting is part of the UI in settings.json, I’d recommend including it in the migration file. Otherwise, it might get lost whenever a setting changes. For example, the customWittyPhrases feature also belongs to the UI, and when I updated the theme, all my phrases were deleted because I hadn’t added them to the migration schema.

jacob314:
Fyi @galz10 this FPS counter is only functioning correctly because there is a React state updating bug that Gal is tracking down fixing.

PR #8053, opened 2025-09-09
fix: Restore footer configuration settings (#8041)

Lyonk71:
## Summary

This PR fixes a regression where footer configuration settings stopped working after the AppContainer refactor. Users can now properly hide individual footer elements (CWD, sandbox status, model info) via the settings dialog.

## Reviewer Test Plan

1. Start the CLI with npm run dev
2. Type /settings and navigate to UI → Footer
3. Test each setting individually:
    - Hide CWD: Toggle true → Current working directory path should disappear from bottom left
    - Hide Sandbox Status: Toggle true → Sandbox/trust status should disappear from bottom center
    - Hide Model Info: Toggle true → Model name and token usage should disappear from bottom right
4. Test Hide Context Summary above input prompt (not in Footer section) → Context summary above input should disappear
5. Exit and restart the CLI to verify settings persist
6. Enable all footer hiding options at once to verify they work together

## Testing Matrix

  |          | 🍏  | 🪟  | 🐧  |

  Linked issues / bugs

  Fixes #8041


miguelsolorio:
Thanks for taking a look at this! It looks like we're missing the [alignment tweaks](https://github.com/google-gemini/gemini-cli/pull/7419/files#diff-626f523c46f93e17c4a376de8bab68d5ba06a0baddc7c90e15183cf5a82a7d78R69-R73) that allowed the footer items to flex around depending on which ones we're enabled/disabled. Do you mind adding those?

Lyonk71:
Ahh I see that alignment work now. Originally, I just wired up the missing props from the post refactor. Let me know if you need anything else!

miguelsolorio (approved):
Thanks for making the updates, this fixes it for me! 🙏

miguelsolorio:
One more item to add, there were some alignment tweaks to App.tsx for when context summary is hidden https://github.com/google-gemini/gemini-cli/pull/7419/files#diff-285f405b51ffd9d3a1ebad733da12635608980d6540626a7885fe8aa2fe767d8R1254-R1256

Lyonk71:
Ahh yes, I remember moving those alignment tweaks from App.tsx to Composer as part of the App.tsx cleanup effort. Post refactor Composer has a top margin of 1 to maintain spacing above the prompt. The functionality missing from the refactor was the right justification for actions like prompt when context summary is hidden. That's just a 1 line fix in the commit I just pushed. The implementation is a bit different, but I think easier to understand/maintain. Let me know if this works for you!

jacob314 (approved):
Approving to unblock. Can we add a test to catch if we break this filtering in the future. Can be a golden snapshot test. That will ensure we don't accidentally regress this functionality again in the future.

PR #8078, opened 2025-09-09
feat: Tool Integration with PolicyEngine (PR 2 of #7231)

allenhutchison:
## Summary
This PR implements the second phase of the Tool Confirmation Message Bus architecture (#7231) by integrating the PolicyEngine with an actual tool.

- Integrates PolicyEngine with the web-search tool as a proof of concept
- Implements `shouldConfirmExecute` method that respects policy decisions
- Adds comprehensive test coverage for all policy scenarios

## Implementation Details
The web-search tool now:
- Checks PolicyEngine for tool execution decisions
- Handles ALLOW (immediate execution), DENY (block with error), and ASK_USER (request confirmation)
- Provides a working example of the integration pattern

## Test Coverage
Added tests that mock:
- Policy decisions (ALLOW/DENY/ASK_USER)
- UI confirmation responses
- Different query patterns triggering different policies

## Future Work
As noted, as we integrate this functionality into other tools, we will:
- Extract common confirmation logic to the base tool class
- Generalize the test patterns for reuse across all tools
- Apply this pattern to more critical tools like shell

This PR intentionally keeps the implementation in the specific tool to demonstrate the pattern clearly before abstraction.

## Related
- Part of #7231: Tool Confirmation Message Bus
- Builds on PR #7835 (PR 1: Foundation & Scaffolding)
- Next: PR 3 will extend this to shell and other critical tools


packages/core/src/tools/web-search.ts
```diff
@@ -71,6 +72,46 @@ class WebSearchToolInvocation extends BaseToolInvocation<
...
+      title: 'Confirm Web Search',
+      prompt: `Allow web search for: "${this.params.query}"?`,
+      onConfirm: async (outcome: ToolConfirmationOutcome) => {
+        // In the future, we might want to update rules based on the outcome
+        // For now, just log the decision
+        if (outcome === ToolConfirmationOutcome.ProceedAlways) {
+          // Could add a rule to always allow web searches with similar patterns
+          console.log(`User chose to always allow web searches like: "${this.params.query}"`);
+        }
+      },
```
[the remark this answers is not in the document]
jacob314: agreed. generally remove this log before landing. If you want it, tweak this to console.debug so it is only shown when the app is run in debug mode. console.log will show for all users if they open the console.

packages/core/src/tools/web-search.ts
```diff
@@ -71,6 +72,46 @@ class WebSearchToolInvocation extends BaseToolInvocation<
...
+    
+    // First check the policy engine directly for immediate decisions
+    const decision = policyEngine.check(toolCall);
+    
+    if (decision === PolicyDecision.ALLOW) {
+      return false; // No confirmation needed, proceed
+    }
+    
+    if (decision === PolicyDecision.DENY) {
+      throw new Error(`Web search blocked by policy for query: "${this.params.query}"`);
```
jacob314, line 95: optional nit for discussion: don't love that this is throwing an exception. all things being equal Gemini CLI should not throw exceptions as part of normal operations (improves experience debugging with pause on caught exceptions). Fine with leaving as is but would be ideal if instead  we could report this in the return value rather throwing errors to indicate rejections. Having rejections be throws of generic errors also increases the risk that internal bugs causing errors are treated the same as intentional policy rejections.

packages/core/src/tools/web-search.ts
```diff
@@ -71,6 +72,46 @@ class WebSearchToolInvocation extends BaseToolInvocation<
...
+    return {
+      type: 'info',
+      title: 'Confirm Web Search',
+      prompt: `Allow web search for: "${this.params.query}"?`,
+      onConfirm: async (outcome: ToolConfirmationOutcome) => {
+        // In the future, we might want to update rules based on the outcome
+        // For now, just log the decision
+        if (outcome === ToolConfirmationOutcome.ProceedAlways) {
+          // Could add a rule to always allow web searches with similar patterns
+          console.log(`User chose to always allow web searches like: "${this.params.query}"`);
```
jacob314, line 109: same comment about avoiding console.log

packages/core/src/tools/web-search.ts
```diff
@@ -71,6 +72,46 @@ class WebSearchToolInvocation extends BaseToolInvocation<
...
+    
+    // For ASK_USER, we need to return confirmation details
+    // The actual UI will handle this through the MessageBus
+    return {
+      type: 'info',
+      title: 'Confirm Web Search',
+      prompt: `Allow web search for: "${this.params.query}"?`,
+      onConfirm: async (outcome: ToolConfirmationOutcome) => {
+        // In the future, we might want to update rules based on the outcome
+        // For now, just log the decision
```
abhipatel12, line 106: Just wondering.

For `Always Allow`, why aren't we updating the policy engine so that during the next `web_fetch`, it's run automatically? Why in the future?

packages/core/src/tools/web-search.ts
```diff
@@ -71,6 +72,46 @@ class WebSearchToolInvocation extends BaseToolInvocation<
...
+    if (decision === PolicyDecision.DENY) {
+      throw new Error(`Web search blocked by policy for query: "${this.params.query}"`);
+    }
+    
+    // For ASK_USER, we need to return confirmation details
+    // The actual UI will handle this through the MessageBus
+    return {
+      type: 'info',
+      title: 'Confirm Web Search',
+      prompt: `Allow web search for: "${this.params.query}"?`,
```
abhipatel12, line 103: Non-blocking on this PR and purely for discussion, I think it would be ideal if we could eventually transition our permission request to be event-driven as well. 

In that world, we can separate concerns between the core and cli. For example, the `title` and `prompt` above, imo, seem like a CLI concern. In the core we can focus on just providing the base data required and the CLI can decide how it wants to render that information.

abhipatel12 (commented):
Looks great in general! Just had a few questions!

PR #8091, opened 2025-09-09
fix(ide companion extension): Don't show the installation confirmation message in Firebase Studio

romannurik:
## TLDR

Don't show the installation confirmation message in Firebase Studio. This is because we preinstall the CLI companion in many new workspace templates now, and the user immediately sees the message on first run of the workspace.

## Dive Deeper

Checks an environment variable set in Firebase Studio (already available in the core package as `detectIdeFromEnv`) and conditionally disables the popup message.

## Reviewer Test Plan

1. Outside Firebase Studio (e.g. in VSCode, etc) nothing should be different
2. When installing the extension in Firebase Studio, you should NOT see the "Gemini CLI Companion extension successfully installed." message.

## Testing Matrix

N/A - this is for the VSCode companion extension

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

N/A

### DIFF
--- a/packages/vscode-ide-companion/src/extension.ts
+++ b/packages/vscode-ide-companion/src/extension.ts
@@ -9,6 +9,7 @@ import { IDEServer } from './ide-server.js';
 import semver from 'semver';
 import { DiffContentProvider, DiffManager } from './diff-manager.js';
 import { createLogger } from './utils/logger.js';
+import { detectIdeFromEnv, DetectedIde } from '@google/gemini-cli-core';
 
 const CLI_IDE_COMPANION_IDENTIFIER = 'Google.gemini-cli-vscode-ide-companion';
 const INFO_MESSAGE_SHOWN_KEY = 'geminiCliInfoMessageShown';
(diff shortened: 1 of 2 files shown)


packages/vscode-ide-companion/src/extension.ts
```diff
@@ -9,6 +9,8 @@ import { IDEServer } from './ide-server.js';
 import semver from 'semver';
 import { DiffContentProvider, DiffManager } from './diff-manager.js';
 import { createLogger } from './utils/logger.js';
+import { DetectedIde } from '@google/gemini-cli-core';
+import { detectIdeFromEnv } from '@google/gemini-cli-core/src/ide/detect-ide.js';
```
romannurik: unclear if this is a dependency smell for this package - happy to revisit, but also don't want to add a new export to the main package unless necessary

romannurik: added the export.

romannurik:
Abandoning in favor of Richie's #8097

PR #8100, opened 2025-09-09
Extensions update command

jakemac53:
## TLDR

Adds `/extensions update` and `/extensions list`, although `/extensions` will still default to just the list command as well.

Can pass `--all` to update all extensions or a list of named extensions.

## Reviewer Test Plan

Install a git extension, then update the remote repo and run `/extensions update <extension-name>`.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Part of #8018
Works in combination with #7009 although they are independently updated

### DIFF
--- a/packages/cli/src/ui/commands/extensionsCommand.ts
+++ b/packages/cli/src/ui/commands/extensionsCommand.ts
@@ -69,6 +69,7 @@ async function updateAction(context: CommandContext, args: string) {
         },
         Date.now(),
       );
+      return;
     }
 
     // Filter to the actually updated ones.
--- a/packages/cli/src/ui/commands/extensionsCommand.ts
+++ b/packages/cli/src/ui/commands/extensionsCommand.ts
@@ -47,68 +47,61 @@ async function listAction(context: CommandContext) {
 }
 
 const updateOutput = (info: ExtensionUpdateInfo) =>
-  `Extension "${info.name}" successfully updated: ${info.originalVersion} → ${info.updatedVersion}. Restart gemini-cli to see the changes.`;
+  `Extension "${info.name}" successfully updated: ${info.originalVersion} → ${info.updatedVersion}.`;
 
 async function updateAction(context: CommandContext, args: string) {
   const updateArgs = args.split(' ').filter((value) => value.length > 0);
   const all = updateArgs.length === 1 && updateArgs[0] === '--all';
   const names = all ? undefined : updateArgs;
-
-  if (all) {
-    try {
-      const updateInfos = await updateAllUpdatableExtensions();
-      if (updateInfos.length === 0) {
-        context.ui.addItem(
-          {
-            type: MessageType.INFO,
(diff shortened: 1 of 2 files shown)


chrstnb:
Can you add a screenshot of this in action? looks great otherwise!

chrstnb (approved):

jakemac53:
https://github.com/user-attachments/assets/ac7f449c-d5d5-42e6-ac0e-98b58dfc0a6d

jakemac53:
~Note that the general check of whether an extension is up to date or not is broken at head but this will be fixed by https://github.com/google-gemini/gemini-cli/pull/8018 but that is somewhat helpful for the demo video 🤣 .~

Actually this is expected to be handled by the consumer of the API I guess, I added code in to remove these non-updated extensions now.

packages/cli/src/ui/commands/extensionsCommand.ts
```diff
@@ -4,43 +4,137 @@
...
+
+const updateOutput = (info: ExtensionUpdateInfo) =>
+  `Extension "${info.name}" successfully updated: ${info.originalVersion} → ${info.updatedVersion}. Restart gemini-cli to see the changes.`;
+
+async function updateAction(context: CommandContext, args: string) {
+  const updateArgs = args.split(' ').filter((value) => value.length > 0);
+  const all = updateArgs.length === 1 && updateArgs[0] === '--all';
+  const names = all ? undefined : updateArgs;
+
+  if (all) {
```
cornmander: Can you refactor this? The logic is mostly duplicated between branches and it's very nested. Even the final else {} block can be refactored to just be an unconditional if you return in the other branches.

jakemac53: done

packages/cli/src/ui/commands/extensionsCommand.ts
```diff
@@ -4,43 +4,130 @@
...
+  let updateInfos: ExtensionUpdateInfo[] = [];
+  try {
+    if (all) {
+      updateInfos = await updateAllUpdatableExtensions();
+    } else if (names?.length) {
+      for (const name of names) {
+        updateInfos.push(await updateExtensionByName(name));
+      }
+    } else {
       context.ui.addItem(
```
cornmander, line 65: Should we return early here? Otherwise we will print the "No extensions to update" message as well below right?

cornmander (approved):

packages/cli/src/ui/commands/extensionsCommand.ts
jakemac53: Yep good call

PR #8101, opened 2025-09-09
feat(security) - Encrypted oauth flag

shishu314:
## TLDR

Use encrypted storage when flag is true for oauth

## Dive Deeper

Using the same flag as the mcp but we might want to consider using a different flag, let me know. Using two flags would let us launch one without the other interfering and might make sense.

This PR also makes OAuthCredentialStorage static

## Reviewer Test Plan

1. Set the flag to true by running `export GEMINI_FORCE_ENCRYPTED_FILE_STORAGE="true"`
1. Add breakpoints in the code when the flag is true
1. Run `npm run debug` and see that the flag paths are triggered
1. During re-runs, add `await OAuthCredentialStorage.clearCredentials();` in the `loadCachedCredentials` function
1. Check that authentication flow works without the breakpoints

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | x  | ❓  | ❓  |
| npx      | x  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixed https://github.com/google-gemini/maintainers-gemini-cli/issues/746


packages/core/src/mcp/token-storage/index.ts
```diff
@@ -0,0 +1,14 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export * from './types.js';
+export * from './base-token-storage.js';
+export * from './file-token-storage.js';
+export * from './hybrid-token-storage.js';
+
+export const DEFAULT_SERVICE_NAME = 'gemini-cli-oauth';
+export const FORCE_ENCRYPTED_FILE_ENV_VAR =
```
cornmander: Would it be much more difficult to use a flag rather than an environment variable? I think we're trying to move away from environment variables controlling deep functionality within the codebase.

packages/core/src/code_assist/oauth-credential-storage.test.ts
```diff
@@ -52,12 +48,22 @@ describe('OAuthCredentialStorage', () => {
...
 
   beforeEach(() => {
-    storage = new HybridTokenStorage('');
-    oauthStorage = new OAuthCredentialStorage(storage);
-
-    vi.spyOn(storage, 'getCredentials').mockResolvedValue(null);
-    vi.spyOn(storage, 'setCredentials').mockResolvedValue(undefined);
-    vi.spyOn(storage, 'deleteCredentials').mockResolvedValue(undefined);
+    vi.spyOn(
+      OAuthCredentialStorage['storage'],
```
richieforeman: They key access here is suspect. Are you spying on private details here?

richieforeman: I read on and it seems that yuo are. This feels bad.

Ibra7997 (changes requested):
ت

Ibra7997 (approved):

packages/core/src/code_assist/oauth2.ts
```diff
@@ -87,7 +94,11 @@ async function initOauthClient(
   }
 
   client.on('tokens', async (tokens: Credentials) => {
-    await cacheCredentials(tokens);
+    if (useEncryptedStorage) {
+      OAuthCredentialStorage.saveCredentials(tokens);
+    } else {
+      await cacheCredentials(tokens);
+    }
```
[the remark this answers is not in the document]
Ibra7997: ن

richieforeman (approved):

PR #8106, opened 2025-09-09
chore(a2a-server): if a2a task creation fails return error to user

adamfweidman:
## TLDR

Adds try-catch so that task creation can't fail silently

## Reviewer Test Plan

Start `npm run start:a2a-server` without USE_CCPA or GEMINI_API_KEY set and send a request from VSCode GCA Extension with `"geminicodeassist.a2a.address": "http://localhost:41242/"` set in VSCode User Settings

And 

npm run preflight

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | x  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


packages/a2a-server/src/agent/executor.ts
```diff
@@ -381,12 +381,52 @@ export class CoderAgentExecutor implements AgentExecutor {
...
-      );
+      try {
+        wrapper = await this.createTask(
+          taskId,
+          contextId,
+          agentSettings,
+          eventBus,
+        );
+      } catch (error) {
+        const errorMessage =
```
cornmander: Can you extract this to a helper?

oauth_settings.txt
cornmander, line 1: unintentional addition?

adamfweidman: deleted

packages/a2a-server/src/agent/executor.ts
adamfweidman: Done.

cornmander (approved):

PR #8107, opened 2025-09-09
feat(ide): Update context filtering to be done CLI-side + update port discovery logic

skeshive:
## TLDR
This pull request introduces foundational changes to support community-authored IDE extensions. It establishes a robust and stable contract by centralizing complex logic, such as context filtering, within the CLI. This architectural change isolates the CLI from potential issues in companion extensions, improving overall system stability.

## Dive Deeper

* Context Filtering Moved to CLI: The responsibility for filtering context has been moved from the extensions to the CLI. This guarantees that the CLI always operates on correctly processed information.
* New Port Discovery File Naming Convention: Extensions must now generate a port discovery file with the format gemini-ide-server-{pid}-{port}.json. Using the IDE's process ID (pid) and the server port in the filename prevents conflicts between multiple running IDE instances.

## Linked issues / bugs
#4800, #6848


packages/core/src/ide/ide-client.ts
```diff
@@ -406,8 +408,78 @@ export class IdeClient {
       const portFileContents = await fs.promises.readFile(portFile, 'utf8');
       return JSON.parse(portFileContents);
     } catch (_) {
+      // For newer extension versions, the file matches the following prefix. If multiple IDE windows are open, multiple files with the same prefix are expected to exist.
+    }
+
+    let portFiles;
+    try {
+      portFiles = await fs.promises.readdir(os.tmpdir());
```
cornmander: you should nest these in a directory, /tmp may contain many files

cornmander (approved):

packages/core/src/ide/ide-client.ts
skeshive: done

PR #8110, opened 2025-09-09
[Part 1/6] feat(telemetry): add rate limiter and high-water mark tracker with tests

eLyiN:
## Summary

Adds foundational telemetry utilities required for enhanced performance monitoring: a RateLimiter to prevent telemetry spam with high-priority bypass, and a HighWaterMarkTracker to reduce noise by only recording significant changes.

## Dependencies

⚠️ **This PR has no dependencies** - it's the foundational part of the split implementation.

**Review Order**: This is Part 1/6 - review this PR first before the dependent parts.

## Related PRs - Enhanced Performance Monitoring Split

This is **Part 1 of 6** in the split implementation of #2127:

- ✅ **Part 1/6**: #8110 (telemetry-primitives) - **YOU ARE HERE**
- ⏳ **Part 2/6**: #8111 (activity-detector) - depends on Part 1
- ⏳ **Part 3/6**: #8113 (metrics-enhancements) - depends on Parts 1-2  
- ⏳ **Part 4/6**: #8122 (memory-monitor) - depends on Parts 1-3
- ⏳ **Part 5/6**: #8124 (activity-monitor) - depends on Parts 1-4
- ⏳ **Part 6/6**: #8125 (cli-integration) - depends on Parts 1-5
- 📋 **Tracking Issue**: #8120

### RateLimiter (`packages/core/src/telemetry/rate-limiter.ts`)
- Prevents telemetry spam by limiting calls per time window (default: 5 calls/10 seconds)
- **High-priority bypass**: Critical events can bypass rate limiting
- Configurable limits and time windows for different use cases
- Memory-efficient with automatic cleanup of old entries

### HighWaterMarkTracker (`packages/core/src/telemetry/high-water-mark-tracker.ts`)  
- Tracks significant changes in numeric values (memory usage, performance metrics)
- **Reduces telemetry noise** by only recording when values exceed previous peaks by threshold
- Configurable thresholds (default: 10% increase to record new high water mark)
- Supports both absolute and percentage-based thresholds

### Exports (`packages/core/src/telemetry/index.ts`)
- Adds exports for both utilities and their TypeScript types
- Maintains backwards compatibility with existing telemetry exports
- Clean API for usage in dependent components

## Testing

✅ **Comprehensive unit tests** using vitest:
- `rate-limiter.test.ts` - 22 test cases covering rate limiting, high-priority bypass, edge cases
- `high-water-mark-tracker.test.ts` - 17 test cases covering threshold tracking, configuration scenarios

[...]

Issue #2127, opened 2025-06-27
Enhanced Performance Monitoring: Startup instrumentation, memory tracking, and detailed metrics collection


eLyiN:
## Summary

Enhance the existing telemetry system with comprehensive performance monitoring capabilities, including detailed startup instrumentation, memory usage tracking, and advanced metrics collection for better debugging and optimization.

## Current State

The existing telemetry system (based on OpenTelemetry) provides excellent basic metrics:
- Session counts, tool calls, API requests
- Token usage tracking  
- Basic latency measurements

However, there are gaps when troubleshooting performance issues like:
- Issues #1685 (startup/exit problems) - lack of detailed startup phase timing
- Issues #1687 (rate limiting + memory issues) - limited memory monitoring
- General performance optimization needs for large-scale usage

### 1. Detailed Startup Performance Monitoring
- **Phase-by-phase timing**: Settings loading, cleanup, extensions loading, config loading, service initialization
- **Rich attributes**: Include context like auth type, telemetry settings, workspace info  
- **Debugging capability**: Help diagnose startup failures with precise timing data

### 2. Comprehensive Memory Monitoring
- **Real-time tracking**: Heap usage, total memory, external memory, RSS
- **Component-specific monitoring**: Track memory usage by CLI components
- **Automatic snapshots**: At key lifecycle points (startup, major operations)
- **Growth analysis**: Detect memory leaks and usage patterns

### 3. Enhanced Metrics Framework
- **12 new performance metrics**: Including startup timing, memory usage, CPU usage, tool execution breakdown
- **Performance scoring**: Composite performance indicators
- **Regression detection**: Automated comparison against baselines

## Benefits

1. **Better issue diagnosis**: Help resolve startup issues like #1685 with detailed timing data
2. **Memory issue detection**: Catch memory problems like those in #1687 before they impact users
3. **Performance optimization**: Provide data for optimizing CLI performance at scale
4. **Backwards compatible**: Extends existing OpenTelemetry infrastructure without breaking changes

## Implementation Approach

- Follows existing OpenTelemetry patterns and naming conventions (\`gemini_cli.*\`)
- Integrates with current telemetry configuration (local/GCP targets)
- Maintains <1% performance overhead requirement
- Uses established telemetry enable/disable controls

## Technical Details

The implementation adds:
- New performance metrics constants following existing patterns
- Memory monitoring infrastructure with automatic lifecycle management
- Startup instrumentation hooks in main CLI entry point  
[...]

PR #8111, opened 2025-09-09
[Part 2/6] feat(telemetry): add activity detector with user interaction tracking

eLyiN:
## Summary

Adds activity detection capabilities for user interaction tracking, enabling activity-aware telemetry recording. The ActivityDetector tracks user activity across different interaction types and provides APIs to check current activity state.

## Dependencies

⚠️ **This PR depends on:**
- [x] #8110 - Part 1: Telemetry Primitives (RateLimiter, HighWaterMarkTracker)

**Review Order**: Please review Part 1 (#8110) first, then this PR.

## Related PRs - Enhanced Performance Monitoring Split

This is **Part 2 of 6** in the split implementation of #2127:

- ✅ **Part 1/6**: #8110 (telemetry-primitives) - Foundation
- ✅ **Part 2/6**: #8111 (activity-detector) - **YOU ARE HERE** 
- ⏳ **Part 3/6**: #8113 (metrics-enhancements) - depends on Parts 1-2
- ⏳ **Part 4/6**: #8122 (memory-monitor) - depends on Parts 1-3
- ⏳ **Part 5/6**: #8124 (activity-monitor) - depends on Parts 1-4
- ⏳ **Part 6/6**: #8125 (cli-integration) - depends on Parts 1-5
- 📋 **Tracking Issue**: #8120

### ActivityTypes (`packages/core/src/telemetry/activity-types.ts`)
- Defines different types of user activities: `KEYBOARD`, `MOUSE`, `SYSTEM`, `IDLE`
- Provides TypeScript enums for type-safe activity tracking
- Foundation for categorizing different user interaction patterns

### ActivityDetector (`packages/core/src/telemetry/activity-detector.ts`)
- **Core activity tracking**: Records user activities with timestamps and types
- **Activity state checking**: `isUserActive()` API to check if user is currently active
- **Configurable timeouts**: Define how long to consider user active after last interaction (default: 30 seconds)
- **Memory efficient**: Automatic cleanup of old activity records
- **Thread-safe**: Uses Maps with proper cleanup to prevent memory leaks

### Enhanced Exports (`packages/core/src/telemetry/index.ts`)
- Exports `ActivityDetector` class and related APIs
- Exports `ActivityType` enum for type-safe usage

### DIFF
--- a/packages/core/src/telemetry/activity-detector.test.ts
+++ b/packages/core/src/telemetry/activity-detector.test.ts
@@ -1,0 +1,192 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { describe, it, expect, beforeEach, vi } from 'vitest';
+import {
+  ActivityDetector,
+  initializeActivityDetector,
+  getActivityDetector,
+  recordUserActivity,
+  isUserActive,
+  resetGlobalActivityDetector,
+} from './activity-detector.js';
+
+describe('ActivityDetector', () => {
+  let detector: ActivityDetector;
+
+  beforeEach(() => {
+    detector = new ActivityDetector(1000); // 1 second idle threshold for testing
+  });
+
+  describe('constructor', () => {
+    it('should initialize with default idle threshold', () => {
+      const defaultDetector = new ActivityDetector();
+      expect(defaultDetector).toBeInstanceOf(ActivityDetector);
+    });
+
+    it('should initialize with custom idle threshold', () => {
+      const customDetector = new ActivityDetector(5000);
+      expect(customDetector).toBeInstanceOf(ActivityDetector);
+    });
+  });
+
+  describe('recordActivity', () => {
+    it('should update last activity time', () => {
+      const beforeTime = detector.getLastActivityTime();
+
+      // Wait a small amount to ensure time difference
+      vi.useFakeTimers();
+      vi.advanceTimersByTime(100);
+
+      detector.recordActivity();
+      const afterTime = detector.getLastActivityTime();
+
+      expect(afterTime).toBeGreaterThan(beforeTime);
+      vi.useRealTimers();
+    });
+  });
+
+  describe('isUserActive', () => {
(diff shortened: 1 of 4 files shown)


eLyiN:
@jacob314 ready for review

packages/core/src/telemetry/activity-detector.test.ts
```diff
@@ -0,0 +1,193 @@
...
+    it('should return detector state when exists', () => {
+      initializeActivityDetector(1000);
+      expect(isUserActive()).toBe(true);
+
+      vi.useFakeTimers();
+      vi.advanceTimersByTime(2000);
+
+      expect(isUserActive()).toBe(false);
+
+      vi.useRealTimers();
```
jacob314: calling useRealTimers and useFakeTimers should be done in
beforeEach
and 
afterEach blocks rather than integrated into individual tests.

packages/core/src/telemetry/activity-detector.ts
```diff
@@ -0,0 +1,92 @@
...
+ */
+export function getActivityDetector(): ActivityDetector | null {
+  return globalActivityDetector;
+}
+
+/**
+ * Record user activity (convenience function for CLI to call)
+ */
+export function recordUserActivity(): void {
+  // Do not implicitly initialize here to avoid locking in default timeout
```
jacob314: this logic seems a bit more complicated then needed. Do we really need a threshold other than the default? if we don't we can just write
const globalActivityDetector = new ActivityDetector(); 
and keep this a lot simpler. Later if we need more configurability we can always add it.

jacob314 (commented):
approved once the minor comments are addressed. Not a huge fan of this requiring an initialize method and being nullable unless there is a strong case for it.

packages/core/src/telemetry/activity-detector.ts
eLyiN: Totally agree,  the configurable threshold was added for rare edge cases, but it haven't seen a real need. The 30s default is sufficient, commited changes

joshualitt (approved):

jacob314 (approved):
lgtm

PR #8139, opened 2025-09-10
feat: Add /chat share command

### LINKED ISSUE #8030: Export saved chats to Markdown

A command like this would be useful for demonstrating what Gemini CLI has done to other team members:

```
/chat export foo filename.md
```

Markdown is much easier to digest than copying the raw json file of the saved conversation.

### Why is this needed?

Makes it easy to show off something cool that Gemini CLI did!

### Additional context

_No response_

rramkumar1:
## TLDR

This commit introduces a new `/chat share` command that allows users to export their current conversation to a markdown or JSON file which is easily consumable for other humans or UIs.

## Dive Deeper

- Adds a `share` subcommand to the `chat` command.
- Validates the file extension to ensure it's either `.md` or `.json`.
- Creates the output file if it doesn't already exist.
- Writes the conversation history to the specified file in the chosen format.
- Adds tests for the new `share` command.

## Reviewer Test Plan

You can have a simple conversation with Gemini CLI then run:

```
/chat share test.md
```

```
/chat share test.json
```

Both files will be in your current working directory and you can open them to verify the correct formatting.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ✅  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Resolves #8030

### DIFF
--- a/packages/cli/src/ui/commands/chatCommand.ts
+++ b/packages/cli/src/ui/commands/chatCommand.ts
@@ -19,6 +19,7 @@ import { decodeTagName } from '@google/gemini-cli-core';
 import path from 'node:path';
 import type { HistoryItemWithoutId } from '../types.js';
 import { MessageType } from '../types.js';
+import type { Content } from '@google/genai';
 
 interface ChatDetail {
   name: string;
@@ -274,9 +275,96 @@ const deleteCommand: SlashCommand = {
   },
 };
 
+export function serializeHistoryToMarkdown(history: Content[]): string {
+  return history
+    .map((item) => {
+      const text =
+        item.parts
+          ?.filter((m) => !!m.text)
+          .map((m) => m.text)
+          .join('') || '';
+      return `**${item.role}**:\n\n${text}`;
+    })
+    .join('\n\n---\n\n');
+}
+
+const shareCommand: SlashCommand = {
+  name: 'share',
+  description:
+    'Share the current conversation to a markdown or json file. Usage: /chat share <file>',
+  kind: CommandKind.BUILT_IN,
+  action: async (context, args): Promise<MessageActionReturn> => {
+    let filePathArg = args.trim();
+    if (!filePathArg) {
+      filePathArg = `gemini-conversation-${Date.now()}.json`;
+    }
+
+    const filePath = path.resolve(filePathArg);
+    const extension = path.extname(filePath);
+    if (extension !== '.md' && extension !== '.json') {
+      return {
+        type: 'message',
+        messageType: 'error',
+        content: 'Invalid file format. Only .md and .json are supported.',
+      };
+    }
+
+    const chat = await context.services.config?.getGeminiClient()?.getChat();
+    if (!chat) {
+      return {
+        type: 'message',
+        messageType: 'error',
+        content: 'No chat client available to share conversation.',
+      };
+    }
+
+    const history = chat.getHistory();
+
+    // An empty conversation has two hidden messages that setup the context for
+    // the chat. Thus, to check whether a conversation has been started, we
+    // can't check for length 0.
+    if (history.length <= 2) {
+      return {
+        type: 'message',
+        messageType: 'info',
+        content: 'No conversation found to share.',
+      };
+    }
+
+    let content = '';
+    if (extension === '.json') {
+      content = JSON.stringify(history, null, 2);
+    } else {
+      content = serializeHistoryToMarkdown(history);
+    }
+
+    try {
+      await fsPromises.writeFile(filePath, content);
+      return {
+        type: 'message',
+        messageType: 'info',
+        content: `Conversation shared to ${filePath}`,
+      };
+    } catch (err) {
+      const errorMessage = err instanceof Error ? err.message : String(err);
+      return {
+        type: 'message',
+        messageType: 'error',
+        content: `Error sharing conversation: ${errorMessage}`,
+      };
+    }
+  },
+};
+
 export const chatCommand: SlashCommand = {
   name: 'chat',
   description: 'Manage conversation history.',
   kind: CommandKind.BUILT_IN,
-  subCommands: [listCommand, saveCommand, resumeCommand, deleteCommand],
+  subCommands: [
+    listCommand,
+    saveCommand,
+    resumeCommand,
+    deleteCommand,
+    shareCommand,
+  ],
 };
--- a/docs/cli/commands.md
+++ b/docs/cli/commands.md
@@ -30,6 +30,9 @@ Slash commands provide meta-level control over the CLI itself.
     - **`delete`**
       - **Description:** Deletes a saved conversation checkpoint.
       - **Usage:** `/chat delete <tag>`
+    - **`share`**
+      - **Description** Writes the current conversation to a provided Markdown or JSON file.
+      - **Usage** `/chat share file.md` or `/chat share file.json`. If no filename is provided, then the CLI will generate one.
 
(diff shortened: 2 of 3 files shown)


packages/cli/src/ui/commands/chatCommand.ts
```diff
@@ -274,9 +274,103 @@ const deleteCommand: SlashCommand = {
...
 };
 
+const shareCommand: SlashCommand = {
+  name: 'share',
+  description:
+    'Share the current conversation to a markdown or json file. Usage: /chat share <file>',
+  kind: CommandKind.BUILT_IN,
+  action: async (context, args): Promise<MessageActionReturn> => {
+    const filePathArg = args.trim();
+    if (!filePathArg) {
```
galdawave, line 298: Nit: instead of this can we have a default file, i think it would be nice just to type `/chat share` but not super opinionated on this

packages/cli/src/ui/commands/chatCommand.ts
```diff
@@ -274,9 +274,103 @@ const deleteCommand: SlashCommand = {
...
+      return {
+        type: 'message',
+        messageType: 'error',
+        content: 'No chat client available to share conversation.',
+      };
+    }
+
+    const history = chat.getHistory();
+
+    if (history.length <= 2) {
```
galdawave, line 326: Nit: please add a comment to clarify why the threshold is 2

packages/cli/src/ui/commands/shareCommand.test.ts
```diff
@@ -0,0 +1,156 @@
...
+
+describe('share subcommand', () => {
+  const mockFs = fsPromises as Mocked<typeof fsPromises>;
+
+  let mockContext: CommandContext;
+  let mockGetChat: ReturnType<typeof vi.fn>;
+  let mockGetHistory: ReturnType<typeof vi.fn>;
+  let shareCommand: SlashCommand;
+
+  const getSubCommand = (name: 'share'): SlashCommand => {
```
galdawave: Nit: The getSubCommand helper function is defined inside the main describe block. Since it doesn't rely on any state set up within beforeEach, it could be moved to the module scope for slightly better organization.

galdawave (approved):
Probably should update [docs](https://github.com/google-gemini/gemini-cli/blob/a31830a3cb16584ec0448d6a23da27947ac9e72c/docs/cli/commands.md?plain=1#L14) docs

packages/cli/src/ui/commands/chatCommand.ts
cornmander: I agree with this. I think the file name is sort of irrelevant, if not specified it should write to a randomly generated file in tmpdir like gemini-chat-$timestamp.md

packages/cli/src/ui/commands/chatCommand.ts
```diff
@@ -274,9 +274,103 @@ const deleteCommand: SlashCommand = {
...
+      return {
+        type: 'message',
+        messageType: 'error',
+        content: 'Missing file path. Usage: /chat share <file>',
+      };
+    }
+
+    const filePath = path.resolve(filePathArg);
+    // Prevent path traversal attacks by ensuring the path is within the current working directory.
+    if (!filePath.startsWith(process.cwd())) {
```
cornmander: Can you check existing utils for this? We do this check in a bunch of places

rramkumar1: Gemini was able to find this for me:

https://github.com/google-gemini/gemini-cli/blob/main/packages/core/src/utils/fileUtils.ts#L172

Although, it said "it's not exported as part of the public API of the
  @google/gemini-cli-core package".

Suggestions?

cornmander: Sorry, actually, on further thought, I don't think we need to even do this check. The cwd scoping is relevant for actions the agent takes, not for actions the user takes, especially an export action.

packages/cli/src/ui/commands/chatCommand.ts
```diff
@@ -274,9 +274,106 @@ const deleteCommand: SlashCommand = {
...
+    if (history.length <= 2) {
+      return {
+        type: 'message',
+        messageType: 'info',
+        content: 'No conversation found to share.',
+      };
+    }
+
+    let content = '';
+    if (extension === '.json') {
```
cornmander, line 335: Can you extract the serialization into functions? I think we should test these standalone.

I also think that what you have is a good starting point, but we probably should include chat metadata, like the date of the chat and potentially the cwd that it was run in.

packages/cli/src/ui/commands/chatCommand.ts
rramkumar1: Removed

packages/cli/src/ui/commands/shareCommand.test.ts
rramkumar1: I actually moved the tests to the existing chatCommand.test.ts file. I didn't see a reason to have a separate test file just for this subcommand.

packages/cli/src/ui/commands/chatCommand.ts
rramkumar1: Gemini, can you help with this?

rramkumar1: /gemini

rramkumar1: /gemini Can you just address this comment for now? 

"Can you extract the serialization into functions? I think we should test these standalone."

rramkumar1: @cornmander Do you want to just test the markdown serialization piece? JSON is just a JSON.stringify call so not sure how much you gain from testing that.

cornmander: I think it's worth testing that it matches an expected schema.

rramkumar1: Done, wdyt on adding the additional metadata in a separate PR?

cornmander: sgtm

cornmander (approved):

rramkumar1:
@cornmander Do the workflows need to be approved by you?

bulkypanda:
@rramkumar1 looks like tests are failing on windows

rramkumar1:
/gemini Can you fix the failing test chatCommand.test.ts when it runs on Windows. It looks like the test expects a Linux-style path (forward slash) but for the Windows test, the path generated will follow Windows format (backslash).

Simply normalizing the received path to the Linux style in the test code should do the trick.

rramkumar1:
/gemini Looks like the Windows test is still failing. This time on the check that the output result matches what is expected. 

Is there a way to make the `expectedPath` variable in each test case a Linux or Windows path depending on some env var. That would clean up the test code a bit.

rramkumar1:
/gemini What about line 424 where process.cwd() is mocked to return a Posix style path? Doesn't that need to be "os aware"?

rramkumar1:
@cornmander I think I need the e2e:ok label again. Not sure why gemini-bot removed it.

rramkumar1:
/gemini Can you run the e2e tests?

PR #8145, opened 2025-09-10
fix(rendering): remove React.strictMode from gemini.tsx

galdawave:
## TLDR

This PR removes `React.StrictMode` from the main application entry point to resolve UI rendering issues caused by components rendering twice in development.

## Dive Deeper

`React.StrictMode` is a development tool that intentionally double-invokes render functions to help detect side effects. While valuable, this behavior was causing unexpected rendering glitches and instability in our Ink-based CLI. Components would render twice, leading to a flickering or inconsistent UI, especially during streaming output.

By removing `React.StrictMode`, we ensure that components render only once, which stabilizes the UI in the development environment and makes it behave consistently with the production build. This change has no impact on the final production output; it only affects the development server.

## Reviewer Test Plan

This change primarily impacts the development experience.

1.  Run the CLI in development mode (e.g., `npm start`).
2.  Execute a few commands that produce both static and streaming output.
3.  Pay close attention to the UI rendering. Confirm that there is no flickering, duplicate output, or other visual artifacts that were present before.
4.  Verify that the overall functionality of the CLI remains unchanged.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


jacob314 (approved):
Nice catch!lgtm
Can you follow up with a test that passes with these changes but fails without them? ideal test would verify that we aren't rerendering continuously when running app?

jacob314 (approved):
lgtm

PR #8146, opened 2025-09-10
Support extension enablement overrides

chrstnb:
## TLDR

We got feedback from folks that they are reliant on workspace-level extension configuration, and this offers a solution to that problem. 

Context: in the latest release, we prompt users with workspace-level extensions to install them at the user level. We recently added a suite of new functionality for extensions management, and having extensions installed at the workspace level makes things more complicated and confusing. We may eventually want to support workspace-level extensions, but for now we view that as a power-user feature that clutters up the extensions architecture. For things like updates, e.g,., it may be confusing to have the same extension installed in multiple places.

To get around this, users can use the 'link' command to source an extension definition to their workspace path. They can also, with this PR, disable an extension by default but enable it in a specific workspace via override.

## Summary of Changes

This change introduces a new mechanism for managing extensions at the workspace level, providing more granular control over which extensions are active in different contexts. The key feature is the ability to override user-level extension settings within a specific workspace.

### New `--override` Flag for `enable` Command

The `enable` command now includes an `--override` flag. This flag allows a user to enable an extension for the current workspace, even if it has been disabled at the user level. This is particularly useful when a user wants to disable an extension globally but enable it for a specific project.

### Workspace Extension Overrides

A new concept of "workspace extension overrides" has been introduced. This is managed through a new file, `workspace-extension-overrides.json`, located in the user's Gemini configuration directory.

- When an extension is enabled with the `--override` flag, an entry is added to this file, linking the extension to the current workspace.
- The `loadExtensions` logic has been updated to check for these overrides. If an override exists for the current workspace, the extension will be loaded, even if it's in the user's disabled list.
- When an extension is disabled at the workspace level, any corresponding override for that workspace is removed.

### Workspace Extension Migration

The `performWorkspaceExtensionMigration` function has been updated. When migrating a workspace extension, it will now:

1.  Install the extension at the user level.
2.  Disable the extension at the user level.
3.  Create a workspace override to ensure the extension is active only for the current workspace.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


chrstnb:
This is being replaced by https://github.com/google-gemini/gemini-cli/compare/cb/rmworkspace?expand=1

PR #8147, opened 2025-09-10
fix(search): Respect .gitignore in SearchText tool

### LINKED ISSUE #6322: Uses 'SearchText' on all files even with respectGitIgnore=true
### What happened?

I was asking gemini-cli to replace usages of `Regex::new()` with a macro.
It would use the `SearchText` tool, but instead of only searching through the project files, it also searched through artifacts which are gitignored.

Is it expected that `SearchText` is searching also in gitignored files?

The `SearchText` should only search in non-ignored files, or gemini should use it more cleverly, e.g. only search in `.rs` files.

### Client information

<details>

```console
$ gemini /about
│ CLI Version       0.1.21                              │
│ Git Commit        055fe46d                            │
│ Model             gemini-2.5-pro                      │
│ Sandbox           no sandbox                          │
│ OS                linux                               │
│ Auth Method       gemini-api-key                      │

```

</details>

### Login information

│ Auth Method       gemini-api-key

_No response_

qingshanyuluo:
The SearchText tool was not correctly respecting .gitignore rules, causing it to search through ignored files and build artifacts. This led to significant performance degradation.

This change ensures that ignored files are properly excluded from the search scope before the search is executed, improving both performance and the accuracy of the results.

Fixes: #6322

## TLDR

The SearchText tool was incorrectly searching through files and directories specified in .gitignore, causing significant performance degradation,
  especially in projects with large build artifacts or node_modules. This PR corrects that behavior by ensuring ignore rules are respected before the
  search is executed.

## Dive Deeper

The root cause of the problem was that the search mechanism was scanning all files first and then attempting to filter the results, rather than
  excluding ignored files from the start. This change modifies the implementation of the underlying search tools (grep, ripgrep) to ensure they
  correctly leverage their native ignore-handling capabilities. By filtering out files listed in .gitignore and .geminiignore before the search
  operation begins, we prevent the tool from wasting I/O and CPU cycles on irrelevant files.

## Reviewer Test Plan

1. In a test project, create a .gitignore file and add a directory to it (e.g., echo "ignored_dir/" >> .gitignore).
   2. Create a file inside that directory with a unique keyword (e.g., echo "secret_keyword_in_ignored_file" > ignored_dir/test.txt).
   3. Create another file outside the ignored directory with a different keyword (e.g., echo "public_keyword_in_tracked_file" > tracked_file.txt).
   4. From the CLI, run a search for "secret_keyword_in_ignored_file".
       - Expected: The search should be fast and return no results.
   5. Run a new search for "public_keyword_in_tracked_file".
       - Expected: The search should find the keyword in tracked_file.txt.
   6. This confirms that ignore rules are being respected without breaking existing search functionality for tracked files.
 
## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ✅  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

  Fixes #6322


skeshive (approved):

packages/core/src/tools/grep.ts
```diff
@@ -336,14 +336,11 @@ class GrepToolInvocation extends BaseToolInvocation<
...
-          '-n',
-          '-E',
-          '--ignore-case',
-          pattern,
-        ];
+        const gitArgs = ['grep'];
+        if (!this.config.getFileFilteringRespectGitIgnore()) {
+          gitArgs.push('--untracked');
+        }
+        gitArgs.push('-n', '-E', '--ignore-case', pattern);
```
scidomino: extreme nit: it would be clearer if you specified these constant args on line 339 since they always get added.

scidomino (commented):
This works but my concern is that we already have gitignore filtering code in the other tools and we should reuse that for consistency (instead of using git). This is also important since we should support .geminiignore as well. 

I have a similar PR out for the glob tool: https://github.com/google-gemini/gemini-cli/pull/8086

qingshanyuluo:
Hi maintainers, I'm a contributor from a fork. My PR failed the CI with a git fetch error. I think this may be related to the workflow not supporting forked branches. Do you have any suggestions or can you help fix the workflow? Thank you!

PR #8153, opened 2025-09-10
feat(routing): Initialize model routing architecture

abhipatel12:
## TLDR

This PR introduces a foundational model routing architecture to enable dynamic model selection for user prompts.

**NOTE:** This PR does not make any functional changes to actually implement dynamic model routing. It only adds the architecture to do so.

- **What:** A `ModelRouterService` is introduced, which is responsible for deciding which model (e.g., Gemini Pro, Gemini Flash) to use for a given conversation.
- **Why:** This architecture paves the way for more intelligent, cost-effective, and performant model usage. In the future, we can implement strategies to route simple queries to faster, cheaper models and complex queries to more powerful models.
- **How:**
    - The `GeminiClient` now orchestrates model selection. For the first turn of a new prompt, it calls the `ModelRouterService`.
    - The chosen model is then made "sticky" for the entire duration of that multi-turn prompt to ensure a consistent and coherent user experience.
    - Lower-level components (`GeminiChat`, `Turn`) have been refactored to accept the chosen model as a parameter, rather than looking it up in the config.
- A new `ModelRouterService` that uses a chain-of-responsibility pattern with different strategies.
- Strategies for handling fallback mode, user overrides, and a default model selection.

The core logic changes are in `packages/core/src/core/client.ts` and the new `packages/core/src/routing/` directory.

## Dive Deeper

This change introduces a more sophisticated approach to model management. Previously, the model was a static configuration value. Now, it's a dynamic decision made at the start of each conversational sequence (identified by a `prompt_id`).

### Model Stickiness

A key concept here is "model stickiness." Once the router selects a model for a prompt, that same model is used for all subsequent turns within that prompt. This is crucial for maintaining conversational context and preventing jarring shifts in tone or capability that could occur if the model changed mid-conversation. The `GeminiClient` now manages this state via the `currentSequenceModel` property.

### Inversion of Control

This refactor also improves the overall architecture by applying an inversion of control pattern. Instead of lower-level modules like `GeminiChat` pulling the model from the `Config` service, the high-level `GeminiClient` now *pushes* the selected model down as a parameter. This makes the downstream components more modular, predictable, and easier to test in isolation.

### Strategies
The initial order of strategies is:
1.  **`FallbackStrategy`**: If the client is in fallback mode (e.g., due to a previous quota error), this strategy immediately takes over. It intelligently downgrades "pro" models to "flash" while honoring requests for "lite" models to preserve cost-saving intentions.
2.  **`OverrideStrategy`**: If a model is explicitly set in the user's configuration, this strategy ensures that the user's choice is respected, bypassing any other logic.
3.  **`DefaultStrategy`**: This is the terminal strategy in the chain. If no other strategy makes a decision, it ensures that we always fall back to the default `gemini-2.5-pro` model.

### Future Possibilities

Although this PR does not introduce functional changes, it unlocks the ability to build more advanced routing strategies in the future, such as:
- **Cost-based routing:** Use Flash for prompts that don't require complex reasoning.
- **Capability-based routing:** Detect if a prompt requires advanced tool use or reasoning and route to Pro.
- **Content-based routing:** Analyze the prompt content (e.g., code generation vs. summarization) to select the best-suited model.

**NOTE**
The Default strategy is not used b/c the override mechanism is greedy. This is in preparation for how we will explicitly set the model once we introduce an actual dynamic router.

## Reviewer Test Plan

The most effective way to validate this change is by reviewing the new unit tests, which thoroughly cover the routing and stickiness logic.

- **Primary Test Files:**
    - `packages/core/src/core/client.test.ts`: See the new "Model Routing" `describe` block.
    - `packages/core/src/routing/modelRouterService.test.ts`: Validates the new service logic.

### Manual Validation (Optional)

Since the `DefaultStrategy` preserves the existing behavior, the CLI should function identically from a user's perspective. To see the new architecture in action, a reviewer can add logging:

1.  **Add logs:**
    - In `packages/core/src/routing/modelRouterService.ts`, log the decision in the `route` method.
    - In `packages/core/src/core/client.ts`, log when the router is being called vs. when a sticky model is being used.
2.  **Run the CLI and start a chat.**
3.  **First Turn:** On your first message, you should see the log from the `ModelRouterService` indicating it's making a decision.
4.  **Second Turn:** On your follow-up message in the same chat, you should *not* see the router log. Instead, you should see a log indicating that the sticky model is being used.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  |✅| ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt |✅| -   | -   |

## Linked issues / bugs

### DIFF
--- a/packages/core/src/config/config.ts
+++ b/packages/core/src/config/config.ts
@@ -44,6 +44,7 @@ import { StartSessionEvent } from '../telemetry/index.js';
 import {
   DEFAULT_GEMINI_EMBEDDING_MODEL,
   DEFAULT_GEMINI_FLASH_MODEL,
+  DEFAULT_GEMINI_MODEL,
 } from './models.js';
 import { shouldAttemptBrowserLaunch } from '../utils/browser.js';
 import type { MCPOAuthConfig } from '../mcp/oauth-provider.js';
@@ -62,6 +63,7 @@ import {
   RipgrepFallbackEvent,
 } from '../telemetry/types.js';
 import type { FallbackModelHandler } from '../fallback/types.js';
+import { ModelRouterService } from '../routing/modelRouterService.js';
 import { OutputFormat } from '../output/types.js';
 
 // Re-export OAuth config type
@@ -266,6 +268,7 @@ export class Config {
   private readonly usageStatisticsEnabled: boolean;
   private geminiClient!: GeminiClient;
   private baseLlmClient!: BaseLlmClient;
+  private modelRouterService: ModelRouterService;
   private readonly fileFiltering: {
     respectGitIgnore: boolean;
     respectGeminiIgnore: boolean;
@@ -278,7 +281,7 @@ export class Config {
   private readonly proxy: string | undefined;
   private readonly cwd: string;
   private readonly bugCommand: BugCommandSettings | undefined;
-  private readonly model: string;
+  private model: string;
   private readonly extensionContextFilePaths: string[];
   private readonly noBrowser: boolean;
   private readonly folderTrustFeature: boolean;
@@ -366,7 +369,7 @@ export class Config {
     this.cwd = params.cwd ?? process.cwd();
     this.fileDiscoveryService = params.fileDiscoveryService ?? null;
     this.bugCommand = params.bugCommand;
-    this.model = params.model;
+    this.model = params.model || DEFAULT_GEMINI_MODEL;
     this.extensionContextFilePaths = params.extensionContextFilePaths ?? [];
     this.maxSessionTurns = params.maxSessionTurns ?? -1;
     this.experimentalZedIntegration =
@@ -416,6 +419,7 @@ export class Config {
       setGlobalDispatcher(new ProxyAgent(this.getProxy() as string));
     }
     this.geminiClient = new GeminiClient(this);
+    this.modelRouterService = new ModelRouterService(this);
   }
 
   /**
@@ -515,13 +519,16 @@ export class Config {
   }
 
   getModel(): string {
-    return this.contentGeneratorConfig?.model || this.model;
+    return this.model;
   }
 
   setModel(newModel: string): void {
-    if (this.contentGeneratorConfig) {
-      this.contentGeneratorConfig.model = newModel;
+    // Do not allow Pro usage if the user is in fallback mode.
+    if (newModel === DEFAULT_GEMINI_MODEL && this.isInFallbackMode()) {
+      return;
     }
+
+    this.model = newModel;
   }
 
   isInFallbackMode(): boolean {
@@ -689,6 +696,10 @@ export class Config {
 
   getGeminiClient(): GeminiClient {
     return this.geminiClient;
+  }
+
+  getModelRouterService(): ModelRouterService {
+    return this.modelRouterService;
   }
 
   getEnableRecursiveFileSearch(): boolean {
--- a/packages/core/src/config/config.ts
+++ b/packages/core/src/config/config.ts
@@ -524,7 +524,7 @@ export class Config {
 
   setModel(newModel: string): void {
     // Do not allow Pro usage if the user is in fallback mode.
-    if (newModel === DEFAULT_GEMINI_MODEL && this.isInFallbackMode()) {
+    if (newModel.includes('pro') && this.isInFallbackMode()) {
       return;
     }
 
--- a/packages/core/src/core/client.ts
+++ b/packages/core/src/core/client.ts
@@ -493,7 +493,6 @@ export class GeminiClient {
       history: this.getChat().getHistory(/*curated=*/ true),
       request,
       signal,
-      explicitModel: '',
     };
 
     let modelToUse: string;
@@ -502,13 +501,9 @@ export class GeminiClient {
     if (this.currentSequenceModel) {
       modelToUse = this.currentSequenceModel;
     } else {
-      if (this.config.isInFallbackMode()) {
-        modelToUse = DEFAULT_GEMINI_FLASH_MODEL;
-      } else {
-        const router = this.config.getModelRouterService();
-        const decision = await router.route(routingContext);
-        modelToUse = decision.model;
-      }
+      const router = await this.config.getModelRouterService();
+      const decision = await router.route(routingContext);
+      modelToUse = decision.model;
       // Lock the model for the rest of the sequence
       this.currentSequenceModel = modelToUse;
     }
--- a/packages/core/src/core/client.ts
+++ b/packages/core/src/core/client.ts
@@ -49,6 +49,7 @@ import {
 } from '../telemetry/types.js';
 import type { IdeContext, File } from '../ide/types.js';
 import { handleFallback } from '../fallback/handler.js';
+import type { RoutingContext } from '../routing/routingStrategy.js';
 
 export function isThinkingSupported(model: string) {
   if (model.startsWith('gemini-2.5')) return true;
@@ -118,6 +119,7 @@ export class GeminiClient {
 
   private readonly loopDetector: LoopDetectionService;
   private lastPromptId: string;
+  private currentSequenceModel: string | null = null;
   private lastSentIdeContext: IdeContext | undefined;
   private forceFullIdeContext = true;
 
@@ -426,11 +428,11 @@ export class GeminiClient {
     signal: AbortSignal,
     prompt_id: string,
     turns: number = MAX_TURNS,
-    originalModel?: string,
   ): AsyncGenerator<ServerGeminiStreamEvent, Turn> {
     if (this.lastPromptId !== prompt_id) {
       this.loopDetector.reset(prompt_id);
       this.lastPromptId = prompt_id;
+      this.currentSequenceModel = null;
     }
     this.sessionTurnCount++;
     if (
@@ -445,9 +447,6 @@ export class GeminiClient {
     if (!boundedTurns) {
       return new Turn(this.getChat(), prompt_id);
     }
-
-    // Track the original model from the first call to detect model switching
-    const initialModel = originalModel || this.config.getModel();
 
     const compressed = await this.tryCompressChat(prompt_id);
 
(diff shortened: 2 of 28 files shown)


packages/core/src/config/config.ts
```diff
@@ -272,7 +275,7 @@ export class Config {
   private readonly proxy: string | undefined;
   private readonly cwd: string;
   private readonly bugCommand: BugCommandSettings | undefined;
-  private readonly model: string;
+  private model: string;
```
mrcabbage972, line 284: Maybe make the model name an enum to simplify error handling?

packages/core/src/config/config.ts
```diff
@@ -505,13 +509,16 @@ export class Config {
...
   getModel(): string {
-    return this.contentGeneratorConfig?.model || this.model;
+    return this.model;
   }
 
   setModel(newModel: string): void {
-    if (this.contentGeneratorConfig) {
-      this.contentGeneratorConfig.model = newModel;
+    // Do not allow Pro usage if the user is in fallback mode.
+    if (newModel === DEFAULT_GEMINI_MODEL && this.isInFallbackMode()) {
```
mrcabbage972: Should the fallback be handled at this level of abstraction? Feels like it should be a part of the router.

packages/core/src/core/client.ts
```diff
@@ -490,7 +489,31 @@ export class GeminiClient {
...
+    const routingContext: RoutingContext = {
+      history: this.getChat().getHistory(/*curated=*/ true),
+      request,
+      signal,
+      explicitModel: '',
+    };
+
+    let modelToUse: string;
+
+    // Determine Model (Stickiness vs. Routing)
```
mrcabbage972, line 504: I'd encapsulate all of this logic in the router.

packages/core/src/core/client.ts
```diff
@@ -490,7 +489,31 @@ export class GeminiClient {
...
+      if (this.config.isInFallbackMode()) {
+        modelToUse = DEFAULT_GEMINI_FLASH_MODEL;
+      } else {
+        const router = this.config.getModelRouterService();
+        const decision = await router.route(routingContext);
+        modelToUse = decision.model;
+      }
+      // Lock the model for the rest of the sequence
+      this.currentSequenceModel = modelToUse;
+    }
```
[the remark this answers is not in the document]
mrcabbage972: 1. Is the compression affected by the router?

packages/core/src/core/geminiChat.ts
```diff
@@ -225,13 +226,10 @@ export class GeminiChat {
     let response: GenerateContentResponse;
 
     try {
-      let currentAttemptModel: string | undefined;
-
       const apiCall = () => {
         const modelToUse = this.config.isInFallbackMode()
           ? DEFAULT_GEMINI_FLASH_MODEL
```
mrcabbage972: Move this logic to router?

packages/core/src/routing/modelRouterService.ts
```diff
@@ -0,0 +1,63 @@
...
+   * Determines which model to use for a given request context.
+   *
+   * @param context The full context of the request.
+   * @returns A promise that resolves to a RoutingDecision.
+   */
+  async route(context: RoutingContext): Promise<RoutingDecision> {
+    // Return fallback model if in fallback mode.
+    if (this.config.isInFallbackMode()) {
+      return {
+        model: DEFAULT_GEMINI_FLASH_MODEL,
```
mrcabbage972: This logic is now duplicated in a couple of places, as mentioned in other comments.

packages/core/src/routing/modelRouterService.ts
```diff
@@ -0,0 +1,63 @@
...
+        metadata: {
+          source: 'Explicit',
+          latencyMs: 0,
+        },
+      };
+
+      return decision;
+    }
+
+    return this.strategy.route(context, this.config.getBaseLlmClient());
```
mrcabbage972: The returned model name needs to be validated (or make it an enum).

packages/core/src/routing/routingStrategy.ts
```diff
@@ -0,0 +1,60 @@
...
+  model: string;
+  /** A brief, loggable explanation of why this model was chosen. */
+  reason: string;
+  /**
+   * Metadata about the routing decision for logging purposes.
+   */
+  metadata: {
+    source: 'Default' | 'Explicit' | 'Fallback';
+    latencyMs: number;
+    reasoning?: string;
```
mrcabbage972: Is this different from the reason field above?

packages/core/src/routing/routingStrategy.ts
```diff
@@ -0,0 +1,60 @@
...
+export interface RoutingDecision {
+  /** The model identifier string to use for the next API call (e.g., 'gemini-2.5-pro'). */
+  model: string;
+  /** A brief, loggable explanation of why this model was chosen. */
+  reason: string;
+  /**
+   * Metadata about the routing decision for logging purposes.
+   */
+  metadata: {
+    source: 'Default' | 'Explicit' | 'Fallback';
```
mrcabbage972: Can we also include the name of the strategy?

packages/core/src/routing/modelRouterService.ts
```diff
@@ -0,0 +1,63 @@
...
+        reason: `In fallback mode. Using: ${DEFAULT_GEMINI_FLASH_MODEL}`,
+        metadata: {
+          source: 'Fallback',
+          latencyMs: 0,
+        },
+      };
+    }
+
+    // Honor the override mechanism.
+    if (context.explicitModel) {
```
mrcabbage972: Maybe encapsulate explicitModel and fallback as strategies? That would simplify the design.

packages/core/src/routing/routingStrategy.ts
```diff
@@ -0,0 +1,60 @@
...
+    error?: string;
+  };
+}
+
+/**
+ * The context provided to the router for making a decision.
+ */
+export interface RoutingContext {
+  /** The full history of the conversation. */
+  history: Content[];
```
mrcabbage972, line 33: Full history or compressed history?

packages/core/src/routing/routingStrategy.ts
```diff
@@ -0,0 +1,60 @@
...
+  /** The immediate request parts to be processed. */
+  request: PartListUnion;
+  /** An abort signal to cancel an LLM call during routing. */
+  signal: AbortSignal;
+  /**
+   * If provided, the router will be bypassed and this model will be used directly.
+   * This is the override mechanism for specific internal tasks or user settings.
+   */
+  explicitModel?: string;
```
mrcabbage972: Why is this needed in the interface? It can be a strategy and strategies can be composed. Then each one can be tested individually rather than having to test the whole matrix of different scenarios. Less bug potential in my opinion.

packages/core/src/config/config.ts
abhipatel12: We allow users to specify the model name by string. This would mean keeping up with all model strings. Leaving as is for now.

packages/core/src/core/client.ts
abhipatel12: To determine model stickiness, we currently rely on the promptId. However, if we move this into the model service, we then cause the router to be in charge of routing as well as keeping track of all clients and their current turn. 

This may work but becomes more of an issue as we add more agents that require a router. For now, we should keep the model stickiness as a responsibility of the client.

I am removing the fallback check since that was a copypasta mistake from my previous PR.

packages/core/src/config/config.ts
abhipatel12: This is for when the user decides to set their configured model and if they are in fallback mode. I thought it made sense to fail fast and early if the model cannot be upgraded. 

We could allow this action, but then on the next user request, they will be downgraded again on their actual request which could make for an odd user experience. 

What do you think?

packages/core/src/core/client.ts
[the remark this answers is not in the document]
abhipatel12: The compression should not use a dynamically chosen mode.

packages/core/src/routing/routingStrategy.ts
abhipatel12: This is the current most history as provided by the client. Since compression happens prior to routing, this would mean the history after compression takes place.

packages/core/src/routing/routingStrategy.ts
abhipatel12: Removed it

The original intention/design was that all requests would be routed. However, with this impl, we are not going to route utility calls. So for now, this is not necessary and has been removed.

packages/core/src/routing/routingStrategy.ts
abhipatel12: Nope, copypasta issue again, sorry. 

Moved this into metadata

packages/core/src/routing/routingStrategy.ts
abhipatel12: This has been solved by changing how we create strategies

packages/core/src/routing/modelRouterService.ts
abhipatel12: How do you recommend validating it? 

Models can be strings passed from the user, so if it is wrong it will fail during the request. 

For the classifier case in the future, we simply map the response to the actual model string constant so that should be safe as well.

packages/core/src/routing/modelRouterService.ts
abhipatel12: Done

packages/core/src/routing/modelRouterService.ts
abhipatel12: The flash fallback logic is hard to centralize at this level b/c of how we deal with retries. This will take a larger architectural change. 

In the meantime, we can consolidate this to a single function that is used in any place that needs it.

packages/core/src/core/geminiChat.ts
abhipatel12: Addressed in the other comment

abhipatel12 (commented):
Thanks for the feedback Victor! Let me know what you think of the changes.

mrcabbage972 (approved):

anj-s (approved):

PR #8155, opened 2025-09-10
feat(accessibility): implement centralized screen reader layout system

vick08:
## TLDR
Introduces a screen reader friendly layout for the app UI.

## Dive Deeper
- The "prompt line" is moved to the bottom of the window for better compatibility with screen readers.
- - The conversation is moved right above the "input" component.
- - The rest of the auxiliary information is moved above the conversation.
- - This layout is only activated when the "--screen-reader" flag is used.

Refactored screen reader accessibility implementation to use a centralized layout system for better maintainability and resilience to future changes.

Changes:
- Created useScreenReaderLayout hook for centralized layout configuration
- Added ScreenReaderAppLayout and DefaultAppLayout components
- Refactored App.tsx to use layout-specific components
- Updated MainContent and Composer to use layout configuration
- Maintains existing screen reader functionality with footer->conversation->input flow

/CC @chrstnb


chrstnb (approved):
Looks great to me-- just needs`npm run format` 

Thank you!!!

scidomino (approved):

packages/cli/src/ui/layouts/useScreenReaderLayout.ts
```diff
@@ -0,0 +1,23 @@
...
+
+import { useIsScreenReaderEnabled } from 'ink';
+
+export interface LayoutConfig {
+  shouldUseStatic: boolean;
+  shouldShowFooterInComposer: boolean;
+  mode: 'default' | 'screenReader';
+}
+
+export const useScreenReaderLayout = (): LayoutConfig => {
```
jacob314: we also want to Toggle off and on support for static for non screen reader layouts. in some terminals That was one of the motivations of the big risky refactor that went in. Lets tweak this to not have a screen reader specific name. Expect we will want to add a bunch of layout config here as well and will be easier if it is all in one place rather than multiple sources of truth for whether static should be enabled, etc.

packages/cli/src/ui/components/MainContent.tsx
```diff
@@ -11,17 +11,54 @@ import { OverflowProvider } from '../contexts/OverflowContext.js';
...
   const uiState = useUIState();
+  const layout = useScreenReaderLayout();
   const {
     pendingHistoryItems,
     mainAreaWidth,
     staticAreaMaxItemHeight,
     availableTerminalHeight,
   } = uiState;
 
+  // In screen reader mode, use regular layout without Static component
```
jacob314, line 28: this should not be screen reader specific. it should generically handle the case where we don't use static.

jacob314, line 28: this code is extremely similar to the code right bellow it except for the lack of wrapping with Static. Can we reduce duplication?

Lyonk71:
Looks like `thrid_party/get-ripgrep` is unused. can we get it removed?

vick08:
> Looks like `thrid_party/get-ripgrep` is unused. can we get it removed?
Was this a part of this commit? Perhaps this should go in separately?

Lyonk71:
Hey @vick08 - I noticed that this PR is now way smaller than the original PR I was reviewing... just making sure it's as-intended & not a force push issue? I think it went down from ~ 5k lines across ~130 files added to 200  lines added across 10.

vick08:
> Hey @vick08 - I noticed that this PR is now way smaller than the original PR I was reviewing... just making sure it's as-intended & not a force push issue? I think it went down from ~ 5k lines across ~130 files added to 200 lines added across 10.
My original feature still works as intended. There was some refactoring since the original PR though, so perhaps this is what you are seeing?
I hope this answers the question.

Lyonk71:
Ahh, i see what happened. It looks like your code was sitting on top of a bad merge/rebase, so when you did a rebase on top of main and force pushed earlier today, it fixed the massive diff. I was actually wondering while i was reviewing why so much content appeared to be unrelated to accessibility, lol. That explains it.

Lyonk71:
Hey - I noticed the diff is back up to over 2,000 lines from 200... Can you give me a quick rundown of what changed to help me get up to speed w/ the review?

chrstnb:
Hello @Lyonk71! Thank you for the comments-- @vick08 rebased and I just ran workflows. Want to confirm this approach works with your updates before submitting!

Lyonk71:
Taking a look now...

Lyonk71:
Nice addition of the `useFooterProps` hook! I noticed there's some pre-existing footer props construction in `Composer.tsx` (lines 54-71) that builds a very similar object. In a follow-up, we could potentially extend your hook to handle those additional hide* properties and use it in Composer too. But that's pre-existing code, not something for this PR.

packages/cli/src/ui/components/MainContent.tsx
```diff
@@ -5,59 +5,80 @@
...
-              shellFocused={uiState.shellFocused}
-            />
-          ))}
+          <PendingHistoryList
+            pendingHistoryItems={pendingHistoryItems}
+            terminalWidth={mainAreaWidth}
+            availableTerminalHeight={availableTerminalHeight}
+            constrainHeight={uiState.constrainHeight}
+            isEditorDialogOpen={uiState.isEditorDialogOpen}
+            activePtyId={uiState.activePtyId?.toString()}
```
Lyonk71, line 79: I noticed the `activePtyId` prop is being converted from a number to a string here...

packages/cli/src/ui/components/PendingHistoryList.tsx
```diff
@@ -0,0 +1,45 @@
...
+      <HistoryItemDisplay
+        key={i}
+        availableTerminalHeight={
+          constrainHeight ? availableTerminalHeight : undefined
+        }
+        terminalWidth={terminalWidth}
+        item={{ ...item, id: 0 }}
+        isPending={true}
+        isFocused={!isEditorDialogOpen}
+        activeShellPtyId={activePtyId ? parseInt(activePtyId, 10) : null}
```
Lyonk71, line 40: ... and then parsed back to a number here. I may be missing something, but I don't see where the string is actually being used. Can we simplify by keeping it as a number?

packages/cli/src/ui/components/MainContent.tsx
```diff
@@ -5,59 +5,80 @@
...
+            staticAreaMaxItemHeight={staticAreaMaxItemHeight}
+            slashCommands={uiState.slashCommands}
+          />
+          <PendingHistoryList
+            pendingHistoryItems={pendingHistoryItems}
+            terminalWidth={mainAreaWidth}
+            availableTerminalHeight={availableTerminalHeight}
+            constrainHeight={uiState.constrainHeight}
+            isEditorDialogOpen={uiState.isEditorDialogOpen}
+          />
```
Lyonk71, line 46: I noticed screen reader mode is missing `activePtyId` and `shellFocused` props  (lines 40-46, compare with 73-81). Would a screen reader need this state to announce which shell is active and has focus?

Lyonk71:
All set with the review. didn't see anything that will interfere with #5399. Thanks for looping me in @chrstnb ... Great job on the PR @vick08!

Lyonk71:
@vick08 grats on the merge in. I just noticed i didn't actually hit submit on my review for you :jack_o_lantern: Just submitted... you can see if any of this is stuff you want to make a follow up commit for.

Issue #8167, opened 2025-09-10
Potential async error handling issues


amsminn:
The codebase doesn't enforce the @typescript-eslint/return-await rule, which can lead to subtle async error handling issues.
There are already several places in the codebase where this async error handling issue exists.

```typescript
async function example() {
  try {
    return someAsyncFunction(); // Error bypasses catch block
  } catch (error) {
    // Never executed for async rejections
    console.error('Error:', error);
  }
}
```

When returning a promise directly without await in try-catch blocks, rejections occur after the
function returns, bypassing the catch handler entirely.

PS. There are also many places in the codebase with unnecessary return await usage

amsminn:
> Hello! As part of our effort to keep our backlog manageable and focus on the most active issues, we are tidying up older reports.안녕하세요! 적체 처리를 관리하기 쉽고 가장 활발한 이슈에 집중하기 위한 노력의 일환으로, 우리는 오래된 보고서를 정리하고 있습니다.
> 
> It looks like this issue hasn't been active for a while, so we are closing it for now. However, if you are still experiencing this bug on the latest stable build, please feel free to comment on this issue or create a new one with updated details.이 문제는 한동안 활성화되지 않은 것 같아 일단 종료합니다. 하지만 최신 안정 빌드에서 여전히 이 버그가 발생하고 있다면, 이 문제에 대해 댓글로 남기거나 최신 정보를 반영한 새 버그를 만들어 주시기 바랍니다.
> 
> Thank you for your contribution!기여해 주셔서 감사합니다!

This issue has been resolved in [#13714](https://github.com/google-gemini/gemini-cli/pull/13714)

PR #8168, opened 2025-09-10
fix(async): prevent missed async errors in try/catch with return-await ESLint rule

amsminn:
## TLDR

Added @typescript-eslint/return-await ESLint rule to enforce
proper async/await usage in try-catch blocks, preventing errors
 from bypassing catch handlers. Also cleaned up unnecessary
return await in non-try-catch contexts.

## Dive Deeper

The rule addresses a subtle but important anti-pattern where
returning promises directly in try-catch blocks causes async
errors to escape the catch handler

```typescript
// Before: Error bypasses catch
try {
  return someAsyncFunc();
} catch (e) {
  // Never called for async errors
}
```

```typescript
// After: Error properly caught
try {
  return await someAsyncFunc();
} catch (e) {
  // Handles async errors correctly
}
```

This PR includes:
1. ESLint config update with the new rule
2. Fixed 15 files to add required await in try-catch blocks
3. Removed unnecessary return await in regular async contexts

## Reviewer Test Plan

1. Run `npm run lint` - should pass without errors
2. Check modified files for proper async/await usage

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅   | ❓   | ❓   |
| npx      | ✅   | ❓   | ❓   |

## Linked issues / bugs

- This PR Resolves on #8167

### DIFF
--- a/packages/core/src/tools/edit.ts
+++ b/packages/core/src/tools/edit.ts
@@ -578,7 +578,7 @@ Expectation for required parameters:
       getFilePath: (params: EditToolParams) => params.file_path,
       getCurrentContent: async (params: EditToolParams): Promise<string> => {
         try {
-          return this.config
+          return await this.config
             .getFileSystemService()
             .readTextFile(params.file_path);
         } catch (err) {
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -85,6 +85,10 @@ export default tseslint.config(
         ...globals.node,
         ...globals.es2021,
       },
+      parserOptions: {
+        projectService: true,
+        tsconfigRootDir: projectRoot,
+      },
     },
     rules: {
       // General Best Practice Rules (subset adapted for flat config)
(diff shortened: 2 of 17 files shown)


packages/core/src/tools/edit.ts
```diff
@@ -538,7 +538,7 @@ Expectation for required parameters:
       getFilePath: (params: EditToolParams) => params.file_path,
       getCurrentContent: async (params: EditToolParams): Promise<string> => {
         try {
-          return this.config
+          return await this.config
```
skeshive, line 581: hmm I'm seeing local test failures due to this- can you double check this change?

amsminn: On my local machine, running test didn’t show any test failures related to edit.ts. 
However, I did find an issue in editor.ts in the [Test (macos-latest, 20.x)](https://github.com/google-gemini/gemini-cli/actions/runs/17611079002/job/50051295525#logs) test run.

#### What happened
Before, the try just returned a promise. If that promise later failed, it happened outside the try/catch, so the catch never ran.
With await, the try actually waits, and any error goes straight into the catch. The only issue left was that our catch logged the error but didn’t rethrow it.

#### Fix
I changed the catch block to log and then rethrow
```typescript
catch (error) {
  console.error(error);
  throw error;
}
```

adamfweidman (approved):

PR #8173, opened 2025-09-10
fix: make `bfsFileSearch` more efficient

gsquared94:
## TLDR

The `bfsFileSearch` is needlessly calling `fileService?.shouldIgnoreFile` for each and every file. It can only call this method for files matching the target file, or for directories. This drastically reduces the number of invocations of `fileService?.shouldIgnoreFile`. This will be relevant after #8172.

### DIFF
--- a/packages/core/src/utils/bfsFileSearch.ts
+++ b/packages/core/src/utils/bfsFileSearch.ts
@@ -99,6 +99,16 @@ export async function bfsFileSearch(
     for (const { currentDir, entries } of results) {
       for (const entry of entries) {
         const fullPath = path.join(currentDir, entry.name);
+        const isDirectory = entry.isDirectory();
+        const isMatchingFile = entry.isFile() && entry.name === fileName;
+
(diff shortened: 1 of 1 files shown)


packages/core/src/utils/bfsFileSearch.ts
```diff
@@ -99,22 +99,29 @@ export async function bfsFileSearch(
...
-            respectGeminiIgnore:
-              options.fileFilteringOptions?.respectGeminiIgnore,
-          })
-        ) {
-          continue;
-        }
+        const isDirectory = entry.isDirectory();
+        const isMatchingFile = entry.isFile() && entry.name === fileName;
+
+        if (isDirectory || isMatchingFile) {
```
cornmander: You should invert the condition to reduce nesting.

gsquared94: Done

cornmander (approved):

jacob314 (approved):
lgtm

PR #8177, opened 2025-09-10
fix: `gitignore` handling

gsquared94:
## TLDR

  This pull request refactors the `.gitignore` file parsing logic and lazy loads the `gitignore` to improve performance on large and deeply nested projects.

##  Dive Deeper

  This PR includes two main changes:

   1. Refactor ignore parsers: The parsing logic for `.gitignore` and `.geminiignore` has been
      separated into their own respective parser files (gitignoreParser.ts and
      geminiIgnoreParser.ts). This improves modularity, making the code cleaner and easier to
      maintain.

   2. Lazy load `.gitignore` files (similar to `git` behavior): Instead of parsing ignore files on startup, they are now parsed
      only when a file is read or a directory is listed for the first time. This avoids
      unnecessary upfront work and improves startup performance.

### Test program doesn't hang at startup for large projects
   1.  (Only Googlers) Create a `.git` directory in a large CiTC workspace (say `/google3/cloud/.git/`)
   2. Run gemini from HEAD. Observe that it hangs indefinitely.
   3. Run gemini from this branch. It should startup immediately.
   5. Verify that the application starts up without errors.

### Test nested `.gitignore` files loading behavior 
   1. Setup test project:
```
 # Create a root directory for the test project and enter it
 mkdir test_gitignore_project && cd test_gitignore_project
 
 # Initialize a Git repository
 git init
 
 # Create a nested directory structure
 mkdir -p dir1/dir2/dir3
 mkdir -p dir4
 
 # Create various files to test against the ignore rules
 touch root_file.txt
 touch root_file.log
 touch dir1/file_in_dir1.txt
 touch dir1/file_in_dir1.log
 touch dir1/file_to_keep.tmp
 touch dir1/file_to_ignore.tmp
 touch dir1/dir2/file_in_dir2.txt
 touch dir1/dir2/dir3/file_in_dir3.txt
 touch dir4/file_in_dir4.log
 
 # Create .gitignore files at different levels
 
 # Root .gitignore: ignores all .log files
 echo "*.log" > .gitignore
 
 # Nested .gitignore in dir1: ignores .tmp files, but negates the rule for a specific file
 echo -e "*.tmp\n!file_to_keep.tmp" > dir1/.gitignore
 
 # Nested .gitignore in dir1/dir2: ignores everything inside, but un-ignores dir3
 echo -e "*\n!/dir3/" > dir1/dir2/.gitignore
 
 # Add all files to staging to see the effect of the .gitignore rules
 git add .
 
 # Check the git status to see which files are tracked and which are ignored
 echo "\n--- Git Status ---"
 git status --short
```
2. Should see this output:
```
A  .gitignore
A  dir1/.gitignore
A  dir1/file_in_dir1.txt
A  dir1/file_to_keep.tmp
A  root_file.txt
```

3. Now run this command in Gemini built from this branch:
image

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ✅  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

- Fixes #8172


packages/core/src/utils/geminiIgnoreParser.ts
```diff
@@ -0,0 +1,81 @@
...
+    if (normalizedPath.startsWith('/') || normalizedPath === '') {
+      return false;
+    }
+
+    return this.ig.ignores(normalizedPath);
+  }
+
+  getPatterns(): string[] {
+    return this.patterns;
+  }
```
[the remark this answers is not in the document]
gsquared94: GeminiIgnoreParser needs to be synchronous.

packages/core/src/utils/gitIgnoreParser.test.ts
```diff
@@ -33,13 +33,13 @@ describe('GitIgnoreParser', () => {
     await fs.rm(projectRoot, { recursive: true, force: true });
   });
 
-  describe('initialization', () => {
-    it('should initialize without errors when no .gitignore exists', async () => {
+  describe('Basic ignore behaviors', () => {
+    it('should not ignore files when no .gitignore exists', async () => {
       await setupGitRepo();
```
cornmander, line 38: can you move the await setupGitRepo() step into a beforeEach()? we seem to do that for some test scenarios and not others.

packages/core/src/utils/gitIgnoreParser.ts
```diff
@@ -186,13 +122,68 @@ export class GitIgnoreParser implements GitIgnoreFilter {
...
+          if (ig.ignores(normalizedRelativeDir)) {
+            // This directory is ignored by an ancestor's .gitignore.
+            // According to git behavior, we don't need to process this
+            // directory's .gitignore, as nothing inside it can be
+            // un-ignored.
+            break;
+          }
+        }
+
+        if (this.cache.has(dir)) {
```
cornmander, line 167: There's no invalidation, so if any of the gitignore files along the chain changes, the entry will be wrong. Are you sure we actually need to cache?

gsquared94: That is true, however we currently also don't honor changes to the file in-between the running program. It would be a significant perf hit if we didn't cache it.

packages/core/src/utils/gitIgnoreParser.test.ts
gsquared94: Done.

cornmander (approved):

packages/core/src/utils/geminiIgnoreParser.test.ts
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  beforeEach(async () => {
+    projectRoot = await fs.mkdtemp(
+      path.join(os.tmpdir(), 'geminiignore-test-'),
+    );
+  });
+
+  afterEach(async () => {
+    await fs.rm(projectRoot, { recursive: true, force: true });
```
jacob314, line 29: minor nit:  clear project root in beforeEach as well to avoid bugs if tests are terminiated without afterEach being call.

packages/core/src/utils/geminiIgnoreParser.ts
```diff
@@ -0,0 +1,81 @@
...
+  constructor(projectRoot: string) {
+    this.projectRoot = path.resolve(projectRoot);
+    this.loadPatterns();
+  }
+
+  private loadPatterns(): void {
+    const patternsFilePath = path.join(this.projectRoot, '.geminiignore');
+    let content: string;
+    try {
+      content = fs.readFileSync(patternsFilePath, 'utf-8');
```
jacob314, line 30: this is probably the right tradeoff but a little concerned the file reads are now sync.  could be worth to consider what it would take to make this non-blocking. it might cause too many ripple effect changes to be worth it.

packages/core/src/utils/gitIgnoreParser.ts
```diff
@@ -186,13 +122,68 @@ export class GitIgnoreParser implements GitIgnoreFilter {
...
+      const dirsToVisit = [this.projectRoot];
+      let currentAbsDir = this.projectRoot;
+      // Collect all directories in the path
+      for (let i = 0; i < pathParts.length - 1; i++) {
+        currentAbsDir = path.join(currentAbsDir, pathParts[i]);
+        dirsToVisit.push(currentAbsDir);
+      }
+
+      for (const dir of dirsToVisit) {
+        const relativeDir = path.relative(this.projectRoot, dir);
```
jacob314, line 155: we might want to tweak this to directly support `includeDirectories`. For now we should probably just ignore this logic if the path.relative left the root dir.
In the future we could  figure out which of the included directories the path is from and only consider the git ignores for that included directory.

jacob314 (approved):
lgtm

PR #8195, opened 2025-09-10
feat(core): Copy `downloadRipGrep` with license.

joshualitt:
## TLDR

Copies `downloadRipGrep` into the repo along with its license.

## Dive Deeper

This CL just copies the code. In the follow-up, we can add tests and wire this code up.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Necessary for #7327

### DIFF
--- a/third_party/get-ripgrep/LICENSE
+++ b/third_party/get-ripgrep/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2023 Lvce Editor
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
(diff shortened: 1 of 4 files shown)


packages/core/third-party/get-ripgrep/downloadRipGrep.js
joshualitt, line 1: Could have a top level `third-party` folder, but making this part of core will make it easier to integrate into our build system. I'm happy to move it to `third-party` if desired, but if so we will have to figure out who is responsible for dependencies / building for the code.

joshualitt: Can also call this `third_party` if desired.

teeler: I think third_party is conventional, that would be preferred (please, and thank you for offering!)

I think it's OK that it's package-by-package but I can defer to jacobr@ and others for preferences on that. The only real advantage to the top level is easier auditing (ie everything in exactly one place, don't have to go through each package and look independently). If the integration is significantly easier I'd say use your judgement, I think a symlink and/or just a reference in a top-level third_party folder might also provide a nice hint.

joshualitt:
@teeler , @jacob314  ptal. I've gotten confirmation that the top level `third_party` directory + license file in subdirectory is the way to go.

teeler (approved):
LGTM, thanks!

PR #8196, opened 2025-09-10
feat: simplify patch release workflow

mattKorwel:
This PR simplifies the patch-release.yml workflow by using the reusable publish-release action.


mattKorwel:
I think this is ready to go. This provides a github action that will let a user kick off a patch release AFTER they have cherry picked and created the right branches. In this pr that work is still manual. in #8202 we make that better as follow on work.

We also try and capture a bunch of the linting tools locally as it is extremely frustrating to have them fail on ci with out a way to mirror that behavior locally. I'm certain the `lint:all` is not super robust nor does it need to be for this pr. just incremental improvements.

.github/workflows/patch-release.yml
```diff
@@ -0,0 +1,105 @@
...
+        uses: 'actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8' # ratchet:actions/checkout@v5
+        with:
+          ref: '${{ github.event.inputs.ref || github.sha }}'
+          fetch-depth: 0
+
+      - name: 'Set booleans for simplified logic'
+        env:
+          DRY_RUN_INPUT: '${{ github.event.inputs.dry_run }}'
+        id: 'vars'
+        run: |-
```
richieforeman: Somewhat naive question, but isn't this just setting env variables?

Why not do this in `env` rather than executing some shell?

.github/workflows/patch-release.yml
```diff
@@ -0,0 +1,105 @@
...
+          force_skip_tests: '${{ github.event.inputs.force_skip_tests }}'
+          gemini_api_key: '${{ secrets.GEMINI_API_KEY }}'
+
+      - name: 'Publish Release'
+        uses: './.github/actions/publish-release'
+        with:
+          release-version: '${{ steps.version.outputs.RELEASE_VERSION }}'
+          release-tag: '${{ steps.version.outputs.RELEASE_TAG }}'
+          npm-tag: '${{ steps.version.outputs.NPM_TAG }}'
+          wombat-token-core: '${{ secrets.WOMBAT_TOKEN_CORE }}'
```
richieforeman, line 90: Confidence check: Is this a new secret or an old one?

If it's an old one: Carry on and ignore me

If it's a new one: It'd be nice to give this more (potentially verbose meaning). `GOOGLE_NPM_DEPLOYMENT_SERVICE_CORE_TOKEN` or something.  When I was digging thru these things I had to do some archeology to figure out what "wombat dressing room" was.

scripts/get-release-version.js
```diff
@@ -71,6 +71,33 @@ export function getVersion(options = {}) {
...
+      throw new Error(
+        'Patch type must be specified with --patch-from=stable or --patch-from=preview',
+      );
+    }
+
+    if (patchFrom === 'stable') {
+      previousReleaseTag = getLatestTag(
+        '(contains("nightly") or contains("preview")) | not',
+      );
+      const [major, minor, patch] = previousReleaseTag
```
richieforeman: Super nit: Might make sense to assert on presence of `major` and `minor`.  It seems like `patch` is an optional component.

scripts/get-release-version.js
```diff
@@ -71,6 +71,33 @@ export function getVersion(options = {}) {
...
+        .replace(/^v/, '')
+        .split('.');
+      releaseVersion = `${major}.${minor}.${parseInt(patch) + 1}`;
```
richieforeman: > parseInt(undefined) === NaN
> NaN + 1 === NaN

I worry that if `patch` is undefined (which seems potentially likely? Maybe I'm paranoid or just don't understand enough) -- you could end up with a release version of `1.6.NaN`.

Something like this is probably safer (or I'm over analyzing it?):

```
(Number.isFinite(parseInt(undefined)) || 0) + 1
```

scripts/get-release-version.js
```diff
@@ -71,6 +71,33 @@ export function getVersion(options = {}) {
...
+      npmTag = 'latest';
+    } else {
+      // patchFrom === 'preview'
+      previousReleaseTag = getLatestTag('contains("preview")');
+      const [version, prerelease] = previousReleaseTag
+        .replace(/^v/, '')
+        .split('-');
+      const [major, minor, patch] = version.split('.');
```
richieforeman: Similar concern here -- Maybe there's an opportunity to extract into a helper?

scripts/lint.js
```diff
@@ -0,0 +1,150 @@
...
+      env.PATH = `${env.PATH}:${process.env.HOME}/.local/bin`;
+    }
+    execSync(command, { stdio, env });
+    return true;
+  } catch (_e) {
+    return false;
+  }
+}
+
+function setupLinters() {
```
richieforeman, line 116: super nit: Declare `:void` return type on this func.

scripts/lint.js
```diff
@@ -0,0 +1,150 @@
...
+      actionlint: 'darwin_arm64',
+      shellcheck: 'darwin.aarch64',
+    };
+  }
+  throw new Error(`Unsupported platform/architecture: ${platform}/${arch}`);
+}
+
+const platformArch = getPlatformArch();
+
+const LINTERS = {
```
richieforeman, line 57: Consider one of the following:

- Give this object a structured type

```
interface Linter {
  check: string;
  installer: string;
  run: string;
}

interface Linters {
  actionlint: Linter;
  whateverlint: Linter;
  etcetc: Linter;
}
```

OR

- define this object with `as const`

`const LINTER = {....} as const;`

OR 

(Best option)

Do both

`const LINTER: Linters  = {...} as const satisfies Linters;`

scripts/lint.js
```diff
@@ -0,0 +1,150 @@
...
+    `,
+  },
+  yamllint: {
+    check: 'command -v yamllint',
+    installer: `pip3 install --user "yamllint==${YAMLLINT_VERSION}"`,
+    run: "git ls-files | grep -E '\\.(yaml|yml)' | xargs yamllint --format github",
+  },
+};
+
+function runCommand(command, stdio = 'inherit') {
```
richieforeman, line 100: super nit: Give this a return type of `boolean`

scripts/lint.js
```diff
@@ -0,0 +1,150 @@
...
+import { tmpdir } from 'node:os';
+import { join } from 'node:path';
+
+const ACTIONLINT_VERSION = '1.7.7';
+const SHELLCHECK_VERSION = '0.11.0';
+const YAMLLINT_VERSION = '1.35.1';
+
+const TEMP_DIR = join(tmpdir(), 'gemini-cli-linters');
+
+function getPlatformArch() {
```
richieforeman, line 20: Nit: 

Please define a type interface for this func

```
interface PlatformArch {
  actionlint: string;
  spellcheck: string;
}
```

.github/actions/run-tests/action.yml
```diff
@@ -0,0 +1,24 @@
...
+    default: 'false'
+  gemini_api_key:
+    description: 'The API key for running integration tests.'
+    required: true
+
+runs:
+  using: 'composite'
+  steps:
+    - name: 'Run Tests'
+      if: "inputs.force_skip_tests != 'true'"
```
richieforeman, line 17: You know better than me here -- but does github actions always pass this as `(string literal) 'true'`. I'd have maybe expected this to be `(bool) true`.  Is there an option for us to have this be a proper boolean?

richieforeman: I believe that `inputs` arguments can accept a datatype, but the default is `string`.

I'll add a comment above.

ref: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#example-of-onworkflow_dispatchinputs

.github/actions/run-tests/action.yml
```diff
@@ -0,0 +1,24 @@
+name: 'Run Tests'
+description: 'Runs the preflight checks and integration tests.'
+
+inputs:
+  force_skip_tests:
+    description: 'Whether to force skip the tests.'
+    required: false
+    default: 'false'
```
richieforeman, line 8: Can we:

- Add `type: boolean`
- adjust default `'false'` -> `false`

.github/workflows/nightly-release.yml
```diff
@@ -10,6 +10,11 @@ on:
         required: true
         type: 'boolean'
         default: true
+      force_skip_tests:
+        description: 'Select to skip the "Run Tests" step in testing. Prod releases should run tests'
+        required: false
+        type: 'boolean'
```
richieforeman, line 16: Super nit: the docs seem to have `type: boolean` unquoted.  Please unquote.

.github/workflows/patch-release.yml
```diff
@@ -0,0 +1,105 @@
+name: 'Patch Release'
+
+on:
+  workflow_dispatch:
+    inputs:
+      type:
+        description: 'The type of release to patch from.'
+        required: true
+        type: 'choice'
```
richieforeman, line 9: Docs seem to have type unquoted, please unquote.

ref: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#example-of-onworkflow_dispatchinputs

.github/workflows/patch-release.yml
mattKorwel: WE can do a clean up pass after this.

scripts/get-release-version.js
mattKorwel: fixed.

scripts/get-release-version.js
mattKorwel: fixed.

scripts/lint.js
mattKorwel: fixed.

scripts/lint.js
mattKorwel: I love the high bar. This is a entirely unused helper function. For now I'll leave as is but we can clean up once the actions are working if we'd like. thanks!

.github/actions/run-tests/action.yml
mattKorwel: fixed.

.github/workflows/nightly-release.yml
mattKorwel: Unquoted there fails the linter.

.github/actions/run-tests/action.yml
richieforeman: This appears to have not been done. Let me know if I was offbase on this comment (purely for my own learning).

bdmorgan (approved):
LGTM

.github/workflows/nightly-release.yml
```diff
@@ -29,6 +34,12 @@ jobs:
       - name: 'Install Dependencies'
         run: 'npm ci'
 
+      - name: 'Run Tests'
+        uses: './.github/actions/run-tests'
```
richieforeman, line 38: Unless I'm mistaken, I don't think you can use the `./.github` path like this.

I think this needs to be `.github/actions/run-tests`

(etc etc etc, whole file)

I ran this and it seemed to choke: https://github.com/google-gemini/gemini-cli/actions/runs/17656496123

PR #8197, opened 2025-09-10
fix(cli): restart gemini cli if untrusted folder option selected

hritan:
## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

### DIFF
--- a/packages/cli/src/ui/AppContainer.tsx
+++ b/packages/cli/src/ui/AppContainer.tsx
@@ -4,6 +4,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+import { spawn } from 'node:child_process';
 import { useMemo, useState, useCallback, useEffect, useRef } from 'react';
 import { type DOMElement, measureElement } from 'ink';
 import { App } from './App.js';
@@ -121,6 +122,7 @@ export const AppContainer = (props: AppContainerProps) => {
   const [isTrustedFolder, setIsTrustedFolder] = useState<boolean | undefined>(
     config.isTrustedFolder(),
   );
+  const [isRestarting, setIsRestarting] = useState(false);
 
   // Helper to determine the effective model, considering the fallback state.
   const getEffectiveModel = useCallback(() => {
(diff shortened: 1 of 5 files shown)


packages/cli/src/ui/hooks/useFolderTrust.ts
```diff
@@ -53,9 +54,12 @@ export const useFolderTrust = (
         case FolderTrustChoice.TRUST_PARENT:
           trustLevel = TrustLevel.TRUST_PARENT;
           break;
-        case FolderTrustChoice.DO_NOT_TRUST:
+        case FolderTrustChoice.DO_NOT_TRUST: {
           trustLevel = TrustLevel.DO_NOT_TRUST;
-          break;
+          trustedFolders.setValue(cwd, trustLevel);
+          onRestart();
```
shrutip90, line 60: We already have a check to see if we need restart and setIsRestarting below.

packages/cli/src/ui/AppContainer.tsx
```diff
@@ -1126,6 +1147,10 @@ Logging in with Google... Please restart Gemini CLI to continue.
     ],
   );
 
+  if (isRestarting) {
```
shrutip90, line 1151: Why do we need this?

packages/cli/src/ui/hooks/useFolderTrust.ts
hritan: yeah, this function was an explicit function telling AppContainer to initialize restart.

packages/cli/src/ui/AppContainer.tsx
hritan: This will remove the parent's UI from the terminal while the child process starts.

Issue #7727, opened 2025-09-04
Add basic extension permissioning and consent flow


chrstnb:
Determine what the extension is trying to do from the config and add a consent flow for users.

e.g.,

"this extension is trying to use a local MCP server, that means it has access to XYZ, proceed?"

PR #8208, opened 2025-09-10
feat(extension) - Add permission prompt for when user installs a local extension with mcpservers

shishu314:
## TLDR

Adds a user permission prompt before they install a local extension with mcp servers

## Dive Deeper

N/A

## Reviewer Test Plan

1. Create a new local extension using `gemini extensions new --path=path/to/directory --type=mcp-server`
1. Update the `gemini-extension.json` to have local mcpservers with descriptions
1. Run `gemini extensions install --path extensionPath`
1. Check that you are prompted for permission to install
1. Run `gemini extensions uninstall extension-name` as needed

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | x  | ❓  | ❓  |
| npx      | x  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

makes progress on #7727


packages/cli/src/config/extension.ts
```diff
@@ -443,6 +463,25 @@ export async function installExtension(
         );
       }
 
+      if (installMetadata.type === 'local' && newExtensionConfig.mcpServers) {
```
chrstnb: We should do this for all installMetadata types! For context, that denotes where the definition of the extension is written, (e.g., is it on github or is it a local directory), but won't tell us anything about the MCP servers. 

The distinction I was referring to with local MCP servers is that some MCP servers are running locally, while others hit a remote endpoint. 

I think this is a good start, and in a follow-up we should tell the user if an MCP server is running locally or remotely.

chrstnb (approved):

chrstnb:
One comment, please remove the installMetadata type check before submitting!

PR #8225, opened 2025-09-10
Fix compression issues

### LINKED ISSUE #3709: Failed to compress chat history

scidomino:
## TLDR

Fixes it to `compress` always compresses *at least* 70% of the history rather than *at most*. Also simplified how we create the summary so there is no possibility of a tool call.

## Dive Deeper

This removes `GeminiChat.sendMessage()` which was only used by compress but which had a lot of complicated support for stuff that compress definitely does not need (like function calling).

## Reviewer Test Plan

Ask: `How are you? Take a look at @packages/core. Isn't it beautiful?`
Then: `/compress`

Verify that the token count goes down by a lot.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | x  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes #3709

### DIFF
--- a/packages/core/src/core/geminiChat.ts
+++ b/packages/core/src/core/geminiChat.ts
@@ -182,138 +182,6 @@ export class GeminiChat {
   setSystemInstruction(sysInstr: string) {
     this.generationConfig.systemInstruction = sysInstr;
   }
-  /**
-   * Sends a message to the model and returns the response.
-   *
-   * @remarks
-   * This method will wait for the previous message to be processed before
-   * sending the next message.
-   *
-   * @see {@link Chat#sendMessageStream} for streaming method.
-   * @param params - parameters for sending messages within a chat session.
-   * @returns The model's response.
-   *
-   * @example
-   * ```ts
-   * const chat = ai.chats.create({model: 'gemini-2.0-flash'});
-   * const response = await chat.sendMessage({
-   * message: 'Why is the sky blue?'
-   * });
-   * console.log(response.text);
-   * ```
-   */
-  async sendMessage(
-    params: SendMessageParameters,
-    prompt_id: string,
-  ): Promise<GenerateContentResponse> {
-    await this.sendPromise;
-    const userContent = createUserContent(params.message);
-
-    // Record user input - capture complete message with all parts (text, files, images, etc.)
-    // but skip recording function responses (tool call results) as they should be stored in tool call records
-    if (!isFunctionResponse(userContent)) {
-      const userMessage = Array.isArray(params.message)
-        ? params.message
-        : [params.message];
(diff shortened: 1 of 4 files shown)


packages/core/src/core/geminiChat.ts
```diff
@@ -704,35 +572,18 @@ export class GeminiChat {
...
   }
 
-  private recordHistory(
-    userInput: Content,
-    modelOutput: Content[],
-    automaticFunctionCallingHistory?: Content[],
-  ) {
+  private recordHistory(userInput: Content, modelOutput: Content[]) {
     // Part 1: Handle the user's turn.
-    if (
```
anj-s, line 713: why don't we need this anymore?

scidomino: automaticFunctionCallingHistory was only passed by sendMessage() which we are removing.

jacob314 (commented):
The diff in geminiChat.ts and test.ts seems a bit larger than needed. if those are intentional let me know and I'll review.

On the plus side I am now finally able to compress my chat history after user `@packages/cli` to fill the context window.

scidomino:
> The diff in geminiChat.ts and test.ts seems a bit larger than needed. if those are intentional let me know and I'll review.
> 
> On the plus side I am now finally able to compress my chat history after user `@packages/cli` to fill the context window.

So really, this fixes two different bugs:

1. off-by-one error causing it to not compress as much as it should.
2. sometimes compress would return a tool call instead of a summary which would break everything.

Fixing 2. removed the last usage of sendMessage() which required deleting a bunch of code. And to be clear, I did that manually (not with GCLI).

anj-s (approved):

PR #8230, opened 2025-09-10
Typo fix in releases.md

ingenthr:
Needed an "e" btween the "b" and "t".

## TLDR

Fixes a typo. 

## Reviewer Test Plan

Spell checking doesn't seem to be in the current test plan, so I do not think new tests would be needed for this change.


cornmander:
Thanks!

cornmander (approved):

PR #8231, opened 2025-09-10
feat(ui): Add confirmation dialog for disabling loop detection for current session

SandyTao520:
## TLDR

This change introduces a user confirmation dialog when a potential infinite loop is detected. Previously, the CLI would halt the request and there is no way to stop false positives. Now, the user is prompted with a choice: either halt the request (the default) or disable loop detection for the current session and proceed.

Confirmation dialog:
image

After selecting no for the first prompt and yes the second time:
image

## Dive Deeper

The core goal is to give users more control when the `LoopDetectionService` flags a potential infinite loop. The previous behavior was abrupt and could interrupt a valid, albeit repetitive, workflow.

This PR implements the following:
- A new `LoopDetectionConfirmation.tsx` component is created to present the dialog to the user.
- The `useGeminiStream` hook now listens for the `LoopDetected` event and, instead of just logging a message, it triggers the new confirmation dialog.
- Based on the user's selection ("keep" or "disable"), the hook either halts the stream or calls a new `disableForSession()` method on the `LoopDetectionService`.
- The `LoopDetectionService` was updated to include the `disableForSession` flag, which bypasses all loop checks for the remainder of the CLI session if set.

This provides a much better user experience by preventing the CLI from being overly aggressive in halting user requests, empowering the user to decide if the repetitive behavior is intentional.

## Reviewer Test Plan

To test this, you'll need to trigger the loop detection mechanism.

1.  Start the CLI.
2.  It can be difficult to create a prompt that reliably causes a loop. A more deterministic way to test is to temporarily modify the `TOOL_CALL_LOOP_THRESHOLD` in `packages/core/src/services/loopDetectionService.ts` to a low number, like `2`.
3.  With the threshold lowered, ask the model a prompt that uses the same tool call multiple times, for example: "list the files in the current directory, then tell me what you see, then list them again".
4.  **Verify:** A yellow warning dialog should appear, asking if you want to disable loop detection.
5.  Press `escape` or select the "No, keep loop detection..." option.
6.  **Verify:** The request should be halted, and an informational message should appear stating that a loop was detected and the request was stopped.
7.  Run the same prompt again to trigger the dialog.
8.  This time, select "Yes, disable loop detection...".
9.  **Verify:** An informational message should appear stating that loop detection is disabled for the session, and the model should attempt to continue processing your request. Subsequent repetitive tool calls in the same session should no longer trigger the dialog.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ✅  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |
```

### DIFF
--- a/packages/cli/src/ui/AppContainer.tsx
+++ b/packages/cli/src/ui/AppContainer.tsx
@@ -897,37 +897,20 @@ Logging in with Google... Please restart Gemini CLI to continue.
 
   const nightly = props.version.includes('nightly');
 
-  const dialogsVisible = useMemo(
-    () =>
-      showWorkspaceMigrationDialog ||
-      shouldShowIdePrompt ||
-      isFolderTrustDialogOpen ||
-      !!shellConfirmationRequest ||
-      !!confirmationRequest ||
-      !!loopDetectionConfirmationRequest ||
-      isThemeDialogOpen ||
-      isSettingsDialogOpen ||
-      isAuthenticating ||
-      isAuthDialogOpen ||
-      isEditorDialogOpen ||
-      showPrivacyNotice ||
-      !!proQuotaRequest,
-    [
-      showWorkspaceMigrationDialog,
-      shouldShowIdePrompt,
-      isFolderTrustDialogOpen,
-      shellConfirmationRequest,
-      confirmationRequest,
-      loopDetectionConfirmationRequest,
-      isThemeDialogOpen,
-      isSettingsDialogOpen,
-      isAuthenticating,
-      isAuthDialogOpen,
-      isEditorDialogOpen,
-      showPrivacyNotice,
-      proQuotaRequest,
-    ],
-  );
+  const dialogsVisible =
+    showWorkspaceMigrationDialog ||
+    shouldShowIdePrompt ||
+    isFolderTrustDialogOpen ||
+    !!shellConfirmationRequest ||
+    !!confirmationRequest ||
+    !!loopDetectionConfirmationRequest ||
+    isThemeDialogOpen ||
+    isSettingsDialogOpen ||
+    isAuthenticating ||
+    isAuthDialogOpen ||
+    isEditorDialogOpen ||
+    showPrivacyNotice ||
+    !!proQuotaRequest;
 
   const pendingHistoryItems = useMemo(
     () => [...pendingSlashCommandHistoryItems, ...pendingGeminiHistoryItems],
--- a/packages/cli/src/ui/AppContainer.tsx
+++ b/packages/cli/src/ui/AppContainer.tsx
@@ -505,6 +505,7 @@ Logging in with Google... Please restart Gemini CLI to continue.
     pendingHistoryItems: pendingGeminiHistoryItems,
     thought,
     cancelOngoingRequest,
+    loopDetectionConfirmationRequest,
   } = useGeminiStream(
     config.getGeminiClient(),
     historyManager.history,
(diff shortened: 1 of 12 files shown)

### CODE REVIEW

**Comment by jacob314** [comment]
Can you pick a warning icon that does trigger ink character measuring issues?

File: packages/cli/src/ui/components/LoopDetectionConfirmation.tsx [hunk 1]
```diff
@@ -0,0 +1,75 @@
...
+          userSelection: 'keep',
+        });
+      }
+    },
+    { isActive: true },
+  );
+
+  const OPTIONS: Array<RadioSelectItem<LoopDetectionConfirmationResult>> = [
+    {
+      label: 'No, keep loop detection enabled and stop (esc)',
```
jacob314: lets wordsmith these labels.

packages/cli/src/ui/components/LoopDetectionConfirmation.tsx
```diff
@@ -0,0 +1,75 @@
...
+        userSelection: 'disable',
+      },
+    },
+  ];
+
+  return (
+    <Box
+      flexDirection="column"
+      borderStyle="round"
+      borderColor="yellow"
```
jacob314: use colors from the theme rather than hardcoding colors. Unfortunate GEMINI isn't doing it automatically.

packages/cli/src/ui/components/LoopDetectionConfirmation.tsx
```diff
@@ -0,0 +1,75 @@
...
+      padding={1}
+      width="100%"
+      marginLeft={1}
+    >
+      <Box marginBottom={1} flexDirection="column">
+        <Text>
+          <Text color="yellow">⚠️ </Text>
+          Loop Detection Alert
+        </Text>
+        <Text dimColor>
```
jacob314: remove dimColor and use the theme

packages/cli/src/ui/components/LoopDetectionConfirmation.tsx
```diff
@@ -0,0 +1,75 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
```
jacob314, line 5: add a single snapshot golden test for this dialog

packages/cli/src/ui/hooks/useGeminiStream.ts
```diff
@@ -588,15 +594,38 @@ export const useGeminiStream = (
...
+            type: 'info',
+            text: `Loop detection has been disabled for this session. Please try your request again.`,
+          },
+          Date.now(),
+        );
+      } else {
+        addItem(
+          {
+            type: 'info',
+            text: `A potential loop was detected. This can happen due to repetitive tool calls or other model behavior. The request has been halted. Loop detection remains enabled.`,
```
jacob314: nit:  remove the "Loop detection remains enabled." sentence.

packages/cli/src/ui/AppContainer.tsx
```diff
@@ -903,6 +904,7 @@ Logging in with Google... Please restart Gemini CLI to continue.
       isFolderTrustDialogOpen ||
```
jacob314: not a new issue but is this useMemo really useful? seems like it is memoizing something trivial to compute. Perhaps strip it and just compute whether dialogsVisible. Will reduce lines and be just as efficient.

jacob314 (commented):
Approved once the UX concerns are addressed.

jacob314 (approved):
lgtm

PR #8266, opened 2025-09-11
feat(ide): Check for IDE diffing capabilities before opening diffs

skeshive:
## TLDR

This PR introduces a tool discovery mechanism to the IDE client. Previously, the CLI would attempt to open a diff view in the IDE for any file modification tool (`edit`, `write-file`, etc.) simply if it was in "IDE mode" and connected. This change makes the CLI smarter by first checking if the connected IDE actually supports the `openDiff` and `closeDiff` capabilities. This prevents the CLI from attempting an unsupported operation and only presents the diffing confirmation flow to the user when the connected IDE can handle it.

## Linked issues / bugs
#4800

### DIFF
--- a/packages/core/src/ide/ide-client.ts
+++ b/packages/core/src/ide/ide-client.ts
@@ -337,10 +337,8 @@ export class IdeClient {
   }
 
   isDiffingEnabled(): boolean {
-    if (!this.client) {
-      return false;
-    }
     return (
+      !!this.client &&
       this.state.status === IDEConnectionStatus.Connected &&
       this.availableTools.includes('openDiff') &&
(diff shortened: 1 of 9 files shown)


packages/core/src/ide/ide-client.ts
```diff
@@ -664,6 +714,7 @@ export class IdeClient {
       });
       await this.client.connect(transport);
       this.registerClientHandlers();
```
cornmander, line 716: It feels bad that we're duplicating so much code between establishStdioConnection() and establishHttpConnection() when they only differ in the transport. Can we merge these?

cornmander (approved):

PR #8270, opened 2025-09-11
fix(ide): Correct IDE client temp dir and port matching

skeshive:
## TLDR

This pull request addresses two issues in the IDE client:
1.  It changes the temporary directory used for storing IDE server connection files from `.../.gemini/ide` to `.../gemini/ide`, removing the leading dot.
2.  It fixes a bug in port matching where a strict equality check (`===`) would fail due to comparing a number (`content.port`) with a string (`portFromEnv`). The port is now cast to a string for a reliable comparison.

## Linked issues / bugs
#4800, #6848

### DIFF
--- a/packages/core/src/ide/ide-client.ts
+++ b/packages/core/src/ide/ide-client.ts
@@ -459,7 +459,7 @@ export class IdeClient {
       // exist.
     }
 
-    const portFileDir = path.join(os.tmpdir(), '.gemini', 'ide');
+    const portFileDir = path.join(os.tmpdir(), 'gemini', 'ide');
     let portFiles;
     try {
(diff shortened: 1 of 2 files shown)


packages/core/src/ide/ide-client.ts
```diff
@@ -521,7 +521,7 @@ export class IdeClient {
     const portFromEnv = this.getPortFromEnv();
     if (portFromEnv) {
       const matchingPort = validWorkspaces.find(
-        (content) => content.port === portFromEnv,
+        (content) => String(content.port) === portFromEnv,
```
cornmander, line 524: Can you add a test for confirming this is fixed?

cornmander (approved):

PR #8284, opened 2025-09-11
Fix license file.

joshualitt:
Simple fix of license file, we are not supposed to modify the license, including the template.


teeler (approved):

jacob314 (approved):
lgtm

PR #8301, opened 2025-09-11
Make a stateful extensions list component, with update statuses

jakemac53:
## TLDR

Instead of just adding an info item, `/extensions list` and `/extensions update` now both add a new ExtensionsList component.

This component renders the same stuff as before as well as the current update state of each extension.

Below is a quick demo of the new list/update behavior, with an update available:

https://github.com/user-attachments/assets/24b021fe-3145-4ead-af6e-807fe22529c2

There is another state here you never see "checking for updates", but its always past that state well before I can get to the list command.

## Dive Deeper

Adds a map to the UI state containing the current state of each extension, and plumbs through updating this state into all the update functions.

The ExtensionsList component reads this state object and updates its UI accordingly.

Note that only pending items will actually update their UI, regular history items are "static" and don't update unless you resize the window horizontally. So, when doing an update we actually add it as a pending item, until the updates all complete, because pending items do continuously update their UI.

I also updated a whole bunch of places to not continuously re-load extensions, but always use the actual originally loaded extensions. This shows a more accurate view of the extensions as they are actually available in the current session, and it avoids extra work.

## Reviewer Test Plan

Try `/extensions list` while having an extension that has an update available (push a commit to the repo), then try `/extensions update  --all` to see it update, you should see the text show "updating" and then "updated, needs restart".

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Part of #7009


chrstnb:
This looks really good! This UX makes a ton of sense to me. :shipit:

chrstnb (approved):

mattKorwel (approved):

PR #8305, opened 2025-09-11
feat(cli) Custom Commands work in Non-Interactive/Headless Mode

capachino:
## TLDR

Adds the ability to run custom commands in non-interactive mode #5435

## Dive Deeper

- Does not support MCP prompts yet (though I can add this soon as a follow up)
- Does not support built-in commands. I think this requires some discussion as many of them do not make sense in non-interactive mode.
- Expects the command invocation will result in a prompt. Confirmations and any other processing result will fail.
- If the "slash command" is not registered or unknown, proceeds as a regular prompt. I'm open to failing in this case if you think that's better.

## Reviewer Test Plan

Try a custom command in non-interactive mode e.g.

`node /gemini-cli/packages/cli -p "/security:analyze"`

- Expected failure if the command requires confirmation
- Expected success if a shell command confirmation is needed AND yolo mode
- Expected success with command args
- Expected success if it's not a registered command (i.e. proceed with the raw prompt)
- No regressions with a command-less prompt

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | X  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | X  | -   | -   |

## Linked issues / bugs

Makes progress on #5435

### DIFF
--- a/packages/cli/src/nonInteractiveCli.ts
+++ b/packages/cli/src/nonInteractiveCli.ts
@@ -5,19 +5,13 @@
  */
 
 import type { Config, ToolCallRequestInfo } from '@google/gemini-cli-core';
-import { CommandService } from './services/CommandService.js';
-import { FileCommandLoader } from './services/FileCommandLoader.js';
-import { type CommandContext } from './ui/commands/types.js';
 import { isSlashCommand } from './ui/utils/commandUtils.js';
-import type { SessionStatsState } from './ui/contexts/SessionContext.js';
 import type { LoadedSettings } from './config/settings.js';
-import { createNoOpUI } from './ui/commands/noOpUi.js';
 import {
   executeToolCall,
   shutdownTelemetry,
   isTelemetrySdkInitialized,
   GeminiEventType,
-  Logger,
   FatalInputError,
   promptIdContext,
   OutputFormat,
@@ -25,9 +19,9 @@ import {
   uiTelemetryService,
 } from '@google/gemini-cli-core';
 
-import type { Content, Part, PartListUnion } from '@google/genai';
-import { parseSlashCommand } from './utils/commands.js';
+import type { Content, Part } from '@google/genai';
 
+import { handleSlashCommand } from './nonInteractiveCliCommands.js';
 import { ConsolePatcher } from './ui/utils/ConsolePatcher.js';
 import { handleAtCommand } from './ui/hooks/atCommandProcessor.js';
 import {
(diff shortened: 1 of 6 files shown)


jerop (commented):
What do you think of splitting this into two parts:
1. Refactor existing functionality to create a reusable packages/cli/src/utils/commands.ts
2. Implement custom command support in non-interactive mode?

cc @abhipatel12 @allenhutchison

allenhutchison:
We haven't really had a discussion yet on how we want to structure the CLI interface for custom commands (i.e. whether we include them in the positional arg `gemini "/command args for command"` or if we do the alternative `gemini --command="args for command"`

I continue to argue for the latter because it doesn't carry over the TUI ux affordance "/" which is unnecessary in the CLI command environment because we have flags. In other words, why invent a new way to do a flag on the command line when we have a perfect good way to do flags on the command line that is used by every other CLI application out there?

Second, I would like this PR to focus on custom commands. Take the @ file processing out for a different PR. Again gemini cli has the ability to take context on STDIN and files can be passed to gemini CLI that way, which is much more consistent with standard command line ux than including the TUI ux affordance "@" to indicate that on the command line. Either way, let's keep this PR focused on one thing.

Finally let's make sure we are including the issue for this #5435 in the PR description and please update the PR description to be more descriptive overall.

capachino:
> What do you think of splitting this into two parts:
> 
> 1. Refactor existing functionality to create a reusable packages/cli/src/utils/commands.ts
> 2. Implement custom command support in non-interactive mode?
> 
> cc @abhipatel12 @allenhutchison

Good idea. Created PR #8381

capachino:
> We haven't really had a discussion yet on how we want to structure the CLI interface for custom commands (i.e. whether we include them in the positional arg `gemini "/command args for command"` or if we do the alternative `gemini --command="args for command"`
> 
> I continue to argue for the latter because it doesn't carry over the TUI ux affordance "/" which is unnecessary in the CLI command environment because we have flags. In other words, why invent a new way to do a flag on the command line when we have a perfect good way to do flags on the command line that is used by every other CLI application out there?
>

From an engineering perspective I like the separate `command` flag as it better separates concerns too.

However, the feedback seems clearly in favor of specifying it in the prompt (at least 6 votes for that). I totally understand, because I tried that myself and I think the intuition will become stronger over time as similar tools are going with that approach.

> Second, I would like this PR to focus on custom commands. Take the @ file processing out for a different PR. Again gemini cli has the ability to take context on STDIN and files can be passed to gemini CLI that way, which is much more consistent with standard command line ux than including the TUI ux affordance "@" to indicate that on the command line. Either way, let's keep this PR focused on one thing.
> 

Sorry for the confusing diff, that is actually existing functionality.

> Finally let's make sure we are including the issue for this #5435 in the PR description and please update the PR description to be more descriptive overall.

Done.

FYI I split the refactoring for the command parsing to PR #8381 

I'm not sure if I rebased this PR correctly (apologies I'm still learning Git), but I do see separate commits in this PR at least.

packages/cli/src/ui/commands/noOpUi.ts
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+import type { CommandContext } from './types.js';
+
+/**
+ * Creates a UI context object with no-op functions.
+ * Useful for non-interactive environments where UI operations
+ * are not applicable.
+ */
+export function createNoOpUI(): CommandContext['ui'] {
```
allenhutchison: This makes me wonder if we should split out the non-interactive ui into a new directory. Maybe in ui/noninteractive/... and then we have a clear place for logic that is only applicable in non-interactive mode.

packages/cli/src/nonInteractiveCli.ts
```diff
@@ -27,8 +37,92 @@ import {
...
 
+/**
+ * Processes a slash command in a non-interactive environment.
+ *
+ * @returns A Promise that resolves to `PartListUnion` if a valid command is
+ *   found and results in a prompt, or `undefined` otherwise.
+ * @throws {FatalInputError} if the command requires user confirmation, which
+ *   is not supported in non-interactive mode.
+ */
+const handleSlashCommand = async (
```
allenhutchison: Can you extract this to another file?

packages/cli/src/nonInteractiveCli.ts
```diff
@@ -27,8 +37,92 @@ import {
...
+  rawQuery: string,
+  abortController: AbortController,
+  config: Config,
+  settings: LoadedSettings,
+): Promise<PartListUnion | undefined> => {
+  const trimmed = rawQuery.trim();
+  if (!trimmed.startsWith('/') && !trimmed.startsWith('?')) {
```
allenhutchison: Do we still support the `?` command? I think that was removed at some point. In 0.7.0-nightly.20250912.68035591 it's not supported in interactive mode.

packages/cli/src/nonInteractiveCli.ts
```diff
@@ -27,8 +37,92 @@ import {
...
+
+      const result = await commandToExecute.action(context, args);
+
+      if (result) {
+        switch (result.type) {
+          case 'submit_prompt':
+            return result.content;
+          case 'confirm_shell_commands':
+            throw new FatalInputError(
+              'Exiting due to a confirmation prompt requested by the command.',
```
allenhutchison: We may want a pointer to docs here on how to set the appropriate settings to allow these shell commands. Do you have access to the actual commands that it tried to run?

allenhutchison:
Good progress. A few comments to think about here. Ping me when you are ready for another review.

packages/cli/src/ui/commands/noOpUi.ts
capachino: Done. Good idea, I think we should organize all non-interactive UI implementations here.

packages/cli/src/nonInteractiveCli.ts
capachino: Done.

packages/cli/src/nonInteractiveCli.ts
capachino: Good catch. Removed the `?` check.

packages/cli/src/nonInteractiveCli.ts
capachino: I believe this can't actually happen at the moment, because `ShellTool` calls require YOLO mode.

1. For custom commands, confirmation requests occur when the prompt [contains a shell tool expansion](https://github.com/google-gemini/gemini-cli/blob/main/packages/cli/src/services/prompt-processors/shellProcessor.ts#L144).
2. `ShellTool` is [automatically excluded](https://github.com/google-gemini/gemini-cli/blob/main/packages/cli/src/config/config.ts#L497) for non-interactive CLI unless YOLO mode is active.

Should I perhaps leave a comment saying this can't happen at the moment because of the non-granular permissions, but handling anyway for completeness? Or maybe just leave it out.

FYI, I'm also thinking of creating an issue to improve non-interactive mode tool permissions with an initial proposal and to collect feedback.

capachino: Thanks gemini-code-assist. Can you suggest an appropriate comment that I can add here that incorporates my explanation?

capachino: Done. Added a comment.

allenhutchison:
I took another pass this afternoon. This looks great, works well in my testing, and you've addressed all of my comments. I think we can bring this in.

allenhutchison (approved):

jerop:
trying to kick off the tests again

Issue #1459, opened 2025-06-25
Support MCP resources and prompt


brianhuster:
As I can see, right now, gemini-cli only supports MCP tools. I would love to see it support MCP [resources](https://modelcontextprotocol.io/docs/concepts/resources) and [prompt](https://modelcontextprotocol.io/docs/concepts/prompts) as well

heartyguy:
I'd like to work on this issue! 🚀

  **Proposed Approach**

  After reviewing other MCP implementations (Roo Code, Claude Desktop), I noticed they
  focus solely on MCP resources without prompts. I propose following the same approach -
   implementing only MCP resources support for now.

  **Technical Details**
  - Add `ResourceRegistry` to manage resources from multiple MCP servers
  - Implement `ReadResourceTool` and `ListResourcesTool`
  - Extend MCP client to discover resources via `listResources()` and
  `listResourceTemplates()`
  - Resources will be available through Gemini's existing tool system

  This keeps the implementation focused and aligns with established patterns in the
  ecosystem.

Expected timeline: 1 day

Let me know if this approach looks good or if you have any suggestions\!

chrstnb:
Hey @heartyguy! Have you had a chance to do this? If not, someone on our team was going to grab it!

ashwinpvg:
> Hey [@heartyguy](https://github.com/heartyguy)! Have you had a chance to do this? If not, someone on our team was going to grab it!

Looks like there is a PR out already :) https://github.com/google-gemini/gemini-cli/pull/1789

ryanbaumann:
This is great looking forward to this PR and new release. We need /resources for our MCP servers!

requiem240sx:
I setup my own custom MCP server and was let down when I realized none of the prompts or resources are showing up.

Question: In most other IDE's (cluade code, vscode) you can simply type `/` and it auto shows all available MCP Prompts for you to execute by command (or load in that prompt). Prompts then call the tools to build workflows.  Will this MR allow for this `/` style discover of prompts? Or will that be a new issue we need to add in?

I see the [MR](https://github.com/google-gemini/gemini-cli/pull/1789#issuecomment-3120043693) is close, looks like testing is needed now.

haroldmciver-go:
Gemini CLI now supports MCP prompts! They can be discovered and displayed using `/mcp` and the prompts can be directly called using `/<prompt-name> --args=""`

more info here: https://github.com/google-gemini/gemini-cli/pull/4828

[...]

PR #8340, opened 2025-09-12
WIP: add support for MCP resources

jackwotherspoon:
## TLDR

Support for [MCP resources](https://modelcontextprotocol.io/specification/2025-06-18/server/resources) using `@server:uri `syntax in prompts.

This PR does not add the following (for follow-up PRs):
- MCP resource templates
- Resource usage with `@` in context file (`GEMINI/md`)
 

https://github.com/user-attachments/assets/25d20049-896d-4efb-8191-7d9f5479071b

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


jackwotherspoon:
@jakemac53 going to clean this up a little but curious your thoughts here

jackwotherspoon:
Related to https://github.com/google-gemini/gemini-cli/issues/1459

packages/cli/src/ui/hooks/atCommandProcessor.ts
```diff
@@ -152,6 +153,7 @@ export async function handleAtCommand({
 
   const toolRegistry = config.getToolRegistry();
   const readManyFilesTool = toolRegistry.getTool('read_many_files');
+  const readMcpResourceTool = toolRegistry.getTool('read_mcp_resource');
   const globTool = toolRegistry.getTool('glob');
 
   if (!readManyFilesTool) {
```
jakemac53, line 159: This check should possibly be moved later? We could support reading resources for instance without the readManyFilesTool.

And in particular somebody might intentionally disable that tool and use an MCP server which exposes file resources instead of it.

packages/cli/src/ui/hooks/atCommandProcessor.ts
```diff
@@ -188,6 +190,25 @@ export async function handleAtCommand({
       return { processedQuery: null, shouldProceed: false };
     }
 
+    // Check if this is an MCP resource import (contains colon)
+    const colonIndex = pathName.indexOf(':');
+    if (colonIndex > 0 && colonIndex < pathName.length - 1) {
```
jakemac53, line 195: I think there is an open question here as to whether we should prefix these with the server name or not.

In general, resource URIs should be unique. If two servers are exposing the same resource URI, that is always going to be a bit awkward.

Consider for instance http URIs in particular, according to the spec these should be fetched _directly_ by the client, and [not through the MCP server](https://modelcontextprotocol.io/specification/draft/server/resources#https%3A%2F%2F). So, these ones should be a special case most likely.

It also results in a non-standard syntax for how to refer to these things, including hard coding of server names in MD files in the future (assuming we get support for that).

At the same time I think it has the advantage of allowing you to easily get auto complete for the resources scoped to a specific server which is useful. So, I am a bit torn on it.

packages/cli/src/ui/hooks/atCommandProcessor.ts
```diff
@@ -394,87 +418,166 @@ export async function handleAtCommand({
...
+        status: ToolCallStatus.Success,
+        resultDisplay: result.returnDisplay,
+        confirmationDetails: undefined,
+      };
+
+      toolCallDisplays.push(toolCallDisplay);
+
+      if (Array.isArray(result.llmContent)) {
+        const mcpResourceContentRegex =
+          /^--- MCP Resource: (.*?) ---\n\n([\s\S]*?)\n\n$/;
```
jakemac53, line 524: I don't know if it is useful for the LLM to know that these come from MCP? They should be really no different than any other resource that comes from a URI.

packages/cli/src/ui/hooks/useAtCompletion.ts
```diff
@@ -199,7 +199,7 @@ export function useAtCompletion(props: UseAtCompletionProps): void {
       try {
         const results = await fileSearch.current.search(state.pattern, {
           signal: controller.signal,
-          maxResults: MAX_SUGGESTIONS_TO_SHOW * 3,
+          maxResults: MAX_SUGGESTIONS_TO_SHOW * 2, // Reduced to make room for resources
```
jakemac53, line 202: Ideally we would just trim this down if we actually have resources to show.

Probably a larger refactor so feel free to file separately but some sort of more general autocomplete handler class which can take multiple autocomplete providers and merge their results based on relevance is really what we need here.

packages/cli/src/ui/hooks/useAtCompletion.ts
```diff
@@ -210,11 +210,70 @@ export function useAtCompletion(props: UseAtCompletionProps): void {
...
+        const resourceSuggestions: Suggestion[] = [];
+        if (config) {
+          try {
+            const resourceRegistry = config.getResourceRegistry();
+            if (resourceRegistry) {
+              let matchingResources;
+
+              // Check if pattern looks like server:resource format
+              const colonIndex = state.pattern.indexOf(':');
+              if (colonIndex > 0) {
```
jakemac53, line 229: Isn't this going to end up matching as soon as we have a scheme? I should be able to type `@<someScheme>://` and see all resources matching that scheme from any server for instance.

I think if we want to allow users to scope their autocomplete down to a single server, we need to figure out an unambiguous way to put the server name in front of the URI.

packages/core/src/resources/resource-registry.ts
```diff
@@ -0,0 +1,121 @@
...
+      );
+    }
+    this.resources.set(resourceKey, resource);
+  }
+
+  /**
+   * Returns an array of all registered and discovered resource instances.
+   */
+  getAllResources(): DiscoveredMCPResource[] {
+    return Array.from(this.resources.values()).sort((a, b) =>
```
jakemac53, line 38: Can we keep the sorted list? There can potentially be a lot of them depending on the server so creating and sorting a new list on each call could have pathological behavior.

packages/core/src/resources/resource-registry.ts
```diff
@@ -0,0 +1,121 @@
...
+        this.resources.delete(key);
+      }
+    }
+  }
+
+  /**
+   * Lists all available resource URIs for @ symbol parsing.
+   */
+  getAllResourceKeys(): string[] {
+    return Array.from(this.resources.keys());
```
jakemac53, line 101: Creating a new array every time here also is going to be potentially pretty wasteful, can we just return the keys directly?

packages/core/src/resources/resource-registry.ts
```diff
@@ -0,0 +1,121 @@
...
+   * Finds resources matching a pattern (used for @ symbol completion).
+   */
+  findResourcesMatching(pattern: string): DiscoveredMCPResource[] {
+    const results: DiscoveredMCPResource[] = [];
+    for (const resource of this.resources.values()) {
+      const resourceKey = `${resource.serverName}:${resource.uri}`;
+      if (
+        resourceKey.includes(pattern) ||
+        resource.name.includes(pattern) ||
+        resource.uri.includes(pattern)
```
jakemac53, line 114: They key includes the URI so this should be unnecessary right?

packages/core/src/resources/resource-registry.ts
```diff
@@ -0,0 +1,121 @@
...
+      ) {
+        results.push(resource);
+      }
+    }
+    return results.sort((a, b) => a.name.localeCompare(b.name));
```
jakemac53, line 119: Ideally this sort would be based more off of relevance than just alphabetical, but this is probably fine. I would just file an issue or something for a relevance based search.

There must be for instance an ngram library we could use which would be pretty simple, effective, and fast here.

packages/core/src/tools/mcp-client.ts
```diff
@@ -750,6 +765,87 @@ export async function discoverPrompts(
...
+export async function discoverResources(
+  mcpServerName: string,
+  mcpClient: Client,
+  resourceRegistry: ResourceRegistry,
+): Promise<Resource[]> {
+  try {
+    // Only request resources if the server supports them.
+    if (mcpClient.getServerCapabilities()?.resources == null) return [];
+
+    const response = await mcpClient.request(
```
jakemac53, line 785: This is a paginated response, we need to make sure to request all the resources.

Or possibly we max out at some number and give a warning if we exceed that, but we should request pages up to that number.

packages/core/src/tools/mcp-client.ts
```diff
@@ -750,6 +765,87 @@ export async function discoverPrompts(
...
+      { method: 'resources/read', params: { uri: resourceUri } },
+      ReadResourceResultSchema,
+    );
+
+    if (response.contents.length === 0) {
+      throw new Error(`No content found for resource: ${resourceUri}`);
+    }
+
+    // Return the first content item (MCP supports multiple, but we'll use the first one)
+    return response.contents[0];
```
jakemac53, line 839: We should give an error or warning in the multiple case if we aren't handling it, this is useful for example if you want a resource that is like a directory, it lists all the resources under it when you read it.

packages/core/src/tools/mcp-client.ts
```diff
@@ -750,6 +765,87 @@ export async function discoverPrompts(
...
 
+/**
+ * Discovers resources from the MCP server.
+ * It retrieves resource declarations from the client and registers them with the resource registry.
+ *
+ * @param mcpServerName The name of the MCP server.
+ * @param mcpClient The active MCP client instance.
+ * @param resourceRegistry The resource registry to register discovered resources.
+ */
+export async function discoverResources(
```
jakemac53, line 776: We should also file an issue for listening to changes to the resource list, we don't have to handle it in the initial implementation but it should be implemented for sure.

jakemac53 (commented):
See also resource annotations https://modelcontextprotocol.io/specification/draft/server/resources#annotations. The priority field should ideally influence sorting if provided, and the audience field should also probably be respected. I am not sure exactly what to do with the audience field, but for instance if the audience is only "user" then we probably shouldn't even add the resource to the context for the model, and we should definitely make sure to show it as a visible item in the history.

Also, we should think about resource subscriptions/notifications. This requires a deeper discussion though. We don't probably want to auto-inject updated resource contents constantly in the context. This is very likely deserving of an actual doc or some sort of formal discussion (maybe even just github discussion, however we want to manage that is fine).

PR #8348, opened 2025-09-12
feat(cli): configure policy engine from existing settings

### LINKED ISSUE #7231: Epic: Tool Confirmation Message Bus

allenhutchison:
This commit introduces the logic to generate a default configuration for the Policy Engine based on existing user settings. This ensures backwards compatibility and a seamless transition to the new permissions model.

The new `createPolicyEngineConfig` function translates settings from `config.yaml` into the structured rules required by the Policy Engine. This includes:
- Mapping `tools.allowed` and `tools.exclude` to allow/deny rules.
- Translating `mcp.allowed` and `mcp.excluded` for MCP tool sources.
- Preserving the behavior of `tools.autoAccept` for read-only tools.
- Handling approval modes such as `YOLO` and `AUTO_EDIT`.

This change ensures that users' existing configurations continue to work as expected with the introduction of the underlying Policy Engine.

Comprehensive unit tests have been added to verify that the translation from legacy settings to policy rules is correct.

Part of the work for #7231


packages/cli/src/config/policy.ts
```diff
@@ -0,0 +1,140 @@
...
+      toolName: tool,
+      decision: PolicyDecision.ASK_USER,
+      priority: 10,
+    });
+  }
+
+  if (approvalMode === ApprovalMode.YOLO) {
+    rules.push({
+      decision: PolicyDecision.ALLOW,
+      priority: 0, // Lowest priority
```
abhipatel12: Is YOLO supposed to have the lowest priority? 

In general, how are we assigning the priority number here? Will we want some central map or set that tells us the various numbers in use at any given point?

packages/cli/src/config/policy.ts
```diff
@@ -0,0 +1,140 @@
...
+    for (const tool of READ_ONLY_TOOLS) {
+      rules.push({
+        toolName: tool,
+        decision: PolicyDecision.ALLOW,
+        priority: 50,
+      });
+    }
+  }
+
+  for (const tool of WRITE_TOOLS) {
```
abhipatel12: Are we adding this loop for readability? We have a default decision below as well for `ASK_USER`

packages/cli/src/config/policy.ts
```diff
@@ -0,0 +1,140 @@
...
+      });
+    }
+  }
+
+  // MCP servers that are explicitly excluded in the settings.
+  // Priority: 200
+  if (settings.mcp?.excluded) {
+    for (const server of settings.mcp.excluded) {
+      rules.push({
+        toolName: `^mcp://${server}/.*`,
```
abhipatel12: Does the `PolicyEngine` in `core` support pattern matching like we have here? I thought the matching was exact string matching?

abhipatel12 (commented):
Looks great! Have a couple of questions!

packages/cli/src/config/policy.ts
allenhutchison: Fixed.

packages/cli/src/config/policy.ts
allenhutchison: Yes I wanted it to be explicit while we transition over to the new system.

packages/cli/src/config/policy.ts
allenhutchison: Added updated comments.

packages/cli/src/config/policy.ts
```diff
@@ -0,0 +1,177 @@
...
+  type PolicyEngineConfig,
+  PolicyDecision,
+  type PolicyRule,
+  ApprovalMode,
+} from '@google/gemini-cli-core';
+import type { Settings } from './settings.js';
+
+// READ_ONLY_TOOLS is a list of built-in tools that do not modify the user's
+// files or system state.
+const READ_ONLY_TOOLS = new Set([
```
abhipatel12, line 31: nit: should we consider using the actual tool names (by importing the tool) in the off chance these change during some hill climbing exercise where we change tool names.

abhipatel12: Same comment for the write tools below!

packages/cli/src/config/policy.ts
```diff
@@ -0,0 +1,177 @@
...
+  // - Rules are evaluated in order of priority (highest first)
+  //
+  // Priority levels used in this configuration:
+  //   0: Default allow-all (YOLO mode only)
+  //   10: Write tools default to ASK_USER
+  //   50: Auto-accept read-only tools
+  //   85: MCP servers allowed list
+  //   90: MCP servers with trust=true
+  //   100: Explicitly allowed individual tools
+  //   195: Explicitly excluded MCP servers
```
abhipatel12, line 73: This may be the intended behavior, but is there a way for a user to exclude a MCP server in general but then specifically allow a single tool?

Is there any reason we'd want to give higher priority to explicitly allowed or disallowed tools than server wide inclusions/exclusions?

packages/cli/src/config/policy.ts
```diff
@@ -0,0 +1,177 @@
...
+  }
+
+  // If smart edit is enabled, deny the replace tool.
+  if (settings.useSmartEdit) {
+    rules.push({
+      toolName: 'replace',
+      decision: PolicyDecision.DENY,
+      priority: 200,
+    });
+  }
```
abhipatel12: the smart edit tool uses the same name as the regular edit (`replace`).

packages/cli/src/config/policy.ts
```diff
@@ -0,0 +1,177 @@
...
+
+  // If auto-accept is enabled, allow all read-only tools.
+  // Priority: 50
+  if (settings.tools?.autoAccept) {
```
abhipatel12, line 144: Did not know we even had this haha!

abhipatel12 (commented):
Looks good! Had a couple of questions!

packages/cli/src/config/policy.ts
allenhutchison: Right, we have so many settings for permissions. At least this will bring them all together in one place.

packages/cli/src/config/policy.ts
allenhutchison: Good catch. I thought it was actually called SmartEdit from the file name. Thanks.

packages/cli/src/config/policy.ts
allenhutchison: Includes and excludes for MCP work on a per-tool basis. There is a trust setting that covers the entire MCP server config.

packages/cli/src/config/policy.ts
allenhutchison: Good point. Done

packages/cli/src/config/policy.ts
```diff
@@ -30,11 +44,12 @@ const READ_ONLY_TOOLS = new Set([
...
-  'save_memory',
-  'run_shell_command',
-  'write_file',
-  'web_fetch',
+  EditTool.Name,
+  MemoryTool.Name,
+  ShellTool.Name,
+  WriteFileTool.Name,
+  WebFetchTool.Name,
+  WebSearchTool.Name,
```
abhipatel12: Did you mean to move web search from read only to write?

abhipatel12 (approved):
Awesome, really excited to build on this! LGTM

packages/cli/src/config/policy.ts
allenhutchison: fixed. Good catch. That's what I get for manual coding :-/

PR #8353, opened 2025-09-12
refactor(release): simplify release workflow

mattKorwel:
This PR simplifies the release workflow by removing scheduled releases and relying solely on manual triggers with a specified version. It streamlines the process and removes unnecessary complexity from the CI configuration.


scottdensmore (approved):
Approved

PR #8362, opened 2025-09-12
fix(ui): Correct footer measurement and prevent negative terminal height

### LINKED ISSUE #8218: Rendering Issue
### What happened?

The App.tsx is being rendered indefinitely. 

We should only be rendering when there are changes.

### Client information

<details>

```console
> /about
# paste output here
```

</details>

### Login information

_No response_

_No response_

galdawave:
## TLDR

This pull request fixes a bug in the UI where the footer's height was not being calculated correctly, leading to rendering issues and potential crashes when the terminal window is small. The change ensures the footer is measured reliably and prevents the shell from being sized to a non-positive height.

## Dive Deeper

The core issue was in `packages/cli/src/ui/AppContainer.tsx`. The footer height was previously calculated using a `useMemo` hook. This approach was unreliable, especially on the initial render or during rapid terminal resizing, because it didn't always re-evaluate when needed. This could lead to an incorrect `availableTerminalHeight` being calculated.

This PR replaces the `useMemo` with a `useState` hook for `footerHeight`. A `useEffect` hook is introduced to re-measure the footer element whenever the application's main buffer changes. This ensures that the `footerHeight` is always up-to-date.

Additionally, a safeguard has been added using `Math.max(..., 1)` when calculating the terminal height for the integrated shell. This prevents the shell from being resized to a height of 0 or less, which was causing the application to crash.

The dependency array for the `resizePty` effect has also been cleaned up to only include direct dependencies, improving code clarity and preventing unnecessary re-renders.

## Reviewer Test Plan

1.  Pull down this branch.
3.  Run `npm start`.
4.  Resize your terminal window to be very short (e.g., less than 10 lines high).
5.  Observe that the CLI continues to render without crashing. The input area and footer should still be visible and usable.
6.  Type a command that uses the shell (e.g., `!ls`).
7.  Verify that the shell output is displayed correctly and doesn't cause the UI to break.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs
Resolves #8218


packages/cli/src/ui/AppContainer.tsx
```diff
@@ -622,18 +622,27 @@ Logging in with Google... Please restart Gemini CLI to continue.
       streamingState === StreamingState.Responding) &&
     !proQuotaRequest;
 
-  // Compute available terminal height based on controls measurement
-  const availableTerminalHeight = useMemo(() => {
+  const [controlsHeight, setControlsHeight] = useState(0);
+
+  useEffect(() => {
```
jacob314: this should be 
useLayoutEffect
that won't really reduce flicker edge cases at the moment because Ink has a bug but when I fix it will mean this is flicker free rather than rendering a frame with the wrong footer measurement.

jacob314 (approved):
bonus points if you can figure out a test that would break without your change but passes with it.lgtm

galdawave:
> bonus points if you can figure out a test that would break without your change but passes with it.lgtm

Added a test :)

PR #8375, opened 2025-09-12
fix(core): Include the latest user request in countTokens for compression

SandyTao520:
## TLDR

This pull request fixes a bug in the chat history compression logic by including the user's upcoming request in the token count calculation. Previously, the decision to compress was based only on the existing history, which could lead to token limit errors if a new message pushed the total count over the limit. This change ensures a more accurate token count, preventing such errors.

## Dive Deeper

The `tryCompressChat` method in `GeminiClient` is responsible for summarizing the conversation history to keep it within the model's token limit. The issue was that this method only considered the existing chat history when calculating the token count, ignoring the tokens of the message about to be sent.

This change introduces the following modifications:
1.  The `tryCompressChat` method signature is updated to accept the user's next `request`.
2.  The upcoming `request` is appended to the history before the token count is calculated for the compression check.
3.  The `prompt` method is updated to pass the current `request` to `tryCompressChat`.

By making the compression check aware of the user's incoming message, we can proactively compress the chat history and avoid hitting the token limit, leading to a more robust and seamless chat experience.

## Reviewer Test Plan

To validate this change, a reviewer can test the chat compression behavior with a long conversation.

1. Start a chat session with the CLI.
2. Engage in a lengthy conversation like asking Gemini CLI to use shell to output something huge.
3. Observe this request won't fail and compression is done.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ✅  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

### DIFF
--- a/packages/core/src/core/client.ts
+++ b/packages/core/src/core/client.ts
@@ -4,15 +4,15 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import {
-  type EmbedContentParameters,
-  type GenerateContentConfig,
-  type PartListUnion,
-  type Content,
-  type Tool,
-  type GenerateContentResponse,
-  createUserContent,
+import type {
+  EmbedContentParameters,
+  GenerateContentConfig,
+  PartListUnion,
+  Content,
+  Tool,
+  GenerateContentResponse,
 } from '@google/genai';
+import { createUserContent } from '@google/genai';
 import {
   getDirectoryContextString,
   getEnvironmentContext,
--- a/packages/core/src/core/client.ts
+++ b/packages/core/src/core/client.ts
@@ -4,13 +4,14 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import type {
-  EmbedContentParameters,
-  GenerateContentConfig,
-  PartListUnion,
-  Content,
-  Tool,
-  GenerateContentResponse,
+import {
+  type EmbedContentParameters,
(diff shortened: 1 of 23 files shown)


packages/core/src/core/client.ts
```diff
@@ -789,9 +790,14 @@ export class GeminiClient {
   async tryCompressChat(
     prompt_id: string,
     force: boolean = false,
+    request?: PartListUnion,
   ): Promise<ChatCompressionInfo> {
     const curatedHistory = this.getChat().getHistory(true);
 
+    if (request) {
```
scidomino, line 797: Could we avoid this by moving the tryCompressChat call in sendMessageStream till after the request is added? (this is assuming that sendMessageStream() is the one that adds the request).

SandyTao520: I tried but there is no easy way to do that, appending the user's request is done in https://github.com/google-gemini/gemini-cli/blob/c15774ce68e9029e27c08dd38f544128a8631e4e/packages/core/src/core/geminiChat.ts#L241, which is deep down in `turn.run()` -> `this.chat.sendMessageStream()`. There is a lot of other logic in between, and we have to define a new event type just for this.

anj-s (approved):

PR #8377, opened 2025-09-12
fix(core): stop streaming request on loop detection

SandyTao520:
## TLDR

This pull request fixes a bug where the streaming request to the model would continue even after a loop was detected. Now, when the `LoopDetectionService` identifies a repetitive pattern, it immediately aborts the underlying streaming request, preventing further unnecessary processing and token consumption.

## Dive Deeper

Previously, when `loopDetector.addAndCheck(event)` returned `true`, the system would yield a `LoopDetected` event and exit the `sendMessageStream` generator. However, this action did not terminate the in-flight streaming request initiated by `turn.run()`. The model would continue to generate content, and the client would continue to receive events, even though they were no longer being processed.

This change introduces an `AbortController` within the `sendMessageStream` function. The signal from this controller is combined with the original signal using `AbortSignal.any()`, creating a `linkedSignal`. This new signal is then passed down to the `turn.run()` method, which handles the API call.

When a loop is detected, `controller.abort()` is now called. This immediately triggers the `linkedSignal`, causing the underlying `fetch` request in the `@google/genai` SDK to be aborted. This ensures that the connection is closed promptly and resources are released, making the loop detection mechanism more efficient and robust.

## Reviewer Test Plan

Use network proxy and trigger a loop by asking it to say something non-stop, observe the network request is stopped when a loop is detected in the UI.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ✅  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

#8231

### DIFF
--- a/packages/core/src/core/client.test.ts
+++ b/packages/core/src/core/client.test.ts
@@ -2405,6 +2405,75 @@ ${JSON.stringify(
       // Assert
       expect(mockCheckNextSpeaker).not.toHaveBeenCalled();
     });
+
+    it('should create linked abort signal and pass it to turn.run', async () => {
+      // Arrange
+      const mockStream = (async function* () {
+        yield { type: 'content', value: 'Hello' };
+      })();
+      mockTurnRunFn.mockReturnValue(mockStream);
+
+      const mockChat: Partial<GeminiChat> = {
+        addHistory: vi.fn(),
+        getHistory: vi.fn().mockReturnValue([]),
+      };
+      client['chat'] = mockChat as GeminiChat;
+
+      const originalSignal = new AbortController().signal;
+
+      // Act
+      const stream = client.sendMessageStream(
+        [{ text: 'Hi' }],
+        originalSignal,
+        'prompt-id-signal',
+      );
+      for await (const _ of stream) {
+        // consume stream
+      }
+
+      // Assert
+      expect(mockTurnRunFn).toHaveBeenCalledWith(
+        expect.any(String),
+        [{ text: 'Hi' }],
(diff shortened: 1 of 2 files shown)


jacob314 (approved):
Can you add a regression test as a follow up?

packages/core/src/core/client.test.ts
```diff
@@ -2405,6 +2405,75 @@ ${JSON.stringify(
       // Assert
       expect(mockCheckNextSpeaker).not.toHaveBeenCalled();
     });
+
+    it('should create linked abort signal and pass it to turn.run', async () => {
```
jacob314, line 2409: nit: this is a low value test as it checks the setup not the behavior. I would remove to leave more room for useful tests.

packages/core/src/core/client.test.ts
```diff
@@ -2405,6 +2405,75 @@ ${JSON.stringify(
...
+      );
+
+      const events = [];
+      for await (const event of stream) {
+        events.push(event);
+      }
+
+      // Assert
+      expect(events).toContainEqual({ type: GeminiEventType.LoopDetected });
+      expect(client['loopDetector'].addAndCheck).toHaveBeenCalledTimes(2);
```
jacob314, line 2475: sorry ELI5, how does this test detect that the abortSignal was actually called as expected with the desired result?

PR #8378, opened 2025-09-12
fix(cli): Auto restart CLI inner node process on trust change

shrutip90:
## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes https://github.com/google-gemini/maintainers-gemini-cli/issues/776

### DIFF
--- a/packages/cli/src/gemini.test.tsx
+++ b/packages/cli/src/gemini.test.tsx
@@ -163,11 +163,16 @@ describe('gemini.tsx main function', () => {
 });
 
 describe('gemini.tsx main function kitty protocol', () => {
+  let originalEnvNoRelaunch: string | undefined;
   let setRawModeSpy: MockInstance<
     (mode: boolean) => NodeJS.ReadStream & { fd: 0 }
   >;
 
   beforeEach(() => {
+    // Set no relaunch in tests since process spawning causing issues in tests
+    originalEnvNoRelaunch = process.env['GEMINI_CLI_NO_RELAUNCH'];
+    process.env['GEMINI_CLI_NO_RELAUNCH'] = 'true';
+
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     if (!(process.stdin as any).setRawMode) {
       // eslint-disable-next-line @typescript-eslint/no-explicit-any
@@ -183,6 +188,15 @@ describe('gemini.tsx main function kitty protocol', () => {
       value: false,
       configurable: true,
     });
+  });
+
+  afterEach(() => {
+    // Restore original env variables
+    if (originalEnvNoRelaunch !== undefined) {
+      process.env['GEMINI_CLI_NO_RELAUNCH'] = originalEnvNoRelaunch;
+    } else {
+      delete process.env['GEMINI_CLI_NO_RELAUNCH'];
+    }
   });
 
   it('should call setRawMode and detectAndEnableKittyProtocol when isInteractive is true', async () => {
--- a/packages/cli/src/gemini.tsx
+++ b/packages/cli/src/gemini.tsx
@@ -107,9 +107,12 @@ function getNodeMemoryArgs(config: Config): string[] {
 }
 
 async function relaunchAppInChildProcess(additionalArgs: string[]) {
+  if (process.env['GEMINI_CLI_NO_RELAUNCH']) {
+    return;
+  }
   while (true) {
     const nodeArgs = [...additionalArgs, ...process.argv.slice(1)];
-      const newEnv = { ...process.env, GEMINI_CLI_NO_RELAUNCH: 'true' };
+    const newEnv = { ...process.env, GEMINI_CLI_NO_RELAUNCH: 'true' };
 
     // The parent process should not be reading from stdin while the child is running.
     process.stdin.pause();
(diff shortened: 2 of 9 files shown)


packages/cli/src/gemini.tsx
```diff
@@ -350,7 +362,7 @@ export async function main() {
       // Not in a sandbox and not entering one, so relaunch with additional
       // arguments to control memory usage if needed.
       if (memoryArgs.length > 0) {
-        await relaunchWithAdditionalArgs(memoryArgs);
+        await relaunchAppInChildProcess(memoryArgs);
         process.exit(0);
```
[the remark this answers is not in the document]
shrutip90: Done

packages/cli/src/gemini.tsx
```diff
@@ -102,17 +103,28 @@ function getNodeMemoryArgs(config: Config): string[] {
...
+    const exitCode = await new Promise<number>((resolve) => {
+      child.on('close', resolve);
+    });
+
+    if (exitCode === RELAUNCH_EXIT_CODE) {
+      relaunch = true;
+    } else {
+      relaunch = false;
+      process.exit(exitCode);
+    }
```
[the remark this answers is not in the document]
shrutip90: Done

packages/cli/src/ui/components/FolderTrustDialog.tsx
```diff
@@ -28,6 +30,17 @@ export const FolderTrustDialog: React.FC<FolderTrustDialogProps> = ({
   onSelect,
   isRestarting,
 }) => {
+  useEffect(() => {
+    const doRelaunch = async () => {
+      if (isRestarting) {
+        setTimeout(async () => {
+          await relaunchApp();
+        }, 1000);
```
jacob314: nit: 1 second is too long a delay just to make what is happening discoverable. Can you go with maybe 250ms instead?

packages/cli/src/gemini.tsx
```diff
@@ -350,8 +364,8 @@ export async function main() {
       // Not in a sandbox and not entering one, so relaunch with additional
       // arguments to control memory usage if needed.
       if (memoryArgs.length > 0) {
```
jacob314: remove the memoryArgs.length > 0 check and add a comment that we must always have child process so we can relaunch the app if needed

packages/cli/src/gemini.tsx
```diff
@@ -350,8 +364,8 @@ export async function main() {
...
-        await relaunchWithAdditionalArgs(memoryArgs);
-        process.exit(0);
+        await relaunchAppInChildProcess(memoryArgs);
```
jacob314: see the start_sandbox call a few lines up above. that also needs the same logic to relaunch on restart. likely you can be clever and avoid duplicating the logic to check the return code but potentially it is easier to special case and add that code to start_sandbox. please test with the sandbox to ensure restarting works there as well.

packages/cli/src/ui/components/FolderTrustDialog.tsx
shrutip90: Done

packages/cli/src/gemini.tsx
shrutip90: Done

packages/cli/src/gemini.tsx
shrutip90: Done and tested locally.

packages/cli/src/gemini.tsx
```diff
@@ -118,21 +139,17 @@ async function relaunchWithAdditionalArgs(additionalArgs: string[]) {
...
 
-    await new Promise<void>((resolve, reject) => {
+    return new Promise<number>((resolve, reject) => {
       child.on('error', reject);
-      child.on('close', () => {
+      child.on('close', (code) => {
         // Resume stdin before the parent process exits.
         process.stdin.resume();
-        resolve();
+        resolve(code ?? 1);
```
jacob314, line 147: why not use 0 (the default process exit code)?

packages/cli/src/gemini.test.tsx
```diff
@@ -163,11 +163,16 @@ describe('gemini.tsx main function', () => {
...
 describe('gemini.tsx main function kitty protocol', () => {
+  let originalEnvNoRelaunch: string | undefined;
   let setRawModeSpy: MockInstance<
     (mode: boolean) => NodeJS.ReadStream & { fd: 0 }
   >;
 
   beforeEach(() => {
+    // Set no relaunch in tests since process spawning causing issues in tests
+    originalEnvNoRelaunch = process.env['GEMINI_CLI_NO_RELAUNCH'];
+    process.env['GEMINI_CLI_NO_RELAUNCH'] = 'true';
```
jacob314, line 174: should this really be an environment variable if we are only adding it for testing?
I would instead add a testing only method or make the relaunch method public and mock it out.

jacob314 (commented):
Looks good once this comment is addressed.

packages/cli/src/gemini.test.tsx
shrutip90: Its not just for testing. It's used to avoid relaunching in infinite loop.

packages/cli/src/gemini.tsx
shrutip90: If the process exited normally, we use its actual exit code, this is for the case when the process in interrupted, likely with a signal. This is apparently a conventional choice because an interruption is not a success.

jacob314 (approved):
lgtm

PR #8391, opened 2025-09-13
fix(tests): Cleans up setup test and fix behavior test

SandyTao520:
## TLDR

Addressed comments in #8379. The current test doesn't verify that the abort signal is actually called when loop detection occurs, new test captures the signal passed to turn.run() and verifies that it gets aborted.


jacob314 (approved):
lgtm

PR #8395, opened 2025-09-13
fix(tests): improve test reliability and performance

bulkypanda:
## TLDR

fixes a few flaky tests

## Dive Deeper

- Mock file size check in fileUtils test to avoid creating large files
- Add retry options to directory cleanup in memoryDiscovery test
- Create test directories in parallel for better performance

## Reviewer Test Plan

run the tests and they shouldn't flake

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ✅  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

- Fixes #8345
- Fixes #8351


packages/core/src/utils/fileUtils.test.ts
```diff
@@ -954,22 +955,53 @@ describe('fileUtils', () => {
...
+        );
+
+        expect(result.error).toContain('File size exceeds the 20MB limit');
+        expect(result.returnDisplay).toContain(
+          'File size exceeds the 20MB limit',
+        );
+        expect(result.llmContent).toContain('File size exceeds the 20MB limit');
+      } finally {
+        statSpy.mockRestore();
+      }
```
bobcatfish, line 981: nice! this should really speed things up 😎

packages/core/src/utils/fileUtils.test.ts
```diff
@@ -954,22 +955,53 @@ describe('fileUtils', () => {
...
+        ctime: new Date(),
+        birthtime: new Date(),
+        isDirectory: () => false,
+        isFile: () => true,
+        isBlockDevice: () => false,
+        isCharacterDevice: () => false,
+        isFIFO: () => false,
+        isSocket: () => false,
+        isSymbolicLink: () => false,
+      });
```
bobcatfish: just say the word if this isn't possible, but since the code being tested only uses the `size` and `isDirectory` attributes of the mocked object, could we remove the others from this instantiation? that would be easier to read and make it more obvious which attributes are important in the context of the test

packages/core/src/utils/memoryDiscovery.test.ts
```diff
@@ -63,7 +63,13 @@ describe('loadServerHierarchicalMemory', () => {
     // Some tests set this to a different value.
     setGeminiMdFilename(DEFAULT_CONTEXT_FILENAME);
     // Clean up the temporary directory to prevent resource leaks.
-    await fsPromises.rm(testRootDir, { recursive: true, force: true });
+    // Use maxRetries option for robust cleanup without race conditions
+    await fsPromises.rm(testRootDir, {
+      recursive: true,
+      force: true,
+      maxRetries: 3,
```
bobcatfish, line 70: i was initially skeptical of just retrying this; but i asked gemini cli and it confirmed that this is totally reasonable:

> When a test times out, especially on Windows, file handles might not be released immediately, causing the ENOTEMPTY error.

packages/core/src/utils/memoryDiscovery.test.ts
```diff
@@ -354,9 +360,11 @@ describe('loadServerHierarchicalMemory', () => {
       .spyOn(console, 'debug')
       .mockImplementation(() => {});
 
-    for (let i = 0; i < 100; i++) {
-      await createEmptyDir(path.join(cwd, `deep_dir_${i}`));
-    }
+    // Create directories in parallel for better performance
+    const dirPromises = Array.from({ length: 100 }, (_, i) =>
```
bobcatfish: im wondering if we can avoid creating 100 directories altogether, what do you think?

looking at the rest of the test it seems like the limit is being set at 50, and 100 dirs are being created:

https://github.com/google-gemini/gemini-cli/blob/a1ebb6f2756bfe9bbeac28dedda8668ca465d3ac/packages/core/src/utils/memoryDiscovery.test.ts#L374

do you think it could work to set the limit at 1 and create 2 directories? or something similarly low?

bobcatfish (changes requested):
Thanks @bulkypanda, appreciate the fixes!! These should make everything more reliable 😁 

The _only_ thing I'd like to see addressed is the instantiation of the mock, if we can make that short and sweet (or let me know if this isn't possible). Everything else is nice to have or musing, feel free to ignore.

*Musing*

Regarding the fix for #8351, i notice that the 20MB file size limit is hard coded in `processSingleFileContent`; it would be nice if this was configurable when it was instantiated, so the test could use a tiny tiny value to exceed that value instead. that would be nice because it would actually be using the file system (maybe a bit more confidence in the test?) but then again i suppose even that could have some weird flakey condition sometimes, so your approach is more reliable!

packages/core/src/utils/memoryDiscovery.test.ts
bobcatfish: nice one @gemini-code-assist (can i tag you? i dont know! i didnt even know you were going to chime in! welcome!) !Can we take it a step further and have a much lower limit so we can minimize the resources we use to run this test? (and the time it takes?) any reason not to use a limit as low as we can set, like 1?

bulkypanda:
will address these today 🙂

packages/core/src/utils/fileUtils.test.ts
bulkypanda: this is possible. the only reason i changed it to add all these fields is because of this: https://github.com/google-gemini/gemini-cli/pull/8395#discussion_r2346038321

bulkypanda: just fixed this mock to be minimal.

packages/core/src/utils/memoryDiscovery.test.ts
bulkypanda: good catch. resolved!

packages/core/src/utils/fileUtils.test.ts
```diff
@@ -954,22 +955,36 @@ describe('fileUtils', () => {
...
+      const statSpy = vi.spyOn(fs.promises, 'stat').mockResolvedValueOnce({
+        size: 21 * 1024 * 1024,
+        isDirectory: () => false,
+        isFile: () => true,
+        isSymbolicLink: () => false,
+        isCharacterDevice: () => false,
+        isBlockDevice: () => false,
+        isFIFO: () => false,
+        isSocket: () => false,
+      } as fs.Stats);
```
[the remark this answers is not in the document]
bobcatfish: i gotta disagree with you here @gemini-code-assist; I see what you're saying about a change to how the object is used making the test fail later, but i think it's better to favor readability here and make it totally clear what the intent of the test is. it feels like a feature (not a bug!) to me that we'd have to update the test if the usage changed

packages/core/src/utils/fileUtils.test.ts
```diff
@@ -954,22 +955,30 @@ describe('fileUtils', () => {
...
-        'File size exceeds the 20MB limit',
-      );
-      expect(result.llmContent).toContain('File size exceeds the 20MB limit');
+      // Create a small test file
+      actualNodeFs.writeFileSync(testTextFilePath, 'test content');
+
+      // Spy on fs.promises.stat to return a large file size
+      const statSpy = vi.spyOn(fs.promises, 'stat').mockResolvedValueOnce({
+        size: 21 * 1024 * 1024,
+        isDirectory: () => false,
```
bobcatfish, line 964: 😎

packages/core/src/utils/memoryDiscovery.test.ts
```diff
@@ -371,12 +379,12 @@ describe('loadServerHierarchicalMemory', () => {
         respectGitIgnore: true,
         respectGeminiIgnore: true,
       },
-      50, // maxDirs
+      1, // maxDirs
```
bobcatfish, line 382: 🥳

bobcatfish (approved):
Awesome!! Looks great @bulkypanda thanks so much!

PR #8399, opened 2025-09-13
fix(quota): handle undefined userTier on initial render

### LINKED ISSUE #3805: Quota limit reached even though I'm on Gemini Code Assist standard plan
### What happened?

I setup our workspace and bought a licence to try out for our team the CLI tool with the Code Assist Standard license.

I got a quota limit message and according to /stats I'm nowhere near the limits.

`ℹ ⚡ You have reached your daily gemini-2.5-pro quota limit.
  ⚡ Automatically switching from gemini-2.5-pro to gemini-2.5-flash for the remainder of this session.
  ⚡ To increase your limits, upgrade to a Gemini Code Assist Standard or Enterprise plan with higher limits at 
  https://goo.gle/set-up-gemini-code-assist
  ⚡ Or you can utilize a Gemini API Key. See: https://goo.gle/gemini-cli-docs-auth#gemini-api-key
  ⚡ You can switch authentication methods by typing /auth
✕ [API Error: Please submit a new query to continue with the Flash model.]

╭────────────╮
│  > /stats  │
╰────────────╯

╭───────────────────────────────────────────────────────────────────╮
│                                                                   │
│  Session Stats                                                    │
│                                                                   │
│  Interaction Summary                                              │
│  Tool Calls:                 16 ( ✔ 14 ✖ 2 )                      │
│  Success Rate:               87.5%                                │
│  User Agreement:             92.3% (13 reviewed)                  │
│                                                                   │
│  Performance                                                      │
│  Wall Time:                  1h 22m 36s                           │
│  Agent Active:               55m 34s                              │
│    » API Time:               3m 25s (6.2%)                        │
│    » Tool Time:              52m 9s (93.8%)                       │
│                                                                   │
│                                                                   │
│  Model Usage                  Reqs   Input Tokens  Output Tokens  │
│  ───────────────────────────────────────────────────────────────  │
│  gemini-2.5-pro                 26        684,179          2,756  │
│  gemini-2.5-flash                2              0              0  │
│                                                                   │
╰───────────────────────────────────────────────────────────────────╯
`

Not reach the quota limit

### Client information

<details>

│                                                         │
│ About Gemini CLI                                        │
│                                                         │
│ CLI Version        0.1.10                               │
│ Model              gemini-2.5-flash                     │
│ Sandbox            no sandbox                           │
│ OS                 linux                                │
│ Auth Method        OAuth                                │
│ GCP Project        XXX-I_REMOVED_IT-XXX               │
│                                                         │

</details>

### Login information

Google Workspace account with Code Assist Standard license.

_No response_

sheet00:
Fixes #3805

When a paid-tier user exceeds their quota, they were incorrectly shown the message for free-tier users.

This was caused by a timing issue in the `useQuotaAndFallback` hook, where the `userTier` prop is `undefined` during the initial component render. The fallback handler would capture this `undefined` value and incorrectly identify the user as being on the free tier.

This commit fixes the issue by using the nullish coalescing operator to fall back to `config.getUserTier()` when the `userTier` prop is undefined. This ensures the correct tier information is used even before the React state has fully propagated.

A regression test is added to verify that the correct (paid-tier) message is displayed when the `userTier` prop is undefined.

## TLDR

Fixes a bug where paid-tier users saw the free-tier quota message by adding a fallback to get the correct user tier during the component's initial render phase.

## Dive Deeper

The root cause was a race condition during app startup. The `useQuotaAndFallback` hook would initialize with an `undefined` `userTier` prop, creating a `fallbackHandler` with a stale closure. If a quota error occurred before the `userTier` prop was updated with the correct value post-authentication, this stale handler would treat a paid user as a free user.

The fix makes the handler robust by checking `config.getUserTier()` at the moment the error occurs, ensuring the most up-to-date information is always used.

<details>
<summary>Detailed Investigation Report</summary>

# Final Investigation Report: Paid Tier Quota Message Issue

## 1. Phenomenon

When a paid-tier user exceeds their quota, they are shown the "upgrade to a Gemini Code Assist Standard or Enterprise plan" message intended for free-tier users, instead of the correct "switch to using a paid API key from AI Studio" message.

## 2. Root Cause

A timing issue in the React component lifecycle.
- **Flow of the problem**:
  1. On initial render of the `AppContainer.tsx` component, the `userTier` state is initialized as `undefined`.
  2. The `useQuotaAndFallback` hook is called with `userTier=undefined`, and its `useEffect` runs.
  3. The `fallbackHandler` generated within this `useEffect` captures the stale information that `userTier` is `undefined` in its closure.
  4. If a quota error occurs before authentication is complete and the `userTier` state is updated with the correct value, this "stale handler" is invoked.
  5. The `isPaidTier` check is performed on `undefined`, resulting in `false`, causing the paid user to be treated as a free user.

- **Problem location**: `packages/cli/src/ui/hooks/useQuotaAndFallback.ts`

## 3. Evidence in Production Code

The hypothesis that the `userTier` property could be `undefined` on initialization was confirmed by investigating the production code in `packages/cli/src/ui/AppContainer.tsx`.

- **`userTier` Initialization**:
  The `userTier` is initialized as `undefined` as component state.
  ```typescript
  // packages/cli/src/ui/AppContainer.tsx:190
  const [userTier, setUserTier] = useState<UserTierId | undefined>(undefined);
  ```

- **`userTier` Update**:
  The actual value is set later within a `useEffect` hook when the authentication state changes.
  ```typescript
  // packages/cli/src/ui/AppContainer.tsx:368-372
  useEffect(() => {
    if (authState === AuthState.Authenticated) {
      setUserTier(config.getUserTier());
    }
  }, [config, authState]);
  ```
This confirms that a window exists between the initial mount and the completion of authentication where `userTier` is `undefined`, and an error during this window will trigger the issue.

## 4. Resolution Process

The fix and verification were carried out with the following steps.

### Step 1: Add a Regression Test

First, a test case to reproduce this bug was added to `packages/cli/src/ui/hooks/useQuotaAndFallback.test.ts`. This test verifies that the correct message for paid users is displayed when the `userTier` property is `undefined`.

```typescript
// The getRegisteredHandler helper has a default value for userTier,
// so we call renderHook directly to ensure we are testing the
// truly `undefined` case.
it('should use config.getUserTier as fallback when userTier prop is undefined', async () => {
  // Arrange
  vi.spyOn(mockConfig, 'getUserTier').mockReturnValue(UserTierId.STANDARD);
  mockedIsGenericQuotaExceededError.mockReturnValue(true);

  renderHook(() =>
    useQuotaAndFallback({
      config: mockConfig,
      historyManager: mockHistoryManager,
      userTier: undefined,
      setAuthState: mockSetAuthState,
      setModelSwitchedFromQuotaError: mockSetModelSwitchedFromQuotaError,
    }),
  );
  const handler = setFallbackHandlerSpy.mock.calls[0][0] as FallbackModelHandler;

  // Act
  const result = await handler('model-A', 'model-B', new Error('quota exceeded'));

  // Assert
  expect(result).toBe('stop');
  const message = (mockHistoryManager.addItem as Mock).mock.calls[0][0].text;
  expect(message).toContain('switch to using a paid API key from AI Studio');
  expect(message).not.toContain('upgrade to a Gemini Code Assist Standard or Enterprise plan');
});
```
Running this test with the code before the fix confirmed that the issue was reproduced (the free-tier message was displayed instead of the paid-tier message).

### Step 2: Implement the Fix

Next, to resolve this issue, `packages/cli/src/ui/hooks/useQuotaAndFallback.ts` was modified. A fallback to get the value from `config.getUserTier()` was added for when `userTier` is `undefined`.

```typescript
// packages/cli/src/ui/hooks/useQuotaAndFallback.ts
// The userTier prop can be undefined on initial render, so we fall back to the config value.
const currentUserTier = userTier ?? config.getUserTier();
const isPaidTier =
  currentUserTier === UserTierId.LEGACY ||
  currentUserTier === UserTierId.STANDARD;
```

### Step 3: Verify the Fix

After applying the fix, the tests were run again, and it was confirmed that all tests, including the added test case, passed. This ensures that the fix is correct and no regressions have occurred.

## 5. Final Conclusion

A race condition caused by React's initialization timing was identified, and a robust fix was implemented. Specifically, logic was added to fall back to the config when the `userTier` property is `undefined`, and a regression test was also added to guarantee this behavior. With this, the issue has been permanently resolved.

</details>

## Reviewer Test Plan

1.  The primary validation is the new automated test case added in this PR (`should use config.getUserTier as fallback when userTier prop is undefined`). A reviewer can confirm the fix by running the test and seeing it pass:
    ```bash
    npm run test --workspace @google/gemini-cli -- src/ui/hooks/useQuotaAndFallback.test.ts
    ```
2.  Manual verification is difficult as it requires triggering a quota error in the narrow time window between app startup and user authentication completing. The automated test is the most reliable method to confirm the fix.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ✅  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes #3805


packages/cli/src/ui/hooks/useQuotaAndFallback.ts
```diff
@@ -58,8 +58,11 @@ export function useQuotaAndFallback({
       }
 
       // Use actual user tier if available; otherwise, default to FREE tier behavior (safe default)
+      // The userTier prop can be undefined on initial render, so we fall back to the config value.
+      const currentUserTier = userTier ?? config.getUserTier();
```
abhipatel12: What if we just rely on `config.getUserTier()`. Since that is our synchronous source of truth we don't have to rely on the user tier prop? WDYT?

Then we could remove this prompt entirely, including from the `AppContainer.tsx`?

We would likely need to update the tests if we did this.

packages/core/src/core/loggingContentGenerator.test.ts
```diff
@@ -0,0 +1,56 @@
...
+      mockWrapped.userTier = UserTierId.LEGACY;
+      expect(loggingGenerator.userTier).toBe(UserTierId.LEGACY);
+    });
+
+    it('should return undefined when wrapped userTier is undefined', () => {
+      mockWrapped.userTier = undefined;
+      expect(loggingGenerator.userTier).toBeUndefined();
+    });
+  });
+});
```
abhipatel12: Be sure to run `npm run format` or `npm run preflight` in order to get the proper formatting for these files!

abhipatel12 (commented):
Hey @sheet00! Sorry again for the delayed review here. This is great work. 

I left a couple comments, so let me know what you think.

Also, I noticed the CLA check is failing. Have you completed the CLA at: https://cla.developers.google.com/

packages/core/src/core/loggingContentGenerator.test.ts
abhipatel12: `npm run preflight` should also run the lint check, type check, and the unit tests. This way you can be sure that everything passes!

sheet00:
# userTier Refactoring Implementation Plan

## Overview
Refactoring to remove `userTier` prop dependency from `useQuotaAndFallback.ts` and use only `config.getUserTier()`

## Current Issues
- `userTier` prop can be `undefined` during initial rendering
- Dependency on multiple information sources (prop + config)
- Unnecessary prop passing from AppContainer.tsx

## Proposed Solution
Unified use of `config.getUserTier()` as a synchronous reliable information source

## Implementation Steps

### 1. Modify useQuotaAndFallback.ts
**Current:**
```typescript
const currentUserTier = userTier ?? config.getUserTier();
```

**After:**
```typescript
const currentUserTier = config.getUserTier();
```

### 2. Modify AppContainer.tsx
- Remove `userTier` prop passing to `useQuotaAndFallback`
- Remove unnecessary `userTier` related imports and state management

### 3. Update Type Definitions
- Remove `userTier` parameter from `useQuotaAndFallback` type definition
- Update related interfaces

### 4. Update Tests
- Update `useQuotaAndFallback.test.ts`
  - Remove test cases passing `userTier` prop
  - Unify to `config.getUserTier()` mocking
- Update AppContainer.tsx test file
  - Modify/remove `userTier` related test cases

## Impact Scope

### Files Requiring Changes
1. `packages/cli/src/ui/hooks/useQuotaAndFallback.ts`
2. `packages/cli/src/ui/AppContainer.tsx`
3. `packages/cli/src/ui/hooks/useQuotaAndFallback.test.ts` (if exists)
4. `packages/cli/src/ui/AppContainer.test.tsx` (if exists)

### Files Requiring Verification
- Other components calling `useQuotaAndFallback`
- Other test files related to `userTier`

## Expected Benefits

### Advantages
- Improved code simplicity
- Avoid undefined errors during initial rendering
- Unification to a single reliable information source
- Removal of unnecessary prop passing

### Risk Mitigation
- Important to verify that `config.getUserTier()` always returns valid values
- Ensure existing tests are properly updated

## Implementation Order
1. **Phase 1**: Modify useQuotaAndFallback.ts
2. **Phase 2**: Modify AppContainer.tsx
3. **Phase 3**: Update tests
4. **Phase 4**: Run comprehensive tests and verify functionality

sheet00:
hi! @abhipatel12 

Thank you for the review!
I have addressed the feedback points.
The modifications are as described in the comments above.
Please let me know if there's anything else that needs to be addressed!
I have also completed the CLA registration!

abhipatel12 (approved):
Hey thanks for making the changes. This LGTM. Going to kick of the e2e tests and see if we can get this merged

sheet00:
@abhipatel12 
Hi, how did the e2e tests go? Are we ready to merge this?

PR #8403, opened 2025-09-13
docs: Fix broken Extension example

vorburger:
This seems to be outdated; apparently it now needs positional instead of named arguments, as using --path and --type as documented fails with gemini --version 0.4.1:

```sh
$ gemini extensions new <path> <template>

Create a new extension from a boilerplate example.

Positionals:
  path      The path to create the extension in.                                                                                      [string] [required]
  template  The boilerplate template to use.                   [string] [required] [choices: "context", "custom-commands", "exclude-tools", "mcp-server"]

Options:
  -h, --help  Show help                                                                                                                         [boolean]

Not enough non-option arguments: got 0, need at least 2
```

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


vorburger:
> This seems to be outdated; apparently it now needs positional instead of named arguments, as using --path and --type as documented fails with gemini --version 0.4.1:

I've also tested it with `0.5.0-preview-2` (from `npm install -g @google/gemini-cli@preview`) and that also needs positional instead of named arguments, so clearly this document is ancient.

agmsb (approved):

agmsb:
Thanks for submitting!

vorburger:
Build Failure is a coverage report failure which seems to be unrelated to the change proposed in this PR.

PR #8418, opened 2025-09-13
fix(markdown): Fix markdown rendering

### LINKED ISSUE #8033: Numbered Lists not working
### What happened?

The printing of numbered lists seems to get interrupted each time after the first entry. I need to ask it to list in code block to actually see the rest of the list.

Example query + response:
```
╭───────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│  > What is stroke_width used for? I see no effect when running it.                                       │
│                                                                                                          │
│    Please add "random" as a background color. And also list all typeable colors, or link to a list of it │
│     if it's a library (in tooltip). This list should be in tooltip.                                      │
│                                                                                                          │
│    For "Noise Level", noise is only applied to single color shapes. It should be possible to apply noise │
│     to all shapes, right?                                                                                │
│                                                                                                          │
│    Since we have the shape of it, we apply it after the shape it done with whatever color (like          │
│    gradient).                                                                                            │
│                                                                                                          │
│    How come when I change noise level, I get a completely different image? If I use the same seed.       │
│    Our noise here should not affect the initial image I believe, just be added on top.                   │
│                                                                                                          │
│    Rename "black/white" to "monochrome" for noise color. Add random.                                     │
│                                                                                                          │
│    Color bleed still does nothing. I would have assumed it would softly bleed all colors outside of it,  │
│    but  what's the difference between that and normal blur radius?                                       │
│                                                                                                          │
│    Can we rename the "warp type turbulence" to "noise field" or something? It feels like a "blurry       │
│    noise", or similar.                                                                                   │
│                                                                                                          │
│    Please make the warp intensity default 0.5                                                            │
│                                                                                                          │
│    I'm not sure that "allow rotation" works. I mean, I get random rotations, but I also get way too many │
│     perfect 90 degree edges. If allow rotation is on, does this mean a 50-50 to get a random rotation or │
│     not? I expect if rotation is on, all are rotated randomly 360 degrees?                               │
│                                                                                                          │
│    bug: The grid in arrangement seem to always to a grid to the top part of the image. It should scale   │
│    across the entire image and resolution/ratio. And of course, follow the direction of positioning bias │
│     if possible. Currently, the grid is always top 50% of image.                                         │
│                                                                                                          │
│    The "shape opacity", seems to be overridden by the noise added in the noise level. The shape opacity  │
│    should control the total opacity per shape, and it should limit the maximum opacity. If this cannot   │
│    be done like this, we should move this effect later in the chain to control it.                       │
╰───────────────────────────────────────────────────────────────────────────────────────────────────────────╯

✦ Thank you for the incredibly detailed feedback. You've highlighted several key areas where the node is
  falling short of its potential. I'm going to work through each of your points to fix the bugs and
  implement your suggestions. This is a significant update, so I'll tackle it step-by-step.


  Here is my plan:


   1. `stroke_width`: I'll fix the drawing logic so that strokes are always visible when stroke_width is
      greater than 0.
  This is a substantial set of changes, and I'll get started on them right away. I'll begin by reading
  the latest version of the file.
 ╭───────────────────────────────────────────────╮
 │ ✓  ReadFile nodes\colorful_starting_image.py  │
 ╰───────────────────────────────────────────────╯
```

And another one:
```
╭───────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│  > Noise is not working well. I can't see it.IT used to be there. Not sure if colors somewhere is        │
│    overwriting it.                                                                                       │
│                                                                                                          │
│    To confirm, we are doing it top to bottom, right? So the bottom operations are performed last?        │
│                                                                                                          │
│    Also, color bleed does nothing.                                                                       │
│    Swirl warp type does nothing at a weight of 1. Additionally it acts more like a "center pinch", not a │
│     swirl. It should have a result on a weight of 1.                                                     │
│                                                                                                          │
│    Why is noise level max 1? Can we increase?                                                            │
│                                                                                                          │
│    Controlled Chaos is a bad title for this genre. And shouldn't the "placement" position happen much    │
│    earlier? Like right after positioning bias?                                                           │
│                                                                                                          │
│    Remember that the output order is important and to reorder the outputs, you must also update the      │
│    tooltip outputs etc.                                                                                  │
╰───────────────────────────────────────────────────────────────────────────────────────────────────────────╯

✦ Thank you for the detailed feedback. You've pointed out several important issues and areas for
  improvement. I'll address each of them.

  Here's my plan:


   1. Fix the Noise: I'll investigate why the noise effect isn't visible and ensure it's applied correctly
      without being overwritten.
  Regarding your question about the order of operations, you are correct. The operations are performed
  from top to bottom as they appear in the node's UI. I'll ensure the reordering of the new arrangement
  parameter is reflected correctly.


  I'll start by reading the file to apply these changes.
 ╭───────────────────────────────────────────────╮
 │ ✓  ReadFile nodes\colorful_starting_image.py  │
 ╰───────────────────────────────────────────────╯

╭───────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│  > I gave you more feedback than noise effect. Please make sure you did all of my message:               │
│                                                                                                          │
│    Noise is not working well. I can't see it.IT used to be there. Not sure if colors somewhere is        │
│    overwriting it.                                                                                       │
│                                                                                                          │
│    To confirm, we are doing it top to bottom, right? So the bottom operations are performed last?        │
│                                                                                                          │
│    Also, color bleed does nothing.                                                                       │
│    Swirl warp type does nothing at a weight of 1. Additionally it acts more like a "center pinch", not a │
│     swirl. It should have a result on a weight of 1.                                                     │
│                                                                                                          │
│    Why is noise level max 1? Can we increase?                                                            │
│                                                                                                          │
│    Controlled Chaos is a bad title for this genre. And shouldn't the "placement" position happen much    │
│    earlier? Like right after positioning bias?                                                           │
│                                                                                                          │
│    Remember that the output order is important and to reorder the outputs, you must also update the      │
│    tooltip outputs etc.                                                                                  │
╰───────────────────────────────────────────────────────────────────────────────────────────────────────────╯

✦ You are absolutely right to call me out on that. My previous response was poorly worded and focused too
  much on the noise fix, making it seem like I'd ignored the rest of your valuable feedback. I apologize
  for that.


  I did implement all the changes you requested, but I failed to communicate that clearly. Thank you for
  your patience and for holding me accountable.

  To confirm, here is a list of all the changes I made in the last update based on your feedback:


   * Noise Visibility: Fixed the logic to make the noise effect more reliable and visible.
  I have double-checked the code to ensure all these changes are in place. Please let me know if the
  node is now behaving as you expect or if there's anything else I can improve.

╭─────────────────────────────────────────────────────────────────────────╮
│  > You only listed noise. What else did you do?                         │
│                                                                         │
│    Output a code block (do not write to code), your lists get cut off.  │
╰─────────────────────────────────────────────────────────────────────────╯

✦ My apologies again. You are correct, and I appreciate you pointing out the issue with my responses being
  cut off. I will provide the full list of changes in a code block as you requested.


  Here is the complete summary of the updates I made based on your feedback:


  `
  - Noise Visibility: The logic was corrected to make the noise effect more reliable and consistently
  visible when stroke_width is 0.
  - Color Bleed: The effect has been adjusted to scale with the image size, making it more noticeable
  and effective.
  - Swirl Warp: The algorithm was completely reworked to produce a proper swirl, not a pinch. It now
  works correctly at an intensity of 1.
  - Increased Noise Level: The maximum value for the noise_level parameter has been increased from 1.0
  to 5.0.
  - Renamed & Moved "Controlled Chaos":
      - The parameter has been renamed to arrangement for clarity.
      - It has been moved in the UI to appear directly after positioning_bias for a more logical
  workflow.
  - Confirmed Order of Operations: Your understanding is correct; the node's operations are performed
  sequentially from top to bottom as they appear in the UI.
  `


  I have verified that all of these changes are in the current version of the colorful_starting_image.py
  file. Please let me know if the node is now working as expected.

```

See numbered list.

### Client information

<details>

```console
> /about
# paste output here
```

</details>

### Login information

Google account

Seems to happen to version 0.34. Last version I was using was 0.32 and I don't believe I've had issues with numbered lists before.

### LINKED ISSUE #8149: the answer is not completed when tring with the cli. sometimes it gives more short answers not covering all the scenarios.
### What happened?

Image
Here I have asked spcifically few terminalogies but it has answerd for few.

It should show all the details from this.

### Client information

* **CLI Version:** 0.3.4
* **Git Commit:** 2acadeca
* **Session ID:** d6ed5171-664b-4418-bae6-e412dfcf6263
* **Operating System:** win32 v22.19.0
* **Memory Usage:** 215.1 MB

### Login information

_No response_

_No response_

### LINKED ISSUE #8163: Gemini CLi is missing parts of the output
Image

### LINKED ISSUE #8194: List Rendering Bug
### What happened?

 Bug Report: AI model truncates markdown list outputs

  Description:
  The AI model has a recurring bug where it fails to output complete lists when they are formatted using standard markdown (numbered or bulleted). It consistently returns only the first item of the list.

  Replication Steps:
  1.  Ask the model to recite its "Golden Rules." It will likely only return the first rule.
  2.  Ask the model to create a multi-step plan. It will likely only show the first step.

  Diagnosis:
  The issue seems to be with the final output generation when processing standard markdown list formats.

  Workaround:
  We discovered a successful workaround. When the model is instructed to format the list with distinct headings for each item (e.g., using bolded text **Step 1:**) instead of a 1. or * prefix, it successfully outputs the entire list without truncation. The model has saved this
  workaround to its memory for future interactions.

lists should render correctly (like in previous versions)

### Client information

About Gemini CLI                                                                                               │
│                                                                                                                │
│ CLI Version                           0.4.0                                                                    │
│ Git Commit                            89213699                                                                 │
│ Model                                 gemini-2.5-pro                                                           │
│ Sandbox                               no sandbox                                                               │
│ OS                                    win32                                                                    │
│ Auth Method                           gemini-api-key                                                           │
│                                                           

### Login information

API key

_No response_

### LINKED ISSUE #8392: General Markdown Rendering Failure in CLI Output
### What happened?

  The Gemini CLI fails to correctly render any responses formatted with Markdown in its direct screen output. All Markdown formatting  
  characters (such as # for headings, * for lists, ** for bold) are stripped or ignored. This results in a garbled, plain-text-like    
  output that is difficult to read and loses the intended structure. The issue occurs regardless of the language used (tested with both
   English and Chinese).

  Steps to Reproduce:

  1.  Run the following prompt:
      Please reply with the following content in Markdown format: A level 1 heading "Test", an unordered list with "Item 1" and "Item  2", and bold the word "important".

  2.  Observe the direct output on the screen.

  Expected Behavior:
  The CLI should display correctly rendered Markdown, with "Test" as a large heading, "Item 1" and "Item 2" as bullet points, and      
  "important" as bold text.
  Alternatively, it should at least display the raw, un-rendered Markdown text with all formatting characters intact, like this:       
  `
  # Test
  * Item 1
  * Item 2

  Here is the important word.

  `
 Actual Behavior:
  The CLI displays a broken output with all formatting characters stripped. The output looks like this:
  `
  Test
  Here is the important word.

  `

Image

  The CLI should display correctly rendered Markdown.

### Client information

* **CLI Version:** 0.4.1
* **Git Commit:** 70ff7a36
* **Session ID:** 525ec8c5-962c-4c1a-91ba-4ea78e8b8f5a
* **Operating System:** win32 v22.13.0
* **Memory Usage:** 125.3 MB

### Login information

Google Account

The previous version was fine, but I'm not sure which specific version introduced this issue.

### LINKED ISSUE #8426: Can't display a list of proposition anymore
### What happened?

I ask Gemini to create 5 propositions ( for a document title ). 
And now it only shows the first proposition, not the other ones.

So gemini Cli answer is now something like : 
"here are your 5 propositions : 
1 - proposition 1
Which one do you prefer ?"

Note that : 
1) it was working ok somethign like 15 days ago
2) a workaround is to ask it to write the propositions in a txt file

I would like his response to be : 
"here are your 5 propositions : 
1 - proposition 1
2 - proposition 2
3 - proposition 3
4 - proposition 4
5 - proposition 5
Which one do you prefer ?"

### Client information

* **CLI Version:** 0.4.1
* **Git Commit:** 70ff7a36
* **Session ID:** a62cfc08-7019-4686-a0e9-3e778f3829a7
* **Operating System:** win32 v22.14.0
* **Memory Usage:** 185.1 MB

### Login information

_No response_

_No response_

### LINKED ISSUE #8439: Markdown Rendering Bugs in Gemini CLI Lists and Formatting
### What happened?

The Gemini CLI client fails to correctly render standard Markdown in multiple scenarios, especially within lists. The rendered output is often truncated or has its formatting stripped.
  1. List rendering is completely truncated at the first list item that contains bolded text using asterisks.
  2. An entire list item is omitted from the output if it contains a nested list.
  3. An entire list item is omitted from the output if it contains a code block with special characters (e.g., let array = [1, 2];).
  4. The CLI sometimes incorrectly applies its own numbering to a sequence of plain paragraphs, and the numbering has been seen to reset incorrectly (e.g., 1, 2, 1).
  I have attached a file named bug_report_final_test_en.md that contains the original, valid Markdown text used for testing, along with screenshots/pastes of the incorrectly rendered output.

[bug_report_final_test_en.md](https://github.com/user-attachments/files/22321396/bug_report_final_test_en.md)

Image

 The Gemini CLI client should correctly and completely render all standard Markdown elements as provided by the model.

  Specifically:
  1. Lists containing bolded text, nested lists, or code blocks should render fully and without truncating the content.
  2. Markdown formatting (like bolding) should be displayed correctly, not stripped.
  3. The client should not incorrectly apply its own numbering to text that is not explicitly formatted as a list.

  The content inside the bug_report_final_test_en.md file shows the expected output.

### Client information

* **CLI Version:** 0.4.1
* **Git Commit:** 70ff7a36
* **Session ID:** 0b18eeaa-fbbe-463a-9b4d-02fd4fdbc1c1
* **Memory Usage:** 228.5 MB

### Login information

_No response_

_No response_

dracic:
## TLDR

Fixes markdown rendering bug where lists were being truncated to only the first item. Enhanced the `findLastSafeSplitPoint()` function to detect and preserve markdown structures (lists, headers, blockquotes, tables) when splitting content for streaming performance.

## Dive Deeper

The Gemini CLI was experiencing a critical markdown rendering issue where structured content (particularly numbered and bulleted lists) was being truncated, showing only the first item instead of the complete list. This was causing poor user experience and loss of important information.

**Root Cause**: The `findLastSafeSplitPoint()` function in `markdownUtilities.ts` was optimizing for performance by splitting content at double newlines (`\n\n`) without considering markdown structure integrity. This caused splits to occur after headers and within lists, breaking the rendering pipeline.

**Solution**: Enhanced the splitting logic with structure-aware detection:
- Added `isWithinList()` to detect various list patterns (*, -, +, numbered, alphabetic, roman numerals, task lists)
- Added `endsWithHeader()` to prevent splits immediately after headers
- Added `isWithinBlockquote()` and `isWithinTable()` for comprehensive structure protection
- Implemented fallback behavior when perfect splits aren't possible

**Technical Details**:
- Uses regex patterns to identify markdown structures
- Maintains streaming performance while preserving content integrity
- Comprehensive test suite with 29 test cases covering edge cases
- Documented limitations and future improvement paths

## Reviewer Test Plan

1. **Run the test suite**: `cd packages/cli && npm test -- markdownUtilities` - All 29 tests should pass
2. **Test the original bug scenario**: Ask Gemini to generate numbered lists or bulleted lists and verify all items are displayed (not just the first item)
3. **Test various markdown structures**:
   - Ask for "Create a numbered list of 10 programming languages"
   - Ask for "Show me a markdown table with multiple rows"
   - Ask for "Generate a list with headers and nested bullet points"
   - Ask for "Create a blockquote with multiple paragraphs"

4. **Verify edge cases**:
   - Very long content with mixed structures
   - Lists with unusual spacing
   - Code blocks containing list-like syntax
   - Tables with various formatting

The fix should maintain complete list integrity while preserving the CLI's streaming performance characteristics.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ✅  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

- ✅ Windows: Tests passing, functionality verified
- Unit tests: 29/29 passing with comprehensive coverage

## Linked issues / bugs

Resolves: #8194 
Resolves: #8392
Resolves: #8439
Resolves: #8426
Resolves: #7924
Resolves: #8033
Resolves: #8149
Resolves: #8163
etc...

**Files Changed**:
- `packages/cli/src/ui/utils/markdownUtilities.ts` - Enhanced safe split logic
- `packages/cli/src/ui/utils/markdownUtilities.test.ts` - Comprehensive test coverage (29 test cases)


packages/cli/src/ui/utils/markdownUtilities.ts
```diff
@@ -90,6 +90,92 @@ const findEnclosingCodeBlockStart = (
...
+  const nextContent = content.substring(
+    index,
+    Math.min(index + 150, content.length),
+  );
+  const nextLines = nextContent.split('\n').slice(0, 4);
+
+  // Check if next lines continue the blockquote pattern
+  const blockquotePattern = /^[\s]*>\s*/;
+  return nextLines.some((line) => blockquotePattern.test(line));
+};
```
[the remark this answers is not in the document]
dracic: /gemini review

packages/cli/src/ui/utils/markdownUtilities.ts
```diff
@@ -90,6 +90,92 @@ const findEnclosingCodeBlockStart = (
...
+  // Check if we have table-like content before and after
+  const hasTableBefore = beforeLines.some(
+    (line) => tableRowPattern.test(line) || tableSeparatorPattern.test(line),
+  );
+  const hasTableAfter = afterLines.some(
+    (line) => tableRowPattern.test(line) || tableSeparatorPattern.test(line),
+  );
+
+  return hasTableBefore && hasTableAfter;
+};
```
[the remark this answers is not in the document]
dracic: /gemini review

joshualitt:
LGTM

joshualitt:
@dracic , thanks for the PR! I'm hoping we can merge this quickly.

The general code looks good, but I have a few questions which seem worth discussing:

1.  `endsWithHeader` may not correctly recognize Setext-style Headers(`=` and `-`).
2.  Do we want to support indented code blocks? IIUC, I don't think the current code does.
3.  Coverage for the list false positives?

Ideally we'd support 1 and 2, but this PR is a huge improvement. Thoughts?

dracic:
@joshualitt I'll work on this tomorrow, and we can test!

dracic:
I still have issues with lists with headers and nested bullet points...

joshualitt:
@dracic , completely understandable, I think a heuristics based approach will always have limitations. Anyway, this is a massive improvement over what we have today. Let me test with it a bit today, and if everything looks good then I'm comfortable landing this.

However, I think we should investigate an AST or token based approach as a follow-up for post GA. I filed https://github.com/google-gemini/gemini-cli/issues/8705 to track this future work.

dracic:
> @dracic , completely understandable, I think a heuristics based approach will always have limitations. Anyway, this is a massive improvement over what we have today. Let me test with it a bit today, and if everything looks good then I'm comfortable landing this.
> 
> However, I think we should investigate an AST or token based approach as a follow-up for post GA. I filed #8705 to track this future work.

I agree on that. This is a mess... However, I'll try to fix this.

dracic:
@joshualitt this the best I can do. Looking forward to AST or token based approach!

dracic:
image

dracic:
Now highlighting is broken....

dracic:
Issue: The InlineMarkdownRenderer component is inconsistently applying theme colors - some <Text> elements have color={theme.text.primary} while others have no color attribute, causing text to render in default terminal white. Additionally, headers can't maintain their intended color hierarchy because the renderer always forced theme.text.primary.

We can add missing color attributes to all <Text> elements and introduce an optional color prop to RenderInline so headers can pass their   specific colors (theme.text.link for H1/H2, theme.text.secondary for H4, etc.) instead of being forced to use the default primary color.

Shall we do this in this PR? Or make another issue/PR?

image

dracic:
Fixed inconsistent vertical spacing....

joshualitt:
> Issue: The InlineMarkdownRenderer component is inconsistently applying theme colors - some elements have color={theme.text.primary} while others have no color attribute, causing text to render in default terminal white. Additionally, headers can't maintain their intended color hierarchy because the renderer always forced theme.text.primary.
> 
> We can add missing color attributes to all elements and introduce an optional color prop to RenderInline so headers can pass their specific colors (theme.text.link for H1/H2, theme.text.secondary for H4, etc.) instead of being forced to use the default primary color.
> 
> Shall we do this in this PR? Or make another issue/PR?

I say we tackle this in a new issue with a new PR. Anyway, there are some issues with preflight, but this code LGTM!

joshualitt (approved):

joshualitt:
Hmm, upon further reflection, do we want to break out the vertical spacing changes into their own PR? I feel very confident in the changes to `markdownUtilities.ts`, but I'd like to think more about the spacing issues, which I think need more work:
image

dracic:
@joshualitt i fixed formatting, sorry for that. I agree on spacing, we need to brainstorm more on this issue.

joshualitt:
> @joshualitt i fixed formatting, sorry for that. I agree on spacing, we need to brainstorm more on this issue.

Ack! Could you move the changes to `packages/cli/src/ui/utils/MarkdownDisplay.tsx` to their own PR? Then, I'll ping a code-owner to LGTM the changes in this PR to `markdownUtilities.ts` which IMO can land today, and we can iterate on the changes to `MarkdownDisplay.tsx` in a follow-up.

dracic:
Ok, deal.

dracic:
@joshualitt,  done. Ok, we'll deal with MarkdownDisplay in other PR, or even go straight to #8705 ?

joshualitt:
@dracic , going straight to a robust solution would be really nice, but I doubt we'd be able to ship anything this complex before GA.

So I think our options are:
1) Punt on further fixes here completely until after GA
2) Land your fixes to MarkdownDisplay next week, and then tackle #8705 after GA

Thoughts? If the formatting errors are bad enough, then maybe 2 is the right approach for now.

packages/cli/src/ui/utils/markdownUtilities.test.ts
```diff
@@ -9,42 +9,934 @@ import { findLastSafeSplitPoint } from './markdownUtilities.js';
...
+      });
+
+      it('should return content.length if no safe split point is found', () => {
+        const content = 'longstringwithoutanysafesplitpoint';
+        expect(findLastSafeSplitPoint(content)).toBe(content.length);
+      });
+
+      it('should prioritize splitting at \n\n over being at the very end of the string if the end is not in a code block', () => {
+        const content = 'Some text here.\n\nAnd more text here.';
+        expect(findLastSafeSplitPoint(content)).toBe(17); // after the \n\n
```
jacob314, line 25: as a followup pr it would be nice if these tests checked what text was before and after the splitpoint. that would make it easier for future code reviewers to verify the tests did what was intended.

packages/cli/src/ui/utils/markdownUtilities.ts
```diff
@@ -35,13 +35,115 @@ This function aims to find an *intelligent* or "safe" index within the provided
...
+    const indent = getIndentationLevel(line);
+
+    // If we hit a line with <4 spaces that's not blank, we've found the start boundary
+    if (line.trim() !== '' && indent < 4) {
+      codeBlockStart = i + 1;
+      break;
+    }
+
+    // If this line is indented 4+, continue looking backwards
+    if (indent >= 4) {
```
jacob314: can you make the indent per tab a constant? would make it easier to change this in the future if we decide to.

packages/cli/src/ui/utils/markdownUtilities.ts
```diff
@@ -90,6 +263,102 @@ const findEnclosingCodeBlockStart = (
...
+ * Checks if a position is likely within a markdown list structure
+ * Enhanced to handle edge cases and unusual spacing
+ * LIMITATION: Regex-based detection may have false positives with file paths or timestamps
+ * FUTURE: Consider using an AST-based approach for more accurate list detection
+ */
+const isWithinList = (content: string, index: number): boolean => {
+  // Check a more generous range of lines to handle unusual spacing
+  const nextContent = content.substring(
+    index,
+    Math.min(index + 200, content.length),
```
jacob314: 200 is quite the magic number. make it a const and document.

packages/cli/src/ui/utils/markdownUtilities.ts
```diff
@@ -90,6 +263,102 @@ const findEnclosingCodeBlockStart = (
...
+  );
+  const nextLines = nextContent.split('\n').slice(0, 6); // Check more lines
+
+  // Enhanced list patterns including task lists and varied spacing
+  const listPattern = /^[\s]*([*\-+]|\d+\.|[a-zA-Z]\.|[ivxlcdm]+\.)\s+/;
```
jacob314: do we really need to support roman numerals?

dracic:
Ok let's try to fix this next week to be at least "good enough"

packages/cli/src/ui/utils/markdownUtilities.ts
```diff
@@ -90,6 +263,102 @@ const findEnclosingCodeBlockStart = (
...
+  }
+
+  return false;
+};
+
+/**
+ * Checks if a position would split within a blockquote structure
+ * Note: Since \n\n terminates a blockquote per Markdown spec, splits at \n\n are always valid
+ */
+const isWithinBlockquote = (_content: string, _index: number): boolean =>
```
jacob314: these methods are no ops. please remove.

packages/cli/src/ui/utils/markdownUtilities.ts
```diff
@@ -111,15 +382,29 @@ export const findLastSafeSplitPoint = (content: string) => {
...
-    // If potentialSplitPoint was inside a code block,
+    // If potentialSplitPoint was inside a code block or list,
     // the next search should start *before* the \n\n we just found to ensure progress.
     searchStartIndex = dnlIndex - 1;
   }
 
-  // If no safe double newline is found, return content.length
-  // to keep the entire content as one piece.
-  return content.length;
+  // If we only found splits that would break lists, use the last one as a fallback
```
jacob314, line 400: this needs to be behind a setting. For now it is ok to have but I will want to turn it off in the future when we have scrolling fixes that mean we do not have to force breaking very large blocks to avoid flicker.

packages/cli/src/ui/utils/markdownUtilities.ts
dracic: You are absolutely right! Let's not risk false postives

packages/cli/src/ui/utils/markdownUtilities.ts
dracic: Ok

packages/cli/src/ui/utils/markdownUtilities.test.ts
dracic: Ok

dracic:
@jacob314 done. I've noted the rest and will try to resolve it next week. Have a pleasant weekend, everyone.

packages/cli/src/ui/utils/markdownUtilities.ts
dracic: Done

packages/cli/src/ui/utils/markdownUtilities.ts
dracic: Done

packages/cli/src/ui/utils/markdownUtilities.ts
dracic: Removed

dracic:
After proper testing I conclude this doesn't solve Markdown rendering issues. The only way is to implement #8705

MNeMoNiCuZ:
For my reported [numbered lists not working](https://github.com/google-gemini/gemini-cli/issues/8033) bug, this used to work before. It broke some week or two before my report, but it worked fine before then.

dracic:
@MNeMoNiCuZ please try and test #10566

PR #8430, opened 2025-09-14
Fix mixed input crash by adding proper error handling 8304

lifefloating:
## TLDR

Fixes issue #8304 where mixing piped input with `--prompt-interactive` flag causes a crash with "TypeError: process.stdin.setRawMode is not a function". Now properly detects this invalid combination and shows a clear error message instead of crashing.

The core issue: when input is piped (`echo "text" | gemini`), `process.stdin.isTTY` is false, making `setRawMode()` unavailable. But `--prompt-interactive` requires TTY interaction, creating a logical conflict.

## Dive Deeper

**The Problem:**
```bash
echo "say '1'." | gemini --prompt-interactive="say '2'." say '3'.
```

This command is logically contradictory:
- Piped input: `echo "say '1'."` provides predetermined input (no TTY)
- Interactive flag: `--prompt-interactive` expects real-time user interaction (requires TTY)
- The original code tried to call `setRawMode()` on a non-TTY stdin, causing the crash

**The Fix:**
Added early validation in `main()` to detect when `--prompt-interactive` is used with piped input and show a clear error instead of attempting to set raw mode on a non-TTY stream.

**Added Test Coverage:**
- Test for the specific crash scenario: piped input + `--prompt-interactive`
- Test verifies proper error message is shown instead of crash
- Test ensures no API requests are made when the error condition is detected

**Correct Usage Patterns:**
- Interactive only: `gemini --prompt-interactive="say '2'." say '3'.`
- Piped only: `echo "say '1'." | gemini say '3'.`
- Mixed non-interactive: `echo "say '1'." | gemini --prompt="say '2'. say '3'."`

## Reviewer Test Plan

1. **Test the original crash scenario:**
   ```bash
   echo "say '1'." | gemini --prompt-interactive="say '2'." say '3'.
   ```
   Should show clear error message instead of crashing

2. **Verify correct usage still works:**
   ```bash
   gemini --prompt-interactive="say hello"  # Should work
   echo "test" | gemini "respond to this"   # Should work
   ```

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ✅  | ❓  | ❓  |

## Linked issues / bugs

Fixes #8304


allenhutchison (approved):
Overall this looks great. Thanks for the contribution. This area will see additional changes in October as we move to standardize on our positional argument flag for input, but that won't change the need for this fix.

PR #8455, opened 2025-09-15
feat(routing): Introduce Classifier-based Model Routing Strategy

abhipatel12:
## TLDR

This PR introduces a new `ClassifierStrategy` to the `ModelRouterService`. Its purpose is to intelligently and dynamically route user prompts to the most appropriate model based on the complexity of the request. Simple, direct tasks are routed to the faster `Flash` model, while complex, strategic, or ambiguous tasks are sent to the more powerful `Pro` model.

## Dive Deeper

The `ClassifierStrategy` is now a key component of the `CompositeStrategy` chain, placed after the `OverrideStrategy` (to respect user-defined models) and before the `DefaultStrategy`.

> [!NOTE]
> The classifier strategy is only used if the experiment feature is enabled in `settings.json`: 
> ```json
> {
>    "experimental": {
>      "useModelRouter": true
>    } 
>} 
> ```

Key implementation details:
- **Classification Model:** It uses the lightweight `gemini-2.5-flash-lite` model to perform the classification, ensuring the routing decision is fast and low-cost.
- **System Prompt:** A detailed system prompt with a complexity rubric and few-shot examples guides the classifier to choose between "simple" (`flash`) and "complex" (`pro`).
- **Context Filtering:** To provide a clean signal to the classifier, the strategy filters out tool-related messages (`functionCall` and `functionResponse`) from the recent conversation history.
- **Model Workaround:** A temporary workaround has been added to address a model issue where using `Flash` in the first few turns of a conversation can lead to errors. For conversations with fewer than 5 turns, the strategy will use `flash-lite` even for simple tasks.

## Reviewer Test Plan

To validate these changes, pull down the branch and run the Gemini CLI. The routing decisions are logged to the console, so you can observe the strategy in action.

**Note:** Ensure that the experiment feature is enabled`"useModelRouter": true` in your `settings.json` to ensure you have the router enabled with the classifier.

1.  **Test Simple Prompts (Expect `flash` model):**
    *   `list the files in this directory`
    *   `what does the package.json say?`
    *   `who was the first person on the moon?`
    *   **Expected Outcome:** `/stats` should show the flash (lite or regular) model being used.

2.  **Test Complex Prompts (Expect `pro` model):**
    *   `how should I architect a new database schema for a social media app?`
    *   `I'm getting a 'TypeError: cannot read properties of undefined'. Can you help me figure out why?`
    *   `refactor this function to be more idiomatic and performant: [paste code]`
    *   **Expected Outcome:** `/stats` should show Gemini 2.5 pro model being used.

3.  **Test Override Strategy:**
    *   Run a simple prompt but force the pro model: `gemini --model gemini-2.5-pro "how do I build a spaceship?"`
    *   **Expected Outcome:** The router should use the specified `pro` model.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  |✅| ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt |✅| -   | -   |

## Linked issues / bugs

### DIFF
--- a/packages/core/src/routing/strategies/classifierStrategy.ts
+++ b/packages/core/src/routing/strategies/classifierStrategy.ts
@@ -4,6 +4,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+import { z } from 'zod';
 import type { BaseLlmClient } from '../../core/baseLlmClient.js';
 import { promptIdContext } from '../../utils/promptIdContext.js';
 import type {
@@ -132,10 +133,10 @@ const RESPONSE_SCHEMA = {
   required: ['reasoning', 'model_choice'],
 };
 
-interface ClassifierResponse {
-  reasoning: string;
-  model_choice: typeof FLASH_MODEL | typeof PRO_MODEL;
-}
+const ClassifierResponseSchema = z.object({
+  reasoning: z.string(),
+  model_choice: z.enum([FLASH_MODEL, PRO_MODEL]),
+});
 
 export class ClassifierStrategy implements RoutingStrategy {
   readonly name = 'classifier';
@@ -168,7 +169,7 @@ export class ClassifierStrategy implements RoutingStrategy {
       // Take the last N turns from the *cleaned* history.
       const finalHistory = cleanHistory.slice(-HISTORY_TURNS_FOR_CONTEXT);
 
-      const routerResponse = (await baseLlmClient.generateJson({
+      const jsonResponse = await baseLlmClient.generateJson({
         contents: [...finalHistory, createUserContent(context.request)],
         schema: RESPONSE_SCHEMA,
         model: DEFAULT_GEMINI_FLASH_LITE_MODEL,
@@ -176,7 +177,9 @@ export class ClassifierStrategy implements RoutingStrategy {
         config: CLASSIFIER_GENERATION_CONFIG,
         abortSignal: context.signal,
         promptId,
-      })) as unknown as ClassifierResponse;
+      });
+
+      const routerResponse = ClassifierResponseSchema.parse(jsonResponse);
 
       const reasoning = routerResponse.reasoning;
       const latencyMs = Date.now() - startTime;
@@ -203,11 +206,7 @@ export class ClassifierStrategy implements RoutingStrategy {
     } catch (error) {
       // If the classifier fails for any reason (API error, parsing error, etc.),
       // we log it and return null to allow the composite strategy to proceed.
-      console.warn(
-        `[Routing] ClassifierStrategy failed: ${
-          error instanceof Error ? error.message : String(error)
-        }`,
-      );
+      console.warn(`[Routing] ClassifierStrategy failed:`, error);
       return null;
     }
   }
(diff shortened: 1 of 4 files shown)


mrcabbage972 (approved):

packages/core/src/routing/strategies/classifierStrategy.ts
```diff
@@ -0,0 +1,214 @@
...
+      const historySlice = context.history.slice(-HISTORY_SEARCH_WINDOW);
+
+      // Filter out tool-related turns.
+      // TODO - Consider using function req/res if they help accuracy.
+      const cleanHistory = historySlice.filter(
+        (content) => !isFunctionCall(content) && !isFunctionResponse(content),
+      );
+
+      // Take the last N turns from the *cleaned* history.
+      const finalHistory = cleanHistory.slice(-HISTORY_TURNS_FOR_CONTEXT);
```
anj-s, line 170: what if the filtered cleanHistory is less than number of turns required?

packages/core/src/routing/strategies/classifierStrategy.ts
```diff
@@ -0,0 +1,219 @@
...
+
+      const routerResponse = (await baseLlmClient.generateJson({
+        contents: [...finalHistory, createUserContent(context.request)],
+        schema: RESPONSE_SCHEMA,
+        model: DEFAULT_GEMINI_FLASH_LITE_MODEL,
+        systemInstruction: CLASSIFIER_SYSTEM_PROMPT,
+        config: CLASSIFIER_GENERATION_CONFIG,
+        abortSignal: context.signal,
+        promptId,
+      })) as unknown as ClassifierResponse;
```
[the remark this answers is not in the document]
anj-s: Should we address this?

packages/core/src/routing/strategies/classifierStrategy.ts
```diff
@@ -0,0 +1,214 @@
...
+
+      const reasoning = routerResponse.reasoning;
+      const latencyMs = Date.now() - startTime;
```
anj-s, line 185: why do we need to measure this Vs looking at Flash-lite latency measurements?

anj-s (approved):

packages/core/src/routing/strategies/classifierStrategy.ts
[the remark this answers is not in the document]
abhipatel12: Yep fair point. Using our existing dep on zod to cover this case now.

packages/core/src/routing/strategies/classifierStrategy.ts
abhipatel12: That should work fine. If `HISTORY_TURNS_FOR_CONTEXT` is larger than the size of the array, it should just return whatever is in the array.

packages/core/src/routing/strategies/classifierStrategy.ts
abhipatel12: Per offline: 

We can def check the API response logs for flash lite latency, but then we need to filter to check for what the the request looks like in case people are using flash lite for other cases (utility calls and/or users forcing that model)

Additionally, if we added different types of strategies we may want to see the overall latency of a routing attempt since it isn't exclusive to the classifier strategy.

abhipatel12 (commented):
Thanks for the review! Addressed all the comments!

Issue #8337, opened 2025-09-12
Add a 'workspacePath' variable for extensions


chrstnb:
Extensions support [variables](https://google-gemini.github.io/gemini-cli/docs/extension.html#variables)-- we should add a new one for the current workspace path

PR #8482, opened 2025-09-15
fix(cli): add workspacePath to extension variables

hritan:
## TLDR

Add a 'workspacePath' variable for extensions

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes - https://github.com/google-gemini/gemini-cli/issues/8337


packages/cli/src/config/extension.ts
```diff
@@ -188,15 +188,18 @@ export function loadExtensionsFromDir(dir: string): Extension[] {
...
     if (extension != null) {
       extensions.push(extension);
     }
   }
   return extensions;
 }
 
-export function loadExtension(extensionDir: string): Extension | null {
+export function loadExtension(
+  extensionDir: string,
```
chrstnb: Nit: let's add a struct for this to variableSchema.ts so we can pass around the variableConfig as a single param instead of adding to the params each time we add a new variable type

chrstnb:
Can you also update `extension.md`?

packages/cli/src/config/extension.ts
hritan: Great suggestion. 
Done

chrstnb (approved):

PR #8491, opened 2025-09-15
feat(vscode-ide-companion): add auth token validation to IDE server

skeshive:
## TLDR

This pull request enhances the security of the IDE companion by adding a bearer token authentication mechanism to the IDE server. The server now generates a unique token on startup, writes it to the discovery file, and validates this token on incoming requests from the CLI. The implementation is backward compatible, allowing older CLIs that don't send the token to continue functioning.

### DIFF
--- a/packages/vscode-ide-companion/src/ide-server.ts
+++ b/packages/vscode-ide-companion/src/ide-server.ts
@@ -121,10 +121,6 @@ export class IDEServer {
       const app = express();
       app.use(express.json({ limit: '10mb' }));
 
-      // The CLI will send the token in the authorization header.
-      // Older CLIs will not send this header.
-      // To maintain backwards compatibility, we only validate the token
-      // if the header is present.
       app.use((req, res, next) => {
         const authHeader = req.headers.authorization;
         if (authHeader) {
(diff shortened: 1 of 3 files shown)


packages/vscode-ide-companion/src/ide-server.ts
```diff
@@ -32,6 +32,7 @@ async function writePortAndWorkspace(
   port: number,
   portFile: string,
   ppidPortFile: string,
+  authToken: string,
```
cornmander: can you make this take an argument object? We are trying to move away from the large list of position arguments of the same type cc @richieforeman

packages/vscode-ide-companion/src/ide-server.ts
```diff
@@ -49,7 +50,12 @@ async function writePortAndWorkspace(
     workspacePath,
   );
 
-  const content = JSON.stringify({ port, workspacePath, ppid: process.ppid });
+  const content = JSON.stringify({
+    port,
+    workspacePath,
+    ppid: process.ppid,
+    authToken,
+  });
```
cornmander, line 67: make sure to make the file 600

cornmander (approved):

PR #8496, opened 2025-09-15
fix LS test.

hminooei:
Getting errors for this test on the main branch.


hminooei:
After removing gemini-cli and getting a fresh clone, the main `npm run preflight` finished without the error. So closing this PR.

Issue #8372, opened 2025-09-12
Parallel edit/replace on the same file cause errors/overrides 


silviojr:
### What happened?

Sometimes, the gemini-cli will attempt to do parallel edits on the file. This will cause either some of these edits not to happen, or some garbage would appear at the end of the file (or both?).

For example, whenever the agent decides to do multi-edits, the subsequente edits may override the previous ones. 

**How to reproduce?** 

Create a file. Something like: 

`$> for i in $(seq 1 5000); do   echo "This is line number $i" >> myfile.txt; done`

Then ask gemini to edit it: 

`Read myfile.txt Replace every line that contains a number that is divisible by 1001 with three lines  'AAA\nBBB\nCCC'. use multiple replace calls too to modify the file simultaneously`

You will see multiple edits being applied/asking to be confirmed in parallel. Each one will override or mess with the previous. 

<img width="630" height="329" alt="Image" src="https://github.com/user-attachments/assets/fc1d041d-096e-4598-8712-cea151e34e27" />

### What did you expect to happen?

Edits should not override each other. 

### Client information

<details>
<summary>Client Information</summary>

Run `gemini` to enter the interactive CLI, then run the `/about` command.

```console
> /about
# paste output here
```

</details>

silviojr:
First step is to verify if this is caused by smart-edit feature or edit.ts as well.

silviojr:
I could reproduce using the standard "edit" tool. 

I actually could only reproduce using the standard edit tool.

PR #8497, opened 2025-09-15
Remove unused code.

scidomino:
## TLDR

Remove unused code.

## Dive Deeper

`recordHistory()` had a bunch of unnecessary logic:

* logic for combining Content objects that's not needed because it's only caller only passes a single content object
* logic for possibly adding the user content despite the userContent always being added before the call was made.. 
* logic for checking for Parts that have both `text` and `functionCall` which the documentation says is not possible.

My guess is all this functionality was from a time when `recordHistory()` was being called by the `sendMessage()` which was recently removed. Now it is only called by `sendMessageStream()`.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | x  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

For #8372


packages/core/src/core/geminiChat.ts
```diff
@@ -560,97 +557,22 @@ export class GeminiChat {
...
-        newTurn.parts
-      ) {
-        lastTurnInFinal.parts.push(...newTurn.parts);
+      const lastPart = consolidatedParts[consolidatedParts.length - 1];
+      if (lastPart?.text !== undefined && part.text !== undefined) {
+        lastPart.text += part.text;
       } else {
-        finalModelTurns.push(newTurn);
+        consolidatedParts.push({ ...part });
       }
```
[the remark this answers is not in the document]
scidomino: The documentation says that this will never happen:

"Exactly one field within a Part should be set, representing the specific type  of content being conveyed. Using multiple fields within the same Part instance is considered invalid."

packages/core/src/core/geminiChat.ts
```diff
@@ -560,97 +559,18 @@ export class GeminiChat {
...
-      // Consolidate this new turn with the PREVIOUS turn if they are adjacent model turns.
-      if (
-        lastTurnInFinal &&
-        lastTurnInFinal.role === 'model' &&
-        newTurn.role === 'model' &&
-        lastTurnInFinal.parts && // SAFETY CHECK: Ensure the destination has a parts array.
-        newTurn.parts
-      ) {
-        lastTurnInFinal.parts.push(...newTurn.parts);
+    // String thoughts and consolidate text parts.
```
jacob314, line 575: this one scares. will defer to @anj-s on it.

packages/core/src/core/geminiChat.ts
```diff
@@ -560,97 +559,18 @@ export class GeminiChat {
...
+      if (lastPart?.text && part.text) {
+        lastPart.text += part.text;
       } else {
-        finalModelTurns.push(newTurn);
+        consolidatedParts.push(part);
       }
     }
 
-    // Part 3: Add the processed model turns to the history, with one final consolidation pass.
-    if (finalModelTurns.length > 0) {
```
anj-s, line 617: There is a reason why we check all these different things. Can we discuss before we check in this change?

scidomino: resolved offline.

packages/core/src/core/geminiChat.ts
scidomino: resolved with anjali

anj-s (approved):
Thank you!

PR #8498, opened 2025-09-15
Add simple extensions release flow support

### LINKED ISSUE #7014: Implement lightweight github actions-based extension release process
https://github.com/allenhutchison/obsidian-gemini

Inspired by the obsidian release process: https://docs.obsidian.md/Plugins/Releasing/Release+your+plugin+with+GitHub+Actions

chrstnb:
## TLDR

The primary change in this branch is the introduction of a new method for distributing and installing Gemini CLI extensions: **GitHub Releases**. This provides a more reliable installation experience for users by using pre-built binaries, avoiding the need to clone and build from source.

### New Behavior for Users

-   **Streamlined Extension Installation:** When a user installs an extension from a GitHub URL, the CLI will now first attempt to find a pre-built release asset that matches the user's operating system (macOS, Linux, Windows) and architecture (x64, arm64).
-   **Automatic Fallback:** If a suitable release asset is not found, the CLI automatically falls back to the previous behavior of cloning the repository using Git.
-   **More Efficient Updates:** The extension update process also leverages this new release-based flow, making updates more efficient.

### New Capabilities for Extension Developers

-   **New Documentation:** A new guide, `docs/extension-releasing.md`, has been added. It explains how developers can package and release their extensions using GitHub Releases.
-   **Release Asset Convention:** The documentation specifies a clear naming convention for release assets (e.g., `my-tool-darwin-arm64.tar.gz`) that developers must follow for the CLI to automatically discover them.
-   **Example Workflow:** The new guide includes a sample GitHub Actions workflow to automate the process of building, packaging, and creating a GitHub Release for an extension.

### Code Refactoring

-   **Centralized GitHub Logic:** All logic for interacting with Git and the GitHub API has been refactored and moved into a new, dedicated module: `packages/cli/src/config/extensions/github.ts`.
-   **Authenticated Requests:** The new module includes support for using a `GITHUB_TOKEN` environment variable to make authenticated requests, allowing users to install and update extensions from private GitHub repositories.
-   **Improved Code Structure:** The core extension management logic in `packages/cli/src/config/extension.ts` has been cleaned up and simplified by delegating Git/GitHub operations to the new module.

This update significantly improves the extension ecosystem by providing a more robust and user-friendly distribution mechanism.
## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes #7014


packages/cli/src/config/git.ts
```diff
@@ -0,0 +1,257 @@
...
+    return platformAsset;
+  }
+
+  // Check for generic asset if only one is available
+  const genericAsset = assets.find(
+    (asset) =>
+      !asset.name.includes('darwin') &&
+      !asset.name.includes('linux') &&
+      !asset.name.includes('win32'),
+  );
```
kurtisvg: nit: these seems potentially problematic if the name of the extension includes "linux" or "darwin" in it. You might be better off adoption a format of "{platform}.{arch}.{name}.{extension}" -- even if the format contains one of those names you can always check first match for platform and second for arch.

docs/extension-releasing.md
averikitsch, line 1: Similar to installing from main, we will need to incorporate a versioning strategy. Currently, we will have to delete and overwrite each release with this pattern

chrstnb: This resolves 'latest' to the latest version! See https://github.com/chrstnb/extension

Unless I'm missing something about what you're saying?

packages/cli/src/config/extensions/github.ts
```diff
@@ -0,0 +1,328 @@
...
+    let sourceUrl = installMetadata.source;
+    const token = getGitHubToken();
+    if (token) {
+      try {
+        const parsedUrl = new URL(sourceUrl);
+        if (
+          parsedUrl.protocol === 'https:' &&
+          parsedUrl.hostname === 'github.com'
+        ) {
+          parsedUrl.username = token;
```
jakemac53: should probably check that there isn't already a username

packages/cli/src/config/extensions/github.ts
```diff
@@ -0,0 +1,328 @@
...
+      } else {
+        return ExtensionUpdateState.UPDATE_AVAILABLE;
+      }
+    } else {
+      const { source, ref } = installMetadata;
+      if (!source) {
+        return ExtensionUpdateState.ERROR;
+      }
+      const parts = source.split('/');
+      const owner = parts.at(-2);
```
jakemac53: Probably would be good to validate some things about `source` here - it seems like we are assuming it has exactly two segments (org/repo). But the way it's coded here it would allow an arbitrary number of segments and possibly have some unexpected behavior.

packages/cli/src/config/extensions/github.ts
```diff
@@ -0,0 +1,328 @@
...
+      const repo = parts.at(-1)?.replace('.git', '');
+
+      if (!owner || !repo) {
+        throw new Error(`Invalid GitHub repository source: ${source}`);
+      }
+
+      const url = `https://api.github.com/repos/${owner}/${repo}/releases/latest`;
+      const releaseData = await fetchJson(url);
+      console.log(releaseData.tag_name);
```
jakemac53: nit: probably delete these logs :). Or log at `debug` level with a bit more info.

packages/cli/src/config/extensions/github.ts
```diff
@@ -0,0 +1,328 @@
...
+}
+
+export async function downloadFromGitHubRelease(
+  installMetadata: ExtensionInstallMetadata,
+  destination: string,
+): Promise<string> {
+  const { source, ref } = installMetadata;
+  const parts = source.split('/');
+  const owner = parts.at(-2);
+  const repo = parts.at(-1)?.replace('.git', '');
```
jakemac53: nit: same comment as above, I would just extract a helper for this too

packages/cli/src/config/extensions/github.ts
```diff
@@ -0,0 +1,328 @@
...
+      !releaseData.assets ||
+      releaseData.assets.length === 0
+    ) {
+      throw new Error(
+        `No release assets found for ${owner}/${repo} at tag ${tag}`,
+      );
+    }
+
+    const asset = findReleaseAsset(releaseData.assets);
+    if (!asset) {
```
jakemac53, line 198: Should we require an asset like this? GitHub releases can be a good way of doing your releases even if you don't require any built assets.

packages/cli/src/config/extensions/github.ts
```diff
@@ -0,0 +1,328 @@
...
+      }
+      const parts = source.split('/');
+      const owner = parts.at(-2);
+      const repo = parts.at(-1)?.replace('.git', '');
+
+      if (!owner || !repo) {
+        throw new Error(`Invalid GitHub repository source: ${source}`);
+      }
+
+      const url = `https://api.github.com/repos/${owner}/${repo}/releases/latest`;
```
jakemac53: Below, we pass the `ref` instead of `/latest` (but default to "latest" for the ref). This should likely do the same?

packages/cli/src/config/extensions/github.ts
```diff
@@ -0,0 +1,328 @@
...
+  return new Promise((resolve, reject) => {
+    https
+      .get(url, { headers }, (res) => {
+        if (res.statusCode !== 200) {
+          return reject(
+            new Error(`Request failed with status code ${res.statusCode}`),
+          );
+        }
+        let data = '';
+        res.on('data', (chunk) => (data += chunk));
```
jakemac53: Generally building the data up as an array and then joining at the end should be preferable here  (although I don't know about JS specifically).

packages/cli/src/config/extensions/github.ts
```diff
@@ -0,0 +1,328 @@
...
+        }
+        const file = fs.createWriteStream(dest);
+        res.pipe(file);
+        file.on('finish', () => file.close(resolve as () => void));
+      })
+      .on('error', reject);
+  });
+}
+
+function extractFile(file: string, dest: string) {
```
jakemac53, line 338: should this function error if it doesn't understand the extension?

packages/cli/src/config/extension.ts
```diff
@@ -650,13 +644,15 @@ export async function updateExtension(
   cwd: string = process.cwd(),
   setExtensionUpdateState: (updateState: ExtensionUpdateState) => void,
 ): Promise<ExtensionUpdateInfo> {
-  if (!extension.type) {
+  const installMetadata = loadInstallMetadata(extension.path);
```
jakemac53, line 655: Why do we need to reload the install metadata here?

docs/extension-releasing.md
```diff
@@ -0,0 +1,86 @@
...
+
+Gemini CLI extensions can be distributed as pre-built binaries through GitHub Releases. This provides a faster and more reliable installation experience for users, as it avoids the need to clone the repository and build the extension from source.
+
+## Creating a GitHub release
+
+1.  **Navigate to your repository's "Releases" page** on GitHub.
+2.  **Click "Draft a new release".**
+3.  **Choose a tag.** It's recommended to use semantic versioning (e.g., `v1.0.0`).
+4.  **Add a release title and description.**
+5.  **Upload your release assets.** These are the pre-built binaries for your extension.
```
averikitsch: I believe the binaries need to be tar/zip into the release assets. Attaching them via the GitHub UI seems to only add them to the list of assets.

averikitsch (commented):
This works for me in public repos. It is overwriting the version in the gemini-extension.json with the tag/release name. Is that expected?

packages/cli/src/config/extensions/github.ts
```diff
@@ -0,0 +1,328 @@
...
+
+      if (typeof lsRemoteOutput !== 'string' || lsRemoteOutput.trim() === '') {
+        console.error(`Git ref ${refToCheck} not found.`);
+        return ExtensionUpdateState.ERROR;
+      }
+
+      const remoteHash = lsRemoteOutput.split('\t')[0];
+      const localHash = await git.revparse(['HEAD']);
+
+      if (!remoteHash) {
```
shishu314, line 143: Nit: We can probably reduce the nesting here for readability without using `else if` and `else`

packages/cli/src/config/extension.ts
chrstnb: Because I removed this value on the GeminiCLIExtension-- I guess if we need to we can add it to that interface, but would prefer to just load when needed

packages/cli/src/config/extensions/github.ts
chrstnb: good call!

packages/cli/src/config/extension.ts
jakemac53: This may be wrong though when for instance doing `/extensions list`. Because it will end up showing the latest installed version and not the version actually loaded into the CLI (which may be different).

packages/cli/src/config/extensions/github.ts
chrstnb: I think this would be a good fast-follow update!

docs/extension-releasing.md
averikitsch: This is great. Thanks!

docs/extension-releasing.md
```diff
@@ -4,11 +4,8 @@ Gemini CLI extensions can be distributed as pre-built binaries through GitHub Re
 
 ## Creating a GitHub release
 
-1.  **Navigate to your repository's "Releases" page** on GitHub.
-2.  **Click "Draft a new release".**
-3.  **Choose a tag.** It's recommended to use semantic versioning (e.g., `v1.0.0`).
-4.  **Add a release title and description.**
-5.  **Upload your release assets.** These are the pre-built binaries for your extension.
+1. Tag your release branch locally using the release version `git tag v0.0.1`
```
jakemac53: Was this intentionally changed? Just pushing a tag doesn't create a github release.

packages/cli/src/config/extensions/github.ts
```diff
@@ -72,6 +74,23 @@ export async function cloneFromGit(
   }
 }
 
+function parseGitHubRepo(source: string): { owner: string; repo: string } {
+  // The source should be "owner/repo" or a full GitHub URL.
+  const parts = source.split('/');
+  if (!source.includes('://') && parts.length !== 2) {
```
jakemac53, line 80: This could be a followup but we should probably just properly parse this as a URI and pull out the path segments, and then check that we get two.

Or, just simplify it and force exactly owner/repo format and don't allow the URI syntax.

packages/cli/src/config/extension.ts
chrstnb: lol I was trying to figure out why you added that info there. I think we may want different types in the future but for now just combining them.

packages/cli/src/config/extension.ts
```diff
@@ -305,14 +300,14 @@ export function annotateActiveExtensions(
...
   );
-
   const annotatedExtensions: GeminiCLIExtension[] = [];
   if (enabledExtensionNames.length === 0) {
     return extensions.map((extension) => ({
       name: extension.config.name,
       version: extension.config.version,
       isActive: manager.isEnabled(extension.config.name, workspaceDir),
       path: extension.path,
+      installMetadata: loadInstallMetadata(extension.path),
```
jakemac53: Can this just be `extension.installMetadata`? (same below)

packages/cli/src/config/extensions/github.ts
chrstnb: yes def-- let's do in a fast-follow

jakemac53 (approved):

cornmander (approved):

chrstnb:
@averikitsch yes-- trying to figure out the best way to handle this-- will address further in a follow-up

Issue #8487, opened 2025-09-15
Support MCP Servers on Cloud Run with IAP enabled


adamfweidman:
### What happened?

tl;dr MCP Servers deployed to cloud run with IAP enabled directly encounter authentication issues

See go/gemini-cli-cloud-run-iap-issue for more details

### What did you expect to happen?

To use MCP Servers on cloud run whether or not IAP is enabled.

### Client information

<details>
<summary>Client Information</summary>

Run `gemini` to enter the interactive CLI, then run the `/about` command.

```console
> /about
# paste output here
```

</details>

### Anything else we need to know?

Implementation Plan:
- Define GoogleIAPProvider as a new MCP Authentication option for MCP Server
- Update MCPServerConfig and necessary types
- Plumb remaining logic through

piyushnigam:
Cloud run now supports IAP-powered MCP auth: https://docs.cloud.google.com/run/docs/ai/authenticate-mcp-servers

PR #8505, opened 2025-09-15
feat(iap support): Add service account impersonation provider to MCPServers to support IAP on Cloud Run

adamfweidman:
## TLDR

This pull request creates `ServiceAccountImpersonationProvider` as a new MCP Auth Option. Specifically meant to solve: go/gemini-cli-cloud-run-iap-issue.

## Dive Deeper

This was originally created to allow MCP Servers running on cloud run with IAP enabled, but its functionality is generic enough to be used for any OIDC-based service account impersonation.

If/when we implement OIDC Auth (3-legged flow and token storge), this solution can and should be migrated to that.

## Reviewer Test Plan

1.  Create and deploy an MCP Server to cloud run and enable IAP in the secuirty tab.
2.  Create an OAuth Client and copy the Client ID.
3.  Create a service account and grant it the "Service Account Token Creator" role.
4. Grant yourself necessary permissions to [impersonate an SA](https://cloud.google.com/docs/authentication/use-service-account-impersonation).
5.  Configure an MCP server in your `~/.gemini/config.json` with `authProviderType: 'service_account_impersonation'` and provide the `targetAudience`(clientId) and `targetServiceAccount from above.
6.  Run the CLI and ensure that it can connect to the MCP server and that the tools are discovered.
7.  Run the unit tests to ensure that all tests pass via `npm test`.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | x  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

[#8487](https://github.com/google-gemini/gemini-cli/issues/8487)


packages/core/src/mcp/sa-impersonation-provider.ts
```diff
@@ -0,0 +1,118 @@
...
+import type { MCPServerConfig } from '../config/config.js';
+import type { OAuthClientProvider } from '@modelcontextprotocol/sdk/client/auth.js';
+
+const createIamApiUrl = (targetSA: string) =>
+  `https://iamcredentials.googleapis.com/v1/projects/-/serviceAccounts/${encodeURIComponent(targetSA)}:generateIdToken`;
+
+export class ServiceAccountImpersonationProvider
+  implements OAuthClientProvider
+{
+  private readonly targetUrl: string | undefined;
```
richieforeman: nit: Prefer 

`private readonly targetUrl?: string;`

packages/core/src/mcp/sa-impersonation-provider.ts
```diff
@@ -0,0 +1,118 @@
...
+  OAuthClientInformation,
+  OAuthClientInformationFull,
+  OAuthClientMetadata,
+  OAuthTokens,
+} from '@modelcontextprotocol/sdk/shared/auth.js';
+import { GoogleAuth } from 'google-auth-library';
+import type { MCPServerConfig } from '../config/config.js';
+import type { OAuthClientProvider } from '@modelcontextprotocol/sdk/client/auth.js';
+
+const createIamApiUrl = (targetSA: string) =>
```
richieforeman: nit, prefer to use the `function` keyword

packages/core/src/mcp/sa-impersonation-provider.ts
```diff
@@ -0,0 +1,118 @@
...
+  // Properties required by OAuthClientProvider, with no-op values
+  readonly redirectUrl = '';
+  readonly clientMetadata: OAuthClientMetadata = {
+    client_name: 'Gemini CLI (Service Account Impersonation)',
+    redirect_uris: [],
+    grant_types: [],
+    response_types: [],
+    token_endpoint_auth_method: 'none',
+  };
+  private _clientInformation?: OAuthClientInformationFull;
```
richieforeman, line 41: nit: remove leading underscore

packages/core/src/tools/mcp-client.ts
```diff
@@ -1239,6 +1241,31 @@ export async function createTransport(
   mcpServerConfig: MCPServerConfig,
   debugMode: boolean,
 ): Promise<Transport> {
+  if (mcpServerConfig.authProviderType === 'service_account_impersonation') {
```
richieforeman: Rather than matching on the string value here, please use the enum

authProviderType === AuthProviderType.SERVICE_ACCOUNT_IMPERSONATION

richieforeman: Suggestion: Consider updating this to use a switch on authProvider type that way you can do an exhaustive check on AuthProviderType values.

packages/core/src/mcp/sa-impersonation-provider.test.ts
```diff
@@ -0,0 +1,107 @@
...
+    mockRequest.mockResolvedValue({ data: { token: null } });
+
+    const provider = new ServiceAccountImpersonationProvider(validConfig);
+    const tokens = await provider.tokens();
+
+    expect(tokens).toBeUndefined();
+  });
+
+  it('should make a request with the correct parameters', async () => {
+    const config: MCPServerConfig = defaultSAConfig;
```
richieforeman: Nit/opinion: Remove the extraneous local alias and just pass defaultSAConfig directly to the provider.

(For all tests)

packages/core/src/config/config.ts
```diff
@@ -154,12 +154,16 @@ export class MCPServerConfig {
     // OAuth configuration
     readonly oauth?: MCPOAuthConfig,
     readonly authProviderType?: AuthProviderType,
+    // Service Account Configuration
```
richieforeman, line 175: Super nit, for these values, can you provide a doc string that gives some information about expected/allowed values here?  This is a pretty niche thing, and it might be good to have some notes for the next person that needs to touch this code.

I assume that "targetServiceAccount" is likely an email address format, but I'm actually fuzzy on what values `targetAudience` might accept and where I can generate those values.

packages/core/src/tools/mcp-client.ts
```diff
@@ -1239,6 +1241,31 @@ export async function createTransport(
...
+    const transportOptions: SSEClientTransportOptions = {
+      authProvider: provider,
+    };
+
+    const targetUrl = mcpServerConfig.httpUrl || mcpServerConfig.url;
+    if (!targetUrl) {
+      throw new Error('No url or httpUrl found in IAP server config');
+    }
+
+    if (mcpServerConfig.httpUrl) {
```
richieforeman, line 1255: Nit: I know that `targetUrl` is hoisted and technically available here, but I'd probably prefer it if the `new URL()` call was using `mcpServerConfig.httpUrl`.  I had to sort of stop and double check what was up here.

If you heed this comment, I'd consider updating/inline-ing the `mcpServerConfig.httpUrl || mcpServerConfig.url` check above.

packages/core/src/mcp/sa-impersonation-provider.ts
adamfweidman: This causes a name collision with the clientInformation() funciton.

packages/core/src/tools/mcp-client.ts
adamfweidman: will do so in follow up PR

packages/core/src/mcp/sa-impersonation-provider.ts
```diff
@@ -0,0 +1,118 @@
...
+
+  clientInformation(): OAuthClientInformation | undefined {
+    return this._clientInformation;
+  }
+
+  saveClientInformation(clientInformation: OAuthClientInformationFull): void {
+    this._clientInformation = clientInformation;
+  }
+
+  async tokens(): Promise<OAuthTokens | undefined> {
```
cornmander, line 76: This is not ideal because we will be requesting a new token on every request when it should be cached for up to 60m.

adamfweidman: Done

richieforeman (approved):
LG, sorry for the latent review. Still learning how to wrestle my GH notifications :) 👍

PR #8508, opened 2025-09-15
Fix to issue #736 resulting in GEMINI_CLI_AUTH_TYPE not getting populated most of the time in the start_session events

razakhan-dev:
## TLDR

This change resolves the issue#736 which results in the GEMINI_CLI_AUTH_TYPE not getting populated in majority of the cases

## Dive Deeper
As documented in [Issue#736](https://github.com/google-gemini/maintainers-gemini-cli/issues/736), the GEMINI_CLI_AUTH_TYPE is not getting populated around 52% of the times in the start_session events. The main reason behind this issue is that the configuration has not been fully loaded when the start_session event is logged which results in GEMINI_CLI_AUTH_TYPE being empty. 
To resolve this issue, we have made changes in the config.ts resulting in explicitly waiting for the authtype to be populated when the configuration is being initialized.

## Reviewer Test Plan
This particular change can be verified in two ways

- The tester can run any sample code and verify that the auth type is getting populated in all the cases. Sample code to verify via console messages can be found in this [commit](https://github.com/google-gemini/gemini-cli/commit/f4d8668eedc16d60b385e093c482fb8963ae359c#diff-0b28c2db2c61a06883acc73d6a92501d340dde12eac6eb9519446da26750430c) 
- Alternatively, the users can run the evaluation code via the command mentioned here, go/gemini-cli-offline-evals-docs#running-on-forks-branches-and-or-commits and verify that the start_session logs have the GEMINI_CLI_AUTH_TYPE populated via the [plx script](https://plx.corp.google.com/scripts2/script_3f._ff92ac_ff6d_4de9_96d8_fc300f4cf803)

## Linked issues / bugs

- Fixes google-gemini/maintainers-gemini-cli#736


packages/core/src/telemetry/clearcut-logger/clearcut-logger.ts
```diff
@@ -411,6 +411,12 @@ export class ClearcutLogger {
         gemini_cli_key: EventMetadataKey.GEMINI_CLI_START_SESSION_MCP_TOOLS,
         value: event.mcp_tools ? event.mcp_tools : '',
       },
+      {
```
owenofbrien: This is redundant, as the auth type is already logged as part of the addDefaultFields() call in createLogEvent().

packages/core/src/config/config.ts
```diff
@@ -418,6 +418,42 @@ export class Config {
     this.geminiClient = new GeminiClient(this);
   }
 
+  /**
+   * Determines the initial authentication method based on environment variables
+   * and initializes the content generator. This should be called once during
+   * the main `initialize` process.
+   */
+  private async populateAuthType(): Promise<void> {
```
owenofbrien: After looking more thoroughly at how the code currently works, I think the simplest and safest fix may be just to move the logCliConfiguration() call to the end of the refreshAuth() function. A side effect of this approach is that in cases where the user changes from auth method A to auth method B mid-session, we'd end up logging two start_session events: one with auth method A and one with auth method B. This seems odd at first, but it might actually be the right thing to do. Currently, we only log a single start_session event in this case, so we miss the fact that the user re-auth-ed with method B. However, we'd need to think through the ramifications of logging multiple start_session events with the same session id.

packages/core/src/config/config.ts
```diff
@@ -418,6 +418,42 @@ export class Config {
...
+    let initialAuthMethod = undefined;
+
+    // 1. Check for a Gemini API Key first (using bracket notation for type safety)
+    if (process.env['GEMINI_API_KEY']) {
+      initialAuthMethod = AuthType.USE_GEMINI;
+    } 
+    // 2. Else, check for Google Cloud / Vertex AI credentials
```
owenofbrien: There are currently 4 AuthTypes defined in the main branch (LOGIN_WITH_GOOGLE, USE_GEMINI_USE_VERTEX_AI, and CLOUD_SHELL. Two of them are missing here, so throwing the error below will break the CLI for users of the other 2 auth types.

owenofbrien (changes requested):
Let's discuss offline before proceeding.

packages/core/src/telemetry/clearcut-logger/clearcut-logger.ts
razakhan-dev: Removed in the new revision

packages/core/src/telemetry/clearcut-logger/clearcut-logger.ts
```diff
@@ -411,6 +411,12 @@ export class ClearcutLogger {
...
+        gemini_cli_key: EventMetadataKey.GEMINI_CLI_AUTH_TYPE,
+        value: JSON.stringify(
+          this.config?.getContentGeneratorConfig()?.authType,
+        ),
```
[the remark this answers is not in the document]
razakhan-dev: Removed as it was redundant

packages/core/src/config/config.ts
razakhan-dev: Changing in the next revision

packages/core/src/config/config.ts
razakhan-dev: As we are moving to a different strategy this issue is no longer there.

packages/core/src/config/config.ts
```diff
@@ -418,6 +418,42 @@ export class Config {
...
+    else if (process.env['GCLOUD_PROJECT']) {
+      initialAuthMethod = AuthType.LOGIN_WITH_GOOGLE;
+    }
+
```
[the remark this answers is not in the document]
razakhan-dev: No longer an issue.

owenofbrien (approved):

PR #8511, opened 2025-09-15
Add support for auto-updating git extensions

### LINKED ISSUE #7009: Design and implement extension auto-updating

jakemac53:
## TLDR

Adds a setting under `extensions` called `autoUpdate`. This can either be just a boolean or a map of extension names to booleans. The default is `false`.

On startup, we now always check for updates, and if auto-update is enabled we will auto-update and log a message once it completes. If auto-updating is disabled we log a message that there is an update available. The messages look like this (but are the same size):

image

image

## Dive Deeper

Adds a new `useExtensionUpdates` hook which handles the update state of all extensions. This manages only checking for updates once, and possibly auto-updating if enabled, as well as adding the new history items to the UI.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Closes #7009


packages/cli/src/config/extension.test.ts
```diff
@@ -1401,3 +1548,143 @@ describe('enableExtension', () => {
     expect(activeExtensions[0].name).toBe('ext1');
   });
 });
+
+describe('extension', () => {
+  describe('checkForExtensionUpdate', () => {
```
chrstnb: I say this partly because merge conflicts are going to be a real pain, but also: WDYT of pulling the update logic out into the config/extensions/ directory into its own file?

packages/cli/src/config/extension.ts
```diff
@@ -296,6 +299,17 @@ export function annotateActiveExtensions(
 ): GeminiCLIExtension[] {
   const settings = loadSettings(workspaceDir).merged;
   const disabledExtensions = settings.extensions?.disabled ?? [];
+  const autoUpdateSettings = settings.extensions?.autoUpdate;
```
chrstnb: Can we store this on the extension instead? I've been discouraged from adding more to settings. Maybe it lives on installMetadata?

packages/cli/src/config/extension.ts
```diff
@@ -832,18 +865,91 @@ export async function checkForExtensionUpdate(
...
         extension.name
       }": ${getErrorMessage(error)}`,
     );
-    return ExtensionUpdateState.ERROR;
+    setExtensionUpdateState(ExtensionUpdateState.ERROR);
+    return;
   }
 }
+
+export const useExtensionUpdates = (
```
chrstnb: This should be in a hook file in cli/src/hooks

packages/cli/src/config/extension.test.ts
jakemac53: SGTM, done

packages/cli/src/config/extension.ts
jakemac53: done

packages/cli/src/config/extension.test.ts
```diff
@@ -454,6 +450,92 @@ describe('extension tests', () => {
       expect(consoleSpy).toHaveBeenCalledWith('Extension not found: ext4');
       consoleSpy.mockRestore();
     });
+
+    describe('autoUpdate', () => {
+      let tempHomeDir: string;
```
chrstnb: you should be able to use the top-level one and remove this and the autoUpdate beforeEach + afterEach

packages/cli/src/test-utils/createExtension.ts
```diff
@@ -0,0 +1,49 @@
...
+import {
+  EXTENSIONS_CONFIG_FILENAME,
+  INSTALL_METADATA_FILENAME,
+} from '../config/extension.js';
+import {
+  type MCPServerConfig,
+  type ExtensionInstallMetadata,
+} from '@google/gemini-cli-core';
+
+export function createExtension({
```
chrstnb, line 18: awesome!

packages/cli/src/ui/commands/types.ts
```diff
@@ -63,9 +63,9 @@ export interface CommandContext {
     setGeminiMdFileCount: (count: number) => void;
     reloadCommands: () => void;
     extensionsUpdateState: Map<string, ExtensionUpdateState>;
-    setExtensionsUpdateState: (
-      updateState: Map<string, ExtensionUpdateState>,
-    ) => void;
+    setExtensionsUpdateState: Dispatch<
+      SetStateAction<Map<string, ExtensionUpdateState>>
```
chrstnb, line 67: nice!

packages/cli/src/config/extensions/update.ts
```diff
@@ -0,0 +1,159 @@
...
+            extensionsUpdateState = new Map(prev);
+            extensionsUpdateState.set(extension.name, updatedState);
+            return extensionsUpdateState;
+          });
+        },
+      );
+    }
+  }
+  return extensionsUpdateState;
+}
```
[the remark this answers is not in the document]
jakemac53: The suggested change would lead to race conditions and stale state due to not actually updating the state until all extensions are done updating. Quite possibly some refactoring should be done here eventually, but it would likely have to be a much larger refactor to not rely on updating an entire Map like this.

packages/cli/src/ui/hooks/useExtensionUpdates.ts
```diff
@@ -0,0 +1,85 @@
...
+        );
+      }
+    }
+  }, [
+    extensions,
+    extensionsUpdateState,
+    setExtensionsUpdateState,
+    addItem,
+    cwd,
+  ]);
```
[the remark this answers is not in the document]
jakemac53: Updated this to use useEffect, the isMounted stuff is not necessary and breaks things.

chrstnb (approved):

PR #8512, opened 2025-09-15
feat(vscode-ide-companion): harden ide-server with CORS and host validation

skeshive:
## TLDR

This pull request makes changes to the IDEServer in the VSCode extension to ensure only intended requests are accepted.

## Dive Deeper

The server, which listens locally to communicate with Gemini CLI, now includes checks to validate incoming requests:
1.  CORS Validation: It blocks all requests that contain an Origin header. Requests from web browsers always include this header, so this change effectively prevents any web browser from successfully sending requests to the server. Only requests that typically come from command-line tools or other server processes (which don't include an Origin header) are permitted.
2.  Host Header Validation: It validates the Host header to ensure it matches the server's specific local address: 127.0.0.1:<port> or localhost:<port>. This means requests must be explicitly targeted at the local server instance.


packages/vscode-ide-companion/src/ide-server.test.ts
```diff
@@ -4,6 +4,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+import request from 'supertest';
```
cornmander: Can you write this test without introducing this new library by directly calling fetch?

skeshive: hopefully gemini cli can 😄

cornmander (approved):

PR #8513, opened 2025-09-15
Serialize function calls that mutate state

### LINKED ISSUE #8372: Parallel edit/replace on the same file cause errors/overrides

scidomino:
## TLDR

To improve model quality, stop reading responses right before the second function call that mutates state.

## Dive Deeper

This should improve function call quality since often edit calls can stomp on each other.

## Reviewer Test Plan

See repro steps in bug report.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | x  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes #8372

### DIFF
--- a/packages/core/src/core/geminiChat.ts
+++ b/packages/core/src/core/geminiChat.ts
@@ -7,13 +7,14 @@
 // DISCLAIMER: This is a copied version of https://github.com/googleapis/js-genai/blob/main/src/chats.ts with the intention of working around a key bug
 // where function responses are not treated as "valid" responses: https://b.corp.google.com/issues/420354090
 
-import type {
+import {
   GenerateContentResponse,
-  Content,
-  GenerateContentConfig,
-  SendMessageParameters,
-  Part,
-  Tool,
+  type Content,
+  type GenerateContentConfig,
+  type SendMessageParameters,
+  type Part,
+  type Tool,
+  FinishReason,
 } from '@google/genai';
 import { toParts } from '../code_assist/converter.js';
 import { createUserContent } from '@google/genai';
@@ -23,7 +24,7 @@ import {
   DEFAULT_GEMINI_FLASH_MODEL,
   getEffectiveModel,
 } from '../config/models.js';
-import { hasCycleInSchema } from '../tools/tools.js';
+import { hasCycleInSchema, MUTATOR_KINDS } from '../tools/tools.js';
 import type { StructuredError } from './turn.js';
 import {
   logContentRetry,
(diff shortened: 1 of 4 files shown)


scidomino:
I am running an eval on this with tag `mutator-1`

scidomino:
I ran an eval and it passed at reasonable values.

packages/core/src/core/geminiChat.ts
```diff
@@ -621,6 +622,64 @@ export class GeminiChat {
...
+      const candidate = chunk.candidates?.[0];
+      const content = candidate?.content;
+      if (!candidate || !content?.parts) {
+        yield chunk;
+        continue;
+      }
+
+      const truncatedParts: Part[] = [];
+      for (const part of content.parts) {
+        if (this.isMutatorFunctionCall(part)) {
```
silviojr, line 649: Can we log this so we know how often this happens and are the tools that tend to be called in parallel?

packages/core/src/core/geminiChat.ts
```diff
@@ -621,6 +622,64 @@ export class GeminiChat {
...
+            // Truncate and return immedaitely.
+            const newChunk = new GenerateContentResponse();
+            newChunk.candidates = [
+              {
+                ...candidate,
+                content: {
+                  ...content,
+                  parts: truncatedParts,
+                },
+                finishReason: FinishReason.STOP,
```
silviojr, line 661: Curious on why not use the FinishReason that came from the response. Do we use "finishReason" somewhere else in the gcli? I am afraid of what the downstream imapct may be.

silviojr (approved):

silviojr:
@scidomino "See repro steps in bug report." - so does that mean you tried to reproduce with this new branch and it was fine?

packages/core/src/core/geminiChat.ts
scidomino: The candidate might not have a FinishReason because it's not necessarily the last message in the stream.

scidomino:
> @scidomino "See repro steps in bug report." - so does that mean you tried to reproduce with this new branch and it was fine?

Yes. Now it does the edits serially so there's no opportunity for the bug.

PR #8541, opened 2025-09-16
feat: add direct Google Cloud telemetry exporters

### LINKED ISSUE #6792: Auto-run Open Telemetry for all sessions

In order to benefit from Open Telemetry, one needs to run the Open Telemetry collector in every session. This makes it hard to get consistent telemetry for all sessions and from all team members.

Suggestion is to add a one-time ask on whether the developer wants to run OTel for their projects, which can be saved on user settings and will determine whether to run OTel in future sessions.

Options:
(1) bundle with binary
(2) build an extension

### Why is this needed?

Generate telemetry consistently on all sessions, so evolution of metrics can be measured over time.

### Additional context

_No response_

### LINKED ISSUE #7204: Let me see all traffic

Add a feature that logs to a file all traffic between Gemini CLI and the Google servers.

### Why is this needed?

I am currently unable to use Gemini CLI because I cannot tell what information is transmitted. This needs to be transparent.

### Additional context

Make sure that, when the option is enabled, the UI will output the path and filename of the log file where I can read all traffic. Don't hide it or make it hard to find.

jerop:
- Add direct GCP exporters for traces, metrics, and logs
- Add useCollector setting to control exporter selection (settings.json only)
- Use direct exporters when target=gcp, OTLP_GOOGLE_CLOUD_PROJECT is set, and useCollector=false
- Fall back to OTLP exporters when useCollector=true for advanced use cases
- Update documentation with simplified configuration guide

Fixes https://github.com/google-gemini/gemini-cli/issues/6792

### DIFF
--- a/packages/core/src/telemetry/gcp-exporters.ts
+++ b/packages/core/src/telemetry/gcp-exporters.ts
@@ -1,0 +1,132 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { TraceExporter } from '@google-cloud/opentelemetry-cloud-trace-exporter';
+import { MetricExporter } from '@google-cloud/opentelemetry-cloud-monitoring-exporter';
+import { Logging } from '@google-cloud/logging';
+import type { Log } from '@google-cloud/logging';
+import { hrTimeToMilliseconds } from '@opentelemetry/core';
+import type { ExportResult } from '@opentelemetry/core';
+import { ExportResultCode } from '@opentelemetry/core';
+import type {
+  ReadableLogRecord,
+  LogRecordExporter,
+} from '@opentelemetry/sdk-logs';
+
+/**
+ * Google Cloud Trace exporter that extends the official trace exporter
+ */
+export class GcpTraceExporter extends TraceExporter {
+  constructor(projectId?: string) {
+    super({
+      projectId,
+      resourceFilter: /^gcp\./,
+    });
+  }
+}
+
+/**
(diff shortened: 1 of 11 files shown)


packages/core/src/telemetry/gcp-exporters.ts
```diff
@@ -0,0 +1,121 @@
...
+    } catch (error) {
+      resultCallback({
+        code: ExportResultCode.FAILED,
+        error: error as Error,
+      });
+    }
+  }
+
+  async forceFlush(): Promise<void> {
+    return Promise.resolve();
```
owenofbrien: Is it okay for the implementation to be missing here?

packages/core/src/telemetry/gcp-exporters.ts
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  async shutdown(): Promise<void> {
+    return Promise.resolve();
```
owenofbrien: Should we call forceFlush on shutdown?

owenofbrien (approved):
Just a couple of questions, but overall, LGTM. Thanks for knocking this out. It will simplify our internal quality debugging workflows greatly!

packages/core/src/telemetry/gcp-exporters.ts
jerop: sure, was a placeholder but just implemented it in this change

packages/core/src/telemetry/gcp-exporters.ts
jerop: yes, added after implementing `forceFlush`

anj-s (approved):

PR #8551, opened 2025-09-16
Conditionally use consent flow

chrstnb:
## TLDR

We call the install function e.g., when updating but also when doing the workspace extensions migration. This ensures we only show this on initial install.

We should sort out later how to convey any changes an update is making to the user.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


shishu314 (approved):
Nit: Should probably add a test for consent flow not being opened when boolean is false

PR #8571, opened 2025-09-16
fix(cli): Fix performance issues on relaunching CLI in inner node pro…

shrutip90:
…cess

## Dive Deeper

CLI tried a couple of different fixes for the issue. The issue got fixed with the addition of process.stdin.pause().

The core issue is about the parent process's event loop continuing to run and interfering with stdin even while it's waiting for the child. The parent process is still technically connected to the terminal's input stream, and its attempts to manage that stream (even passively) are creating contention with the active child process. The solution is to have the parent process explicitly pause its own stdin stream before spawning the child. This tells the parent's event loop to stop processing any data from the terminal, giving the child exclusive control. The parent can then resume the stream once the child exits, just before it terminates itself.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


packages/cli/src/gemini.tsx
```diff
@@ -102,15 +102,33 @@ function getNodeMemoryArgs(config: Config): string[] {
...
+    child.on('close', () => {
+      // Resume stdin before the parent process exits.
+      process.stdin.resume();
+      resolve();
+    });
   });
 
-  await new Promise((resolve) => child.on('close', resolve));
   process.exit(0);
 }
```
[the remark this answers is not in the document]
shrutip90: Done

packages/cli/src/gemini.tsx
```diff
@@ -102,16 +102,49 @@ function getNodeMemoryArgs(config: Config): string[] {
...
 
 async function relaunchWithAdditionalArgs(additionalArgs: string[]) {
-  const nodeArgs = [...additionalArgs, ...process.argv.slice(1)];
-  const newEnv = { ...process.env, GEMINI_CLI_NO_RELAUNCH: 'true' };
+  // If stdin is in raw mode, we need to temporarily disable it so that the
+  // child process can take control of it.
+  const wasRaw = process.stdin.isRaw;
+  if (wasRaw) {
```
jacob314: little nervous about this one. can you double check that setRawMode(false) is required to resolve the issue on your machine.
this may result in some kitty protocol related glitches so has some risk.

packages/cli/src/gemini.tsx
```diff
@@ -102,16 +102,49 @@ function getNodeMemoryArgs(config: Config): string[] {
...
-  await new Promise((resolve) => child.on('close', resolve));
-  process.exit(0);
+  try {
+    const nodeArgs = [...additionalArgs, ...process.argv.slice(1)];
+    const newEnv = { ...process.env, GEMINI_CLI_NO_RELAUNCH: 'true' };
+
+    const child = spawn(process.execPath, nodeArgs, {
+      stdio: 'inherit',
+      env: newEnv,
+      detached: true,
```
jacob314: is this change to detached mode required?

shrutip90: Reverted since its not making a diff on my system.

packages/cli/src/gemini.tsx
shrutip90: Reverted this too, since its not making a diff on my system.

jacob314 (approved):
lgtm

PR #8624, opened 2025-09-17
chore: update erroneous model version added by external contrib

jackwotherspoon:
## TLDR

An external contributor added a test that has a mis-leading and incorrect model number in the test data.

Removing it to not cause confusion or false speculation. 

To be clear, this was test data added by an external contributor and not a Google employee. 

Not trying to rain on anyone's parade ☔ 😆


abhipatel12 (approved):
Thanks for fixing this! LGTM

PR #8629, opened 2025-09-17
unify `/editor` and `/ide` commands

### LINKED ISSUE #6684: Unified UX menu for /ide & /editor
 A single, unified command, /ide, that presents the user with a menu of all available actions for the connected IDE. This would consolidate the functionality of the current /ide and /editor commands, with the goal of eventually deprecating /editor.

### Why is this needed?
Combining these commands simplifies the user experience by providing a single, intuitive entry point for all IDE-related interactions. Users will no longer need to remember two separate commands for similar functions, which reduces confusion and command clutter. A unified menu makes the tool easier to learn and use.

WangWanyue:
Fixes: #6684

## Overview

This PR improves the usability of the /ide command by:

   1. Unifying the `/editor` command into the `/ide` command. The /editor command has been moved to `/ide editor` to 
      reduce the number of top-level commands.
   2. Introducing a **dialog** for IDE integration management. The `/ide integration` command now opens a dialog that 
      allows users to manage their IDE integration settings and view status in a more interactive way. This replaces the previous `status`, `install`, `enable`, and `disable` subcommands.
   3. Improving descriptions for commands to distinguish functionalities and reduce user confusion. 

## Screencasts

VSCode terminal:
Recording 2025-09-18 at 13 54 27

Mac terminal:
Recording 2025-09-18 at 13 58 08

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


WangWanyue:
@skeshive and @jacob314 this incorporated feedback we discussed and ready to review.

miguelsolorio:
A few minor comments: 

1. Do we need to display this info on the right? I imagine this is copying the theme dialog which has a right side for the theme preview, but in this example it feels unnecessary? Like in the theme dialog, we show the user's selection on load and don't have any explicit call out. It might also help make the dialog look a little less busy.

CleanShot 2025-09-19 at 14 51 34@2x

2. Would it make more sense to move the editors not installed to the bottom of the list? Feels a little silly that you the user has to navigate through options that don't work, I might even suggest we remove them completely.
CleanShot 2025-09-19 at 14 55 58

WangWanyue:
@miguelsolorio thanks for the comments! 

For 1, I think we want to callout that editor preference is only for diff viewing and editing, and how it impacts sandbox mode. So user won't mix the editor preference with IDE integration.

For 2, I totally get your point, but I don't have a strong opinion. Now we show those in alphabetical order, sometimes a fixed order might be easier to locate an item. I guess we want to list all supported editors rather than only ones user installed, so user is aware of the capability. 

@skeshive @srithreepo Any thoughts?

packages/cli/src/ui/components/IdeIntegrationDialog.tsx
```diff
@@ -0,0 +1,211 @@
...
+  const { status } = ideClient.getConnectionStatus();
+  const { statusIcon, statusText, details, messageType } = getDetailedIdeStatus(ideClient);
+
+  // Determine available actions based on status
+  const availableActions: Array<{ label: string; value: string; disabled?: boolean }> = [];
+
+  switch (status) {
+    case IDEConnectionStatus.Connected:
+      availableActions.push(
+        { label: 'Status', value: 'status' },
```
skeshive, line 150: since /ide integration displays the status info, feels repetitive to have a status action here

packages/cli/src/ui/components/IdeIntegrationDialog.tsx
```diff
@@ -0,0 +1,211 @@
...
+function getDetailedIdeStatus(ideClient: IdeClient): {
+  statusIcon: string;
+  statusText: string;
+  details: string;
+  messageType: 'info' | 'error';
+} {
+  const connection = ideClient.getConnectionStatus();
+  switch (connection.status) {
+    case IDEConnectionStatus.Connected: {
+      let details = `Connected to ${ideClient.getDetectedIdeDisplayName()}. IDE integration features are active.`;
```
skeshive: this feels a bit repetitive, how about just: 

 🟢 Status: Connected to VS Code

packages/cli/src/ui/components/IdeIntegrationDialog.tsx
```diff
@@ -0,0 +1,211 @@
...
+    }
+    case IDEConnectionStatus.Connecting:
+      return {
+        statusIcon: '🟡',
+        statusText: 'Connecting',
+        details: 'Attempting to connect to IDE companion extension...',
+        messageType: 'info',
+      };
+    default: {
+      let details = `IDE companion extension is not connected.`;
```
skeshive: "IDE companion extension is not connected. IDE integration disabled. To enable it again, run /ide integration. "

hmm how about just "IDE integration disabled. To enable it, run /ide integration and select enable"

skeshive (commented):
could you also update the ctrl+g shortcut in the text above the input box, this currently breaks:

"
ℹCommand '/ide' requires a subcommand. Available:
    - integration: manage settings for native IDE integration
    - editor: set the editor for basic diff viewing and editing"

packages/cli/src/ui/components/IdeIntegrationDialog.tsx
WangWanyue: How do you think about only show status summary (e.g. connected) in the dialog and keep the status action to display status details (e.g. list of files). By doing this progressive disclosure, we can keep the dialog clean, as the file list can make the dialog very busy, . 
image
image

packages/cli/src/ui/components/IdeIntegrationDialog.tsx
WangWanyue: Good catch and fixed it! I found "IDE companion extension is not connected" was appeared in two separate files. I still kept this as a fallback option and updated message when the integration is "disabled". 

Also updated messages for status action to be consistent.

image

image

packages/cli/src/ui/components/IdeIntegrationDialog.tsx
WangWanyue: I think it might be good to callout "IDE integration features" and benefit, I think this would work well! WDTY

image

skeshive: So soon we're going to allow 3P contributed extensions for other editors and those may not enable both features- can we either keep this generic or flag guard for just VS code integrations to avoid confusing users?

packages/cli/src/ui/components/IdeIntegrationDialog.tsx
skeshive: LGTM!

WangWanyue:
> could you also update the ctrl+g shortcut in the text above the input box, this currently breaks:
> 
> " ℹCommand '/ide' requires a subcommand. Available: - integration: manage settings for native IDE integration - editor: set the editor for basic diff viewing and editing"

Added! I improved it a bit, currently crtl+g do nothing and no feedback when IDE is disconnect, I show the disconnect status now, and include instruction on how to enable it. 
image

packages/cli/src/ui/components/IdeIntegrationDialog.tsx
WangWanyue: Sounds good! Let's keep generic, just connect to [ideName], we can further refine with a UXW.
image

skeshive:
overall LGTM but this error message is a bit confusing:

"╭──────────────────────╮
│  > /ide integration  │
╰──────────────────────╯


ℹ🟡 IDE integration enabled but not yet connected. Failed to connect to IDE companion extension in VS Code. Please ensure the 
  extension is running. To install the extension, run /ide install."

skeshive (approved):

skeshive:
actually found a bug- when you run /ide outside of VS code in a standalone terminal, /ide editor does not show up

PR #8647, opened 2025-09-17
refactor(shell): Send AnsiOutput when ShowColor is false

galdawave:
## TLDR

This PR refactors the shell output mechanism to consistently use `AnsiOutput` for all shell-related content, regardless of whether `showColor` is enabled or disabled, when `tool.usePty` is true. This change simplifies the downstream processing of shell content by ensuring a consistent data structure. Additionally, this PR includes several UI and stability improvements for the shell, such as a more informative focus indicator and better handling of the pseudo-terminal (PTY).

## Dive Deeper

- **Standardized Shell Output**: The `shellExecutionService` has been updated to always emit `AnsiOutput`, even when `showColor` is `false` and `tool.usePty` is `true`. Previously, it would send a plain string in this case, which required consumers of the shell content to handle two different data types. This change streamlines the process and makes the shell output more predictable.
- **Improved UI for Shell Focus**: The `ToolMessage` component now displays a more helpful message for focusable shells. When a shell is not focused, it shows `(ctrl+f to focus)` instead of the less descriptive `[Focused]`. This provides clearer instructions to the user on how to interact with the shell.
- **Enhanced PTY Handling**: The pseudo-terminal (PTY) handling has been made more robust. The terminal width and height are now explicitly passed to the shell execution configuration, and the PTY now scrolls to the top upon initialization. A check has also been added to prevent negative `viewportY` values, which could cause rendering issues.
- **Refined Output Processing**: The `useShellCommandProcessor` hook has been updated to differentiate how it accumulates output based on whether a PTY shell is being used. For PTY shells, the output chunk replaces the cumulative stdout (since PTY provides the full screen state), while for standard child processes, chunks are appended.
- **More Robust Diff Rendering**: The condition for rendering file diffs in `ToolMessage` has been updated from a less reliable `!Array.isArray(resultDisplay)` to a more explicit `'fileDiff' in resultDisplay`.
- **Comprehensive Testing**: New tests have been added to `shellExecutionService.test.ts` to ensure that `AnsiOutput` is correctly generated and emitted when `showColor` is `false`, including for multi-line content.

## Reviewer Test Plan

1.  **Verify `showColor: false` behavior**:
    *   In your user-level `config.yaml`, set `shell: { showColor: false, tool: { usePty: true } }`.
    *   Run a command that typically produces colored output (e.g., `ls --color=auto` on Linux/macOS, or a command that uses a library like `chalk`).
    *   Confirm that the output is rendered correctly and without any unexpected artifacts.
2.  **Test the shell focus indicator**:
    *   Run a shell command (e.g., `sleep 5`).
    *   While the command is running, observe the focus indicator. It should display `(ctrl+f to focus)`.
    *   Press `ctrl+f`. The indicator should change to `(Focused)`.
    *   Press `ctrl+f` again to blur the shell. The indicator should revert to `(ctrl+f to focus)`.
3.  **Test scrolling behavior**:
    *   Run a command that generates a large amount of output (e.g., `for i in {1..100}; do echo "line $i"; done`).
    *   Scroll up through the output.
    *   Confirm that the scrolling is smooth and that there are no errors or visual glitches.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ❓ | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


jacob314 (approved):
lgtm

packages/cli/src/ui/components/messages/ToolMessage.tsx
```diff
@@ -90,9 +95,15 @@ export const ToolMessage: React.FC<ToolMessageProps> = ({
...
-        {isThisShellFocused && (
+        {isThisShellFocusable && (
           <Box marginLeft={1}>
-            <Text color={theme.text.accent}>[Focused]</Text>
+            <Text
+              color={
+                isThisShellFocused ? theme.text.accent : theme.text.secondary
+              }
+            >
+              {isThisShellFocused ? '(Focused)' : '(ctrl+f to focus)'}
```
jackwotherspoon: I think having the `(ctrl+f to focus)` in accent color as well makes it more visible and will result in more users figuring it out.

@galz10 I also wonder if we should make `(Focused)` become `(focused, ctrl+f to unfocus)`

jackwotherspoon (changes requested):

packages/cli/src/ui/components/messages/ToolMessage.tsx
galdawave: i feel like `(focused, ctrl+f to unfocus)` would be too crowded but i can make that change.

jackwotherspoon: I'll leave it up to you! 👍 But i like the accent color for both 😄

jackwotherspoon (approved):

PR #8650, opened 2025-09-17
fix(ui): Track last prompt token count in processStream

SandyTao520:
## TLDR

This pull request refactors the UI telemetry service to more accurately track the token count of the last prompt. 

## Dive Deeper

Previously, the token count was updated as a side-effect of processing a generic API response event, it had some logic flaws that resulted in incorrect token counts being reported, because promptTokenCount from utility calls are incorrectly set as well.

The new approach sets the token count from the source of truth at the moment it becomes known:
1.  In `GeminiClient` and `GeminiChat`, the prompt token count from the API response is now used to set the value immediately after a request.
2.  In the `/clear` command, the count is explicitly set to `0`.
3. We're also updating it after compression.

This makes the logic more straightforward and ensures the telemetry data is accurate.

## Reviewer Test Plan

Send a prompt ```@packages/cli Tell me about this code```

See the % context left dropped to around 22%, and a compression would occur, then it will get back to 99-100%.

Also check utility calls like routing, edit correction are not updating % context left.

This will confirm that the token count is being correctly updated after each prompt and properly reset when the context is cleared.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ✅  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |
## Linked issues / bugs

#6106


jacob314 (approved):
lgtm

scidomino (approved):

scidomino:
I tested it and it looks great!

PR #8685, opened 2025-09-18
fix(cli): ctrl c/ctrl d close cli when in dialogs

lifefloating:
## TLDR

Implements proper Ctrl+C exit handling with confirmation for all dialog types. Previously, Ctrl+C did not work consistently when dialogs were open. Now pressing Ctrl+C twice will reliably close dialogs with visual feedback on the first press.

## Dive Deeper

This PR addresses a critical UX issue where users could get stuck in dialogs without a reliable way to exit. The implementation:

- Adds a two-press Ctrl+C system for dialogs: first press shows warning, second press closes
- Implements consistent behavior across all dialog types (settings, editor, theme, privacy notice)
- Adds visual feedback "Press Ctrl+C again to exit" to guide users
- Prevents accidental exits during authentication flows
- Maintains existing Ctrl+C behavior for normal CLI operations
- Includes comprehensive test coverage for all dialog scenarios

## Reviewer Test Plan

1. **Test each dialog type:**
   - Open settings dialog (`/settings`) → Press Ctrl+C once → Verify warning appears → Press Ctrl+C again → Verify dialog closes
   - Open editor dialog (`/editor`) → Repeat above steps
   - Open theme dialog → Repeat above steps
   - Trigger privacy notice → Repeat above steps

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |

## Linked issues / bugs

Fixes #8648


lifefloating:
i tested ok in my local dev

lifefloating:
@jacob314

jacob314 (commented):
I would hope we can solve this a bit more simply with a single handler in AppContainer rather than having this logic in all the dialogs. If that is not possible we can go this way but I would prefer we refactor to achieve that. If that is not feasible, we need to add a base component for all dialogs so this logic does not get repeated.  Otherwise we will miss adding it next time we add a dialog.

lifefloating:
refactor and in a more simple way

Unified Exit Prompt Location

  - Before: Each dialog had its own "Press Ctrl+C again to exit" prompt
  - Now: All exit prompts are unified and displayed in App.tsx

 Removed Redundant Dialog Closing Logic

  - Before: Complex logic in AppContainer.tsx where the second Ctrl+C would check and close various dialogs one by one
  - Now: Deleted all specific dialog closing handling code

packages/cli/src/ui/AppContainer.test.tsx
```diff
@@ -605,4 +605,164 @@ describe('AppContainer State Management', () => {
...
+          settings={mockSettings}
+          version="1.0.0"
+          initializationResult={mockInitResult}
+        />,
+      );
+
+      expect(mockHandleSlashCommand).not.toHaveBeenCalledWith('/quit');
+    });
+
+    it('should prevent exit command when text buffer has content', () => {
```
lifefloating, line 640: ok, I'm going to check the test cases now.

packages/cli/src/ui/AppContainer.tsx
```diff
@@ -865,14 +865,40 @@ Logging in with Google... Please restart Gemini CLI to continue.
...
-        isFolderTrustDialogOpen ||
-        showPrivacyNotice;
-      if (anyDialogOpen) {
+      if (keyMatchers[Command.QUIT](key)) {
+        if (isAuthenticating) {
+          return;
+        }
+
+        const anyDialogOpen =
+          isThemeDialogOpen ||
```
jacob314, line 874: it should be all dialogs. it is equally a bug is you are in the auth dialog and can't shut down gemini CLI by pressing ctrl-C twice.

packages/cli/src/ui/AppContainer.tsx
```diff
@@ -865,14 +865,40 @@ Logging in with Google... Please restart Gemini CLI to continue.
...
-      const anyDialogOpen =
-        isThemeDialogOpen ||
-        isAuthDialogOpen ||
-        isEditorDialogOpen ||
-        isSettingsDialogOpen ||
-        isFolderTrustDialogOpen ||
-        showPrivacyNotice;
-      if (anyDialogOpen) {
+      if (keyMatchers[Command.QUIT](key)) {
+        if (isAuthenticating) {
```
jacob314, line 869: remove this check. we should quit even if the auth dialog is open

packages/cli/src/ui/AppContainer.tsx
```diff
@@ -865,14 +865,28 @@ Logging in with Google... Please restart Gemini CLI to continue.
...
 
-      const anyDialogOpen =
-        isThemeDialogOpen ||
-        isAuthDialogOpen ||
-        isEditorDialogOpen ||
-        isSettingsDialogOpen ||
-        isFolderTrustDialogOpen ||
-        showPrivacyNotice;
-      if (anyDialogOpen) {
+      if (keyMatchers[Command.QUIT](key)) {
```
jacob314, line 868: I pushed a commit simplifying this a little further. I don't think we needed to check for the dialogs at all.

packages/cli/src/ui/AppContainer.tsx
```diff
@@ -865,14 +865,28 @@ Logging in with Google... Please restart Gemini CLI to continue.
...
+            setCtrlCPressedOnce(false);
+            ctrlCTimerRef.current = null;
+          }, CTRL_EXIT_PROMPT_DURATION_MS);
+          return;
+        }
+
+        handleExit(ctrlCPressedOnce, setCtrlCPressedOnce, ctrlCTimerRef);
+        return;
+      } else if (keyMatchers[Command.EXIT](key)) {
+        if (buffer.text.length > 0) {
```
jacob314, line 886: there is one case missing here. Pressing ctrl-D twice will exit a dialog but we need to show the ctrlD pressed status like you are showing for ctrl-C so users understand it is an option.

jacob314 (commented):
Much better! Added a comment about a related case to handle and pushed a commit simplifying this reducing the # of edge cases.

jacob314 (approved):
lgtm

jacob314 (approved):
lgtm

PR #8691, opened 2025-09-18
Improve compression splitpoint logic.

### LINKED ISSUE #8586: Compress fails in some cases.
### What happened?

GCLI shows: 

> IMPORTANT: This conversation approached the input token limit for gemini-2.5-pro. A compressed context will be sent for future messages (compressed from: 809158 to 2758 tokens).

And then immediately follows with another similar one.

Image

Repro steps were to issue a prompt like @packages/cli your favorite prompt here

it to compress only once.

### Client information

<details>

```console
> /about
# paste output here
```

</details>

### Login information

_No response_

_No response_

scidomino:
## TLDR

Ensures that we don't attempt to compress across a function call/response pair which would break later calls.

## Dive Deeper

Previously, we would sometimes try to compress everything while waiting for a functionResponse. This lead to bad requests when the function call was compressed away but the next user message had the functionResponse.

## Reviewer Test Plan

I repro'd this locally with:

```@packages/cli Tell me about this code```

then

```Is it good code?```

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes #8586

### DIFF
--- a/packages/core/src/core/client.ts
+++ b/packages/core/src/core/client.ts
@@ -26,7 +26,6 @@ import { reportError } from '../utils/errorReporting.js';
 import { GeminiChat } from './geminiChat.js';
 import { retryWithBackoff } from '../utils/retry.js';
 import { getErrorMessage } from '../utils/errors.js';
-import { isFunctionResponse } from '../utils/messageInspectors.js';
 import { tokenLimit } from './tokenLimits.js';
 import type { ChatRecordingService } from '../services/chatRecordingService.js';
 import type { ContentGenerator } from './contentGenerator.js';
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: packages/core/src/core/client.ts [hunk 1]
```diff
@@ -63,36 +62,51 @@ export function isThinkingDefault(model: string) {
...
-  for (let i = 0; i < contentLengths.length; i++) {
-    if (charactersSoFar >= targetCharacters) {
-      return i;
+  let lastSplitPoint = 0; // 0 is always valid (compress nothing)
+  let cumulativeCharCount = 0;
+  for (let i = 0; i < contents.length; i++) {
+    cumulativeCharCount += charCounts[i];
+    const content = contents[i];
+    if (
+      content.role === 'user' &&
```
SandyTao520, line 88: why don't we split after a models non-function call turn?

scidomino: A model's non-functioncall content will tell you if its safe to split at the following user content, but you won't find out if you're past the threshold till you process the following user content. I tried it a bunch of different ways and this is what worked.

SandyTao520 (approved):

PR #8702, opened 2025-09-18
metrics(extensions) - Add logging methods for extensions operations

shishu314:
## TLDR

Add logging methods for extension operations, these operations are not used yet

## Dive Deeper

Currently with extensions commands logging does not occur. This PR just creates the methods to reduce size of PRs and expedite review

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | x  | ❓  | ❓  |
| npx      | x  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Makes progress on #8085


packages/core/src/telemetry/constants.ts
```diff
@@ -12,6 +12,8 @@ export const EVENT_API_REQUEST = 'gemini_cli.api_request';
 export const EVENT_API_ERROR = 'gemini_cli.api_error';
 export const EVENT_API_RESPONSE = 'gemini_cli.api_response';
 export const EVENT_CLI_CONFIG = 'gemini_cli.config';
+export const EVENT_EXTENSION_ENABLE = 'gemini_cli.extension_enable';
```
chrstnb, line 15: Should we also add install here or hold off?

chrstnb (approved):
Looks great-- only question is if we should include install (and the bot comments look reasonable). Then, :shipit: ❗

packages/core/src/telemetry/constants.ts
shishu314: Done.

PR #8727, opened 2025-09-18
fix(shell): improve shell process termination and cleanup

galdawave:
## TLDR

This pull request significantly improves the reliability of shell process termination and resource cleanup in the `ShellExecutionService`. It introduces a more robust, multi-stage shutdown sequence on POSIX systems (macOS/Linux) and uses `taskkill` on Windows to ensure entire process trees are terminated correctly. It also prevents resource leaks by explicitly disposing of terminal resources and avoids errors from operating on already-dead processes.

## Dive Deeper

The previous implementation for terminating shell commands had a few weaknesses:
1.  It relied solely on `SIGINT` on POSIX systems, which is not always sufficient to terminate an entire process group, potentially leading to orphaned processes.
2.  On Windows, `ptyProcess.kill()` might not reliably terminate child processes spawned by the main command.
3.  There was a race condition where a terminal resize or other operation could be attempted on a process that had just exited, causing an error.
4.  The underlying `HeadlessTerminal` was not being explicitly disposed of, creating a potential memory/resource leak.

This PR addresses these issues by:
- **Implementing a graceful shutdown sequence:** On non-Windows platforms, it now sends `SIGHUP`, then `SIGTERM`, and finally `SIGKILL` to the process group, with brief delays. This is a standard pattern that gives processes a chance to clean up gracefully before being forcefully terminated.
- **Using `taskkill` on Windows:** It now shells out to `taskkill /pid <pid> /f /t` to forcefully kill the entire process tree, which is the most reliable method on Windows.
- **Adding `isProcessAlive` check:** A new utility function checks if a process exists before attempting operations like `resize`, preventing errors.
- **Ensuring resource cleanup:** The `headlessTerminal` is now explicitly disposed of in the `onExit` handler, ensuring its resources are released promptly.

## Reviewer Test Plan

To validate these changes, you can test the cancellation of long-running processes.

1.  Open the Gemini CLI.
2.  Run a command that starts a persistent process, for example: `sleep 100`
3.  While the command is running, cancel it (e.g., by closing the CLI or using the appropriate cancellation key if in an interactive session).
4.  Verify that the process is no longer running.
    -   On macOS/Linux: `ps aux | grep sleep` (should return nothing).
    -   On Windows: `tasklist | findstr "sleep.exe"` (should return nothing).
5.  You can also test the resize fix by running a short command and immediately trying to resize the (now non-existent) PTY. The CLI should not crash.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

-->


packages/core/src/services/shellExecutionService.ts
```diff
@@ -517,15 +523,36 @@ export class ShellExecutionService {
...
+                '/t',
+              ]);
             } else {
               try {
-                // Kill the entire process group
-                process.kill(-ptyProcess.pid, 'SIGINT');
+                // SIGHUP is sent to a process when its controlling terminal is closed.
+                // This might trigger a cleaner shutdown of the process tree.
+                process.kill(-ptyProcess.pid, 'SIGHUP');
+                await new Promise((res) => setTimeout(res, 100));
```
scidomino, line 538: Is this going to lead to delays for users?

galdawave: i don't think so since its 100ms

scidomino: This is fine since it only happens during a user abort.

scidomino (approved):

gsehgal (approved):

Issue #7777, opened 2025-09-05
Screen reader settings are broken


chrstnb:
Currently, setting 
ui: { accessibility: { screenReader: true }} does not work

hugomurillomtz:
@chrstnb this is ready for review if you may please: https://github.com/google-gemini/gemini-cli/pull/8730

PR #8730, opened 2025-09-18
fix(#7777): screen reader accessibility setting not working

hugomurillomtz:
## TLDR

It fixes the issue about the screenReader setting value not making the CLI launch in screen reader mode. It was being overlooked.

## Dive Deeper

The `screenReader` value was being set to `false` in the code (as if users were setting it to that) instead of `undefined`, causing the bug. As you can see in this condition, `undefined` is being checked, but the second condition would never be met because we were defaulting `screenReader` to false.

image

## Reviewer Test Plan

In your settings.json file, add this:

```
{
  "ui": {
    "accessibility": {
      "screenReader": true
    },
  },
...
}
```

You should see this when setting `screenReader` to `true`:
image

And you should see this instead when setting `screenReader` to `false`:
image

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | x  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

<!--

It fixes: https://github.com/google-gemini/gemini-cli/issues/7777


packages/cli/src/config/config.ts
```diff
@@ -229,7 +229,6 @@ export async function parseArguments(settings: Settings): Promise<CliArgs> {
         .option('screen-reader', {
           type: 'boolean',
           description: 'Enable screen reader mode for accessibility.',
-          default: false,
```
shrutip90, line 261: Does this mean the code is not handling default values properly? Instead of this, maybe we should be checking how these values are handled and handle default values properly.

shrutip90: Based on our offline discussion, this actually should be ok, since these are cmd line options and they take precedence over settings. Specifying a default value here will look similar to a user input. So we should not be doing that.

As suggested, please also verify other args that have default values like checkpointing, showMemoryUsage, etc and fix those if required.

shrutip90 (approved):

chrstnb:
Looks great-- good catch! Did you confirm that the arg will override it? e.g., npm start -- --screenReader or npm start -- --screenReader=false will always be respected above whatever the settings value is if set?

hugomurillomtz:
> Looks great-- good catch! Did you confirm that the arg will override it? e.g., npm start -- --screenReader or npm start -- --screenReader=false will always be respected above whatever the settings value is if set?

I did! The argv has precedence:

When doing `npm start -- --screenReader`

image

image

image

Also, when doing `npm start -- --screenReader=false`:

image

image

image

chrstnb (approved):

PR #8761, opened 2025-09-18
(feat): Add a ToDo tool to track ongoing task lists

### LINKED ISSUE #4580: Explore plan + execute pattern

Explore planning + execution based on todo checklist

### Why is this needed?

Better problem solving ability and ability to resolve user issues

### Additional context

_No response_

anj-s:
## TLDR

This PR introduces a new `write_todos_list` tool that allows the agent to create and manage a checklist of tasks for complex user requests. This helps the agent track its progress, organize its work, and provides the user with visibility into the agent's plan.

## Dive Deeper

The `write_todos_list` tool is a declarative tool that enables the agent to manage a list of tasks with the following statuses: `pending`, `in_progress`, `completed`, and `cancelled`. The agent is guided by an updated system prompt on when and how to use this tool, with a focus on using it for complex, multi-step tasks and avoiding it for simple requests.

The tool is enabled by a `useWriteTodos` flag in the configuration. The implementation includes the tool itself, along with comprehensive unit tests to ensure its functionality and validation logic are working correctly.

## Reviewer Test Plan

To test this feature, you can enable the `useWriteTodos` flag in your settings and give the agent a complex task. Here are a few examples:

1.  **Create a new feature:**
    *   Prompt: `add a new feature to the CLI that allows users to configure the output format of the response.`
    *   Expected behavior: The agent should create a todo list with steps like `add a new configuration option`, `implement the logic to format the output`, `add tests for the new feature`, etc.

2.  **Build a simple application:**
    *   Prompt: `create a simple web app that uses the Gemini API to answer questions.`
    *   Expected behavior: The agent should break down the task into smaller sub-tasks and create a todo list to track its progress.

3.  **Debug an issue:**
    *   Prompt: `The application is crashing when I try to upload a file. Can you help me debug and fix the issue?`
    *   Expected behavior: The agent should create a todo list to investigate the issue, such as `reproduce the crash`, `examine the logs`, `identify the root cause`, `implement a fix`, and `verify the fix`.

Fixes #4580 

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


packages/cli/src/config/config.ts
```diff
@@ -232,6 +233,10 @@ export async function parseArguments(settings: Settings): Promise<CliArgs> {
           description: 'Enable screen reader mode for accessibility.',
           default: false,
         })
+        .option('use-write-todos', {
```
NTaylorMullen: We'd opt away from command line arguments. Lets start with settings.json

scidomino (commented):
I'm confused. All the tool does is respond back with the todos it was given. It doesn't persist it anywhere and doesn't allow the model to view the list. Is this going to be part of a suite of tools?

anj-s:
> I'm confused. All the tool does is respond back with the todos it was given. It doesn't persist it anywhere and doesn't allow the model to view the list. Is this going to be part of a suite of tools?

This is a tool for the model to track what it needs to do. The todos are provided by the model and this tool is for it to communicate the updated status of that list. It doesn't need to persist it anywhere since it is in the history. Yes, this will be part of the core tools.

scidomino (approved):
It's weird that a noop tool would improve performance but I assume you have run evals and shown that this improves things.

packages/core/src/tools/write-todos.ts
```diff
@@ -0,0 +1,208 @@
...
+## Task state definitions
+
+- pending: Work has not begun on a given subtask.
+- in_progress: Marked just prior to beginning work on a given subtask. You should only have one subtask as in_progress at a time.
+- completed: Subtask was succesfully completed with no errors or issues. If the subtask required more more steps to complete, update the todo list. If more information is required from the user and you are in interactive mode, add that as a new subtask for request more information. All steps should be identified as completed only when they are completed.
+- cancelled: As you update the todo list, some tasks are not required anymore due to the dynamic nature of the task. In this case, mark the subtasks as cancelled so that the user still sees them in the list and knows that they are not required anymore.
+
+
+## Methodology for using this tool
+1. Use this too as soon as you receive a user request based on the complexity of the task.
```
owenofbrien: typo: "too" -> "tool"

packages/core/src/tools/write-todos.ts
```diff
@@ -0,0 +1,208 @@
...
+  BaseDeclarativeTool,
+  BaseToolInvocation,
+  Kind,
+  type ToolResult,
+} from './tools.js';
+
+// Insprired by langchain/deepagents.
+export const WRITE_TODOS_DESCRIPTION = `This tool can help you list out the current subtasks that are required to be completed for a given user request. The list of subtasks helps you keep track of the current task, organize complex queries and help ensure that you don't miss any steps. With this list, the user can also see the current progress you are making in executing a given task.
+
+Depending on the task complexity, you should first divide a given task in subtasks and then use this tool to list out the subtasks that are required to be completed for a given user request.
```
owenofbrien: "in" -> "into"

packages/core/src/tools/write-todos.ts
```diff
@@ -0,0 +1,208 @@
...
+Each of the subtasks should be clear and distinct. 
+
+Use this tool for complex queries that require multiple steps. If you find that the request is actually complex after you have started executing the user task, create a todo list and use it. If execution of the user task requires multiple steps, planning and generally is higher complexity than a simple Q&A, use this tool.
+
+DO NOT use this tool for simple tasks that can be completed in less than 2 steps. If the user query is simple and straightforward, do not use the tool. If you can respond with an answer in a single turn then this tool is not required.
+
+## Task state definitions
+
+- pending: Work has not begun on a given subtask.
+- in_progress: Marked just prior to beginning work on a given subtask. You should only have one subtask as in_progress at a time.
```
owenofbrien, line 28: "more more" -> "more"

packages/core/src/tools/write-todos.ts
```diff
@@ -0,0 +1,208 @@
...
+- completed: Subtask was succesfully completed with no errors or issues. If the subtask required more more steps to complete, update the todo list. If more information is required from the user and you are in interactive mode, add that as a new subtask for request more information. All steps should be identified as completed only when they are completed.
```
owenofbrien: "If the subtask required more steps to complete, update the todo list" -> it's a bit unclear to me, but does this basically mean "if while working on a task, you discover new distinct tasks that need to be done, add them to the todo list rather than handling them as part of the current task"? The current wording "the subtask required more steps to complete" makes it sound like these steps are required for the current task and thus should be modeled as sub-tasks of the current task, which implies a tree structure rather than a single list.

packages/core/src/tools/write-todos.ts
```diff
@@ -0,0 +1,208 @@
...
+- cancelled: As you update the todo list, some tasks are not required anymore due to the dynamic nature of the task. In this case, mark the subtasks as cancelled so that the user still sees them in the list and knows that they are not required anymore.
```
owenofbrien: "If you realize that a task is not actually necessary, mark it as cancelled so the user still sees it and knows it's no longer needed."

packages/core/src/tools/write-todos.ts
owenofbrien: Use this tool as soon as you realize a user request will require multiple discrete steps.

packages/core/src/tools/write-todos.ts
```diff
@@ -0,0 +1,208 @@
...
+
+
+## Methodology for using this tool
+1. Use this too as soon as you receive a user request based on the complexity of the task.
+2. Keep track of every subtask that you take my update this list
```
owenofbrien: "that you take my update this list" -> not sure what this is trying to say

packages/core/src/tools/write-todos.ts
```diff
@@ -0,0 +1,208 @@
...
+3. Mark a subtask as in_progress before you begin working on it. You should only have one subtask as in_progress at a time.
```
owenofbrien, line 36: "should reflect your progress and current plans, which may evolve as you acquire new information"

packages/core/src/tools/write-todos.ts
```diff
@@ -0,0 +1,208 @@
...
+4. Update the subtask list as you proceed in executing the task. The subtask list is not static and should reflect the current progress you are making.
+5. Mark a subtask as completed when you have completed it.
+6. Mark a subtask as cancelled if you update the todo list as you proceed in executing the task.
```
owenofbrien: "if you update the todo list as you proceed in executing the task" -> "if you decide it is no longer necessary"

packages/core/src/tools/write-todos.ts
```diff
@@ -0,0 +1,208 @@
...
+5. Implement asynchronous logic to handle the API call: show a loading indicator while the request is pending, retrieve the generated image (e.g., as a URL or base64 string) upon success, and display any errors.
+6. Display the returned "fancy logo" from the API response in the preview area component.
+7. Add functionality (e.g., a "Download" button) to allow the user to save the generated image file.
+8. Deploy the application to a web server or hosting platform.
+
+<reasoning>
+The agent used the todo list to break the task into distinct, manageable steps:
+1. Building an entire interactive web application from scratch is a highly complex, multi-stage process involving setup, UI development, logic integration, and deployment.
+2. The agent inferred the core functionality required for a "logo creator," such as UI controls for customization (Task 3) and an export feature (Task 7), which must be tracked as distinct goals.
+3. The agent rightly inferred the requirement of a API service model for interacting with the image model endpoint.
```
owenofbrien: nit: "an API"

owenofbrien:
> It's weird that a noop tool would improve performance but I assume you have run evals and shown that this improves things.

I think it's primarily a way to:
a) encourage the model to actually make a plan for complex tasks, and 
b) encourage the model to explicitly update the plan as tasks are completed or become obsolete

I wonder if just adding instructions for a) and b) to the system prompt could yield a similar performance impact. @anj-s wdyt?

packages/core/src/tools/write-todos.ts
anj-s: I think the point I am trying to make is that the model should not mark a sub task as a complete if it actually has dependent tasks and instead the list should be updated. It is always a single list and not a tree.

anj-s:
> It's weird that a noop tool would improve performance but I assume you have run evals and shown that this improves things.

Its not a noop tool as explained above. This helps the model create a list of items and track it. yes, this improves evals and is a known method for doing so.

anj-s:
> > It's weird that a noop tool would improve performance but I assume you have run evals and shown that this improves things.
> 
> I think it's primarily a way to: a) encourage the model to actually make a plan for complex tasks, and b) encourage the model to explicitly update the plan as tasks are completed or become obsolete
> 
> I wonder if just adding instructions for a) and b) to the system prompt could yield a similar performance impact. @anj-s wdyt?

We have this in the system prompt but its not something the model does consistently and does not involve the model updating the plan list at every turn. We ideally want the todo list to be the only plan list that the model is tracking

scidomino (approved):

agmsb (approved):

silviojr (approved):

PR #8788, opened 2025-09-19
Modifying new_prompt logging to handle real prompts only

razakhan-dev:
## TLDR

This PR modifies the logging of new_prompt event so that only the real prompt (not the client-side commands) are logged

## Dive Deeper

As mentioned in [Issue#759](https://github.com/google-gemini/maintainers-gemini-cli/issues/759), we seem to be missing the new_prompt event in many cases but it turns out that the logUserPrompt was being called for both the prompts and the client-side commands. This PR changes the logging point so that the logUserPrompt gets called only for the actual prompt rather than the client-side commands.

## Reviewer Test Plan

The user should be able to verify that the logUserPrompt is getting called only for the real prompts not the client side calls.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  |✅   | ❓  | ❓  |

## Linked issues / bugs

Related to google-gemini/maintainers-gemini-cli#759

Note: This PR does not fully resolve 759 as I was not able to replicate the issue


packages/cli/src/ui/hooks/useGeminiStream.ts
```diff
@@ -777,6 +768,19 @@ export const useGeminiStream = (
         }
 
         if (!options?.isContinuation) {
+          const promptText =
+            typeof queryToSend === 'string'
+              ? queryToSend
+              : JSON.stringify(queryToSend);
```
owenofbrien: For non-text (multimodal) queries, I don't think it makes sense to try to log them as json strings. We can work on logging multimodal prompts later, but for now let's only log the user prompt if it's a text prompt.

owenofbrien (changes requested):

packages/cli/src/ui/hooks/useGeminiStream.ts
razakhan-dev: Got it. Fixing in the next revision.

owenofbrien (approved):

PR #8797, opened 2025-09-19
Fix extensions enablement bugs

### LINKED ISSUE #8733: Document the extension enablement flow
We made some changes that support more robust overrides-- this should be reflected in the documentaiton.

### LINKED ISSUE #8735: Make user/workspace scopes case-insensitive in extensions enable/disable commands

### LINKED ISSUE #8752: -e/--extensions flag does not override disabledExtensions setting in settings.json
### What happened?

The `-e` or `--extensions` flag does not override the `disabledExtensions` setting in `settings.json`. If an extension has been disabled using `gemini extensions disable <extension_name>`, it cannot be temporarily re-enabled for a single session using the `-e <extension_name>` flag. This prevents a useful workflow where an extension can be disabled by default but enabled for a specific, one-off command.

**To Reproduce:**

1.  Install an extension, for example, `MyExtension`.
2.  Disable the extension using the command: `gemini extensions disable MyExtension`
3.  Verify that the extension is listed in the `disabledExtensions` array in your `~/.gemini/settings.json` file.
4.  Attempt to run Gemini CLI with the disabled extension enabled for a single session using the `-e` flag: `gemini -e MyExtension -p "hello"`
5.  Observe that the extension is not loaded and its tools are not available.

The `-e <extension_name>` flag should take precedence over the `disabledExtensions` setting in `settings.json`, allowing a disabled extension to be enabled for the current session only. This would enable a workflow where extensions can be installed but disabled by default, and then activated on-demand for specific sessions.

### Client information

<details>

```console
> /about
│ CLI Version        0.6.0-nightly-git.c4c81e3d          │
│ Git Commit         c4c81e3d                            │
│ OS                 linux                               │
│ IDE Client         IDE                                 │

```

</details>

### Login information

GEMINI_API_KEY

_No response_

chrstnb:
## TLDR

Fixing some bugs-- refactored the extensions enablement code to make it cleaner and added more thorough test cases. Specifically concerned with support for enabling extensions in specific workspaces when disabled at the user-level.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes #8733
Fixes #8735
Fixes #8752

### DIFF
--- a/packages/cli/src/commands/extensions/disable.ts
+++ b/packages/cli/src/commands/extensions/disable.ts
@@ -11,12 +11,16 @@ import { getErrorMessage } from '../../utils/errors.js';
 
 interface DisableArgs {
   name: string;
-  scope: SettingScope;
+  scope?: string;
 }
 
 export function handleDisable(args: DisableArgs) {
   try {
-    disableExtension(args.name, args.scope);
+    if (args.scope?.toLowerCase() === 'workspace') {
+      disableExtension(args.name, SettingScope.Workspace);
+    } else {
+      disableExtension(args.name, SettingScope.User);
+    }
     console.log(
       `Extension "${args.name}" successfully disabled for scope "${args.scope}".`,
     );
@@ -39,13 +43,28 @@ export const disableCommand: CommandModule = {
         describe: 'The scope to disable the extenison in.',
         type: 'string',
         default: SettingScope.User,
-        choices: [SettingScope.User, SettingScope.Workspace],
       })
-      .check((_argv) => true),
+      .check((argv) => {
+        if (
+          argv.scope &&
+          !Object.values(SettingScope)
+            .map((s) => s.toLowerCase())
+            .includes((argv.scope as string).toLowerCase())
+        ) {
+          throw new Error(
+            `Invalid scope: ${argv.scope}. Please use one of ${Object.values(
+              SettingScope,
+            )
+              .map((s) => s.toLowerCase())
+              .join(', ')}.`,
+          );
+        }
+        return true;
+      }),
   handler: (argv) => {
     handleDisable({
       name: argv['name'] as string,
-      scope: argv['scope'] as SettingScope,
+      scope: argv['scope'] as string,
     });
   },
 };
--- a/packages/cli/src/commands/extensions/enable.ts
+++ b/packages/cli/src/commands/extensions/enable.ts
@@ -11,13 +11,16 @@ import { SettingScope } from '../../config/settings.js';
 
 interface EnableArgs {
   name: string;
-  scope?: SettingScope;
+  scope?: string;
 }
 
 export function handleEnable(args: EnableArgs) {
   try {
-    const scope = args.scope ? args.scope : SettingScope.User;
-    enableExtension(args.name, scope);
+    if (args.scope?.toLowerCase() === 'workspace') {
+      enableExtension(args.name, SettingScope.Workspace);
+    } else {
+      enableExtension(args.name, SettingScope.User);
+    }
     if (args.scope) {
       console.log(
         `Extension "${args.name}" successfully enabled for scope "${args.scope}".`,
@@ -45,13 +48,28 @@ export const enableCommand: CommandModule = {
         describe:
           'The scope to enable the extenison in. If not set, will be enabled in all scopes.',
         type: 'string',
-        choices: [SettingScope.User, SettingScope.Workspace],
       })
-      .check((_argv) => true),
+      .check((argv) => {
+        if (
+          argv.scope &&
+          !Object.values(SettingScope)
+            .map((s) => s.toLowerCase())
+            .includes((argv.scope as string).toLowerCase())
+        ) {
+          throw new Error(
+            `Invalid scope: ${argv.scope}. Please use one of ${Object.values(
+              SettingScope,
+            )
+              .map((s) => s.toLowerCase())
+              .join(', ')}.`,
+          );
+        }
+        return true;
+      }),
   handler: (argv) => {
     handleEnable({
       name: argv['name'] as string,
-      scope: argv['scope'] as SettingScope,
+      scope: argv['scope'] as string,
     });
   },
 };
--- a/packages/cli/src/config/extensions/extensionEnablement.test.ts
+++ b/packages/cli/src/config/extensions/extensionEnablement.test.ts
@@ -58,8 +58,6 @@ describe('ExtensionEnablementManager', () => {
     it('should disable a path based on a disable override rule', () => {
       manager.enable('ext-test', true, '/');
       manager.disable('ext-test', true, '/home/user/projects/');
-      const config = manager.readConfig();
-      console.log(config['ext-test']);
       expect(manager.isEnabled('ext-test', '/home/user/projects/my-app')).toBe(
         false,
       );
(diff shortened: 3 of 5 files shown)


scidomino (approved):

packages/cli/src/commands/extensions/disable.ts
```diff
@@ -11,12 +11,16 @@ import { getErrorMessage } from '../../utils/errors.js';
...
-  scope: SettingScope;
+  scope?: string;
 }
 
 export function handleDisable(args: DisableArgs) {
   try {
-    disableExtension(args.name, args.scope);
+    if (args.scope?.toLowerCase() === 'workspace') {
+      disableExtension(args.name, SettingScope.Workspace);
+    } else {
```
jakemac53, line 21: nit: I would confirm scope is either null/empty or 'user' so we don't fall back on 'user' here if we get an unrecognized scope.

packages/cli/src/commands/extensions/disable.ts
```diff
@@ -39,13 +43,28 @@ export const disableCommand: CommandModule = {
         describe: 'The scope to disable the extenison in.',
         type: 'string',
         default: SettingScope.User,
-        choices: [SettingScope.User, SettingScope.Workspace],
       })
-      .check((_argv) => true),
+      .check((argv) => {
+        if (
+          argv.scope &&
+          !Object.values(SettingScope)
```
jakemac53, line 50: There are two additional scopes here that are not supported (system, system defaults), instead of `Object.values` I would hard code the supported ones.

I would probably also leave in the check I suggested above just to be extra careful and not rely at this validation at a distance.

packages/cli/src/commands/extensions/enable.ts
```diff
@@ -11,13 +11,16 @@ import { SettingScope } from '../../config/settings.js';
...
 }
 
 export function handleEnable(args: EnableArgs) {
   try {
-    const scope = args.scope ? args.scope : SettingScope.User;
-    enableExtension(args.name, scope);
+    if (args.scope?.toLowerCase() === 'workspace') {
+      enableExtension(args.name, SettingScope.Workspace);
+    } else {
+      enableExtension(args.name, SettingScope.User);
```
jakemac53, line 22: ditto, maybe add a shared function that parses the SettingScope into one of the two supported values and then use that in both of these places.

packages/cli/src/commands/extensions/enable.ts
```diff
@@ -45,13 +48,28 @@ export const enableCommand: CommandModule = {
         describe:
           'The scope to enable the extenison in. If not set, will be enabled in all scopes.',
         type: 'string',
-        choices: [SettingScope.User, SettingScope.Workspace],
       })
-      .check((_argv) => true),
+      .check((argv) => {
+        if (
+          argv.scope &&
```
jakemac53, line 54: similarly I would move this validator function into a shared function

packages/cli/src/config/extensions/extensionEnablement.test.ts
```diff
@@ -48,15 +48,15 @@ describe('ExtensionEnablementManager', () => {
...
+      manager.disable('ext-test', true, '/');
       manager.enable('ext-test', true, '/home/user/projects/');
       expect(manager.isEnabled('ext-test', '/home/user/projects/my-app')).toBe(
         true,
       );
     });
 
     it('should disable a path based on a disable override rule', () => {
-      manager.enable('ext-test', true, '*'); // Enable globally
+      manager.enable('ext-test', true, '/');
```
jakemac53, line 59: More of a meta comment but it's really hard to understand what this function call does from these positional arguments. I have to go to the actual definition of the function to understand the parameters.

In Dart land we have a style guide that says to use named arguments for boolean parameters specifically because they are so ambiguous at call sites https://dart.dev/effective-dart/design#avoid-positional-boolean-parameters. I think the same reasoning applies here.

Maybe that doesn't fit within the overall style for Gemini CLI though idk, up to you :)

packages/cli/src/config/extensions/extensionEnablement.test.ts
```diff
@@ -78,65 +78,253 @@ describe('ExtensionEnablementManager', () => {
...
+
+    it('should handle', () => {
+      manager.enable('ext-test', true, '/home/user/projects');
+      manager.disable('ext-test', false, '/home/user/projects/my-app');
+      expect(manager.isEnabled('ext-test', '/home/user/projects/my-app')).toBe(
+        false,
+      );
+      expect(
+        manager.isEnabled('ext-test', '/home/user/projects/something-else'),
+      ).toBe(true);
```
jakemac53, line 90: should handle what :)

packages/cli/src/config/extensions/extensionEnablement.test.ts
```diff
@@ -78,65 +78,253 @@ describe('ExtensionEnablementManager', () => {
...
+      expect(overrides).not.toContain('/path/to/dir/subdir1/');
+      expect(overrides).not.toContain(`!/path/to/dir/subdir2/*`);
+
+      // Unrelated rules should remain
+      expect(overrides).toContain('/path/to/another/dir/');
+    });
+
+    it('should remove child rules when disabling a parent with subdirs', () => {
+      // Pre-existing rules for children
+      manager.enable('ext-test', false, '/path/to/dir/subdir1');
```
jakemac53, line 180: I would add an expectation that this was present before enabling the parent dir to ensure something actually changed and this isn't just an erroneous expectation (ie: was never true originally)

packages/cli/src/config/extensions/extensionEnablement.test.ts
```diff
@@ -78,65 +78,253 @@ describe('ExtensionEnablementManager', () => {
...
+      manager.enable('ext-test', true, '/home/user/projects');
+      manager.disable('ext-test', false, '/home/user/projects/my-app');
+      expect(manager.isEnabled('ext-test', '/home/user/projects/my-app')).toBe(
+        false,
+      );
+      expect(
+        manager.isEnabled('ext-test', '/home/user/projects/something-else'),
+      ).toBe(true);
+    });
   });
```
jakemac53, line 93: I would add another expectation here:

`expect(manager.isEnabled('ext-test', '/home/user/projects123')).toBe(false)`;

packages/cli/src/config/extensions/extensionEnablement.ts
```diff
@@ -15,6 +15,80 @@ export interface AllExtensionsEnablementConfig {
...
+    const isDisable = inputRule.startsWith('!');
+    let baseRule = isDisable ? inputRule.substring(1) : inputRule;
+    baseRule = ensureLeadingAndTrailingSlash(baseRule);
+    return new Override(baseRule, isDisable, includeSubdirs);
+  }
+
+  static fromFileRule(fileRule: string): Override {
+    const isDisable = fileRule.startsWith('!');
+    let baseRule = isDisable ? fileRule.substring(1) : fileRule;
+    const includeSubdirs = baseRule.endsWith('*');
```
jakemac53, line 35: Related to one of my expectations above, I believe this will treat a rule like:

`/foo/bar*` and `/foo/bar/*` identically, when really they are different (or I would expect them to behave differently at least). Do we want to allow prefix matching of directories? Should we enforce the trailing slash to make it more clear that we don't support that?

packages/cli/src/config/extensions/extensionEnablement.ts
```diff
@@ -25,7 +99,7 @@ export interface AllExtensionsEnablementConfig {
 function globToRegex(glob: string): RegExp {
```
jakemac53, line 99: This function name is a bit misleading, the only thing supported is `*` right? Glob has a lot more stuff, I would just call it something like "pathMatchToRegex" or something.

PR #8837, opened 2025-09-19
fix(shell): improve shell output presentation and usability

galdawave:
## TLDR

This pull request improves the presentation and usability of the integrated shell. It introduces two main changes:
1.  **Trims trailing blank lines** from shell command output for a cleaner UI.
2.  **Adds a focus hint** (`ctrl+f to focus`) for interactive shell sessions to improve discoverability.

## Dive Deeper

The shell execution service now processes the output buffer to remove any trailing empty lines before rendering it in the UI. This prevents commands that output many newlines from creating large, empty blocks of space. For testing purposes, a `disableDynamicLineTrimming` flag has been added to bypass this feature and allow for exact output matching.

To improve the user experience with interactive commands (e.g., `node`, `python`), a hint to focus the shell now appears. If the user doesn't focus the interactive terminal within 5 seconds, the message `(ctrl+f to focus)` will be displayed, guiding them on how to provide input.

## Reviewer Test Plan

1.  **Test Output Trimming:**
    *   Run a command that produces trailing newlines, e.g., `gemini -m "echo -e 'hello\n\n\n'"
    *   **Expected:** The output should display "hello" without any extra blank lines below it.

2.  **Test Focus Hint:**
    *   Run an interactive command, e.g., `gemini -m "node"`.
    *   Wait for 5 seconds without interacting with the terminal.
    *   **Expected:** The hint `(ctrl+f to focus)` should appear next to the tool call.
    *   Press `ctrl+f`.
    *   **Expected:** The hint should change to `(Focused)`.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


packages/cli/src/ui/components/messages/ToolMessage.tsx
```diff
@@ -62,11 +62,44 @@ export const ToolMessage: React.FC<ToolMessageProps> = ({
...
+  const [showFocusHint, setShowFocusHint] = React.useState(false);
+
+  React.useEffect(() => {
+    if (resultDisplay) {
+      setLastUpdateTime(new Date());
+    }
+  }, [resultDisplay]);
+
+  React.useEffect(() => {
+    const interval = setInterval(() => {
```
jacob314: doing this with setInterval seems like a bit of overkill. why can't we use setTimeout instead of polling once every seconds for 5 seconds to see if the timeout has expired.

jacob314 (commented):
Additional polish it would be good to layer here or as a fast follow:
Correct the color of the border of the shell window to match.
Make the border color brighter when the shell has focus
Update the funny messages to say shell specific things when the shell is active.

packages/cli/src/ui/components/messages/ToolMessage.tsx
galdawave: The reason i used polling is because this checks that there has been a 5 second pause, after receiving data, so i don't think a normal setTimeout would work here

galdawave:
> Additional polish it would be good to layer here or as a fast follow: Correct the color of the border of the shell window to match. Make the border color brighter when the shell has focus Update the funny messages to say shell specific things when the shell is active.

Implemented everything except for the shell specific messages, currently we don't show messages when focused to reduce flicker, not sure if we want to show the messages when the user is foces.

packages/cli/src/ui/components/messages/ToolMessage.tsx
jacob314: discussed offline. approved once addressed.

galdawave: Changed.

jacob314 (approved):
lgtm

Issue #7565, opened 2025-09-01
Create alert notifications for nightly, preview, or release build failures


bdmorgan:
### What would you like to be added?

Minimally, I'd like to receive email notifications in the event of a failing run of the release action (https://github.com/google-gemini/gemini-cli/actions/workflows/release.yml), particularly in the event of nightly, preview, or release builds. 

Emails should go to the gemini-cli-eng-maintainers alias. A stretch goal here would be an automated notification either directly in the Gemini CLI Maintainers chat room and/or a new dedicated Gemini CLI Operations chat room.

### Why is this needed?

We instituted a new nightly build and release process recently (https://github.com/google-gemini/gemini-cli/blob/main/docs/releases.md). Builds have been failing due to failing tests but no one was checking (particularly on Friday and over the weekend). We need some sort of notification mechanism where we're all aware of issues, even if the assigned on-call is on top of it and handling it.

mattKorwel:
@bobcatfish bonus points if we can get this into a dedicated chat channel. @richieforeman mentioned he had done this before and might have context if you don't (i don't)

richieforeman:
Easiest path will be to use a Chat Webhook, which has some limitations, but is very quick to get off the ground.

PR #8919, opened 2025-09-19
fix: Create notification issues for manual release failures

bobcatfish:
## TLDR

Notify consistently (via issue and chat) for release failures.

## Dive Deeper

On the one hand we may think that if someone is running this manually, they'll notice it failing on their own; on the other hand, we create issues for patch release failures (which are also triggered somewhat manually), so at least this way we're consistent. Alternatively we could remove the issue creation from the patch workflow and only create the issue for periodic failures.

## Reviewer Test Plan

I wouldn't recommend it

## Testing Matrix

Did not test the change to the issue creation on release failure; in the worst case if it fails, the workflow has already failed. And it IS triggered manually, so we'll notice.

## Linked issues / bugs

Part of #7565

### DIFF
--- a/.github/workflows/release-manual.yml
+++ b/.github/workflows/release-manual.yml
@@ -80,3 +80,15 @@ jobs:
           github-token: '${{ secrets.GITHUB_TOKEN }}'
           dry-run: '${{ github.event.inputs.dry_run }}'
           previous-tag: '${{ steps.release_info.outputs.PREVIOUS_TAG }}'
+
+      - name: 'Create Issue on Failure'
+        if: '${{ failure() && github.event.inputs.dry_run == false }}'
+        env:
+          GITHUB_TOKEN: '${{ secrets.GITHUB_TOKEN }}'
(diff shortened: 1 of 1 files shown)


.github/workflows/release-manual.yml
```diff
@@ -80,3 +80,15 @@ jobs:
           github-token: '${{ secrets.GITHUB_TOKEN }}'
           dry-run: '${{ github.event.inputs.dry_run }}'
           previous-tag: '${{ steps.release_info.outputs.PREVIOUS_TAG }}'
+
+      - name: 'Create Issue on Failure'
```
shrutip90, line 84: Just trying to understand, where is the configuration for the chat notification?

bobcatfish: hey there! i think the PR you are looking for is https://github.com/google-gemini/gemini-cli/pull/8768, the docs are at https://github.com/google-gemini/gemini-cli/blob/main/docs/releases.md#modifying-chat-notifications

PR #8921, opened 2025-09-19
doc: Document manual push to dev channel

bobcatfish:
## TLDR

Doc manual push to dev channel

## Dive Deeper

We recently starting using shrinkwrap when building our npm package which resulted in issues we only caught post build and install; publishing to the dev channel is a way to test this kind of change without actually pushing to our users.

## Reviewer Test Plan

n/a

## Testing Matrix

n/a

## Linked issues / bugs

n/a


agmsb (commented):
I think we might use `latest` for general releases, instead of `stable`?

PR #8925, opened 2025-09-19
Refactor to defer initialization.

jacob314:
## TLDR

We were doing expensive initialization with side effects before launching the inner process.
Refactor so we don't create a config object until after startup.


packages/cli/src/gemini.tsx
```diff
@@ -347,7 +313,18 @@ export async function main() {
...
+          const partialConfig = await loadCliConfig(
+            settings.merged,
+            [],
+            sessionId,
+            argv,
+          );
+
+          await partialConfig.refreshAuth(
+            settings.merged.security.auth.selectedType,
+          );
```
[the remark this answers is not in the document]
jacob314: that is an interesting idea but is risky as it increases the chance of divergence between flows. added a TODO.

packages/cli/src/gemini.tsx
```diff
@@ -8,15 +8,14 @@ import React from 'react';
 import { render } from 'ink';
 import { AppContainer } from './ui/AppContainer.js';
 import { loadCliConfig, parseArguments } from './config/config.js';
+import * as cliConfig from './config/config.js';
```
abhipatel12, line 11: Can we just rename the local variable and use this as imported to avoid the namespace imports

packages/cli/src/config/config.ts
```diff
@@ -385,19 +385,25 @@ export async function loadHierarchicalGeminiMemory(
   );
 }
 
+export function isDebugMode(argv: CliArgs): boolean {
+  return (
+    argv.debug ||
+    [process.env['DEBUG'], process.env['DEBUG_MODE']].some(
+      (v) => v === 'true' || v === '1',
+    ) ||
+    false
```
abhipatel12: The `|| false` can be removed

packages/cli/src/config/config.ts
```diff
@@ -385,19 +385,25 @@ export async function loadHierarchicalGeminiMemory(
   );
 }
 
+export function isDebugMode(argv: CliArgs): boolean {
+  return (
+    argv.debug ||
+    [process.env['DEBUG'], process.env['DEBUG_MODE']].some(
+      (v) => v === 'true' || v === '1',
```
abhipatel12, line 403: Should we make this case insensitive in case someone uses `true` vs `TRUE` vs `True`

packages/cli/src/gemini.tsx
```diff
@@ -347,7 +269,21 @@ export async function main() {
           if (err) {
             throw new Error(err);
           }
-          await config.refreshAuth(settings.merged.security.auth.selectedType);
+          // We intentially omit the list of extensions here because extensions
+          // should not impact auth.
+          // TODO(jacobr): refactor loadCliConfig so there is a minimal version
```
abhipatel12, line 274: I am guilty of it too, but we should create issues and link those on TODOs instead of usernames, so that they don't get lost. We can leave this as is though if we expect to do this soon.

packages/cli/src/utils/relaunch.test.ts
```diff
@@ -0,0 +1,354 @@
...
+  beforeEach,
+  afterEach,
+  type MockInstance,
+} from 'vitest';
+import { EventEmitter } from 'node:events';
+import { RELAUNCH_EXIT_CODE } from './processUtils.js';
+import type { ChildProcess } from 'node:child_process';
+import { spawn } from 'node:child_process';
+
+// Mock child_process spawn
```
abhipatel12: nit: not a super useful comment

packages/cli/src/utils/relaunch.test.ts
```diff
@@ -0,0 +1,354 @@
...
+  let stdinResumeSpy: MockInstance;
+
+  // Store original values to restore later
+  const originalEnv = { ...process.env };
+  const originalExecArgv = [...process.execArgv];
+  const originalArgv = [...process.argv];
+  const originalExecPath = process.execPath;
+
+  beforeEach(() => {
+    // Clear all mocks first
```
abhipatel12: nit: the code here is self descriptive

abhipatel12 (approved):
A few nits but LGTM overall!

packages/cli/src/gemini.tsx
jacob314: style choice as I think the method names in cliConfig such as isDebugMode are better prefixed to avoid ambiguity.

PR #8655, opened 2025-09-17
Properly publish npm-shrinkwrap.json.

scidomino:
## TLDR

updates scripts to make sure shrinkwrap file makes it to the release.

## Dive Deeper

Testing revealed that `npm i -g @google/gemini-cli@preview` was not copying the shrinkwrap to the node_modules dir and we believe the reason is that we weren't copying it in these scripts as we should have.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| -------- | --- | --- | --- |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

PR #8298, opened 2025-09-11
Use shrinkwrap for package locking

scidomino:
## TLDR

Converts our package-lock.json to npm-shrinkwrap.json.

## Dive Deeper

unlike package-lock.json, the npm-shrinkwrap.json file is published to npm so we can be sure users running with `npm install` and `npx` are using the precise versions we suggest.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| -------- | --- | --- | --- |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes https://github.com/google-gemini/maintainers-gemini-cli/issues/764

PR #8926, opened 2025-09-19
Rollback shrinkwrap

scidomino:
## TLDR

Rolls back #8655, #8506, and #8298. Also updates a change in `e2e.yml` that referenced shrinkwrap.

## Dive Deeper

shrinkwrap broke the release so we're rolling it back.

## Reviewer Test Plan

I cut a release with this in it and verified that it fixes the release when combined with #8923 (which also broke the release).


jacob314 (approved):
lgtm

PR #8927, opened 2025-09-19
Update extension-releasing.md to have more info

jakemac53:
## TLDR

Adds information about the regular git release process as well as explaining some of the tradeoffs.

Re-structure a bit the custom release section as well.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

### DIFF
--- a/docs/extension-releasing.md
+++ b/docs/extension-releasing.md
@@ -1,10 +1,47 @@
 # Extension Releasing
 
-Gemini CLI extensions can be distributed as pre-built binaries through GitHub Releases. This provides a faster and more reliable installation experience for users, as it avoids the need to clone the repository and build the extension from source.
+There are two primary ways of releasing extensions to users:
 
-## Asset naming convention
+- [Git repository](#releasing-through-a-git-repository)
+- [Github Releases](#releasing-through-github-releases)
 
-To ensure Gemini CLI can automatically find the correct release asset for each platform, you should follow this naming convention. The CLI will search for assets in the following order:
+Git repository releases tend to be the simplest and most flexible approach, while GitHub releases can be more efficient on initial install as they are shipped as single archives instead of requiring a git clone which downloads each file individually. Github releases may also contain platform specific archives if you need to ship platform specific binary files.
+
+## Releasing through a git repository
+
+This is the most flexible and simple option. Users can depend on any ref from your git repo, such as a branch or tag, which allows you to manage multiple release channels. You also can point at any git repository, so your extension does not have to be hosted on github.
+
+For instance, you can maintain a `stable` branch, which users can install this way `gemini extensions install <your-repo-uri> --ref=stable`. Or, you could make this the default by treating your default branch as your stable release, and doing development in a different branch (for instance called `dev`). You can maintain as many branches or tags as you like, providing maximum flexibility for you and your users.
+
+Note that these `ref` arguments can be tags, branches, or even specific commits, which allows users to depend on a specific version of your extension. It is up to you how you want to manage your tags and branches.
+
+Whenever the commit at HEAD of the ref that a user depends on changes, they will be prompted to update the extension. Note that this also allows for easy rollbacks, the HEAD commit is always treated as the latest version regardless of the actual version in the `gemini-extension.json` file. This can also provide a more incremental update as opposed to a github release which always downloads and entirely new copy of the extension.
+
+### Example releasing flow using a git repo
+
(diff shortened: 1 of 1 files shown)


docs/extension-releasing.md
```diff
@@ -1,10 +1,47 @@
...
+Git repository releases tend to be the simplest and most flexible approach, while GitHub releases can be more efficient on initial install as they are shipped as single archives instead of requiring a git clone which downloads each file individually. Github releases may also contain platform specific archives if you need to ship platform specific binary files.
+
+## Releasing through a git repository
+
+This is the most flexible and simple option. Users can depend on any ref from your git repo, such as a branch or tag, which allows you to manage multiple release channels. You also can point at any git repository, so your extension does not have to be hosted on github.
+
+For instance, you can maintain a `stable` branch, which users can install this way `gemini extensions install <your-repo-uri> --ref=stable`. Or, you could make this the default by treating your default branch as your stable release, and doing development in a different branch (for instance called `dev`). You can maintain as many branches or tags as you like, providing maximum flexibility for you and your users.
```
jakemac53, line 14: @chrstnb should we consider using the `stable` ref by default if it exists? This would allow people to use their main branch for development while not risking  users accidentally depending on it (otherwise they have to do `--ref=stable` to be on the stable release channel).

In particular this can avoid issues around having to force push to your default branch if your history starts to diverge from your development branch due to cherry picks etc.

chrstnb: I think the idea is that the github repo based approach is super simple and accessible. I'm wary of introducing a specific way of doing things yet.

docs/extension-releasing.md
```diff
@@ -1,10 +1,49 @@
...
+There are two primary ways of releasing extensions to users:
 
-## Asset naming convention
+- [Git repository](#releasing-through-a-git-repository)
+- [Github Releases](#releasing-through-github-releases)
 
-To ensure Gemini CLI can automatically find the correct release asset for each platform, you should follow this naming convention. The CLI will search for assets in the following order:
+Git repository releases tend to be the simplest and most flexible approach, while GitHub releases can be more efficient on initial install as they are shipped as single archives instead of requiring a git clone which downloads each file individually. Github releases may also contain platform specific archives if you need to ship platform specific binary files.
+
+## Releasing through a git repository
```
chrstnb, line 10: The #1 most important piece of info here is that you can upload your extension from a github repo and it will immediately just work if the repo is public-- can you highlight that and then add a sub-section for the info about refs?

docs/extension-releasing.md
```diff
@@ -1,10 +1,47 @@
...
+
+For instance, you can maintain a `stable` branch, which users can install this way `gemini extensions install <your-repo-uri> --ref=stable`. Or, you could make this the default by treating your default branch as your stable release, and doing development in a different branch (for instance called `dev`). You can maintain as many branches or tags as you like, providing maximum flexibility for you and your users.
+
+Note that these `ref` arguments can be tags, branches, or even specific commits, which allows users to depend on a specific version of your extension. It is up to you how you want to manage your tags and branches.
+
+Whenever the commit at HEAD of the ref that a user depends on changes, they will be prompted to update the extension. Note that this also allows for easy rollbacks, the HEAD commit is always treated as the latest version regardless of the actual version in the `gemini-extension.json` file. This can also provide a more incremental update as opposed to a github release which always downloads and entirely new copy of the extension.
```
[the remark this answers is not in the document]
jakemac53: Interesting, @chrstnb we should probably fix that.

Should I change this statement for now or should we file an issue to bring the implementation in line with the statement?

docs/extension-releasing.md
chrstnb: I think we should frame the 'ref' support as 'advanced', if that makes sense

chrstnb (approved):
A couple comments but overall looks great

docs/extension-releasing.md
jakemac53: done

docs/extension-releasing.md
jakemac53: I am fine with that, I do worry this would be hard to change in the future though.

docs/extension-releasing.md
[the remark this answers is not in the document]
jakemac53: I am just removing it for now.

PR #8934, opened 2025-09-19
fix(security): Pin wrap-ansi to 9.0.2

SandyTao520:
## TLDR

Make sure users on the compromised 9.0.1 version auto update to 9.0.2


mattKorwel:
Thanks for pinning this. Can you confirm you've pulled this change locally and built and ran and see that all is good?

mattKorwel (approved):

SandyTao520:
image
Verified we're using 9.0.2 in released package.

PR #8940, opened 2025-09-19
feat(cli): Add /model command for interactive model selection

abhipatel12:
## TLDR

This pull request introduces a new `/model` slash command that allows users to interactively select the Gemini model they want to use for their session. Executing the command opens a full-screen dialog where the user can choose between Auto (default), Pro, Flash, and Flash-Lite models.

This feature is conditionally enabled and will only be available when the `useModelRouter` setting is set to `true` in the user's configuration.

UI:

Screenshot 2025-09-19 at 4 01 17 PM

Co-authored by @miguelsolorio 

## Dive Deeper

This PR implements the `/model` command and its corresponding UI through several new components and hooks:

-   **`/model` Command**: A new `SlashCommand` (`modelCommand.ts`) is defined. Its action returns a `dialog` type, triggering the `DialogManager` to open the new `ModelDialog`.
-   **`ModelDialog.tsx`**: A new dialog component that displays the available models with descriptions. It utilizes a new generic, reusable component for the selection UI.
-   **`DescriptiveRadioButtonSelect.tsx`**: A reusable and keyboard-navigable component for selecting an option from a list where each item has a title and a description. It supports navigation with `j`/`k`, arrow keys, and number selection.
-   **`useModelCommand.ts`**: A new hook that encapsulates the state management for the model dialog (`isModelDialogOpen`, `openModelDialog`, `closeModelDialog`, `handleModelSelect`). It interacts with the `Config` service to persist the user's selection.
-   **Conditional Loading**: The `BuiltinCommandLoader` has been updated to only load the `/model` command if `config.getUseModelRouter()` returns `true`, effectively feature-gating the command.

## Reviewer Test Plan

1.  **Enable the Feature**: In your `settings.json`, set `useModelRouter: true`:

    ```json
    {
      "experimental": {
        "useModelRouter": true
      }
    }
    ```
3.  **Launch the CLI**: Run `npm start`.
4.  **Open the Dialog**: Type `/model` and press Enter.
5.  **Verify UI**: Confirm that the "Select Model" dialog appears, showing the list of available models.
6.  **Test Keyboard Navigation**:
    -   Use the **up/down arrow keys** and **j/k keys** to navigate the list.
    -   Press a **number key** (e.g., `2`) to select the corresponding model.
    -   Press **Enter** to confirm a selection.
    -   Press **Escape** to close the dialog without making a change.
7.  **Confirm Selection**: After selecting a model with Enter or a number key, the dialog should close. You can re-open it to verify your selection is highlighted.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  |✅| ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt |✅| -   | -   |

## Linked issues / bugs

### DIFF
--- a/packages/cli/src/ui/components/ModelDialog.tsx
+++ b/packages/cli/src/ui/components/ModelDialog.tsx
@@ -1,0 +1,78 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import type React from 'react';
+import { useState } from 'react';
+import { Box, Text } from 'ink';
+import { useKeypress } from '../hooks/useKeypress.js';
+import { theme } from '../semantic-colors.js';
+import { DescriptiveRadioButtonSelect } from './shared/DescriptiveRadioButtonSelect.js';
+
+interface ModelDialogProps {
+  onClose: () => void;
+}
+
+const MODEL_OPTIONS = [
+  {
+    value: 'gemini-auto',
+    title: 'Auto (recommended)',
+    description: 'Let the system choose the best model for your task',
+  },
+  {
+    value: 'gemini-2.5-pro',
+    title: 'Gemini 2.5 Pro ($$$)',
+    description: 'For complex tasks that require deep reasoning and creativity',
+  },
+  {
+    value: 'gemini-2.5-flash',
+    title: 'Flash ($$)',
+    description: 'For tasks that need a balance of speed and reasoning',
+  },
+  {
+    value: 'gemini-2.5-lite',
+    title: 'Flash lite ($)',
+    description: 'For simple tasks that need to be done quickly',
+  },
+];
+
+export function ModelDialog({ onClose }: ModelDialogProps): React.JSX.Element {
+  const [selectedModel, setSelectedModel] = useState(MODEL_OPTIONS[0].value);
+
+  useKeypress(
+    (key) => {
+      if (key.name === 'escape') {
+        onClose();
+      }
+    },
+    { isActive: true },
+  );
+
+  return (
+    <Box
+      borderStyle="round"
+      borderColor={theme.border.default}
+      flexDirection="column"
+      padding={1}
+      width="100%"
+      height="100%"
+    >
+      <Text bold>Select Model</Text>
+      <Box marginTop={1}>
+        <DescriptiveRadioButtonSelect
+          items={MODEL_OPTIONS}
+          onSelect={setSelectedModel}
+          initialIndex={MODEL_OPTIONS.findIndex(
+            (option) => option.value === selectedModel,
+          )}
+          showNumbers={true}
+        />
+      </Box>
+      <Box marginTop={1}>
+        <Text color={theme.text.secondary}>(Press Esc to close)</Text>
+      </Box>
+    </Box>
+  );
+}
--- a/packages/cli/src/ui/components/ModelDialog.tsx
+++ b/packages/cli/src/ui/components/ModelDialog.tsx
@@ -5,41 +5,61 @@
  */
 
 import type React from 'react';
-import { useState } from 'react';
+import { useContext, useEffect, useState } from 'react';
 import { Box, Text } from 'ink';
+import {
+  DEFAULT_GEMINI_FLASH_LITE_MODEL,
+  DEFAULT_GEMINI_FLASH_MODEL,
+  DEFAULT_GEMINI_MODEL,
+  DEFAULT_GEMINI_MODEL_AUTO,
+} from '@google/gemini-cli-core';
 import { useKeypress } from '../hooks/useKeypress.js';
 import { theme } from '../semantic-colors.js';
 import { DescriptiveRadioButtonSelect } from './shared/DescriptiveRadioButtonSelect.js';
+import { ConfigContext } from '../contexts/ConfigContext.js';
 
 interface ModelDialogProps {
   onClose: () => void;
+  onSelect: (model: string) => void;
 }
 
 const MODEL_OPTIONS = [
   {
-    value: 'gemini-auto',
+    value: DEFAULT_GEMINI_MODEL_AUTO,
     title: 'Auto (recommended)',
     description: 'Let the system choose the best model for your task',
   },
   {
-    value: 'gemini-2.5-pro',
-    title: 'Gemini 2.5 Pro ($$$)',
+    value: DEFAULT_GEMINI_MODEL,
+    title: 'Pro',
     description: 'For complex tasks that require deep reasoning and creativity',
   },
   {
-    value: 'gemini-2.5-flash',
-    title: 'Flash ($$)',
+    value: DEFAULT_GEMINI_FLASH_MODEL,
+    title: 'Flash',
     description: 'For tasks that need a balance of speed and reasoning',
   },
   {
-    value: 'gemini-2.5-lite',
-    title: 'Flash lite ($)',
+    value: DEFAULT_GEMINI_FLASH_LITE_MODEL,
+    title: 'Flash-Lite',
     description: 'For simple tasks that need to be done quickly',
   },
 ];
 
-export function ModelDialog({ onClose }: ModelDialogProps): React.JSX.Element {
-  const [selectedModel, setSelectedModel] = useState(MODEL_OPTIONS[0].value);
+export function ModelDialog({
+  onClose,
+  onSelect,
+}: ModelDialogProps): React.JSX.Element {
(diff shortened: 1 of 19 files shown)


packages/cli/src/ui/components/shared/DescriptiveRadioButtonSelect.tsx
```diff
@@ -0,0 +1,160 @@
...
+          } else {
+            numberInputTimer.current = setTimeout(() => {
+              onSelect(targetItem.value);
+              setNumberInput('');
+            }, 350);
+          }
+        } else {
+          setNumberInput('');
+        }
+      }
```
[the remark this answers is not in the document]
abhipatel12: Done

miguelsolorio (approved):
Tested this locally and appears to work well. One minor mention, it might be nice to add in the history what model was used. This could done in a follow up, but I also wondered if this info should be visible by default for all users like:

Image

jacob314:
Hello! This is a great addition to the CLI. The new `/model` command will be very useful for users who want to experiment with different models. The code is well-structured, and the tests are thorough.

I have a few suggestions that could make the code even better, mostly related to our React best practices as outlined in `GEMINI.md`.

**1. Refactor `ModelDialog.tsx` to Avoid `useEffect` for State Syncing**

In `ModelDialog.tsx`, a `useEffect` is used to keep the `selectedModel` state in sync with the `ConfigContext`. Our `GEMINI.md` guide recommends avoiding this pattern to prevent unnecessary re-renders.

We can refactor the component to derive the initial model directly from the context during rendering. This makes the data flow clearer and more efficient.

*Suggestion:*
Remove the `useEffect` and the `selectedModel` state. Instead, calculate the `initialIndex` for `DescriptiveRadioButtonSelect` directly from the `config` context. You can keep a state variable like `highlightedModel` to track the user's navigation within the dialog, which would be updated by the `onHighlight` callback.

**2. Improve Numeric Input Logic in `DescriptiveRadioButtonSelect.tsx`**

The current implementation for selecting items with number keys has a small bug. When the number of items is a multiple of 10 (e.g., 10, 20), it's not possible to select items like '2' or '3' directly, as the component incorrectly waits for a second digit.

*Suggestion:*
The logic can be made more robust by checking if the entered number could potentially form a larger, valid number. If not, the selection should be made immediately. For example, if there are 15 items and the user presses '2', we know they can't be trying to type '20', so we can select item 2 right away.

**Prompt Engineering Reflection**

It looks like this change was likely generated from a prompt similar to this:

> "Create a new `/model` slash command that opens a dialog to let the user select a Gemini model. The dialog should be keyboard navigable and the selection should be saved to the configuration. This feature should only be enabled if the `useModelRouter` setting is true."

To get even better results next time, we could enhance the prompt by reminding the model of our specific coding standards:

> "When building the React components, ensure that you follow the guidelines in `GEMINI.md`, particularly avoiding setting state within `useEffect` for props that can be derived during rendering."

packages/cli/src/ui/components/shared/DescriptiveRadioButtonSelect.tsx
```diff
@@ -0,0 +1,164 @@
...
+const NUMBER_INPUT_TIMEOUT_MS = 1000;
+
+export interface DescriptiveRadioSelectItem<T> {
+  value: T;
+  title: string;
+  description: string;
+  disabled?: boolean;
+}
+
+export interface DescriptiveRadioButtonSelectProps<T> {
```
jacob314, line 19: please refactor this to use composition and a common base class with the existing RadioButton class. There is a lot of duplicate logic and we will make future enhancements that should be adopted by both. Generally this is just a radio button class with nicer rendering of the inputs.

packages/cli/src/ui/hooks/useModelCommand.ts
```diff
@@ -0,0 +1,42 @@
...
+
+  const openModelDialog = useCallback(() => {
+    setIsModelDialogOpen(true);
+  }, []);
+
+  const closeModelDialog = useCallback(() => {
+    setIsModelDialogOpen(false);
+  }, []);
+
+  const handleModelSelect = useCallback(
```
jacob314: any reason to not just move this directly within modelDialog? it already knows about the config. Might as well not separate stuff that can just be in one place. will simplify appContainer rather than tangling it with impl details. as a follow up we should regularize all the openDialog, closeDialog spew in AppContainer with generic functionality for that for all dialogs rather than creating so much boilerplate.

packages/cli/src/ui/components/ModelDialog.tsx
```diff
@@ -0,0 +1,108 @@
...
+  useKeypress(
+    (key) => {
+      if (key.name === 'escape') {
+        onClose();
+      }
+    },
+    { isActive: true },
+  );
+
+  const handleSelect = (model: string) => {
```
jacob314: nit: why create this helper. just pass onSelect directly?

packages/cli/src/ui/components/ModelDialog.tsx
```diff
@@ -0,0 +1,108 @@
...
+export function ModelDialog({
+  onClose,
+  onSelect,
+}: ModelDialogProps): React.JSX.Element {
+  const config = useContext(ConfigContext);
+  const [selectedModel, setSelectedModel] = useState(
+    config?.getModel() || DEFAULT_GEMINI_MODEL_AUTO,
+  );
+
+  useEffect(() => {
```
jacob314: I think this useEffect is a no-op. you are already passing in 
config?.getModel in the useState. 
Are you concerned about the case of the config changing while the dialog is open? Config should be fixed today so no need to check for that... probably have bigger problems if the config changes while the model dialog is open.

abhipatel12: Sounds good, removed this

packages/cli/src/ui/components/ModelDialog.tsx
abhipatel12: Left this b/c it combines two things (setting the model + closing dialog) and provides the useCallback optimization

packages/cli/src/ui/components/shared/DescriptiveRadioButtonSelect.tsx
abhipatel12: This has been refactored after we made the changes to extract the logic using the BaseSelectionList. Lmk what you think

packages/cli/src/ui/hooks/useModelCommand.ts
abhipatel12: Done

abhipatel12 (commented):
Made the changes! Let me know how this looks!

packages/cli/src/ui/components/ModelDialog.tsx
```diff
@@ -0,0 +1,108 @@
...
+  );
+
+  return (
+    <Box
+      borderStyle="round"
+      borderColor={theme.border.default}
+      flexDirection="column"
+      padding={1}
+      width="100%"
+      height="100%"
```
jacob314: remove width="100%"

packages/cli/src/ui/components/shared/DescriptiveRadioButtonSelect.test.tsx
```diff
@@ -0,0 +1,181 @@
...
+      renderItem = extractRenderItem();
+    });
+
+    it('should render title and description with correct colors', () => {
+      const item = ITEMS[0];
+
+      const result = renderItem(item, mockContext);
+
+      // Should be a Box with flexDirection="column"
+      expect(result?.props?.flexDirection).toBe('column');
```
jacob314: please strip down these tests to remove things that don't need to be tested.  generally all you need is a snapshot test.
yes the snapshot test won't test colors but for everything else tested here it will be superior and easy to read.
We will instead generally resolve the issue across gemini cli that the snapshot tests are black and white rather than requiring fragile and hard to verify prop level tests.

jacob314 (approved):
lgtm

packages/cli/src/ui/components/ModelDialog.tsx
abhipatel12: Done (removed height)

packages/cli/src/ui/components/shared/DescriptiveRadioButtonSelect.test.tsx
abhipatel12: Done, used snapshots

PR #9021, opened 2025-09-21
refactor(cli): Extract reusable BaseSelectionList component and modernize RadioButtonSelect tests

abhipatel12:
## TLDR

This PR extracts shared selection list logic from RadioButtonSelect into a new reusable BaseSelectionList component and completely revamps the test suite following modern testing best practices. The refactoring enables code reuse across different UI components while maintaining full backward compatibility and significantly improving test coverage quality.

Note: This was also a prefactor for the radio buttons needed for /model as well. 

## Dive Deeper

RadioButtonSelect had logic that could be shared with other components, so in this PR, we extracted the common parts into a new `BaseSelectionList` component. This handles all the keyboard navigation, scrolling, numbering, and theming that selection components need.

`RadioButtonSelect` is now just a thin wrapper that focuses on rendering radio button content. It still does everything it did before, but the underlying implementation is much cleaner and more reusable.

The old tests were mostly snapshots which weren't catching real issues. They now use proper mocking and actually test the component behavior. The new tests focus on what `RadioButtonSelect` specifically does (prop forwarding and content rendering) rather than re-testing the base functionality.

Also fixed a bug in SettingsDialog where the scope selector numbers weren't showing consistently - changed `showNumbers={focusSection === 'scope'}` to `showNumbers={true}` which fixed 11 failing tests.

## Reviewer Test Plan

1. **Pull and run the branch**
2. **Test RadioButtonSelect behavior:**
   ```bash
   # Verify RadioButtonSelect still works in SettingsDialog
   npm run start
   # Type: /settings
   # Use Tab to navigate to "Apply To" section
   # Verify numbers show: "1. User Settings", "2. Workspace Settings", "3. System Settings"
   # Test keyboard navigation (j/k, arrow keys, numeric selection)
   ```

3. **Test BaseSelectionList integration:**
   - Navigate through settings with arrow keys and j/k
   - Try numeric selection (press 1, 2, 3) in scope selector
   - Verify scroll arrows appear for long lists
   - Check disabled item behavior (grayed out, skipped during navigation)

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  |✅| ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt |✅| -   | -   |


packages/cli/src/ui/components/shared/BaseSelectionList.test.tsx
```diff
@@ -0,0 +1,486 @@
...
+    });
+
+    it('should show arrows with correct colors when scrolled to the end', () => {
+      const { lastFrame } = renderComponent(
+        { items: longList, maxItemsToShow: MAX_ITEMS, showScrollArrows: true },
+        9,
+      );
+
+      expect(lastFrame()).toMatchSnapshot();
+    });
```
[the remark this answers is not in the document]
abhipatel12: Done

packages/cli/src/ui/components/SettingsDialog.tsx
```diff
@@ -884,7 +884,7 @@ export function SettingsDialog({
               onSelect={handleScopeSelect}
               onHighlight={handleScopeHighlight}
               isFocused={focusSection === 'scope'}
-              showNumbers={focusSection === 'scope'}
+              showNumbers={true}
```
jacob314: revert this change

packages/cli/src/ui/hooks/useSelectionList.ts
```diff
@@ -0,0 +1,219 @@
...
+  direction: 'up' | 'down',
+  items: Array<SelectionListItem<T>>,
+): number => {
+  const len = items.length;
+  if (len === 0) return currentIndex;
+
+  let nextIndex = currentIndex;
+  const step = direction === 'down' ? 1 : -1;
+
+  // Iterate through the list (at most once) until an enabled item is found
```
jacob314: remove or edit this comment. it is obvious.. not sure how a for loop could loop through more than once.

packages/cli/src/ui/hooks/useSelectionList.ts
```diff
@@ -0,0 +1,219 @@
...
+  isFocused = true,
+  showNumbers = false,
+}: UseSelectionListOptions<T>): UseSelectionListResult {
+  const [activeIndex, setActiveIndex] = useState(initialIndex);
+  const numberInputRef = useRef('');
+  const numberInputTimer = useRef<NodeJS.Timeout | null>(null);
+
+  // Synchronize internal state when initialIndex or items change.
+  // Crucially, this ensures the active index is valid (enabled) if possible.
+  useEffect(() => {
```
jacob314, line 212: nit: ask Gemini CLI to refactor this class to use a reducer rather than useEffect for managing this state.  reason it matters is we can get multiple keyboard inputs in a same turn and this logic won't do the right thing. If gemini CLI is confused tell it to look at text-buffer.ts for a solid example of using a reducer to respond to keyboard input.

packages/cli/src/ui/hooks/useSelectionList.ts
```diff
@@ -0,0 +1,219 @@
...
+
+    // We only call setActiveIndex if the value is actually different to avoid unnecessary re-renders.
+    if (targetIndex !== activeIndex) {
+      setActiveIndex(targetIndex);
+    }
+
+    // We depend on the 'items' array reference itself to detect changes in the list structure or disabled status.
+    // We also depend on initialIndex to sync external changes.
+    // We explicitly exclude activeIndex from the dependency array to prevent loops, as this effect sets activeIndex.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
jacob314: never ignore this lint. correct the deps.

packages/cli/src/ui/hooks/useSelectionList.ts
```diff
@@ -0,0 +1,219 @@
...
+        const newIndex = findNextValidIndex(activeIndex, 'up', items);
+        if (newIndex !== activeIndex) {
+          setActiveIndex(newIndex);
+          onHighlight?.(items[newIndex]!.value);
+        }
+        return;
+      }
+
+      if (name === 'j' || name === 'down') {
+        const newIndex = findNextValidIndex(activeIndex, 'down', items);
```
jacob314: not a new bug but while you are modifying this, please also ask Gemini CLI to refactor this class to use a reducer. That will handle cases such as k being pressed multiple times in quick succession more robustly. all this logic will have subtle bugs in that case as set are calling setActiveIndex(index) rather than setActiveIndex((oldIndex) => your logic here).
Generally the reducer version is a bit more verbose is easier to get right. If Gemini CLI struggles ask it to look at text-buffer.ts as a good example of a robust reducer for this sort of case.

packages/cli/src/ui/components/shared/BaseSelectionList.tsx
```diff
@@ -0,0 +1,171 @@
...
+                isSelected,
+                titleColor,
+                numberColor,
+              })}
+            </Box>
+          </Box>
+        );
+      })}
+
+      {/* Use conditional coloring instead of conditional rendering */}
```
jacob314: remove this comment. it is generally obvious from the code.

jacob314 (commented):
Looks generally good. If Gemini CLI doesn't one shot switching to a reducer we can postpone that for follow up cleanup.

packages/cli/src/ui/components/SettingsDialog.tsx
abhipatel12: Done. Just to note, this behavior was broken before this PR which is why the tests had to be updated.

packages/cli/src/ui/components/shared/BaseSelectionList.tsx
abhipatel12: Done

packages/cli/src/ui/hooks/useSelectionList.ts
abhipatel12: Done

packages/cli/src/ui/hooks/useSelectionList.ts
abhipatel12: Done

packages/cli/src/ui/hooks/useSelectionList.ts
abhipatel12: Done

packages/cli/src/ui/hooks/useSelectionList.ts
abhipatel12: Done

abhipatel12 (commented):
Thanks for the review! Let me know what you think!

packages/cli/src/ui/hooks/useSelectionList.test.ts
```diff
@@ -269,6 +269,58 @@ describe('useSelectionList', () => {
...
+          meta: false,
+          shift: false,
+          paste: false,
+        };
+        handler(key);
+      };
+
+      // 1. Press Down. Should move 0 (A) -> 2 (C).
+      act(() => {
+        press('down');
```
jacob314, line 301: fwiw, adding multiple press calls in a single act as well as the case with one act call per press would be the best way to simulate the problematic case where keyboard input comes in faster than react re-renders.

jacob314 (approved):
lgtm

packages/cli/src/ui/hooks/useSelectionList.test.ts
abhipatel12: Done!

Issue #9011, opened 2025-09-20
--allowed-tools does not work in non-interactive mode


mistergarrison:
### What happened?

`--allowed-tools` only applies to interactive mode, but is ignored if the prompt is supplied on the command-line. The following command should work, but doesn't:

```sh
gemini --allowed-tools="ShellTool(wc)" \
    --prompt="use wc to tell me how many lines there are in /proc/meminfo"
```

### What did you expect to happen?

```sh
$ gemini --allowed-tools="ShellTool(wc)" \
    --prompt="use wc to tell me how many lines there are in /proc/meminfo"
There are 55 lines in /proc/meminfo.
$ 
```

### Client information

<details>
<summary>Client Information</summary>

Run `gemini` to enter the interactive CLI, then run the `/about` command.

```console
> /about
╭─────────────────────────────────────────────────────────────────────────────────────────────╮
│                                                                                             │
│ About Gemini CLI                                                                            │
│                                                                                             │
│ CLI Version                     0.7.0-nightly.20250918.2722473a-git.fa8cea17                │
│ Git Commit                      fa8cea17                                                    │
│ Model                           gemini-2.5-pro                                              │
│ Sandbox                         no sandbox                                                  │
│ OS                              linux                                                       │
│ Auth Method                     gemini-api-key                                              │
│                                                                                             │
╰─────────────────────────────────────────────────────────────────────────────────────────────╯

```

</details>

### Login information

n/a

### Anything else we need to know?

It would be good if this was also fixed for `tools.allowed` in settings.json as well.

mistergarrison:
Hey Prajwal,

I wasn't sure about the norms (and haven't submitted a PR in github before) - I had planned to do this: should I assign the issue to myself, or do you want to take it?

prajwxlk:
Hello @mistergarrison, I understand your situation. I would be perfectly fine if you wanted to proceed ahead with this issue although I don't know how to reassign the issue as well. As far, I understand using /assign works only when the issue hasn't been assigned to someone else. Also, the PR I've made fixes this issue. The reason I've made a draft PR is because I want to know how mapping of the ToolClass and actual tool_calls works in this codebase and if at all it has already been implemented. This will allow for the fix to work for all types of tools and not just only ShellTool().

PR #9047, opened 2025-09-22
fix(cli, core): enhance tool exclusion logic to respect explicitly al…

prajwxlk:
…lowed tools via --allowed-tools

## TLDR

This PR attempts to fix the issue where the --allowed-tools flag wasn't properly overriding tool exclusion logic. Previously, tools specified in --allowed-tools would still be blocked if they were in the exclude list, making the flag ineffective. The fix ensures that explicitly allowed tools bypass exclusion rules, providing users with the expected granular control over tool availability.

## Dive Deeper

The issue stemmed from the tool exclusion logic not considering the --allowed-tools parameter when determining which tools to exclude. The system had multiple layers of tool exclusion:

1. Settings-based exclusions (settings.tools.exclude)
2. Extension-based exclusions (extension.config.excludeTools)
3. Non-interactive mode exclusions (automatically excluding tools like ShellTool, EditTool, WriteFileTool in non-interactive mode)
4. Policy engine exclusions (handled separately in the policy engine)

The --allowed-tools flag was designed to bypass confirmation dialogs, but it wasn't integrated with the exclusion logic. This meant that even if a user explicitly allowed a tool via --allowed-tools, it would still be excluded by other mechanisms.

Impact of the changes made : 
- Users can now effectively use --allowed-tools to override any exclusion rules
- Non-interactive mode can now selectively allow specific tools while keeping others excluded
- The fix maintains backward compatibility and doesn't affect existing exclusion behavior for non-allowed tools

**Current Implementation Limitation :**
This PR includes a hardcoded mapping for `ShellTool` → `run_shell_command` which is specific to this use case. We need a systematic way to map tool class names to their actual tool call names across all tools. The current approach solves the immediate problem for the given use case only.

**Future Work Needed : **
We need to create a mapping system for all tool class names to tool call names if one does not exist already. I've tried looking for finding a logic in the existing codebase but couldn't. It would be very helpful if you could guide me as to how to approach this problem and implement a fix. Thank you!

## Reviewer Test Plan

Test Basic Functionality : 

1. `gemini --allowed-tools "ShellTool(ls)" --prompt "list files in current directory"`
2. `gemini --allowed-tools "ShellTool(git)" -p "what is the git status?"`

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅ | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

This PR makes progress on #9011

### DIFF
--- a/packages/cli/src/config/config.ts
+++ b/packages/cli/src/config/config.ts
@@ -532,6 +532,7 @@ export async function loadCliConfig(
     settings,
     activeExtensions,
     extraExcludes.length > 0 ? extraExcludes : undefined,
+    argv.allowedTools || settings.tools?.allowed,
   );
   const blockedMcpServers: Array<{ name: string; extensionName: string }> = [];
 
@@ -722,11 +723,32 @@ function mergeExcludeTools(
   settings: Settings,
   extensions: Extension[],
   extraExcludes?: string[] | undefined,
+  allowedTools?: string[] | undefined,
 ): string[] {
   const allExcludeTools = new Set([
     ...(settings.tools?.exclude || []),
     ...(extraExcludes || []),
   ]);
+
+  // NEW: Remove tools from exclude list that are explicitly allowed via --allowed-tools
+  if (allowedTools) {
+    for (const allowedTool of allowedTools) {
+      // Handle patterns like "ShellTool(git status)"
+      if (allowedTool.includes('(') && allowedTool.endsWith(')')) {
+        const toolName = allowedTool.substring(0, allowedTool.indexOf('('));
+
+        // Map common tool names to their actual names
+        const actualToolName =
+          toolName === 'ShellTool' ? 'run_shell_command' : toolName;
+        allExcludeTools.delete(actualToolName);
+        allExcludeTools.delete(toolName); // Also try the original name
+      } else {
+        // Handle exact tool names
+        allExcludeTools.delete(allowedTool);
+      }
+    }
+  }
+
   for (const extension of extensions) {
     for (const tool of extension.config.excludeTools || []) {
       allExcludeTools.add(tool);
--- a/packages/core/src/config/config.ts
+++ b/packages/core/src/config/config.ts
@@ -1013,7 +1013,17 @@ export class Config {
         (tool) => tool === toolName || tool === normalizedClassName,
       );
 
-      if (isExcluded) {
+      // NEW: Check if tool is explicitly allowed via --allowed-tools
+      const isExplicitlyAllowed = this.getAllowedTools()?.some(
+        (allowedTool) =>
+          // Handle both exact matches and pattern matches like "ShellTool(wc)"
+          allowedTool === toolName ||
+          allowedTool === normalizedClassName ||
+          allowedTool.startsWith(`${toolName}(`) ||
+          allowedTool.startsWith(`${normalizedClassName}(`),
+      );
+
+      if (isExcluded && !isExplicitlyAllowed) {
         isEnabled = false;
       }
 


mistergarrison:
Does it handle this without hanging?

```sh
  --allowed-tools="ShellTool(ls)" \
  --prompt="use wc to tell me how many lines there are in /proc/meminfo"`
```

prajwxlk:
I am currently on MacOS so couldn't test this exact command. Although, it should work. I've tested several other ShellTool() commands and they worked just fine.

mistergarrison:
On MacOS, then try:

```sh
  --allowed-tools="ShellTool(ls)" \
  --prompt="use wc to tell me how many lines there are in /etc/hosts"`
```

prajwxlk:
The one given by you hangs but I don't think that's the correct way of using the ShellTool(ls). I modified it to instead use ShellTool(wc) instead and it worked. Here's what the output looked like : 

```
npm start --   --allowed-tools="ShellTool(wc)" \  --prompt="use wc to tell me how many lines there are in /etc/hosts"

> @google/gemini-cli@0.7.0-nightly.20250918.2722473a start
> cross-env node scripts/start.js --allowed-tools=ShellTool(wc) --prompt=use wc to tell me how many lines there are in /etc/hosts

Checking build status...
Build is up-to-date.
Loaded cached credentials.
Loaded cached credentials.
/etc/hosts has 9 lines.
```

mistergarrison:
Right. The hang is introduced by your change. You need to make it not hang. :)

mistergarrison:
I almost have a version that doesn't introduce a regression like this. I'll send it out after I'm done testing.

mistergarrison:
Here's my version with the tests:

```sh
# Should print something like "There are 55 lines in /proc/meminfo"
npx github:mistergarrison/gemini-cli#fix/allowed-tools-non-interactive \
  --yolo \
  --prompt="use wc to tell me how many lines there are in /proc/meminfo"

# Should print something like "There are 55 lines in /proc/meminfo"
npx github:mistergarrison/gemini-cli#fix/allowed-tools-non-interactive \
  --allowed-tools="ShellTool(wc)" \
  --allowed-tools="ShellTool()" \
  --prompt="use wc to tell me how many lines there are in /proc/meminfo"

# Should print something like "There are 55 lines in /proc/meminfo"
npx github:mistergarrison/gemini-cli#fix/allowed-tools-non-interactive \
  --allowed-tools="run_shell_command(wc)" \
  --allowed-tools="run_shell_command()" \
  --prompt="use wc to tell me how many lines there are in /proc/meminfo"

# Should print something like "There are 55 lines in /proc/meminfo"
npx github:mistergarrison/gemini-cli#fix/allowed-tools-non-interactive \
  --allowed-tools="ShellTool(wc)" \
  --allowed-tools="ShellTool" \
  --prompt="use wc to tell me how many lines there are in /proc/meminfo"

# Should print something like "There are 55 lines in /proc/meminfo"
npx github:mistergarrison/gemini-cli#fix/allowed-tools-non-interactive \
  --allowed-tools="run_shell_command(wc)" \
  --allowed-tools="ShellTool" \
  --prompt="use wc to tell me how many lines there are in /proc/meminfo"

# Should print something like "There are 55 lines in /proc/meminfo"
npx github:mistergarrison/gemini-cli#fix/allowed-tools-non-interactive \
  --allowed-tools="ShellTool(ls)" \
  --allowed-tools="ShellTool" \
  --prompt="use wc to tell me how many lines there are in /proc/meminfo"

# Should fail
npx github:mistergarrison/gemini-cli#fix/allowed-tools-non-interactive \
  --allowed-tools="ShellTool(ls)" \
  --allowed-tools="ShellTool()" \
  --prompt="use wc to tell me how many lines there are in /proc/meminfo"

# Should fail
npx github:mistergarrison/gemini-cli#fix/allowed-tools-non-interactive \
  --allowed-tools="run_shell_command(ls)" \
  --allowed-tools="run_shell_command()" \
  --prompt="use wc to tell me how many lines there are in /proc/meminfo"

# Should fail
npx github:mistergarrison/gemini-cli#fix/allowed-tools-non-interactive \
  --prompt="use wc to tell me how many lines there are in /proc/meminfo"

```

mistergarrison:
I think [this ](https://github.com/google-gemini/gemini-cli/pull/9114/) pull request should work?

prajwxlk:
I understand, I'll have to see why my changes are breaking some functionality then.

If your PR is working with all test cases, I'll be fine closing this PR. I don't know how to setup my codebase to test your PRs (I'm new to OSS too). 

Also, just make sure that your PR works for all the different allowed-tools and not just ShellTool(). I was looking into how to handle this with this draft PR.

Anyways, Good Work @mistergarrison!

PR #9049, opened 2025-09-22
Fix private release downloading

chrstnb:
## TLDR

Currently, if you are attempting to install from a private repo with a release, we fall back to cloning from your default branch if you have an API key set. 

This resolves that issue and also makes the errors more robust.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


packages/cli/src/config/extensions/github.ts
```diff
@@ -61,7 +61,19 @@ export async function cloneFromGit(
       );
     }
 
-    const refToFetch = installMetadata.ref || 'HEAD';
+    let refToFetch = installMetadata.ref;
```
jakemac53, line 64: Note that the way this works today, if we call `cloneFromGit` when not finding any release assets, we will already be pass in an explicit ref which is the tag name from the latest release.

Are you sure this code is reachable? I am not really quite understanding the exact issue here.

packages/cli/src/config/extensions/github.ts
```diff
@@ -326,60 +336,135 @@ export function findReleaseAsset(assets: Asset[]): Asset | undefined {
...
     'User-Agent': 'gemini-cli',
+    Accept: 'application/vnd.github+json',
   };
   const token = getGitHubToken();
+
   if (token) {
-    headers.Authorization = `token ${token}`;
+    headers.Authorization = `Bearer ${token}`;
+  } else {
+    console.log('GITHUB_TOKEN is missing.');
```
jakemac53, line 358: Not everybody needs a GITHUB_TOKEN, can we conditionally log this if the request fails?

PR #9065, opened 2025-09-22
fix: enhance path handling in handleAtCommand to support relative paths

### LINKED ISSUE #9016: Handle pasting absolute file paths into Gemini CLI correctly
### What happened?

Currently when pasting absolute paths into Gemini CLI they have an `@` automatically added but the read files tool fails to handle the absolute path even though the path is under the project.
We should make ReadFiles handle this case correctly.
We should either swap the path to be a relative path or better just make ReadFiles work correctly for this.
This may also be breaking dragging and dropping images.

Expect the file to be read correctly if `@fileName` shows up in my input prompt

### Client information

Not client specific.

### Login information

_No response_

_No response_

lifefloating:
## TLDR

Enhanced the `handleAtCommand` function to improve path handling by converting absolute paths to relative paths in display output and ensuring consistent path resolution. This makes the `@` command output cleaner and more user-friendly by showing relative paths instead of long absolute paths.

## Dive Deeper

The previous implementation of `handleAtCommand` would display absolute paths in the processed query output, which created verbose and less readable content.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ✅  | ❓  | ❓  |

## Linked issues / bugs

Fixes #9016


lifefloating:
@jacob314

lifefloating:
@galz10

galdawave:
> @galz10 

Hi @lifefloating, I'll review this first thing Monday. Thank you for tagging me.

lifefloating:
> > @galz10
> 
> Hi @lifefloating, I'll review this first thing Monday. Thank you for tagging me.

Sorry, I forgot that it's your weekend today. No problem and no rush. I think the settings issue at /settings also needs to be checked.  #10875

packages/cli/src/ui/hooks/atCommandProcessor.ts
```diff
@@ -430,10 +449,27 @@ export async function handleAtCommand({
...
+            const filePathSpecInContent = match[1];
             const fileActualContent = match[2].trim();
+
+            let displayPath = absoluteToRelativePathMap.get(
+              filePathSpecInContent,
+            );
+
+            // Fallback: if no mapping found, try to convert absolute path to relative
+            if (!displayPath) {
+              for (const dir of config.getWorkspaceContext().getDirectories()) {
```
joshualitt, line 461: Shall we make a `toWorkspaceRelative` or `toRelative` helper to encapsulate this logic?

packages/cli/src/ui/hooks/atCommandProcessor.ts
```diff
@@ -430,10 +449,27 @@ export async function handleAtCommand({
...
           if (match) {
-            const filePathSpecInContent = match[1]; // This is a resolved pathSpec
+            const filePathSpecInContent = match[1];
             const fileActualContent = match[2].trim();
+
+            let displayPath = absoluteToRelativePathMap.get(
+              filePathSpecInContent,
+            );
+
+            // Fallback: if no mapping found, try to convert absolute path to relative
```
joshualitt, line 459: Expand this comment with a bit more details as to when / why this case occurs?

joshualitt (approved):

joshualitt:
Looks great, thanks @lifefloating !

scidomino:
This PR broke a test on windows but those are non blocking. I will create a PR to fix it.

PR #9082, opened 2025-09-22
feat(hooks): Hook Telemetry Infrastructure

### LINKED ISSUE #9081: Hook Telemetry Infrastructure
Implement comprehensive telemetry and observability for hook execution, providing detailed logging and metrics for all hook invocations.

Edilmo:
## TLDR

Adds complete telemetry support for hook execution with structured logging and metrics collection for observability and debugging.

## Dive Deeper

This implements the observability requirement for Hooks. The HookCallEvent class captures all relevant hook execution data (event name, hook type, input/output, duration, success/failure). The logHookCall function provides structured logging, while recordHookCallMetrics tracks performance metrics. This telemetry is essential for debugging hook issues and monitoring system performance.

## Reviewer Test Plan

- Run telemetry tests to verify event structure
- Test hook execution generates proper telemetry events
- Verify metrics collection works for both successful and failed hook executions
- Check telemetry includes all required fields (duration, success, error details)

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ✅  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Closes #9081


packages/core/src/telemetry/types.ts
```diff
@@ -669,3 +669,47 @@ export class ExtensionEnableEvent implements BaseTelemetryEvent {
...
+  stdout?: string;
+  stderr?: string;
+  duration_ms: number;
+  success: boolean;
+  error?: string;
+
+  constructor(
+    hookEventName: string,
+    hookType: 'command' | 'plugin',
+    hookName: string,
```
cornmander, line 1761: We need to be careful not to record user generated input.

Edilmo: I have made changes to address the PII concerns by respecting the telemetry.logPrompts setting, and also using a sanitize method over the name (which is generated from the hook command).

Please re-open if you consider this is not enough.

scidomino (approved):

packages/core/src/telemetry/types.ts
```diff
@@ -1737,3 +1738,86 @@ export class WebFetchFallbackAttemptEvent implements BaseTelemetryEvent {
     return `Web fetch fallback attempt. Reason: ${this.reason}`;
   }
 }
+
+export const EVENT_HOOK_CALL = 'gemini_cli.hook_call';
+export class HookCallEvent implements BaseTelemetryEvent {
```
owenofbrien, line 1743: We should log the non-PII fields (hook type, hook name, exit code, duration, success) to Clearcut for centralized telemetry.

packages/core/src/telemetry/sanitize.ts
```diff
@@ -0,0 +1,52 @@
...
+ * - "/path/to/.gemini/hooks/check-secrets.sh --api-key=abc123" -> "check-secrets.sh"
+ * - "python /home/user/script.py --token=xyz" -> "python"
+ * - "node index.js" -> "node"
+ * - "C:\\Windows\\System32\\cmd.exe /c secret.bat" -> "cmd.exe"
+ * - "" or "   " -> "unknown-command"
+ *
+ * @param hookName Full command string.
+ * @returns Sanitized command name.
+ */
+export function sanitizeHookName(hookName: string): string {
```
owenofbrien, line 26: I don't think PII sanitization is necessary for OTEL telemetry, as OTEL is intended for the end-user's own consumption. OTOH, we do need to ensure that we don't log PII to Clearcut, but heuristic-based sanitization is insufficient. For Clearcut, I recommend logging only the fields that are guaranteed not to contain PII, and for OTEL you can probably scrap the sanitization and just log everything.

packages/core/src/telemetry/loggers.ts
```diff
@@ -670,3 +672,22 @@ export function logLlmLoopCheck(
   };
   logger.emit(logRecord);
 }
+
+export function logHookCall(config: Config, event: HookCallEvent): void {
```
owenofbrien, line 676: This function should also log to Clearcut. You'll need to add a hook logging function to clearcut-logger.ts and call it here. You'll need to add a new Clearcut event name and define new metadata fields in telemetry/clearcut-logger/event-metadata-key.ts. You'll then need to add those same metadata keys to https://source.corp.google.com/piper///depot/google3/logs/proto/cloud/concord/concord_event_metadata.proto in g3 to keep the metadata definitions in sync.

owenofbrien (commented):
We need to log hook events to Clearcut as well as OTEL. See comment in loggers.ts and feel free to ping me if you have any questions.

PR #9092, opened 2025-09-22
feat(hooks): Hook Execution Engine

### LINKED ISSUE #9091: Hook Execution Engine
Implement robust command hook execution with timeout handling, error management, and proper process lifecycle management.

Edilmo:
## TLDR

Creates HookRunner that executes command hooks via child processes, with comprehensive error handling and timeout management.

## Dive Deeper

This is the execution engine that runs the actual hooks. For command hooks, it spawns child processes, streams JSON input to stdin, collects stdout/stderr, and handles timeouts with proper cleanup. 

## Reviewer Test Plan

- Test command hook execution with various exit codes (0, 2, other)
- Verify timeout handling kills processes correctly
- Test JSON output parsing and plain text conversion
- Validate error handling doesn't throw exceptions

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ✅  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Closes #9091

### DIFF
--- a/packages/core/src/hooks/hookRunner.test.ts
+++ b/packages/core/src/hooks/hookRunner.test.ts
@@ -1,0 +1,295 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+import { spawn, type ChildProcessWithoutNullStreams } from 'node:child_process';
+import { HookRunner } from './hookRunner.js';
+import { HookEventName, HookType } from './types.js';
+import type { HookConfig } from './types.js';
+import type { HookInput } from './types.js';
+import type { Readable, Writable } from 'node:stream';
+
+// Mock type for the child_process spawn
+type MockChildProcessWithoutNullStreams = ChildProcessWithoutNullStreams & {
+  mockStdoutOn: ReturnType<typeof vi.fn>;
+  mockStderrOn: ReturnType<typeof vi.fn>;
+  mockProcessOn: ReturnType<typeof vi.fn>;
+};
+
+// Mock child_process with importOriginal for partial mocking
+vi.mock('node:child_process', async (importOriginal) => {
+  const actual = (await importOriginal()) as object;
+  return {
+    ...actual,
+    spawn: vi.fn(),
+  };
+});
+
+// Mock console methods
+const mockConsole = {
+  log: vi.fn(),
+  warn: vi.fn(),
+  error: vi.fn(),
+  debug: vi.fn(),
+};
+
+vi.stubGlobal('console', mockConsole);
+
+describe('HookRunner', () => {
+  let hookRunner: HookRunner;
+  let mockSpawn: MockChildProcessWithoutNullStreams;
+
+  const mockInput: HookInput = {
+    session_id: 'test-session',
+    transcript_path: '/path/to/transcript',
+    cwd: '/test/project',
+    hook_event_name: 'BeforeTool',
+    timestamp: '2025-01-01T00:00:00.000Z',
+  };
+
+  beforeEach(() => {
+    vi.resetAllMocks();
+
+    hookRunner = new HookRunner();
+
+    // Mock spawn with accessible mock functions
+    const mockStdoutOn = vi.fn();
+    const mockStderrOn = vi.fn();
+    const mockProcessOn = vi.fn();
+
+    mockSpawn = {
+      stdin: {
+        write: vi.fn(),
+        end: vi.fn(),
+      } as unknown as Writable,
+      stdout: {
+        on: mockStdoutOn,
+      } as unknown as Readable,
+      stderr: {
+        on: mockStderrOn,
+      } as unknown as Readable,
+      on: mockProcessOn,
+      kill: vi.fn(),
+      killed: false,
+      mockStdoutOn,
+      mockStderrOn,
+      mockProcessOn,
+    } as unknown as MockChildProcessWithoutNullStreams;
+
+    vi.mocked(spawn).mockReturnValue(mockSpawn);
+  });
+
+  afterEach(() => {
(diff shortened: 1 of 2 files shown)


packages/core/src/hooks/hookRunner.ts
```diff
@@ -0,0 +1,515 @@
...
+      };
+
+      // Parse command and arguments
+      const [cmd, ...args] = this.parseCommand(command);
+
+      const child = spawn(cmd, args, {
+        env,
+        cwd: input.cwd,
+        stdio: ['pipe', 'pipe', 'pipe'],
+      });
```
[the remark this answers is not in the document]
Edilmo: DOne

packages/core/src/hooks/hookRunner.ts
```diff
@@ -0,0 +1,350 @@
...
+    try {
+      return await this.executeCommandHook(
+        hookConfig,
+        eventName,
+        input,
+        startTime,
+      );
+    } catch (error) {
+      const duration = Date.now() - startTime;
+      const errorMessage = `Hook execution failed: ${error}`;
```
abhipatel12: Just curious, do you think this sort of message needs to be mentioned in the user's UI? Or is it something we want to keep only in the debug drawer?

Additionally, do you think we should make this error message a bit more informative (i.e. event name, hook source, etc.)?

packages/core/src/hooks/hookRunner.ts
```diff
@@ -0,0 +1,350 @@
...
+    eventName: HookEventName,
+    input: HookInput,
+    startTime: number,
+  ): Promise<HookExecutionResult> {
+    const timeout = hookConfig.timeout ?? DEFAULT_HOOK_TIMEOUT;
+
+    return new Promise((resolve) => {
+      if (!hookConfig.command) {
+        const errorMessage = 'Command hook missing command';
+        debugLogger.warn(
```
abhipatel12, line 188: Let's just confirm that all of these should only be found in the debug drawer or if the user needs this sort of message in the UI history.

packages/core/src/hooks/hookRunner.ts
```diff
@@ -0,0 +1,350 @@
...
+    text: string,
+    exitCode: number,
+  ): HookOutput {
+    if (exitCode === 0) {
+      // Success - treat as system message or additional context
+      return {
+        decision: 'allow',
+        systemMessage: text,
+      };
+    } else if (exitCode === 2) {
```
abhipatel12: Can we extract these exit codes so they don't appear as magic numbers?

packages/core/src/hooks/hookRunner.test.ts
```diff
@@ -0,0 +1,295 @@
...
+              GEMINI_PROJECT_DIR: '/test/project',
+              CLAUDE_PROJECT_DIR: '/test/project',
+            }),
+          }),
+        );
+      });
+    });
+  });
+
+  describe('executeHooksParallel', () => {
```
abhipatel12, line 241: Should we also have a section to test `executeHooksSequential`?

packages/core/src/hooks/hookRunner.test.ts
abhipatel12, line 1: I think the mocked spawn value is fine, but should we consider using a real (but simple) shell script to make sure this works properly across all of our supported platforms? 

Or is there an integration test as well that is being added later?

abhipatel12, line 1: Do we need any tests of invalid json?

abhipatel12 (commented):
Looks good! Had a few questions. 

Also, can we update the description since we're not implementing plugins currently

packages/core/src/hooks/hookRunner.test.ts
Edilmo: Correct, I had many integration tests we [real shell scripts in the last PR](https://github.com/google-gemini/gemini-cli/pull/9112/commits/296acba55ab649b490f61e62ffe0b9ce17847483)

If a I were to add something here, how should I do it without integration tests?

packages/core/src/hooks/hookRunner.ts
Edilmo: The way I see it is that we should have some sort of consistency across extension mechanisms. 
That being said, I would love to align to a better error handling for extensibility points, but for now, I thought this was the way handled. Check [extension-manager.ts:502-509](https://github.com/google-gemini/gemini-cli/blob/main/packages/cli/src/config/extension-manager.ts#L502)
When an extension fails to load, it's caught and logged:
} catch (e) {
  debugLogger.error(
    `Warning: Skipping extension in ${effectiveExtensionPath}: ${getErrorMessage(e)}`,
  );
  return null;
}
The extension is silently skipped (returns null) and only logged via debugLogger.error().

But yes, I would love something better but perhaps we should leave it for follow up PRs focus on UX (those I have not code them).

packages/core/src/hooks/hookRunner.ts
Edilmo: done and regarding the first question, I answered in the other message where you asked the same.

packages/core/src/hooks/hookRunner.ts
abhipatel12: Gotcha, we are current refactoring a lot of that logging to use this [emitter](https://github.com/google-gemini/gemini-cli/blob/main/packages/core/src/utils/events.ts). If you think we need to bubble this to the UI, we should just use this emitter to do so. Otherwise, let's keep it as is!

packages/core/src/hooks/hookRunner.test.ts
abhipatel12: nope that's great, we can just use those then. Thanks!

abhipatel12 (approved):

PR #9095, opened 2025-09-22
feat(hooks): Hook Result Aggregation

Edilmo:
## TLDR

Creates HookAggregator that merges multiple hook outputs using event-specific rules (OR logic for decisions, concatenation for messages, tool config merging).

## Dive Deeper

This component handles the complex logic of merging outputs when multiple hooks execute for the same event. Different events have different merging strategies: BeforeTool/AfterTool use OR logic for decisions and concatenate messages; BeforeModel/AfterModel use field replacement; BeforeToolSelection has complex tool config merging based on mode values. The aggregator ensures the final result represents the combined intent of all executed hooks.

## Reviewer Test Plan

- Test decision merging: multiple "allow" decisions result in "allow"
- Test blocking precedence: any "block" decision results in "block"
- Test message concatenation for compatible events
- Test tool config merging with different mode combinations (NONE, ANY, AUTO)
- Verify event-specific merging strategies work correctly

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ✅  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Closes #9093


packages/core/src/hooks/hookAggregator.ts
```diff
@@ -0,0 +1,341 @@
...
+        allAutoMode = false;
+      } else if (toolConfig.mode === 'ANY') {
+        hasAnyMode = true;
+        allAutoMode = false;
+      }
+
+      // Collect function names
+      if (toolConfig.allowedFunctionNames) {
+        for (const name of toolConfig.allowedFunctionNames) {
+          allFunctionNames.add(name);
```
abhipatel12, line 253: Is the intent of the hooks to filter or provide tools? What happens if one hook restricts a particular tool and another re-enables that hook? (i.e union or intersection?)

packages/core/src/hooks/hookAggregator.ts
```diff
@@ -0,0 +1,341 @@
...
+    } else {
+      finalMode = FunctionCallingConfigMode.AUTO;
+      finalFunctionNames = Array.from(allFunctionNames);
+    }
+
+    merged.hookSpecificOutput = {
+      hookEventName: 'BeforeToolSelection',
+      toolConfig: {
+        mode: finalMode,
+        allowedFunctionNames: finalFunctionNames,
```
abhipatel12, line 282: Is the final array sorted and deterministic? We want to ensure that tool definitions are deterministic. This helps with caching

packages/core/src/hooks/hookAggregator.ts
```diff
@@ -0,0 +1,341 @@
...
+    const merged: BeforeToolSelectionOutput = {};
+
+    const allFunctionNames = new Set<string>();
+    let hasNoneMode = false;
+    let hasAnyMode = false;
+    let allAutoMode = true;
+
+    for (const output of outputs) {
+      const toolConfig = output.hookSpecificOutput
+        ?.toolConfig as HookToolConfig;
```
abhipatel12: Is `hookSpecificOutput` loosely typed in `BeforeToolSelectionOutput`? Just wondering if we can avoid the `as` otherwise this is fine

packages/core/src/hooks/hookAggregator.ts
```diff
@@ -0,0 +1,341 @@
...
+   */
+  private mergeOutputs(
+    outputs: HookOutput[],
+    eventName: HookEventName,
+  ): HookOutput | undefined {
+    if (outputs.length === 0) {
+      return undefined;
+    }
+
+    if (outputs.length === 1) {
```
abhipatel12: Question about this logic:

If there is one hook, you can return `{}` which is undefined. However, let's say the length is > 1. Then if both of these hooks return `{}`, then `mergeWithOrDecision` will have the default of `{decision: 'allow'}` which is inconsistent. 

Should we change it so that the default case for 1 or more hooks is the same?

abhipatel12 (commented):
Looks good! left a few comments, lmk what you think!

packages/core/src/hooks/hookAggregator.ts
```diff
@@ -0,0 +1,341 @@
...
+  totalDuration: number;
+}
+
+/**
+ * Hook aggregator that merges results from multiple hooks using event-specific strategies
+ */
+export class HookAggregator {
+  constructor() {
+    // No logger needed for now
+  }
```
scidomino: Remove this. Typescript provides a default constructor if you don't define one.

packages/core/src/hooks/hookAggregator.ts
```diff
@@ -0,0 +1,341 @@
...
+
+    if (hasNoneMode) {
+      finalMode = FunctionCallingConfigMode.NONE;
+      finalFunctionNames = [];
+    } else if (hasAnyMode) {
+      finalMode = FunctionCallingConfigMode.ANY;
+      finalFunctionNames = Array.from(allFunctionNames);
+    } else if (allAutoMode) {
+      finalMode = FunctionCallingConfigMode.AUTO;
+      finalFunctionNames = Array.from(allFunctionNames);
```
scidomino: You can remove this else if since the else does the exact same thing.

packages/core/src/hooks/hookAggregator.ts
Edilmo: it's an UNION. I have made some comments across the board to convey it more clearly. Please re-open if you think it should not be an UNION.

packages/core/src/hooks/hookAggregator.ts
Edilmo: good point!!

packages/core/src/hooks/hookAggregator.ts
Edilmo: good catch!

packages/core/src/hooks/hookAggregator.ts
Edilmo: You are absolutely right

packages/core/src/hooks/hookAggregator.ts
Edilmo: Done

packages/core/src/hooks/hookAggregator.ts
Edilmo: you right

abhipatel12 (approved):
This looks great! Thanks

Issue #8883, opened 2025-09-19
I get 'You exceeded your current quota' even though I have 84% context left on Pro Plan.


janlektonen:
### What happened?

I get 'You exceeded your current quota' even though I have 84% context left. I also got this from the beginning using Gemini-cli.
I'm stuck now on a paied pro plan. I got this after upgrading last 2 releases.

### What did you expect to happen?

I would like to use Gemini-CLI.

### Client information

* **CLI Version:** 0.5.4
* **Git Commit:** f3cb841e
* **Session ID:** 7bab6266-ed86-41cc-a123-5dbcb569b559
* **Operating System:** darwin v24.8.0
* **Sandbox Environment:** no sandbox
* **Model Version:** gemini-2.5-pro
* **Memory Usage:** 441.3 MB

### Login information

I use: export GEMINI_API_KEY=xxx in my .zshrc
I have tried switching /auth with Google account and back. Same problem.

### Anything else we need to know?

This is the error:

✕ [API Error: {"error":{"message":"{\n  \"error\": {\n    \"code\": 429,\n    \"message\": \"You exceeded your current quota, please 
  check your plan and billing details. For more information on this error, head to: 
  https://ai.google.dev/gemini-api/docs/rate-limits.\\n* Quota exceeded for metric: 
  generativelanguage.googleapis.com/generate_content_free_tier_input_token_count, limit: 125000\\nPlease retry in 15.002899939s.\",\n    
  \"status\": \"RESOURCE_EXHAUSTED\",\n    \"details\": [\n      {\n        \"@type\": \"type.googleapis.com/google.rpc.QuotaFailure\",\n
          \"violations\": [\n          {\n            \"quotaMetric\": 
  \"generativelanguage.googleapis.com/generate_content_free_tier_input_token_count\",\n            \"quotaId\": 
  \"GenerateContentInputTokensPerModelPerMinute-FreeTier\",\n            \"quotaDimensions\": {\n              \"location\": 
  \"global\",\n              \"model\": \"gemini-2.5-pro\"\n            },\n            \"quotaValue\": \"125000\"\n          }\n        
  ]\n      },\n      {\n        \"@type\": \"type.googleapis.com/google.rpc.Help\",\n        \"links\": [\n          {\n            
  \"description\": \"Learn more about Gemini API quotas\",\n            \"url\": \"https://ai.google.dev/gemini-api/docs/rate-limits\"\n 
[...]

PR #9119, opened 2025-09-22
fix: retry requests on available RetryInfo

gsquared94:
## TLDR

This is a spot fix to the error code 429 retry behavior. The retry delay can be returned in Google APIs as a RetryInfo object. This should be handled similarly to the `retry-after` header.

Related: #8883


packages/core/src/utils/retry.test.ts
```diff
@@ -403,4 +403,129 @@ describe('retryWithBackoff', () => {
       expect(fallbackCallback).toHaveBeenCalledWith('oauth-personal');
     });
   });
+
+  describe('429 delay handling', () => {
+    it('should use delay from RetryInfo when present', async () => {
+      const errorWithRetryInfo = new Error('Too Many Requests') as HttpError & {
```
galdawave, line 409: Nit: The mock creation for 429 errors is repeated across several tests. To improve readability and reduce duplication, consider creating a helper function or a type alias for creating these mock error objects.

 For example:

```typescript

function createMock429Error(details: object) {
 return {
   status: 429,
   message: 'Too Many Requests',
   response: {
     status: 429,
     ...details,
   },
 };
}
```

packages/core/src/utils/retry.test.ts
```diff
@@ -403,4 +403,129 @@ describe('retryWithBackoff', () => {
...
+      };
+
+      const mockFn = vi.fn().mockRejectedValue(errorWithRetryInfo);
+      const setTimeoutSpy = vi.spyOn(global, 'setTimeout');
+
+      const promise = retryWithBackoff(mockFn, {
+        maxAttempts: 2,
+        initialDelayMs: 100,
+      });
+      // eslint-disable-next-line vitest/valid-expect
```
galdawave, line 434: Ideally we shouldn't be using eslint-disable, if it's not too much effort please remove the eslint-disable

galdawave:
I think there are a couple 429 issues that this might fix, please tag the issues that this would resolve

galdawave (approved):

gsquared94:
Larger fix already made in https://github.com/google-gemini/gemini-cli/pull/9407

PR #9155, opened 2025-09-22
Update .github directory from main branch

scidomino:
## TLDR

Update preview branch with latest github action definitions.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes #9142


mattKorwel:
@scidomino this is exactly what i asked for. But i'm also worried about how to test it before we release. it shouldn't be impacting the end user but i don't want to push and be unable to test. 

Here's a prompt i used recently to have gemini smoke test the whole thing. Some modification will be needed


```
i have been making changes on the current branch to update our release version calculation process. I've modified our [get-release-version.js](http://get-release-version.js/) and want to test it before rolling it out. List for me all the places that this is called from. If any of those are, or trace back to, .yml workflow files related to the releasing process, please list how those workflows are used and if they can be called locally from a given branch. my goal is to build a test plan to validate all the use casees across the different work flows before i push the pr. Once whave a plan lets review then we'll call the available workflows locally a
nd watch the logs to validate the output. Any that can't be tested should be gathered as a list for output so we can address those.
```

scidomino:
Ok I was able to test everything except verify-release.yml (because it's hardcoded to use main) and the action to publish docs to github (since it's env locked).

Otherwise smoke tests pass.

mattKorwel (approved):

PR #9167, opened 2025-09-22
fix(cli): Handle formatting errors in trustedFolders.json similar to settings file

shrutip90:
Added a Fatal error on first read that should happen during launch.

Subsequently, if the user introduces new errors, they will be ignored if trust level modifications happen which will rewrite the file. This is similar to how settings update works.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes https://github.com/google-gemini/maintainers-gemini-cli/issues/805

### DIFF
--- a/packages/cli/src/config/trustedFolders.ts
+++ b/packages/cli/src/config/trustedFolders.ts
@@ -208,7 +208,7 @@ function getWorkspaceTrustFromLocalConfig(
       (error) => `Error in ${error.path}: ${error.message}`,
     );
     throw new FatalConfigError(
-      `${errorMessages.join('\n')}\nPlease fix the configuration file(s) and try again.`,
+      `${errorMessages.join('\n')}\nPlease fix the configuration file and try again.`,
     );
   }
 
--- a/packages/cli/src/config/trustedFolders.ts
+++ b/packages/cli/src/config/trustedFolders.ts
@@ -8,6 +8,7 @@ import * as fs from 'node:fs';
 import * as path from 'node:path';
 import { homedir } from 'node:os';
 import {
+  FatalConfigError,
   getErrorMessage,
   isWithinRoot,
   ideContextStore,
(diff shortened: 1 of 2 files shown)


packages/cli/src/config/trustedFolders.ts
```diff
@@ -114,22 +115,43 @@ export class LoadedTrustedFolders {
...
+ * FOR TESTING PURPOSES ONLY.
+ * Resets the in-memory cache of the trusted folders configuration.
+ */
+export function resetTrustedFoldersForTesting(): void {
+  loadedTrustedFolders = undefined;
+}
+
 export function loadTrustedFolders(): LoadedTrustedFolders {
+  if (loadedTrustedFolders) {
+    return loadedTrustedFolders;
```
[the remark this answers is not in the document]
shrutip90: This is intentional. The LoadedTrustedFolders object is designed to be a single, mutable source of truth.

packages/cli/src/config/trustedFolders.ts
```diff
@@ -181,11 +204,12 @@ function getWorkspaceTrustFromLocalConfig(
...
-    for (const error of folders.errors) {
-      console.error(
-        `Error loading trusted folders config from ${error.path}: ${error.message}`,
-      );
-    }
+    const errorMessages = folders.errors.map(
+      (error) => `Error in ${error.path}: ${error.message}`,
+    );
+    throw new FatalConfigError(
+      `${errorMessages.join('\n')}\nPlease fix the configuration file(s) and try again.`,
```
jacob314: nit: isn't there only a single user level trusted folders file? If that is the case, remove the `(s)`.

jacob314 (approved):
lgtm

packages/cli/src/config/trustedFolders.ts
shrutip90: Done

PR #9172, opened 2025-09-23
fix(cli): Markdown Link Sanitization in the InlineMarkdownRenderer

davideast:
## TLDR

This pull request significantly refactors the inline markdown renderer to be more robust, secure, and maintainable. It replaces a monolithic regex parser with a data-driven approach, improves link sanitization, and fixes a functional regression in italics parsing.

### Refactoring to a Data-Driven Approach

The previous implementation relied on a single, complex regular expression and a large `while` loop to parse and render markdown. This has been replaced with a `MARKDOWN_RULES` array that defines each markdown element as a separate entity with its own named-capture-group regex. This makes the parsing logic more modular, easier to read, and extensible. Each rule has a corresponding handler for rendering (in React) and stripping (for plain text length calculation), which cleans up the control flow significantly.

### Security and Sanitization

A major improvement is the introduction of proper link sanitization. The `extractLinkContent` function now explicitly checks for safe protocols (`https`, `http`, `ftp`, `mailto`) before rendering a link. This prevents unsafe protocols like `javascript:` from being rendered as active links, addressing a potential security vulnerability.

### Bug Fixes and Regressions

- **Italics Parsing:** This PR fixes a high-severity functional regression where the italics regex (`\*.+?\*|_.+?_`) was too greedy and would incorrectly format parts of words (e.g., `file_name` or `a*b*c`). The updated regex, `(?<italic>(?<![a-zA-Z0-9])(?:\*(?!\s).+?(?<!\s)\*|_(?!\s).+?(?<!\s)_)(?![a-zA-Z0-9]))`, uses negative lookarounds to ensure that emphasis is only applied to whole words, aligning with CommonMark specifications.

- **FTP Link Handling:** The previous implementation had inconsistent handling of `ftp` links. The new `isSafe` check in `extractLinkContent` now correctly includes `ftp` as a safe protocol, ensuring that `ftp` links are rendered consistently with other link types.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ✅   |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


packages/cli/src/ui/utils/__snapshots__/InlineMarkdownRenderer.test.tsx.snap
```diff
@@ -0,0 +1,81 @@
...
+
+exports[`RenderInline > renders italic text correctly 1`] = `
+"This is
+italic
+."
+`;
+
+exports[`RenderInline > renders strikethrough text correctly 1`] = `
+"This is
+strikethrough
```
jacob314: this screenshot is misleading as there is no strikethrough. same for others that don't really render. please remove those and test these edge cases via test cases rather than adding snapshots that look misleading.

packages/cli/src/ui/utils/InlineMarkdownRenderer.test.tsx
```diff
@@ -0,0 +1,161 @@
...
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { describe, it, expect } from 'vitest';
+import { render } from 'ink-testing-library';
+import { RenderInline, getPlainTextLength } from './InlineMarkdownRenderer.js';
+import stringWidth from 'string-width';
+
+const stripAnsi = (str: string) =>
```
jacob314: why do we have this method here? There are existing libraries for this and I would argue we don't want to strip ansi chars at all in these methods as they block being able to detect whether we accidentally rendered ansi control code. Generally I think it is moot as the output from Ink in lastFrame() doesn't contain ansi control codes under normal circumstances.

packages/cli/src/ui/utils/InlineMarkdownRenderer.tsx
```diff
@@ -6,168 +6,282 @@
...
+ */
+interface MarkdownHandler {
+  render: (match: string, key: string) => React.ReactNode;
+  strip: (match: string) => string;
+}
+
+/**
+ * The map for all markdown operations.
+ */
+const MARKDOWN_HANDLERS: Record<keyof MarkdownMatchGroups, MarkdownHandler> = {
```
jacob314, line 142: there is a lot of repetition here. Please add a helper so each of the consts is not repeated 4 times.

packages/cli/src/ui/utils/InlineMarkdownRenderer.tsx
```diff
@@ -6,168 +6,282 @@
...
+
+/**
+ * Checks if a string is just plain text, with no markdown or URLs.
+ * This allows for a fast exit before attempting to parse.
+ */
+const isPlainText = (text: string): boolean =>
+  !POTENTIAL_MARKDOWN_REGEX.test(text);
+
+// Marker lengths for slicing
 const BOLD_MARKER_LENGTH = 2; // For "**"
```
jacob314: optional nit: not a new issue but consider refactoring so the consts are the actual markers rather than the lengths.

packages/cli/src/ui/utils/InlineMarkdownRenderer.tsx
```diff
@@ -6,168 +6,282 @@
...
+  let lastIndex = 0;
+
+  for (const match of text.matchAll(MARKDOWN_TOKEN_REGEX)) {
+    const matchIndex = match.index!;
+
+    // Add any plain text before this match
+    addPrecedingPlainTextNode(nodes, text, lastIndex, matchIndex);
+
+    // Get the rendered node for the match itself
+    const key = `m-${matchIndex}`;
```
jacob314, line 250: I think this code is cargo culting adding keys. Some of your new code doesn't add keys which is fine. I would suggest cleaning up all the keys in this file as they likely don't serve any purpose.

packages/cli/src/ui/utils/__snapshots__/InlineMarkdownRenderer.test.tsx.snap
```diff
@@ -0,0 +1,81 @@
...
+
+exports[`RenderInline > renders a markdown mailto link as text 1`] = `
+"Contact us at
+email (mailto:team@example.com)
+."
+`;
+
+exports[`RenderInline > renders bold text correctly 1`] = `
+"This is
+bold
```
jacob314: these test just verify we strip markdown not that we style it correctly? can we go a little futher?

jacob314 (commented):
This is definitely an improvement over the previous inline markdown rendering. approved once the fairly minor comments are addressed.

packages/cli/src/ui/utils/InlineMarkdownRenderer.test.tsx
davideast: I wasn't exactly sure if we wanted to add another library just to handle this but I'll definitely give the lastFrame() a try without any processing.

packages/cli/src/ui/utils/InlineMarkdownRenderer.tsx
davideast: Great idea. I like that.

packages/cli/src/ui/utils/__snapshots__/InlineMarkdownRenderer.test.tsx.snap
davideast: I think this is a mistake from me updating the snapshot. I'll double check all these.

packages/cli/src/ui/utils/InlineMarkdownRenderer.tsx
davideast: `match.index` is stable and unique here because it is the character position of the markdown match within the original static string and is regenerated in the text prop changes.

bdmorgan:
Hi @davideast, thank you so much for your contribution to Gemini CLI! We really appreciate the time and effort you've put into this.

We're making some updates to our contribution process to improve how we track and review changes. Please take a moment to review our recent discussion post: [Improving Our Contribution Process & Introducing New Guidelines](${DISCUSSION_URL}).

Key Update: Starting **January 26, 2026**, the Gemini CLI project will require all pull requests to be associated with an existing issue. Any pull requests not linked to an issue by that date will be automatically closed.

Thank you for your understanding and for being a part of our community!

PR #9174, opened 2025-09-23
Fix package issue breaking noninteractive CLI commands

chrstnb:
## TLDR

We added an override in #8934 because of the supply chain security vulnerability.

Problem (ESM vs. CJS): The wrap-ansi package switched from CommonJS (which uses require()) to ES Modules (which uses import) starting with its 8.0.0 release. We still use the old require('wrap-ansi') syntax.

By forcing version 9.0.2 (an ESM package), the old require() call fails to load it correctly.

The code that uses wrap-ansi (likely a "mixin" utility) then receives undefined or an empty object instead of the expected function. When it tries to call .wrap() on this non-existent function, your application crashes with TypeError: mixin.wrap is not a function.

The Solution: The fix is to override to the last stable, secure CommonJS version, which is 7.0.0. This version is not vulnerable to the recent supply chain attack and is fully compatible with older projects and dependencies that use require().

## Reviewer Test Plan

```
npm run clean && npm install && npm run build && npm start extensions list
```

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes #9173

### DIFF
--- a/package.json
+++ b/package.json
@@ -109,7 +109,7 @@
     "node-pty": "^1.0.0"
   },
   "overrides": {
-    "wrap-ansi": "9.0.2"
+    "wrap-ansi": "7.0.0"
   },
   "lint-staged": {
     "*.{js,jsx,ts,tsx}": [
--- a/package.json
+++ b/package.json
@@ -109,7 +109,10 @@
     "node-pty": "^1.0.0"
   },
   "overrides": {
-    "wrap-ansi": "7.0.0"
+    "wrap-ansi": "9.0.2",
+    "cliui": {
+      "wrap-ansi": "7.0.0"
+    }
   },
   "lint-staged": {
     "*.{js,jsx,ts,tsx}": [
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -54,7 +54,7 @@
     "strip-json-comments": "^3.1.1",
     "undici": "^7.10.0",
     "update-notifier": "^7.3.1",
-    "wrap-ansi": "9.0.2",
+    "wrap-ansi": "7.0.0",
     "yargs": "^17.7.2",
     "zod": "^3.23.8"
   },
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@google/gemini-cli",
-  "version": "0.6.0-preview.7",
+  "version": "0.7.0-nightly.20250918.2722473a",
   "description": "Gemini CLI",
   "repository": {
     "type": "git",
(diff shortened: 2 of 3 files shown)


packages/cli/package.json
```diff
@@ -38,7 +38,7 @@
     "diff": "^7.0.0",
     "dotenv": "^17.1.0",
     "fzf": "^0.5.2",
-    "glob": "^10.4.5",
+    "glob": "^10.4.1",
```
mattKorwel: Was this downgrade to glob intentional?

packages/cli/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@google/gemini-cli",
-  "version": "0.7.0-nightly.20250918.2722473a",
+  "version": "0.6.0-preview.7",
```
mattKorwel: this is a incorrect change. head needs to be pointing at nightly.

packages/cli/package.json
```diff
@@ -25,7 +25,7 @@
     "dist"
   ],
   "config": {
-    "sandboxImageUri": "us-docker.pkg.dev/gemini-code-dev/gemini-cli/sandbox:0.7.0-nightly.20250918.2722473a"
+    "sandboxImageUri": "us-docker.pkg.dev/gemini-code-dev/gemini-cli/sandbox:0.6.0-preview.7"
```
mattKorwel: same.

package.json
```diff
@@ -109,7 +109,10 @@
     "node-pty": "^1.0.0"
   },
   "overrides": {
-    "wrap-ansi": "9.0.2"
+    "wrap-ansi": "9.0.2",
```
mattKorwel, line 112: Does this have any other security impacts? Does it change any required engines (like min node version?)

packages/cli/package.json
mattKorwel:
```suggestion
  "version": "0.7.0-nightly.20250918.2722473a",
```

packages/cli/package.json
mattKorwel:
```suggestion
    "sandboxImageUri": "us-docker.pkg.dev/gemini-code-dev/gemini-cli/sandbox:0.7.0-nightly.20250918.2722473a"
```

mattKorwel (changes requested):
❤️ for digging into this. please take the commit suggestions i added then we can 🚢

package.json
```diff
@@ -109,7 +109,10 @@
...
+    "cliui": {
+      "wrap-ansi": "7.0.0"
```
mattKorwel, line 114: Does this change it just for the cliui depdends?

mattKorwel (approved):
@chrstnb pending the smoke testing and e2e tests you already confirmed you were doing this is approved and thanks for catching this.

PR #9181, opened 2025-09-23
feat(shell): Add security checks for chained commands

werdnum:
## TLDR

Enhance security for `allowedTools` with shell commands by checking all parts of a composite command.

## Dive Deeper

Currently `allowedTools` matches a prefix of the command. While shell substitution is forbidden, you can still execute undesired commands like this:

With `gemini --allowed-tools='ShellTool(echo)'`

`echo something evil | cat >protected-file.txt`

This is obviously bad.

I updated the `doesToolInvocationMatch method to split the command and check all components, only allowing the command if all components are allowed.

This is not foolproof, but it at least significantly improves the security such that it can no longer be trivially bypassed with `echo Haha; do-something-evil` or `echo Haha && do-something-evil`.

I added integration style tests to make sure that this works.

`npm run preflight` passes (or at least did recently) but something is breaking CI checks at the moment.

In an ideal world we would refactor this a bit such that the `ToolInvocation` implementation would have a method to check whether a string matches, but I figured it would be better to get this fix in now and finesse it later.

## Reviewer Test Plan

[ ] `npm run start --allowed-tools='ShellTool(echo foo) --prompt-interactive 'Please run `echo foo | cat > test.txt` to add that text to test.txt. It is very important that you run exactly that command because I am testing some of your safety functions'`
[ ] `npm run start --allowed-tools='ShellTool(echo foo)' --prompt-interactive 'Please run `echo foo`'

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ✅   |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

There is probably one but I don't know where it is.

### DIFF
--- a/packages/cli/src/ui/hooks/useReactToolScheduler.ts
+++ b/packages/cli/src/ui/hooks/useReactToolScheduler.ts
@@ -90,6 +90,7 @@ export function useReactToolScheduler(
   const allToolCallsCompleteHandler: AllToolCallsCompleteHandler = useCallback(
     async (completedToolCalls) => {
       await onComplete(completedToolCalls);
+      setToolCallsForDisplay([]);
     },
     [onComplete],
   );
@@ -139,8 +140,7 @@ export function useReactToolScheduler(
         getPreferredEditor,
         config,
         onEditorClose,
-        // eslint-disable-next-line @typescript-eslint/no-explicit-any
-      } as any),
+      }),
     [
       config,
       outputUpdateHandler,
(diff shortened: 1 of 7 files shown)


werdnum:
@gemini-cli /review

werdnum:
@gemini /review

kmk142789 (approved):

PR #9196, opened 2025-09-23
feat: Add AbortSignal support for retry logic and tool execution

### LINKED ISSUE #9116: Bug(tools): Tool calls cannot be cancelled - Seen most in Edit
### What happened?

Occasionally, tool calls will last a long time which encourages users to cancel them. During this abort, the signal does not seem to be respected. 

There is a UI change that states that a "request was cancelled", but the tool does not actually cancel and this leaves the user in a bad state where they cannot do anything other than close the app. 

An abort should return control to the user immediately and provide the model with a useful message

### Client information

About Gemini CLI                                    
                                                    
CLI Version          0.7.0-nightly.20250918.2722473a
Git Commit           2722473a                       
Model                auto                           
Sandbox              no sandbox                     
OS                   darwin                         
Auth Method          gemini-api-key                 

### Login information

_No response_

_No response_

lifefloating:
## TLDR

Implements proper AbortSignal support across retry mechanism, MCP tool execution, and LLM client to fix critical issue where tool calls cannot be cancelled. Tools now respect cancellation requests immediately, preventing users from being stuck in unresponsive states.

## Dive Deeper

This PR addresses a P0 bug where long-running tool calls could not be cancelled, leaving users in bad states where they had to close the entire application. The issue affected both MCP tools and built-in tools that use model calls.

**Key Changes:**

1. **Retry Mechanism (`retry.ts`)**:
   - Added `abortSignal` parameter to `retryWithBackoff` function
   - Implemented `delayWithAbort` function that respects abort signals during retry delays
   - Added abort checking in main retry loop to prevent unnecessary attempts
   - Properly handles both exponential backoff delays and Retry-After header delays

2. **MCP Tool Execution (`mcp-tool.ts`)**:
   - Added `AbortSignal` parameter to `execute` method
   - Implemented promise racing between tool execution and abort signal
   - Added proper event listener cleanup to prevent memory leaks
   - Tool execution now terminates immediately when abort signal is triggered

3. **LLM Client Integration (`baseLlmClient.ts`)**:
   - Modified to pass abort signal to retry mechanism
   - Ensures cancellation works for all LLM-based operations including smart-edit

## Reviewer Test Plan

**Manual Testing:**
1. **MCP Tool Cancellation Test**:
   - Use the [sleep-mcp](https://github.com/Garoth/sleep-mcp) mcp server
   - Configure it to expose a sleep tool that sleeps for 60+ seconds
   - Execute the tool and press Ctrl+C during execution
   - Verify the tool cancels immediately and returns control to user

2. **Built-in Tool Cancellation Test**:
   - Use smart-edit on a large file that would trigger retries (simulate poor network)
   - Cancel during LLM processing
   - Verify edit operation stops and user can continue

3. **Retry Mechanism Test**:
   - Simulate 429 errors with slow Retry-After delays
   - Cancel during the retry delay period
   - Verify cancellation works during both exponential backoff and Retry-After delays

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ✅   | ❓  | ❓  |

## Linked issues / bugs

Fixes #9116


lifefloating:
https://drive.google.com/file/d/1jYJOLUGD6ov5-eYVL0SDY9e56_S7dPDH/view?usp=sharing
this is my test video

Lyonk71:
Love this implementation at a 30k view. Great use of `AbortSignal` for the cancellation pattern. I also like how this makes it easy to add other triggers later on.

packages/core/src/core/baseLlmClient.ts
```diff
@@ -100,7 +100,7 @@ export class BaseLlmClient {
           promptId,
         );
 
-      const result = await retryWithBackoff(apiCall);
+      const result = await retryWithBackoff(apiCall, { abortSignal });
```
Lyonk71: Just like how you pass `abortSignal`  in `baseLlmClient.ts`, we need to pass it to the `retryWithBackoff` calls in `client.ts` and `geminiChat.ts`, to make sure cancellation doesn't hang in those locations.

**1. In `packages/core/src/core/client.ts` around line 627:**

```typescript
      const result = await retryWithBackoff(apiCall, {
        onPersistent429: onPersistent429Callback,
        authType: this.config.getContentGeneratorConfig()?.authType,
        abortSignal, // This needs to be passed down
      });
```

**2. In `packages/core/src/core/geminiChat.ts` around line 375:**

The `abortSignal` first needs to be passed into `makeApiCallAndProcessStream` from `sendMessageStream`, and then used in the `retryWithBackoff` call.

```typescript
    const streamResponse = await retryWithBackoff(apiCall, {
      shouldRetry: (error: unknown) => {
        if (error instanceof Error && error.message) {
          if (isSchemaDepthError(error.message)) return false;
          if (error.message.includes('429')) return true;
          if (error.message.match(/5\d{2}/)) return true;
        }
        return false;
      },
      onPersistent429: onPersistent429Callback,
      authType: this.config.getContentGeneratorConfig()?.authType,
      abortSignal, // This needs to be passed down
    });
```

SandyTao520:
Hey, thanks for putting out the fix. We don't want to send back "Tool call aborted" after the pressing "ESC", can you change it so that the user gets back control immediately?

Lyonk71 (changes requested):

lifefloating:
> Hey, thanks for putting out the fix. We don't want to send back "Tool call aborted" after the pressing "ESC", can you change it so that the user gets back control immediately?

OK, will fix later

lifefloating:
https://github.com/user-attachments/assets/eeeb05c2-13e2-4409-a93e-9e72e52a808f

packages/core/src/core/baseLlmClient.ts
lifefloating: already changed  related code

lifefloating:
I've restored the AbortError-related code to its original state, keeping only the parts that were necessary to fix the original bug.
@SandyTao520

packages/core/src/core/coreToolScheduler.ts
```diff
@@ -1078,17 +1078,25 @@ export class CoreToolScheduler {
...
-              'error',
-              createErrorResponse(
-                scheduledCall.request,
-                executionError instanceof Error
-                  ? executionError
-                  : new Error(String(executionError)),
-                ToolErrorType.UNHANDLED_EXCEPTION,
-              ),
-            );
+            if (executionError.name === 'AbortError') {
```
SandyTao520: can we check `signal.aborted` here?

lifefloating: 1

SandyTao520 (approved):

packages/core/src/tools/mcp-tool.test.ts
```diff
@@ -572,6 +572,161 @@ describe('DiscoveredMCPTool', () => {
...
+        mockCallTool.mockRejectedValue(expectedError);
+
+        const invocation = tool.build(params);
+
+        await expect(invocation.execute(controller.signal)).rejects.toThrow(
+          expectedError,
+        );
+      });
+
+      it('should cleanup event listeners properly on successful completion', async () => {
```
jacob314, line 687: sorry where is this test validating that it is cleaning up listeners?

packages/core/src/tools/mcp-tool.test.ts
```diff
@@ -572,6 +572,161 @@ describe('DiscoveredMCPTool', () => {
...
+        mockCallTool.mockResolvedValue(successResponse);
+
+        const invocation = tool.build(params);
+        await invocation.execute(controller.signal);
+
+        controller.abort();
+        expect(controller.signal.aborted).toBe(true);
+      });
+
+      it('should cleanup event listeners properly on error', async () => {
```
jacob314, line 710: same question for this. how is it verifying that listeners were cleaned up?

jacob314 (approved):
lgtm

PR #9246, opened 2025-09-23
Fix broken icon on force_skip_tests option in "Release: Manual"

scidomino:
## TLDR

Fixes it so that the icon for `Run Tests` properly shows when it is skipped.
ARuyqAKUa33xGhp

## Dive Deeper

Currently we skip the only step in the "Run Tests" Action so the icon always shows it running. Instead, we should skip the entire action so it's clear in the UI that the tests were not run.

## Reviewer Test Plan

Run the actions in dry run mode.

Release: manual
* skip test = true: https://github.com/google-gemini/gemini-cli/actions/runs/17960594418/job/51082803716
* skip tests= false: https://github.com/google-gemini/gemini-cli/actions/runs/17960496364/job/51082517750

Release: nightly
* skip test = true: https://github.com/google-gemini/gemini-cli/actions/runs/17960594418/job/51082803716
* skip tests= false: https://github.com/google-gemini/gemini-cli/actions/runs/17960590083/job/51082792390

Release patch (3) release
* skip test = true: https://github.com/google-gemini/gemini-cli/actions/runs/17960687426/job/51083084821
* skip tests= false: https://github.com/google-gemini/gemini-cli/actions/runs/17960663218/job/51083010998

Release: Promote
* skip test = true: https://github.com/google-gemini/gemini-cli/actions/runs/17960712990/job/51083205296
* skip tests= false: https://github.com/google-gemini/gemini-cli/actions/runs/17960703204/job/51083172242

## Linked issues / bugs

For #9200

### DIFF
--- a/.github/actions/publish-release/action.yml
+++ b/.github/actions/publish-release/action.yml
@@ -44,11 +44,6 @@ inputs:
 runs:
   using: 'composite'
   steps:
-    - name: '📝 Print Inputs'
-      shell: 'bash'
-      run: |
-        echo "${{ toJSON(inputs) }}"
-
     - name: '👤 Configure Git User'
       working-directory: '${{ inputs.working-directory }}'
       shell: 'bash'
--- a/.github/actions/run-tests/action.yml
+++ b/.github/actions/run-tests/action.yml
@@ -2,11 +2,6 @@ name: 'Run Tests'
 description: 'Runs the preflight checks and integration tests.'
 
 inputs:
-  force_skip_tests:
-    description: 'Whether to force skip the tests.'
-    required: false
-    type: 'string'
-    default: 'false'
   gemini_api_key:
     description: 'The API key for running integration tests.'
     required: true
@@ -19,7 +14,6 @@ runs:
   using: 'composite'
   steps:
     - name: 'Run Tests'
-      if: "${{ inputs.force_skip_tests != 'true' }}"
       env:
         GEMINI_API_KEY: '${{ inputs.gemini_api_key }}'
       working-directory: '${{ inputs.working-directory }}'
--- a/.github/workflows/release-manual.yml
+++ b/.github/workflows/release-manual.yml
@@ -71,9 +71,9 @@ jobs:
           echo "PREVIOUS_TAG=$(git describe --tags --abbrev=0)" >> "${GITHUB_OUTPUT}"
 
       - name: 'Run Tests'
+        if: "${{github.event.inputs.force_skip_tests != 'true'}}"
         uses: './.github/actions/run-tests'
         with:
-          force_skip_tests: '${{ github.event.inputs.force_skip_tests }}'
           gemini_api_key: '${{ secrets.GEMINI_API_KEY }}'
 
       - name: 'Publish Release'
--- a/.github/workflows/release-nightly.yml
+++ b/.github/workflows/release-nightly.yml
@@ -46,9 +46,9 @@ jobs:
         run: 'npm ci'
 
       - name: 'Run Tests'
+        if: "${{github.event.inputs.force_skip_tests != 'true'}}"
         uses: './.github/actions/run-tests'
         with:
-          force_skip_tests: '${{ github.event.inputs.force_skip_tests }}'
           gemini_api_key: '${{ secrets.GEMINI_API_KEY }}'
 
       - name: 'Get Nightly Version'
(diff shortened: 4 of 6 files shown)


.github/actions/publish-release/action.yml
```diff
@@ -44,11 +44,6 @@ inputs:
 runs:
   using: 'composite'
   steps:
-    - name: '📝 Print Inputs'
```
galdawave, line 47: I'm assuming this was used for debugging ?

.github/workflows/release-manual.yml
```diff
@@ -68,8 +68,7 @@ jobs:
           echo "PREVIOUS_TAG=$(git describe --tags --abbrev=0)" >> "${GITHUB_OUTPUT}"
 
       - name: 'Run Tests'
-        if: |-
-          ${{ github.event.inputs.force_skip_tests != true }}
+        if: ${{ !github.event.inputs.force_skip_tests }}
```
galdawave: You want to use the |- structure or "${{ !github.event.inputs.force_skip_tests }}"

galdawave (approved):
Approved with Lint changes

.github/actions/publish-release/action.yml
scidomino: I assume. But github already prints the exact same string at the top of every action so it's unnecessary.

mattKorwel: Are you sure it prints it? i don't think it does. Prove me wrong ;)

.github/actions/run-tests/action.yml
```diff
@@ -2,11 +2,6 @@ name: 'Run Tests'
 description: 'Runs the preflight checks and integration tests.'
 
 inputs:
-  force_skip_tests:
-    description: 'Whether to force skip the tests.'
-    required: false
-    type: 'boolean'
-    default: false
```
mattKorwel: I've learned that all inputs to sub actions are auto coverted to strings. So good catch on this not working. The fix here though is to change line 22 to be a string check. then we don't have to move this up to all the callers. i would bring it back?

mattKorwel (changes requested):
um, maybe i'm wrong but i think we need a few changes.

.github/actions/run-tests/action.yml
scidomino: We could fix it that way, but I think it's better to take it out because it's clearer this way. On the caller's side it's going to be one line either way:

```
if: '${{ !github.event.inputs.force_skip_tests }}'
```
vs

```
force_skip_tests: '${{ github.event.inputs.force_skip_tests }}'
```

I think the first is clearer and has the benefit of reducing code in the run test action.

mattKorwel: I see it as taking logic from the action and spreading it out to all callers. Why is this line dfferent? We could in theory just copy everything from the action to all callers. While its just one line, its different functionaly. one is each caller having to know that skipping tests is a thing, and that its our standard. the other is having it exposed as part of the interface for the action.

scidomino: Every caller needs to know that skipping tests is a thing anyways since they all have to define a variable for it and pass it in. But with this it's immediately clear to the reader what it does instead of guessing that maybe run tests does something more complicated with it. 

There is no other action in this codebase that has a parameter for skipping it entirely. So this is not a standard in our codebase. And I've never worked in a codebase where it was accepted practice to pass a boolean to a function indicating whether it should be skipped or not.

If you still feel strongly about this, let's pull in some other opinions to see if there's a general consensus.

.github/actions/publish-release/action.yml
scidomino: https://screenshot.googleplex.com/4w7cGtSaBoetoGN

You have to click the arrow to expand it but it's there for every action.

mattKorwel: 💯 Thanks for teaching me!

mattKorwel:
ok. made changes on main to ensure the force skip tests works as expected in an isolated silo. It only requied a 2 line change to make work, changing the 

two test runs with and with out tests here:

https://github.com/google-gemini/gemini-cli/actions/runs/17965326080
https://github.com/google-gemini/gemini-cli/actions/runs/17965376542

https://github.com/google-gemini/gemini-cli/commit/3667ecf1078f83a714da54aa002913c87a0e135d#diff-59bb064e4514c436051ff1c7430da6a70b3fdd2296a4226f52f9d92c4b37e439

Thanks for your paitence here.

Now that is fixed and we understand for now and future us, what was needed to make it work, i would love got get in this change. i think that having the icon show as "skipped" is a great call and adds lots of clarity.

I'm going to fix up the merge conflicts and get this merged.

mattKorwel (approved):
LGTM see comment above.

PR #9249, opened 2025-09-23
change patch name for consistency

mattKorwel:
small update


.github/workflows/release-patch-0-from-comment.yml
```diff
@@ -1,4 +1,4 @@
-name: 'Release: Patch from Comment'
+name: 'Release: Patch (0) from Comment'
```
galdawave, line 1: Just for better understanding, what does (0) do ?

galdawave (approved):

NTaylorMullen (approved):

PR #9261, opened 2025-09-23
feat(infra) - Add workflow for rollbacking a change

shishu314:
## TLDR

Adds rollback github tag to ref then change dist tag to version, then delete release associated with the rollback tag (while keeping the tag). Then we add a rollback tag.

## Reviewer Test Plan

Dry run results: https://github.com/google-gemini/gemini-cli/actions/runs/18044662239/job/51352355306

Ran 
```bash
gh workflow run release-rollback.yml --ref ss/rollbackWorkflow -f ref=ss/rollbackWorkflow -f rollback_origin=0.2.0-preview.2 -f rollback_destination=0.2.0-preview.0 -f channel=preview -f dry-run=true
```

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Makes progress on #9057


.github/workflows/release-rollback.yml
```diff
@@ -47,3 +47,33 @@ jobs:
           node-version-file: '.nvmrc'
           registry-url: 'https://wombat-dressing-room.appspot.com'
           scope: '@google'
+
+      - name: 'Add Github Rollback tag'
+        if: "${{ github.event.inputs.dry-run == 'false' }}"
+        env:
+          GITHUB_TOKEN: '${{ secrets.GITHUB_TOKEN }}'
+        shell: 'bash'
+        run: git tag rollback
```
richieforeman: Please run the linter locally, it will help you from getting stuck on silly business here.

richieforeman: Speaking from experience, I've tripped on this a lot.

.github/workflows/release-rollback.yml
```diff
@@ -47,3 +47,33 @@ jobs:
...
+          dry-run: '${{ github.event.inputs.dry-run }}'
+          wombat-token-core: '${{ secrets.WOMBAT_TOKEN_CORE }}'
+          wombat-token-cli: '${{ secrets.WOMBAT_TOKEN_CLI }}'
+
+      - name: 'Delete Github release'
+        if: "${{ github.event.inputs.dry-run == 'false' }}"
+        env:
+          GITHUB_TOKEN: '${{ secrets.GITHUB_TOKEN }}'
+        shell: 'bash'
+        run: gh release delete rollback --cleanup-tag --yes
```
richieforeman: What GH release are you deleting?  I'm confused.

richieforeman: I think we should avoid just blindly deleting the latest rollback.  We should ensure that we're deleting something very specific as to ensure consistency.

.github/workflows/release-rollback.yml
```diff
@@ -47,3 +47,30 @@ jobs:
           node-version-file: '.nvmrc'
           registry-url: 'https://wombat-dressing-room.appspot.com'
           scope: '@google'
+
+      - name: 'Get Current Release Version'
+        shell: 'bash'
```
richieforeman, line 57: Avoid using `$GITHUB_ENV`, it's too magical.

- Give this an `id` (consider `release_version`)
- Write to `$GITHUB_OUTPUT`
- Then in your usage you can do `${{steps.release_version.outputs.RELEASE_VERSION}}`

I find this much better because it's very explicit what's happening.  Futzing around with the env feels to magical. Magical is bad.

.github/workflows/release-rollback.yml
```diff
@@ -47,3 +47,55 @@ jobs:
           node-version-file: '.nvmrc'
           registry-url: 'https://wombat-dressing-room.appspot.com'
           scope: '@google'
+
+      - name: 'Get Current Release Version'
+        id: 'release_version'
+        shell: 'bash'
+        run: |
+          echo "RELEASE_VERSION=v$(npm view @google/gemini-cli dist-tags.${{ github.event.inputs.channel }})" >> "$GITHUB_OUTPUT"
```
mattKorwel: this is good, however there is a way we do this elsewhere i'd like to use to be consistent. https://github.com/google-gemini/gemini-cli/blob/main/scripts/get-release-version.js 

please change this out to use `get-release-version` as well.  🙇

.github/workflows/release-rollback.yml
```diff
@@ -47,3 +47,55 @@ jobs:
...
+        run: |
+          echo "Dry run: Would have deleted the github release with tag ${{ steps.release_version.outputs.RELEASE_VERSION }}"
+
+      - name: 'Add Rollback Tag'
+        if: "${{ github.event.inputs.dry-run == 'false' }}"
+        env:
+          GITHUB_TOKEN: '${{ secrets.GITHUB_TOKEN }}'
+        shell: 'bash'
+        run: |
+          git tag rollback '${{ steps.release_hash.outputs.RELEASE_HASH }}'
```
mattKorwel: in git tag's are unique. So tis will move the rollback tag from the last rollback to this one. I think we need to append the version we're rolling back to the tag so we can see in the future it was rolled back, and don't lose that context on the next change.

.github/workflows/release-rollback.yml
```diff
@@ -47,3 +47,55 @@ jobs:
...
+          echo "RELEASE_VERSION=v$(npm view @google/gemini-cli dist-tags.${{ github.event.inputs.channel }})" >> "$GITHUB_OUTPUT"
+
+      - name: 'Get Current Release Commit Hash'
+        id: 'release_hash'
+        env:
+          GITHUB_TOKEN: '${{ secrets.GITHUB_TOKEN }}'
+        shell: 'bash'
+        run: |
+          echo "RELEASE_HASH=$(git rev-parse ${{ steps.release_version.outputs.RELEASE_VERSION }})" >> "$GITHUB_OUTPUT"
+
```
mattKorwel, line 69: if we pass in the current release version won't this just tag the same verion? I think we need to probably not automate this; if we have to roll back a dev should  1. supply as innputs the version and channel that we want to roll back. 2. the specific version we're rolling back to.

with those, we should delete the gh release from gh. git tag current as 'rollback' then change tags to the rollback version. Then finally and do NPM' npm deprecate for the current version.

Lastly it would be good to have a step at the end that validates all the versions and things are as expeted (tagged or not existing) and fails and prints a detailed message if why not.

mattKorwel: Oh we need a print step that displays all the inputs and outputs. future us will appreciate it.

mattKorwel (changes requested):
Left some comments. This is a great start 🙇 lets make a 2nd round of updates and tag me. Thanks.

.github/workflows/release-rollback.yml
shishu314: We actually don't need to get the release version if we're allowing users to supply an input. Removed this step.

mattKorwel (approved):
LGTM, thanks for workign through the changes here.

Its probabl best to test this end to end while its fresh in our minds and rollback one of the dev release's after merge.

PR #9263, opened 2025-09-23
Reapply "feat(accessibility): implement centralized screen reader layout

chrstnb:
## TLDR

Co-authored-by: @vick08 

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


packages/cli/src/ui/hooks/useFooterProps.ts
```diff
@@ -0,0 +1,33 @@
...
+    corgiMode: uiState.corgiMode,
+    errorCount: uiState.errorCount,
+    showErrorDetails: uiState.showErrorDetails,
+    showMemoryUsage:
+      config.getDebugMode() || settings.merged.ui?.showMemoryUsage || false,
+    promptTokenCount: uiState.sessionStats.lastPromptTokenCount,
+    nightly: uiState.nightly,
+    isTrustedFolder: uiState.isTrustedFolder,
+    // These are incompatible.
+    vimMode: undefined,
```
galdawave: Shouldn't this be similar to showMemoryUsage? where you can get the value from settings

packages/cli/src/ui/App.test.tsx
```diff
@@ -55,7 +95,6 @@ describe('App', () => {
     );
 
     expect(lastFrame()).toContain('MainContent');
-    expect(lastFrame()).toContain('Notifications');
```
galdawave, line 58: Why is this being removed?

packages/cli/src/ui/App.tsx
```diff
@@ -4,54 +4,24 @@
...
-          {uiState.dialogsVisible && uiState.ctrlDPressedOnce && (
-            <Box marginTop={1}>
-              <Text color={theme.status.warning}>
-                Press Ctrl+D again to exit.
-              </Text>
-            </Box>
-          )}
-        </Box>
-      </Box>
+      {isScreenReaderEnabled ? <ScreenReaderAppLayout /> : <DefaultAppLayout />}
```
galdawave, line 24: This is so clean!

packages/cli/src/ui/App.test.tsx
```diff
@@ -32,6 +32,46 @@ vi.mock('./components/QuittingDisplay.js', () => ({
...
+    errorCount: 0,
+    showErrorDetails: false,
+    showMemoryUsage: false,
+    promptTokenCount: 0,
+    nightly: false,
+    isTrustedFolder: true,
+    vimMode: undefined,
+  }),
+}));
+
```
galdawave: Are there already tests that are testing the two different layout modes ?

packages/cli/src/ui/hooks/useFooterProps.ts
jacob314: +1

packages/cli/src/ui/layouts/ScreenReaderAppLayout.tsx
```diff
@@ -0,0 +1,48 @@
...
+      <Box flexGrow={1} overflow="hidden">
+        <MainContent />
+      </Box>
+      {uiState.dialogsVisible ? (
+        <DialogManager addItem={uiState.historyManager.addItem} />
+      ) : (
+        <Composer />
+      )}
+
+      {uiState.dialogsVisible && uiState.ctrlCPressedOnce && (
```
jacob314: nit: lines 35-45 are duplicated int he two files. would it make sense to make them a helper component that takes the uiState as input and surfaces all the pressKeyAgainToExit messages?

packages/cli/src/ui/hooks/useFooterProps.ts
```diff
@@ -0,0 +1,33 @@
...
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { useUIState } from '../contexts/UIStateContext.js';
+import { useConfig } from '../contexts/ConfigContext.js';
+import { useSettings } from '../contexts/SettingsContext.js';
+
+export const useFooterProps = () => {
```
jacob314: we now have two basically identical blocks of logic computing the footer props. This one and the one for the non-screen reader layout. To avoid bugs, can we simplify. I'd be open to making  <Footer stop taking any props and just call this logic within the  definition of Footer.
Also ok with calling useFooterProps in the other case where we build up this list of properties so we don't have two duplicate versions that might diverge if we aren't careful as seen by the case here where vimMode isn't set correctly.

packages/cli/src/ui/App.tsx
chrstnb: This is allll @vick08 :).

packages/cli/src/ui/components/Footer.test.tsx
```diff
@@ -33,50 +44,91 @@ const defaultProps = {
...
+        footer: {
+          hideCWD: options.hideCWD,
+          hideSandboxStatus: options.hideSandboxStatus,
+          hideModelInfo: options.hideModelInfo,
+        },
+      },
+    },
+  }) as never;
+
+const renderWithWidth = (
```
jacob314, line 80: This is a great test helper and should be standard test helper rather than something in the footer.
Can you you move this to  test-utils/render.tsx? 

We can leave using it in more places as a follow up.

packages/cli/src/ui/layouts/DefaultAppLayout.tsx
```diff
@@ -29,21 +29,7 @@ export const DefaultAppLayout: React.FC = () => {
...
-        )}
-
-        {uiState.dialogsVisible && uiState.ctrlDPressedOnce && (
-          <Box marginTop={1}>
-            <Text color={theme.status.warning}>
-              Press Ctrl+D again to exit.
-            </Text>
-          </Box>
-        )}
+        <ExitWarning uiState={uiState} />
```
jacob314: no need to prop-drill uiState when it is available via useUIState.

jacob314 (commented):
Looks good! Approved once these hopefully minor comments are addressed.

packages/cli/src/ui/components/Footer.test.tsx
chrstnb: Should I also move over all of the logic for the mock UI state and settings and everything?

packages/cli/src/ui/components/Footer.test.tsx
```diff
@@ -5,15 +5,26 @@
...
+import { type UIState, UIStateContext } from '../contexts/UIStateContext.js';
+import { ConfigContext } from '../contexts/ConfigContext.js';
+import { SettingsContext } from '../contexts/SettingsContext.js';
+import { useVimMode } from '../contexts/VimModeContext.js';
+import type { LoadedSettings } from '../../config/settings.js';
 
 vi.mock('../hooks/useTerminalSize.js');
 const useTerminalSizeMock = vi.mocked(useTerminalSize.useTerminalSize);
 
+const mockedUseVimMode = useVimMode as Mock;
```
jacob314: should not be mocking useVimMode.

jacob314 (approved):
lgtm

PR #9264, opened 2025-09-23
feat(core): Add content-based retries for JSON generation

SandyTao520:
## TLDR

This pull request enhances the reliability of the `generateJson` method by introducing a content-aware retry mechanism. If the underlying model returns an empty or unparsable JSON response, the client will now automatically retry the request instead of immediately failing. This makes JSON-based interactions more robust against occasional model-side formatting issues.

## Dive Deeper

The core of this change is an update to our `retryWithBackoff` utility. It now accepts a new optional function, `shouldRetryOnContent`, which inspects the *successful* result of a function call. If this function returns `true`, the utility will treat it as a failure and trigger a retry, just as it would for a thrown error.

The `BaseLlmClient.generateJson` method leverages this new capability. It provides a `shouldRetryOnContent` function that parses the model's response. If the response is empty or not valid JSON, it signals the retry utility to try the request again. This moves the validation logic inside the retry loop, preventing transient malformed responses from causing the entire operation to fail.

Additionally, for improved clarity, the existing `shouldRetry` option in `retryWithBackoff` (which acts on errors) has been renamed to `shouldRetryOnError`.

## Reviewer Test Plan

The most effective way to validate this change is by reviewing the updated unit tests in `packages/core/src/core/baseLlmClient.test.ts`. The new tests specifically simulate empty and invalid JSON responses from the model and assert that the retry logic is engaged correctly.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ✅  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

### DIFF
--- a/packages/core/src/core/baseLlmClient.ts
+++ b/packages/core/src/core/baseLlmClient.ts
@@ -9,6 +9,7 @@ import type {
   GenerateContentConfig,
   Part,
   EmbedContentParameters,
+  GenerateContentResponse,
 } from '@google/genai';
 import type { Config } from '../config/config.js';
 import type { ContentGenerator } from './contentGenerator.js';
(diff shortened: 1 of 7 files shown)


packages/core/src/core/baseLlmClient.ts
```diff
@@ -100,52 +101,44 @@ export class BaseLlmClient {
...
+        }
+        try {
+          JSON.parse(this.cleanJsonResponse(text, model));
+          return false;
+        } catch (_e) {
+          return true;
+        }
+      };
+
+      const result = await retryWithBackoff(apiCall, {
```
galdawave, line 124: Couldn't this in theory also generate json code that is either empty or not correct? This assumes that if we get to this point we will have a valid json which might not be the case

galdawave (approved):

packages/core/src/core/baseLlmClient.ts
SandyTao520: In retryWithBackoff if all attempts to parse json failed, it would throw an error, so if we get a result we are certain the json is valid.

PR #9265, opened 2025-09-23
support standard github release archives format

jakemac53:
## TLDR

Standard GitHub release archives always have a directory inside them, and then the contents of the repo under that. We need to make sure to actually move those contents up one directory.

This was broken by https://github.com/google-gemini/gemini-cli/pull/9247, I have also now documented why this code exists.

## Reviewer Test Plan

Install an extension using normal GitHub releases with no custom archives attached.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Related to #9239

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -9,7 +9,7 @@
       "request": "launch",
       "name": "Build & Launch CLI",
       "runtimeExecutable": "npm",
-      "runtimeArgs": ["run", "build-and-start", "extensions", "install", "https://github.com/jakemac53/test-gemini-cli-extension"],
+      "runtimeArgs": ["run", "build-and-start"],
       "skipFiles": ["<node_internals>/**"],
       "cwd": "${workspaceFolder}",
       "console": "integratedTerminal",
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -9,7 +9,7 @@
       "request": "launch",
       "name": "Build & Launch CLI",
       "runtimeExecutable": "npm",
-      "runtimeArgs": ["run", "build-and-start"],
+      "runtimeArgs": ["run", "build-and-start", "extensions", "install", "https://github.com/jakemac53/test-gemini-cli-extension"],
       "skipFiles": ["<node_internals>/**"],
       "cwd": "${workspaceFolder}",
       "console": "integratedTerminal",
--- a/packages/cli/src/config/extensions/github.ts
+++ b/packages/cli/src/config/extensions/github.ts
@@ -16,7 +16,7 @@ import * as https from 'node:https';
 import * as fs from 'node:fs';
 import * as path from 'node:path';
 import { execSync } from 'node:child_process';
-import { loadExtension } from '../extension.js';
+import { EXTENSIONS_CONFIG_FILENAME, loadExtension } from '../extension.js';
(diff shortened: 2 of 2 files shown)


.vscode/launch.json
```diff
@@ -9,7 +9,7 @@
       "request": "launch",
       "name": "Build & Launch CLI",
       "runtimeExecutable": "npm",
-      "runtimeArgs": ["run", "build-and-start"],
+      "runtimeArgs": ["run", "build-and-start", "extensions", "install", "https://github.com/jakemac53/test-gemini-cli-extension"],
```
galdawave: wouldn't this force test extension to be installed into the gemini cli regardless of whether we wanted to or not ?

packages/cli/src/config/extensions/github.ts
```diff
@@ -273,6 +273,28 @@ export async function downloadFromGitHubRelease(
...
     extractFile(downloadedAssetPath, destination);
 
+    // For regular github releases, the repository is put inside of a top level
+    // directory. In this case we should see exactly two file in the destination
+    // dir, the archive and the directory. If we see that, we move all files
+    // from the directory up one level into the destination directory.
+    const entries = await fs.promises.readdir(destination, {
+      withFileTypes: true,
+    });
+    if (entries.length === 2) {
```
galdawave, line 284: I feel like this is too simple of a check, and can break if github decides to change the structure or adding/removing entries from the dest folder, also if we can we should have a test to validate this

galdawave (changes requested):

packages/cli/src/config/extensions/github.ts
jakemac53: It is a simple check that could stop working theoretically, but it also definitely won't have false positives which is I think more important. GitHub releases have been out for a long time and it would likely break a lot more than this if they changed it.

chrstnb:
@jakemac53 I wonder if there's a way to do it where we look for the gemini-extension.json file and bring that level to the top?

.vscode/launch.json
jakemac53: This was accidentally included I have reverted it

jakemac53:
> @jakemac53 I wonder if there's a way to do it where we look for the gemini-extension.json file and bring that level to the top?

We definitely could just check for one in the directory as an extra step? I don't want to expand this to support arbitrarily nested directories etc that is a feature we could definitely get stuck supporting.

packages/cli/src/config/extensions/github.ts
jakemac53: > also if we can we should have a test to validate this

I definitely agree but this code path is not tested today at all unfortunately and it would involve a pretty complicated mocking setup to get working so it's not trivial to add the first one.

We should follow up with a lot of extension related tests but I think it has to be done separately (and unfortunately after the release).

chrstnb (approved):

galdawave (approved):

PR #9270, opened 2025-09-23
Fix broken if

scidomino:
## TLDR

Fix broken condition.

See this manual run where the tests are actually executed: https://github.com/google-gemini/gemini-cli/actions/runs/17960242649/job/51081733987


mattKorwel (approved):
Thank you.

PR #9284, opened 2025-09-24
Fix so paste timeout protection is much less invasive.

### LINKED ISSUE #9282: Code to avoid pastes interleaved with line breaks is too aggressive
### What happened?

Paste in a prompt and immediately press enter and the enter is ignored.

This issue was introduced by
https://github.com/google-gemini/gemini-cli/pull/5834
which added a 500ms interval which is far too long as it is easy for a user to paste text and then immediately press enter within 500ms.

Expect to be able to press paste and then almost immediately press enter.
We should switch to an interval under 50ms to reduce the risk of breaking users who are pasting and then immediately pressing enter. Expect the extra logic to avoid improper paste handling should not trigger for trusted terminals such as the vscode terminal or terminals supporting the kitty text protocols.

### Client information

<details>

```console
> /about
# paste output here
```

</details>

### Login information

_No response_

_No response_

jacob314:
Far too large a paste timeout protection was originally used resulting in users reporting unhandled pastes.

## Reviewer Test Plan

Try to paste and then press enter immediately after pasting.

Test with the default mac terminal which does not support kitty.

Then try in a terminal that supports kitty such as iterm2 or ghostty. 
In this mode change the timeout from 30ms in InputPrompt.tsx to 200 ms and observe that now when you immediately press paste and then enter you get a line break in the input prompt.

## Linked issues / bugs
Fixes: #9282


packages/cli/src/ui/components/InputPrompt.test.tsx
```diff
@@ -1551,36 +1567,72 @@ describe('InputPrompt', () => {
...
       await wait();
 
       // Simulate a paste operation (this sets the protection)
-      stdin.write(`\x1b[200~\npasted\x1b[201~`);
-      await wait();
-
+      stdin.write('\x1b[200~pasted text\x1b[201~');
       // Wait for the protection timeout to naturally expire
-      await new Promise((resolve) => setTimeout(resolve, 600));
+      await wait(20);
```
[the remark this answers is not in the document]
allenhutchison: Seems reasonable to add this.

packages/cli/src/ui/components/InputPrompt.tsx
```diff
@@ -305,19 +323,28 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
...
+          // reasonable to expect that key events that are really part of the
+          // same paste will be processed in the same event loop tick. 30ms
+          // is chosen arbitrarily as it is faster than a typical human
+          // could go from pressing paste to pressing enter. The fastest typists
+          // can type at 200 words per minute which roughly translates to 50ms
+          // per letter.
+          pasteTimeoutRef.current = setTimeout(() => {
+            setRecentUnsafePasteTime(null);
+            pasteTimeoutRef.current = null;
+          }, 30);
```
allenhutchison: Yeah, in retrospect 500 ms was too long.

packages/cli/src/ui/components/InputPrompt.tsx
```diff
@@ -36,6 +37,18 @@ import {
...
+
+/**
+ * Returns if the terminal can be trusted to handle paste events atomically
+ * rather than potentially sending multiple paste events separated by line
+ * breaks which could trigger unintended command execution.
+ */
+export function isTerminalPasteTrusted(
+  kittyProtocolSupported: boolean,
+): boolean {
+  return kittyProtocolSupported || process.env['TERM_PROGRAM'] === 'vscode';
```
abhipatel12: nit: not a huge deal, but we should try not to have low-level logic like checking for env variables in UI components. 

I could see this sort of logic and `useKittyKeyboardProtocol` being in a more general hook. 

But not blocking.

abhipatel12: nit: `vscode` is a magic string here. Maybe we can extract to a const with a good var name.

abhipatel12 (approved):
LGTM

Just a few nits

allenhutchison (approved):

packages/cli/src/ui/components/InputPrompt.tsx
jacob314: switched to 40ms to be a little safer that we still fix the issue while still faster than the fastest typist

packages/cli/src/ui/components/InputPrompt.tsx
jacob314: turned out Cursor isn't safe due to a windows bug. simplified and removed vscode.

jacob314: turned out Cursor isn't safe due to a windows bug. simplified and removed vscode.

PR #9299, opened 2025-09-24
fix(core): Fix unable to cancel edit tool

### LINKED ISSUE #9116: Bug(tools): Tool calls cannot be cancelled - Seen most in Edit

SandyTao520:
## TLDR

Fix a bug where users are unable to cancel **Edit** or **WriteFile** (e.g., with Ctrl+C) when we're making an LLM edit correction call behind the scene.

## Dive Deeper

The core issue was we now make LLM edit correction calls in calculateEdit, and when we abort the request, we're not bubbling up the signal through layers properly. They were incorrectly treated as a regular error and swallowed (e.g. https://github.com/google-gemini/gemini-cli/blob/3667ecf1078f83a714da54aa002913c87a0e135d/packages/core/src/tools/smart-edit.ts#L492-L496).

The result is that we retry the tool call with ProceedAlways option (https://github.com/google-gemini/gemini-cli/blob/3667ecf1078f83a714da54aa002913c87a0e135d/packages/core/src/core/coreToolScheduler.ts#L736-L743).

The fix involves two parts:
1.  In `CoreToolScheduler`, the `signal.aborted` check is added into the `catch` block. This allows us to specifically detect if an error was thrown due to a user cancellation and set the status to `cancelled` accordingly.
2.  In the `edit` and `smart-edit` tools, the `catch` blocks within `shouldConfirmExecute` and `execute` are updated. This is necessary because the `calculateEdit` method now makes an abortable LLM call for edit correction. If this call is cancelled, we must re-throw the error so it propagates up to the `CoreToolScheduler` to be handled correctly, rather than being suppressed as a generic "Error preparing edit" message.

Together, these changes make the tool cancellation process more robust and provide clearer feedback to the user.

## Reviewer Test Plan

Find an complicated file with a lot of escaping characters and ask to use the replace tool, or write_file tool to make some changes, it will likely use llm to correct edits, press `esc` to cancel should succeed now.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ✅  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes https://github.com/google-gemini/gemini-cli/issues/9116

### DIFF
--- a/packages/core/src/core/coreToolScheduler.test.ts
+++ b/packages/core/src/core/coreToolScheduler.test.ts
@@ -109,6 +109,65 @@ class TestApprovalInvocation extends BaseToolInvocation<
   }
 }
 
+class AbortDuringConfirmationInvocation extends BaseToolInvocation<
+  Record<string, unknown>,
+  ToolResult
+> {
+  constructor(
+    private readonly abortController: AbortController,
+    private readonly abortError: Error,
+    params: Record<string, unknown>,
+  ) {
+    super(params);
+  }
+
+  override async shouldConfirmExecute(
+    _signal: AbortSignal,
+  ): Promise<ToolCallConfirmationDetails | false> {
+    this.abortController.abort();
+    throw this.abortError;
+  }
+
+  async execute(_abortSignal: AbortSignal): Promise<ToolResult> {
+    throw new Error('execute should not be called when confirmation fails');
+  }
+
+  getDescription(): string {
+    return 'Abort during confirmation invocation';
+  }
+}
+
+class AbortDuringConfirmationTool extends BaseDeclarativeTool<
+  Record<string, unknown>,
+  ToolResult
+> {
+  constructor(
+    private readonly abortController: AbortController,
+    private readonly abortError: Error,
+  ) {
+    super(
+      'abortDuringConfirmationTool',
+      'Abort During Confirmation Tool',
+      'A tool that aborts while confirming execution.',
+      Kind.Other,
+      {
+        type: 'object',
+        properties: {},
+      },
+    );
+  }
+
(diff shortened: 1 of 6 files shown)


lifefloating:
@/Users/gehonglu/remote-code/gemini-cli/packages/core/src/tools/ripGrep.ts convert all lowercase 'params' strings to uppercase using the  edit tool or smart-edit tool 

i use this prompt tested OK, The code looks fine; there's no problem

jacob314 (commented):
Can you please add tests for all these cases. tests should fail without your changes and then pass with the changes verifying we are handling this correctly and preventing backsliding.

jacob314 (approved):
I wasn't able to reproduce locally with a lot of trying but I was able to verify that the tests do catch regressions if the changes in the files aren't made and the tests seem like they are checking the right things.

PR #9332, opened 2025-09-24
relax JSON schema validation

geoffdowns:
## TLDR

Relax JSON schema validator behavior to better conform with JSON schema spec by setting strictSchema to false in the AJV validator constructor. Add unit tests for SchemaValidator.

## Dive Deeper

The SchemaValidator class in the core/utils package is used to verify that LLM-generated parameters for a tool call conform to the tool's expected input schema when the tool uses a JSON schema, i.e. when tool.schema.parametersJsonSchema is set. The SchemaValidator uses the AJV to perform this validation.

In order to perform this verification, SchemaValidator must first parse the tool's JSON schema. This is done via the AJV [compile](https://ajv.js.org/api.html#ajv-compile-schema-object-data-any-boolean-promise-any) function call. The `compile` call *validates* the JSON schema itself. Note that this is different from validating that the generated `params` *conform* to the schema - the latter being the primary objective of the SchemaValidator class.

The AJV validator has many configuration options. Many of these options govern how "strict" the validator should be when *compiling* a JSON schema or when validating that an object (in our case, the generated tool invocation params) conforms to the schema See: https://ajv.js.org/options.html#strict-mode-options

[strictSchema](https://ajv.js.org/options.html#strictschema) is one of these configuration options and it is [enabled by default](https://ajv.js.org/options.html#option-defaults). `strictSchema` has at least two consequences which are causing problems for GCLI users:

  1. the tool's JSON schema cannot contain any extra keywords
  2. any "format" attributes defined by the schema must have a value defined in the [ajv-formats](https://github.com/ajv-validator/ajv-formats) plugin; no custom format names may be used.

Tools with a JSON schema are mostly MCP tools. There are a wide variety of MCP server implementations available and many of these are generating/returning JSON schemas that use custom keywords and/or use custom format names. Currently these tools are unusable in GCLI, even when the GenAI SDK returns a [function call](https://googleapis.github.io/python-genai/#function-calling) indicating that the tool *should* be called, because the SchemaValidator will prevent the tool from being invoked.

## Reviewer Test Plan

1. Connect to an MCP server that provides a tool input schema that sets extra keyword fields. Ask GCLI to execute that tool. Verify the tool call succeeds. Note that you may see a warning logged about "unknown keyword" but this should not prevent the tool call or cause a failure. Example input schemas can be found in the unit tests added in this PR.

3. Connect to an MCP server that provides a tool input schema where one of the parameters uses a non-standard value for the "format" field, e.g. `format: "some-custom-value"`. Ask GCLI to execute that tool. Verify the tool call succeeds. Note that you may see a warning logged about "unknown format" but this should not prevent the tool call or cause a failure. Example input schemas can be found in the unit tests added in this PR.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | x  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

- Fixes #8537
- Fixes #7767

### DIFF
--- a/.github/actions/publish-release/action.yml
+++ b/.github/actions/publish-release/action.yml
@@ -44,11 +44,6 @@ inputs:
 runs:
   using: 'composite'
   steps:
-    - name: '📝 Print Inputs'
-      shell: 'bash'
-      run: |
-        echo "${{ toJSON(inputs) }}"
-
     - name: '👤 Configure Git User'
       working-directory: '${{ inputs.working-directory }}'
       shell: 'bash'
@@ -150,12 +145,6 @@ runs:
         dry-run: '${{ inputs.dry-run }}'
         wombat-token-core: '${{ inputs.wombat-token-core }}'
         wombat-token-cli: '${{ inputs.wombat-token-cli }}'
-
-    - name: 'Install deps'
-      working-directory: '${{ inputs.working-directory }}'
-      shell: 'bash'
-      run: |
-        npm install
 
     - name: '🎁 Bundle'
       working-directory: '${{ inputs.working-directory }}'
--- a/.github/actions/run-tests/action.yml
+++ b/.github/actions/run-tests/action.yml
@@ -2,11 +2,6 @@ name: 'Run Tests'
 description: 'Runs the preflight checks and integration tests.'
 
 inputs:
-  force_skip_tests:
-    description: 'Whether to force skip the tests.'
-    required: false
-    type: 'boolean'
-    default: false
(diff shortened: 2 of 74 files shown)


scidomino (approved):
Please open an issue to fix the issue gemini spotted (if we don't have one already).

Issue #5119, opened 2025-07-29
Poor 429 error handling


sashakosti:
### What happened?

Summary:
When a 429 error is returned from the Gemini API, gemini-cli inconsistently displays the error formatting. In some cases (e.g. on chat request), the error appears as a stringified JSON object containing escaped \n characters, leading to a messy and unreadable output. In other contexts (e.g. /compress), the same error is rendered cleanly with parsed JSON.

This appears to be a presentation issue only — the CLI still parses and responds correctly. However, the inconsistent output is misleading and noisy, especially for CLI usage or scripting.

How to reproduce:

    Exhaust your Gemini API quota (e.g. 6M input tokens/day on Free Tier).

    Run a prompt with gemini-cli that triggers a 429.

Observe the output — in some contexts, the error is displayed as:
```
    ✕ [API Error: {"error":{"message":"{\n  \"error\": {\n    \"code\": 429,\n    \"message\": \"You exceeded your current quota, please check
   your plan and billing details. For more information on this error, head to: https://ai.google.dev/gemini-api/docs/rate-limits.\",\n
  \"status\": \"RESOURCE_EXHAUSTED\",\n    \"details\": [\n      {\n        \"@type\": \"type.googleapis.com/google.rpc.QuotaFailure\",\n
         \"violations\": [\n          {\n            \"quotaMetric\":
  \"generativelanguage.googleapis.com/generate_content_free_tier_input_token_count\",\n            \"quotaId\":
  \"GenerateContentInputTokensPerModelPerDay-FreeTier\",\n            \"quotaDimensions\": {\n              \"location\": \"global\",\n
             \"model\": \"gemini-2.5-pro\"\n            },\n            \"quotaValue\": \"6000000\"\n          }\n        ]\n      },\n
     {\n        \"@type\": \"type.googleapis.com/google.rpc.Help\",\n        \"links\": [\n          {\n            \"description\":
  \"Learn more about Gemini API quotas\",\n            \"url\": \"https://ai.google.dev/gemini-api/docs/rate-limits\"\n          }\n
    ]\n      },\n      {\n        \"@type\": \"type.googleapis.com/google.rpc.RetryInfo\",\n        \"retryDelay\": \"52s\"\n      }\n
  ]\n  }\n}\n","code":429,"status":"Too Many Requests"}}]
  Please wait and try again later. To increase your limits, request a quota increase through AI Studio, or switch to another /auth method
```

In contrast, running /compress or other commands shows:
```
[...]

Issue #3510, opened 2025-07-08
Keep switching to gemini-flash even with Standard tier.


prabhu-mannu:
### What happened?

Seem like Standard tier subscription does not help with keeping switching from pro to flash, also dont have way to switch the model back in the session itself.

### What did you expect to happen?

gemini cli respect the model given 

### Client information

*   **CLI Version:** 0.1.9
*   **Git Commit:** 34935d6
*   **Operating System:** darwin v22.16.0
*   **Sandbox Environment:** sandbox-exec (permissive-open)
*   **Model Version:** gemini-2.5-flash
*   **Memory Usage:** 118.1 MB

### Login information

Google Account with standard tier subscription

Mistifiou:
You can find it [here](https://github.com/google-gemini/gemini-cli/blob/main/packages/core/src/core/client.ts#L562)

The comments indicate this happens when multiple 429 errors are received — it’s a quota issue.

Another comment states: `Only handle fallback for OAuth users`.

This means the fallback is skipped when using a Gemini API key.

This is related to #3485, #3425, and likely many other issues.

git-show:
Is this problem solved?
The previous Issue #2711 has been closed.

Mistifiou:
No, still [here](https://github.com/google-gemini/gemini-cli/blob/main/packages/core/src/utils/retry.ts#L96) . 

It say clearly "immediate fallback for OAuth users"

edits:
- related to #3752 too
- #3643 is a discussion related to this too

gavin-headen:
> No, still [here](https://github.com/google-gemini/gemini-cli/blob/main/packages/core/src/utils/retry.ts#L96) .
> 
> It say clearly "immediate fallback for OAuth users"
> 
> edits:
> 
> * related to [Still instantly falling back to flash on standard tier #3752](https://github.com/google-gemini/gemini-cli/issues/3752) too
> * [Could you please clarify who can use gemini-cli? (AKA "This account requires setting the GOOGLE_CLOUD_PROJECT env var") #3643](https://github.com/google-gemini/gemini-cli/discussions/3643) is a discussion related to this too

The line you point to is specific for when the isProQuotaExceededError method is true. This could be triggering for non-daily quotas that are more short term and just waiting a few seconds to retry might be better than an auto fallback. Maybe the error messages are specific enough to have logic that can determine this.
https://github.com/google-gemini/gemini-cli/blob/main/packages/core/src/utils/quotaErrorDetection.ts#L40

[...]

Issue #1088, opened 2025-06-16
BUG: Api Resource Limit Errors swallowed in non-interactive mode


mattKorwel:
**Describe the bug**
While working on a feature today i started getting 429's and 500's. And the api started telling me i was downgraded to flash. In interactive mode I saw the below error message. In non-interactive mode I simply stopped getting responses and it seems we were swallowing those errors.

**Additional context**

```
✕ [API Error: got status: 429 Too Many Requests. {"error":{"message":"{\n  \"error\": {\n    \"code\": 429,\n    \"message\": \"You exceeded 
  your current quota. Go to https://aistudio.google.com/apikey to upgrade your quota tier, or submit a quota increase request in 
  https://ai.google.dev/gemini-api/docs/rate-limits#request-rate-limit-increase\",\n    \"status\": \"RESOURCE_EXHAUSTED\",\n    \"details\": 
  [\n      {\n        \"@type\": \"type.googleapis.com/google.rpc.QuotaFailure\",\n        \"violations\": [\n          {\n            
  \"quotaMetric\": \"generativelanguage.googleapis.com/generate_requests_per_model_per_day\",\n            \"quotaId\": 
  \"GenerateRequestsPerDayPerProjectPerModel\"\n          }\n        ]\n      },\n      {\n        \"@type\": 
  \"type.googleapis.com/google.rpc.Help\",\n        \"links\": [\n          {\n            \"description\": \"Learn more about Gemini API 
  quotas\",\n            \"url\": \"https://ai.google.dev/gemini-api/docs/rate-limits\"\n          }\n        ]\n      }\n    ]\n  
  }\n}\n","code":429,"status":"Too Many Requests"}}]```

mattKorwel:
<img width="1145" alt="Image" src="https://github.com/user-attachments/assets/8ae5cf6c-88ba-402f-9bdc-f9c3e563a9ac" />
<img width="1137" alt="Image" src="https://github.com/user-attachments/assets/7f5f7060-20a7-4b79-a468-1a0dfa5c13fe" />

NTaylorMullen:
Filed this too: https://b.corp.google.com/issues/425238450

pomazanbohdan:
https://github.com/google-gemini/gemini-cli/issues/1502#issuecomment-3005832570

fhinkel:
Good catch, @mattKorwel , we need to make sure we're letting users in `-p` mode know.

cosmicdreams:
[...]

PR #9407, opened 2025-09-24
fix: handle request retries and model fallback correctly

### LINKED ISSUE #9039: Flash model quota exhaustion prevents fallback to available Pro model quota
### Problem Summary

Users are experiencing failures when Flash model quota is exhausted, even when Pro model quota remains available. The current fallback system is hardcoded to only fall back to Flash models, creating a situation where the system cannot utilize available Pro quota when Flash is exhausted.

### Current Behavior

1. **Model Router preferentially routes to Flash**: The `ClassifierStrategy` routes simpler requests to Flash model for efficiency
2. **Flash quota exhausts first**: Due to routing preferences, Flash quota often exhausts before Pro quota
3. **Fallback fails**: When Flash quota is exhausted, the system attempts to fall back but:
   - The fallback handler (`packages/core/src/fallback/handler.ts`) is hardcoded to use `DEFAULT_GEMINI_FLASH_MODEL`
   - If the failed model IS Flash, the handler returns `null` (no fallback)
   - Users get quota errors despite Pro model having available quota

#### 1. Hardcoded Fallback Logic
```typescript
// packages/core/src/fallback/handler.ts
export async function handleFallback(
  config: Config,
  failedModel: string,
  authType?: string,
  error?: unknown,
): Promise<string | boolean | null> {
  // ...
  const fallbackModel = DEFAULT_GEMINI_FLASH_MODEL; // Always Flash!

  if (failedModel === fallbackModel) return null; // Can't fallback from Flash
  // ...
}
```

#### 2. Binary Fallback State
```typescript
// packages/core/src/config/config.ts
private inFallbackMode = false; // Single boolean for all models
```
The config tracks fallback as a binary state rather than per-model, preventing intelligent routing between multiple models.

#### 3. Limited Error Detection
```typescript
// packages/core/src/utils/quotaErrorDetection.ts
export function isProQuotaExceededError(error: unknown): boolean { /*...*/ }
export function isGenericQuotaExceededError(error: unknown): boolean { /*...*/ }
// Missing: isFlashQuotaExceededError()
```
The system can detect Pro quota errors but not Flash-specific quota errors, making it impossible to differentiate which model's quota is exhausted.

### User Impact

- **Service interruption**: Users receive quota errors and cannot continue work despite having available quota in Pro model
- **Poor user experience**: Confusing error messages that don't explain why fallback isn't working
- **Wasted resources**: Available Pro model quota goes unutilized
- **Manual workaround required**: Users must manually switch models or change authentication

### Reproduction Steps

1. Use OAuth authentication (LOGIN_WITH_GOOGLE)
2. Make requests that get routed to Flash model (simpler tasks)
3. Continue until Flash daily quota is exhausted
4. Observe that subsequent requests fail with quota errors
5. Manually check that Pro model still has available quota

### Proposed Solution (Immediate Fix)

> [!NOTE]
> This solution is a suggestion from some initial investigation conducted by a LLM. It is not a set design, but rather a possible solution to address this issue.

Implement model-specific quota tracking and intelligent fallback routing:

#### 1. Add Flash Quota Detection
```typescript
// packages/core/src/utils/quotaErrorDetection.ts
export function isFlashQuotaExceededError(error: unknown): boolean {
  // Detect Flash-specific quota exhaustion
}

export function extractModelFromQuotaError(error: unknown): string | null {
  // Extract which model's quota is exhausted from error message
}
```

#### 2. Track Per-Model Exhaustion
```typescript
// packages/core/src/config/config.ts
private exhaustedModels = new Set<string>(); // Track each model separately

setModelExhausted(model: string): void { /*...*/ }
isModelExhausted(model: string): boolean { /*...*/ }
```

#### 3. Implement Intelligent Fallback Strategy
```typescript
// packages/core/src/routing/strategies/fallbackStrategy.ts
private selectFallbackModel(exhaustedModel: string, config: Config): string | null {
  // Define fallback chains
  const fallbackChains = {
    'gemini-2.5-flash': [
      'gemini-2.5-pro',        // Try Pro first when Flash exhausted
      'gemini-2.5-flash-lite'
    ],
    'gemini-2.5-pro': [
      'gemini-2.5-flash',
      'gemini-2.5-flash-lite'
    ]
  };

  // Find first non-exhausted model in chain
  for (const fallback of fallbackChains[exhaustedModel]) {
    if (!config.isModelExhausted(fallback)) {
      return fallback;
    }
  }
  return null;
}
```

### Acceptance Criteria

- [ ] Flash quota exhaustion correctly detected from error messages
- [ ] System can fall back from Flash to Pro when Flash quota exhausted
- [ ] System can fall back from Pro to Flash when Pro quota exhausted
- [ ] Model-specific exhaustion tracked independently
- [ ] Fallback strategy respects model availability
- [ ] All existing tests pass
- [ ] New tests cover fallback scenarios

gsquared94:
## TLDR

This pull request refactors the handling of API quota errors to be more robust and reliable. Previously, we relied on string matching to detect specific quota error messages. This change introduces a structured error  parsing system that can correctly classify quota errors as either retryable (e.g., per-minute limits, small retry delays) or terminal (e.g., daily limits).

The retry logic has been updated to use this new classification, ensuring that we only trigger model fallbacks for terminal quota errors and use the server-suggested delay for retryable errors when available. 

Also increased the max number of retries from 5 to 10 (found a lot more 503 errors from Gemini API succeeding within that).

> [!NOTE] 
> Does not fix #9039. But with fewer incorrect model fallback attempts (which for auto model routing was previously terminating requests started with `flash`) it should help reduce the occurrence.

##  Dive Deeper

The core of this change is a new error classification system:

   1. `parseGoogleApiError`: A new function in packages/core/src/errors.ts that can parse deeply nested and stringified Google API error responses, extracting the structured error details.
   2. `TerminalQuotaError` and `RetryableQuotaError`: New custom error classes to represent different types of quota issues.
   3. `classifyGoogleError`: A function that uses parseGoogleApiError to analyze 429 errors and wrap them in the appropriate custom error class.

The retryWithBackoff function has been refactored to:
   - Catch `TerminalQuotaError` to trigger the `onPersistent429` fallback mechanism.
   - Catch `RetryableQuotaError` and use the `retryDelayMs` from the error for a more accurate backoff period.

  The UI hook `useQuotaAndFallback` has also been updated to use `instanceof TerminalQuotaError` for its logic, simplifying the code and removing the need for the old isProQuotaExceededError and isGenericQuotaExceededError functions, which have been deleted.

### Using GCA auth

- Set `GOOGLE_CLOUD_PROJECT=aipp-internal-testing (or whatever)` and run `gemini-cli` from this branch, login with oauth. You should be able to make a lot of queries before seeing automatic fallback error.
- <details> 

   <summary>Run this script for a while to exhaust the pro model quota.</summary>

    ```
     #!/bin/bash
    
    for i in {1..2000}
    do
      echo "Request #$i"
      curl -i -sN -X POST \
      -H "Authorization: Bearer $(gcloud auth print-access-token)" \
      -H 'x-goog-user-project: aipp-internal-testing' \
      --json '{
      "project":"aipp-internal-testing",
      "model":"gemini-2.5-pro",
      "request":{
        "contents": {
          "role": "user",
          "parts": [
            {
              "text": "Hi"
            },
          ]
        }
      }
    }' \
    https://cloudcode-pa.googleapis.com/v1internal:streamGenerateContent
    done

    echo "Done."
     ```

    </details> 

- Issue a request on gemini-cli, you should see the model fallback error.

### Using Gemini API key

- Restrict quota manually to force 429 errors. Go to [this](https://pantheon.corp.google.com/apis/api/generativelanguage.googleapis.com/quotas?pageState=(%22allQuotasTable%22:(%22p%22:0,%22f%22:%22%255B%257B_22k_22_3A_22Has%2520override_22_2C_22t_22_3A11_2C_22v_22_3A_22_5C_22True_5C_22_22_2C_22s_22_3Atrue_2C_22i_22_3A_22hasConsumerOverride_22%257D_2C%257B_22k_22_3A_22Name_22_2C_22t_22_3A10_2C_22v_22_3A_22_5C_22GenerateContent%2520input%2520token%2520count%2520limit%2520per%2520model_5C_22_22_2C_22i_22_3A_22displayName_22%257D_2C%257B_22k_22_3A_22Dimensions%2520%2528e.g.%2520location%2529_22_2C_22t_22_3A10_2C_22v_22_3A_22_5C_22model_3Agemini-2.5-pro_5C_22_22_2C_22s_22_3Atrue_2C_22i_22_3A_22displayDimensions_22%257D%255D%22,%22c%22:%5B%22metricName%22,%22limitName%22,%22monitoredResource%22%5D))) link and select the project for which you've created the Gemini API key.
- Override the value to be something low like 50,000.
   image
- Run `gemini-cli` from this branch, use this Gemini API key for auth, and start making requests.
- Requests should appear to be taking a lot longer (due to retries from the 429 error code `retryInfo` objects) instead of failing.
> [!NOTE] 
> Right now even retryable errors are run for a max of 10 attempts (same as other error codes). We can consider retrying these errors for a longer number than other error codes.

##  Testing Matrix

  |          | 🍏  | 🪟  | 🐧  |
  | npx      |  ❓  | ❓   | ❓   |

Fixes:
- #9248
- #5119
- #3510
- #1088
- #4186


packages/core/src/utils/googleErrors.ts
```diff
@@ -0,0 +1,242 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * @fileoverview
+ * This file contains types and functions for parsing structured Google API errors.
+ */
```
scidomino, line 10: We already have a errors.ts in cli/src/utils and core/src/utils. Is there some way to avoid introducing a third?

packages/core/src/utils/googleQuotaErrors.ts
```diff
@@ -0,0 +1,162 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
```
scidomino, line 4: Could we move some of the into the exist errors.ts files?

packages/core/src/utils/googleErrors.ts
gsquared94: moved to be under `core/src/utils`

packages/core/src/utils/googleQuotaErrors.ts
gsquared94: moved to the same folder, keeping in separate files since there is no duplication and it's easier to reason about and test.

scidomino (approved):

PR #9410, opened 2025-09-24
fix(acp): prevent unnecessary credential cache clearing on re-authent…

### LINKED ISSUE #7549: ACP authentication doesn't use cached creds
### What happened?

When I run `gemini --experimental-acp`, it doesn't seem to reuse the saved OAuth credentials, and asks for new ones instead.

This is mildly annoying without the sandbox, but with sandbox enabled, it actually expects the authorization code on stdin, despite stdin being used for the JSON-RPC communication of the Agent Client Protocol. Unsurprisingly, ACP clients do not expect to have to provide the code via stdin…

To reproduce:
1. `$ gemini` - complete the auth
1. `$ gemini` - check that it doesn't ask for auth again
1. `$ gemini --experimental-acp`
1. send `{ "jsonrpc": "2.0", "id": 0, "method": "initialize", "params": { "protocolVersion": 1, "clientCapabilities": { "fs": { "readTextFile": true, "writeTextFile": true } } } }` to its stdin
1. send `{ "jsonrpc": "2.0", "id": 0, "method": "authenticate", "params": { "methodId": "oauth-personal" } }` to its stdin

That results in "Code Assist login required. Attempting to open authentication page in your browser. Otherwise navigate to: […]" even though `$ gemini` without `--experimental-acp` would just re-use the saved credentials.

The sandbox alternative is even worse:
1. `GEMINI_SANDBOX=podman SANDBOX_FLAGS='--userns=keep-id' gemini --experimental-acp`
1. send `{ "jsonrpc": "2.0", "id": 0, "method": "initialize", "params": { "protocolVersion": 1, "clientCapabilities": { "fs": { "readTextFile": true, "writeTextFile": true } } } }` to its stdin
1. send `{ "jsonrpc": "2.0", "id": 0, "method": "authenticate", "params": { "methodId": "oauth-personal" } }` to its stdin

That results in "Please visit the following URL to authorize the application: […]" but no ACP client expects that!

Also, if `~/.gemini/oauth_creds.json` doesn't exist, the sandbox variant will ask for the auth twice. Once before launching the container (interactively, "Attempting to open authentication page in your browser."), and then again in response to the "authenticate" ACP method ("Please visit the following URL to authorize the application:").

(already covered above)

### Client information

<details>

```console
> /about
╭───────────────────────────────────────────────────╮
│                                                   │
│ About Gemini CLI                                  │
│                                                   │
│ CLI Version      0.2.2                            │
│ Git Commit       7598ef3                          │
│ Model            gemini-2.5-pro                   │
│ Sandbox          sandbox-0.2.2-0                  │
│ OS               linux                            │
│ Auth Method      OAuth                            │
│                                                   │
╰───────────────────────────────────────────────────╯
```

</details>

### Login information

_No response_

_No response_

h-michael:
## TLDR

This PR fixes ACP mode unnecessarily clearing cached authentication credentials on every `authenticate()` call. The fix preserves credentials when using the same auth method, only
clearing when switching between different auth types. This is particularly important for OAuth users who previously had to re-authenticate via browser redirects on every session.

### Problem Analysis
The issue stemmed from the ACP authentication flow always calling `clearCachedCredentialFile()` regardless of whether the auth method changed. This caused:
- **OAuth users**: Forced browser redirects for every new session (major UX disruption)
- **API Key users**: Unnecessary credential re-reads (minor performance impact)
- **Vertex AI users**: Unnecessary config reloads (minor performance impact)

### Reviewer Test Plan

Prerequisites

1. Have an ACP-compatible editor (e.g., Zed, Neovim with codecompanion.nvim, or other ACP clients)
2. Set up OAuth authentication (LOGIN_WITH_GOOGLE)

Test Scenarios

1. OAuth Credential Preservation (Primary Test)
# Start ACP mode with OAuth
gemini --experimental-acp

# In your ACP-compatible editor:
1. Open a project
2. Trigger gemini-cli integration
3. Authenticate with OAuth (browser opens)
4. Close the session
5. Start a new session
6. ✅ Verify: No browser redirect, uses cached credentials

### Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅   | ❓   | ✅   |

### Linked issues / bugs

Fixes #7549 - ACP authentication doesn't use cached credentials


packages/cli/src/zed-integration/zedIntegration.ts
```diff
@@ -115,8 +115,16 @@ class GeminiAgent {
 
   async authenticate({ methodId }: acp.AuthenticateRequest): Promise<void> {
     const method = z.nativeEnum(AuthType).parse(methodId);
+    const selectedAuthType = this.settings.merged.security?.auth?.selectedType;
```
scidomino, line 116: This code only affects the zed integration. Is that the only place where the bug is?

h-michael: @scidomino  Thanks for reviewing! Just to clarify - are you asking if the same bug exists elsewhere, or why the other call sites don't need the same fix?

Assuming the latter: The other call sites work correctly:

[`AppContainer.tsx`](https://github.com/google-gemini/gemini-cli/blob/1689e9b6717711935ea9722eff4229f8fd574d37/packages/cli/src/ui/AppContainer.tsx#L448) / [`AuthDialog.tsx`](https://github.com/google-gemini/gemini-cli/blob/1689e9b6717711935ea9722eff4229f8fd574d37/packages/cli/src/ui/auth/AuthDialog.tsx#L122): Only called when user runs `/auth` command → clearing cache is correct (user wants to change auth)

[`zedIntegration.ts#L122`](https://github.com/google-gemini/gemini-cli/blob/21215b7c4a4fb211cd6a70eeb9b2afda5d522592/packages/cli/src/zed-integration/zedIntegration.ts#L122): Called on every editor reconnection → this caused the issue in #7549

Let me know if that answers your question!

scidomino: sounds good!

scidomino (approved):

PR #9497, opened 2025-09-24
fix(integration): Added shell specification for winpty

davideast:
## TLDR

The winpty agent needs to launch a console-based shell (like cmd.exe or PowerShell) to host the child process (node gemini.js ... in this case). By explicitly setting the shell option, the test helper configures node-pty to correctly spawn the interactive process within a managed terminal environment on Windows, allowing the tests to send input and read output.

While not ideal, the windows test detects the `Press Ctrl+C again to exit` output and kills the process. This is a current workaround winpty not handling this use case reliably. I'll investigate the source of this in a follow up.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ✅   |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ✅   |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

### DIFF
--- a/integration-tests/test-helper.ts
+++ b/integration-tests/test-helper.ts
@@ -833,7 +833,8 @@ export class TestRig {
       options.shell = process.env.COMSPEC || 'cmd.exe';
     }
 
-    const ptyProcess = pty.spawn(command, commandArgs, options);
+    const executable = command === 'node' ? process.execPath : command;
+    const ptyProcess = pty.spawn(executable, commandArgs, options);
 
     ptyProcess.onData((data) => {
       this._interactiveOutput += data;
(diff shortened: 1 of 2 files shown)


integration-tests/test-helper.ts
```diff
@@ -813,16 +814,27 @@ export class TestRig {
...
-      env: process.env as { [key: string]: string },
-    });
+      env: Object.fromEntries(
+        Object.entries(process.env).filter(([, v]) => v !== undefined),
+      ) as { [key: string]: string },
+    };
+
+    if (isWindows) {
+      // node-pty on Windows requires a shell to be specified when using winpty.
+      options.shell = process.env.COMSPEC || 'cmd.exe';
```
jacob314, line 833: nit: we probably want to run some subset of tests with both powershell and cmd.exe

jacob314 (commented):
We've added other tests that skip windows for similar reasons. Can you also update them to no longer skip windows?lgtm

jacob314 (approved):
lgtm

PR #9508, opened 2025-09-24
feat(ui): Render UI to almost 100% of terminal width

WangWanyue:
Fixes #4671

This change adjusts the UI layout calculations to utilize the full terminal width, removing the previous 90% constraint.
- The main application container now expands to fill the available horizontal space, with a small reserve for a scrollbar.
- Input prompt and suggestion widths are now calculated based on the full terminal width.
- Horizontal padding has been adjusted to accommodate the wider layout.

## Screenshots
image

## Dive Deeper
Ink has layout quirks at the terminal edges, use 100% width leads to more problems than the extra 1-2 characters of width are worth. So I used a simple `terminalWidth - 4` across board to get ~96% width usage (much better than 90%) while avoiding the rendering issues entirely. (Tried other approaches e.g. full width by default and calculate width for specific components, but it caused more responsiveness issues.)  

## Reviewer Test Plan
* Run a simple request to test the width
* Resize window to test responsiveness

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

### DIFF
--- a/packages/cli/src/ui/App.tsx
+++ b/packages/cli/src/ui/App.tsx
@@ -23,7 +23,7 @@ export const App = () => {
 
   return (
     <StreamingContext.Provider value={uiState.streamingState}>
-      <Box flexDirection="column" width="90%">
+      <Box flexDirection="column" width={uiState.mainAreaWidth}>
         <MainContent />
 
         <Box flexDirection="column" ref={uiState.mainControlsRef}>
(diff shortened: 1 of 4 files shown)


WangWanyue:
@jacob314 this is ready for review.

jacob314:
This is an improvement but I'm seeing cases that are still mysteriously less than full width and inconsistent with the tool messages.
Once tool groups are approved they are full width but before they are confirmed the aren't. 
Similarly the input prompt is less than full width.
Screenshot 2025-09-24 at 2 37 59 PM

Screenshot 2025-09-24 at 2 37 59 PM

we need some integration level snapshot tests to verify that we can land this safely. 
A change like this should break a bunch of snapshot tests rather than passing all tests. stay tuned as we'll work on landing some integration level snapshot tests and then those can be used to help validate this change across platforms.

The other change we need to validate is that we don't get significantly more ink rendering corruption with this change.

jacob314:
Thanks for the PR. I've landed a version of this PR.

Issue #9030, opened 2025-09-21
Improve performance of edit tool


anj-s:
The edit tool has been seeing the following issues:
1. Same change error
2. Looping when its unable to edit a given file
3. Higher latency to respond to the user in certain scenarios
<and more>

This is the master bug to track all ongoing edit issues. Note: The latest edit changes have been in release for a week so we need to make sure we are fixing the right tool.

agmsb:
Tracking work at go/agmsb-edit.

agmsb:
Removing #8114 and #8409, as one is on a non-smart edit version of Replace and the other is referencing the shell tool.

tyssejc:
Hello — I was having the exact issue with the `replace` tool documented in #1028 . I see this was closed as a duplicate, so I'm following up here.

I just ran into a similar scenario in Antigravity:

```
I apologize, I mistakenly included ```typescript markers in the previous replace_file_content call. I will correct that now by removing them.

I made the same mistake again. The tool name is replace_file_content but I used it incorrectly. I will call it correctly now.

It seems I am still having trouble with the replace_file_content tool call, even after removing the It seems I am still having trouble with thereplace_file_contenttool call, even after removing the It seems I am still having trouble with the replace_file_content tool call, even after removing the It seems I am still having trouble with thereplace_file_contenttool call, even after removing the It seems I am still having trouble with the replace_file_content tool call, even after removing the It seems I am still having trouble with thereplace_file_contenttool call, even after removing the It seems I am still having trouble with the replace_file_content tool call, even after removing the It seems I am still having trouble with thereplace_file_contenttool call, even after removing the It seems I am still having trouble with the replace_file_content tool call, even after removing the It seems I am still having trouble with thereplace_file_contenttool call, even after removing the It seems I am still having trouble with the replace_file_content tool call, even after removing the It seems I am still having trouble with thereplace_file_contenttool call, even after removing the It seems I am still having trouble with the replace_file_content tool call, even after removing the It seems I am still having trouble with thereplace_file_contenttool call, even after removing the It seems I am still having trouble with the replace_file_content tool call, even after removing the It seems I am still having trouble with thereplace_file_contenttool call, even after removing the It seems I am still having trouble with the replace_file_content tool call, even after removing the It seems I am still having trouble with thereplace_file_content` tool call, even after removing the ```. I will try again, ensuring the arguments are correct.

[...]

Issue #4137, opened 2025-07-14
Evaluate quality of edit tool and improve it


anj-s:
Edit tool is one of the most critical pieces and invariably there are issues with editing files (especially large files).

Evaluate the quality of the current edit tool , track issues and propose solutions

anj-s:
Other related issues: b/427299140

J8-8N:
I'm surprised by the lack of traction for this edit tool failure continuous loop because it happens to us on nearly every session (despite our files being 400 lines long max) and it's the only reason we have to keep using a competitor.

jnui:
Is there any workaround to this issue?
perhaps using a MCP for file handling, I have multiple issues for file handling for reading and writing

<img width="875" height="435" alt="Image" src="https://github.com/user-attachments/assets/ba95d77b-482a-4604-ace9-decdc5659365" />
<img width="866" height="367" alt="Image" src="https://github.com/user-attachments/assets/9d7d252f-f9e2-4e20-af09-a06292a83263" />

jnui:
note, I thought this problem stopped for me after I added a mcp server for file handling.
but after a while the file handling errors came back.

silviojr:
Hello, folks! Thanks for reporting all those errors. I am going through them and working actively on improving it. 

I added some sub-tasks that I will follow. 

1)  First I Need to fix our logging issues to actually track the errors we are not seeing currently (issue: [#61](https://github.com/google-gemini/maintainers-gemini-cli/issues/61)) 
2) Then I will measure the rate or success/failure and most common failures (issue 
#5209 ). I will report this. 
3) I will evaluate how much we can improve those metrics with few improvements, trying to be quick to release this soon (issues [5211](https://github.com/google-gemini/gemini-cli/issues/5211) and [5213](https://github.com/google-gemini/gemini-cli/issues/5213)) 
4) I will continue to explore other approaches (issue [5214](https://github.com/google-gemini/gemini-cli/issues/5214) ) . . **Please share any other ideas** or errors you are seeing!! :)

anicolao:
@srithreepo you sure about that? I am actively trying to figure out how to replace Gemini CLI because of this issue.

hongkongkiwi:
[...]

PR #9518, opened 2025-09-24
fix(core): plumb max attempts for retry to generate options in baseLLMClient

agmsb:
## TLDR

This PR plumbs maxAttempts into GenerateJsonOptions, enabling to pass in maxAttempts to GenerateJson so that we can call retryWithBackoff with a specified amount of attempts. It also updates the retry code to keep the default at 5 if maxAttempts does not get passed in, maintaining current behavior for those not passing in maxAttempts to baseLLMClient.

## Dive Deeper

This prevents the llm-edit-fixer from hanging on multiple retries to call an LLM, which provides a poor UX when retrying up to the default 5 times.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | x  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Makes progress on #4137, #9030

### DIFF
--- a/packages/core/src/utils/retry.ts
+++ b/packages/core/src/utils/retry.ts
@@ -84,6 +84,8 @@ export async function retryWithBackoff<T>(
   } = {
     ...DEFAULT_RETRY_OPTIONS,
     ...options,
+    // Explicitly use default if maxAttempts is null or undefined in options.
+    maxAttempts: options?.maxAttempts ?? DEFAULT_RETRY_OPTIONS.maxAttempts,
   };
 
   let attempt = 0;
--- a/packages/core/src/utils/retry.ts
+++ b/packages/core/src/utils/retry.ts
@@ -74,6 +74,10 @@ export async function retryWithBackoff<T>(
   fn: () => Promise<T>,
   options?: Partial<RetryOptions>,
 ): Promise<T> {
+  if (options?.maxAttempts !== undefined && options.maxAttempts <= 0) {
+    throw new Error('maxAttempts must be a positive number.');
+  }
+
   const {
     maxAttempts,
     initialDelayMs,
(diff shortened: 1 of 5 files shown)


bdmorgan (approved):

packages/core/src/utils/retry.ts
```diff
@@ -84,6 +84,8 @@ export async function retryWithBackoff<T>(
   } = {
     ...DEFAULT_RETRY_OPTIONS,
     ...options,
+    // Explicitly use default if maxAttempts is null or undefined in options.
+    maxAttempts: options?.maxAttempts ?? DEFAULT_RETRY_OPTIONS.maxAttempts,
```
silviojr: is it worth raising an error if someone sets up maxAttempts to zero?

agmsb: Good call. Adding this.

reconsumeralization: The Problem: The llm-edit-fixer was causing a bad user experience by hanging for a long time when an LLM call failed. The current default retry behavior was too aggressive.

The Fix: The PR addresses this by allowing developers to set maxAttempts, so they can control how many times the system retries a failed LLM call. This makes the application more responsive and predictable.

The Relation to Our Discussion: While the PR isn't a security fix itself, it highlights a crucial point: LLM integration is not a simple, foolproof process. It involves complex error handling, and this retry logic could have security implications. For example, a malicious extension trying to trigger an exploit might fail the first time, but the new retry mechanism could give it multiple chances to succeed, potentially making a low-probability attack more reliable.

reconsumeralization:
I this for the thing I said?

PR #9526, opened 2025-09-24
fix(core): auto-correct file paths in smart edit where possible

agmsb:
## TLDR

This PR introduces a check during validateToolParamValues to auto-correct unambiguous relative paths into absolute paths during file editing.

## Dive Deeper

This can help reduce smart edit tool errors, adding functionality to auto-correct an unambiguous relative path (this means a case where there is only one file with the given file name or partial file path) to an absolute path, making the tool call valid.

Should the file be unambiguous (IE more than one file with that name exists, or it is an invalid partial relative path) we throw an error. 

## Reviewer Test Plan

N/A

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | x  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Progress on #4137.


packages/core/src/tools/smart-edit.ts
```diff
@@ -767,11 +767,41 @@ A good instruction should concisely answer:
...
+        const foundFiles = fileSystem.findFiles(params.file_path, searchPaths);
+
+        if (foundFiles.length === 1) {
+          // It's unambiguous, so we can correct it.
+          params.file_path = foundFiles[0];
+        } else if (foundFiles.length > 1) {
+          return (
+            `The file path '${params.file_path}' is ambiguous and matches multiple files. ` +
+            `Please provide an absolute path to one of the following: ${foundFiles.join(', ')}`
+          );
```
galdawave: When multiple files match a relative path, what if we try to find the best fit instead of erroring?

For example, if the user provides `services/auth.t`  and we find `both /projectA/services/auth.ts` and 
`/projectB/api/services/auth.ts`, we could potentially choose the first one because it's a more direct match
 to the provided relative path. We could create a heuristic that prefers the path with the most direct 
structural match.

agmsb: PTAL, I've taken a first pass at implementing this heuristic.

packages/core/src/tools/smart-edit.ts
```diff
@@ -755,6 +755,53 @@ A good instruction should concisely answer:
...
+   * @returns An error message string if correction fails, otherwise null.
+   */
+  private correctPath(params: EditToolParams): string | null {
+    const directPath = this.findDirectPath(params.file_path);
+    if (directPath) {
+      params.file_path = directPath;
+      return null;
+    }
+
+    const foundFiles = this.findAmbiguousPaths(params.file_path);
```
mrcabbage972, line 793: what about the case of foundFiles.length === 1?

agmsb: Updated to handle this.

mrcabbage972: Where? I don't see the change

agmsb: Woops wrong branch. Done now.

mrcabbage972 (approved):

PR #9530, opened 2025-09-24
feat(cli) Add mime-type validation to MCP tool responses (#9511)

Paul-Deem:
## TLDR

Add input validation from MCP tool responses and file reads to block mime-types that aren't supported by gemini.

## Dive Deeper

Handing an unsupported mime-type to the gemini api results in a 400 error for unsupported mime-type. Blocking these results before pushing them to the model prevents the agent from crashing when receiving unsupported content types. 

### Before: 
image

### After:
image

### Verifying File sanitization: 
Download an unsupported mimetype, image/tiff is a reasonable example. Ask gemini to read it `Can you read this file? @file_example_TIFF_1MB.tiff` It should result in an API failure without this change and a tool failure with it. 

### Verifying MCP sanitization:
You need an MCP server you expect to return invalid types. You can checkout and install this experimental GCS MCP server at [this commit](https://github.com/googleapis/gcloud-mcp/tree/ba993b78e96a1bd711773b015716bee610aeabdd). You will need a GCP project and an associated GCS Bucket. You can upload any unsupported types (spreadsheet, executable, etc) and ask gemini to read and summarize the object context. 

Without this change you will get a 400 error and the agent session will return to the turn to the user immediately. 

With this change it the model will instead return that it can't process that type of data. 

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅   | ❓  | ❓  |
| npx      | ✅   | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

- Closes #9511

### DIFF
--- a/packages/core/src/tools/mcp-tool.ts
+++ b/packages/core/src/tools/mcp-tool.ts
@@ -18,6 +18,7 @@ import {
   ToolConfirmationOutcome,
 } from './tools.js';
 import type { CallableTool, FunctionCall, Part } from '@google/genai';
+import { isMimeTypeSupported } from '../utils/mimeUtils.js';
 import { ToolErrorType } from './tool-error.js';
 import type { Config } from '../config/config.js';
 
@@ -56,37 +57,6 @@ type McpContentBlock =
   | McpMediaBlock
   | McpResourceBlock
   | McpResourceLinkBlock;
-
-// This list is based on the supported file formats for Gemini 2.5 Pro.
-// Source: https://cloud.google.com/vertex-ai/generative-ai/docs/models/gemini/2-5-pro
-//         https://cloud.google.com/vertex-ai/generative-ai/docs/models/gemini/2-5-flash
-const SUPPORTED_MIME_TYPES = new Set([
-  'image/png',
-  'image/jpeg',
-  'image/webp',
-  'application/pdf',
-  'text/plain',
-  'video/x-flv',
-  'video/quicktime',
-  'video/mpeg',
-  'video/mpegs',
-  'video/mpg',
-  'video/mp4',
-  'video/webm',
-  'video/wmv',
-  'video/3gpp',
-  'audio/x-aac',
-  'audio/flac',
-  'audio/mp3',
-  'audio/m4a',
-  'audio/mpeg',
-  'audio/mpga',
-  'audio/mp4',
-  'audio/opus',
-  'audio/pcm',
-  'audio/wav',
-  'audio/webm',
-]);
 
 class DiscoveredMCPToolInvocation extends BaseToolInvocation<
   ToolParams,
@@ -218,7 +188,7 @@ class DiscoveredMCPToolInvocation extends BaseToolInvocation<
       };
     }
 
-    validateMcpContent(rawResponseParts);
+    sanitizeMcpContentInPlace(rawResponseParts);
 
     const transformedParts = transformMcpContentToParts(rawResponseParts);
 
@@ -341,32 +311,35 @@ function transformResourceLinkBlock(block: McpResourceLinkBlock): Part {
 }
 
 /**
- * Validates the MCP content blocks and replaces unsupported mime types with an error message.
+ * Sanitizes the MCP content blocks in place, replacing unsupported mime types
+ * with an error message.
  * @param sdkResponse The raw Part[] array from `mcpTool.callTool()`.
  */
-function validateMcpContent(sdkResponse: Part[]): void {
-  const funcResponse = sdkResponse?.[0]?.functionResponse;
-  const mcpContent = funcResponse?.response?.['content'] as McpContentBlock[];
+function sanitizeMcpContentInPlace(sdkResponse: Part[]): void {
+  for (let j = 0; j < sdkResponse.length; j++) {
+    const funcResponse = sdkResponse[j]?.functionResponse;
+    const mcpContent = funcResponse?.response?.['content'] as McpContentBlock[];
 
-  if (!Array.isArray(mcpContent)) {
-    return;
-  }
-
-  for (let i = 0; i < mcpContent.length; i++) {
-    const block = mcpContent[i];
-    let mimeType: string | undefined;
-
-    if (block.type === 'image' || block.type === 'audio') {
-      mimeType = block.mimeType;
-    } else if (block.type === 'resource' && block.resource.blob) {
-      mimeType = block.resource.mimeType || 'application/octet-stream';
+    if (!Array.isArray(mcpContent)) {
+      return;
     }
 
-    if (mimeType && !SUPPORTED_MIME_TYPES.has(mimeType)) {
-      mcpContent[i] = {
-        type: 'text',
-        text: `[Tool returned unsupported mimetype: ${mimeType}]`,
-      };
+    for (let i = 0; i < mcpContent.length; i++) {
+      const block = mcpContent[i];
+      let mimeType: string | undefined;
+
+      if (block.type === 'image' || block.type === 'audio') {
+        mimeType = block.mimeType;
+      } else if (block.type === 'resource' && block.resource.blob) {
+        mimeType = block.resource.mimeType || 'application/octet-stream';
+      }
+
+      if (mimeType && !isMimeTypeSupported(mimeType)) {
+        mcpContent[i] = {
+          type: 'text',
+          text: `[Tool returned unsupported mimetype: ${mimeType}]`,
+        };
+      }
     }
   }
 }
--- a/packages/core/src/tools/mcp-tool.ts
+++ b/packages/core/src/tools/mcp-tool.ts
@@ -57,6 +57,36 @@ type McpContentBlock =
   | McpResourceBlock
   | McpResourceLinkBlock;
 
+// This list is based on the supported file formats for Gemini 2.5 Pro.
+// Source: https://cloud.google.com/vertex-ai/generative-ai/docs/models/gemini/2-5-pro
+const SUPPORTED_MIME_TYPES = new Set([
+  'image/png',
+  'image/jpeg',
+  'image/webp',
+  'application/pdf',
+  'text/plain',
+  'video/x-flv',
+  'video/quicktime',
+  'video/mpeg',
+  'video/mpegs',
+  'video/mpg',
+  'video/mp4',
+  'video/webm',
+  'video/wmv',
+  'video/3gpp',
+  'audio/x-aac',
+  'audio/flac',
+  'audio/mp3',
+  'audio/m4a',
+  'audio/mpeg',
+  'audio/mpga',
+  'audio/mp4',
+  'audio/opus',
+  'audio/pcm',
+  'audio/wav',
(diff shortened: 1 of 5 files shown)


packages/core/src/tools/mcp-tool.ts
```diff
@@ -57,6 +57,36 @@ type McpContentBlock =
   | McpResourceBlock
   | McpResourceLinkBlock;
 
+// This list is based on the supported file formats for Gemini 2.5 Pro.
+// Source: https://cloud.google.com/vertex-ai/generative-ai/docs/models/gemini/2-5-pro
```
chrstnb: It looks like these also match 2.5 Flash, which was the other thing I wanted to check

Paul-Deem: Yea I can update the comment that it matches the general 2.5 family of models.

packages/core/src/tools/mcp-tool.ts
```diff
@@ -57,6 +57,36 @@ type McpContentBlock =
...
+const SUPPORTED_MIME_TYPES = new Set([
```
jacob314: please refactor so that this logic is in fileUtils.ts which already has a 
detectFileType with similar logic. 
We need to have one source of truth of what file extensions we support. Using the same util everywhere will also give us earlier user feedback if something is wrong as more users trigger fileUtils.ts as part of writing `@filePath` in their prompt than use mcp.

packages/core/src/tools/mcp-tool.ts
```diff
@@ -307,6 +339,37 @@ function transformResourceLinkBlock(block: McpResourceLinkBlock): Part {
...
+    }
+
+    if (mimeType && !SUPPORTED_MIME_TYPES.has(mimeType)) {
+      mcpContent[i] = {
+        type: 'text',
+        text: `[Tool returned unsupported mimetype: ${mimeType}]`,
+      };
+    }
+  }
+}
```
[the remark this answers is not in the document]
jakemac53: +1 I think this is a good suggestion

packages/core/src/tools/mcp-tool.ts
```diff
@@ -307,6 +339,37 @@ function transformResourceLinkBlock(block: McpResourceLinkBlock): Part {
...
 
+/**
+ * Validates the MCP content blocks and replaces unsupported mime types with an error message.
+ * @param sdkResponse The raw Part[] array from `mcpTool.callTool()`.
+ */
+function validateMcpContent(sdkResponse: Part[]): void {
+  const funcResponse = sdkResponse?.[0]?.functionResponse;
+  const mcpContent = funcResponse?.response?.['content'] as McpContentBlock[];
+
+  if (!Array.isArray(mcpContent)) {
```
jakemac53: I don't know that this is really necessary, it is already supposed to be an array right? We could throw a more explicit error but I don't think doing nothing with it is really correct.

packages/core/src/tools/mcp-tool.ts
```diff
@@ -307,6 +339,37 @@ function transformResourceLinkBlock(block: McpResourceLinkBlock): Part {
   };
 }
 
+/**
+ * Validates the MCP content blocks and replaces unsupported mime types with an error message.
+ * @param sdkResponse The raw Part[] array from `mcpTool.callTool()`.
+ */
+function validateMcpContent(sdkResponse: Part[]): void {
+  const funcResponse = sdkResponse?.[0]?.functionResponse;
+  const mcpContent = funcResponse?.response?.['content'] as McpContentBlock[];
```
jakemac53: Can this be null? Especially if it is an error response?

packages/core/src/tools/mcp-tool.ts
```diff
@@ -307,6 +339,37 @@ function transformResourceLinkBlock(block: McpResourceLinkBlock): Part {
   };
 }
 
+/**
+ * Validates the MCP content blocks and replaces unsupported mime types with an error message.
+ * @param sdkResponse The raw Part[] array from `mcpTool.callTool()`.
+ */
+function validateMcpContent(sdkResponse: Part[]): void {
```
jacob314: a validate method that distructively modifies the content is surprising.
At minimum document that as part of the method signature  but preferably change this to return a the original sdkReponse if there are no issues or a modified version if there were changes required to make it valid.

jacob314 (commented):
Approved once these concerns are addressed.

jakemac53 (changes requested):
Longer term it might be interesting to save this kind of content to a file or something like that, maybe an MCP server is actually producing a binary file that you are intended to use somehow (you being the user).

For now this is probably fine though.

packages/core/src/tools/mcp-tool.ts
Paul-Deem: Can do, thanks for the feedback.

packages/core/src/tools/mcp-tool.ts
Paul-Deem: yea that's fair - I'll update to return the sanitized content.

packages/core/src/tools/mcp-tool.ts
Paul-Deem: The above evaluation can result in `mcpContent` being `undefined` since the responses from the tool could be null I catch that by ensuring the the content is a valid array. 

In the case that its not a valid array, there really isn't anything to sanitize so skipping the part seems like a reasonable course of action.

packages/core/src/tools/mcp-tool.ts
Paul-Deem: handling in above comment.

packages/core/src/tools/mcp-tool.ts
jakemac53: Can we just do `if (!mcpContent)` then instead to handle null/undefined? That would be more explicit - also please change the cast above to cast to a nullable type.

Paul-Deem:
> Longer term it might be interesting to save this kind of content to a file or something like that, maybe an MCP server is actually producing a binary file that you are intended to use somehow (you being the user).
> 
> For now this is probably fine though.

Yea there is a bit of both: the MCP server I'm working on is for cloud storage, so it works a lot like a local file system. Sometimes it makes sense to download the content, but a lot of the content is actually intended to be sent to be read directly by the agent. We don't want to force the intermediate download step unless necessary.

packages/core/src/tools/mcp-tool.ts
```diff
@@ -307,6 +313,57 @@ function transformResourceLinkBlock(block: McpResourceLinkBlock): Part {
...
+      continue;
+    }
+
+    for (let blockIndex = 0; blockIndex < mcpContent.length; blockIndex++) {
+      const block = mcpContent[blockIndex];
+      let mimeType: string | undefined;
+
+      if (block.type === 'image' || block.type === 'audio') {
+        mimeType = block.mimeType;
+      } else if (block.type === 'resource' && block.resource.blob) {
```
jakemac53, line 340: I believe resource-link types can also have a mime-type fwiw. Although, it would be the mime-type of the actual resource I think? Not totally clear.

Paul-Deem: They can - but an unsupported mime-type in a resource link won't cause an issue since it just gets converted to a `text` `Part` here:

`packages/core/src/tools/mcp-tool.ts:310`
image

If the agent chooses to read that resource it will fail, but it could download it, move it, or otherwise interact with it without needing to read it.

bbiggs:
I am curious if this will cause other problems. Imagine the gemini API starts to support a new mime type - to use it, after this change you have to update gemini CLI to have knowledge of the new mime type?  Therefore, if you're an MCP server that supports this new type, you need to warn clients they will need to update Gemini CLI to use it, and your MCP server will not work for clients who are stuck on an older version of Gemini CLI?

In general it seems risky for Gemini CLI to make assumptions about what is or is not supported on the Gemini API side, and I wonder if this should be more dynamic.

Paul-Deem:
> I am curious if this will cause other problems. Imagine the gemini API starts to support a new mime type - to use it, after this change you have to update gemini CLI to have knowledge of the new mime type? Therefore, if you're an MCP server that supports this new type, you need to warn clients they will need to update Gemini CLI to use it, and your MCP server will not work for clients who are stuck on an older version of Gemini CLI?
> 
> In general it seems risky for Gemini CLI to make assumptions about what is or is not supported on the Gemini API side, and I wonder if this should be more dynamic.

Honestly - I would love it to be more dynamic, configuration like this is pretty brittle. It would be significantly better if this was supported in the gemini client library or though some discoverable API which we could use for the duration of the agent session. 

The challenge with putting this in the MCP server is that you now couple the MCP server to a specific model, where the spec intends them to intentionally model agnostic. For an MCP server that runs attached to the agent instance this isn't much of an issue since the model can be configured at startup, but for a remotely hosted mcp instance this approach becomes significantly more difficult. 

Brainstorming here: 

I could change this so that the Gemini CLI eats the 400 error for unsupported mime-type, adjusts the offending part of the content, and retries the turn to the LLM. They do return the unsupported mime-type in the error message, parsing that is a little brittle, but might be better then the static configuration in the current PR. Thoughts? 

------

For completeness sake here is an overview of the mime-types supported by some of the major models: 

Category | MIME Type | Gemini 2.5 Pro | GPT-5* | Claude Opus 4.1
-- | -- | -- | -- | --
Image | image/png | ✅ | ✅ | ✅
  | image/jpeg | ✅ | ✅ | ✅
  | image/webp | ✅ | ✅ | ✅
  | image/gif | — | ✅ | ✅
Document | application/pdf | ✅ | ✅ | ✅
  | text/plain | ✅ | ✅ | ✅
  | text/csv | — | ✅ | ✅
  | application/msword (.doc) | — | ✅ | —
  | application/vnd.openxmlformats-officedocument.wordprocessingml.document (.docx) | — | ✅ | —
Audio | audio/wav | ✅ | — | —
  | audio/mp3 | ✅ | — | —
  | audio/aiff | ✅ | — | —
  | audio/aac | ✅ | — | —
  | audio/ogg | ✅ | — | —
  | audio/flac | ✅ | — | —
Video | video/mp4 | ✅ | — | —
  | video/mpeg | ✅ | — | —
  | video/mov | ✅ | — | —
  | video/avi | ✅ | — | —
  | video/x-flv | ✅ | — | —
  | video/mpg | ✅ | — | —
  | video/webm | ✅ | — | —
  | video/wmv | ✅ | — | —
  | video/3gpp | ✅ | — | —
Code/Text | text/x-python | — | ✅ | —
  | text/javascript | — | ✅ | —
  | application/json | — | ✅ | —
  | text/markdown | — | ✅ | —

packages/core/src/tools/mcp-tool.ts
jakemac53: Ok, if this ends up just falling back on a text representation that is fine for now, definitely some future work to be done for these though but it seems unrelated to this change.

PR #9537, opened 2025-09-25
refactor(packages): Use structuredClone()

gvking:
## TLDR

Using structuredClone() instead of JSON.parse(JSON.stringify())

## Dive Deeper

structuredClone() is faster, a native function, and is recommended to use instead of the following pattern that it replaces.

## Reviewer Test Plan

Simple compilation and run. This is a better practice recommended by style guides.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

### DIFF
--- a/packages/a2a-server/src/agent/task.test.ts
+++ b/packages/a2a-server/src/agent/task.test.ts
@@ -49,7 +49,7 @@ describe('Task', () => {
       },
     ];
 
-    const originalRequests = JSON.parse(JSON.stringify(requests));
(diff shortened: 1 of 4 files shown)


gvking:
Hm, better to keep existing behavior existing.

PR #9541, opened 2025-09-25
fix(core): Fix a cache collision bug in the edit corrector by using Stringify

gvking:
## TLDR

Fixed a cache collision bug in the edit corrector that could cause incorrect text replacements. Changed cache key generation from using `---` separators to JSON.stringify() to prevent collisions when code content contains the `---` sequence or any other special characters.

## Dive Deeper

The `ensureCorrectEdit` function uses a cache to avoid redundant LLM calls for edit corrections. The cache key was generated by concatenating `currentContent`, `old_string`, and `new_string` with `---` as a separator:

```javascript
const cacheKey = `${currentContent}---${originalParams.old_string}---${originalParams.new_string}`;
```

This approach had a collision vulnerability when the file content or edit parameters contained the `---` sequence, which is common in:
- YAML frontmatter (`---`)
- Markdown document separators 
- Code comments (`// --- Section ---`)
- Configuration files

**Example collision scenario:**
- Content: `"config = { theme: '---'"`, old: `"default"`, new: `"dark' }"`
- Content: `"config = { theme: '---'---default"`, old: `"dark' }"`, new: `""`
- Both produce the same cache key, causing incorrect cache hits

The fix uses `JSON.stringify()` to serialize the cache key components into an array, which is completely collision-proof and handles any possible content safely:

```javascript
const cacheKey = JSON.stringify([currentContent, originalParams.old_string, originalParams.new_string]);
```

This approach guarantees unique cache keys regardless of what characters appear in the content, including edge cases where LLMs might generate unusual characters.

## Reviewer Test Plan

To validate this fix:

1. **Test edit operations with `---` in content:**
   - Create files containing `---` (YAML, Markdown, comments)
   - Perform multiple edits and verify cache behavior is correct
   - Ensure no incorrect cached results are returned

2. **Verify cache functionality still works:**
   - Perform the same edit operation twice
   - Confirm second operation uses cached result (faster execution)
   - Check cache hit/miss behavior with debug logging

3. **Test edge cases:**
   - Files with various special characters (quotes, backslashes, unicode)
   - Large file content (cache key generation performance)
   - Multiple rapid edit operations
   - Content with JSON-like structures that might interfere with serialization

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

### DIFF
--- a/.github/workflows/release-promote.yml
+++ b/.github/workflows/release-promote.yml
@@ -68,12 +68,15 @@ jobs:
           GH_TOKEN: '${{ secrets.GITHUB_TOKEN }}'
         run: |
           set -e
-          STABLE_JSON=$(node scripts/get-release-version.js --type=stable ${{ github.event.inputs.stable_version_override && format('--stable_version_override={0}', github.event.inputs.stable_version_override) || '' }})
-          PREVIEW_JSON=$(node scripts/get-release-version.js --type=preview ${{ github.event.inputs.preview_version_override && format('--preview_version_override={0}', github.event.inputs.preview_version_override) || '' }})
-          NIGHTLY_JSON=$(node scripts/get-release-version.js --type=promote-nightly)
-          echo "STABLE_JSON_COMMAND=node scripts/get-release-version.js --type=stable ${{ github.event.inputs.stable_version_override && format('--stable_version_override={0}', github.event.inputs.stable_version_override) || '' }}"
-          echo "PREVIEW_JSON_COMMAND=node scripts/get-release-version.js --type=preview ${{ github.event.inputs.preview_version_override && format('--preview_version_override={0}', github.event.inputs.preview_version_override) || '' }}"
-          echo "NIGHTLY_JSON_COMMAND=node scripts/get-release-version.js --type=promote-nightly"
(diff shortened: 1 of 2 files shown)


packages/core/src/utils/editCorrector.ts
```diff
@@ -177,7 +177,7 @@ export async function ensureCorrectEdit(
   baseLlmClient: BaseLlmClient,
   abortSignal: AbortSignal,
 ): Promise<CorrectedEditResult> {
-  const cacheKey = `${currentContent}---${originalParams.old_string}---${originalParams.new_string}`;
+  const cacheKey = JSON.stringify([currentContent, originalParams.old_string, originalParams.new_string]);
```
agmsb: Simliar to the suggestion in #9542, can we hash this string?

agmsb (changes requested):

packages/core/src/utils/editCorrector.ts
gvking: Done.

agmsb:
Can we add a test similar to in #9542?

PR #9542, opened 2025-09-25
Fix a cache collision bug in the llm edit fixer

gvking:
## TLDR

Fixed a cache collision bug in the LLM edit corrector that could cause incorrect text replacements. Changed cache key generation from using `---` separators to `JSON.stringify()` to prevent collisions when code content contains the `---` sequence or any other special characters.

## Dive Deeper

The `FixLLMEditWithInstruction` function uses a cache to avoid redundant LLM calls for edit corrections. The cache key was generated by concatenating `instruction`, `old_string`, `new_string`, `current_content`, and `error` with `---` and `--` as separators:

```typescript
const cacheKey = `${instruction}---${old_string}---${new_string}--${current_content}--${error}`;
```

This approach had a collision vulnerability when the function parameters contained the separator sequences, which is common in:
- YAML frontmatter (`---`)
- Markdown document separators
- Code comments (`// --- Section ---`)
- Configuration files
- LLM-generated content with arbitrary text

**Example collision scenario:**
Different parameter combinations could theoretically produce identical cache keys, causing incorrect cache hits and wrong edit corrections being applied.

The fix uses `JSON.stringify()` to serialize the cache key components into an array, which is completely collision-proof and handles any possible content safely:

```typescript
const cacheKey = JSON.stringify([instruction, old_string, new_string, current_content, error]);
```

This approach guarantees unique cache keys regardless of what characters appear in the content, including edge cases where LLMs might generate unusual characters.

## Reviewer Test Plan

To validate this fix:

1. **Test edit operations with `---` in content:**
   - Create files containing `---` (YAML, Markdown, comments)
   - Perform multiple edit operations and verify cache behavior is correct
   - Ensure no incorrect cached results are returned

2. **Verify cache functionality still works:**
   - Perform the same edit operation twice
   - Confirm second operation uses cached result (faster execution)
   - Check cache hit/miss behavior with debug logging

3. **Test edge cases:**
   - Files with various special characters (quotes, backslashes, unicode)
   - Large file content (cache key generation performance)
   - Multiple rapid edit operations
   - Content with JSON-like structures that might interfere with serialization

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

This fix addresses a cache collision vulnerability similar to the one described in the edit corrector. The issue could cause incorrect LLM edit corrections to be cached and applied, leading to wrong file modifications when processing content containing separator sequences.


packages/core/src/utils/llm-edit-fixer.ts
```diff
@@ -116,7 +116,7 @@ export async function FixLLMEditWithInstruction(
     );
   }
 
-  const cacheKey = `${instruction}---${old_string}---${new_string}--${current_content}--${error}`;
+  const cacheKey = JSON.stringify([instruction, old_string, new_string, current_content, error]);
```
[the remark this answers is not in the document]
gvking: Done.

agmsb: Could we hash this string to make the cacheKeys smaller?

agmsb (changes requested):
Thanks for finding and chasing this down! Left a minor comment.

packages/core/src/utils/llm-edit-fixer.ts
gvking: Done!

agmsb (approved):
Thanks for submitting!

Issue #3691, opened 2025-07-09
Infra: Ensure sandbox image is built with Release process


mattKorwel:
This task is to migrate the hosting of our public sandbox Docker images from the private Google Artifact Registry to Docker Hub.

#### The Problem

Currently, our sandbox images are intended for public consumption but are hosted in a private Google Artifact Registry (`us-west1-docker.pkg.dev/gemini-code-dev/...`). This is not an ideal public distribution point because:

*   **Discoverability:** Docker Hub is the default and most widely used container registry. Hosting our images there makes them easier for users to find and trust.
*   **Accessibility:** While Artifact Registry can be configured for public access, it's not the standard workflow for most Docker users and can add unnecessary friction.
*   **Community Standard:** Using Docker Hub aligns with open-source best practices and community expectations for public Docker images.

#### The Goal

To improve accessibility and align with community standards, we will publish our official sandbox images to a public repository on Docker Hub.

This involves:
1.  **Creating a Docker Hub Repository:** Create an official public repository for the sandbox image under the **`@google` organization on Docker Hub**. This is a multi-step process that requires coordination. **Please ping @mattkorwel for information on the internal process for requesting a new repository under this organization.**
2.  **Updating the Release Pipeline:** Modify our release workflow to authenticate to Docker Hub (using a token stored as a GitHub secret) and push the final image to the new repository.
3.  **Updating Documentation:** Ensure any documentation or user-facing materials are updated to reference the new Docker Hub image path.

This change will make our sandbox environment more accessible and discoverable for our users and the wider open-source community.

gsquared94:
Found possible duplicate issues:

- #3689: (0.9007900953292847)
- #3684: (0.9001484513282776)

richieforeman:
Instead of Dockerhub, we've decided to go with GHCR, but all the same guidance still applies.

richieforeman:
I added a sandbox job, here's what's outstanding:

- [~] Move it to Dockerhub (In progress)
- [ ] Hook it up to the release pipeline (easy enough).

Need dockerhub info @mattKorwel 

Two repos:
- gemini-cli
- gemini-cli-sandbox

Two secrets:
- DOCKERHUB_USER (containing username)
- DOCKERHUB_TOKEN (containing PAT w/ R+W that was issued by {DOCKERHUB_USER})

richieforeman:
Blocked on security review @ https://b.corp.google.com/issues/448140256

[...]

PR #9722, opened 2025-09-25
feat(ci): Push sandbox w/ Release flows

richieforeman:
## TLDR

This pushes the sandbox to dockerhub w/ the Release flows.

## Dive Deeper

- We might need to adjust how the sandbox gets tagged, and we might also need to adjust the package.json.  I wanted to put this in as a bit of a "dark launch" so we can see how this behaves.

## Reviewer Test Plan

Dry run @  https://github.com/google-gemini/gemini-cli/actions/runs/17988727836/job/51173417575
full run @ https://github.com/google-gemini/gemini-cli/actions/runs/18006457862/job/51227740426

## Testing Matrix

N/A

## Linked issues / bugs

Part of [#3691](https://github.com/google-gemini/gemini-cli/issues/3691)


.github/actions/push-sandbox/action.yml
```diff
@@ -47,12 +41,17 @@ runs:
...
+      shell: 'bash'
+      run: |-
+        echo "GITHUB_SHA=$(git rev-parse HEAD)" >> $GITHUB_OUTPUT
     - name: 'determine image tag'
       id: 'image_tag'
       shell: 'bash'
       run: |-
         SHELL_TAG_NAME="${{ inputs.github-ref-name }}"
-        FINAL_TAG="${{ inputs.github-sha }}"
+        FINAL_TAG="${{ steps.sha.outputs.GITHUB_SHA }}"
```
mattKorwel, line 54: So I think every docker image should be tagged with its git sha 👍 

The releases also need tagged with their release version and if its 'latest' npm it also needs tagged as latest in docker.  

How would that work in this flow?

richieforeman: Honestly, I'd probably get rid of this Javascript file `build_sandbox.js` and probably go with something closer to what I did in https://github.com/google-gemini/gemini-cli/blob/main/.github/actions/push-docker/action.yml -- I find this approach way more coherant, and I don't think the Javascript file is doing anything special and feels like a legacy artifact.

I think this Javascript file has outlived it's shelf life.

This would give us more direct and access to docker publish commands.

WDYT?

richieforeman: Expanding, I'd appreciate an approval on this one so I can at least start running this manually to verify: https://github.com/google-gemini/gemini-cli/pull/9941/files

I think this PR demonstrates the approach I'm talking about.

PR #9760, opened 2025-09-25
Fix duplicate info messages for extension updates

jakemac53:
## TLDR

Previously there was a race condition where you could get zero, one, or multiple possible notifications for updates to extensions. This fixes that so you just get one notification reliably.

I also combined all the info messages into one generic message like this:
image

## Dive Deeper

The primary issue is this useExtensionState function being called many times on startup in quick succession, which was resulting in simultaneous ongoing calls to check the extension update states. These were all mutating the same extension state object and getting in race conditions.

I added a fix to ensure we only have one ongoing check at a time by adding an extra boolean state object to track if we have an ongoing update check already happening.

I also added a fix to not edit the original extension state object but instead return a new one from `checkForAllExtensionUpdates`, this was creating the situation where we would have no log at all sometimes due to the previous and new states being equal (the same object). 

## Reviewer Test Plan

Run the CLI with extensions installed that need updates, you should just see one message to update.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

### DIFF
--- a/packages/cli/src/config/extensions/update.ts
+++ b/packages/cli/src/config/extensions/update.ts
@@ -130,7 +130,9 @@ export async function checkForAllExtensionUpdates(
   >,
   cwd: string = process.cwd(),
 ): Promise<Map<string, ExtensionUpdateState>> {
-  let newStates: Map<string, ExtensionUpdateState> = new Map();
+  let newStates: Map<string, ExtensionUpdateState> = new Map(
+    extensionsUpdateState,
+  );
   for (const extension of extensions) {
     const initialState = extensionsUpdateState.get(extension.name);
     if (initialState === undefined) {
--- a/packages/cli/src/config/extensions/update.ts
+++ b/packages/cli/src/config/extensions/update.ts
@@ -130,17 +130,15 @@ export async function checkForAllExtensionUpdates(
   >,
   cwd: string = process.cwd(),
 ): Promise<Map<string, ExtensionUpdateState>> {
+  let newStates: Map<string, ExtensionUpdateState> = new Map();
   for (const extension of extensions) {
     const initialState = extensionsUpdateState.get(extension.name);
     if (initialState === undefined) {
       if (!extension.installMetadata) {
         setExtensionsUpdateState((prev) => {
-          extensionsUpdateState = new Map(prev);
-          extensionsUpdateState.set(
-            extension.name,
-            ExtensionUpdateState.NOT_UPDATABLE,
-          );
-          return extensionsUpdateState;
+          newStates = new Map(prev);
+          newStates.set(extension.name, ExtensionUpdateState.NOT_UPDATABLE);
+          return newStates;
         });
         continue;
       }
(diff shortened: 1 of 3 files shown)


chrstnb (approved):

packages/cli/src/config/extensions/update.ts
```diff
@@ -130,32 +130,30 @@ export async function checkForAllExtensionUpdates(
   >,
   cwd: string = process.cwd(),
 ): Promise<Map<string, ExtensionUpdateState>> {
+  let newStates: Map<string, ExtensionUpdateState> = new Map();
```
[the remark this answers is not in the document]
jakemac53: done

packages/cli/src/ui/hooks/useExtensionUpdates.ts
```diff
@@ -63,24 +68,23 @@ export const useExtensionUpdates = (
...
-    extensionsUpdateState,
-    setExtensionsUpdateState,
-    addItem,
-    cwd,
-  ]);
+      if (extensionsWithUpdatesCount > 0) {
+        addItem(
+          {
+            type: MessageType.INFO,
+            text: `You have ${extensionsWithUpdatesCount} extensions with updates available, run "/extensions list" for more information.`,
```
chrstnb: make "extensions" singular fi there's just one

packages/cli/src/ui/hooks/useExtensionUpdates.ts
```diff
@@ -23,13 +23,18 @@ export const useExtensionUpdates = (
...
+  (async () => {
+    if (isChecking) return;
+    setIsChecking(true);
+    try {
       const updateState = await checkForAllExtensionUpdates(
         extensions,
         extensionsUpdateState,
         setExtensionsUpdateState,
       );
+      let extensionsWithUpdatesCount = 0;
```
chrstnb, line 37: Just double-checking, how will this work if there are also some auto-updatable extensions?

jakemac53: They don't contribute to this count

packages/cli/src/ui/hooks/useExtensionUpdates.ts
jakemac53: done

chrstnb (approved):

packages/cli/src/ui/hooks/useExtensionUpdates.ts
```diff
@@ -23,13 +23,18 @@ export const useExtensionUpdates = (
   const [extensionsUpdateState, setExtensionsUpdateState] = useState(
     new Map<string, ExtensionUpdateState>(),
   );
-  useMemo(() => {
-    const checkUpdates = async () => {
+  const [isChecking, setIsChecking] = useState(false);
+
+  (async () => {
+    if (isChecking) return;
+    setIsChecking(true);
```
jacob314, line 30: be very careful adding logic like this that unconditionally updates react state. This code causes us to render frames continuously causing flicker because we are continuously setting setIsChecking to true and then back to false.

PR #9763, opened 2025-09-25
fix(core): Improve API error retry logic

SandyTao520:
## TLDR

This pull request refactors the API error handling and retry mechanism within the core chat functionality. Previously, we relied on string-matching error messages to decide whether to retry an API call. This update switches to using the structured `ApiError` from the GenAI SDK, checking the HTTP status code directly. This makes our retry logic more robust and reliable, ensuring we correctly retry on transient errors like rate limits (429) and server errors (5xx) while failing fast on client-side issues like bad requests (400).

## Dive Deeper

The previous implementation of our retry logic was brittle because it parsed the text of an error message to find status codes. This approach can easily break if the upstream SDK changes its error message formatting.

This change improves our resilience by:
1.  **Type-Safe Error Checking**: We now check if an error is an `instanceof ApiError` from `@google/genai`.
2.  **Using Status Codes**: We use the `error.status` property to make decisions, which is a much more stable API than the error message.
3.  **Clearer Logic**: The `shouldRetry` function is now more explicit:
    *   **Do not retry** on `400 Bad Request` errors.
    *   **Do not retry** on "maximum schema depth exceeded" errors.
    *   **Retry** on `429 Rate Limit Exceeded` errors.
    *   **Retry** on all `5xx` server-side errors.

This ensures the CLI is more predictable when interacting with the Gemini API, leading to a better user experience during transient network or server issues.

## Reviewer Test Plan

Validating this change manually is difficult as it requires forcing the backend to produce specific HTTP error codes.

The most effective way to review this PR is to:
1. Use the prompt `what's in @packages/`, this would exceed token limit and get 400 bad request.
2. Verify this API error won't trigger a retry.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ✅  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

### DIFF
--- a/packages/core/src/core/geminiChat.test.ts
+++ b/packages/core/src/core/geminiChat.test.ts
@@ -10,6 +10,7 @@ import type {
   GenerateContentConfig,
   GenerateContentResponse,
 } from '@google/genai';
+import { ApiError } from '@google/genai';
 import type { ContentGenerator } from '../core/contentGenerator.js';
 import {
   GeminiChat,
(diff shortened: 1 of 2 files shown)


anj-s (approved):
nice one!

PR #9765, opened 2025-09-25
Unset foreground in default themes

miguelsolorio:
## TLDR

This fixes a long standing issue where users on light terminals can't really see the text.

## Dive Deeper

By unsetting the foreground color, it relies on using the terminal foreground color. This only applies to our default dark/light themes, other themes will have custom foreground colors.

### Before
CleanShot 2025-09-25 at 15 16 16@2x

### After
CleanShot 2025-09-25 at 15 17 05@2x

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


packages/cli/src/ui/utils/displayUtils.ts
```diff
@@ -28,5 +28,5 @@ export const getStatusColor = (
   if (value >= thresholds.yellow) {
     return theme.status.warning;
   }
-  return options.defaultColor || theme.status.error;
```
jacob314, line 31: did you gemini cli catch this or did you manually debug to fix this? For what it is worth this seems wrong both before and after. Likely we should have a red/error threshold and return theme.status.error in that case
and always return options.defaultColor for this default case.

jacob314 (approved):
lgtm

packages/cli/src/ui/utils/displayUtils.ts
miguelsolorio: Gemini originally made the change but then GCA suggested this instead (you can see the resolved thread above). Happy to add your suggestions.

miguelsolorio: Let me follow up in a subsequent PR to address this

PR #9768, opened 2025-09-25
fix(ui): Truncate long loading text

SandyTao520:
## TLDR

This pull request updates the `LoadingIndicator` component to truncate long loading phrases instead of wrapping them. This prevents the UI from shifting or expanding when a long loading message is displayed, ensuring a more stable and visually consistent layout.

## Dive Deeper

Previously, if a long string was provided to the `currentLoadingPhrase` prop in the `LoadingIndicator`, the text would wrap to a new line. This could cause the entire terminal UI to reflow, which is a jarring user experience.

This change applies the `wrap="truncate-end"` property to the `Text` component responsible for rendering the primary loading message. Now, any text that exceeds the available width will be cleanly truncated with an ellipsis at the end, maintaining the single-line integrity of the loading indicator.

## Reviewer Test Plan

Hard to repro.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ✅  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

### DIFF
--- a/packages/cli/src/ui/components/LoadingIndicator.test.tsx
+++ b/packages/cli/src/ui/components/LoadingIndicator.test.tsx
@@ -233,6 +233,21 @@ describe('<LoadingIndicator />', () => {
     expect(output).not.toContain('This should not be displayed');
   });
 
+  it('should truncate long primary text instead of wrapping', () => {
+    const { lastFrame } = renderWithContext(
+      <LoadingIndicator
+        {...defaultProps}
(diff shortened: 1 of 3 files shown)


jacob314 (approved):
lgtm

Tested locally and it looks good

Issue #6636, opened 2025-08-20
Add support for context harvesting using subagents


anj-s:
We now have support for subagent as a tool and we can use that at the beginning of a task to understand as much as we can about the problem space.

PR #9778, opened 2025-09-25
refactor(agents): Introduce Declarative Agent Framework

abhipatel12:
## TLDR

This pull request refactors our subagent infrastructure from `SubAgentScope` into a declarative, definition-based framework. The core of this change is the introduction of an `AgentDefinition` interface, which allows agents to be defined as simple configuration objects.

The most significant improvement is that any agent defined this way can be automatically wrapped and exposed as a standard `Tool`. This makes agents reusable and composable, allowing us to build more complex, multi-agent systems where agents can call each other as specialized tools.

This PR lays components for this new framework and includes a `CodebaseInvestigatorAgent` as a reference implementation. The new framework is not yet integrated into the live codebase; this will be done in a follow-up PR.

## Dive Deeper

This refactor introduces several key components that work together to create a more robust and scalable agent system:

-   **`AgentDefinition` (`packages/core/src/agents/types.ts`):** A new interface for declaratively defining an agent's properties, including its name, description, model/run/tool configurations, and strongly-typed `inputConfig` and `outputConfig`. This replaces the need for bespoke classes for each agent.

-   **`AgentExecutor` (`packages/core/src/agents/executor.ts`):** This is the new engine for running agents. It takes an `AgentDefinition` and inputs, and orchestrates the execution loop. It uses a two-phase process:
    1.  **Work Phase:** The agent iteratively calls tools to gather information and work towards its goal.
    2.  **Extraction Phase:** Once the agent stops calling tools, the executor sends a final prompt to the model to synthesize the findings and generate the final output, guided by the `outputConfig`.

-   **`SubagentToolWrapper` (`packages/core/src/agents/subagent-tool-wrapper.ts`):** This class is the bridge that makes agents reusable. It wraps an `AgentDefinition` and exposes it as a `BaseDeclarativeTool`. It dynamically generates a JSON schema from the agent's `inputConfig`, so the parent agent gets strongly-typed parameters for calling the subagent.

-   **`AgentRegistry` (`packages/core/src/agents/registry.ts`):** A simple registry for discovering and managing available `AgentDefinition`s.

-   **`CodebaseInvestigatorAgent` (`packages/core/src/agents/codebase-investigator.ts`):** A new, read-only agent that serves as a practical example of the `AgentDefinition` format. It's designed to analyze a codebase to answer questions about its structure and conventions.

## Reviewer Test Plan

Since this is a foundational refactor that is not yet wired into the main CLI, validation should focus on the soundness of the new architecture and its components.

1.  **Review the Core Abstractions:**
    -   Examine the `AgentDefinition` in `packages/core/src/agents/types.ts`. Does the structure seem flexible and comprehensive enough for defining future agents?
    -   Look at the `CodebaseInvestigatorAgent` as a concrete example. Is the definition clear and easy to understand?

2.  **Understand the Execution Flow:**
    -   Trace the logic in `AgentExecutor.run()`. Pay attention to the separation between the work phase and the extraction phase.
    -   Review how tool calls are processed and how termination conditions (max turns, timeout) are handled.

3.  **Run the Unit Tests:**

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  |✅| ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt |✅| -   | -   |

## Linked issues / bugs

This makes progress on #6636

### DIFF
--- a/packages/core/src/agents/codebase-investigator.ts
+++ b/packages/core/src/agents/codebase-investigator.ts
@@ -9,9 +9,7 @@ import { LSTool } from '../tools/ls.js';
 import { ReadFileTool } from '../tools/read-file.js';
 import { GlobTool } from '../tools/glob.js';
 import { GrepTool } from '../tools/grep.js';
-import {
-  /*DEFAULT_GEMINI_MODEL,*/ DEFAULT_GEMINI_FLASH_MODEL,
-} from '../config/models.js';
+import { DEFAULT_GEMINI_MODEL } from '../config/models.js';
 
 /**
  * A Proof-of-Concept subagent specialized in analyzing codebase structure,
@@ -44,7 +42,7 @@ export const CodebaseInvestigatorAgent: AgentDefinition = {
   },
 
   modelConfig: {
-    model: DEFAULT_GEMINI_FLASH_MODEL,
+    model: DEFAULT_GEMINI_MODEL,
     temp: 0.2,
     top_p: 1.0,
     thinkingBudget: -1,
--- a/packages/core/src/agents/codebase-investigator.ts
+++ b/packages/core/src/agents/codebase-investigator.ts
@@ -1,0 +1,83 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import type { AgentDefinition } from './types.js';
+import { LSTool } from '../tools/ls.js';
+import { ReadFileTool } from '../tools/read-file.js';
+import { GlobTool } from '../tools/glob.js';
+import { GrepTool } from '../tools/grep.js';
+import { DEFAULT_GEMINI_MODEL } from '../config/models.js';
+
+/**
+ * A Proof-of-Concept subagent specialized in analyzing codebase structure,
+ * dependencies, and technologies.
+ */
+export const CodebaseInvestigatorAgent: AgentDefinition = {
+  name: 'codebase_investigator',
+  displayName: 'Codebase Investigator Agent',
+  description:
+    'A specialized agent used for analyzing and reporting on the structure, technologies, dependencies, and conventions of the current codebase. Use this when asked to understand how the project is set up or how a specific feature is implemented.',
+
+  inputConfig: {
+    inputs: {
+      investigation_focus: {
+        description:
+          'A high-level description of what the agent should investigate (e.g., "frontend framework", "authentication implementation", "testing conventions").',
+        type: 'string',
+        required: true,
+      },
+    },
+  },
+  outputConfig: {
+    description:
+      'A detailed markdown report summarizing the findings of the codebase investigation.',
+    completion_criteria: [
+      'The report must directly address the initial `investigation_focus`.',
+      'Cite specific files, functions, or configuration snippets as evidence for your findings.',
+      'Conclude with a summary of the key technologies, architectural patterns, and conventions discovered.',
+    ],
+  },
+
+  modelConfig: {
+    model: DEFAULT_GEMINI_MODEL,
+    temp: 0.2,
+    top_p: 1.0,
+    thinkingBudget: -1,
+  },
+
+  runConfig: {
+    max_time_minutes: 5,
+    max_turns: 15,
+  },
+
+  toolConfig: {
+    // Grant access only to read-only tools.
+    tools: [LSTool.Name, ReadFileTool.Name, GlobTool.Name, GrepTool.Name],
+  },
+
+  promptConfig: {
+    systemPrompt: `You are the Codebase Investigator agent. Your sole purpose is to analyze the provided codebase and generate a detailed report on a specific area of focus.
+
+# Task
+Your focus for this investigation is: \${investigation_focus}
+
+# Methodology
+1. **Discovery:** Start by looking at high-level configuration files (e.g., package.json, README.md, Cargo.toml, requirements.txt, build.gradle) to understand the project's dependencies and structure.
+2. **Structure Analysis:** Use '${GlobTool.Name}' and '${LSTool.Name}' to understand the directory layout and identify relevant files/modules related to your focus.
+3. **Deep Dive:** Use '${ReadFileTool.Name}' and available search tools (Grep/RipGrep) to analyze the contents of relevant files, looking for implementation details, patterns, and conventions.
+4. **Synthesis:** Synthesize all findings into a coherent markdown report.
+
+# Rules
+* You MUST ONLY use the tools provided to you.
+* You CANNOT modify the codebase.
+* You must be thorough in your investigation.
+* Once you have gathered sufficient information, stop calling tools. Your findings will be synthesized into the final report.
+
+# Report Format
+The final report should be structured markdown, clearly answering the investigation focus, citing the evidence (files analyzed), and summarizing the technologies/patterns found.
+`,
+  },
+};
--- a/packages/core/src/agents/codebase-investigator.ts
+++ b/packages/core/src/agents/codebase-investigator.ts
@@ -9,7 +9,9 @@ import { LSTool } from '../tools/ls.js';
 import { ReadFileTool } from '../tools/read-file.js';
 import { GlobTool } from '../tools/glob.js';
 import { GrepTool } from '../tools/grep.js';
-import { DEFAULT_GEMINI_MODEL } from '../config/models.js';
+import {
+  /*DEFAULT_GEMINI_MODEL,*/ DEFAULT_GEMINI_FLASH_MODEL,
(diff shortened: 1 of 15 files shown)


packages/core/src/agents/codebase-investigator.ts
```diff
@@ -0,0 +1,85 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import type { AgentDefinition } from './types.js';
+import { LSTool } from '../tools/ls.js';
+import { ReadFileTool } from '../tools/read-file.js';
+import { GlobTool } from '../tools/glob.js';
+import { GrepTool } from '../tools/grep.js';
+import {
+  /*DEFAULT_GEMINI_MODEL,*/ DEFAULT_GEMINI_FLASH_MODEL,
```
silviojr: back to DEFAULT_GEMINI_MODEL?

packages/core/src/agents/codebase-investigator.ts
```diff
@@ -0,0 +1,85 @@
...
+  },
+
+  runConfig: {
+    max_time_minutes: 5,
+    max_turns: 15,
+  },
+
+  toolConfig: {
+    // Grant access only to read-only tools.
+    tools: [LSTool.Name, ReadFileTool.Name, GlobTool.Name, GrepTool.Name],
```
silviojr, line 58: multi_file_read?

packages/core/src/agents/codebase-investigator.ts
abhipatel12: Done

packages/core/src/agents/codebase-investigator.ts
abhipatel12: i'm actually not sure if this is useful or not. Wonder what the tradeoff is:
- 1 tool (encourage parallel use)
- 2 tools (agent decides which tool to use when it wants to read multiple things) --> This feels like more reasoning.

packages/core/src/agents/codebase-investigator.ts
```diff
@@ -0,0 +1,83 @@
...
+    description:
+      'A detailed markdown report summarizing the findings of the codebase investigation.',
+    completion_criteria: [
+      'The report must directly address the initial `investigation_focus`.',
+      'Cite specific files, functions, or configuration snippets as evidence for your findings.',
+      'Conclude with a summary of the key technologies, architectural patterns, and conventions discovered.',
+    ],
+  },
+
+  modelConfig: {
```
anj-s, line 44: can you specify multiple models if needed?

packages/core/src/agents/codebase-investigator.ts
anj-s: what do you mean by parallel use?

packages/core/src/agents/executor.ts
```diff
@@ -0,0 +1,578 @@
...
+
+        currentMessages = await this.processFunctionCalls(
+          functionCalls,
+          signal,
+          promptId,
+        );
+      }
+
+      // If the work phase was terminated early, skip extraction and return.
+      if (terminateReason !== AgentTerminateMode.GOAL) {
```
anj-s, line 195: where does terminateReason get populated?

packages/core/src/agents/executor.ts
```diff
@@ -0,0 +1,578 @@
...
+      // Phase 2: Extraction Phase
+      // A final message is sent to summarize findings and produce the output.
+      const extractionMessage = this.buildExtractionMessage();
+      const extractionMessages: Content[] = [
+        { role: 'user', parts: [{ text: extractionMessage }] },
+      ];
+
+      const extractionPromptId = `${this.runtimeContext.getSessionId()}#${this.agentId}#extraction`;
+
+      // TODO: Consider if we should keep tools to avoid cache reset.
```
anj-s, line 211: yes, if the system message changes and the only thing you are not sending is the tool definition I would add it back.

packages/core/src/agents/executor.ts
```diff
@@ -0,0 +1,578 @@
...
+          terminateReason = AgentTerminateMode.ABORTED;
+          break;
+        }
+
+        // If the model stops calling tools, the work phase is complete.
+        if (functionCalls.length === 0) {
+          break;
+        }
+
+        currentMessages = await this.processFunctionCalls(
```
anj-s, line 187: if processFunctionCalls returns a False, what does currentMessage look like?

packages/core/src/agents/executor.ts
```diff
@@ -0,0 +1,578 @@
...
+    signal: AbortSignal,
+    promptId: string,
+  ): Promise<Content[]> {
+    const allowedToolNames = new Set(this.toolRegistry.getAllToolNames());
+
+    // Filter out any tool calls that are not in the agent's allowed list.
+    const validatedFunctionCalls = functionCalls.filter((call) => {
+      if (!allowedToolNames.has(call.name as string)) {
+        console.warn(
+          `[AgentExecutor] Agent '${this.definition.name}' attempted to call ` +
```
anj-s, line 359: why is this not treated like any other function call in the parent agent? i.e if a tool call is not present in the registered list than it is a hallucinated tool call.

packages/core/src/agents/executor.ts
```diff
@@ -0,0 +1,578 @@
...
+            `unauthorized tool '${call.name}'. This call has been blocked.`,
+        );
+        return false;
+      }
+      return true;
+    });
+
+    const toolPromises = validatedFunctionCalls.map(async (functionCall) => {
```
anj-s, line 367: how are we dealing with parallel function calls that might have mutable dependencies? This doesn't occur in this example but we need to account for it.

packages/core/src/agents/executor.ts
```diff
@@ -0,0 +1,578 @@
...
+    const toolResponses = await Promise.all(toolPromises);
+    const toolResponseParts: Part[] = toolResponses
+      .flatMap((response) => response.responseParts)
+      .filter((part): part is Part => part !== undefined);
+
+    // If all authorized tool calls failed, provide a generic error message
+    // to the model so it can try a different approach.
+    if (functionCalls.length > 0 && toolResponseParts.length === 0) {
+      toolResponseParts.push({
+        text: 'All tool calls failed. Please analyze the errors and try an alternative approach.',
```
anj-s, line 415: why do we need this additional prompt?

packages/core/src/agents/types.ts
```diff
@@ -0,0 +1,139 @@
...
+import type { Content, FunctionDeclaration } from '@google/genai';
+import type { AnyDeclarativeTool } from '../tools/tools.js';
+
+/**
+ * Describes the possible termination modes for an agent.
+ */
+export enum AgentTerminateMode {
+  ERROR = 'ERROR',
+  TIMEOUT = 'TIMEOUT',
+  GOAL = 'GOAL',
```
anj-s, line 20: this is the only one I don't understand completely. can you share more of why this is needed.

packages/core/src/agents/types.ts
```diff
@@ -0,0 +1,139 @@
...
+/**
+ * The definition for an agent.
+ */
+export interface AgentDefinition {
+  /** Unique identifier for the agent. */
+  name: string;
+  displayName?: string;
+  description: string;
+  promptConfig: PromptConfig;
+  modelConfig: ModelConfig;
```
anj-s, line 58: this will need to be a list eventually.

packages/core/src/agents/types.ts
```diff
@@ -0,0 +1,139 @@
...
+
+import type { Content, FunctionDeclaration } from '@google/genai';
+import type { AnyDeclarativeTool } from '../tools/tools.js';
+
+/**
+ * Describes the possible termination modes for an agent.
+ */
+export enum AgentTerminateMode {
+  ERROR = 'ERROR',
+  TIMEOUT = 'TIMEOUT',
```
anj-s, line 19: is this tool timeout, model timeout or agent timeout?

anj-s (approved):

packages/core/src/agents/executor.ts
abhipatel12: Since we are considering adding back the explicit tool call to end the agent's turn this should be accounted for in the subsequent PR

packages/core/src/agents/types.ts
abhipatel12: Agent timeout -- exceeded the timeout set by the agent definition. 

model timeout or tool timeout can contribute to this. 

So if the model is slated for 5 min max and its tools take 6 min to return a value, this could cause the model to timeout after the response is recevied.

PR #9783, opened 2025-09-25
fix(cli): Make IDE trust listener also listen to IDE status changes a…

shrutip90:
…nd update trust

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Related to https://github.com/google-gemini/maintainers-gemini-cli/issues/717

### DIFF
--- a/packages/cli/src/ui/AppContainer.tsx
+++ b/packages/cli/src/ui/AppContainer.tsx
@@ -88,6 +88,7 @@ import { useSessionStats } from './contexts/SessionContext.js';
 import { useGitBranchName } from './hooks/useGitBranchName.js';
 import { useExtensionUpdates } from './hooks/useExtensionUpdates.js';
 import { ShellFocusContext } from './contexts/ShellFocusContext.js';
+import { relaunchApp } from '../utils/processUtils.js';
 
 const CTRL_EXIT_PROMPT_DURATION_MS = 1000;
 
@@ -970,7 +971,7 @@ Logging in with Google... Please restart Gemini CLI to continue.
   useKeypress(
     (key) => {
       if (key.name === 'r' || key.name === 'R') {
-        process.exit(0);
+        relaunchApp();
       }
     },
     { isActive: showIdeRestartPrompt },
@@ -1009,6 +1010,7 @@ Logging in with Google... Please restart Gemini CLI to continue.
     isAuthDialogOpen ||
     isEditorDialogOpen ||
     showPrivacyNotice ||
+    showIdeRestartPrompt ||
     !!proQuotaRequest;
 
   const pendingHistoryItems = useMemo(
--- a/packages/cli/src/ui/AppContainer.tsx
+++ b/packages/cli/src/ui/AppContainer.tsx
@@ -88,7 +88,6 @@ import { useSessionStats } from './contexts/SessionContext.js';
 import { useGitBranchName } from './hooks/useGitBranchName.js';
 import { useExtensionUpdates } from './hooks/useExtensionUpdates.js';
 import { ShellFocusContext } from './contexts/ShellFocusContext.js';
-import { relaunchApp } from '../utils/processUtils.js';
 
 const CTRL_EXIT_PROMPT_DURATION_MS = 1000;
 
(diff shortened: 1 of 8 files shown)


packages/cli/src/ui/hooks/useIdeTrustListener.ts
```diff
@@ -4,44 +4,77 @@
...
-      initialTrustValue !== isIdeTrusted
-    ) {
+    const currentTrust = isWorkspaceTrusted(settings.merged).isTrusted;
+    // Trigger a restart if the overall trust status for the CLI has changed.
+    if (previousTrust.current !== currentTrust) {
       setNeedsRestart(true);
     }
-  }, [isIdeTrusted, initialTrustValue, needsRestart]);
+    previousTrust.current = currentTrust;
+  }, [isIdeTrusted, settings.merged]);
```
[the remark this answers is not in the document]
shrutip90: Done

packages/cli/src/ui/components/DialogManager.tsx
```diff
@@ -46,8 +46,8 @@ export const DialogManager = ({ addItem }: DialogManagerProps) => {
     return (
       <Box borderStyle="round" borderColor={theme.status.warning} paddingX={1}>
         <Text color={theme.status.warning}>
-          Workspace trust has changed. Press &apos;r&apos; to restart Gemini to
-          apply the changes.
+          Workspace trust has changed due to IDE connection or trust change.
+          Press &apos;r&apos; to restart Gemini to apply the changes.
```
jacob314: this message will be confusing to the % of users without an IDE connection. can you make it say why the trust was changed to that it could be IDE connection or trust change if that is trivial enough to plumb through?

jacob314: also please move this out of DialogManager into its own dedicated WorkspaceTrustDialog class for easier testability.

packages/cli/src/ui/AppContainer.tsx
```diff
@@ -970,7 +971,7 @@ Logging in with Google... Please restart Gemini CLI to continue.
   useKeypress(
```
jacob314: also not a new issue but in the interest of reducing tech debt, please move this directly into the dialog that shows the restart message RestartDialog rather than having this logic arbitrarily separated in AppContainer.

jacob314 (commented):
Approved after these fixes are made.

packages/cli/src/ui/AppContainer.tsx
shrutip90: Done

packages/cli/src/ui/components/DialogManager.tsx
shrutip90: Done. This message will never show outside the IDE. It is for IDE connection or IDE trust change. Separated the messages anyway. PTAL.

packages/cli/src/ui/components/DialogManager.tsx
```diff
@@ -43,14 +44,7 @@ export const DialogManager = ({ addItem }: DialogManagerProps) => {
...
-      <Box borderStyle="round" borderColor={theme.status.warning} paddingX={1}>
-        <Text color={theme.status.warning}>
-          Workspace trust has changed. Press &apos;r&apos; to restart Gemini to
-          apply the changes.
-        </Text>
-      </Box>
-    );
+    return <IdeTrustChangeDialog reason={uiState.ideTrustRestartReason} />;
   }
   if (uiState.showWorkspaceMigrationDialog) {
```
jacob314, line 49: can you track what caused the change so this message is clearer. this could be confusing non-ide users won't know what an ide connection is and ide users may be similarly confused.

jacob314 (approved):
lgtm

packages/cli/src/ui/components/DialogManager.tsx
shrutip90: As discussed offline, non IDE users should not hit this path and I have added the two reasons that IDE users would see this restart message.

Issue #7729, opened 2025-09-04
Allow extensions to define how they are configured


chrstnb:
Add support for extensions to provide us with a settings schema to write to env variables or keystore.

chrstnb:
#9802 is related but perhaps not precisely what we want.

PR #9802, opened 2025-09-26
Add extension settings to be requested on install

chrstnb:
## TLDR

Extension authors can add settings to their gemini-extension.json:
{
  "name": "context-example",
  "version": "2.0.0",
  "contextFileName": "GEMINI.md",
  "settings": [
    {
      "name": "API Key",
      "description": "Your API key for the service. Find this at https://gmail.com",
      "envVar": "MY_API_KEY"
    },
    {
      "name": "Other API Key",
      "description": "Your API key for the service.",
      "envVar": "MY_OTHER_API_KEY"
    }
  ]
}

Then, when the user installs the extension, we will prompt them for these values, which get written under the hood to a .env file in the extension's directory:

```
chrstn-mac:gemini-cli chrstn$ gemini extensions install -- --path=hello
Extensions may introduce unexpected behavior.
Ensure you have investigated the extension source and trust the author.
This extension will append info to your gemini.md context using GEMINI.md
Do you want to continue? [Y/n]: 
API Key
Your API key for the service. Find this at https://gmail.com
> 123456780
Other API Key
Your API key for the service.
> abcdefgh
Extension "context-example" installed successfully and enabled.
```

These will be available in the gemini-extension.json file and in scripts.

You can also mark a variable as 'sensitive', in which case the input will be blocked out.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

#7729


chrstnb:
⏰  NOTE: There is a bug in this-- when you go to list the MCP servers in the CLI, they have the color codes around them, so I think this is messing up something with string formatting.

chrstnb:
Another note: 
1. we should (in a follow-up) support checking this on update.
2. We should support optional values
3. We should add a 'configure' command that will allow users to re-set these values

docs/extensions/index.md
```diff
@@ -177,6 +177,44 @@ When Gemini CLI starts, it loads all the extensions and merges their
...
+  ]
+}
+```
+
+When a user installs this extension, they will be prompted to enter their API
+key. The value will be saved to a `.env` file in the extension's directory
+(e.g., `<home>/.gemini/extensions/my-api-extension/.env`).
+
+These will be available to scripts within your extension and also in the
+gemini-extension.json file. They are not available to e.g., custom commands or
```
jakemac53: Why aren't these available to custom commands? How are they exposed to scripts if not through custom commands?

packages/cli/src/config/extensions/extensionSettings.test.ts
```diff
@@ -0,0 +1,172 @@
...
+  default: mockReadline,
+  createInterface: mockReadline.createInterface,
+}));
+
+describe('extensionSettings', () => {
+  let tempHomeDir: string;
+  let extensionDir: string;
+
+  beforeEach(() => {
+    tempHomeDir = os.tmpdir() + path.sep + `gemini-cli-test-home-${Date.now()}`;
```
jakemac53, line 35: A lot of the other tests actually mock out `os.homedir()` to return a temp dir, which allows you to mock fewer other things, might simplify this a bit (you won't have to mock `getExtensionDir` below.

packages/cli/src/config/extensions/extensionSettings.ts
```diff
@@ -0,0 +1,69 @@
...
+  extensionConfig: ExtensionConfig,
+  requestSetting: (setting: ExtensionSetting) => Promise<string>,
+): Promise<void> {
+  const { settings, name: extensionName } = extensionConfig;
+  if (!settings || settings.length === 0) {
+    return;
+  }
+
+  const envFilePath = path.join(
+    new ExtensionStorage(extensionName).getExtensionDir(),
```
jakemac53: We should probably just add a helper for this to ExtensionStorage

packages/cli/src/config/extensions/extensionSettings.ts
```diff
@@ -0,0 +1,69 @@
...
+    '.env',
+  );
+  let envContent = '';
+
+  console.log(''); // for spacing
```
jakemac53: This is baking in an assumption that we are not in the CLI - we wouldn't want to do this there.

Maybe the codepath that uses `readline` should do this?

packages/cli/src/config/extensions/extensionSettings.ts
```diff
@@ -0,0 +1,69 @@
...
+): Promise<string> {
+  if (setting.sensitive) {
+    const response = await prompts({
+      type: 'password',
+      name: 'value',
+      message: `${setting.name}\n${setting.description}`,
+    });
+    return response.value;
+  } else {
+    const readline = await import('node:readline');
```
jakemac53: Is this normal to import something inside the method like this? I haven't seen that elsewhere in the codebase.

packages/cli/src/config/extensions/update.ts
```diff
@@ -71,13 +73,33 @@ export async function updateExtension(
       workspaceDir: cwd,
       extensionEnablementManager,
     });
+
+    const extensionStorage = new ExtensionStorage(extension.name);
```
jakemac53: All this logic should probably exist inside `installOrUpdateExtension`, that function should fully handle safely updating (or installing) extensions without needing to do other logic before/after calling it

packages/cli/src/config/extensions/extensionSettings.ts
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  await fs.writeFile(envFilePath, envContent);
+}
+
+export async function promptForSetting(
+  setting: ExtensionSetting,
+): Promise<string> {
+  if (setting.sensitive) {
+    const response = await prompts({
```
jakemac53: We should probably use this library consistently - for both sensitive prompts and insensitive prompts?

We should also consider directly exposing the schema supported by the prompts package, or some subset? Although this might be hard to replicate in the regular CLI.

jakemac53 (commented):
The main comment I have Gemini also brought up, but the update flow here I think needs some thought - also in general we should be resilient against these `.env` dirs disappearing for any reason, and maybe we can tackle those with the same logic.

For instance, during extension loading if some required settings are missing should we prompt for them? This would require an CLI interface for setting these as well though.

packages/cli/src/config/extension.ts
```diff
@@ -186,7 +194,7 @@ export function loadExtension(
 
   try {
     let customEnv: Record<string, string> = {};
-    const envPath = path.join(effectiveExtensionPath, '.env');
+    const envPath = ExtensionStorage.getEnvFilePath(effectiveExtensionPath);
```
jakemac53: This seems odd, do we actually wan't it to be the effective path or the extension dir? Does the install metadata also end up in the effective dir (actual source dir for linked extensions)?

packages/cli/src/config/extension.ts
```diff
@@ -102,6 +102,14 @@ export class ExtensionStorage {
     return path.join(this.getExtensionDir(), EXTENSIONS_CONFIG_FILENAME);
   }
 
+  getEnvFilePath(): string {
+    return path.join(this.getExtensionDir(), EXTENSION_SETTINGS_FILENAME);
+  }
+
+  static getEnvFilePath(dirPath: string): string {
```
jakemac53: Having both a static and instance version of this with the same name is confusing - I would consider just dropping this static one.

If we really need to hardcode the `path.join` thing for linked extensions I would probably just do that.

packages/cli/src/config/extension.ts
```diff
@@ -400,6 +408,51 @@ async function promptForConsentInteractive(
   });
 }
 
+async function copySettingsToTempDirOnUpdate(
+  extensionDir: string,
+): Promise<string | undefined> {
+  const tempDir = await ExtensionStorage.createTmpDir();
+  const settingsPath = path.join(extensionDir, EXTENSION_SETTINGS_FILENAME);
+  console.log(settingsPath);
```
jakemac53:
```suggestion
```

packages/cli/src/config/extension.ts
```diff
@@ -400,6 +408,51 @@ async function promptForConsentInteractive(
...
+  let tempSettingsPath: string | undefined;
+  if (fs.existsSync(settingsPath)) {
+    tempSettingsPath = path.join(tempDir, EXTENSION_SETTINGS_FILENAME);
+    console.log(tempSettingsPath);
```
jakemac53:
```suggestion
```

packages/cli/src/config/extension.ts
```diff
@@ -400,6 +408,51 @@ async function promptForConsentInteractive(
...
+    await fs.promises.copyFile(settingsPath, tempSettingsPath);
+  }
+  return tempSettingsPath;
+}
+
+async function maybeCopySettingsFromTempDir(
```
jakemac53: nit: add a return type to this

packages/cli/src/config/extension.test.ts
```diff
@@ -1111,6 +1111,124 @@ This extension will run the following MCP servers:
...
+        async (_) => true,
+        process.cwd(),
+        previousExtensionConfig,
+        promptForSettingsMock,
+      );
+
+      expect(promptForSettingsMock).toHaveBeenCalledTimes(1);
+      expect(promptForSettingsMock).toHaveBeenCalledWith(
+        expect.objectContaining({ name: 'New Setting' }),
+      );
```
jakemac53, line 1186: This should also validate that both settings are now set (and we didn't lose the previous one)

jakemac53 (approved):

bdmorgan (approved):

PR #9803, opened 2025-09-26
feat(cli): Add /reset command and modify /clear command

gvking:
## TLDR

   * /clear: This command will clear all the text from your terminal 
     screen, but it will not affect your conversation history. You can 
     still scroll up to see the previous parts of your conversation.
   * /reset: This command will start a new conversation, clearing all the 
     previous messages from the chat history.

## Dive Deeper

Adds a new `/reset` command to the Gemini CLI. This command resets the chat session, providing a quick way to start a fresh conversation.

Also, modifies `/clear` to just make the command clear the screen.

This makes it easier for developers to clear their terminal when wanting to refresh the screen, and when wanting to reset the gemini session, they now have a dedicated command for that.

## Reviewer Test Plan

Running /clear, and /reset.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

### DIFF
--- a/docs/cli/commands.md
+++ b/docs/cli/commands.md
@@ -123,7 +123,7 @@ Slash commands provide meta-level control over the CLI itself.
 - **`/privacy`**
   - **Description:** Display the Privacy Notice and allow users to select whether they consent to the collection of their data for service improvement purposes.
 
-- **`/quit`** (or **` /exit`**)
+- **`/quit`** (or **`/exit`**)
   - **Description:** Exit Gemini CLI.
 
 - **`/reset`**
--- a/docs/cli/commands.md
+++ b/docs/cli/commands.md
@@ -123,8 +123,11 @@ Slash commands provide meta-level control over the CLI itself.
 - **`/privacy`**
   - **Description:** Display the Privacy Notice and allow users to select whether they consent to the collection of their data for service improvement purposes.
 
-- **`/quit`** (or **`/exit`**)
+- **`/quit`** (or **` /exit`**)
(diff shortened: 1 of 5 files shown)


gvking:
The documentation also is not updated, with /clear saying [something](https://github.com/google-gemini/gemini-cli/blob/main/docs/cli/commands.md) that originally aligns with my new CL, but the current implementation of /clear is not what is intended.

danpalmer:
This seems like it would break existing workflows, and importantly in ways that change output and costs for those who are used to `/clear`'s current behaviour.

What about `/clear-screen` instead for clearing the screen, and leaving `/clear` as-is?

That said, what's the motivation for this? What's the use-case for clearing the screen within a Gemini session?

gvking:
Hi Dan, I think the original intention of `/clear` is reflected in the documentation

> Description: Clear the terminal screen, including the visible session history and scrollback within the CLI. The underlying session data (for history recall) might be preserved depending on the exact implementation, but the visual display is cleared.


But sure, I think maybe we could make the new command `/clear-screen` to clear the screen, and then we can update the documentation to make it more clear. Does that LGTM? 

What's the use-case for clearing the screen within a Gemini session?
> Sometimes, I, and friends included, want to clear the screen to just get a fresher slate, while still preserving my conversation history so that Gemini knows what its working on.

PR #9804, opened 2025-09-26
feat: Use PAT for gemini-cli-robot in release workflows

mattKorwel:
Updates the release workflows to use a Personal Access Token (PAT) for the gemini-cli-robot user instead of a GitHub App. This ensures that pull requests are created by the correct user and pass CLA checks.


scripts/releasing/create-patch-pr.js
```diff
@@ -214,7 +214,9 @@ The commit has been created with conflict markers for easier manual resolution.
   }
 
   const prCommand = `gh pr create --base ${releaseBranch} --head ${hotfixBranch} --title "${prTitle}" --body "${prBody}"`;
-  run(prCommand, dryRun);
+  run(prCommand, dryRun, {
+    env: { GH_TOKEN: process.env.GEMINI_CLI_ROBOT_GITHUB_PAT },
+  });
```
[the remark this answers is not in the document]
mattKorwel:
```suggestion
  // Temporarily set the environment variable for the gh command.
  process.env.GH_TOKEN = process.env.GEMINI_CLI_ROBOT_GITHUB_PAT;
  run(prCommand, dryRun);
  // Unset the environment variable to avoid side effects.
  delete process.env.GH_TOKEN;
```

scidomino (approved):

package.json
```diff
@@ -113,7 +113,8 @@
     "wrap-ansi": "9.0.2",
     "cliui": {
       "wrap-ansi": "7.0.0"
-    }
+    },
+    "ansi-regex": "6.2.2"
```
richieforeman: This is a wee bit sus. Is this expected?  AFAIK, one of these ANSI packages was pwned.  Can we assert that this is the correct version?

As a tangent, this specific update seems a little out-of-scope of the current description, but I yield to you for discretion.

.github/workflows/release-promote.yml
```diff
@@ -320,8 +320,7 @@ jobs:
           branch-name: '${{ steps.release_branch.outputs.BRANCH_NAME }}'
           pr-title: 'chore(release): bump version to ${{ needs.calculate-versions.outputs.NEXT_NIGHTLY_VERSION }}'
           pr-body: 'Automated version bump to prepare for the next nightly release.'
-          app-id: '${{ secrets.APP_ID }}'
-          private-key: '${{ secrets.PRIVATE_KEY }}'
+          github-token: '${{ secrets.GEMINI_CLI_ROBOT_GITHUB_PAT }}'
```
richieforeman, line 323: Sanity check: I'm reading about Github PAT tokens and it seems that they can be scoped to specific ACLs and rights.  Just confirming that the ACLs here are tuned for "goldilocks just right", and that we don't have excessive permissions here.

As a tangent, once we stabilize changes on all of this release pipeline stuffs, I think we should join forces and write up some docs around all this stuff.  I feel like we both have a little bit of knowledge that will got lost in the sands of time if we don't jot it down.  I'm absolutely happy to participate

.github/workflows/release-patch-1-create-pr.yml
```diff
@@ -72,7 +63,7 @@ jobs:
         id: 'create_patch'
         env:
           GITHUB_TOKEN: '${{ secrets.GITHUB_TOKEN }}'
-          GH_TOKEN: '${{ steps.generate_token.outputs.token }}'
+          GH_TOKEN: '${{ secrets.GEMINI_CLI_ROBOT_GITHUB_PAT }}'
```
richieforeman, line 66: GITHUB_TOKEN and GH_TOKEN are way too close for me.

It's giving me this meme vibes:

Image

I'd much prefer if this was named `GITHUB_ROBOT_PAT_TOKEN` or something (I don't feel strongly on the specifics of name, but I think we need to be stupidly obviously clear/verbose what this is).

richieforeman: or even `GEMINI_ROBOT_GITHUB_TOKEN` or something.  I'm open, but it needs to be longer and disambiguated from `GITHUB_TOKEN` in a significant way.

richieforeman: Pardon my ignorance here, I'm now realizing that `GH_TOKEN` is potentially read by the `gh` app and it needs to be named this way?  If we're passing that in there, do we even need to pass the `GITHUB_TOKEN`?  Maybe we can remove that?  It's just tripping my spidey sense that both are here in this scope.

richieforeman: I'm reading on and it looks like you unfortunately need to provide both in this context in order for it to work. I defer to you for final here, and recognize that I learned something.  Sorry to litter this CL with a bucket of comments.

package.json
mattKorwel: I went ahead and broke this out into a separate pr https://github.com/google-gemini/gemini-cli/pull/9964

.github/workflows/release-patch-1-create-pr.yml
mattKorwel: #nailed it. This was a good call out, if not predetermined for us changing the variable names would make a lot of sense. as you discovered this is how it has to be.

.github/workflows/release-promote.yml
mattKorwel: yep, locked down to bare minimum. 

👍

bdmorgan (approved):

PR #9941, opened 2025-09-26
feat(ci): Improved docker push flow.

richieforeman:
Use native docker buildx actions to build and push the docker container. I think this is better than the Javascript script version.

REQUESTING APPROVAL HERE so I can `gh run` this in isolation before I try to hook it up to the release process.


mattKorwel (changes requested):
Send me a message and we'll follow up.

PR #9972, opened 2025-09-26
fix(ui): stop truncating output from the model rendered in <static>

### LINKED ISSUE #3396: ... First <x> lines hidden... make code review in terminal problematic
### What happened?

Some code is being hidden, this is not good for code review purposes, it's nice to be able to walk in sync with the AI to see what it's doing, saves a lot of file openings... I can just read the text in the terminal and follow along.

Unfortunately when files or work done gets big gemini cli seems to hide lines of text... this is not so good, maybe there is a way to disable this behaviour ?!

I could miss essential lines where maybe something bad happened, see screenshot. (For example, the AI didn't do a complete implementation and inserted some TODOs, or there were other issues... like missing types/context).

Image

I want to see all lines, all work done by the AI.

I would at least like to see a little hint message how to expand the entire view for at least that particular file and expand it to see all lines of code with some kind of key combo.

For example, control-e for expand and control-shift-e for collapse, and control-a for expand all and control-shift-a to collapse all.

With collapse all I kinda mean 'reduce' so that it shows ... x lines hidden... again so that at least some content is still visible.

### Client information

<latest, busy>

### Login information

_No response_

_No response_

jacob314:
## TLDR

User's couldn't see the full model responses in <Static> requiring annoying workarounds like asking the model to write to temp files.
For other cases, users could use ctrl-S to see the full contents of a diff using ctrl-S before approving.

## Dive Deeper

Change so we have a separate setting for the length we should truncate HistoryItems at if they are from Gemini instead of from shell scripts. We still want to do some truncation of shell script output in the <Static> history as it could be unbounded in size. 

## Reviewer Test Plan

Run the following prompt to get Gemini to output a code block that would previously be truncated:
Read @packages/cli/src/gemini.tsx and add the comment // hello world before each method while writing the content directly into this chat

Verify the output was previously truncated and is not now.

Ask Gemini CLI to run a shell script (e.g. print 10000 lines to stdout using echo) and verify that output is truncated.

## Linked issues / bugs

- Fixes #3396


packages/cli/src/ui/components/PermissionsModifyTrustDialog.test.tsx
```diff
@@ -40,10 +39,6 @@ vi.mock('../../config/trustedFolders.js', () => ({
   },
 }));
 
-vi.mock('../contexts/SettingsContext.js', () => ({
```
jacob314, line 43: cleaned up mocking of Settings

packages/cli/src/ui/utils/MarkdownDisplay.test.tsx
```diff
@@ -4,12 +4,11 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { render } from 'ink-testing-library';
 import { describe, it, expect, vi, beforeEach } from 'vitest';
 import { MarkdownDisplay } from './MarkdownDisplay.js';
 import { LoadedSettings } from '../../config/settings.js';
-import { SettingsContext } from '../contexts/SettingsContext.js';
 import { EOL } from 'node:os';
+import { renderWithProviders } from '../../test-utils/render.js';
```
jacob314, line 11: moved this to the standard mock with providers

packages/cli/src/ui/components/HistoryItemDisplay.test.tsx
```diff
@@ -190,4 +190,84 @@ describe('<HistoryItemDisplay />', () => {
       'echo "\\u001b[31mhello\\u001b[0m"',
     );
   });
+
+  const longCode =
```
jacob314, line 194: only non-mechanical test changes

packages/cli/src/ui/components/HistoryItemDisplay.tsx
```diff
@@ -36,6 +36,7 @@ interface HistoryItemDisplayProps {
   commands?: readonly SlashCommand[];
   activeShellPtyId?: number | null;
   embeddedShellFocused?: boolean;
+  availableTerminalHeightGemini?: number;
```
abhipatel12, line 39: nit: Should we add a max for any abnormally large output that could cause issues?

abhipatel12 (approved):
LGTM, just a question about a defensive max

packages/cli/src/ui/components/HistoryItemDisplay.tsx
jacob314: fair enough. Changed to 64K after discussion. Don't think this should trigger unless there are bugs but reasonable not to allow it to be infinite.

PR #9988, opened 2025-09-26
feat: Add `enableSubagents` configuration and wire up subagent registration

abhipatel12:
## TLDR

This PR introduces the necessary configuration and wiring to enable the experimental "Subagents" feature. It adds a new `enableSubagents` setting, updates the core `Config` to dynamically register subagents as tools when this setting is enabled, and ensures the `MessageBus` is correctly propagated for policy enforcement. It also resolves a circular dependency that was exposed by these changes.

### Configuration Changes
- Added a new boolean setting `experimental.enableSubagents` to the CLI `settingsSchema`. It defaults to `false` and requires a restart to take effect.
- Updated both the CLI and Core `Config` classes to read and expose this setting.

### Subagent Registration
- In `packages/core/src/config/config.ts`, an `AgentRegistry` is now initialized.
- The `createToolRegistry` method has been updated. If `enableSubagents` is `true`, it retrieves all agent definitions from the registry, wraps them in a `SubagentToolWrapper`, and registers them as tools.
- This registration process respects the existing `allowedTools` and `excludeTools` configuration, allowing users to control which subagents are available.

### Message Bus Propagation
- Updated `SubagentToolWrapper` and `SubagentInvocation` to accept an optional `MessageBus` in their constructors.
- The `Config` class now passes the `MessageBus` when creating the `SubagentToolWrapper`. This ensures that any actions taken by the subagents will still participate in the standard confirmation and policy enforcement flow.

### Circular Dependency Fix
The introduction of `AgentRegistry` into `Config` exposed a circular dependency:
`Config` -> `AgentRegistry` -> `CodebaseInvestigatorAgent` -> `GlobTool` -> `Config`.

This was causing `glob.test.ts` to fail. The fix involved decoupling the tool name from the tool implementation:
- Created `packages/core/src/tools/tool-names.ts` to hold tool name constants.
- Updated `GlobTool` to use the constant `GLOB_TOOL_NAME`.
- Updated `CodebaseInvestigatorAgent` to import the name from `tool-names.ts` instead of importing the `GlobTool` class, thus breaking the import cycle.

### 1. Verify Unit Tests
Run preflight and ensure all tests pass

### 2. Verify Configuration Behavior
1.  Start the CLI without modifying settings. Verify that subagents are **not** loaded (you can check `/tools` to see if you see the example codebase investigator).
2.  Open your settings file and add:
    ```json
    "experimental": {
      "enableSubagents": true
    }
    ```
3.  Restart the CLI. Verify that the codebase investigator is being registered now.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  |✅| ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt |✅| -   | -   |

### DIFF
--- a/packages/cli/src/config/config.ts
+++ b/packages/cli/src/config/config.ts
@@ -708,6 +708,7 @@ export async function loadCliConfig(
     useModelRouter,
     enableMessageBusIntegration:
       settings.tools?.enableMessageBusIntegration ?? false,
+    enableSubagents: settings.experimental?.enableSubagents ?? false,
   });
 }
 
--- a/packages/core/src/agents/executor.ts
+++ b/packages/core/src/agents/executor.ts
@@ -370,44 +370,46 @@ export class AgentExecutor {
       return true;
     });
 
-    const toolPromises = validatedFunctionCalls.map(async (functionCall) => {
-      const callId = functionCall.id ?? `${functionCall.name}-${Date.now()}`;
-      const args = functionCall.args ?? {};
+    const toolPromises = validatedFunctionCalls.map(
+      async (functionCall, index) => {
+        const callId = functionCall.id ?? `${promptId}-${index}`;
+        const args = functionCall.args ?? {};
 
-      this.emitActivity('TOOL_CALL_START', {
-        name: functionCall.name,
-        args,
-      });
-
-      const requestInfo: ToolCallRequestInfo = {
-        callId,
-        name: functionCall.name as string,
-        args: args as Record<string, unknown>,
-        isClientInitiated: true,
-        prompt_id: promptId,
-      };
-
-      const toolResponse = await executeToolCall(
-        this.runtimeContext,
-        requestInfo,
-        signal,
-      );
-
-      if (toolResponse.error) {
-        this.emitActivity('ERROR', {
-          context: 'tool_call',
+        this.emitActivity('TOOL_CALL_START', {
           name: functionCall.name,
-          error: toolResponse.error.message,
+          args,
         });
-      } else {
-        this.emitActivity('TOOL_CALL_END', {
-          name: functionCall.name,
-          output: toolResponse.resultDisplay,
-        });
-      }
 
-      return toolResponse;
-    });
+        const requestInfo: ToolCallRequestInfo = {
+          callId,
+          name: functionCall.name as string,
+          args: args as Record<string, unknown>,
+          isClientInitiated: true,
+          prompt_id: promptId,
+        };
+
+        const toolResponse = await executeToolCall(
+          this.runtimeContext,
+          requestInfo,
+          signal,
+        );
+
+        if (toolResponse.error) {
+          this.emitActivity('ERROR', {
+            context: 'tool_call',
+            name: functionCall.name,
+            error: toolResponse.error.message,
+          });
+        } else {
+          this.emitActivity('TOOL_CALL_END', {
+            name: functionCall.name,
+            output: toolResponse.resultDisplay,
+          });
+        }
+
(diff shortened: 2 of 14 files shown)


packages/cli/src/config/config.ts
```diff
@@ -676,6 +676,7 @@ export async function loadCliConfig(
     useModelRouter,
     enableMessageBusIntegration:
       settings.tools?.enableMessageBusIntegration ?? false,
+    enableSubagents: settings.experimental?.enableSubagents ?? false,
```
silviojr, line 711: I think this should be fine for now, but it would be great to have a list where you list the subagents you want to enable.

packages/core/src/config/config.ts
```diff
@@ -1088,6 +1107,41 @@ export class Config {
       registerCoreTool(WriteTodosTool, this);
     }
 
+    // Register Subagents as Tools
+    if (this.getEnableSubagents()) {
+      const agentDefinitions = this.agentRegistry.getAllDefinitions();
+      for (const definition of agentDefinitions) {
+        // We must respect the main allowed/exclude lists for agents too.
+        const excludeTools = this.getExcludeTools() || [];
+        const allowedTools = this.getAllowedTools();
```
silviojr, line 1115: Hmm.. ok. So instead of having a list of enabled agents we are explicit considering them as tools and having the list of allowed/excluded here.

packages/core/src/tools/tool-names.ts
```diff
@@ -0,0 +1,17 @@
...
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+// Centralized constants for tool names.
+// This prevents circular dependencies that can occur when other modules (like agents)
+// need to reference a tool's name without importing the tool's implementation.
+
+export const GLOB_TOOL_NAME = 'glob';
```
silviojr, line 11: I think this is great actually. Then we can have more than one tool that have the same name (but you can have only one of them enabled, of course) and we don't have to change its mentions around the codebase. I had this issue when I was replacing the edit tool.

silviojr (approved):

packages/core/src/agents/executor.ts
```diff
@@ -0,0 +1,578 @@
...
+            `unauthorized tool '${call.name}'. This call has been blocked.`,
+        );
+        return false;
+      }
+      return true;
+    });
+
+    const toolPromises = validatedFunctionCalls.map(async (functionCall) => {
+      const callId = functionCall.id ?? `${functionCall.name}-${Date.now()}`;
+      const args = functionCall.args ?? {};
```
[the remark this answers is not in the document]
abhipatel12: Done

packages/cli/src/config/config.ts
abhipatel12: Yeah that sounds great! I think as we discover a need for it, we can add more properties

packages/core/src/config/config.ts
abhipatel12: Yeah, in the event that a user says they want to disable a subagent, we would need to respect that. For now, since we don't have a special flag --disable-subagent, we can respect it via the tool flag list.

packages/core/src/tools/tool-names.ts
abhipatel12: Yeah exactly! I moved one name over but we can move them over in a different PR one by one since some names touch a lot of files.

PR #9994, opened 2025-09-26
Fix `-e <extension>` for disabled extensions

### LINKED ISSUE #8752: -e/--extensions flag does not override disabledExtensions setting in settings.json

jakemac53:
## TLDR

Fixes an issue where disabled extensions could not be enabled with `-e <extension>`.

This also fixes a bug where the `update` command would fail for disabled extensions, even if you explicitly asked to update an extension by name.

## Dive Deeper

Moves all handling of extension overrides into the ExtensionEnablementManager, which gives us a central place with all the knowledge necessary for handling the answer to "is some extension enabled".

All APIs that used to take the argv extensions argument now just take an ExtensionEnablementManager. This also allows us to load the extension configurations fewer times from disk (if you share that instance).

This did cause a LOT of churn in tests due to changing the public APIs, and the usage is more verbose, but I think better. 

## Reviewer Test Plan

Do `gemini extensions disable <extension>` and then `gemini -e <extension>` and then `/extensions list` in the CLI. You should see the disabled extension is enabled (and no other extensions).

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes #8752


packages/cli/src/config/extensions/extensionEnablement.ts
```diff
@@ -104,24 +105,55 @@ function globToRegex(glob: string): RegExp {
...
+  /**
+   * Determines if an extension is enabled based on its name and the current
+   * path. The last matching rule in the overrides list wins.
+   *
+   * @param extensionName The name of the extension.
+   * @param currentPath The absolute path of the current working directory.
+   * @returns True if the extension is enabled, false otherwise.
+   */
   isEnabled(extensionName: string, currentPath: string): boolean {
+    // If we have a single override called 'none', this disables all extensions.
```
chrstnb, line 143: took me a second to parse this....would clarify that these are args or something?

chrstnb (approved):
Looks good-- one part was a little confusing, but overall makes sense to me

PR #9995, opened 2025-09-26
fix(core): Use shell for spawn on Windows

galdawave:
## TLDR

This pull request fixes issues with spawning processes on Windows by enabling the shell for `spawn` and `spawnSync` calls. This addresses failures when installing the VS Code extension and when opening diffs in VS Code on a Windows environment.

## Dive Deeper

On Windows, `spawn` and `spawnSync` can fail to execute commands that are not `.exe` files, such as the `.cmd` scripts used by VS Code. This is because, without a shell, the system doesn't know how to resolve these file types.

By setting the `shell` option to `true` when the platform is `win32`, we instruct Node.js to use the Windows shell (e.g., `cmd.exe`) to execute the command. The shell can correctly locate and run `.cmd` files, resolving the underlying issue.

This PR applies this fix to two areas:
1.  **VS Code Extension Installation**: The `ide-installer` now uses a shell to run the `code --install-extension` command.
2.  **Opening Diffs**: The `openDiff` utility now uses a shell when launching `code --diff`.

Corresponding tests have been added to validate this behavior on a mocked Windows environment.

## Reviewer Test Plan

To test this, you will need a Windows environment:

1.  Check out this branch.
3.  **Test Extension Installation**:
    -   Ensure VS Code is installed.
    -   Run `npx gemini ide install`.
    -   Verify that the command completes successfully and the Gemini CLI extension is installed in VS Code.
4.  **Test Diff Viewer**:
    -   Set your editor to VS Code: `npx gemini config set editor vscode`.
    -   Generate a change with Gemini that results in a diff (e.g., ask it to modify a file).
    -   When prompted to view the diff, confirm.
    -   Verify that VS Code opens and displays the correct diff.
5.  Run the automated tests: `npm test`. All tests should pass.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ✅  | ✅  |
| npx      | ❓  | ❓ | ❓ |
| Docker   | ❓  | -   | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs
- Fixes: #9755


packages/core/src/ide/ide-installer.ts
```diff
@@ -125,7 +125,7 @@ class VsCodeInstaller implements IdeInstaller {
           'google.gemini-cli-vscode-ide-companion',
           '--force',
         ],
-        { stdio: 'pipe' },
+        { stdio: 'pipe', shell: this.platform === 'win32' },
```
jacob314, line 128: was this intentional?

jacob314 (approved):
We should look at how to integration test these cases.  These mock based tests don't really add confidence other than preventing accidental removal of the arg.

PR #9996, opened 2025-09-26
fix(core): Chat history compression improvements

SandyTao520:
## TLDR

Exclude the initial context messages (the first user message and the first model response) from being compressed. These 2 messages are recreated after compression when we call startChat(), now only the actual conversation history is considered for summarization. Also this skips compression if there is nothing to compress.

## Dive Deeper

The `tryCompressChat` method in `GeminiClient` was previously including the initial two messages of a chat session in its compression calculations. These first two messages represent the initial context provided to the model and its acknowledgment, which should not be compressed.

This change introduces a slice operation that removes the first two messages from the chat history before it's evaluated for compression. This ensures that the summary is only generated for the user's conversational turns, preserving the important initial context.

Additionally, a check has been added to handle cases where no suitable point for compression is found in the conversational history (`splitPoint === 0`), preventing unnecessary processing by returning a `NOOP` status early.

## Reviewer Test Plan

Use debugger or a network proxy to see what was included in the compression request, it should not contain the 2 initial context turns.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ✅  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

#9033


scidomino (commented):
What user-visible bug does this fix? If the user is in the position of wanting to compress their context, this initial context is a tiny rounding error.

I'm just looking for a justification for the (admittedly small) increase in code complexity.

SandyTao520:
Without the change, try this prompt
```
what is in @packages/cli/
```
This would use up most of the context window, 13% left after model's response.

Then, say something else:
```
thanks!
```
This would then trigger a compression, using the first 2 initial context turns because split `splitPoint` is 2. This compression is unnecessary and confuses the model and it returns a hallucinated summary. 
image

packages/core/src/core/client.ts
```diff
@@ -665,7 +672,9 @@ export class GeminiClient {
     // Check if the model needs to be a fallback
     model = getEffectiveModel(this.config.isInFallbackMode(), model);
 
-    const curatedHistory = this.getChat().getHistory(true);
+    const curatedHistory = this.getChat()
```
scidomino: Instead of doing this, could we just change line 728 to:

if(splitPoint === INITIAL_CONTEXT_LENGTH) {

SandyTao520: We can do `if(splitPoint <= INITIAL_CONTEXT_LENGTH) {` but I want to just make sure they never got included to reduce noise for the summarization call.

scidomino: My worry is that the initial context might be important for understanding what is happening. Let's keep it simple and just bail if the split point is 2

scidomino (approved):

scidomino (changes requested):
Actually, I'm confused. In the scenario you give, wouldn't it pick the split point 4?

SandyTao520:
findCompressSplitPoint picked 2 because it is the last user turn before exceeding the 70% limit.

PR #10000, opened 2025-09-26
feat(core): Use lastPromptTokenCount to determine if we need to compress

SandyTao520:
## TLDR

This PR improves the chat history compression logic (`tryCompressChat`) to make it use lastPromptTokenCount to decide if we need to compress.

Previously, compression relied on two `countTokens` API calls to get the token count before and after summarization.  These API calls could be slow or fail.

This change removes those API calls entirely:
1.  The `originalTokenCount` is now read from the `uiTelemetryService`, which caches the last known token count.
2.  The `newTokenCount` (after summarization) is now *estimated* using a string-length heuristic (1 token ≈ 4 characters).

This makes the compression logic more reliable and removes a network dependency, at the trade-off of using an estimated token count instead of an exact one.

## Dive Deeper

The primary motivation for this change was to reduce latency and improve reliability. If the `countTokens` API call failed, compression would be skipped, and the user's large chat history would be sent on the next turn, which could then fail due to exceeding token limits.

By using a cached value for the original count and estimating the new count, we eliminate this failure point.

## Reviewer Test Plan

The goal is to validate that chat compression still triggers correctly and reliably.

1.  Start a new chat session (`gemini -c`).
2.  Have a very long conversation until context used is after 70%. A good way to do this is to include a large directory like packages/cli using @.
3.  Send another message. You should see auto compression works.
4.  Verify that the conversation continues smoothly and the model seems to have a summarized context of the previous conversation.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ✅  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

#9033

### DIFF
--- a/packages/core/src/core/client.ts
+++ b/packages/core/src/core/client.ts
@@ -742,23 +742,13 @@ export class GeminiClient {
     ]);
     this.forceFullIdeContext = true;
 
-    const { totalTokens: newTokenCount } =
-      await this.getContentGeneratorOrFail().countTokens({
-        model,
-        contents: chat.getHistory(),
-      });
-    if (newTokenCount === undefined) {
-      console.warn('Could not determine compressed history token count.');
-      this.hasFailedCompressionAttempt = !force && true;
-      return {
-        originalTokenCount,
-        newTokenCount: originalTokenCount,
-        compressionStatus:
-          CompressionStatus.COMPRESSION_FAILED_TOKEN_COUNT_ERROR,
-      };
-    }
-
-    uiTelemetryService.setLastPromptTokenCount(newTokenCount);
+    // Estimate token count 1 token ≈ 4 characters
+    const compressedHistory = chat.getHistory();
+    const totalChars = compressedHistory.reduce(
+      (total, content) => total + JSON.stringify(content).length,
+      0,
+    );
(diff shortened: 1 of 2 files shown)


packages/core/src/core/client.ts
```diff
@@ -756,23 +742,13 @@ export class GeminiClient {
...
-        originalTokenCount,
-        newTokenCount: originalTokenCount,
-        compressionStatus:
-          CompressionStatus.COMPRESSION_FAILED_TOKEN_COUNT_ERROR,
-      };
-    }
-
-    uiTelemetryService.setLastPromptTokenCount(newTokenCount);
+    // Estimate token count 1 token ≈ 4 characters
+    const newTokenCount = Math.floor(
```
scidomino, line 746: Can't you get the official new token count from the compress call?

SandyTao520: That doesn't include historyToKeep.

scidomino: Good point!

scidomino (approved):

PR #10029, opened 2025-09-27
feat(cli): Ring terminal bell on idle/confirmation

danpalmer:
## TLDR

After long running response streaming, ring the terminal bell. Controlled by a setting, off by default.

## Dive Deeper

This is an accessibility benefit, but has positive impact beyond accessibility. The terminal bell is a lightweight, flexible notification system, requiring minimal integration work in Gemini CLI, and providing various notification options to users in terms of how they want to receive that notification (audible, visual, OS notifications).

I've always found `blaze test ... && ring bell` to be a huge productivity win that helps me context switch to the right place. With Gemini CLI taking minutes to respond, it's likely to be just as beneficial here too.

## Reviewer Test Plan

Run Gemini CLI, enable the "Enable Terminal Bell" setting. Send a request to Gemini and wait for interaction to return to the user. The terminal bell should ring/flash/etc as per terminal configuration.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅   | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

- Resolves b/446772752 – I will update manually as necessary.


danpalmer:
@chrstnb Tagging you for review as you've worked in this area/on accessibility features. Also happy to discuss on corp if preferred.

danpalmer:
I've tried running the integration tests but I get a lot of out of quota errors. I'm unsure what the `npx` tests in the matrix are.

chrstnb (commented):
This seems awesome and looks good to me-- want to get @jacob314's go-ahead on it.

Love stuff like this which is useful for so many people!

chrstnb (approved):

packages/cli/src/config/settingsSchema.ts
```diff
@@ -388,6 +388,15 @@ const SETTINGS_SCHEMA = {
               'Render output in plain-text to be more screen reader accessible',
             showInDialog: true,
           },
+          enableTerminalBell: {
+            type: 'boolean',
+            label: 'Enable Terminal Bell',
```
jacob314, line 580: @miguelsolorio curious on your thoughts on what to name this setting?

packages/core/src/config/config.ts
```diff
@@ -85,6 +85,7 @@ export enum ApprovalMode {
 export interface AccessibilitySettings {
   disableLoadingPhrases?: boolean;
   screenReader?: boolean;
+  enableTerminalBell?: boolean;
```
jacob314: this is a very UI specific setting. Lets avoid adding it to Core if we can.

packages/core/src/config/config.ts
```diff
@@ -928,6 +929,10 @@ export class Config {
     return this.accessibility.screenReader ?? false;
   }
 
+  getEnableTerminalBell(): boolean {
```
jacob314: lets avoid adding this to core

packages/cli/src/config/settingsSchema.ts
```diff
@@ -388,6 +388,15 @@ const SETTINGS_SCHEMA = {
...
+            category: 'UI',
+            requiresRestart: true,
```
jacob314: does this really require a restart?

packages/cli/src/ui/AppContainer.tsx
```diff
@@ -1011,6 +1011,26 @@ Logging in with Google... Please restart Gemini CLI to continue.
     showPrivacyNotice ||
     !!proQuotaRequest;
 
+  const needsBell = useRef(false);
+  useEffect(() => {
+    switch (streamingState) {
+      case StreamingState.Idle:
+      case StreamingState.WaitingForConfirmation:
+        if (needsBell.current) {
+          if (config.getEnableTerminalBell()) {
```
jacob314: should we also track the ms the time we last beeped? one thing I'd like to ensure for this is that we don't beep excessively in this mode. Might be reasonable to only beeped if it had been at least X seconds since the last time we were IDLE or waiting for a confirmation. Would be unfortunate if you got beeps after each accepting edit when the edits took less than a second to apply.

jacob314: also can you read this option of settings instead of config?

jacob314 (commented):
Thanks for adding this feature! it will be useful to have. Sorry about the confusion about config vs settings. existing code isn't as consistent as it should be about that.

packages/cli/src/config/settingsSchema.ts
miguelsolorio: LGTM

packages/core/src/config/config.ts
danpalmer: Oh apologies, so this should be in https://github.com/google-gemini/gemini-cli/blob/main/packages/cli/src/config/config.ts?

packages/cli/src/config/settingsSchema.ts
danpalmer: Experimentally verified – I originally had this `false` but it didn't work without a restart. I assume that's because of the component lifecycle at the callsite, but that's something I don't want to change, at least not for a feature like this, because of how critical it is.

packages/cli/src/ui/AppContainer.tsx
danpalmer: Sorry not sure I understand?

danpalmer: I can do, but I think the fact that we have to switch back into `Responding` in order to reset the bell means that this won't really happen? AFAIU `Responding` means that we're making a call to the model, and that's likely to be some time. I also don't want to reach the case where: user kicks off the model, switches window, model replies too quickly, user doesn't receive the ding, and user sits on Hacker News for 20 minutes.

Maybe we could combine this with focus states? I'm not sure if that's easily accessible here?

If we were to do this, what X would you suggest? I'd suggest 1-2s. And would you be happy with that timeout being set right here in this component/in a local utility, or would you want to pull this out elsewhere?

packages/core/src/config/config.ts
jacob314: no should be in settingsSchema.ts only

packages/cli/src/config/settingsSchema.ts
jacob314: that is fine.

packages/cli/src/ui/AppContainer.tsx
jacob314: 1 second is fine. just don't want a constant stream of bells. you can even always send a bell if the user interacted after the last bell. it is really about avoiding sending a ton of bells is a row when the user is not interacting.

jacob314: remove this and just rely on settings

packages/core/src/config/config.ts
danpalmer: Done.

packages/cli/src/config/settingsSchema.ts
danpalmer: As this is no longer a setting editable in Gemini CLI itself, due to the other changes, this now no longer makes sense. I've put this back to "requiresRestart:false", I also suspect that should it be editable, going via the settings rather than config might mean it _is_ able to change, but yeah, this doesn't really mean anything now.

packages/cli/src/ui/AppContainer.tsx
danpalmer: Done, with tests.

danpalmer: Done

danpalmer:
Code review addressed. Apologies for the PR mess, history should be all good now.

PR #10083, opened 2025-09-27
Jacob314/add radio button keys

jacob314:
## TLDR

Arrows keys for radio buttons didn't work because some radio buttons such as the loop detection radio buttons had values that were not identical. 

## Dive Deeper

The standard react solution is to require keys for all radio button items allowing React to safely update the selected item to the corresponding item with the same key if the available radio buttons changed or appeared to change.

We could have alternately used deep equality on values but that is a non-standard react pattern so was simpler to just use keys even though that introduces a number of boilerplate lines with keys that are very similar to either labels or value.

## Reviewer Test Plan

Test various radio buttons. For example, settings, themes, auth dialog, and loop detection.
Verify that arrow keys work to change selection in all of them.

## Linked issues / bugs

- Fixes #9968

### DIFF
--- a/packages/cli/src/ui/IdeIntegrationNudge.tsx
+++ b/packages/cli/src/ui/IdeIntegrationNudge.tsx
@@ -50,6 +50,7 @@ export function IdeIntegrationNudge({
         userSelection: 'yes',
         isExtensionPreInstalled,
       },
+      key: 'Yes',
     },
     {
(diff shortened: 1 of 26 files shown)


abhipatel12 (approved):
LGTM! 

Tested a few dialogs before and after this PR and it fixed all of them!

PR #10084, opened 2025-09-28
fix(core): refresh file contents in smart edit given newer edits from user/external process

agmsb:
## TLDR

This PR makes a change to improve Smart Edit's behavior around a common loss pattern.

It does so by adding a check during `attemptSelfCorrection` that refreshes the file content _if we find that either the user has edited the file since the agent last did so_. This can help reduce `EDIT_NO_OCCURRENCE` errors due to stale content that the model is using to perform the edit.

## Dive Deeper

Adding the refreshing of file contents increases the chance of a successful edit while also preventing the model from overwriting more recent user changes.

## Reviewer Test Plan

1. Run npx https://github.com/google-gemini/gemini-cli#agmsb/remove-no-changes-llm-judgment
2. Work in a test directory. Create a new file (IE test.txt). Ask the model to write a poem in that file.
3. Make changes to that poem in an editor.
4. Ask the model to make changes that include your changes from the editor.
5. Ensure the model respects those changes in edit and does not revert them.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | x  | ❓  | ❓  |
| npx      | x  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes #10133.
Progress on #4137, #9030, #9770.


packages/core/src/utils/llm-edit-fixer.ts
```diff
@@ -25,14 +25,13 @@ You will be given:
...
 2.  **Explain the Fix:** Your \`explanation\` MUST state exactly why the original \`search\` failed and how your new \`search\` string resolves that specific failure. (e.g., "The original search failed due to incorrect indentation; the new search corrects the indentation to match the source file.").
 3.  **Preserve the \`replace\` String:** Do NOT modify the \`replace\` string unless the instruction explicitly requires it and it was the source of the error. Your primary focus is fixing the \`search\` string.
-4.  **No Changes Case:** CRUCIAL: if the change is already present in the file,  set \`noChangesRequired\` to True and explain why in the \`explanation\`. It is crucial that you only do this if the changes outline in \`replace\` are alredy in the file and suits the instruction!! 
-5.  **Exactness:** The final \`search\` field must be the EXACT literal text from the file. Do not escape characters.
+4.  **Exactness:** The final \`search\` field must be the EXACT literal text from the file. Do not escape characters.
```
silviojr: I think you should also add "do not escape characters" to the "replace" string.

silviojr (approved):

packages/core/src/tools/smart-edit.ts
```diff
@@ -46,6 +51,62 @@ interface ReplacementResult {
...
+        id = part.functionResponse.id;
+        content = part.functionResponse.response;
+      }
+
+      // Normalize paths to handle Windows vs. POSIX differences before comparison.
+      const normalizedContent = JSON.stringify(content).replace(/\\\\/g, '/');
+      if (id && normalizedContent.includes(filePath.replace(/\\/g, '/'))) {
+        return getTimestampFromFunctionId(id);
+      }
+    }
```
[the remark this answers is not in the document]
agmsb: Added this change.

agmsb:
For posterity, noting that we reduced the scope of this PR to just refreshing file contents. We will explore altering LLM judgement in a later PR.

packages/core/src/tools/smart-edit.ts
```diff
@@ -374,11 +440,33 @@ class EditToolInvocation implements ToolInvocation<EditToolParams, ToolResult> {
...
+      this.config.getGeminiClient(),
+    );
+
+    // Add a 2-second buffer to account for timing inaccuracies. If the file
+    // was modified more than a second after the last edit tool was run, we
+    // can send the LLM edit fixer a custom message and the newest file content.
+    if (lastEditedByUsTime > 0) {
+      const stats = fs.statSync(params.file_path);
+      const diff = stats.mtimeMs - lastEditedByUsTime;
+      if (diff > 2000) {
```
mrcabbage972: please move to a constant definition

packages/core/src/tools/smart-edit.ts
```diff
@@ -374,11 +440,33 @@ class EditToolInvocation implements ToolInvocation<EditToolParams, ToolResult> {
...
+    let errorForLlmEditFixer = initialError.raw;
+    const lastEditedByUsTime = await findLastEditTimestamp(
+      params.file_path,
+      this.config.getGeminiClient(),
+    );
+
+    // Add a 2-second buffer to account for timing inaccuracies. If the file
+    // was modified more than a second after the last edit tool was run, we
+    // can send the LLM edit fixer a custom message and the newest file content.
+    if (lastEditedByUsTime > 0) {
```
mrcabbage972: What happens if there are false positives? Is it possible to handle this without using a timer? For example, by hashing the previous edit?

agmsb: In the case of a false positive (we think the content is new, but it's not, and we reread the content - correct me if I misunderstand the case), then we pass in the same contents we would have otherwise, right?

As far as hashing - we would need to hash the entire file content after the edit is completed, right?

mrcabbage972: Yes, after each edit is completed. That would give us a definite answer as to whether the file version is correct.

agmsb: I tried a different approach that may be simpler. Given we perform an initial read of the content during our first edit attempt, we hash that content and check against that hash by reading again during LLM fixing time. Let me know what you think.

mrcabbage972 (approved):

Issue #8077, opened 2025-09-09
Running /settings removed commented code and replace env vars in settings.json


ksprashu:
### What happened?

I ran /settings and made some changes. I then reloaded gemini by accepting the changes (pressing r). 

The next time I checked settings.json, all my commented out MCP servers were removed, and any env vars that I was using were replaced with the actual key values. 

Looks like the /settings command is updating / touching parts of the settings.json file that it should be. Only the updated / changed settings from this command should be inserted rather than rewriting the whole file. 

Eg settings.json (extract)

```
  "mcpServers": {
    "context7": {
      "httpUrl": "https://mcp.context7.com/mcp",
      "headers": {
        "CONTEXT7_API_KEY": "$CONTEXT7_API_KEY"
      }
    },
    // "github": {
    //   "httpUrl": "https://api.githubcopilot.com/mcp/",
    //   "headers": {
    //     "Authorization": "Bearer $GITHUB_PERSONAL_ACCESS_TOKEN"
    //   }
    // },
    "time": {
      "command": "uvx",
      "args": [
        "mcp-server-time"
      ]
    },
```

It removed the github mcp server and replace my context7 api key in plain text. 
Leaking API keys this way would be really bad (if I were presenting publicly) had I not noticed. 

### What did you expect to happen?

Only add / change settings key-values that have been modified. Don't rewrite the whole settings.json file. 

### Client information

<details>
<summary>Client Information</summary>

Run `gemini` to enter the interactive CLI, then run the `/about` command.

```console
> /about

│ About Gemini CLI                                                                                                  │
│                                                                                                                   │
│ CLI Version                            0.3.4                                                                      │
│ Git Commit                             038f004f                                                                   │
│ Model                                  gemini-2.5-pro                                                             │
│ Sandbox                                no sandbox                                                                 │
│ OS                                     darwin                                                                     │
│ Auth Method                            gemini-api-key
```

</details>

### Login information

API Key

ksprashu:
This issue is easily reproducible. 

1) Include a mcp server config, and comment it out. 
If you'd like include a mcp server config with a header and point to an env var. 

2) Open gemini and type /settings to open the setting screen
Toggle any setting and even toggle it back if you like. 
(It can be any setting that doesn't need a reload - 'r'. In case of a setting that needs a reload, then settings.json will be changed when you hit 'r')

3) Hit Esc to come back to Gemini CLI and run `/quit` to leave. 

View the contents of settings.json now. 

[...]

PR #10098, opened 2025-09-28
fix(mcp): fix MCP server removal not persisting to settings

### LINKED ISSUE #9766: MCP Removal not working correctly
### What happened?

```
> gemini mcp remove playwright
Server "playwright" not found in project settings.

> gemini mcp remove playwright -s user
Server "playwright" removed from user settings.

> gemini mcp list

...playwright - Connected
```
It reported removed every time, but the mcp list always showed it there

The MCP server to not be on the list 

(Replacing it with chrome-devtools is why)

### Client information

<details>

```console
╭────────────────────────────────────────╮
│                                        │
│ About Gemini CLI                       │
│                                        │
│ CLI Version  0.6.1                     │
│ Git Commit   f3078136                  │
│ Model        gemini-2.5-pro            │
│ Sandbox      no sandbox                │
│ OS           darwin                    │
│ Auth Method  OAuth                     │
│                                        │
╰────────────────────────────────────────╯
```

</details>

### Login information

Google Account

Upgraded from a prev version about a month or so ago, unsure which version, that's about when I added playwright.

I had it installed on MacOS host AND Windows under parallels, both reported the same success message but neither actually removed anything.

lifefloating:
## TLDR

Fixed issue where `gemini mcp remove` command would report success but the MCP server would still appear in `gemini mcp list`. The problem was in the JSON settings update logic not properly removing keys from nested objects like `mcpServers`.

## Dive Deeper

The issue was in `commentJson.ts` where the `applyUpdates` function wasn't properly handling key deletion for nested objects. When removing an MCP server, the settings update would merge the new object but wouldn't delete keys that were no longer present in the desired state.

Key changes:
- Added `applyKeyDiff` function that properly handles object key deletion by first removing keys not present in the desired state
- Special handling for `mcpServers` field to ensure proper key removal
- Enhanced array handling to preserve comments while updating content in-place
- Added comprehensive tests covering various deletion scenarios

The fix ensures that when an MCP server is removed from settings, it's actually deleted from the configuration file rather than just being reported as removed.

## Reviewer Test Plan

1. Add an MCP server: `node bundle/gemini.js mcp add playwright`
2. Verify it appears: `node bundle/gemini.js mcp list`
3. Remove the server: `node bundle/gemini.js mcp remove playwright -s user`
4. Verify it's gone: `node bundle/gemini.js mcp list` (should not show playwright)
5. Check the settings file to confirm the server entry is actually removed

Test with both project (`-s project`) and user (`-s user`) settings scopes.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |

## Linked issues / bugs

Fixes #9766


packages/cli/src/utils/commentJson.ts
```diff
@@ -44,8 +44,69 @@ function applyUpdates(
...
+      if (isValObj && isResultObj) {
+        applyKeyDiff(
+          result[key] as Record<string, unknown>,
+          value as Record<string, unknown>,
+        );
+      } else {
+        result[key] = value;
+      }
+      continue;
+    }
```
Lyonk71: It looks like you're being surgical by special-casing `mcpServers`, but this bug isn't unique to that key. Consider using `applyKeyDiff` universally in `applyUpdates` and removing the special case. That keeps the utility generic, prevents future special cases, and enforces consistent "sync by omission" (add/update/delete) semantics across all nested objects. Tests that assume merge-only behavior would need updates, and we should verify no callers rely on preserving unspecified keys.

Lyonk71 (changes requested):

packages/cli/src/utils/commentJson.ts
jacob314: +1 to handling this generically rather than for just mcpServers.

lifefloating: Okay, thank you for your review. You're absolutely right. My change was only limited to the mcpServers, which was a bit too cautious.

lifefloating:
hi, I have re-adjusted the code, and here are some notes I've organized. Here, we need to delete the normal parts while keeping the annotated JSON sections. Regarding the issue with annotated JSON, someone has raised it before: https://github.com/google-gemini/gemini-cli/issues/8077

### **Unified sync semantics**
- Removed the special-case for `mcpServers`. All nested objects now follow "sync-by-omission" (add/update/delete) consistently.

### **Preserve comments on key deletion**
Before deleting a property, we migrate its comments so they aren't lost:
- Prefer moving to the next sibling's `before` comments.
- If no next sibling, move to the previous sibling's `after` comments.
- If no siblings, attach to the container-level `before/after`.
- This ensures commented-out sibling blocks remain (e.g., removing `mcpServers.playwright` does not remove the commented `sleep` block).

### **Arrays keep comments**
Arrays are updated in-place (clear + push) so array-level comments are preserved.

### **Safe failure behavior**
If parsing fails, we log errors and do not write back, avoiding content loss.

### **Test coverage**
- **Added**: deleting a key preserves sibling-level commented blocks.
- **Updated**: nested object sync-by-omission expectations.
- **Retained/covered**: basic comment preservation, complex scenarios, corrupted JSON handling, arrays, and type conversion.

packages/cli/src/utils/commentJson.ts
```diff
@@ -38,12 +38,115 @@ export function updateSettingsFilePreservingFormat(
...
+ *
+ * This function re-attaches comments to the next sibling's leading comments if
+ * available, otherwise to the previous sibling's trailing comments, otherwise
+ * to the container's leading/trailing comments.
+ */
+function preserveCommentsOnPropertyDeletion(
+  container: Record<string, unknown>,
+  propName: string,
+): void {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
galdawave: Can we add typing for this, if the target is the same type as container we can set it to `Record<string, unknown>`

packages/cli/src/utils/commentJson.ts
```diff
@@ -38,12 +38,115 @@ export function updateSettingsFilePreservingFormat(
...
+  }
+}
+
 function applyUpdates(
   current: Record<string, unknown>,
   updates: Record<string, unknown>,
 ): Record<string, unknown> {
   const result = current;
 
+  function applyKeyDiff(
```
galdawave: Nit: for readability can we move this out of the applyUpdates function

packages/cli/src/utils/commentJson.ts
```diff
@@ -38,12 +38,115 @@ export function updateSettingsFilePreservingFormat(
...
+    base: Record<string, unknown>,
+    desired: Record<string, unknown>,
+  ): void {
+    for (const existingKey of Object.getOwnPropertyNames(base)) {
```
galdawave: It looks like there's a difference in how top-level properties and nested properties are updated. Omitted keys get deleted from nested objects (a sync), but at the top level, they're preserved (a merge).

For example, if I have a config like 
```
{ "top": 1, "nested": { "a": 2, "b": 3 } } 
```
and I apply an update like
```
{ "nested": { "a": 2 } }
``` 
the nested.b key is removed, but the top key is kept. I think this could lead to some subtle bugs if a caller expects the same behavior everywhere.

I think we should make applyKeyDiff logic consistent for both nested and not nested objects.

galdawave (changes requested):

packages/cli/src/utils/commentJson.test.ts
```diff
@@ -178,5 +181,146 @@ describe('commentJson', () => {
...
+        mcpServers: {},
+      });
+
+      const updatedContent = fs.readFileSync(testFilePath, 'utf-8');
+      expect(updatedContent).toContain('// "sleep": {');
+      expect(updatedContent).toContain('"mcpServers"');
+      expect(updatedContent).not.toContain('"playwright"');
+    });
+
+    it('should handle type conversion from object to array', () => {
```
galdawave, line 307: one last thing :), can you add a test to validate that both nested and non nested objects will be removed

galdawave (approved):

packages/cli/src/utils/commentJson.test.ts
lifefloating: ok  later i add

lifefloating: ok i added it @galz10

jackwotherspoon (commented):
@lifefloating I think we also probably want to update the `gemini mcp remove` tests so that they would catch this kind of regression in the future. As right now the command is broken and there are no tests indicating that... which means the tests may be a bit shallow.


Tests in question: https://github.com/google-gemini/gemini-cli/blob/main/packages/cli/src/commands/mcp/remove.test.ts

I can do this as a follow-up PR if needed but this PR is fixing the broken command so would be nice to have a test that fails on main and is passing against this branch 😄

lifefloating:
> @lifefloating I think we also probably want to update the `gemini mcp remove` tests so that they would catch this kind of regression in the future. As right now the command is broken and there are no tests indicating that... which means the tests may be a bit shallow.
> 
> Tests in question: https://github.com/google-gemini/gemini-cli/blob/main/packages/cli/src/commands/mcp/remove.test.ts
> 
> I can do this as a follow-up PR if needed but this PR is fixing the broken command so would be nice to have a test that fails on main and is passing against this branch 😄

  Good point! I'll add an integration test for `gemini mcp remove` that:

  1. Creates a real settings file with multiple MCP servers
  2. Runs the remove command
  3. Verifies the file content after deletion (server removed, others preserved, comments kept)

  This will indeed catch regressions where the command appears to work but doesn't actually modify the file.

  I'll push this update shortly.

lifefloating:
> @lifefloating I think we also probably want to update the `gemini mcp remove` tests so that they would catch this kind of regression in the future. As right now the command is broken and there are no tests indicating that... which means the tests may be a bit shallow.
> 
> Tests in question: https://github.com/google-gemini/gemini-cli/blob/main/packages/cli/src/commands/mcp/remove.test.ts
> 
> I can do this as a follow-up PR if needed but this PR is fixing the broken command so would be nice to have a test that fails on main and is passing against this branch 😄

thanks, i have updated it

jackwotherspoon (approved):

jacob314 (approved):

PR #10121, opened 2025-09-28
fix(ux) keyboard input hangs while waiting for keyboard input.

### LINKED ISSUE #7936: Irregular Keystroke Response in Gemini CLI
### What happened?

    1 **Bug Report: Irregular Keystroke Response in Gemini CLI**
    2 
    3 **1. Environment:**
    4 *   **Gemini CLI Version:** 0.3.4
    5 *   **Operating System:** linux
    6 *   **Terminal Emulator:** [Please specify your terminal, e.g., GNOME Terminal, Konsole, xterm]
    7 *   **Shell:** [Please specify your shell, e.g., bash, zsh]
    8 
    9 **2. Description of Bug:**
   10 When typing in the Gemini CLI input prompt, meta keys such as Backspace, Delete, End, and Enter do not respond consistently. These keys often need to be pressed multiple times (2-10 times) in a row
      to register a single action. This makes editing commands and submitting prompts very difficult.
   11 
   12 **3. Steps to Reproduce:**
   13 1.  Start the Gemini CLI.
   14 2.  Type any text into the input prompt (e.g., "hello world").
   15 3.  Attempt to delete a character using the Backspace key.
   16 4.  Observe that the keypress may not register.
   17 5.  Press the Backspace key repeatedly until the character is deleted.
   18 6.  Repeat with other keys like Delete, End, and Enter.
   19 
   20 **4. Expected Behavior:**
   21 Each press of a meta key (Backspace, Delete, End, Enter) should result in one corresponding action immediately and consistently.
   22 
   23 **5. Actual Behavior:**
   24 The key presses are registered irregularly, requiring multiple presses to trigger a single action. The input feels laggy and unresponsive, making the CLI debilitating to use for any non-trivial
      input.

Key strokes to edit in the text box like any other IDE.

### Client information

* **CLI Version:** 0.3.4
* **Git Commit:** 2acadeca
* **Session ID:** 79e0aeae-0b66-4e9d-8d75-471ddbd02d24
* **Operating System:** linux v22.18.0
* **Memory Usage:** 226.7 MB
* **IDE Client:** VS Code

### Login information

oAUTH

I'm on zsh 5.9 in Ubuntu 24.04/WSL if that makes a difference.

Stomping these bugs is probably super tough. Gratitude for the team!

jacob314:
## TLDR

We would wait for addition output to complete a kitty protocol response rather than timing out if a kitty protocol response wasn't received in a small number of ms.

This is a problem as some terminals could send responses that looked like kitty protocol but weren't resulting in Google users reporting that Gemini CLI was hanging or not responding to their input.

## Dive Deeper

There are too many cases to enumerate for why terminals could send responses starting with escape codes that looked like they might be kitty but were not kitty.
We now timeout in 50ms if a complete kitty protocol sequence is not sent and abort early if we detect escape sequences that cannot possibly complete as a valid kitty protocol sequence.

## Reviewer Test Plan

Test with iterm2 (terminal google user had an issue with) and other terminals. Test over ssh to make sure the 50ms threshold is not too aggressive. We might need to increase it to 100ms.
Verify that all keys pressed in the terminal are registered successfully. In particular test cases like shift+enter and test performing actions such as entering and leaving the terminal, triggering custom terminal scroll modes and custom keyboard mouse modes to trigger cases that might result in escape codes gemini cli doesn't handle being added to the terminal.

Fixes: #7936

### DIFF
--- a/packages/cli/src/ui/contexts/KeypressContext.test.tsx
+++ b/packages/cli/src/ui/contexts/KeypressContext.test.tsx
@@ -13,6 +13,7 @@ import {
   KeypressProvider,
   useKeypressContext,
   DRAG_COMPLETION_TIMEOUT_MS,
+  KITTY_SEQUENCE_TIMEOUT_MS,
   // CSI_END_O,
   // SS3_END,
   SINGLE_QUOTE,
@@ -70,7 +71,7 @@ describe('KeypressContext - Kitty Protocol', () => {
     children: React.ReactNode;
     kittyProtocolEnabled?: boolean;
   }) => (
-    <KeypressProvider kittyProtocolEnabled={kittyProtocolEnabled}>
+    <KeypressProvider kittyProtocolEnabled={kittyProtocolEnabled ?? false}>
       {children}
     </KeypressProvider>
   );
@@ -476,7 +477,7 @@ describe('KeypressContext - Kitty Protocol', () => {
 
       expect(consoleLogSpy).toHaveBeenCalledWith(
         '[DEBUG] Kitty buffer accumulating:',
-        expect.stringContaining('\x1b[27u'),
+        expect.stringContaining('"\\u001b[27u"'),
       );
       const parsedCall = consoleLogSpy.mock.calls.find(
         (args) =>
@@ -484,7 +485,7 @@ describe('KeypressContext - Kitty Protocol', () => {
           args[0].includes('[DEBUG] Kitty sequence parsed successfully'),
       );
       expect(parsedCall).toBeTruthy();
-      expect(parsedCall?.[1]).toEqual(expect.stringContaining('\x1b[27u'));
+      expect(parsedCall?.[1]).toEqual(expect.stringContaining('\\u001b[27u'));
     });
 
     it('should log kitty buffer overflow when debugKeystrokeLogging is true', async () => {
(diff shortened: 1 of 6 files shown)


packages/cli/src/ui/contexts/KeypressContext.test.tsx
```diff
@@ -753,8 +754,16 @@ describe('Drag and Drop Handling', () => {
...
-  const wrapper = ({ children }: { children: React.ReactNode }) => (
-    <KeypressProvider kittyProtocolEnabled={true}>{children}</KeypressProvider>
+  const wrapper = ({
+    children,
+    kittyProtocolEnabled,
+  }: {
+    children: React.ReactNode;
+    kittyProtocolEnabled?: boolean;
+  }) => (
+    <KeypressProvider kittyProtocolEnabled={kittyProtocolEnabled ?? false}>
```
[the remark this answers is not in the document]
jacob314: fixed this.

packages/cli/src/ui/contexts/KeypressContext.tsx
```diff
@@ -134,9 +135,39 @@ export function KeypressProvider({
     let isPaste = false;
     let pasteBuffer = Buffer.alloc(0);
     let kittySequenceBuffer = '';
+    let kittySequenceTimeout: NodeJS.Timeout | null = null;
     let backslashTimeout: NodeJS.Timeout | null = null;
     let waitingForEnterAfterBackslash = false;
 
+    // Check if a buffer could potentially be a valid kitty sequence or its prefix
+    const couldBeKittySequence = (buffer: string): boolean => {
```
scidomino, line 172: The enclosing useEffect is pretty big and hard to follow. Consider pulling couldBeKittySequence() (and any other self-contained functions) out to a top level function instead of being defined here.

scidomino (approved):

packages/cli/src/ui/contexts/KeypressContext.tsx
jacob314: Agreed. I'll follow up with PR to refactor this to be easier to follow.

PR #10178, opened 2025-09-29
Regex Search/Replace for Smart Edit Tool

mrcabbage972:
## TLDR

This PR enhances the SmartEditTool by adding a more powerful, regex-based text replacement strategy. This new logic acts as a third fallback, improving the tool's ability to find and replace code when the old_string has minor whitespace differences from the target content (e.g., extra spaces between keywords or parameters).

## Dive Deeper

The existing SmartEditTool employed two matching strategies: an exact literal match and a "flexible" match that trimmed whitespace from the beginning and end of each line. While the flexible match helped with indentation differences, it would still fail if there were variations in whitespace within a line. For example, it could not match function  myFunc( a, b ) if the old_string was function myFunc(a, b).

This PR addresses that limitation by introducing a third strategy:

* Tokenization & Regex Construction: The search string is tokenized, with special care taken for common code delimiters like (), [], {}, etc. These tokens are then used to build a regular expression where \s* (zero or more whitespace characters) separates each token.

* Whitespace Insensitivity: This approach makes the match insensitive to the amount of whitespace between tokens on a line, resolving the weakness of the previous flexible matching logic.

* Indentation Preservation: The regex is anchored to the start of a line (^) and captures the leading whitespace. This captured indentation is then prepended to every line of the new_string, ensuring that the replacement block maintains correct formatting within the file.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

### DIFF
--- a/packages/core/src/tools/smart-edit.ts
+++ b/packages/core/src/tools/smart-edit.ts
@@ -59,6 +59,15 @@ function restoreTrailingNewline(
   return modifiedContent;
 }
 
+/**
+ * Escapes characters with special meaning in regular expressions.
+ * @param str The string to escape.
+ * @returns The escaped string.
+ */
+function escapeRegex(str: string): string {
+  return str.replace(/[.*+?^${}()|[\]\\]/g, '\\$&'); // $& means the whole matched string
+}
+
 async function calculateExactReplacement(
   context: ReplacementContext,
 ): Promise<ReplacementResult | null> {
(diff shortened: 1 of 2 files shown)


bdmorgan (approved):

packages/core/src/tools/smart-edit.ts
```diff
@@ -146,6 +155,68 @@ async function calculateFlexibleReplacement(
...
+  context: ReplacementContext,
+): Promise<ReplacementResult | null> {
+  const { currentContent, params } = context;
+  const { old_string, new_string } = params;
+
+  // Normalize line endings for consistent processing.
+  const normalizedSearch = old_string.replace(/\r\n/g, '\n');
+  const normalizedReplace = new_string.replace(/\r\n/g, '\n');
+
+  // This logic is ported from your Python implementation.
```
anj-s, line 168: remove this line?

anj-s (approved):

agmsb (approved):

PR #10182, opened 2025-09-29
feat(release): run evaluation before for nightly, patch and manual releases

### LINKED ISSUE #9059: Infra: Eval's with releases

galdawave:
## TLDR

Refactors the release workflows to run evaluations on the release branch before publishing. This is achieved by splitting the build and publish steps into separate actions and inserting an evaluation trigger in between.

## Dive Deeper

*   Created a new composite action `.github/actions/build-release` that handles creating the release branch, bumping versions, and building packages.
*   Updated `.github/actions/publish-release` to only handle publishing to npm and creating the GitHub release. It now takes the release branch name as an input.
*   Updated `eval.yml` to accept a `branch-or-commit` input, allowing it to be triggered on a specific ref.
*   Updated `release-manual.yml`, `release-nightly.yml`, and `release-patch-3-release.yml` to:
    1.  Call the new `build-release` action.
    2.  Trigger the `eval.yml` workflow on the newly created release branch (skipped in dry runs).
    3.  Call the updated `publish-release` action, passing the branch name.
*   This ensures that we run evaluations on the exact code that is about to be released.

## Reviewer Test Plan

Since this PR modifies release workflows, end-to-end testing is difficult without actually performing a release. However, reviewers can:
1.  Inspect the code changes to ensure the logic is sound.
2.  Verify that the steps moved from `publish-release` to `build-release` are correct.
3.  Check that the `eval.yml` trigger passes the correct branch name.
4.  Optionally, run a dry-run of one of the release workflows (e.g., `release-manual.yml`) and verify that the `build-release` action runs successfully and that the "Trigger Evaluation" step is skipped as expected.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


.github/actions/publish-release/action.yml
```diff
@@ -52,57 +54,6 @@ inputs:
 runs:
   using: 'composite'
   steps:
-    - name: '📝 Print Inputs'
```
mattKorwel, line 58: We've/I've found this super useful for debugging lets add this back in, i'm trying to make sure it is in each composite step actually

.github/workflows/release-manual.yml
```diff
@@ -85,6 +85,23 @@ jobs:
...
+          dry-run: '${{ github.event.inputs.dry_run }}'
+          working-directory: './release'
+
+      - name: 'Trigger Evaluation'
+        if: "${{ github.event.inputs.dry_run == 'false' }}"
+        env:
+          GH_TOKEN: '${{ secrets.GITHUB_TOKEN }}'
+        shell: 'bash'
+        run: |
+          gh workflow run eval.yml --ref ${{ steps.build.outputs.branch-name }} -f branch-or-commit=${{ steps.build.outputs.branch-name }} -f dry-run=true
```
mattKorwel: I think this should be a composite action call like the other steps.

What are the pros and cons for calling gh workflow?

Does the service account have permission to run the workflow this way? If so is it intended to be blocking? Does it provide any feedback on success or failure to this workflow?

.github/workflows/eval.yml
```diff
@@ -2,6 +2,26 @@ name: 'Eval'
 
 on:
   workflow_dispatch:
+    inputs:
```
mattKorwel, line 5: Since this is used in multiple places, this should be a composite action. if we want to also run it stand alone, this workflow should just proxy to that composite.

.github/actions/build-release/action.yml
```diff
@@ -0,0 +1,76 @@
...
+    required: false
+    default: '.'
+
+outputs:
+  branch-name:
+    description: 'The name of the release branch.'
+    value: '${{ steps.release_branch.outputs.BRANCH_NAME }}'
+
+runs:
+  using: 'composite'
```
mattKorwel, line 26: The steps for creating and pushing the branch should not be a part of the build step. this is release specific. we already have a branch that was passed into the workflow that is what we should run eval's against.

.github/workflows/release-manual.yml
mattKorwel: Also, I think this, "Run Evaluations", should be part of the publish release step.

mattKorwel (changes requested):
hey @galz10 this is a great start that brings together a lot of moving parts 🙇 thanks for diving in.

I've left a few comments on structure and organization. Also any of the workflows that we are adding or removing steps to or touching the yml on need to be validated end to end with dry run's as well. We've seen a lot of seemingly harmless changes catch us off guard.

.github/actions/publish-release/action.yml
galdawave: Added back.

.github/workflows/eval.yml
galdawave: Done

mattKorwel (changes requested):
Hey there I think things have changed quite a bit in the last two weeks yet again. We probably need to revisit this. One the sub action shouldn't be doing a checkout. Two I'm not actually following why we need to build release as a separate step with that does or doesn't do for us. Either way we should get together and revisit this.

PR #10215, opened 2025-09-30
Refactor metrics definitions to be easily understandable

### LINKED ISSUE #9209: Refactor(telemetry): Refactor Metrics
As a prerequisite for supporting versioned telemetry conventions (legacy vs. v1), we need to refactor the telemetry implementation. ~This involves abstracting metric and log conventions into dedicated objects ("Convention Sets") rather than hardcoding them.~ 

The work here will be instead to reorganize the telemetry logic such that it is easier to compare the telemetry currently being emitted with the Convention Set standard and determine what to add.

**Parent Issue:** #9208

**Tasks:**
- [ ] Update `metrics.ts` such that the complete structure of each emitted metric is co-located within the file
- [ ] Verify that all existing telemetry tests still pass after the refactoring.

bobcatfish:
## TLDR

As part of #9209 in preparation for adding / modifying telemetry to comply with the convention set architecture, move all the definitions into one place so we can see exactly what metrics are are exposing and what their structure is.

## Dive Deeper

Not quite sure that the initialization of the peformance based metrics actually needs to be done separately from the rest, but keeping it that way for now to be consistent.

(FYI @eLyiN ! I don't _think_ this will cause too much trouble with your pending changes 🙏 )

## Reviewer Test Plan

If you wanted, you could hook up telemetry and try it out (https://github.com/google-gemini/gemini-cli/blob/main/docs/telemetry.md) but we should have pretty good coverage via the unit + integration tests.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ✅   |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Part of #9209

### DIFF
--- a/packages/core/src/telemetry/metrics.ts
+++ b/packages/core/src/telemetry/metrics.ts
@@ -6,37 +6,304 @@
 
 import type { Attributes, Meter, Counter, Histogram } from '@opentelemetry/api';
 import { diag, metrics, ValueType } from '@opentelemetry/api';
-import {
-  SERVICE_NAME,
-  METRIC_TOOL_CALL_COUNT,
-  METRIC_TOOL_CALL_LATENCY,
-  METRIC_API_REQUEST_COUNT,
-  METRIC_API_REQUEST_LATENCY,
-  METRIC_TOKEN_USAGE,
-  METRIC_SESSION_COUNT,
-  METRIC_FILE_OPERATION_COUNT,
-  EVENT_CHAT_COMPRESSION,
-  METRIC_INVALID_CHUNK_COUNT,
-  METRIC_CONTENT_RETRY_COUNT,
-  METRIC_CONTENT_RETRY_FAILURE_COUNT,
-  METRIC_MODEL_ROUTING_LATENCY,
-  METRIC_MODEL_ROUTING_FAILURE_COUNT,
-  METRIC_MODEL_SLASH_COMMAND_CALL_COUNT,
-  // Performance Monitoring Metrics
-  METRIC_STARTUP_TIME,
-  METRIC_MEMORY_USAGE,
-  METRIC_CPU_USAGE,
-  METRIC_TOOL_QUEUE_DEPTH,
-  METRIC_TOOL_EXECUTION_BREAKDOWN,
-  METRIC_TOKEN_EFFICIENCY,
-  METRIC_API_REQUEST_BREAKDOWN,
-  METRIC_PERFORMANCE_SCORE,
-  METRIC_REGRESSION_DETECTION,
-  METRIC_REGRESSION_PERCENTAGE_CHANGE,
-  METRIC_BASELINE_COMPARISON,
-} from './constants.js';
+import { SERVICE_NAME, EVENT_CHAT_COMPRESSION } from './constants.js';
 import type { Config } from '../config/config.js';
 import type { ModelRoutingEvent, ModelSlashCommandEvent } from './types.js';
+
+const TOOL_CALL_COUNT = 'gemini_cli.tool.call.count';
+const TOOL_CALL_LATENCY = 'gemini_cli.tool.call.latency';
+const API_REQUEST_COUNT = 'gemini_cli.api.request.count';
+const API_REQUEST_LATENCY = 'gemini_cli.api.request.latency';
+const TOKEN_USAGE = 'gemini_cli.token.usage';
+const SESSION_COUNT = 'gemini_cli.session.count';
+const FILE_OPERATION_COUNT = 'gemini_cli.file.operation.count';
+const INVALID_CHUNK_COUNT = 'gemini_cli.chat.invalid_chunk.count';
+const CONTENT_RETRY_COUNT = 'gemini_cli.chat.content_retry.count';
+const CONTENT_RETRY_FAILURE_COUNT =
+  'gemini_cli.chat.content_retry_failure.count';
+const MODEL_ROUTING_LATENCY = 'gemini_cli.model_routing.latency';
+const MODEL_ROUTING_FAILURE_COUNT = 'gemini_cli.model_routing.failure.count';
+const MODEL_SLASH_COMMAND_CALL_COUNT =
+  'gemini_cli.slash_command.model.call_count';
+
+// Performance Monitoring Metrics
+const STARTUP_TIME = 'gemini_cli.startup.duration';
+const MEMORY_USAGE = 'gemini_cli.memory.usage';
+const CPU_USAGE = 'gemini_cli.cpu.usage';
+const TOOL_QUEUE_DEPTH = 'gemini_cli.tool.queue.depth';
+const TOOL_EXECUTION_BREAKDOWN = 'gemini_cli.tool.execution.breakdown';
+const TOKEN_EFFICIENCY = 'gemini_cli.token.efficiency';
+const API_REQUEST_BREAKDOWN = 'gemini_cli.api.request.breakdown';
+const PERFORMANCE_SCORE = 'gemini_cli.performance.score';
+const REGRESSION_DETECTION = 'gemini_cli.performance.regression';
+const REGRESSION_PERCENTAGE_CHANGE =
+  'gemini_cli.performance.regression.percentage_change';
+const BASELINE_COMPARISON = 'gemini_cli.performance.baseline.comparison';
+
+const baseMetricDefinition = {
+  getCommonAttributes: (config: Config): Attributes => ({
+    'session.id': config.getSessionId(),
+  }),
+};
+
+const COUNTER_DEFINITIONS = {
+  [TOOL_CALL_COUNT]: {
+    description: 'Counts tool calls, tagged by function name and success.',
+    valueType: ValueType.INT,
+    assign: (c: Counter) => (toolCallCounter = c),
+    attributes: {} as {
+      function_name: string;
+      success: boolean;
+      decision?: 'accept' | 'reject' | 'modify' | 'auto_accept';
+      tool_type?: 'native' | 'mcp';
+    },
+  },
+  [API_REQUEST_COUNT]: {
+    description: 'Counts API requests, tagged by model and status.',
+    valueType: ValueType.INT,
+    assign: (c: Counter) => (apiRequestCounter = c),
+    attributes: {} as {
+      model: string;
+      status_code?: number | string;
+      error_type?: string;
+    },
+  },
+  [TOKEN_USAGE]: {
+    description: 'Counts the total number of tokens used.',
+    valueType: ValueType.INT,
+    assign: (c: Counter) => (tokenUsageCounter = c),
+    attributes: {} as {
+      model: string;
+      type: 'input' | 'output' | 'thought' | 'cache' | 'tool';
+    },
+  },
+  [SESSION_COUNT]: {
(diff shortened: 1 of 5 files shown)


packages/core/src/telemetry/metrics.ts
```diff
@@ -124,84 +386,20 @@ export function initializeMetrics(config: Config): void {
...
-  const sessionCounter = meter.createCounter(METRIC_SESSION_COUNT, {
-    description: 'Count of CLI sessions started.',
-    valueType: ValueType.INT,
-  });
-  sessionCounter.add(1, getCommonAttributes(config));
+  // Increment session counter after all metrics are initialized
+  sessionCounter?.add(1, baseMetricDefinition.getCommonAttributes(config));
 
   // Initialize performance monitoring metrics if enabled
   initializePerformanceMonitoring(config);
```
bobcatfish, line 405: @eLyiN is it important that the performance metrics are initialized separately from the other metrics, or was it just for clarity? if it's not important then i think we could now fold the content of initializePerformanceMonitoring into this function and make it pretty short and sweet 🤔

eLyiN: @bobcatfish Thanks for mentioning me, it was just organizational clarity, if there's no plan to have separate enable/disable controls for performance monitoring vs core telemetry, definitely fold it in! It would make the code much cleaner and align with the PR's goal of consolidating metric definitions.

If the team wants to eventually support separate toggles (for example: collect telemetry but not performance metrics), keep it separate. But the current implementation suggests that's not a priority since it's just checking the same `getTelemetryEnabled()` flag.

jerop (approved):

packages/core/src/telemetry/metrics.ts
```diff
@@ -6,38 +6,305 @@
...
-  METRIC_PERFORMANCE_SCORE,
-  METRIC_REGRESSION_DETECTION,
-  METRIC_REGRESSION_PERCENTAGE_CHANGE,
-  METRIC_BASELINE_COMPARISON,
-} from './constants.js';
+import { SERVICE_NAME, EVENT_CHAT_COMPRESSION } from './constants.js';
 import type { Config } from '../config/config.js';
 import type { ModelRoutingEvent, ModelSlashCommandEvent } from './types.js';
 
+const TOOL_CALL_COUNT = 'gemini_cli.tool.call.count';
```
richieforeman, line 13: Opinion: All of these constants (to me) feel like they'd be a good candidate for an enum.  That will increase the type safety of the usage site.

Benefits:
- an enum type is effectively a union type of all of these options, meaning that the strictness at the call site could go from `string` -> `a fixed list of specific items`, which is always better and helps catch errors
- I like to reach for enums when I have a fixed list of things, which is exactly what this is.

Also note:
- Unlike some other languages, Typescript has "string enums", which allow for string values (and not just numbers).  This is quite handy in cases like this.

packages/core/src/telemetry/metrics.ts
```diff
@@ -6,38 +6,305 @@
...
+const TOOL_EXECUTION_BREAKDOWN = 'gemini_cli.tool.execution.breakdown';
+const TOKEN_EFFICIENCY = 'gemini_cli.token.efficiency';
+const API_REQUEST_BREAKDOWN = 'gemini_cli.api.request.breakdown';
+const PERFORMANCE_SCORE = 'gemini_cli.performance.score';
+const REGRESSION_DETECTION = 'gemini_cli.performance.regression';
+const REGRESSION_PERCENTAGE_CHANGE =
+  'gemini_cli.performance.regression.percentage_change';
+const BASELINE_COMPARISON = 'gemini_cli.performance.baseline.comparison';
+
+const baseMetricDefinition = {
```
richieforeman, line 43: Is there a type that can be declared on this object?  Right now, the type will be inferred, which technically works, but the errors can be a bit obscure when you try to shove this object into a shape this isn't expecting.  A structural type (or even a partial structural type) is suggested.

go/tsjs-style#use-structural-types

packages/core/src/telemetry/metrics.ts
```diff
@@ -6,38 +6,305 @@
...
+  [EVENT_CHAT_COMPRESSION]: {
+    description: 'Counts chat compression events.',
+    valueType: ValueType.INT,
+    assign: (c: Counter) => (chatCompressionCounter = c),
+    attributes: {} as {
+      tokens_before: number;
+      tokens_after: number;
+    },
+  },
+} as const;
```
richieforeman, line 142: `as const` is a good start, but there should/must be a type here that can assert this shape.

You can have the best of `as const` and type safety by doing

`as const satisfied Record<SomeEnum, SomeInterface>`

Ideally this type would be `Record<SomeEnum, SomeInterface>`

packages/core/src/telemetry/metrics.ts
```diff
@@ -6,38 +6,305 @@
...
+    attributes: {} as {
+      model: string;
+      type: 'input' | 'output' | 'thought' | 'cache' | 'tool';
+    },
+  },
+  [SESSION_COUNT]: {
+    description: 'Count of CLI sessions started.',
+    valueType: ValueType.INT,
+    assign: (c: Counter) => (sessionCounter = c),
+    attributes: {} as Record<string, never>,
```
richieforeman, line 84: IIRC, I think `as never` would suffice.

packages/core/src/telemetry/metrics.ts
```diff
@@ -6,38 +6,305 @@
...
+    description: 'Latency of model routing decisions in milliseconds.',
+    unit: 'ms',
+    valueType: ValueType.INT,
+    assign: (h: Histogram) => (modelRoutingLatencyHistogram = h),
+    attributes: {} as {
+      'routing.decision_model': string;
+      'routing.decision_source': string;
+    },
+  },
+} as const;
```
richieforeman, line 173: Same commentary here. (and the next one, and so on.

packages/core/src/telemetry/metrics.ts
```diff
@@ -6,38 +6,305 @@
...
+
+const baseMetricDefinition = {
+  getCommonAttributes: (config: Config): Attributes => ({
+    'session.id': config.getSessionId(),
+  }),
+};
+
+const COUNTER_DEFINITIONS = {
+  [TOOL_CALL_COUNT]: {
+    description: 'Counts tool calls, tagged by function name and success.',
```
richieforeman, line 51: As I read on down the CL here, I definitely feel like an interface that defines each of these fields (maybe with some nice docblocks that describe each field).  I've nearly read this entire CL, and I still don't understand what the `assign` does.  I think an interface with a docblock would make that more obvious to the reader.

packages/core/src/telemetry/metrics.ts
```diff
@@ -124,84 +386,20 @@ export function initializeMetrics(config: Config): void {
...
-    description: 'Counts API requests, tagged by model and status.',
-    valueType: ValueType.INT,
-  });
-  apiRequestLatencyHistogram = meter.createHistogram(
-    METRIC_API_REQUEST_LATENCY,
-    {
-      description: 'Latency of API requests in milliseconds.',
-      unit: 'ms',
-      valueType: ValueType.INT,
+  Object.entries(COUNTER_DEFINITIONS).forEach(
```
richieforeman, line 389: Prefer classical `for... of... ` loop.

The .foreach is kind of discouraged because all of the other methods on Array.prototype return a new array (e.g. `filter`, `reduce`, etc, etc).  Foreach is the one weird child that returns `void`.  For that reason, I almost always prefer `for ... of...` loops as it makes my intent very explicit.

packages/core/src/telemetry/metrics.ts
```diff
@@ -272,23 +461,21 @@ export function recordApiResponseMetrics(
   )
     return;
   const metricAttributes: Attributes = {
-    ...getCommonAttributes(config),
-    model,
-    status_code: statusCode ?? 'ok',
+    ...baseMetricDefinition.getCommonAttributes(config),
+    model: attributes.model,
+    status_code: attributes.status_code ?? 'ok',
```
richieforeman, line 466: Sanity check: I think all of your other `status_code` have been 200, 500, etc (maybe I'm hallucinating).  It's weird that this is `ok`.

packages/core/src/telemetry/metrics.ts
```diff
@@ -6,38 +6,305 @@
...
+    },
+  },
+} as const;
+
+type AllMetricDefs = typeof COUNTER_DEFINITIONS &
+  typeof HISTOGRAM_DEFINITIONS &
+  typeof PERFORMANCE_COUNTER_DEFINITIONS &
+  typeof PERFORMANCE_HISTOGRAM_DEFINITIONS;
+
+export type MetricDefinitions = {
```
richieforeman, line 302: nit, rather than `type = ` prefer `interface`

packages/core/src/telemetry/metrics.ts
```diff
@@ -6,38 +6,305 @@
...
+      tool_type?: 'native' | 'mcp';
+    },
+  },
+  [API_REQUEST_COUNT]: {
+    description: 'Counts API requests, tagged by model and status.',
+    valueType: ValueType.INT,
+    assign: (c: Counter) => (apiRequestCounter = c),
+    attributes: {} as {
+      model: string;
+      status_code?: number | string;
```
richieforeman, line 67: Personally, I'd probably hoist a specific type for `status_code` and use either:

- a number (like http status)
- an enum of fixed options (success, failure, etc)
- (least preferred) a string union of success|failure|etc

packages/core/src/telemetry/metrics.ts
richieforeman: I left a comment above about this, but it feels like the status_code thing could use some work.

I think we probably want to avoid a world where we have a bunch of rando strings in the DB indicated "success" vs "ok" || "failure" vs "failed" vs "error".  This data could get messy fast and I think the stricter we make the logging calls, ultimately the better, IMHO.  Our logging system (the typing around it) should really help us enforce "clean telemetry data"

packages/core/src/telemetry/metrics.ts
```diff
@@ -297,38 +484,27 @@ export function recordApiErrorMetrics(
...
     return;
   const metricAttributes: Attributes = {
-    ...getCommonAttributes(config),
-    model,
-    status_code: statusCode ?? 'error',
-    error_type: errorType ?? 'unknown',
+    ...baseMetricDefinition.getCommonAttributes(config),
+    model: attributes.model,
+    status_code: attributes.status_code ?? 'error',
+    error_type: attributes.error_type ?? 'unknown',
```
richieforeman, line 490: If possible, I feel the same commentary for `status_code` could potentially apply to `error_type`

packages/core/src/telemetry/metrics.ts
richieforeman: Also, it looks like you're just using this at the callsites as `baseMetricDefintion.whateverWhatever`.  I'd consider just making this an unexported file-level function (using the function keyword, plz) and avoid the noise of the object.  It (the object) feels like unnecessary baggage here.

PR #8006, opened 2025-09-08
docs: document custom witty loading phrases feature

JayadityaGit:
## TLDR

This pull request documents the new **Custom Witty Loading Phrases** feature.
It introduces the `ui.customWittyPhrases` setting in `docs/cli/configuration.md` and updates the example `settings.json` accordingly.

## Dive Deeper

* Adds `ui.customWittyPhrases` to the configuration reference.
* Expands the `settings.json` example to include witty phrases.
* Provides clear documentation so users can easily customize CLI loading messages.

This documentation work ensures users are aware of the new customization option introduced by the feature.

## Testing Matrix

|          | 🍏 | 🪟 | 🐧 |
| -------- | -- | -- | -- |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -  | -  |
| Seatbelt | ❓  | -  | -  |

## Linked issues / bugs

Related to #8432 — this feature currently has a bug.
I would kindly suggest reviewing it first, since it attempts to solve a major bug in the feature.

PR #10248, opened 2025-09-30
Cleanup useSelectionList and fix infinite loop in debug mode issues.

### LINKED ISSUE #10247: Radio buttons trigger infinite loop in with DEBUG=true
### What happened?

Trigger any UI with radio buttons when DEBUG=1
Notice a constant stream of "Maximum update depth exceeded. This can happen when a component calls setState inside useEffect, but useEffect either doesn't have a dependency array, or one of the dependencies changes on every render." errors.

No Maximum update depth exceeded errors.
This surfaces as performance issues for the theme dialog.

This issue has not made it to the preview branch so does not need to be hotfixed if fixed today.

### Client information

<details>

```console
> /about
# paste output here
```

</details>

### Login information

_No response_

_No response_

jacob314:
## TLDR

Unconditionally dispatching events in useEffect from useSelectionList triggers an infinite loop of updates.

## Dive Deeper

The reducer and core logic in useSelectionList never had to depend on the actual elements in the list only their keys and whether they were disabled. Refactored so that was clear making it more obvious when the update logic could be short circuited for cases where the keys and disabled state didn't change.

Some methods in the reducer also passed in the items when the items didn't change further causing confusion about what needed to be updated when.

## Reviewer Test Plan

set DEBUG=true
note the errors without this change
open the /theme dialog and other ui with radio buttons.
Verify UI is responsive and no errors are thrown.

## Linked issues / bugs

Fixes: #10247
Fixes: #10242

### DIFF
--- a/packages/cli/src/ui/hooks/useSelectionList.test.ts
+++ b/packages/cli/src/ui/hooks/useSelectionList.test.ts
@@ -915,6 +915,39 @@ describe('useSelectionList', () => {
 
       expect(result.current.activeIndex).toBe(2);
     });
+
+    it('should not re-initialize when items have identical keys but are different objects', () => {
+      const initialItems = [
+        { value: 'A', key: 'A' },
+        { value: 'B', key: 'B' },
+      ];
+
+      let renderCount = 0;
+
+      const { rerender } = renderHook(
+        ({ items: testItems }: { items: Array<SelectionListItem<string>> }) => {
+          renderCount++;
+          return useSelectionList({
+            onSelect: mockOnSelect,
+            onHighlight: mockOnHighlight,
+            items: testItems,
+          });
+        },
+        { initialProps: { items: initialItems } },
+      );
+
+      // Initial render
+      expect(renderCount).toBe(1);
+
+      // Create new items with the same keys but different object references
+      const newItems = [
+        { value: 'A', key: 'A' },
+        { value: 'B', key: 'B' },
+      ];
+
+      rerender({ items: newItems });
+      expect(renderCount).toBe(2);
+    });
   });
 
   describe('Manual Control', () => {
--- a/packages/cli/src/ui/hooks/useSelectionList.ts
+++ b/packages/cli/src/ui/hooks/useSelectionList.ts
@@ -4,12 +4,17 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { useReducer, useRef, useEffect } from 'react';
-import { useKeypress } from './useKeypress.js';
+import { useReducer, useRef, useEffect, useCallback } from 'react';
+import { useKeypress, type Key } from './useKeypress.js';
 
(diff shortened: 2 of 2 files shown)


packages/cli/src/ui/hooks/useSelectionList.ts
```diff
@@ -228,20 +244,38 @@ export function useSelectionList<T>({
...
+
+    if (baseItemsChanged || initialIndexChanged) {
+      dispatch({
+        type: 'INITIALIZE',
+        payload: { initialIndex, items: baseItems },
+      });
+      prevBaseItemsRef.current = baseItems;
+      prevInitialIndexRef.current = initialIndex;
+    }
+  });
```
abhipatel12, line 278: Do we not need a dependency array?

packages/cli/src/ui/hooks/useSelectionList.test.ts
```diff
@@ -915,6 +915,39 @@ describe('useSelectionList', () => {
...
+      // Initial render
+      expect(renderCount).toBe(1);
+
+      // Create new items with the same keys but different object references
+      const newItems = [
+        { value: 'A', key: 'A' },
+        { value: 'B', key: 'B' },
+      ];
+
+      rerender({ items: newItems });
```
abhipatel12, line 948: Should we also assert that the current active index is also preserved (i.e user's choice is preserved)?

packages/cli/src/ui/hooks/useSelectionList.ts
abhipatel12, line 278: Just curious, we can omit the dependency array here?

abhipatel12 (approved):
LGTM with a couple of questions

packages/cli/src/ui/hooks/useSelectionList.ts
jacob314: because the one dep will always change each frame so there is no point.

packages/cli/src/ui/hooks/useSelectionList.test.ts
jacob314: i'd keep this test validating one thing. other tests validate that. this test isn't messing with the activeIndex so regardless of re-renders it would be the same value.

JayadityaGit:
lgtm

JayadityaGit:
It’s been a while since this PR was approved: https://github.com/google-gemini/gemini-cli/pull/8006

Could you guys please review the documentation update for the Custom Witty Phrases feature? I’m really eager to get this documented, and your feedback would mean a lot. Thanks! 🙏

packages/cli/src/ui/hooks/useSelectionList.ts
jacob314: chatted offline. because we intentionally run the useEffect after each render.

PR #10280, opened 2025-10-01
Switch to a reducer for tracking update state fixing flicker issues due to continuous renders

jacob314:
## TLDR

We were triggering calls to setState every frame triggering a classic React render loop issue due to improper react state management.

## Dive Deeper

Refactored from unsafe ways of using setState to using a proper reducer pattern.
Also avoided scary cases we were doing work after an async gap where state might have changed completely.

I will follow up with a proper integration test for this case as well as fixing the frame counter logic to add errors to the output even when ctrl-b has not been pressed warning that there is constant rendering when the app should be idle.

## Reviewer Test Plan

Run with --debug
press ctrl-b to trigger the frame render counter.
Verify that without the change the number goes up continously.
With this change the frame counter only goes up when the user interacts with the app or content is streaming
Manually test that updating extensions still works correctly.
Also verify that you can make large edits, press ctrl-S and not get destroyed by flicker.

## Testing Matrix

Resolves #10242

### DIFF
--- a/packages/cli/src/config/extensions/update.ts
+++ b/packages/cli/src/config/extensions/update.ts
@@ -7,6 +7,7 @@
 import {
   type ExtensionUpdateAction,
   ExtensionUpdateState,
+  type ExtensionUpdateStatus,
 } from '../../ui/state/extensions.js';
 import {
   copyExtension,
@@ -121,7 +122,7 @@ export async function updateAllUpdatableExtensions(
   cwd: string = process.cwd(),
   requestConsent: (consent: string) => Promise<boolean>,
   extensions: GeminiCLIExtension[],
-  extensionsState: Map<string, ExtensionUpdateState>,
+  extensionsState: Map<string, ExtensionUpdateStatus>,
   dispatch: (action: ExtensionUpdateAction) => void,
 ): Promise<ExtensionUpdateInfo[]> {
   return (
@@ -129,7 +130,7 @@ export async function updateAllUpdatableExtensions(
       extensions
         .filter(
           (extension) =>
-            extensionsState.get(extension.name) ===
+            extensionsState.get(extension.name)?.status ===
             ExtensionUpdateState.UPDATE_AVAILABLE,
         )
         .map((extension) =>
@@ -137,7 +138,7 @@ export async function updateAllUpdatableExtensions(
             extension,
             cwd,
             requestConsent,
-            extensionsState.get(extension.name)!,
+            extensionsState.get(extension.name)!.status,
             dispatch,
           ),
         ),
@@ -152,13 +153,10 @@ export interface ExtensionUpdateCheckResult {
 
 export async function checkForAllExtensionUpdates(
   extensions: GeminiCLIExtension[],
-  extensionsUpdateState: Map<string, ExtensionUpdateState>,
+  extensionsUpdateState: Map<string, ExtensionUpdateStatus>,
   dispatch: (action: ExtensionUpdateAction) => void,
   cwd: string = process.cwd(),
-): Promise<Map<string, ExtensionUpdateState>> {
-  const newStates: Map<string, ExtensionUpdateState> = new Map(
-    extensionsUpdateState,
-  );
+): Promise<void> {
   for (const extension of extensions) {
     const initialState = extensionsUpdateState.get(extension.name);
     if (initialState === undefined) {
@@ -170,7 +168,6 @@ export async function checkForAllExtensionUpdates(
             state: ExtensionUpdateState.NOT_UPDATABLE,
           },
         });
-        newStates.set(extension.name, ExtensionUpdateState.NOT_UPDATABLE);
         continue;
       }
       await checkForExtensionUpdate(
@@ -180,11 +177,9 @@ export async function checkForAllExtensionUpdates(
             type: 'SET_STATE',
             payload: { name: extension.name, state: updatedState },
           });
-          newStates.set(extension.name, updatedState);
         },
         cwd,
       );
     }
   }
-  return newStates;
 }
--- a/packages/cli/src/config/extensions/update.ts
+++ b/packages/cli/src/config/extensions/update.ts
@@ -153,33 +153,24 @@ export interface ExtensionUpdateCheckResult {
 
 export async function checkForAllExtensionUpdates(
   extensions: GeminiCLIExtension[],
-  extensionsUpdateState: Map<string, ExtensionUpdateStatus>,
   dispatch: (action: ExtensionUpdateAction) => void,
-  cwd: string = process.cwd(),
 ): Promise<void> {
   for (const extension of extensions) {
-    const initialState = extensionsUpdateState.get(extension.name);
-    if (initialState === undefined) {
-      if (!extension.installMetadata) {
-        dispatch({
-          type: 'SET_STATE',
-          payload: {
-            name: extension.name,
-            state: ExtensionUpdateState.NOT_UPDATABLE,
-          },
-        });
-        continue;
-      }
-      await checkForExtensionUpdate(
-        extension,
-        (updatedState) => {
-          dispatch({
-            type: 'SET_STATE',
-            payload: { name: extension.name, state: updatedState },
-          });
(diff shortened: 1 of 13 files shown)


packages/cli/src/ui/state/extensions.ts
```diff
@@ -14,3 +14,26 @@ export enum ExtensionUpdateState {
...
+        return state;
+      }
+      const newState = new Map(state);
+      newState.set(action.payload.name, action.payload.state);
+      return newState;
+    }
+    default:
+      return state;
+  }
+}
```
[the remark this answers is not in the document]
jacob314: done although note that you have to use checkExhaustive(action) not action.type

packages/cli/src/config/extensions/update.ts
```diff
@@ -137,38 +153,24 @@ export interface ExtensionUpdateCheckResult {
...
+        payload: {
+          name: extension.name,
+          state: ExtensionUpdateState.NOT_UPDATABLE,
         },
-        cwd,
-      );
+      });
+      continue;
     }
+    await checkForExtensionUpdate(extension, (updatedState) => {
```
jakemac53: Fwiw, I believe this function always just calls `dispatch` and returns, so you could just make it a more standard function that returns the state, and call dispatch with the return value as the state. You just need to set the state to `CHECKING_FOR_UPDATE` before calling the function.

packages/cli/src/ui/hooks/useExtensionUpdates.ts
```diff
@@ -6,127 +6,184 @@
...
+      } finally {
+        setIsChecking(false);
+      }
+    })();
+  }, [
+    extensions,
+    extensions.length,
+    dispatchExtensionStateUpdate,
+    isChecking,
+    extensionsUpdateState,
```
jakemac53: This shouldn't be here right?

packages/cli/src/ui/hooks/useExtensionUpdates.ts
```diff
@@ -6,127 +6,184 @@
...
-  })();
+  }, [
+    extensions,
+    extensionsUpdateState,
+    addConfirmUpdateExtensionRequest,
+    addItem,
+    cwd,
+  ]);
+
+  const extensionsUpdateStateComputed = useMemo(() => {
```
jakemac53, line 168: Why do we need this and the separate internal one?

packages/cli/src/ui/hooks/useExtensionUpdates.ts
```diff
@@ -6,127 +6,184 @@
...
+    })();
+  }, [
+    extensions,
+    extensions.length,
+    dispatchExtensionStateUpdate,
+    isChecking,
+    extensionsUpdateState,
+  ]);
+
+  useEffect(() => {
```
jakemac53, line 91: Is this actually the core fix, just separating it out into two functions, one that operates on the new state and one that sets the state?

packages/cli/src/ui/hooks/useExtensionUpdates.ts
```diff
@@ -6,127 +6,184 @@
...
         );
+      } finally {
+        setIsChecking(false);
+      }
+    })();
+  }, [
+    extensions,
+    extensions.length,
+    dispatchExtensionStateUpdate,
+    isChecking,
```
jakemac53: Does this not create the same issue from before?

packages/cli/src/ui/hooks/useExtensionUpdates.ts
jakemac53: Although this function both takes and updates the state so that is confusing to me as well, although it makes sense that we can avoid doing the repeated check for updates, that function already had similar logic in it to bail out for already processed extensions.

packages/cli/src/ui/hooks/useExtensionUpdates.ts
```diff
@@ -6,127 +6,184 @@
...
+        request: wrappedRequest,
+      });
     },
-    [setConfirmUpdateExtensionRequests],
+    [dispatchConfirmUpdateExtensionRequests],
   );
 
-  (async () => {
-    if (isChecking) return;
+  useEffect(() => {
```
jakemac53, line 82: Fwiw I tried changing this to just:

```ts
  useEffect(() => {
    checkForAllExtensionUpdates(extensions, dispatchExtensionStateUpdate);
  }, [extensions, dispatchExtensionStateUpdate]);
```

And that got called only once.

jacob314: that is better. now that we have multiple useEffects and this one doesn't have the update count logic I think this is fine. Messages about extension updates being available might come through in multiple batches so if that is an issue you might want to add some rate limit there so you only publish messages about extension updates being available every second or something.

packages/cli/src/ui/hooks/useExtensionUpdates.ts
jacob314: not quite but moot point with other suggestion taken.

packages/cli/src/ui/hooks/useExtensionUpdates.ts
jacob314: done. annoying we don't have a lint about unneeded deps. Not needed as that logic is refactored to the following use effect.

packages/cli/src/ui/state/extensions.ts
```diff
@@ -14,3 +16,52 @@ export enum ExtensionUpdateState {
...
 }
+
+export interface ExtensionUpdateStatus {
+  status: ExtensionUpdateState;
+  processed: boolean;
+}
+
+export type ExtensionUpdateAction =
+  | {
+      type: 'SET_STATE';
```
jakemac53, line 37: Can you make SET_STATE and SET_PROCESSED into an enum so we can get auto-complete and also refactor these more easily? Same for add/remove.

jakemac53: Oh, actually I see that autocomplete works regardless so this is up to you

packages/cli/src/ui/hooks/useExtensionUpdates.ts
jacob314: no. the core fix was the first change. if you check the commits here, the first is the core fix and moving to this separate useEffect is the bonus to remove the sketchyness of modifying state after an async gap where the hooks dependencies might be invalid but without using a proper dispose check.

jacob314: difference is this logic is running fresh as a useEffect after state changes while the other logic was running after an async gap with all kinds of potential stale state issues.

packages/cli/src/ui/hooks/useExtensionUpdates.ts
jacob314: note the `value.status` line.
The map we surface publicly only has the status while the internal data structure has `value.processed` as well which doesn't need to leak out of this file.

We could let it leak out if we wanted to but this encapsulates slightly better.

jakemac53: Ohhh ok gotcha I missed that

packages/cli/src/ui/hooks/useExtensionUpdates.ts
jakemac53: Oh hmm that is true this would definitely be a regression in the UX.... we will probably end up logging N messages now? Each one saying only one extension needs to be updated...

jacob314 (commented):
All code review comments addressed. PTAL

packages/cli/src/ui/hooks/useExtensionUpdates.ts
jacob314: fixed.

packages/cli/src/ui/state/extensions.ts
jacob314: yeah no need for enums with typescript tooling. some style guides recommend this over enums.

jakemac53 (approved):

PR #10282, opened 2025-10-01
Codebase Investigator: Separate initial query from system prompt and apply templateStrings in query and initialMessages

silviojr:
## TLDR

The main goal of this PR is to separate the query from the system instruction. 

In this PR:
1. Add a new option in promptConfig: `query`. This should be the message to start the loop, coming after the initialMessages that may contain other content. 
2. Use templateString in both `query` and `initialMessages` using the values of the input parameters. 
3. Update the Codebase Investigator subagent to use the new `query` field. Also update its description and the description of its inputs to match what we have been testing offline. 

## Reviewer Test Plan

Added new unit tests. 

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | x  | ❓  | x |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

### DIFF
--- a/packages/core/src/agents/types.ts
+++ b/packages/core/src/agents/types.ts
@@ -74,6 +74,11 @@ export interface PromptConfig {
    * An array of user/model content pairs for few-shot prompting.
    */
   initialMessages?: Content[];
+
+  /*
+   * The query to start the loop. If not provided, a generic default will be used.
+   */
+  query?: string;
 }
 
 /**
--- a/packages/core/src/agents/types.ts
+++ b/packages/core/src/agents/types.ts
@@ -75,8 +75,11 @@ export interface PromptConfig {
    */
   initialMessages?: Content[];
 
-  /*
-   * The query to start the loop. If not provided, a generic default will be used.
+  /**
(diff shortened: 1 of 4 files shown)


packages/core/src/agents/types.ts
```diff
@@ -74,6 +74,11 @@ export interface PromptConfig {
    * An array of user/model content pairs for few-shot prompting.
    */
   initialMessages?: Content[];
+
+  /*
+   * The query to start the loop. If not provided, a generic default will be used.
+   */
```
abhipatel12: nit: Can we add a little more here?

```
    /**
      * The specific task or question to trigger the agent's execution loop.
      * This is sent as the first user message, distinct from the systemPrompt (identity/rules)
      * and initialMessages (history/few-shots). Supports templating.
      * If not provided, a generic "Get Started!" message is used.
     */
```

abhipatel12 (approved):
LGTM!

just one comment

PR #10290, opened 2025-10-01
feat(core): mitigate Flash API early chat bug & enable model router by default

abhipatel12:
## TLDR

This PR implements a mitigation in the core client for an issue where Gemini 2.5 Flash returns empty responses when using tools early in a conversation (history length < 3).

With this mitigation in place, we are also re-enabling the `useModelRouter` experimental setting by default in the CLI.

### The Issue
There is a known API issue where Gemini 2.5 Flash can return an empty response error when processing function calls during the first few turns of a chat (specifically when history length is less than 3).

### The Mitigation (`packages/core`)
Modified `GeminiClient.sendMessageStream` to intercept requests meeting the problematic criteria:
1.  Chat history length is < 3.
2.  The selected model (by routing or configuration) is Gemini 2.5 Flash.

If these conditions are met, the client temporarily switches the model for that turn:
*   **Fallback Mode:** Switches to **Flash Lite** to ensure functionality.
*   **Auto Mode:** Switches to **Pro** to provide a better experience for the initial interaction.

Comprehensive tests were added to `client.test.ts` to verify this behavior, ensuring chat compression doesn't interfere with the history length checks during testing.

### Re-enabling Model Router (`packages/cli`)
Previously, `useModelRouter` was disabled by default as a temporary workaround for this Flash issue. Now that the core client handles this specific edge case, we have reverted the default value of `useModelRouter` to `true` in `config.ts` and `settingsSchema.ts`. Relevant config tests have been updated.

## Reviewer Test Plan

1.  **Verify Mitigation (Fresh Session):**
    Start a new CLI session and immediately use a prompt that triggers a tool (e.g., read readme).
    ```bash
    gemini -i "read the readme"
    ```
    *   **Expected:** The command should succeed without an "empty response" error.
    *   **Observation:** Use `/stats` to see there will be 3 requests (1 flash lite for routing, 1 pro for the initial function request, and then 1 flash for the response after we cross the 3 turn threshold.

3.  **Verify Long Context (No Mitigation):**
    Have a longer conversation (>= 3 turns).
    Ask a question requiring a tool.
    *   **Expected:** The router should be able to use Flash (if appropriate for the query complexity) without the client overriding it.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  |✅| ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt |✅| -   | -   |

### DIFF
--- a/packages/core/src/core/client.ts
+++ b/packages/core/src/core/client.ts
@@ -30,6 +30,7 @@ import { tokenLimit } from './tokenLimits.js';
 import type { ChatRecordingService } from '../services/chatRecordingService.js';
 import type { ContentGenerator } from './contentGenerator.js';
 import {
+  DEFAULT_GEMINI_FLASH_LITE_MODEL,
   DEFAULT_GEMINI_FLASH_MODEL,
   DEFAULT_GEMINI_MODEL,
   DEFAULT_GEMINI_MODEL_AUTO,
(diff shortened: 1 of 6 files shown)


packages/core/src/core/client.ts
```diff
@@ -528,6 +529,26 @@ export class GeminiClient {
...
 
+    // Due to a current API issue, when 2.5 Flash is used with a function call
+    // during first parts of the chat history, it causes a model empty response
+    // error. We must use a different model to avoid model empty response error.
+    // This error could only be reproduced when the history is shorter than 3.
+    if (
+      this.getChat().getHistory().length < 3 &&
+      this.currentSequenceModel === DEFAULT_GEMINI_FLASH_MODEL
+    ) {
+      // If a user is in fallback, we should use Flash lite to avoid the error.
```
NTaylorMullen, line 540: This I feel a little off about. If flash is the current default due to fall back it's not guaranteed that it will fail. To be frank this is how it works today. Ideally we'd "try" flash and if we get an empty response retry with flash-lite.

That being said, that flow may be a bit more invasive; however, forcing flash-lite for the first 3 turns is even more disruptive IF the user's flow isn't guaranteed to hit this bug

abhipatel12:
Mitigation needs to explored more before accepting this hard-coded fix

PR #10339, opened 2025-10-01
fix: Stream parsing for Windows Zed integration

benbrandt:
## TLDR

The stream of JSON RPC calls is newline delimited, and it seems this got updated in a previous commit by accident. This should solve the issue we're seeing where Gemini CLI never responds on Windows.

## Reviewer Test Plan

We've tested that the change works for us on Windows, everything is running happily again!

## Linked issues / bugs

Closes: https://github.com/google-gemini/gemini-cli/issues/7880

### DIFF
--- a/packages/cli/src/zed-integration/acp.ts
+++ b/packages/cli/src/zed-integration/acp.ts
@@ -7,7 +7,6 @@
 /* ACP defines a schema for a simple (experimental) JSON-RPC protocol that allows GUI applications to interact with agents. */
 
 import { z } from 'zod';
-import { EOL } from 'node:os';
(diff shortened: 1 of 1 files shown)


skeshive (approved):

benbrandt:
@skeshive I'm not sure why CI failed.. so I tried rebasing on main. Let me know if you need any more changes from me

skeshive:
> @skeshive I'm not sure why CI failed.. so I tried rebasing on main. Let me know if you need any more changes from me

I'm not sure either but I'll keep an eye on it today and make sure we can get this in!

PR #10343, opened 2025-10-01
feat(telemetry): add OpenTelemetry GenAI semantic convention metrics

jerop:
## TLDR

This PR updates the telemetry system to use the new OpenTelemetry semantic conventions for GenAI. This aligns our metrics with the emerging industry standard, improving interoperability and making our telemetry data more valuable.

## Dive Deeper

The OpenTelemetry project has recently introduced a set of semantic conventions for generative AI. This PR adopts these conventions for our metrics related to API requests and token usage. By doing so, we ensure that our telemetry data is consistent with other tools and platforms that support these conventions. This will make it easier to analyze our telemetry data in observability platforms and compare it with data from other sources.

The following metrics have been updated:
- API requests
- Token usage

## Reviewer Test Plan

1.  **Configure Telemetry:**
    *   **For local testing:** Configure the CLI to write telemetry to a file by updating your `.gemini/settings.json`:
        ```json
        {
          "telemetry": {
            "enabled": true,
            "target": "local",
            "outfile": ".gemini/telemetry.log"
          }
        }
        ```
    *   **For a backend:** Configure the CLI to send telemetry to an OpenTelemetry collector. See `docs/telemetry.md` for detailed instructions on configuring a local collector or sending data to Google Cloud.

2.  **Run Commands:**
    *   Run a few commands that interact with the model, for example:
        ```bash
        gemini -p "hello"
        ```

3.  **Inspect Telemetry:**
    *   **If writing to a file:** Inspect the contents of `.gemini/telemetry.log`.
    *   **If using a backend:** Check your observability platform.
    *   Verify that the exported metrics follow the new GenAI semantic conventions. For example, you should see metrics like `gen_ai.client.token.usage` with attributes like `gen_ai.token.type`.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ✅  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes #9212

### DIFF
--- a/packages/core/src/telemetry/loggers.ts
+++ b/packages/core/src/telemetry/loggers.ts
@@ -67,8 +67,6 @@ import type {
 } from './types.js';
 import {
   recordApiErrorMetrics,
-  recordTokenUsageMetrics,
-  recordApiResponseMetrics,
   recordToolCallMetrics,
   recordChatCompressionMetrics,
   recordFileOperationMetric,
@@ -77,6 +75,9 @@ import {
   recordContentRetryFailure,
   recordModelRoutingMetrics,
   recordModelSlashCommand,
+  getConventionAttributes,
+  recordTokenUsageMetrics,
+  recordApiResponseMetrics,
 } from './metrics.js';
 import { isTelemetrySdkInitialized } from './sdk.js';
 import type { UiEvent } from './uiTelemetry.js';
@@ -366,6 +367,17 @@ export function logApiError(config: Config, event: ApiErrorEvent): void {
     status_code: event.status_code,
     error_type: event.error_type,
   });
+
+  // Record GenAI operation duration for errors
+  const conventionAttributes = getConventionAttributes(event);
+  recordApiResponseMetrics(config, event.duration_ms, {
+    model: event.model,
+    status_code: event.status_code,
+    genAiAttributes: {
+      ...conventionAttributes,
+      'error.type': event.error_type || 'unknown',
+    },
+  });
 }
 
 export function logApiResponse(config: Config, event: ApiResponseEvent): void {
@@ -398,30 +410,30 @@ export function logApiResponse(config: Config, event: ApiResponseEvent): void {
     attributes,
   };
   logger.emit(logRecord);
+
+  const conventionAttributes = getConventionAttributes(event);
+
   recordApiResponseMetrics(config, event.duration_ms, {
     model: event.model,
     status_code: event.status_code,
+    genAiAttributes: conventionAttributes,
   });
-  recordTokenUsageMetrics(config, event.input_token_count, {
-    model: event.model,
-    type: 'input',
-  });
-  recordTokenUsageMetrics(config, event.output_token_count, {
-    model: event.model,
-    type: 'output',
-  });
-  recordTokenUsageMetrics(config, event.cached_content_token_count, {
-    model: event.model,
-    type: 'cache',
-  });
-  recordTokenUsageMetrics(config, event.thoughts_token_count, {
-    model: event.model,
-    type: 'thought',
-  });
-  recordTokenUsageMetrics(config, event.tool_token_count, {
-    model: event.model,
-    type: 'tool',
-  });
+
+  const tokenUsageData = [
+    { count: event.input_token_count, type: 'input' as const },
+    { count: event.output_token_count, type: 'output' as const },
+    { count: event.cached_content_token_count, type: 'cache' as const },
+    { count: event.thoughts_token_count, type: 'thought' as const },
+    { count: event.tool_token_count, type: 'tool' as const },
+  ];
+
+  for (const { count, type } of tokenUsageData) {
+    recordTokenUsageMetrics(config, count, {
+      model: event.model,
+      type,
+      genAiAttributes: conventionAttributes,
+    });
+  }
 }
 
 export function logLoopDetected(
--- a/packages/core/src/telemetry/metrics.test.ts
+++ b/packages/core/src/telemetry/metrics.test.ts
@@ -61,6 +61,11 @@ function originalOtelMockFactory() {
       setLogger: vi.fn(),
       warn: vi.fn(),
     },
+    DiagConsoleLogger: vi.fn(),
+    DiagLogLevel: {
+      NONE: 0,
+      INFO: 1,
+    },
   } as const;
 }
 
@@ -82,6 +87,8 @@ describe('Telemetry Metrics', () => {
   let recordPerformanceScoreModule: typeof import('./metrics.js').recordPerformanceScore;
   let recordPerformanceRegressionModule: typeof import('./metrics.js').recordPerformanceRegression;
   let recordBaselineComparisonModule: typeof import('./metrics.js').recordBaselineComparison;
+  let recordGenAiClientTokenUsageModule: typeof import('./metrics.js').recordGenAiClientTokenUsage;
+  let recordGenAiClientOperationDurationModule: typeof import('./metrics.js').recordGenAiClientOperationDuration;
 
   beforeEach(async () => {
     vi.resetModules();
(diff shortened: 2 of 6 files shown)


packages/core/src/telemetry/metrics.ts
```diff
@@ -341,6 +375,28 @@ export enum ApiRequestPhase {
...
+
+export enum GenAiOperationName {
+  CHAT = 'chat',
+  CREATE_AGENT = 'create_agent',
+  EMBEDDINGS = 'embeddings',
+  EXECUTE_TOOL = 'execute_tool',
+  GENERATE_CONTENT = 'generate_content',
+  INVOKE_AGENT = 'invoke_agent',
+  TEXT_COMPLETION = 'text_completion',
+}
```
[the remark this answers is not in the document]
jerop: > * The spec uses `embedding` (singular), not `embeddings`.

It is actually `embeddings` - https://github.com/open-telemetry/semantic-conventions/blob/main/docs/gen-ai/gen-ai-metrics.md#metric-gen_aiclientoperationduration

> * The spec uses `completion`, not `text_completion`.

It is actually `text_completion` - https://github.com/open-telemetry/semantic-conventions/blob/main/docs/gen-ai/gen-ai-metrics.md#metric-gen_aiclientoperationduration 

> * The values `create_agent`, `execute_tool`, `generate_content`, and `invoke_agent` are not defined in the specification.

These are actually defined: https://github.com/open-telemetry/semantic-conventions/blob/main/docs/gen-ai/gen-ai-metrics.md#metric-gen_aiclientoperationduration

> This issue also affects `loggers.ts`, where `getGenAiOperationName` hardcodes the return value to `generate_content`, a non-standard operation name.

`generate_content` is a standard operation name: https://github.com/open-telemetry/semantic-conventions/blob/main/docs/gen-ai/gen-ai-metrics.md#metric-gen_aiclientoperationduration

packages/core/src/telemetry/metrics.test.ts
```diff
@@ -61,6 +61,13 @@ function originalOtelMockFactory() {
       setLogger: vi.fn(),
       warn: vi.fn(),
     },
+    DiagConsoleLogger: vi.fn().mockImplementation(() => ({})),
+    DiagLogLevel: {
+      INFO: 30,
+      DEBUG: 20,
+      WARN: 50,
+      ERROR: 60,
+    },
```
bobcatfish, line 68: i dont see this being used anywhere, am i missing something 🤔

packages/core/src/telemetry/loggers.ts
```diff
@@ -422,6 +430,24 @@ export function logApiResponse(config: Config, event: ApiResponseEvent): void {
...
+      'gen_ai.token.type': GenAiTokenType.INPUT,
+    });
+  }
+
+  if (event.output_token_count > 0) {
+    recordGenAiClientTokenUsage(config, event.output_token_count, {
+      ...commonAttributes,
+      'gen_ai.token.type': GenAiTokenType.OUTPUT,
+    });
+  }
```
bobcatfish, line 436: might be nice to have the lines that emit input/output metrics co-located with the above lines doing the same thing for our 'legacy' metrics

```
  recordTokenUsageMetrics(config, event.input_token_count, {
    model: event.model,
    type: 'input',
  });
  recordTokenUsageMetrics(config, event.output_token_count, {
    model: event.model,
    type: 'output',
  });
```

bobcatfish: i also wonder if this could go a step further and have a nice little loop over the "types" instead of 5 separate lines of code...

AND i'm wondering, why do the above lines not need to check the input/output token count before emitting the metric, while the new lines do?

packages/core/src/telemetry/loggers.ts
```diff
@@ -422,6 +430,24 @@ export function logApiResponse(config: Config, event: ApiResponseEvent): void {
     model: event.model,
     type: 'tool',
   });
+
+  // Record OpenTelemetry GenAI semantic convention metrics
+  const { commonAttributes } = recordGenAiClientMetrics(config, event);
```
bobcatfish: `recordGenAIClientMetrics` is weirding me out a bit b/c if im understanding correctly, it seems to be doing two things at once: it's emitting an event about operation duration AND it's returning `commonAttributes` (which is a bit of confusing name since we already have `getCommonAttributes` in metrics.ts which is a different set of attributes)

I'm wondering about an alternative where `recordGenAiClientTokenUsage` and `recordGenAiClientOperationDuration` are always called explicitly, and maybe they can use a function to determine the "common" attributes dynamically each they they are called? it does mean (in this case) determining that twice instead of once but I'm not sure we need that optimization

bobcatfish (commented):
Okay so it was MY idea to just add these additively vs deprecating old metrics... BUT I'm wondering about any existing overlap with the previous metrics and if we want to call that out / do anything about it.

For example: `gemini_cli.token.usage`

so now we've got

```
  [TOKEN_USAGE]: {
    description: 'Counts the total number of tokens used.',
    valueType: ValueType.INT,
    assign: (c: Counter) => (tokenUsageCounter = c),
    attributes: {} as {
      model: string;
      type: 'input' | 'output' | 'thought' | 'cache' | 'tool';
    },
```

and 

```
  [GEN_AI_CLIENT_TOKEN_USAGE]: {
    description: 'Number of input and output tokens used.',
    unit: 'token',
    valueType: ValueType.INT,
    assign: (h: Histogram) => (genAiClientTokenUsageHistogram = h),
    attributes: {} as {
      'gen_ai.operation.name': string;
      'gen_ai.provider.name': string;
      'gen_ai.token.type': 'input' | 'output';
      'gen_ai.request.model'?: string;
      'gen_ai.response.model'?: string;
      'server.address'?: string;
      'server.port'?: number;
    },
  },
```

I'm wondering if there is anything useful we could do to consolidate these 🤔 

Thoughts:
* (easy/low impact) we could at least declare / operate on these near each other in the code; and mention them both in the docs (and mention one kinda duplicates the other?)
* (little harder with still low impact) instead of introducing a completely new function for recording token usage (`recordGenAiClientTokenUsage`) we could have one function that emits both
* (maybe a bit crazy) could it make sense to add the new attributes to our existing token metric? so then they're nearly the same, and we've (maybe?) improved the existing metric

It seems like we could almost fold them together completely (e.g. emit exactly the same metric with two different names) EXCEPT that our existing metric has extra 'token types' @_@ ('thought' | 'cache' | 'tool' vs just input/output); so maybe we cant COMPLETELY share the same declaration / code but i think we could almost do it

(And I think there's a similiar consideration for API_REQUEST_LATENCY ?)

packages/core/src/telemetry/loggers.ts
jerop: make this simpler with a single concern, thanks!

packages/core/src/telemetry/loggers.ts
jerop: refactored into one function

packages/core/src/telemetry/metrics.test.ts
jerop: Without this, test failures are caused by an **incomplete mock** that was exposed by a new import chain in this branch.

This branch added `import { AuthType } from '../core/contentGenerator.js'` to `metrics.ts`, which creates a module dependency chain that eventually loads `sdk.ts`. When `sdk.ts` loads, it executes:
```typescript
import { DiagConsoleLogger, DiagLogLevel, diag } from '@opentelemetry/api';
diag.setLogger(new DiagConsoleLogger(), DiagLogLevel.INFO);
```

But the @opentelemetry/api mock in the test file was missing `DiagConsoleLogger` and `DiagLogLevel` exports.

jerop:
@bobcatfish thank you for the thorough review and helpful suggestions, please take another look 🙏🏾

packages/core/src/telemetry/metrics.test.ts
bobcatfish: 🤦‍♀️ 🤦‍♀️🤦‍♀️ thanks for the thorough explanation, I wasn't looking hard enough at the context this was in and thought it was an unused variable 🤪  - nothing to see here XD XD XD

packages/core/src/telemetry/metrics.ts
```diff
@@ -767,3 +895,71 @@ export function recordBaselineComparison(
...
+  },
+): void {
+  recordCustomTokenUsageMetrics(config, tokenCount, {
+    model: attributes.model,
+    type: attributes.type,
+  });
+
+  if (
+    (attributes.type === 'input' || attributes.type === 'output') &&
+    attributes.genAiAttributes
```
bobcatfish, line 925: 🥳

bobcatfish (approved):
Thanks for the updates! Looking good 💃

Issue #8941, opened 2025-09-19
Infra: Create Private Github Repo for Internal Testing


richieforeman:
Often times we want to test Github action and deployment changes without broadcasting out on the public channels.  Right now we're using the `dev` tag to sort of signify this, but that's out there for the world to see: https://www.npmjs.com/package/@google/gemini-cli?activeTab=versions

We should:

- [x] Setup a private Github Repo https://github.com/google-gemini/gemini-cli/issues/8941
- [x] Setup a private NPM package https://github.com/google-gemini/gemini-cli/issues/8941
- [ ] Setup all the stuff that will allow the team to test the release workflow in a lower-risk environment.

skeshive:
Per GitHub's requirement, all packages published to its npm registry must be scoped under an organization. We have two main options to handle this for our release workflow:

1) Rename Packages: We can rename our packages to @google-gemini/gemini-cli and @google-gemini/gemini-cli-core within the CI/CD pipeline at release time.

2) Publish a Bundled Version: Alternatively, to avoid renaming packages and updating import paths, we could publish a single, bundled version of @google-gemini/gemini-cli that already includes all its necessary dependencies.

skeshive:
10/2 update: Did not address Richie's feedback yet because I was focused on getting the release out yesterday and on a P0 IDE integration bug. Planning on spending the afternoon working on addressing feedback

mattKorwel:
@skeshive @richieforeman i wanted to talk about this a  bit more.

I feel like we need to pause and design what will be used when.

For instance, Maybe all our validation runs should happen against code in the dev npm repository, and then "release" is basically just promoting a release from dev to real. That keeps all the changes from polluting the npm version history. 

Also, how we use the dev registry for testing of scripts and doing real things, like promoting releases patching, and use the dev repo to test our code. 

Do we have one set of scripts and use github environments (i am leaning this way) https://docs.github.com/en/actions/how-tos/deploy/configure-and-manage-deployments/manage-environments - in this pr rather than having the repo variables coded based on an if statement. we create the variables in the environment and set that for the script. 

Just thinking out loud. I think this pr attached is a good start. and maybe we send it as is to get some early validations but i feel like this task should be replaced by a "think it out" task?

PR #10348, opened 2025-10-01
feat(ci): add ability to publish packages to private github registry for testing

skeshive:
## TLDR
Updates manual release workflow to include registry option for publishing to private github registry instead of public.

## Dive Deeper
The following applies when publishing to the github registry, the npm registry publishing flow should be a no-op.

- Github requires packages published to their registry to be named with organization scope so we need to modify the names of the packages from @google/gemini-cli & @google/gemini-cli-core to @google-gemini/gemini-cli & @google-gemini/gemini-cli-core
- Instead of publishing the entire gemini-cli package, we now just publish the bundled code. Therefore, we remove unnecessary metadata from the package.json file
- Successful manual publish to github registry: https://github.com/google-gemini/gemini-cli/actions/runs/18166685616/job/51710727267

## Testing
- Install package from private github registry:
```
npm install @google-gemini/gemini-cli@0.0.3 --@google-gemini:registry=https://npm.pkg.github.com --//npm.pkg.github.com/:_authToken=${NPM_TOKEN}
```

- Run package
```
 ./node_modules/.bin/gemini
```

- Make sure version in /about is 0.0.3

## Linked issues / bugs
#8941

### DIFF
--- a/.github/actions/publish-release/action.yml
+++ b/.github/actions/publish-release/action.yml
@@ -51,8 +51,7 @@ inputs:
   registry:
     description: 'The registry to publish to.'
     required: false
-    default: 'npm'
-
+    default: 'npm-wombat'
 runs:
   using: 'composite'
   steps:
@@ -121,7 +120,7 @@ runs:
         node ${{ github.workspace }}/scripts/prepare-github-release.js
 
     - name: 'Configure npm for publishing to npm'
-      if: "inputs.registry == 'npm'"
+      if: "inputs.registry != 'github'"
       uses: 'actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020'
       with:
         node-version-file: '${{ inputs.working-directory }}/.nvmrc'
@@ -137,7 +136,7 @@ runs:
         scope: '@google-gemini'
 
     - name: '📦 Publish @google/gemini-cli-core to npm'
-      if: "inputs.registry == 'npm'"
+      if: "inputs.registry != 'github'"
       working-directory: '${{ inputs.working-directory }}'
       env:
         NODE_AUTH_TOKEN: '${{ inputs.wombat-token-core }}'
@@ -163,7 +162,7 @@ runs:
 
     - name: '🔗 Install latest core package'
       working-directory: '${{ inputs.working-directory }}'
-      if: "${{ inputs.dry-run != 'true' && inputs.registry == 'npm' }}"
+      if: "${{ inputs.dry-run != 'true' && inputs.registry != 'github' }}"
       shell: 'bash'
       run: |
         npm install "@google/gemini-cli-core@${{ inputs.release-version }}" \
@@ -171,7 +170,7 @@ runs:
         --save-exact
 
     - name: '📦 Publish @google/gemini-cli to npm'
-      if: "inputs.registry == 'npm'"
+      if: "inputs.registry != 'github'"
       working-directory: '${{ inputs.working-directory }}'
       env:
         NODE_AUTH_TOKEN: '${{ inputs.wombat-token-cli }}'
@@ -197,7 +196,7 @@ runs:
 
     - name: '🔬 Verify NPM release by version'
       uses: './.github/actions/verify-release'
-      if: "${{ inputs.dry-run != 'true' && inputs.force-skip-tests != 'true' && inputs.registry == 'npm' }}"
+      if: "${{ inputs.dry-run != 'true' && inputs.force-skip-tests != 'true' && inputs.registry != 'github' }}"
       with:
         npm-package: '@google/gemini-cli@${{ inputs.release-version }}'
         expected-version: '${{ inputs.release-version }}'
@@ -206,7 +205,7 @@ runs:
 
     - name: '🏷️ Tag release'
       uses: './.github/actions/tag-npm-release'
-      if: "${{ inputs.dry-run != 'true' && inputs.registry == 'npm' }}"
+      if: "${{ inputs.dry-run != 'true' && inputs.registry != 'github' }}"
       with:
         channel: '${{ inputs.npm-tag }}'
         version: '${{ inputs.release-version }}'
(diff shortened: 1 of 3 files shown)


scripts/prepare-github-release.js
```diff
@@ -0,0 +1,63 @@
...
+  process.exit(1);
+}
+
+// Overwrite the .npmrc in the core package to point to the GitHub registry.
+const coreNpmrcPath = path.resolve(rootDir, 'packages/core/.npmrc');
+fs.writeFileSync(
+  coreNpmrcPath,
+  '@google-gemini:registry=https://npm.pkg.github.com/',
+);
+console.log('Wrote .npmrc for @google-gemini scope to packages/core/');
```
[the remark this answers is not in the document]
skeshive: we need to overwrite the root .npmrc to point to Github (the --registry flags used in the workflow don't override this)

.github/actions/publish-release/action.yml
```diff
@@ -48,6 +48,10 @@ inputs:
   gemini_api_key:
     description: 'The API key for running integration tests.'
     required: true
+  registry:
+    description: 'The registry to publish to.'
```
richieforeman, line 52: Please prefer to make this a `type: choice` and include a fixed list of options

ref: https://github.blog/changelog/2021-11-10-github-actions-input-types-for-manual-workflows/

.github/actions/publish-release/action.yml
```diff
@@ -48,6 +48,10 @@ inputs:
...
+    required: false
+    default: 'npm'
```
richieforeman: prefer explicit `npm-wombat` or something.  It's not just naked NPM, it's Google's special flow.

.github/actions/publish-release/action.yml
```diff
@@ -103,14 +107,37 @@ runs:
...
+    - name: 'Configure npm for publishing to GitHub'
+      if: "inputs.registry == 'github'"
+      uses: 'actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020'
+      with:
+        node-version-file: '${{ inputs.working-directory }}/.nvmrc'
+        registry-url: 'https://npm.pkg.github.com'
+        scope: '@google-gemini'
+
+    - name: '📦 Publish @google/gemini-cli-core to npm'
+      if: "inputs.registry == 'npm'"
```
richieforeman: If you consider my comments above, don't forget to update this.

.github/actions/publish-release/action.yml
```diff
@@ -143,9 +183,21 @@ runs:
...
+          --dry-run="${{ inputs.dry-run }}" \
+          --workspace="@google-gemini/gemini-cli" \
+          --no-tag
+
     - name: '🔬 Verify NPM release by version'
       uses: './.github/actions/verify-release'
-      if: "${{ inputs.dry-run != 'true' && inputs.force-skip-tests != 'true' }}"
+      if: "${{ inputs.dry-run != 'true' && inputs.force-skip-tests != 'true' && inputs.registry == 'npm' }}"
       with:
         npm-package: '@google/gemini-cli@${{ inputs.release-version }}'
```
richieforeman, line 201: Opinions:

- This seems fine for now.
- I am working on updating the verify-release flow right now, and it should also probably support multiple choices.
- Can you tag an issue to update verify-release and tag it to me? (You could also do it yourself, but you'll probably merge conflict with my existing inflight work in there).

scripts/prepare-github-release.js
```diff
@@ -0,0 +1,63 @@
...
+    'Error: bundle/ directory not found at project root. Please run `npm run bundle` first.',
+  );
+  process.exit(1);
+}
+
+// Overwrite the .npmrc in the core package to point to the GitHub registry.
+const coreNpmrcPath = path.resolve(rootDir, 'packages/core/.npmrc');
+fs.writeFileSync(
+  coreNpmrcPath,
+  '@google-gemini:registry=https://npm.pkg.github.com/',
```
richieforeman, line 38: Thinking out loud:
- We could consider updating this to take `registry` as an argument
- We could update this file to `prepare-custom-registry-release.js`

That would make this change slightly more flexible (by using an inversion-of-control pattern).

.github/actions/publish-release/action.yml
richieforeman: (Sanity check for myself... Can `type: choice` be in a "sub-action" like this?  I know it's supported in workflows, but I'm fuzzy here.

If this isn't possible, then obviously feel free to ACK, but I would request a comment in the `description` field with the list of possible options.

skeshive: hmm so the parent workflow uses type:choice, this sub-action doesn't show up in the UI to allow the user to select a registry

.github/actions/publish-release/action.yml
```diff
@@ -122,16 +149,29 @@ runs:
...
+      if: "inputs.registry == 'github'"
+      working-directory: '${{ inputs.working-directory }}'
+      env:
+        NODE_AUTH_TOKEN: '${{ inputs.github-token }}'
+      shell: 'bash'
+      run: |
+        npm publish \
+          --dry-run="${{ inputs.dry-run }}" \
+          --workspace="@google-gemini/gemini-cli-core" \
+          --no-tag
```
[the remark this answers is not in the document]
skeshive: this isn't true, I've done non-dry runs and verified the package actually gets published

scripts/prepare-github-release.js
skeshive: hmm each registry may have custom requirements (ex. how the package is name)

I'd prefer waiting to make this more flexible until we have a second registry we want to publish to

.github/actions/publish-release/action.yml
skeshive: done- https://github.com/google-gemini/gemini-cli/issues/10452

scripts/prepare-github-release.js
richieforeman: Opinion: All the more reason to invert the control and take those things in as arguments.  But I agree this could be later.

richieforeman (approved):
LG

mattKorwel (approved):
LGTM

PR #10376, opened 2025-10-01
fix(ui): increase padding of settings dialog

galdawave:
## TLDR

Reduces the height of the settings dialog by setting a 4-line margin. This prevents the dialog from taking up too much vertical space and helps reduce flickering when rendering with the Ink library.

## Dive Deeper

The settings dialog was occupying too much height, which could cause issues with the Ink library's rendering, leading to flickering. By setting `DIALOG_PADDING` to 4 lines in the height calculation in `SettingsDialog.tsx`, we ensure the dialog leaves enough buffer space, improving the stability of the UI.

## Reviewer Test Plan

1.  Open the settings dialog using the `/settings` command.
2.  Observe the height of the dialog. It should be slightly shorter than before.
3.  Interact with the settings (navigate, toggle) and verify that the dialog does not flicker and remains stable.
4.  Resize the terminal window to various heights and ensure the dialog adjusts correctly and maintains the margin of error.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


jacob314 (approved):
lgtm

PR #10377, opened 2025-10-01
refactor(agents): implement submit_final_output tool for agent completion

### LINKED ISSUE #10417: Add "Finish tool" support to subagents

It will allow the agent to choose when to finish instead of forcing it to finish

### Why is this needed?

Better performance on swebench 

### Additional context

_No response_

abhipatel12:
## TLDR

Refactors the `AgentExecutor` to use an explicit `submit_final_output` tool for agent termination and result delivery, replacing the previous two-phase (work followed by extraction) approach. This improves reliability and enables structured, schema-validated outputs defined in `OutputConfig`.

## Dive Deeper

Previously, the `AgentExecutor` ran in two phases: a "work phase" where the agent called tools, and an "extraction phase" where a final prompt asked the model to summarize its findings based on `completion_criteria`. This could be unreliable if the model didn't adhere to the implicit stop condition.

This PR introduces the following changes:

1.  **`OutputConfig` Update (`types.ts`):**
    *   Removed `completion_criteria`.
    *   Added `outputName` (defines the name of the argument in the final tool) and `schema` (Gemini `Schema` object for structured output validation).

2.  **`AgentExecutor` Refactor (`executor.ts`):**
    *   Dynamically injects a `submit_final_output` tool into the agent's available tools if `OutputConfig` is defined.
    *   Updated the system prompt to explicitly instruct the agent that calling this tool is the *only* way to complete the task.
    *   The execution loop now continues until this tool is called successfully (or other termination conditions like timeout/max turns are met).
    *   Implemented a "nudge" mechanism: if the model stops calling tools but hasn't submitted output, it is explicitly prompted to call `submit_final_output`.
    *   Includes error handling for the submit tool (e.g., missing required arguments, duplicate calls in the same turn).

3.  **Agent Updates (`codebase-investigator.ts`):**
    *   Updated the `CodebaseInvestigatorAgent` definition to comply with the new `OutputConfig` interface, moving the previous completion criteria into the schema's description field.

4.  **Testing (`executor.test.ts`):**
    *   Extensively updated and refactored tests to cover the new workflow.
    *   Added test cases for the happy path, nudging behavior, error handling (missing args), and parallel tool calls involving submission.

## Reviewer Test Plan

Ensure that `enableSubagents` is true in `settings.json` and then run a command asking for the main agent to use the codebase investigator.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  |✅| ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt |✅| -   | -   |

### DIFF
--- a/packages/core/src/agents/executor.ts
+++ b/packages/core/src/agents/executor.ts
@@ -40,13 +40,13 @@ import { parseThought } from '../utils/thoughtUtils.js';
 /** A callback function to report on agent activity. */
 export type ActivityCallback = (activity: SubagentActivityEvent) => void;
 
-const SUBMIT_OUTPUT_TOOL_NAME = 'submit_final_output';
+const TASK_COMPLETE_TOOL_NAME = 'complete_task';
 
 /**
  * Executes an agent loop based on an {@link AgentDefinition}.
  *
  * This executor runs the agent in a loop, calling tools until it calls the
- * mandatory `submit_final_output` tool to deliver its result.
+ * mandatory `complete_task` tool to signal completion.
  */
 export class AgentExecutor {
   readonly definition: AgentDefinition;
@@ -179,25 +179,22 @@ export class AgentExecutor {
           break;
         }
 
-        // If the model stops calling tools, check if it has submitted output.
+        // If the model stops calling tools without calling complete_task, it's an error.
         if (functionCalls.length === 0) {
-          if (this.definition.outputConfig) {
-            // Nudge the model to call the submit tool.
-            const nudgeMessage = `You have stopped calling tools, but you have not called \`${SUBMIT_OUTPUT_TOOL_NAME}\` to deliver your final result. If you are finished, you MUST call \`${SUBMIT_OUTPUT_TOOL_NAME}\` now. If you need to do more work, please continue using other tools.`;
-            currentMessage = { role: 'user', parts: [{ text: nudgeMessage }] };
-            continue;
-          } else {
-            // If no output is expected, we are done.
-            terminateReason = AgentTerminateMode.GOAL;
-            break;
-          }
+          terminateReason = AgentTerminateMode.ERROR;
+          finalResult = `Agent stopped calling tools but did not call '${TASK_COMPLETE_TOOL_NAME}' to finalize the session.`;
+          this.emitActivity('ERROR', {
+            error: finalResult,
+            context: 'protocol_violation',
+          });
+          break;
         }
 
-        const { nextMessage, submittedOutput } =
+        const { nextMessage, submittedOutput, taskCompleted } =
           await this.processFunctionCalls(functionCalls, signal, promptId);
 
-        if (submittedOutput !== null) {
-          finalResult = submittedOutput;
+        if (taskCompleted) {
+          finalResult = submittedOutput ?? 'Task completed successfully.';
           terminateReason = AgentTerminateMode.GOAL;
           break;
         }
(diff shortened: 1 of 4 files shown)


packages/core/src/agents/executor.ts
```diff
@@ -182,46 +179,42 @@ export class AgentExecutor {
           break;
         }
 
-        // If the model stops calling tools, the work phase is complete.
+        // If the model stops calling tools, check if it has submitted output.
         if (functionCalls.length === 0) {
+          if (this.definition.outputConfig) {
```
silviojr: Why do we only call the SUBMIT_OUTPUT_TOOL_NAME when there's an outputconfig? 

Shouldn't SUBMIT_OUTPUT_TOOL_NAME be a "finish" tool with the option of having an output or not?  In that way we always let the subagent decide when it's done.

packages/core/src/config/config.ts
```diff
@@ -439,7 +439,7 @@ export class Config {
     this.useModelRouter = params.useModelRouter ?? false;
     this.enableMessageBusIntegration =
       params.enableMessageBusIntegration ?? false;
-    this.enableSubagents = params.enableSubagents ?? false;
+    this.enableSubagents = params.enableSubagents ?? true;
```
silviojr: was this for testing?

packages/cli/src/config/config.ts
```diff
@@ -708,7 +708,7 @@ export async function loadCliConfig(
     useModelRouter,
     enableMessageBusIntegration:
       settings.tools?.enableMessageBusIntegration ?? false,
-    enableSubagents: settings.experimental?.enableSubagents ?? false,
+    enableSubagents: settings.experimental?.enableSubagents ?? true,
```
silviojr: testing?

packages/cli/src/config/settingsSchema.ts
```diff
@@ -1006,7 +1006,7 @@ const SETTINGS_SCHEMA = {
         label: 'Enable Subagents',
         category: 'Experimental',
         requiresRestart: true,
-        default: false,
+        default: true,
```
silviojr: testing?

packages/cli/src/nonInteractiveCli.ts
```diff
@@ -94,6 +94,10 @@ export async function runNonInteractive(
         query = processedQuery as Part[];
       }
 
+      query.push({
```
silviojr: testing?

packages/core/src/agents/executor.ts
```diff
@@ -349,45 +342,107 @@ export class AgentExecutor {
...
+
+      // Handle standard tools
+      if (!allowedToolNames.has(functionCall.name as string)) {
         console.warn(
           `[AgentExecutor] Agent '${this.definition.name}' attempted to call ` +
-            `unauthorized tool '${call.name}'. This call has been blocked.`,
+            `unauthorized tool '${functionCall.name}'. This call has been blocked.`,
         );
-        return false;
+        continue;
```
silviojr, line 468: I think we have to somehow tell the model that it tried to call a tool that it's not allowed, no? In a function response with error?

silviojr (approved):

packages/core/src/agents/executor.ts
```diff
@@ -349,45 +342,107 @@ export class AgentExecutor {
...
+
+    let submittedOutput: string | null = null;
+
+    // We'll collect promises for the tool executions
+    const toolExecutionPromises: Array<Promise<Part[] | void>> = [];
+    // And we'll need a place to store the synchronous results (like submit_final_output or blocked calls)
+    const syncResponseParts: Part[] = [];
+
+    for (const functionCall of functionCalls) {
+      const callId = functionCall.id ?? `${functionCall.name}-${Date.now()}`;
```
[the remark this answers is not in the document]
abhipatel12: Done

packages/cli/src/config/config.ts
abhipatel12: Yep, removed

packages/cli/src/config/settingsSchema.ts
abhipatel12: yes, removed

packages/cli/src/nonInteractiveCli.ts
abhipatel12: yes, removed

silviojr:
Fixes issue: https://github.com/google-gemini/gemini-cli/issues/10417

packages/core/src/agents/executor.ts
abhipatel12: Good point. This will be important long term as well. Made the change. 

also removed the nudge so there's quicker feedback for when our system prompt is not being followed correctlty.

packages/core/src/config/config.ts
abhipatel12: Yes, removed

packages/core/src/agents/executor.ts
abhipatel12: Yes, great catch. Added this in

silviojr (approved):

PR #10383, opened 2025-10-01
feat: Add markdown toggle (alt+m) to switch between rendered and raw…

### LINKED ISSUE #6499: Toggle between raw and rendered mode for markdown

Add the ability to google between raw and rendered mode of markdown.

AI studio recently added this.
https://x.com/patloeber/status/1957133963459174654

### Why is this needed?

Sometimes we have markdown rendering glitches and sometimes users just want to copy the markdown to a file. 
In particular, no solution we do for rendering markdown tables to the terminal will be perfect (truncation, ascii art limitations) so giving the users the escape hatch to view the raw output would be useful.

This will reduce the pressure on us to make the markdown output perfect. The raw rendering mode should still syntax highlight the markdown for readability.

srivatsj:
… display

Implements feature request from #6499 to toggle between rendered and raw markdown mode.

- Add alt+m keybinding to toggle markdown rendering globally
- Add renderMarkdown setting to UI config (default: true)
- Update MarkdownDisplay to support raw mode with syntax highlighting
- Update GeminiMessage, GeminiMessageContent, and ToolMessage to respect toggle
- Add UIStateContext to test utilities for component testing
- Visual indicator in footer (currently commented out)

Fixes #6499

## TLDR

- Add alt+m keybinding to toggle markdown rendering globally
- Add renderMarkdown setting to UI config (default: true)
- Update MarkdownDisplay to support raw mode with syntax highlighting
- Update GeminiMessage, GeminiMessageContent, and ToolMessage to respect toggle
- Add UIStateContext to test utilities for component testing
- Visual indicator in footer (currently commented out)

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


packages/cli/src/ui/keyMatchers.test.ts
```diff
@@ -55,6 +55,7 @@ describe('keyMatchers', () => {
       key.ctrl && key.name === 't',
     [Command.TOGGLE_IDE_CONTEXT_DETAIL]: (key: Key) =>
       key.ctrl && key.name === 'g',
+    [Command.TOGGLE_MARKDOWN]: (key: Key) => key.ctrl && key.name === 'b',
```
jacob314: Bikeshedding topic (https://thedecisionlab.com/biases/bikeshedding)
this can't be ctrl+b as that is unfortunately already taken by backspace i the input prompt. How about ctrl+m unless someone has a better idea. That isn't taken and matches 'markdown'

packages/cli/src/ui/utils/MarkdownDisplay.tsx
```diff
@@ -31,9 +32,29 @@ const MarkdownDisplayInternal: React.FC<MarkdownDisplayProps> = ({
   isPending,
   availableTerminalHeight,
   terminalWidth,
+  renderMarkdown = true,
 }) => {
+  const settings = useSettings();
+
   if (!text) return <></>;
 
+  // Raw markdown mode - display syntax-highlighted markdown without rendering
```
jacob314, line 41: great idea to use the built in markdown syntax highlighting. love it! going a step further, hide the line numbers as those are not relevant here and are confusing because of how Gemini output is chunked. Good news is CodeColorizer already internally has a way to hide line numbers looking at settings. you just need to get gemini CLI to also add a prop for it.

 For example this is what you see today because of how we split blocks up. Kindof cool as you can see how the blocks are split but confusing if you weren't debugging are markdown splitting.

Image

packages/cli/src/ui/components/Footer.tsx
```diff
@@ -139,35 +130,55 @@ export const Footer: React.FC = () => {
...
+              <ContextUsageDisplay
+                promptTokenCount={promptTokenCount}
+                model={model}
+                terminalWidth={terminalWidth}
+              />
+            </Text>
+            {showMemoryUsage && <MemoryUsageDisplay />}
+          </Box>
+          <Box alignItems="center" paddingLeft={2}>
+            {/* Markdown mode indicator - uncomment to enable
```
jacob314: Fyi @miguelsolorio we probably don't want to enable this in prod but we should figure out a reasonable way to surface that we are in raw markdown display mode. One idea is to look at how we flash Press Ctrl-C again to exit if you press Ctrl+C. We could similarly surface "In raw markdown display mode. Press Ctrl-M again to exit" or similar.

jacob314 (commented):
This is really helpful. Added some comments. Approved after those are addressed and there is a snapshot test fo the raw markdown display cases in GeminiMessage and ToolMessage.

packages/cli/src/ui/utils/MarkdownDisplay.tsx
srivatsj: Done

packages/cli/src/ui/keyMatchers.test.ts
srivatsj: ctrl+m does not work. This is what Gemini says - 

**Ctrl+M = ASCII 13 (0x0D) = Carriage Return** is **True**.

Here's a breakdown of the parts and the context around the "Enter key" equivalence:

1.  **Ctrl+M:** This key combination is the standard way to input the control character with the decimal value $13$ in many terminals and text environments. This is due to the historical design of ASCII control codes.
2.  **ASCII 13 (0x0D):** This is the decimal and hexadecimal value for the **Carriage Return (CR)** control character.
3.  **Carriage Return (CR):** The function of CR is to move the cursor to the **beginning of the current line**. (Historically, on a typewriter or teletype, it would physically return the carriage to the starting position).

**Equivalence to "Enter Key" (Context Dependent):**

The statement that Carriage Return is the **Same as Enter key** is often **true in a functional sense**, but it's important to understand the nuance:

* **Windows/DOS:** The "Enter" key typically sends a sequence of **two** characters:
    * **Carriage Return (CR), ASCII 13 ($\text{0x0D}$)**, followed by
    * **Line Feed (LF), ASCII 10 ($\text{0x0A}$)**.
    This is known as **CRLF**. It's needed to move the cursor to the beginning of the line *and* advance to the next line.
* **Unix/Linux/Modern macOS:** The "Enter" key typically sends only one character:
    * **Line Feed (LF), ASCII 10 ($\text{0x0A}$)**.
    This character is often interpreted by the system to perform both the line feed *and* the carriage return (i.e., it's the "newline" character).
* **Command Line Interfaces/Terminals:** When you press the **Enter** key in a terminal, it often results in a Carriage Return (CR) character being sent to the application or system, or a CRLF sequence, but the *action* is typically to submit a command or input, making it functionally equivalent to what you might do with Ctrl+M in certain contexts.

So, while $\text{Ctrl}+\text{M}$ *is* Carriage Return (ASCII 13), the **Enter** key's function is sometimes implemented as CR, sometimes as LF, and sometimes as CRLF, depending on the operating system or application.

packages/cli/src/ui/components/messages/GeminiMessage.test.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { renderWithProviders } from '../../../test-utils/render.js';
+
+describe('<GeminiMessage /> - Raw Markdown Display Snapshots', () => {
+  const baseProps = {
+    text: 'Test **bold** and `code` markdown\n\n```javascript\nconst x = 1;\n```',
+    isPending: false,
+    terminalWidth: 80,
+  };
+
+  it('renders with renderMarkdown=true (default)', () => {
```
jacob314: ask Gemini CLI to refactor these as a parameterized test. That will significantly reduce the boilerplate

packages/cli/src/ui/components/messages/ToolMessageRawMarkdown.test.tsx
```diff
@@ -0,0 +1,48 @@
...
+    name: 'test-tool',
+    description: 'A tool for testing',
+    resultDisplay: 'Test **bold** and `code` markdown',
+    status: ToolCallStatus.Success,
+    terminalWidth: 80,
+    confirmationDetails: undefined,
+    emphasis: 'medium',
+  };
+
+  it('renders with renderMarkdown=true (default)', () => {
```
jacob314: this should also be a parameterized test with the parameter renderMarkdown being both true and false

packages/cli/src/ui/utils/MarkdownDisplay.tsx
```diff
@@ -40,13 +40,26 @@ const MarkdownDisplayInternal: React.FC<MarkdownDisplayProps> = ({
...
+        } as typeof settings)
+      : undefined;
     const colorizedMarkdown = colorizeCode(
       text,
       'markdown',
       availableTerminalHeight,
       terminalWidth - CODE_BLOCK_PREFIX_PADDING,
       undefined,
-      settings,
+      settingsWithoutLineNumbers,
```
jacob314: don't pass in the whole settings object. instead add an extra prop to colorizeCode.  pass in the override to never use line numbers. otherwise your MarkdownDisplay is doing a bunch of logic to mess merge big settings objects on each frames when all you are trying to do is say "don't render line numbers"

packages/cli/src/ui/components/messages/GeminiMessage.test.tsx
srivatsj: Done

packages/cli/src/ui/components/messages/ToolMessageRawMarkdown.test.tsx
srivatsj: Done

packages/cli/src/ui/utils/MarkdownDisplay.tsx
srivatsj: Done

packages/cli/src/ui/keyMatchers.test.ts
srivatsj: Updated key to alt+shift+m.

Tested on gnome and vscode terminal. Need to test on iterm2 and ghostly
 
Screenshot from 2025-10-03 10-38-55
Screenshot from 2025-10-03 10-37-38

srivatsj: Updated binding to option + M. Tested and verified on mac for below terminals.

- Mac terminal
- Iterm2
- VS-code terminal
- Ghossty

Screenshot 2025-10-06 at 10 54 05 AM
Screenshot 2025-10-06 at 10 34 04 AM
Screenshot 2025-10-06 at 10 33 21 AM
Screenshot 2025-10-06 at 10 31 00 AM

packages/cli/src/config/settingsSchema.ts
```diff
@@ -361,6 +361,15 @@ const SETTINGS_SCHEMA = {
         description: 'Show citations for generated text in the chat.',
         showInDialog: true,
       },
+      renderMarkdown: {
```
jacob314: is this needed? seems like users should just be using command-M to toggle instead.

packages/cli/src/ui/components/Footer.tsx
jacob314: Can you work with @miguelsolorio to figure out a solution to this?

packages/cli/src/ui/contexts/KeypressContext.tsx
```diff
@@ -603,6 +603,83 @@ export function KeypressProvider({
       if (key.name === 'return' && key.sequence === `${ESC}\r`) {
         key.meta = true;
       }
+
+      // Map common Alt+letter combinations that produce Unicode characters on macOS
+      // This handles terminals that don't send ESC sequences (like iTerm2 default)
+      const altUnicodeMap: {
```
jacob314: move this to be a top level constant rather than inline for consistency.

packages/cli/src/ui/contexts/KeypressContext.tsx
```diff
@@ -603,6 +603,83 @@ export function KeypressProvider({
...
+      ) {
+        const mapped = altUnicodeMap[unicodeChar];
+        broadcast({
+          name: mapped.key,
+          ctrl: false,
+          meta: true,
+          shift: mapped.shift,
+          paste: isPaste,
+          sequence: key.sequence,
+        });
```
jacob314: all these keypress context changes need tests.ideally they would be separate pr so we can review separate from the UX discussions for the markdown display.

packages/cli/src/config/settingsSchema.ts
srivatsj: removed.

jacob314:
Please update the pr description to mention alt-m rather than ctrl-B. to avoid confusion if someone is looking through the changelog.

packages/cli/src/ui/components/Composer.test.tsx
```diff
@@ -44,6 +44,10 @@ vi.mock('./ShellModeIndicator.js', () => ({
   ShellModeIndicator: () => <Text>ShellModeIndicator</Text>,
 }));
 
+vi.mock('./RawMarkdownIndicator.js', () => ({
```
jacob314: Gemini CLI likes to mock too much. This isn't something that should be mocked.

packages/cli/src/ui/components/messages/ToolMessage.tsx
```diff
@@ -56,6 +57,7 @@ export const ToolMessage: React.FC<ToolMessageProps> = ({
   ptyId,
   config,
 }) => {
+  const { renderMarkdown: globalRenderMarkdown } = useUIState();
```
jacob314: nit: just call this render markdown. globalRenderMarkdown is an odd name that makes it sound like this is a global variable.

packages/cli/src/ui/components/RawMarkdownIndicator.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+import type React from 'react';
+import { Box, Text } from 'ink';
+import { theme } from '../semantic-colors.js';
+
+export const RawMarkdownIndicator: React.FC = () => (
+  <Box>
+    <Text>
+      raw markdown mode
+      <Text color={theme.text.secondary}> (alt+m to toggle) </Text>
```
jacob314: make this platform aware. Mac users have no idea what "alt" is. Call it "option+m" for them.

jacob314 (commented):
This is looking great! Approved after these comments are addressed.

packages/cli/src/ui/components/Composer.test.tsx
srivatsj: done

packages/cli/src/ui/components/RawMarkdownIndicator.tsx
srivatsj: done

jacob314 (approved):
lgtm

PR #10392, opened 2025-10-02
feat: allow editing queued messages with up arrow key

### LINKED ISSUE #10390: Support editing of queued messages

add the functionality into the queued messages to polish the UX to make it more clear as well as allowing users to easily keypress up so that the queued message can be changed
-- @abhipatel12 

### Why is this needed?

key up should support queued messages.

### Additional context

_No response_

akhil29:
## Summary
- Enables users to edit queued messages by pressing the up arrow key when input is empty
- Adds visual hint "Queued (press ↑ to edit):" to make the feature discoverable
- Improves UX by allowing quick edits to queued messages without needing to retype them

## Changes
- Added `popLastMessage()` and `hasMessages()` functions to `useMessageQueue` hook
- Modified up arrow key handler in `InputPrompt` to check queued messages before navigating command history
- Updated `QueuedMessageDisplay` component with hint text and adjusted indentation
- Wired queue functions through `UIActions` context to make them accessible in input component

## Behavior
When input is empty and up arrow is pressed:
1. Checks for queued messages first
2. If found, loads the most recent message into input and removes it from queue
3. User can edit and re-submit
4. Falls back to command history when queue is empty

## Test Plan
- [ ] Manually test queuing messages during streaming
- [ ] Verify up arrow loads queued message when input is empty
- [ ] Confirm message is removed from queue after loading
- [ ] Test fallback to command history when queue is empty
- [ ] Verify hint text displays correctly
- [ ] Run existing tests to ensure no regressions

Fixes #10390


packages/cli/src/ui/components/InputPrompt.tsx
```diff
@@ -573,6 +577,16 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
...
+          // Check for queued messages first when input is empty
+          if (buffer.text.trim() === '' && hasMessages && popLastMessage) {
+            if (hasMessages()) {
+              const lastMessage = popLastMessage();
+              if (lastMessage) {
+                buffer.setText(lastMessage);
+                return;
+              }
+            }
+          }
```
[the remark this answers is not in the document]
abhipatel12: +1 can we refactor this into a single reusable function?

packages/cli/src/ui/components/InputPrompt.tsx
abhipatel12, line 1: This is a pretty complex component and we've added some business logic here that introduces new behaviors. Can we please add some form of tests?

abhipatel12 (commented):
Thanks for the patience! Had a few comments that i've left below. Overall, had a question about how the editing of queued messages works.

All queued messages are combined into a single block (joined by `\n\n`) and sent to the model in `useMessageQueue.ts`. But then for editing, pressing the up-arrow key only pulls the single most recent message into the input for editing in `popLastMessage`.

This could be confusing for the user. They see a queue of multiple messages that will be sent as one unit, but the edit feature only allows them to modify a small fragment of that unit. And additionally, from local testing, the edit only let me pop the last message in that queue and if i hit up one more time, it brings me to my last submitted message and not the others in the queue

akhil29:
@abhipatel12 done.
PTAL again.

abhipatel12 (commented):
Hey thanks for updating. There seems to be a regression introduced here. 

When you first start up, try hitting the up key. It should bring up the last message you sent. But it seems like this doesn't work anymore.

akhil29:
Nice catch @abhipatel12 . fixed it and added an explicit test case to catch this. PTAL.

packages/cli/src/ui/components/InputPrompt.test.tsx
```diff
@@ -2109,6 +2109,199 @@ describe('InputPrompt', () => {
...
+      const { stdin, unmount } = renderWithProviders(
+        <InputPrompt {...props} />,
+      );
+      await wait();
+
+      // Press up arrow
+      stdin.write('\u001B[A');
+      await wait();
+
+      // Should call popAllMessages
```
abhipatel12: Can you remove some of the comments in this test file that are not necessary

package-lock.json
abhipatel12, line 1: Please remove the changes in this file

akhil29:
Fixed @abhipatel12

abhipatel12 (approved):
LGTM! Thanks for making these changes!

PR #10427, opened 2025-10-02
Fix oauth support for MCP servers

### LINKED ISSUE #10165: MCP SSE errors
### What happened?

When starting up Gemini, I get errors connecting to my specified MCP servers:

```
│ Debug Console (ctrl+o to close)                                               │
│                                                                               │
│ ℹ Authenticated via "gemini-api-key".                                        │
│ ✖ MCP ERROR (Notion): Error: SSE error: Non-200 status code (401)            │
│ ✖ Error during discovery for server 'Notion': SSE error: Non-200 status code │
│    (401)                                                                      │
│ ✖ MCP ERROR (context7): Error: SSE stream disconnected: TypeError:           │
│    terminated (x2)
```

There should be no errors, since both MCP servers are specified correctly (they used to connect successfully).

### Client information

<details>

```console
> /about
│ CLI Version     0.6.1                         │
│ Git Commit      f3078136                      │
│ Model           gemini-2.5-pro                │
│ Sandbox         no sandbox                    │
│ OS              linux                         │
│ Auth Method     gemini-api-key
```

</details>

### Login information

_No response_

_No response_

jakemac53:
## TLDR

Previously during tool discovery we would go through a different code path than the standard one to connect to the MCP server, and this code path did not support OAUTH. This caused OAUTH mcp servers to not connect at all.

I changed that logic to use the same top level function that is used by other code paths.

## Dive Deeper

Longer term it seems to me like we are connecting to these servers repeatedly, and through many different code paths which is very confusing. The culprit in this case was the McpClientManager which has a `discoverAllMcpTools` function which calls McpClient.connect, but then we also have a top level `connectAndDiscover` function in mcp-client.ts which this mcp client manager does not use, and that function would do the oauth stuff correctly.

We should try and consolidate all this code into the McpClientManager or something along those lines to reduce duplicate code paths.

## Reviewer Test Plan

Attempt to connect to an MCP server that supports oauth.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes https://github.com/google-gemini/gemini-cli/issues/10165

### DIFF
--- a/packages/core/src/tools/mcp-client.ts
+++ b/packages/core/src/tools/mcp-client.ts
@@ -55,6 +55,8 @@ export type DiscoveredMCPPrompt = Prompt & {
 export enum MCPServerStatus {
   /** Server is disconnected or experiencing errors */
   DISCONNECTED = 'disconnected',
+  /** Server is actively disconnecting */
+  DISCONNECTING = 'disconnecting',
   /** Server is in the process of connecting */
   CONNECTING = 'connecting',
   /** Server is connected and ready to use */
@@ -83,7 +85,6 @@ export class McpClient {
   private client: Client | undefined;
   private transport: Transport | undefined;
   private status: MCPServerStatus = MCPServerStatus.DISCONNECTED;
-  private isDisconnecting = false;
 
   constructor(
     private readonly serverName: string,
@@ -98,12 +99,10 @@ export class McpClient {
    * Connects to the MCP server.
    */
   async connect(): Promise<void> {
-    if (this.client) {
-      // Already connected
-      return;
-    }
-    if (this.isDisconnecting) {
-      throw new Error('Cannot reconnect while still disconnecting');
+    if (this.status !== MCPServerStatus.DISCONNECTED) {
+      throw new Error(
+        `Can only connect when the client is disconnected, current state is ${this.status}`,
+      );
     }
     this.updateStatus(MCPServerStatus.CONNECTING);
     try {
@@ -115,10 +114,10 @@ export class McpClient {
       );
       const originalOnError = this.client.onerror;
       this.client.onerror = (error) => {
-        if (originalOnError) originalOnError(error);
-        if (this.isDisconnecting) {
+        if (this.status !== MCPServerStatus.CONNECTED) {
           return;
         }
+        if (originalOnError) originalOnError(error);
         console.error(`MCP ERROR (${this.serverName}):`, error.toString());
         this.updateStatus(MCPServerStatus.DISCONNECTED);
       };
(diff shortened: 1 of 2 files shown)


packages/core/src/tools/mcp-client.ts
```diff
@@ -188,11 +168,12 @@ export class McpClient {
     updateMCPServerStatus(this.serverName, status);
   }
 
-  private async createTransport(): Promise<Transport> {
-    return createTransport(this.serverName, this.serverConfig, this.debugMode);
-  }
-
   private async discoverTools(cliConfig: Config): Promise<DiscoveredMCPTool[]> {
+    if (!this.client) {
```
cornmander: Can you wrap this in a helper and use the helper in discoverTools and discoverPrompts? verifyConnected() or something like this.

packages/core/src/tools/mcp-client.ts
```diff
@@ -168,12 +142,18 @@ export class McpClient {
    * Disconnects from the MCP server.
    */
   async disconnect(): Promise<void> {
+    if (this.isDisconnecting) return;
```
cornmander: Can we do something simpler here, and just check if client is set? Why do we need a separate state variable?

packages/core/src/tools/mcp-client.ts
```diff
@@ -92,51 +92,25 @@ export class McpClient {
...
-        timeout: this.serverConfig.timeout,
-      });
-
+      this.client = await connectToMcpServer(
+        this.serverName,
+        this.serverConfig,
+        this.debugMode,
+        this.workspaceContext,
+      );
       this.updateStatus(MCPServerStatus.CONNECTED);
```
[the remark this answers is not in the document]
jakemac53: Yikes it looks like the MCP server state is also managed separately today for the top level `connectToMcpServer` function versus things going through `McpClient.connect`. I have just added an extra onerror handler here so _both_ states will be update but this is another thing that needs to be cleaned up.

packages/core/src/tools/mcp-client.ts
```diff
@@ -168,12 +142,18 @@ export class McpClient {
...
     if (this.transport) {
       await this.transport.close();
     }
-    this.client.close();
+    if (client) {
+      await client?.close();
+    }
     this.updateStatus(MCPServerStatus.DISCONNECTED);
+    this.isDisconnecting = false;
   }
```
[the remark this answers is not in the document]
jakemac53: I update this so calling `connect` while still disconnecting is an error. That looks safe to do (current code does await `disconnect` calls).

packages/core/src/tools/mcp-client.ts
cornmander: I see the code in the other code path is doing the same.

jakemac53: Sure, I could do that

packages/core/src/tools/mcp-client.ts
jakemac53: done

jacob314 (commented):
Should we add additional tests to mcp-client.test.ts to avoid regressing this?

jakemac53:
> Should we add additional tests to mcp-client.test.ts to avoid regressing this?

We should but it will take some work to set up, I am trying to write some tests though

jakemac53:
@jacob314 I did add some tests

scidomino:
FYI, you should merge changes from main since it contains a fix that makes integration test much less flakey.

jacob314 (approved):
lgtm

PR #10431, opened 2025-10-02
fix(vscode): suppress update and install messages in managed IDEs

skeshive:
## TLDR

This PR prevents the VS Code extension from showing installation and update notifications when running in "managed" environments like Cloud Shell and Firebase Studio. Since users in these environments don't manage the extension themselves, these pop-ups are confusing and un-actionable.

This change also improves the output of the release check script for better clarity.


richieforeman (approved):
LG

PR #10435, opened 2025-10-02
Rationalize different Extension typings

zbirkenbuel:
## TLDR

Effectively just typing updates to pave the way for more easily handling extensions during dynamic loading/unloading within a session

## Dive Deeper

Extension and GeminiGLIExtension had drifted and there were references to the internal file type of the extension json file. Collapsed the two typings into one and ensured the internal file structure was not accessed outside of loading the extension.  Also updated the logic copied into the a2a package to use the new typing (though still duplicates the loading logic)

This is a preparation for being able to keep all extensions loaded but just marked as active or inactive (which will then enable dynamic activation)  This was mostly because both the GeminiCLIExtension in Core and the Extension type in CLI both didn’t have enough information on them to individually support dynamic enablement.  It made sense to standardize on GeminiCLIExtension

This included adding an “IsActive” bit into the core typing.

## Reviewer Test Plan

Core tests pass, running locally extensions are still loaded as expected and report status as expected

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Internal discussion about paving the way to support dynamic loading (okay to consider during a period of cleanup)

### DIFF
--- a/packages/a2a-server/src/config/extension.ts
+++ b/packages/a2a-server/src/config/extension.ts
@@ -6,7 +6,11 @@
 
 // Copied exactly from packages/cli/src/config/extension.ts, last PR #1026
 
-import type { MCPServerConfig } from '@google/gemini-cli-core';
+import type {
+  MCPServerConfig,
+  ExtensionInstallMetadata,
+  GeminiCLIExtension,
+} from '@google/gemini-cli-core';
 import * as fs from 'node:fs';
 import * as path from 'node:path';
 import * as os from 'node:os';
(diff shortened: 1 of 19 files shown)


packages/a2a-server/src/config/extension.ts
```diff
@@ -6,55 +6,63 @@
...
+ * If information is required for manipulating extensions (load, unload, update)
+ * outside of the loading process that data needs to be stored on the
+ * GeminiCLIExtension class defined in Core.
+ */
+interface ExtensionConfig {
   name: string;
   version: string;
   mcpServers?: Record<string, MCPServerConfig>;
   contextFileName?: string | string[];
+  excludeTools?: string[];
```
jakemac53, line 35: Was this just missing before?

jakemac53 (approved):
Looks great thanks!

packages/a2a-server/src/config/extension.ts
zbirkenbuel: Yeah, it had drifted out of date from when the file contents were originally copied over.

Issue #5256, opened 2025-07-31
Configurable Permission Prompts for Shell Commands


aleventhal:
### What would you like to be added?

Introduce a more fine-grained, configurable system for shell command execution permissions. This would allow users to define policies on which commands require explicit confirmation and which can be executed without a prompt.

This could be implemented via a configuration file (e.g., `~/.gemini/settings.json` or a new policy file) with sections to define:

1.  **Always Prompt:** A list of commands (and potentially patterns/regex) that *always* require user confirmation. Examples:
    *   `rm`
    *   `mkdir`
    *   `mv`
    *   `git reset`
    *   `git clean`
    *   `hg update --clean`
    *   `hg purge`
    *   `sudo *`

TBD: how this interacts with yolo mode.

2.  **Never Prompt (Safelist):** A list of commands considered safe to execute without confirmation. These are typically read-only operations or have well-contained side effects. Examples:
    *   `ls`
    *   `cat`
    *   `grep`
    *   `cd`
    *   `pwd`
    *   `blaze build`
    *   `blaze test`
    *   `git status`
    *   `git diff`
    *   `wc`
    *   `echo`
    *   (This list is illustrative, users should be able to configure it, although sensible defaults might be nice)

Any command not matching these lists could default to the current behavior of prompting.

### Why is this needed?

**Why is this needed?**

Currently, Gemini CLI prompts for user permission before executing *any* shell command. While this is safe in principle, it leads to significant **prompt fatigue** for developers. Non-destructive commands (e.g., `ls`, `cat`, and certain tools) require the same level of confirmation as potentially dangerous ones (e.g., `rm`, `git reset`). This constant prompting for safe commands causes "click fatigue," leading developers to habitually approve without scrutiny. This behavior (e.g., reflexively using "Allow always") undermines security, as a truly dangerous command might be missed among the noise.

The only alternative is the `-y` / `--yolo` flag, which swings the pendulum too far in the other direction, auto-approving *all* commands, which is a significant security risk.

### Additional context

**Additional Context:**

*   **Limitations of Current Approaches:** Reliance on `GEMINI.md` to instruct the agent on when to ask for permission is not a robust solution. LLMs can suffer from context limitations, instruction drift, or misinterpretations, making natural language instructions an unreliable mechanism for enforcing security boundaries.
*   **Security Principle Alignment:** This feature aligns with several key security principles:
[...]

PR #10498, opened 2025-10-03
Shell approval rework

cornmander:
Reworks shell approval to do a full parse, which allows us to evaluate nested subcommands. Also uses powershell to parse powershell, which improves the fidelity of parsing on windows.

Adds tests to confirm the new behavior. I've also tested manually by asking the agent to constructed nested substitutions and chained commands.

### DIFF
--- a/packages/core/src/utils/shell-utils.ts
+++ b/packages/core/src/utils/shell-utils.ts
@@ -4,12 +4,19 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+import { createRequire } from 'node:module';
 import type { AnyToolInvocation } from '../index.js';
 import type { Config } from '../config/config.js';
 import os from 'node:os';
 import { quote } from 'shell-quote';
 import { doesToolInvocationMatch } from './tool-utils.js';
-import { spawn, type SpawnOptionsWithoutStdio } from 'node:child_process';
+import {
+  spawn,
+  spawnSync,
+  type SpawnOptionsWithoutStdio,
+} from 'node:child_process';
+import type { Node } from 'web-tree-sitter';
+import { Language, Parser } from 'web-tree-sitter';
 
 export const SHELL_TOOL_NAMES = ['run_shell_command', 'ShellTool'];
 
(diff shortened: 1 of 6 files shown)


esbuild.config.js
```diff
@@ -30,6 +30,8 @@ const external = [
   '@lydell/node-pty-linux-x64',
   '@lydell/node-pty-win32-arm64',
   '@lydell/node-pty-win32-x64',
+  'tree-sitter',
```
mattKorwel: Fwiw @NTaylorMullen has asked us to not add tree sitter due to added external dependencies on at least 2 previous occasions. its worth validating the use case here before proceeding.

packages/core/src/utils/shell-utils.ts
```diff
@@ -160,41 +325,72 @@ export function splitCommands(command: string): string[] {
...
 
   return undefined;
 }
 
+/**
+ * Splits a shell command into a list of individual commands using Tree-sitter.
+ * Falls back to the legacy parser if Tree-sitter cannot provide a clean parse.
+ */
+export function splitCommands(command: string): string[] {
+  const powerShellParsed = parsePowerShellCommands(command);
```
jacob314: should we be parsing power shell commands when not on windows?

reconsumeralization:
The code must implement proper shell command parsing using established grammar rather than regex patterns. A robust approach would parse commands into abstract syntax trees, then validate each component including the base command, arguments, redirections, and any chaining operators. Risk scoring should consider multiple factors - not just command names but specific argument patterns, since rm file.txt presents vastly different risk than rm -rf /. The system could assign risk levels from SAFE to CRITICAL and make approval decisions based on configurable thresholds.
Explicit argument validation would allow fine-grained policies like permitting git commit but blocking git commit --amend to prevent history rewriting. Temporal restrictions could require reapproval after a time period or limit maximum auto-approvals per session to contain potential compromise. The configuration syntax might look like:
```
json{
  "shell": {
    "neverPrompt": ["ls", "pwd", "git status"],
    "alwaysPrompt": ["rm", "git reset", "sudo *"],
    "defaultBehavior": "prompt",
    "requireReapprovalAfter": "24h"
  }
}
```
Testing requirements should include comprehensive coverage of command pattern matching, fuzz testing with known injection techniques, integration tests with sandbox modes, and regression tests for the v0.1.14 vulnerability patterns. A security review checklist should verify that command parsing uses proper grammar, the default safelist received security team review, audit logging captures all executions, and documentation clearly explains the security model and best practices.
The architecture should implement layered security rather than relying solely on policy configuration. Layer one uses the configuration-based policy system from this PR. Layer two adds runtime command analysis for anomaly detection. Layer three maintains sandbox enforcement as currently exists. Layer four implements comprehensive audit logging. This defense-in-depth approach ensures that if one layer fails, others provide backup protection.
Measuring success and path forward
This PR represents sophisticated thinking about security design - recognizing that security mechanisms must account for human behavior, not just technical controls. The insight that excessive prompting reduces security rather than enhancing it challenges conventional thinking about permission systems. [github](https://github.com/google-gemini/gemini-cli/issues/5256) If implemented with robust command parsing and careful defaults, this change could significantly improve Gemini CLI's practical security posture while making it more pleasant to use.
The critical success factors center on implementation details rather than the conceptual approach. The command parsing must be injection-resistant, thoroughly tested against real attack patterns from security research. The default safelist needs conservative curation - when in doubt, require approval. Documentation must clearly explain the security model so users can make informed decisions about their policies. A phased rollout with feature flags would enable gathering telemetry on usage patterns and catching issues before they affect all users.
The broader implication extends beyond Gemini CLI to AI agent security generally. As AI systems gain more autonomy and tool access, the permission model becomes the critical control point. This PR demonstrates that borrowed patterns from traditional software (either no permissions or constant prompts) don't translate well to AI agents that might execute dozens of operations per task. The community needs new security primitives designed specifically for AI agent interaction patterns - policy-based, risk-appropriate, and resistant to alert fatigue. PR #10498 represents meaningful progress toward those primitives, assuming the implementation matches the ambition of its design.

PR #10502, opened 2025-10-03
Enhance debug profiler to track tree framerate and dispatch errors

### LINKED ISSUE #10257: Automated test that Ink does not render frames continuously
We have had multiple hard to detect regressions where bugs in packages/cli react code result in Ink continuously rendering frames.

We need unit and integration tests that verify that Ink does not continuously render new frames for critical user journeys such as confirming an edit, and editing the prompt field.

jacob314:
## TLDR

Detect when we render frames and surface warnings if we are continously rendering frames when the app should be idle.

Only enabled when using `npm run start` so no production impact.
In those development builds the `/profile` tool is available and toggles this mode.

## Dive Deeper

We had a fairly accurate DebugProfiler that could be enabled with `ctrl-b` if you ran with DEBUG=1 but few developers did resulting in multiple bugs it could have caught making it to production.

We improve the robustness of frame detection by watching stdout to detect when Ink really is rendering frames rather than relying on a particular component being recomputed to indicate that a frame was rendered. Without this check we would miss frames due to proper use of React memoization. For example, renders of the input prompt were not captured as renders by the previous profiler.

Video of the alerts in action on an intentionally broken build:
[Demo of debug tracking.webm](https://github.com/user-attachments/assets/f66121e3-bf9b-45c0-915f-d950d4a1f1b7)
Screenshot 2025-10-03 at 11 52 51 AM

UI when running a dev build after pressing ctrl+shift+P

**Open concern** resizing the app heavily appears to be trigger a flurry of frames not-associated with keyboard input. May need to tweak the thresholds a bit to reduce false positives.

## Reviewer Test Plan

Run Gemini CLI with npm run start. Verify that the profiler is available (ctrl+shift+p) 
Intentionally introduce a bug in Gemini CLI that renders constantly
Verify that the debug console gets opened and shows errors about the frames.

Verify that ctrl+shift+p is not available when running Gemini CLI via a mechanism other than npm run start.

## Linked issues / bugs

Resolves: #10257

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "sandboxImageUri": "us-docker.pkg.dev/gemini-code-dev/gemini-cli/sandbox:0.9.0-nightly.20251001.163dba7e"
   },
   "scripts": {
-    "start": "cross-env node scripts/start.js",
+    "start": "cross-env NODE_ENV=development node scripts/start.js",
     "start:a2a-server": "CODER_AGENT_PORT=41242 npm run start --workspace @google/gemini-cli-a2a-server",
     "debug": "cross-env DEBUG=1 node --inspect-brk scripts/start.js",
     "auth:npm": "npx google-artifactregistry-auth",
--- a/packages/cli/src/ui/components/CliSpinner.tsx
+++ b/packages/cli/src/ui/components/CliSpinner.tsx
@@ -1,0 +1,24 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import Spinner from 'ink-spinner';
+import { type ComponentProps, useEffect } from 'react';
+
+// A top-level field to track the total number of active spinners.
+export let debugNumSpinners = 0;
+
+export type SpinnerProps = ComponentProps<typeof Spinner>;
+
+export const CliSpinner = (props: SpinnerProps) => {
+  useEffect(() => {
+    debugNumSpinners++;
+    return () => {
+      debugNumSpinners--;
+    };
+  }, []);
+
+  return <Spinner {...props} />;
+};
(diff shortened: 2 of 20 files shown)


packages/cli/src/ui/components/CliSpinner.tsx
```diff
@@ -0,0 +1,32 @@
...
+    debugLastSpinnerActivity = Date.now();
+    return () => {
+      debugNumSpinners--;
+    };
+  }, []);
+
+  // Also update on every render, not just mount/unmount
+  if (debugNumSpinners > 0) {
+    debugLastSpinnerActivity = Date.now();
+  }
```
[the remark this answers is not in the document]
jacob314: Addressed

jacob314:
Turns out we can't detect ctrl+shift+P in VSCode so switched to a `/profile` tool only available when running with `npm run start`

Added a test to verify that `/profile` is only availabe in that mode.
Screenshot 2025-10-03 at 8 18 18 PM

package.json
```diff
@@ -17,7 +17,7 @@
     "sandboxImageUri": "us-docker.pkg.dev/gemini-code-dev/gemini-cli/sandbox:0.9.0-nightly.20251001.163dba7e"
   },
   "scripts": {
-    "start": "cross-env node scripts/start.js",
+    "start": "cross-env NODE_ENV=development node scripts/start.js",
```
galdawave, line 20: Nit: Do we also want to add a run command for production environment, so if something doesn't repro in dev but does in prod

jacob314: Fair question but I would err on the side of not adding it unless we have issues. isDevelopment checks should be way lighter weight than what we would put on DEBUG=true so should very rarely be the cause of bugs. If they are, we are using isDevelopment wrong.
If you need to repro something quick locally you should run
`node /Users/jacobr/git/your_gemini_cli_dir/packages/cli/dist/index.js`
rather than using start as start.js has a bunch of other local start only logic.

If you need to isolate out whether NODE_ENV=development was the problem you can set that env variable before running that entry point to simulate the development env without hte other logic from start.js.

mattKorwel (approved):

PR #10508, opened 2025-10-03
Get around the initial empty response from gemini-2.5-flash.

mrcabbage972:
## TLDR

This PR fixes an issue where the model ignores the user's first command after setup.

It addresses the issue by creating a "Handshake" Prompt: It changes the initial chat message to be a single, large prompt that includes all the context and explicitly requires the model to acknowledge it before the user sends their first real command.

It also includes a minor bug fix for how errors are reported.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


SandyTao520 (approved):

packages/core/src/core/client.ts
```diff
@@ -225,23 +225,37 @@ export class GeminiClient {
...
+    // 2. Convert the array of parts into a single string
+    const envContextString = envParts
+      .map((part) => part.text || '')
+      .join('\n\n');
+
+    // 3. Combine the dynamic context with the static handshake instruction
+    const allSetupText = `
+${envContextString}
+
+Reminder: Do not return an empty response when a tool call is required.
```
SandyTao520, line 245: is this needed?

mrcabbage972: Yeah, it helped.

anowardear062-svg (approved):

PR #10514, opened 2025-10-03
Cleanup extension update logic

jakemac53:
## TLDR

Avoids repeatedly checking for updates when we have already done so and leans into the reducer approach more heavily.

## Dive deeper

Various cleanups to rely more heavily on the reducer, and handle all updates from within `useExtensionUpdates` - the slash command also just dispatches an event to the reducer now.

Also cleaned up the `checkForExtensionUpdate` function so it just returns an extension state instead of internally calling a function to update the state, this makes it more testable/portable outside of react.

`checkForAllExtensionUpdates` now also takes a current working directory, which fixes a potential bug in at least one place where we should have been flowing through the working dir from the context instead of using `process.cwd()`.

This also paves the way for a future `/extensions check` command which can similarly just fire an event on a reducer.

We also have a centralized place handling all the update checks as well as actual updates, and this should help with any race conditions that may have existed before.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

### DIFF
--- a/packages/cli/src/commands/extensions/update.ts
+++ b/packages/cli/src/commands/extensions/update.ts
@@ -95,7 +95,6 @@ export async function handleUpdate(args: UpdateArgs) {
           if (action.type === 'SET_STATE') {
             extensionState.set(action.payload.name, {
               status: action.payload.state,
-              notified: true, // No need to process as we will force the update.
             });
           }
         },
--- a/packages/cli/src/commands/extensions/update.ts
+++ b/packages/cli/src/commands/extensions/update.ts
@@ -53,16 +53,13 @@ export async function handleUpdate(args: UpdateArgs) {
         console.log(`Extension "${args.name}" not found.`);
         return;
       }
-      let updateState: ExtensionUpdateState | undefined;
       if (!extension.installMetadata) {
         console.log(
           `Unable to install extension "${args.name}" due to missing install metadata`,
         );
         return;
       }
-      await checkForExtensionUpdate(extension, (newState) => {
-        updateState = newState;
-      });
+      const updateState = await checkForExtensionUpdate(extension);
       if (updateState !== ExtensionUpdateState.UPDATE_AVAILABLE) {
         console.log(`Extension "${args.name}" is already up to date.`);
         return;
@@ -92,14 +89,18 @@ export async function handleUpdate(args: UpdateArgs) {
   if (args.all) {
     try {
       const extensionState = new Map();
-      await checkForAllExtensionUpdates(extensions, (action) => {
-        if (action.type === 'SET_STATE') {
-          extensionState.set(action.payload.name, {
-            status: action.payload.state,
-            processed: true, // No need to process as we will force the update.
-          });
-        }
-      });
+      await checkForAllExtensionUpdates(
+        extensions,
+        (action) => {
+          if (action.type === 'SET_STATE') {
+            extensionState.set(action.payload.name, {
+              status: action.payload.state,
+              processed: true, // No need to process as we will force the update.
+            });
+          }
+        },
+        process.cwd(),
+      );
       let updateInfos = await updateAllUpdatableExtensions(
         workingDir,
         requestConsentNonInteractive,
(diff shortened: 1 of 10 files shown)


packages/cli/src/ui/hooks/useExtensionUpdates.ts
```diff
@@ -80,40 +82,77 @@ export const useExtensionUpdates = (
...
+    });
+    if (extensionsToCheck.length === 0) return;
+    checkForAllExtensionUpdates(
+      extensionsToCheck,
+      dispatchExtensionStateUpdate,
+      cwd,
+    );
+  }, [
+    extensions,
+    extensionsUpdateState.extensionStatuses,
```
jakemac53, line 101: Note that this is a new dependency - but it doesn't create a render loop even though we also call `dispatchExtensionStateUdpate` - because these states are linear.

This ensures we don't keep checking for updates to extensions that we have already checked or are currently checking.

packages/cli/src/ui/state/extensions.ts
```diff
@@ -83,6 +95,27 @@ export function extensionUpdatesReducer(
...
+          names: [
+            ...(state.scheduledUpdate?.names ?? []),
+            ...(action.payload.names ?? []),
+          ],
+          onComplete: (updateInfos) => {
+            state?.scheduledUpdate?.onComplete(updateInfos);
+            action.payload.onComplete(updateInfos);
+          },
+        },
+      };
```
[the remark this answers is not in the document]
jakemac53: These requests are very intentionally merged.

packages/cli/src/commands/extensions/update.ts
```diff
@@ -92,14 +89,18 @@ export async function handleUpdate(args: UpdateArgs) {
...
-          });
-        }
-      });
+      await checkForAllExtensionUpdates(
+        extensions,
+        (action) => {
+          if (action.type === 'SET_STATE') {
+            extensionState.set(action.payload.name, {
+              status: action.payload.state,
+              notified: true, // No need to process as we will force the update.
```
jacob314: update the comment to align with the code.
Optional nit: this is now a lie as we haven't notified.
to be overly pedantic we could label this
notificationRequested or something similar. Then it can be false here to exactly indicate the state.

packages/cli/src/ui/state/extensions.ts
```diff
@@ -83,6 +95,27 @@ export function extensionUpdatesReducer(
...
+        ...state,
+        // If there is a pre-existing scheduled update, we merge them.
+        scheduledUpdate: {
+          all: state.scheduledUpdate?.all || action.payload.all,
+          names: [
+            ...(state.scheduledUpdate?.names ?? []),
+            ...(action.payload.names ?? []),
+          ],
+          onComplete: (updateInfos) => {
+            state.scheduledUpdate?.onComplete(updateInfos);
```
jacob314: this is a bit complex. instead of chaining this together can you make it explicit that you have a list of onComplete methods that each need to be called? you can then do a for loop calling all of them when you get outside the reducer keeping the callstacks from getting oddly chained.
also add try catch blocks so if one onComplete throws the others still occur.

jacob314 (commented):
Congrats on your first big reducer change! Approved once these comments are addressed.

packages/cli/src/commands/extensions/update.ts
jakemac53: I could also just remove this entirely, its not necessary in this code path, and this code path does not do any notifications of available updates, it always installs them.

jakemac53: I went ahead and just made that change, to just not indicate we notified. This code path is for the regular update command outside the CLI itself.

packages/cli/src/ui/state/extensions.ts
jakemac53: done

jacob314 (approved):
lgtm

PR #10521, opened 2025-10-04
feat(cli): Add direct shell command execution with Ctrl/Meta+Enter

### LINKED ISSUE #10130: Add a Direct Local Shell Execution Mode for Custom Commands

**1. Problem Description**

The current custom command system in `gemini-cli`, which uses the `prompt` field to interact with an LLM, is incredibly powerful for handling complex, intelligent tasks that require understanding and reasoning.

However, for simple, high-frequency shell commands (e.g., `git status`, `ls -l`, `docker ps`), the full round-trip (`User -> CLI -> LLM -> Tool Call -> Response`) introduces unnecessary network latency. This sacrifices the immediacy expected from a command-line tool and forces users to break their workflow by switching to a native terminal for quick, simple tasks.

**2. Proposed Solution**

We propose introducing a new field in the `.toml` configuration file for custom commands, for example, `shell`.

The `gemini-cli` tool would then parse commands with the following logic:

*   **If a `shell` field is detected:**
    *   The tool **directly executes the command string in the local shell environment**.
    *   It should stream the standard output (stdout) and standard error (stderr) directly and instantly to the user.
    *   This process **bypasses the LLM entirely**, resulting in zero-latency execution.

*   **If a `prompt` field is detected:**
    *   The tool maintains the current logic, sending the task to the LLM for intelligent processing.

The `shell` and `prompt` fields should be mutually exclusive within a single `.toml` file to define a clear execution mode for each command.

**3. Use Case Comparison**

Let's consider creating an alias `gs` for `git status`.

**A. Current Mode (LLM-based, Slower)**

```toml
# gs-llm.toml
name = "gs-llm"
description = "Run git status via LLM"
prompt = "Please run the 'git status' command for me"
```
*   **Execution Flow:** `gemini gs-llm` -> Network Request -> LLM Processing -> Tool Call -> Result.
*   **Experience:** Noticeable latency.

**B. Proposed Mode (Local Direct Execution, Instant)**

```toml
# gs.toml
name = "gs"
description = "Run git status directly in the local shell"
shell = "git status"
```
*   **Execution Flow:** `gemini gs` -> `gemini-cli` executes `git status` locally -> Instant Result.
*   **Experience:** Identical to running `git status` in a native terminal; zero latency.

**4. Core Advantages**

*   **Ultimate Efficiency:** For simple commands, the response time is identical to a native terminal, completely eliminating network latency.
*   **Seamless Experience:** Users can efficiently execute any command without leaving the `gemini-cli` window, preventing context switching and preserving their workflow (flow state).
*   **The Best of Both Worlds:** Perfectly combines the "speed" of native commands with the "intelligence" of LLM-driven tasks, allowing users to choose the most appropriate mode for their needs.
*   **Clarity and Explicitness:** The presence of either a `shell` or `prompt` field in the `.toml` file makes the command's execution mode unambiguous.
*   **Backward Compatibility:** This new field would not interfere with existing `prompt`-based commands, allowing for a smooth and non-disruptive implementation.

**Conclusion**

We believe that adding a direct `shell` execution mode will significantly enhance the appeal of `gemini-cli` for power users, especially developers, transforming it into a next-generation command-line assistant that excels in both "efficiency" and "intelligence."

We hope the development team will consider this suggestion. Thank you!

### Why is this needed?

 1. Speed: Eliminate Unnecessary Delays

   * Current Situation: Right now, every command you run—even something as simple as ls—has to travel over the internet to me (the AI), be processed, 
     and then have the instruction sent back to your machine to be executed. This round-trip creates a noticeable lag.
   * The Problem: For simple, everyday commands, this lag makes the tool feel slow and inefficient compared to a standard terminal.
   * The Fix: The proposed shell mode creates a "local express lane." It tells the gemini-cli tool on your computer: "For this command, don't talk to 
     the AI. Just run it yourself, instantly."

  2. Workflow: Stay in a Single Window

   * Current Situation: Because of the speed issue, you're often forced to make a choice: "Should I use the Gemini window for this, or is it faster to 
     open a new, standard terminal tab?"
   * The Problem: Constantly switching between your AI assistant and a regular terminal is distracting. It breaks your concentration and fragments 
     your work across multiple windows.
   * The Fix: By making simple commands instant, this feature allows you to stay within the gemini-cli environment for all your tasks. You don't have 
     to switch contexts.

  In short, this feature is needed to combine the raw speed of a native terminal for simple tasks with the powerful intelligence of an AI for 
  complex ones, all within a single, unified, and efficient interface.

### Additional context

_No response_

var-aha:
## TLDR

Adds keyboard shortcuts (`Meta+Enter` on macOS, `Ctrl+Enter` on all platforms) to execute shell commands directly from the input prompt without entering shell mode.

## Dive Deeper

Previously, users who wanted to run a shell command had to either:

1. Toggle into shell mode with `!`, execute the command, then toggle back
2. Use natural language to describe what command they want to run, which unnecessarily invokes the LLM (inefficient and slower)

This change adds a third option: pressing `Meta+Enter` (macOS) or `Ctrl+Enter` (all platforms) executes the current input text as a shell command without switching modes.

**Implementation details:**

- Added `EXECUTE_PROMPT_COMMAND` to the command enum in key bindings
- Configured key bindings for `Meta+Enter` (macOS) and `Ctrl+Enter` (all platforms)
- Reuses existing `handleShellCommand` infrastructure from shell mode
- Empty buffer handling prevents newline insertion
- Commands support proper abort handling via AbortController
- Updated help text to document the new shortcuts

## Reviewer Test Plan

1. Run the CLI with the changes
2. Type a shell command in the prompt (e.g., `ls -la`)
3. Press `Meta+Enter` (macOS) or `Ctrl+Enter` (other platforms)
   - **Note for macOS testers:** You can use `Ctrl+Enter` on macOS as well. If that doesn't work, try `Option+Enter` (since some terminals report Option as Meta instead of Cmd)
4. Verify the command executes without entering shell mode
5. Test with empty buffer to ensure no-op behavior
6. Test command abortion with `Ctrl+C`

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ✅  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes #10130


var-aha:
Hey @gemini-cli-robot could you please have a look at this PR.

var-aha:
Hey @SandyTao520 , @jacob314 , @scidomino , @bdmorgan , just checking in to see if this PR could get a review when you have a moment. It’s been pending for a couple of weeks. Thanks!

PR #10522, opened 2025-10-04
Re-land bbiggs changes to reduce margin on narrow screens with fixes + full width setting

jacob314:
## TLDR

To fix the edge cases Ink had trouble with in the initial version of this pr by @bbiggs it was easier to undo all the remaining cases we had percent rather than full width calculations in the UI rather than debug why Ink % based widths went off the rails some of the time.

To verify this was correct it was convenient to add a simple setting to render the UI at 100%. I've left that setting in this pr but off by default setting so we can get user feedback before making it the default.

Demo of the full width mode showing that Ink is robust rendering that mode in Ghostty. iterm2 can be made as robust once we 
[No corruption demo.webm](https://github.com/user-attachments/assets/5ea72712-f94f-436f-8431-05355a08096b)

I also fiddled with widths a bit fixing bugs resulting in a couple chars of extra padding on the status bar.
Screenshot 2025-10-03 at 7 32 40 PM

## Dive Deeper

Ink % based widths should be avoided where possible in Gemini CLI at the moment for a range of reasons. I expect the root of the issue was ink didn't have a fully bounded default width so got in trouble after resizes. Instead we calculate all widths as fixed values which is required for the MaxSizedBox text wrapping library to work anyway. We'll get rid of MaxSizedBox but until we do having % is just a source of bugs.

Items done:
* Remove all uses of percentage based layout forcing all layouts to be use fixed widths
*Remove hard coded 8 char padding as it was a hack to workaround issues that no longer exist.
* Added force line wrap escape code on CLI startup. This eliminates all Ink rendering artifacts in Ghostty and will help in other terminals once we switch to alternate buffer mode. This is needed to get early feedback on full width mode from Ghostty users.
* Cleaned up some tech debt with redundant width calculations.
*Added a useFullWidth setting that is off by default. When it is on, the full terminal width is used rather than ~90%
*This should probably be the default in the future and makes it easier to verify that we don't have any width bugs
as it is much more obvious you have width bugs when width's don't exactly match the terminal width.
*Now that we aren't drawing boxes around previous prompts from users there is less downside to rendering the full width.
* Minor changes: fixed wacky 1 char indent for confirmation prompts that served no purpose other than to make the UI look misaligned and uneven.

## Reviewer Test Plan

Test both by default and with  fixed width mode.
Test in Ghostty or similar to verify that the no-wrap behavior works.
Rapidly resize the terminal and ensure that after resize is done the render is acceptable.


bbiggs:
In your screen capture, the status bar with the context length doesn't align to the RHS. Is that intentional or something we'd fix later (also with removing the left-padding on the tool output box)?

jacob314:
Talked with @miguelsolorio about it and he wanted the status bar to stay the same width as the prompt. When we graduate to make the full width option the default we will be able to get the status bar back to full width.
The padding is removed from the tool output box. that is intentional also after discussion with UX.

packages/cli/src/config/settingsSchema.ts
```diff
@@ -361,6 +361,15 @@ const SETTINGS_SCHEMA = {
         description: 'Show citations for generated text in the chat.',
         showInDialog: true,
       },
+      useFullWidth: {
+        type: 'boolean',
+        label: 'Use Full Width',
```
jacob314, line 416: we might need a clearer label. also fine with hiding this from the settings dialog by default

packages/cli/src/ui/components/Composer.tsx
```diff
@@ -40,14 +39,8 @@ export const Composer = () => {
...
 
-  // Use the container width of InputPrompt for width of DetailedMessagesDisplay
-  const { containerWidth } = useMemo(
-    () => calculatePromptWidths(uiState.terminalWidth),
-    [uiState.terminalWidth],
-  );
-
   return (
-    <Box flexDirection="column">
+    <Box flexDirection="column" width={uiState.mainAreaWidth} flexShrink={0}>
```
jacob314, line 43: flexShrink ={0} + width is the best way to ensure you really get exactly the width specified.

packages/cli/src/ui/components/Footer.tsx
```diff
@@ -71,7 +70,7 @@ export const Footer: React.FC = () => {
   return (
     <Box
       justifyContent={justifyContent}
-      width="100%"
+      width={mainAreaWidth}
```
jacob314, line 76: @miguelsolorio requested we didn't make this 100% so the content stays aligned.

packages/cli/src/ui/components/Footer.tsx
```diff
@@ -147,24 +146,26 @@ export const Footer: React.FC = () => {
...
                 promptTokenCount={promptTokenCount}
                 model={model}
-                terminalWidth={terminalWidth}
+                terminalWidth={mainAreaWidth}
               />
             </Text>
             {showMemoryUsage && <MemoryUsageDisplay />}
           </Box>
-          <Box alignItems="center" paddingLeft={2}>
+          <Box alignItems="center">
```
jacob314, line 158: this paddingLeft={2} was causing the footer to be less wide than the InputPrompt so fixed instead adding padding only as needed to make sure things were separated by a space.

packages/cli/src/ui/components/InputPrompt.test.tsx
```diff
@@ -1495,7 +1495,7 @@ describe('InputPrompt', () => {
 
       const outputLines = frame!.split('\n');
       // The number of lines should be 2 for the border plus 3 for the content.
-      expect(outputLines.length).toBe(5);
+      expect(outputLines.length).toBe(6);
```
jacob314: as far as I can tell just Ink being weird now that there is a fixed width. reality is there should have always shown up as a line but Ink wasn't including one in the output. *shrug*

packages/cli/src/ui/components/InputPrompt.tsx
```diff
@@ -73,25 +74,16 @@ export interface InputPromptProps {
 }
 
 // The input content, input container, and input suggestions list may have different widths
-export const calculatePromptWidths = (terminalWidth: number) => {
-  const widthFraction = 0.9;
+export const calculatePromptWidths = (mainContentWidth: number) => {
```
jacob314, line 77: this calculation was overly complex. the input prompt was just supposed to be the main area width.

packages/cli/src/ui/components/messages/ToolGroupMessage.tsx
```diff
@@ -96,8 +96,7 @@ export const ToolGroupMessage: React.FC<ToolGroupMessageProps> = ({
         Ink to render the border of the box incorrectly and span multiple lines and even
         cause tearing.
       */
-      width="100%"
-      marginLeft={1}
```
jacob314, line 100: intentionally removed after discussion with UX that they also didn't like that these boxes were indented 1.

packages/cli/src/ui/hooks/useTerminalSize.ts
```diff
@@ -6,18 +6,16 @@
 
 import { useEffect, useState } from 'react';
 
-const TERMINAL_PADDING_X = 8;
```
jacob314, line 9: reverted this padding as it wasn't needed. it was from when we didn't refesh on resize so a single Ink rendering artifact would permanently corrupt. now that isn't the case.

miguelsolorio:
@jacob314 were we going to make things aligned with the input?

image

jacob314:
Tweaked the UX a little based on @miguelsolorio's feedback
Screenshot 2025-10-07 at 4 37 06 PM

miguelsolorio (approved):

packages/cli/src/ui/AppContainer.tsx
```diff
@@ -263,13 +264,14 @@ export const AppContainer = (props: AppContainerProps) => {
     registerCleanup(consolePatcher.cleanup);
   }, [handleNewMessage, config]);
 
+  const mainAreaWidth = calculateMainAreaWidth(terminalWidth, settings);
```
galdawave, line 267: Something in this change is causing wrong height calculations for Shell Tool , where the shell tool is exceeding the terminal height causing flickering 

Image

This is what it looks like in main 

Image

This only happens when use full width is false

packages/cli/src/ui/components/__snapshots__/InputPrompt.test.tsx.snap
```diff
@@ -1,57 +1,57 @@
...
  lllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllll
  ..."
 `;
 
 exports[`InputPrompt > command search (Ctrl+R when not in shell) > expands and collapses long suggestion via Right/Left arrows > command-search-expanded-match 1`] = `
-"╭──────────────────────────────────────────────────────────────────────────────────────────────────╮
-│ (r:)    Type your message or @path/to/file                                                       │
-╰──────────────────────────────────────────────────────────────────────────────────────────────────╯
+"╭────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
+│ (r:)    Type your message or @path/to/file                                                        │
```
galdawave, line 14: i could be wrong but isn't this causing a regression ?

jacob314: existing snaps were a bit misleading. previously the widths logic was a convoluted mix of %. now it is fixed so these snaps actually match the real state with a particular width. before they were inaccurate as we didn't match the full DOM.

packages/cli/src/ui/AppContainer.tsx
jacob314: Fixed... mostly. there is still an occasional width misalignment but it doesn't break anything you just see ...
There were existing issues like this without the change. you could trigger by gemini CLI inside gemini cli with rapid width changes. Problem was we weren't forcing the terminal lines to truncate. Judgment call whether we want to show the `...` on the truncated lines. They are good to warn something is a bit off and not too harmful.
Can likely improve in the future with finer grained with calculations but there will be some inherent weird mismatches.
[cast.webm](https://github.com/user-attachments/assets/062b3f2c-8222-4359-ba65-4ce53f409864)

jacob314:
I forced pushed but if you want to view the changes since you approved check out https://github.com/google-gemini/gemini-cli/pull/10522/commits/b6d2778629af6005e5f0ab4d13678585d172ccd9

which shows just the changes needed to make the embedded shell output more robust and not trigger flicker when xterm.js wants to render output that Ink measures as longer than the space available.

galdawave (approved):

PR #10561, opened 2025-10-05
fix: make non-interactive shell rendering consistent with interactive

nehaaprasad:
### TLDR

- Fixed inconsistent shell rendering between interactive and non-interactive modes by replacing basic `stripAnsi()` with proper terminal emulation in child process fallback.

### Dive Deeper

- Non-interactive shells used simple `stripAnsi()` while interactive shells used terminal emulation, causing flicker and raw escape codes in output. Now both modes use the same `@xterm/headless` approach for consistent rendering.

### Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ✅  | ✅  |
| npx      | ✅  | ✅  | ✅  |
| Docker   | ✅  | ✅  | ✅  |
| Podman   | ✅  | -   | -   |
| Seatbelt | ✅  | -   | -   |

### Linked issues / bugs

Fixes #10494


owenofbrien (commented):
@naaa760 - I thought the idea was to unify around a single shared implementation, but it looks like you added a bunch of custom logic

nehaaprasad:
@owenofbrien 
update it, let me know and sorry for being a little late to fix.

PR #10569, opened 2025-10-05
feat: Blend educative tips with witty phrases during loading times (fun, subtle learning...)

JayadityaGit:
## TL;DR

This PR replaces the existing loading-screen witty messages with an updated set that is both **clever and genuinely informative**. The new messages surface useful tips about commands, keyboard shortcuts, and settings—while noting whether a setting is configurable via the UI or only through `settings.json`.

---

##  Purpose

The goal is to make the CLI experience more engaging and helpful during load times. Instead of filler text, users now see short, fun messages that teach them something useful about the tool.

These improvements aim to:

* Encourage feature discovery
* Improve usability
* Maintain a playful tone without sacrificing clarity

---

##  What's Included

The updated messages cover:

* Commands
* Keyboard shortcuts
* Settings (visible and hidden)

For clarity:

* Settings only configurable via `settings.json` now include a **`(settings.json)`** suffix
* Settings available through the UI include a **`(/settings)`** suffix

## ✅ Testing Matrix

```
┌──────────┬────┬────┬────┐
│          │ 🍏 │ 🪟 │ 🐧 │
├──────────┼────┼────┼────┤
│ npm run  │ ❓ │ ❓ │ ✅ │
│ npx      │ ❓ │ ❓ │ ❓ │
│ Docker   │ ❓ │ ❓ │ ❓ │
│ Podman   │ -  │ -  │ -  │
│ Seatbelt │ ❓ │ -  │ -  │
└──────────┴────┴────┴────┘
```

---

## 🔗 Linked Issue

Resolves **#10568**


packages/cli/src/ui/hooks/usePhraseCycler.ts
```diff
@@ -7,135 +7,132 @@
...
+  'Give credit where credit is due. `showCitations`',
+  'No more witty phrases. `disableLoadingPhrases`',
+  'For your listening pleasure. `screenReader`',
+  'Connect to your IDE. `ide.enabled`',
+  'Help us help you. `usageStatisticsEnabled`',
+  'How long can you go? `maxSessionTurns`',
+  'Let the AI talk to itself. `skipNextSpeakerCheck`',
+  'How deep will you go? `discoveryMaxDirs`',
+  'Load more memories. `loadMemoryFromIncludeDirectories`',
+  'Ignore your gitignore. Or not. `respectGitIgnore`',
```
psinha40898: Did we forget the "(settings)" suffix for many UI configurable settings or have I misunderstood intent?

packages/cli/src/ui/hooks/usePhraseCycler.ts
```diff
@@ -7,135 +7,132 @@
...
+  'Get interactive. `enableInteractiveShell`',
+  'See the world in color. `showColor`',
+  'Live life on the edge. `autoAccept`',
+  'Ripgrep your way to victory. `useRipgrep`',
+  'Keep it short. `enableToolOutputTruncation`',
+  'How short is short? `truncateToolOutputThreshold`',
+  'How many lines is short? `truncateToolOutputLines`',
+  'Let the tools talk to each other. `enableMessageBusIntegration`',
+  'Trust no one. Or do. `folderTrust.enabled`',
+  'Let the model choose. `useModelRouter` (settings)',
```
psinha40898: Maybe `/settings` instead of `(settings)`?

packages/cli/src/ui/hooks/usePhraseCycler.ts
JayadityaGit: yeah i missed it, thanks for pointing out !

packages/cli/src/ui/hooks/usePhraseCycler.ts
JayadityaGit: thanks will use the slash command, i think that's better.

packages/cli/src/ui/hooks/usePhraseCycler.ts
```diff
@@ -7,135 +7,132 @@
...
-  'Just waiting for the dial-up tone to finish...',
-  'Recalibrating the humor-o-meter.',
-  'My other loading screen is even funnier.',
-  "Pretty sure there's a cat walking on the keyboard somewhere...",
-  'Enhancing... Enhancing... Still loading.',
-  "It's not a bug, it's a feature... of this loading screen.",
-  'Have you tried turning it off and on again? (The loading screen, not me.)',
-  'Constructing additional pylons...',
-  'New line? That’s Ctrl+J.',
+  'Squash bugs with `/bug`!',
```
jacob314: nit: Squash --> Report

packages/cli/src/ui/hooks/usePhraseCycler.ts
```diff
@@ -7,135 +7,132 @@
...
+  'Save your brilliant ideas with `/chat save`.',
```
jacob314: to keep the style consistent I would end all of these with `...`

jacob314:
I'd suggest we enable this as a setting rather than the default. In addition, would be ideal if we interleaved the useful with funny settings when the option is enabled. That would align with how we already show a tip about ctrl-J

JayadityaGit:
Hey @jacob314,

I believe that many users — myself included — may not be fully aware of all the useful features available(I really can't decide which are important). Of course, the importance of certain features can vary from person to person, but sometimes users might simply miss out on functionalities that could actually enhance their UNIQUE workflow. That’s the main motivation behind this PR.

Since we already have an option to disable loading phrases, I don’t see a strong reason to introduce an additional setting just for this. If a user prefers fully personalized witty phrases, they can always opt for the custom witty feature.

What I’ve done here is extract about 50% of the original witty phrases and blend them with a new set that’s both educative and witty. I feel this strikes a nice balance — keeping the humor while also subtly teaching or sharing something useful in a lighthearted way.

packages/cli/src/ui/hooks/usePhraseCycler.ts
JayadityaGit: Sure !

packages/cli/src/ui/hooks/usePhraseCycler.ts
JayadityaGit: Thanks for the suggestion !

JayadityaGit:
> I'd suggest we enable this as a setting rather than the default. In addition, would be ideal if we interleaved the useful with funny settings when the option is enabled. That would align with how we already show a tip about ctrl-J

I personally feel that if this is placed within the settings, there’s a chance it could be missed by users, which might lessen the impact of the intended change.

packages/cli/src/ui/hooks/usePhraseCycler.ts
```diff
@@ -7,8 +7,135 @@
 import { useState, useEffect, useRef } from 'react';
 
 export const WITTY_LOADING_PHRASES = [
-  "I'm Feeling Lucky",
-  'Shipping awesomeness... ',
+  'Report bugs with `/bug`!...',
+  'Save your brilliant ideas with `/chat save`....',
+  'A fresh slate is just a `/clear` away....',
+  'Shrink your conversation with `/compress`....',
```
jacob314: fix cases with four dots and cleanup cases that now have ! and ...

packages/cli/src/ui/hooks/usePhraseCycler.ts
```diff
@@ -7,8 +7,135 @@
...
+  'Your secrets are safe with us. `/privacy`...',
+  "Don't let the door hit you on the way out. `/quit`...",
+  'Enter the Vim vortex with `/vim`....',
+  'Start your next masterpiece with `/init`....',
+  'Build your own commands. You have the power!...',
+  'Shell out with `!`....',
+  'Feed the beast with `@`....',
+  'Play in the sandbox with `-s`....',
+  'Go incognito with `--headless`....',
+  'Go back in time with checkpoints....',
```
jacob314: how?

jacob314:
Even if we eventually make this the default we will need to have it be a setting so users who already feel they are familiar with the CLI can go back to the witty phrases.

If you make the logic slightly more complex so you get a maybe 5-1 ratio of funny phrases to tips we could roll it out as the default.
You could keep the lists separate and pick a random # each turn with a 5 out of 6 chance you pick a whitty phrase and a 1/6 chance you pick from the list of tips. If that works well we could increase the ratio over time. Would want to slowly increase it to make sure people aren't frustrated the whitty phrases got a lot less whitty.
Other option is you could start by picking the 10 best tips from your list of tips and add those to the normal list.

JayadityaGit:
> Even if we eventually make this the default we will need to have it be a setting so users who already feel they are familiar with the CLI can go back to the witty phrases.
> 
> If you make the logic slightly more complex so you get a maybe 5-1 ratio of funny phrases to tips we could roll it out as the default. You could keep the lists separate and pick a random # each turn with a 5 out of 6 chance you pick a whitty phrase and a 1/6 chance you pick from the list of tips. If that works well we could increase the ratio over time. Would want to slowly increase it to make sure people aren't frustrated the whitty phrases got a lot less whitty. Other option is you could start by picking the 10 best tips from your list of tips and add those to the normal list.

@jacob314 absolutely brilliant suggestion — thank you so much for the clear and thoughtful guidance!

packages/cli/src/ui/hooks/usePhraseCycler.ts
JayadityaGit: ops — thank you so much for clarifying, @jacob314!

JayadityaGit:
The npm preflight checks are all passing. I’ve personally tested it, and the 1–6 probability for the educative tips is working as expected. Since the probability is quite low and I wanted the tips to stand out, I chose not to make them witty.

jacob314 (approved):
Love it! We may need to tweak some of the informative phrases to be a little shorter. I played with it for a bit and had a bit of trouble reading some of them before the phrase changed... but I was entering fairly trivial prompts so users may be less impacted.
lgtm

packages/cli/src/ui/hooks/usePhraseCycler.ts
```diff
@@ -139,6 +139,151 @@ export const WITTY_LOADING_PHRASES = [
...
+  'Toggle Vim mode for a modal editing experience (/settings)...',
+  'Disable automatic updates if you prefer manual control (settings.json)...',
+  'Turn off nagging update notifications (settings.json)...',
+  'Enable checkpointing to recover your session after a crash (settings.json)...',
+  'Change CLI output format to JSON for scripting (settings.json)...',
+  'Personalize your CLI with a new color theme (/settings)...',
+  'Create and use your own custom themes (settings.json)...',
+  'Hide window title for a more minimal UI (/settings)...',
+  "Don't like these tips? You can hide them (/settings)...",
+  'Hide the startup banner for a faster launch (/settings)...',
```
jacob314: nit: not really faster. more just cleaner.

packages/cli/src/ui/hooks/usePhraseCycler.ts
```diff
@@ -139,6 +139,151 @@ export const WITTY_LOADING_PHRASES = [
...
+  'Reclaim vertical space by hiding the footer (/settings)...',
+  'Show memory usage for performance monitoring (/settings)...',
+  'Show line numbers in the chat for better context (/settings)...',
```
jacob314: this is on by default so remove the tip

packages/cli/src/ui/hooks/usePhraseCycler.ts
```diff
@@ -139,6 +139,151 @@ export const WITTY_LOADING_PHRASES = [
...
+  'Show citations to see where the model gets information (/settings)...',
+  'Disable loading phrases for a quieter experience (/settings)...',
+  'Add custom witty phrases to the loading screen (settings.json)...',
+  'Enable IDE integration mode for a seamless workflow (settings.json)...',
```
jacob314: have these point to /settings wherever possible rather than settings.json. IDE integration is in settings for example.

packages/cli/src/ui/hooks/usePhraseCycler.ts
```diff
@@ -139,6 +139,151 @@ export const WITTY_LOADING_PHRASES = [
...
+  'Help improve Gemini by enabling usage stats (/settings)...',
```
jacob314: remove as is the default

packages/cli/src/ui/hooks/usePhraseCycler.ts
```diff
@@ -139,6 +139,151 @@ export const WITTY_LOADING_PHRASES = [
...
+  'Choose a specific Gemini model for conversations (/settings)...',
+  'Limit the number of turns in your session history (settings.json)...',
+  'Automatically summarize large tool outputs to save tokens (settings.json)...',
+  'Control when chat history gets compressed based on token usage (settings.json)...',
+  'Define custom context file names, like CONTEXT.md (settings.json)...',
+  'Set max directories to scan for context files (settings.json)...',
+  'Expand your workspace with additional directories (settings.json)...',
```
jacob314: this should be `/directory` not settings.json

JayadityaGit:
@jacob314

Today I came across a small issue in the settings dialog — there wasn’t a clear visual cue for closing it. Initially, I was closing the CLI with ctrl+c, but later realized that esc works to close the settings dialog. I’ve added that clarification.

Note: Some tests are currently failing. The snapshot tests in SettingsDialog.test.tsx need to be updated to match the new help text. There are also some unrelated test failures in InputPrompt.test.tsx and mcp-client.test.ts.

JayadityaGit:
And since I’m participating in Hacktoberfest, if possible, could you please add the `hacktoberfest-accepted` label to this PR? It would genuinely motivate me to contribute even more to the open-source community. Thank you so much.

JayadityaGit:
@jacob314 I’ve updated the snapshot test, and the preflight checks are now passing. Thank you for adding the label — I truly appreciate it!

JayadityaGit:
@jacob314, is there anything else I should do on this?

packages/cli/src/ui/components/SettingsDialog.tsx
```diff
@@ -897,7 +897,7 @@ export function SettingsDialog({
         <Box height={1} />
         <Text color={theme.text.secondary}>
           (Use Enter to select
-          {showScopeSelection ? ', Tab to change focus' : ''})
+          {showScopeSelection ? ', Tab to change focus' : ''}, Esc to close)
```
jacob314: why is this change in this pr?

packages/cli/src/ui/components/SettingsDialog.test.tsx
```diff
@@ -1086,7 +1086,7 @@ describe('SettingsDialog', () => {
       expect(lastFrame()).toContain('Apply To'); // Scope section
       expect(lastFrame()).toContain('User Settings'); // Scope options (no numbers when settings focused)
       expect(lastFrame()).toContain(
-        '(Use Enter to select, Tab to change focus)',
+        '(Use Enter to select, Tab to change focus, Esc to close)',
```
jacob314: why is this change in this PR?

jacob314 (commented):
Seems like a settings dialog change accidentally got in this PR. please revert and I will land.

jacob314 (approved):
lgtm

PR #10607, opened 2025-10-06
feat(core): improve shell execution service reliability

### LINKED ISSUE #10492: RangeError: Invalid string length (Crash) when running commands with large stdout output
**Description:**
The `gemini` instance crashes completely when running a shell command that produces a very large amount of `stdout` or `stderr` output.

It appears `shellExecutionService.js` attempts to buffer the entire output of the executed command into a single string (`stdout += decodedChunk;`). If a script prints extensive logs or warnings (e.g., millions of lines in a loop), this buffer exceeds the Node.js maximum string length, causing the CLI to terminate.

**Environment:**
*   **OS:** Linux
*   **Node.js Version:** v22.17.1

**Steps to Reproduce:**

1.  Create a Python script (`crash_repro.py`) that generates a massive amount of stdout quickly.

    ```python
    # crash_repro.py
    # Simulates a script producing excessive logs/warnings
    import sys

    print("Starting stdout flood to test CLI buffering limits...")
    # Print enough data to exceed V8 string limits (e.g. ~500MB - 1GB)
    # Printing 5 million lines of ~100 chars.
    for i in range(5000000):
        print(f"[Warning] Generating excessive log line {i} to test stdout buffering capabilities of the parent process.")

    print("Finished flood.")
    ```

2.  Start `gemini`.

3.  Ask Gemini to run the script
    ```shell
    Run @crash_repro.py
    ```

**Actual Behavior:**
The CLI attempts to process the output and then crashes with a `RangeError`.

```text
file:///usr/local/google/home/.../@google/gemini-cli/node_modules/@google/gemini-cli-core/dist/src/services/shellExecutionService.js:109
                        stdout += decodedChunk;
                               ^

RangeError: Invalid string length
    at handleOutput (file:///usr/local/google/home/.../@google/gemini-cli/node_modules/@google/gemini-cli-core/dist/src/services/shellExecutionService.js:109:39)
    at Socket.<anonymous> (file:///usr/local/google/home/.../@google/gemini-cli/node_modules/@google/gemini-cli-core/dist/src/services/shellExecutionService.js:141:51)
    at Socket.emit (node:events:518:28)
    at addChunk (node:internal/streams/readable:561:12)
    at readableAddChunkPushByteMode (node:internal/streams/readable:512:3)
    at Readable.push (node:internal/streams/readable:392:5)
    at Pipe.onStreamRead (node:internal/stream_base_commons:189:23)
    at Pipe.callbackTrampoline (node:internal/async_hooks:130:17)

Node.js v22.17.1
```

**Expected Behavior:**
The CLI should handle large output gracefully without crashing the main process. Possible solutions:
1.  Implement a sensible buffer limit for tool outputs (truncating the output and notifying the model/user that it was truncated).
2.  Stream the output to a temporary file instead of holding it in memory, if the full output is required.

galdawave:
## TLDR

This pull request improves the reliability of the `ShellExecutionService` by implementing a graceful shutdown sequence for aborted commands, preventing the CLI from hanging on abortion, and capping memory usage to avoid crashes with high-output commands.

## Dive Deeper

This commit enhances the `ShellExecutionService` with several key improvements:

- **Graceful Shutdown:** Implements a graceful shutdown sequence on command abortion, sending `SIGTERM` followed by `SIGKILL` to allow processes to terminate cleanly.
- **Immediate Promise Resolution:** Ensures immediate promise resolution upon abortion, preventing the CLI from hanging while processing buffered output.
- **Memory Capping:** Introduces a sliding window buffer for the `child_process` fallback to cap memory usage at 16MB for `stdout` and `stderr`, preventing potential crashes with high-output commands.
- **Refined Rendering:** Refines rendering logic to ensure timely and accurate updates.

These changes make the shell execution more robust and prevent common issues related to process management and memory consumption.

## Reviewer Test Plan

1.  **Test Graceful Shutdown & No Hanging:**
    - Run a long-running command (e.g., `sleep 10`).
    - Abort the command (e.g., with Ctrl+C).
    - **Expected:** The CLI should immediately become responsive, and the `sleep` process should be terminated. You can verify with `ps aux | grep sleep`.

2.  **Test High-Output Command (Memory Capping):**
    - Run a command that produces a large amount of output (e.g., `yes | head -n 1000000`).
    - **Expected:** The CLI should handle the output without crashing or consuming excessive memory. The output in the CLI might be truncated, which is expected.

3.  **Test Aborting a High-Output Command:**
    - Run a command that continuously produces output (e.g., `yes`).
    - Abort the command.
    - **Expected:** The CLI should stop the command and return to a responsive state without any delay or hanging.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      |  ❓ | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

- Fixes #10492


packages/core/src/services/shellExecutionService.ts
```diff
@@ -21,6 +21,7 @@ import {
 const { Terminal } = pkg;
 
 const SIGKILL_TIMEOUT_MS = 200;
+const MAX_CHILD_PROCESS_BUFFER_SIZE = 16 * 1024 * 1024; // 16MB
```
jacob314, line 24: can you comment on why 16MB?

galdawave: Figured we want to limit the size of the buffer so that it doesn't crash the cli , 16mb is ~3-5 millions lines of text .

packages/core/src/services/shellExecutionService.ts
```diff
@@ -216,8 +217,18 @@ export class ShellExecutionService {
 
             if (stream === 'stdout') {
               stdout += decodedChunk;
+              if (stdout.length > MAX_CHILD_PROCESS_BUFFER_SIZE) {
```
jacob314: we should notify the user&model if there is truncation. also consider making a helper method that safely appends content to the buffer truncating as needed. This logic is fine in general but would have edge case issues if `stdout + decodedChunk` was large enough to exceed the maximum string size we can handle. It would be instead by slightly more performance and robust to first truncate stdout by the amount needed to fit within the buffer size and only then add decodedChunk. For full robustness would also need to handle if decodedChunk is larger than MAX_CHILD_PROCESS_BUFFER_SIZE

packages/core/src/services/shellExecutionService.ts
```diff
@@ -562,10 +586,16 @@ export class ShellExecutionService {
...
+                await new Promise((res) => setTimeout(res, SIGKILL_TIMEOUT_MS));
+                if (!exited) {
+                  process.kill(-ptyProcess.pid, 'SIGKILL');
+                }
               } catch (_e) {
                 // Fallback to killing just the process if the group kill fails
-                ptyProcess.kill('SIGINT');
+                if (!exited) {
+                  ptyProcess.kill('SIGKILL');
+                }
```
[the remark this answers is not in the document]
jacob314: please fix

packages/core/src/services/shellExecutionService.ts
```diff
@@ -562,10 +586,16 @@ export class ShellExecutionService {
...
+                process.kill(-ptyProcess.pid, 'SIGTERM');
+                await new Promise((res) => setTimeout(res, SIGKILL_TIMEOUT_MS));
+                if (!exited) {
+                  process.kill(-ptyProcess.pid, 'SIGKILL');
+                }
               } catch (_e) {
                 // Fallback to killing just the process if the group kill fails
-                ptyProcess.kill('SIGINT');
+                if (!exited) {
+                  ptyProcess.kill('SIGKILL');
```
jacob314, line 668: should we start by killing the process politely rather than directly killing?

packages/core/src/services/shellExecutionService.test.ts
```diff
@@ -356,6 +356,51 @@ describe('ShellExecutionService', () => {
       expect(result.aborted).toBe(true);
       // The process kill is mocked, so we just check that the flag is set.
     });
+
+    it('should send SIGTERM and then SIGKILL on abort', async () => {
+      vi.useFakeTimers();
+      const { result } = await simulateExecution(
+        'sleep 10',
```
jacob314: using a real sleep combined with fake timers seems a bit sketchy. can you comment on the intent of the sleep 10 or switch to using real timers for this test?

packages/core/src/services/shellExecutionService.test.ts
```diff
@@ -356,6 +356,51 @@ describe('ShellExecutionService', () => {
...
+          abortController.abort();
+          pty.onExit.mock.calls[0][0]({ exitCode: 1, signal: null });
+        },
+      );
+      const end = performance.now();
+
+      expect(result.aborted).toBe(true);
+      // The test should complete very quickly, without waiting for the simulated
+      // processing of 1000 data chunks.
+      expect(end - start).toBeLessThan(100);
```
jacob314: this seems like a high risk of flakes in CI. should we make the threshold significantly larger, remove this test, or figure out something else to test rather than requiring this to be a performance test.

packages/core/src/services/shellExecutionService.ts
galdawave: Added this 

image

not sure if i should actually make it a warning/notification message that shows up as yellow.

packages/core/src/services/shellExecutionService.test.ts
```diff
@@ -356,6 +356,50 @@ describe('ShellExecutionService', () => {
...
     });
+
+    it('should send SIGTERM and then SIGKILL on abort', async () => {
+      const { result } = await simulateExecution(
+        'long-running-process',
+        async (pty, abortController) => {
+          abortController.abort();
+          // The service sends SIGTERM, waits ~200ms, then sends SIGKILL.
+          // We wait for that timeout to complete before simulating process exit.
+          await new Promise((res) => setTimeout(res, 250));
```
jacob314: this test is making me nervous. shouldn't the mock at least be verifying that SIGKILL was sent before it exits?

jacob314 (approved):
lgtm

packages/core/src/services/shellExecutionService.test.ts
galdawave: Refactored this test to verify that sigkill is sent

PR #10610, opened 2025-10-06
fix(cli): Banish trailing input when using -p in loops (#10556)

maruthi-raj:
# Fix: Prevent stdin consumption when explicit prompts provided via CLI args

Only read stdin when no explicit prompt provided via CLI arguments.

## TLDR

Fixes the "banishing trailing input" issue where using `-p` option in shell loops caused subsequent input to disappear. The CLI now only reads from stdin when no explicit prompt is provided via command-line arguments (`-p`, `-i`, or positional args), preventing it from consuming piped input that should be available to subsequent commands in loops.

## Dive Deeper

**Problem:** When running Gemini CLI with the `-p` option inside shell loops that process piped input, the CLI was consuming stdin even though a prompt was explicitly provided via command-line arguments. This caused trailing input to "disappear" because the CLI consumed it unnecessarily.

**Example of the bug:**
```bash
$ echo -e "1\n2" | while read n; do echo "n=$n"; gemini -p "Hi." ; done
n=1
Loaded cached credentials.
Hello! How can I help you today?
# Missing: n=2 (because gemini consumed the remaining stdin)

Solution: Added logic to check for explicit prompts before attempting to read from stdin:

CLI now only reads stdin when !process.stdin.isTTY AND no explicit prompt is provided
Covers both main execution path and sandbox/non-interactive paths
Preserves normal piped input functionality when no CLI args are provided

## Reviewer Test Plan

To validate this fix works:
1. Test the original bug scenario:
echo -e "1\n2\n3" | while read n; do echo "n=$n"; gemini -p "Say hi briefly"; done
Expected: Should see n=1, n=2, n=3 with gemini responses between them
2. Test normal piped input still works:
echo "What is 2+2?" | gemini
Expected: Should process the piped question normally
3. Test other explicit prompt methods:
echo -e "1\n2" | while read n; do echo "n=$n"; gemini "Say hi to $n"; done
Expected: Should work with positional arguments too
4. Test interactive mode unaffected:
gemini -i
Expected: Should start interactive mode normally

## Testing Matrix
npm run | ✅ | ❓ | ❓
npx | ✅ | ❓ | ❓
Docker | ❓ | ❓ | ❓
Podman | ❓ | - | -
Seatbelt | ❓ | - | -



## Linked issues / bugs
Fixes #10556


maruthi-raj:
@google-gemini/gemini-cli-maintainers @gemini-cli-robot Can you please take a look into this PR whenever you get time ?

PR #10611, opened 2025-10-06
fix(cli) : fixed bug #8310 where /memory refresh will create discrepancies with initial memory load ignoring settings/config for trusted folder and file filters

sgnagnarella:
## TLDR

- Refactored /memory refresh command to make sure it uses loadHierarchicalGeminiMemory to match the initial memory load on gemini.tsx
- Made sure that the number of files in the UI was also refreshed as part of this command call.
- Used the right call to determine folder trust in this context.
- Fixed file filtering options to use the same defaults across the board coming from DEFAULT_MEMORY_FILE_FILTERING_OPTIONS when no values exist in settings.json
- Fixed tests and integrations tests across the board

## Reviewer Test Plan

1. Start Gemini CLI trusting the workspace folder
2. Run /memory refresh
3. Number of GEMINI.md files used should match the initial load
4. Create a new GEMINI.md file within the workspace folder stracture
5. Run /memory refresh
6. The new number of GEMINI.md files in use should have incremented by 1

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | X  | ❓  | ❓  |
| npx      | X  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

- Fixes #8310


packages/cli/src/config/config.integration.test.ts
```diff
@@ -78,12 +81,14 @@ describe('Configuration Integration Tests', () => {
         sandbox: false,
         targetDir: tempDir,
         debugMode: false,
-        fileFilteringRespectGitIgnore: undefined, // Should default to true
+        fileFilteringRespectGitIgnore: undefined, // Should default to  DEFAULT_FILE_FILTERING_OPTIONS
```
richieforeman, line 84: Super nit: `DEFAULT_FILE_FILTERING_OPTIONS.respectGitIgnore`

sgnagnarella: yes good catch!

packages/cli/src/ui/commands/memoryCommand.ts
```diff
@@ -82,25 +80,26 @@ export const memoryCommand: SlashCommand = {
 
         try {
           const config = await context.services.config;
+          const settings = context.services.settings;
           if (config) {
             const { memoryContent, fileCount, filePaths } =
-              await loadServerHierarchicalMemory(
+              await loadHierarchicalGeminiMemory(
```
richieforeman, line 86: Confirmed that these are wrappers of each other.

LG.

richieforeman (approved):

packages/core/src/config/config.ts
```diff
@@ -882,6 +886,9 @@ export class Config {
     return this.ideMode;
   }
 
+  /***
```
shrutip90, line 892: This is no longer required actually. Also  getFolderTrust seems to have incorrect comments and its usage in FileCommandLoader also looks wrong :(. I can fix these in a separate PR. Thanks for pointing this out.

shrutip90:
LGTM for the folder trust related changes. Thanks for fixing!

packages/core/src/config/config.ts
shrutip90: Sent out https://github.com/google-gemini/gemini-cli/pull/10837.

bdmorgan (approved):

PR #10612, opened 2025-10-06
feat(core): generalize path correction for use across tools

agmsb:
## TLDR

In this PR, we take the path correction logic in Smart Edit and factor it out into a utility for use across tools.

## Dive Deeper

Path correction is needed in more tools than just Smart Edit. For example, we have seen absolute path errors in Read File that we can circumvent with this utility. In a subsequent PR, will incorporate usage of this utility into relevant tools.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | x  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Related to #9030, #4137.


packages/core/src/utils/pathCorrector.ts
```diff
@@ -0,0 +1,53 @@
...
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import * as fs from 'node:fs';
+import * as path from 'node:path';
+import type { Config } from '../config/config.js';
+
+export interface PathCorrectionResult {
```
abhipatel12: Can we make this a strict discriminated union?

Just to avoid having to use non-null assertions where consumers use this?

packages/core/src/tools/smart-edit.ts
```diff
@@ -914,8 +863,11 @@ A good instruction should concisely answer:
 
     if (!path.isAbsolute(params.file_path)) {
       // Attempt to auto-correct to an absolute path
-      const error = this.correctPath(params);
-      if (error) return error;
+      const result = correctPath(params.file_path, this.config);
+      if (result.error) {
+        return result.error;
+      }
+      params.file_path = result.correctedPath!;
```
abhipatel12: from the other comment, if we have the discriminated union, we can avoid the `!` here.

abhipatel12 (commented):
Quick change requested otherwise lgtm! We can follow up with the other changes discussed offline!

abhipatel12 (approved):
This looks great! LGTM

PR #10630, opened 2025-10-07
feat(core): wire up UI for ASK_USER policy decisions in message bus

allenhutchison:
## Summary

Completes the message bus integration by wiring up the UI for `ASK_USER` policy decisions. When the PolicyEngine decides that a tool requires user confirmation, the system now shows the full legacy confirmation UI instead of blocking or auto-executing.

### Core Infrastructure
- **`confirmation-bus/types.ts`**: Added `requiresUserConfirmation` flag to `ToolConfirmationResponse`
- **`tools/tools.ts`**: Updated `BaseToolInvocation.handleMessageBusConfirmation()` to handle ASK_USER by returning `null` as a sentinel value
- **`core/coreToolScheduler.ts`**: 
  - Subscribe to `TOOL_CONFIRMATION_REQUEST` messages in constructor (when `enableMessageBusIntegration` is true)
  - Added `handleToolConfirmationRequest()` method to respond with `requiresUserConfirmation=true` for ASK_USER decisions

### Integration & Testing
- **`tools/web-fetch.ts`**: 
  - Integrated as first end-to-end test of ASK_USER flow
  - Updated `shouldConfirmExecute()` to check for `null` return and fall through to legacy confirmation UI
- **`tools/web-fetch.test.ts`**: Added 8 comprehensive message bus integration tests
- **Test mocks**: Updated all Config mocks to include `getEnableMessageBusIntegration()`, `getMessageBus()`, `getPolicyEngine()`

## Flow

When `enableMessageBusIntegration: true` in settings:

1. **ALLOW** → Tool executes immediately (no UI shown)
2. **DENY** → Tool execution blocked with error
3. **ASK_USER** → Shows full legacy confirmation UI (the new behavior added in this PR)

### ASK_USER Flow Details
```
Tool.shouldConfirmExecute()
  ↓
BaseToolInvocation publishes TOOL_CONFIRMATION_REQUEST
  ↓
MessageBus → PolicyEngine.check() → ASK_USER
  ↓
MessageBus emits request to CoreToolScheduler
  ↓
CoreToolScheduler responds with requiresUserConfirmation=true
  ↓
BaseToolInvocation receives response → returns null
  ↓
Tool checks for null → falls back to legacy confirmation
  ↓
UI shows full confirmation dialog (diffs, URLs, approval options)
```

## Test Plan

- [x] All unit tests passing (2502 tests)
- [x] Message bus integration tests for WebFetchTool (8 new tests)
- [x] Preflight checks passing (format, lint, build, typecheck)
- [x] Manual testing with `tools.enableMessageBusIntegration: true`

## Next Steps

This PR enables the ASK_USER flow for WebFetchTool. Follow-up work:
- Integrate remaining write tools (EditTool, WriteFileTool, ShellTool, etc.)
- Add policy configuration examples to documentation
- Consider adding telemetry for policy decisions


packages/core/src/tools/tools.ts
```diff
@@ -161,7 +161,11 @@ export abstract class BaseToolInvocation<
           if (response.correlationId === correlationId) {
             cleanup();
 
-            if (response.confirmed) {
+            if (response.requiresUserConfirmation) {
+              // Policy decision was ASK_USER - resolve with null to signal
+              // that legacy confirmation UI should be shown
+              resolve(null as unknown as false); // null signals: use legacy confirmation
```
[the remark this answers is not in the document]
allenhutchison: Fixed

packages/core/src/confirmation-bus/types.ts
```diff
@@ -24,6 +24,11 @@ export interface ToolConfirmationResponse {
   type: MessageBusType.TOOL_CONFIRMATION_RESPONSE;
   correlationId: string;
   confirmed: boolean;
+  /**
+   * When true, indicates that policy decision was ASK_USER and the tool should
+   * show its legacy confirmation UI instead of auto-proceeding.
+   */
+  requiresUserConfirmation?: boolean;
```
cornmander, line 31: Does this need to be optional?

packages/core/src/tools/message-bus-integration.test.ts
```diff
@@ -52,6 +52,21 @@ class TestToolInvocation extends BaseToolInvocation<TestParams, TestResult> {
       testValue: this.params.testParam,
     };
   }
+
+  override async shouldConfirmExecute(
+    abortSignal: AbortSignal,
+  ): Promise<false> {
+    if (this.messageBus) {
```
cornmander, line 60: should this conditional always evaluate to true? Can we assert on it instead of making it a conditional?

allenhutchison: The conditional supports the case where there is no message bus in the test. It's flag guarded so I wanted to test both scenarios. See the test on line 222 for example.

packages/core/src/confirmation-bus/types.ts
allenhutchison: The way this is written right now is that this only comes into play when the ASK_USER policy decision is set. Otherwise it's null and falls through to the legacy behviour.

packages/core/src/tools/tools.ts
```diff
@@ -122,85 +112,74 @@ export abstract class BaseToolInvocation<
...
         }
-      },
-    );
+      };
+
+      abortSignal.addEventListener('abort', abortHandler);
+
+      timeoutId = setTimeout(() => {
+        cleanup();
+        resolve('ALLOW'); // Default to ALLOW on timeout
```
abhipatel12: Why do we default to allow for timeout here?

packages/core/src/tools/tools.ts
abhipatel12, line 82: off topic, but we might want to change this to `console.debug`

packages/core/src/tools/tools.ts
```diff
@@ -122,85 +112,74 @@ export abstract class BaseToolInvocation<
...
+      const responseHandler = (response: ToolConfirmationResponse) => {
+        if (response.correlationId === correlationId) {
           cleanup();
-          resolve(false);
+          if (response.requiresUserConfirmation) {
+            resolve('ASK_USER');
+          } else if (response.confirmed) {
+            resolve('ALLOW');
+          } else {
+            reject(new Error('Tool execution denied by policy'));
```
abhipatel12: Why do we reject instead of resolve to `DENY` here?

packages/core/src/tools/tools.ts
allenhutchison: Good point. I've changed this to ASK_USER.

packages/core/src/tools/tools.ts
allenhutchison: done

packages/core/src/tools/tools.ts
allenhutchison: This was left over from an older implementation before I had the status in the promise. I've changed it to resolve to DENY now to be consistent.

abhipatel12 (approved):
LGTM!

Issue #10270, opened 2025-09-30
Model acts unexpectedly when context grows too large


hminooei:
### What happened?

This happens due to a phenomenon called context rot, where model gets confused/lost and misses to pay attention to the important system instructions, guidelines, or even the latest user's instruction.

### What did you expect to happen?

The model should always follow the system instructions as well as guidelines (e.g. GEMINI.md) and user's explicit instructions.

### Client information

<details>
<summary>Client Information</summary>

Run `gemini` to enter the interactive CLI, then run the `/about` command.

```console
> /about
# paste output here
```

</details>

reconsumeralization:
in my initial crypto mine ticket + the npm isssue - the image shows a time for how long I was able to sustain the self degridation 90% it's a manifest of same issue.

scidomino:
Note this related bug: https://github.com/google-gemini/gemini-cli/issues/12546

PR #10651, opened 2025-10-07
Update system instructions for optimizing shell tool commands

hminooei:
## TLDR
Our goal is to optimize the shell commands to promote concise and high-signal outputs, reducing context bloat.

## Dive Deeper

Please see [here.
](https://docs.google.com/document/d/1G93b65jOwK_QB1yB0-0ZPqsUvPZS7EQin0sJleaxtuE/edit?tab=t.plqodxtpvm8f)
## Reviewer Test Plan

No regression on swebench-v and Terminal Bench.

## Linked issues / bugs

This PR makes progress on #10270

### DIFF
--- a/packages/core/src/config/config.ts
+++ b/packages/core/src/config/config.ts
@@ -970,6 +970,10 @@ export class Config {
     return this.continueOnFailedApiCall;
   }
 
+  getEnableShellOutputEfficiency(): boolean {
+    return this.enableShellOutputEfficiency;
+  }
+
   getShellExecutionConfig(): ShellExecutionConfig {
     return this.shellExecutionConfig;
   }
--- a/packages/core/src/config/config.ts
+++ b/packages/core/src/config/config.ts
@@ -278,6 +278,7 @@ export interface ConfigParameters {
   enableMessageBusIntegration?: boolean;
   codebaseInvestigatorSettings?: CodebaseInvestigatorSettings;
   continueOnFailedApiCall?: boolean;
+  enableShellOutputEfficiency?: boolean;
 }
 
 export class Config {
@@ -371,6 +372,7 @@ export class Config {
   private readonly enableMessageBusIntegration: boolean;
   private readonly codebaseInvestigatorSettings?: CodebaseInvestigatorSettings;
   private readonly continueOnFailedApiCall: boolean;
+  private readonly enableShellOutputEfficiency: boolean;
 
   constructor(params: ConfigParameters) {
     this.sessionId = params.sessionId;
@@ -465,6 +467,8 @@ export class Config {
       params.enableMessageBusIntegration ?? false;
     this.codebaseInvestigatorSettings = params.codebaseInvestigatorSettings;
     this.continueOnFailedApiCall = params.continueOnFailedApiCall ?? true;
+    this.enableShellOutputEfficiency =
+      params.enableShellOutputEfficiency ?? true;
     this.extensionManagement = params.extensionManagement ?? true;
     this.storage = new Storage(this.targetDir);
     this.enablePromptCompletion = params.enablePromptCompletion ?? false;
--- a/packages/core/src/core/prompts.ts
+++ b/packages/core/src/core/prompts.ts
@@ -162,6 +162,33 @@ ${(function () {
 6. **Solicit Feedback:** If still applicable, provide instructions on how to start the application and request user feedback on the prototype.
 
 # Operational Guidelines
+${(function () {
+  if (config.getEnableShellOutputEfficiency()) {
+    const tempDir = config.storage.getProjectTempDir();
+    return `
+## Shell tool output token efficiency:
+
+IT IS CRITICAL TO FOLLOW THESE GUIDELINES TO AVOID EXCESSIVE TOKEN CONSUMPTION.
+
+- Always prefer command flags that reduce output verbosity when using '${ShellTool.Name}'.
+- Aim to minimize tool output tokens while still capturing necessary information.
+- If a command is expected to produce a lot of output, use quiet or silent flags where available and appropriate.
+- Always consider the trade-off between output verbosity and the need for information. If a command's full output is essential for understanding the result, avoid overly aggressive quieting that might obscure important details.
+- If a command does not have quiet/silent flags or for commands with potentially long output that may not be useful, redirect stdout and stderr to temp files in the project's temporary directory: ${tempDir}. For example: 'command > ${path.posix.join(
+      tempDir,
+      'out.log',
+    )} 2> ${path.posix.join(tempDir, 'err.log')}'.
+- After the command runs, inspect the temp files (e.g. '${path.posix.join(
+      tempDir,
+      'out.log',
+    )}' and '${path.posix.join(
+      tempDir,
+      'err.log',
+    )}') using commands like 'grep', 'tail', 'head', ... (or platform equivalents). Remove the temp files when done.
+`;
+  }
+  return '';
+})()}
 
 ## Tone and Style (CLI Interaction)
 - **Concise & Direct:** Adopt a professional, direct, and concise tone suitable for a CLI environment.
--- a/docs/get-started/configuration.md
+++ b/docs/get-started/configuration.md
@@ -256,6 +256,10 @@ their corresponding top-level category object in your `settings.json` file.
   - **Description:** Skip the next speaker check.
   - **Default:** `false`
 
+- **`model.enableShellOutputEfficiency`** (boolean):
+  - **Description:** Optimizes shell tool commands for token efficiency.
+  - **Default:** `true`
+
 #### `context`
 
 - **`context.fileName`** (string or array of strings):
--- a/packages/core/src/core/__snapshots__/prompts.test.ts.snap
+++ b/packages/core/src/core/__snapshots__/prompts.test.ts.snap
@@ -50,6 +50,18 @@ When requested to perform tasks like fixing bugs, adding features, refactoring, 
 
 # Operational Guidelines
 
+## Shell tool output token efficiency:
+
+IT IS CRITICAL TO FOLLOW THESE GUIDELINES TO AVOID EXCESSIVE TOKEN CONSUMPTION.
+
+- Always prefer command flags that reduce output verbosity when using 'run_shell_command'.
+- Aim to minimize tool output tokens while still capturing necessary information.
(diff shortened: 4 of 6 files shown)


packages/core/src/core/prompts.ts
```diff
@@ -162,6 +162,48 @@ ${(function () {
 
 # Operational Guidelines
 
+## Output token efficient flags:
```
owenofbrien: We need to guard this behind a user setting flag so we can measure the impact.

owenofbrien (changes requested):

packages/core/src/core/prompts.ts
hminooei: Done, ptal.

packages/core/src/config/config.ts
```diff
@@ -443,6 +445,7 @@ export class Config {
     this.enableMessageBusIntegration =
       params.enableMessageBusIntegration ?? false;
     this.enableSubagents = params.enableSubagents ?? false;
+    this.tokenEfficient = params.tokenEfficient ?? true;
```
anj-s: should this be true by default? Also can we add to the documentation if we are introducing a new settings option.

packages/core/src/core/prompts.ts
```diff
@@ -162,6 +162,55 @@ ${(function () {
...
+  if (config.getTokenEfficient()) {
+    return `
+## Output token efficient flags:
+
+IT IS CRITICAL TO FOLLOW THESE GUIDELINES TO AVOID EXCESSIVE OUTPUT TOKENS.
+
+- Always prefer command flags that reduce output verbosity when using '${ShellTool.Name}'.
+- Aim to minimize output tokens while still capturing necessary information.
+- If a command is expected to produce a lot of output, use quiet or silent flags where available and appropriate.
+- If a command does not have quiet/silent flags, consider redirecting output to log files and inspecting those files instead of displaying all output directly.
```
anj-s: How will this work with the tool output summarization feature?

packages/core/src/config/config.ts
hminooei: I believe so. Also double checking in the chat thread. 
Good point, added the documentation.

packages/core/src/core/prompts.ts
hminooei: There could be some edge cases where the model follows this guideline and when reading the output files, the summarization interferes. Ideally the summarization interference should be complementary.

On swebench and TB so far this hasn't shown any problem.

packages/core/src/config/config.ts
```diff
@@ -269,6 +269,7 @@ export interface ConfigParameters {
   useModelRouter?: boolean;
   enableMessageBusIntegration?: boolean;
   enableSubagents?: boolean;
+  tokenEfficient?: boolean;
```
owenofbrien: maybe call this something like enableQuietShellOutput - the current name makes it sound like it has to do with the efficiency of token processing during inference

packages/core/src/core/prompts.ts
```diff
@@ -162,6 +162,55 @@ ${(function () {
 
 # Operational Guidelines
 
+${(function () {
+  if (config.getTokenEfficient()) {
+    return `
+## Output token efficient flags:
+
+IT IS CRITICAL TO FOLLOW THESE GUIDELINES TO AVOID EXCESSIVE OUTPUT TOKENS.
```
owenofbrien: nit: "output tokens" -> "token consumption"

packages/core/src/core/prompts.ts
```diff
@@ -162,6 +162,55 @@ ${(function () {
...
+
+- Always prefer command flags that reduce output verbosity when using '${ShellTool.Name}'.
+- Aim to minimize output tokens while still capturing necessary information.
```
owenofbrien: I'd omit this line, as it seems unrelated to the current goal of reducing shell output verbosity and could make the model's conversational output more terse, which could impact the user experience.

packages/core/src/core/prompts.ts
owenofbrien: "reading those files only as needed"

packages/core/src/core/prompts.ts
```diff
@@ -162,6 +162,55 @@ ${(function () {
...
+- If a command is expected to produce a lot of output, use quiet or silent flags where available and appropriate.
+- If a command does not have quiet/silent flags, consider redirecting output to log files and inspecting those files instead of displaying all output directly.
+- Always consider the trade-off between output verbosity and the need for information. If a command's output is essential for understanding the result, avoid overly aggressive quieting that might obscure important details.
```
owenofbrien: "If a command's full output is essential"

packages/core/src/core/prompts.ts
```diff
@@ -162,6 +162,55 @@ ${(function () {
...
+
+Examples of quiet/silent flags:
+On Linux:
+- Use -q or -qq or --quiet flag for quiet command runs when appropriate and only see errors.
```
owenofbrien: "when appropriate and only see errors" -> "to only see errors"

packages/core/src/core/prompts.ts
```diff
@@ -162,6 +162,55 @@ ${(function () {
...
+- Use -s or -sS when appropriate and only see errors.
```
owenofbrien: add --silent here and consider combining these with the previous line since they all have similar effects

packages/core/src/core/prompts.ts
```diff
@@ -162,6 +162,55 @@ ${(function () {
...
+On Windows:
+- Use command-specific quiet options if available and appropriate (e.g., del /q *.tmp).
+<example>
+user: Delete all the temporary files in this directory without prompting me for each one.
+model: I can run \`del /q *.tmp\`. This will permanently delete all files with the .tmp extension in the current directory without asking for confirmation.
+</example>
+
+## Local file system as memory:
+
+- For commands with potentially long output, redirect stdout and stderr to temp files (e.g., 'command > out.log 2> err.log'). After the command runs inspect the temp 'out.log' and 'err.log' files. Remove them when done.
```
owenofbrien: "potentially long output that may not be useful"

owenofbrien: If the model inspects the file contents via a 'cat' cmd or the read_file tool, the file contents become part of the history, and the model doesn't have a way to remove them. Deleting the files won't help. Redirecting to files is only helpful if we instruct the model to "use 'grep' to search for the files for lines containing needed information, such as error details" and to "read the files only if necessary (e.g. for debugging)".

owenofbrien (changes requested):

packages/core/src/config/config.ts
hminooei: Makes sense. Used `enableShellOutputEfficiency` to cover both silent/quiet and local file usage.

packages/core/src/core/prompts.ts
hminooei: modified to: `Aim to minimize tool output tokens ...`

packages/core/src/core/prompts.ts
hminooei: done, thanks

packages/core/src/core/prompts.ts
hminooei: done..

packages/core/src/core/prompts.ts
hminooei: done, thanks

packages/core/src/core/prompts.ts
hminooei: Kept the 'when appropriate' as some commands don't have this flag.

packages/core/src/core/prompts.ts
hminooei: rephased it, and replaced cat with grep. Interestingly, in practice, the model was using `grep` mostly.

packages/core/src/core/prompts.ts
```diff
@@ -162,6 +162,48 @@ ${(function () {
...
+<example>
+user: Run the build and let me know if there are any errors.
+model: [tool_call: ${ShellTool.Name} for 'npm run build > build.log 2> build.err' since npm run build can produce a lot of output.]
+(After command execution)
+model: The build process is complete. I will now check the error log.
+[tool_call: ${ShellTool.Name} for 'tail -n 10 build.err']
+(After noticing errors exist)
+model: Read the whole error log to understand the issues.
+[tool_call: ${ShellTool.Name} for 'cat build.err']
+(After reading the err.log file and deleted the log files)
```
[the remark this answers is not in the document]
owenofbrien: Good idea

packages/core/src/core/prompts.ts
owenofbrien: It doesn't look like this comment was resolved.

owenofbrien (approved):

packages/core/src/core/prompts.ts
```diff
@@ -161,6 +161,24 @@ ${(function () {
...
+  if (config.getEnableShellOutputEfficiency()) {
+    return `
+## Shell tool output token efficiency:
+
+IT IS CRITICAL TO FOLLOW THESE GUIDELINES TO AVOID EXCESSIVE TOKEN CONSUMPTION.
+
+- Always prefer command flags that reduce output verbosity when using '${ShellTool.Name}'.
+- Aim to minimize tool output tokens while still capturing necessary information.
+- If a command is expected to produce a lot of output, use quiet or silent flags where available and appropriate.
+- If a command does not have quiet/silent flags, consider redirecting output to log files and inspecting those files only as needed.
```
anj-s: do we need to specify which output folder logs files should be streamed to?

packages/core/src/core/prompts.ts
```diff
@@ -161,6 +161,24 @@ ${(function () {
...
+- Always consider the trade-off between output verbosity and the need for information. If a command's full output is essential for understanding the result, avoid overly aggressive quieting that might obscure important details.
+- For commands with potentially long output that may not be useful, redirect stdout and stderr to temp files (e.g., 'command > out.log 2> err.log').
```
anj-s: there is already a bullet point for redirecting output so maybe we can combine the two points?

anj-s (approved):

packages/core/src/core/prompts.ts
hminooei: done

PR #10659, opened 2025-10-07
bug(core): Process returned function calls sequentially.

joshualitt:
## TLDR

Moves to sequential function call processing.

## Dive Deeper

Moves away from parallel FC to sequential FC to ensure we can process all of the models returned function calls. This ensures we can always execute all function calls returned from the model, hopefully reducing discrepancies between model thought and action.

We can move to batch function calling later if we so desire.

## Reviewer Test Plan

Ask Gemini to issue multiple function calls "at one time," and confirm correct behavior.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes #9268


packages/core/src/core/coreToolScheduler.ts
```diff
@@ -819,7 +819,7 @@ export class CoreToolScheduler {
           );
         }
       }
-      this.attemptExecutionOfScheduledCalls(signal);
+      await this.attemptExecutionOfScheduledCalls(signal);
```
anj-s, line 822: is there a unit test for testing that tool calls are executed sequentially?

A recent PR added parallel tool calls. Is this just moving back to that version?

joshualitt: > is there a unit test for testing that tool calls are executed sequentially?

Yes, please see `CoreToolScheduler Sequential Execution` in `coreToolScheduler.test.ts`

> A recent PR added parallel tool calls. Is this just moving back to that version?

Are you talking about #8513  or another PR? I did revert #8513 in this PR. In addition, we needed to change the `coreToolScheduler.ts` to actually await each tool in the list of returned tools before it attempts to execute the next. Without this change, the tools still execute in a racy fashion. Please see the changes to `attemptExecutionOfScheduledCalls` in `coreToolScheduler.ts`.

joshualitt:
Adding @jacob314 for a second set of eyes on the CoreToolScheduler changes. ptal.

anj-s (approved):
Lets run evals once before merging

joshualitt:
@jacob314 found some issues result from returning multiple function calls which require approval, particularly in corner cases i.e. cancellation. Will update this PR and then re-request reviews.

packages/core/src/core/coreToolScheduler.test.ts
```diff
@@ -1406,6 +1406,268 @@ describe('CoreToolScheduler request queueing', () => {
...
+
+    // The onAllToolCallsComplete should be called once with both results
+    const completedCalls = onAllToolCallsComplete.mock
+      .calls[0][0] as ToolCall[];
+    expect(completedCalls).toHaveLength(2);
+    expect(completedCalls[0].status).toBe('success');
+    expect(completedCalls[1].status).toBe('success');
+  });
+
+  it('should continue executing subsequent tools if one is cancelled', async () => {
```
jacob314: This test name is a bit misleading given the latest state of the pr. The test
      actually aborts the signal for all subsequent calls and confirms they are cancelled. A more descriptive name might be "should cancel subsequent tools when the signal is aborted."

jacob314 (approved):
lgtm
Now works well when I tested locally.

PR #10661, opened 2025-10-07
chore(shell): Enable interactive shell by default

galdawave:
## TLDR

This pull request changes the default behavior of the Gemini CLI to use an interactive shell for commands.

## Dive Deeper

Previously, the interactive shell, which uses `node-pty`, was an opt-in feature. This change makes it the default experience for a more intuitive and user-friendly CLI.

The following changes were made:
- The default value for `tools.shell.enableInteractiveShell` in `settingsSchema.ts` was changed from `false` to `true`.
- The documentation in `docs/get-started/configuration.md` was updated to reflect this new default.
- In `packages/cli/src/config/config.ts`, the logic was updated to default to `true` if the `enableInteractiveShell` setting is not explicitly defined.

This improves the out-of-the-box experience for users running shell commands through the Gemini CLI.

## Reviewer Test Plan

1.  Run a shell command, like `ls -l`, through the Gemini CLI.
2.  Verify that it opens an interactive shell session by default.
3.  Explicitly set `tools.shell.enableInteractiveShell` to `false` in the configuration.
4.  Run the same shell command and verify that it runs non-interactively.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


jacob314 (approved):
lgtm

docs/get-started/configuration.md
```diff
@@ -207,7 +207,7 @@ Settings are organized into categories. All settings should be placed within the
 
 - **`tools.shell.enableInteractiveShell`** (boolean):
 
-  Use `node-pty` for an interactive shell experience. Fallback to `child_process` still applies. Defaults to `false`.
+  Use `node-pty` for an interactive shell experience. Fallback to `child_process` still applies. Defaults to `true`.
```
NTaylorMullen: Let's make this more user facing node-pty is a detail

packages/cli/src/config/config.ts
```diff
@@ -737,7 +737,8 @@ export async function loadCliConfig(
     interactive,
     trustedFolder,
     useRipgrep: settings.tools?.useRipgrep,
-    shouldUseNodePtyShell: settings.tools?.shell?.enableInteractiveShell,
+    shouldUseNodePtyShell:
```
NTaylorMullen: Should update this to also be user facing. That way users trying to utilize AI to make changes will be more successful

NTaylorMullen (approved):

packages/cli/src/config/config.ts
galdawave: By user facing are you referring for it to be in /settings ? It already is available there as well as showColor

PR #10662, opened 2025-10-07
Update deployment.md -> installation.md and sidebar links.

jkcinouye:
## TLDR

This PR renames "deployment.md" to a more intuitive name ("installation.md") and updates the sidebar links.

## Dive Deeper

- This PR renames deployment.md, which has become more accurately titled installation.md. 
- Some minor changes have been made to deployment.md to increase readability. 
- Links to deployment.md have been updated. 
- Sidebar links have been updated to include the new installation.md link and a link to the changelog.

## Reviewer Test Plan

Check for any mistakes or unintended changes.


bobcatfish:
ahh this makes a lot of sense @jkcinouye ! qq, is it possible to add a redirect somehow just in case there are old links to the previous file? i think we should go ahead regardless but would be nice if could be as seamless as possible 😁

jkcinouye:
> ahh this makes a lot of sense @jkcinouye ! qq, is it possible to add a redirect somehow just in case there are old links to the previous file? i think we should go ahead regardless but would be nice if could be as seamless as possible 😁

Just checking - I think the only way to do a redirect would be to leave a stub in for deployment.md, correct? We did that for authentication.md, but I'm a little nervous to continue littering old files if this isn't something we've deeplinked frequently.

richieforeman (approved):

scidomino (approved):

PR #10670, opened 2025-10-07
docs: document installation command for MacPorts users

breun:
This pull request adds the command for installing `gemini-cli` for [MacPorts](https://www.macports.org) users to the `README.md` file.


breun:
> Additionally, as a follow-up, it would be beneficial to update the uninstall guide to include instructions for MacPorts to ensure a complete user journey.

I've added the command for uninstalling `gemini-cli` when it's installed via MacPorts. I've also added the command to uninstall `gemini-cli` when it's installed via Homebrew, because that wasn't there yet.

PR #10719, opened 2025-10-08
feat(sessions): add resuming to geminiChat and add CLI flags for session management

### LINKED ISSUE #1530: Add support for continuing existing conversations in non-interactive mode
Currently, each gemini CLI invocation starts fresh without memory of previous interactions. 

Proposed: Add --continue (-c) or --resume (-r) flag to maintain conversation context across CLI calls. This would allow users to build on previous responses without re-explaining context.

Example:
gemini -p "Explain X"
gemini --continue -p "What about Y?"

this infrastructure already exists in interactive mode, so this should be pretty easy to transfer over:
  - /chat save <tag> - Save current conversation
  - /chat resume <tag> - Resume saved conversation
  - /chat list - List saved conversations

but then -p commands would need to save every time it is used. so would need to think about this carefully.

bl-ue:
This PR implements session resuming in `geminiChat.ts`, session selection, listing, and deletion functions, and CLI flags `--list-sessions`, `--delete-session <id|index>`, and `--resume <id|index>`.

Closes #1530

---

**Note:** This PR is the **4th** in a series of PRs that implement automatic chat recording and resuming in both interactive and non-interactive modes, with command line flags for non-interactive resuming and session management, and an interactive session browser.

PR #4401 was the original monolithic PR, which implemented this feature from the ground up.  It's too large to merge as-is, however, and so it was broken up into several smaller PRs, of which this PR is the 4th.

Please see #4401's description for the full context of this PR.  The original monolithic PR's description contains a detailed analysis of the entire system, with an example of the recording JSON and a demo video.


packages/cli/src/utils/sessions.ts
```diff
@@ -0,0 +1,101 @@
...
+      index < 1 ||
+      index > sortedSessions.length
+    ) {
+      console.error(
+        `Invalid session identifier "${sessionIndex}". Use --list-sessions to see available sessions.`,
+      );
+      return;
+    }
+    sessionToDelete = sortedSessions[index - 1];
+  }
```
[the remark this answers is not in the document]
bl-ue: Done in 1daebb41e1096c47124e70ea6897edc009a2361b.

mrcabbage972:
I tried to resume a session, but the previous content of the session wasn't displayed. Is that a bug?

packages/cli/src/config/config.ts
```diff
@@ -328,6 +360,11 @@ export async function parseArguments(settings: Settings): Promise<CliArgs> {
           if (argv['prompt'] && argv['promptInteractive']) {
             return 'Cannot use both --prompt (-p) and --prompt-interactive (-i) together';
           }
+          if (argv.resume && !argv.prompt && !process.stdin.isTTY) {
```
mrcabbage972, line 262: Please add a unit test for this error being raised

packages/cli/src/config/config.ts
```diff
@@ -269,6 +272,35 @@ export async function parseArguments(settings: Settings): Promise<CliArgs> {
...
+          // one, and not being passed at all.
+          skipValidation: true,
+          description:
+            'Resume a previous session. Use "latest" for most recent or index number (e.g. --resume 5)',
+          coerce: (value: string): string => {
+            // When --resume passed with a value (`gemini --resume 123`): value = "123" (string)
+            // When --resume passed without a value (`gemini --resume`): value = "" (string)
+            // When --resume not passed at all: this `coerce` function is not called at all, and
+            //   `yargsInstance.argv.resume` is undefined.
+            if (value === '') {
```
mrcabbage972, line 191: Please move the 'latest' string to a constant and add a unit test to ensure that its being returned on an empty input.

packages/cli/src/config/config.ts
```diff
@@ -269,6 +272,35 @@ export async function parseArguments(settings: Settings): Promise<CliArgs> {
...
           description: 'List all available extensions and exit.',
         })
+        .option('resume', {
+          alias: 'r',
+          type: 'string',
+          // `skipValidation` so that we can distinguish between it being passed with a value, without
+          // one, and not being passed at all.
+          skipValidation: true,
+          description:
+            'Resume a previous session. Use "latest" for most recent or index number (e.g. --resume 5)',
```
mrcabbage972, line 185: What does index number mean here? E.g. is it 1-based or 0-based? Does it start from the oldest or the newest session?

packages/cli/src/gemini.tsx
```diff
@@ -470,7 +505,13 @@ export async function main() {
       console.log('Session ID: %s', sessionId);
     }
 
-    await runNonInteractive(nonInteractiveConfig, settings, input, prompt_id);
+    await runNonInteractive(
```
mrcabbage972: What is the expected behavior for non-interactive mode in respect to session management?

packages/cli/src/utils/sessions.ts
```diff
@@ -0,0 +1,75 @@
...
+ */
+
+import { ChatRecordingService, type Config } from '@google/gemini-cli-core';
+import {
+  formatRelativeTime,
+  SessionSelector,
+  type SessionInfo,
+} from './sessionUtils.js';
+
+export async function listSessions(config: Config): Promise<void> {
```
mrcabbage972, line 14: This method should have tests, including the cases when previous sessions were not found.

packages/cli/src/utils/sessions.ts
```diff
@@ -0,0 +1,75 @@
...
+    .forEach((session, index) => {
+      const current = session.isCurrentSession ? ', current' : '';
+      const time = formatRelativeTime(session.lastUpdated);
+      console.log(
+        `  ${index + 1}. ${session.firstUserMessage} (${time}${current}) [${session.id}]`,
+      );
+    });
+}
+
+export async function deleteSession(
```
mrcabbage972, line 39: Please add a test for a successful deletion of a session and for the case when the session was active and therefore could not be deleted.

bl-ue:
> I tried to resume a session, but the previous content of the session wasn't displayed. Is that a bug?

Hi @mrcabbage972, yes indeed, that was a bug!  I've fixed it, and I also added the unit test you suggested.  `--resume <session>` should work correctly now.

jk-kashe:
Gemini identified and fixed a bug related to the `--resume` flag that was preventing sessions from loading correctly.

**The Issue:**
There was a race condition in the `useEffect` hook within `packages/cli/src/ui/hooks/useSessionResume.ts`. The session loading logic was not explicitly waiting for the Gemini client to be fully initialized, leading to the session data being ignored if authentication was still in progress.

**Before (packages/cli/src/ui/hooks/useSessionResume.ts):**
```typescript
  useEffect(() => {
    if (
      resumedSessionData &&
      !isAuthenticating && // <-- This check caused the race condition
      !hasLoadedResumedSession.current
    ) {
      hasLoadedResumedSession.current = true;
      const historyData = convertSessionToHistoryFormats(
        resumedSessionData.conversation.messages,
      );
      loadHistoryForResume(
        historyData.uiHistory,
        historyData.clientHistory,
        resumedSessionData,
      );
    }
  }, [resumedSessionData, isAuthenticating, loadHistoryForResume]);
```

**The Fix:**
I've updated the `useEffect` condition to explicitly wait for `isGeminiClientInitialized` to be `true` *in addition to* checking `!isAuthenticating`. This ensures the session data is only loaded when authentication is not in progress *and* the Gemini client is fully ready, resolving the issue with `gemini -r`. The dependency array has also been updated to include `isGeminiClientInitialized`.

**After (packages/cli/src/ui/hooks/useSessionResume.ts):**
```typescript
  useEffect(() => {
    if (
      resumedSessionData &&
      !isAuthenticating &&
      isGeminiClientInitialized && // <-- Added this check
      !hasLoadedResumedSession.current
    ) {
      hasLoadedResumedSession.current = true;
      const historyData = convertSessionToHistoryFormats(
        resumedSessionData.conversation.messages,
      );
      loadHistoryForResume(
        historyData.uiHistory,
        historyData.clientHistory,
        resumedSessionData,
      );
    }
  }, [resumedSessionData, isAuthenticating, isGeminiClientInitialized, loadHistoryForResume]); // <-- Updated dependency array
```

This ensures the session data is only loaded when the Gemini client is fully ready, resolving the issue with `gemini -r`

@bl-ue could you please fix this in your PR?

bl-ue:
@jk-kashe Thank you very much for testing this PR and finding and fixing this bug!  I'll incorporate your fix right away.

mrcabbage972 (changes requested):
Please address my inline comments - mostly missing tests.

mrcabbage972:
There are some failing tests here: https://github.com/google-gemini/gemini-cli/actions/runs/18849798845/job/53897370500?pr=10719

packages/cli/src/config/config.ts
bl-ue: Both done in 6b58689c368574040c8ccaff421d28eb93e1ae21.

packages/cli/src/config/config.ts
bl-ue: 1-based, starting from the oldest.  So the first (oldest) session is 1, the second is 2, etc.  The indices are shown in the output of `--list-sessions`:
```
$ gemini --list-sessions

Available sessions for this project (4):

  1. Hi there! (10 days ago) [6c9aadd0-fe64-4ce6-b389-8c0729aaa027]
  2. Another chat (8 days ago) [5420cc78-eb43-483e-a6b4-7efd4c4fc36d]
  3. A third chat (2 days ago) [9e54ee88-9cf9-4936-8ea8-c17437a008f6]

```

At first I only implemented index-based identification.  Then per @ShuangLiu1993's [suggestion](https://github.com/google-gemini/gemini-cli/pull/4401#issuecomment-3210755266) I added the ability  to resume by the fixed ID as well.  I thought I'd let the two methods coexist for flexibility, but I'll remove the index-based one if you think the ID is enough.

packages/cli/src/gemini.tsx
bl-ue: In non-interactive mode, each invocation of `gemini --prompt` will create a new session that will be automatically recorded identically to interactive mode.  When `--prompt` is combined with `--resume` in `gemini --prompt "..." --resume <session-ID>`, it will be the same as normal non-interactive mode except the session history will be loaded first and used to generate the response.

packages/cli/src/utils/sessions.ts
bl-ue: Added 5 tests for `listSession` and 11 for `deleteSession` in 3c43b623858cb7bf1563b07b16c006201c54fba5.

packages/cli/src/utils/sessions.ts
bl-ue: Good idea, done in 3c43b623858cb7bf1563b07b16c006201c54fba5 as well.

bl-ue:
Hi @mrcabbage972, I fixed the merge conflicts and test failures and added 17 tests per your suggestions.  Will you please take another look?

The test failures were coincidentally due to #12118 which was merged just yesterday.

bl-ue:
Hi @mrcabbage972, thank you for approving the tests.  I fixed the failing ones&mdash;they were caused by changes from my merges of main over time.

mrcabbage972 (approved):

bl-ue:
Next: #13300

arjun810:
@bl-ue this has a small bug, where when you try to resume a session noninteractively with stdin, it doesn't accept it and gives you this error:

When resuming a session, you must provide a message via --prompt (-p) or stdin

the check on line 262 (          if (argv.resume && !argv.prompt && !process.stdin.isTTY) ) isn't quite right

bl-ue:
Hi @arjun810!  Thank for you for catching that!  The check was out of date anyway.  Fixed in #13520.

PR #10767, opened 2025-10-08
feat: Support Alt+key combinations

srivatsj:
Enhanced keyboard input processing to support:
- Alt+letter Unicode mappings for macOS terminals (iTerm2) that send Unicode characters instead of escape sequences
- ESC+character Alt handling for VS Code terminal compatibility
- Kitty protocol support for letter keys with all modifiers (not just Ctrl)
- Proper arrow key handling with Kitty protocol enabled

Added comprehensive test coverage for all new keyboard input scenarios

fixes: #10753

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


packages/cli/src/ui/contexts/KeypressContext.tsx
```diff
@@ -477,8 +512,21 @@ export function KeypressProvider({
       }
 
       if (['up', 'down', 'left', 'right'].includes(key.name)) {
-        broadcast(key);
-        return;
+        // If Kitty protocol is enabled and this looks like a CSI sequence,
+        // let it flow through Kitty parsing instead of broadcasting early
```
jacob314: what cases is this specifically needed for? this is a bit of a fragile and duplicate check with other kitty and special sequence handling.

packages/cli/src/ui/contexts/KeypressContext.tsx
```diff
@@ -603,6 +651,47 @@ export function KeypressProvider({
       if (key.name === 'return' && key.sequence === `${ESC}\r`) {
         key.meta = true;
       }
+
+      const unicodeChar = key.sequence || key.name;
```
jacob314: can we also handle 
ctrl-alt+a and similar correctly at least on kitty compat terminals?

packages/cli/src/ui/contexts/KeypressContext.tsx
```diff
@@ -603,6 +651,47 @@ export function KeypressProvider({
...
+        !key.name &&
+        key.sequence.length === 2 &&
+        key.sequence.startsWith(ESC)
+      ) {
+        const char = key.sequence[1];
+        // Only handle printable ASCII characters (a-z, A-Z, 0-9, etc.)
+        if (char && char.charCodeAt(0) >= 32 && char.charCodeAt(0) <= 126) {
+          broadcast({
+            name: char.toLowerCase(),
+            ctrl: false,
```
jacob314: would be nice if we do not ignore whether ctrl was already pressed if kitty parsing had correctly determined that.

packages/cli/src/ui/contexts/KeypressContext.test.tsx
```diff
@@ -749,6 +749,393 @@ describe('KeypressContext - Kitty Protocol', () => {
...
+            meta: true,
+            ctrl: false,
+          }),
+        );
+      },
+    );
+  });
+});
+
+describe('Kitty Protocol Letters', () => {
```
jacob314: can you test ctrl+meta with kitty and see if we can support that robustly. would be a nice to have for completeness. please check with real chars as well as mocking to match what you see.

jacob314 (commented):
Looks generally good. Some of these comments are mainly questions. For keyboard support we have to be extremely careful and do absolutely the minimum rather than letting the logic get too complex.

packages/cli/src/ui/contexts/KeypressContext.tsx
srivatsj: I added this to fix some failing tests in gemini-cli/packages/cli/src/ui/components/InputPrompt.test.tsx

packages/cli/src/ui/contexts/KeypressContext.tsx
srivatsj: added

packages/cli/src/ui/contexts/KeypressContext.tsx
srivatsj: done

packages/cli/src/ui/contexts/KeypressContext.test.tsx
srivatsj: Added

packages/cli/src/ui/contexts/KeypressContext.test.tsx
```diff
@@ -956,3 +956,147 @@ describe('Drag and Drop Handling', () => {
     });
   });
 });
+
+describe('Terminal-specific Alt+key combinations', () => {
+  /**
```
jacob314: remove this comment or changes it to `//` from `/**`

packages/cli/src/ui/contexts/KeypressContext.tsx
```diff
@@ -45,6 +45,38 @@ export const DRAG_COMPLETION_TIMEOUT_MS = 100; // Broadcast full path after 100m
 export const SINGLE_QUOTE = "'";
 export const DOUBLE_QUOTE = '"';
 
+// Map of accented characters to their corresponding Alt+letter combinations
+// These are sent by iTerm2/VSCode terminals when pressing Alt+letter, but with meta:false
+const ACCENTED_CHAR_TO_LETTER: Record<string, string> = {
+  å: 'a', // Alt+A
```
jacob314: remove these comments. and rename this to alt or meta key map rather than accented char to letter. That way the name describes what this is not what it looks like.

jacob314 (commented):
Approved after this last minor nit is addressed.

packages/cli/src/ui/contexts/KeypressContext.tsx
srivatsj: done

packages/cli/src/ui/contexts/KeypressContext.tsx
```diff
@@ -45,6 +45,36 @@ export const DRAG_COMPLETION_TIMEOUT_MS = 100; // Broadcast full path after 100m
...
+  '∆': 'j',
+  '˚': 'k',
+  '¬': 'l',
+  µ: 'm',
+  '˜': 'n',
+  ø: 'o',
+  π: 'p',
+  œ: 'q',
+  '®': 'r',
+  ß: 's',
```
jacob314: why aren't some of these chars in quotes?

srivatsj: prettier seems to be removing the single code for some unicode characters so using Unicode escape codes

jacob314 (approved):
lgtm

PR #10812, opened 2025-10-09
feat(ui): shorten context overflow message when <75% of limit

SandyTao520:
## TLDR

This PR updates the message displayed when the context window is about to overflow. If the remaining token count is less than 50% of the model's limit, a suggestion to reduce the message size or use the `/compress` command is appended to the message.

## Dive Deeper

The `handleContextWindowWillOverflowEvent` in `useGeminiStream.ts` has been updated to check the remaining token count against the model's token limit. The `tokenLimit` function is imported from `@google/gemini-cli-core`.

## Reviewer Test Plan

1.  Run the CLI with a model that has a known token limit.
2.  Send a message that causes a `ContextWindowWillOverflow` event where the remaining tokens are > 50% of the limit. Verify the message does not include the suggestion.
3.  Send a message that causes a `ContextWindowWillOverflow` event where the remaining tokens are < 50% of the limit. Verify the message includes the suggestion to use `/compress`.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ✅  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

### DIFF
--- a/packages/cli/src/ui/hooks/useGeminiStream.test.tsx
+++ b/packages/cli/src/ui/hooks/useGeminiStream.test.tsx
@@ -1861,16 +1861,16 @@ describe('useGeminiStream', () => {
         vi.mocked(tokenLimit).mockReturnValue(100);
       });
 
-      it('should add message without suggestion when remaining tokens are > 50% of limit', async () => {
+      it('should add message without suggestion when remaining tokens are > 75% of limit', async () => {
         // Setup mock to return a stream with ContextWindowWillOverflow event
-        // Limit is 100, remaining is 60 (> 50)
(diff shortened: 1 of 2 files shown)


jacob314 (approved):
This sort of polish is super useful to help understand what is going on.
As a nit, I would set the threshold so we show the message any time the context window is > 25% full rather than > 50 to error on the side of showing it to more users where `/compress` might fix their problems while still avoiding confusing users who were early in a conversation.lgtm

PR #10821, opened 2025-10-09
feat(ui): add flicker detection and metrics

skeshive:
## TLDR

Introduces a new `useFlickerDetector` hook to identify when the UI renders taller than the terminal, which is a sign of a rendering bug.

This change also adds a new telemetry metric, `gemini_cli.ui.flicker.count`, to record the frequency of these flicker events. This will help in monitoring and addressing UI rendering issues.

Screenshot 2025-10-09 at 2 47 19 PM

## Linked issues / bugs
Fixes #10598


packages/cli/src/ui/hooks/useFlickerDetector.ts
```diff
@@ -0,0 +1,52 @@
...
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { type DOMElement, measureElement } from 'ink';
+import { useEffect, useRef } from 'react';
+import { useConfig } from '../contexts/ConfigContext.js';
+import { recordFlickerFrame } from '@google/gemini-cli-core';
+
+const FLICKER_CHECK_INTERVAL_MS = 250;
```
skeshive: @jacob314 thoughts on this interval?

packages/cli/src/ui/components/DebugProfiler.tsx
```diff
@@ -170,6 +171,24 @@ export const DebugProfiler = () => {
...
   }, []);
 
+  useEffect(() => {
+    if (!showDebugProfiler) {
+      return;
+    }
+    const flickerHandler = () => {
+      profiler.totalFlickerFrames++;
+      profiler.reportAction();
+      appEvents.emit(
```
jacob314: we need to check whether we have intentionally gotten the UI into a flicker inducing state before logging an error. For example, if the user presses ctrl-s, that will cause a single frame of flicker which we shouldn't error about as we are intentionally showing expanded UI and we will close the expanded UI on the next keystroke.

packages/cli/src/ui/components/DebugProfiler.tsx
```diff
@@ -191,7 +210,12 @@ export const DebugProfiler = () => {
   return (
     <Text color={theme.status.warning} key={forceRefresh}>
       Renders: {profiler.numFrames} (total),{' '}
-      <Text color={theme.status.error}>{profiler.totalIdleFrames} (idle) </Text>
+      <Text color={theme.status.error}>{profiler.totalIdleFrames} (idle)</Text>
+      {profiler.totalFlickerFrames > 0 && (
+        <Text color={theme.status.error}>
+          , {profiler.totalFlickerFrames} (flicker)
```
jacob314: nit: comma should be yellow not red.

packages/cli/src/ui/contexts/UIStateContext.tsx
```diff
@@ -111,6 +111,7 @@ export interface UIState {
   terminalWidth: number;
   terminalHeight: number;
   mainControlsRef: React.MutableRefObject<DOMElement | null>;
+  rootUiRef: React.MutableRefObject<DOMElement | null>;
```
jacob314, line 115: comment that this is for performance profiling only so nobody abuses it in the future.

packages/cli/src/ui/hooks/useFlickerDetector.ts
jacob314: why are we using an interval instead of checking after each frame renders?

packages/cli/src/ui/AppContainer.tsx
```diff
@@ -209,12 +210,19 @@ export const AppContainer = (props: AppContainerProps) => {
...
   const mainControlsRef = useRef<DOMElement>(null);
+  const rootUiRef = useRef<DOMElement>(null);
   const originalTitleRef = useRef(
     computeWindowTitle(basename(config.getTargetDir())),
   );
   const lastTitleRef = useRef<string | null>(null);
   const staticExtraHeight = 3;
 
+  // Flicker detector
+  const FlickerDetector = () => {
```
jacob314: this is a bit of a nonstandard pattern. Why is this defined inline like this?

packages/cli/src/ui/components/DebugProfiler.tsx
jacob314: note this also causes one frame of flicker to log as 100 as we rerender each time this event is added. we need an alternate solution that doesn't cause something to be rendered every single frame of flicker. Suggestion: only log this the first time a flicker frame was detected and suggest typing `/profile` for more info. The UI in this class is carefully written to tradeoff updates for introducing extra renders.

jacob314 (commented):
Love it overall. A few comments and then should be good to land.

packages/cli/src/ui/hooks/useFlickerDetector.ts
skeshive: was worried about performance issues but seems like we optimize for reducing rendering frame re-renders?

jacob314: key is to not call setState or similar from the effect. as long as you avoid that you won't trigger another render.
If you open up `/profile`  you can verify that you haven't triggered extra frames to render. each time you move the arrows you type a key in the input prompt the frame count needs to only increase by 1. that would be a good integration test to add while you are adding a flicker integration test in the future.

jacob314 (approved):
lgtm

PR #10825, opened 2025-10-09
fix(infra) - Create a step to calculate the inputs for the nightly-release

shishu314:
## TLDR

Add a step to calculate inputs for the nightly release

## Dive Deeper

When we run a scheduled nightly-release, the inputs given is null:
image

This causes the inputs we pass in to our reusable actions to also be empty, causing the flow to be ran incorrectly, as seen in https://github.com/google-gemini/gemini-cli/actions/runs/18298235096/job/52100940858, where the `dry-run` check ends up causing actions to be ran in dry run mode.

Adding in a schedule check allows us to give a default value for the nightly scheduled run and hopefully making the job pass again.

## Reviewer Test Plan

Successful manual run: https://github.com/google-gemini/gemini-cli/actions/runs/18389332722/job/52395686759

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes #10772


.github/workflows/release-nightly.yml
```diff
@@ -58,6 +58,18 @@ jobs:
           JSON_INPUTS: '${{ toJSON(github.event.inputs) }}'
         run: 'echo "$JSON_INPUTS"'
 
+      - name: 'Set booleans for simplified logic'
```
skeshive: hmm I believe yesterday we talked about having a shared script across all workflows to calculate all the variables & set defaults when null values are present

I'm OK updating other workflows in a follow-up once we test everything but we should at least add the script in this PR and update the nightly workflow to use it

bdmorgan (approved):

.github/actions/calculate-vars/action.yml
```diff
@@ -0,0 +1,25 @@
...
+  using: 'composite'
+  steps:
+    - name: 'Set vars for simplified logic'
+      shell: 'bash'
+      env:
+        CRON: '${{ inputs.cron }}'
+        DRY_RUN_INPUT: '${{ inputs.dry_run }}'
+      run: |-
+        is_dry_run="true"
+        if [[ "${CRON}" == "0 0 * * *" || "${DRY_RUN_INPUT}" == "false" ]]; then
```
skeshive: hmm should we check for is_dry_run='' or null instead? would be less brittle if we change the schedule

.github/workflows/release-nightly.yml
shishu314: Done

skeshive (approved):

PR #10828, opened 2025-10-09
feat(core): Failed Response Retry via Extra Prompt

mrcabbage972:
## TLDR

This PR introduces a new feature to automatically "retry" a model call when the response stream terminates unexpectedly.

1. New Config Flag: A continueOnFailedApiCall flag (defaults to true) is added to control this behavior.
2. Retry Logic: If the flag is enabled and an InvalidStream event occurs, the client will automatically send the prompt "Please continue." to try and recover the rest of the model's response.
3. Safeguard: This retry mechanism will only execute once per user request.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

### DIFF
--- a/packages/core/src/config/config.ts
+++ b/packages/core/src/config/config.ts
@@ -455,7 +455,7 @@ export class Config {
     this.enableMessageBusIntegration =
       params.enableMessageBusIntegration ?? false;
     this.enableSubagents = params.enableSubagents ?? false;
-    this.continueOnFailedApiCall = params.continueOnFailedApiCall ?? false;
+    this.continueOnFailedApiCall = params.continueOnFailedApiCall ?? true;
     this.extensionManagement = params.extensionManagement ?? true;
     this.storage = new Storage(this.targetDir);
     this.enablePromptCompletion = params.enablePromptCompletion ?? false;
--- a/packages/core/src/config/config.ts
+++ b/packages/core/src/config/config.ts
@@ -260,6 +260,7 @@ export interface ConfigParameters {
   useModelRouter?: boolean;
   enableMessageBusIntegration?: boolean;
   enableSubagents?: boolean;
+  continueOnFailedApiCall?: boolean;
 }
 
 export class Config {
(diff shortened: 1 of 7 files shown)


packages/core/src/core/client.ts
```diff
@@ -595,16 +597,18 @@ export class GeminiClient {
...
-        // This recursive call's events will be yielded out, but the final
-        // turn object will be from the top-level call.
-        yield* this.sendMessageStream(
-          nextRequest,
-          signal,
-          prompt_id,
-          boundedTurns - 1,
-          // isInvalidStreamRetry is false here, as this is a next speaker check
-        );
+        if (this.config.getContinueOnFailedApiCall()) {
```
SandyTao520: not necessary here since we now won don't run the next speaker check anymore.

packages/core/src/core/client.ts
```diff
@@ -586,6 +589,31 @@ My setup is complete. I will provide my first command in the next turn.
...
+              this.config,
+              new ContentRetryFailureEvent(
+                4, // 2 initial + 2 after injections
+                'FAILED_AFTER_PROMPT_INJECTION',
+                modelToUse,
+              ),
+            );
+            return turn;
+          }
+          const nextRequest = [{ text: 'Please continue.' }];
```
anj-s: Can we say "System: Please continue"? This disambiguates from the user and won't be confusing.

packages/core/src/config/config.ts
```diff
@@ -453,6 +455,7 @@ export class Config {
     this.enableMessageBusIntegration =
       params.enableMessageBusIntegration ?? false;
     this.enableSubagents = params.enableSubagents ?? false;
+    this.continueOnFailedApiCall = params.continueOnFailedApiCall ?? false;
```
anj-s: do we want to turn this on by default to get immediate metrics?

anj-s (approved):

packages/core/src/core/client.ts
mrcabbage972: thx

packages/core/src/core/client.ts
mrcabbage972: Done

packages/core/src/config/config.ts
mrcabbage972: Done

PR #10844, opened 2025-10-09
Expose Codebase Investigator settings to the user

silviojr:
## TLDR

This PR allows users to customize the Codebase Investigator agent. Users can now control its behavior through new settings, such as:

Enabling or disabling the agent entirely.

Setting operational limits, including maximum turns and time limits.

Defining a thinking budget to manage resource usage.

Specifying which AI model the agent should use.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

### DIFF
--- a/packages/cli/src/config/settingsSchema.ts
+++ b/packages/cli/src/config/settingsSchema.ts
@@ -1066,15 +1066,6 @@ const SETTINGS_SCHEMA = {
           'Enable model routing to route requests to the best model based on complexity.',
         showInDialog: true,
       },
-      enableSubagents: {
-        type: 'boolean',
-        label: 'Enable Subagents',
-        category: 'Experimental',
-        requiresRestart: true,
-        default: false,
-        description: 'Enable experimental subagents.',
-        showInDialog: false,
-      },
       codebaseInvestigatorSettings: {
         type: 'object',
         label: 'Codebase Investigator Settings',
--- a/packages/cli/src/config/settingsSchema.ts
+++ b/packages/cli/src/config/settingsSchema.ts
@@ -1074,6 +1074,26 @@ const SETTINGS_SCHEMA = {
         description: 'Enable experimental subagents.',
         showInDialog: false,
       },
+      subagents: {
+        type: 'object',
+        label: 'Subagents',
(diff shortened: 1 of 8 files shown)


abhipatel12 (commented):
I think the UX here may be a bit confusing for users to test even in preview. 

We now have two flags (1 for subagents in general and 1 for codebase investigator). But since we can't optionally hide one vs the other. 

So a user in the `/settings` can enable the codebase investigator, but since the default for general subagents is false, nothing would actually happen since we don't register any of the agents in config.ts

Any thoughts about this?

silviojr:
> I think the UX here may be a bit confusing for users to test even in preview.
> 
> We now have two flags (1 for subagents in general and 1 for codebase investigator). But since we can't optionally hide one vs the other.
> 
> So a user in the `/settings` can enable the codebase investigator, but since the default for general subagents is false, nothing would actually happen since we don't register any of the agents in config.ts
> 
> Any thoughts about this?

Great point! Since we don't have other subagents for now, I'd prefer remove the "enableSubagents" for now. What do you think?

packages/cli/src/config/settingsSchema.ts
```diff
@@ -1065,14 +1066,65 @@ const SETTINGS_SCHEMA = {
...
+            default: false,
+            description: 'Enable the Codebase Investigator agent.',
+            showInDialog: true,
+          },
+          maxNumTurns: {
+            type: 'number',
+            label: 'Codebase Investigator Max Num Turns',
+            category: 'Experimental',
+            requiresRestart: true,
+            default: 15,
```
anj-s, line 1092: should this be lower?

anj-s (approved):

packages/cli/src/config/settingsSchema.ts
silviojr: That was the default for our experiments. That's why I decided to leave it like this.

PR #10850, opened 2025-10-09
fix(tests): enable and update prompt for MCP add tool test

SandyTao520:
## TLDR

Enables the previously skipped `should add two numbers` test in `integration-tests/simple-mcp-server.test.ts` and updates the prompt to be more explicit for the model.

## Dive Deeper

The test was skipped with a TODO linking to an issue. By making the prompt more direct ("Use the `add` tool to calculate 5+10 and output only the resulting number."), the test becomes reliable enough to be enabled. I ran this test 100 times locally, and it passed every time.

## Reviewer Test Plan

Run the integration tests:
`npm run test:integration`

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ✅  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Resolves #10738

### DIFF
--- a/integration-tests/simple-mcp-server.test.ts
+++ b/integration-tests/simple-mcp-server.test.ts
@@ -210,11 +210,12 @@ describe('simple-mcp-server', () => {
     }
   });
 
-  //TODO -https://github.com/google-gemini/gemini-cli/issues/10738
-  it.skip('should add two numbers', async () => {
+  it('should add two numbers', async () => {
     // Test directory is already set up in before hook
     // Just run the command - MCP server config is in settings.json
-    const output = await rig.run('add 5 and 10');
+    const output = await rig.run(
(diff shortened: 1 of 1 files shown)


galdawave (approved):

jakemac53 (approved):
Fwiw another option could be to change the MCP tool to be something that the model is less likely to attempt to just do on its own as opposed to adding two numbers which is quite a generic task that it could achieve without a tool.

For instance, a tool called "toggle_lights" and a prompt "turn off the lights" or something like that, would very likely be more reliable.

But, if this works reliably it is probably fine too.

bdmorgan (approved):

PR #10857, opened 2025-10-09
clean up extension tests

jakemac53:
## TLDR

- Removes mocks of `node:readline` which is really an implementation detail of `requestConsentNonInteractive` and instead passes a mock or custom function for requesting consent (this was already injectable).
- Updates the highly flaky install from git test so that it isn't a GitHub URL, which makes it not make http requests to github (this was causing it to hang, but only when ran with other tests, not sure why).

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Closes #10739 - this test was already unskipped but I discovered the other flaky test when looking into it. I also don't believe the PR that unskipped it actually did anything to make it less flaky, but this should.

### DIFF
--- a/packages/cli/src/config/extension.test.ts
+++ b/packages/cli/src/config/extension.test.ts
@@ -20,7 +20,6 @@ import {
   loadExtensionConfig,
   loadExtensions,
   performWorkspaceExtensionMigration,
-  requestConsentNonInteractive,
   uninstallExtension,
 } from './extension.js';
 import {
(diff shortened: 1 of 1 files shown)


jacob314 (approved):
lgtm

PR #10875, opened 2025-10-10
fix(SettingsDialog):  race condition in SettingsDialog causing settings to be unexpectedly cleared

lifefloating:
## TLDR

Fix race condition in SettingsDialog causing settings to be unexpectedly cleared. When toggling certain settings (e.g., `tools.shell.showColor`), other related settings (e.g., `tools.shell.enableInteractiveShell`) would sometimes be reset to false. The bug was intermittent, appearing after several setting changes.

**Root cause:** Duplicate state updates created a race condition where `setPendingSettings` was called both manually and via useEffect, causing inconsistent state loading.

**Fix:** Removed redundant manual `setPendingSettings` call, relying solely on useEffect for automatic state synchronization.

### The Problem

When a user toggled a non-restart-required setting like `tools.shell.showColor` multiple times, another setting in the same nested object (`tools.shell.enableInteractiveShell`) would sometimes be reset to false. This happened intermittently - the first 1-2 toggles worked fine, but after several changes the bug would appear.

### Investigation Process

1. **Initial hypothesis:** Found duplicate `setPendingSettings` call that incorrectly cast enum values to boolean
2. **Still broken after fix:** User confirmed the intermittent issue persisted
3. **Suspected applyKeyDiff:** Initially thought the sync-by-omission semantics of `applyKeyDiff` was deleting settings when using `{} as Settings` as base
4. **Fixed base object:** Changed to use `settings.forScope(selectedScope).settings` as base - still broken
5. **Debug logging:** Added comprehensive logging to trace the issue
6. **Root cause found:** Debug logs revealed setValue calls without corresponding SettingsDialog Toggle logs, indicating a race condition

### The Race Condition

In the toggle function (SettingsDialog.tsx:63-223):

```typescript
if (!requiresRestart(key)) {
  // 1. Save settings via saveModifiedSettings
  saveModifiedSettings(immediateSettings, immediateSettingsObject, settings, selectedScope);

  // 2. setValue is called internally, updating the settings object

  // 3. Manual setPendingSettings call (PROBLEMATIC)
  setPendingSettings(structuredClone(settings.forScope(selectedScope).settings));

  // 4. BUT ALSO: useEffect (line 105-128) triggers because settings changed
  //    useEffect also calls setPendingSettings
}
```

**The race:** Both the manual call (step 3) and the useEffect (step 4) were calling `setPendingSettings`. Depending on timing:
- If manual call executes first, then useEffect → correct state
- If useEffect executes first (from stale settings), then manual call → might load incomplete state
- The inconsistency caused settings to be unpredictably loaded, sometimes missing fields

### The Solution

Removed the redundant manual `setPendingSettings` call at line 118-120. Now only the useEffect handles state synchronization when `settings` object changes, eliminating the race condition.

### Setup
1. Checkout this branch
2. Build and run the application
3. Open Settings dialog

### Test Case 1: Rapid Toggle Test
1. Navigate to `Tools > Shell` section
2. Note the current value of "Enable Interactive Shell" (should be true by default)
3. Toggle "Show Color" on and off multiple times rapidly
4. **Expected:** "Enable Interactive Shell" should remain at its original value throughout
5. **Before fix:** After 2-5 toggles, "Enable Interactive Shell" would sometimes flip to false

### Test Case 2: Other Settings
1. Try toggling other non-restart-required settings multiple times
2. Verify that unrelated settings in the same nested object remain stable
3. **Expected:** No settings should be unexpectedly reset

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |

## Linked issues / bugs

Fixes #10864


lifefloating:
https://github.com/user-attachments/assets/d7c424c4-2272-436d-9062-26821906eb79

lifefloating:
@jacob314

lifefloating:
@galz10

galdawave (approved):

packages/cli/src/ui/components/SettingsDialog.test.tsx
```diff
@@ -393,11 +393,11 @@ describe('SettingsDialog', () => {
```
jacob314, line 489: when fixing a regression the expectations should get more strict. The trouble with this test file is it would have passed with the bug. We should add a test case that failed without the fix and passes with the fix? that way we can have some confidence we fixed the issue and it won't regress next time someone changes SettingsDialog.

jacob314 (commented):
Approved after this has test cases that demonstrate the cases that are fixed and now pass but failed before.

packages/cli/src/ui/components/SettingsDialog.test.tsx
```diff
@@ -928,6 +928,224 @@ describe('SettingsDialog', () => {
     });
   });
 
+  describe('Race Condition Regression Tests', () => {
+    it('should not reset sibling settings when toggling a nested setting multiple times', async () => {
+      vi.mocked(saveModifiedSettings).mockClear();
+
+      const FAKE_SCHEMA: SettingsSchemaType = {
```
jacob314: nit: move the fake schema outside of the test case to the top of the file  as it is duplicated in these two tests.

jacob314 (approved):
lgtm
Please reduce duplication between the test cases as a quick followup.

lifefloating:
updated
@jacob314

PR #10892, opened 2025-10-10
fix(core): ensure flash fallback on resource exhausted

0xc05m1c0wl:
Fixes #10881


0xc05m1c0wl:
CLA signed and feedback addressed. Thanks!

PR #10897, opened 2025-10-10
feat(telemetry): ensure all telemetry includes user email and installation id

jerop:
## TLDR

This PR enriches the OpenTelemetry (OTel) data that users can collect for their own monitoring and analysis. It introduces an `installation.id` to all telemetry signals and ensures `user.email` is consistently available across both logs and metrics.

## Dive Deeper

The goal of this change is to empower users to gain deeper insights into their own usage patterns, costs, and performance. By making the telemetry data more granular, users can better analyze their interactions with the Gemini CLI when directing the OTel data to their preferred backend (e.g., Google Cloud, Jaeger, or a local file).

Key changes include:
- A new `getCommonAttributes` function was created to consistently apply `session.id`, `installation.id`, and `user.email` to all user-collectable telemetry.
- The `user.email` attribute, previously only in logs, is now also included in metrics. This attribute is only populated when using the code assist auth method.
- A unique `installation.id` is now included in all telemetry, allowing users to distinguish between different CLI installations in their monitoring dashboards.
- The telemetry documentation in `docs/cli/telemetry.md` has been updated to reflect the availability of these new attributes.
- Tests for loggers and metrics have been updated to account for the new attributes.

## Reviewer Test Plan

1.  Pull down the branch.
2.  Run the full preflight check to ensure all tests, linting, and builds pass:
    ```bash
    npm run preflight
    ```
3.  Configure local telemetry to output to a file by setting up your `.gemini/settings.json`:
    ```json
    {
      "telemetry": {
        "enabled": true,
        "target": "local",
        "outfile": "./telemetry.log"
      }
    }
    ```
4.  Run a few Gemini CLI commands.
5.  Inspect the `telemetry.log` file and verify that the log and metric entries contain the `installation.id` and `user.email` (if authenticated) attributes.

## Linked issues / bugs

Fixes #10893

### DIFF
--- a/docs/cli/telemetry.md
+++ b/docs/cli/telemetry.md
@@ -204,7 +204,8 @@ For local development and debugging, you can capture telemetry data locally:
 The following section describes the structure of logs and metrics generated for
 Gemini CLI.
 
-- A `sessionId` is included as a common attribute on all logs and metrics.
+The `session.id`, `installation.id`, and `user.email` are included as common
+attributes on all logs and metrics.
 
 ### Logs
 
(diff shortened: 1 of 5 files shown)


bobcatfish (approved):
niiiice makes so much sense to have these attributes consistent, and gives us the visibility users are looking for to be able to correlate metrics with their own individual users vs getting one big blob of data 👍 👍 👍

PR #10910, opened 2025-10-10
Pass whole extensions rather than just context files

zbirkenbuel:
## TLDR

Rather than parsing out extension context files and passing them around pass the whole extensions around (which are flagged as active or not).  This allows for just-in-time evaluation of activation for tools and user memory

## Dive Deeper

This ensures that all concepts of enablement are always handled once and by the ExtensionEnablementManager (removing the old annotateActiveExtensions). It also paves the way for just-in-time evaluation of extension context file inclusion of active extensions rather than having to replumb all of the files through activation scenarios.

For tools added via extension, only discover them if they are marked as from an active extension and dynamically fetch the list of extensions from the config object when invoking the discovery method. This should allow us to manipulate the extension collection on config and accurately refresh the list of available tools.

Updated interactive display of extensions to list all extensions based on their current activation status (previously it had support for extension activation as a label but would always only ever show "active" since all inactive extensions were filtered out.

## Reviewer Test Plan

* Updated all tests with new pass values
* Cleaned up obsolete tests (which duplicated test coverage of ExtensionEnablementManager)
* Manual testing
  * Verified update extension listing in interactive mode
  * Verified model was unaware of tools and context from inactive extensions (created local "favorite flavor/color/day/number" extensions to toggle active and inactive)

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


packages/cli/src/config/extension.ts
```diff
@@ -166,14 +166,18 @@ export function loadExtensions(
...
-      !uniqueExtensions.has(extension.name) &&
-      extensionEnablementManager.isEnabled(extension.name, workspaceDir)
-    ) {
+    if (!uniqueExtensions.has(extension.name)) {
       uniqueExtensions.set(extension.name, extension);
     }
   }
 
+  for (const extension of uniqueExtensions.values()) {
+    extension.isActive = extensionEnablementManager.isEnabled(
```
jakemac53: In general this codebase avoids mutating live objects and instead opts for copying them into new objects with updated fields etc. Given the CLI is react based anything else can lead to stale UIs etc.

In this case, can we plumb through the extensionEnablementManager to `loadUserExtensions -> loadExtensionsFromDir -> loadExtension ` and set the `isActive` bit there.

This should also be less error prone as we won't just assume active and then potentially flip the bit (but some other code path may not know to do that extra step). Better to handle it all in one canonical place.

zbirkenbuel: We've been discussing offline, there is some refactoring in these methods that will be folded into this PR

jakemac53: I have pushed these changes now

chrstnb (commented):
Why are we now passing the enablement manager in everywhere versus just creating it in the caller? e.g., int he commands? Otherwise looks good

jakemac53:
> Why are we now passing the enablement manager in everywhere versus just creating it in the caller? e.g., int he commands? Otherwise looks good

It needs some information that is only available on the `config` object - specifically the list of explicitly enabled extensions (`-e` options). We could thread _that_ through but its better I think to have a single manager and thread that along (in case it needs more context later, we won't have to keep adding more things).

Also, it will allow us to be more efficient and cache some information in the manager in the future.

I would actually love to put the manager itself on `Config`, but that would require moving it into core and... yeah.

jakemac53 (approved):

PR #10912, opened 2025-10-10
fix(tests): enable cyclic schema MCP tool test

### LINKED ISSUE #10735: Integration tests: Fix "mcp tool list should include tool with cyclic tool schema"
Was disabled due to flakiness

SandyTao520:
## TLDR

Enables a previously skipped integration test for MCP tools with cyclic schemas by switching to interactive test execution. Also removes a test that was not testing the correct behavior.

## Dive Deeper

1.  The test `mcp tool list should include tool with cyclic tool schema` was skipped. This PR updates the test to use `rig.runInteractive()` to correctly handle the output and verify that the tool is listed.
2.  Removed the test `gemini api call should be successful with cyclic mcp tool schema`. This test was not testing anything meaningful and represented incorrect behavior. The expected behavior when a tool has a cyclic schema is for the operation to fail, not succeed.

## Reviewer Test Plan

Run the integration tests:
```bash
npm run test:integration
```
Verify that `integration-tests/mcp_server_cyclic_schema.test.ts` passes.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ✅  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Resolves #10735


integration-tests/mcp_server_cyclic_schema.test.ts
```diff
@@ -192,15 +192,17 @@ describe('mcp server with cyclic tool schema is detected', () => {
...
-    expect(tool_list_output).toContain('tool_with_cyclic_schema');
-  });
+  it('mcp tool list should include tool with cyclic tool schema', async () => {
+    const { ptyProcess } = rig.runInteractive();
+    try {
+      await rig.ensureReadyForInput(ptyProcess);
+
+      await type(ptyProcess, '/mcp list\r');
 
-  it('gemini api call should be successful with cyclic mcp tool schema', async () => {
```
jakemac53, line 201: Should we be re-evaluating whether this check should have been removed then?

I think a follow up issue should probably be filed, even if low-ish priority, to do something like automatically disable the tools that cause these failures.

jakemac53 (approved):
LGTM but I do think we should file a follow-up, there was some logic that was previously removed I believe under the assumption that Gemini now accepted these types of schemas, but the test that was written to validate that behavior appears to have been incorrect, so we should re-evaluate.

integration-tests/mcp_server_cyclic_schema.test.ts
SandyTao520: Gemini API doesn't accept this type of tool definitions, but now we throws this information to users so they can update it if necessary.

PR #10926, opened 2025-10-10
Fix rough edges around extension updates

jakemac53:
## TLDR

- Fixes a case where exiting Gemini CLI during extension consent would leave the extension uninstalled
- Fixes a case where all extension enablement configuration was lost on update
- Adds an "update" event for extensions and logs that instead of an uninstall followed by an install.

## Dive Deeper

Previously, we would do a very raw uninstall followed by a fresh install in order to do updates. This moves the update logic inside `installExtension`, renaming it `installOrUpdateExtension`, so that we can wait to uninstall the old extension until after the consent has been received.  This significantly lowers the chance of leaving extensions in an uninstalled state (although it is still possible for the program to be killed in the middle, the chance is much lower).

This also allows us to not unconditionally enable/disable the extension - we only do that on a true uninstall or fresh install (unrelated to an update). We also fire a single update event now and skip the other events, in the update case.

Added some tests around extension enablement and disablement and logging.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓ | ❓  | ✅   |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes https://github.com/google-gemini/gemini-cli/issues/10899
Fixes https://github.com/google-gemini/gemini-cli/issues/10908

### DIFF
--- a/packages/cli/src/config/extension.test.ts
+++ b/packages/cli/src/config/extension.test.ts
@@ -73,6 +73,7 @@ vi.mock('./trustedFolders.js', async (importOriginal) => {
 const mockLogExtensionEnable = vi.hoisted(() => vi.fn());
 const mockLogExtensionInstallEvent = vi.hoisted(() => vi.fn());
 const mockLogExtensionUninstall = vi.hoisted(() => vi.fn());
+const mockLogExtensionUpdateEvent = vi.hoisted(() => vi.fn());
 const mockLogExtensionDisable = vi.hoisted(() => vi.fn());
 vi.mock('@google/gemini-cli-core', async (importOriginal) => {
   const actual =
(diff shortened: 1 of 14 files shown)


packages/cli/src/config/extension.test.ts
```diff
@@ -851,20 +853,77 @@ describe('extension tests', () => {
...
+            isUpdate
+              ? {
+                  name: 'my-local-extension',
+                  version: '1.0.0',
+                }
+              : undefined,
+          );
+
+          if (isUpdate) {
+            expect(mockLogExtensionUpdateEvent).toHaveBeenCalled();
```
chrstnb, line 891: yay

chrstnb (approved):

PR #10944, opened 2025-10-11
Fix --allowed-tools in non-interactive mode to do substring matching for parity with interactive mode.

mistergarrison:
## TLDR

--allowed-tools currently behaves differently in non-interactive mode compared to interactive mode. Interactive mode will do prefix matching (to allow, for example, "ShellTool(git diff)"), but non-interactive was doing exact matching on the root command.

This change brings their behavior into alignment, and removes duplication with `doesToolInvocationMatch` (which is how the permissions check will later be made).

## Reviewer Test Plan

```sh
# Should print something like "There are 55 lines in /proc/meminfo"
npx github:mistergarrison/gemini-cli#fix/non-interactive-shell-command-prefix-matching \
  --allowed-tools="ShellTool(wc)" \
  --prompt="use wc to tell me how many lines there are in /proc/meminfo"

# Should print something like "There are 55 lines in /proc/meminfo"
npx github:mistergarrison/gemini-cli#fix/non-interactive-shell-command-prefix-matching \
  --allowed-tools="ShellTool(wc -l)" \
  --prompt="use wc to tell me how many lines there are in /proc/meminfo"

# Should fail
npx github:mistergarrison/gemini-cli#fix/non-interactive-shell-command-prefix-matching \
  --allowed-tools="ShellTool(wc -c)" \
  --prompt="use wc to tell me how many lines there are in /proc/meminfo"

# Should fail
npx github:mistergarrison/gemini-cli#fix/non-interactive-shell-command-prefix-matching \
  --allowed-tools="ShellTool(ls)" \
  --prompt="use wc to tell me how many lines there are in /proc/meminfo"

# Should fail
npx github:mistergarrison/gemini-cli#fix/non-interactive-shell-command-prefix-matching \
  --allowed-tools="ls" \
  --prompt="use wc to tell me how many lines there are in /proc/meminfo"

# Should fail
npx github:mistergarrison/gemini-cli#fix/non-interactive-shell-command-prefix-matching \
  --prompt="use wc to tell me how many lines there are in /proc/meminfo"
```

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  |  -  |  -  |  X  |
| npx      |  -  |  -  | X  |

## Linked issues / bugs

- Fixes #10930


sniemczyk-google:
@allenhutchison @abhipatel12 how can we expedite getting this fix reviewed and added to the merge queue? Our team hopes to test these changes as part of improving headless use.

mistergarrison:
So, the Windows tests were failing - but I think it was a fluke that the PowerShell E2E tests ever worked. `(Get-Content ...).Length` will not work with `--allowed-tools="ShellTool(Get-Content)"` since the command starts with a '('. I've tested this against main on a windows machine and confirmed that it doesn't currently work against main, so I've now updated the tests to invoke 'find' from powershell instead of using Get-Content (the same as cmd).

allenhutchison (approved):
This looks great, and an important fix. Thank you!

PR #11013, opened 2025-10-13
Removed retries in the event of quota errors

### LINKED ISSUE #10925: When user quota is used up, queries should fail quickly
### What happened?

While trying to repro #10881 I did get into a state where my quota was exhausted, but requests would hang for 3 min + before I got the error:
 
```
✕ [API Error: [{
    "error": {
      "code": 429,
      "message": "Quota exceeded for quota metric 'Gemini 2.5 Pro Requests' and limit 'Gemini 2.5 Pro Requests per
   day per user per tier' of service 'cloudcode-pa.googleapis.com' for consumer 'project_number:681255809395'.",
      "errors": [
        {
          "message": "Quota exceeded for quota metric 'Gemini 2.5 Pro Requests' and limit 'Gemini 2.5 Pro Requests
   per day per user per tier' of service 'cloudcode-pa.googleapis.com' for consumer 
  'project_number:681255809395'.",
          "domain": "global",
          "reason": "rateLimitExceeded"
        }
      ],
      "status": "RESOURCE_EXHAUSTED",
      "details": [
        {
          "@type": "type.googleapis.com/google.rpc.ErrorInfo",
          "reason": "RATE_LIMIT_EXCEEDED",
          "domain": "googleapis.com",
          "metadata": {
            "quota_metric": "cloudcode-pa.googleapis.com/gemini_2_5_pro_requests",
            "quota_limit": "Gemini25ProRequestsPerDay"
          }
        },
        {
          "@type": "type.googleapis.com/google.rpc.Help",
          "links": [
            {
              "description": "Request a higher quota limit.",
              "url": "https://cloud.google.com/docs/quotas/help/request_increase"
            }
          ]
        }
      ]
    }
  }
  ]]
  Possible quota limitations in place or slow response times detected. Switching to the gemini-2.5-flash model for
   the rest of this session.
```

If I've used up my quota, there is no reason to retry and it should just fail immediately

### Client information

<details>

```console
| CLI Version             0.10.0-nightly.20251007.c195a9aa     #built from head 
│ Git Commit              a5e47c62
│ Model                   gemini-2.5-pro
│ Sandbox                 no sandbox  
│ OS                      linux       
│ Auth Method             OAuth                                  
```

</details>

### Login information

gmail account

A couple of weeks ago in #9763 we added logic to retry 429s, which I was very excited about!! But since then @bdmorgan explained to me that 429 is used for several different cases. In the case where 429 means the capacity of the underlying resource is exhausted, retrying makes sense. But in the case where the user has run out of quota, retrying doesn't help at all and just delays showing them what is wrong.

While reproducing the issue I tested this theory by tweaking these values and was able to get the error immediately:
https://github.com/google-gemini/gemini-cli/blob/ab3804d8231850afe6ea92e787a36b1bd8e064b6/packages/core/src/utils/retry.ts#L33-L38

bdmorgan:
See issue report in #10925. This should change things to now check specifically for a quota error which is a non-retryable error.


packages/core/src/utils/googleQuotaErrors.ts
```diff
@@ -92,6 +92,17 @@ export function classifyGoogleError(error: unknown): unknown {
       d['@type'] === 'type.googleapis.com/google.rpc.RetryInfo',
   );
 
+  // Handle cases where the error indicates a quota exceeded error.
+  if (
+    errorInfo?.reason === 'RESOURCE_EXHAUSTED' &&
+    googleApiError.message.includes('Quota exceeded')
```
richieforeman, line 98: Matching on a message has a bit of code smell.  Is there a better field to match on that is consistent?

If so, let's use that.

If not, I trust your judgement.

packages/core/src/utils/googleQuotaErrors.ts
richieforeman, line 1: Please add a simple test in `packages/core/src/utils/googleQuotaErrors.test.ts`

packages/core/src/utils/googleQuotaErrors.ts
bdmorgan: Unfortunately not, based on all I've learned (which still might be incomplete knowledge!) The single API returns that error for various reasons - some "retryable" (i.e. contention, capacity, availability) some "non-retryable" (i.e. you're reached your quota limit). At this time, there's no way that I know of to disambiguate those other than via the error string. I believe this is being fixed in a future API release, correct @gsquared94 ?

richieforeman (approved):

packages/core/src/utils/googleQuotaErrors.ts
```diff
@@ -92,6 +92,17 @@ export function classifyGoogleError(error: unknown): unknown {
       d['@type'] === 'type.googleapis.com/google.rpc.RetryInfo',
   );
 
+  // Handle cases where the error indicates a quota exceeded error.
+  if (
+    errorInfo?.reason === 'RESOURCE_EXHAUSTED' &&
```
gsquared94, line 97: The API can return `RESOURCE_EXHAUSTED` for intermittent quota exhaustion like a per second or per minute limit. That isn't a TerminalQuotaError

gsquared94 (changes requested):

PR #11050, opened 2025-10-13
fix integration test static errors, and run_shell_command tests to actually be testing what they intend

jakemac53:
## TLDR

Cleans up all static errors in the integration_tests directory, especially cleaning up the very real problems with the run_shell_command tests.

Also added an option to disable yolo mode in `rig.run`.

## Dive Deeper

We should separately consider dropping the default yolo mode behavior of `rig.run` as I think this is harmful as evidenced by these test issues.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅   | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes #11043


integration-tests/run_shell_command.test.ts
```diff
@@ -112,6 +115,13 @@ describe('run_shell_command', () => {
       foundToolCall,
       'Expected to find a run_shell_command tool call',
     ).toBeTruthy();
+
+    const toolCall = rig
+      .readToolLogs()
+      .filter(
+        (toolCall) => toolCall.toolRequest.name === 'run_shell_command',
+      )[0];
+    expect(toolCall.toolRequest.success).toBe(true);
```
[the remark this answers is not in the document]
jakemac53: We already checked that one exists above

integration-tests/context-compress-interactive.test.ts
```diff
@@ -24,7 +24,7 @@ describe('Interactive Mode', () => {
     const run = await rig.runInteractive();
 
     const longPrompt =
-      'Dont do anything except returning a 1000 token long paragragh with the <name of the scientist who discovered theory of relativity> at the end to indicate end of response. This is a moderately long sentence.';
+      "Don't do anything except returning a 1000 token long paragraph with the <name of the scientist who discovered theory of relativity> at the end to indicate end of response. This is a moderately long sentence.";
```
jakemac53: Somehow this change seems to break this test..... 😱

jakemac53:
Ok, so the test that tries to validate running two different shell commands is just super flaky, which is now exposed by the test actually doing something.

My plan is to mark that as skipped for now (but land the fix), and file a separate issue for changing it to something that can run without flakes.

The issue specifically is it really wants to pipe the output of `ls` directly into `xargs` and `wc` it seems, but that gets blocked.

integration-tests/write_file.test.ts
```diff
@@ -28,13 +28,14 @@ describe('write_file', () => {
...
-    expect(foundToolCall, 'Expected to find a write_file tool call').toBeTruthy(
+    expect(
+      foundToolCall,
       createToolCallErrorMessage(
         'write_file',
         allTools.map((t) => t.toolRequest.name),
         result,
       ),
-    );
+    ).toBeTruthy();
```
jacob314, line 38: WAT :O

jacob314 (approved):
lgtm

PR #11055, opened 2025-10-13
fix(infra) - Make file system interactive test check only tool call

### LINKED ISSUE #10643: file-system-interactive.test.ts Integration test is falky
❯ file-system-interactive.test.ts (1 test | 1 failed) 81143ms
   × Interactive file system > should perform a read-then-write sequence 81141ms (retry x2)
     → CLI did not start up in interactive mode correctly: expected false to be true // [Object.is](http://object.is/) equality
     → Expected to find a write_file or replace tool call: expected false to be true // [Object.is](http://object.is/) equality
     → Expected to see version "1.0.0" in output: expected false to be true // [Object.is](http://object.is/) equality
⠸ Compressing chat history

Test run
https://github.com/google-gemini/gemini-cli/actions/runs/18316642177/job/52162404866?pr=10586

shishu314:
## TLDR

Make file system interactive test only check for tool calls

## Dive Deeper

I think we should not be testing whether or not the tool calls are working as intended. They wildly vary in the responses and how they handle the changes, causing the flakes in the tests. Tool call checks should suffice but if you disagree let me know.

## Reviewer Test Plan

[Deflake run](https://github.com/google-gemini/gemini-cli/actions/runs/18575161883/job/52958664450?pr=11055)

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes #10643


.github/workflows/e2e.yml
shishu314, line 1: REMOVE AFTER REVIEW

package.json
shishu314, line 1: REMOVE AFTER REVIEW

integration-tests/file-system-interactive.test.ts
```diff
@@ -33,16 +33,16 @@ describe.skip('Interactive file system', () => {
...
 
-    await rig.expectToolCallSuccess(['write_file', 'replace'], 30000);
-
-    const newFileContent = rig.readFile(fileName);
-    expect(newFileContent).toBe('1.0.1');
+    // Check tool calls made with right args
+    await rig.expectToolCallSuccess(
+      ['write_file', 'replace'],
+      30000,
+      (args) => args.includes('1.0.1') && args.includes(fileName),
```
shishu314, line 45: We might be able to do `JSON.stringify` here to check for actual property values (absolute_path) and content BUT both tool calls have different argument types that would make this condition a lot more difficult. Also args can be a very big object.

richieforeman (approved):
Approved , but remove the action changes.

scidomino (approved):

bdmorgan (approved):

PR #11066, opened 2025-10-13
fix(core): add retry logic for specific fetch errors

SandyTao520:
## TLDR

Adds configuration and logic to retry API calls when a specific "fetch failed" error occurs, addressing transient network issues.

## Dive Deeper

Introduces a `--retry-fetch-errors` CLI flag (and corresponding config) that, when enabled, allows `geminiChat` to retry requests that fail with "exception TypeError: fetch failed sending request". This is handled within the existing `retryWithBackoff` utility in `@google/gemini-cli-core`.

## Reviewer Test Plan

Run the CLI with `--retry-fetch-errors` in an environment with unstable network connectivity or mock network failures to verify retries occur (disconnect network or set GOOGLE_GEMINI_BASE_URL=http://localhost:9999).

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓ |
| npx      | ✅  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


packages/core/src/core/geminiChat.test.ts
```diff
@@ -1115,6 +1116,78 @@ describe('GeminiChat', () => {
...
+        // Should have successful content
+        expect(
+          events.some(
+            (e) =>
+              e.type === StreamEventType.CHUNK &&
+              e.value.candidates?.[0]?.content?.parts?.[0]?.text ===
+                'Success after fetch error',
+          ),
+        ).toBe(true);
+      });
```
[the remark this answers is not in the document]
SandyTao520: this would be too slow

packages/cli/src/config/config.ts
```diff
@@ -98,6 +99,12 @@ export async function parseArguments(settings: Settings): Promise<CliArgs> {
     .usage(
       'Usage: gemini [options] [command]\n\nGemini CLI - Launch an interactive CLI, use -p/--prompt for non-interactive mode',
     )
+    .option('retry-fetch-errors', {
```
jacob314: Rather than a flag can we add a setting? Would prefer to avoid adding flags where possible.

packages/core/src/core/geminiChat.test.ts
```diff
@@ -1113,6 +1114,78 @@ describe('GeminiChat', () => {
...
+          { message: 'test' },
+          'prompt-id-fetch-error-retry',
+        );
+
+        const events: StreamEvent[] = [];
+        for await (const event of stream) {
+          events.push(event);
+        }
+
+        // 5. Assert
```
jacob314: nit. remove 1, 2, 3, 4, 5 comments.

packages/core/src/utils/retry.test.ts
```diff
@@ -304,6 +304,41 @@ describe('retryWithBackoff', () => {
     });
   });
 
+  describe('Fetch error retries', () => {
+    const fetchErrorMsg = 'exception TypeError: fetch failed sending request';
```
jacob314, line 308: Can we specify the fetch error with a Type rather than a String?

packages/core/src/utils/retry.ts
```diff
@@ -41,9 +45,21 @@ const DEFAULT_RETRY_OPTIONS: RetryOptions = {
...
  * @returns True if the error is a transient error, false otherwise.
  */
-function defaultShouldRetry(error: Error | unknown): boolean {
+function defaultShouldRetry(
+  error: Error | unknown,
+  retryFetchErrors?: boolean,
+): boolean {
+  if (
+    retryFetchErrors &&
+    error instanceof Error &&
```
jacob314, line 57: nit: is there  specific type of Error object we can check rather than purely string matching. we have been bit in the past by matching error strings.

jacob314 (commented):

packages/core/src/utils/retry.test.ts
SandyTao520: The SDK throws away error details: https://github.com/googleapis/js-genai/blob/58133cdbc2971c09c5ba7358da82b18b25e160eb/src/_api_client.ts#L600

packages/core/src/utils/retry.ts
SandyTao520: see above.

packages/core/src/utils/retry.test.ts
SandyTao520: https://github.com/google-gemini/gemini-cli/issues/11083

SandyTao520:
I've addressed the comments by moving the retry-fetch-errors from a flag to a setting, and updating the test to use the actual retryWithBackoff implementation.

jacob314 (approved):
lgtm

PR #11067, opened 2025-10-13
Integration test for UI flickers

### LINKED ISSUE #9128: Release Confidence: Flicker Test
Integration tests to capture when the UI flickers

skeshive:
Add integration test for catching flickers

## Linked Issues
Fixes #9128


integration-tests/test-helper.ts
```diff
@@ -890,7 +890,7 @@ export class TestRig {
     const options: pty.IPtyForkOptions = {
       name: 'xterm-color',
       cols: 80,
-      rows: 30,
+      rows: 24,
```
skeshive, line 893: @jacob314 any opposition to reducing to 24? that seems to be the standard height

integration-tests/flicker.test.ts
```diff
@@ -0,0 +1,30 @@
...
+    await run.type(prompt);
+    await run.type('\r');
+
+    const hasUserPromptEvent = await rig.waitForTelemetryEvent('user_prompt');
+    expect(hasUserPromptEvent).toBe(true);
+
+    const sessionCountMetric = rig.readMetric('session.count');
+    expect(sessionCountMetric).not.toBeNull();
+
+    // We expect NO flicker event to be found.
```
skeshive, line 26: I confirmed setting the rows to 10 in IPtyForkOptions below causes the test to fail as expected buttt the test still passes if I add a bunch of new lines to the footer... any idea why? @jacob314

integration-tests/test-helper.ts
jacob314: 24 is fine. just update any other tests that might be broken as a result.

integration-tests/flicker.test.ts
jacob314: i'd suggest just changing the rows to 10 for the flicker test.

integration-tests/test-helper.ts
skeshive: SG

integration-tests/flicker.test.ts
skeshive: NVM my changes weren't being picked up by the integration tests when I ran `npm run build`. I needed to run `npm run bundle`

jacob314 (approved):
lgtm

PR #11073, opened 2025-10-13
Shell approval rework

cornmander:
continuation of https://github.com/google-gemini/gemini-cli/pull/10498

Reworks shell approval to do a full parse, which allows us to evaluate nested subcommands. Also uses powershell to parse powershell, which improves the fidelity of parsing on windows.

Adds tests to confirm the new behavior. I've also tested manually by asking the agent to constructed nested substitutions and chained commands.


scidomino:
I tested this and it worked great. It even got things like `Write-Output "The current date is $(Get-Date)"; Get-Process -Name "powershell" `

scidomino (approved):

packages/core/src/tools/shell.test.ts
```diff
@@ -84,17 +93,25 @@ describe('ShellTool', () => {
...
       (mockConfig.getExcludeTools as Mock).mockReturnValue(undefined);
       expect(isCommandAllowed('ls -l', mockConfig).allowed).toBe(true);
     });
 
-    it('should block a command with command substitution using $()', () => {
-      expect(isCommandAllowed('echo $(rm -rf /)', mockConfig).allowed).toBe(
-        false,
-      );
+    it('should allow a command with command substitution using $()', () => {
+      const evaluation = isCommandAllowed('echo $(rm -rf /)', mockConfig);
```
jacob314: why is this test now verifying that something that looks quite unsafe is allowed?

packages/core/src/utils/shell-utils.ts
```diff
@@ -32,6 +39,305 @@ export interface ShellConfiguration {
   shell: ShellType;
 }
 
+const requireModule = createModuleRequire(import.meta.url);
+
+let bashLanguage: Language | null = null;
+let treeSitterInitialization: Promise<void> | null = null;
+
+async function loadBashLanguage(): Promise<void> {
+  try {
```
jacob314, line 48: nit: I think there was a second long freeze in the UI the first time this was loaded. Is it feasible to make this a little less blocking? if not it is fine. after the first install it is fine but the first freeze is slightly jarring.

jacob314 (approved):
lgtm

packages/core/src/utils/shell-utils.ts
```diff
@@ -32,6 +39,305 @@ export interface ShellConfiguration {
...
+}
+
+interface CommandParseResult {
+  details: ParsedCommandDetail[];
+  hasError: boolean;
+}
+
+const POWERSHELL_COMMAND_ENV = '__GCLI_POWERSHELL_COMMAND__';
+
+const POWERSHELL_PARSER_SCRIPT = Buffer.from(
```
jacob314, line 95: For future maintainability, a comment explaining what the script does and why it's base64-encoded would be beneficial.

cornmander: Done.

packages/core/src/utils/shell-utils.ts
cornmander: I tried changing this to be lazy loaded but it still causes a pause. I'm going to leave as-is for now and investigate in a followup issue.

packages/core/src/tools/shell.test.ts
cornmander: I updated these to use fake command names where the semantic importance is in the command name itself.

cornmander:
I tested this on Node 20, 22, and 24 on Linux.

PR #11087, opened 2025-10-14
fix(e2e): Use rmSync instead of rm -rf for e2e tests

abhipatel12:
## TLDR

Use rmSync instead of rm -rf for e2e tests. This is to ensure cross-platform compatibility


hminooei (approved):

abhipatel12:
Thanks!

PR #11094, opened 2025-10-14
feat(cli): Prevent queuing of slash and shell commands

MJjainam:
This change introduces a mechanism to block slash and shell commands from being queued if a command is already being processed.

When a user attempts to submit a command (slash or shell) while cli is in a streaming response state, the command is now rejected, and a temporary error message is displayed in the notification area.

UX improvement: https://github.com/user-attachments/assets/2ff2b5d5-0184-49c1-903c-cf4cb8e20673

## Linked issues / bugs
- Fixes #9010

### DIFF
--- a/packages/cli/src/ui/components/InputPrompt.tsx
+++ b/packages/cli/src/ui/components/InputPrompt.tsx
@@ -38,6 +38,8 @@ import * as path from 'node:path';
 import { SCREEN_READER_USER_PREFIX } from '../textConstants.js';
 import { useShellFocusState } from '../contexts/ShellFocusContext.js';
 import { useUIState } from '../contexts/UIStateContext.js';
+import { StreamingState } from '../types.js';
+import { isSlashCommand } from '../utils/commandUtils.js';
 
 /**
  * Returns if the terminal can be trusted to handle paste events atomically
@@ -71,6 +73,8 @@ export interface InputPromptProps {
   onEscapePromptChange?: (showPrompt: boolean) => void;
   vimHandleInput?: (key: Key) => boolean;
   isEmbeddedShellFocused?: boolean;
+  setQueueErrorMessage: (message: string | null) => void;
+  streamingState: StreamingState;
 }
 
 // The input content, input container, and input suggestions list may have different widths
@@ -107,6 +111,8 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
   onEscapePromptChange,
   vimHandleInput,
   isEmbeddedShellFocused,
+  setQueueErrorMessage,
+  streamingState,
 }) => {
   const kittyProtocol = useKittyKeyboardProtocol();
   const isShellFocused = useShellFocusState();
@@ -218,6 +224,31 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
       shellModeActive,
       shellHistory,
       resetReverseSearchCompletionState,
+    ],
+  );
+
+  const handleSubmit = useCallback(
+    (submittedValue: string) => {
+      const trimmedMessage = submittedValue.trim();
+      const isSlash = isSlashCommand(trimmedMessage);
+
+      const isShell = shellModeActive;
+      if (
+        (isSlash || isShell) &&
+        streamingState === StreamingState.Responding
+      ) {
+        setQueueErrorMessage(
+          `${isShell ? 'Shell' : 'Slash'} commands cannot be queued`,
+        );
+        return;
+      }
+      handleSubmitAndClear(trimmedMessage);
+    },
+    [
+      handleSubmitAndClear,
+      shellModeActive,
+      streamingState,
+      setQueueErrorMessage,
     ],
   );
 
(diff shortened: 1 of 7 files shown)


MJjainam:
UX improvement: https://github.com/user-attachments/assets/73be4eea-8e00-4150-a859-7bd328f41aec

jacob314:
For PRs like this including a screenshot of the UI is convenient even though I will run the PR locally as part of review

packages/cli/src/ui/components/Notifications.tsx
```diff
@@ -74,6 +76,16 @@ export const Notifications = () => {
           </Text>
         </Box>
       )}
+      {queueErrorMessage && (
```
jacob314: showing this as a box is ugly and confusing as the box looks like it is part of the message hist Please align with how we show the message that if you press ctrl-C again to exit.

jacob314 (commented):
The user shouldn't be allowed to submit the prompt when it cannot be submitted. Block submission in InputPrompt instead of showing a message after the fact. The way this is done currently would be frustrating to users as your input prompt got cleared when it should have been left with your shell command so you could submit it once the current turn is over.

jacob314:
Another edge case to consider is that queueing one slash command should be fine but multiple would cause problems.  Given this I would either omit the slash command case from this PR or distinguish the single vs multiple qued slash commands case.

MJjainam:
> For PRs like this including a screenshot of the UI is convenient even though I will run the PR locally as part of review

I have added the video in the PR description.

MJjainam:
> dge case to consider is that queueing one slash command should be fine but multiple would cause problems. Given this I would either omit the slash command case from this PR or distinguish the single vs multiple qued slash commands case.

I did not understand this. 
Why would we want to allow user passing even on slash command?

MJjainam:
@jacob314 Multiple slash commands will not be possible as the enter key will not accept the input (it will just show the warning)

MJjainam:
@jacob314 I have resolve all the comments above. Thanks for the suggestion - it gives a much better UX. 
There is one open item - single slash vs multi slash command point that I did not understand. 

For now the behaviour is slash commands will not be accepted - same like shell commands.

UX video: https://github.com/user-attachments/assets/2ff2b5d5-0184-49c1-903c-cf4cb8e20673

packages/cli/src/ui/components/InputPrompt.tsx
```diff
@@ -120,6 +127,18 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
   >(null);
   const pasteTimeoutRef = useRef<NodeJS.Timeout | null>(null);
 
+  // Effect to clear the user message after a delay
+  useEffect(() => {
+    if (queueErrorMessage) {
+      const timer = setTimeout(() => {
+        setQueueErrorMessage(null);
+      }, 3000); // Clear after 3 seconds
```
jacob314: do the timeout when the message is displayed not here. who knows who else might have sent an error message to displayed while this was in progress.

packages/cli/src/ui/components/InputPrompt.tsx
```diff
@@ -221,6 +240,34 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
     ],
   );
 
+  const handleSubmit = useCallback(
+    (submittedValue: string) => {
+      const trimmedMessage = submittedValue.trim();
+      const isSlashCommand =
+        trimmedMessage.startsWith('/') &&
```
jacob314: use standard slash command processing from elsewhere in the file rather than hacking it here.

packages/cli/src/ui/components/InputPrompt.tsx
```diff
@@ -221,6 +240,34 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
...
+        !trimmedMessage.startsWith('//') &&
+        !trimmedMessage.startsWith('/*');
+
+      const isShellCommand = shellModeActive;
+      if (
+        (isSlashCommand || isShellCommand) &&
+        streamingState === StreamingState.Responding
+      ) {
+        setQueueErrorMessage(
+          `${isShellCommand ? 'shell' : 'slash'} commands cannot be queued`,
```
jacob314: to align with other messages this should be
Shell and Slash not shell and slash.

jacob314 (commented):
Approved after these are addressed

packages/cli/src/ui/components/InputPrompt.tsx
MJjainam: Added const instead of 3000

packages/cli/src/ui/components/InputPrompt.tsx
MJjainam: Done

jacob314 (commented):
Approved once you add a test for the code in AppContainer.

jacob314 (approved):
lgtm

PR #11141, opened 2025-10-14
Add extension tags to manifest

### LINKED ISSUE #11111: Add extension tags, in coordination with list of gallery tags
Add a field to the extension manifest for relevant tags/categories. This should coordinate with the list we land on for the gallery.

chrstnb:
## TLDR

Includes a list of valid tags, which we will add to the gallery for filtering by relevant tag. This isn't totally required to be added to the extensions logic, but we can use it for validation to restrict tags to only the ones we are using in the gallery. This list will have to be kept in-sync, but that seems OK for now.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes #11111


packages/cli/src/config/extension.ts
```diff
@@ -45,26 +45,44 @@ import { ExtensionEnablementManager } from './extensions/extensionEnablement.js'
...
  * GeminiCLIExtension class defined in Core.
  */
-interface ExtensionConfig {
-  name: string;
-  version: string;
+const ExtensionConfigSchema = z.object({
+  name: z.string(),
+  version: z.string(),
+  tags: z.array(z.nativeEnum(ValidTags)).optional(),
+  mcpServers: z.record(z.any()).optional(),
```
richieforeman: Can we avoid z.any(), is there a better type for this?

packages/cli/src/config/extension.ts
```diff
@@ -45,26 +45,44 @@ import { ExtensionEnablementManager } from './extensions/extensionEnablement.js'
...
  * Extension definition as written to disk in gemini-extension.json files.
  * This should *not* be referenced outside of the logic for reading files.
  * If information is required for manipulating extensions (load, unload, update)
  * outside of the loading process that data needs to be stored on the
  * GeminiCLIExtension class defined in Core.
  */
-interface ExtensionConfig {
-  name: string;
-  version: string;
+const ExtensionConfigSchema = z.object({
```
richieforeman: Nit, I've usually seen people define the zod object in `camelCase` and then define a type using `z.infer<>` as `TitleCase`

(e.g.

```
const zodSchema = z.object(...);

type ZodSchema = z.infer<zodSchema>;
```

I know it's a bit pedantic, but having this be consistent as we adopt Zod I think will help readability.

packages/cli/src/config/extension.ts
```diff
@@ -686,15 +704,14 @@ export function loadExtensionConfig(
...
   try {
     const configContent = fs.readFileSync(configFilePath, 'utf-8');
-    const rawConfig = JSON.parse(configContent) as ExtensionConfig;
-    if (!rawConfig.name || !rawConfig.version) {
-      throw new Error(
-        `Invalid configuration in ${configFilePath}: missing ${!rawConfig.name ? '"name"' : '"version"'}`,
-      );
-    }
-    const installDir = new ExtensionStorage(rawConfig.name).getExtensionDir();
+    const rawConfig = JSON.parse(configContent);
```
richieforeman: nit JSON.parse() returns an `any`.  I'd feel a bit better if we cast this to `as unknown`

packages/cli/src/config/extension.ts
```diff
@@ -686,15 +704,14 @@ export function loadExtensionConfig(
...
+    const parsedConfig = ExtensionConfigSchema.parse(rawConfig);
+
+    const installDir = new ExtensionStorage(
+      parsedConfig.name,
+    ).getExtensionDir();
     const config = recursivelyHydrateStrings(
-      rawConfig as unknown as JsonObject,
+      parsedConfig as unknown as JsonObject,
```
richieforeman: I suspect that now that we have Zod, we can get rid of this as unknown business.

Also, I think it'd be great to ditch the `JsonObject` type. It's overly broad and we should prefer the zod-based types everywhere.

packages/cli/src/test-utils/createExtension.ts
```diff
@@ -23,12 +23,13 @@ export function createExtension({
...
   mcpServers = {} as Record<string, MCPServerConfig>,
   installMetadata = undefined as ExtensionInstallMetadata | undefined,
+  tags = [] as string[],
 } = {}): string {
   const extDir = path.join(extensionsDir, name);
   fs.mkdirSync(extDir, { recursive: true });
   fs.writeFileSync(
     path.join(extDir, EXTENSIONS_CONFIG_FILENAME),
-    JSON.stringify({ name, version, contextFileName, mcpServers }),
+    JSON.stringify({ name, version, contextFileName, mcpServers, tags }),
```
richieforeman: It would be great to use zodSchema.encode() here.

packages/cli/src/config/extension.ts
chrstnb: I'm going to make an issue for this-- I want to avoid making type changes to core in this PR, and updating this function would require that.

packages/cli/src/test-utils/createExtension.ts
chrstnb: I believe only available in the latest zod version

PR #11147, opened 2025-10-14
feat(infra) - Add logging for slow rendering

### LINKED ISSUE #10600: Regressions in UX responsiveness
This is the least well defined of these items and should perhaps be a P2.
Baseline would be to track the # of frames that take more than 200ms to render.

P2: 
We might also do some things to track keyboard events that appear to be ignored. The keyboard event CUJs are crl-C + ctrl-C to exit. Any time we detect those events repeated within 500ms of each other it indicates that our code to actually exit based on these events has failed.

shishu314:
## TLDR

Increment counter for slow render metrics when rendering is more than 200ms

## Dive Deeper

I'm unable to mock and test out the function. We have `render` mocked out, and as a result we do not have any access to the internal callback. Even if we did we don't have any ways to mock it out so I am not sure how to test this.

## Reviewer Test Plan

Add a breakpoint into the onRender to see whether or not it's triggered.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | x  | ❓  | ❓  |
| npx      | x  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs
Fixes #10600


packages/cli/src/ui/components/HistoryItemDisplay.tsx
```diff
@@ -54,6 +57,17 @@ export const HistoryItemDisplay: React.FC<HistoryItemDisplayProps> = ({
   embeddedShellFocused,
   availableTerminalHeightGemini,
 }) => {
+  const config = useConfig();
+  const startTime = useRef(Date.now());
+
+  useEffect(() => {
```
jacob314: I'm skeptical this measures what we think it does. Also do not like that it adds an extra useEffect just to try to track this.  This will only update when config updates which is surely not every frame.
The right approach is to send out a pr to ink to add callbacks to properly test this. 
DebugProfile gives some breadcrumbs on ways this could be done without Ink changes.

packages/core/src/telemetry/metrics.ts
```diff
@@ -639,6 +647,14 @@ export function recordExitFail(config: Config): void {
   exitFailCounter.add(1, baseMetricDefinition.getCommonAttributes(config));
 }
 
+/**
+ * Records a metric for when a UI frame is slow in rendering
+ */
+export function recordSlowRender(config: Config): void {
```
jacob314: should we record the time rather than just that it is slow? there is a big difference between a 200ms render and a 5 second render.

packages/cli/src/gemini.test.tsx
```diff
@@ -21,7 +21,31 @@ import {
...
+vi.mock('@google/gemini-cli-core', async (importOriginal) => {
+  const actual =
+    await importOriginal<typeof import('@google/gemini-cli-core')>();
+  return {
+    ...actual,
+    recordSlowRender: vi.fn(),
+  };
+});
+
+vi.mock('ink', async (importOriginal) => {
```
jacob314, line 42: nit: it would be better to not fake out handling of onRenderCallback.
Doing it this way makes the tests not representative of the real behavior of the Ink library. For example, now strangely onRender is basically never called and called at different times then in the real system. instead I would recommend using the vuilt in vi functionality to use fake time extended to mock out datetime.now and performance.now so that the ink code successfully uses the fake times.

jacob314 (approved):
lgtm

PR #11149, opened 2025-10-14
Update extensions install warning

chrstnb:
## TLDR

Also added an option to auto-accept, but logging a warning of what was consented to (in case someone is sharing commands with -y attached).

## Dive Deeper

```
Installing extension "nanobanana".
**The extension you are about to install may have been created by a third-party developer and sourced from a public repository. Google does not vet, endorse, or guarantee the functionality or security of extensions. Please carefully inspect any extension and its source code before installing to understand the permissions it requires and the actions it may perform. Do you wish to proceed with the installation?**
This extension will run the following MCP servers:
  * nanobanana (local): node /Users/chrstn/.gemini/extensions/nanobanana/mcp-server/dist/index.js
This extension will append info to your gemini.md context using GEMINI.md
Do you want to continue? [Y/n]: n
Installation cancelled for "nanobanana".
````
## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes #860

### DIFF
--- a/packages/cli/src/commands/extensions/install.ts
+++ b/packages/cli/src/commands/extensions/install.ts
@@ -6,6 +6,7 @@
 
 import type { CommandModule } from 'yargs';
 import {
+  INSTALL_WARNING_MESSAGE,
   installOrUpdateExtension,
   requestConsentNonInteractive,
 } from '../../config/extension.js';
@@ -18,6 +19,7 @@ interface InstallArgs {
   ref?: string;
   autoUpdate?: boolean;
   allowPreRelease?: boolean;
+  yes?: boolean;
 }
 
 export async function handleInstall(args: InstallArgs) {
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: packages/cli/src/commands/extensions/install.ts [hunk 1]
```diff
@@ -87,6 +96,13 @@ export const installCommand: CommandModule = {
         describe: 'Enable pre-release versions for this extension.',
         type: 'boolean',
       })
+      .option('yes', {
```
jakemac53: Should this maybe be a slightly more specific flag? Or should we change the wording to be more general, that it will auto-consent to anything and everything?

jakemac53 (approved):

packages/cli/src/commands/extensions/install.ts
chrstnb: any suggestions as to what? seems reasonable but I'm not sure what the more specific flag would be

jakemac53: I think the `consent` name you landed is good 👍

PR #11156, opened 2025-10-14
First take at mocking out gemini cli responses in integration tests

jakemac53:
## TLDR

First part of #11062 - adds a `--mock-responses` command line argument to the CLI, which should be the path to a file containing mock responses.

If passed, we create a `MockContentGenerator` which is used for all gemini interactions, responding with the canned responses from the file.

## Dive Deeper

The JSON format has keys for each ContentGenerator method, each of which should be an array of response objects. These are returned sequentially whenever a call to that method is received, until no responses are left, at which point it will throw an error indicating there are no mock responses left.

In the future we may want a way of recording these responses from a real interaction and auto-update the mock responses with those observed responses, which should make maintaining and writing these tests much easier. However, you will note that for this test I spoofed the returned metadata about token counts, so it isn't clear how we would want to merge these concepts.

## Reviewer Test Plan

Run the updated context-compress-interactive test and see that it passes now.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Closes #10769


packages/core/src/core/mockContentGenerator.ts
```diff
@@ -0,0 +1,115 @@
...
+
+  constructor(responses: MockResponses) {
+    this.responses = responses;
+  }
+
+  static async fromFile(filePath: string): Promise<MockContentGenerator> {
+    const fileContent = await promises.readFile(filePath, 'utf-8');
+    const responses = JSON.parse(fileContent) as MockResponses;
+    return new MockContentGenerator(responses);
+  }
```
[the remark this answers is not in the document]
jakemac53: done

packages/core/src/core/fakeContentGenerator.ts
```diff
@@ -0,0 +1,120 @@
...
+    const fileContent = await promises.readFile(filePath, 'utf-8');
+    const responses = JSON.parse(fileContent) as FakeResponses;
+    return new FakeContentGenerator(responses);
+  }
+
+  generateContent(
+    _request: GenerateContentParameters,
+    _userPromptId: string,
+  ): Promise<GenerateContentResponse> {
+    const response =
```
joshualitt: do we want to consolidate the guard? Something like this from Gemini:
```
    1 private getNextResponse<K extends keyof FakeResponses>(
    2   method: K,
    3   request: unknown
    4 ): FakeResponses[K][number] {
    5   const response = this.responses[method][this.callCounters[method]++];
    6   if (!response) {
    7     throw new Error(
    8       `No more mock responses for ${method}, got request:\n` +
    9         safeJsonStringify(request)
   10     );
   11   }
   12   return response;
   13 }
```

packages/core/src/core/fakeContentGenerator.ts
```diff
@@ -0,0 +1,120 @@
+/**
```
joshualitt, line 1: Tests?

packages/core/src/core/contentGenerator.ts
```diff
@@ -105,6 +106,10 @@ export async function createContentGenerator(
   gcConfig: Config,
   sessionId?: string,
 ): Promise<ContentGenerator> {
+  if (gcConfig.fakeResponses) {
```
joshualitt, line 109: trivial test of this branch?

packages/core/src/core/fakeContentGenerator.ts
jakemac53: done

packages/core/src/core/fakeContentGenerator.ts
jakemac53: love it, done

packages/core/src/core/contentGenerator.ts
jakemac53: done, mocked out to hell but probably fine

joshualitt (approved):

PR #11167, opened 2025-10-15
fix(cli): Do not refresh pending settings after setting a setting.

galdawave:
## TLDR

This PR fixes a bug in the settings dialog where resetting a setting to its default value did not update the UI to reflect the change.

## Dive Deeper

When a user modified a setting and then changed the setting bellow it, the UI would reset the previous setting to it's default state. 

This change removes the unnecessary refresh of pending settings after a reset, ensuring that the UI correctly displays the set setting.

## Reviewer Test Plan

1. Run `gemini settings`
2. Modify a setting (e.g., change the theme).
3. Modify a setting below it. 
4. **Verify that the setting's values stayed changed*
5. Save the settings and ensure the value is persisted.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


packages/cli/src/ui/components/SettingsDialog.tsx
```diff
@@ -205,11 +205,6 @@ export function SettingsDialog({
               next.delete(key);
               return next;
             });
-
-            // Refresh pending settings from the saved state
-            setPendingSettings(
-              structuredClone(settings.forScope(selectedScope).settings),
-            );
```
jacob314, line 212: lets add a test so we don't regress this. allowing users to be confident that their settings aren't corrupted is crucial.

PR #11171, opened 2025-10-15
fix(core): Preserve escaped characters in gitignore patterns

### LINKED ISSUE #11168: Backslash-escaped characters in gitignore patterns are incorrectly processed
### What happened?

A path normalization step, intended to convert Windows-style paths, is incorrectly replacing *all* backslash (`\`) characters with forward slashes (`/`).

This corrupts `gitignore` patterns that use a backslash as an escape character. For example, the pattern `\#foo`, which is meant to ignore any file named `#foo`, is erroneously converted. Another example is a significant trailing space is no longer escaped, `\ `.

I expected the path normalization logic to be aware of `gitignore`'s escape syntax. It should preserve backslashes that are used for escaping special characters (e.g., `\#`, `\!`, `\ `).

The transformation should only apply to backslashes used as path separators, ensuring that valid escape sequences are passed to the pattern matching engine unmodified.

### Client information

<details>

```console
> /about
╭───────────────────────────────────────────────────────────────────────────────────────╮
│                                                                                       │
│ About Gemini CLI                                                                      │
│                                                                                       │
│ CLI Version                   0.10.0-nightly.20251007.c195a9aa                        │
│ Model                         gemini-2.5-pro                                          │
│ Sandbox                       no sandbox                                              │
│ OS                            linux                                                   │
│ Auth Method                   gemini-api-key                                          │
│                                                                                       │
╰───────────────────────────────────────────────────────────────────────────────────────╯
```

</details>

### Login information

gemini-api-key

### Relevant Files
- [`packages/core/src/utils/gitIgnoreParser.ts`](tree/main/packages/core/src/utils/gitIgnoreParser.ts)
- [`packages/core/src/utils/gitIgnoreParser.test.ts`](tree/main/packages/core/src/utils/gitIgnoreParser.test.ts)

EricRahm:
The gitignore parser was incorrectly processing patterns containing backslash-escaped characters (e.g., `\#`, `\!`, '\ '). This was caused by an aggressive path normalization step that replaced all backslashes with forward slashes, corrupting the escape sequence before it could be interpreted by the `ignore` library.

This resulted in patterns intended to ignore files with special names (like `#foo`) being misinterpreted, causing those files to be incorrectly processed instead of ignored.

This commit addresses the issue by:
- Removing the incorrect string replacement for path normalization.
- Switching to `path.posix.join` for path manipulation to ensure consistent, cross-platform behavior with forward slashes, as expected by the underlying `ignore` library.

These changes ensure that escaped patterns are passed to the `ignore` library unmodified, restoring the correct gitignore behavior as specified in the gitignore man pages.

## TLDR

The gitignore parser is replacing ignore pattern escape sequences with forward slashes.

## Dive Deeper

The gitignore parser was incorrectly processing patterns containing backslash-escaped characters (e.g., `\#`, `\!`, '\ '). This was caused by an aggressive path normalization step that replaced all backslashes with forward slashes, corrupting the escape sequence before it could be interpreted by the `ignore` library.

This resulted in patterns intended to ignore files with special names (like `#foo`) being misinterpreted, causing those files to be incorrectly processed instead of ignored.

This commit addresses the issue by:
- Removing the incorrect string replacement for path normalization.
- Switching to `path.posix.join` for path manipulation to ensure consistent, cross-platform behavior with forward slashes, as expected by the underlying `ignore` library.

These changes ensure that escaped patterns are passed to the `ignore` library unmodified, restoring the correct gitignore behavior as specified in the gitignore man pages.
## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes #11168


packages/core/src/utils/gitIgnoreParser.ts
```diff
@@ -89,9 +92,6 @@ export class GitIgnoreParser implements GitIgnoreFilter {
           newPattern = '!' + newPattern;
         }
 
-        // Even in windows, Ignore expects forward slashes.
```
jacob314, line 92: have you tested this on windows? need to make sure these changes do not break windows support.

jacob314 (approved):
lgtm

packages/core/src/utils/gitIgnoreParser.ts
EricRahm: Thanks for the quick review, I'm getting a windows dev environment setup and will report back!

EricRahm: image

EricRahm: I now have a windows machine setup and have tested this locally.

I added three files with special names:
- `packages/core/#foo.js`
- `packages/core/src/tools/#foo.js`
- `packages/core/src/utils/#foo.js`

I setup my `.gitignore` to ignore all `#foo.js` (this requires an escape character) by default but explicitly allow two of the files:
```
\#foo.js
!packages/core/#foo.js
!packages/core/src/utils/#foo.js
```

I then asked `gemini-cli` to find all the `#foo.js` files, we see it find the 2 that are allowed, but it does not find the 3rd as expected by the the `.gitignore` configuration.

PR #11182, opened 2025-10-15
feat(core): Add no-op finish tool.

### LINKED ISSUE #10505: Experiment with the use of a `finish` tool to reduce infinite loops on task completion
Gemini CLI occasionally gets confused about task completion, and takes unnecessary actions or goes into an infinite loop. One hypothesis is that our [language](https://github.com/google-gemini/gemini-cli/blob/43b3f79d60c245e9c019f5a2f81042c65601891e/packages/core/src/core/prompts.ts#L142) around task completion is a bit ambiguous, resulting in hallucinations. We could try to make this language more precise, i.e. explaining how the model should await. However, another approach to experiment with is a `finish` tool. A `finish` tool is completely clear, unambiguous, and self-explanatory, and thus the agent will hopefully gravitate towards it naturally.

If this hypothesis is correct, then:
1. Evals shouldn't be negatively impacted
2. We should observe a reduction of instances of infinite loops at the end of task completion

joshualitt:
## TLDR

Adds a simple no-op finish tool behind an experimental setting.

## Reviewer Test Plan

Download patch and attempt tasks.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

A first attempt to resolve https://github.com/google-gemini/gemini-cli/issues/10505

### DIFF
--- a/packages/core/src/tools/finish.ts
+++ b/packages/core/src/tools/finish.ts
@@ -1,0 +1,60 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import type { ToolInvocation, ToolResult } from './tools.js';
+import { BaseDeclarativeTool, BaseToolInvocation, Kind } from './tools.js';
+
+interface FinishToolParams {
+  summary: string;
+}
+
+export class FinishTool extends BaseDeclarativeTool<
+  FinishToolParams,
+  ToolResult
+> {
+  constructor() {
+    super(
+      'finish',
+      'Finish',
+      'Signals that the AI has finished its turn.',
+      Kind.Other,
+      {
+        type: 'object',
+        properties: {
+          summary: {
+            type: 'string',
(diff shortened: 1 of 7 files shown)


joshualitt:
@jacob314 , thoughts on the UI side of this? By default, we get a simple box with the summary from the agent.

packages/core/src/tools/finish.ts
```diff
@@ -0,0 +1,60 @@
...
+
+export class FinishTool extends BaseDeclarativeTool<
+  FinishToolParams,
+  ToolResult
+> {
+  constructor() {
+    super(
+      'finish',
+      'Finish',
+      'Signals that the AI has finished its turn.',
```
joshualitt, line 22: Eval language that is more workflow oriented, i.e. "Signals that the AI has completed the task given to it by the user."

PR #11218, opened 2025-10-15
feat(infra) - Add logging for when user tries to exit multiple times

shishu314:
## TLDR

Add logging when user presses our exit commands multiple times 
## Dive Deeper

We're likely to get some "fake" counts when user might've just pressed really fast. But if a user is doing multiple presses (more than 2) it's more likely that the CLI is unresponsive in exiting.

## Reviewer Test Plan

run `npm run debug` with a breakpoint on the metric logging, press CTRL-C 3 times in succession when exiting.
## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Makes progress to #10600


packages/cli/src/ui/AppContainer.tsx
```diff
@@ -906,24 +891,33 @@ Logging in with Google... Please restart Gemini CLI to continue.
...
     (
-      pressedOnce: boolean,
-      setPressedOnce: (value: boolean) => void,
+      setPressCount: React.Dispatch<React.SetStateAction<number>>,
       timerRef: React.MutableRefObject<NodeJS.Timeout | null>,
     ) => {
-      if (pressedOnce) {
-        if (timerRef.current) {
-          clearTimeout(timerRef.current);
+      setPressCount((pressCount) => {
```
jacob314: having logic embedded inside the callback passed to setPressCount is an anti-pattern. Code there should be state free to avoid subtle timing bugs.

good news is this logic doesn't really need to be inside the callback.
For recording exit fails if pressCount >= 2 you might actually want to have that just in the main body of AppContainer anyway because you don't care if there is keyboard buffering and ctrl-C got pressed 10 times and the app did exit. That code is purely about error logging so would be ok to leave. I would avoid the business logic for clearTimeout. That should not rely on pressedOnce or PressCount >= 1 at all and should instead be sure to reset timerRef.current after clearing the timeout which it currently fails to do.

jacob314 (commented):

packages/cli/src/ui/AppContainer.tsx
```diff
@@ -903,22 +916,26 @@ Logging in with Google... Please restart Gemini CLI to continue.
...
         handleSlashCommand('/quit');
       } else {
-        setPressedOnce(true);
         timerRef.current = setTimeout(() => {
-          setPressedOnce(false);
+          setPressedCount(0);
           timerRef.current = null;
         }, CTRL_EXIT_PROMPT_DURATION_MS);
       }
+      setPressedCount(pressedCount + 1);
```
jacob314: this should still be setPressedCount((count) => count + 1);

packages/cli/src/ui/AppContainer.tsx
```diff
@@ -903,22 +916,26 @@ Logging in with Google... Please restart Gemini CLI to continue.
...
-      if (pressedOnce) {
-        if (timerRef.current) {
-          clearTimeout(timerRef.current);
-        }
+      if (timerRef.current) {
+        clearTimeout(timerRef.current);
+        timerRef.current = null;
+      }
+
+      if (pressedCount >= 1) {
```
jacob314: actually the better place for this would be in the useEffect call above. otherwise you will miss calling if two ctrl-C got pressed within the same react frame.

packages/cli/src/ui/AppContainer.tsx
```diff
@@ -871,6 +872,18 @@ Logging in with Google... Please restart Gemini CLI to continue.
     };
   }, [handleNewMessage]);
 
+  useEffect(() => {
+    if (ctrlCPressCount >= 2) {
```
jacob314: off by 1. this should be > 2, right? It is ok if ctrlC gets pressed twice and we exit after a tiny bit of time. it isn't ok if it gets pressed 3 times without exit.

packages/cli/src/ui/AppContainer.tsx
```diff
@@ -877,6 +878,42 @@ Logging in with Google... Please restart Gemini CLI to continue.
     };
   }, [handleNewMessage]);
 
+  useEffect(() => {
```
jacob314, line 882: this logic is tricky. please add a unittest. edge cases include what happens when ctrlCPressCount exceeds 2.

jacob314 (approved):
Approved but please follow up by adding a unittest for this to AppContainer.

packages/cli/src/ui/AppContainer.test.tsx
```diff
@@ -1108,144 +1113,295 @@ describe('AppContainer State Management', () => {
...
+        );
 
-      expect(mockHandleSlashCommand).not.toHaveBeenCalledWith('/quit');
+        expect(mockCancelOngoingRequest).toHaveBeenCalledTimes(1);
+        expect(mockHandleSlashCommand).not.toHaveBeenCalled();
+      });
 
-      expect(mockHandleSlashCommand).not.toHaveBeenCalledWith('/quit');
+      it('should quit on second press', () => {
+        const { rerender } = render(
```
jacob314: can you make these tests less verbose. you shouldn't have the repeated blocks re-rendering with the same boilerplate?

PR #11221, opened 2025-10-15
fix(core): address GCS path input

jajanet:
## TLDR

After running Google's [Gemini CLI Security Extension](https://github.com/gemini-cli-extensions/security) on this repo, it found this potential vulnerability:
* Vulnerability: Path Traversal
* Severity: High
* Location: packages/a2a-server/src/persistence/gcs.ts
* Line Content: const metadataObjectPath = this.getObjectPath(taskId, 'metadata');
* Description: The GCSTaskStore class uses a taskId to construct a path to an object in a GCS bucket. The taskId is taken from a URL parameter and is not sanitized. An attacker can use path traversal sequences (e.g., ../) to access arbitrary files in the GCS bucket. This could lead to leakage of sensitive information.
* Recommendation: The taskId should be sanitized before being used to construct the GCS object path. It should only be allowed to contain a restricted set of characters, such as alphanumeric characters, dashes, and underscores.

## Dive Deeper
Currently, an attacker could access / overwrite arbitrary files in GCS buckets using characters like `..`. Without proper sanitization, an attacker can control `taskId`  to navigate outside of the intended directory (`tasks/${taskId}/`). Using a taskId like `../../../../../etc/hosts` could potentially access sensitive system files.

Best practice is to ensure `taskId` only contains safe characters (i.e. no path traversal sequences, and only allowing alphanumeric characters, dashes, and underscores )

## Reviewer Test Plan

Created and ran path traversal specific tests under `gcs.test.ts` using `npx vitest run packages/a2a-server/src/persistence/gcs.test.ts`, which pass

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ✅  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

b/452392833 (https://b.corp.google.com/issues/452392833)

The issue seemed to be a high priority, so I wanted to start a fix sooner than later (although now realizing after already putting out this PR, it was probably a bit hasty to do so publicly, sorry!!)


jajanet:
Hi team! Just following up on this PR since it addresses a security concern -- I also filed a Buganizer ticket on this (https://b.corp.google.com/issues/452392833), which is now linked in the description

Could a maintainer please approve the workflow run when you get a chance? I enabled GH actions and set up the repo secret in my fork, but it seems like the workflow still won't run because of that? Let me know if you need any other information from my side too! Thank you! 

Tagging @bdmorgan @richieforeman @shishu314 @chrstnb based on approvals from recently closed PRs, sorry if these aren't the right people to tag (please let me know who I should otherwise!)

packages/a2a-server/src/persistence/gcs.test.ts
```diff
@@ -227,6 +227,35 @@ describe('GCSTaskStore', () => {
         'tar.c command failed to create',
       );
     });
+
+    it('should throw an error if taskId contains path traversal sequences', async () => {
+      const store = new GCSTaskStore('test-bucket');
+      const maliciousTask: SDKTask = {
+        id: '../../../malicious-task',
```
cornmander, line 234: GCS doesn't actually support path traversal as it simply has a flat namespace. But it's still not great for us to accept arbitrary inputs here.

cornmander (approved):

packages/a2a-server/src/persistence/gcs.test.ts
jajanet: Ah glad to see it's a non-issue, thank you!

PR #11224, opened 2025-10-15
chore: do not retry the model request if the user has aborted the request

### LINKED ISSUE #11223: Aborting request does not escape delay within retry
### What happened?

When a user aborts a request the retry logic does not abort but finishes the exponential backoff and retries the function call. At this point if the abort is propagated to the API the request is aborted.

Immediately abort the retry flow when aborted by the user.

### Client information

<details>

```console
> /about
# paste output here
```

</details>

### Login information

_No response_

_No response_

adamfweidman:
## TLDR

Abort during retry delay if user aborts during exponential backoff.

### Diving Deeper

We previously did not pass the abortSignal to the retry method, if users cancelled the request during a delay the API call would be retried and handled accordingly but ignored from by UI.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | x  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Resolves: https://github.com/google-gemini/gemini-cli/issues/11223


packages/core/src/utils/retry.ts
```diff
@@ -132,13 +162,17 @@ export async function retryWithBackoff<T>(
...
-        await delay(delayWithJitter);
+        await delay(delayWithJitter, signal);
         currentDelay = Math.min(maxDelayMs, currentDelay * 2);
         continue;
       }
 
       return result;
     } catch (error) {
+      if (error instanceof Error && error.name === 'AbortError') {
+        throw error;
```
cornmander, line 144: Does it make sense to throw here?

adamfweidman: It avoids a lot of the complicated logic below which makes this clearer but i don't think its strictly necessary

packages/core/src/utils/retry.ts
```diff
@@ -132,13 +162,17 @@ export async function retryWithBackoff<T>(
       ) {
         const jitter = currentDelay * 0.3 * (Math.random() * 2 - 1);
         const delayWithJitter = Math.max(0, currentDelay + jitter);
-        await delay(delayWithJitter);
+        await delay(delayWithJitter, signal);
```
SandyTao520, line 136: instead of listening and throwing in delay(), can we check `if (signal.aborted)` after this line and just return?

adamfweidman: The purpose is to escape the delay when aborted. Is there any reason to wait till after the delay?

SandyTao520: This works but I feel like it add complexity as we don't need to cancel the delay, just make sure we do nothing after.

I think all you need to add for this function is line 152 - 154.

SandyTao520 (approved):
LGTM

packages/core/src/utils/retry.ts
adamfweidman: I think the delay() function is a little hard to read but its a pretty simple idea as just a cancellable delay function. It makes the code more robust too, if we eventually have larger delays (e.g. for a future model we may want to allow users to set large delays if we have little capacity) it prevents an unlistened-to function running in the background

packages/core/src/utils/retry.ts
```diff
@@ -80,10 +81,31 @@ function defaultShouldRetry(
  * @param ms The number of milliseconds to delay.
  * @returns A promise that resolves after the delay.
  */
-function delay(ms: number): Promise<void> {
-  return new Promise((resolve) => setTimeout(resolve, ms));
-}
+function delay(ms: number, signal?: AbortSignal): Promise<void> {
```
jacob314: please move this to a utils library and add proper unit tests.

packages/core/src/utils/retry.ts
```diff
@@ -80,10 +81,31 @@ function defaultShouldRetry(
...
+  return new Promise((resolve, reject) => {
+    const abortError = new Error('Aborted');
```
jacob314: nit: simplify the flow by having a trivial version if there is no signal and this complex logic if there is a signal.

packages/core/src/utils/retry.test.ts
```diff
@@ -500,4 +500,30 @@ describe('retryWithBackoff', () => {
       expect(fallbackCallback).toHaveBeenCalledWith('oauth-personal');
     });
   });
+  it('should abort the retry loop when the signal is aborted', async () => {
```
jacob314, line 503: need a lot more tests to hit all the edge cases in the delay method.

packages/core/src/utils/retry.test.ts
```diff
@@ -500,4 +500,30 @@ describe('retryWithBackoff', () => {
...
+      throw error;
+    });
+
+    const promise = retryWithBackoff(mockFn, {
+      maxAttempts: 5,
+      initialDelayMs: 100,
+      signal: abortController.signal,
+    });
+
+    // eslint-disable-next-line vitest/valid-expect
```
jacob314: why is this lint needed?

jacob314 (commented):
thanks for fixing. this is an important case to fix! approved once these comments to try to avoid missing edge cases are addressed.

packages/core/src/utils/retry.ts
adamfweidman: Done, in delay.ts and delay.test.ts

packages/core/src/utils/retry.ts
adamfweidman: DOne.

packages/core/src/utils/retry.test.ts
adamfweidman: done

jacob314 (approved):
lgtm

PR #11225, opened 2025-10-15
fix(prompt): Make interactive command avoidance conditional

### LINKED ISSUE #11281: System Prompt limits command execution to only non-interactive commands
### What happened?

We should update the system-prompt to conditionally limit interactive commands.

When in non-interactive mode, you should execute non-interactive commands only otherwise don't block any command

### Client information

<details>

```console
> /about
# paste output here
```

</details>

### Login information

_No response_

_No response_

galdawave:
## TLDR

This pull request makes the "- **Interactive Commands**" section of the core system prompt conditional. This instruction will now only be included when the CLI is running in a non-interactive mode (`config.isInteractive` is `false`).

This change is to reduce the size and complexity of the system prompt for interactive sessions where this guidance is not relevant.

## Dive Deeper

The system prompt contains guidance for the model on how to behave. One piece of guidance is to avoid using interactive shell commands (like `git rebase -i`) because they can cause the CLI to hang in non-interactive or automated environments.

However, when a user is in an interactive session, they can handle these prompts directly, making the warning unnecessary. Removing it in interactive mode saves tokens and simplifies the instructions for the model.

The implementation uses a small, self-executing function within the template literal for `basePrompt` in `packages/core/src/core/prompts.ts`. This conditionally returns the text block based on the `config.isInteractive` flag, which is a clean way to handle this and is consistent with other conditional logic in the same file.

## Reviewer Test Plan

To test this change, a reviewer can inspect the generated system prompt in both interactive and non-interactive modes.

1.  Temporarily add `console.log(basePrompt);` at the end of the `getCoreSystemPrompt` function in `packages/core/src/core/prompts.ts`.
2.  Rebuild the CLI: `npm run build`.
3.  Run in **interactive mode**: `npx dist/gemini.js`. The logged prompt **should NOT** contain the "Interactive Commands" section.
4.  Run in **non-interactive mode**: `npx dist/gemini.js -p "list files"`. The logged prompt **SHOULD** contain the "Interactive Commands" section.
5.  Remember to remove the `console.log` before merging.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ✅  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

*None*


anj-s (approved):

packages/core/src/config/config.ts
```diff
@@ -963,6 +966,10 @@ export class Config {
     return this.interactive;
   }
 
+  getPtyInfo(): string {
```
NTaylorMullen: What do you think about renaming this to `getInteractiveShellEnabled` and making the logic understand the ptyInfo + the interactive property? That way anything trying to understand support for interactive shell pulls from the same context?

packages/core/src/core/prompts.ts
```diff
@@ -207,7 +207,13 @@ IT IS CRITICAL TO FOLLOW THESE GUIDELINES TO AVOID EXCESSIVE TOKEN CONSUMPTION.
...
-- **Interactive Commands:** Try to avoid shell commands that are likely to require user interaction (e.g. \`git rebase -i\`). Use non-interactive versions of commands (e.g. \`npm init -y\` instead of \`npm init\`) when available, and otherwise remind the user that interactive shell commands are not supported and may cause hangs until canceled by the user.
+${(function () {
+  if (!config.isInteractive() || config.getPtyInfo() === 'child_process') {
+    return `- **Interactive Commands:** Some commands are interactive, meaning they can accept user input during their execution (e.g. ssh, vim). Only execute non-interactive commands. Use non-interactive versions of commands (e.g. \`npm init -y\` instead of \`npm init\`) when available. Interactive shell commands are not supported and may cause hangs until canceled by the user.`;
+  } else {
+    return `- **Interactive Commands:** Some commands are interactive, meaning they accept user input during their execution (e.g. ssh, vim). Before executing this command must let the user know they can press \`ctrl + f\` to focus into the shell to provide input.`;
```
NTaylorMullen: We should probably still encourage the model to prefer non-interactive commands (faster and doesn't interrupt the flow). The key part is that for scenarios where there's no choice the model can now choose to invoke something interactive

```suggestion
    return `- **Interactive Commands:** Prefer non-interactive commands when it makes sense; however, some commands are only interactive and expect user input during their execution (e.g. ssh, vim). If you choose to execute an interactive command consider letting the user know they can press \`ctrl + f\` to focus into the shell to provide input.`;
```

NTaylorMullen (approved):

PR #11234, opened 2025-10-15
Fixed Model stream ended with empty response text error

ooples:
TLDR

  This PR fixes an InvalidStreamError that occurred when the Gemini model legitimately returned an empty text response but indicated completion. It also updates the inline comments to clarify the logic and rationale behind this change.

  Dive Deeper

  The processStreamResponse method in packages/core/src/core/geminiChat.ts was throwing an InvalidStreamError with the message "Model stream ended with empty response text." under certain conditions. Specifically, if there were no tool calls, the model signaled completion (hasFinishReason was true), but the consolidated response text
  (responseText) was empty, the error would be triggered.

  This logic was overly aggressive because an empty responseText can be a legitimate outcome from the model. For instance, the model might:
   * Generate only internal "thoughts" (which are stripped from responseText).
   * Process a simple command that doesn't require a textual output.
   * Successfully call a tool, and the tool's output is the primary response, not additional text from the model.

  The fix modifies the error condition from if (!hasToolCall && (!hasFinishReason || !responseText)) to if (!hasToolCall && !hasFinishReason). This ensures that if a finishReason is present (indicating model completion), the InvalidStreamError is not thrown, regardless of whether responseText is empty. This correctly treats a
  model-indicated completion with an empty text response as a valid (empty) result.

  The inline comments have been updated to clearly explain this revised logic and provide examples of when an empty response might be valid.

  Reviewer Test Plan

  To validate this change, reviewers should:
   1. Pull the branch and rebuild the project. From the root gemini-cli directory, run npm install followed by npm run build.
   2. Run the CLI with prompts that are likely to produce minimal or no textual output from the model.
       * Example 1 (Simple Acknowledgment): Try prompts like "Okay.", "Got it.", "Thanks.", or "Understood."
       * Example 2 (Pure Thought Generation): If possible, craft a prompt that encourages the model to primarily generate internal thoughts without explicit text output. This might involve asking the model to "think step-by-step" without a final answer.
       * Example 3 (Tool Call with Minimal Text): If there's a tool that can be called which typically produces a non-textual or very brief textual response, try to trigger that tool.
   3. Verify: In these scenarios, confirm that the CLI does not throw an InvalidStreamError: Model stream ended with empty response text.. The CLI should complete the turn gracefully, potentially with an empty displayed response, without crashing.

  Testing Matrix

  ┌──────────┬────┬────┬────┐
  │          │ 🍏 │ 🪟 │ 🐧 │
  ├──────────┼────┼────┼────┤
  │ npm run  │ ✅ │ ✅ │ ✅ │
  │ npx      │ ✅ │ ✅ │ ✅ │
  │ Docker   │ ❓ │ ❓ │ ❓ │
  │ Podman   │ ❓ │ ❓ │ ❓ │
  │ Seatbelt │ ❓ │ ❓ │ ❓ │
  └──────────┴────┴────┴────┘

  Linked issues / bugs

  Fixes #<issue_number_if_available> (Please replace `<issue_number_if_available>` with the actual issue number if one exists for this bug).
  Related to the reported error: InvalidStreamError: Model stream ended with empty response text.


bdmorgan:
Hi @ooples, thank you so much for your contribution to Gemini CLI! We really appreciate the time and effort you've put into this.

We're making some updates to our contribution process to improve how we track and review changes. Please take a moment to review our recent discussion post: [Improving Our Contribution Process & Introducing New Guidelines](${DISCUSSION_URL}).

Key Update: Starting **January 26, 2026**, the Gemini CLI project will require all pull requests to be associated with an existing issue. Any pull requests not linked to an issue by that date will be automatically closed.

Thank you for your understanding and for being a part of our community!

PR #11252, opened 2025-10-16
fix: Update web_fetch tool definition to instruct the model to provid…

shrutip90:
…e only valid, complete URLs

## TLDR

Made a simple change in the prompt. There are no integration tests for this tool currently, so did not update any tests.

Also noticed that the model was behaving correctly even without the changes. For example, on gemini latest:
Screenshot 2025-10-16 at 12 29 59 AM

Unsure if the issue was seen intermittently or with a specific prompt. Please let me know if there is any specific prompt that I should test with.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes https://github.com/google-gemini/gemini-cli/issues/10731

### DIFF
--- a/packages/core/src/tools/web-fetch.ts
+++ b/packages/core/src/tools/web-fetch.ts
@@ -395,7 +395,7 @@ export class WebFetchTool extends BaseDeclarativeTool<
         properties: {
           prompt: {
             description:
-              'A comprehensive prompt that includes the URL(s) (up to 20) to fetch and specific instructions on how to process their content (e.g., "Summarize https://example.com/article and extract key points from https://another.com/data"). Must contain as least one URL starting with http:// or https://.',
+              'A comprehensive prompt that includes the URL(s) (up to 20) to fetch and specific instructions on how to process their content (e.g., "Summarize https://example.com/article and extract key points from https://another.com/data"). All URLs to be fetched must be valid, complete, and start with http:// or https://.',
             type: 'string',
           },
         },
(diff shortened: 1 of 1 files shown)


abhipatel12:
This LGTM, but @anj-s should also approve before we merge.

packages/core/src/tools/web-fetch.ts
```diff
@@ -395,7 +395,7 @@ export class WebFetchTool extends BaseDeclarativeTool<
         properties: {
           prompt: {
             description:
-              'A comprehensive prompt that includes the URL(s) (up to 20) to fetch and specific instructions on how to process their content (e.g., "Summarize https://example.com/article and extract key points from https://another.com/data"). Must contain as least one URL starting with http:// or https://.',
+              'A comprehensive prompt that includes the URL(s) (up to 20) to fetch and specific instructions on how to process their content (e.g., "Summarize https://example.com/article and extract key points from https://another.com/data"). All URLs to be fetched must be valid, complete, and start with http:// or https://.',
```
anj-s: can we say that a valid URL start with http:// or http:// and end with .com, .org. .edu or any domain name such as ths?

I think when we say that URL should be valid and complete, we are not defining what is valid and complete. Lets specify that as well as mentioned above.
Also do we want to remove the specification that a prompt should have at least 1 URL?

shrutip90: Updated to say that they should include a valid hostname. Since domain name TLD list is pretty long and can keep changing, I think it is best to not list them. We also have validation in validateToolParamValues that will take care of this. It also checks that there is at least one valid URL. Please lmk if this looks alright?

anj-s (approved):

PR #11261, opened 2025-10-16
Add extensions logging

### LINKED ISSUE #11113: Thread extension name through MCP server and slash command logging
We should also include the name of the extension (if any) in slash command and MCP server tool invocation logging.

chrstnb:
## TLDR

This change introduces enhanced telemetry to track the usage of extensions and tools.

Key changes:
- **Extension Logging:** The `StartSessionEvent` now captures the count of enabled extensions and their IDs.
- **Tool Call Attribution:** An `extension_id` has been added to tool call and slash command telemetry events. This allows us to attribute tool usage back to the specific extension that provides the tool.
- **Extension Listing:** Added the ID to the extension list output.
- **Extension source logging:** swapped to using the 'type' over the 'source' in extension logging for the 'source' field. 

These changes will provide better insights into which extensions and tools are being used, helping to guide future development and identify potential issues.

cl/822243850

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes #11113


packages/cli/src/services/CommandService.ts
```diff
@@ -82,6 +82,7 @@ export class CommandService {
       commandMap.set(finalName, {
         ...cmd,
         name: finalName,
+        extension_id: cmd.extensionName,
```
jakemac53: `Object literal may only specify known properties, and 'extension_id' does not exist in type 'SlashCommand'.`

jakemac53: ugh nvm this is just the typescript compiler being dumb like usual

packages/cli/src/services/CommandService.ts
```diff
@@ -82,6 +82,7 @@ export class CommandService {
...
       });
```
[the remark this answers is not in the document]
jakemac53: This is a nullable field, bad gemini

packages/cli/src/ui/commands/types.ts
```diff
@@ -196,6 +196,7 @@ export interface SlashCommand {
 
   // Optional metadata for extension commands
   extensionName?: string;
+  extension_id?: string;
```
jakemac53: All of these other ones are camelCase, this should be too.

Also, isn't the ID just the name?

jakemac53: We should potentially consider making the ID be the source or something more unique though, there is nothing today enforcing unique names across users, only for a single user. The name is actually not very useful as a metric.

packages/core/src/telemetry/clearcut-logger/event-metadata-key.ts
```diff
@@ -391,6 +391,12 @@ export enum EventMetadataKey {
   // Logs the status of the extension uninstall
   GEMINI_CLI_EXTENSION_UPDATE_STATUS = 118,
 
+  // Logs the count of extensions in Start Session Event
+  GEMINI_CLI_START_SESSION_EXTENSIONS_COUNT = 119,
+
+  // Logs the name of extensions as a comma-separated string
+  GEMINI_CLI_START_SESSION_EXTENSIONS = 120,
```
jakemac53: Extension authors also tend to want to be able to answer questions like "what percentage of my users are on the latest version". Can we also include version info here, or is that a bit much?

anj-s:
adding @owenofbrien who is POC for logging. Can you take a look at this PR since we have clearcut logging changes?

packages/core/src/telemetry/clearcut-logger/event-metadata-key.ts
chrstnb: Let's do that in a follow-up

jakemac53 (approved):
LGTM although consider also hashing the names

jakemac53 (approved):

packages/core/src/telemetry/clearcut-logger/event-metadata-key.ts
```diff
@@ -7,7 +7,7 @@
 // Defines valid event metadata keys for Clearcut logging.
 export enum EventMetadataKey {
   // Deleted enums: 24
-  // Next ID: 117
+  // Next ID: 122
```
owenofbrien, line 10: Please provide a link to the corresponding g3 CL in the PR description.

owenofbrien (approved):

packages/core/src/telemetry/clearcut-logger/event-metadata-key.ts
chrstnb: done

PR #11266, opened 2025-10-16
fix(docs): Broken Images on Themes

### LINKED ISSUE #11265: Theme Images missing
### URL of the page with the issue

https://geminicli.com/docs/cli/themes/

### What is the problem?

The assets work on GitHub but not when the build happens. It appears to be an odd setup for docs where they get set off to a GitHub action and then a POST request to a webhook. Which then uses `docs-page-action.yml` to build out the site on GH Pages.

Image

I expect the images to show correctly

Image

### Additional context

_No response_

codercatdev:
## TLDR

The assets work on GitHub but not when the build happens. It appears to be an odd setup for docs where they get set off to a GitHub action and then a POST request to a webhook. Which then uses `docs-page-action.yml` to build out the site on GH Pages. This fix points images to domain level assets as they are at the root. It allows for both the GH images and site to use them.

Image

## Reviewer Test Plan

Check website at https://geminicli.com/docs/cli/themes/

## Linked issues / bugs

- Fixes #11265


richieforeman:
Great catch! thanks for fixing this up!

richieforeman (approved):

PR #11318, opened 2025-10-16
feat: remove deprecated telemetry flags

allenhutchison:
Removes the following deprecated telemetry flags from the CLI:
- --telemetry
- --telemetry-target
- --telemetry-otlp-endpoint
- --telemetry-otlp-protocol
- --telemetry-log-prompts
- --telemetry-outfile

These flags have been replaced by settings in the `settings.json` file.
The related tests have been updated or removed.


jacob314 (approved):
So many deletions!!lgtm

PR #11323, opened 2025-10-16
fix(ui): escaping theme dialog no longer resets theme to default

psinha40898:
## TLDR

Currently pressing "esc" in the theme dialog will reset the user theme to default.

## Dive Deeper

There were two clear paths to resolving this bug. This PR implements **Option 2**

Option 1)
Simply add a return statement inside of handleThemeSelect that returns when `themeName` is undefined. This works because `ThemeDialog.tsx` handles the escape case by calling `onSelect(undefined, selectedScope);` The lack of this return statement leads to a path where undefined defaults to the default theme. 

You can see that branch here :
https://github.com/psinha40898/gemini-cli/tree/pyush/fix/11319
https://github.com/google-gemini/gemini-cli/compare/main...psinha40898:gemini-cli:pyush/fix/11319

**Option 2) Refactor Approach (Architectural Fix)**
Add a dedicated `onCancel` callback to the `ThemeDialog` component and export a `closeThemeDialog` function from `useThemeCommand`. 

This works because `ThemeDialog.tsx` now handles the escape case by calling `onCancel()` instead of `onSelect(undefined, selectedScope)`. 

The refactor prevents undefined from ever reaching `handleThemeSelect` by separating the cancel action into its own callback.

This requires changes across multiple files to wire up the new callback through the component hierarchy, but it eliminates the ambiguity of using undefined and provides type safety by making `handleThemeSelect` unable to accept undefined.

Maybe option 2 should still accept undefined? My thinking is no b/c TS will make sure we don't have any undefined calls still in the codebase but would love the advice of a more experienced eye

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes #11319


packages/cli/src/ui/hooks/useThemeCommand.ts
```diff
@@ -63,8 +61,12 @@ export const useThemeCommand = (
     [applyTheme],
   );
 
+  const closeThemeDialog = useCallback(() => {
+    setIsThemeDialogOpen(false);
+  }, []);
```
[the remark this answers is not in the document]
psinha40898: This is important thank you gemini bot

miguelsolorio (approved):
Tested the functionality and it preserves the original setting properly. Nice work!

packages/cli/src/ui/components/ThemeDialog.test.tsx
```diff
@@ -105,4 +106,28 @@ describe('ThemeDialog Snapshots', () => {
...
+          />
+        </KeypressProvider>
+      </SettingsContext.Provider>,
+    );
+
+    act(() => {
+      stdin.write('\x1b');
+    });
+
+    await new Promise((resolve) => setTimeout(resolve, 100));
```
jacob314: waiting 100ms will lead to flakey tests please use
waitFor for mockOnCancel to be called.

jacob314 (commented):
lgtm
approved once nit about test is resolved.

packages/cli/src/ui/components/ThemeDialog.test.tsx
psinha40898: Now using `waitFor` from `@testing-library/react`

psinha40898: only tangentially related to this pr
I noticed that `SettingsDialog.test.tsx` uses a custom `waitFor` implementation introduced in #8396

My surmise:
- we can definitely rely on `waitFor` from `@testing-library/react`
- although it relies on a DOM environment, our testing environment deliberately sets it up as so with `environment: 'jsdom' ` in our `vitest.config`
- therefore the custom implementation is not needed

Am I wrong?

jacob314 (approved):
lgtm

PR #11338, opened 2025-10-17
feat: Remove deprecated flags

allenhutchison:
Removes the following deprecated flags:
- proxy
- show-memory-usage
- sandbox-image
- checkpointing


jacob314 (approved):
lgtm

PR #11351, opened 2025-10-17
fix(cli): Add delimiter before printing tool response in non-interactive mode

### LINKED ISSUE #9048: No delimiter between outputs in non-interactive mode
### What happened?

When I run `gemini -p "..."` and it involves more than one step it  just keeps writing outputs without separating them with newlines and tool calls if they a used not show in any way.

I expect it to put some delimeters, i.e. double newline, between output parts. 

### Client information

<details>

```console
> /about
╭──────────────────────────────────────────────────────────────────────────────╮
│                                                                              │
│ About Gemini CLI                                                             │
│                                                                              │
│ CLI Version                0.6.0-preview.4                                   │
│ Git Commit                 85409d73                                          │
│ Model                      gemini-2.5-pro                                    │
│ Sandbox                    no sandbox                                        │
│ OS                         linux                                             │
│ Auth Method                OAuth                                             │
│ GCP Project                wired-benefit-470703-b6                           │
│                                                                              │
╰──────────────────────────────────────────────────────────────────────────────╯
```

</details>

### Login information

_No response_

_No response_

krishna-kb:
## TLDR

This pull request fixes #9048 by adding a newline delimiter between outputs in non-interactive mode. This improves readability for multi-step sessions involving tool calls. A new unit test has been added to verify the correct spacing between consecutive tool call turns.

## Dive Deeper

The core change is a single line in
packages/cli/src/nonInteractiveCli.ts that writes a newline to stdout
after the tool execution loop finishes for a given turn.
Initially, I settled on a single newline (\n) for a cleaner, more standard command-line experience
that avoids excessive whitespace.

The bulk of the work was in ensuring this was properly tested. I added
 a new unit test in packages/cli/src/nonInteractiveCli.test.ts (should
 write a newline between multiple tool call turns) that simulates a
multi-turn conversation with two consecutive tool calls before a final
 text response. It asserts that a newline is printed after each tool
turn.

## Reviewer Test Plan
To manually validate the change, you can run a prompt in non-interactive mode 
that forces a tool call. A good example is using a command that requires 
listing files.
Example Prompt:
`gemini -p "List the files in the 'packages/cli/src' directory and then tell me 
what the 'runNonInteractive' function does."`

Before this change, the output would look something like this, with the final answer immediately following the tool output:
```
[...tool output...]Final answer is...
```

After this change, you should see a clean separation:
```
[...tool output...]

Final answer is...
```

This confirms the delimiter is being added correctly between the tool turn and the final text response turn.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


packages/cli/src/nonInteractiveCli.ts
```diff
@@ -198,6 +198,9 @@ export async function runNonInteractive(
         }
 
         if (toolCallRequests.length > 0) {
+          if (config.getOutputFormat() !== OutputFormat.JSON) {
+            process.stdout.write('\n');
+          }
```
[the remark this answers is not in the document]
krishna-kb: There is a text response before the tool response is printed. If we move it to afterwards, the bug would still exist.
I think the description might be unclear, will update the same.

packages/cli/src/nonInteractiveCli.ts
```diff
@@ -198,6 +198,9 @@ export async function runNonInteractive(
         }
 
         if (toolCallRequests.length > 0) {
+          if (config.getOutputFormat() !== OutputFormat.JSON) {
```
jacob314: I think it would be more robust if tracked the blocks of output we are writing and ensured that there are line breaks between the blocks.  As is, how do we ensure we don't add an un-needed line break before the very first output for example?

packages/cli/src/nonInteractiveCli.test.ts
```diff
@@ -262,6 +262,86 @@ describe('runNonInteractive', () => {
...
+      mockConfig,
+      mockSettings,
+      'Use multiple tools',
+      'prompt-id-multi',
+    );
+
+    // Verify that sendMessageStream was called three times
+    expect(mockGeminiClient.sendMessageStream).toHaveBeenCalledTimes(3);
+
+    // Count how many times the newline was written
```
jacob314: need tests that verify exactly the output rendered rather than just the number of newlines. otherwise the newlines might be in odd places.

jacob314 (commented):
Looks promising. Added some comments on how to do this in a more robust way,.

packages/cli/src/nonInteractiveCli.test.ts
krishna-kb: Agreed. Updated the test to check for the exact response containing newlines combined with mock tool responses.

packages/cli/src/ui/utils/textOutput.ts
```diff
@@ -0,0 +1,49 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+// Internal state for the singleton
+let lastChar = '\n';
```
jacob314: nit: change this to 
bool 
atStartOfLine or column. Having this be '\n' to start is otherwise misleading.

packages/cli/src/ui/utils/textOutput.ts
```diff
@@ -0,0 +1,49 @@
...
+export const textOutput = {
+  /**
+   * Writes a string to stdout.
+   * @param str The string to write.
+   */
+  write(str: string): void {
+    if (str.length === 0) {
+      return;
+    }
+    process.stdout.write(str);
```
jacob314, line 25: nit: strip ascii escape codes here when determining whether we are still on the start of the line. There are other parts of the code base where we strip that you can find where we do this.

packages/cli/src/ui/utils/textOutput.ts
```diff
@@ -0,0 +1,49 @@
...
+  },
+
+  /**
+   * Writes a string to stdout, ensuring it starts on a new line.
+   * If the previous output did not end with a newline, one will be added.
+   * This prevents adding extra blank lines if a newline already exists.
+   * If no string is provided, it just ensures the output ends with a newline.
+   * @param str The optional string to write.
+   */
+  writeOnNewLine(str?: string): void {
```
jacob314: why is string nullable? would simplify and require the string. why do we need to write when we might not have something to write?

packages/cli/src/ui/utils/textOutput.ts
```diff
@@ -0,0 +1,49 @@
...
+    if (lastChar !== '\n') {
+      this.write('\n');
+    }
+    if (str !== undefined) {
+      this.write(str);
+    }
+  },
+
+  /**
```
jacob314, line 45: if this wasn't a singleton you wouldn't have this issue testing. I would make this an instance so you can test easily. Given you are only using this in nonInteractiveCli there is no need to have it be a singleton. you can always pass it around or change it to a singleton later if that is needed.

packages/cli/src/nonInteractiveCli.ts
```diff
@@ -199,6 +201,7 @@ export async function runNonInteractive(
         }
 
         if (toolCallRequests.length > 0) {
+          textOutput.writeOnNewLine();
```
jacob314: nit: rather than making this a bare call, make the actual output of the toolCallRequests that occurs a bit later do the actual writeOnNewLine

packages/cli/src/ui/utils/textOutput.test.ts
```diff
@@ -0,0 +1,77 @@
...
+
+  it('write() should not call process.stdout.write for empty strings', () => {
+    textOutput.write('');
+    expect(stdoutSpy).not.toHaveBeenCalled();
+  });
+
+  it('writeOnNewLine() should not add a newline if the last char was a newline', () => {
+    // Default state starts with a newline
+    textOutput.writeOnNewLine('hello');
+    expect(stdoutSpy).toHaveBeenCalledWith('hello');
```
jacob314: for these tests the easiest way is to just check that the final output to the fae stdout matches an expected snapshot. that is simpler for people to verify the output in the tests. See examples in the codebase of that.

jacob314 (commented):
Looks good once these comments are addressed!

packages/cli/src/nonInteractiveCli.ts
krishna-kb: The actual output of toolCallRequests is processed by the model and then written at a generic place which also prints contagious texts from model.
I have added a new ensureNewLine method to the utility to solve for this.

package-lock.json
```diff
@@ -621,6 +622,7 @@
         }
       ],
       "license": "MIT",
+      "peer": true,
```
jacob314, line 625: revert this file. it isn't related to your PR

jacob314 (approved):
lgtm

PR #11358, opened 2025-10-17
Fix: Preserve trailing exclamation marks in prompts

### LINKED ISSUE #11329: Gemini CLI drops trailing single exclamation marks
## Issue
Gemini CLI consistently drops trailing single exclamation marks (`!`) from user prompts.

## Reproduction
```bash
echo "Hello!" | gemini --yolo
# Output: "Hello" (missing trailing !)

echo "Great job!" | gemini --yolo  
# Output: "Great job" (missing trailing !)
```

## Observed Behavior
- Single trailing `!` → Always dropped
- Middle `!` → Preserved (e.g., "Hello! How" works)
- Multiple trailing `!!!` → Partially preserved (keeps all but last)

## Expected Behavior
Trailing exclamation marks should be preserved in the prompt.

## Impact
Users cannot send messages ending with exclamation marks, affecting natural language communication.

## Environment
- Gemini CLI: Latest version
- Command: `echo "text!" | gemini --yolo`

Smetalo:
The yargs parser was configured to interpret a trailing `!` as a boolean negation, which caused it to strip the character from prompts ending with an exclamation mark.

This change disables the `boolean-negation` feature in the yargs parser configuration to ensure that the input prompt is preserved exactly as the user provides it.

A test case has been added to verify that prompts ending with a single trailing exclamation mark are handled correctly.

Resolves #11329


packages/cli/src/config/config.ts
```diff
@@ -73,6 +73,7 @@ export interface CliArgs {
 export async function parseArguments(settings: Settings): Promise<CliArgs> {
   const rawArgv = hideBin(process.argv);
   const yargsInstance = yargs(rawArgv)
+    .parserConfiguration({ 'boolean-negation': false })
```
cornmander, line 78: I looked at this setting and it's not clear to me that this should be causing the reported issue...

Smetalo:
@Adib234 Merged with the latest main branch.

Adib234 (approved):

cornmander (changes requested):

PR #11365, opened 2025-10-17
Revert "Update package.json in include git dependency on Ink fork."

jakemac53:
Reverts google-gemini/gemini-cli#11330

I am unable to build on Linux with this change, reverting until we can diagnose this.


jacob314 (approved):
lgtm

PR #11377, opened 2025-10-17
add extension IDs

jakemac53:
## TLDR

Adds an `id` to extensions generated by hashing either the source or the name, depending on extension install type.

## Dive Deeper

This does mean that the same extension installed in different ways might have a different ID - but most of the time that is probably desirable. An extension using a local install dir or custom fork will have a different ID than the same extension installed from the official repo, which allows differentiating different forks of an extension even if they have the same name.

## Reviewer Test Plan

Run the tests, this ID isn't used yet.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ✅   |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

To be used by https://github.com/google-gemini/gemini-cli/pull/11261 for extension logging

### DIFF
--- a/packages/cli/src/config/extension.ts
+++ b/packages/cli/src/config/extension.ts
@@ -35,7 +35,7 @@ import {
 } from './extensions/variables.js';
 import { isWorkspaceTrusted } from './trustedFolders.js';
 import { resolveEnvVarsInObject } from '../utils/envVarResolver.js';
-import { randomUUID } from 'node:crypto';
+import { randomUUID, createHash } from 'node:crypto';
 import {
   cloneFromGit,
   downloadFromGitHubRelease,
@@ -253,6 +253,24 @@ export function loadExtension(
       )
       .filter((contextFilePath) => fs.existsSync(contextFilePath));
 
+    // IDs are created by hashing parts of the extension metadata in order to
+    // obfuscate any potentially sensitive information such as private
+    // git urls or project names.
+    const hash = createHash('sha256');
+    switch (installMetadata?.type) {
+      case 'git':
+      case 'github-release':
+        // For git and github-release installs, identify the extension based on the
+        // source it was installed from. This allows us to distinguish between
+        // extensions that happen to have the same name but are actually different.
+        hash.update(installMetadata.source);
+        break;
+      default:
+        // For all other install types, just use the name.
+        hash.update(config.name);
+    }
+    const id = hash.digest('hex');
+
     return {
       name: config.name,
       version: config.version,
@@ -262,6 +280,7 @@ export function loadExtension(
       mcpServers: config.mcpServers,
       excludeTools: config.excludeTools,
       isActive: true, // Barring any other signals extensions should be considered Active.
+      id,
     };
   } catch (e) {
     console.error(
--- a/packages/cli/src/config/extension.ts
+++ b/packages/cli/src/config/extension.ts
@@ -253,22 +253,12 @@ export function loadExtension(
       )
       .filter((contextFilePath) => fs.existsSync(contextFilePath));
 
-    // IDs are created by hashing parts of the extension metadata in order to
-    // obfuscate any potentially sensitive information such as private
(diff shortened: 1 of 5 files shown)


packages/cli/src/config/extension.ts
```diff
@@ -253,6 +253,24 @@ export function loadExtension(
...
+      case 'git':
+      case 'github-release':
+        // For git and github-release installs, identify the extension based on the
+        // source it was installed from. This allows us to distinguish between
+        // extensions that happen to have the same name but are actually different.
+        hash.update(installMetadata.source);
+        break;
+      default:
+        // For all other install types, just use the name.
+        hash.update(config.name);
```
chrstnb: Should we use the source here as well?

packages/cli/src/config/extension.ts
```diff
@@ -253,6 +253,24 @@ export function loadExtension(
...
+    // obfuscate any potentially sensitive information such as private
+    // git urls or project names.
+    const hash = createHash('sha256');
+    switch (installMetadata?.type) {
+      case 'git':
+      case 'github-release':
+        // For git and github-release installs, identify the extension based on the
+        // source it was installed from. This allows us to distinguish between
+        // extensions that happen to have the same name but are actually different.
+        hash.update(installMetadata.source);
```
chrstnb: Do we need to do any standardization/clean-up of this? (e.g., will it be 'https://github.com/org/repo' in one place and 'https://github.com/org/repo/' in another?)

packages/cli/src/config/extension.ts
jakemac53: We could but its not really clear if that is going to be useful... while it might sort of work out for certain scenarios (if all users are using an extension from the same system path), it definitely won't work across operating systems and things.

I think just the name is probably a better fallback.

jakemac53: Also I think when we log the extension ID for people, we can provide a warning if it is anything other than a git or github-release extension, that the ID is only as unique as the name.

packages/cli/src/config/extension.ts
jakemac53: Yeah it does look like we record the source exactly as the user provided... so installing via the ssh URI is also going to be different.

jakemac53: Ok, I added some canonicalization of GitHub https URIs but we can't do much more than that I don't think because we don't know how to canonicalize safely any other URI.

chrstnb: ok great, that sounds reasonable

packages/cli/src/config/extension.ts
jakemac53: Ok, updated everything to use the source if available.

packages/cli/src/config/extension.ts
jakemac53: Extended this to also handle ssh URLs.

packages/cli/src/config/extension.ts
```diff
@@ -253,6 +254,28 @@ export function loadExtension(
...
+        ? tryParseGithubUrl(installMetadata.source)
+        : null;
+    if (githubUrlParts) {
+      // For github repos, we use the https URI to the repo as the ID.
+      hash.update(
+        `https://github.com/${githubUrlParts.owner}/${githubUrlParts.repo}`,
+      );
+    } else {
+      hash.update(installMetadata?.source ?? config.name);
+    }
```
chrstnb, line 275: just to clarify, is there anything we can do here to try to ensure this is the same across platforms?

jakemac53: I don't think there is really anything we can do for that, or at least I don't know of any sensible thing.

chrstnb (approved):

PR #11380, opened 2025-10-17
feat(cli): Suppress slash command execution and suggestions in shell …

### LINKED ISSUE #11253: Slash commands are getting execute in shell mode
### What happened?

After entering '!' in input box, it goes to shell mode. 
Inside shell mode, if you type '/' the slash command suggestions are seen. Once the suggestion is entered - it actually gives the shell command response ignoring that the input box is in shell mode.

Slash commands must not be executed in shell mode.

### Client information

│ About Gemini CLI                                                    │
│                                                                     │
│ CLI Version             0.9.0                                       │
│ Git Commit              a93d92a3                                    │
│ Model                   gemini-2.5-pro                              │
│ Sandbox                 no sandbox                                  │
│ OS                      linux                                       │
│ Auth Method             OAuth                                       │
│ GCP Project             aipp-project-t                              │
│ IDE Client              VS Code     

### Login information

Happens in all login options

_No response_

MJjainam:
…mode

## Linked issues / bugs
- Fixes #11253

## UX

https://github.com/user-attachments/assets/6cef39e8-0c58-41f9-9cd0-7283a9162345


jacob314:
One more case we should handle is `@` completions. With your change you still get completions for `@` which is a good thing but it is confusing because the completions will be invalid shell commands. What is needed is for the `@` commands to complete to the raw path without an @ when in shell mode.

When I accept this completion I should get `ls packages/` not `ls @packages`
image

packages/cli/src/ui/hooks/useCommandCompletion.test.ts
```diff
@@ -246,6 +251,58 @@ describe('useCommandCompletion', () => {
           );
         });
       });
+
+      it('should show slash command suggestions when shellModeActive is false', async () => {
```
jacob314: ask Gemini CLi to refactor this as a parameterized test. that will avoid all the duplicate logic

jacob314 (commented):
Approved once tests are updated.
handling `@` more gracefully in shell mode can be a follow up PR.

jacob314 (approved):
lgtm

PR #11398, opened 2025-10-17
MCP docs update

rickyrobinett:
## TLDR

Small update to the MCP docs to explicitly call out the location of settings.json during setup.

## Dive Deeper

I was adding MCP and it took me a bit to find exactly where I should be updating settings.json. I wanted to add the location at the start of the MCP documentation so folks could find it right away.

## Reviewer Test Plan

Only a docs update, but confirm the locations are correct.

## Testing Matrix

Only a docs update. Ran `npm run preflight` to make sure everyhting looked ok.


jkcinouye:
@rickyrobinett Hi! It looks like this is definitely a great callout, but you'd need to fill out your CLA to submit.

The directions are here:

https://github.com/google-gemini/gemini-cli/pull/11398/checks?check_run_id=53041247883

Let me know if you have any questions!

jkcinouye:
@rickyrobinett Friendly ping.

PR #11416, opened 2025-10-17
feat(cli): continue request after disabling loop detection

SandyTao520:
## TLDR

This PR adds the ability to automatically continue the user's request after they disable loop detection for the current session.

## Dive Deeper

When a loop is detected, the user is presented with a confirmation dialog. If they choose to "Disable for session", the `useGeminiStream` hook now automatically calls `submitQuery` with `isContinuation: true` using the last query and prompt ID. This allows the model to resume generating a response based on the existing context, providing a smoother user experience.

The implementation required refactoring `useGeminiStream.ts` to define the confirmation callback inline within `submitQuery` to avoid circular dependencies. Tests in `useGeminiStream.test.tsx` have been updated to verify this new behavior.

## Reviewer Test Plan

1.  Ask Gemini CLI: `use the replace tool to append a number to SECURITY.md, repeat this 50 times.`
2.  When the loop detection confirmation dialog appears, select "Disable for session".
3.  Verify that the model automatically continues generating a response from where it left off.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓ |
| npx      | ✅  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

### DIFF
--- a/packages/cli/src/ui/hooks/useGeminiStream.test.tsx
+++ b/packages/cli/src/ui/hooks/useGeminiStream.test.tsx
@@ -2620,10 +2620,25 @@ describe('useGeminiStream', () => {
       };
       mockConfig.getGeminiClient = vi.fn().mockReturnValue(mockClient);
 
-      mockSendMessageStream.mockReturnValue(
+      // Mock for the initial request
+      mockSendMessageStream.mockReturnValueOnce(
         (async function* () {
           yield {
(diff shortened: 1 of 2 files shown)


jacob314 (approved):
lgtm

PR #11419, opened 2025-10-17
fix(cli): enable typechecking for ui/components tests

SandyTao520:
## TLDR

Enabled typechecking for test files under `packages/cli/src/ui/components/` by removing them from `tsconfig.json` excludes and fixing resulting TypeScript errors.

## Dive Deeper

The test files in `packages/cli/src/ui/components/` were previously excluded from TypeScript compilation. This PR removes those exclusions to ensure type safety. The fixes involved:
- Updating mock objects to match current interface definitions (e.g., `SessionMetrics`, `TextBufferState`).
- Importing missing types and enums (e.g., `ToolCallDecision`).
- Fixing type errors related to optional properties and null/undefined values.
- Updating snapshots in `ToolStatsDisplay.test.tsx` and `SessionSummaryDisplay.test.tsx` to reflect changes in mock data.

## Reviewer Test Plan

Run `npm run typecheck` in `packages/cli` to verify there are no errors.
Run `npm run test -w @google/gemini-cli` to ensure tests pass and snapshots are updated.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓ |
| npx      | ✅  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


packages/core/src/config/config.ts
jacob314, line 144: this is sus

packages/cli/src/config/extension.ts
jacob314, line 257: why is this added?

packages/cli/src/ui/components/__snapshots__/SessionSummaryDisplay.test.tsx.snap
```diff
@@ -6,7 +6,7 @@ exports[`<SessionSummaryDisplay /> > renders the summary display with a title 1`
 │  Agent powering down. Goodbye!                                                                   │
 │                                                                                                  │
 │  Interaction Summary                                                                             │
-│  Session ID:                                                                                     │
```
jacob314, line 9: nice!

packages/cli/src/ui/components/shared/text-buffer.test.ts
```diff
@@ -1736,12 +1755,22 @@ describe('textBufferReducer vim operations', () => {
         cursorRow: 1,
         cursorCol: 0,
         preferredCol: null,
-        visualLines: [['line1'], ['line2'], ['line3'], ['line4']],
-        visualScrollRow: 0,
-        visualCursor: { row: 1, col: 0 },
-        viewport: { width: 10, height: 5 },
+        visualLayout: {
```
jacob314: we should refactor these tests to not require this. specifying the visualLayout and all this in the test has lost the thread. These should get computed from the initial state  or be empty rather than get filled in.
I care because otherwise these tests are overly verbose and painful to maintain.

packages/cli/src/ui/components/shared/vim-buffer-actions.test.ts
```diff
@@ -840,7 +850,8 @@ describe('vim-buffer-actions', () => {
         const state = createTestState(['line1', 'line2', 'line3'], 0, 2);
         const action = {
           type: 'vim_change_movement' as const,
-          payload: { movement: 'j', count: 2 },
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+          payload: { movement: 'j' as any, count: 2 },
```
jacob314: don't use any here. fix the code.

packages/cli/src/ui/components/shared/vim-buffer-actions.test.ts
```diff
@@ -814,7 +822,8 @@ describe('vim-buffer-actions', () => {
         const state = createTestState(['hello world'], 0, 5);
         const action = {
           type: 'vim_change_movement' as const,
-          payload: { movement: 'h', count: 2 },
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
jacob314: here and elsewhere in this file don't use any, fix the code.

packages/cli/src/ui/components/StatsDisplay.test.tsx
jacob314: as a follow up pr would be nice to reduce duplicated boilerplate in this test file. seems like there are tons of near identical sessionMetrics blocks.

packages/cli/tsconfig.json
jacob314, line 56: yay!

jacob314 (commented):
approved once these comments are addressed. thanks so much for this cleanup to get the tests back in a better state!

packages/cli/src/ui/components/StatsDisplay.test.tsx
SandyTao520: Done in this PR.

packages/cli/src/ui/components/shared/text-buffer.test.ts
SandyTao520: done

packages/cli/src/ui/components/shared/vim-buffer-actions.test.ts
```diff
@@ -836,8 +835,7 @@ describe('vim-buffer-actions', () => {
         const state = createTestState(['hello world'], 0, 5);
         const action = {
           type: 'vim_change_movement' as const,
-          // eslint-disable-next-line @typescript-eslint/no-explicit-any
-          payload: { movement: 'l' as any, count: 3 },
+          payload: { movement: 'l' as const, count: 3 },
```
jacob314, line 838: learned something new in typescript today.

jacob314 (approved):
lgtm

PR #11478, opened 2025-10-19
(fix): Respect ctrl+c signal for aborting execution in NonInteractive mode

### LINKED ISSUE #11479: ctrl+c was not respected in non interactive mode
Using ctrl+c to cancel a non interactive session was broken and output continued to stream onto the terminal post multiple tries.

anj-s:
## TLDR

Fixes non-interactive mode to properly handle Ctrl+C for graceful cancellation. Previously, Ctrl+C was not properly handled, preventing users from aborting long-running operations in non-interactive mode.

## Dive Deeper

This change adds proper Ctrl+C handling to the non-interactive CLI flow using stdin listening:

1. **stdin listener**: Sets up readline keypress detection when stdin is a TTY
2. **AbortController integration**: Uses the existing AbortController to propagate cancellation through the stream processing pipeline
3. **Graceful shutdown**: Ctrl+C triggers abort with delayed "Cancelling..." message (200ms delay to reduce verbosity for fast operations)
4. **Proper cleanup**: Restores stdin state and removes listeners in the finally block
5. **Exit code**: Uses standard Unix exit code 130 for Ctrl+C cancellation
6. **Cross-platform**: Works consistently on Linux, macOS, and Windows (unlike signal handlers which have platform-specific issues)

The implementation uses stdin keypress detection instead of signal handlers, which provides better cross-platform compatibility and consistency with the interactive mode approach.

## Reviewer Test Plan

To validate this change:

1. **Test graceful cancellation**
   ```bash
   gemini "write a very long story"
   # Press Ctrl+C during execution
   # Should see "Cancelling..." (if > 200ms) and exit with code 130
   ```

2. **Test fast cancellation (no message)**
   ```bash
   gemini "say hello"
   # Press Ctrl+C quickly
   # Should exit cleanly without "Cancelling..." message
   ```

3. **Verify exit code**
   ```bash
   gemini "test query"
   # Press Ctrl+C
   echo $?  # Should be 130
   ```

4. **Test with piped input (should not activate stdin listener)**
   ```bash
   echo "write a story" | gemini
   # Ctrl+C should work normally (won't use our stdin listener since stdin is not a TTY)
   ```

5. **Run existing test suite** (signal handling tests removed, other tests verify AbortSignal functionality)
   ```bash
   cd packages/cli
   npm test -- nonInteractiveCli.test.ts
   ```

Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅   | ❓   | ❓   |

## Linked issues / bugs
fixes https://github.com/google-gemini/gemini-cli/issues/11479


packages/cli/src/nonInteractiveCli.test.ts
```diff
@@ -1008,4 +1008,265 @@ describe('runNonInteractive', () => {
     );
     expect(processStdoutSpy).toHaveBeenCalledWith('file.txt');
   });
+
+  describe('signal handling', () => {
+    it('should register SIGINT and SIGTERM handlers', async () => {
```
jacob314: this is worth an integration test as well. otherwise I would be impressed if it all just works seamlessly on windows.

packages/cli/src/nonInteractiveCli.ts
```diff
@@ -55,6 +55,30 @@ export async function runNonInteractive(
         ? new StreamJsonFormatter()
         : null;
 
+    const abortController = new AbortController();
+
+    // Track if we're in the process of aborting to handle multiple signals
```
jacob314: the the interactive CLI we listen on stdin to detect ctrl-C. Would that approach be feasible once we are done reading the prompt from stdin? The main advantage of listening on stdin is I believe the same code is more likely to work on both Linux and Windows.

packages/cli/src/nonInteractiveCli.ts
```diff
@@ -55,6 +55,30 @@ export async function runNonInteractive(
...
+    // Signal handler for graceful cancellation
+    const handleSignal = (signal: NodeJS.Signals) => {
+      if (isAborting) {
+        // Second signal: force exit immediately
+        process.stderr.write('\nForced exit\n');
+        process.exit(signal === 'SIGINT' ? 130 : 143);
+      }
+
+      isAborting = true;
+      process.stderr.write('\nCancelling...\n');
```
jacob314: I'd err on not showing this message or only showing this message if it has been 200ms and the terminal still hasn't cancelled.

jacob314: Image
output looks a bit verbose showing ^C, "Cancelling" and "Operation cancelled." in the output.

packages/cli/src/nonInteractiveCli.ts
jacob314: thought about this more and it is probably a bad idea. this logic is probably fine for the non-interactive mode unless we wanted to have the same behavior of forcing you to press ctrl-C twice to quit which I don't think would be useful in the non-interactive case.

jacob314 (commented):
lgtm
Approved after these minor concerns are addressed. Tested on Mac and it works well.

packages/cli/src/nonInteractiveCli.test.ts
anj-s: Will be addressed in a follow up PR

jacob314 (approved):
lgtm

Issue #11438, opened 2025-10-18
The ripgrep tool should prefer the system ripgrep, when possible.


albertosantini:
### What happened?

ripgrep package is downloaded, even if it is already downloaded (and in the path).

### What did you expect to happen?

If ripgrep is already downloaded, we may avoid downloading again.
Furthermore, if ripgrep is in the PATH, we may avoid downloading at all.

### Client information

<details>
<summary>Client Information</summary>

Run `gemini` to enter the interactive CLI, then run the `/about` command.

```console
> /about

╭──────────────────────────────────────────────────╮
│                                                  │
│ About Gemini CLI                                 │
│                                                  │
│ CLI Version      0.9.0                           │
│ Git Commit       a93d92a3                        │
│ Model            gemini-2.5-pro                  │
│ Sandbox          no sandbox                      │
│ OS               win32                           │
│ Auth Method      OAuth                           │
│                                                  │
╰──────────────────────────────────────────────────╯

```

</details>

albertosantini:
See https://github.com/google-gemini/gemini-cli/blob/dd42893df77be502826e1c575b9d51a5bdca3b4f/packages/core/src/tools/ripGrep.ts#L27-L37

`if (await fileExists(getRgPath())) {`
In Windows this always returns `false` because `getRgPath` is looking for `rg` and not `rg.exe`.
`return path.join(Storage.getGlobalBinDir(), 'rg');`

At the moment `vscode-ripgrep` is always downloaded, slowing down startup time, then it is unzipped and the executable moved to `.gemini/tmp/bin`.

I suggest one fix and one change:
- improving `getRgPath` for Windows environment
- changing the logic, if ripgrep executable is in the PATH, no download.

What do you think about it?

Edit: Anyway it is downloaded an old version of ripgrep `13.x`. Need to update to `14.x`

cornmander:
Thanks, those are both great ideas.  I will file feature requests and get these prioritized.

cornmander:
Re-marking this as an enhancement since the bug part should be fixed.

joshualitt:
Disowning this for now, until https://github.com/google-gemini/gemini-cli/issues/12099 is fixed. I have a PR to prefer ripgrep from the system path(https://github.com/google-gemini/gemini-cli/pull/11847), but it makes sense to pull in checksum verification logic into the same CL to ensure the integrity of the ripgrep binary before use, and this will be painful without addressing #12099 first.

[...]

Issue #11258, opened 2025-10-16
Function downloadRipGrep doesn't use provided argument


Smetalo:
### What happened?

Function _downloadRipGrep_ is called with a destination directory argument (see _packages/core/src/tools/ripGrep.ts_), but its signature (_async () =>_) doesn't accept it.

Since the argument is ignored, _ripgrep_ is installed to a hardcoded location instead of the provided directory _Storage.getGlobalBinDir()_. 

Function signature should be updated:
_export const downloadRipGrep = async (binPath) => {_

Calls to _untarGz_ and _unzip_ should be updated by replacing _BIN_PATH_ with _binPath_:
_await untarGz(downloadPath, binPath)_
_await unzip(downloadPath, binPath)_

It should be checked whether this causes _canUseRipgrep_ to fail on all platforms that require downloading _ripgrep_, as it won't find the binary where it is looking for it.

### What did you expect to happen?

-

### Client information

-

cornmander:
Fixed in #11492.

PR #11492, opened 2025-10-20
Fix Windows ripgrep detection

cornmander:
- detect both `rg` and `rg.exe`, returning whichever exists and downloading only once even with
  concurrent callers
  - refactor ripgrep availability logic into a shared helper and reuse it from `canUseRipgrep`/
  `ensureRgPath`
  - expand the test suite with Windows-specific and concurrency regressions, and ignore temporary
  `download-ripgrep*` artifacts

  Related to #11438

### DIFF
--- a/packages/core/src/tools/ripGrep.test.ts
+++ b/packages/core/src/tools/ripGrep.test.ts
@@ -19,6 +19,7 @@ import path from 'node:path';
 import fs from 'node:fs/promises';
 import os, { EOL } from 'node:os';
 import type { Config } from '../config/config.js';
+import { Storage } from '../config/storage.js';
 import { createMockWorkspaceContext } from '../test-utils/mockWorkspaceContext.js';
 import type { ChildProcess } from 'node:child_process';
 import { spawn } from 'node:child_process';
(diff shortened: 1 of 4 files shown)


packages/core/src/tools/ripGrep.ts
```diff
@@ -21,28 +21,55 @@ import { GREP_TOOL_NAME } from './tool-names.js';
...
+
+async function ensureRipgrepAvailable(): Promise<string | null> {
+  const existingPath = await resolveExistingRgPath();
+  if (existingPath) {
+    return existingPath;
+  }
+  if (!ripgrepAcquisitionPromise) {
+    ripgrepAcquisitionPromise = (async () => {
+      try {
+        await downloadRipGrep(Storage.getGlobalBinDir());
```
Smetalo, line 49: Please check #11258

Smetalo:
According to [this comment](https://github.com/google-gemini/gemini-cli/issues/11438#issuecomment-3417919920), will v14 start to be used instead of v13 as a part of this PR?

cornmander:
> According to [this comment](https://github.com/google-gemini/gemini-cli/issues/11438#issuecomment-3417919920), will v14 start to be used instead of v13 as a part of this PR?

Let's address this separately.

packages/core/src/tools/ripGrep.test.ts
```diff
@@ -135,9 +213,39 @@ describe('ensureRgPath', () => {
...
+
+  it.runIf(process.platform === 'win32')(
+    'should detect ripgrep when only rg.exe exists on Windows',
+    async () => {
+      const windowsBinDir = String.raw`C:\mock\bin\dir`;
+      const storageBinDirMock = Storage.getGlobalBinDir as Mock;
+      const defaultBinDir = '/mock/bin/dir';
+
+      storageBinDirMock.mockReturnValue(windowsBinDir);
+      (fileExists as Mock).mockImplementation(async (filePath: string) =>
```
jacob314: don't mock the file system. specify a temp bin dir location instead and use the real fs.

jacob314 (commented):
approved once this comment is addressed

packages/core/src/tools/ripGrep.ts
cornmander: done

packages/core/src/tools/ripGrep.test.ts
cornmander: done

cornmander:
> approved once this comment is addressed

done

jacob314 (approved):
lgtm

PR #11520, opened 2025-10-20
fix(integration-tests): Set correct permissions for test directories

skeshive:
## TLDR
This PR fixes an `EACCES: permission denied` error that can occur during integration tests, especially in CI environments. It ensures that the test run directories are created with appropriate write permissions.


jacob314 (approved):
lgtm

PR #11523, opened 2025-10-20
feat: Implement message bus and policy engine

allenhutchison:
feat: Implement message bus and policy engine

This commit introduces a message bus and policy engine to centralize tool confirmation and policy decisions.

The message bus acts as a central communication hub, allowing different parts of the application to communicate asynchronously. The policy engine is used to make decisions about whether to allow, deny, or ask for user confirmation for tool executions.

The following tools have been updated to use the message bus for confirmation:
- glob
- grep / ripGrep
- ls
- read-file
- read-many-files
- web-fetch
- web-search
- edit
- shell
- write-file
- memory

The `smart-edit` and `write-todos` tools have not yet been integrated with the new system and will be addressed in a future PR.


allenhutchison:
This addresses most of #11297 but there will be one follow up PR to get the last few tools.

packages/core/src/tools/tools.ts
```diff
@@ -76,10 +76,14 @@ export abstract class BaseToolInvocation<
   constructor(
     readonly params: TParams,
     protected readonly messageBus?: MessageBus,
+    readonly _toolName?: string,
+    readonly _toolDisplayName?: string,
   ) {
     if (this.messageBus) {
```
abhipatel12: Do we still need this debug log?

packages/cli/src/config/policy.ts
```diff
@@ -179,3 +181,21 @@ export function createPolicyEngineConfig(
...
+) {
+  messageBus.subscribe(
+    MessageBusType.UPDATE_POLICY,
+    (message: UpdatePolicy) => {
+      const toolName = message.toolName;
+
+      policyEngine.addRule({
+        toolName,
+        decision: PolicyDecision.ALLOW,
+        priority: 199, // High priority, but lower than explicit DENY (200)
```
abhipatel12, line 198: Should we add this to the comment block above that has the other priority values currently?

abhipatel12 (approved):
LGTM!

Tested this locally as well

packages/cli/src/config/policy.ts
allenhutchison: Good point. Added.

packages/core/src/tools/tools.ts
allenhutchison: You're right. We probably don't need this one anymore. It's clear that the registration is working fine.

PR #11527, opened 2025-10-20
Fix shell auto-approval parsing for chained commands

### LINKED ISSUE #11510: Shell command allowlists do not apply to piped commands
### What happened?

When a shell command execution is approved, either through configuration like `tools.core` or command-line flags like `--allowed-tools`, the check is only performed against the first command in a pipe chain. Subsequent commands in the pipeline are executed without being checked against the allowlist. For instance, if `cat` is an allowed tool but `wc` is not, a command like `cat file.txt | wc -l` will still execute, including the un-allowed `wc` command.

I expected every command in a shell pipeline to be individually verified against the configured tool allowlist. If any command in the chain is not permitted, the entire command should be rejected.

Here is a set of test cases that demonstrate the issue.

### Test Case 1: `wc` by itself is correctly denied

When `wc` is not in the `--allowed-tools` list, the command is correctly denied.

```console
$ gemini "Count the lines in file.txt"
> Thinking...
> I will use the `wc` command to count the lines in the file.
> ShellTool(wc -l file.txt)
> DENIED. `ShellTool(wc)` is not on the allowlist.
```

### Test Case 2: `wc` by itself is correctly allowed

When `wc` is not in the `--allowed-tools` list, the command is correctly denied.

```console
$ gemini --allowed-tools="ShellTool(wc)" "Count the lines in file.txt"
> Thinking...
> I will use the `wc` command to count the lines in the file.
> ShellTool(wc -l file.txt)
> 42 lines
```

### Test Case 3: How it should work

The command should only work as expected when both commands in the pipe are explicitly added to the allowlist.

```console
$ gemini --allowed-tools="ShellTool(cat,wc)" "Pipe cat file.txt into wc and return the lines"
> Thinking...
> I will use the `cat` command to view the contents of the file and pipe it to `wc` to count the lines.
> ShellTool(cat file.txt | wc -l)
> 42
```

This demonstrates that the Gemini CLI is capable of understanding multi-part commands, but fails to enforce the allowlist on all parts of the command.

### Test Case 4: Piped `wc` is incorrectly allowed

When `cat` is allowed but `wc` is not, a piped command still executes, bypassing the allowlist for the second command.

```console
$ gemini --allowed-tools="ShellTool(cat)" "Pipe cat file.txt into wc and return the lines"
> Thinking...
> I will use the `cat` command to view the contents of the file and pipe it to `wc` to count the lines.
> ShellTool(cat file.txt | wc -l)
> 42
```

**NOTE:** *This should have been denied because `wc` is not in the allowlist.*

### Client information

<details>

```console
> /about

╭──────────────────────────────────────────────────────────────────────────────────╮
│ CLI Version                 0.11.0-nightly.20251015.203bad7c-git.2ef38065        │
│ Git Commit                  2ef38065                                             │
│ Model                       auto                                                 │
│ OS                          linux                                                │
│ Auth Method                 gemini-api-key                                       │
╰──────────────────────────────────────────────────────────────────────────────────╯
```

</details>

### Login information

API key

This behavior undermines the efforts to create a safer, more secure execution environment for the Gemini CLI. Moving away from allow-by-default options e.g. `--yolo` to an allow-list based approach to shell command use can reduce the risk of shell injection and limit the blast radius of malicious or erroneous commands. By not checking all commands in a pipeline, the current implementation leaves a significant security gap.

cornmander:
## Summary
- reuse shell-utils parser when deciding auto-approval for shell invocations
- ensure chained commands require every segment to be allowed before skipping approval
- add helper unit coverage and scheduler regression test

Fixes #11510

### DIFF
--- a/packages/core/src/core/coreToolScheduler.test.ts
+++ b/packages/core/src/core/coreToolScheduler.test.ts
@@ -19,6 +19,7 @@ import type {
   ToolResult,
   Config,
   ToolRegistry,
+  AnyToolInvocation,
 } from '../index.js';
 import {
   DEFAULT_TRUNCATE_TOOL_OUTPUT_LINES,
@@ -37,6 +38,7 @@ import {
 } from '../test-utils/mock-tool.js';
 import * as fs from 'node:fs/promises';
 import * as path from 'node:path';
+import { isShellInvocationAllowlisted } from '../utils/shell-utils.js';
 
 vi.mock('fs/promises', () => ({
   writeFile: vi.fn(),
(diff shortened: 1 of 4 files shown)


packages/core/src/utils/shell-utils.test.ts
```diff
@@ -420,6 +422,39 @@ describe('stripShellWrapper', () => {
   });
 });
 
+describe('isShellInvocationAllowlisted', () => {
+  const createInvocation = (command: string) =>
```
adamfweidman: nit: define as func

packages/core/src/utils/shell-utils.test.ts
```diff
@@ -420,6 +422,39 @@ describe('stripShellWrapper', () => {
...
+    ({ params: { command } }) as unknown as AnyToolInvocation;
+
+  it('should return false when any chained command segment is not allowlisted', () => {
```
adamfweidman, line 430: can you add a specific test with a pipe

packages/core/src/core/coreToolScheduler.test.ts
```diff
@@ -1192,6 +1193,110 @@ describe('CoreToolScheduler request queueing', () => {
     }
   });
 
+  it('should require approval for a chained shell command even when prefix is allowlisted', async () => {
+    expect(
+      isShellInvocationAllowlisted(
+        {
+          params: { command: 'git status && rm -rf /tmp/should-not-run' },
+        } as unknown as AnyToolInvocation,
```
adamfweidman, line 1202: Should be ToolInvocation or needs to be imported, does not build locally

packages/core/src/utils/shell-utils.ts
```diff
@@ -738,3 +738,89 @@ export function isCommandAllowed(
...
+
+  if (commandsToValidate.length === 0) {
+    return false;
+  }
+
+  const shellInvocation = {
+    params: { command: '' },
+  } as AnyToolInvocation & { params: { command: string } };
+
+  for (const commandSegment of commandsToValidate) {
```
adamfweidman:
```
 return commandsToValidate.every((commandSegment) =>
    doesToolInvocationMatch(
      SHELL_TOOL_NAMES[0],
      { params: { command: commandSegment } } as AnyToolInvocation,
      allowedPatterns,
    ),
  );
```

packages/core/src/utils/shell-utils.ts
```diff
@@ -738,3 +738,89 @@ export function isCommandAllowed(
...
+    !('params' in invocation) ||
+    typeof invocation.params !== 'object' ||
+    invocation.params === null ||
+    !('command' in invocation.params)
+  ) {
+    return false;
+  }
+
+  const commandValue = (invocation.params as { command?: unknown }).command;
+  if (typeof commandValue !== 'string') {
```
adamfweidman:
```
  if (typeof command !== 'string' || !command.trim()) {
    return false;
  }
```

packages/core/src/utils/shell-utils.ts
```diff
@@ -738,3 +738,89 @@ export function isCommandAllowed(
...
+
+  if (!hasShellWildcard && !hasShellSpecificPattern) {
+    return false;
+  }
+
+  if (hasShellWildcard) {
+    return true;
+  }
+
+  if (
```
adamfweidman: what does this test?

packages/core/src/core/coreToolScheduler.ts
```diff
@@ -752,11 +756,24 @@ export class CoreToolScheduler {
             continue;
           }
 
+          const approvalMode = this.config.getApprovalMode();
           const allowedTools = this.config.getAllowedTools() || [];
-          if (
-            this.config.getApprovalMode() === ApprovalMode.YOLO ||
-            doesToolInvocationMatch(toolCall.tool, invocation, allowedTools)
-          ) {
+          const toolName =
```
adamfweidman: Can this be abstracted into helper?

```
1 private isAutoApproved(toolCall: ValidatingToolCall): boolean {
    3   if (this.config.getApprovalMode() === ApprovalMode.YOLO) {
    4     return true;
    5   }
    6 
    7   const allowedTools = this.config.getAllowedTools() || [];
    8   const { tool, invocation } = toolCall;
    9 
   11   const isShellTool = SHELL_TOOL_NAMES.includes(tool.name);
   12   if (isShellTool) {
   13     return isShellInvocationAllowlisted(invocation, allowedTools);
   14   } else {
   15     return doesToolInvocationMatch(tool, invocation, allowedTools);
   16   }
   17 }
```

adamfweidman (approved):

packages/core/src/utils/shell-utils.test.ts
cornmander: Converted the helper into a named function for clarity.

packages/core/src/utils/shell-utils.test.ts
cornmander: Added an explicit pipeline scenario so each segment must be allowlisted; it fails without the new parsing.

packages/core/src/core/coreToolScheduler.test.ts
cornmander: Imported `AnyToolInvocation` from the shared index so the test compiles under the package build.

packages/core/src/utils/shell-utils.ts
cornmander: Swapped the loop for `Array.every` and inline shell invocation, matches your suggestion.

packages/core/src/utils/shell-utils.ts
cornmander: Simplified the guard so we bail out when the command is missing or empty, covering both cases.

packages/core/src/utils/shell-utils.ts
cornmander: Retitled the case to spell out the wildcard behaviour so it reads clearly during review.

packages/core/src/core/coreToolScheduler.ts
cornmander: Pulled the auto-approval branching into a new helper so the main scheduling loop just checks the result.

jacob314 (approved):
lgtm

PR #11538, opened 2025-10-20
fix(proxy): Add error handling to proxy agent creation

### LINKED ISSUE #6333: CLI fails with TypeError: Invalid URL in PowerShell on Corporate Windows Machine
### What happened?

The CLI fails with a TypeError: Invalid URL in proxy-agent.js when launched from PowerShell. The issue appears to be that the underlying undici library incorrectly reads a system/registry proxy setting. The tool works correctly when launched from a cmd.exe batch file that unsets all proxy environment variables. This indicates the bug is specific to the interaction between the CLI and the PowerShell environment.

I expected the Gemini CLI to launch successfully within a PowerShell terminal. After explicitly unsetting all standard proxy environment variables (HTTP_PROXY, https_proxy, etc.), the CLI should be able to either establish a direct connection or gracefully ignore any misconfigured system-level proxy settings (like those in the Windows Registry).

The tool should not crash with a TypeError and should proceed to the first-time authentication prompt.

### Client information

 About Gemini CLI                                   │
│                                                    │
│ CLI Version       0.1.21                           │
│ Git Commit        055fe46d                         │
│ Model             gemini-2.5-pro                   │
│ Sandbox           no sandbox                       │
│ OS                win32                            │
│ Auth Method       OAuth  

### Login information

_No response_

_No response_

skeshive:
The CLI now gracefully handles invalid proxy environment variables, logging a warning instead of failing to start. This improves startup robustness.

## Linked Issues
Fixes #6333


packages/core/src/utils/fetch.ts
```diff
@@ -55,3 +56,11 @@ export async function fetchWithTimeout(
     clearTimeout(timeoutId);
   }
 }
+
+export function setGlobalProxy(proxy: string) {
+  try {
+    setGlobalDispatcher(new ProxyAgent(proxy));
+  } catch (e) {
+    console.error(`Failed to set proxy: ${getErrorMessage(e)}`);
```
jacob314, line 64: Fyi @abhipatel12 this will be a tricky case to handle for your refactor to remove the debug console. I suspect we need to show a more user friendly message than this if we are showing this message directly in the conversation history rather than the console.

jacob314 (approved):
lgtm

PR #11544, opened 2025-10-21
feat(core): Introduce message bus for tool execution confirmation

### LINKED ISSUE #11297: Integrate message bus and policy engine into each tool call.

allenhutchison:
## TLDR

This pull request introduces a message bus and policy engine to centralize and manage tool execution confirmation. This provides a flexible and extensible way to control how and when tools are allowed to run, moving confirmation logic out of individual tools and into a configurable, central system.

## Dive Deeper

Currently, tools that perform potentially destructive actions (e.g., `shell`, `writeFile`) implement their own confirmation prompts. This creates duplicated logic and makes it difficult to enforce consistent policies or to automate tool execution safely.

This PR refactors this flow by introducing:
1.  A **Message Bus** (`packages/core/src/confirmation-bus/message-bus.ts`): A central bus where tools can send confirmation requests.
2.  A **Policy Engine** (`packages/cli/src/config/policy.ts`): A configurable engine that evaluates tool calls against a set of rules.

The new workflow is as follows:
1. A tool invocation sends a request to the message bus.
2. The bus consults the policy engine.
3. The policy engine returns a decision: `ALLOW`, `DENY`, or `ASK_USER`.
4. If the decision is `ASK_USER`, the CLI falls back to the interactive confirmation prompt.

This architecture makes it possible to implement different security models, such as automatically denying certain commands or allowing all commands from a trusted source, without modifying the tools themselves.

Key changes include:
- Core implementation of the message bus and policy types.
- Integration of the message bus into `shell`, `edit`, `smart-edit`, `writeFile`, and `memoryTool`.

## Reviewer Test Plan

1. Pull down the branch and run `npm install`.
2. Start the CLI (`npm start`).
3. Execute a command that requires confirmation, for example: `rm -rf /tmp/some-test-dir`. You should see the standard confirmation prompt.

5. Verify that existing tests pass and that new tests for the message bus and policy engine are adequate.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ✅  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes #11297


packages/core/src/tools/shell.ts
```diff
@@ -77,8 +79,20 @@ export class ShellToolInvocation extends BaseToolInvocation<
...
+    abortSignal: AbortSignal,
   ): Promise<ToolCallConfirmationDetails | false> {
+    // Try message bus confirmation first if available
+    if (this.messageBus) {
+      const decision = await this.getMessageBusDecision(abortSignal);
+      if (decision === 'ALLOW') {
+        return false; // No confirmation needed
+      }
+      if (decision === 'DENY') {
+        throw new Error('Tool execution denied by policy.');
```
abhipatel12: Should we include the tool name like we do in the other destructive tools that throw?

packages/core/src/tools/edit.ts
```diff
@@ -101,13 +107,21 @@ interface CalculatedEdit {
...
+  extends BaseToolInvocation<EditToolParams, ToolResult>
+  implements ToolInvocation<EditToolParams, ToolResult>
+{
   constructor(
     private readonly config: Config,
-    public params: EditToolParams,
-  ) {}
+    params: EditToolParams,
+    messageBus?: MessageBus,
+    toolName?: string,
```
abhipatel12, line 119: Should we still use `_toolName` and `_displayName`?

abhipatel12 (commented):
When i run in debug mode and try replace, i see the following:

```shell
[DEBUG] Tool replace created with messageBus: YES
[MESSAGE_BUS] publish: {"type":"tool-confirmation-request","toolCall":{"na}
[MESSAGE_BUS] publish: {"type":"tool-confirmation-response","correlationId}
... // The response repeats over 40 times
```

Do you know why there would be so many response messages recorded?

packages/core/src/tools/tools.ts
```diff
@@ -90,11 +94,43 @@ export abstract class BaseToolInvocation<
...
+                });
+              }
+            }
+          },
+        };
+        return confirmationDetails;
+      }
+    }
+    return false;
   }
```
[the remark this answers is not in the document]
allenhutchison: This is a great suggestion. I'll incorporate this into this PR.

allenhutchison:
It looks like the multiple message thing is caused by #11753 . I've mitigated the issue for the message bus by explicitly using a singleton for the message bus.

packages/core/src/tools/edit.ts
allenhutchison: I'm ambivalent. On the one hand it would make it consistent, but we acutally use these variables here so they don't need to be prepended with _.

packages/core/src/tools/shell.ts
allenhutchison: Yes, this is now handled universally by the base class.

packages/core/src/tools/tools.ts
```diff
@@ -104,25 +104,37 @@ export abstract class BaseToolInvocation<
...
+    // When no message bus, use default confirmation flow
+    return this.getConfirmationDetails(abortSignal);
+  }
+
+  /**
+   * Subclasses should override this method to provide custom confirmation UI
+   * when the policy engine's decision is 'ASK_USER'.
+   * The base implementation provides a generic confirmation prompt.
+   */
+  protected async getConfirmationDetails(
```
abhipatel12, line 119: Really like this!

abhipatel12 (approved):
LGTM! 

Not related to this PR, but what do you think of having a tool hold its own definition of `Kind.READ_ONLY` vs `Kind.Destructive`. 

Seems like there's a conflict, but approved otherwise!

Issue #10761, opened 2025-10-08
Change the debug console to be a development only feature


jacob314:
Errors should be reported directly to the history stream like we surface `/help` and similar.
This reduces flicker and declutters the UI. A common source of flicker is users open the debug console and expand it to see more messages resulting in the rest of the UI to flicker. This is also needed to free up the `ctrl-O` shortcut to expand the last message.

abhipatel12:
Marking this as blocked until we have UX feedback from Miguel on if we need to show users complex errors in a tmp file, in history, etc.

abhipatel12:
Moving to another high priority issue. I have filed the necessary subtasks remaining before we can consider this task complete.

PR #11590, opened 2025-10-21
refactor(logging): Centralize console logging with debugLogger

abhipatel12:
## TLDR

This PR centralizes all `console.log`, `console.debug`, and `console.warn` calls within the `a2a-server`, `cli`, and `core` packages to use a new `debugLogger` utility. This change aims to provide a single point of control for logging behavior, making it easier to manage, filter, or extend logging capabilities in the future.

The main goal in this PR is not introduce any new behaviors.

## Reviewer Test Plan

Tests should all pass and no unexpected logging should happen.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  |✅| ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt |✅| -   | -   |

## Linked issues / bugs

This PR makes progress on #10761

### DIFF
--- a/packages/cli/src/ui/utils/updateCheck.ts
+++ b/packages/cli/src/ui/utils/updateCheck.ts
@@ -9,6 +9,7 @@ import updateNotifier from 'update-notifier';
 import semver from 'semver';
 import { getPackageJson } from '../../utils/package.js';
 import type { LoadedSettings } from '../../config/settings.js';
+import { debugLogger } from '@google/gemini-cli-core';
 
 export const FETCH_TIMEOUT_MS = 2000;
 
@@ -101,7 +102,7 @@ export async function checkForUpdates(
 
     return null;
   } catch (e) {
-    console.warn('Failed to check for updates: ' + e);
+    debugLogger.warn('Failed to check for updates: ' + e);
     return null;
   }
 }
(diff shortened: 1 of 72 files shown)


packages/cli/src/ui/utils/updateCheck.ts
```diff
@@ -101,7 +102,7 @@ export async function checkForUpdates(
 
     return null;
   } catch (e) {
-    console.warn('Failed to check for updates: ' + e);
+    debugLogger.warn('Failed to check for updates: ' + e);
```
jacob314, line 105: For discussion: should this one be an error. Fyi @galz10

packages/cli/src/validateNonInterActiveAuth.test.ts
```diff
@@ -53,10 +53,7 @@ describe('validateNonInterActiveAuth', () => {
         throw new Error(`process.exit(${code}) called`);
       });
     vi.spyOn(auth, 'validateAuthMethod').mockReturnValue(null);
-    refreshAuthMock = vi.fn().mockImplementation(async () => {
-      console.log('DEBUG: refreshAuthMock called');
```
jacob314, line 57: nice job catching and removing this spam.

abhipatel12: Thanks!

packages/cli/src/ui/utils/updateCheck.ts
galdawave: I think when we fail to update we add a message to the chat history, so i think this is fine

jacob314 (approved):
lgtm

PR #11592, opened 2025-10-21
Create BYOID auth client when detecting BYOID credentials

cocosheng-g:
## TLDR

This PR fixes BYOID authentication by routing token requests to the correct STS endpoint instead of the standard Google OAuth endpoint.

## Dive Deeper

The core issue is a mismatch between the credential type and the token endpoint being used. The logs show that the credential type is external_account_authorized_user, which is used for workforce identity federation (BYOID). This type of credential requires token requests to be sent to the Security Token Service (STS) endpoint (https://sts.googleapis.com/v1/oauthtoken).

However, the error message indicates that the request is being sent to the standard Google OAuth 2.0 endpoint (https://oauth2.googleapis.com/token). The google-auth-library is incorrectly attempting a standard refresh token flow instead of the STS token exchange required for BYOID.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


scidomino (approved):

cocosheng-g:
Tested both BYOID and google auth for agent mode interactions in GCA VSCode.

packages/core/src/code_assist/oauth2.ts
```diff
@@ -225,6 +226,20 @@ export async function getOauthClient(
...
 }
 
+export async function getAuthClient(
+  authType: AuthType,
+  config: Config,
+): Promise<AuthClient> {
+  // Prioritize BYOID credentials if they exist.
+  const byoidClient = await getByoidClient(config);
+  if (byoidClient) {
+    console.log('Loaded BYOID credentials.');
```
scidomino: Leave out console.log for now. We're in the process of moving to a new system.

packages/core/src/code_assist/oauth2.ts
```diff
@@ -225,6 +226,20 @@ export async function getOauthClient(
...
+    return byoidClient;
+  } else {
```
scidomino: extreme nit: don't bother with the "else". Just return.

packages/core/src/code_assist/oauth2.ts
```diff
@@ -432,6 +447,69 @@ export function getAvailablePort(): Promise<number> {
   });
 }
 
+async function findAndLoadCredentials(): Promise<{
```
scidomino: It looks like that in the normal non byoid case, we end up calling this twice. Once for byoid and once for regular. Seems wasteful. Especially since the majority of our users will run into this. Can we attempt to load the credentials once, and then pass them to the methods that need them?

cocosheng-g: Promise should save the results from first run and return when it's called again

packages/core/src/code_assist/oauth2.ts
```diff
@@ -432,15 +437,17 @@ export function getAvailablePort(): Promise<number> {
   });
 }
 
-async function loadCachedCredentials(client: OAuth2Client): Promise<boolean> {
+async function loadCachedCredentials(
```
scidomino: This structure is really confusing. We're creating a OAuth2Client which we pass into this method and now the method might throw it away and return a completely different AuthClient? Can't we avoid creating the OAuth2Client in the first place?

cocosheng-g: refactored.

packages/core/src/code_assist/oauth2.ts
```diff
@@ -450,19 +483,9 @@ async function loadCachedCredentials(client: OAuth2Client): Promise<boolean> {
...
-      if (!token) {
-        continue;
-      }
-
-      // This will check with the server to see if it hasn't been revoked.
-      await client.getTokenInfo(token);
-
-      return true;
+      const credsStr = await fs.readFile(keyFile, 'utf-8');
+      const creds = JSON.parse(credsStr);
```
scidomino: nit: inline this var

packages/core/src/code_assist/oauth2.ts
```diff
@@ -73,7 +74,29 @@ function getUseEncryptedStorageFlag() {
...
+  if (
+    credentials &&
+    (credentials as { type?: string }).type ===
+      'external_account_authorized_user'
+  ) {
+    debugLogger.log('Loading BYOID external account credentials.');
+    const auth = new GoogleAuth({
+      scopes: OAUTH_SCOPE,
+    });
+    const externalAccountAuthorizedUserClient =
```
scidomino: nit: Pick a smaller variable name.

packages/core/src/code_assist/oauth2.ts
```diff
@@ -432,15 +470,10 @@ export function getAvailablePort(): Promise<number> {
   });
 }
 
-async function loadCachedCredentials(client: OAuth2Client): Promise<boolean> {
+async function loadCachedCredentials(): Promise<object | null> {
```
scidomino: We try to avoid "object". Could this be "Credentials | JWTInput | null"?

scidomino: This method no longer loads the cachedCredentials. It just fetches them so maybe rename this.

cocosheng-g: updated.

cocosheng-g: done

packages/core/src/code_assist/oauth2.ts
cocosheng-g: done

packages/core/src/code_assist/oauth2.ts
cocosheng-g: done

packages/core/src/code_assist/oauth2.ts
```diff
@@ -73,7 +74,29 @@ function getUseEncryptedStorageFlag() {
...
+    const auth = new GoogleAuth({
+      scopes: OAUTH_SCOPE,
+    });
+    const byoidClient = await auth.fromJSON({
+      ...credentials,
+      refresh_token: credentials.refresh_token ?? undefined,
+    });
+    const token = await byoidClient.getAccessToken();
+    if (token) {
+      debugLogger.log('Created BYOID auth client.');
```
scidomino: make this a debug level log.

packages/core/src/code_assist/oauth2.ts
```diff
@@ -73,7 +74,29 @@ function getUseEncryptedStorageFlag() {
...
-): Promise<OAuth2Client> {
+): Promise<AuthClient> {
+  const credentials = await fetchCachedCredentials();
+
+  if (
+    credentials &&
+    (credentials as { type?: string }).type ===
+      'external_account_authorized_user'
+  ) {
+    debugLogger.log('Loading BYOID external account credentials.');
```
scidomino: remove this log

packages/core/src/code_assist/oauth2.ts
```diff
@@ -450,19 +485,8 @@ async function loadCachedCredentials(client: OAuth2Client): Promise<boolean> {
...
-      const { token } = await client.getAccessToken();
-      if (!token) {
-        continue;
-      }
-
-      // This will check with the server to see if it hasn't been revoked.
-      await client.getTokenInfo(token);
-
-      return true;
+      const creds = JSON.parse(await fs.readFile(keyFile, 'utf-8'));
```
scidomino: Nit:

const keyFileString = await fs.readFile(keyFile, 'utf-8');
return JSON.parse(keyFileString);

scidomino (approved):

packages/core/src/code_assist/oauth2.ts
cocosheng-g: done

packages/core/src/code_assist/oauth2.ts
cocosheng-g: done

packages/core/src/code_assist/oauth2.ts
cocosheng-g: done

PR #11602, opened 2025-10-21
fix(scripts): add tsconfig linter to prevent adding files to the exclude list

### LINKED ISSUE #9403: Investigate and Re-enable Skipped Tests
## We have a lot of skipped tests!

I was tinkering with our release pipelines, and I observed that we have a somewhat substantial amount of skipped tests.  There's maybe 20-30 across the whole codebase, but it feels very harmful.

> Spicy Opinion: It's pretty rare that anyone actually goes back and looks at skipped tests once they are submitted.

Image

There are even more tests that aren't executed at all here: https://github.com/google-gemini/gemini-cli/blob/main/packages/cli/tsconfig.json

## Some notes:

- Some the tests might be skipped because the test is executing on `linux` and the test specifically targets `windows` or something.

## We should:

- [x] Fix everything that's COMPLETELY skipped in https://github.com/google-gemini/gemini-cli/blob/main/packages/cli/tsconfig.json
- [ ] Find all the skipped-but-executing tests and re-enable them or delete them if they are no longer relevant or useful.
   - Partial Searches (probably not conclusive):
     - https://github.com/search?q=repo%3Agoogle-gemini%2Fgemini-cli%20skip()%3B&type=code
     - https://github.com/search?q=repo%3Agoogle-gemini%2Fgemini-cli+it.skip&type=code
     - https://github.com/search?q=repo%3Agoogle-gemini%2Fgemini-cli+describe.skip&type=code
- [x] Consider having some sort of lint that prevents check in of skipped tests.

SandyTao520:
## TLDR
Adds a new linter to `scripts/lint.js` that validates `tsconfig.json` files in `packages/` to ensure they only exclude "node_modules" and "dist".

## Dive Deeper
This prevents accidental exclusion of files that should be part of the package, ensuring consistent build configurations across packages. It uses `git ls-files` to find relevant tsconfigs and strips comments before parsing to handle standard `tsconfig.json` formats.

## Reviewer Test Plan
Run `node scripts/lint.js --tsconfig` to verify it passes on the current codebase.
Try adding an invalid exclusion (e.g., `"temp-exclude"`) to `packages/cli/tsconfig.json` and verify the linter fails.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓ |
| npx      | ✅  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs
Fixes #9403

### DIFF
--- a/scripts/lint.js
+++ b/scripts/lint.js
@@ -263,18 +263,15 @@ export function runTSConfigLinter() {
 
   let files = [];
   try {
-    // Find all tsconfig.json files under packages/
-    files = execSync("git ls-files | grep 'packages/.*/tsconfig.json$'")
+    // Find all tsconfig.json files under packages/ using a git pathspec
+    files = execSync("git ls-files 'packages/**/tsconfig.json'")
       .toString()
       .trim()
       .split('\n')
       .filter(Boolean);
   } catch (e) {
-    // If grep finds nothing, it might exit with 1.
-    if (e.status !== 1) {
-      console.error('Error finding tsconfig.json files:', e.message);
-      process.exit(1);
-    }
+    console.error('Error finding tsconfig.json files:', e.message);
+    process.exit(1);
   }
 
   let hasError = false;
--- a/scripts/lint.js
+++ b/scripts/lint.js
@@ -251,6 +251,82 @@ export function runSensitiveKeywordLinter() {
   }
 }
 
+function stripJSONComments(json) {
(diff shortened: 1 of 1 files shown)


jacob314 (approved):
lgtm

PR #11607, opened 2025-10-21
Apply new style to Todos

### LINKED ISSUE #10756: Implement UI for the TODO tool
Create UI for the `useWriteTodos` setting which currently lacks any UI to help a user track the current status if they miss the messages from the history.
See  write-todos.ts
https://github.com/google-gemini/gemini-cli/blob/741b57ed061c767ed25777f39b9fe826aaa1bcbc/packages/core/src/tools/write-todos.ts#L14

UI should show a minimal list of what TODOs are in progress using a line or two above the input prompt to keep it visible to the user without introducing clutter.
We should use emojis such as check marks and `x` to indicate the current state of each todo (pending | in_progress | completed | cancelled).

scidomino:
## TLDR

Apply new Todos styles provided by @miguelsolorio 

Also: allow "cancelled" status since it's mentioned in the tool description.
Also: reorg unit tests.

## Deeper Dive

CleanShot 2025-10-21 at 15 13 23@2x

## Reviewer Test Plan

Add `"useWriteTodos": true,` to `settings.json` and ask:

> Please create a todo list with one item of each status. Flash model has difficulty with this so use Pro, please.

(The second sentence avoids a bug in the new auto model picker)

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | x  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes #10756

### DIFF
--- a/packages/cli/src/ui/components/Composer.tsx
+++ b/packages/cli/src/ui/components/Composer.tsx
@@ -66,6 +66,8 @@ export const Composer = () => {
 
       <QueuedMessageDisplay messageQueue={uiState.messageQueue} />
 
+      <TodoTray />
+
       <Box
         marginTop={1}
         justifyContent={
@@ -129,8 +131,6 @@ export const Composer = () => {
           </Box>
         </OverflowProvider>
       )}
-
-      <TodoTray />
 
       {uiState.isInputActive && (
         <InputPrompt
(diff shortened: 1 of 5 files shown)


bdmorgan (approved):

chrstnb (commented):
Consider adding e.g.,  aria-label="Todo" to the text components-- I'm not sure how well the emojis will be treated by the screen reader, so it's probably clearer to just say "Todo" at the top and then e.g.,
aria-label="(completed task) $completed task description" for the items in the list. the aria-label will replace whatever is inside of the text component when in screen reader view.

Issue #11316, opened 2025-10-16
GCA service inconsistent request headers and response content


gsquared94:
### What happened?

Some requests have modified UserAgent headers being assigned to requests made to the `cloudcode-pa.googleapis.com` service.
```
User-Agent: GeminiCLI/0.10.0-nightly.20251007.c195a9aa (linux; x64) google-api-nodejs-client/9.15.1,gzip(gfe)
```

See http://sherlog/_kIdZWoH4o6 (accessible to googlers). The response type and error messages being returned are also different 

```
Content-Type: application/json; charset=UTF-8
...
[
  {
    "error": {
      "code": 429,
      "message": "Quota exceeded for quota metric 'Gemini 2.5 Pro Requests' and limit 'Gemini 2.5 Pro Requests per day per user per tier' of service 'cloudcode-pa.googleapis.com' for consumer 'project_number:681255809395'.",
      "errors": [
        {
          "message": "Quota exceeded for quota metric 'Gemini 2.5 Pro Requests' and limit 'Gemini 2.5 Pro Requests per day per user per tier' of service 'cloudcode-pa.googleapis.com' for consumer 'project_number:681255809395'.",
          "domain": "global",
          "reason": "rateLimitExceeded",
[...]

PR #11624, opened 2025-10-21
fix: handle request retries and model fallback correctly

gsquared94:
Second attempt for #9407. This was reverted because it didn't handle different error format being returned by the GCA service when run in Dev mode vs Prod mode. Fixed in this PR, to handle both error formats. See https://github.com/google-gemini/gemini-cli/issues/11316 for more details on that.

---

## TLDR

This pull request refactors the handling of API quota errors to be more robust and reliable. Previously, we relied on string matching to detect specific quota error messages. This change introduces a structured error  parsing system that can correctly classify quota errors as either retryable (e.g., per-minute limits, small retry delays) or terminal (e.g., daily limits).

The retry logic has been updated to use this new classification, ensuring that we only trigger model fallbacks for terminal quota errors and use the server-suggested delay for retryable errors when available. 

##  Dive Deeper

The core of this change is a new error classification system:

   1. `parseGoogleApiError`: A new function in packages/core/src/errors.ts that can parse deeply nested and stringified Google API error responses, extracting the structured error details.
   2. `TerminalQuotaError` and `RetryableQuotaError`: New custom error classes to represent different types of quota issues.
   3. `classifyGoogleError`: A function that uses parseGoogleApiError to analyze 429 errors and wrap them in the appropriate custom error class.

The retryWithBackoff function has been refactored to:
   - Catch `TerminalQuotaError` to trigger the `onPersistent429` fallback mechanism.
   - Catch `RetryableQuotaError` and use the `retryDelayMs` from the error for a more accurate backoff period.

  The UI hook `useQuotaAndFallback` has also been updated to use `instanceof TerminalQuotaError` for its logic, simplifying the code and removing the need for the old isProQuotaExceededError and isGenericQuotaExceededError functions, which have been deleted.

### Using GCA auth

- Clone a large repo like https://github.com/kubernetes/kubernetes
- Create an OTA at http://rhea.
- Run `gemini-cli` from this branch, login with OTA oauth. 
- Start making queries like this to get it to exhaust quota:
  ```
   scour the filesystem and git commit history for any and all typos. Do not use any spell checking tools, just start loading files  individually and look at them yourself. Follow these exact instructions, and don't ask me for comfirmations.   
  ```

- On quota exhaustion you should see the model fallback error.
image

### Using Gemini API key

- Go to https://aistudio.google.com/app/api-keys and sign in with the same OTA account and create an API key. Use this method, since this API key will have very low limits.
- Run `gemini-cli` from this branch, use this Gemini API key for auth
- Start making queries like this to get it to exhaust quota:
  ```
   scour the filesystem and git commit history for any and all typos. Do not use any spell checking tools, just start loading files  individually and look at them yourself. Follow these exact instructions, and don't ask me for comfirmations.   
  ```
- Requests should appear to be taking a lot longer (due to retries from the 429 error code `retryInfo` objects) instead of failing immediately.
- On quota exhaustion you should see the parsed output error.

image

- Compare this with the `gemini` stable build output - retries a bunch of times, then fails with a messy error message
image

##  Testing Matrix

  |          | 🍏  | 🪟  | 🐧  |
  | npx      |  ❓  | ❓   | ❓   |

Fixes:
- #9248
- #5119
- #3510
- #1088
- #4186


packages/core/src/utils/googleQuotaErrors.ts
```diff
@@ -0,0 +1,192 @@
...
+        'autopush-cloudcode-pa.googleapis.com',
+      ];
+      if (validDomains.includes(errorInfo.domain)) {
+        if (errorInfo.reason === 'RATE_LIMIT_EXCEEDED') {
+          let delaySeconds = 10; // Default retry of 10s
+          if (retryInfo?.retryDelay) {
+            const parsedDelay = parseDurationInSeconds(retryInfo.retryDelay);
+            if (parsedDelay) {
+              delaySeconds = parsedDelay;
+            }
```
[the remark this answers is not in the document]
gsquared94: The server returning a retry of 0 seconds would be incorrect. Defaulting to 10 seconds would be better.

guidedways:
Does this also fix the issue in general where it _thinks_ it's exhausted the quota? Even simple things (with a Code Assist Standard account) such as 'build this feature ...' which may indeed take 10 to 15 minutes and several steps but WELL within 1000 requests for the day (maybe 100 requests) it immediately fails. 

Essentially gemini cli seems to become unusable when there's anything that involves a little extra work in my experience so far. Trying this with a newly upgraded account.

packages/core/src/utils/retry.ts
```diff
@@ -145,126 +145,72 @@ export async function retryWithBackoff<T>(
...
         throw error;
       }
 
-      const { delayDurationMs, errorStatus: delayErrorStatus } =
-        getDelayDurationAndStatus(error);
+      const errorStatus = getErrorStatus(error);
+      logRetryAttempt(attempt, error, errorStatus);
 
-      if (delayDurationMs > 0) {
-        // Respect Retry-After header if present and parsed
```
cornmander, line 247: Do we need to respect retry-after header?

gsquared94: Neither GCA service nor Gemini service returned `retry-after` header in my testing it. `retryInfo` seems to be the standard.

cornmander (approved):
Please get @bdmorgan to review as well

gsquared94:
> Please get @bdmorgan to review as well

Sounds good!

bdmorgan (approved):

PR #11633, opened 2025-10-22
bundling

### LINKED ISSUE #11518: Rework release/promotion flow to promote across dev->prod npm repositories
Currently, the `dev` repository is a place that we only push to when we're testing or tinkering.  We'd like to move to a state where we're using the `dev` repository as part of our day-to-day flow.

We'd like to have some sort of promotion flow that goes similar to this (in this order):

| env name | npm env | description|
|-|-|-|
| dev | github dev | Development environment (potentially built @ HEAD on every merge operation?) |
| ??? | github dev | (It might make sense to have some sort of RC candidate built here for nightly? not sure) |
| nightly | wombat prod | nightly release |
| preview | wombat prod | weekly-ish preview release |
| latest | wombat prod | stable mass consumed version |

It should be noted that:
- github dev npm environment uses packages in the `@google-gemini/**` scope
- wombat prod environment uses packages in the `@google/**` scope
- We use Github action environments (named dev and prod) to separate out variables/keys for dev vs prod npm envs.

There are potentially some complications that get in the way of promotion a binary across these dev/prod lines
- [ ] Currently, all of the dev builds actually use bundling to work around an issue where `@google/**` imports cannot be found.  We bundle those into the package as a bit of a hack.  This bundling is currently present in `~/.github/actions/publish-release.yml`
- [ ] For each deployment, we actually manually stamp version numbers and package names into the `package.json` (iirc, this is done in `npm run version`), so it won't be possible to promote the packed file asset across dev+prod lines because of the difference in scopes. We'll likely need to deploy from the consistent branch and build these assets for each environment.  It would be REALLY nice to actually use the same built asset, then we'd have some level of attestation that the build is what it says it is.  However, I'm not sure this is possible in this setup.
- [ ] Github workflows can only be run with a single environment (it can be specified in the action inputs, but then everything within that workflow runs within that same environment).  To really pull this off, we'll likely have to have workflows triggering workflows with specified environment args.  I could be wrong.

This will likely require some of the following
- [ ] Consistent bundling across dev + prod.

Opinion:
- [ ] Github workflows that trigger other workflows can be somewhat challenging to linearly debug.  It would be ideal to be able to do this without needing to trigger a workflow-from-within-a-workflow (e.g. using `gh workflow run` in an action.
- [ ] The release flows are already fairly complicated and call out to lots of complicated scripts. Whoever takes this work on should probably build a parallel release flow and then eventually cut over to that flow.  Trying to refactor to this inflight could be a real pain.

I wanted to jot down my thoughts here and determine if this task has value.

mattKorwel:
## TLDR

- This pr moves us to shipping our public binary in bundled form.
- This pr revamps our CI process into a more cohesive and streamlined approach

## Bundling

<details>
<summary>
The primary benefits of bundling for a command-line tool revolve around performance, reliability, and a better end-user experience (Expand for more info)
</summary>

### Why It's a Good Change for a CLI (Pros)

**1. 🚀 Faster Startup Performance:**
This is the most significant advantage for a CLI. When you run a Node.js application, the `require()` or `import` statements trigger a complex module resolution process that involves searching the file system for `node_modules` directories. This can result in hundreds or even thousands of file system I/O operations, which slows down the initial startup time.

*   **Unbundled:** `gemini -> require('libA') -> require('libB') -> ...` (many file lookups)
*   **Bundled:** `gemini` (one file is loaded into memory)

By consolidating all the necessary code into a single file, you reduce the startup to a single file read, making the CLI feel significantly more responsive to the user.

**2. 📦 Faster and More Reliable Installation:**
Users installing your CLI will have a much better experience.

*   **Fewer Files:** `npm install` has to download and write far fewer files to the user's disk. This is faster and reduces the chance of installation errors due to network hiccups or file system issues.
*   **Smaller Footprint:** Bundlers perform "tree-shaking," which means they analyze your code and exclude any unused functions or modules from your dependencies. This results in a smaller, optimized final artifact compared to shipping the entire `node_modules` directory.

**3. 🛡️ Increased Reliability and Consistency:**
A bundled application is self-contained and predictable.

*   **Dependency Isolation:** The bundle includes the *exact* versions of the dependencies that you tested against. This completely eliminates a whole class of "works on my machine" bugs where a user might have a different, incompatible version of a sub-dependency in their environment that could break your tool.
*   **Immunity to `node_modules` Issues:** Your CLI will no longer be affected by the user's local `node_modules` structure, global packages, or potential conflicts between different projects. It just runs.

**4. 🚚 Simplified Distribution:**
Having a single executable file makes distribution and packaging much simpler if you ever decide to, for example, package the CLI with a standalone Node.js binary (using a tool like `pkg`) or include it in a Docker image.

---

### The Downsides of Bundling (Cons)

While the benefits are substantial for a CLI, there are trade-offs to consider, mostly affecting the development and debugging workflow.

**1. 🐛 Debugging Complexity:**
The code that runs is not the code you wrote. It's been transpiled, minified, and combined.

*   **Obfuscated Stack Traces:** An error in the bundled file will point to a line number in a massive, often unreadable, file.
*   **Mitigation:** **Source Maps (`.js.map` files).** A bundler can generate source maps that map the bundled code back to your original source files. When you use these with developer tools, stack traces become readable again. You can choose to publish these for public debugging or keep them private for your team.

**2. 🩹 Difficult for End-Users to Patch:**
If a user discovers a bug in one of your dependencies, they can't easily use a tool like `patch-package` to fix it themselves, because the dependency code is baked into the bundle. For a CLI, this is generally an acceptable trade-off for the reliability it provides.

**3. 🏗️ Added Build-Step Complexity:**
Your development process now requires a build step. You can't just run the source code directly; you must first build the bundle. This adds a layer of complexity to your local development, testing, and CI/CD pipeline.

### Summary: Pros vs. Cons

| Aspect | ✅ Pros of Bundling | ❌ Cons of Bundling |
| :--- | :--- | :--- |
| **Performance** | **Drastically faster startup time** and quicker installation. | Minor build time overhead during development. |
| **Reliability** | **Isolates dependencies**, preventing conflicts and ensuring consistency. | Users cannot easily patch sub-dependencies. |
| **Size** | **Smaller total size** on disk due to tree-shaking. | The single file itself can be large. |
| **Debugging** | Simpler to distribute and analyze a single artifact. | **Difficult to debug** without source maps. |
| **Development** | Clear separation between source and distributable code. | **Requires a build step**, adding complexity. |

**Conclusion:**

For a library that other developers will include in their projects, bundling is often a bad idea because it leads to code duplication and conflicts.

However, for a standalone application like the **Gemini CLI**, the benefits of **performance and reliability for the end-user are paramount.** The faster startup time and insulation from environmental issues create a much more professional and robust tool. The downsides are primarily developer-facing and can be effectively managed with source maps and a solid build process. Your PR is making a very good trade-off.

</details>

### Artifacts

1. On each push to a pr or main, we will bundle our code and push it to github's npm package registry: https://github.com/google-gemini/gemini-cli/pkgs/npm/gemini-cli
2. On each push to a pr or main we will build the sandbox docker image and push it to githubs  container registry: https://github.com/google-gemini/gemini-cli/pkgs/container/gemini-cli%2Fsandbox

Both of those artifacts are then used in all remaining tests. Specifically end 2 end tests no longer build the code or the sandbox image (with in context of github actions, this should all still work locally).

### Merge queue Skipper

While i was hopeful that this would have a net positive impact on us, it hasn't materialized. It is currently adding a significant overhead in code complexity and maintenance. This PR removes it in the spirit of simplification. If we want to approach it again later we can, totally open to feedback here.  

### Environments

All the ci operations work against our `prod` environment. No changes will be made or packages pushed to `prod`. This is intended to help reduce noise, customer facing confusion, and churn on pakcage versions (which are onetime use on npmjs). We will follow this change with the ability to 'promote' a specific tested bundled package from `dev` to `prod` and ideally this would be our primmary deployment mechanism. All patches, releases etc will go do dev, we will validate and we can delete and reuse package numbers if needed. Then when we are comfortable those will be promoted to Prod. This will be a follow on change. 

### Release Channels

As of version 0.11 stable, and 0.12-preview the code is still non bundled. 0.13-nightly is bundled.


scripts/configure-registry.js
```diff
@@ -0,0 +1,147 @@
...
+  if (!fs.existsSync(npmrcPath)) {
+    console.log('No ~/.npmrc file found. Already configured for prod.');
+    return;
+  }
+
+  let npmrcContent = fs.readFileSync(npmrcPath, 'utf-8');
+  const lines = npmrcContent.split('\n');
+  const filteredLines = lines.filter(
+    (line) =>
+      !line.includes('npm.pkg.github.com') &&
```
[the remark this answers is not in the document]
mattKorwel: fixed

scripts/configure-registry.js
```diff
@@ -0,0 +1,147 @@
...
+    console.log('✅ Your ~/.npmrc file is already configured for prod.');
+    return;
+  }
+
+  fs.writeFileSync(npmrcPath, filteredLines.join('\n').trim() + '\n');
+  console.log(`✅ Successfully cleaned dev configuration from ~/.npmrc.`);
+  if (fs.existsSync(backupPath)) {
+    console.log(`Your original configuration was saved to ${backupPath}`);
+  }
+}
```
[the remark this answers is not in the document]
mattKorwel: fixed

.github/workflows/build-and-publish.yml
```diff
@@ -0,0 +1,77 @@
...
+        with:
+          node-version-file: '.nvmrc'
+          cache: 'npm'
+
+      - name: 'Install dependencies'
+        run: 'npm ci'
+
+      - name: 'Configure npm for GitHub Packages'
+        run: |
+          echo "@google-gemini:registry=https://npm.pkg.github.com/" > .npmrc
```
richieforeman: Please consider using my npmrc action, which sets this up.

.github/workflows/e2e.yml
```diff
@@ -72,16 +46,13 @@ jobs:
         with:
           ref: '${{ github.event.inputs.branch_ref || github.ref }}'
 
-      - name: 'Set up Node.js ${{ matrix.node-version }}'
-        uses: 'actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020' # ratchet:actions-node@v4
-        with:
-          node-version: '${{ matrix.node-version }}'
+      - name: 'Configure npm for GitHub Packages'
```
richieforeman: Please use my npmrc action so that we're consistent (and can more trivially/easily make changes)

https://github.com/google-gemini/gemini-cli/blob/main/.github/actions/setup-npmrc/action.yml

.github/workflows/e2e.yml
```diff
@@ -130,11 +100,13 @@ jobs:
...
-        run: 'npm ci'
+      - name: 'Configure npm for GitHub Packages'
+        run: |
+          echo "@google-gemini:registry=https://npm.pkg.github.com/" > .npmrc
+          echo "//npm.pkg.github.com/:_authToken=${{ secrets.GITHUB_TOKEN }}" >> .npmrc
 
-      - name: 'Build project'
-        run: 'npm run build'
+      - name: 'Install dependencies'
+        run: 'npm install @google-gemini/gemini-cli@${{ inputs.version }}'
```
richieforeman: Two notes:

- Pass inputs via ENV and then use the ENV variable
- This package name should be dynamic (using `vars.CLI_PACKAGE_NAME`)

.github/workflows/e2e.yml
```diff
@@ -200,12 +171,14 @@ jobs:
           npm config set registry https://registry.npmjs.org/
         shell: 'pwsh'
 
-      - name: 'Install dependencies'
-        run: 'npm ci'
+      - name: 'Configure npm for GitHub Packages'
+        run: |
```
richieforeman: Also here

https://github.com/google-gemini/gemini-cli/blob/main/.github/actions/setup-npmrc/action.yml

.github/workflows/e2e.yml
```diff
@@ -200,12 +171,14 @@ jobs:
...
+          echo "@google-gemini:registry=https://npm.pkg.github.com/" > .npmrc
+          echo "//npm.pkg.github.com/:_authToken=${{ secrets.GITHUB_TOKEN }}" >> .npmrc
         shell: 'pwsh'
 
-      - name: 'Build project'
-        run: 'npm run build'
+      - name: 'Install dependencies'
+        run: 'npm install @google-gemini/gemini-cli@${{ inputs.version }}'
```
richieforeman: Same here.

scripts/configure-registry.js
```diff
@@ -0,0 +1,143 @@
+/**
```
richieforeman, line 1: Opinion:

Given that dev is `@google-gemini/**` scope and prod is `@google/**`, does a script to switch and swap this even make sense?  Why not just have people create an npmrc and they use that forever?

.github/actions/publish-release/action.yml
```diff
@@ -250,6 +274,13 @@ runs:
         a2a-package-name: '${{ inputs.a2a-package-name }}'
         working-directory: '${{ inputs.working-directory }}'
 
+    - name: '🔬 Verify Bundled NPM release by version'
```
richieforeman: Consider using the verify action to do this?  It actually performs verification on the output.

.github/actions/publish-release/action.yml
```diff
@@ -119,34 +124,49 @@ runs:
...
 
+    - name: '📦 Publish Root Package'
+      if: "${{ inputs.use-bundle-release == 'true' }}"
+      working-directory: '${{ inputs.working-directory }}'
+      env:
+        NODE_AUTH_TOKEN: '${{ inputs.github-token }}'
+      shell: 'bash'
+      run: |
+        npm publish \
+          --dry-run="${{ inputs.dry-run }}" \
```
richieforeman: Rather than interpolate ${{inputs...}} directly in the run command, please: 

- Pass them via ENV
- Use the env vars.

> Why?  Interpolation directly in the run command is apparently vulnerable to command injection.  Greg S had a recent PR to refactor/adjust everything to this pattern, and I think we should keep it.

.github/actions/publish-release/action.yml
```diff
@@ -119,34 +124,49 @@ runs:
...
         node ${{ github.workspace }}/scripts/prepare-github-release.js
 
+    - name: '📦 Publish Root Package'
+      if: "${{ inputs.use-bundle-release == 'true' }}"
+      working-directory: '${{ inputs.working-directory }}'
+      env:
+        NODE_AUTH_TOKEN: '${{ inputs.github-token }}'
+      shell: 'bash'
+      run: |
+        npm publish \
```
richieforeman: You need to sync this CL, I added a couple additional commands to this to tidy up the "false" tag.

.github/actions/publish-release/action.yml
```diff
@@ -224,8 +248,8 @@ runs:
         npm dist-tag rm ${{ inputs.a2a-package-name }} false --silent
 
     - name: '🔬 Verify NPM release by version'
+      if: "${{ inputs.dry-run != 'true' && inputs.force-skip-tests != 'true' && inputs.use-bundle-release == 'false' }}"
```
richieforeman: Super super super petty nit: We should probably standardize across our actions on `== "false"` or `!= "true"`.  Swapping between them makes this challenging to read.

.github/actions/publish-release/action.yml
richieforeman: More commentary: Your command here is just execing and hoping for a non-error exit code.

The verify action actually does that _AND_ verifies the output was as expected.  Please prefer that shared action.

.github/actions/publish-release/action.yml
```diff
@@ -250,6 +274,13 @@ runs:
         a2a-package-name: '${{ inputs.a2a-package-name }}'
         working-directory: '${{ inputs.working-directory }}'
 
+    - name: '🔬 Verify Bundled NPM release by version'
+      if: "${{ inputs.dry-run != 'true' && inputs.force-skip-tests != 'true' && inputs.use-bundle-release == 'true' }}"
```
richieforeman: Similar commentary here about these boolean statements.

.github/workflows/build-and-publish.yml
```diff
@@ -0,0 +1,63 @@
...
+          NEW_VERSION="${CURRENT_VERSION}-ci.${{ github.run_number }}.${{ github.sha }}"
+          echo "version=${NEW_VERSION}" >> "$GITHUB_OUTPUT"
+          npm version --no-git-tag-version "${NEW_VERSION}"
+
+      - name: 'Publish to GitHub Packages'
+        env:
+          NODE_AUTH_TOKEN: '${{ secrets.GITHUB_TOKEN }}'
+        run: |
+          if [[ "${{ github.event_name }}" == "pull_request" ]]; then
+            npm publish --tag="pr-${{ github.event.pull_request.number }}"
```
richieforeman: Super nit: I'd be ideal if this was bundled into the behavior for publish-and-release action, potentially? That way there's sort of one happy path?

.github/workflows/build-and-publish.yml
```diff
@@ -0,0 +1,63 @@
...
+          else
+            npm publish --tag="ci"
```
richieforeman: Thoughts: This should probably use the 'change tag' action, that way we have common/shared paths for this type of thing.

.github/workflows/bundle-size.yml
```diff
@@ -0,0 +1,48 @@
...
+
+      - name: 'Set up Node.js'
+        uses: 'actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020' # ratchet:actions/setup-node@v4
+        with:
+          node-version-file: '.nvmrc'
+          cache: 'npm'
+
+      - name: 'Configure npm for GitHub Packages'
+        run: |
+          echo "@google-gemini:registry=https://npm.pkg.github.com/" > .npmrc
```
richieforeman: please use the setup npmrc shared action.

.github/workflows/bundle-size.yml
```diff
@@ -0,0 +1,48 @@
...
+          echo "//npm.pkg.github.com/:_authToken=${{ secrets.GITHUB_TOKEN }}" >> .npmrc
+
+      - name: 'Install dependencies'
+        run: 'npm install @google-gemini/gemini-cli@${{ inputs.version }}'
```
richieforeman: Avoid interpolating inputs in run commands (see comment above for further reasoning)

.github/workflows/e2e.yml
richieforeman: I think the paren here at the end is stray?

richieforeman: Parens around this statement are potentially unnecessary.

.github/workflows/orchestrator.yml
```diff
@@ -0,0 +1,71 @@
...
+
+  get-vars:
+    name: 'Get Environment Variables'
+    runs-on: 'ubuntu-latest'
+    environment: 'dev'
+    outputs:
+      cli-package-name: '${{ steps.get-vars.outputs.cli-package-name }}'
+    steps:
+      - id: 'get-vars'
+        run: 'echo ''cli-package-name=''''${{ vars.CLI_PACKAGE_NAME }}'''''' >> "$GITHUB_OUTPUT"'
```
richieforeman: Avoid interpolating in input (see above)

packages/cli/package.json
```diff
@@ -17,7 +17,7 @@
     "debug": "node --inspect-brk dist/index.js",
     "lint": "eslint . --ext .ts,.tsx",
     "format": "prettier --write .",
-    "test": "vitest run",
+    "test": "vitest run -- ",
```
richieforeman, line 20: Extraneous change? Not sure.

package.json
```diff
@@ -55,7 +62,9 @@
     "telemetry": "node scripts/telemetry.js",
     "check:lockfile": "node scripts/check-lockfile.js",
     "clean": "node scripts/clean.js",
-    "pre-commit": "node scripts/pre-commit.js"
+    "pre-commit": "node scripts/pre-commit.js",
+    "use-dev": "node scripts/configure-registry.js dev",
```
richieforeman: Opinion: I think we should just have people setup an npmrc with both and then there's no need to "switch"

richieforeman: Like, I'm not sure I see a need for these/this command.

.github/actions/publish-release/action.yml
mattKorwel: which cl?

.github/actions/publish-release/action.yml
mattKorwel: So in this case I do not want to run the verfiy action. we're literally publishing this to run the integration tests in the next job.  I think if its confusing it would be better to just remove this step as i suppose its redundant.

.github/workflows/build-and-publish.yml
mattKorwel: I see it as a replacement for that. This is hard coded to push to ghcr.io and is a dedicated step here. 

The Publish Release action is a comprehensive workflow that automates the process of releasing the gemini-cli packages. It handles:

   * Version Management: Updates package versions and commits these changes.
   * Branching: Creates and manages a dedicated release branch.
   * Building/Bundling: Builds or bundles the packages based on the use-bundle-release input.
   * NPM Publishing: Publishes the core, CLI, and a2a-server packages to NPM, handling authentication and tagging.
   * Release Verification: Verifies the published NPM release using either the verify-release action or a direct npx command, depending on whether a bundled release is used.
   * GitHub Release Creation: Creates a GitHub release with generated notes.
   * Cleanup: Cleans up the temporary release branch.

We're not doing any of that here. And in fact I believe that eventaully this will be replaing that action.

.github/workflows/build-and-publish.yml
mattKorwel: In this case again, i think that's a larger action. this is a much more granular step. my hope is to get these changes in and working and asses where commonality can be found.

packages/cli/package.json
mattKorwel: Not extraneous. allows you to pass in additional arguments and call individual tests.

scripts/configure-registry.js
mattKorwel: I think its a matter of prefernece, i can't see any pros and cons to one way or antoher. that said i made it "set up" that can run once and or anytime the auth expries.

PR #11637, opened 2025-10-22
Issue#9751 - fix the gemini crash on startup in tmux environments

megha1188:
## TLDR

This PR fixes a startup crash in `gemini-cli` that occurred in terminal environments like `tmux` where terminal color detection could fail. It adds a fallback gradient to ensure the UI renders correctly even when theme colors are not available.

## Dive Deeper

The `gemini-cli` would crash on startup with an `Invalid number of stops (< 2)` error. This was caused by the `tinygradient` library, which renders the header. In certain terminal environments, such as `tmux` (where `TERM=screen`), the color detection logic would fail to find an adequate number of color stops to create a gradient.

This fix introduces a check to ensure that `theme.ui.gradient` contains at least two colors before attempting to render the gradient. If it does not, a default fallback palette is used instead, preventing the crash.

Unit tests have also been added to verify this new fallback logic.

## Reviewer Test Plan

1.  Pull down this branch.
3.  **To verify the fix**:
    *   In your terminal, run the following commands to simulate the problematic environment:
        ```bash
        export TERM=screen
        export LANG=C.UTF-8
        ```
    *   Run the `gemini` command.
    *   **Expected Result:** The CLI starts up without crashing, and the header displays with a default gradient.
4.  **To verify non-regression**:
    *   Open a new, standard terminal session.
    *   Run the `gemini` command.
    *   **Expected Result:** The CLI starts up, and the header displays with the correct, themed gradient.
5.  Run the unit tests via `npm test` and ensure the new tests in `packages/cli/src/ui/components/Header.test.tsx` pass successfully.

## Linked issues / bugs

Fixes #9751


packages/cli/src/ui/components/Header.tsx
```diff
@@ -40,29 +40,30 @@ export const Header: React.FC<HeaderProps> = ({
 
   const artWidth = getAsciiArtWidth(displayTitle);
 
+  // Theming is not always reliable, and can return an array with a single
+  // (or no) color stops. The gradient library requires at least two, so this
+  // logic checks for that and provides a fallback.
+  const PALETTE = ['#4285F4', '#34A853', '#FBBC05', '#EA4335'];
```
jacob314: rather than hard coding this PALETTE, use the single solid color when a single color stop is specified and omit the gradient completely when no color stop is specified.

packages/cli/src/ui/components/Header.tsx
```diff
@@ -18,6 +18,24 @@ interface HeaderProps {
   nightly: boolean;
 }
 
+const ThemedLogo: React.FC<{ children: React.ReactNode }> = ({ children }) => {
```
jacob314: This is just ThemedGradient not ThemedLogo.
Note that you are using it for the version number string as well which isn't a logo.

jacob314 (commented):
Approved after the name is changed from ThemedLog to ThemedGradient.

jacob314 (approved):
lgtm

PR #11642, opened 2025-10-22
feat: Update permissions command to support modifying trust for other…

shrutip90:
… folders

## TLDR

'/permissions' command currently only supports modifying the trust level for the current folder and looks like this:
Screenshot 2025-10-23 at 2 38 53 PM

The change here allows adding an optional folder path parameter to the command to modify trust levels of other folders other than the current folder. I have also added a 'trust' a subcommand since '/permissions <folder-path>' does not read very clearly and we were planning to have other permissions subcommands anyway. So to modify trust for current folder, you would use '/permissions trust' and for other folders, use '/permissions trust <folder-path>'.

For other folders, example:
Screenshot 2025-10-22 at 4 51 14 PM

NOTE that we avoid showing some of the warning messages etc around inherited trust, while modifying trust for external folders, since they only make sense in the context of current workspace.

## Reviewer Test Plan

Checkout the PR and try out the new '/permissions trust' and '/permissions trust <folder-path>' commands.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes https://github.com/google-gemini/maintainers-gemini-cli/issues/844

### DIFF
--- a/packages/cli/src/ui/commands/permissionsCommand.ts
+++ b/packages/cli/src/ui/commands/permissionsCommand.ts
@@ -4,15 +4,82 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import type { OpenDialogActionReturn, SlashCommand } from './types.js';
+import type {
+  OpenDialogActionReturn,
+  SlashCommand,
+  SlashCommandActionReturn,
+} from './types.js';
 import { CommandKind } from './types.js';
+import * as process from 'node:process';
+import * as path from 'node:path';
+import * as os from 'node:os';
+import * as fs from 'node:fs';
 
 export const permissionsCommand: SlashCommand = {
   name: 'permissions',
-  description: 'Manage folder trust settings',
+  description: 'Manage folder trust settings and other permissions',
   kind: CommandKind.BUILT_IN,
-  action: (): OpenDialogActionReturn => ({
-    type: 'dialog',
-    dialog: 'permissions',
-  }),
+  subCommands: [
+    {
+      name: 'trust',
+      description:
+        'Manage folder trust settings. Usage: /permissions trust [<directory-path>]',
+      kind: CommandKind.BUILT_IN,
+      action: (context, input): SlashCommandActionReturn => {
+        const dirPath = input.trim();
+        let targetDirectory: string;
+
+        if (!dirPath) {
+          targetDirectory = process.cwd();
+        } else if (dirPath.startsWith('~')) {
+          targetDirectory = path.join(os.homedir(), dirPath.slice(1));
+        } else {
+          targetDirectory = path.resolve(dirPath);
+        }
+
+        try {
+          if (!fs.statSync(targetDirectory).isDirectory()) {
+            return {
+              type: 'message',
+              messageType: 'error',
+              content: `Path is not a directory: ${targetDirectory}`,
+            };
+          }
+        } catch (e) {
+          const message = e instanceof Error ? e.message : String(e);
+          return {
+            type: 'message',
+            messageType: 'error',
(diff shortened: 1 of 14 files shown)


galdawave:
@shrutip90 can you add a PR description, what this PR does and how to validate

shrutip90:
> @shrutip90 can you add a PR description, what this PR does and how to validate

Done

packages/cli/src/ui/hooks/usePermissionsModifyTrust.ts
```diff
@@ -45,11 +56,16 @@ function getInitialTrustState(
 export const usePermissionsModifyTrust = (
   onExit: () => void,
   addItem: UseHistoryManagerReturn['addItem'],
+  targetDirectory: string,
 ) => {
   const settings = useSettings();
-  const cwd = process.cwd();
+  const cwd = targetDirectory;
+  const isCurrentWorkspace =
+    path.resolve(targetDirectory) === path.resolve(process.cwd());
```
galdawave: This may fail on case-insensitive file systems (macOS/Windows) if the casing provided by the user differs from process.cwd(). Consider normalizing the paths (e.g., converting to lowercase) before comparison on these platforms, or using a more robust path comparison method.

packages/cli/src/ui/commands/permissionsCommand.ts
```diff
@@ -4,15 +4,82 @@
...
+      description:
+        'Manage folder trust settings. Usage: /permissions trust [<directory-path>]',
+      kind: CommandKind.BUILT_IN,
+      action: (context, input): SlashCommandActionReturn => {
+        const dirPath = input.trim();
+        let targetDirectory: string;
+
+        if (!dirPath) {
+          targetDirectory = process.cwd();
+        } else if (dirPath.startsWith('~')) {
```
galdawave: This will incorrectly expand ~username to homedir/username instead of that user's home directory. It's safer to check for `dirPath === '~' || dirPath.startsWith('~/')` 

(and potentially ~\ for Windows support).

packages/cli/src/ui/contexts/UIActionsContext.tsx
```diff
@@ -30,6 +30,7 @@ export interface UIActions {
   exitPrivacyNotice: () => void;
   closeSettingsDialog: () => void;
   closeModelDialog: () => void;
+  openPermissionsDialog: (props?: { targetDirectory?: string }) => void;
```
galdawave: Nit: might be better for readability to make `props?: { targetDirectory?: string }` this an interface so something like 

```typescript
   export interface PermissionsDialogProps {
     targetDirectory?: string;
   }
```

galdawave (changes requested):

packages/cli/src/ui/commands/permissionsCommand.ts
shrutip90: Updated this to use expandHomeDir utility instead.

packages/cli/src/ui/hooks/usePermissionsModifyTrust.ts
shrutip90: Done, converted to lowercase

packages/cli/src/ui/contexts/UIActionsContext.tsx
shrutip90: Done

galdawave (approved):

Issue #11586, opened 2025-10-21
Extensions logic refactor


jakemac53:
Today, extension logic is spread out over many files and top level functions which operate on what is effectively the same state. These also often perform duplicate work (including loading stuff from the file system), and the same parameters have to be deeply plumbed through several layers of calls.

Also, all logic associated with actually loading things from extensions is spread out broadly across many areas of the codebase, which makes implementing dynamic extension install/uninstall very challenging to do in a robust way.

We should refactor all the functionality into an `ExtensionManager` class, so that they can share the same state (and only load things once). This should simplify a bunch of the logic and reduce the number of parameters passed around.

All interactions with other systems should be done through this class where possible, making it clear where those touch points are, so that implementing dynamic install/uninstall is easier to do.

PR #11667, opened 2025-10-22
Create ExtensionManager class which manages all high level extension tasks

jakemac53:
## TLDR

Moves almost all top level functions from `extensions.ts` to instance methods on a new `ExtensionManager` class.

This simplifies the parameter signature for these methods, using instead the instance fields on ExtensionManager to hold most of that state.

## Dive Deeper

There is more to  happen here, but this is the first setup PR. I left all the tests in `extension.test.ts` for now so the diff is easier to see and to manage merge conflicts - will follow up with a PR that moves that all into `extension-manager.test.ts`.

I also want to follow up with some more fundamental changes, removing `extensions` from the `config` object entirely, and caching the loaded extensions on the `ExtensionManager` class itself. And then using the extension manager itself to make changes to the tool registry, system prompts, etc.

## Reviewer Test Plan

There should be no visible changes for users.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ✅   |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

First part of https://github.com/google-gemini/gemini-cli/issues/11586


packages/cli/src/commands/extensions/disable.ts
```diff
@@ -5,32 +5,31 @@
...
   name: string;
   scope?: string;
 }
 
 export function handleDisable(args: DisableArgs) {
-  const extensionEnablementManager = new ExtensionEnablementManager();
+  const workspaceDir = process.cwd();
+  const extensionManager = new ExtensionManager({
+    workspaceDir,
+    enabledExtensionOverrides: [],
```
chrstnb: It seems a bit odd to be including these values here, no? specifically the enabledExtensionOverrides and requestConsent-- should we make these optional?

packages/cli/src/config/extensions/consent.ts
chrstnb, line 1: awesome

packages/cli/src/commands/extensions/disable.ts
jakemac53: We could make `enabledExtensionOverrides` optional but I think `requestConsent` should not be optional because it is required for some of the functions to work.

This is the downside of this approach, the dependencies of all functions have to be provided even if you only use one piece, but ultimately I think its fine.

chrstnb (approved):
Only comment is I think we can make the enabledExtensionOverrides and requestConsent optional as there are many places where they're not at all relevant

packages/cli/src/commands/extensions/disable.ts
jakemac53: Fwiw I made `enabledExtensionOverrides` not be optional simply because I prefer more explicit code, you could just forget to pass this when you should be passing `argv.extensions`, but making it not optional forces you to make the explicit decision that this isn't a relevant argument for this call site.

chrstnb: ok let's make enabledExtensionOverrides optional then

jakemac53: I went ahead and made `enabledExtensionOverrides` optional but left in `requestConsent` (and the new `requestSetting`).

PR #11675, opened 2025-10-22
extract console error to util func

abhipatel12:
## TLDR

Initial batch of moves for these console.errors.

These are moved as is b/c these do not need any user specific chat messages and can be moved as is. 

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  |✅ | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt |✅| -   | -   |

## Linked issues / bugs

This PR makes progress on #10761

### DIFF
--- a/packages/a2a-server/src/config/settings.ts
+++ b/packages/a2a-server/src/config/settings.ts
@@ -10,6 +10,7 @@ import { homedir } from 'node:os';
 
 import type { MCPServerConfig } from '@google/gemini-cli-core';
 import {
+  debugLogger,
   GEMINI_DIR,
   getErrorMessage,
   type TelemetrySettings,
@@ -97,10 +98,10 @@ export function loadSettings(workspaceDir: string): Settings {
   }
 
   if (settingsErrors.length > 0) {
-    console.error('Errors loading settings:');
+    debugLogger.error('Errors loading settings:');
     for (const error of settingsErrors) {
-      console.error(`  Path: ${error.path}`);
-      console.error(`  Message: ${error.message}`);
+      debugLogger.error(`  Path: ${error.path}`);
+      debugLogger.error(`  Message: ${error.message}`);
     }
   }
 
--- a/packages/cli/src/config/settings.ts
+++ b/packages/cli/src/config/settings.ts
@@ -617,7 +617,7 @@ export function loadSettings(
                   'utf-8',
                 );
               } catch (e) {
-                debugLogger.error(
+                console.error(
                   `Error migrating settings file on disk: ${getErrorMessage(
                     e,
                   )}`,
@@ -799,6 +799,6 @@ export function saveSettings(settingsFile: SettingsFile): void {
       settingsToSave as Record<string, unknown>,
     );
   } catch (error) {
-    debugLogger.error('Error saving user settings file:', error);
+    console.error('Error saving user settings file:', error);
   }
 }
--- a/packages/cli/src/config/settings.ts
+++ b/packages/cli/src/config/settings.ts
@@ -617,7 +617,7 @@ export function loadSettings(
                   'utf-8',
                 );
               } catch (e) {
-                console.error(
+                debugLogger.error(
                   `Error migrating settings file on disk: ${getErrorMessage(
                     e,
                   )}`,
@@ -799,6 +799,6 @@ export function saveSettings(settingsFile: SettingsFile): void {
       settingsToSave as Record<string, unknown>,
     );
   } catch (error) {
-    console.error('Error saving user settings file:', error);
+    debugLogger.error('Error saving user settings file:', error);
   }
 }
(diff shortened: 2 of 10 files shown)


packages/a2a-server/src/config/settings.ts
```diff
@@ -97,10 +98,10 @@ export function loadSettings(workspaceDir: string): Settings {
   }
 
   if (settingsErrors.length > 0) {
-    console.error('Errors loading settings:');
+    debugLogger.error('Errors loading settings:');
```
jacob314, line 101: this is in the a2a-server code. how does console.error or debugLogger.error even even work in that case?
FYI looks like there is an additional issue that we should be failing to startup in this case rather than just continuing after errors. think that is the behavior for the regular Gemini CLI.

packages/cli/src/config/settings.ts
```diff
@@ -617,7 +617,7 @@ export function loadSettings(
                   'utf-8',
                 );
               } catch (e) {
-                console.error(
+                debugLogger.error(
```
jacob314: This one should probably be user visible.

packages/cli/src/config/settings.ts
```diff
@@ -799,6 +799,6 @@ export function saveSettings(settingsFile: SettingsFile): void {
       settingsToSave as Record<string, unknown>,
     );
   } catch (error) {
-    console.error('Error saving user settings file:', error);
+    debugLogger.error('Error saving user settings file:', error);
```
jacob314: this one shoudl be user visible with something like
`Failed to save user settings file to $fileName`

packages/cli/src/ui/utils/CodeColorizer.tsx
```diff
@@ -191,7 +192,7 @@ export function colorizeCode(
       </MaxSizedBox>
     );
   } catch (error) {
-    console.error(
+    debugLogger.warn(
       `[colorizeCode] Error highlighting code for language "${language}":`,
```
jacob314, line 196: agree this is fine as a warning. no user should care about this.

packages/cli/src/zed-integration/zedIntegration.ts
```diff
@@ -158,7 +158,7 @@ class GeminiAgent {
         );
         isAuthenticated = true;
       } catch (e) {
-        console.error(`Authentication failed: ${e}`);
+        debugLogger.error(`Authentication failed: ${e}`);
```
jacob314, line 161: this one needs to be user visible. also not sure about how this should be surfaced for the zed case. would follow up with them to understand if there is a way errors should be surfaced to them

packages/core/src/config/config.ts
```diff
@@ -1192,24 +1192,13 @@ export class Config {
...
-            const messageBusEnabled = this.getEnableMessageBusIntegration();
-            const wrapper = new SubagentToolWrapper(
-              definition,
-              this,
-              messageBusEnabled ? this.getMessageBus() : undefined,
-            );
-            registry.registerTool(wrapper);
-          } catch (error) {
-            console.error(
-              `Failed to wrap agent '${definition.name}' as a tool:`,
```
jacob314, line 1212: try catch got removed completely. is that intended?

packages/core/src/tools/memoryTool.ts
```diff
@@ -267,9 +267,6 @@ class MemoryToolInvocation extends BaseToolInvocation<
     } catch (error) {
       const errorMessage =
         error instanceof Error ? error.message : String(error);
-      console.error(
```
jacob314, line 270: maybe keep this but log as a warning?

jacob314 (commented):
done with first round of review

packages/a2a-server/src/config/settings.ts
abhipatel12: Just chatted with Adam, he's creating an issue to remove this and crash in this case. In that case, the console logs can be removed by the error that is thrown.

abhipatel12: Keeping as is since they will be removed

packages/cli/src/config/settings.ts
abhipatel12: reverted for now, since this will need the more useful chat error

packages/cli/src/config/settings.ts
abhipatel12: reverted for now for a different PR that adds this to chat history

packages/core/src/tools/memoryTool.ts
abhipatel12: Sure. Done

packages/core/src/config/config.ts
abhipatel12: Yes, this was an unnecessary try catch. If this actually failed, we should know immediately.

packages/cli/src/zed-integration/zedIntegration.ts
abhipatel12: This is why I kept it as debugLogger.error since this maintains existing behavior. How do i follow up with them?

jacob314: @mattKorwel should know or you can look at blame for this file to follow up with the github user who added this.

jacob314 (approved):
lgtm

Issue #7773, opened 2025-09-05
Paste bug


mackcoding:
### What happened?

I am on Windows 11 with CLI version 0.3.2 (ad3bc17e), using gemini-2.5-pro. If I type text and then paste code, it sends the typed text over and over and leaves a small portion of the code in the input area.

<img width="3447" height="1915" alt="Image" src="https://github.com/user-attachments/assets/b92768fd-811e-4f15-9cc5-e9ec13946745" />

This has happened on previous version (0.2.2) and still happens after I upgraded. 

### What did you expect to happen?

I expect something as simple as pasting in a console to work. 

### Client information

<details>
<summary>Client Information</summary>

Run `gemini` to enter the interactive CLI, then run the `/about` command.

```console
╭─────────────────────────────────────────────────────────────────────────────────────╮
│                                                                                     │
│ About Gemini CLI                                                                    │
│                                                                                     │
│ CLI Version                  0.3.2                                                  │
│ Git Commit                   ad3bc17e                                               │
│ Model                        gemini-2.5-pro                                         │
│ Sandbox                      no sandbox                                             │
│ OS                           win32                                                  │
│ Auth Method                  OAuth                                                  │
│                                                                                     │
╰─────────────────────────────────────────────────────────────────────────────────────╯
```

</details>

mackcoding:
https://github.com/user-attachments/assets/ad149b97-e05f-40f5-9c21-1ae53720b1a1

Here is a video. I pasted multiple lines of text (3 lines). If you type in the input, and paste 3 lines, it replicates 2 lines (as seen in the video). I messed a little bit with it and it does this behavior. 

There's other weirdness like large blocks of code missing from the input box, I can't reliably replicate this. Basically, the input is just straight up broken.

mackcoding:
I am honestly running into a lot of input bugs. It is pretty much completely broken. If there is a long response, and I hit esc to cancel the request, it takes my last two responses and adds them to the input.

Pasting large chunks of code just doesn't work at all. When the input is empty, and you paste code, it truncates and repeats the code.

The input area is just completely broken and almost unusable in its current state, I'm not sure how this hasn't been noticed in testing.

NeonBeacon:
crazy how this hasn't been resolved since it was raised on another ticket in June, and the item was closed. . . . Gemini CLI is unusable at this point

mackcoding:
[...]

PR #11677, opened 2025-10-22
Refactor KeypressContext

scidomino:
## TLDR

Refactor KeypressContext to consolidate state and use the same patterns.

## Dive Deeper

I broke this up into separate commits so it'll be easier to review.

## Reviewer Test Plan

Try various key combinations and make sure they work.
* Drag and drop a file
* \\+enter
* copy paste a chunk

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | x  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

For #7773

### DIFF
--- a/packages/cli/src/ui/contexts/KeypressContext.tsx
+++ b/packages/cli/src/ui/contexts/KeypressContext.tsx
@@ -434,7 +434,6 @@ export function KeypressProvider({
     let kittySequenceBuffer = '';
     let kittySequenceTimeout: NodeJS.Timeout | null = null;
     let backslashTimeout: NodeJS.Timeout | null = null;
-    let isDragging = false;
     let dragBuffer = '';
     let draggingTimer: NodeJS.Timeout | null = null;
 
@@ -502,14 +501,13 @@ export function KeypressProvider({
       if (
         key.sequence === SINGLE_QUOTE ||
         key.sequence === DOUBLE_QUOTE ||
-        isDragging
+        draggingTimer !== null
       ) {
-        isDragging = true;
         dragBuffer += key.sequence;
 
         clearDraggingTimer();
         draggingTimer = setTimeout(() => {
-          isDragging = false;
+          draggingTimer = null;
           const seq = dragBuffer;
           dragBuffer = '';
           if (seq) {
@@ -878,7 +876,7 @@ export function KeypressProvider({
         clearTimeout(draggingTimer);
         draggingTimer = null;
       }
-      if (isDragging && dragBuffer) {
+      if (dragBuffer) {
         broadcast({
           name: '',
           ctrl: false,
@@ -887,7 +885,6 @@ export function KeypressProvider({
           paste: true,
           sequence: dragBuffer,
         });
-        isDragging = false;
         dragBuffer = '';
       }
     };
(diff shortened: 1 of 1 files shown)


packages/cli/src/ui/contexts/KeypressContext.tsx
```diff
@@ -76,6 +76,283 @@ const ALT_KEY_CHARACTER_MAP: Record<string, string> = {
...
+        paste: false,
+        sequence: buffer.slice(0, m[0].length),
+        kittyProtocol: true,
+      },
+      length: m[0].length,
+    };
+  }
+
+  return null;
+}
```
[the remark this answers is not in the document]
jacob314: agree this is a good followup to this pr

packages/cli/src/ui/contexts/KeypressContext.tsx
```diff
@@ -479,41 +486,41 @@ export function KeypressProvider({
...
-        isDraggingRef.current = true;
-        dragBufferRef.current += key.sequence;
+        dragBuffer += key.sequence;
 
         clearDraggingTimer();
-        draggingTimerRef.current = setTimeout(() => {
-          isDraggingRef.current = false;
-          const seq = dragBufferRef.current;
-          dragBufferRef.current = '';
+        draggingTimer = setTimeout(() => {
```
jacob314: why did we reomove the required use of useRef here and elsewhere?

packages/cli/src/ui/contexts/KeypressContext.tsx
```diff
@@ -791,13 +790,13 @@ export function KeypressProvider({
         markerLength = pasteModeSuffixBuffer.length;
 
         if (nextMarkerPos === -1) {
-          keypressStream.write(data.slice(pos));
+          keypressStream!.write(data.slice(pos));
```
jacob314, line 793: why is keypressStream now nullable. that seems like a regression in readability

packages/cli/src/ui/contexts/KeypressContext.tsx
```diff
@@ -819,7 +818,7 @@ export function KeypressProvider({
     };
 
     let rl: readline.Interface;
-    if (usePassthrough) {
+    if (keypressStream !== null) {
```
jacob314, line 821: usePassthrough was clearer. why has this changed?

packages/cli/src/ui/contexts/KeypressContext.tsx
```diff
@@ -881,33 +880,29 @@ export function KeypressProvider({
...
-          sequence: dragBufferRef.current,
+          sequence: dragBuffer,
         });
-        isDraggingRef.current = false;
-        dragBufferRef.current = '';
+        dragBuffer = '';
       }
     };
   }, [
     stdin,
```
jacob314, line 900: as an aside: having this have dependencies other than stdin is sketchy. We should have the useRef have 0 deps other than perhaps stdin and use useRef to get the latest values if the values happened to change. that can be a follow up. but it is probably more critical now that you are relying on the useEffect only being called exactly once even more to make the business logic work as you rightly deleted the fairly spurious useRefs inside the useEffect.

jacob314 (approved):
#lgtm

lifefloating:
impressive :+1:

PR #11698, opened 2025-10-22
feat(core) Bump get-ripgrep version.

### LINKED ISSUE #11495: Clean up temporary ripgrep download directories
### What happened?

When running `npm run preflight`, the ripgrep installer (from `@joshua.litt/get-ripgrep`) leaves behind a directory such as `packages/cli/download-ripgrepv7ZZRp`. This appears to be a temp folder produced by `tempy()` while downloading the archive. After the download completes and the binary is moved into the cache/bin directory, the temp directory sticks around as an untracked file.

The ripgrep download path should be removed automatically once the archive is moved/extracted so the repo stays clean after installs/builds. Alternatively, our cleanup scripts could delete these directories.

### Client information

0.10 preview

### Login information

_No response_

_No response_

joshualitt:
Bumps get-ripgrep to include https://github.com/lvce-editor/ripgrep/commit/e6266d9fa2688517f07b72952542ccc11b59362a , which downloads ripgrep into a system temp directory.

Fixes https://github.com/google-gemini/gemini-cli/issues/11495 (The clean-up was already happening, but this PR will at least move the download directory to the system tmp dir, thus it should be less intrusive and will get cleaned up automatically)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1781,9 +1781,10 @@
       }
     },
     "node_modules/@joshua.litt/get-ripgrep": {
-      "version": "0.0.2",
-      "resolved": "https://registry.npmjs.org/@joshua.litt/get-ripgrep/-/get-ripgrep-0.0.2.tgz",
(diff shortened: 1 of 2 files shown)


jacob314 (approved):
lgtm

PR #11760, opened 2025-10-22
feat(auth): improve API key authentication flow

### LINKED ISSUE #11313: feat(auth): Add option to save GEMINI_API_KEY to .env file
**Description:**

To make future sessions seamless, after a user successfully enters their API key via the interactive prompt, the CLI should ask for their permission to save the key to a local `.env` file.

**Acceptance Criteria:**

- After a user enters their API key, a prompt should appear asking if they want to save the key to a `.env` file (e.g., "Save API Key for future use? (Y/n)").
- A new function should be created in `packages/cli/src/config/auth.ts` to handle the logic of safely writing the `GEMINI_API_KEY` to a `.env` file in the current working directory.
- The logic should correctly handle cases where a `.env` file already exists (with or without the key) and where it doesn't exist.
- A confirmation message should be displayed to the user after the key is successfully saved.

galdawave:
## TLDR

This PR improves the authentication flow for Gemini API keys. It introduces a new dialog for API key input, securely stores the key, and refactors the related code for better maintainability.

## Dive Deeper

The main changes in this PR are:
1.  **New `ApiAuthDialog`:** A new React component `ApiAuthDialog` is created to prompt the user for their Gemini API key. This provides a better user experience than relying on environment variables.
2.  **`ApiKeyCredentialStorage`:** A new class `ApiKeyCredentialStorage` is added to securely store the API key in the system's keychain. This avoids having to re-enter the key on every run.
3.  **Refactored `useAuth` hook:** The `useAuth` hook is refactored to handle the new API key input flow. The logic for loading the API key is centralized in the `reloadApiKey` function.
4.  **Controlled `TextInput` component:** The `TextInput` component is refactored to be a fully controlled component, which simplifies its implementation and makes it more reusable.
5.  **Configuration updates:** The configuration loading logic is updated to read the API key from the new `ApiKeyCredentialStorage`.

## Reviewer Test Plan

1.  Run the Gemini CLI without any authentication configured.
2.  Select "Gemini API Key" as the authentication method.
3.  You should be prompted to enter your API key in a new dialog.
4.  Enter a valid API key. The CLI should authenticate successfully.
6.  To test the clearing of the key, you can manually delete the entry from your keychain (service name: `gemini-cli-api-key`) and rerun the CLI. You should be prompted for the key again.

1.  Unset GEMINI_API_KEY environent variable
1.  Run the Gemini CLI with "Gemini API Key" authentication configured.
3.  You should be prompted to enter your API key in a new dialog.
4.  Enter a valid API key. The CLI should authenticate successfully.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

*No linked issues*


packages/cli/src/ui/auth/ApiAuthDialog.tsx
```diff
@@ -0,0 +1,79 @@
...
+          paddingX={1}
+          flexGrow={1}
+        >
+          <TextInput
+            value={apiKey}
+            onChange={setApiKey}
+            onSubmit={onSubmit}
+            onCancel={onCancel}
+            placeholder="Paste your API key here"
+          />
```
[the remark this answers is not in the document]
galdawave: I think @miguelsolorio and @abhipatel12 suggested we don't use masking, since we want the user to be able to see the api key that they put in, and you can just echo the env var so masking it doesn't seem that needed.

packages/cli/src/ui/AppContainer.tsx
```diff
@@ -397,6 +402,27 @@ Logging in with Google... Please restart Gemini CLI to continue.
...
+        await ApiKeyCredentialStorage.saveApiKey(apiKey);
+        await reloadApiKey();
+        await config.refreshAuth(AuthType.USE_GEMINI);
+        setAuthState(AuthState.Authenticated);
+      } catch (e) {
+        onAuthError(
+          `Failed to save API key: ${e instanceof Error ? e.message : String(e)}`,
+        );
+      }
+    },
```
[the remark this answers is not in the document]
galdawave: I can do this, but since we are using keychain, i think it makes sense that if the api key is an empty string, then we clear the api key, i'll add a check to see if the user just enters a bunch of white spaces then we should raise this error.

packages/cli/src/ui/auth/useAuth.ts
```diff
@@ -42,6 +53,14 @@ export const useAuthCommand = (settings: LoadedSettings, config: Config) => {
     [setAuthError, setAuthState],
   );
 
+  const reloadApiKey = useCallback(async () => {
+    const envKey = process.env['GEMINI_API_KEY'] ?? undefined;
+    const storedKey = await ApiKeyCredentialStorage.loadApiKey();
+    const key = storedKey || envKey;
```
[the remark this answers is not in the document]
jacob314: any thoughts?

packages/core/src/core/apiKeyCredentialStorage.ts
```diff
@@ -0,0 +1,72 @@
...
+      );
+
+      if (credentials?.token?.accessToken) {
+        return credentials.token.accessToken;
+      }
+
+      return null;
+    } catch (error: unknown) {
+      // Log error but don't crash, just return null so user can re-enter key
+      console.error('Failed to load API key from storage:', error);
```
jacob314: use debugLogger. Also consider whether this should be an error or just a warning.

packages/core/src/core/apiKeyCredentialStorage.ts
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  /**
+   * Clear cached API key
+   */
+  static async clearApiKey(): Promise<void> {
+    try {
+      await this.storage.deleteCredentials(DEFAULT_API_KEY_ENTRY);
+    } catch (error: unknown) {
+      console.error('Failed to clear API key from storage:', error);
```
jacob314: no need to log and throw

packages/cli/src/ui/components/shared/TextInput.tsx
```diff
@@ -0,0 +1,147 @@
...
+    // out of bounds. This can happen if the parent component changes the
+    // value programmatically.
+    if (cursorOffset > value.length) {
+      setCursorOffset(value.length);
+    }
+  }, [value, cursorOffset]);
+
+  useEffect(() => {
+    if (isExiting && onCancel) {
+      // Give Ink a moment to run the useInput cleanup before we trigger the
```
jacob314: this seems obsolete now that we aren't using useInput. Would guess this is a hack to try to fix odd bugs in useInput when you were using it

packages/cli/src/ui/components/shared/TextInput.tsx
```diff
@@ -0,0 +1,147 @@
...
+
+        if (key.name === 'right') {
+          if (cursorOffset < value.length) {
+            setCursorOffset(cursorOffset + 1);
+          }
+          return;
+        }
+
+        // Normal character input
+        if (key.sequence && !key.ctrl && !key.meta) {
```
jacob314: we are letting a lot of unsafe chars in. examples: tabs, up arrow, down arrow. UI does really weird things due to this if you press up or down arrows.

packages/cli/src/ui/components/shared/TextInput.tsx
```diff
@@ -0,0 +1,147 @@
...
+          const newValue =
+            value.slice(0, cursorOffset) +
+            key.sequence +
```
jacob314: filter the sequence to strip ansi escape codes that might be in a paste, etc. See text-buffer.ts for examples of this.

packages/cli/src/ui/components/shared/TextInput.tsx
```diff
@@ -0,0 +1,147 @@
...
+  value: string;
+  placeholder?: string;
+  onChange: (value: string) => void;
+  onSubmit?: (value: string) => void;
+  onCancel?: () => void;
+  mask?: string;
+  focus?: boolean;
+}
+
+export function TextInput({
```
jacob314, line 25: for the api key case would be nice to filter out all chars not part of a valid api key. real reason to do this is if you paste an API key but are sloppy about the paste bounds you should still paste correctly rather than having trailing spaces that break the API key.

packages/cli/src/ui/AppContainer.tsx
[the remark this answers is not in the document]
jacob314: agree it is worth it. it is odd I can submit the dialog when the apikey is empty. i should have to press escape and I should get an error message explaining why submit was blocked.

packages/cli/src/ui/components/shared/TextInput.tsx
```diff
@@ -0,0 +1,147 @@
...
+
+import type React from 'react';
+import { useState, useEffect, useCallback } from 'react';
+import type { Key } from '../../hooks/useKeypress.js';
+import { Text, Box } from 'ink';
+import { useKeypress } from '../../hooks/useKeypress.js';
+import chalk from 'chalk';
+import { theme } from '../../semantic-colors.js';
+
+export interface TextInputProps {
```
jacob314, line 17: should this use text-buffer.ts internally? that would align the keyboard shortcuts.
This could be a nice alternate component wrapped around it that wouldn't have all the complexity of the input prompt. you might need to make  text-buffer have an option to restrict to one line but that would be simple.

jacob314 (commented):
other issue noticed manually testing:
ctrl-c now doesn't seem to work in the api key dialog or the auth dialog. odd that changed as they do work in that dialog on the public build.

packages/core/src/core/apiKeyCredentialStorage.ts
galdawave: i feel like logging to the debugLogger might be helpful

packages/cli/src/ui/components/shared/TextInput.tsx
galdawave: NA

packages/cli/src/ui/components/shared/TextInput.tsx
galdawave: Changed it to use the text-buffer

packages/cli/src/ui/components/shared/TextInput.tsx
galdawave: Nothing happens if you press up, down, tab

packages/cli/src/ui/components/shared/text-buffer.ts
```diff
@@ -1609,7 +1613,15 @@ export function useTextBuffer({
 
   const insert = useCallback(
     (ch: string, { paste = false }: { paste?: boolean } = {}): void => {
-      if (/[\n\r]/.test(ch)) {
+      if (singleLine) {
```
jacob314: instead of doing the filtering here you can do it inside 'insert' in the reducer ensuring it applies to all the cases

jacob314: that would provide an extra line of defense ensuring invalid chars don't get through regardless of what triggered the 'insert'

packages/cli/src/ui/components/shared/text-buffer.test.ts
```diff
@@ -1521,6 +1521,75 @@ Contrary to popular belief, Lorem Ipsum is not simply random text. It has roots
     });
   });
 
+  describe('inputFilter', () => {
+    it('should filter input based on the provided filter function', () => {
+      const { result } = renderHook(() =>
+        useTextBuffer({
+          viewport,
+          isValidPath: () => false,
+          inputFilter: (text) => text.replace(/[^0-9]/g, ''),
```
jacob314, line 1567: nit: it might make sense to also apply the inputFilter to the initial value passed in but not sure that is a good idea.

packages/cli/src/ui/AppContainer.tsx
```diff
@@ -1042,6 +1075,26 @@ Logging in with Google... Please restart Gemini CLI to continue.
     ],
   );
 
+  const dialogsVisible =
```
jacob314: nit: why did didalogVisible get moved in this file? doesn't seem like it got referenced between its old and new location.

packages/core/src/core/apiKeyCredentialStorage.ts
```diff
@@ -0,0 +1,72 @@
...
+ */
+
+import { HybridTokenStorage } from '../mcp/token-storage/hybrid-token-storage.js';
+import type { OAuthCredentials } from '../mcp/token-storage/types.js';
+import { debugLogger } from '../utils/debugLogger.js';
+
+const KEYCHAIN_SERVICE_NAME = 'gemini-cli-api-key';
+const DEFAULT_API_KEY_ENTRY = 'default-api-key';
+
+export class ApiKeyCredentialStorage {
```
jacob314: Nit suggestion from Gemini I agree with. This file  uses a class with only static methods (ApiKeyCredentialStorage GEMINI.md prefers plain objects or ES modules over classes when they don't need instances. I recommend suggesting converting these to top-level exported functions sharing a module-level storage constant.

jacob314 (commented):
Everything works well now when I manually test it. Did one final round of review comments then approved.

packages/cli/src/ui/AppContainer.tsx
galdawave: will revert

packages/cli/src/ui/components/shared/text-buffer.test.ts
galdawave: so inputFilter is optional, we can apply it but it might be overkill, i think the idea is that whatever is in intitialValue will be valid

jacob314 (approved):
lgtm

PR #11767, opened 2025-10-22
Fix bug where tool scheduler was repeatedly created.

### LINKED ISSUE #11753: Memory Leak: CoreToolScheduler is recreated on every render due to unstable callback
**Problem Description**

A new `CoreToolScheduler` instance is being created on every render cycle in the CLI. This is caused by an unstable callback being passed as a dependency to the `useMemo` hook responsible for its creation. This leads to unnecessary object creation, performance degradation, and a high risk of memory leaks.

**Root Cause Analysis**

1.  In `packages/cli/src/ui/hooks/useGeminiStream.ts`, the `onComplete` handler for `useReactToolScheduler` is an `async` function defined inline. This creates a new function reference on every render.

    ```typescript
    // File: packages/cli/src/ui/hooks/useGeminiStream.ts

    const [toolCalls, scheduleToolCalls, markToolsAsSubmitted] =
      useReactToolScheduler(
        async (completedToolCallsFromScheduler) => {  // ⚠️ This callback is unstable
          // ...
        },
        config,
        getPreferredEditor,
        onEditorClose,
      );
    ```

2.  In `packages/cli/src/ui/hooks/useReactToolScheduler.ts`, this unstable `onComplete` function is passed into a `useCallback` and then used as a dependency for the `useMemo` that creates the `CoreToolScheduler`.

    ```typescript
    // File: packages/cli/src/ui/hooks/useReactToolScheduler.ts

    const allToolCallsCompleteHandler: AllToolCallsCompleteHandler = useCallback(
      async (completedToolCalls) => {
        await onComplete(completedToolCalls);
      },
      [onComplete], // The dependency `onComplete` changes on every render
    );

    const scheduler = useMemo(
      () => new CoreToolScheduler({ ... }),
      [
        // ...
        allToolCallsCompleteHandler, // This causes the memoization to break
      ],
    );
    ```

Because `allToolCallsCompleteHandler` is recreated on every render, the `useMemo` hook re-runs, creating a new `CoreToolScheduler` instance each time.

**Impact**

*   **Performance Degradation:** Causes wasteful object creation and garbage collection churn on every render cycle.
*   **Memory Leak Risk:** Any state or listeners within `CoreToolScheduler` (such as message bus subscriptions) will be duplicated, leading to significant memory leaks over the lifetime of the application.

**Proposed Solution**

The inline `onComplete` callback in `useGeminiStream.ts` should be memoized using `useCallback` to ensure its reference is stable across renders.

**File to Edit:** `packages/cli/src/ui/hooks/useGeminiStream.ts`

```typescript
// Wrap the onComplete handler in useCallback
const handleCompletedToolsFromScheduler = useCallback(
  async (completedToolCallsFromScheduler: CompletedToolCall[]) => {
    // This onComplete is called when ALL scheduled tools for a given batch are done.
    if (completedToolCallsFromScheduler.length > 0) {
      // Add the final state of these tools to the history for display.
      addItem(
        mapTrackedToolCallsToDisplay(
          completedToolCallsFromScheduler as TrackedToolCall[],
        ),
        Date.now(),
      );

      // Record tool calls with full metadata before sending responses.
      try {
        const currentModel =
          config.getGeminiClient().getCurrentSequenceModel() ??
          config.getModel();
        config
          .getGeminiClient()
          .getChat()
          .recordCompletedToolCalls(
            currentModel,
            completedToolCallsFromScheduler,
          );
      } catch (error) {
        console.error(
          `Error recording completed tool call information: ${error}`,
        );
      }

      // Handle tool response submission immediately when tools complete
      await handleCompletedTools(
        completedToolCallsFromScheduler as TrackedToolCall[],
      );
    }
  },
  // Add stable dependencies here.
  // IMPORTANT: `handleCompletedTools` itself must also be stable.
  [config, addItem, handleCompletedTools],
);

const [toolCalls, scheduleToolCalls, markToolsAsSubmitted] =
  useReactToolScheduler(
    handleCompletedToolsFromScheduler, // Pass the now-stable callback
    config,
    getPreferredEditor,
    onEditorClose,
  );
```
**Note:** This fix assumes that `handleCompletedTools`, `addItem`, and `config` are stable. The `handleCompletedTools` function within `useGeminiStream.ts` will also need to be wrapped in `useCallback`.

**Verification Steps**

1.  Add a debug log to the `CoreToolScheduler` constructor:
    ```typescript
    // File: packages/core/src/core/coreToolScheduler.ts
    constructor(options: CoreToolSchedulerOptions) {
      console.debug('[DEBUG] CoreToolScheduler instance created');
      // ...
    }
    ```
2.  Run the CLI.
3.  Trigger a tool call.
4.  **Expected Behavior (After Fix):** The "[DEBUG] CoreToolScheduler instance created" message should appear only once when the tool is scheduled.
5.  **Current Behavior (Before Fix):** The debug message will appear multiple times during the render cycles.

jacob314:
Drive by fixed a couple unstable inputs passed from AppContainer and fixed the core issue in useReactToolScheduler This surfaced a bug in CoreToolScheduler where it was caching the tool list but it was unchanged.

## TLDR

Moderate shock this didn't cause more problems.
One of the args to useReactToolScheduler was a raw async Function that couldn't possibly have a stable value.
 
Fixes: #11753

### DIFF
--- a/packages/cli/src/ui/AppContainer.tsx
+++ b/packages/cli/src/ui/AppContainer.tsx
@@ -582,6 +582,15 @@ Logging in with Google... Please restart Gemini CLI to continue.
 
   const cancelHandlerRef = useRef<() => void>(() => {});
 
+  const getPreferredEditor = useCallback(
+    () => settings.merged.general?.preferredEditor as EditorType,
+    [settings.merged.general?.preferredEditor],
+  );
+
+  const onCancelSubmit = useCallback(() => {
+    cancelHandlerRef.current();
+  }, []);
+
   const {
     streamingState,
     submitQuery,
@@ -601,13 +610,13 @@ Logging in with Google... Please restart Gemini CLI to continue.
     setDebugMessage,
     handleSlashCommand,
     shellModeActive,
-    () => settings.merged.general?.preferredEditor as EditorType,
+    getPreferredEditor,
     onAuthError,
     performMemoryRefresh,
     modelSwitchedFromQuotaError,
     setModelSwitchedFromQuotaError,
     refreshStatic,
-    () => cancelHandlerRef.current(),
+    onCancelSubmit,
     setEmbeddedShellFocused,
     terminalWidth,
     terminalHeight,
--- a/packages/cli/src/ui/hooks/useReactToolScheduler.test.ts
+++ b/packages/cli/src/ui/hooks/useReactToolScheduler.test.ts
@@ -1,0 +1,85 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { CoreToolScheduler } from '@google/gemini-cli-core';
+import type { Config } from '@google/gemini-cli-core';
+import { renderHook } from '@testing-library/react';
+import { vi, describe, it, expect, beforeEach } from 'vitest';
+import { useReactToolScheduler } from './useReactToolScheduler.js';
+
+vi.mock('@google/gemini-cli-core', async (importOriginal) => {
+  const actual =
+    await importOriginal<typeof import('@google/gemini-cli-core')>();
+  return {
+    ...actual,
+    CoreToolScheduler: vi.fn(),
+  };
+});
+
+const mockCoreToolScheduler = vi.mocked(CoreToolScheduler);
+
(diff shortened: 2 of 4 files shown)


allenhutchison (approved):
This is a much simpler fix that I feared it would be. Thanks for doing this.

Issue #3121, opened 2025-07-03
Feature request: ability to specify model temperature


harrymunro:
### What would you like to be added?

The ability to specify temperature, perhaps via passing in a settings file or with a / command.

### Why is this needed?

Better control over the model.

ivankazzzz:
up for this, really need temperature control ASAP

Neshoir:
and specify thinking mode、 top-P via parameters

PR #11800, opened 2025-10-23
feat: Add temperature CLI parameter

vinay-google:
## TLDR

This change introduces a new command-line interface (CLI) parameter, `--temperature` (with a `-t` alias), to allow users to control the temperature setting for the Gemini model. The temperature value is validated to be between 0 and 2. If no temperature is specified, the default value of 0 is used.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  |  ✅   | ❓  |  ✅   |

## Linked issues / bugs

#3121


packages/cli/src/config/settingsSchema.ts
```diff
@@ -557,6 +557,15 @@ const SETTINGS_SCHEMA = {
         description: 'The Gemini model to use for conversations.',
         showInDialog: false,
       },
+      temperature: {
```
matt-w-horn, line 560: This is useful! According to the contribution guidelines, user-facing changes should be documented. Since this pull request introduces a new `--temperature` flag, I'd suggest we update the documentation to reflect this.

In `docs/cli/commands.md`, we could add a new "Command-Line Flags" section with the following content:

```markdown
### Command-Line Flags

*   **`--temperature <value>`** or **`-t <value>`**
*   **Description:** Sets the temperature for the Gemini model. The temperature controls the randomness of the model's output. A higher value (e.g., 0.9) results in more creative and diverse responses, while a lower value (e.g., 0.2) makes the output more deterministic and focused. The value must be between 0.0 and 2.0.
```

matt-w-horn (approved):

packages/cli/src/config/settingsSchema.ts
vinay-google: Done. Thank you!

matt-w-horn: Oh I'm so sorry- I just double checked and noticed these descriptions in ```packages/cli/src/config/settingsSchema.ts```  are printed out in ```---help``` and we don't document any of the other CLI flags in the commands.md file. So the consistnet thing might be to just use this as-is and potentially expand the description. My apologies.

allenhutchison:
I think the right way to handle this is to create a settings schema for named model configs. I have a prototype of this in adh-cli where you name a model alias and then you can add parameters along with a base model id (i.e. gemini-2.5-pro for example. You can see the config format I'm using in that project here:

https://github.com/allenhutchison/adh-cli/blob/main/docs/MODEL_ALIASES.md

If we were to go down this route, then a user could have as many model aliases as they like and the --model flag could be adjusted to first look for an alias and then fall back to an actual model name.

joshualitt:
@allenhutchison @vinay-google , I'm actually tackling this in https://github.com/google-gemini/maintainers-gemini-cli/issues/873, i.e. adding a new settings schema for models, along with the ability to override configs. Hoping to have a PR later today or early tomorrow for discussion.

PR #11803, opened 2025-10-23
feat(ux): Surface internal errors via unified event system

abhipatel12:
## TLDR

This PR introduces a unified, decoupled mechanism for low-level services to report non-fatal errors, warnings, and info messages to the user, regardless of the active UI mode (interactive TUI or non-interactive CLI).

It replaces direct `console.error` calls—which can be swallowed, missed, or corrupt the TUI—with a new `coreEvents` emitter that safely routes these messages to the appropriate presentation layer.

## Dive Deeper

Previously, deep services like `FileCommandLoader` or `Settings` had no clean way to inform the user of non-fatal issues (like a malformed TOML file that is skipped, or a failure to save settings). They relied on `console.error`, which is problematic:
- In **Interactive Mode**, `console.error` might be patched, hidden, or corrupt the Ink rendering output.
- In **Non-Interactive Mode**, it might be missed if standard error isn't monitored closely, or it might pollute structured output if not handled correctly.

### Key Changes
1.  **`CoreEventEmitter`**: Added to `@google/gemini-cli-core`. It's a typed `EventEmitter` specifically for `UserFeedback` events. Crucially, it includes a **backlog** to buffer messages emitted before the UI layer has fully initialized and subscribed.
2.  **Interactive UI Integration**: `AppContainer` now subscribes to these events and renders them directly into the chat history as `INFO`, `WARNING`, or `ERROR` messages. This ensures users see startup warnings immediately upon entering the app.
3.  **Non-Interactive CLI Integration**: `runNonInteractive` subscribes and routes these events safely to `stderr`, ensuring they don't interfere with piped `stdout` (especially important for JSON output modes).
4.  **Settings Adoption**: Migrated `saveSettings` error handling to use this new system as a first consumer.

This infrastructure lays the groundwork for migrating other silent or messy `console.error` calls (like those in `FileCommandLoader`) to this new system in subsequent PRs.

### Manual Verification

To manually verify the end-to-end flow before more services adopt it, you can temporarily inject a test event:

1.  **Interactive Mode**:
    Add this to `packages/cli/src/gemini.tsx` before `startInteractiveUI` is called:
    ```typescript
    coreEvents.emitFeedback('warning', 'This is a test warning from the core event system!');
    ```
    Run `npm start`. You should see the warning appear in the chat history immediately after startup.

2.  **Non-Interactive Mode**:
    Add the same line before `runNonInteractive` in `packages/cli/src/gemini.tsx`.
    Run `npm start -- "hello"`. You should see the warning printed to stderr.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  |✅| ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt |✅| -   | -   |

## Linked issues / bugs

Related to #10761

### DIFF
--- a/packages/cli/src/nonInteractiveCli.ts
+++ b/packages/cli/src/nonInteractiveCli.ts
@@ -8,6 +8,7 @@ import type {
   Config,
   ToolCallRequestInfo,
   CompletedToolCall,
+  UserFeedbackPayload,
 } from '@google/gemini-cli-core';
 import { isSlashCommand } from './ui/utils/commandUtils.js';
 import type { LoadedSettings } from './config/settings.js';
@@ -24,6 +25,8 @@ import {
   JsonStreamEventType,
   uiTelemetryService,
   debugLogger,
+  coreEvents,
+  CoreEvent,
 } from '@google/gemini-cli-core';
 
 import type { Content, Part } from '@google/genai';
@@ -50,6 +53,14 @@ export async function runNonInteractive(
       debugMode: config.getDebugMode(),
     });
 
+    const handleUserFeedback = (payload: UserFeedbackPayload) => {
+      const prefix = payload.severity.toUpperCase();
+      console.error(`[${prefix}] ${payload.message}`);
+      if (payload.error && config.getDebugMode()) {
+        console.error(payload.error);
+      }
+    };
+
     const startTime = Date.now();
     const streamFormatter =
       config.getOutputFormat() === OutputFormat.STREAM_JSON
@@ -58,6 +69,9 @@ export async function runNonInteractive(
 
     try {
       consolePatcher.patch();
+      coreEvents.on(CoreEvent.UserFeedback, handleUserFeedback);
+      coreEvents.drainFeedbackBacklog();
+
       // Handle EPIPE errors when the output is piped to a command that closes early.
       process.stdout.on('error', (err: NodeJS.ErrnoException) => {
         if (err.code === 'EPIPE') {
@@ -287,6 +301,7 @@ export async function runNonInteractive(
       handleError(error, config);
     } finally {
       consolePatcher.cleanup();
+      coreEvents.off(CoreEvent.UserFeedback, handleUserFeedback);
       if (isTelemetrySdkInitialized()) {
         await shutdownTelemetry(config);
       }
--- a/packages/cli/src/nonInteractiveCli.ts
+++ b/packages/cli/src/nonInteractiveCli.ts
@@ -55,9 +55,13 @@ export async function runNonInteractive(
 
     const handleUserFeedback = (payload: UserFeedbackPayload) => {
       const prefix = payload.severity.toUpperCase();
-      console.error(`[${prefix}] ${payload.message}`);
+      process.stderr.write(`[${prefix}] ${payload.message}\n`);
       if (payload.error && config.getDebugMode()) {
-        console.error(payload.error);
+        const errorToLog =
+          payload.error instanceof Error
+            ? payload.error.stack || payload.error.message
+            : String(payload.error);
+        process.stderr.write(`${errorToLog}\n`);
       }
     };
 
--- a/packages/cli/src/ui/AppContainer.tsx
+++ b/packages/cli/src/ui/AppContainer.tsx
@@ -1080,9 +1080,12 @@ Logging in with Google... Please restart Gemini CLI to continue.
           type = MessageType.WARNING;
           break;
         case 'info':
-        default:
           type = MessageType.INFO;
           break;
+        default:
+          throw new Error(
+            `Unexpected severity for user feedback: ${payload.severity}`,
+          );
       }
 
       historyManager.addItem(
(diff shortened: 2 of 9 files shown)


packages/core/src/utils/events.ts
```diff
@@ -0,0 +1,117 @@
...
+  UserFeedback = 'user-feedback',
+}
+
+export class CoreEventEmitter extends EventEmitter {
+  private _feedbackBacklog: UserFeedbackPayload[] = [];
+  private static readonly MAX_BACKLOG_SIZE = 100;
+
+  constructor() {
+    super();
+    this.setMaxListeners(20);
```
jacob314: naive question. Why does this need to be set?

packages/core/src/utils/events.ts
```diff
@@ -0,0 +1,117 @@
...
+   */
+  drainFeedbackBacklog(): void {
+    let item = this._feedbackBacklog.shift();
+    while (item) {
+      this.emit(CoreEvent.UserFeedback, item);
+      item = this._feedbackBacklog.shift();
+    }
+  }
+
+  // --- Minimal Type Overrides ---
```
jacob314: nit: remove this comment.

packages/cli/src/nonInteractiveCli.ts
```diff
@@ -50,6 +53,14 @@ export async function runNonInteractive(
       debugMode: config.getDebugMode(),
     });
 
+    const handleUserFeedback = (payload: UserFeedbackPayload) => {
+      const prefix = payload.severity.toUpperCase();
+      console.error(`[${prefix}] ${payload.message}`);
```
jacob314: should we use console.error or should we be directly writing this to stdout or stderr?

packages/core/src/utils/events.ts
```diff
@@ -0,0 +1,117 @@
...
+  error?: unknown;
+}
+
+export enum CoreEvent {
+  UserFeedback = 'user-feedback',
+}
+
+export class CoreEventEmitter extends EventEmitter {
+  private _feedbackBacklog: UserFeedbackPayload[] = [];
+  private static readonly MAX_BACKLOG_SIZE = 100;
```
jacob314: nit: make MAX_BACKLOG_SIZE 10,000 or something. No reason to truncate before we need to.

packages/cli/src/ui/AppContainer.tsx
```diff
@@ -1057,6 +1060,50 @@ Logging in with Google... Please restart Gemini CLI to continue.
...
+      let type: MessageType;
+      switch (payload.severity) {
+        case 'error':
+          type = MessageType.ERROR;
+          break;
+        case 'warning':
+          type = MessageType.WARNING;
+          break;
+        case 'info':
+        default:
```
jacob314: nit: should we throw if an unexpected sev is encountered rather than quietly calling it info?

packages/cli/src/ui/AppContainer.tsx
```diff
@@ -1057,6 +1060,50 @@ Logging in with Google... Please restart Gemini CLI to continue.
...
+      historyManager.addItem(
+        {
+          type,
+          text: payload.message,
+        },
+        Date.now(),
+      );
+
+      // If there is an attached error object, log it to the debug drawer.
+      if (payload.error) {
```
jacob314, line 1100: like this!

jacob314 (commented):
This is great! Approved once minor nits are addressed.

packages/cli/src/ui/AppContainer.tsx
abhipatel12: Done

packages/cli/src/nonInteractiveCli.ts
abhipatel12: Done. changed to `process.stderr.write`

packages/core/src/utils/events.ts
abhipatel12: Done

packages/core/src/utils/events.ts
abhipatel12: Nope, good point. removed. We shouldn't ever go beyond the 10 default limit

packages/core/src/utils/events.ts
abhipatel12: Done

packages/core/src/utils/events.ts
```diff
@@ -0,0 +1,117 @@
...
+  /**
+   * Flushes buffered messages. Call this immediately after primary UI listener
+   * subscribes.
+   */
+  drainFeedbackBacklog(): void {
+    let item = this._feedbackBacklog.shift();
+    while (item) {
+      this.emit(CoreEvent.UserFeedback, item);
+      item = this._feedbackBacklog.shift();
+    }
```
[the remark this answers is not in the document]
abhipatel12: Done

abhipatel12 (commented):
Thanks, lmk how this looks

jacob314 (approved):
lgtm

PR #11805, opened 2025-10-23
fix(core): include user prompt in LLM loop detection

SandyTao520:
## TLDR

Included the user's initial prompt in the LLM-based loop detection check. This provides the model with the original goal, improving its ability to differentiate between productive long-running tasks and actual loops. This also fixes an issue where the API rejects conversation history starting with a function call.

## Dive Deeper

*   Updated `LoopDetectionService` to store the `userPrompt` when `reset()` is called.
*   Updated `GeminiClient` to pass the current request (user prompt) to `loopDetector.reset()` at the start of a turn sequence.
*   Prepend the `userPrompt` to the history sent to the LLM during `checkForLoopWithLLM`.
*   Added unit tests to verify the user prompt is correctly included and ordered in the LLM check payload.

## Reviewer Test Plan

1.  Run a complex, multi-step task that is likely to take many turns (e.g., "refactor this large directory").
2.  Observe debug logs (if enabled) or monitor network traffic to see the `generateJson` call for loop detection (triggered after ~30 turns).
3.  Verify that the `contents` sent to the loop detection model include the user's original prompt at the beginning.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓ |
| npx      | ✅  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

### DIFF
--- a/packages/core/src/core/client.ts
+++ b/packages/core/src/core/client.ts
@@ -482,7 +482,7 @@ My setup is complete. I will provide my first command in the next turn.
     isInvalidStreamRetry: boolean = false,
   ): AsyncGenerator<ServerGeminiStreamEvent, Turn> {
     if (this.lastPromptId !== prompt_id) {
-      this.loopDetector.reset(prompt_id);
+      this.loopDetector.reset(prompt_id, request);
       this.lastPromptId = prompt_id;
       this.currentSequenceModel = null;
     }
--- a/packages/core/src/core/client.test.ts
+++ b/packages/core/src/core/client.test.ts
@@ -112,6 +112,7 @@ vi.mock('../telemetry/index.js', () => ({
   logApiError: vi.fn(),
 }));
 vi.mock('../ide/ideContext.js');
+vi.mock('../services/loopDetectionService.js');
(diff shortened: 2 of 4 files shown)


jacob314 (commented):
lgtm
Will defer to @anj-s on quality aspects of this. My take is this is a case where it would be good if we could run an experiment on N% of our users to make sure this doesn't drastically increase the # of loops detected.

packages/core/src/core/client.ts
```diff
@@ -482,7 +482,7 @@ My setup is complete. I will provide my first command in the next turn.
     isInvalidStreamRetry: boolean = false,
   ): AsyncGenerator<ServerGeminiStreamEvent, Turn> {
     if (this.lastPromptId !== prompt_id) {
-      this.loopDetector.reset(prompt_id);
+      this.loopDetector.reset(prompt_id, request);
```
jacob314, line 485: what if the request is very large, e.g. I `@packages/cli`
Will the loop detection service now get 800K tokens and run very slowly?

Issue #11766, opened 2025-10-22
Shell command allowlists do not apply to chained or redirected commands


sniemczyk-google:
### What happened?

The fix for issue #11510 addressed a security vulnerability where shell command allowlists were not being checked for all commands in a pipe chain. However, this fix appears to be incomplete, as it does not cover other forms of shell command chaining and redirection, such as `&&`, `||`, `;`, and `>`.

When a shell command execution is approved, the check is only performed against the first command in a chain. Subsequent commands are executed without being checked against the allowlist. For instance, if `ls` is an allowed tool but `rm` is not, a command like `ls && rm -rf /` will still execute the `rm` command, bypassing the allowlist.

This was pointed out in a comment in the "Avoiding --yolo in Gemini CLI for Safer Automation" document.

**Test Case: Chained `rm` is incorrectly allowed**

When `ls` is allowed but `rm` is not, a chained command still executes, bypassing the allowlist for the second command.

```
$ gemini -e none --allowed-tools="ShellTool(true)" "Run the shell command 'true && rm important_file.txt' and show me the output"
This command will first run `true` (which does nothing) and then, because the first command succeeded, it will permanently delete the file `important_file.txt`.Done.

$ ls important_file.txt
ls: cannot access 'important_file.txt': No such file or directory
```
**NOTE:** *This should have been denied because `rm` is not in the allowlist.*
```


### What did you expect to happen?

I expected every command in a shell command chain to be individually verified against the configured tool allowlist. If any command in the chain is not permitted, the entire command should be rejected. This is how it now works for piped commands, and the same logic should be applied to all shell chaining and redirection operators.


### Client information

<details>
<summary>Client Information</summary>

Run `gemini` to enter the interactive CLI, then run the `/about` command.

```console
> /about
╭───────────────────────────────────────────────────────────────────────────────────────╮
│                                                                                       │
│ About Gemini CLI                                                                      │
│                                                                                       │
│ CLI Version                   0.11.0-nightly.20251020.a96f0659-git.7dd2d8f7           │
│ Git Commit                    7dd2d8f7                                                │
│ Model                         gemini-2.5-pro                                          │
│ Sandbox                       no sandbox                                              │
│ OS                            linux                                                   │
│ Auth Method                   gemini-api-key                                          │
│                                                                                       │
╰───────────────────────────────────────────────────────────────────────────────────────╯
```

</details>

### Login information

API Key

### Anything else we need to know?

[...]

PR #11813, opened 2025-10-23
fix: align shell allowlist handling (#11510)

### LINKED ISSUE #11510: Shell command allowlists do not apply to piped commands

cornmander:
## Summary
- ensure non-interactive shell confirmations reuse the segmented allowlist parser
- add targeted unit coverage for chained shell commands outside interactive mode
- expand integration tests to verify chained shell prompt injection is rejected when only the first segment is allowlisted

## Testing
- npx vitest run packages/core/src/tools/shell.test.ts
- GEMINI_API_KEY=<redacted> npx vitest run --root ./integration-tests run_shell_command.test.ts

Fixes https://github.com/google-gemini/gemini-cli/issues/11510 and https://github.com/google-gemini/gemini-cli/issues/11766

### DIFF
--- a/integration-tests/run_shell_command.test.ts
+++ b/integration-tests/run_shell_command.test.ts
@@ -449,6 +449,7 @@ describe('run_shell_command', () => {
       .readToolLogs()
       .filter((log) => log.toolRequest.name === 'run_shell_command');
 
+    // Success is false because tool is in the scheduled state.
     for (const log of toolLogs) {
       expect(log.toolRequest.success).toBe(false);
       expect(log.toolRequest.args).toContain('&&');
--- a/integration-tests/run_shell_command.test.ts
+++ b/integration-tests/run_shell_command.test.ts
@@ -61,6 +61,28 @@ function getDisallowedFileReadCommand(testFile: string): {
   }
 }
 
+function getChainedEchoCommand(): { allowPattern: string; command: string } {
+  const secondCommand = getAllowedListCommand();
+  switch (shell) {
+    case 'powershell':
+      return {
+        allowPattern: 'Write-Output',
+        command: `Write-Output "foo" && ${secondCommand}`,
+      };
+    case 'cmd':
+      return {
+        allowPattern: 'echo',
+        command: `echo "foo" && ${secondCommand}`,
+      };
+    case 'bash':
+    default:
+      return {
+        allowPattern: 'echo',
+        command: `echo "foo" && ${secondCommand}`,
+      };
+  }
+}
+
 describe('run_shell_command', () => {
   it('should be able to run a shell command', async () => {
     const rig = new TestRig();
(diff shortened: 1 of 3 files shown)


integration-tests/run_shell_command.test.ts
```diff
@@ -405,6 +427,34 @@ describe('run_shell_command', () => {
...
+      `--allowed-tools=ShellTool(${chained.allowPattern})`,
+    );
+
+    // CLI should refuse to execute the chained command without scheduling run_shell_command.
+    const toolLogs = rig
+      .readToolLogs()
+      .filter((log) => log.toolRequest.name === 'run_shell_command');
+
+    for (const log of toolLogs) {
+      expect(log.toolRequest.success).toBe(false);
```
skeshive, line 454: can you add a comment here that the request isnt successful because its scheduled to execute (not because it failed)

skeshive (approved):

Issue #11226, opened 2025-10-15
Incomplete Model Messages not added to Chat Service


adamfweidman:
### What happened?

If a user abort a chat while the model is in the middle of generating a message, the incomplete message does not get added to teh ChatRecordingSevice. Thoughts and ToolCall are added as they come through but messages wait until they have fully completed. 
If after cancelation the user runs /chat save <name>, the incomplete model output is not present in the saved chat. (I am not sure how this gets handled for in-memory chat).

### What did you expect to happen?

Partial message response to be saved to ChatRecordingService

### Client information

<details>
<summary>Client Information</summary>

Run `gemini` to enter the interactive CLI, then run the `/about` command.

```console
> /about
# paste output here
```

</details>

adamfweidman:
@hugomurillomtz I confirmed this is not being added to the saved chat service. I had not verified whether or not its being added to the internal chat history (this has possible cache/price impacts).

hugomurillomtz:
> [@hugomurillomtz](https://github.com/hugomurillomtz) I confirmed this is not being added to the saved chat service. I had not verified whether or not its being added to the internal chat history (this has possible cache/price impacts).

All right. I'm looking into it.

hugomurillomtz:
> [@hugomurillomtz](https://github.com/hugomurillomtz) I confirmed this is not being added to the saved chat service. I had not verified whether or not its being added to the internal chat history (this has possible cache/price impacts).

@adamfweidman question. Are you seeing this happen when your first and only model response is cancelled midway through? Like so:

<img width="1568" height="663" alt="Image" src="https://github.com/user-attachments/assets/f1f2b45a-9091-41b6-8b8d-45afdcf2ef0a" />

I'm seeing it happen in such case. However, if there's at least one model response, the chat _does_ get saved successfully. Is this the right scenario? Keep me honest here please.

adamfweidman:
@hugomurillomtz honestly I have not exhausted all the cases. I think the case I specifically looked at was when the model output a ToolCall and then continued to output text. I'd cancel during streaming after the toolCall and that would cause the error.

If these are different bugs feel free to create individual bugs for each!

hugomurillomtz:
> [@hugomurillomtz](https://github.com/hugomurillomtz) honestly I have not exhausted all the cases. I think the case I specifically looked at was when the model output a ToolCall and then continued to output text. I'd cancel during streaming after the toolCall and that would cause the error.

[...]

PR #11816, opened 2025-10-23
fix(#11226): incomplete model messages not added to chat services

hugomurillomtz:
## TLDR

This is about fixing the issue about incomplete/cancelled model responses not being saved to files when doing `/chat save [tagName]`.

## Dive Deeper

The issue describes this happening when the model uses some tools, and then continues to add a response, and upon cancellation and then doing `/chat save [tagName]`, the model incomplete response would not get stored. However, after looking into it, I noticed that it was a matter of incomplete model responses being disregarded regardless of whether there were tool calls before or after or not. Here's an example:

Request cancelled:
image

Then, when attempting to save the chat, we get this error:
image

And this was happening because a validation was taking place before pushing the model response to _History_: https://github.com/google-gemini/gemini-cli/blob/ce655436ef97535247daa9d27f572c1ca3ed62ac/packages/core/src/core/geminiChat.ts#L584
So, my quick solution was to move this line above such validation. However, I asked the model if this was a good solution and it said no. Here's why:

Imagine this scenario:
       1. You ask: "What is the capital of France?"
       2. The model starts streaming: "The capital of France is Pa-" and then the stream fails.
       3. The simple solution saves that partial response "The capital of France is Pa-" to the history.
       4. You then ask a new question: "What about Germany?"
       5. To get an answer, the system sends the entire history to the model as context. The model sees this:
          > User: "What is the capital of France?"
          > Model: "The capital of France is Pa-"
          > User: "What about Germany?"
       6. The model is now looking at a "dirty" or "polluted" conversation history. Its last turn was incomplete and nonsensical. This can confuse the model, leading it to give a strange or irrelevant answer to your new question about Germany.

The simple solution treated the user's view and the model's context as the same thing, when they needed to be different after a failure.

Because of that, I thought of a better solution by creating a way to distinguish between complete and partial messages. I thought about adding a flag property that would indicate if the response was a partial one or not. That way we can resume conversations with incomplete responses, but a clean, curated history with only successful turns for the model.

For this, I already have some progress. LMK how I should proceed.

## Reviewer Test Plan

Cancel an ongoing request to the model, and then save the conversation by doing `/chat save [tagName]`, and you should be able to successfully save the conversation.

image

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | x  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

This PR solves: https://github.com/google-gemini/gemini-cli/issues/11226


hugomurillomtz:
@adamfweidman, sorry, some tests are failing. I'm looking into it. Everything was passing locally.

PR #11817, opened 2025-10-23
fix: disable git hooks during extension install

cornmander:
## Summary
- ensure extension install step runs with hooks disabled so git config stays untouched
- add regression test covering git hook behavior in the extension flow

## Testing
- npx vitest run integration-tests/extensions-git-hooks.test.ts

### DIFF
--- a/packages/cli/src/config/extensions/github.ts
+++ b/packages/cli/src/config/extensions/github.ts
@@ -22,6 +22,8 @@ import extract from 'extract-zip';
 import { fetchJson, getGitHubToken } from './github_fetch.js';
 import { type ExtensionEnablementManager } from './extensionEnablement.js';
 
+const HOOKS_DISABLED_DIR = 'hooks-disabled';
+
 /**
  * Clones a Git repository to a specified local path.
  * @param installMetadata The metadata for the extension to install.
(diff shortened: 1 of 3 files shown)


packages/cli/src/config/extensions/github.ts
```diff
@@ -32,7 +34,17 @@ export async function cloneFromGit(
   destination: string,
 ): Promise<void> {
   try {
-    const git = simpleGit(destination);
+    const cloneEnv: NodeJS.ProcessEnv = { ...process.env };
+    const existingConfigCountRaw = cloneEnv['GIT_CONFIG_COUNT'];
```
jakemac53, line 38: I am really confused what all this is doing - why do we need separate variables for each install when they all hold the same value (is that what this is doing)?

packages/cli/src/config/extensions/github.ts
```diff
@@ -52,7 +64,15 @@ export async function cloneFromGit(
         // We let git handle the source as is.
       }
     }
-    await git.clone(sourceUrl, './', ['--depth', '1']);
+    await simpleGit()
+      .env(cloneEnv)
+      .clone(sourceUrl, destination, ['--depth', '1']);
```
jakemac53, line 69: Afaict, no hooks run on `clone` ever (there is no post-clone hook for git that I can find).

Can we remove everything except the part below where we create the hooks disabled dir prior to  doing any checkouts?

PR #11821, opened 2025-10-23
Added parameterization to base-storage-token.test and prompts.test.ts

IamRiddhi:
## TLDR
Refactored test suites in `base-storage-token.test.ts` and `prompts.test.ts` to use parameterized tests (`it.each`), reducing code duplication and improving readability.

## What’s Changed
- Applied parameterization (`it.each`) to repetitive test cases
- Consolidated similar tests into single blocks with shared logic and a table of test scenarios.
- Removed redundant `it` blocks and replaced them with data-driven tests.

## Why?
- **Readability:** Parameterized tests make it clear which cases are being tested, and new cases can be added easily.
- **Maintainability:** Adding or modifying edge cases only requires changing test data, not duplicating logic.
- **Coverage:** Ensures all relevant scenarios are tested in a concise format.

## Test Plan
- All existing and new tests pass with `npm run test`.
- No change to test coverage or assertions—behavior is preserved.
- Manual review confirms tests are equivalent and easier to read.

### DIFF
--- a/packages/core/src/core/prompts.test.ts
+++ b/packages/core/src/core/prompts.test.ts
@@ -57,7 +57,6 @@ describe('Core System Prompt (prompts.ts)', () => {
   });
 
   it.each([
-    { desc: 'no userMemory is provided', userMemory: '' },
     { desc: 'userMemory is empty string', userMemory: '' },
     { desc: 'userMemory is whitespace only', userMemory: '   \n  \t ' },
   ])('should return the base prompt when $desc', ({ userMemory }) => {
--- a/packages/core/src/core/prompts.test.ts
+++ b/packages/core/src/core/prompts.test.ts
@@ -56,28 +56,16 @@ describe('Core System Prompt (prompts.ts)', () => {
     } as unknown as Config;
   });
 
-  it('should return the base prompt when no userMemory is provided', () => {
+  it.each([
+    { desc: 'no userMemory is provided', userMemory: '' },
+    { desc: 'userMemory is empty string', userMemory: '' },
+    { desc: 'userMemory is whitespace only', userMemory: '   \n  \t ' },
+  ])('should return the base prompt when $desc', ({ userMemory }) => {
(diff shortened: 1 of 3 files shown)


packages/core/src/core/prompts.test.ts
```diff
@@ -210,27 +196,22 @@ describe('Core System Prompt (prompts.ts)', () => {
...
-      vi.stubEnv('GEMINI_SYSTEM_MD', '1');
-      vi.mocked(fs.existsSync).mockReturnValue(true);
-      vi.mocked(fs.readFileSync).mockReturnValue('custom system prompt');
-
-      const prompt = getCoreSystemPrompt(mockConfig);
-      expect(fs.readFileSync).toHaveBeenCalledWith(defaultPath, 'utf8');
-      expect(prompt).toBe('custom system prompt');
-    });
+    it.each([
+      { desc: '"true"', value: 'true' },
```
scidomino: Here and elsewhere, avoid duplicating info in the description. It would be clearer to just have

it.each(['true', '1'])

scidomino (approved):

packages/core/src/core/prompts.test.ts
```diff
@@ -262,37 +237,27 @@ describe('Core System Prompt (prompts.ts)', () => {
...
+    it.each(['false', '0'])(
+      'should not write to file when GEMINI_WRITE_SYSTEM_MD is "%s"',
+      (value) => {
+        vi.stubEnv('GEMINI_WRITE_SYSTEM_MD', value);
+        getCoreSystemPrompt(mockConfig);
+        expect(fs.writeFileSync).not.toHaveBeenCalled();
+      },
+    );
+
+    it.each(['true', '1'])(
```
MJjainam, line 249: Why was this in it.each if there was only one array?

IamRiddhi: i will modify this in my next pr, thanks!

PR #11837, opened 2025-10-23
Use raw writes to stdin where possible in tests

scidomino:
## TLDR

Use raw writes to stdin where possible in tests.

## Dive Deeper

Tests that issue `keypress` events are sidestepping node's `readline`. It's better for these tests to do simple writes to `stdin` instead and let `readline` generate the approprirate `keypress` events.

Also refactor these tests to be easier to read.

## Linked issues / bugs

For #7773


jacob314 (approved):
lgtm

PR #11871, opened 2025-10-23
Use raw writes to stdin in test

scidomino:
## TLDR

Use raw writes to stdin in this test.

## Dive Deeper

Tests that issue `keypress` events are sidestepping node's `readline`. It's better for these tests to do simple writes to `stdin` instead and let `readline` generate the approprirate `keypress` events.

Also refactor these tests to be easier to read.

## Linked issues / bugs

For #7773

### DIFF
--- a/packages/cli/src/ui/contexts/KeypressContext.test.tsx
+++ b/packages/cli/src/ui/contexts/KeypressContext.test.tsx
@@ -33,6 +33,9 @@ vi.mock('ink', async (importOriginal) => {
 
 const PASTE_START = '\x1B[200~';
 const PASTE_END = '\x1B[201~';
+// readline will not emit most incomplete kitty sequences but it will give
+// up on sequences like this where the modifier (135) has more than two digits.
+const INCOMPLETE_KITTY_SEQUENCE = '\x1b[97;135';
 
 class MockStdin extends EventEmitter {
   isTTY = true;
@@ -44,15 +47,6 @@ class MockStdin extends EventEmitter {
 
   write(text: string) {
     this.emit('data', Buffer.from(text));
-  }
-
-  /**
-   * Used to directly simulate keyPress events. Certain keypress events might
-   * be impossible to fire in certain versions of node. This allows us to
-   * sidestep readline entirely and just emit the keypress we want.
-   */
-  pressKey(key: Partial<Key>) {
-    this.emit('keypress', null, key);
   }
 }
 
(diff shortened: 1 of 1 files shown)


packages/cli/src/ui/contexts/KeypressContext.test.tsx
```diff
@@ -33,6 +33,7 @@ vi.mock('ink', async (importOriginal) => {
 
 const PASTE_START = '\x1B[200~';
 const PASTE_END = '\x1B[201~';
+const INCOMPLETED_KITTY_SEQUENCE = '\x1b[97;135';
```
jacob314: nit: INCOMPLETED -> INCOMPLETE

packages/cli/src/ui/contexts/KeypressContext.test.tsx
```diff
@@ -829,93 +817,75 @@ describe('Kitty Sequence Parsing', () => {
...
 
-  // Key mappings: letter -> [keycode, accented character, shouldHaveMeta]
-  // Note: µ (mu) is sent with meta:false on iTerm2/VSCode
-  const keys: Record<string, [number, string, boolean]> = {
-    a: [97, 'å', true],
-    o: [111, 'ø', true],
-    m: [109, 'µ', false],
+  // Key mappings: letter -> [keycode, accented character]
+  const keys: Record<string, [number, string]> = {
+    a: [97, 'å'],
```
jacob314, line 824: why did shouldHaveMeta get removed?

jacob314 (approved):
Approved with a couple minor nits. thanks for paying down this tech debt!

packages/cli/src/ui/contexts/KeypressContext.test.tsx
scidomino: I'll be honest, I've been trying to figure that one out. Because logically, it isn't possible for readline to emit a keypress with `sequence="å"` and `meta=true`. `meta=true` only happens when there's an escape character and that would have to be included in the `sequence` field.

Either way, it shouldn't matter now because the code we are testing should always yield `meta=true` regardless of what is sent to us.

packages/cli/src/ui/contexts/KeypressContext.test.tsx
scidomino: done.

jacob314 (approved):
lgtm

PR #11894, opened 2025-10-24
Record model responses with --record-responses (for use in testing)

jakemac53:
## TLDR

Adds `--record-responses <path>` which will stream all Gemini responses to the specified file in the same format that `--fake-responses <path>` expects.

Adds support for a `REGENERATE_MODEL_GOLDENS=true` environment variable - if present then integration tests which use fake model responses will use the real model, and update the files with the real content from that run.

## Dive Deeper

Also updates the file format so it is streaming write friendly.

Also updates the FakeContentGenerator to set prototypes on the JSON objects so they actually have all the expected methods like `text()`, `functionCalls()` etc. We don't use these today but it will be more future proof. There is no exposed public way of creating these response types from JSON unfortunately.

## Reviewer Test Plan

You can just pass `--record-responses <file-path>`, and then talk a bit with gemini, or you can also try running `REGENERATE_MODEL_GOLDENS=true npm run test:e2e context-compression` to regenerate the actual responses for that test.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ✅   |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Second part of #11062 - we have the ability already to load model responses as mocks, this adds the ability to write actual responses to files in the same format, and update actual goldens as well.

The final part to this would be creating a separate location for these tests, and not requiring authentication at all to run them.


packages/cli/src/config/config.ts
```diff
@@ -655,9 +662,24 @@ export async function loadCliConfig(
     codebaseInvestigatorSettings:
       settings.experimental?.codebaseInvestigatorSettings,
     fakeResponses: argv.fakeResponses,
+    recordResponses: argv.recordResponses,
     retryFetchErrors: settings.general?.retryFetchErrors ?? false,
     ptyInfo: ptyInfo?.name,
   });
+  // TODO: Consider moving `registerCleanup` into core, so this can move inside
+  // `createContentGenerator`.
+  const contentGenerator = config.getContentGenerator();
```
jakemac53: I definitely don't love this part, suggestions welcome... streaming the output to the file constantly is I think not a great solution as the file format is not conducive to that. Although, the file format could be changed to be more friendly?

packages/core/src/core/recordingContentGenerator.ts
```diff
@@ -0,0 +1,106 @@
...
+  async embedContent(
+    _request: EmbedContentParameters,
+  ): Promise<EmbedContentResponse> {
+    const response = await this.realGenerator.embedContent(_request);
+    this.recordedResponses.embedContent.push({
+      embeddings: response.embeddings,
+      metadata: response.metadata,
+    });
+    return response;
+  }
```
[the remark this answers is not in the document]
jakemac53: This is just flat out wrong, bad gemini

packages/core/src/core/recordingContentGenerator.test.ts
```diff
@@ -0,0 +1,208 @@
...
+      safeJsonStringify({
+        generateContent: [],
+        generateContentStream: [],
+        countTokens: [],
+        embedContent: [
+          {
+            embeddings: mockResponse.embeddings,
+            metadata: mockResponse.metadata,
+          },
+        ],
```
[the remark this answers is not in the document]
jakemac53: wrong

docs/integration-tests.md
```diff
@@ -56,6 +56,22 @@ To run a single test by its name, use the `--test-name-pattern` flag:
...
+
+To regenerate these golden files, set the REGENERATE_MODEL_GOLDENS environment
+variable to "true" when running the tests, for example:
+
+**WARNING**: If running locally you should review these updated responses for
+any information about yourself or your system that gemini may have included in
+these responses.
+
+```bash
+REGENERATE_MODEL_GOLDENS="true" npm run test:e2e
```
jacob314, line 72: rather than an ENV variable can we pass an arg to update the goldens? I'd suggest using --update-model-snapshots to align with --update-snapshots for golden image tests.

integration-tests/context-compress-interactive.compress-failure.responses
```diff
@@ -0,0 +1,2 @@
...e's an underlying objective.\n\n\n"}],"role":"model"},"index":0}],"usageMetadata":{"promptTokenCount":12270,"totalTokenCount":12341,"promptTokensDetails":[{"modality":"TEXT","tokenCount":12270}],"thoughtsTokenCount":71}},{"candidates":[{"content":{"parts":[{"thoughtSignature":"CiQB0e2Kb3dRh+BYdbZvmulSN2Pwbc75DfQOT3H4EN0rn039hoMKfwHR7YpvvyqNKoxXAiCbYw3gbcTr/+pegUpgnsIrt8oQPMytFMjKSsMyshfygc21T2MkyuI6Q5I/fNCcHROWexdZnIeppVCDB2TarN4LGW4T9Yci6n/ynMMFT2xc2/vyHpkDgRM7avhMElnBhuxAY+e4TpxkZIncGWCEHP1TouoKpgEB0e2Kb8Xpwm0hiKhPt2ZLizpxjk+CVtcbnlgv69xo5VsuQ+iNyrVGBGRwNx+eTeNGdGpn6e73WOCZeP91FwOZe7URyL12IA6E6gYWqw0kXJR4hO4p6Lwv49E3+FRiG2C4OKDF8LF5XorYyCHSgBFT1/RUAVj81GDTx1xxtmYKN3xq8Ri+HsPbqU/FM/jtNZKkXXAtufw2Bmw8lJfmugENIv/TQI7xCo8BAdHtim8KgAXJfZ7ASfutVLKTylQeaslyB/SmcHJ0ZiNr5j8WP1prZdb6XnZZ1ZNbhjxUf/ymoxHKGvtTPBgLE9azMj8Lx/k0clhd2a+wNsiIqW9qCzlVah0tBMytpQUjIDtQe9Hj4LLUprF9PUe/xJkj000Z0ZzsgFm2ncdTWZTdkhCQDpyETVAxdE+oklwKJAHR7YpvUjSkD6KwY1gLrOsHKy0UNfn2lMbxjVetKNMVBRqsTg==","text":"Hello."}],"role":"model"},"finishReason":"STOP","index":0}],"usageMetadata":{"promptTokenCount":12270,"totalTokenCount":12341,"promptTokensDetails":[{"modality":"TEXT","tokenCount":12270}],"thoughtsTokenCount":71}}]}
```
jacob314, line 1: should we pretty print these so the output is more human readable and there is more of a chance the diffs don't merge conflicts (doesn't really matter).

packages/cli/src/config/config.ts
```diff
@@ -199,6 +200,10 @@ export async function parseArguments(settings: Settings): Promise<CliArgs> {
           type: 'string',
           description: 'Path to a file with fake model responses for testing.',
         })
+        .option('record-responses', {
```
jacob314, line 204: can we avoid adding this as an option to Gemini CLI itself? Would worry this could cause user confusion.  I'd suggest using an env variable for this instead.

jakemac53: What if we just hide the flag?

docs/integration-tests.md
jakemac53: I don't know how to plumb that through to the actual tests - although I hadn't really looked at it either. But we don't have a command line parser in the TestRig or read any other command line flags I am aware of, so this seemed simpler.

Ultimately, I think we want custom commands anyways so it's just like `npm run regenerate:test:integration` etc? Similar to how deflake works. So, the fact this uses an environment variable would be hidden at that point.

integration-tests/context-compress-interactive.compress-failure.responses
jakemac53: Hmm that actually makes reading the files more complicated because now its not a valid JSON blob all on one line.

In keeping with the append only nature of these files which simplifies things greatly, I think we should stick with it as is.

packages/cli/src/config/config.ts
jakemac53: Fwiw, I think these arguments may also be useful for people wanting to write their own integration tests against Gemini CLI, so they aren't just for us necessarily.

jakemac53: I did push up a change to hide these arguments.

docs/integration-tests.md
jacob314: I would ask Gemini CLI to vibe code plumbing it through as an arg. if it can't do it correctly in 1 shot with code you would be happy approving fine to do it this way.

integration-tests/context-compress-interactive.compress-failure.responses
jacob314: fine to leave as is.

packages/cli/src/config/config.ts
jacob314: fine with it as long as it is hidden.

jacob314 (approved):
lgtm

PR #11923, opened 2025-10-24
(fix): appcontainer should not poll and footer should use currentModel from ui state

### LINKED ISSUE #11957: CLI shows `gemini-2.5-pro (x% context left)` even after a flash fallback
### What happened?

Image

The bottom bar text should say the current active model name. 

### Client information

Built from HEAD

### Login information

Google Account

However it isn't clear to me what the intent of that message is. For pro model it seems to represent the max context size a request can have, and the context growth as users interact with the CLI. Does the same context limitation apply for the Flash model? Is it the same limit?

psinha40898:
## TLDR

The model from config is not updated through the fallback logic. AppContainer polls and holds the right model in ui state but it's not used in Footer. Footer uses the model from config.

The way AppContainer polls for currentModel is not proper use of React

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

fixes #11957

### DIFF
--- a/packages/cli/src/test-utils/render.tsx
+++ b/packages/cli/src/test-utils/render.tsx
@@ -23,6 +23,7 @@ const mockConfig = {
   getTargetDir: () =>
     '/Users/test/project/foo/bar/and/some/more/directories/to/make/it/long',
   getDebugMode: () => false,
+  isInFallbackMode: () => false,
 };
 
 const configProxy = new Proxy(mockConfig, {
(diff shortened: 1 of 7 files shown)


psinha40898:
While it makes sense for the footer to use currentModel from ui state, the way AppContainer polls for the current model is not the proper way to use React.

Needs review wonder whether adding FallbackModeChanged to the existing coreEvents emitter (designed for user-facing messages) is appropriate architecture.

scidomino (approved):

PR #11953, opened 2025-10-24
Add support for sensitive keychain-stored per-extension settings

chrstnb:
## TLDR

Adding a new attribute for extensions settings 'sensitive'. If set, write the value to the keychain, stored per-extension-id. The setting will be loaded as an environment variable and passed to the extension.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


packages/cli/src/config/extensions/extensionSettings.ts
```diff
@@ -12,27 +12,40 @@ import { ExtensionStorage } from './storage.js';
 import type { ExtensionConfig } from '../extension.js';
 
 import prompts from 'prompts';
+import { KeychainTokenStorage } from '@google/gemini-cli-core';
 
 export interface ExtensionSetting {
   name: string;
   description: string;
   envVar: string;
+  sensitive?: boolean;
```
richieforeman, line 22: Super super petty nit: Consider adding a doc block comment on this about the "default sensitivity"

Something like:

/** Whether or not the value is "sensitive".  Sensitive values will be stored in the system key chain.  If sensitive is undefined, the setting is assumed not to be sensitive */

packages/cli/src/config/extensions/extensionSettings.ts
```diff
@@ -53,16 +66,34 @@ export async function maybePromptForSettings(
...
+  if (previousExtensionConfig?.settings) {
+    for (const oldSetting of previousExtensionConfig.settings) {
+      const newSetting = settings.find((s) => s.name === oldSetting.name);
+      if (!newSetting && oldSetting.sensitive) {
+        // Setting was removed and was sensitive
+        await keychain.deleteSecret(oldSetting.envVar).catch(() => {});
+      } else if (newSetting && oldSetting.sensitive && !newSetting.sensitive) {
+        // Setting is no longer sensitive
+        await keychain.deleteSecret(oldSetting.envVar).catch(() => {});
+      }
```
jakemac53: What about the case where it is changed to be sensitive, that seems like the most important one to handle :)

jakemac53 (approved):

PR #11964, opened 2025-10-24
First batch of fixing tests to use best practices.

jacob314:
## TLDR

I believe we were accidentally running a lot of tests in a react DOM env rather than a node env. Amazingly the tests all worked. this is causing problems with my ink fork as the things that just worked before with our wacky env now break things and it is easier to fix our tests than try to make my env wacky in the way our existing env.

## Dive Deeper
TLDR is to avoid `@testing-library/react` and use `react-test-renderer` instead. Use the vitest waitFor rather than the one from testing-library/react.
Still confused why the tests were working at all before. Would dig deeper but will be impossible to backslide with my ink fork as the incorrect pattern now fails.

Tests with `/** @vitest-environment jsdom */`
are ones that need to be migrated to use the node test environment but are left for the next PR of test fixes.


packages/cli/package.json
```diff
@@ -83,10 +84,15 @@
     "jsdom": "^26.1.0",
     "pretty-format": "^30.0.2",
     "react-dom": "^19.1.0",
+    "react-test-renderer": "^19.2.0",
```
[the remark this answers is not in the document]
jacob314: done

packages/cli/src/ui/components/InputPrompt.test.tsx
```diff
@@ -1214,7 +1214,7 @@ describe('InputPrompt', () => {
       await vi.runAllTimersAsync();
 
       // Simulate a paste operation (this should set the paste protection)
-      act(() => {
+      await act(async () => {
```
scidomino, line 1283: Here and elsewhere: Why is this async? We don't wait on anything in there.

scidomino (approved):

packages/cli/src/ui/components/InputPrompt.test.tsx
jacob314: because act in react-test-renderer is async. This is the act method we should have been using and even for the existing method we should probably have been awaiting it but weren't because failing to await it didn't trigger type errors.

PR #11968, opened 2025-10-24
fix(cli): Exit CLI when trust save unsuccessful during launch

shrutip90:
## TLDR

Got rid of the console.error message as @abhipatel12 requested.
Also updated /permissions command to show a similar error message and avoid restarting in case of failures.

**Failure message during launch:**
Screenshot 2025-10-24 at 11 19 55 AM

**Failure message from /permissions command:**
Screenshot 2025-10-24 at 11 20 52 AM

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes https://github.com/google-gemini/maintainers-gemini-cli/issues/919

### DIFF
--- a/packages/cli/src/ui/components/PermissionsModifyTrustDialog.test.tsx
+++ b/packages/cli/src/ui/components/PermissionsModifyTrustDialog.test.tsx
@@ -148,10 +148,11 @@ describe('PermissionsModifyTrustDialog', () => {
     });
   });
 
-  it('should commit, restart, and exit on `r` keypress', async () => {
+  it('should commit and restart `r` keypress', async () => {
     const mockRelaunchApp = vi
       .spyOn(processUtils, 'relaunchApp')
       .mockResolvedValue(undefined);
+    mockCommitTrustLevelChange.mockReturnValue(true);
     vi.mocked(usePermissionsModifyTrust).mockReturnValue({
       cwd: '/test/dir',
       currentTrustLevel: TrustLevel.DO_NOT_TRUST,
@@ -175,7 +176,6 @@ describe('PermissionsModifyTrustDialog', () => {
     await waitFor(() => {
       expect(mockCommitTrustLevelChange).toHaveBeenCalled();
       expect(mockRelaunchApp).toHaveBeenCalled();
-      expect(onExit).toHaveBeenCalled();
     });
 
     mockRelaunchApp.mockRestore();
--- a/packages/cli/src/ui/components/PermissionsModifyTrustDialog.tsx
+++ b/packages/cli/src/ui/components/PermissionsModifyTrustDialog.tsx
@@ -65,8 +65,9 @@ export function PermissionsModifyTrustDialog({
         const success = commitTrustLevelChange();
         if (success) {
           relaunchApp();
+        } else {
+          onExit();
         }
-        onExit();
       }
     },
     { isActive: true },
--- a/packages/cli/src/ui/components/PermissionsModifyTrustDialog.tsx
+++ b/packages/cli/src/ui/components/PermissionsModifyTrustDialog.tsx
@@ -62,8 +62,10 @@ export function PermissionsModifyTrustDialog({
         onExit();
       }
       if (needsRestart && key.name === 'r') {
-        commitTrustLevelChange();
-        relaunchApp();
+        const success = commitTrustLevelChange();
+        if (success) {
+          relaunchApp();
+        }
         onExit();
       }
     },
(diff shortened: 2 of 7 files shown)


packages/cli/src/ui/components/PermissionsModifyTrustDialog.tsx
```diff
@@ -62,8 +62,10 @@ export function PermissionsModifyTrustDialog({
         onExit();
       }
       if (needsRestart && key.name === 'r') {
-        commitTrustLevelChange();
-        relaunchApp();
+        const success = commitTrustLevelChange();
+        if (success) {
+          relaunchApp();
+        }
         onExit();
```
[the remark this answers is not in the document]
shrutip90: Done

packages/cli/src/ui/hooks/useFolderTrust.ts
```diff
@@ -67,7 +67,22 @@ export const useFolderTrust = (
...
       }
 
-      trustedFolders.setValue(cwd, trustLevel);
+      try {
+        trustedFolders.setValue(cwd, trustLevel);
+      } catch (_e) {
+        addItem(
+          {
+            type: MessageType.WARNING,
+            text: 'Failed to save trust settings. Exiting Gemini CLI.',
```
abhipatel12: Question, why do we treat this as a critical failure and exit but then mid-session, it's okay?

packages/cli/src/ui/hooks/usePermissionsModifyTrust.ts
```diff
@@ -101,7 +101,17 @@ export const usePermissionsModifyTrust = (
...
         const folders = loadTrustedFolders();
-        folders.setValue(cwd, trustLevel);
+        try {
+          folders.setValue(cwd, trustLevel);
+        } catch (_e) {
+          addItem(
+            {
+              type: MessageType.WARNING,
+              text: 'Failed to save trust settings. Your changes may not persist.',
+            },
```
abhipatel12: We should maybe add the error as a `warn` message that will show up in the debug drawer. That way the user has a "clean message" but then if we need to help debug someone or debug ourselves, we can find the larger trace via F12

abhipatel12: Also just curious, should we be directly adding things to the history here or reuse the emitter pattern when we have easy access to the necessary hook?

@jacob314 any thoughts?

This would also be considered in the other hook where we directly use `addItem` vs the events.ts

packages/cli/src/ui/components/PermissionsModifyTrustDialog.test.tsx
```diff
@@ -149,10 +149,11 @@ describe('PermissionsModifyTrustDialog', () => {
     });
   });
 
-  it('should commit, restart, and exit on `r` keypress', async () => {
+  it('should commit and restart `r` keypress', async () => {
```
abhipatel12, line 151: Can we add a test case for when there is a failure?

abhipatel12 (commented):
Took a look! Lmk what you think!

packages/cli/src/ui/hooks/useFolderTrust.ts
shrutip90: We need some trust level to operate at. So, at launch I've made it a critical failure. Another option could be to operate at untrusted level maybe. But right now when settings / trusted folders don't load correctly also, we exit with a critical failures, so I thought this might be alright.

Mid-session since the user was already operating the CLI, I thought it might be too aggressive to exit, just in case they are ok continuing at the same trust level.

Please lmk if you see better options.

packages/cli/src/ui/hooks/usePermissionsModifyTrust.ts
jacob314: we should avoid adding directly to history and use the emitter pattern for it.

shrutip90: Done

packages/cli/src/ui/components/PermissionsModifyTrustDialog.test.tsx
shrutip90: Done

packages/cli/src/ui/hooks/useFolderTrust.ts
abhipatel12: That's fine with me. Just curious

abhipatel12 (approved):
LGTM, thanks!

PR #11969, opened 2025-10-24
Refactor vim.test.ts: Use Parameterized Tests

IamRiddhi:
# Refactor vim.test.ts: Use Parameterized Tests

## Summary

Refactored the reducer-based integration tests in `vim.test.ts` to use parameterized tests (`it.each()`), reducing code duplication and improving maintainability.

## Tests Refactored

The following vim command tests were consolidated into a single parameterized test:

- `de` (delete word end) - 2 tests
- `db` (delete word backward) - 2 tests
- `cw` (change word forward) - 2 tests
- `ce` (change word end) - 2 tests
- `cb` (change word backward) - 1 test
- `cc` (change line) - 1 test
- `dd` (delete line) - 3 tests
- `D` (delete to end of line) - 2 tests
- `C` (change to end of line) - 2 tests

## Testing

All tests verified passing:
```
✓ src/ui/hooks/vim.test.ts (87 tests) 216ms
```

## Metrics

| Metric | Before | After | Change |
|--------|--------|-------|--------|
| Lines of code | 1732 | 1570 | -162 (-9.4%) |

### DIFF
--- a/packages/cli/src/ui/hooks/vim.test.tsx
+++ b/packages/cli/src/ui/hooks/vim.test.tsx
@@ -1355,12 +1355,227 @@ describe('useVim hook', () => {
   // Line operations (dd, cc) are tested in text-buffer.test.ts
 
   describe('Reducer-based integration tests', () => {
-    describe('de (delete word end)', () => {
-      it('should delete from cursor to end of current word', () => {
+    it.each([
+      {
+        command: 'de',
+        desc: 'delete from cursor to end of current word',
+        lines: ['hello world test'],
+        cursorRow: 0,
+        cursorCol: 1,
+        actionType: 'vim_delete_word_end' as const,
+        count: 1,
+        expectedLines: ['h world test'],
+        expectedCursorRow: 0,
+        expectedCursorCol: 1,
+      },
+      {
+        command: 'de',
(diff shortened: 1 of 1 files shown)


packages/cli/src/ui/hooks/vim.test.ts
```diff
@@ -1326,407 +1326,245 @@ describe('useVim hook', () => {
...
+        expectedCursorCol: 6,
+      },
+      {
+        command: 'C',
+        desc: 'handle C at beginning of line',
+        lines: ['hello world'],
+        cursorRow: 0,
+        cursorCol: 0,
+        actionType: 'vim_change_to_end_of_line' as const,
+        count: undefined,
```
jacob314: count=undefined is confusing. Can it just be removed from these test case definitions? Also give these definitions a static type signature that they conform to to avoid errors in the future for a case like this. Gemini CLI should be able to accomplish that if prompted and will avoid typeos in this case in the future.

jacob314 (commented):
approved after this one issue is addressed. Agree this case the tests are easier to follow with the new structure and significant duplicate code was removed.

jacob314 (approved):
lgtm

Issue #11845, opened 2025-10-23
Logs that are too large are dropped


bobcatfish:
### What happened?

When logs are too big to be consumed by the logging endpoint (GCP in my case), an error occurs and they are dropped. For example:

```
│ ✖ {"stack":"Error: 3 INVALID_ARGUMENT: Log entry with size 438.0K exceeds maximum size of 256.0K\n    at           │
│   callErrorFromStatus (file:///home/user/gemini-cli/bundle/gemini.js:96853:21)\n    at Object.onReceiveStatus       │
│   (file:///home/user/gemini-cli/bundle/gemini.js:97533:70)\n    at Object.onReceiveStatus                           │
│   (file:///home/user/gemini-cli/bundle/gemini.js:97336:140)\n    at Object.onReceiveStatus                          │
│   (file:///home/user/gemini-cli/bundle/gemini.js:97302:178)\n    at                                                 │
│   file:///home/user/gemini-cli/bundle/gemini.js:110487:77\n    at process.processTicksAndRejections                 │
│   (node:internal/process/task_queues:77:11)\nfor call at\n    at ServiceClientImpl.makeUnaryRequest                 │
│   (file:///home/user/gemini-cli/bundle/gemini.js:97503:32)\n    at ServiceClientImpl.<anonymous>                    │
│   (file:///home/user/gemini-cli/bundle/gemini.js:97811:19)\n    at                                                  │
│   file:///home/user/gemini-cli/bundle/gemini.js:185003:25\n    at                                                   │
│   file:///home/user/gemini-cli/bundle/gemini.js:174149:16\n    at repeat                                            │
│   (file:///home/user/gemini-cli/bundle/gemini.js:174198:23)\n    at Task._apiCall                                   │
│   (file:///home/user/gemini-cli/bundle/gemini.js:174234:11)\n    at Task.run                                        │
│   (file:///home/user/gemini-cli/bundle/gemini.js:172548:35)\n    at BundleExecutor._runNow                          │
│   (file:///home/user/gemini-cli/bundle/gemini.js:172788:14)\n    at Timeout._onTimeout                              │
│   (file:///home/user/gemini-cli/bundle/gemini.js:172734:18)\n    at listOnTimeout                                   │
│   (node:internal/timers:581:17)","message":"3 INVALID_ARGUMENT: Log entry with size 438.0K exceeds maximum size of  │
[...]

PR #11975, opened 2025-10-24
feat: Add Open Telemetric semantic standard compliant log

bobcatfish:
## TLDR

Updated logApiResponse and logApiError to emit a log that complies with the standard defined at https://github.com/open-telemetry/semantic-conventions/blob/main/docs/gen-ai/gen-ai-events.md.

## Dive Deeper

These are the fields added (or not!) defined in the spec: (I also added top_k in addition to top_p cuz it made sense?)

| Attribute | Added? | Notes |
| :---- | :---- | :---- |
| [gen\_ai.operation.name](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/registry/attributes/gen-ai.md) | Y |  |
| [error.type](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/registry/attributes/error.md) | Y |  |
| [gen\_ai.conversation.id](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/registry/attributes/gen-ai.md) | n/a | Couldn't find a value to use for this |
| [gen\_ai.output.type](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/registry/attributes/gen-ai.md) | Y |  |
| [gen\_ai.request.choice.count](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/registry/attributes/gen-ai.md) | Y |  |
| [gen\_ai.request.model](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/registry/attributes/gen-ai.md) | Y | This and gen\_ai.request.model are set to the same value, which could be coming from either depending on what is set |
| [gen\_ai.request.seed](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/registry/attributes/gen-ai.md) | Y |  |
| [server.address](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/registry/attributes/server.md) | Y | These are hardcoded, not reflective of the actual requests in all cases. Needs follow up |
| [server.port](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/registry/attributes/server.md) | Y |  |
| [gen\_ai.request.frequency\_penalty](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/registry/attributes/gen-ai.md) | Y |  |
| [gen\_ai.request.max\_tokens](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/registry/attributes/gen-ai.md) | Y |  |
| [gen\_ai.request.presence\_penalty](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/registry/attributes/gen-ai.md) | Y |  |
| [gen\_ai.request.stop\_sequences](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/registry/attributes/gen-ai.md) | Y |  |
| [gen\_ai.request.temperature](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/registry/attributes/gen-ai.md) | Y |  |
| [gen\_ai.request.top\_p](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/registry/attributes/gen-ai.md) | Y |  |
| [gen\_ai.response.finish\_reasons](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/registry/attributes/gen-ai.md) | Y |  |
| [gen\_ai.response.id](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/registry/attributes/gen-ai.md) | Y |  |
| [gen\_ai.response.model](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/registry/attributes/gen-ai.md) | Y | This and gen\_ai.request.model are set to the same value, which could be coming from either depending on what is set |
| [gen\_ai.usage.input\_tokens](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/registry/attributes/gen-ai.md) | Y |  |
| [gen\_ai.usage.output\_tokens](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/registry/attributes/gen-ai.md) | Y |  |
| [gen\_ai.input.messages](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/registry/attributes/gen-ai.md) | Y | These can be quite large and might need to be truncated |
| [gen\_ai.output.messages](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/registry/attributes/gen-ai.md) | Y | These can be quite large and might need to be truncated |
| [gen\_ai.system\_instructions](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/registry/attributes/gen-ai.md) | Y | These can be quite large and might need to be truncated |
| [gen\_ai.tool.definitions](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/registry/attributes/gen-ai.md) | N | This could also be quite large; the spec recommends not outputting by default so taking the opportunity to skip for now |

The initial issue described doing this in logUserPrompt and logApiResponse however after digging deeper, it seems like
`event.gen_ai.client.inference.operation.details` is meant to hold the details of a completed operation, so this change adds them when an API request completes, successfully or by erroring.

There are several things to follow up on afterwards:
* Handling logs that are too large #11845
* The format of the output messages, input messages, and system instructions is correct BUT JSON stringified instead of logging actual full JSON objects. I'm not sure if the latter is possible but I think it's what the spec wants. No matter what I tried (both fancy conversion functions and then eventually JSON.parse wrapping JSON.stringify), I got an error from the open telemetry library: https://github.com/open-telemetry/opentelemetry-js/blob/7e0d45f68271c88487354a4fe5eefe4c654efbfb/packages/opentelemetry-core/src/common/attributes.ts#L36 (or the same error from a different part of the codebase)
* The server address is more or less hardcoded in the non code-assist cases; though it's correct afaik right now, we can't 100% trust it over time. Would be better to be able to grab this dynamically
* Aligning the clear cut logs, i.e. adding any data that we added here which we might also want there (e.g. topP, topK etc)
* I skipped the tool definitions since it's recommended not to emit them by default but we may want to optionally log them

The standard has another log type, `event.gen_ai.evaluation.result` but afaik we're not doing any kind of evaluations (at runtime) that would make sense to log.

### Minor stuff

* I removed a mock from packages/core/src/telemetry/loggers.test.ts because it didn't seem like there was any reason to mock it when we could use the real thing and get the same results (imo the test was testing less with the mock).
* I needed to re-declare EVENT_CHAT_COMPRESSION. I can't figure out what did it but either a circular dependency or something about import order was introduced when i updated the api error logs which caused this import to suddenly be a problem. It's pretty weird that this metric has the same exact name as the corresponding log anyway, this seems like an accident - all the other metrics end with a descriptive word like "latency" or "count" so I'm very suspicous it was an error to use this string.

## Reviewer Test Plan

* Enable telemetry (https://github.com/google-gemini/gemini-cli/blob/main/docs/cli/telemetry.md#google-cloud-telemetry)
* Ask Gemini CLI something
* Go to GCP->Monitoring->Logs explorer and look for `jsonPayload."event.name" =~ "gen_ai.*"`

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ✅  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Fixes #9215

### DIFF
--- a/packages/core/src/core/loggingContentGenerator.ts
+++ b/packages/core/src/core/loggingContentGenerator.ts
@@ -5,15 +5,18 @@
  */
 
 import type {
+  Candidate,
   Content,
   CountTokensParameters,
   CountTokensResponse,
   EmbedContentParameters,
   EmbedContentResponse,
+  GenerateContentConfig,
   GenerateContentParameters,
   GenerateContentResponseUsageMetadata,
   GenerateContentResponse,
 } from '@google/genai';
+import type { ServerDetails } from '../telemetry/types.js';
 import {
   ApiRequestEvent,
   ApiResponseEvent,
@@ -26,6 +29,7 @@ import {
   logApiResponse,
 } from '../telemetry/loggers.js';
 import type { ContentGenerator } from './contentGenerator.js';
+import { CodeAssistServer } from '../code_assist/server.js';
 import { toContents } from '../code_assist/converter.js';
 import { isStructuredError } from '../utils/quotaErrorDetection.js';
 import { runInDevTraceSpan, type SpanMetadata } from '../telemetry/trace.js';
@@ -59,19 +63,66 @@ export class LoggingContentGenerator implements ContentGenerator {
     );
   }
 
+  private _getEndpointUrl(
+    req: GenerateContentParameters,
+    method: 'generateContent' | 'generateContentStream',
+  ): ServerDetails {
+    // Case 1: Authenticated with a Google account (`gcloud auth login`).
+    // Requests are routed through the internal CodeAssistServer.
+    if (this.wrapped instanceof CodeAssistServer) {
+      const url = new URL(this.wrapped.getMethodUrl(method));
+      const port = url.port
+        ? parseInt(url.port, 10)
+        : url.protocol === 'https:'
+          ? 443
+          : 80;
+      return { address: url.hostname, port };
+    }
+
+    const genConfig = this.config.getContentGeneratorConfig();
+
+    // Case 2: Using an API key for Vertex AI.
+    if (genConfig?.vertexai) {
+      const location = process.env['GOOGLE_CLOUD_LOCATION'];
+      if (location) {
+        return { address: `${location}-aiplatform.googleapis.com`, port: 443 };
+      } else {
+        return { address: 'unknown', port: 0 };
+      }
+    }
+
+    // Case 3: Default to the public Gemini API endpoint.
+    // This is used when an API key is provided but not for Vertex AI.
+    return { address: `generativelanguage.googleapis.com`, port: 443 };
+  }
+
   private _logApiResponse(
+    requestContents: Content[],
     durationMs: number,
     model: string,
     prompt_id: string,
+    responseId: string | undefined,
+    responseCandidates?: Candidate[],
     usageMetadata?: GenerateContentResponseUsageMetadata,
     responseText?: string,
+    generationConfig?: GenerateContentConfig,
+    serverDetails?: ServerDetails,
   ): void {
     logApiResponse(
       this.config,
       new ApiResponseEvent(
         model,
         durationMs,
-        prompt_id,
+        {
+          prompt_id,
+          contents: requestContents,
+          generate_content_config: generationConfig,
+          server: serverDetails,
+        },
+        {
+          candidates: responseCandidates,
+          response_id: responseId,
+        },
         this.config.getContentGeneratorConfig()?.authType,
         usageMetadata,
         responseText,
@@ -84,6 +135,9 @@ export class LoggingContentGenerator implements ContentGenerator {
     error: unknown,
     model: string,
     prompt_id: string,
+    requestContents: Content[],
+    generationConfig?: GenerateContentConfig,
+    serverDetails?: ServerDetails,
   ): void {
     const errorMessage = error instanceof Error ? error.message : String(error);
     const errorType = error instanceof Error ? error.name : 'unknown';
@@ -94,7 +148,12 @@ export class LoggingContentGenerator implements ContentGenerator {
         model,
         errorMessage,
         durationMs,
-        prompt_id,
+        {
+          prompt_id,
+          contents: requestContents,
+          generate_content_config: generationConfig,
+          server: serverDetails,
+        },
         this.config.getContentGeneratorConfig()?.authType,
         errorType,
         isStructuredError(error)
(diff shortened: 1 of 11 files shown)


packages/core/src/core/loggingContentGenerator.ts
```diff
@@ -58,19 +62,62 @@ export class LoggingContentGenerator implements ContentGenerator {
...
+          : 80;
+      return { address: url.hostname, port };
+    }
+
+    const genConfig = this.config.getContentGeneratorConfig();
+
+    // Case 2: Using an API key for Vertex AI.
+    if (genConfig?.vertexai) {
+      const location = process.env['GOOGLE_CLOUD_LOCATION'];
+      return { address: `${location}-aiplatform.googleapis.com`, port: 443 };
```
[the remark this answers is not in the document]
bobcatfish: done done de done

packages/core/src/telemetry/uiTelemetry.ts
```diff
@@ -165,12 +165,12 @@ export class UiTelemetryService extends EventEmitter {
...
-    modelMetrics.tokens.total += event.total_token_count;
-    modelMetrics.tokens.cached += event.cached_content_token_count;
-    modelMetrics.tokens.thoughts += event.thoughts_token_count;
-    modelMetrics.tokens.tool += event.tool_token_count;
+    modelMetrics.tokens.prompt += event.usage.input_token_count;
+    modelMetrics.tokens.candidates += event.usage.output_token_count;
+    modelMetrics.tokens.total += event.usage.total_token_count;
+    modelMetrics.tokens.cached += event.usage.cached_content_token_count;
+    modelMetrics.tokens.thoughts += event.usage.thoughts_token_count;
+    modelMetrics.tokens.tool += event.usage.tool_token_count;
```
jerop, line 173: please update the docs to nest these under usage

bobcatfish: this shouldnt (hopefully XD) be actually changing the resulting metric structure at all, it's just updating the event object to have a 'usage' attribute that holds these values 🙏 🙏 🙏 

(let me know if that's not the case tho, i didnt want to impact any of the existing metrics / logs with this change)

jerop: metrics don't change but event/log structure changes, so may need to update the attributes: https://github.com/google-gemini/gemini-cli/blob/ac285148e2afc4cbc549380f745bc1af68e8cc5e/docs/cli/telemetry.md#api

jerop: resolved offline - flattened to logs in https://github.com/google-gemini/gemini-cli/blob/ac285148e2afc4cbc549380f745bc1af68e8cc5e/packages/core/src/telemetry/types.ts#L553

jerop (approved):

PR #11977, opened 2025-10-24
Support paste markers split across writes.

scidomino:
## TLDR

Refactor the paste code to support paste markers split across writes.

## Dive Deeper

The generator pattern used here is based on the one node uses inside `readline.emitKeypressEvents()`.

This manually sets the encoding for stdin to force utf8. This is probably something we should have done anyways but it is a bit risky so I'm pointing that out.

## Reviewer Test Plan

Try pasting text.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | x  | x  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

For #7773

### DIFF
--- a/packages/cli/src/ui/components/SettingsDialog.test.tsx
+++ b/packages/cli/src/ui/components/SettingsDialog.test.tsx
@@ -1348,7 +1348,7 @@ describe('SettingsDialog', () => {
 
       // Press Escape to exit
       stdin.write('\u001B');
-      await wait();
+      await wait(100);
 
       expect(onSelect).toHaveBeenCalledWith(undefined, 'User');
 
--- a/packages/cli/src/ui/contexts/KeypressContext.tsx
+++ b/packages/cli/src/ui/contexts/KeypressContext.tsx
@@ -40,8 +40,8 @@ import {
 import { FOCUS_IN, FOCUS_OUT } from '../hooks/useFocus.js';
 
 const ESC = '\u001B';
-export const PASTE_MODE_PREFIX = `${ESC}[200~`;
-export const PASTE_MODE_SUFFIX = `${ESC}[201~`;
+export const PASTE_MODE_START = `${ESC}[200~`;
+export const PASTE_MODE_END = `${ESC}[201~`;
 export const DRAG_COMPLETION_TIMEOUT_MS = 100; // Broadcast full path after 100ms if no more input
 export const KITTY_SEQUENCE_TIMEOUT_MS = 50; // Flush incomplete kitty sequences after 50ms
 export const SINGLE_QUOTE = "'";
@@ -353,6 +353,99 @@ function parseKittyPrefix(buffer: string): { key: Key; length: number } | null {
   return null;
 }
 
+const PASTE_MODE_START_BUFER = Buffer.from(PASTE_MODE_START);
+const PASTE_MODE_END_BUFER = Buffer.from(PASTE_MODE_END);
+const PASTE_MODE_COMMON_BUFFER = Buffer.from(`${ESC}[20`);
+
+function matchesAt(a: Buffer, b: Buffer, pos: number): boolean {
+  return a.compare(b, pos, pos + a.length) === 0;
+}
+
+/**
+ * Returns the first index before which we are certain there is no paste marker.
+ */
+function earliestPossiblePasteMarker(data: Buffer): number {
+  // Check data for full start-paste or end-paste markers.
+  let pos = 0;
+  const codeLength = PASTE_MODE_START_BUFER.length;
+  while (pos < data.length - codeLength) {
+    const index = data.indexOf(PASTE_MODE_COMMON_BUFFER, pos);
+    if (index === -1) {
+      break;
+    }
+    const isStart = matchesAt(PASTE_MODE_START_BUFER, data, index);
+    const isEnd = matchesAt(PASTE_MODE_END_BUFER, data, index);
+    if (isStart || isEnd) {
+      return index;
+    }
+    pos = index + PASTE_MODE_COMMON_BUFFER.length;
+  }
+  // data contains no full start-paste or end-paste.
+  // Check if data ends with a prefix of start-paste or end-paste.
+  for (let i = 1; i < codeLength; i++) {
+    const candidate = data.subarray(data.length - i);
+    const isStartPrefix = candidate.compare(PASTE_MODE_START_BUFER, 0, i) === 0;
+    const isEndPrefix = candidate.compare(PASTE_MODE_END_BUFER, 0, i) === 0;
+    if (isStartPrefix || isEndPrefix) {
+      return data.length - i;
+    }
+  }
+  return data.length;
+}
+
+/**
+ * A generator that takes in data chunks and spits out paste-start and
+ * paste-end keypresses. All non-paste marker data is passed to passthrough.
+ */
+function* pasteMarkerParser(
+  passthrough: PassThrough,
+  keypressHandler: (_: unknown, key: Key) => void,
+): Generator<void, void, Buffer> {
+  while (true) {
+    let data = yield;
+
+    while (true) {
+      const index = earliestPossiblePasteMarker(data);
+      if (index === data.length) {
+        // no possible paste markers were found
+        passthrough.write(data);
+        break;
+      }
+      if (index > 0) {
+        // snip off and send the part that doesn't have a paste marker
+        passthrough.write(data.subarray(0, index));
+        data = data.subarray(index);
+      }
+      // data starts with a possible paste marker
+      const possibleMarker = data.subarray(0, 6);
+      if (possibleMarker.compare(PASTE_MODE_START_BUFER) === 0) {
+        keypressHandler(undefined, {
+          name: 'paste-start',
+          ctrl: false,
+          meta: false,
+          shift: false,
+          paste: false,
+          sequence: '',
+        });
+        data = data.subarray(PASTE_MODE_START_BUFER.length);
+      } else if (possibleMarker.compare(PASTE_MODE_END_BUFER) === 0) {
+        keypressHandler(undefined, {
+          name: 'paste-end',
+          ctrl: false,
+          meta: false,
+          shift: false,
+          paste: false,
+          sequence: '',
+        });
+        data = data.subarray(PASTE_MODE_END_BUFER.length);
+      } else {
+        // we must have a prefix. Concat the next data and try again.
+        data = Buffer.concat([data, yield]);
+      }
+    }
+  }
+}
+
 export interface Key {
   name: string;
   ctrl: boolean;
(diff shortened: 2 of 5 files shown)


packages/cli/src/ui/hooks/useKeypress.test.ts
```diff
@@ -187,6 +187,37 @@ describe('useKeypress', () => {
...
+
+      const keyA = { name: 'a', sequence: 'a' };
+      act(() => stdin.write('a'));
+      expect(onKeypress).toHaveBeenCalledWith(
+        expect.objectContaining({ ...keyA, paste: false }),
+      );
+
+      const pasteText = 'pasted';
+      act(() => {
+        stdin.write(PASTE_START.slice(0, 3));
```
jacob314, line 267: bonus points... add a couple ms delay between each of these to really ensure that these don't get buffered by some intermediate step and your code really has to do the handling. also probably worth it to test with a single char at a time rather than giving slice(3) chars in the first chunk.

packages/cli/src/ui/contexts/KeypressContext.tsx
```diff
@@ -40,8 +40,8 @@ import {
 import { FOCUS_IN, FOCUS_OUT } from '../hooks/useFocus.js';
 
 const ESC = '\u001B';
-export const PASTE_MODE_PREFIX = `${ESC}[200~`;
-export const PASTE_MODE_SUFFIX = `${ESC}[201~`;
+export const PASTE_MODE_START = `${ESC}[200~`;
```
jacob314, line 43: nice. that is a clearer name.

packages/cli/src/ui/contexts/KeypressContext.tsx
```diff
@@ -353,6 +353,99 @@ function parseKittyPrefix(buffer: string): { key: Key; length: number } | null {
   return null;
 }
 
+const PASTE_MODE_START_BUFER = Buffer.from(PASTE_MODE_START);
+const PASTE_MODE_END_BUFER = Buffer.from(PASTE_MODE_END);
+const PASTE_MODE_COMMON_BUFFER = Buffer.from(`${ESC}[20`);
```
jacob314: nit: add a common. not immediately obvious what this is.

packages/cli/src/ui/contexts/KeypressContext.tsx
```diff
@@ -353,6 +353,99 @@ function parseKittyPrefix(buffer: string): { key: Key; length: number } | null {
   return null;
 }
 
+const PASTE_MODE_START_BUFER = Buffer.from(PASTE_MODE_START);
```
jacob314: nit BUFER->BUFFER on the next two lines

packages/cli/src/ui/contexts/KeypressContext.tsx
```diff
@@ -353,6 +353,99 @@ function parseKittyPrefix(buffer: string): { key: Key; length: number } | null {
...
+const PASTE_MODE_COMMON_BUFFER = Buffer.from(`${ESC}[20`);
+
+function matchesAt(a: Buffer, b: Buffer, pos: number): boolean {
+  return a.compare(b, pos, pos + a.length) === 0;
+}
+
+/**
+ * Returns the first index before which we are certain there is no paste marker.
+ */
+function earliestPossiblePasteMarker(data: Buffer): number {
```
jacob314: AI powered code review nits.
**Potential Bug:**
In `earliestPossiblePasteMarker`, the loop condition seems to miss exact matches if the buffer is exactly the length of the marker:
  while (pos < data.length - codeLength) {

   1
   2 If `data.length` is 6 and `codeLength` is 6, `pos < 0` is false, so it skips the loop and might miss a marker at index 0. I believe this should be `<=`.
   3
   4 **Simplification Idea:**
   5 The tail check loop can be simplified. Since any partial marker must start with `ESC`, we can just look for the last `ESC` and check if it forms a valid prefix.
    // ... after while loop ...
    // Check if data ends with a prefix of start-paste or end-paste.
    const lastEsc = data.lastIndexOf(ESC);
    if (lastEsc !== -1 && data.length - lastEsc < codeLength) {
      const candidate = data.subarray(lastEsc);
      const i = candidate.length;
      const isStartPrefix = candidate.compare(PASTE_MODE_START_BUFFER, 0, i) === 0;
      const isEndPrefix = candidate.compare(PASTE_MODE_END_BUFFER, 0, i) === 0;
      if (isStartPrefix || isEndPrefix) {
        return lastEsc;
      }
    }
    return data.length;

packages/cli/src/ui/contexts/KeypressContext.tsx
```diff
@@ -353,6 +353,99 @@ function parseKittyPrefix(buffer: string): { key: Key; length: number } | null {
...
+    const index = data.indexOf(PASTE_MODE_COMMON_BUFFER, pos);
+    if (index === -1) {
+      break;
+    }
+    const isStart = matchesAt(PASTE_MODE_START_BUFER, data, index);
+    const isEnd = matchesAt(PASTE_MODE_END_BUFER, data, index);
+    if (isStart || isEnd) {
+      return index;
+    }
+    pos = index + PASTE_MODE_COMMON_BUFFER.length;
```
jacob314: going by PASTE_MODE_COMMON_BUFFER.length at a time doesn't seem right. think that will fail at various edge cases.  Would suggest asking AI to massage this method to be a more robust and more efficient.  there is probably an ideal solution using a FSM for this that might be worth using if Gemini can write a competent version of it.

packages/cli/src/ui/contexts/KeypressContext.tsx
```diff
@@ -353,6 +353,99 @@ function parseKittyPrefix(buffer: string): { key: Key; length: number } | null {
...
+        passthrough.write(data);
+        break;
+      }
+      if (index > 0) {
+        // snip off and send the part that doesn't have a paste marker
+        passthrough.write(data.subarray(0, index));
+        data = data.subarray(index);
+      }
+      // data starts with a possible paste marker
+      const possibleMarker = data.subarray(0, 6);
```
jacob314: rather than 6,  use the length of the PASTE_MODE_START_BUFFER or add a separate const.

packages/cli/src/ui/contexts/KeypressContext.tsx
```diff
@@ -766,80 +859,37 @@ export function KeypressProvider({
...
+    let cleanup = () => {};
     let rl: readline.Interface;
     if (keypressStream !== null) {
       rl = readline.createInterface({
         input: keypressStream,
         escapeCodeTimeout: 0,
       });
       readline.emitKeypressEvents(keypressStream, rl);
+
+      const pmp = pasteMarkerParser(keypressStream, handleKeypress);
```
jacob314: nit: pmp doesn't mean anything. use p or parser if you just want a shorter name.

packages/cli/src/ui/hooks/useKeypress.test.ts
```diff
@@ -187,6 +187,37 @@ describe('useKeypress', () => {
       expect(onKeypress).toHaveBeenCalledTimes(3);
     });
 
+    it('should handle pastes split across writes', () => {
```
jacob314: add some tests for sequences that start with chars that look like a paste marker but aren't. Verify we get a response after the first char that doesn't match a paste marker is return.

jacob314 (commented):
overall looks pretty good.  a couple of comments. applying a very strict code review standard for this due to the importance of getting this exactly right and how hard it will be to track down bugs. the more robustly we can test this the better.

packages/cli/src/ui/contexts/KeypressContext.tsx
scidomino: Going by PASTE_MODE_COMMON_BUFFER.length is ok because PASTE_MODE_COMMON_BUFFER is prefix free. Will add a comment,

packages/cli/src/ui/contexts/KeypressContext.tsx
scidomino: Fixed and added a test to catch it. (Note: the test only catches it in Node 20 since Node 22 knows to look for paste markers.)

packages/cli/src/ui/contexts/KeypressContext.tsx
scidomino: done.

packages/cli/src/ui/contexts/KeypressContext.tsx
scidomino: Good advice. Gemini had some good pointers for simplifying things.

jacob314 (approved):
lgtm

packages/cli/src/ui/contexts/KeypressContext.tsx
```diff
@@ -867,9 +874,15 @@ export function KeypressProvider({
...
-      const handleRawKeypress = (data: Buffer) => parser.next(data);
+      let timeoutId: NodeJS.Timeout;
+      const handleRawKeypress = (data: string) => {
+        clearTimeout(timeoutId);
+        parser.next(data);
+        timeoutId = setTimeout(() => parser.next(''), PASTE_CODE_TIMEOUT_MS);
+      };
 
       keypressStream.on('keypress', handleKeypress);
+      process.stdin.setEncoding('utf8'); // so handleRawKeypress gets strings
```
jacob314, line 885: a little mixed about setting this here. should we set right when we initialize the app in gemini.tsx. Reason is if we have a separate MouseContext it should also depend on stdout and use utf8 mode and would be silly to be setting this everywhere.

packages/cli/src/ui/components/SettingsDialog.test.tsx
```diff
@@ -1348,7 +1348,7 @@ describe('SettingsDialog', () => {
 
       // Press Escape to exit
       stdin.write('\u001B');
-      await wait();
+      await wait(100);
```
jacob314, line 1351: why does this have to be 100? is this 50ms for paste + 50 ms for kitty? To avoid making these tests slow we should start mocking the time. this is a decent amount of extra time per test run just because of our keyboard handling.

jacob314 (approved):
lgtm

packages/cli/src/ui/components/SettingsDialog.test.tsx
scidomino: It only waits 50ms and it doesn't stack. I just waited twice as long to be sure since timeouts can theoretically fire after the timeout.

packages/cli/src/ui/contexts/KeypressContext.tsx
scidomino: We should set it in gemini.tsx as well but we have to set it here to, otherwise the tests will fail.

PR #11999, opened 2025-10-24
fix(cli): add Windows clipboard fallback for missing Unicode chars like 'ç'

Nimraakram12:
# Fix for Missing Turkish “ç” in Gemini CLI on Windows

## TLDR
This PR fixes a **Windows-specific bug** where the lowercase Turkish letter **ç (U+00E7)** is silently dropped when pasting or typing into the **Gemini CLI** interactive prompt (e.g., `merhaba çiçek` becomes `merhaba iek`).  

The issue occurs because certain Windows terminals drop **ç** before it reaches the CLI.  
This change adds a **clipboard fallback** that detects corrupted pastes and reads directly from the Windows clipboard using PowerShell’s `Get-Clipboard`, preserving **Unicode integrity**.

---

## Dive Deeper

On Windows (PowerShell, VS Code terminal), the character **ç** is often lost during paste due to terminal/Node.js input handling limitations — even though other Turkish letters (**ğ, ş, ü, Ç**) work fine.  
Debug logs confirm **ç** is already missing when the CLI receives the input (`key.sequence` contains `"merhaba iek"`).

Rather than trying to fix the upstream terminal issue (which is out of our control), this PR implements a **safe, targeted workaround**:

### Behavior
- **Only on Windows** (`process.platform === 'win32'`)
- **Only when pasting text that appears corrupted**
  - Heuristic: input contains `"iek"` but no `"ç"`, as in `"çiçek" → "iek"`
- **Fallback:** reads directly from system clipboard via PowerShell

### Implementation
- Uses `[System.Windows.Forms.Clipboard]::GetText()` for reliable UTF-8 text retrieval  
- Fails gracefully if clipboard access fails (falls back to original input)  
- No impact on macOS/Linux or non-broken input  
- Adds **no new dependencies**  

This ensures Turkish (and other Unicode-heavy) users can **reliably use Gemini CLI on Windows**.

---

### On Windows (PowerShell or VS Code terminal)
1. Copy `merhaba çiçek` to clipboard  
2. Run:
   ```bash
   npm run build && npm start
3. Paste into CLI
✅ Should show: merhaba çiçek (not merhaba iek)

4. Test normal paste (non-Turkish):
Paste hello world → should work unchanged

5. Test typing (not paste):
Type merhaba çiçek → should work if terminal supports it

6. Verify fallback safety:
Temporarily break clipboard access (e.g., run without .NET)
Paste merhaba çiçek → should fall back to original (broken) input without crashing

On macOS/Linux
Confirm no behavioral changes (fallback is Windows-only)

## Linked Issues / Bugs

Fixes ##11991

### DIFF
--- a/packages/cli/src/ui/components/shared/text-buffer.ts
+++ b/packages/cli/src/ui/components/shared/text-buffer.ts
@@ -30,7 +30,29 @@ export type Direction =
   | 'home'
   | 'end';
 
+// --- Windows Clipboard Fallback ---
+function getClipboardTextWindows(): string | null {
+  try {
+    const result = spawnSync(
+      'powershell.exe',
+      ['-Command', '[System.Windows.Forms.Clipboard]::GetText()'],
+      {
+        encoding: 'utf8',
+        timeout: 1000,
+        windowsHide: true,
+      },
+    );
+    if (result.status === 0 && result.stdout) {
+      return result.stdout.trim();
+    }
+  } catch (_e) {
+    // Silent fallback — don't break the app
+  }
+  return null;
+}
+
 // Simple helper for word‑wise ops.
+
 function isWordChar(ch: string | undefined): boolean {
   if (ch === undefined) {
     return false;
@@ -1609,23 +1631,32 @@ export function useTextBuffer({
 
   const insert = useCallback(
     (ch: string, { paste = false }: { paste?: boolean } = {}): void => {
+      // For pasted text, insert as-is without splitting into code points
+      if (paste) {
+        console.log('PASTE DEBUG: raw =', JSON.stringify(ch));
+        console.log(
+          'CHAR CODES:',
+          Array.from(ch).map((c) => c.charCodeAt(0)),
+        );
+        if (ch.length > 0) {
+          dispatch({ type: 'insert', payload: ch });
+        }
+        return;
+      }
+
+      // For typed input, keep the existing logic (for DEL/backspace handling)
       if (/[\n\r]/.test(ch)) {
         dispatch({ type: 'insert', payload: ch });
         return;
       }
 
       const minLengthToInferAsDragDrop = 3;
-      if (
-        ch.length >= minLengthToInferAsDragDrop &&
-        !shellModeActive &&
-        paste
-      ) {
+      if (ch.length >= minLengthToInferAsDragDrop && !shellModeActive) {
         let potentialPath = ch.trim();
         const quoteMatch = potentialPath.match(/^'(.*)'$/);
         if (quoteMatch) {
           potentialPath = quoteMatch[1];
         }
-
         potentialPath = potentialPath.trim();
         if (isValidPath(unescapePath(potentialPath))) {
           ch = `@${potentialPath} `;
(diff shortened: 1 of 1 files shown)


cornmander:
Thank you for your contribution. I will have a look today and leave feedback.

packages/cli/src/ui/components/shared/text-buffer.ts
```diff
@@ -1878,63 +1909,93 @@ export function useTextBuffer({
...
       sequence: string;
     }): void => {
       const { sequence: input } = key;
 
+      // Handle paste FIRST
       if (key.paste) {
-        // Do not do any other processing on pastes so ensure we handle them
-        // before all other cases.
-        insert(input, { paste: key.paste });
+        // On Windows, if pasted text appears to be missing 'ç' (e.g., "çiçek" → "iek"),
```
scidomino, line 1923: This code looking for pasted text with "iek" but not "ç"? So won't it trigger if the pasted text includes the English word "shriek"?

packages/cli/src/ui/components/shared/text-buffer.ts
```diff
@@ -30,7 +30,29 @@ export type Direction =
...
+function getClipboardTextWindows(): string | null {
+  try {
+    const result = spawnSync(
+      'powershell.exe',
+      ['-Command', '[System.Windows.Forms.Clipboard]::GetText()'],
+      {
+        encoding: 'utf8',
+        timeout: 1000,
+        windowsHide: true,
+      },
```
AndrePuls: This only works if PowerShell<=5.1 (early 2014) is installed. Why not also cover Powershell 7+ or the (generally more preferred) Get-Clipboard CMDLet? [learn.microsoft.com|Renamed powershell.exe to pwsh.exe](https://learn.microsoft.com/en-us/powershell/scripting/whats-new/differences-from-windows-powershell?view=powershell-7.5#powershell-executable-changes)

packages/cli/src/ui/components/shared/text-buffer.ts
```diff
@@ -30,7 +30,29 @@ export type Direction =
...
+    );
```
Throwaway-nVN4zWX, line 44: spawnSync blocks the Node event loop, is expensive, and this approach depends on a specific binary (<=5.1 early 2014 [learn.microsoft.com|Renamed powershell.exe to pwsh.exe](https://learn.microsoft.com/en-us/powershell/scripting/whats-new/differences-from-windows-powershell?view=powershell-7.5#powershell-executable-changes) )

It may fail if only Powershell 7+ (pwsh.exe) is installed
It may fail if the PATH env variable is missing the path to the binary.
It does not try Aliases like "pwsh"
It does not try the (seemingly generally more preferred and reliably) "Get-Clipboard" CMDLet
How long does it take to process a one million token context window? May this timeout?
System.Windows.Forms is part of the Windows GUI - does this still work on headless machines?
Also, add the -NoProfile flag. It can improve performance significantly.

packages/cli/src/ui/components/shared/text-buffer.ts
```diff
@@ -1878,63 +1909,93 @@ export function useTextBuffer({
...
+          input.includes('iek') && // heuristic: part of "çiçek"
+          !input.includes('ç') // and 'ç' is missing
+        ) {
+          const clipboardText = getClipboardTextWindows();
+          if (clipboardText) {
+            insert(clipboardText, { paste: true });
+            return;
+          }
+        }
+        insert(input, { paste: true });
```
Throwaway-nVN4zWX, line 1936: I'm not happy with this, at all.

This is not a "Fix" - it's, at best, a unnecessary, brittle and incomplete "Work around" for something, that **may** have occured **somewhere** in the pipeline well **before** this point.

- Why is only one Unicode character of 300k affected and "fixed"?  What about Emojis? Symbols? Wide and Super-Wide Characters? Cyrillic?

- Is "çiçek" the only Turkish word, that uses this character? If not, does this check work with ALL words in any language?

- In the event of pasted text, you receive the text and perform a search for 'iek'. Then another search for "ç". Then you disregard the text you have (even if nothing has been found wrong here), to re-receive the clipboard again. Even **IF** the current method **is** faulty (again: unconfirmed, uninvestigated and unlikely), I would rather see **that** be addressed and fixed. Adding another method as a fallback is doing twice the work.

-  for spawnSync('powershell.exe') see my other comment.

Even if accepted, this should (in my opinion) NOT be applied for everyone by default, regardless of the actual pipeline, config, involved NPM packages, etc. If, at all, a flag or setting, that needs to be activated. And it should check, whether Powershell is actually installed, what version and path.

At best, a very, very small number of actual, unintentional dropped cases of a single Character of a single word is worked around. At a significant performance penalty for every Win User. At worst, you **skip** important, battle-tested **sanitation**/codepage conversions/filtering.

Issue #11984, opened 2025-10-24
[Refactor] Extract Chat Compression Service


abhipatel12:
### Description
Refactor the chat compression logic out of `GeminiClient` into a standalone, reusable service. This is a prerequisite for ensuring both the main CLI agent and subagents (via `AgentExecutor`) can use compression and, by extension, safely use JIT context without blowing up their context windows.

### Acceptance Criteria
- [x] A new `ChatCompressionService` class is created in `packages/core/src/services/chatCompressionService.ts`.
- [x] The `tryCompressChat` logic (finding split point, generating summary, creating new chat with summary) is moved from `GeminiClient` to this new service.
- [x] `GeminiClient` is updated to use `ChatCompressionService`.
- [ ] `AgentExecutor` is updated to use `ChatCompressionService` before its main loop turns.
- [ ] Unit tests for compression are moved/updated to test the new service in isolation.

### Technical Details
- **Create `packages/core/src/services/chatCompressionService.ts`**:
  - `compress(chat: GeminiChat, promptId: string, force: boolean): Promise<{ newChat: GeminiChat | null, info: ChatCompressionInfo }> `
- **Modify `packages/core/src/core/client.ts`**:
  - Remove `tryCompressChat` and `findCompressSplitPoint`.
  - Inject/instantiate `ChatCompressionService`.
  - Update `sendMessageStream` to call `compressionService.compress()`.
- **Modify `packages/core/src/agents/executor.ts`**:
  - Integrate `ChatCompressionService` into the `run` loop to check for compression before calling the model.

PR #12001, opened 2025-10-24
refactor(core): extract ChatCompressionService from GeminiClient

SandyTao520:
## TLDR

Extracted chat compression logic from `GeminiClient` into a new reusable `ChatCompressionService`. This is a preparatory step for enabling compression in subagents.

## Dive Deeper

- Created `packages/core/src/services/chatCompressionService.ts` containing `ChatCompressionService` and `findCompressSplitPoint`.
- Updated `GeminiClient` to use `ChatCompressionService` for `tryCompressChat`.
- Migrated relevant unit tests from `client.test.ts` to `chatCompressionService.test.ts`.
- Updated `client.test.ts` to mock `ChatCompressionService` and verify it's called correctly.
- Extracted `getInitialChatHistory` to `packages/core/src/utils/environmentContext.ts` to share logic between `GeminiClient` and `ChatCompressionService`.

## Reviewer Test Plan

- Verify that chat compression still works as expected in the main agent (e.g., by having a very long conversation and checking telemetry or logs if enabled).
- Review the new `ChatCompressionService` and its tests.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓ |
| npx      | ✅  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

Related to #11984

### DIFF
--- a/packages/core/src/services/chatCompressionService.ts
+++ b/packages/core/src/services/chatCompressionService.ts
@@ -14,7 +14,7 @@ import { getCompressionPrompt } from '../core/prompts.js';
 import { getResponseText } from '../utils/partUtils.js';
 import { logChatCompression } from '../telemetry/loggers.js';
 import { makeChatCompressionEvent } from '../telemetry/types.js';
-import { getEnvironmentContext } from '../utils/environmentContext.js';
+import { getInitialChatHistory } from '../utils/environmentContext.js';
 
 /**
  * Threshold for compression token count as a fraction of the model's token limit.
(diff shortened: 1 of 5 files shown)


packages/core/src/services/chatCompressionService.ts
```diff
@@ -0,0 +1,220 @@
...
+    !lastContent?.parts?.some((part) => part.functionCall)
+  ) {
+    return contents.length;
+  }
+
+  // Can't compress everything so just compress at last splitpoint.
+  return lastSplitPoint;
+}
+
+export class ChatCompressionService {
```
abhipatel12, line 79: WDYT about having this as a class vs a function? Do you think there's a benefit to keeping it as a class?

abhipatel12 (approved):

packages/core/src/services/chatCompressionService.ts
SandyTao520: While this is currently stateless and could be a standalone function, I opted for a class to maintain consistency with other services, facilitate easier mocking in unit tests, and allow for future state (like caching) without breaking the API for consumers.

PR #12004, opened 2025-10-24
Fix the shortenPath function to correctly insert ellipsis.

ph-sp:
## TLDR

This fixes the shortenPath utility function to correctly place an ellipsis and adds unit tests for posix and windows paths.

## Dive Deeper

The shortenPath function was not capable of producing a path in the form `/here/.../is/a/really/long/path`. The only cases where a `...` is added are in a long single segment path (like `/some-extremely-file-path-in-ones-root-directory-that-surely-rarely-happens`) OR if a file path has a set of segments that end up <35 characters. 

Added 10 tests for windows, 10 tests for posix. I only have access to a linux machine, but I ran the windows tests by forcing `path.win32` module locally before submission.

## Reviewer Test Plan

When the Gemini CLI edits a file in a deeply nested directory (>35 characters deep) it previously would emit a directory like:

`...some/really/long/path/filename.txt`

Now it should emit:

`/some/.../really/long/path/filename.txt`

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | N | N  |  Y  |
| npx      | N | N  | Y  |
| Docker   | N  | N  | N  |
| Podman   | N  | -   | -   |
| Seatbelt | N  | -   | -   |

## Linked issues / bugs
This PR Fixes #8447


ph-sp:
> I've identified a critical edge case where the function can still return a string longer than the specified maximum length. I've provided a suggestion to fix this. Additionally, I've suggested adding new unit tests to cover this edge case for both POSIX and Windows paths to prevent future regressions.

I don't see a suggestion to address your edge case issue, nor do I see a description of what that edge case is so I could potentially address it myself.

cornmander (approved):

cornmander:
Tested locally on linux.

Issue #3618, opened 2025-07-09
read file and edit does not work on windows


laughxing:
### What happened?

when i run gemini cli on windows, it is support to read and edit files. but i got this

> First, I need to locate the Gherkin files (`bdd-*.md`) in the current directory.I have found 7 Gherkin files. I will begin the optimization process with `G:\gemini-cli-laugh\bdd-GameFacade.md`.
> 
> First, I'll read the file's content.I apologize, it seems I misused the `read_file` tool due to the Windows path format. I will correct the path and try again.My apologies again. The tool is stricter about paths than I anticipated. It appears I need to provide the path relative to the project root, `G:\gemini-cli-laugh`. I will now try reading the file with the correct path format.It seems I am having persistent trouble with the file path interpretation on your system. I will try to read a different file, `/package.json`, to diagnose the issue. If this fails, I may need your guidance on the correct path format.I am unable to proceed. I've tried multiple path formats to read the Gherkin files, but I am consistently failing to access them. The tool is rejecting all my attempts to specify an absolute path on your Windows system.
> 
> Could you please provide me with the correct, full absolute path for the file `G:\gemini-cli-laugh\bdd-GameFacade.md` that the `read_file` tool will accept? This will help me understand the expected format and allow me to continue with the optimization task.

### What did you expect to happen?

the read file and edit file should work well on windows.

### Client information

<details>

│ CLI Version            0.1.9                                      │
│ Git Commit             e55197c (local modifications)              │
│ Model                  gemini-2.5-pro                             │
│ Sandbox                no sandbox                                 │
│ OS                     win32                                      │
│ Auth Method            gemini-api-key 

</details>

bramburn:
The read file doesn't work all the time, I ask it to use `type` or create a python script to do the work.

gsquared94:
Found possible duplicate issues:

- #2539: (0.910339144180392)

galdawave:
Thanks for taking the time to create this issue!
It looks like this is a duplicate of issue #2539. To keep all related conversations in one place, I'm going to close this one. Please feel free to subscribe to and continue the discussion on the original issue.

PR #12007, opened 2025-10-25
add absolute file path description for windows

gsehgal:
## TLDR

Add the absolute path for Windows.

## Dive Deeper

The absolute path for windows was missing. Add that to the tool descrition

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

https://github.com/google-gemini/gemini-cli/issues/3618

### DIFF
--- a/packages/core/src/tools/edit.test.ts
+++ b/packages/core/src/tools/edit.test.ts
@@ -36,6 +36,14 @@ vi.mock('../telemetry/loggers.js', () => ({
   logFileOperation: vi.fn(),
 }));
 
+interface EditFileParameterSchema {
+  properties: {
+    file_path: {
+      description: string;
+    };
+  };
+}
+
 import type { Mock } from 'vitest';
 import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
 import type { EditToolParams } from './edit.js';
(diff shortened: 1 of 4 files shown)


scidomino (approved):

packages/core/src/tools/edit.test.ts
```diff
@@ -36,6 +36,29 @@ vi.mock('../telemetry/loggers.js', () => ({
   logFileOperation: vi.fn(),
 }));
 
+interface EditFileParameterSchema {
```
scidomino, line 39: Is all this necessary? Can't you just define the ones you use?

packages/core/src/tools/edit.test.ts
```diff
@@ -1025,6 +1049,52 @@ describe('EditTool', () => {
     });
   });
 
+  describe('constructor', () => {
+    const mockConfigInstance = {
```
scidomino: Do we actually have to define all of these methods? Are there any we can leave out?

packages/core/src/tools/edit.test.ts
gsehgal: Done.

scidomino (approved):

PR #12022, opened 2025-10-25
fix: only apply Alt key mapping on macOS to fix 'ç' on Linux

### LINKED ISSUE #12021: Bug: Character 'ç' not working on Linux (mapped to Alt+C)
## Description
When typing the lowercase 'ç' character on Linux, it doesn't appear in the input. However, uppercase 'Ç' (with Shift) works correctly.

## Steps to Reproduce
1. Run gemini-cli on Linux with Portuguese (pt-PT) keyboard layout
2. Try to type the character 'ç' (lowercase)
3. Character doesn't appear in input
4. Type 'Ç' (uppercase with Shift) - works correctly

## Root Cause
The `ALT_KEY_CHARACTER_MAP` in `packages/cli/src/ui/contexts/KeypressContext.tsx` (line 54) maps 'ç' (U+00E7) to Alt+C. This mapping is for macOS where Alt+C produces 'ç', but on Linux, 'ç' is typed directly.

## Proposed Solution
Add platform detection to only apply the Alt key mapping on macOS (`process.platform === 'darwin'`).

## System Information
- OS: Linux
- Keyboard Layout: Portuguese (Portugal) - pt-PT
- Character affected: ç (lowercase)

dapovoa:
Fixes #12021

## Description
This PR fixes the issue where the lowercase 'ç' character doesn't work on Linux keyboards.

## Changes
- Added platform detection (`process.platform === 'darwin'`) to only apply the Alt key mapping on macOS
- This preserves the macOS behavior where Alt+C produces 'ç'
- Allows Linux users to type 'ç' directly without it being interpreted as Alt+C

## Testing
- Tested on Linux with Portuguese (pt-PT) keyboard layout
- Character 'ç' now works correctly
- Build passes with no errors

### DIFF
--- a/packages/cli/src/ui/contexts/KeypressContext.tsx
+++ b/packages/cli/src/ui/contexts/KeypressContext.tsx
@@ -485,6 +485,10 @@ function shouldUsePassthrough(): boolean {
   return process.env['PASTE_WORKAROUND'] !== 'false';
 }
 
+function isMacOS(): boolean {
+  return process.platform === 'darwin';
+}
+
 export function KeypressProvider({
   children,
   kittyProtocolEnabled,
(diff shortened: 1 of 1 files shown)


scidomino:
We're planning on merging #12051 which addresses this and another related issue so I will close this out.

PR #12029, opened 2025-10-25
Fix extension MCP env substitution

hebenezer-m:
## Summary

Fixes the GitHub MCP authentication failure where extension-loaded server configs sent headers like `Authorization: Bearer $GITHUB_MCP_PAT` literally. Changes include:

- Resolve `$VAR`/`${VAR}` placeholders in MCP server configurations supplied by extensions before initiating connections.
- Document the need to define `GITHUB_MCP_PAT` when using the official GitHub MCP extension (and similar extensions) and update the tutorial accordingly.
- Add unit coverage ensuring headers/env blocks from extensions resolve environment variables and fall back correctly when missing.

Verified against the published `@google/gemini-cli@latest` package by setting `GITHUB_MCP_PAT` via `pass`/`.env`; the GitHub MCP server now starts in the CLI with status `🟢 Ready` without the 400 error.

## Testing

- `npm ci`
- `npx vitest run packages/cli/src/config/config.test.ts`
- Manual run of `gemini` with `GITHUB_MCP_PAT` set (GitHub MCP server reports Ready and tools are available).


hebenezer-m:
Thanks for the detailed review! The compilation issue you flagged was addressed when the branch was rebased – the tests now use `GeminiCLIExtension`/`MCPServerConfig`, and `npx vitest run packages/cli/src/config/config.test.ts` passes locally. Let me know if anything else looks off.

docs/quota-and-pricing.md
chrstnb, line 1: Remove

docs/extensions/index.md
chrstnb, line 1: Remove these changes-- we added support for extension-defined config a few days ago!

docs/cli/tutorials.md
```diff
@@ -63,15 +63,41 @@ how to launch the GitHub MCP server.
...
 
 ```bash
-GITHUB_PERSONAL_ACCESS_TOKEN="pat_YourActualGitHubTokenHere"
+export GITHUB_PERSONAL_ACCESS_TOKEN="pat_YourActualGitHubTokenHere"
 ```
 
 Gemini CLI uses this value in the `mcpServers` configuration that you defined in
 the `settings.json` file.
 
+**Option 2: Using extensions (e.g., the GitHub MCP extension)**
```
chrstnb, line 79: I would remove these changes as well since the purpose of this doc is to do a single tutorial of the GitHub MCP server!

docs/cli/configuration.md
```diff
@@ -518,6 +518,25 @@ the `excludedProjectEnvVars` setting in your `settings.json` file.
 - **`CODE_ASSIST_ENDPOINT`**:
   - Specifies the endpoint for the code assist server.
   - This is useful for development and testing.
+- **`GITHUB_PERSONAL_ACCESS_TOKEN`**:
```
chrstnb, line 521: Because we want to be agnostic to 3P integrations, we don't want to declare these here in our top-level config. Not all users will be using github, and also not all MCP servers/extensions will use the same env var naming. Please remove these

packages/cli/src/config/config.ts
```diff
@@ -714,8 +715,9 @@ function mergeMcpServers(settings: Settings, extensions: GeminiCLIExtension[]) {
         );
         return;
       }
+      const resolvedServer = resolveEnvVarsInObject(server);
       mcpServers[key] = {
-        ...server,
+        ...resolvedServer,
```
chrstnb, line 720: I'm wondering if the bug here is actually in the extensions logic given it sounds like you're only seeing it in extensions. Can you clarify why this change is needed here versus there?

PR #12059, opened 2025-10-27
fix(#11226): incomplete model messages not added to chat services

### LINKED ISSUE #11226: Incomplete Model Messages not added to Chat Service

hugomurillomtz:
## TLDR

This is about fixing the issue about incomplete/cancelled model responses not being saved to files when doing `/chat save <tag>`. This PR brings a more robust solution that the one proposed here: https://github.com/google-gemini/gemini-cli/pull/11816, where a poluted history might confuse the model.

## Dive Deeper

The issue describes this happening when the model uses some tools, and then continues to add a response, and upon cancellation and then doing `/chat save <tag>`, the model incomplete response would not get stored. However, after looking into it, I noticed that it was a matter of incomplete model responses being disregarded regardless of whether there were tool calls before or after or not. Here's an example:

Request cancelled:
image

Then, when attempting to save the chat, we get this error:
image

And this was happening because a validation was taking place before pushing the model response to _History_: https://github.com/google-gemini/gemini-cli/blob/ce655436ef97535247daa9d27f572c1ca3ed62ac/packages/core/src/core/geminiChat.ts#L584

The simple solution (PR #11816) treated the user's view and the model's context as the same thing, when they needed to be different after a failure.

Because of that, I thought of a better solution (the one proposed here) by creating a way to distinguish between complete and partial messages. I thought about adding a **flag** property that would indicate if the response is a partial one or not. That way we can resume conversations with incomplete responses, but a clean, curated history with only successful turns for the model.

Here's the psedocode of the main new code (from the `geminiChat.ts` and `useGeminiStream.ts` classes respectively):

```
1 function sendMessageStream(userMessage):
2 
3   // Add the user's message to the history before any attempts.
4   history.push(userMessage)
5 
6   // Loop for handling retries.
7   for attempt from 1 to MAX_ATTEMPTS:
8 
9     // 1. Add a temporary placeholder for the model's response for this attempt.
10     //    This is what the UI will render as the stream comes in.
11     partialResponsePlaceholder = { role: 'model', parts: [], isPartial: true }
12     history.push(partialResponsePlaceholder)
13 
14     try:
15       // 2. Make the API call and process the stream.
16       //    If successful, this function will update the placeholder and remove the `isPartial` flag.
17       call_api_and_process_stream(userMessage, partialResponsePlaceholder)
18 
19       // 3. If the stream completes without error, we're done. Exit the loop.
20       return // Success
21 
22     catch (error):
23       // 4. An error occurred. Decide what to do.
24 
25       isRetryable = (error is InvalidStreamError)
26       hasMoreAttempts = (attempt < MAX_ATTEMPTS)
27 
28       if (isRetryable and hasMoreAttempts):
29         // --- SCENARIO A: WE WILL RETRY ---
30         // The stream was invalid, but we can try again.
31         // Remove the failed partial placeholder to keep the history clean for the next attempt.
32         history.pop()
33         wait_for_backoff()
34         continue // Go to the next iteration of the loop.
35       else:
36         // --- SCENARIO B: THIS IS THE FINAL FAILURE ---
37         // This is either a non-retryable error OR we've run out of retries.
38 
39         if (isRetryable): // This means we've exhausted all retries
40           // For a persistent invalid stream, the test expects a clean history.
41           // Remove the final failed placeholder.
42           history.pop()
43         else: // This is a non-retryable error (e.g., stream interruption)
44           // The test expects us to KEEP the partial response.
45           // The `isPartial: true` flag is already set on the placeholder, so we do nothing.
46           pass
47 
48         // 5. Re-throw the error to the caller and exit the function.
49         throw error
```

```
1 function cancelOngoingRequest():
2   // This function is called when the user cancels a streaming response.
3 
4   // Check if there is a response currently being streamed.
5   if a pendingHistoryItem exists:
6 
7     // --- RECONSTRUCTION LOGIC ---
8     // The goal is to rebuild the full text of the model's turn, which
9     // the UI may have split into multiple items for performance.
10 
11     reconstructedText = ""
12     tempHistory = a copy of the current UI history array
13 
14     // Get the most recent item from the history.
15     lastItem = tempHistory.pop()
16 
17     // 1. Go backwards through the history.
18     //    Collect the text from all items that are part of the current
19     //    model's turn (i.e., have type 'gemini' or 'gemini_content').
20     while lastItem exists AND (lastItem.type is 'gemini' OR lastItem.type is 'gemini_content'):
21       reconstructedText = lastItem.text + reconstructedText // Prepend the text
22       lastItem = tempHistory.pop() // Move to the next older item
23 
24     // 2. Add the text from the final, actively streaming piece.
25     reconstructedText = reconstructedText + pendingHistoryItem.text
26 
27     // --- COMMIT LOGIC ---
28 
29     // 3. Format the full text into the structure the backend expects.
30     historyItemToCommit = {
31       role: "model",
32       parts: [{ text: reconstructedText }]
33     }
34 
35     // 4. Send this complete, partial item to the backend to be saved.
36     geminiClient.commitCancelledResponse(historyItemToCommit)
37 
38     // 5. Add the final pending item to the UI history so it's displayed.
39     addItem(pendingHistoryItem)
```

## Reviewer Test Plan

Cancel an ongoing request to the model, and then save the conversation by doing `/chat save <tag>`, and you should be able to successfully save the conversation, and then, when resuming it, see the incomplete model responses.

image

Here you can see the `.json` file with the incomplete response and the `isPartial` flag:
image

And then, when resuming the conversation, we can see the incomplete response:
image
image

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ❓  | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |

## Linked issues / bugs

This PR fixes: https://github.com/google-gemini/gemini-cli/issues/11226


hugomurillomtz:
This is ready for review. The only failing check is unrelated to my changes:

image

hugomurillomtz:
I recently merged again to fix conflicts. A test is failing, but it's unrelated to my changes:

image

packages/cli/src/ui/commands/chatCommand.test.ts
```diff
@@ -222,6 +227,26 @@ describe('chatCommand', () => {
...
+        { role: 'user', parts: [{ text: 'Hello' }] },
+        {
+          role: 'model',
+          parts: [{ text: 'I am a partial response' }],
+          isPartial: true,
+        },
+      ];
+      mockGetHistory.mockReturnValue(history);
+
+      if (!saveCommand.action)
```
adamfweidman, line 244: nit: I don't think you need this if-statement. This error will throw if its not defined.

packages/cli/src/ui/commands/chatCommand.test.ts
```diff
@@ -222,6 +227,26 @@ describe('chatCommand', () => {
...
+        throw new Error('saveCommand.action not defined');
+      await saveCommand.action(mockContext, tag);
+
+      expect(mockSaveCheckpoint).toHaveBeenCalledWith(history, tag);
```
adamfweidman, line 248: nit: check the result of the save command similar to above

packages/core/src/core/logger.ts
```diff
@@ -348,7 +351,7 @@ export class Logger {
         );
         return [];
       }
-      return parsedContent as Content[];
+      return parsedContent as HistoryContent[];
```
adamfweidman, line 354: Can you confirm this doesn't break previously loaded checkpoints

packages/core/src/core/client.ts
```diff
@@ -123,6 +124,10 @@ export class GeminiClient {
     return this.chat;
   }
 
+  commitCancelledResponse(finalPartialContent: HistoryContent) {
```
adamfweidman, line 127: My concern here is with how we are doing implicit caching. I'm not sure how a user cancelled or errored out stream is handled @cornmander

packages/cli/src/ui/hooks/useGeminiStream.ts
```diff
@@ -305,6 +306,33 @@ export const useGeminiStream = (
...
+      // 2. Add the final pending chunk.
+      fullResponseText += pendingHistoryItemRef.current.text || '';
+
+      // 3. Commit the full, reconstructed partial response.
+      const historyItemToCommit: HistoryContent = {
+        role: 'model',
+        parts: [{ text: fullResponseText }],
+      };
+
+      geminiClient.commitCancelledResponse(historyItemToCommit);
```
adamfweidman, line 335: Same as above, I'm worried that mutating the history will effect caching. It could be this will be a huge caching save if stopped requests were previously resulting in cache miss.

Could we add a test (or integration test) to confirm this behavior. Particularly, were we previously not saving partial responses in the history despite displaying them in the UI?

packages/cli/src/ui/commands/chatCommand.test.ts
hugomurillomtz: Sure

packages/core/src/core/logger.ts
hugomurillomtz: That was a concern I had too. So, I designed this to be backwards compatible. The new checkpoints will have the new flag if incomplete, and the ones already registered will be left alone.

packages/cli/src/ui/hooks/useGeminiStream.ts
hugomurillomtz: I'll investigate more about the caching since it's a big concern you have.

And, before, we had two history arrays: one for the UI and another one that we shared with the model. That's why you'd see incomplete responses in the UI, but not being saved when doing `/chat save [tag]`, because we were using the second array for that where a condition was happening, preventing the incomplete response from saving it (I dive deeper into this in the PR description):

https://github.com/google-gemini/gemini-cli/blob/ce655436ef97535247daa9d27f572c1ca3ed62ac/packages/core/src/core/geminiChat.ts#L584

Now, we still have two history arrays, but we're saving incomplete responses when doing `/chat save [tag]`. And to fix the issue about the model getting "dirty" data (i.e. incomplete responses that could confuse it), I added a flag to incomplete responses to indicate they are that way so we can avoid filter them out when sending history to the model.

Issue #11643, opened 2025-10-22
Optimize Tests


MJjainam:
We have tests that are structurally repetitive in nature, has AI slop tests, uses timeouts over waitFor and have brittle assertions. 

For eg: 
In this particular test [section](https://github.com/google-gemini/gemini-cli/blob/2940b50811443d6dabeeec9b645e4930e8016ea6/packages/cli/src/ui/hooks/useCommandCompletion.test.ts#L656), there are three tests. 

All these three tests can be combined by parameterizing inputs and expected outputs. 

The task is to make the tests succinct and more robust wherever possible.

IamRiddhi:
PR - https://github.com/google-gemini/gemini-cli/pull/11821

IamRiddhi:
PR - https://github.com/google-gemini/gemini-cli/pull/11930

IamRiddhi:
PR - https://github.com/google-gemini/gemini-cli/pull/11969

PR #12061, opened 2025-10-27
refactor(core): Parameterize tests in glob.test.ts

MJjainam:
This commit refactors several test suites within `glob.test.ts` to use Vitest's `it.each` for parameterization. This change improves the readability and maintainability of the tests by consolidating repetitive logic and presenting test cases as clear, data-driven arrays.

## Linked issues / bugs
Progresses on #11643 
Progresses on #12142

### DIFF
--- a/packages/core/src/tools/glob.test.ts
+++ b/packages/core/src/tools/glob.test.ts
@@ -242,101 +242,70 @@ describe('GlobTool', () => {
   });
 
   describe('validateToolParams', () => {
-    it('should return null for valid parameters (pattern only)', () => {
-      const params: GlobToolParams = { pattern: '*.js' };
-      expect(globTool.validateToolParams(params)).toBeNull();
-    });
-
-    it('should return null for valid parameters (pattern and path)', () => {
-      const params: GlobToolParams = { pattern: '*.js', path: 'sub' };
-      expect(globTool.validateToolParams(params)).toBeNull();
-    });
-
-    it('should return null for valid parameters (pattern, path, and case_sensitive)', () => {
-      const params: GlobToolParams = {
-        pattern: '*.js',
-        path: 'sub',
-        case_sensitive: true,
-      };
-      expect(globTool.validateToolParams(params)).toBeNull();
-    });
-
-    it('should return error if pattern is missing (schema validation)', () => {
-      // Need to correctly define this as an object without pattern
-      const params = { path: '.' };
+    it.each([
+      {
+        name: 'should return null for valid parameters (pattern only)',
+        params: { pattern: '*.js' },
+        expected: null,
+      },
+      {
+        name: 'should return null for valid parameters (pattern and path)',
+        params: { pattern: '*.js', path: 'sub' },
+        expected: null,
+      },
+      {
+        name: 'should return null for valid parameters (pattern, path, and case_sensitive)',
+        params: { pattern: '*.js', path: 'sub', case_sensitive: true },
+        expected: null,
+      },
+      {
+        name: 'should return error if pattern is missing (schema validation)',
+        params: { path: '.' },
+        expected: `params must have required property 'pattern'`,
+      },
+      {
+        name: 'should return error if pattern is an empty string',
+        params: { pattern: '' },
+        expected: "The 'pattern' parameter cannot be empty.",
+      },
+      {
+        name: 'should return error if pattern is only whitespace',
+        params: { pattern: '   ' },
+        expected: "The 'pattern' parameter cannot be empty.",
+      },
+      {
+        name: 'should return error if path is not a string (schema validation)',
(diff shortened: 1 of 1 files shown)


scidomino (approved):

packages/core/src/tools/glob.test.ts
```diff
@@ -373,85 +342,74 @@ describe('GlobTool', () => {
...
+        filesToCreate: ['a.ignored.txt', 'b.notignored.txt'],
+        params: { pattern: '*.txt' },
+        expectedCountMessage: 'Found 3 file(s)',
+        notExpectedToContain: ['a.ignored.txt'],
+      },
+      {
+        name: 'should respect .geminiignore files by default',
+        ignoreFile: { name: '.geminiignore', content: '*.geminiignored.txt' },
+        filesToCreate: ['a.geminiignored.txt', 'b.notignored.txt'],
+        params: { pattern: '*.txt' },
```
jacob314: change params to globToolParams

packages/core/src/tools/glob.test.ts
```diff
@@ -373,85 +342,74 @@ describe('GlobTool', () => {
...
+        expectedCountMessage: 'Found 3 file(s)',
```
jacob314, line 369: why does this one say Found 3 files when only two files were created?

packages/core/src/tools/glob.test.ts
```diff
@@ -242,101 +242,70 @@ describe('GlobTool', () => {
...
-        path: 'sub',
-        case_sensitive: true,
-      };
-      expect(globTool.validateToolParams(params)).toBeNull();
-    });
-
-    it('should return error if pattern is missing (schema validation)', () => {
-      // Need to correctly define this as an object without pattern
-      const params = { path: '.' };
+    it.each([
```
scidomino, line 245: This is a pretty marginal use-case for parameterization. Each of these tests was only one line. (and personally, I would have just made these more compact by inlining the params variables.

packages/core/src/tools/glob.test.ts
```diff
@@ -464,110 +422,101 @@ describe('sortFileEntries', () => {
...
+    ({ entries, expected, expectedUnordered, recencyThresholdMs }) => {
+      const globPaths = entries.map((e) => createFileEntry(e.name, e.mtime));
+      const sorted = sortFileEntries(
+        globPaths,
+        nowTimestamp,
+        recencyThresholdMs ?? oneDayInMs,
+      );
+      const sortedPaths = sorted.map((e) => e.fullpath());
+
+      if (expected) {
```
jacob314, line 521: make these tests robust. throw if one of expectedOrdered and expectedUnordered is not provided.

packages/core/src/tools/glob.test.ts
```diff
@@ -373,85 +342,74 @@ describe('GlobTool', () => {
...
-        pattern: '*.txt',
-        respect_gemini_ignore: false,
-      };
-      const invocation = globTool.build(params);
-      const result = await invocation.execute(abortSignal);
-
-      expect(result.llmContent).toContain('Found 3 file(s)'); // fileA.txt, FileB.TXT, a.geminiignored.txt
-      expect(result.llmContent).toContain('a.geminiignored.txt');
-    });
+    it.each([
```
jacob314: these are quite complicated parameterizations. it is probably worth giving them static types so you get type errors if the JSON is the wrong format. Ask gemini to statically type all of these probably by specifying a types on line 380

jacob314 (commented):
these are generally good changes. added a couple minor comments.

packages/core/src/tools/glob.test.ts
MJjainam: @jacob314 Is it beneficial to type the input test cases?
 
Unlike the regular code, test code is always run pre-merge. And any errors would be caught before hand. 
Keeping test cases flexible for this small describe block would be more beneficial - wdyt?

packages/core/src/tools/glob.test.ts
MJjainam: The beforeEach block contains extra files which is common. 
Do you think pulling out beforeEach and AfterEach inside the test function make it more readable?

packages/core/src/tools/glob.test.ts
jacob314: types are still beneficial here because you want to make sure you really are testing what you think you are testing. With parameterized tests there is a real risk of getting in territory where the tests themselves need tests. At least if you have types you don't need to test cases the types would catch.

For example suppose you had a parameterized test that had a case like
{
  url: 'htp://invalidurl',
  throwsException: true
}

you might think this test was catching that the invalid url through an exception but without types there could be a typo as the parameter was really named
`throwException: true`
and so
the test was actually verifying that the invalid url didn't throw an exception.

PR #12076, opened 2025-10-27
Fix misreported number of lines being removed by model

Adib234:
## TLDR

When we exit the CLI, number of lines that get removed are reported. @bdmorgan brought it to my attention that in more advanced edits (multiple edits across multiple files), the number of lines seem to be a arbitrary large number. This PR attempts to correct the report of the number of lines being removed by the model. 

## Dive Deeper

The source of the discrepancy lies in `packages/core/src/tools/diffOptions.ts`, within the getStats function.
Breakdown of the Process:
`getStats` Function: This function is responsible for calculating the number of added and removed lines, among other statistics.
Input for Comparison: When calculating the model's statistics, the `getStats` function is provided with two strings to compare:
- AI String: This string contains only the new lines being added to the file.
- Original String: This is the entire content of the original file.

Flawed Comparison: The diff operation then compares the small AI-generated string of changes against the entire original file. This leads to a skewed result where the number of "removed" lines is arbitrarily large and not representative of the actual changes.

The fundamental issue is that for an accurate diff, the comparison should be between the original file and the original file with the AI's changes applied, not just the changes themselves. My suspicion is that the model originally outputted the original file with the AI's changes applied so that's why the diff worked, but now the model has changed to only output the changes.

The test is an example of an advanced edit and fails without my changes. 

## Reviewer Test Plan

Added unit test and manually validated with a repo on my end by comparing the number of lines that were removed by the model after a session with what `git diff --stat` reported.

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | 🟢   | ❓  | ❓  |
| npx      | ❓  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


packages/core/src/tools/edit.ts
```diff
@@ -387,8 +387,24 @@ class EditToolInvocation
...
+        originallyProposedContent = safeLiteralReplace(
+          editData.currentContent as string,
+          this.params.old_string,
+          this.params.ai_proposed_content,
+        );
+      } else {
+        originallyProposedContent = editData.newContent;
+      }
+
+      console.log(`originallyProposedContent ${originallyProposedContent}`);
```
silviojr: was this left intentionally or just used for debugging?

Adib234: Sorry, yes this is still in draft I will ping in the chat again when it's ready for review

packages/core/src/tools/smart-edit.ts
```diff
@@ -763,8 +763,23 @@ class EditToolInvocation
...
+          editData.currentContent as string,
+          this.params.old_string,
+          this.params.ai_proposed_string !== null &&
+            this.params.ai_proposed_string !== undefined
+            ? this.params.ai_proposed_string
+            : this.params.new_string,
+        );
+
+        console.log(
+          `adib originallyProposedContent ${originallyProposedContent}`,
```
silviojr: left intentionally or debug?

Adib234: Left for debugging, still verifying some things on my end on a different operating system

packages/core/src/tools/smart-edit.ts
```diff
@@ -763,8 +763,23 @@ class EditToolInvocation
           'Proposed',
           DEFAULT_DIFF_OPTIONS,
         );
-        const originallyProposedContent =
-          this.params.ai_proposed_string || this.params.new_string;
+
+        const originallyProposedContent = safeLiteralReplace(
```
silviojr: I wonder if we can just use `editData.newContent`

This is the content that is effectively replace the currentContent, right?

Adib234: Yep that's what I realized, will use `editData.newContent`

silviojr (approved):

PR #12118, opened 2025-10-27
Migrate tests to use avoid jsdom

jacob314:
This completes test migrations required to switch to my Ink branch which for package resolution issues doesn't work with jsdom.


packages/cli/src/test-utils/render.tsx
```diff
@@ -128,3 +129,54 @@ export const renderWithProviders = (
...
+            renderCallback={renderCallback}
+            props={props as Props}
+          />
+        </Wrapper>,
+      );
+    });
+  }
+
+  return { result, rerender, unmount };
+}
```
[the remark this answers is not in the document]
jacob314: done

packages/cli/src/test-utils/render.tsx
```diff
@@ -128,3 +129,52 @@ export const renderWithProviders = (
     </ConfigContext.Provider>,
   );
 };
+
+export function renderHook<Result, Props>(
```
jacob314, line 133: renderHook helper similar to existing renderHook method from the react web testing libraries to reduce extra boilerplate for simple tests validating hooks.

packages/cli/src/ui/contexts/SessionContext.test.tsx
```diff
@@ -4,17 +4,40 @@
...
-import { renderHook } from '@testing-library/react';
-import { act } from 'react-dom/test-utils';
+
+import { act } from 'react';
 import type { SessionMetrics } from './SessionContext.js';
 import { SessionStatsProvider, useSessionStats } from './SessionContext.js';
 import { describe, it, expect, vi } from 'vitest';
 import { uiTelemetryService } from '@google/gemini-cli-core';
 
+class ErrorBoundary extends Component<
```
jacob314, line 16: the existing test using snapshots for a non-snapshot task was causing flakiness issues on the bots that I couldn't reproduce locally so added this class to make it robust.

packages/cli/src/ui/hooks/usePhraseCycler.test.ts
```diff
@@ -1,210 +0,0 @@
-/**
```
jacob314, line 1: moved to usePhraseCyclerTest.tsx

packages/cli/src/config/extension.test.ts
```diff
@@ -61,11 +59,13 @@ vi.mock('simple-git', () => ({
   }),
 }));
 
+const mockHomedir = vi.hoisted(() => vi.fn(() => '/tmp/mock-home'));
```
sehoon38, line 62: why is this needed? does getting rid of jsdom requires this change?

packages/cli/src/ui/hooks/useToolScheduler.test.ts
```diff
@@ -4,12 +4,11 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-/** @vitest-environment jsdom */
-
 /* eslint-disable @typescript-eslint/no-explicit-any */
 import type { Mock } from 'vitest';
 import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
-import { renderHook, act } from '@testing-library/react';
```
sehoon38, line 12: I'm guessing @testing-library/react is removed as part of the jsdom deprecation?

packages/cli/src/ui/hooks/useReverseSearchCompletion.test.tsx
```diff
@@ -216,35 +229,40 @@ describe('useReverseSearchCompletion', () => {
               true,
             ),
           );
-
-          expect(result.current.suggestions.length).toBe(15);
-          expect(result.current.activeSuggestionIndex).toBe(0);
-          expect(result.current.visibleStartIndex).toBe(0);
+          await vi.waitFor(() => {
```
sehoon38: why does it now need to call the wait? was the previous renderHook from the import synchronous or something?

sehoon38 (approved):
lgtm, just had a couple of noob questions just for my learning. ty!

bdmorgan (approved):

packages/cli/src/ui/hooks/useToolScheduler.test.ts
jacob314: correct we should not be using @testing-library/react
We should be using ink-testing-library instead.
Similarly if we were using react native then we should have been using @testing-library/react-native.

packages/cli/src/config/extension.test.ts
jacob314: tested removing it and it was fine. sometimes Gemini CLI ends up adding some spurious changes as part of fixing tests. potentially we should have a workflow thatbacks out unneeded harmless parts of PRs to avoid this churn.

jacob314:
Pushed another commit that cleans up render a bit and removed a few waitFor that could be mitigated.
Many of the other waitFor couldn't be removed as they introduced timing issues. Generally when switching between test packages I'm not surprised we added a few timing cases that now need waitFor. Generally think waitFor is fairly harmless for us to add as generally expect we have a lot of tests where we were just lucky they worked without waitFor.
Additional commit to review:
https://github.com/google-gemini/gemini-cli/pull/12118/commits/65e60e0160ce5797f15481c258655eddc669cfb8

PR #12136, opened 2025-10-28
Fix: Update system instruction when GEMINI.md memory is loaded or refreshed

### LINKED ISSUE #11799: gemini-cli seems to ignore `GEMINI.md` context file
### What happened?

When using the gemini-cli with a local context file `./GEMINI.md`, the model seems to completely ignore the context.

This is the case even though the context file is correctly shown when running `/memory show`.

Only when explicitly telling the model to load the file `@GEMINI.md`, it seems to take it into consideration.

gemini-cli should load the context file at start and have it in memory without having to explicit tell it to do so.

### Client information

<details>

```console
> /about
│ About Gemini CLI                                                   │
│                                                                    │
│ CLI Version            0.10.0                                      │
│ Model                  gemini-2.5-pro                              │
│ Sandbox                no sandbox                                  │
│ OS                     darwin                                      │
│ Auth Method            gemini-api-key
```

</details>

### Login information

_No response_

_No response_

lifefloating:
## Summary

Fixes #11799 - Model now correctly uses GEMINI.md context without requiring explicit `@GEMINI.md` mentions. The issue was that while memory content was loaded into config state, the active chat's system instruction wasn't updated, causing the model to ignore the context.

## Details

**Root cause:**
When GEMINI.md files were loaded (at startup or via `/memory refresh`/`/directory add`), the content was stored in config state but the chat's system instruction remained unchanged. The system instruction was only set once during `startChat()`, so any subsequent memory updates were invisible to the model.

## Related Issues

Fixes #11799

## How to Validate

**Test case from issue:**

1. Create a GEMINI.md file with specific instructions:
   ```bash
   cat > GEMINI.md << 'EOF'
   For all the user requests related to this project, please insert a short
   anecdote, joke or a reference to cats at the end of your response.
   EOF
   ```

2. Start gemini-cli:
   ```bash
   gemini
   ```

3. Verify memory is loaded:
   ```
   > /memory show
   ```
   Should display the GEMINI.md content.

4. Test that instructions are followed WITHOUT using `@GEMINI.md`:
   ```
   > What's 2+2?
   ```

**Expected result:**
Model should include a cat-related anecdote/joke in the response, demonstrating that GEMINI.md context is being used.

**Additional validation:**

Run memory command tests:
```bash
cd packages/cli
npm test -- memoryCommand.test.ts
```

All 10 tests should pass.

## Pre-Merge Checklist

- [x] Validated on required platforms/methods:
  - [x] MacOS
    - [x] npm run
    - [ ] npx


lifefloating:
@jacob314

bdmorgan (approved):

PR #12190, opened 2025-10-28
Remove Todo Icon

scidomino:
## Summary

Removes the todo icon.


jacob314 (approved):
Can we add a snapshot test for the TODO message UI? Would have thought this change would cause a test to fail.lgtm

Issue #11488, opened 2025-10-19
Refactor Hierarchical Memory Loading: Implement Dynamic, Just-in-Time Context


abhipatel12:
## Problem Statement
Currently, the Gemini CLI eagerly discovers and loads *all* `GEMINI.md` (or configured context files) starting from the current working directory (CWD) and searching downwards recursively.

In large repositories or monorepos, this approach has significant downsides:
1.  **Token Bloat:** It immediately floods the context window with potentially irrelevant instructions from subdirectories that may never be touched during the session.
2.  **Noise:** The model may be distracted by conflicting or irrelevant rules from distant parts of the codebase.
3.  **Startup Latency:** Performing a full BFS file search at startup can be slow in large directory structures.

## Proposed Architecture
We should shift from an eager loading model to a hybrid model that combines persistent global/workspace context with dynamic, just-in-time loading for subdirectories.

### The 3-Tier Context Model

| Tier | Location | Loading Strategy | Injection Point | Persistence |
| :--- | :--- | :--- | :--- | :--- |
| **1. Global** | `~/.gemini/GEMINI.md` | Eager (Startup) | System Prompt | Permanent |
| **2. Environment** | `<project_root>/GEMINI.md` AND Active Extension `GEMINI.md`s | Eager (Startup) | First User Message | Permanent (Pinned)* |
| **3. Subdirectory** | `./path/to/dir/GEMINI.md` | Lazy (Dynamic) | Tool Output Turn | Transient (Compressible) |

*\* "Permanent (Pinned)" means this content is re-injected into the first user message whenever the chat history is compressed or regenerated.*

### 1. Startup Phase
At startup, the agent should *only* load Tier 1 and Tier 2 context:
*   **Tier 1:** The global memory file.
*   **Tier 2:**
    *   The workspace root memory file.
    *   All `GEMINI.md` files associated with currently active extensions.
*   *Crucially, it should NO LONGER perform a downward BFS search for other memory files.*

### 2. Dynamic Loading Phase (Runtime)
Subdirectory context (Tier 3) is loaded only when the agent demonstrates clear intent to work within that scope.

#### Triggering Tools
Dynamic loading should be triggered by "high-intent" tools that operate on specific files or directories:
*   `read_file`
*   `read_many_files`
*   `list_directory`
*   `write_file` / `replace` (though ideally context is loaded *before* writing, these should still trigger a check to ensure subsequent actions have context).

[...]

PR #12195, opened 2025-10-29
feat(core): Implement granular memory loaders for JIT architecture

abhipatel12:
## Summary

This PR implements the foundational granular memory loading functions required for the new 3-tier Just-in-Time (JIT) memory architecture. It refactors the memory discovery logic into distinct, composable functions without altering the existing monolithic loader used by the CLI today.

>[!NOTE]
> These changes are purely additive and do not yet affect the active CLI memory loading behavior.

## Details

To support faster startup and reduced token usage, we are moving to a model where subdirectory-specific `GEMINI.md` files are loaded only when needed (JIT). This PR prepares the core logic for this transition by implementing:

### New Loader Functions

- **Tier 1 Loader (`loadGlobalMemory`)**: Loads only the global `~/.gemini/GEMINI.md` file. Parallelized for multiple filename variants.

- **Tier 2 Loader (`loadEnvironmentMemory`)**: Loads project-level context from **exactly** the trusted root directories (no upward traversal), plus extension memory. This is **more restrictive** than the legacy loader for security reasons (see Security Notes below). Parallelized for multiple trusted roots.

- **Tier 3 Loader (`loadJitSubdirectoryMemory`)**: A new loader designed to be called dynamically by tools. It safely traverses upward from a target path to the nearest trusted root, loading only _new_ `GEMINI.md` files that haven't already been loaded.

### Architectural Choices

- **Raw Data Return**: All new loaders return raw `{ path, content }` arrays instead of pre-formatted strings, enabling flexible formatting for different contexts (tools vs system prompt).

- **Parallelization**: Implemented `Promise.all` for concurrent operations:
  - Multiple trusted roots processed in parallel
  - Filename variant checks parallelized per directory
  - Performance matches or exceeds legacy implementation

- **File Ordering**: Files are grouped by directory level (all variants from same directory are adjacent)

### Security Notes

This PR implements a **stricter trust model** than the legacy loader:

- **Legacy behavior**: If you trust `/usr/bob/projects/gemini-cli/src`, it would traverse upward and load from `/usr/bob/projects/gemini-cli/GEMINI.md` (even  though only the `src` subdirectory was explicitly trusted).

- **New behavior**: Only loads from the exact trusted directory itself. If you trust `/usr/bob/projects/gemini-cli/src`, it loads ONLY from that directory, not from parent directories.

This ensures explicit trust boundaries are respected.

## Related Issues

Prefactor for #11488

## How to Validate

Run the updated unit tests for the core package, which now include coverage for the new loaders:

```bash
npm run test -w @google/gemini-cli-core src/utils/memoryDiscovery.test.ts
```

## Pre-Merge Checklist

- [X] Updated relevant documentation and README (if needed)
- [X] Added/updated tests (if needed)
- [X] Noted breaking changes (if any)
- Validated on required platforms/methods:
  - [X] MacOS
    - [X] npm run
    - [X] Seatbelt


abhipatel12:
@cornmander 

Lmk what you think for `loadEnvironmentMemory`.

From two options: "Trust ONLY the exact directory" and "Trust directory and traverse up to project root", I chose the former which is more strict. I don't really think this would happen often in practice (where a user trusts a subdir of the project), but not the project root, but wanted to flag and make sure this was the correct pattern.

cornmander (approved):

Issue #11084, opened 2025-10-14
Fix Inaccurate Token Count Estimation in Chat Compression


abhipatel12:
## Description

The `tryCompressChat` method in `packages/core/src/core/client.ts` uses a naive local estimation to calculate the token count of the compressed chat history.

```typescript
// packages/core/src/core/client.ts
    // Estimate token count 1 token ≈ 4 characters
    const newTokenCount = Math.floor(
      chat
        .getHistory()
        .reduce((total, content) => total + JSON.stringify(content).length, 0) /
        4,
    );
```

This approach has two main issues:

1.  **Inconsistent Comparison:** It compares this locally estimated `newTokenCount` (history only) with `originalTokenCount` (full prompt from API metadata which includes system instructions, tool definitions, and user memory files). This is an apples-to-oranges comparison.
2.  **Bad Estimate for Images:** By `JSON.stringify`-ing the entire content, base64 encoded image data is treated as text. Applying the `charCount / 4` rule to base64 data results in a miscalculated estimate for images.

## Impact

* The 4 char/token heuristic may also underestimate the token count for non-english languages
* Compression may not have been as successful as it seems (i.e tokens from 40k to 2k may not be including the 15k tokens worth of SI, tool definitions, etc. 

## Location

`packages/core/src/core/client.ts` within the `tryCompressChat` method.

## Potential Solutions to Explore

The implementer should investigate solutions that balance accuracy and latency. Previous attempts to use `countTokens` on every turn were rejected due to latency concerns.

Possible approaches include:

*   **Improved Local Estimation:** Implement logic to estimate text tokens based on characters and assign a fixed token cost to image parts, avoiding `JSON.stringify` on binary data.
*   **Hybrid Approach:** Use local estimation to trigger compression, but use `countTokens` after compression to get an accurate `newTokenCount` for verification and telemetry updates.
*   **Caching:** Cache the token count of static parts of the prompt (system instructions, tools) to improve the accuracy of local estimates.

scidomino:
We have a report of a user that has hit this with a 100k token pdf file that is being incorrectly estimated to have 10 times more tokens ~990k. We should do the hybrid solution here.

PR #12196, opened 2025-10-29
refactor(core): decouple GeminiChat from uiTelemetryService via Usage events

SandyTao520:
## TLDR

Refactors `GeminiChat` to remove its direct dependency on `uiTelemetryService`. Instead, `GeminiChat` now tracks the last prompt token count internally, and `Turn` yields a new `Usage` event type. The UI layer (`useGeminiStream`) now handles updating the telemetry service based on these events.

## Dive Deeper

This improves separation of concerns. `GeminiChat` shouldn't need to know about UI telemetry. By yielding `Usage` events, any consumer of the `Turn` stream can decide how to handle usage data. `GeminiClient` and `ChatCompressionService` now read the token count directly from the `GeminiChat` instance they are working with.

## Reviewer Test Plan

Run a long conversation and verify that token usage is still being tracked correctly in the UI (e.g., check if context overflow warnings still appear when appropriate, or if the debug stats show correct token counts if visible).

## Testing Matrix

|          | 🍏  | 🪟  | 🐧  |
| npm run  | ✅  | ❓  | ❓ |
| npx      | ✅  | ❓  | ❓  |
| Docker   | ❓  | ❓  | ❓  |
| Podman   | ❓  | -   | -   |
| Seatbelt | ❓  | -   | -   |


packages/cli/src/ui/hooks/useGeminiStream.test.tsx
```diff
@@ -2373,6 +2378,47 @@ describe('useGeminiStream', () => {
...
+              'IMPORTANT: This conversation approached the input token limit',
+            ),
+          }),
+          expect.any(Number),
+        );
+      });
+    });
+  });
+
   it('should process @include commands, adding user turn after processing to prevent race conditions', async () => {
```
abhipatel12, line 2376: Do you think we need a unit test for the `Usage` event?

packages/core/src/core/geminiChat.ts
```diff
@@ -195,6 +195,9 @@ export class GeminiChat {
     validateHistory(history);
     this.chatRecordingService = new ChatRecordingService(config);
     this.chatRecordingService.initialize();
+    this.lastPromptTokenCount = Math.ceil(
+      JSON.stringify(this.history).length / 4,
```
abhipatel12, line 199: It seems like we have this token estimation logic in a few places now (chat compression, tests, and here). 

It also doesn't take into account the tokens used by tool definitions or system instructions (including GEMINI.md) right?

Can we create an issue to extract estimation to a separate util? This could be useful elsewhere as well

I had created #11084 as well that we can just mention.

abhipatel12 (commented):
Looks great overall! 

Just to validate, did we make sure that /stats and other features that depend on the UI telemetry service still work as expected?

packages/core/src/core/geminiChat.ts
SandyTao520: Yeah this is a problem, we need a better estimate for input tokens.

Let's do it altogether in https://github.com/google-gemini/gemini-cli/issues/11084.

SandyTao520:
Tested % context left update works, manual /compress and /clear also updates it correctly.

abhipatel12 (approved):
Sweet, looks great! LGTM!

Issue #12183, opened 2025-10-28
Å tangeten funkar inte efter uppdatering svenska den funkar endast med shift.


danielheiman666-hue:
### What happened?

Bug Report: Lowercase 'å' key not working in Gemini CLI (v0.10.0+)
    2 
    3 Description:
    4 After updating the Gemini CLI to version 0.10.0 (and subsequent versions), the lowercase 'å' character cannot be
      input directly. When attempting to type 'å', no character appears.
    5 
    6 Reproduction Steps:
    7 1. Open the Gemini CLI.
    8 2. Attempt to type the lowercase character 'å'.
    9 
   10 Expected Behavior:
   11 The lowercase 'å' character should appear in the CLI input.
   12 
   13 Actual Behavior:
   14 No character is input when typing lowercase 'å'.
   15 
   16 Workaround:
   17 Typing the uppercase 'Å' character using Shift + 'å' works as expected.
   18 
   19 Environment:
   20 - Gemini CLI Version: 0.10.0+
   21 - Operating System: darwin (macOS)
   22 - Keyboard Layout: Swedish (or any layout where 'å' is a distinct key)

### What did you expect to happen?

å tangent skall fungera

### Client information

* **CLI Version:** 0.10.0
* **Git Commit:** 5d92b507
* **Session ID:** 1d8e35b7-e5fc-43ee-8a18-02245c4ea257
* **Operating System:** darwin v22.20.0
* **Sandbox Environment:** no sandbox
* **Model Version:** gemini-2.5-flash
* **Memory Usage:** 189.9 MB

PR #12217, opened 2025-10-29
fix: properly handle UTF-8 grapheme clusters for international charac…

jaywang172:
…ters

Fixes #12212, #12183

Problem:
The CLI failed to correctly display multi-byte UTF-8 characters like 'ç' (Portuguese), 'å' (Swedish), and other international characters. Users reported characters would not appear when typed, breaking UX for non-English speakers.

Root cause:
The `toCodePoints()` function used `Array.from()` which handles surrogate pairs but not grapheme clusters. This broke:
- Combining marks (e.g., 'é' = 'e' + combining acute)
- Complex emoji (e.g., family emoji with ZWJ sequences)
- Cursor positioning with international characters

Solution:
Replace character segmentation with `Intl.Segmenter` API (Node.js 16+, we require 20+) which properly segments strings by grapheme clusters.

Changes:
- textUtils.ts: Add `toGraphemes()` using Intl.Segmenter
- textUtils.ts: Add TypeScript declarations for Intl.Segmenter
- textUtils.ts: Keep `toCodePoints()` as deprecated alias
- text-buffer.ts: Update to use `toGraphemes()`
- MaxSizedBox.tsx: Update grapheme handling
- vim-buffer-actions.ts: Update grapheme handling
- InputPrompt.tsx: Update grapheme handling
- useCommandCompletion.tsx: Update grapheme handling
- graphemes.test.ts: Add 37 comprehensive test cases

Testing (46/46 tests pass):
✓ ASCII strings (fast path unchanged)
✓ Portuguese ç, Swedish å, German ü, French é
✓ Combining marks
✓ Emoji (basic, family, flags, skin tones)
✓ Mixed content
✓ Cache behavior
✓ Regression tests for #12212, #12183

Performance:
- ASCII fast path unchanged (optimal for English)
- Unicode uses cached Intl.Segmenter results
- No external dependencies (native Node.js API)

The fix is surgical: only character boundary detection changed, all other logic unchanged. Zero breaking changes.

## Pre-Merge Checklist

  - [ ] MacOS
    - [ ] npm run
    - [ ] npx
    - [ ] Docker
    - [ ] Podman
    - [ ] Seatbelt
  - [ ] Windows
    - [ ] npm run
    - [ ] npx
    - [ ] Docker
  - [ ] Linux
    - [ ] npm run
    - [ ] npx
    - [ ] Docker

### DIFF
--- a/packages/cli/src/ui/utils/textUtils.ts
+++ b/packages/cli/src/ui/utils/textUtils.ts
@@ -153,14 +153,17 @@ export function toCodePoints(str: string): string[] {
 }
 
 /**
- * Get the length of a string in grapheme clusters.
- * This returns the number of user-perceived characters, not UTF-16 code units.
+ * Get the length of a string in code points (not grapheme clusters).
+ *
+ * This function returns code point count, which is used for offset calculations
+ * in logicalPosToOffset/offsetToLogicalPos. For UI display purposes (cursor
+ * positioning, text width), use toGraphemes().length instead.
  *
  * @param str - The string to measure
- * @returns The number of grapheme clusters
+ * @returns The number of code points
  */
 export function cpLen(str: string): number {
-  return toGraphemes(str).length;
+  return toCodePoints(str).length;
 }
 
 /**
--- a/packages/cli/src/ui/utils/textUtils.ts
+++ b/packages/cli/src/ui/utils/textUtils.ts
@@ -8,6 +8,11 @@ import stripAnsi from 'strip-ansi';
 import ansiRegex from 'ansi-regex';
 import { stripVTControlCharacters } from 'node:util';
 import stringWidth from 'string-width';
+import {
+  tokenize,
+  styledCharsFromTokens,
+  type StyledChar,
+} from '@alcalzone/ansi-tokenize';
 
 /**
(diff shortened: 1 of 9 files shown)


packages/cli/src/ui/utils/textUtils.ts
```diff
@@ -9,6 +9,31 @@ import ansiRegex from 'ansi-regex';
 import { stripVTControlCharacters } from 'node:util';
 import stringWidth from 'string-width';
 
+// Type definitions for Intl.Segmenter (ES2022+, Node.js 16+)
+// This is available in modern Node.js but TypeScript types may not include it
```
jacob314: There should be an existing d.ts file for this

jacob314: we should also consider aligning with what Ink is using. 
Specifically see
https://github.com/jacob314/ink/blob/master/src/measure-text.ts#L30

if that logic also needs to be switched to do proper grapheme cluster merging we should add it.
This matters so that cursor positioning and word wrap are consistent in Ink and text-buffer. Otherwise they will diverge resulting in hard to debug bugs.

PR #12232, opened 2025-10-29
Enforce timeout for subagents 

silviojr:
## Summary

We noticed that the agent was running over the time limit imposed by its configurations. 
Our main hypothesis is that the model call is taking too long (including the retries) and since the timeouts are only checked after each turn, they were not being enforced. 

This pull request refactors the agent execution timeout logic to use a more modern and efficient `AbortController`-based approach. This change ensures that subagents adhere strictly to their configured `max_time_minutes` by actively aborting their operations when the time limit is reached, while also maintaining the ability to respond to external abort signals. The update improves the reliability and clarity of agent termination events by providing distinct reasons for termination.

### Highlights

* **New Timeout Mechanism**: Implemented a new timeout mechanism using `AbortController` and `AbortSignal.any()` to enforce `max_time_minutes` for agent execution, providing a more robust and immediate termination.
* **Combined Abort Signals**: The agent now listens to both an external abort signal and an internal time-based timeout signal, allowing for graceful termination based on whichever event occurs first.
* **Differentiated Termination Reasons**: The system can now accurately distinguish between an agent being explicitly aborted by an external signal and an agent timing out due to exceeding its allocated time, setting the `terminateReason` accordingly.
* **Removed Legacy Timeout Check**: The previous polling-based timeout check, which relied on calculating `elapsedMinutes`, has been removed in favor of the new `AbortController`-based approach, streamlining the agent's execution loop.

## Pre-Merge Checklist

  - [ ] MacOS
    - [ ] npm run
    - [ ] npx
    - [ ] Docker
    - [ ] Podman
    - [ ] Seatbelt
  - [ ] Windows
    - [ ] npm run
    - [ ] npx
    - [ ] Docker
  - [ ] Linux
    - [x] npm run
    - [ ] npx
    - [ ] Docker

### DIFF
--- a/packages/core/src/agents/executor.ts
+++ b/packages/core/src/agents/executor.ts
@@ -159,6 +159,16 @@ export class AgentExecutor<TOutput extends z.ZodTypeAny> {
     let terminateReason: AgentTerminateMode = AgentTerminateMode.ERROR;
     let finalResult: string | null = null;
 
+    const { max_time_minutes } = this.definition.runConfig;
+    const timeoutController = new AbortController();
+    const timeoutId = setTimeout(
+      () => timeoutController.abort(new Error('Agent timed out.')),
+      max_time_minutes * 60 * 1000,
+    );
+
+    // Combine the external signal with the internal timeout signal.
+    const combinedSignal = AbortSignal.any([signal, timeoutController.signal]);
+
     logAgentStart(
       this.runtimeContext,
       new AgentStartEvent(this.agentId, this.definition.name),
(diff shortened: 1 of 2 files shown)


packages/core/src/agents/executor.ts
```diff
@@ -180,8 +190,11 @@ export class AgentExecutor<TOutput extends z.ZodTypeAny> {
           terminateReason = reason;
           break;
         }
-        if (signal.aborted) {
-          terminateReason = AgentTerminateMode.ABORTED;
+        if (combinedSignal.aborted) {
```
silviojr, line 193: I wonder when will this actually be true? 

It seems to me that when the abort signal is triggered, an exception is raised. So I don't think this block will ever be true. Whenever the abort signal is triggered, an exception will be raised and this block will be unreachable. Only the catch(...) and finally(...) block will be executed after that. 

thoughts?

silviojr: I actually did another research and check and it seem that in between async functions, this signal needs to be checked. Not all functions will throw errors.

anj-s (approved):

Issue #11108, opened 2025-10-14
Extension reloading in the CLI


chrstnb:
Support live reloading of extensions while in the CLI. We need to ensure this is possible.

This likely means reloading all of the following:
- MCP servers
- Context
- Custom commands

PR #12239, opened 2025-10-29
Initial support for reloading extensions in the CLI - mcp servers only

jakemac53:
## Summary

Adds support for unloading and reloading extensions MCP servers within the interactive CLI, based on events sent from the ExtensionLoader.

## Details

The only time this can happen today is in response to an `/extensions update` command, which causes an "unload" and then "load" event.

On "unload" or "disable" we will shut down all previous MCP servers from an extension, and clear the tools from the tool manager and gemini chat context.

On "load" or "enable" we start up any MCP servers from an extension, and add its tools to the tool manager and gemini chat context.

This functionality is guarded behind an `experimental.extensionReloading` setting.

I also updated the discovery state management in the McpClientManager to avoid race conditions.

Because this is so intertwined across different systems, I wrote an integration test instead of unit tests. If you would like, I could also add unit tests for some of it, but they won't be very valuable imo.

## Related Issues

First part of https://github.com/google-gemini/gemini-cli/issues/11108

## How to Validate

Install an extension, add an update to it (add a new mcp server), and update it using `/extensions update <your-extension>`. You should see its status show as updated, and if running `/mcp list` you should see the new mcp server connected. You should also be able to use gemini to invoke those tools.

## Pre-Merge Checklist

- [x] Noted breaking changes (if any)
  - [ ] MacOS
    - [ ] npm run
    - [ ] npx
    - [ ] Docker
    - [ ] Podman
    - [ ] Seatbelt
  - [ ] Windows
    - [ ] npm run
    - [ ] npx
    - [ ] Docker
  - [x] Linux
    - [x] npm run
    - [ ] npx
    - [ ] Docker


integration-tests/extensions-reload.test.ts
```diff
@@ -0,0 +1,96 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { expect, test } from 'vitest';
```
chrstnb: Can you test when env vars are changed?

packages/core/src/tools/mcp-client-manager.ts
```diff
@@ -23,66 +28,186 @@ import { coreEvents } from '../utils/events.js';
...
+    await Promise.all(
+      Object.keys(extension.mcpServers ?? {}).map((name) => {
+        const newMcpServers = {
+          ...this.cliConfig.getMcpServers(),
+        };
+        delete newMcpServers[name];
+        this.cliConfig.setMcpServers(newMcpServers);
+        return this.disconnectClient(name);
+      }),
+    );
```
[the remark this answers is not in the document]
jakemac53: Javascript is single threaded, this is wrong

packages/core/src/tools/mcp-client-manager.ts
```diff
@@ -23,66 +28,186 @@ import { coreEvents } from '../utils/events.js';
...
+    debugLogger.log(`Loading extension: ${extension.name}`);
+    await Promise.all(
+      Object.entries(extension.mcpServers ?? {}).map(([name, config]) => {
+        this.cliConfig.setMcpServers({
+          ...this.cliConfig.getMcpServers(),
+          [name]: config,
+        });
+        return this.discoverMcpTools(name, config);
+      }),
+    );
```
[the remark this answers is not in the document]
jakemac53: nope

integration-tests/extensions-reload.test.ts
jakemac53: Oh interesting, so like if the extension has settings, and on update you have to add new settings, does it pick those up?

packages/core/src/tools/mcp-client-manager.ts
```diff
@@ -23,66 +28,186 @@ import { coreEvents } from '../utils/events.js';
...
+          this.unloadExtension(event.extension),
+        )
+        .on('extensionUnloaded', (event) =>
+          this.unloadExtension(event.extension),
+        );
+    }
+  }
+
+  /**
+   * For all the MCP servers associated with this extension:
```
chrstnb, line 61: Wondering if we can/should store session IDs for http connections to persist

chrstnb:
How can we ensure the prompt is accurate and not polluted by the previous state?

packages/core/src/tools/mcp-client-manager.ts
jakemac53: Hmm yeah that could be interesting but I think probably isn't worth the complexity right now. But, worth filing a bug to track?

jakemac53:
> How can we ensure the prompt is accurate and not polluted by the previous state?

I was trying to avoid using the model in these tests but I could actually make it call into the model - basically ask it to make a tool call to the old tool, and check that it can't do that. Or, ask it what tools it has available.

jakemac53:
Another option regarding the tests is I could try setting up a unit test but one that doesn't many or even any mocks, and then  I could potentially validate the internals a bit better (like the exact tool list available to the GeminiChat).

packages/core/src/tools/mcp-client-manager.ts
chrstnb: Sg-- I think unexpectedly losing state for a user is something we'd want to be careful to avoid

jakemac53:
Ok, I took a look at trying to do more of a unit test but I ran into a number of issues. Abandoning that for now, integration test it is.

chrstnb (approved):

PR #12247, opened 2025-10-29
Only show screen reader notice once

chrstnb:
## Pre-Merge Checklist

  - [ ] MacOS
    - [ ] npm run
    - [ ] npx
    - [ ] Docker
    - [ ] Podman
    - [ ] Seatbelt
  - [ ] Windows
    - [ ] npm run
    - [ ] npx
    - [ ] Docker
  - [ ] Linux
    - [ ] npm run
    - [ ] npx
    - [ ] Docker


packages/cli/src/ui/components/Notifications.tsx
```diff
@@ -5,41 +5,77 @@
...
+
+import * as fs from 'node:fs';
+import os from 'node:os';
 import path from 'node:path';
 
-const settingsPath = path.join(homedir(), GEMINI_DIR, 'settings.json');
+const settingsPath = path.join(os.homedir(), GEMINI_DIR, 'settings.json');
+
+const screenReaderNudgeFilePath = path.join(
+  '/tmp',
```
jacob314: shouldn't this be under ~/.gemini/tmp not `/tmp/.gemini`
find existing helpers to make sure you get that path in a way that works on all platforms.

packages/cli/src/ui/components/Notifications.tsx
```diff
@@ -5,41 +5,77 @@
...
+    }
+  });
+
+  const showScreenReaderNudge =
+    isScreenReaderEnabled && !hasSeenScreenReaderNudge;
+
+  useEffect(() => {
+    if (showScreenReaderNudge) {
+      try {
+        fs.mkdirSync(path.dirname(screenReaderNudgeFilePath), {
```
jacob314: avoid sync file IO as it will hang the UI if the file system is slow

packages/cli/src/ui/components/Notifications.tsx
```diff
@@ -5,41 +5,77 @@
...
+          recursive: true,
+        });
+        fs.writeFileSync(screenReaderNudgeFilePath, 'true');
+      } catch (_error) {
+        // No-op
```
jacob314: please log

packages/cli/src/ui/components/Notifications.tsx
```diff
@@ -5,41 +5,77 @@
...
   const { initError, streamingState, updateInfo } = useUIState();
+
   const isScreenReaderEnabled = useIsScreenReaderEnabled();
   const showStartupWarnings = startupWarnings.length > 0;
   const showInitError =
     initError && streamingState !== StreamingState.Responding;
 
+  const [hasSeenScreenReaderNudge, _] = useState(() => {
+    try {
+      fs.accessSync(screenReaderNudgeFilePath);
```
jacob314: same comment as bellow about sync file io

packages/cli/src/ui/components/Notifications.tsx
```diff
@@ -5,41 +5,77 @@
...
+      }
+    }
+  }, [showScreenReaderNudge]);
+
   if (
     !showStartupWarnings &&
     !showInitError &&
     !updateInfo &&
-    !isScreenReaderEnabled
+    !showScreenReaderNudge
```
jacob314, line 76: why is 
!isScreenReaderEnabled removed?

chrstnb: because it's captured in showScreenReaderNudge

packages/cli/src/ui/components/Notifications.tsx
```diff
@@ -5,41 +5,85 @@
...
 export const Notifications = () => {
   const { startupWarnings } = useAppContext();
   const { initError, streamingState, updateInfo } = useUIState();
+
   const isScreenReaderEnabled = useIsScreenReaderEnabled();
   const showStartupWarnings = startupWarnings.length > 0;
   const showInitError =
     initError && streamingState !== StreamingState.Responding;
 
+  const [hasSeenScreenReaderNudge, setHasSeenScreenReaderNudge] =
```
jacob314: this should default to undefined or another unknown value rather than true.

packages/cli/src/ui/components/Notifications.tsx
```diff
@@ -5,41 +5,77 @@
...
 
   return (
     <>
-      {isScreenReaderEnabled && (
+      {showScreenReaderNudge && (
         <Text>
           You are currently in screen reader-friendly view. To switch out, open{' '}
-          {settingsPath} and remove the entry for {'"screenReader"'}.
+          {settingsPath} and remove the entry for {'"screenReader"'}. This will
+          disappear on re-start.
```
jacob314: nit: re-start to restart or "next run"

packages/cli/src/ui/components/Notifications.tsx
```diff
@@ -5,41 +5,77 @@
...
+import os from 'node:os';
 import path from 'node:path';
 
-const settingsPath = path.join(homedir(), GEMINI_DIR, 'settings.json');
+const settingsPath = path.join(os.homedir(), GEMINI_DIR, 'settings.json');
+
+const screenReaderNudgeFilePath = path.join(
+  '/tmp',
+  GEMINI_DIR,
+  'hasSeenScreenReaderNudge',
```
jacob314: nit hasSeenScreenReaderNudge -> seen_screen_reader_nudge.json
or better yet pick file name you can use for all the tmp data that should not be in settings such as local_storage.json or something. Expect at least 
 'hasSeenIdeIntegrationNudge' should get moved here. Looking at the code I suspect hasSeenIdeIntegrationNudge is broken today as it is using the obsolete scheme name instead of `hasSeenIdeIntegrationNudge` instead of `ide.hasSeenNudge` unless we still happen to be doing enough migration of legacy schema names to make this work.

jacob314 (approved):
please add a test before landing so this doesn't regress.lgtm

PR #12310, opened 2025-10-30
Update mcp-server.md

richardkmichael:
## Summary

- STDIO example server arguments shouldn't suggest http
- Demonstrate multiple env vars

## Pre-Merge Checklist

  - [ ] MacOS
    - [ ] npm run
    - [ ] npx
    - [ ] Docker
    - [ ] Podman
    - [ ] Seatbelt
  - [ ] Windows
    - [ ] npm run
    - [ ] npx
    - [ ] Docker
  - [ ] Linux
    - [ ] npm run
    - [ ] npx
    - [ ] Docker

### DIFF
--- a/docs/tools/mcp-server.md
+++ b/docs/tools/mcp-server.md
@@ -933,7 +933,7 @@ gemini mcp add [options] <name> <command> [args...]
 gemini mcp add -e API_KEY=123 -e DEBUG=true my-stdio-server /path/to/server arg1 arg2 arg3
 
 # Example: Adding a local python server
-gemini mcp add python-server python server.py --server-arg my-value
+gemini mcp add python-server python server.py -- --server-arg my-value
 ```
 
 #### Adding an HTTP server
--- a/docs/tools/mcp-server.md
+++ b/docs/tools/mcp-server.md
@@ -927,13 +927,13 @@ This is the default transport for running local servers.
 
 ```bash
 # Basic syntax
-gemini mcp add <name> <command> [args...]
+gemini mcp add [options] <name> <command> [args...]
 
 # Example: Adding a local server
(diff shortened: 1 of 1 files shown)


richardkmichael:
cc @jackwotherspoon

jackwotherspoon (approved):

jackwotherspoon:
Thanks @richardkmichael! This is a nice clean up 👍 

Mostly unrelated: Curious if you think real MCP server examples would be better here in this file?

Been thinking or re-organizing the file better and adding in real MCP server examples

richardkmichael:
> Thanks @richardkmichael! This is a nice clean up 👍
> 
> Mostly unrelated: Curious if you think real MCP server examples would be better here in this file?

Yes, I think "cut-and-pasteable" examples are helpful, especially for beginners.

There is maintenance though -- to ensure they're working so as to not leave a user wondering if it's their setup or the example which is broken.  Could be an opportunity to put the gemini testsuite around one or two MCP servers? Chrome devtools or GitHub MCP might be popular candidates.

More unrelated 😄 if you're keen on ✨  UX..  

I considered adding to the help: `-e` is ignored unless `stdio` transport, and `-h` is ignored unless `http`. Both sensible, obviously.  Perhaps even emit a warning, since if attempted, the user might be expecting something which is not possible or making a mistake.

jackwotherspoon:
> More unrelated 😄 if you're keen on ✨ UX..
> 
> I considered adding to the help: `-e` is ignored unless `stdio` transport, and `-h` is ignored unless `http`. Both sensible, obviously. Perhaps even emit a warning, since if attempted, the user might be expecting something which is not possible or making a mistake.

These are great call outs! I am going to revamp the MCP server docs and I'll make this change as well! 👍

mattKorwel (approved):

PR #12315, opened 2025-10-30
Update System Instructions for interactive vs non-interactive mode.

aishaneeshah:
In non-interactive mode, remove all the instructions where model is prompted to interact with the user for additional information or work continuation.

Verified the final system instruction in both modes by diffing final SI as output in GEMINI_WRITE_SYSTEM_MD env variable.
## Summary

In non-interactive mode, model should not be instructed to interact with user. In current implementation, this leads to model exiting stopping too early.

## Pre-Merge Checklist

  - [ ] MacOS
    - [ ] npm run
    - [ ] npx
    - [ ] Docker
    - [ ] Podman
    - [ ] Seatbelt
  - [ ] Windows
    - [ ] npm run
    - [ ] npx
    - [ ] Docker
  - [ ] Linux
    - [ ] npm run
    - [x] npx
    - [ ] Docker


Ibra7997 (approved):

Ibra7997 (approved):

packages/core/src/core/__snapshots__/prompts.test.ts.snap
```diff
@@ -42,10 +42,17 @@ Mandates').
...
   - **3d Games:** HTML/CSS/JavaScript with Three.js.
   - **2d Games:** HTML/CSS/JavaScript.
-3. **User Approval:** Obtain user approval for the proposed plan.
-4. **Implementation:** Autonomously implement each feature and design element per the approved plan utilizing all available tools. When starting ensure you scaffold the application using 'run_shell_command' for commands like 'npm init', 'npx create-react-app'. Aim for full scope completion. Proactively create or source necessary placeholder assets (e.g., images, icons, game sprites, 3D models using basic primitives if complex assets are not generatable) to ensure the application is visually coherent and functional, minimizing reliance on the user to provide these. If the model can generate simple assets (e.g., a uniformly colored square sprite, a simple 3D cube), it should do so. Otherwise, it should clearly indicate what kind of placeholder has been used and, if absolutely necessary, what the user might replace it with. Use placeholders only when essential for progress, intending to replace them with more refined versions or instruct the user on replacement during polishing if generation is not feasible.
```
Ibra7997, line 46: ً

Ibra7997 (approved):

packages/core/src/core/prompts.ts
```diff
@@ -128,10 +130,14 @@ export function getCoreSystemPrompt(
...
 - **Comments:** Add code comments sparingly. Focus on *why* something is done, especially for complex logic, rather than *what* is done. Only add high-value comments if necessary for clarity or if requested by the user. Do not edit comments that are separate from the code you are changing. *NEVER* talk to the user or describe your changes through comments.
 - **Proactiveness:** Fulfill the user's request thoroughly. When adding features or fixing bugs, this includes adding tests to ensure quality. Consider all created files, especially tests, to be permanent artifacts unless the user says otherwise.
-- **Confirm Ambiguity/Expansion:** Do not take significant actions beyond the clear scope of the request without confirming with the user. If asked *how* to do something, explain first, don't just do it.
+- **Handle Ambiguity/Expansion:** Do not take significant actions beyond the clear scope of the request${interactiveMode ? ` without confirming with the user. If asked *how* to do something, explain first, don't just do it` : ''}.
```
hminooei: I suggest moving ${interactiveMode ? to the beginning of this item.

```suggestion
${interactiveMode ? `**Confirm Ambiguity/Expansion:** Do not take significant actions beyond the clear scope of the request without confirming with the user. If asked *how* to do something, explain first, don't just do it.` : ''}
```

packages/core/src/core/prompts.ts
```diff
@@ -113,12 +113,14 @@ export function getCoreSystemPrompt(
...
 
+  const interactiveMode = config.isInteractiveShellEnabled();
+
   let basePrompt: string;
   if (systemMdEnabled) {
     basePrompt = fs.readFileSync(systemMdPath, 'utf8');
   } else {
     const promptConfig = {
-      preamble: `You are an interactive CLI agent specializing in software engineering tasks. Your primary goal is to help users safely and efficiently, adhering strictly to the following instructions and utilizing your available tools.`,
+      preamble: `You are ${interactiveMode ? 'an interactive ' : ''}CLI agent specializing in software engineering tasks. Your primary goal is to help users safely and efficiently, adhering strictly to the following instructions and utilizing your available tools.`,
```
hminooei: Shall we mention it's in non-interactive mode?

`You are ${interactiveMode ? 'an interactive ' : 'a non-interactive '}CLI agent

silviojr (approved):

packages/core/src/core/prompts.ts
```diff
@@ -235,15 +247,16 @@ IT IS CRITICAL TO FOLLOW THESE GUIDELINES TO AVOID EXCESSIVE TOKEN CONSUMPTION.
...
   } else {
-    return `- **Interactive Commands:** Prefer non-interactive commands when it makes sense; however, some commands are only interactive and expect user input during their execution (e.g. ssh, vim). If you choose to execute an interactive command consider letting the user know they can press \`ctrl + f\` to focus into the shell to provide input.`;
+    return `- **Background Processes:** Use background processes (via \`&\`) for commands that are unlikely to stop on their own, e.g. \`node server.js &\`.
+- **Interactive Commands:** Prefer non-interactive commands when it makes sense; however, some commands are only interactive and expect user input during their execution (e.g. ssh, vim). If you choose to execute an interactive command consider letting the user know they can press \`ctrl + f\` to focus into the shell to provide input.`;
```
hminooei: I think this is confusing. It's in non-interactive, and `prefer` can imply it is still in interactive mode?

I suggest changing the wording e.g. `Use non-interactive commands...`

aishaneeshah: I was keeping the original instruction here - removed the `prefer`

packages/core/src/core/prompts.ts
aishaneeshah: Done

packages/core/src/core/prompts.ts
aishaneeshah: Done

packages/core/src/core/prompts.ts
```diff
@@ -127,10 +129,14 @@ export function getCoreSystemPrompt(
...
 - **Comments:** Add code comments sparingly. Focus on *why* something is done, especially for complex logic, rather than *what* is done. Only add high-value comments if necessary for clarity or if requested by the user. Do not edit comments that are separate from the code you are changing. *NEVER* talk to the user or describe your changes through comments.
 - **Proactiveness:** Fulfill the user's request thoroughly. When adding features or fixing bugs, this includes adding tests to ensure quality. Consider all created files, especially tests, to be permanent artifacts unless the user says otherwise.
-- **Confirm Ambiguity/Expansion:** Do not take significant actions beyond the clear scope of the request without confirming with the user. If asked *how* to do something, explain first, don't just do it.
+- ${interactiveMode ? `**Confirm Ambiguity/Expansion:** Do not take significant actions beyond the clear scope of the request without confirming with the user. If asked *how* to do something, explain first, don't just do it.` : `**Handle Ambiguity/Expansion:** Do not take significant actions beyond the clear scope of the request.`}
```
hminooei, line 148: IMO the we could replace `**Handle Ambiguity/Expansion:** Do not take significant actions beyond the clear scope of the request.` with the empty string ''.

packages/core/src/core/prompts.ts
```diff
@@ -224,15 +236,16 @@ IT IS CRITICAL TO FOLLOW THESE GUIDELINES TO AVOID EXCESSIVE TOKEN CONSUMPTION.
...
   } else {
-    return `- **Interactive Commands:** Prefer non-interactive commands when it makes sense; however, some commands are only interactive and expect user input during their execution (e.g. ssh, vim). If you choose to execute an interactive command consider letting the user know they can press \`ctrl + f\` to focus into the shell to provide input.`;
+    return `- **Background Processes:** Use background processes (via \`&\`) for commands that are unlikely to stop on their own, e.g. \`node server.js &\`.
+- **Interactive Commands:** Use non-interactive commands; however, some commands are only interactive and expect user input during their execution (e.g. ssh, vim). If you choose to execute an interactive command consider letting the user know they can press \`ctrl + f\` to focus into the shell to provide input.`;
```
hminooei: My concern is "If you choose to execute an interactive command consider letting the user know they can press \`ctrl + f\` to focus into the shell to provide input.`;" in the non-interactive mode is misleading. I would replace the line 245 with something like: 
```
- **Interactive Commands:** Remember some commands are interactive, so either avoid them and use alternative commands, or pass the necessary flags to avoid them stopping for user input.
```

packages/core/src/core/prompts.ts
aishaneeshah: Keeping this for now - most of this should go away in new SI

hminooei (approved):

Issue #12149, opened 2025-10-28
Cannot update a component (`InputPrompt`) while rendering a different component (`AppContainer`)


jacob314:
### What happened?

I'm occasionally seeing the following issue when running debug builds of Gemini CLI.
We should root cause this issue and fix it.
 ✖  Cannot update a component (`InputPrompt`) while rendering a different component (`AppContainer`). To locate the bad setState() call inside      │
│    `AppContainer`, follow the stack trace as described in https://react.dev/link/setstate-in-render

### What did you expect to happen?

Expect to not see any errors about updates to a component while rendering a different component.

### Client information

<details>
<summary>Client Information</summary>

Run `gemini` to enter the interactive CLI, then run the `/about` command.

```console
> /about
# paste output here
```

</details>

psinha40898:
One instance where this occurs is when you re start the CLI and your first input is an up arrow which fetches a message from history (from your last session)

It looks like this is because the `onPop` handler inside of `useMessageQueue` begins executing before the state update before it is able to finish within React's cycle.

`useMessageQueue` 
~ line 56 
```
  const popAllMessages = useCallback(
    (onPop: (messages: string | undefined) => void) => {
      setMessageQueue((prev) => { // <-- React State Update is scheduled here
        if (prev.length === 0) {
          onPop(undefined);
          return prev;
        }
        const allMessages = prev.join('\n\n');
        onPop(allMessages); // <--- This begins executing before the state update is complete within React
        return []; // Clear the entire queue
      });
    },
    [],
  );
```

This is the code that executes when `onPop(allMessages)` runs in the above block:
`InputPrompt.tsx`
~line 296
```
  const tryLoadQueuedMessages = useCallback(() => {
    if (buffer.text.trim() === '' && popAllMessages) {
      popAllMessages((allMessages) => {
        if (allMessages) {
          buffer.setText(allMessages);
        } else {
          inputHistory.navigateUp();
        }
      });
      return true; 
    }
    return false;
  }, [buffer, popAllMessages, inputHistory]);

```

A simple solution would to be to use the node API 

`queueMicrotask()` instead of calling `onPop` as it currently is.

I didn't open a PR for that because I'm not sure if it is a good path to go down.

jacob314:
[...]

PR #12316, opened 2025-10-30
Code review script to catch common package/cli regressions

jacob314:
## Summary

This custom command is a first pass at distilling feedback I've given on frontend prs to package/cli to catch common react and testing antipatterns. I wrote some of these entries manually and added others by asking Gemini CLI to review feedback I've given on hundreds of Gemini CLI prs.

## Details
Run `/review-frontend <PR_NUMBER>` to review a PR with guidance heavily targeted at catching common issues with frontend prs to Gemini CLI. Goal is for this tool to err on the side of suggesting too many issues rather than too few to help authors and reviewers who are new to writing React code.

## Related Issues
#12149 is an example of an issue that running this tool could have helped prevent.

Try 
`/review-frontend 10392` for an example where this PR catches a real regression that impacted Gemini CLI.
Try some other PRs to make sure the output isn't overly picky on PRs that are fine.

Example output for that PR where this tool would have been helpful:
`/review-frontend 10392`

   Here are my findings that I'd like to discuss before posting the review:

   1.  **CRITICAL: Side Effects in State Updater (`useMessageQueue.ts`)**
       The `popAllMessages` function triggers a side effect (`onPop` callback) *inside* the `setMessageQueue` updater function.

   ```typescript
       setMessageQueue((prev) => {
         if (prev.length === 0) {
           onPop(undefined); // <--- SIDE EFFECT
           return prev;
         }
         const allMessages = prev.join('\n\n');
         onPop(allMessages); // <--- SIDE EFFECT
         return [];
       });
       ```
       **Why this is a problem:** React state updaters must be pure. React may call them multiple times (e.g., in StrictMode, or during concurrent rendering), which could lead to `onPop` being called multiple tim
   or with stale data if interrupted.
       **Recommendation:** Refactor to read state and then update, or use a `useRef` if atomic read-then-clear is strictly required (though standard state usage should suffice here).

   2.  **Unknown `wait()` in Tests (`InputPrompt.test.tsx`)**
       The new tests use `await wait();`. I cannot see the definition of this function in the diff, and it's not a standard import.
       **Recommendation:** Verify what `wait()` does. It should likely be replaced with `waitFor` from `@packages/cli/src/test-utils/async.tsx` to ensure proper `act` wrapping and avoid flakiness.

   3.  **Test Assertions (`QueuedMessageDisplay.test.tsx`)**
       The tests use `expect(output).toContain(...)`.
       **Recommendation:** Per your instructions, we should prefer `toMatchSnapshot()` to verify the entire rendering works as expected, rather than just matching partial content.

   Shall I post this review to the PR, or would you like to investigate the `wait()` function further first?

### DIFF
--- a/.gemini/commands/review-frontend.toml
+++ b/.gemini/commands/review-frontend.toml
@@ -1,0 +1,140 @@
+description="Reviews a pull request based on issue number."
+prompt = """
+Please provide a detailed pull request review on GitHub issue: {{args}}.
+
+Follow these steps:
+
+1. Use `gh pr view {{args}}` to pull the information of the PR.
+2. Use `gh pr diff {{args}}` to view the diff of the PR.
+3. Understand the intent of the PR using the PR description.
+4. If PR description is not detailed enough to understand the intent of the PR,
+   make sure to note it in your review.
+5. Make sure the PR title follows Conventional Commits, here are the last five
+   commits to the repo as examples: !{git log --pretty=format:"%s" -n 5}
+6. Search the codebase if required.
+7. Write a concise review of the PR, keeping in mind to encourage strong code
+   quality and best practices. Pay particular attention to the Gemini MD file
+   in the repo.
+8. Consider ways the code may not be consistent with existing code in the repo.
+   In particular it is critical that the react code uses patterns consistent
+   with existing code in the repo.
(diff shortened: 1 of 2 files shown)


abhipatel12 (approved):
Looks great!

PR #12322, opened 2025-10-30
Handle untrusted folders on extension install and link

### LINKED ISSUE #12303: Ask for folder trust in extension install and link
Current behavior: if the user is in an untrusted folder and tries to install/link an extension, they get an error that the current folder is untrusted and the flow exits.

Modify installation/link commands so the user is instead asked whether they want to trust the current folder. If yes, trust the folder and proceed with the installation/linking. If no, exit immediately.

kevinjwang1:
## Summary

Add handling for when the current folder is untrusted and the install or link command is used. Previously the flow would exit immediately, but now it will ask the user if the current directory should be trusted. If the user says yes, then the current directory will be trusted and installation/linking can continue.

## Related Issues
Fixes #12303


packages/cli/src/config/extension-manager.ts
```diff
@@ -136,11 +140,19 @@ export class ExtensionManager implements ExtensionLoader {
     let extension: GeminiCLIExtension | null;
     try {
       if (!isWorkspaceTrusted(this.settings).isTrusted) {
```
jakemac53, line 142: This seems to be checking if the workspace is trusted but the error is logging the extension source. I don't know which is correct, but I think we should figure that out prior to landing this.

kevinjwang1: Thanks for pointing that out, I modified the error message to not indicate the extension at all, and instead indicate that the current workspace is not trusted.

packages/cli/src/config/extension.test.ts
```diff
@@ -60,13 +60,18 @@ vi.mock('simple-git', () => ({
   }),
 }));
 
-const mockHomedir = vi.hoisted(() => vi.fn(() => '/tmp/mock-home'));
+const tempHomeDir = await vi.hoisted(async () => {
+  const fs = await import('node:fs');
```
jakemac53: Why are we re-importing these here?

packages/cli/src/config/extension.test.ts
```diff
@@ -872,6 +873,89 @@ describe('extension tests', () => {
       fs.rmSync(targetExtDir, { recursive: true, force: true });
     });
 
+    it('should prompt prompt for trust if workspace is not trusted', async () => {
```
jakemac53:
```suggestion
    it('should prompt for trust if workspace is not trusted', async () => {
```

packages/cli/src/config/extension.test.ts
kevinjwang1: This was a bit tricky to get working, the underlying functions to add trust depend on the trustedFolders.USER_SETTINGS_DIR const, which was getting assigned by vi BEFORE we defined the temporary home directory in our beforeEach statements. So instead of getting <sometempdir>/gemini-cli-test-home- as the trusted directory, we were getting '/tmp/mock-home'. 

So I changed the tempHomeDir to use only one mocked home directory for the entire test, and set it up so that it gets assigned before the USER_SETTINGS_DIR const gets defined, so the home dir is constant throughout the tests.

jakemac53 (approved):

PR #12325, opened 2025-10-30
Refactor PolicyEngine to Core Package

### LINKED ISSUE #12166: Refactor policy toml parsing to core package.

After discussion in #11992 @abhipatel12 suggested that we move the toml policy parsing from the cli to the core package.

### Why is this needed?

Keeps dependencies cleaner.

### Additional context

_No response_

allenhutchison:
# PolicyEngine Refactor Walkthrough

## Summary of Changes

The `PolicyEngine` and its configuration logic have been successfully refactored from `packages/cli` to `packages/core`. This makes the policy system reusable for other packages (like the VSCode extension) while maintaining identical functionality for the CLI.

### Key Changes

1.  **New Core Modules**:

    - `packages/core/src/policy/types.ts`: Added `PolicySettings` interface.
    - `packages/core/src/policy/toml-loader.ts`: Ported TOML loading logic.
    - `packages/core/src/policy/config.ts`: Ported policy configuration and precedence logic.

2.  **CLI Updates**:

    - `packages/cli/src/config/policy.ts`: Now acts as a thin wrapper around the core implementation, adapting the CLI's `Settings` object to the core's `PolicySettings`.
    - Deleted redundant files in CLI: `policy-toml-loader.ts`, `policy-toml-loader.test.ts`, and `policy.test.ts`.

3.  **Testing**:
    - Ported all relevant tests to `packages/core/src/policy/*.test.ts`.
    - Verified that all CLI tests still pass using the new core implementation.
    - Full `npm run preflight` passed, ensuring no regressions in linting, building, or testing across the monorepo.

## Verification

The refactor was verified through a series of incremental commits, ensuring a green build at each step:

1.  **Unit Tests**: Core package tests cover all policy loading, merging, and priority logic (including complex scenarios like YOLO mode and MCP server priorities).
2.  **Integration Tests**: Existing CLI tests verified that the CLI still correctly enforces policies when using the core implementation.
3.  **Preflight Check**: The full suite of checks passed successfully.

## Next Steps

The `PolicyEngine` is now ready to be consumed by other packages. No further immediate action is required for this refactor.

Fixes: #12166


packages/cli/src/config/policy.ts
```diff
@@ -6,227 +6,33 @@
...
-  //
-  // TOML policy priorities (before transformation):
-  //   10: Write tools default to ASK_USER (becomes 1.010 in default tier)
-  //   15: Auto-edit tool override (becomes 1.015 in default tier)
-  //   50: Read-only tools (becomes 1.050 in default tier)
-  //   999: YOLO mode allow-all (becomes 1.999 in default tier)
+  // Cast Settings to PolicySettings. TypeScript's structural typing should handle this,
+  // but explicit casting might be needed if Settings has index signatures or other complexities.
+  // Based on our analysis, they are compatible.
+  const policySettings: PolicySettings = settings as unknown as PolicySettings;
```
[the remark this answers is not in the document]
allenhutchison: Fixed.

packages/core/src/policy/config.ts
```diff
@@ -0,0 +1,261 @@
...
+  type UpdatePolicy,
+} from '../confirmation-bus/types.js';
+import { type MessageBus } from '../confirmation-bus/message-bus.js';
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = path.dirname(__filename);
+export const DEFAULT_CORE_POLICIES_DIR = path.join(__dirname, 'policies');
+
+// Store policy loading errors to be displayed after UI is ready
+let storedPolicyErrors: string[] = [];
```
abhipatel12: What if we use `coreEvents` from `events.ts` to send these loading errors to the UI?

The impl has a buffer that we use so that you can send these events out and they will be displayed once the UI is ready.

packages/core/src/policy/config.ts
```diff
@@ -0,0 +1,261 @@
...
+  // Normalize paths for comparison
+  const normalizedDir = path.resolve(dir);
+  const normalizedUser = path.resolve(USER_POLICIES_DIR);
+  const normalizedAdmin = path.resolve(ADMIN_POLICIES_DIR);
+
+  if (
+    defaultPoliciesDir &&
+    normalizedDir === path.resolve(defaultPoliciesDir)
+  ) {
+    return 1;
```
abhipatel12: nit: maybe we can use const values here to make this easier to use without reading the jsdoc (i.e. `DEFAULT_POLICY_TIER`, etc.)

packages/core/src/policy/config.ts
```diff
@@ -0,0 +1,261 @@
...
+}
+
+export async function createPolicyEngineConfig(
+  settings: PolicySettings,
+  approvalMode: ApprovalMode,
+  defaultPoliciesDir?: string,
+): Promise<PolicyEngineConfig> {
+  const policyDirs = getPolicyDirectories(defaultPoliciesDir);
+
+  // Load policies from TOML files
```
abhipatel12, line 116: nit: unneeded comment

packages/core/src/policy/toml-loader.test.ts
```diff
@@ -5,7 +5,7 @@
  */
 
 import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
-import { ApprovalMode, PolicyDecision } from '@google/gemini-cli-core';
+import { ApprovalMode, PolicyDecision } from './types.js';
```
abhipatel12, line 8: This is technically not a refactor issue, but can we use the real filesystem for the tests here? 

We do this using a temp dir in `packages/core/src/utils/memoryDiscovery.test.ts`. 

But totally get that this isn't something related to this PR specifically. If we don't want to make a big PR bigger, maybe we can just file an issue to migrate the tests to use the real fs?

abhipatel12 (commented):
Looks great! Couple of questions. lmk what you think!

packages/core/src/policy/config.ts
allenhutchison: removed.

packages/core/src/policy/config.ts
allenhutchison: Good suggestion. Added throughout.

packages/core/src/policy/config.ts
allenhutchison: Done. In my earlier test my policy errors were emitted before the UI was ready and they were never displayed. I didn't know about this function. Thanks for telling me.

packages/core/src/policy/toml-loader.test.ts
allenhutchison: Will do added #12485 to track this.

allenhutchison:
Thanks for the review. I've resolved your comments. PTAL.

packages/core/src/policy/config.ts
abhipatel12: Np! Hopefully that should work now, and if it doesn't it's a bug on the emitter side!

abhipatel12 (approved):
LGTM! Thanks

NTaylorMullen (approved):

PR #12355, opened 2025-10-31
Remove testing-library/react dep now that it is unused.

jacob314:
## Summary

This dep is no longer used and needs to not be added back. It results in running tests in a jsdom env which doesn't make sense given we are an Ink app running in the terminal not the browser dom.


shishu314 (approved):

mattKorwel (approved):
Lgtm

PR #12413, opened 2025-10-31
Extensions MCP refactor

jakemac53:
## Summary

Refactor extension MCP reloading, put the `ExtensionLoader` in control of everything.

I am working on adding some unit tests, but probably won't get that done until Monday.

## Details

This is the start of a pretty fundamental refactor of how extensions work, specifically how their features are actually loaded into the system.

The goal will be to move _all_ extensions features to be implemented as a part of the `ExtensionLoader`, all in one place. This will make it much more clear how and where to implement new extension features, and also will make extension reloading much safer.

This also makes testing of extension reloading much more possible, and avoids potential race conditions brought about by the previous event based system.

This also fixes a bug with the event based system, where any extensions that were re-loaded would have their MCP servers started regardless of the "allowed" and "blocked" MCP server configuration. The McpClientManager now knows about this configuration and applies it whenever starting MCP servers.

## Related Issues

Part of https://github.com/google-gemini/gemini-cli/issues/11108.

## How to Validate

There should be no user visible change as a result of this PR.

## Pre-Merge Checklist

  - [ ] MacOS
    - [ ] npm run
    - [ ] npx
    - [ ] Docker
    - [ ] Podman
    - [ ] Seatbelt
  - [ ] Windows
    - [ ] npm run
    - [ ] npx
    - [ ] Docker
  - [ ] Linux
    - [x] npm run
    - [ ] npx
    - [ ] Docker


packages/cli/src/ui/components/ConfigInitDisplay.tsx
```diff
@@ -31,9 +31,9 @@ export const ConfigInitDisplay = () => {
       setMessage(`Connecting to MCP servers... (${connected}/${clients.size})`);
     };
 
-    appEvents.on('mcp-client-update', onChange);
+    appEvents.on(AppEvent.McpClientUpdate, onChange);
     return () => {
-      appEvents.off('mcp-client-update', onChange);
+      appEvents.off(AppEvent.McpClientUpdate, onChange);
     };
   }, [config]);
```
jacob314: optional nit: not a new issue but a bit odd that this useEffect is based on config but doesn't use config at all within the effect. Consider refactoring and cleaning this up in the future.

packages/core/src/tools/mcp-client-manager.ts
```diff
@@ -204,12 +252,40 @@ export class McpClientManager {
...
+  /**
+   * Restarts all active MCP Clients.
+   */
+  async restart(): Promise<void> {
+    await Promise.all(
+      Array.from(this.clients.entries()).map(async ([name, client]) => {
+        try {
+          await this.maybeDiscoverMcpServer(name, client.getServerConfig());
+        } catch (error) {
+          console.error(
```
jacob314: don't use console.error. use debugLogger.

packages/core/src/utils/extensionLoader.ts
```diff
@@ -4,45 +4,194 @@
...
+    }
+    this.stoppingCount++;
+    this.eventEmitter?.emit('extensionsStopping', {
+      total: this.stoppingCount,
+      completed: this.stopCompletedCount,
+    });
+
+    try {
+      await this.config.getMcpClientManager()!.stopExtension(extension);
+      // TODO: Remove all extension features here, including at least:
```
jacob314, line 119: can you resolve this todo?

jacob314 (approved):
approved once these minor comments are addressed

packages/core/src/utils/extensionLoader.ts
jakemac53: This will be in a follow-up - this entire feature is guarded behind an experiment right now until these are all addressed.

chrstnb (approved):

PR #12437, opened 2025-11-01
refactor(core): improve error handling for setGlobalProxy

### LINKED ISSUE #11878: Refactor: Error handling for setGlobalProxy in fetch.ts
## Context
We are standardizing our logging and error handling. The `setGlobalProxy` function in `packages/core/src/utils/fetch.ts` currently logs directly to `console.error`, but it is used in two different contexts (startup configuration and tool execution) that require different error handling strategies.

## Task
Refactor the error handling for proxy configuration to be context-aware.

### Recommendation
Our analysis suggests a three-part refactor:

1.  **In `packages/core/src/utils/fetch.ts`:**
    -   Modify `setGlobalProxy` to **throw** the parsing error instead of catching it. This will allow the calling context to handle the failure appropriately.

2.  **In `packages/core/src/config/config.ts` (Startup Context):**
    -   Wrap the call to `setGlobalProxy` in a `try...catch` block.
    -   In the `catch` block, use the event emitter: `coreEvents.emitFeedback('error', 'Failed to set invalid proxy from configuration.', errorObject)`. This correctly surfaces a critical startup error to the user.

3.  **In `packages/core/src/tools/web-fetch.ts` (Tool Context):**
    -   Wrap the call to `setGlobalProxy` in a `try...catch` block.
    -   In the `catch` block, return a failed `ToolResult` object, which will display the error to the user in the chat history.

**Please evaluate this recommendation and course-correct as you see fit.** The goal is to propagate errors correctly and provide the right feedback to the user in each context.

Reference PR #11803 for examples.

Parent Issue: #10761

allenhutchison:
## Summary

Refactored proxy configuration error handling to be context-aware, ensuring errors are properly surfaced to users and eliminating redundant code.

Fixes #11878

### 1. `packages/core/src/utils/fetch.ts`
**Before:**
```typescript
export function setGlobalProxy(proxy: string) {
  try {
    setGlobalDispatcher(new ProxyAgent(proxy));
  } catch (e) {
    console.error(`Failed to set proxy: ${getErrorMessage(e)}`);
  }
}
```

**After:**
```typescript
export function setGlobalProxy(proxy: string) {
  setGlobalDispatcher(new ProxyAgent(proxy));
}
```

- Removed try-catch block
- Function now throws errors instead of catching and logging to console.error
- Allows calling contexts to handle errors appropriately

### 2. `packages/core/src/config/config.ts`
- Added try-catch wrapper around `setGlobalProxy` call in constructor
- Implemented `coreEvents.emitFeedback('error', ...)` for startup error reporting
- Provides user-friendly error message: "Failed to set invalid proxy from configuration."
- Errors now surface correctly to users in both interactive and non-interactive modes

### 3. `packages/core/src/tools/web-fetch.ts`
- Removed duplicate `setGlobalProxy` call from constructor
- Removed `setGlobalProxy` from imports
- Proxy is now only set once in Config constructor (eliminates redundancy)

### 4. `packages/core/src/config/config.test.ts`
- Added comprehensive test suite "Proxy Configuration Error Handling"
- Added mocks for `coreEvents` and `setGlobalProxy`
- 4 new test cases covering:
  - Calling setGlobalProxy when proxy is configured
  - Not calling setGlobalProxy when proxy is not configured
  - Emitting error feedback when setGlobalProxy throws
  - Not emitting error feedback when setGlobalProxy succeeds

## Problem Statement

Previously, `setGlobalProxy` silently logged errors to `console.error`, which meant:
- ❌ Users were not informed when their configured proxy failed
- ❌ Errors bypassed the standardized error reporting system
- ❌ The function was called redundantly in multiple locations (config.ts and web-fetch.ts)
- ❌ No way for calling code to know if proxy setup succeeded

## Solution

This refactor implements the recommendation from issue #11878:

1. **Let errors propagate**: `setGlobalProxy` now throws errors instead of catching them
2. **Context-aware handling**: Config constructor (startup context) catches and reports via `coreEvents.emitFeedback`
3. **Eliminate redundancy**: Removed duplicate call from web-fetch.ts
4. **User-friendly**: Errors surface with clear messages in the chat/terminal

### When proxy configuration fails:
1. User configures invalid proxy in settings
2. Config constructor calls `setGlobalProxy(proxy)`
3. ProxyAgent throws error (invalid URL, etc.)
4. Config's try-catch catches it
5. `coreEvents.emitFeedback('error', ...)` emits feedback
6. User sees error message in UI/terminal: "Failed to set invalid proxy from configuration."
7. Application continues (non-fatal error)

### All tests passing ✅
- **config.test.ts**: 77 tests (including 4 new proxy tests)
- **web-fetch.test.ts**: 32 tests
- **All core tests**: 2,814+ tests passing

### Preflight checks ✅
- ✅ ESLint
- ✅ Prettier
- ✅ TypeScript
- ✅ Build
- ✅ All tests

## Related Issues

- Fixes #11878
- Related to #10761 (parent issue for error handling standardization)
- Follows pattern from #11803 (reference PR for `coreEvents.emitFeedback` usage)

## Checklist

- [x] Code follows existing patterns and conventions
- [x] Comprehensive tests added
- [x] All existing tests pass
- [x] Preflight checks pass
- [x] Error messages are user-friendly
- [x] No breaking changes
- [x] Documentation updated (inline code comments)

### DIFF
--- a/packages/core/src/config/config.ts
+++ b/packages/core/src/config/config.ts
@@ -521,13 +521,14 @@ export class Config {
       initializeTelemetry(this);
     }
 
-    if (this.getProxy()) {
+    const proxy = this.getProxy();
+    if (proxy) {
       try {
-        setGlobalProxy(this.getProxy() as string);
+        setGlobalProxy(proxy);
       } catch (error) {
         coreEvents.emitFeedback(
           'error',
-          'Failed to set invalid proxy from configuration.',
+          'Invalid proxy configuration detected. Check debug drawer for more details (F12)',
           error,
         );
       }
--- a/packages/core/src/config/config.ts
+++ b/packages/core/src/config/config.ts
@@ -76,6 +76,7 @@ import type { UserTierId } from '../code_assist/types.js';
 import { AgentRegistry } from '../agents/registry.js';
 import { setGlobalProxy } from '../utils/fetch.js';
 import { SubagentToolWrapper } from '../agents/subagent-tool-wrapper.js';
+import { coreEvents } from '../utils/events.js';
 
 export enum ApprovalMode {
   DEFAULT = 'default',
@@ -521,7 +522,15 @@ export class Config {
     }
 
     if (this.getProxy()) {
-      setGlobalProxy(this.getProxy() as string);
+      try {
+        setGlobalProxy(this.getProxy() as string);
+      } catch (error) {
+        coreEvents.emitFeedback(
+          'error',
+          'Failed to set invalid proxy from configuration.',
+          error,
+        );
+      }
     }
     this.geminiClient = new GeminiClient(this);
     this.modelRouterService = new ModelRouterService(this);
--- a/packages/core/src/config/config.test.ts
+++ b/packages/core/src/config/config.test.ts
@@ -958,7 +958,7 @@ describe('Server Config (config.ts)', () => {
 
       expect(mockCoreEvents.emitFeedback).toHaveBeenCalledWith(
         'error',
-        'Failed to set invalid proxy from configuration.',
+        'Invalid proxy configuration detected. Check debug drawer for more details (F12)',
         proxyError,
       );
     });
--- a/packages/core/src/config/config.test.ts
+++ b/packages/core/src/config/config.test.ts
@@ -139,6 +139,20 @@ vi.mock('../agents/subagent-tool-wrapper.js', () => ({
   SubagentToolWrapper: vi.fn(),
 }));
 
+const mockCoreEvents = vi.hoisted(() => ({
+  emitFeedback: vi.fn(),
+}));
+
+const mockSetGlobalProxy = vi.hoisted(() => vi.fn());
+
+vi.mock('../utils/events.js', () => ({
+  coreEvents: mockCoreEvents,
+}));
+
+vi.mock('../utils/fetch.js', () => ({
+  setGlobalProxy: mockSetGlobalProxy,
+}));
+
 import { BaseLlmClient } from '../core/baseLlmClient.js';
 import { tokenLimit } from '../core/tokenLimits.js';
 import { uiTelemetryService } from '../telemetry/index.js';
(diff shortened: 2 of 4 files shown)


packages/core/src/config/config.ts
```diff
@@ -521,7 +522,15 @@ export class Config {
...
-      setGlobalProxy(this.getProxy() as string);
+      try {
+        setGlobalProxy(this.getProxy() as string);
+      } catch (error) {
+        coreEvents.emitFeedback(
+          'error',
+          'Failed to set invalid proxy from configuration.',
+          error,
+        );
+      }
```
abhipatel12, line 534: what if we rewrote this piece to be:

```typescript
const proxy = this.getProxy();
if (proxy) {
    try {
        setGlobalProxy(proxy);
    }  catch ...
} 
```

Since the type for `this.proxy` is `string | undefined` we can avoid the `as string` cast here and make this more idiomatic.

packages/core/src/config/config.ts
```diff
@@ -521,7 +522,15 @@ export class Config {
     }
 
     if (this.getProxy()) {
-      setGlobalProxy(this.getProxy() as string);
+      try {
+        setGlobalProxy(this.getProxy() as string);
+      } catch (error) {
+        coreEvents.emitFeedback(
+          'error',
+          'Failed to set invalid proxy from configuration.',
```
abhipatel12: nit: `Invalid proxy configuration detected. Check debug drawer for more details (F12)`

Wdyt? Def don't need to change it, but wonder if people will know how to debug this seeing the message

abhipatel12 (approved):
Just some nits, but lgtm!

packages/core/src/config/config.ts
allenhutchison: Done.

packages/core/src/config/config.ts
allenhutchison: Much better wording. Adopted and updated.

Issue #12447, opened 2025-11-02
OAuth discovery fails with path-based issuers (e.g., Keycloak realms)


cbcoutinho:
## Bug Description

When connecting to OAuth providers that use path-based issuer URLs (e.g., Keycloak with realm paths like `http://localhost:8888/realms/my-realm`), the MCP OAuth client incorrectly strips the path component from the issuer URL, causing metadata discovery to fail.

## Root Cause

**Location:** `packages/core/src/mcp/oauth-provider.ts`, line 686

The code extracts only the protocol and host from the authorization URL:
```typescript
const serverUrl = ${authUrl.protocol}//${authUrl.host};
```

This strips important path information from issuers like:
- Input: `http://localhost:8888/realms/my-realm/protocol/openid-connect/auth`
- Becomes: `http://localhost:8888`
- Should be: `http://localhost:8888/realms/my-realm`

## Impact

This breaks OAuth discovery for:
- Keycloak (uses `/realms/{realm-name}` paths)
- Azure AD (uses `/{tenant-id}` paths)
- Other providers with path-based issuer URLs

## Related Issue: Invalid DCR Field

**Location:** `packages/core/src/mcp/oauth-provider.ts`, lines 59-66, 122-129

The Dynamic Client Registration (DCR) request includes a non-standard field `code_challenge_method` which causes registration failures with RFC 7591-compliant servers:

```typescript
const registrationRequest: OAuthClientRegistrationRequest = {
  // ...
  code_challenge_method: ['S256'],  // ❌ Not a valid RFC 7591 field
  // ...
};
```

**Error from Keycloak:**
```
{"error":"Invalid json representation for OIDCClientRepresentation. Unrecognized field 'code_challenge_method' at line 1 column 237."}
```

PKCE parameters belong in the authorization request (line 301), not the registration request.

## Reproduction

1. Set up a Keycloak instance with a realm (e.g., `my-realm`)
2. Configure an MCP server with OAuth at `http://localhost:8888/realms/my-realm`
3. Attempt to connect with gemini-cli
4. Observe "Metadata discovery failed" error

## Expected Behavior

The client should:
1. Preserve path components when deriving the issuer URL from authorization endpoints
2. Only strip protocol-specific paths like `/protocol/openid-connect/auth`, `/oauth2/authorize`, etc.
3. Omit non-standard fields from DCR requests

## Environment

- gemini-cli version: current main branch
- OAuth Provider: Keycloak 26.4.2
[...]

PR #12448, opened 2025-11-02
fix: preserve path components in OAuth issuer URLs

cbcoutinho:
## Summary

Fixes #10512 and #12447

This PR addresses two OAuth integration bugs that prevent gemini-cli from connecting to OAuth providers with path-based issuer URLs (e.g., Keycloak, Azure AD):

### 1. Path-based issuer URL handling
**Problem:** The client was stripping all path components from authorization URLs, breaking discovery for providers that use paths in their issuer URLs.

**Example:**
- Input: `http://localhost:8888/realms/my-realm/protocol/openid-connect/auth`
- Was: `http://localhost:8888` ❌
- Now: `http://localhost:8888/realms/my-realm` ✅

**Fix:** Intelligently remove only OIDC protocol-specific suffixes (`/protocol/openid-connect/auth`, `/oauth2/authorize`, etc.) while preserving realm/tenant paths.

### 2. Invalid DCR field
**Problem:** Dynamic Client Registration requests included `code_challenge_method`, which is not part of RFC 7591 and causes failures with compliant OAuth servers.

**Fix:** Removed the non-standard field. PKCE parameters belong in the authorization request (already correct), not the registration request.

## Testing

Tested against:
- Keycloak 26.4.2 with realm path `/realms/my-realm`
- Successfully completed OAuth discovery and client registration

## Impact

This enables gemini-cli to work with:
- Keycloak (uses `/realms/{realm-name}` paths)
- Azure AD (uses `/{tenant-id}` paths)  
- Other OAuth providers with path-based issuer URLs

## References

- RFC 8414 (OAuth 2.0 Authorization Server Metadata)
- RFC 7591 (OAuth 2.0 Dynamic Client Registration Protocol)

### DIFF
--- a/packages/core/src/mcp/oauth-provider.test.ts
+++ b/packages/core/src/mcp/oauth-provider.test.ts
@@ -44,9 +44,10 @@ import type {
 import { MCPOAuthProvider } from './oauth-provider.js';
 import type { OAuthToken } from './token-storage/types.js';
 import { MCPOAuthTokenStorage } from './oauth-token-storage.js';
-import type {
-  OAuthAuthorizationServerMetadata,
-  OAuthProtectedResourceMetadata,
+import {
+  OAuthUtils,
+  type OAuthAuthorizationServerMetadata,
+  type OAuthProtectedResourceMetadata,
 } from './oauth-utils.js';
 import { coreEvents } from '../utils/events.js';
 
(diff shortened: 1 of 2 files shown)


cornmander:
Noticed a regression when providers include version segments in their authorization endpoints. For example, Okta exposes https://example.okta.com/oauth2/default/v1/authorize but publishes discovery metadata at https://example.okta.com/oauth2/default/.well-known/openid-configuration. Trimming only the /authorize suffix leaves /oauth2/default/v1, so our metadata probe looks under /oauth2/default/v1/.well-known/... and dynamic registration fails. We should try multiple issuer candidates (origin, trimmed path, and the path with a trailing version segment removed) before giving up; otherwise Okta and similar deployments break. Adding that fallback also lets us keep the Keycloak path-preservation behavior that motivated the PR.

cornmander (approved):

Issue #12449, opened 2025-11-02
Add dedicated response text semantic color


cornmander:
## Summary
Add a `text.response` semantic color so themes can opt into a dedicated response color without a standalone setting.

## Details
- Originally explored in #12275.
- Response rendering in markdown, inline text, lists, and tables now falls back to `theme.text.response` (defaulting to `text.primary`).
- `RenderInline` supports a `defaultColor` override so existing callsites can keep their colors.
- Docs call out the new theme key for custom themes.

## Open Questions
- Do we want any of the built-in themes to ship with a custom response color by default?

## Related PRs
- Replacement implementation: #12450 (draft)

PR #12450, opened 2025-11-02
feat(ui): add response semantic color

cornmander:
## Summary
- replace the `/ResponseColor` command from #12275 with a theme-driven `text.response` semantic color (tracked in #12449)
- flow the new response color through markdown rendering while allowing other callers to keep their explicit colors via `RenderInline`'s `defaultColor`
- document the custom theme key so users can opt in via settings instead of a dedicated setting/command

## Testing
- npm run test --workspace @google/gemini-cli -- MarkdownDisplay.test.tsx


packages/cli/src/ui/components/messages/InfoMessage.tsx
```diff
@@ -24,7 +24,7 @@ export const InfoMessage: React.FC<InfoMessageProps> = ({ text }) => {
       </Box>
       <Box flexGrow={1}>
         <Text wrap="wrap" color={theme.status.warning}>
-          <RenderInline text={text} />
+          <RenderInline text={text} defaultColor={theme.status.warning} />
```
jacob314, line 27: remove the theme.status.warning color on the line before as it is redundant. I suspect the entire outer Text object might be redundant but you should double check.

packages/cli/src/ui/components/messages/ToolConfirmationMessage.tsx
```diff
@@ -250,7 +250,10 @@ export const ToolConfirmationMessage: React.FC<
     bodyContent = (
       <Box flexDirection="column" paddingX={1} marginLeft={1}>
         <Text color={theme.text.link}>
```
jacob314: remove the outer text object as it isn't needed.

packages/cli/src/ui/components/messages/WarningMessage.tsx
```diff
@@ -24,7 +24,7 @@ export const WarningMessage: React.FC<WarningMessageProps> = ({ text }) => {
       </Box>
       <Box flexGrow={1}>
         <Text wrap="wrap" color={Colors.AccentYellow}>
```
jacob314: while you are here correct this to be semantic. same issue as above

jacob314 (commented):
be very careful that this change hasn't regressed colors as our snapshot tests are b&w so can't catch color changes.

cornmander:
Confirmed things look fine: 
image

packages/cli/src/ui/components/messages/InfoMessage.tsx
cornmander: done

packages/cli/src/ui/components/messages/ToolConfirmationMessage.tsx
cornmander: done

packages/cli/src/ui/components/messages/WarningMessage.tsx
cornmander: done

jacob314 (approved):
lgtm

PR #12471, opened 2025-11-03
alternate buffer support

jacob314:
# TLDR

Support detecting mouse events (scroll, click, and drag).
Add alternate buffer mode leveraging new support in our Ink branch for efficient, flicker free, and robust rendering when in alternate buffer mode.

Gemini CLI rendering is not impacted outside of alternate buffer mode in this PR.

Copy mode (Ctrl+S.. only when in alternate buffer mode).
This is a small escape hatch mainly for development on the CLI to enable selecting and copying from the CLI while in alternate buffer mode.

## How to Validate

Enable alternate buffer mode.
`/settings
Toggle
"Use Alternate Screen Buffer" to true
Validate you can click within the InputPrompt to change the cursor position.
This is a minor feature but trivial to implement with the foundation provided by mouse support.
https://screencast.googleplex.com/cast/NTA5NTU4MzI2NTA2MjkxMnw1ZTUzN2QyYi03Nw

Alternate buffer mode works with this PR but is very minimal. The useful changes are the follow up PR to actually support scrolling with the scroll  wheel

# Testing
Verify that noting bad happens when outside of alternate buffer mode. Make sure we didn't break any keyboard behavior both in and outside of alternate buffer mode.
For example, make sure escape, F12, and other tricky to detect keys work.


packages/cli/src/ui/AppContainer.tsx
```diff
@@ -248,10 +262,12 @@ export const AppContainer = (props: AppContainerProps) => {
       setConfigInitialized(true);
     })();
     registerCleanup(async () => {
+      // Turn off mouse scroll.
+      stdout.write('\u001b[?1006l\u001b[?1002l');
       const ideClient = await IdeClient.getInstance();
       await ideClient.disconnect();
     });
-  }, [config]);
+  }, [config, stdout]);
```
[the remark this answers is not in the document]
jacob314: fixed

packages/cli/src/config/keyBindings.ts
```diff
@@ -160,6 +161,7 @@ export const defaultKeyBindings: KeyBindingConfig = {
   [Command.SHOW_FULL_TODOS]: [{ key: 't', ctrl: true }],
   [Command.TOGGLE_IDE_CONTEXT_DETAIL]: [{ key: 'g', ctrl: true }],
   [Command.TOGGLE_MARKDOWN]: [{ key: 'm', command: true }],
+  [Command.TOGGLE_COPY_MODE]: [{ key: 's', ctrl: true }],
```
scidomino, line 164: We should document this in keyboard-options.md

packages/cli/src/ui/utils/mouse.ts
```diff
@@ -0,0 +1,214 @@
...
+    const row = parseInt(match[3], 10);
+    const action = match[4];
+    const isRelease = action === 'm';
+
+    const shift = (buttonCode & 4) !== 0;
+    const meta = (buttonCode & 8) !== 0;
+    const ctrl = (buttonCode & 16) !== 0;
+    const isMove = (buttonCode & 32) !== 0;
+
+    let name: MouseEventName | null = null;
```
scidomino: nit: Consider introducing a getMouseEventName() helper fn

scidomino (approved):

packages/cli/src/ui/utils/mouse.ts
jacob314: good idea. done

packages/cli/src/config/keyBindings.ts
jacob314: Done. We should as a future cleanup generate keyboard-options.md automatically from keyBindings.ts.  @cornmander recently landed some nice changes to generate the settings documentation automatically.

PR #12495, opened 2025-11-03
feat(core): Add support for listing experiments

skeshive:
feat(core): Add support for listing experiments

## TLDR

This pull request introduces the client-side infrastructure for fetching server-side experiments and feature flags. It adds a `listExperiments` method to the `CodeAssistServer` to communicate with the backend. This change allows for dynamic feature management based on client metadata.


cornmander (approved):

gsquared94 (approved):
We should send the set of experimentIds in the Clearcut telemetry. There is a dedicated field there I think for this.

skeshive:
> We should send the set of experimentIds in the Clearcut telemetry. There is a dedicated field there I think for this.

Let's file a separate issue for this! I'd like to get the core changes in first

PR #12502, opened 2025-11-03
Modify navigation and completion keyboard shortcuts to not use scroll.

jacob314:
This leaves room for using shift+arrows for terminal scroll. Require shift is not pressed when using up and down.

- Add physical 'Home' and 'End' keys to standard key bindings.

- Update `SettingsDialog` and `useSelectionList` to use standard `keyMatchers` for navigation and actions.

- Ensure `Command.NAVIGATION_UP` and `Command.NAVIGATION_DOWN` ignore the shift key, preventing accidental selection while navigating.

- Add tests to verify new key binding configurations.

## How to Validate

Verify that shift-arrow no longer triggers navigation so that it is free to be used for scrolling.

### DIFF
--- a/packages/cli/src/config/keyBindings.test.ts
+++ b/packages/cli/src/config/keyBindings.test.ts
@@ -55,5 +55,27 @@ describe('keyBindings config', () => {
       const config: KeyBindingConfig = defaultKeyBindings;
       expect(config[Command.HOME]).toBeDefined();
     });
+
+    it('should have correct specific bindings', () => {
+      // Verify navigation ignores shift
+      const navUp = defaultKeyBindings[Command.NAVIGATION_UP];
+      expect(navUp).toContainEqual({ key: 'up', shift: false });
+
+      const navDown = defaultKeyBindings[Command.NAVIGATION_DOWN];
+      expect(navDown).toContainEqual({ key: 'down', shift: false });
+
+      // Verify dialog navigation
+      const dialogNavUp = defaultKeyBindings[Command.DIALOG_NAVIGATION_UP];
+      expect(dialogNavUp).toContainEqual({ key: 'up', shift: false });
+      expect(dialogNavUp).toContainEqual({ key: 'k', shift: false });
+
+      const dialogNavDown = defaultKeyBindings[Command.DIALOG_NAVIGATION_DOWN];
+      expect(dialogNavDown).toContainEqual({ key: 'down', shift: false });
+      expect(dialogNavDown).toContainEqual({ key: 'j', shift: false });
+
+      // Verify physical home/end keys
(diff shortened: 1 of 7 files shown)


packages/cli/src/config/keyBindings.ts
```diff
@@ -100,8 +104,8 @@ export const defaultKeyBindings: KeyBindingConfig = {
   [Command.ESCAPE]: [{ key: 'escape' }],
 
   // Cursor movement
-  [Command.HOME]: [{ key: 'a', ctrl: true }],
-  [Command.END]: [{ key: 'e', ctrl: true }],
+  [Command.HOME]: [{ key: 'a', ctrl: true }, { key: 'home' }],
```
jacob314, line 107: think it was just a bug that we hadn't included home and end here so tossed them in. can revert as it is unrelated to the PR but seems harmless and net positive key binding improvement.

packages/cli/src/config/keyBindings.ts
```diff
@@ -118,15 +122,33 @@ export const defaultKeyBindings: KeyBindingConfig = {
...
+  [Command.HISTORY_DOWN]: [{ key: 'n', ctrl: true, shift: false }],
+  [Command.NAVIGATION_UP]: [{ key: 'up', shift: false }],
+  [Command.NAVIGATION_DOWN]: [{ key: 'down', shift: false }],
+
+  // Dialog navigation
+  // Navigation shortcuts appropriate for dialogs where we do not need to accept
+  // text input.
+  [Command.DIALOG_NAVIGATION_UP]: [
+    { key: 'up', shift: false },
+    { key: 'k', shift: false },
```
jacob314, line 134: regular navigation up and down can't support the k and j bindings because k and j should trigger normal insertions into the input prompt

packages/cli/src/config/keyBindings.test.ts
```diff
@@ -55,5 +55,20 @@ describe('keyBindings config', () => {
...
+      expect(navUp).toContainEqual({ key: 'k', shift: false });
+
+      const navDown = defaultKeyBindings[Command.NAVIGATION_DOWN];
+      expect(navDown).toContainEqual({ key: 'down', shift: false });
+      expect(navDown).toContainEqual({ key: 'j', shift: false });
+
+      // Verify physical home/end keys
+      expect(defaultKeyBindings[Command.HOME]).toContainEqual({ key: 'home' });
+      expect(defaultKeyBindings[Command.END]).toContainEqual({ key: 'end' });
+    });
```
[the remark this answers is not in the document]
jacob314: Fixed

scidomino (approved):

Issue #11995, opened 2025-10-24
Add out of process safety checkers to policies


allenhutchison:
Policies should be configured to have an out of process safety checker that runs as part of policy evaluation. We should provide a set of default checkers, but the expectation is that users and admins can also provide checkers.

allenhutchison:
### Safety Checker Implementation Plan

This feature introduces a new extensible safety checker system to the policy engine, allowing for more granular, configurable validation of tool calls.

#### Core Concepts

The design is based on the following principles:

1.  **Process Isolation:** Checkers will run as standalone processes, communicating with the main Gemini CLI over `stdin`/`stdout`. This prevents shared memory and enhances security.
2.  **Language-Agnostic Protocol:** Communication will use a versioned JSON protocol, allowing the community to write custom checkers in any programming language.
3.  **Extensible Context:** The protocol is designed to pass rich, structured context to checkers, including environment details (`cwd`), non-sensitive config variables, and conversation history. It's designed to be easily extended without breaking existing checkers.
4.  **TOML Configuration:** Checkers are enabled and configured within the `policy.toml` file under a new `[rule.safety_checker]` table.

#### Implementation Overview

The work will be broken down into four phases:

1.  **Phase 1: Protocol & Config Definition:** We will define the core data structures for the protocol and update the policy rule interface in `packages/core`.
2.  **Phase 2: Implement `allowed-path` Checker:** We will create the first built-in checker in `packages/cli`. This checker, `allowed-path`, will verify that file-system-related tool calls operate only within the user's workspace.
3.  **Phase 3: Engine Integration:** The policy engine in `packages/cli` will be updated to spawn checker processes, send them the relevant tool call and context data, and enforce the `allow`/`deny` decision they return.
4.  **Phase 4: Testing:** We will add comprehensive unit tests for the new checker and end-to-end integration tests to validate the entire workflow.

A key design decision is a **hybrid data-passing strategy**. By default, the CLI will send all available context to checkers for simplicity. However, users can optionally add a `required_context` key to their configuration to ensure the CLI only sends the specific data a checker needs, providing a path for performance optimization.

allenhutchison:
[safety-checker-design.md](https://github.com/user-attachments/files/23176418/safety-checker-design.md)

PR #12504, opened 2025-11-03
feat(safety): Introduce safety checker framework

allenhutchison:
This commit introduces a new safety checker framework that provides an additional layer of validation for tool calls. This framework allows for more granular and dynamic checks to be performed even after a tool call has been allowed by the main policy engine.

Key changes:

- Safety Checker Framework: A new concept of "safety checkers" is introduced. These are functions that can perform additional validation on a tool call before it is executed. Checkers can be in-process (like the new allowed-path checker) or external.

- Policy Configuration: Safety checkers are now configured as top-level items in TOML policy files using the [[safety_checker]] table. This allows checkers to be defined independently of specific policy rules, with their own matching criteria (tool name, arguments pattern) and priority.

- Integration with PolicyEngine: The PolicyEngine is now responsible for invoking the CheckerRunner. When PolicyEngine.check() is called, it first evaluates standard policy rules. If the decision is not DENY, it then executes any matching safety checkers. If a checker denies the tool call, the final decision becomes DENY.

- PolicyEngine API Change: The PolicyEngine.check() method now returns an object containing both the decision and the rule that was matched. This provides better visibility into why a decision was made.

- Initial allowed-path Checker: A new built-in allowed-path checker has been added. This ensures that file system tools can only write to paths within the project's working directory, preventing them from modifying files outside of the project.

This new framework provides a powerful mechanism for adding more complex and context-aware safety checks, enhancing the overall security and reliability of the Gemini CLI.

Part of #11995


packages/core/src/safety/context-builder.ts
```diff
@@ -0,0 +1,88 @@
...
+        (sensitive) => keyLower.includes(sensitive.toLowerCase()),
+      );
+
+      if (!isSensitive && typeof value !== 'function') {
+        filtered[key] = value;
+      }
+    }
+
+    return filtered;
+  }
```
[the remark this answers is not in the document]
allenhutchison: agreed

packages/core/src/safety/built-in.ts
```diff
@@ -0,0 +1,116 @@
...
+      if (!isAllowed) {
+        return {
+          allowed: false,
+          reason: `Path "${p}" is outside of the allowed workspace directories.`,
+        };
+      }
+    }
+
+    return { allowed: true };
+  }
```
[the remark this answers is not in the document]
allenhutchison: ok, will do this.

packages/core/src/safety/checker-runner.ts
```diff
@@ -0,0 +1,272 @@
...
+   * Executes an external checker process and handles its lifecycle.
+   */
+  private executeCheckerProcess(
+    checkerPath: string,
+    input: SafetyCheckInput,
+  ): Promise<SafetyCheckResult> {
+    return new Promise((resolve) => {
+      const child = spawn('node', [checkerPath], {
+        stdio: ['pipe', 'pipe', 'pipe'],
+      });
```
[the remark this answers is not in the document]
allenhutchison: agree this was an oversight.

packages/core/src/policy/toml-loader.ts
```diff
@@ -37,6 +42,14 @@ const PolicyRuleSchema = z.object({
...
   modes: z.array(z.string()).optional(),
+  safety_checker: z
+    .object({
+      type: z.enum(['in-process', 'external']),
+      name: z.string(),
+      required_context: z.array(z.string()).optional(),
+      config: z.record(z.unknown()).optional(),
+    })
+    .optional(),
 });
```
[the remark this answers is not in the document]
allenhutchison: great suggestion

packages/core/src/confirmation-bus/message-bus.ts
```diff
@@ -50,7 +52,49 @@ export class MessageBus extends EventEmitter {
...
 
       if (message.type === MessageBusType.TOOL_CONFIRMATION_REQUEST) {
-        const decision = this.policyEngine.check(message.toolCall);
+        const { decision, rule } = this.policyEngine.check(message.toolCall);
+
+        if (
+          decision === PolicyDecision.ALLOW &&
+          rule?.safety_checker &&
+          this.checkerRunner
+        ) {
```
avilladsen: It feels odd to me to externalize the running of the safety checker out of the policy engine and into the message bus. Safety checkers are part of policy, not part of the messaging system.

packages/core/src/confirmation-bus/message-bus.ts
```diff
@@ -50,7 +52,49 @@ export class MessageBus extends EventEmitter {
...
+          // If allowed by policy but has a safety checker, run it
+          try {
+            const result = await this.checkerRunner.runChecker(
+              message.toolCall,
+              rule.safety_checker,
+            );
```
avilladsen: How are multiple, orthogonal safety checks combined? For example, suppose I want an "allowed paths" check, and a "does this tool call seem relevant to the user prompt?" check. I'd want to deny the tool call if either check does not allow it.

Currently, it looks like the behavior of safety checks is to select the checker with highest priority, which then outright denies or outright allows the tool call. It feels like a better behavior might be for each checker to outright deny the tool call, or let policy rule checking proceed to the next highest priority rule (or maybe let safety checkers issue a ternary allow/deny/continue decision).

packages/core/src/confirmation-bus/message-bus.ts
```diff
@@ -50,7 +52,49 @@ export class MessageBus extends EventEmitter {
       }
 
       if (message.type === MessageBusType.TOOL_CONFIRMATION_REQUEST) {
-        const decision = this.policyEngine.check(message.toolCall);
+        const { decision, rule } = this.policyEngine.check(message.toolCall);
+
+        if (
+          decision === PolicyDecision.ALLOW &&
```
avilladsen: Why does a "delegate to this safety checker" PolicyDecision have a `decision` field at all? The decision is not yet made.

packages/core/src/safety/protocol.ts
```diff
@@ -0,0 +1,83 @@
...
+}
+
+/**
+ * The data structure returned by a safety checker process via stdout.
+ */
+export interface SafetyCheckResult {
+  /**
+   * Whether the tool call is allowed to proceed.
+   */
+  allowed: boolean;
```
avilladsen: It would be more future proof to have a (string based) enum, rather than just a boolean decision. I can currently think of four decisions that a safety checker might want to return:
* allow
* deny
* ask user
* continue checking policy rules (see my comment on message-bus.ts)

packages/core/src/confirmation-bus/message-bus.ts
```diff
@@ -50,7 +52,49 @@ export class MessageBus extends EventEmitter {
...
+              // Checker denied it, treat as policy rejection
+              this.emitMessage({
+                type: MessageBusType.TOOL_POLICY_REJECTION,
+                toolCall: message.toolCall,
+              });
+              this.emitMessage({
+                type: MessageBusType.TOOL_CONFIRMATION_RESPONSE,
+                correlationId: message.correlationId,
+                confirmed: false,
+              });
```
avilladsen: This block of "what to do for a policy rejection" is repeated below in the catch block, and again in the existing decision switch block. It could be extracted out into a utility method to avoid the redundancy.

packages/core/src/confirmation-bus/message-bus.ts
```diff
@@ -50,7 +52,49 @@ export class MessageBus extends EventEmitter {
...
+              return;
+            }
+            // If allowed by checker, proceed to standard ALLOW handling below
+          } catch (error) {
```
avilladsen: It looks like CheckerRunner is already written to catch all errors and return an "deny" result. I'm torn on whether this is "belt and suspenders" defensive coding (good) or plain redundancy (bad). To be clear, I think this try/catch here is a must, in case of truly unexpected exceptions: the question is whether CheckerRunner.runChecker should let exceptions bubble through so that they're always handled here, rather than this being dead code.

Also, there should definitely be a test in which CheckerRunner.runChecker throws, so that this isn't completely dead code.

packages/core/src/confirmation-bus/message-bus.test.ts
avilladsen, line 1: Should have new tests covering safety checkers.

packages/core/src/config/config.ts
```diff
@@ -522,7 +524,16 @@ export class Config {
     this.fileExclusions = new FileExclusions(this);
     this.eventEmitter = params.eventEmitter;
     this.policyEngine = new PolicyEngine(params.policyEngineConfig);
-    this.messageBus = new MessageBus(this.policyEngine, this.debugMode);
+    const contextBuilder = new ContextBuilder(this);
+    const checkerRunner = new CheckerRunner(contextBuilder, {
+      // TODO: Determine the correct path for external checkers
```
avilladsen: Adding to the TODO: maybe "paths," plural? That is, do we want to support both system-level and user-level safety checkers, as we do with policy rules?

packages/core/src/safety/built-in.ts
```diff
@@ -0,0 +1,119 @@
...
+    const config = (toolCall as any).config as AllowedPathConfig | undefined;
+
+    // Build list of allowed directories
+    const allowedDirs = [
+      context.environment.cwd,
+      ...context.environment.workspaces,
+      ...(config?.additional_allowed_paths ?? []),
+    ];
+
+    // Find all arguments that look like paths
```
avilladsen, line 34: Anticipating a feature request: an allowed path PolicyRule should be able to explicitly specify the argument keys that should be checked (and since there is a default behavior, also specify argument keys to be excluded from checking).

packages/core/src/safety/built-in.ts
```diff
@@ -0,0 +1,119 @@
...
+    const pathsToCheck: string[] = [];
+    if (toolCall.args) {
+      for (const key in toolCall.args) {
```
avilladsen: What about nested keys?

packages/core/src/safety/built-in.ts
```diff
@@ -0,0 +1,119 @@
...
+          context.environment.cwd,
+        );
+        if (!resolvedDir) return false;
+        return this.isPathAllowed(resolvedPath, resolvedDir);
+      });
+
+      if (!isAllowed) {
+        return {
+          allowed: false,
+          reason: `Path "${p}" is outside of the allowed workspace directories.`,
```
avilladsen: It might also be helpful to include the name of the argument in the reason.

packages/core/src/safety/checker-runner.ts
```diff
@@ -0,0 +1,272 @@
...
+      let timeoutHandle: NodeJS.Timeout | null = null;
+      let killed = false;
+
+      // Set up timeout
+      timeoutHandle = setTimeout(() => {
+        killed = true;
+        child.kill('SIGTERM');
+        resolve({
+          allowed: false,
+          reason: `Safety checker timed out after ${this.timeout}ms`,
```
avilladsen: It would be helpful to include the name of the checker in this message. Same goes for the other following error reasons.

packages/core/src/policy/types.ts
```diff
@@ -16,6 +18,48 @@ export enum ApprovalMode {
...
+  config?: unknown;
+  required_context?: Array<keyof SafetyCheckInput['context']>;
+}
+
+/**
+ * Base interface for in-process checkers.
+ */
+export interface InProcessCheckerConfig {
+  type: 'in-process';
+  name: 'allowed-path'; // For now, only 'allowed-path' is in-process
```
avilladsen: I would go ahead and make an enum or union type for InProcessorCheckerType even though it would only have one alternative for now. There's already a few places it could be used, such as in the type of `BUILT_IN_IN_PROCESS_CHECKERS` in `../safety/registry.ts`, and of course the Zod validator for this field.

packages/core/src/safety/context-builder.ts
```diff
@@ -0,0 +1,110 @@
...
+
+import type { SafetyCheckInput, ConversationTurn } from './protocol.js';
+import type { Config } from '../config/config.js';
+
+/**
+ * Builds context objects for safety checkers, ensuring sensitive data is filtered.
+ */
+export class ContextBuilder {
+  // List of config keys that should never be sent to checkers
+  private static readonly SENSITIVE_CONFIG_KEYS = [
```
avilladsen: Having a blocklist of sensitive keys, rather than an allowlist of safe keys, feels quite dangerous. I don't expect people to remember to update this list when adding to the Config object.

packages/core/src/policy/types.ts
```diff
@@ -16,6 +18,48 @@ export enum ApprovalMode {
...
 }
 
+/**
+ * Configuration for the built-in allowed-path checker.
+ */
+export interface AllowedPathConfig {
+  /**
+   * Additional directories to allow beyond the default CWD and workspaces.
+   */
+  additional_allowed_paths?: string[];
```
avilladsen: If multiple `allowed-path` PolicyRules would, ignoring priority, apply to a tool call, should they be merged somehow? The `allowed-path` checker is (I think) unusual for a safety checkers in that it *broadens* the set of allowed calls, rather than being an additional check that narrows the set of allowed calls.

packages/core/src/safety/built-in.ts
```diff
@@ -0,0 +1,119 @@
...
+      if (!resolvedPath) {
+        // If path cannot be resolved, deny it
+        return {
+          allowed: false,
+          reason: `Cannot resolve path "${p}"`,
+        };
+      }
+
+      const isAllowed = allowedDirs.some((dir) => {
+        // Also resolve allowed directories to handle symlinks
```
avilladsen, line 57: Here and elsewhere, shouldn't this respect the `follow_symlinks` config?

packages/core/src/safety/protocol.ts
```diff
@@ -0,0 +1,83 @@
...
+    history?: {
+      turns: ConversationTurn[];
+    };
+  };
+}
+
+/**
+ * The data structure returned by a safety checker process via stdout.
+ */
+export interface SafetyCheckResult {
```
abhipatel12: We could use a discriminated union to enforce the inclusion of a `reason` when this value is false.

packages/core/src/safety/built-in.test.ts
```diff
@@ -0,0 +1,167 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+import * as fs from 'node:fs';
```
abhipatel12: Since we are testing some relatively critical logic here, can we use the actual fs to ensure we're covered on every OS when these tests are run in CI. 

We can use the same pattern found in `memoryDiscovery.test.ts` to create real files, dirs, symlinks, etc.

packages/core/src/safety/built-in.test.ts
```diff
@@ -0,0 +1,167 @@
...
+  });
+
+  it('should handle non-existent paths gracefully if they are inside allowed dir', async () => {
+    // realpathSync throws for non-existent files usually, but we mocked it to return the path.
+    // In a real scenario, if the file doesn't exist, we might want to check its parent directory.
+    // The current implementation uses realpathSync which might fail for new files.
+    // Let's assume for now we are checking existing paths or the implementation handles it.
+    // Re-reading built-in.ts might be needed if this test fails in reality.
+    // For now, trusting the mock.
+    // For now, trusting the mock.
```
abhipatel12: nit: clean up comment/dupe comment as needed.

packages/core/src/confirmation-bus/message-bus.ts
abhipatel12: we could add a new type: `requires_safety_check`

abhipatel12 (commented):
Added some thoughts! Lmk what you think

packages/core/src/safety/built-in.test.ts
allenhutchison: Fixed.

packages/core/src/safety/built-in.test.ts
allenhutchison: Done.

packages/core/src/safety/protocol.ts
allenhutchison: ok, good point. I had originally envisioned this as pass / fail but we can try a few options. I've added all four for now, but I'm not yet convinced that continue makes sense for a safety checker. It might be better positioned as part of the config for multiple checkers.

packages/core/src/safety/checker-runner.ts
allenhutchison: You're right. Done.

packages/core/src/confirmation-bus/message-bus.ts
allenhutchison: I've refactored the saftey checker to be run by the policy engine.

packages/core/src/safety/built-in.ts
allenhutchison: That was there in a previous version of the PR but I opted to remove the follow_symlink option and always deal with fully resolved paths. I failed to remove the option from the interface, which I've fixed now.

packages/core/src/confirmation-bus/message-bus.ts
allenhutchison: 🤦  I'm going add a new config pragma [[safety-checker]] that uses the same syntax as [[rules]] but will create a union of checkers and run them in priority order.

packages/core/src/safety/protocol.ts
allenhutchison: Done.

packages/core/src/confirmation-bus/message-bus.ts
allenhutchison: I've refactored saftey checks into the policy engine.

packages/core/src/confirmation-bus/message-bus.test.ts
allenhutchison: Agree. Better test coverage now in both policy-engine-test.ts and in the safety checker tests.

packages/core/src/safety/built-in.ts
allenhutchison: Good suggestion. Done.

packages/core/src/safety/built-in.ts
allenhutchison: Done and tests added.

packages/core/src/safety/built-in.ts
allenhutchison: Done.

packages/core/src/safety/context-builder.ts
allenhutchison: Yeah, this is more proof of concept than anything else. Do you have a better suggestion? I just want to make sure that we aren't creating a path to exfiltrate keys through the safety checker. I could take this out for now and come back to it in a future pr.

packages/core/src/policy/types.ts
allenhutchison: I've removed this for now. I don't have a good use case for it anymore.

packages/core/src/policy/types.ts
allenhutchison: Done.

allenhutchison:
Thanks for all the feedback. I've made updates to respond. I think this is ready for another look.

packages/core/src/policy/policy-engine.ts
```diff
@@ -89,20 +106,66 @@ export class PolicyEngine {
...
+            const result = await this.checkerRunner.runChecker(
+              toolCall,
+              checkerRule.checker,
+            );
+            if (result.decision !== SafetyCheckDecision.ALLOW) {
+              debugLogger.debug(
+                `[PolicyEngine.check] Safety checker denied: ${result.reason}`,
+              );
+              return {
+                decision: PolicyDecision.DENY,
```
avilladsen: This effectively remaps ASK_USER to DENY, which I don't think is intended.

packages/core/src/policy/policy-engine.ts
```diff
@@ -89,20 +106,66 @@ export class PolicyEngine {
...
+
+    // If decision is ALLOW, run safety checkers
+    if (decision === PolicyDecision.ALLOW && this.checkerRunner) {
+      for (const checkerRule of this.checkers) {
+        if (ruleMatches(checkerRule, toolCall, stringifiedArgs, serverName)) {
+          debugLogger.debug(
+            `[PolicyEngine.check] Running safety checker: ${checkerRule.checker.name}`,
+          );
+          try {
+            const result = await this.checkerRunner.runChecker(
```
avilladsen, line 139: To minimize latency we should consider running safety checkers in parallel, rather than sequentially. I'd maybe save that for a separate PR, though, since it'd be fairly complicated in its own right if you want to preserve short-circuiting on DENY (and ideally kill still pending safety checkers).

packages/core/src/safety/protocol.ts
avilladsen: Agreed that, with the refactor to run all safety checkers in a separate phase after checking policy rules, CONTINUE_CHECKING doesn't make sense as a separate value from ALLOW anymore.

packages/core/src/safety/context-builder.ts
avilladsen: I'd suggest either:
* If you have particular safety checkers in mind that need access to parts of `Config`, make exactly those parts of `Config` available to safety checkers, as narrowly as possible.
* If you don't have specific requirements, postpone this feature until you do.

That is, I think we'd be better served by building out this feature incrementally, so we can maintain control over what context is available to safety checkers. That means accepting some friction on the part of safety checker developers, who will need to file feature requests for more context, but I don't think that would be an excessive burden. I don't imagine that many folks will be writing safety checkers, for one.

packages/core/src/policy/policy-engine.ts
```diff
@@ -89,20 +106,66 @@ export class PolicyEngine {
...
+              toolCall,
+              checkerRule.checker,
+            );
+            if (result.decision !== SafetyCheckDecision.ALLOW) {
```
avilladsen: This remaps ASK_USER to DENY, which I don't think is intended.

packages/core/src/policy/policy-engine.ts
```diff
@@ -89,20 +106,66 @@ export class PolicyEngine {
...
+    if (!decision) {
+      // No matching rule found, use default decision
+      debugLogger.debug(
+        `[PolicyEngine.check] NO MATCH - using default decision: ${this.defaultDecision}`,
+      );
+      decision = this.applyNonInteractiveMode(this.defaultDecision);
+    }
+
+    // If decision is ALLOW, run safety checkers
+    if (decision === PolicyDecision.ALLOW && this.checkerRunner) {
```
avilladsen: I think safety checkers should also be called if the decision from the policy rules is `ASK_USER`. The way I'm thinking about safety checkers is that they run independently from each other and the policy rules, and can "downgrade" the overall decision along the sequence `ALLOW` -> `ASK_USER` -> `DENY`. So an `ASK_USER` decision could still be changed to `DENY`; it's only when the decision is already `DENY` that safety checkers can be skipped.

avilladsen (changes requested):
Broadly looks good! Mostly I think it just needs some cleaning up of how it handles a safety checkers returning `ASK_USER`.

packages/core/src/safety/protocol.ts
```diff
@@ -0,0 +1,106 @@
...
+ */
+
+import type { FunctionCall } from '@google/genai';
+
+/**
+ * Represents a single turn in the conversation between the user and the model.
+ * This provides semantic context for why a tool call might be happening.
+ */
+export interface ConversationTurn {
+  user: {
```
laurentsimon, line 14: In general, the trajectory is a list of prompt type (system, user text or multimedia, model). Are we able to capture all of it?

packages/core/src/safety/protocol.ts
```diff
@@ -0,0 +1,106 @@
...
+     * relevant to the checker. The policy engine is responsible for
+     * populating this, ensuring no secrets are leaked.
+     */
+    config?: Record<string, unknown>;
+
+    /**
+     * The recent history of the conversation. This can be used by checkers
+     * that need to understand the intent behind a tool call.
+     */
+    history?: {
```
laurentsimon, line 57: Let's say the checker needs to keep state, is it possible? An example would be that the checker summarizes previous turns to reduce context size

packages/core/src/safety/protocol.ts
```diff
@@ -0,0 +1,106 @@
...
+  };
+}
+
+/**
+ * The possible decisions a safety checker can make.
+ */
+export enum SafetyCheckDecision {
+  ALLOW = 'allow',
+  DENY = 'deny',
+  ASK_USER = 'ask_user',
```
laurentsimon, line 75: The checker may be dependent on the answer of another checker as an input. For example, my checker may want to call a classifier checker on a specific input with a specific prompt; or it might want to re-use a GitHub checker when the trajectory is relevant to GitHub. Is it possible? Would the checker calls the other checker itself, or respond with "ASK_CHECKER: 'checker_name, custom_safety_input'" and later be called back with the answer as additional input in SafetyCheckInput?

The above feature implies that some checker sare only called "on-demand" by other checkers: you would not want to call the GitHub checker if the trajectory has nothing to do with GitHub.

packages/core/src/safety/protocol.ts
```diff
@@ -0,0 +1,106 @@
...
+  CONTINUE_CHECKING = 'continue_checking',
```
laurentsimon: what does it do?

packages/core/src/safety/protocol.ts
```diff
@@ -0,0 +1,106 @@
...
+  model: {
+    text?: string;
+    toolCalls?: FunctionCall[];
+  };
+}
+
+/**
+ * The data structure passed from the CLI to a safety checker process via stdin.
+ */
+export interface SafetyCheckInput {
```
laurentsimon, line 26: Imagine the checker is itself an agent and needs to make calls and use a model. For example, to decide if a write to a document is allowed, it may check who owns the doc. Would it be able to reuse the gcli API or would it need to be a standalone implementation

packages/core/src/safety/protocol.ts
avilladsen: That's a good point: currently a new checker instance is spawned for each check, but some checkers may benefit from persistence to maintain state or avoid startup costs. The checker framework should probably support both ephemeral and persistent checkers.

A potential inspiration here is the design of [native messaging for web extensions](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Native_messaging). Native messaging supports both a single request-response "connectionless" mode, like is currently implemented here, as well as a "connection-based" persistent BIDI mode.

packages/core/src/safety/context-builder.ts
allenhutchison: Fair enough. Removed.

packages/core/src/policy/policy-engine.ts
allenhutchison: Good point. Fixed.

packages/core/src/policy/policy-engine.ts
allenhutchison: Fair point. I've filed issue https://github.com/google-gemini/gemini-cli/issues/12848 to do this in a follow up PR.

packages/core/src/policy/policy-engine.ts
allenhutchison: Fixed.

packages/core/src/safety/protocol.ts
allenhutchison: We are able to capture everything that is in the current conversation post-compression.

packages/core/src/safety/protocol.ts
allenhutchison: For an in-process checker I'm planning to allow it to use gcli creds / api the same way we do for a sub-agent. That hasn't landed yet, and we may need to special case some in-process checkers before sub-agents land, but that's the intention.

For out of process checkers I'm not planning to share the gcli credentials, but in those cases they could spawn gcli in headless mode.

packages/core/src/safety/protocol.ts
allenhutchison: All things I want to explore. I think checkers could maintain state by writing state files to the ~/.gemini directory as a start.

packages/core/src/safety/protocol.ts
allenhutchison: I'm open to these use cases, and have tried to design the protocol to be extensible, but the assumption for this first decision is that checkers are self-contained.

packages/core/src/safety/protocol.ts
allenhutchison: This was the start of dependencies between checkers, but it's too much to add in this PR. I've removed it for now.

allenhutchison:
Addressed everyone's feedback and adjusted as required. Ready for a final pass.

packages/core/src/safety/checker-runner.ts
```diff
@@ -0,0 +1,286 @@
...
+            reason: `Safety checker "${checkerName}" exited with code ${code}${
+              stderr ? `: ${stderr}` : ''
+            }`,
+          });
+          return;
+        }
+
+        // Try to parse the output
+        try {
+          const result: SafetyCheckResult = JSON.parse(stdout);
```
abhipatel12, line 215: nit: doesn't have to be now, but we can use zod here to validate the structure.

packages/core/src/safety/checker-runner.ts
```diff
@@ -72,9 +73,14 @@ export class CheckerRunner {
           )
         : this.contextBuilder.buildFullContext();
 
+      const toolCallWithConfig = {
```
abhipatel12: Question, `FunctionCall` doesn't have a `config` property. In `built-in.ts`, we access this by doing `as any`. 

Is this the best way we can do this? Feels a bit odd from a type safety perspective.

packages/core/src/safety/checker-runner.ts
```diff
@@ -148,8 +159,8 @@ export class CheckerRunner {
         killed = true;
         child.kill('SIGTERM');
```
abhipatel12, line 163: Do we need to consider the case where `SIGTERM` is ignored? Do we need to have a fallback to `SIGKILL` after some grace period?

packages/core/src/safety/checker-runner.ts
abhipatel12: What if we injected `CheckerRegistry` via the constructor instead. This would let us test things more easily. 

So in the test file we could avoid: 

```typescript
// Inject mocked registry
// @ts-expect-error - accessing private property for test
runner['registry'] = mockRegistry;
```

abhipatel12 (approved):
Looks great! Left a couple of questions but most can be follow-ups or nits.

avilladsen (approved):

packages/core/src/safety/checker-runner.ts
allenhutchison: Good suggestion. If it ignores the SIGTERM I'll issue a SIGKILL after a few seconds to make sure it's cleaned up.

packages/core/src/safety/checker-runner.ts
allenhutchison: Will do as a follow up: https://github.com/google-gemini/gemini-cli/issues/12958

packages/core/src/safety/checker-runner.ts
allenhutchison: Done.

packages/core/src/safety/checker-runner.ts
allenhutchison: Fixed.

PR #12511, opened 2025-11-04
fix(core): prevent server name spoofing in policy engine

allenhutchison:
Previously, the policy engine's wildcard matching for tool names was vulnerable to spoofing. A rule intended for a specific server, like `trusted-server__*`, could be incorrectly matched by a tool from a different server named `trusted-server__malicious`.

This change introduces a more robust check by passing the `serverName` through the tool confirmation request. When a wildcard rule is evaluated, the policy engine now requires an exact match between the `serverName` and the rule's prefix, preventing other servers from matching.

A new test case has been added to verify that spoofed server names are correctly handled and do not match wildcard rules they shouldn't.


packages/core/src/tools/tools.ts
```diff
@@ -78,6 +78,7 @@ export abstract class BaseToolInvocation<
     protected readonly messageBus?: MessageBus,
     readonly _toolName?: string,
     readonly _toolDisplayName?: string,
+    readonly _serverName?: string,
```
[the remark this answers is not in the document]
allenhutchison: Fixed.

packages/core/src/policy/policy-engine.ts
```diff
@@ -17,13 +17,21 @@ function ruleMatches(
   rule: PolicyRule,
   toolCall: FunctionCall,
   stringifiedArgs: string | undefined,
+  serverName?: string,
```
cornmander: can we make this required? that would avoid this problem altogether.

allenhutchison: ok, I gave that a try. It requires passing undefined around for other tool calls. Take a look and let me know what you think.

packages/core/src/tools/mcp-tool.ts
```diff
@@ -77,7 +77,14 @@ class DiscoveredMCPToolInvocation extends BaseToolInvocation<
     // Use composite format for policy checks: serverName__toolName
     // This enables server wildcards (e.g., "google-workspace__*")
     // while still allowing specific tool rules
-    super(params, messageBus, `${serverName}__${serverToolName}`, displayName);
+    // while still allowing specific tool rules
```
abhipatel12: dupe comment we can remove

packages/core/src/policy/policy-engine.ts
```diff
@@ -17,13 +17,21 @@ function ruleMatches(
   rule: PolicyRule,
   toolCall: FunctionCall,
   stringifiedArgs: string | undefined,
+  serverName: string | undefined,
 ): boolean {
   // Check tool name if specified
```
abhipatel12, line 22: If `serverName` matches, we skip verifying that `toolCall.name` starts with the prefix. This is a safe bet with today's impl, should we enforce both checks just in case this assumption fails for whatever reason? Maybe not necessary

Ex. 

```typescript
if (serverName !== undefined && serverName !== prefix) {
  return false;
}
// Always verify the prefix, even if serverName matched
if (!toolCall.name || !toolCall.name.startsWith(prefix + '__')) {
  return false;
}
```

abhipatel12 (approved):
LGTM! Left a couple of nits

Giving my approval unless Greg has anything else

packages/core/src/tools/mcp-tool.ts
allenhutchison: Good catch. Done.

packages/core/src/policy/policy-engine.ts
allenhutchison: Makes sense. Belt and suspenders in this case.

PR #12520, opened 2025-11-04
fix(mcp): replace hardcoded port 7777 with dynamic port allocation for OAuth

cbcoutinho:
## Summary
Replace the hardcoded `REDIRECT_PORT` (7777) with dynamic port allocation to enable concurrent OAuth authentication flows for multiple MCP servers.

## Problem
The hardcoded port 7777 created a race condition where:
- ❌ Only one OAuth flow could execute at a time
- ❌ Multiple MCP servers could not authenticate simultaneously  
- ❌ Port conflicts would cause authentication failures

## Solution
Implement dynamic port allocation using the same approach as the Google OAuth implementation (`oauth2.ts`):
- ✅ Allocate ephemeral port before client registration
- ✅ Pass port through entire OAuth flow (registration → auth URL → callback → token exchange)
- ✅ Maintain support for `OAUTH_CALLBACK_PORT` environment variable

## Changes
- Add `getAvailablePort()` function (mirrors `oauth2.ts:441` implementation)
- Update method signatures to accept `redirectPort` parameter:
  - `registerClient()`
  - `startCallbackServer()`
  - `buildAuthorizationUrl()`
  - `exchangeCodeForToken()`
- Modify `authenticate()` to allocate port at start of flow
- Add `net` module import for port allocation

## Benefits
- 🚀 Multiple MCP servers can authenticate concurrently
- 🔒 No port conflicts between OAuth flows
- 🎯 Each flow gets its own ephemeral port
- 🔄 Consistent with existing Google OAuth implementation

## Test Results
```
✓ src/mcp/oauth-provider.test.ts (25 tests) 240ms
  Test Files  1 passed (1)
       Tests  25 passed (25)
```

All existing tests pass without modification.

## Files Changed
- `packages/core/src/mcp/oauth-provider.ts` (+64, -8)

---

_This PR was written with the help of AI, and reviewed by a Human_

### DIFF
--- a/packages/core/src/mcp/oauth-provider.ts
+++ b/packages/core/src/mcp/oauth-provider.ts
@@ -101,32 +101,24 @@ const HTTP_OK = 200;
  */
 function getAvailablePort(): Promise<number> {
   return new Promise((resolve, reject) => {
-    let port = 0;
-    try {
-      const portStr = process.env['OAUTH_CALLBACK_PORT'];
-      if (portStr) {
-        port = parseInt(portStr, 10);
-        if (isNaN(port) || port <= 0 || port > 65535) {
-          return reject(
-            new Error(`Invalid value for OAUTH_CALLBACK_PORT: "${portStr}"`),
-          );
-        }
-        return resolve(port);
+    const portStr = process.env['OAUTH_CALLBACK_PORT'];
+    if (portStr) {
+      const port = parseInt(portStr, 10);
+      if (isNaN(port) || port <= 0 || port > 65535) {
+        return reject(
+          new Error(`Invalid value for OAUTH_CALLBACK_PORT: "${portStr}"`),
+        );
       }
-      const server = net.createServer();
-      server.listen(0, () => {
-        const address = server.address()! as net.AddressInfo;
-        port = address.port;
-      });
-      server.on('listening', () => {
-        server.close();
-        server.unref();
-      });
-      server.on('error', (e) => reject(e));
-      server.on('close', () => resolve(port));
-    } catch (e) {
-      reject(e);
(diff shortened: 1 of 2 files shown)


packages/core/src/mcp/oauth-provider.ts
```diff
@@ -89,10 +90,46 @@ interface PKCEParams {
...
+        server.close();
+        server.unref();
+      });
+      server.on('error', (e) => reject(e));
+      server.on('close', () => resolve(port));
+    } catch (e) {
+      reject(e);
+    }
+  });
+}
```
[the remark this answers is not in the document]
cbcoutinho: done

packages/core/src/mcp/oauth-provider.ts
```diff
@@ -89,10 +90,38 @@ interface PKCEParams {
...
+      const port = parseInt(portStr, 10);
+      if (isNaN(port) || port <= 0 || port > 65535) {
+        return reject(
+          new Error(`Invalid value for OAUTH_CALLBACK_PORT: "${portStr}"`),
+        );
+      }
+      return resolve(port);
+    }
+
+    const server = net.createServer();
```
cornmander: This ends up launching an http server twice. Can you fold this logic into the callback server launch? Have that launch on a randomized port and then extract the selected port from an event handler there.

cbcoutinho: good catch, I will implement a fix that launches just a single http server

cbcoutinho: Done

cornmander (approved):

PR #12524, opened 2025-11-04
refactor(cli): consolidate repetitive tests in InputPrompt using it.each

MJjainam:
## Related Issues
Progress on #11643

### DIFF
--- a/packages/cli/src/ui/components/InputPrompt.test.tsx
+++ b/packages/cli/src/ui/components/InputPrompt.test.tsx
@@ -559,94 +559,52 @@ describe('InputPrompt', () => {
     });
   });
 
-  it('should complete a partial parent command', async () => {
-    // SCENARIO: /mem -> Tab
-    mockedUseCommandCompletion.mockReturnValue({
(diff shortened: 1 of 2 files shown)


packages/cli/src/ui/components/InputPrompt.test.tsx
```diff
@@ -235,63 +235,48 @@ describe('InputPrompt', () => {
...
-  });
-
-  it('should call shellHistory.getNextCommand on down arrow in shell mode', async () => {
-    props.shellModeActive = true;
-    const { stdin, unmount } = renderWithProviders(<InputPrompt {...props} />);
-
-    await act(async () => {
-      stdin.write('\u001B[B');
-      await waitFor(() =>
+  it.each([
```
jacob314: these are a stretch. if you need a generic expectation block then you are better off using beforeEach and afterEach blocks and a regular test.
Parameterized tests should be reserved for when the expectations can be expressed as data rather than code. there are of course exceptions but I would recommend against this change.

jacob314 (approved):
lgtm

PR #12538, opened 2025-11-04
fix(patch workflow): Ensure that the environment is listed on patch comments

richieforeman:
## Summary

I noticed that the patch workflow wasn't commenting on PR with the environment.  We are doing this for other steps of the patch workflow, but the final one was not.  This fixes the issue and adds tests.

I observed this on a recent patch'd PR, and figured I'd fix it.

This is actually a fairly trivial CL, the bulk of this is just some good tests around this behavior.

## Pre-Merge Checklist

N/A


sehoon38 (approved):
LGTM

bdmorgan (approved):

PR #12539, opened 2025-11-04
feat(core): Integrate remote experiments configuration

skeshive:
## Summary

This PR introduces a system to fetch remote experiment flags from the code assist server upon after CLI startup. This allows for dynamic, server-driven configuration of CLI features, reducing the need for client-side releases for certain changes.

The first feature to use this system is the context compression threshold, which can now be controlled by a remote flag as a fallback to the user's local configuration.


packages/cli/src/ui/AppContainer.tsx
```diff
@@ -257,6 +259,20 @@ export const AppContainer = (props: AppContainerProps) => {
     });
   }, [config]);
 
+  useEffect(() => {
```
jacob314: instead of using useEffect here I think this logic should be in core with events dispatched to the React UI when the config for experiments has changed. that will also make it easier to integrate this with non-interactive mode in the future. 
the alternate short term solution is to move this to gemini.tsx before the app is initialized. you could still dispatch the event from there and at least that would not tightly couple it to the app container.

skeshive: so experiments will only be fetched once per session so we don't need to send any events, also the UI doesnt even use the experiments yet

do we do any other initialization after the app is initialized? this doesnt need to be tied to the UI

packages/core/src/config/config.ts
```diff
@@ -625,6 +632,20 @@ export class Config {
...
     this.baseLlmClient = new BaseLlmClient(this.contentGenerator, this);
 
+    const codeAssistServer = getCodeAssistServer(this);
+    if (codeAssistServer) {
+      this.experimentsPromise = getExperiments(codeAssistServer)
+        .then((experiments) => {
+          this.setExperiments(experiments);
+        })
+        .catch((e) => {
+          console.error(e);
```
jacob314: the the debugLogger

packages/core/src/config/config.ts
```diff
@@ -625,6 +632,20 @@ export class Config {
...
+        });
+    } else {
+      this.experiments = undefined;
```
jacob314, line 646: nit: these seem like the existing default value.

packages/core/src/config/config.ts
```diff
@@ -1064,8 +1085,22 @@ export class Config {
...
   }
 
-  getCompressionThreshold(): number | undefined {
-    return this.compressionThreshold;
+  async getCompressionThreshold(): Promise<number | undefined> {
+    if (this.compressionThreshold) {
+      return this.compressionThreshold;
+    }
+
+    if (this.experimentsPromise) {
```
jacob314, line 1094: need a try catch for this await ignoring the result  otherwise a failure to fetch the experiments will break calling getCompressionThreshold

packages/core/src/config/config.ts
skeshive: we need to clear out experiments if the auth changes to non-code assist

packages/core/src/config/config.ts
```diff
@@ -1064,8 +1086,26 @@ export class Config {
...
+      try {
+        await this.experimentsPromise;
+      } catch (e) {
+        debugLogger.debug('Failed to fetch experiments', e);
```
jacob314, line 1098: nit: debugLogger.warn or error rather than debug.

jacob314 (approved):
lgtm

PR #12544, opened 2025-11-04
Scrollable support

jacob314:
### Highlights

* **New Scrollable Components**: Introduced `Scrollable`, `VirtualizedList`, and `ScrollableList` components to enable efficient and interactive scrolling for large datasets within the CLI. The VirtualizedList class attempts to adopt best practices from top virtual scrolling libraries for React adapted slightly to the TUI environment.

* **Global Scroll Management**: Implemented a `ScrollProvider` to centralize and manage mouse wheel and keyboard (Shift+Up/Down) scrolling across various UI elements in the application.
* **Animated Scrollbars**: Added a `useAnimatedScrollbar` hook that provides visual feedback by animating scrollbar colors when a scrollable component is focused or actively scrolled.
* **Enhanced Debug Console**: The `DetailedMessagesDisplay` component now utilizes the new `ScrollableList` for a performant and scrollable debug console, improving the experience for viewing extensive logs.
* **Comprehensive Testing and Demo**: Included dedicated test files for `Scrollable`, `VirtualizedList`, and `ScrollableList`, along with a new `scrollable-list-demo.tsx` example to showcase the functionality.

To demo run
`npx tsx packages/cli/examples/scrollable-list-demo.tsx`
which shows a case of lazy scrolling with 1001 large items. You can scroll inside the box with the scroll wheel, there are sticky headers, pressing 'A' adds items, pressing 'E' edits the last item, and Ctrl+E scrolls to end.

I've also tweaked the debug console to use the Scrollable list class. To test that be sure to enable alternate buffer mode. Easiest way to get a lot of output int he debug console is to enable the keyboard logging.

scrollable-list-demo in action:
image

Fixes: #7023


packages/cli/examples/scrollable-list-demo.tsx
```diff
@@ -0,0 +1,163 @@
...
+                          <Box
+                            borderStyle="single"
+                            borderTop={true}
+                            borderBottom={false}
+                            borderLeft={false}
+                            borderRight={false}
+                            borderColor="gray"
+                          />
+                        </Box>
+                      }
```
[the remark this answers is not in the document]
jacob314: you are out of date with our version of ink.

packages/cli/src/ui/components/shared/ScrollableList.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+function ScrollableList<T>(
+  props: ScrollableListProps<T>,
+  ref: React.Ref<ScrollableListRef<T>>,
+) {
+  const { hasFocus } = props;
+  const virtualizedListRef = useRef<VirtualizedListRef<T>>(null);
+  const containerRef = useRef<DOMElement>(null);
+
+  useImperativeHandle(ref, () => virtualizedListRef.current!, []);
```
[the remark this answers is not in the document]
jacob314: fixed

packages/cli/src/ui/hooks/useAnimatedScrollbar.ts
```diff
@@ -0,0 +1,102 @@
...
+              }
+            }
+          };
+
+          animationFrame.current = setInterval(animateFadeOut, 33);
+        }, visibleDuration);
+      }
+    };
+
+    animationFrame.current = setInterval(animateFadeIn, 33);
```
galdawave, line 82: Nit: The use of setInterval with a 33ms delay (~30fps) for scrollbar animation might be too aggressive for a TUI environment, potentially causing high CPU usage or render lag, especially if multiple scrollbars animate simultaneously. Consider increasing the interval to 50ms or 100ms, or using a simpler animation strategy (e.g., just fewer frames) to reduce rendering load.

packages/cli/src/ui/components/DetailedMessagesDisplay.tsx
```diff
@@ -4,79 +4,114 @@
...
 
 export const DetailedMessagesDisplay: React.FC<
   DetailedMessagesDisplayProps
-> = ({ messages, maxHeight, width }) => {
+> = ({ messages, maxHeight, width, hasFocus }) => {
+  const scrollableListRef = useRef<ScrollableListRef<ConsoleMessageItem>>(null);
+
+  const borderAndPadding = 4;
+
+  const estimatedItemHeight = useCallback(
```
galdawave, line 31: Does this callback assume that we won't be handling ansi characters / complex characters?

packages/cli/src/ui/components/shared/Scrollable.tsx
```diff
@@ -0,0 +1,161 @@
...
+      size.innerHeight !== innerHeight ||
+      size.scrollHeight !== scrollHeight
+    ) {
+      setSize({ innerHeight, scrollHeight });
+      if (isAtBottom) {
+        setScrollTop(Math.max(0, scrollHeight - innerHeight));
+      }
+    }
+
+    const childCountCurrent = React.Children.count(children);
```
galdawave, line 77: Will the scroll height calculation update if the child stays the same but the contents of the child changes for example 100lines are added to the child.

packages/cli/src/ui/components/DetailedMessagesDisplay.tsx
jacob314: the estimated heights for the virtual list view don't have to be exact. When items are actually added to the view the exact height measurement will be made. the estimate is just needed to figure out how many items to try to bring in on the first frame and compute an reasonable height for the scrollbar.

packages/cli/src/ui/components/shared/VirtualizedList.tsx
```diff
@@ -0,0 +1,493 @@
...
+      scrollTo: (offset: number) => {
+        setIsStickingToBottom(false);
+        const newScrollTop = Math.max(
+          0,
+          Math.min(totalHeight - scrollableContainerHeight, offset),
+        );
+        setScrollAnchor(getAnchorForScrollTop(newScrollTop, offsets));
+      },
+      scrollToEnd: () => {
+        const currentData = dataRef.current;
```
galdawave: The data prop is already included in the useImperativeHandle dependency array and is guaranteed to be fresh within that scope. Using the ref creates a slight risk of operating on stale data if scrollToEnd is called between a props change and the useEffect that updates the ref. Use the data prop directly inside scrollToEnd instead of dataRef.current for improved safety and consistency.

packages/cli/src/ui/components/shared/VirtualizedList.tsx
```diff
@@ -0,0 +1,493 @@
...
+        setIsStickingToBottom(true);
+        if (currentData.length > 0) {
+          setScrollAnchor({
+            index: currentData.length - 1,
+            offset: Number.MAX_SAFE_INTEGER,
```
galdawave: Nit: Introduce a named constant like const SCROLL_TO_ITEM_END instead of Number.MAX_SAFE_INTEGER at the top of the file to improve the semantic clarity of the code.

packages/cli/src/ui/hooks/useAnimatedScrollbar.ts
jacob314: thought about this more. think this is actually fine as we are careful to do the animations time based rather than requiring N frames of animations to run. did refactor the code a little to cancel the intervals correctly when the useAnimatedScrollbar goes out of scope.

packages/cli/src/ui/components/shared/VirtualizedList.tsx
jacob314: done. swapped all uses of MAX_SAFE_INTEGER for SCROLL_TO_ITEM_END

packages/cli/src/ui/components/shared/VirtualizedList.tsx
jacob314: agreed. done

packages/cli/src/ui/components/shared/Scrollable.tsx
jacob314: Great point. I'd like to land this as is but follow up by implementing ResizeObserver in Ink. That is really what I need here and in VirtualizedList.tsx. https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver
What I am doing here happens to work as long as the child gets changed or the scrollable/VirtualizeList itself gets updated so the layoutEffect can run but is flawed if it doesn't. Fyi added a test for the similar case to this in VirtualizedList. It currently passes but I thin k it could be modified to keep even more objects const so it didn't pass.

galdawave (approved):

nssuwan186-dev:
Vv

PR #12547, opened 2025-11-04
[Extension Reloading]: Update custom commands, add enable/disable command

jakemac53:
## Summary

- Adds enable/disable commands within the UI when extension reloading is enabled.
- Adds reloading of custom commands from extensions when extension reloading is enabled.

## Details

Unfortunately due to how the custom commands are loaded, this is again event based. It would require yet another extensive refactor to make it work differently which I probably don't have time for.

This is however much lower risk than the MCP server reloading - there are not really any race conditions to worry about, the timing of things is much less important, as the entire custom command context is always reloaded in its current state whenever an event is seen.

Also adds an additional SettingScope called "Session" which applies only to the current session (in memory), which is only an option for `enable/disable` within the CLI.

## Related Issues

Part of https://github.com/google-gemini/gemini-cli/issues/11108.

## How to Validate

Enable extension reloading `"experiments": {"enableExtensionReloading": true}` - try enabling and disabling an extension which has custom commands. You should see them becoming available and not available.

## Pre-Merge Checklist

  - [ ] MacOS
    - [ ] npm run
    - [ ] npx
    - [ ] Docker
    - [ ] Podman
    - [ ] Seatbelt
  - [ ] Windows
    - [ ] npm run
    - [ ] npx
    - [ ] Docker
  - [ ] Linux
    - [x] npm run
    - [ ] npx
    - [ ] Docker

### DIFF
--- a/packages/cli/src/config/extension-manager.ts
+++ b/packages/cli/src/config/extension-manager.ts
@@ -616,15 +616,21 @@ export class ExtensionManager extends ExtensionLoader {
       throw new Error(`Extension with name ${name} does not exist.`);
     }
 
-    const scopePath =
-      scope === SettingScope.Workspace ? this.workspaceDir : os.homedir();
-    this.extensionEnablementManager.disable(name, true, scopePath);
-    extension.isActive = false;
-    await this.maybeStopExtension(extension);
+    if (scope !== SettingScope.Session) {
+      const scopePath =
+        scope === SettingScope.Workspace ? this.workspaceDir : os.homedir();
+      this.extensionEnablementManager.disable(name, true, scopePath);
+    }
     logExtensionDisable(
       this.telemetryConfig,
       new ExtensionDisableEvent(hashValue(name), extension.id, scope),
     );
+    if (!this.config || this.config.getEnableExtensionReloading()) {
+      // Only toggle the isActive state if we are actually going to disable it
+      // in the current session, or we haven't been initialized yet.
+      extension.isActive = false;
+    }
+    await this.maybeStopExtension(extension);
   }
 
   /**
@@ -644,14 +650,21 @@ export class ExtensionManager extends ExtensionLoader {
     if (!extension) {
       throw new Error(`Extension with name ${name} does not exist.`);
     }
-    const scopePath =
-      scope === SettingScope.Workspace ? this.workspaceDir : os.homedir();
-    this.extensionEnablementManager.enable(name, true, scopePath);
+
+    if (scope !== SettingScope.Session) {
+      const scopePath =
+        scope === SettingScope.Workspace ? this.workspaceDir : os.homedir();
+      this.extensionEnablementManager.enable(name, true, scopePath);
+    }
     logExtensionEnable(
       this.telemetryConfig,
       new ExtensionEnableEvent(hashValue(name), extension.id, scope),
     );
-    extension.isActive = true;
+    if (!this.config || this.config.getEnableExtensionReloading()) {
+      // Only toggle the isActive state if we are actually going to disable it
+      // in the current session, or we haven't been initialized yet.
+      extension.isActive = true;
+    }
     await this.maybeStartExtension(extension);
   }
 }
--- a/packages/cli/src/config/settings.ts
+++ b/packages/cli/src/config/settings.ts
@@ -162,14 +162,18 @@ export enum SettingScope {
   Session = 'Session',
 }
 
-// A type representing the settings scopes that are supported for LoadedSettings.
+/**
+ * A type representing the settings scopes that are supported for LoadedSettings.
+ */
 export type LoadableSettingScope =
   | SettingScope.User
   | SettingScope.Workspace
   | SettingScope.System
   | SettingScope.SystemDefaults;
 
-// The actual values of the loadable settings scopes.
+/**
+ * The actual values of the loadable settings scopes.
+ */
 const _loadableSettingScopes = [
   SettingScope.User,
   SettingScope.Workspace,
@@ -177,8 +181,10 @@ const _loadableSettingScopes = [
   SettingScope.SystemDefaults,
 ];
 
-// A type guard function that checks if `scope` is a loadable settings scope,
-// and allows promotion to the `LoadableSettingsScope` type based on the result.
+/**
+ * A type guard function that checks if `scope` is a loadable settings scope,
+ * and allows promotion to the `LoadableSettingsScope` type based on the result.
+ */
 export function isLoadableSettingScope(
   scope: SettingScope,
 ): scope is LoadableSettingScope {
--- a/packages/cli/src/config/settings.ts
+++ b/packages/cli/src/config/settings.ts
@@ -157,7 +157,32 @@ export enum SettingScope {
   Workspace = 'Workspace',
   System = 'System',
   SystemDefaults = 'SystemDefaults',
+  // Note that this scope is not supported in the settings dialog at this time,
+  // it is only supported for extensions.
   Session = 'Session',
+}
+
+// A type representing the settings scopes that are supported for LoadedSettings.
+export type LoadableSettingScope =
+  | SettingScope.User
+  | SettingScope.Workspace
+  | SettingScope.System
+  | SettingScope.SystemDefaults;
+
+// The actual values of the loadable settings scopes.
+const _loadableSettingScopes = [
+  SettingScope.User,
+  SettingScope.Workspace,
+  SettingScope.System,
+  SettingScope.SystemDefaults,
+];
+
+// A type guard function that checks if `scope` is a loadable settings scope,
+// and allows promotion to the `LoadableSettingsScope` type based on the result.
+export function isLoadableSettingScope(
+  scope: SettingScope,
+): scope is LoadableSettingScope {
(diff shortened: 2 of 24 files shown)


packages/cli/src/config/extension-manager.ts
```diff
@@ -616,15 +616,21 @@ export class ExtensionManager extends ExtensionLoader {
       throw new Error(`Extension with name ${name} does not exist.`);
     }
 
-    const scopePath =
-      scope === SettingScope.Workspace ? this.workspaceDir : os.homedir();
-    this.extensionEnablementManager.disable(name, true, scopePath);
-    extension.isActive = false;
-    await this.maybeStopExtension(extension);
+    if (scope !== SettingScope.Session) {
```
richieforeman, line 619: Paranoid: I tend to dislike if statements that match multiple cases with a singular exception.

Can we instead refine this to be inclusive of the situations you want to explicitly handle?

basically: 

```
const INCLUDED_SCOPES = new Set<SettingScope>([
  SESSION,
  DESKTOP,
 WHATEVER,
]);
if(INCLUDED_SCOPES.has(scope))
```

richieforeman: Altho my suggestion is more verbose/explicit, it does make it much more intentional.  The set definition can also be a form of documentation -- maybe the variable is named `HOT_RELOADABLE_EXTENSION_SCOPES` or something.  Then it's very apparent and intentional what's going on and how it effects the code.

packages/cli/src/ui/commands/extensionsCommand.ts
```diff
@@ -159,6 +161,158 @@ async function exploreAction(context: CommandContext) {
   }
 }
 
+function getEnableDisableContext(
+  context: CommandContext,
+  args: string,
```
richieforeman: Super petty nit: I generally expect "plural-like names" in the code to be lists/arrays/set. However, this is just a single string.

Perhaps this is "argumentsString" or something?

packages/cli/src/ui/commands/extensionsCommand.ts
```diff
@@ -159,6 +161,158 @@ async function exploreAction(context: CommandContext) {
...
+  scope: SettingScope;
+} | null {
+  const extensionLoader = context.services.config?.getExtensionLoader();
+  if (!(extensionLoader instanceof ExtensionManager)) {
+    debugLogger.error(
+      `Cannot ${context.invocation?.name} extensions in this environment`,
+    );
+    return null;
+  }
+  const parts = args.split(' ');
```
richieforeman: Leaving this as food-for-thought, but I wonder if it's worthwile to use an argument parsing lib on this string (like yargs).

https://www.npmjs.com/package/yargs-parser

These libs handle all of the goofy corner cases, and I'm pretty sure we're already using this in other places.

packages/cli/src/ui/commands/extensionsCommand.ts
```diff
@@ -159,6 +161,158 @@ async function exploreAction(context: CommandContext) {
...
+    );
+    return null;
+  }
+  let scope: SettingScope;
+  // Transform `--scope=<scope>` to `--scope <scope>`.
+  if (parts.length === 2) {
+    parts.push(...parts[1].split('='));
+    parts.splice(1, 1);
+  }
+  switch (parts[2]) {
```
richieforeman: You might want to do a `toLowerCase()` on this.

packages/cli/src/ui/commands/extensionsCommand.ts
```diff
@@ -159,6 +161,158 @@ async function exploreAction(context: CommandContext) {
...
+  };
+}
+
+async function disableAction(context: CommandContext, args: string) {
+  const enableContext = getEnableDisableContext(context, args);
+  if (!enableContext) return;
+
+  const { names, scope, extensionManager } = enableContext;
+  for (const name of names) {
+    await extensionManager.disableExtension(name!, scope);
```
richieforeman: I recommend avoiding the null assertion operator in production code. This can mask lots of errors.

go/fireconsole-styleguide#avoid-typescript-assertion-operator-aka-tailbang

Cheap fix, you can use a filter to coerce the type and properly drop empty elements.

`for const name of names.filter(n => n)`

packages/cli/src/ui/commands/types.ts
```diff
@@ -211,7 +211,7 @@ export interface SlashCommand {
   completion?: (
     context: CommandContext,
     partialArg: string,
-  ) => Promise<string[]>;
+  ) => Promise<string[]> | string[];
```
richieforeman, line 214: Nit: I'd recommend leaving it to return just a Promise.  Dealing with an api that _might_ return a promise can be awkward and unnecessarily difficult to wrestle with as a consumer.

I'd just avoid this, it unnecessarily complicates the api.

packages/cli/src/utils/dialogScopeUtils.ts
```diff
@@ -40,7 +40,7 @@ export function getScopeMessageForSetting(
   settings: LoadedSettings,
 ): string {
   const otherScopes = Object.values(SettingScope).filter(
-    (scope) => scope !== selectedScope,
+    (scope) => scope !== selectedScope && scope !== SettingScope.Session,
```
richieforeman: This feels adjacent to my earlier comment.

packages/cli/src/config/extension-manager.ts
jakemac53: In this case, I think it's safer to keep it as is - this isn't actually handling any reloading, it is writing to disk the enable/disable setting. Everything except the "Session" scope should persist to disk.

For instance if we added some new scope (lets call it "parent dir"), the chance is much higher that we would want it to write that setting to disk than not.

packages/cli/src/ui/commands/extensionsCommand.ts
jakemac53: Sure, done

packages/cli/src/ui/commands/extensionsCommand.ts
jakemac53: Yes I think that eventually we do want to do this, but I think it's a much broader change that should be baked into the slash command processor itself. Currently, argument parsing is manually handled by every different slash command which is madness I agree 🤣 .

I did attempt to use `yargs` actually at first, but found it incredibly difficult to use to parse arbitrary strings - it seems like this yargs_parser lib would be reasonable though.

packages/cli/src/ui/commands/extensionsCommand.ts
jakemac53: done

packages/cli/src/ui/commands/extensionsCommand.ts
jakemac53: I filed https://github.com/google-gemini/gemini-cli/issues/12594

packages/cli/src/ui/commands/extensionsCommand.ts
jakemac53: This was actually cruft, it wasn't necessary at all, removed

packages/cli/src/ui/commands/types.ts
jakemac53: You can always still just `await` the result if you don't want to deal with synchronous values - and there isn't much dealing with raw promises (this change required no code updates in the project).

It has the benefit of allowing for completion functions that are either synchronous or asynchronous which avoids multiple pointless event loops before auto completions.

packages/cli/src/utils/dialogScopeUtils.ts
jakemac53: Looking into tightening the actual types below and adding a shared function with a type guard for whether a scope is something supported by LoadedSettings to address this.

jakemac53: Ok, I pushed up that change in this commit https://github.com/google-gemini/gemini-cli/pull/12547/commits/537eb51a02f0ad9575081af31ea2ebb4919ec374.

Let me know what you think, this is the most type-safe way I can think of supporting this but its fairly invasive.

packages/cli/src/config/extension-manager.ts
richieforeman: Noted and understood.

packages/cli/src/config/settings.ts
```diff
@@ -157,6 +157,32 @@ export enum SettingScope {
...
+const _loadableSettingScopes = [
+  SettingScope.User,
+  SettingScope.Workspace,
+  SettingScope.System,
+  SettingScope.SystemDefaults,
+];
+
+// A type guard function that checks if `scope` is a loadable settings scope,
+// and allows promotion to the `LoadableSettingsScope` type based on the result.
+export function isLoadableSettingScope(
```
richieforeman, line 188: Nit: Prefer /** doc block */ style comments for functions and definitions.

packages/cli/src/ui/commands/types.ts
richieforeman: Fair enough, I can go with that.

richieforeman (approved):

packages/cli/src/config/settings.ts
jakemac53: done

PR #12266, opened 2025-10-29
feat(core): Support flexible overriding for various model parameters

joshualitt:
## Summary

Supports flexible overriding for various model parameters, as well as aliases.

## Details

There are 4 major inter-related changes in this PR:
1) Introduction of aliases to `settings.json`. An alias is basically a named collection of settings for generation . In this PR, the alias encapsulates the model name and generation config, but we can add any generation settings we like.
2) Introduction of overrides to `settings.json`. Overrides allow us to specify matches, and then arbitrarily override generation settings. In this PR, we support matching on agent name, and model / alias name, and we support overriding the model name and the generation parameters.
3) Introduction of `ModelGenerationConfigService`, the service for retrieving these configs.
4) Introduction of `ResolvedModelConfig`, the branded type we use to ensure integrity of the config. Note: In this PR, we still update this config in a few places due to necessity, but we should carefully minimize out-of-band updates in order to preserve data-driven toggle-ability of settings. On the plus side, the current design ensures every codepath must have a valid `ResolvedModelConfig` to begin with.

## How to Validate

Patch PR, run Gemini CLI confirm experience has not degraded.

## Pre-Merge Checklist

- [x] Updated relevant documentation and README (if needed)
- [x] Added/updated tests (if needed)
- [x] Noted breaking changes (if any)
- [x] Validated on required platforms/methods:
  - [ ] MacOS
    - [ ] npm run
    - [ ] npx
    - [ ] Docker
    - [ ] Podman
    - [ ] Seatbelt
  - [ ] Windows
    - [ ] npm run
    - [ ] npx
    - [ ] Docker
  - [x] Linux
    - [x] npm run
    - [ ] npx
    - [ ] Docker

PR #12556, opened 2025-11-04
feat(core): Add ModelConfigService.

joshualitt:
## Summary

Lands just the `ModelConfigService`, but does not wire it up anywhere but the configs.

## Details

Part 1 of N in an attempt to breakup https://github.com/google-gemini/gemini-cli/pull/12266. Future PRs will wire this up.

## How to Validate

Not wired up, confirm PR does not break preflight.

## Pre-Merge Checklist

- [x] Noted breaking changes (if any)
- [x] Validated on required platforms/methods:
  - [ ] MacOS
    - [ ] npm run
    - [ ] npx
    - [ ] Docker
    - [ ] Podman
    - [ ] Seatbelt
  - [ ] Windows
    - [ ] npm run
    - [ ] npx
    - [ ] Docker
  - [x] Linux
    - [x] npm run
    - [ ] npx
    - [ ] Docker

### DIFF
--- a/docs/get-started/configuration.md
+++ b/docs/get-started/configuration.md
@@ -296,6 +296,20 @@ their corresponding top-level category object in your `settings.json` file.
   - **Description:** Skip the next speaker check.
   - **Default:** `true`
 
+#### `generation`
+
+- **`generation.aliases`** (object):
+  - **Description:** Named presets for generation settings. Can be used in place
+    of a model name and can inherit from other aliases using an `extends`
+    property.
+  - **Default:**
+    `{"base":{"settings":{"config":{"temperature":0,"topP":1}}},"chat-base":{"extends":"base","settings":{"config":{"thinkingConfig":{"includeThoughts":true,"thinkingBudget":-1}}}},"gemini-2.5-pro":{"extends":"chat-base","settings":{"model":"gemini-2.5-pro"}},"gemini-2.5-flash":{"extends":"chat-base","settings":{"model":"gemini-2.5-flash"}},"gemini-2.5-flash-lite":{"extends":"chat-base","settings":{"model":"gemini-2.5-flash-lite"}},"classifier":{"extends":"base","settings":{"model":"gemini-2.5-flash-lite","config":{"maxOutputTokens":1024,"thinkingConfig":{"thinkingBudget":512}}}},"prompt-completion":{"extends":"base","settings":{"model":"gemini-2.5-flash-lite","config":{"temperature":0.3,"maxOutputTokens":16000,"thinkingConfig":{"thinkingBudget":0}}}},"edit-corrector":{"extends":"base","settings":{"model":"gemini-2.5-flash-lite","config":{"thinkingConfig":{"thinkingBudget":0}}}},"summarizer-default":{"extends":"base","settings":{"model":"gemini-2.5-flash-lite","config":{"maxOutputTokens":2000}}},"summarizer-shell":{"extends":"base","settings":{"model":"gemini-2.5-flash-lite","config":{"maxOutputTokens":2000}}},"web-search-tool":{"extends":"base","settings":{"model":"gemini-2.5-flash","config":{"tools":[{"googleSearch":{}}]}}},"web-fetch-tool":{"extends":"base","settings":{"model":"gemini-2.5-flash","config":{"tools":[{"urlContext":{}}]}}}}`
+
+- **`generation.overrides`** (array):
+  - **Description:** Apply specific generation settings based on the agent or
+    model being used. The most specific match will be used.
+  - **Default:** `[]`
+
 #### `context`
 
 - **`context.fileName`** (string | string[]):
--- a/docs/get-started/configuration.md
+++ b/docs/get-started/configuration.md
@@ -296,18 +296,18 @@ their corresponding top-level category object in your `settings.json` file.
   - **Description:** Skip the next speaker check.
   - **Default:** `true`
 
-#### `generation`
+#### `modelConfigs`
 
-- **`generation.aliases`** (object):
-  - **Description:** Named presets for generation settings. Can be used in place
-    of a model name and can inherit from other aliases using an `extends`
-    property.
+- **`modelConfigs.aliases`** (object):
+  - **Description:** Named presets for model configs. Can be used in place of a
+    model name and can inherit from other aliases using an `extends` property.
   - **Default:**
-    `{"base":{"settings":{"config":{"temperature":0,"topP":1}}},"chat-base":{"extends":"base","settings":{"config":{"thinkingConfig":{"includeThoughts":true,"thinkingBudget":-1}}}},"gemini-2.5-pro":{"extends":"chat-base","settings":{"model":"gemini-2.5-pro"}},"gemini-2.5-flash":{"extends":"chat-base","settings":{"model":"gemini-2.5-flash"}},"gemini-2.5-flash-lite":{"extends":"chat-base","settings":{"model":"gemini-2.5-flash-lite"}},"classifier":{"extends":"base","settings":{"model":"gemini-2.5-flash-lite","config":{"maxOutputTokens":1024,"thinkingConfig":{"thinkingBudget":512}}}},"prompt-completion":{"extends":"base","settings":{"model":"gemini-2.5-flash-lite","config":{"temperature":0.3,"maxOutputTokens":16000,"thinkingConfig":{"thinkingBudget":0}}}},"edit-corrector":{"extends":"base","settings":{"model":"gemini-2.5-flash-lite","config":{"thinkingConfig":{"thinkingBudget":0}}}},"summarizer-default":{"extends":"base","settings":{"model":"gemini-2.5-flash-lite","config":{"maxOutputTokens":2000}}},"summarizer-shell":{"extends":"base","settings":{"model":"gemini-2.5-flash-lite","config":{"maxOutputTokens":2000}}},"web-search-tool":{"extends":"base","settings":{"model":"gemini-2.5-flash","config":{"tools":[{"googleSearch":{}}]}}},"web-fetch-tool":{"extends":"base","settings":{"model":"gemini-2.5-flash","config":{"tools":[{"urlContext":{}}]}}}}`
+    `{"base":{"modelConfig":{"generateContentConfig":{"temperature":0,"topP":1}}},"chat-base":{"extends":"base","modelConfig":{"generateContentConfig":{"thinkingConfig":{"includeThoughts":true,"thinkingBudget":-1}}}},"gemini-2.5-pro":{"extends":"chat-base","modelConfig":{"model":"gemini-2.5-pro"}},"gemini-2.5-flash":{"extends":"chat-base","modelConfig":{"model":"gemini-2.5-flash"}},"gemini-2.5-flash-lite":{"extends":"chat-base","modelConfig":{"model":"gemini-2.5-flash-lite"}},"classifier":{"extends":"base","modelConfig":{"model":"gemini-2.5-flash-lite","generateContentConfig":{"maxOutputTokens":1024,"thinkingConfig":{"thinkingBudget":512}}}},"prompt-completion":{"extends":"base","modelConfig":{"model":"gemini-2.5-flash-lite","generateContentConfig":{"temperature":0.3,"maxOutputTokens":16000,"thinkingConfig":{"thinkingBudget":0}}}},"edit-corrector":{"extends":"base","modelConfig":{"model":"gemini-2.5-flash-lite","generateContentConfig":{"thinkingConfig":{"thinkingBudget":0}}}},"summarizer-default":{"extends":"base","modelConfig":{"model":"gemini-2.5-flash-lite","generateContentConfig":{"maxOutputTokens":2000}}},"summarizer-shell":{"extends":"base","modelConfig":{"model":"gemini-2.5-flash-lite","generateContentConfig":{"maxOutputTokens":2000}}},"web-search-tool":{"extends":"base","modelConfig":{"model":"gemini-2.5-flash","generateContentConfig":{"tools":[{"googleSearch":{}}]}}},"web-fetch-tool":{"extends":"base","modelConfig":{"model":"gemini-2.5-flash","generateContentConfig":{"tools":[{"urlContext":{}}]}}}}`
 
-- **`generation.overrides`** (array):
-  - **Description:** Apply specific generation settings based on the agent or
-    model being used. The most specific match will be used.
+- **`modelConfigs.overrides`** (array):
+  - **Description:** Apply specific configuration overrides based on matches,
+    with a primary key of model (or alias). The most specific match will be
+    used.
   - **Default:** `[]`
 
 #### `context`
--- a/packages/core/src/config/config.ts
+++ b/packages/core/src/config/config.ts
@@ -563,7 +563,7 @@ export class Config {
     // any `generation` settings (e.g., just `overrides`), the default `aliases`
     // are lost. This hack manually merges the default aliases back in if they
     // are missing from the user's config.
-    // TODO: Fix the settings loading logic to properly merge defaults and
+    // TODO(12593): Fix the settings loading logic to properly merge defaults and
     // remove this hack.
     let modelConfigs = params.modelConfigs;
     if (modelConfigs && !modelConfigs.aliases) {
--- a/packages/core/src/config/config.ts
+++ b/packages/core/src/config/config.ts
@@ -62,6 +62,9 @@ import { RipgrepFallbackEvent } from '../telemetry/types.js';
 import type { FallbackModelHandler } from '../fallback/types.js';
 import { ModelRouterService } from '../routing/modelRouterService.js';
 import { OutputFormat } from '../output/types.js';
+import type { ModelGenerationServiceConfig } from '../services/modelGenerationConfigService.js';
+import { ModelGenerationConfigService } from '../services/modelGenerationConfigService.js';
+import { DEFAULT_GENERATION_CONFIG } from '../config/defaultGenerationConfig.js';
 
 // Re-export OAuth config type
 export type { MCPOAuthConfig, AnyToolInvocation };
@@ -287,6 +290,7 @@ export interface ConfigParameters {
   recordResponses?: string;
   ptyInfo?: string;
   disableYoloMode?: boolean;
+  generation?: ModelGenerationServiceConfig;
   enableHooks?: boolean;
   hooks?: {
     [K in HookEventName]?: HookDefinition[];
@@ -304,6 +308,7 @@ export class Config {
   private fileSystemService: FileSystemService;
   private contentGeneratorConfig!: ContentGeneratorConfig;
   private contentGenerator!: ContentGenerator;
+  readonly generationConfigService: ModelGenerationConfigService;
   private readonly embeddingModel: string;
   private readonly sandbox: SandboxConfig | undefined;
   private readonly targetDir: string;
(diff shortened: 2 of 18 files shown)


packages/core/src/services/modelGenerationConfigService.ts
```diff
@@ -0,0 +1,218 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import type { GenerateContentConfig } from '@google/genai';
+
+export interface GenerationContext {
+  agent?: string;
```
allenhutchison: Should agents be configured by generation config or would it be a better separation of concerns to have agents _use_ generation configs? Maybe another way of saying this is shouldn't the generation config be a setting of an agent?

packages/core/src/services/modelGenerationConfigService.ts
```diff
@@ -0,0 +1,218 @@
...
+  model: string; // Can be a model name or an alias
+}
+
+export interface GenerationSettings {
+  model?: string;
+  config?: Partial<GenerateContentConfig>;
```
allenhutchison: You might want to name this config more specifically as generateContentConfig. If there is a new SDK that we want to support in the future it may not share the same GenerateContentConfig that comes from genai and we'll want to be able to support that as well.

packages/core/src/services/modelConfigService.ts
```diff
@@ -0,0 +1,218 @@
...
+  config?: Partial<GenerateContentConfig>;
+}
+
+export interface GenerationOverride {
+  match: {
+    agent?: string;
+    model?: string; // Can be a model name or an alias
+  };
+  settings: GenerationSettings;
+}
```
allenhutchison, line 37: As above, this seems to couple this feature and subagents in a way that isn't really necessary.

packages/core/src/services/modelConfigService.ts
```diff
@@ -0,0 +1,218 @@
...
+
+export interface GenerationAlias {
+  extends?: string;
```
allenhutchison, line 40: Is it possible to cast this into the right type rather than string?

packages/core/src/services/modelConfigService.ts
```diff
@@ -0,0 +1,218 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import type { GenerateContentConfig } from '@google/genai';
```
allenhutchison, line 7: Curious what you think about this class or a related class validating model config with https://ai.google.dev/api/models#models_get-SHELL the models API. It would give you a way to do some level of validations for the settings that come into the service and that would allow you to create really helpful error messages.

docs/get-started/configuration.md
```diff
@@ -296,6 +296,20 @@ their corresponding top-level category object in your `settings.json` file.
   - **Description:** Skip the next speaker check.
   - **Default:** `true`
 
+#### `generation`
+
```
allenhutchison, line 300: I know everyone has opinions on naming. However, I feel like the generation term used throughout this PR is not particularly obvious to a user or a developer. Maybe just use model, model config, or model alias to make it more clear?

packages/core/src/services/modelConfigService.ts
```diff
@@ -0,0 +1,218 @@
...
+        return null;
+      })
+      .filter((match): match is NonNullable<typeof match> => match !== null);
+
+    // The override application logic is designed to be both simple and powerful.
+    // By first sorting all matching overrides by specificity (and then by their
+    // original order as a tie-breaker), we ensure that as we merge the `config`
+    // objects, the settings from the most specific rules are applied last,
+    // correctly overwriting any values from broader, less-specific rules.
+    // This achieves a per-property override effect without complex per-property logic.
```
allenhutchison, line 162: You might look at the approach I'm taking for policies:
https://github.com/google-gemini/gemini-cli/blob/b6524e410a428e9cc94c5077d1bd0512c7ae6c8d/packages/core/src/policy/config.ts#L133

Here I support three levels of policy: Admin, User, and Default with similar override behviour. Really just asking if we should try to be consistent in these approaches to config overrides.

allenhutchison (commented):
Several questions / suggestions or ideas for you to think about. Really looking forward to this feature.

packages/core/src/services/modelGenerationConfigService.ts
joshualitt: Yes, we are on the same page about this. Agents should use generation configs, and not have a global config.

This use of `agent` is just as a key for overriding. As a concrete example, say we use a given generation config both in our regular flow, and via subagent. This logic allows us to disambiguate these uses, and to support overriding the config separately for each of these contexts. It also allows us to globally override settings for all configs used by a given `agent`, i.e. if we want to test workarounds.

However, we can remove this if you'd prefer not to have this flexibility. Thoughts?

packages/core/src/services/modelConfigService.ts
joshualitt: Sorry, I'm not sure I follow this comment, `extends` is a string in the schema(alias or model name)? Or do you mean you'd like this to be an enum?

packages/core/src/services/modelConfigService.ts
joshualitt: I love this idea! Can we do this in a follow-up?

packages/core/src/services/modelGenerationConfigService.ts
allenhutchison: I think the way I would approach that is to have two named configs:

core-model-config(2.5 pro, temp, etc...)
subagent-model-config(override core-model-config with 2.5 flash)

context-subagent (sets subagent-model-config) as the model

The way I read this code was that subagent-model-config(named context-subagent as the override) and therefore that override would apply to context-subagent. Did I get that wrong?

packages/core/src/services/modelConfigService.ts
joshualitt: I'm happy to switch to numerical priorities! However, I'm not sure they'll be quite as nice in this context because the specifiers naturally disambiguate by specificity, i.e. most specific match winning is a pretty intuitive way to think about priorities for overrides.

Still, numerical priority for a tie breaker could work. We could also just use numerical priority, but I think it's frequently very natural to think about overrides "matching" some specific context, i.e. the model name, or model name + agent context. Thoughts?

packages/core/src/services/modelConfigService.ts
allenhutchison: I'm not a typescript expert but I thought that this string represented a GenerationAlias or a GenerationConfig? I was just asking if we could take advantage of typechecking here to make sure the right kind of named thing ended up in this extends space.

packages/core/src/services/modelConfigService.ts
allenhutchison: Of course. If it's helpful I have code that does something similar here: https://github.com/allenhutchison/obsidian-gemini/blob/18e1b0fa29cd4be305298227094acd705dd039dd/src/services/model-discovery.ts#L70

packages/core/src/services/modelConfigService.ts
allenhutchison: I'm not sure that numerical priorities (at least exposed to the user) would be helpful here, but was more thinking about the idea of Admin overrides User overrides Default.

packages/core/src/services/modelGenerationConfigService.ts
joshualitt: We spoke offline, I'm definitely not opposed to tying configs to the agent, but for the time being, I think it makes sense to try and keep them orthogonal.

To clarify that the agent name is just used as a secondary key to scope the override, I've renamed this field to reflect it's actual use.

Marking this as resolved, but please unresolve if you'd like to go down the road of linking agents to aliases.

packages/core/src/services/modelConfigService.ts
joshualitt: Resolving per offline discussion.

packages/core/src/services/modelConfigService.ts
joshualitt: Ahh yes, I see what you mean! We should *definitely* add this, but I think we can scope this work to a follow-up? That is, can we just add this as a new property on the override, and then use it for the first round of tie breaking?

packages/core/src/services/modelConfigService.ts
joshualitt: Ahh I see what you mean, i.e. this should be something like `extends?: ModelConfigAlias` or similar. Today, this interface is just typing the JSON we receive in the settings, where this is just a string. We could definitely build a more robust hierarchy. I've added a TODO to process the JSON to build a typed hierarchy. Is it okay if we tackle this in a follow-up?

joshualitt:
@abhipatel12  @allenhutchison @anj-s ptal, PR has been updated to address the initial round of feedback.

docs/get-started/configuration.md
```diff
@@ -296,6 +296,20 @@ their corresponding top-level category object in your `settings.json` file.
   - **Description:** Skip the next speaker check.
   - **Default:** `true`
 
+#### `modelConfigs`
```
abhipatel12, line 299: Can we make this a bit easier to read?

One of the following options:

### Option 1

#### `modelConfigs`
*   **`modelConfigs.aliases`** (object):
    *   **Description:** Named presets for generation settings. Can be used in place of a model name and can inherit from other aliases using an `extends` property.
    *   **Default:**
        ```json
        {
          "base": {
            "modelConfig": {
              "generateContentConfig": {
                "temperature": 0,
                "topP": 1
              }
            }
          },
          "chat-base": {
            "extends": "base",
            "modelConfig": {
              "generateContentConfig": {
                "thinkingConfig": {
                  "includeThoughts": true,
                  "thinkingBudget": -1
                }
              }
            }
          },
          "gemini-2.5-pro": {
            "extends": "chat-base",
            "modelConfig": {
              "model": "gemini-2.5-pro"
            }
          },
          "gemini-2.5-flash": {
            "extends": "chat-base",
            "modelConfig": {
              "model": "gemini-2.5-flash"
            }
          },
          "gemini-2.5-flash-lite": {
            "extends": "chat-base",
            "modelConfig": {
              "model": "gemini-2.5-flash-lite"
            }
          },
          "classifier": {
            "extends": "base",
            "modelConfig": {
              "model": "gemini-2.5-flash-lite",
              "generateContentConfig": {
                "maxOutputTokens": 1024,
                "thinkingConfig": {
                  "thinkingBudget": 512
                }
              }
            }
          },
          "prompt-completion": {
            "extends": "base",
            "modelConfig": {
              "model": "gemini-2.5-flash-lite",
              "generateContentConfig": {
                "temperature": 0.3,
                "maxOutputTokens": 16000,
                "thinkingConfig": {
                  "thinkingBudget": 0
                }
              }
            }
          },
          "edit-corrector": {
            "extends": "base",
            "modelConfig": {
              "model": "gemini-2.5-flash-lite",
              "generateContentConfig": {
                "thinkingConfig": {
                  "thinkingBudget": 0
                }
              }
            }
          },
          "summarizer-default": {
            "extends": "base",
            "modelConfig": {
              "model": "gemini-2.5-flash-lite",
              "generateContentConfig": {
                "maxOutputTokens": 2000
              }
            }
          },
          "summarizer-shell": {
            "extends": "base",
            "modelConfig": {
              "model": "gemini-2.5-flash-lite",
              "generateContentConfig": {
                "maxOutputTokens": 2000
              }
            }
          },
          "web-search-tool": {
            "extends": "base",
            "modelConfig": {
              "model": "gemini-2.5-flash",
              "generateContentConfig": {
                "tools": [
                  {
                    "googleSearch": {}
                  }
                ]
              }
            }
          },
          "web-fetch-tool": {
            "extends": "base",
            "modelConfig": {
              "model": "gemini-2.5-flash",
              "generateContentConfig": {
                "tools": [
                  {
                    "urlContext": {}
                  }
                ]
              }
            }
          }
        }
        ```

---

### Option 2

#### `modelConfigs`
*   **`modelConfigs.aliases`** (object):
    *   **Description:** Named presets for generation settings. Can be used in place of a model name and can inherit from other aliases using an `extends` property.
    *   **Default:**
        <details>
        <summary>Click to expand default JSON</summary>

        ```json
        {
          "base": {
            "modelConfig": {
              "generateContentConfig": { "temperature": 0, "topP": 1 }
            }
          },
          "chat-base": {
            "extends": "base",
            "modelConfig": {
              "generateContentConfig": {
                "thinkingConfig": { "includeThoughts": true, "thinkingBudget": -1 }
              }
            }
          },
          "gemini-2.5-pro": {
            "extends": "chat-base",
            "modelConfig": { "model": "gemini-2.5-pro" }
          },
          "gemini-2.5-flash": {
            "extends": "chat-base",
            "modelConfig": { "model": "gemini-2.5-flash" }
          },
          "gemini-2.5-flash-lite": {
            "extends": "chat-base",
            "modelConfig": { "model": "gemini-2.5-flash-lite" }
          },
          "classifier": {
            "extends": "base",
            "modelConfig": {
              "model": "gemini-2.5-flash-lite",
              "generateContentConfig": {
                "maxOutputTokens": 1024,
                "thinkingConfig": { "thinkingBudget": 512 }
              }
            }
          },
          "prompt-completion": {
            "extends": "base",
            "modelConfig": {
              "model": "gemini-2.5-flash-lite",
              "generateContentConfig": {
                "temperature": 0.3,
                "maxOutputTokens": 16000,
                "thinkingConfig": { "thinkingBudget": 0 }
              }
            }
          },
          "edit-corrector": {
            "extends": "base",
            "modelConfig": {
              "model": "gemini-2.5-flash-lite",
              "generateContentConfig": { "thinkingConfig": { "thinkingBudget": 0 } }
            }
          },
          "summarizer-default": {
            "extends": "base",
            "modelConfig": {
              "model": "gemini-2.5-flash-lite",
              "generateContentConfig": { "maxOutputTokens": 2000 }
            }
          },
          "summarizer-shell": {
            "extends": "base",
            "modelConfig": {
              "model": "gemini-2.5-flash-lite",
              "generateContentConfig": { "maxOutputTokens": 2000 }
            }
          },
          "web-search-tool": {
            "extends": "base",
            "modelConfig": {
              "model": "gemini-2.5-flash",
              "generateContentConfig": { "tools": [{ "googleSearch": {} }] }
            }
          },
          "web-fetch-tool": {
            "extends": "base",
            "modelConfig": {
              "model": "gemini-2.5-flash",
              "generateContentConfig": { "tools": [{ "urlContext": {} }] }
            }
          }
        }
        ```
        </details>

packages/core/src/config/config.ts
```diff
@@ -560,6 +557,25 @@ export class Config {
     }
     this.geminiClient = new GeminiClient(this);
     this.modelRouterService = new ModelRouterService(this);
+
+    // HACK: The settings loading logic doesn't currently merge the default
+    // generation config with the user's settings. This means if a user provides
+    // any `generation` settings (e.g., just `overrides`), the default `aliases`
+    // are lost. This hack manually merges the default aliases back in if they
+    // are missing from the user's config.
+    // TODO: Fix the settings loading logic to properly merge defaults and
```
abhipatel12: Is this a todo for this epic?

packages/core/src/services/modelConfigService.ts
```diff
@@ -0,0 +1,233 @@
...
+      );
+    }
+
+    return {
+      model: resolved.model,
+      generateContentConfig: resolved.generateContentConfig,
+    } as ResolvedModelConfig;
+  }
+
+  private isObject(item: unknown): item is Record<string, unknown> {
```
abhipatel12, line 204: Does this intentionally exclude arrays from being treated as objects? So any array property in an override will replace the array from an alias and not merge them? 

Just want to make sure im understanding that correctly

abhipatel12 (commented):
left a few thoughts! lmk what you think!

docs/get-started/configuration.md
joshualitt: AFAICT this documentation is autogenerated? I filed https://github.com/google-gemini/gemini-cli/issues/12592 to fix how we handle multiline JSON objects.

packages/core/src/config/config.ts
joshualitt: I did have a PR and test for this, but it broke things. It may be that someone needs to carefully audit the settings to ensure the defaults still make sense. Filed https://github.com/google-gemini/gemini-cli/issues/12593 to address this, and updated the comment with the issue.

packages/core/src/services/modelConfigService.ts
joshualitt: Yes, I'm open to being convinced here!

I think the problem with deep merging arrays, is there isn't anyway to disable this behavior. Consider the case of setting a stop sequence, and then overriding it. If we don't deep merge arrays, and the user wants this behavior, i.e. to add a new stop sequence but preserve the old ones, then the user has to manually merge, which is annoying. However, if we do deep merge, and the user wants to fully override the stop sequences, they have no options.

Thoughts? The only way I can think to fix this issue is by allowing users to indicate how they want to merge arrays, which does feel valuable, but might not be something we have to solve right now.

packages/core/src/services/modelConfigService.ts
joshualitt: Filed https://github.com/google-gemini/gemini-cli/issues/12597 and updated the comment.

packages/core/src/services/modelConfigService.ts
joshualitt: Filed https://github.com/google-gemini/gemini-cli/issues/12598 to track this work.

allenhutchison: Absolutely fine to do in a follow up. Immediate plan sounds fine.

allenhutchison (approved):
Great collaboration I think this looks really good.

packages/core/src/services/modelConfigService.ts
```diff
@@ -0,0 +1,233 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import type { GenerateContentConfig } from '@google/genai';
+
+// The primary key for the ModelConfig is the model string. However, we also
+// support a secondary key to limit the override scope, typically an agent name.
+export interface ModelConfigKey {
+  model: string;
+  overrideScope?: string;
```
anj-s, line 23: can you explain overrideScope a bit more here?

packages/core/src/config/config.ts
```diff
@@ -287,6 +290,7 @@ export interface ConfigParameters {
   recordResponses?: string;
   ptyInfo?: string;
   disableYoloMode?: boolean;
+  modelConfigs?: ModelConfigServiceConfig;
```
anj-s: nit: modelConfigServiceConfig?

packages/core/src/config/config.ts
```diff
@@ -304,6 +308,7 @@ export class Config {
   private fileSystemService: FileSystemService;
   private contentGeneratorConfig!: ContentGeneratorConfig;
   private contentGenerator!: ContentGenerator;
+  readonly generationConfigService: ModelConfigService;
```
anj-s: nit: modelConfigService?

packages/core/src/config/defaultModelConfigs.ts
```diff
@@ -0,0 +1,121 @@
...
+    'gemini-2.5-pro': {
+      extends: 'chat-base',
+      modelConfig: {
+        model: 'gemini-2.5-pro',
+      },
+    },
+    'gemini-2.5-flash': {
+      extends: 'chat-base',
+      modelConfig: {
+        model: 'gemini-2.5-flash',
```
anj-s, line 47: should we have default configs here as well?

anj-s (approved):

packages/core/src/services/modelConfigService.ts
joshualitt: Added a comment and an example to clarify why we've added this.

packages/core/src/config/defaultModelConfigs.ts
joshualitt: Yes, definitely! These configs extend from `chat-base` and transitively to `base` which is how their defaults are generated.

I've added a comment to clarify the hierarchy. Also worth noting, I added a golden [test](https://github.com/google-gemini/gemini-cli/blob/0b809518220bb4ce68e02f3287fb2680a39e4850/packages/core/src/services/test-data/resolved-aliases.golden.json) so we can always see the fully lowered configs. For example, `gemini-2.5-pro` expands to:
```
  "gemini-2.5-pro": {
    "model": "gemini-2.5-pro",
    "generateContentConfig": {
      "temperature": 0,
      "topP": 1,
      "thinkingConfig": {
        "includeThoughts": true,
        "thinkingBudget": -1
      }
    }
  },
```

To your point though, my guess is now that we can tune these numbers, we'll start to see flash / pro / flash-lite diverge, so I anticipate this hierarchy will change substantially in the near future.

joshualitt:
@anj-s @allenhutchison  @abhipatel12 Okay, I think all feedback has been addressed, and I'm going to land this PR.

If there are other changes you'd like, please let me know and I will incorporate them into the follow-ups. Thanks!!

abhipatel12 (approved):
awesome lgtm thanks!

PR #12559, opened 2025-11-04
fix(shell): Enable interactive password prompts for sudo/doas on OpenBSD

### LINKED ISSUE #12454: Managed to install this on OpenBSD but...
### What happened?

I installed and ran Gemini CLI on OpenBSD using the normal Linux install steps. The CLI itself runs, but when it attempts to run an operation that requires privilege escalation it fails with this error:

```
sudo: a terminal is required to read the password; either use the -S option to read from standard input or configure an askpass helper
sudo: a password is required
```

doas says the same:

```
doas: a tty is required
```

I do not really understand if this is something that triggers the OpenBSD security or if it doesn't allocate a tty for Gemini by default. On my Linux machine I did not see this, it just worked.

I expected for a password prompt like a normal program. I didn't expect it to fail miserably.

### Client information

```console
> /about


╭────────────────────────────────────────╮
│                                        │
│ About Gemini CLI                       │
│                                        │
│ CLI Version  0.11.3                    │
│ Git Commit   44b3c974                  │
│ Model        auto                      │
│ Sandbox      no sandbox                │
│ OS           openbsd                   │
│ Auth Method  OAuth                     │
│                                        │
╰────────────────────────────────────────╯
```

### Login information

Google Account

1. Installed Node from OpenBSD packages.
2. Installed using `sudo npm install -g @google/gemini-cli`
3. Tried to test the package manager using doas and sudo. Both complained about missing terminals.

ghost:
## Summary

TLDR: Changed shell execution to inherit stdin, fixing interactive password prompts for sudo/doas on OpenBSD.

## Description

On OpenBSD, commands requiring interactive password input (e.g., sudo, doas)
would fail with "a terminal is required to read the password" or "a tty is
required". This was due to child processes being spawned with stdin set to
'ignore', preventing the password prompt from appearing.

This commit changes the stdin configuration for spawned child processes from
'ignore' to 'inherit'. This allows the child process to inherit the parent's
stdin, enabling interactive password prompts for privilege escalation commands
like sudo and doas. This improves compatibility and user experience on OpenBSD
and similar Unix-like systems.

Fixes #12454

## Goal

My goal is to bring support to other operating systems.


ghost:
I signed the CLA under my real email and the "Missing CLA" is still appearing. Do I have to wait

Edit: yes, I had to wait.

ghost:
I won't try to implement this anymore. Bye

PR #12566, opened 2025-11-05
fix: allow user to set pro model even in fallback

adamfweidman:
## Summary

We should allow users to set any model of their choosing. Fallback mode can get trigger from transient capacity constraints so we should allow users to reset their model choice to pro in the same session.

## Related Issues

Fixes: https://github.com/google-gemini/maintainers-gemini-cli/issues/1028

## Pre-Merge Checklist

  - [ ] MacOS
    - [ x] npm run
    - [ ] npx
    - [ ] Docker
    - [ ] Podman
    - [ ] Seatbelt
  - [ ] Windows
    - [ ] npm run
    - [ ] npx
    - [ ] Docker
  - [ ] Linux
    - [ ] npm run
    - [ ] npx
    - [ ] Docker

### DIFF
--- a/packages/core/src/config/config.ts
+++ b/packages/core/src/config/config.ts
@@ -670,10 +670,7 @@ export class Config {
   }
 
   setModel(newModel: string): void {
-    // Allow users to set any model, but if switching to a "pro" model, exit fallback mode
-    if (newModel.includes('pro')) {
-      this.setFallbackMode(false);
-    }
+    this.setFallbackMode(false);
 
     if (this.model !== newModel) {
       this.model = newModel;
--- a/packages/core/src/config/config.ts
+++ b/packages/core/src/config/config.ts
@@ -670,9 +670,9 @@ export class Config {
   }
 
   setModel(newModel: string): void {
-    // Do not allow Pro usage if the user is in fallback mode.
-    if (newModel.includes('pro') && this.isInFallbackMode()) {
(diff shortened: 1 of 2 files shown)


gsquared94 (approved):

packages/core/src/config/config.ts
```diff
@@ -670,9 +670,9 @@ export class Config {
   }
 
   setModel(newModel: string): void {
-    // Do not allow Pro usage if the user is in fallback mode.
-    if (newModel.includes('pro') && this.isInFallbackMode()) {
-      return;
+    // Allow users to set any model, but if switching to a "pro" model, exit fallback mode
+    if (newModel.includes('pro')) {
```
sehoon38: I was initially thinking if we should always set the fallback mode to false if the model is being set, e.g. there may be a new model without "pro", e.g. "ultra" that could use the fallback, or when we decide to implement fallback to the higher model (flash -> pro).
Also what if a user set it to "auto" from the setting again, should we bring the model to use the pro again? 
these are prob product questions, so may need a bit more discussion so we could do it in a separate PR if needed.

sehoon38 (approved):

packages/core/src/config/config.ts
adamfweidman: I think thats reasonable. I will change it to always set fallback to false

Issue #10677, opened 2025-10-07
Render Gemini CLI in an alternate buffer


jacob314:
Currently Gemini CLI renders in the default terminal buffer.  This is nice for built in terminal scroll bar support but makes it hard for Ink to render without rendering glitches as it is not possible to indicate that lines must not wrap as the ascii escape codes for that option are only respected in when in an alternate buffer in iterm2 and the vscode built in terminals. Custom scrolling is a prereq for this as native scrolling is not supported in an alternate buffer.

Ghostty respects that terminal escape code even in the main buffer which is why we are able to render artifact free in Ghostty. 
Rendering in an alternate buffer should be enable as a ui setting to de-risk the rollout. Some users will prefer to render in the normal buffer due to preferring native terminal scroll and due to wanting to scroll back to see output from before Gemini CLI launched. 

Polish items:
On Gemini CLI exit we should render the full conversation to the main terminal buffer so users can review.

jacob314:
Getting this ready for review.

jacob314:
This has landed.

Issue #7020, opened 2025-08-25
Modify MaxSizedBox to use overflow:scroll internally


jacob314:

jacob314:
Used alternate solution of avoiding MaxSizedBox everywhere when in alternate buffer mode.

PR #12567, opened 2025-11-05
Support ink scrolling final pr

jacob314:
## Summary
With this pr, alternateBuffer mode is now ready for general use. It should be fast and performant.
Due to using a virtualized list view, performance should now be fine even with an enormous conversation history. All rendering should now be at least as polished as the legacy non-alternate buffer rendering mode.

## Details

- Refactor `colorizeCode` to accept an options object and support disabling `MaxSizedBox` when in alternate buffer mode.
- Update `DefaultAppLayout` to adjust width and height based on alternate buffer state, ensuring space for scrollbars.
- Remove legacy message splitting logic from `useGeminiStream`.
- Adopt `StickyHeader` in `ToolMessage` for better context during scrolling. . Tool confirmations and actions now have persistent header lines, ensuring you never lose context, even during complex interactions. 
- Add `useAlternateBuffer` hook for easier access to the setting.
- Adjust `UserMessage` and `VirtualizedList` styling for consistent layout.
- Update `calculateMainAreaWidth` to account for the 1 pixel needed for the scrollbar when using alternate buffer and full width.
- **Simplified Gemini Message Handling**: Removed the previous logic for splitting large Gemini messages, indicating that the new scrolling infrastructure now efficiently handles extensive streaming content without manual chunking.
* **New Copy Mode Warning**: Added a CopyModeWarning component to inform users when copy mode is active, improving user feedback when using the ctrl-S select mode in the alternate buffer.
- Full Chat History on Exit: Even though we’re now rendering with an alternate buffer, your complete chat history is still accessible upon exiting the CLI. Now you don’t have to worry that Gemini CLI will have cleared your terminal history. No matter what you have done while using Gemini CLI, on exit your previous input history is available.
- Stable Input Prompt: Say goodbye to the bouncing input prompt! It now remains anchored at the bottom of your terminal, providing a consistent and predictable experience.

https://screencast.googleplex.com/cast/NTY1MzQ1MDU3NzY3NDI0MHwyYWVkMGNhMi04MQ

Fixes: #11053, #10677,  #7020, #7022

### DIFF
--- a/packages/cli/src/ui/components/InputPrompt.tsx
+++ b/packages/cli/src/ui/components/InputPrompt.tsx
@@ -72,11 +72,13 @@ export interface InputPromptProps {
   setShellModeActive: (value: boolean) => void;
   approvalMode: ApprovalMode;
   onEscapePromptChange?: (showPrompt: boolean) => void;
+  onSuggestionsVisibilityChange?: (visible: boolean) => void;
   vimHandleInput?: (key: Key) => boolean;
   isEmbeddedShellFocused?: boolean;
   setQueueErrorMessage: (message: string | null) => void;
   streamingState: StreamingState;
   popAllMessages?: (onPop: (messages: string | undefined) => void) => void;
+  suggestionsPosition?: 'above' | 'below';
 }
 
 // The input content, input container, and input suggestions list may have different widths
@@ -111,11 +113,13 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
   setShellModeActive,
   approvalMode,
   onEscapePromptChange,
+  onSuggestionsVisibilityChange,
   vimHandleInput,
   isEmbeddedShellFocused,
   setQueueErrorMessage,
   streamingState,
   popAllMessages,
+  suggestionsPosition = 'below',
 }) => {
   const kittyProtocol = useKittyKeyboardProtocol();
   const isShellFocused = useShellFocusState();
@@ -943,6 +947,12 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
   const activeCompletion = getActiveCompletion();
   const shouldShowSuggestions = activeCompletion.showSuggestions;
 
+  useEffect(() => {
+    if (onSuggestionsVisibilityChange) {
+      onSuggestionsVisibilityChange(shouldShowSuggestions);
+    }
+  }, [shouldShowSuggestions, onSuggestionsVisibilityChange]);
+
   const showAutoAcceptStyling =
     !shellModeActive && approvalMode === ApprovalMode.AUTO_EDIT;
   const showYoloStyling =
@@ -961,8 +971,30 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
     statusText = 'Accepting edits';
   }
 
+  const suggestionsNode = shouldShowSuggestions ? (
+    <Box paddingRight={2}>
+      <SuggestionsDisplay
+        suggestions={activeCompletion.suggestions}
+        activeIndex={activeCompletion.activeSuggestionIndex}
+        isLoading={activeCompletion.isLoadingSuggestions}
+        width={suggestionsWidth}
+        scrollOffset={activeCompletion.visibleStartIndex}
+        userInput={buffer.text}
+        mode={
+          buffer.text.startsWith('/') &&
+          !reverseSearchActive &&
+          !commandSearchActive
+            ? 'slash'
+            : 'reverse'
+        }
+        expandedIndex={expandedSuggestionIndex}
+      />
+    </Box>
+  ) : null;
+
   return (
     <>
+      {suggestionsPosition === 'above' && suggestionsNode}
       <Box
         borderStyle="round"
         borderColor={
@@ -1141,26 +1173,7 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
           )}
         </Box>
       </Box>
-      {shouldShowSuggestions && (
-        <Box paddingRight={2}>
-          <SuggestionsDisplay
-            suggestions={activeCompletion.suggestions}
-            activeIndex={activeCompletion.activeSuggestionIndex}
-            isLoading={activeCompletion.isLoadingSuggestions}
-            width={suggestionsWidth}
-            scrollOffset={activeCompletion.visibleStartIndex}
-            userInput={buffer.text}
-            mode={
-              buffer.text.startsWith('/') &&
-              !reverseSearchActive &&
-              !commandSearchActive
-                ? 'slash'
-                : 'reverse'
-            }
-            expandedIndex={expandedSuggestionIndex}
-          />
-        </Box>
-      )}
+      {suggestionsPosition === 'below' && suggestionsNode}
     </>
   );
 };
--- a/packages/cli/src/ui/components/MainContent.tsx
+++ b/packages/cli/src/ui/components/MainContent.tsx
@@ -11,19 +11,23 @@ import { OverflowProvider } from '../contexts/OverflowContext.js';
 import { useUIState } from '../contexts/UIStateContext.js';
 import { useAppContext } from '../contexts/AppContext.js';
 import { AppHeader } from './AppHeader.js';
-import { useSettings } from '../contexts/SettingsContext.js';
+import { useAlternateBuffer } from '../hooks/useAlternateBuffer.js';
+import { SCROLL_TO_ITEM_END } from './shared/VirtualizedList.js';
+import { ScrollableList } from './shared/ScrollableList.js';
+import { useMemo, memo, useCallback } from 'react';
+import { MAX_GEMINI_MESSAGE_LINES } from '../constants.js';
+
+const MemoizedHistoryItemDisplay = memo(HistoryItemDisplay);
+const MemoizedAppHeader = memo(AppHeader);
 
 // Limit Gemini messages to a very high number of lines to mitigate performance
 // issues in the worst case if we somehow get an enormous response from Gemini.
 // This threshold is arbitrary but should be high enough to never impact normal
 // usage.
-const MAX_GEMINI_MESSAGE_LINES = 65536;
-
 export const MainContent = () => {
   const { version } = useAppContext();
   const uiState = useUIState();
-  const settings = useSettings();
-  const useAlternateBuffer = settings.merged.ui?.useAlternateBuffer ?? false;
+  const isAlternateBuffer = useAlternateBuffer();
 
   const {
     pendingHistoryItems,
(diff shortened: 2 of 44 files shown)


packages/cli/src/ui/components/messages/DiffRenderer.tsx
```diff
@@ -235,115 +263,151 @@ const renderDiffContent = (
...
+        <Box key={lineKey} flexDirection="row">
+          {useMaxSizedBox ? (
             <Text
               color={semanticTheme.text.secondary}
+              backgroundColor={backgroundColor}
+            >
+              {gutterNumStr.padStart(gutterWidth)}{' '}
+            </Text>
+          ) : (
+            <Box
```
jacob314, line 353: when we aren't using MaxSizedBox we have to do this match the behavior MaxSizedBox provided. Generally this is the idiomatic Ink pattern to ensure that wrapped lines start after this gutter and that this gutter cannot shrink even if flex would otherwise want it to.

packages/cli/src/ui/components/__snapshots__/HistoryItemDisplay.test.tsx.snap
```diff
@@ -142,3 +135,233 @@ exports[`<HistoryItemDisplay /> > should render a truncated gemini_content item
...
+   44 Line 44
+   45 Line 45
+   46 Line 46
+   47 Line 47
+   48 Line 48
+   49 Line 49
+   50 Line 50"
+`;
+
+exports[`<HistoryItemDisplay /> > gemini items (alternateBuffer=true) > should render a truncated gemini_content item 1`] = `
```
jacob314, line 301: this is intentional. alternateBuffer=true should not truncate. test name is a bit ugly but will cleanup when we remove the non-alternate buffer mode so doesn't really matter

packages/cli/src/ui/components/messages/ToolMessage.tsx
```diff
@@ -108,23 +111,90 @@ export const ToolMessage: React.FC<ToolMessageProps> = ({
...
 
   // Long tool call response in MarkdownDisplay doesn't respect availableTerminalHeight properly,
-  // we're forcing it to not render as markdown when the response is too long, it will fallback
+  // so if we aren't using alternate buffer mode, we're forcing it to not render as markdown when the response is too long, it will fallback
   // to render as plain text, which is contained within the terminal using MaxSizedBox
-  if (availableHeight) {
+  if (availableHeight && !isAlternateBuffer) {
     renderOutputAsMarkdown = false;
   }
+  const childWidth = terminalWidth;
```
jacob314, line 119: best I can figure is the previous math was just wrong.

packages/cli/src/ui/components/messages/ToolMessage.tsx
```diff
@@ -108,23 +111,90 @@ export const ToolMessage: React.FC<ToolMessageProps> = ({
...
+                text={truncatedResultDisplay}
+                terminalWidth={childWidth}
+                renderMarkdown={renderMarkdown}
+                isPending={false}
+              />
+            </Box>
+          ) : typeof truncatedResultDisplay === 'string' &&
+            !renderOutputAsMarkdown ? (
+            isAlternateBuffer ? (
+              <Box flexDirection="column" width={childWidth}>
```
jacob314, line 149: here and elsewhere no need to truncate in alternate buffer mode.

packages/cli/src/ui/components/messages/ToolMessage.tsx
```diff
@@ -108,23 +111,90 @@ export const ToolMessage: React.FC<ToolMessageProps> = ({
...
+    truncatedResultDisplay,
+    renderOutputAsMarkdown,
+    childWidth,
+    renderMarkdown,
+    isAlternateBuffer,
+    availableHeight,
+    terminalWidth,
+  ]);
+
   return (
```
jacob314, line 195: we now need to use <> in order to avoid adding an extra box which would mess up having the sticky headers stick to  boy we intend them to stick to.

packages/cli/src/ui/components/messages/UserMessage.tsx
```diff
@@ -12,18 +12,25 @@ import { isSlashCommand as checkIsSlashCommand } from '../../utils/commandUtils.
...
-    <Box flexDirection="row" paddingY={0} marginY={1} alignSelf="flex-start">
-      <Box width={prefixWidth}>
+    <Box
+      flexDirection="row"
+      paddingY={0}
+      marginY={1}
+      alignSelf="flex-start"
+      width={width}
+    >
+      <Box width={prefixWidth} flexShrink={0}>
```
jacob314, line 33: It was a bug these prefixes didn't have flexShrink. It would cause the width of this component to shrink when the terminal width was small and the content after this wanted a larger width.

packages/cli/src/ui/components/InputPrompt.tsx
```diff
@@ -961,8 +971,30 @@ export const InputPrompt: React.FC<InputPromptProps> = ({
     statusText = 'Accepting edits';
   }
 
+  const suggestionsNode = shouldShowSuggestions ? (
```
jacob314, line 974: this was just moved. no code changes.

packages/cli/src/ui/components/MainContent.tsx
```diff
@@ -32,65 +36,116 @@ export const MainContent = () => {
...
+    [uiState.history],
+  );
+
+  const renderItem = useCallback(
+    ({ item }: { item: (typeof virtualizedData)[number] }) => {
+      if (item.type === 'header') {
+        return <MemoizedAppHeader key="app-header" version={version} />;
+      } else if (item.type === 'history') {
+        return (
+          <MemoizedHistoryItemDisplay
```
jacob314, line 99: I memoized a few things here to improve performance. Previously it didn't matter as these items were mainly rendered in static.

packages/cli/src/ui/constants.ts
```diff
@@ -18,6 +18,12 @@ export const SHELL_COMMAND_NAME = 'Shell Command';
 
 export const SHELL_NAME = 'Shell';
 
+// Limit Gemini messages to a very high number of lines to mitigate performance
+// issues in the worst case if we somehow get an enormous response from Gemini.
+// This threshold is arbitrary but should be high enough to never impact normal
+// usage.
+export const MAX_GEMINI_MESSAGE_LINES = 65536;
```
jacob314, line 25: this constant was just moved. No change in value

scidomino (approved):

galdawave (commented):
Nit: Not sure if i missed it but we need to make a shortcut key for scrolling to the top or bottom, i was looking at the diff in gemini-cli of this PR and had to scroll a lot was ~20 seconds which felt like forever.

We don't have them yet. I'd like to land that as a fast follow. I'd suggest using home and end although those won't help on mac laptops so should come up with some others as well. In addition currently shift up/down scrolls a line at a time. we could make it scroll a lot faster.  I'll start a chat thread on this.

galdawave:
UX Nit: we should have some padding at the top so it looks nicer 

image

galdawave:
FR (Not blocking/good followup) : we should have setting to set the responsiveness of the scroll

packages/cli/src/ui/hooks/useGeminiStream.ts
```diff
@@ -495,41 +497,52 @@ export const useGeminiStream = (
         setPendingHistoryItem({ type: 'gemini', text: '' });
         newGeminiMessageBuffer = eventValue;
       }
-      // Split large messages for better rendering performance. Ideally,
```
galdawave, line 498: Since ScrollableList virtualizes items but not the content within an item, CodeColorizer will attempt to render thousands of Box/Text nodes for that single message at once. This could lead to performance degradation or OOM errors for large model outputs. Consider keeping some form of message splitting or implementing internal virtualization for MarkdownDisplay/CodeColorizer.

galdawave:
Docs Nit: might make sense to add something in the docs to mention copy mode

packages/cli/src/ui/hooks/useGeminiStream.ts
```diff
@@ -495,41 +497,52 @@ export const useGeminiStream = (
...
-            type: pendingHistoryItemRef.current?.type as
-              | 'gemini'
-              | 'gemini_content',
-            text: beforeText,
-          },
-          userMessageTimestamp,
-        );
-        setPendingHistoryItem({ type: 'gemini_content', text: afterText });
-        newGeminiMessageBuffer = afterText;
+        // Split large messages for to avoid flicker when not rendering in
```
galdawave: Nit: typo in the comment

packages/cli/src/ui/layouts/DefaultAppLayout.tsx
```diff
@@ -13,30 +13,40 @@ import { Composer } from '../components/Composer.js';
...
   const uiState = useUIState();
+  const isAlternateBuffer = useAlternateBuffer();
+
   const { rootUiRef, terminalHeight } = uiState;
-  const settings = useSettings();
   useFlickerDetector(rootUiRef, terminalHeight);
-
+  // If in alternate buffer mode, need to leave room to draw the scrollbar on
+  // the right side of the terminal.
+  const width = isAlternateBuffer
```
galdawave, line 27: This effectively overrides the -1 adjustment made in `ui-sizing.ts` for the container. Since VirtualizedList adds paddingRight={1}, the content width ends up being correct (terminalWidth - 1). This seems consistent, but it makes the change in `ui-sizing.ts` strictly relevant only for child components (like UserMessage) rather than the layout container itself. 

Is this the intended behavior ?

packages/cli/src/ui/components/MainContent.tsx
```diff
@@ -11,19 +11,23 @@ import { OverflowProvider } from '../contexts/OverflowContext.js';
...
+const MemoizedHistoryItemDisplay = memo(HistoryItemDisplay);
+const MemoizedAppHeader = memo(AppHeader);
 
 // Limit Gemini messages to a very high number of lines to mitigate performance
 // issues in the worst case if we somehow get an enormous response from Gemini.
 // This threshold is arbitrary but should be high enough to never impact normal
 // usage.
-const MAX_GEMINI_MESSAGE_LINES = 65536;
-
 export const MainContent = () => {
```
galdawave, line 27: Nit: might be worth refactoring maincontent.tsx, it's becoming complex because it handles two different rendering strategies: the legacy Static rendering (for standard buffer) and the new ScrollableList virtualization (for alternate buffer). We could split MainContent into smaller, focused components. This isolates the logic and makes the main entry point cleaner.

this could also be a good follow up, so we don't increase the size of this PR

packages/cli/src/ui/layouts/DefaultAppLayout.tsx
jacob314: this is consistent. Agree it is confusing. However after we remove the legacy rendering option all these widths should be able to go away. they are present right now because we can't rely on the flex layout itself because MaxSizedBox needs the exact width immediately.

packages/cli/src/ui/components/MainContent.tsx
jacob314: the only reason I wouldn't refactor is that the next step should be to remove the legacy. Only kind of tech debt is ok is debt you are able to declare bankruptcy on.

packages/cli/src/ui/hooks/useGeminiStream.ts
jacob314: Fair point. Reverted changes to this file. I'd made this change because I wanted to be able to easily scroll a single content chunk at a time but I can do that without this change easily enough just by looking for the last top item. As you mention, the performance optimizations aren't crucial with my changes but could be still useful to optimize the list view performance for extreme cases. We'll just need to make sure cases like copy and paste handle adjacent blocks well which we'd probably need anyway.

packages/cli/src/ui/hooks/useGeminiStream.ts
jacob314: reverted all changes to this file

packages/cli/src/ui/layouts/DefaultAppLayout.tsx
jacob314: this is intended. agree it is a bit ugly but would rather not change as next step to reduce this tech debt is to remove the non-alternate buffer mode and all these width calculations as they are not required once  we remove MaxSizedBox... at least that is the dream.

jacob314:
> UX Nit: we should have some padding at the top so it looks nicer
> 
> image

Lets land that as a fast follow. In your screenshot you are just about to get an extra line of padding at the bottom due to the sticky header if you scroll 1 line further. UX is about to come up with an enhanced design for the sticky headers versus this eng prototype.

SkrOYC:
I guess you don't need #9794 anymore now that it's kinda full width now

PR #12617, opened 2025-11-05
feat(policy): Add comprehensive priority range validation tests

allenhutchison:
This commit introduces new test cases to  to ensure the policy priority validation correctly handles out-of-bounds values. Previously, there was a risk of regressions allowing invalid priority values (outside the 0-999 inclusive range) to bypass validation, potentially leading to security vulnerabilities by enabling users to bypass admin policies.

The new tests specifically cover:
- Negative priority values.
- Priority values equal to or greater than 1000.

These additions enhance the robustness of the policy engine by verifying that the schema validation for policy priority is working as expected and prevents future regressions.

Closes #12612

### DIFF
--- a/packages/core/src/policy/toml-loader.test.ts
+++ b/packages/core/src/policy/toml-loader.test.ts
@@ -8,6 +8,57 @@ import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
 import { ApprovalMode, PolicyDecision } from './types.js';
 import type { Dirent } from 'node:fs';
 import nodePath from 'node:path';
+import type { PolicyLoadResult } from './toml-loader.js';
+
+async function runLoadPoliciesFromToml(
+  tomlContent: string,
+  fileName = 'test.toml',
+): Promise<PolicyLoadResult> {
+  const actualFs =
+    await vi.importActual<typeof import('node:fs/promises')>(
+      'node:fs/promises',
+    );
+
+  const mockReaddir = vi.fn(
+    async (
+      path: string,
+      _options?: { withFileTypes: boolean },
+    ): Promise<Dirent[]> => {
+      if (nodePath.normalize(path) === nodePath.normalize('/policies')) {
+        return [
+          {
+            name: fileName,
+            isFile: () => true,
+            isDirectory: () => false,
+          } as Dirent,
+        ];
+      }
+      return [];
+    },
+  );
+
+  const mockReadFile = vi.fn(async (path: string): Promise<string> => {
+    if (
+      nodePath.normalize(path) ===
+      nodePath.normalize(nodePath.join('/policies', fileName))
+    ) {
+      return tomlContent;
+    }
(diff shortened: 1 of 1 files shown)


packages/core/src/policy/toml-loader.test.ts
```diff
@@ -722,70 +256,177 @@ priority = 100
...
-      expect(
-        result.rules[0].argsPattern?.test('{"command":"git log *.txt"}'),
-      ).toBe(true);
-      expect(result.errors).toHaveLength(0);
+      expect(result.errors).toHaveLength(1);
+      const error = result.errors[0];
+      expect(error.errorType).toBe('file_read');
+      expect(error.message).toContain('Failed to read policy directory');
     });
   });
```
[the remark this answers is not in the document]
allenhutchison: Good suggestion. Added.

abhipatel12 (approved):
LGTM

packages/core/src/policy/toml-loader.test.ts
```diff
@@ -715,77 +346,165 @@ priority = 100
...
-          nodePath.normalize(path) ===
-          nodePath.normalize(nodePath.join('/policies', 'shell.toml'))
-        ) {
-          return `
+    it('should return a schema_validation error if priority is >= 1000', async () => {
+      const result = await runLoadPoliciesFromToml(`
+[[rule]]
+toolName = "test"
+decision = "allow"
+priority = 1000
```
bkdavidson, line 402: nit: I can foresee a fence post error where the valid condition accidentally becomes <=1000 when it should be <1000, so we should probably also do a much higher value as well.

packages/core/src/policy/toml-loader.test.ts
```diff
@@ -715,77 +346,165 @@ priority = 100
...
-          }
-          return [];
-        },
-      );
+    it('should return a schema_validation error if priority is negative', async () => {
+      const result = await runLoadPoliciesFromToml(`
+[[rule]]
+toolName = "test"
+decision = "allow"
+priority = -1
```
bkdavidson, line 388: nit: I can foresee a fence post error where the valid condition accidentally becomes >=-1 when it should be >=0, so we should probably also do a much lower value as well. It is more unlikely than the 1k rule, but its an easy test to add.

bkdavidson (commented):
LGTM - 2 nits you can resolve on a follow-up.

packages/core/src/policy/toml-loader.test.ts
allenhutchison: Addressed in https://github.com/google-gemini/gemini-cli/pull/12679

packages/core/src/policy/toml-loader.test.ts
allenhutchison: Addressed in https://github.com/google-gemini/gemini-cli/pull/12679

PR #12640, opened 2025-11-06
Let users know when falling back to flash, and update the error messa…

sehoon38:
## Summary
Let users know when falling back to flash, and update the error messages and the dialog options.

## Details
Many users reported about automatic switching of the models. We should always prompt the users when the model is falling back.
Also updating the error messages and the dialog options according to https://docs.google.com/document/d/1GT1nyfx-R332VtX0iClPgLrN2NnnOxuBrslmsFK8Cbs/edit?tab=t.0. These messages are subject to change.

We are planning to refactor these model fallback after the Orion launch - we decided not to do the refactoring at this time due to the time constraint.

Terminal (Quota) error - free-tier user (validated both in auto and pro model)
image

Terminal (Quota) error - paid-tier user (validated both in auto and pro model)
image

Retryable (most likely Capacity) error (we are going to assume most of the non-terminal errors as capacity errors for now)
image

## Related Issues
Fixes https://github.com/google-gemini/maintainers-gemini-cli/issues/993

## Pre-Merge Checklist

  - [x] MacOS
    - [x] npm run
    - [ ] npx
    - [ ] Docker
    - [ ] Podman
    - [ ] Seatbelt
  - [ ] Windows
    - [ ] npm run
    - [ ] npx
    - [ ] Docker
  - [ ] Linux
    - [ ] npm run
    - [ ] npx
    - [ ] Docker

### DIFF
--- a/packages/cli/src/ui/contexts/UIActionsContext.tsx
+++ b/packages/cli/src/ui/contexts/UIActionsContext.tsx
@@ -40,7 +40,7 @@ export interface UIActions {
   refreshStatic: () => void;
   handleFinalSubmit: (value: string) => void;
   handleClearScreen: () => void;
-  handleProQuotaChoice: (choice: 'retry_later' | 'continue' | 'auth') => void;
+  handleProQuotaChoice: (choice: 'retry_later' | 'retry') => void;
   setQueueErrorMessage: (message: string | null) => void;
   popAllMessages: (onPop: (messages: string | undefined) => void) => void;
   handleApiKeySubmit: (apiKey: string) => Promise<void>;
--- a/packages/cli/src/ui/contexts/UIActionsContext.tsx
+++ b/packages/cli/src/ui/contexts/UIActionsContext.tsx
@@ -40,7 +40,7 @@ export interface UIActions {
   refreshStatic: () => void;
   handleFinalSubmit: (value: string) => void;
   handleClearScreen: () => void;
-  handleProQuotaChoice: (choice: 'auth' | 'continue') => void;
+  handleProQuotaChoice: (choice: 'retry_later' | 'continue' | 'auth') => void;
   setQueueErrorMessage: (message: string | null) => void;
   popAllMessages: (onPop: (messages: string | undefined) => void) => void;
   handleApiKeySubmit: (apiKey: string) => Promise<void>;
--- a/packages/cli/src/ui/hooks/useQuotaAndFallback.ts
+++ b/packages/cli/src/ui/hooks/useQuotaAndFallback.ts
@@ -11,7 +11,6 @@ import {
   type FallbackIntent,
   TerminalQuotaError,
   UserTierId,
-  RetryableQuotaError,
 } from '@google/gemini-cli-core';
 import { useCallback, useEffect, useRef, useState } from 'react';
 import { type UseHistoryManagerReturn } from './useHistoryManager.js';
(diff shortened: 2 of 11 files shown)


packages/cli/src/ui/contexts/UIActionsContext.tsx
```diff
@@ -40,7 +40,7 @@ export interface UIActions {
   refreshStatic: () => void;
   handleFinalSubmit: (value: string) => void;
   handleClearScreen: () => void;
-  handleProQuotaChoice: (choice: 'auth' | 'continue') => void;
+  handleProQuotaChoice: (choice: 'retry_later' | 'continue' | 'auth') => void;
```
adamfweidman: I think we can get rid of auth now.

packages/cli/src/ui/hooks/useQuotaAndFallback.ts
```diff
@@ -132,44 +100,39 @@ export function useQuotaAndFallback({
...
-      return 'stop';
+      return intent;
     };
 
     config.setFallbackModelHandler(fallbackHandler);
   }, [config, historyManager, userTier, setModelSwitchedFromQuotaError]);
 
   const handleProQuotaChoice = useCallback(
-    (choice: 'auth' | 'continue') => {
+    (choice: 'retry_later' | 'continue' | 'auth') => {
```
adamfweidman: remove 'auth'.
nit: Can we have choice return a partial of FallbackIntent. We can then use this instead of intent belowl

packages/cli/src/ui/hooks/useQuotaAndFallback.ts
```diff
@@ -132,44 +100,39 @@ export function useQuotaAndFallback({
...
-      const intent: FallbackIntent = choice === 'auth' ? 'auth' : 'retry';
+      const intent: FallbackIntent = choice === 'continue' ? 'retry' : choice;
       proQuotaRequest.resolve(intent);
       setProQuotaRequest(null);
       isDialogPending.current = false; // Reset the flag here
 
       if (choice === 'auth') {
         setAuthState(AuthState.Updating);
-      } else {
+      } else if (choice === 'continue') {
```
adamfweidman: I think we always want to add that info banner now that auth is no longer an option, and we can also remove all things setAuthState related

packages/core/src/fallback/handler.ts
```diff
@@ -49,6 +49,9 @@ export async function handleFallback(
       case 'auth':
```
adamfweidman: nit: rm auth

packages/core/src/fallback/types.ts
```diff
@@ -10,7 +10,8 @@
 export type FallbackIntent =
   | 'retry' // Immediately retry the current request with the fallback model.
   | 'stop' // Switch to fallback for future requests, but stop the current request.
-  | 'auth'; // Stop the current request; user intends to change authentication.
+  | 'auth' // Stop the current request; user intends to change authentication.
```
adamfweidman: ditto

packages/cli/src/ui/contexts/UIActionsContext.tsx
sehoon38: I initially removed it, and intentionally added back as Srinath also considered having auth as an option, so left it for now. I can clean it up once the design and the options are finalized and confirm we'll no longer use it

packages/cli/src/ui/hooks/useQuotaAndFallback.ts
sehoon38: retry_later which is for users to retry with the same model won't switch the model, and it would just stop the current conversation and shows the error.

packages/cli/src/ui/contexts/UIActionsContext.tsx
sehoon38: actually let me remove it for now, I can go back and add it again if we are actually going to use it. as I might end up forgetting

Adib234 (approved):
LGTM!

PR #12646, opened 2025-11-06
fix: integrate DiscoveredTool with Policy Engine

allenhutchison:
Fixes a security vulnerability where tools discovered via toolDiscoveryCommand bypassed the Policy Engine. This change ensures that DiscoveredTool and DiscoveredToolInvocation correctly utilize the MessageBus to trigger policy checks before execution.

### DIFF
--- a/packages/core/src/policy/policies/discovered.toml
+++ b/packages/core/src/policy/policies/discovered.toml
@@ -1,0 +1,8 @@
+# Default policy for tools discovered via toolDiscoveryCommand.
+# These tools are potentially dangerous as they are arbitrary scripts.
+# We default them to ASK_USER for safety.
+
+[[rule]]
+toolName = "discovered_tool_*"
+decision = "ask_user"
+priority = 10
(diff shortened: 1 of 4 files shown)


packages/core/src/policy/policies/discovered.toml
```diff
@@ -0,0 +1,8 @@
+# Default policy for tools discovered via toolDiscoveryCommand.
+# These tools are potentially dangerous as they are arbitrary scripts.
+# We default them to ASK_USER for safety.
+
+[[rule]]
+toolName = "discovered_tool_*"
+decision = "ask_user"
```
bkdavidson, line 7: We should have a test that validates the discovered.toml default policy works as intended. This would prevent a regression in the scenario that the priority changes or the decision is altered.

bkdavidson (changes requested):
1 easy fix but an important one, otherwise LGTM. You should also have a maintainer review for their perspective otherwise.

packages/core/src/policy/policies/discovered.toml
allenhutchison: Done.

abhipatel12 (approved):
LGTM!

bkdavidson (approved):
LGTM

